PR #57, opened 2021-11-16
Template and automate the process of creating a new AI service

### LINKED ISSUE #44: Re-architecture the AI service design and set up process
* shouldn't need to connect the private gitea repo
* do we need to host the docker registry image?
* can we simplify the process of adding a new AI model? automate more of it? less copy & paste and search and replace.

ypkang:
The `create_ai_service.sh` will automatically create the scaffolding for a new AI service. It saves a lot of manual copy pasting.

Closes #44

### DIFF
--- a/ai_serving/template/Dockerfile
+++ b/ai_serving/template/Dockerfile
@@ -1,0 +1,23 @@
+FROM ubuntu:20.04
+WORKDIR /
+ENV DEBIAN_FRONTEND=noninteractive
+
+# OS dependencies
+RUN apt update; apt -y upgrade
+RUN apt -y install --no-install-recommends python3.8 python3-pip vim; apt-get clean; rm -rf /var/lib/apt/lists/*
+
+# Python dependenicies
+COPY requirements.txt requirements.txt
+RUN pip3 install -r requirements.txt
+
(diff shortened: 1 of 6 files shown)


ypkang:
@marsninja let me know what you think of this. It is not much real "templating" but more so just to save manual copy/past/search and replace required for adding a new AI model.

If you have other ideas on how we can template this better, I am definitely open to it!

ypkang:
This is ready for review.

marsninja (changes requested):
Hi Yipig, can you write a readme for the ai_serving directory that gives an overview of the approach?

ypkang:
> Hi Yipig, can you write a readme for the ai_serving directory that gives an overview of the approach?

@marsninja Added a simple README under `ai_serving/`

ai_serving/template/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM ubuntu:20.04
```
marsninja, line 1: Can we try switching this to python:3-slim, thats what jaseci is based on these days and would save image size for public dockerhub. Give it a try , if too tricky we can keep ubuntu.

marsninja (changes requested):

PR #69, opened 2021-12-06
Add uswgi for use server

ypkang:


marsninja (approved):
cool, I think we should really brianstorm the next generation of auto generating pods

marsninja (approved):
LGTM

PR #72, opened 2021-12-12
Added try-else block semantics and tests, all passing

marsninja:


marsninja:
@asimsalim This is for you ;-), try-else blocks are in!! with error catching, check out the test case.

PR #84, opened 2021-12-22
refactor get_global_actions

panikingginoo12:


marsninja:
Twas fun pair coding this. We should do more @panikingginoo12 !!

PR #90, opened 2022-01-19
Django Migrations for 1.0 -> 1.3

ypkang:


ypkang:
for action, `name` should stay the same, kind doesn't matter.

ypkang:
& gets the UUID from the node object. * gets the object from the UUID.

as admin, create a new sentinel, set it to global
for each user, set an active graph, call sentinel_active_global
One potential problem is some users for some reason have multiple graphs, we need to resolve that.

PR #99, opened 2022-02-01
encoders initial commit

AshishMahendra:
Created the module for Bi and Poly encoders


marsninja (commented):
LGTM as a first commit

PR #100, opened 2022-02-01
update for config api and requirements added

AshishMahendra:


AshishMahendra:
Old version of Jaseci used

PR #103, opened 2022-02-03
Add issue and PR templates for github

ypkang:


ypkang:
Configured this via github UI. This is not needed anymore.

PR #107, opened 2022-02-15
Refactor and pod-ify biencoder

ypkang:
Refactor of the biencoder implementation from @AshishMahendra 

- Naming convention (e.g. camel case vs snake-case)
- General code clean up
- Liniting
- Put unit tests under the same testcase framework as rest of jaseci
- Add a yaml file to put this in a pod.


ypkang:
ToDo:

- [x ] Get all unit tests to pass and in the same unittest framework as reset of jaseci
- [ ] consolidate the different config_setup in bi, train and evaluate
- [ ] add model upload/download API
- [x] ~~update docker image~~ and k8s manifest
- [x] test with load remote actions
- [] unwrap the functions without jsonresponse
- 
Question to clarify on:
* the value given in the current test cases don't pass
*

PR #108, opened 2022-02-16
Args qol updates

marsninja:
@ypkang Can you review this. Im thinking you can start reviewing my PRs as a learning tool. We can have calls to talk through them, for fun, learning, and some feedback for me :-).

### DIFF
--- a/jaseci_core/jaseci/api/alias_api.py
+++ b/jaseci_core/jaseci/api/alias_api.py
@@ -23,7 +23,7 @@ class alias_api():
     def __init__(self):
         self.alias_map = {}
 
-    @interface.private_api()
+    @interface.private_api(cli_args=['name'])
     def alias_register(self, name: str, value: str):
         """Create string to string alias mapping that caller can use.
 
@@ -60,7 +60,7 @@ class alias_api():
         """
         return self.alias_map
 
-    @interface.private_api()
+    @interface.private_api(cli_args=['name'])
     def alias_delete(self, name: str):
         """Delete an active string to string alias mapping.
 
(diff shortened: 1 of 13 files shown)


ypkang:
> @ypkang Can you review this. Im thinking you can start reviewing my PRs as a learning tool. We can have calls to talk through them, for fun, learning, and some feedback for me :-).

Love it! Looking through this one right now.

jaseci_core/jaseci/api/alias_api.py
```diff
@@ -23,7 +23,7 @@ class alias_api():
     def __init__(self):
         self.alias_map = {}
 
-    @interface.private_api()
+    @interface.private_api(cli_args=['name'])
```
ypkang, line 26: So this means that `alias register` will have the default arguments `name`, correct? 

```
alias register NAME
```

jaseci_core/jaseci/api/architype_api.py
```diff
@@ -20,7 +20,7 @@ class architype_api():
     a Jaseci instance is aware of.
     """
 
-    @interface.private_api()
+    @interface.private_api(cli_args=['code'])
```
ypkang, line 23: Can I have multiple `cli_args`? I'd assume no because then their position would matter, right?

ypkang (approved):

ypkang:
This looks good! I just have a couple of clarifying questions to help me understand better how this works. Left them as comments inline.

jaseci_core/jaseci/api/alias_api.py
marsninja: Correct! No need for `alias register -name NAME -value VALUE` you can just do `alias register NAME -value VALUE`

jaseci_core/jaseci/api/architype_api.py
marsninja: You can have multiple, just would need to keep in mind whatever is in the list will have to be the first args to the given command in that order then you'd use the `-param {value}` in any order for other params.

PR #114, opened 2022-03-01
productionization changes for entity extraction

AshishMahendra:
This PR makes the necessary changes to make the Flair-based entity extraction AI service production-ready, including:

- Use the `jaseci/ai-ready-img` docker image, same as the rest of AI services
- Put it in a pod
- add model loading and saving API
- other coding standard/formatting changes.


ypkang:
I added the yaml file and updated the ai-ready docker image. Also made some minor tweaks and fixed a few typos. Two things need attention @AshishMahendra 
1. The service download the pretrained taggers to a random temp directory everytime. So as a result, whenever I stand-up the pod, I will have to wait for a new download
```
2022-03-02 19:43:41,131 https://nlp.informatik.hu-berlin.de/resources/models/tars-ner/tars-ner.pt not found in cache, downloading to /tmp/tmpgw6knaxc
```
The `/tmp/` directory is different everytime. @AshishMahendra can we make this a fixed path so that the system can detect it was downloaded before?
2. The pod requires more than 6GB memory. Is this how big roberta is? maybe we should switch to bert or a smaller version of roberta -- 6G is going to create some deployment challenges.

ypkang (approved):
🚢 ing this.

ypkang:
> I added the yaml file and updated the ai-ready docker image. Also made some minor tweaks and fixed a few typos. Two things need attention @AshishMahendra
> 
> 1. The service download the pretrained taggers to a random temp directory everytime. So as a result, whenever I stand-up the pod, I will have to wait for a new download
> 
> ```
> 2022-03-02 19:43:41,131 https://nlp.informatik.hu-berlin.de/resources/models/tars-ner/tars-ner.pt not found in cache, downloading to /tmp/tmpgw6knaxc
> ```
> 
> The `/tmp/` directory is different everytime. @AshishMahendra can we make this a fixed path so that the system can detect it was downloaded before? 2. The pod requires more than 6GB memory. Is this how big roberta is? maybe we should switch to bert or a smaller version of roberta -- 6G is going to create some deployment challenges.

Closing the loop on these points. The model download into temp directory is consistent with how the other AI models work as well so that's oaky. The model size is still a little concerning but we will go with the Roberta model for now because it was fine-tuned on entity data. Switching to a normal bert might see some significant accuracy degredation.

PR #120, opened 2022-03-07
Fix: Changed some misleading info in the Docs

Gorgeous-Patrick:
1. The port info updated. Make sure that it is always port 8888 throughout the doc.
2. Added a few words for Linux installation for kubernetes (just in case that someone not using Ubuntu is confused).


marsninja (approved):
LGTM!!

PR #122, opened 2022-03-08
Fix: namespace problem in monitoring document

Gorgeous-Patrick:
I think that at this point, the document has completely given up teaching people to establish prometheus in a namespace. 
Fix:
1. Deleted '-n prometheus' param in the commands given completely.
2. \<xxx\> is considered a macro substitution in markdown and we did not define \<namespace\> macro. So it is ignored. I replaced it with \\\<namespace\\\> and added some comments about the usage.

### DIFF
--- a/support/monitoring/grafana/grafana.yaml
+++ b/support/monitoring/grafana/grafana.yaml
@@ -5,6 +5,5 @@ datasources:
     - name: Prometheus
       type: prometheus
       url: http://prometheus-server.default.svc.cluster.local
-      # url: http://localhost:8080
       access: proxy
       isDefault: true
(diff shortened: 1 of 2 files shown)


support/monitoring/grafana/grafana.yaml
```diff
@@ -5,5 +5,6 @@ datasources:
     - name: Prometheus
       type: prometheus
       url: http://prometheus-server.default.svc.cluster.local
+      # url: http://localhost:8080
```
ypkang: Remove this commented line.

ypkang (changes requested):
@Gorgeous-Patrick good changes! Just one tiny thing to fix. I left comment on the line.

support/monitoring/grafana/grafana.yaml
Gorgeous-Patrick: Resolved

ypkang (approved):
LGTM

PR #125, opened 2022-03-14
Actions merge

marsninja:
Now uses a single decorators for both local actions and remote actions, everything should work as normal.

### DIFF
--- a/jaseci_core/jaseci/actions/live_actions.py
+++ b/jaseci_core/jaseci/actions/live_actions.py
@@ -3,22 +3,30 @@ General action base class with automation for hot loading
 """
 from importlib.util import spec_from_file_location, module_from_spec
 from jaseci.utils.utils import logger
-from jaseci.actions.remote_actions import ACTIONS_SPEC_LOC
+from jaseci.actions.remote_actions import ACTIONS_SPEC_LOC, serv_actions
 import requests
+import inspect
 
 live_actions = {}
+possible_remote = []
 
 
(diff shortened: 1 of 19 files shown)


jaseci_core/jaseci/actions/live_actions.py
```diff
@@ -4,21 +4,29 @@
...
 live_actions = {}
 
 
-def jaseci_action(act_group=None):
+def jaseci_action(act_group=None, aliases=list(), allow_remote=False):
     """Decorator for Jaseci Action interface"""
+    caller_globals = dict(inspect.getmembers(
+        inspect.stack()[1][0]))["f_globals"]
+    if allow_remote and 'serv_actions' not in caller_globals:
+        caller_globals['serv_actions'] = serv_actions
```
ypkang, line 19: I am curious what does this `caller_globals` do and why is this change conditioned if the action is allowed to be remote or not?

ypkang (changes requested):
There is one issue with this change -- the unit tests for the jskit service rely on the `app` variable to function and it seems like in this version, the `app` variable is no longer available in each server's own python file.

jaseci_core/jaseci/actions/live_actions.py
marsninja: Yea, you'd have to use that function serve_actions.

ypkang (approved):
LGTM.

PR #126, opened 2022-03-14
initial commit for the tutorials

AshishMahendra:


jskit/encoders/tests/test_biencoder.py
```diff
@@ -26,13 +26,13 @@ def tearDown(self) -> None:
...
         response = self.client.post(
-            "/cos_sim_score/",
+            "/cosine_sim/",
             json=test_cos_sim_request
         )
         self.assertEqual(response.status_code, 200)
         self.assertEqual(
             response.json(),
-            0.49492838978767395
+            0.04607951029096483
```
marsninja: I would round this value to 2 digits after the decimal then do the equality check just in case there are slight rounding differences.

marsninja (approved):
Fine for now!

jskit/encoders/tests/test_biencoder.py
marsninja: @AshishMahendra Update this when you get a chance and I can merge.

marsninja (approved):
LGTM

PR #131, opened 2022-03-15
Fix: walker run command  Delete: a duplicated part in Adding an AI Component

Gorgeous-Patrick:
1. It seems that the `walker run -name` command has deprecated so I fixed it.
2. Deleted a duplicated part ( a mistake ).


marsninja (approved):
LGTM

PR #133, opened 2022-03-15
Changes for test data and API

AshishMahendra:


jskit/encoders/tests/test_biencoder.py
```diff
@@ -26,13 +26,13 @@ def tearDown(self) -> None:
...
         response = self.client.post(
-            "/cos_sim_score/",
+            "/cosine_sim/",
             json=test_cos_sim_request
         )
         self.assertEqual(response.status_code, 200)
         self.assertEqual(
             response.json(),
-            0.49492838978767395
+            0.04607951029096483
```
marsninja: Same here, I'd round to 2 digits after decimal and do the equality check on the rounded values just in case there are rounding issues at the least significant digits in these values.

Issue #145, opened 2022-03-19
Add the option of CosineSimilarity loss for bi-encoder training


StefanHeng:
CosineSimilarity is a commonly used loss in training. I'd like to have that option for training the jskit::BiEncoder, for potentially comparing the results with those described in papers. 

The [existing implementation](https://github.com/Jaseci-Labs/jaseci/blob/main/jskit/encoders/utils/models.py#L227-L230) maximizes the dot product. We just need to add an additional normalization. 

For now, I'll align to the existing implementation by supporting only positive label cases. 

Another note: as linked above, correct me if I'm wrong - looks like we're computing the dot product between *every possible* context-candidate pair and then filtering them with `eye`. This is effectively the same as just computing the dot product between each *corresponding* pair, which is more efficient.

ypkang:
Thanks for contributing @StefanHeng. Looking forward to the PR!

PR #147, opened 2022-03-20
Update loss function for jskit::BiEncoder 

StefanHeng:
Update: efficient dot product loss; New: cosine similarity loss

See [issue 145](https://github.com/Jaseci-Labs/jaseci/issues/145).

### DIFF
--- a/jskit/encoders/utils/models.py
+++ b/jskit/encoders/utils/models.py
@@ -176,6 +176,8 @@ class BiEncoder(BertPreTrainedModel):
         else:
             self.cont_bert = kwargs['cont_bert']
             self.cand_bert = kwargs['cand_bert']
+        self.loss_type = kwargs.get('loss_type', 'dot-prod')
+        assert self.loss_type in ['dot-prod', 'cos-sim']
 
     def forward(self, context_input_ids=None, context_input_masks=None,
                 candidate_input_ids=None, candidate_input_masks=None,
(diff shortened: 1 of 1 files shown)


jskit/encoders/utils/models.py
```diff
@@ -224,11 +226,14 @@ def forward(self, context_input_ids=None, context_input_masks=None,
...
-            mask = torch.eye(context_input_ids.size(0)).to(
-                context_input_ids.device)
-            loss = F.log_softmax(dot_product, dim=-1) * mask
+            if self.loss_type == 'dot-prod':
+                dot_product = torch.sum(context_vec * candidate_vec, dim=-1)  # row-wise dot product
+                loss = F.log_softmax(dot_product, dim=-1)
+            else:  # `cos-sim`
+                loss_fn = torch.nn.CosineEmbeddingLoss()
+                batch_size = candidate_vec.size(0)
+                # TODO: support negative pairs
```
ypkang, line 235: So this doesn't work with negative examples?

ypkang:
The changes look good. @AshishMahendra could you take a look here too and give it a test?

AshishMahendra:
@StefanHeng did you test for the loss and accuracy, after implementing ?

AshishMahendra (commented):
Seems like loss value is not reducing and accuracy score have gone down

jskit/encoders/utils/models.py
StefanHeng: https://github.com/Jaseci-Labs/jaseci/pull/147/commits/cf7c5afeb8828d9a20cbdbc56ab769e6a78d88cf#diff-eef0e0493716527b9006d448f729d8c4ff5f4a52988f05999bf95907d0e0e134R236

Here we assign only positive labels as the target. With respect to loss we can have it work with negative examples by just matching the labels from the batch. However, supporting negative examples requires more changes in the entire forward pass. 

So no it doesn't work with negative examples now.

jskit/encoders/utils/models.py
```diff
@@ -224,11 +226,14 @@ def forward(self, context_input_ids=None, context_input_masks=None,
                     return candidate_vec
         if labels is not None and mode == "train":
             candidate_vec = candidate_vec.squeeze(1)
-            dot_product = torch.matmul(
-                context_vec, candidate_vec.t())  # [bs, bs]
-            mask = torch.eye(context_input_ids.size(0)).to(
-                context_input_ids.device)
-            loss = F.log_softmax(dot_product, dim=-1) * mask
+            if self.loss_type == 'dot-prod':
```
ypkang, line 229: @StefanHeng any reason why you also modified the dot product code as well?

StefanHeng: IRRC as we discussed last time and I mentioned in the issue 145, we’re effectively computing the dot product for each corresponding pair of context and candidate vectors. I made the change cos I believe it’s more efficient

StefanHeng:
I could take a look when I'm free (probably after I graduate). I'll need to run experiments to see why the drop in performance. 

Per our meeting with @marsninja yesterday, I want to focus on the paper (and schoolwork) for now. We'll get into trying loss functions.

ypkang:
> I could take a look when I'm free (probably after I graduate). I'll need to run experiments to see why the drop in performance.
> 
> Per our meeting with @marsninja yesterday, I want to focus on the paper (and schoolwork) for now. We'll get into trying loss functions.

Sounds good. @AshishMahendra is going to take this over and incorporate the changes.

ypkang:
This is now superseded by #205

PR #148, opened 2022-03-21
Updated mode for  bi-encoder

AshishMahendra:
Please the details of the evaluation
image

### DIFF
--- a/jskit/encoders/utils/config.cfg
+++ b/jskit/encoders/utils/config.cfg
@@ -1,14 +1,14 @@
 [TRAIN_PARAMETERS]
 max_contexts_length = 128
-MAX_CANDIDATE_LENGTH = 64
-train_batch_size = 32
-eval_batch_size = 2
-MAX_HISTORY = 4
+max_candidate_length = 64
+train_batch_size = 8
+eval_batch_size = 1
+max_history = 4
 learning_rate = 0.00005
 weight_decay = 0.1
 warmup_steps = 2000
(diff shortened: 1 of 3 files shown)


jskit/encoders/utils/config.cfg
```diff
@@ -18,8 +18,8 @@ gpu = 0
 basepath = log_output
 
 [MODEL_PARAMETERS]
-architecture = bi
+architecture = bi (obselete)
```
ypkang: Why not just remove it if it's not being used anymore?

jskit/encoders/utils/config.cfg
```diff
@@ -1,14 +1,14 @@
 [TRAIN_PARAMETERS]
 max_contexts_length = 128
-MAX_CANDIDATE_LENGTH = 64
-train_batch_size = 32
-eval_batch_size = 2
-MAX_HISTORY = 4
+max_candidate_length = 64
+train_batch_size = 8
+eval_batch_size = 1
+max_history = 4
```
ypkang, line 6: are you sure this should be lower case? if you look at `train.py` its all reading using upper case? can you confirm and make sure train/infer work?

ypkang (changes requested):

jskit/encoders/utils/config.cfg
AshishMahendra: Yes, it does work

ypkang (approved):
LGTM

PR #153, opened 2022-03-21
Feature/readme update2

codedbychavez:
# Change log
1. Improved README.md based on feedback from Jason.
2. Fixed quick start error in docs.


marsninja (changes requested):
@codedbychavez Do we need the package-lock.json and yarn.lock? I think those are supposed to be gitignored. Check real quick and let me know

marsninja (approved):
LGTM, Thanks @codedbychavez for a great start to the Readme!

PR #157, opened 2022-03-23
Make edge attributes work in DOT

ypkang:
DOT code like the following will successfully save the edge attribute in their context now
```
graph_root -> node_1 [edge=special, name=edge_1]
```

### DIFF
--- a/jaseci_core/jaseci/graph/edge.py
+++ b/jaseci_core/jaseci/graph/edge.py
@@ -123,6 +123,18 @@ class edge(element, anchored):
                 return False
         return True
 
+    def set_context(self, ctx, arch=None):
+        """Assign values to context of edge"""
+        if (arch is None):
+            arch = self
+        for i in ctx.keys():
+            if (i not in arch.context.keys()):
+                logger.warning(str(f"{i} not a context member of {self}"))
+                continue
+            else:
+                self.context[i] = ctx[i]
+        self.save()
+
     def destroy(self):
         """
         Destroys self from memory and persistent storage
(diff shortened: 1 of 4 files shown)


jaseci_core/jaseci/graph/edge.py
```diff
@@ -123,6 +123,18 @@ def connects(self, source=None, target=None, ignore_direction=False):
                 return False
         return True
 
+    def set_context(self, ctx, arch=None):
```
ypkang, line 126: @marsninja I copied this function over from the node class. I am not 100% sure what the `arch` parameter is used for though...

jaseci_core/jaseci/jac/interpreter/architype_interp.py
```diff
@@ -243,7 +244,7 @@ def run_dot_a_list(self, jac_ast):
             lhs_id = self.run_dot_id(kid[0])
             kid = kid[1:]
             if (len(kid) == 0 or kid[0].token_text() != '='):
-                # If there is rhs, treat it as a boolean value of True
+                # If there is no rhs, treat it as a boolean value of True
```
ypkang, line 252: Noticed this was a typo.

marsninja (approved):
Very nice leet haxorisim @ypkang

jaseci_core/jaseci/graph/edge.py
marsninja: perfect, arch is for updated architypes if the edge itself is old.

PR #171, opened 2022-03-25
Improvement for production-grade bi-encoders

AshishMahendra:


jskit/encoders/bi.py
```diff
@@ -101,16 +105,40 @@ def cosine_sim(vec_a: List[float], vec_b: List[float], meta):
 
 
 @jaseci_action(act_group=['bi_enc'], allow_remote=True)
-def infer(contexts: List, candidates: List):
+def infer(contexts: Union[List[str], List[List[float]]], candidates: Union[List[str], List[List[float]]]):
```
ypkang: Let's make it an explicit function parameter what is being passed in. It will make this API easier to pick up and you can use that parameter to do the logic accordingly so you don't have to do all the list data checking stuff. You can do something like this
```
def infer(contexts, candidates, context_type, candidate_type):
```
context_type can be "text" or "embedding"

jskit/encoders/bi.py
```diff
@@ -101,16 +105,40 @@ def cosine_sim(vec_a: List[float], vec_b: List[float], meta):
...
+        cand_embed = []
+        predicted_candidates = []
+        cand_embed = get_candidate_emb(candidates)
+        for data in contexts:
+            score_dat = []
+            for lbl in cand_embed:
+                score_dat.append(cosine_sim(
+                    vec_a=data, vec_b=lbl, meta="string"))
+            predicted_candidates.append(candidates[np.argmax(score_dat)])
+    else:
```
ypkang: Also should support passing in both embeddings to candidates and contexts.

jskit/encoders/bi.py
```diff
@@ -101,16 +105,40 @@ def cosine_sim(vec_a: List[float], vec_b: List[float], meta):
...
+        raise HTTPException(status_code=404, detail=str(
+            "Invalid argument"))
```
ypkang: With the explicit function parameter, you can also make this error message more detailed.

jskit/encoders/utils/tokenizer.py
```diff
@@ -96,8 +96,76 @@ def batchify_join_str(self, batch):
         return contexts_token_ids_list_batch, contexts_input_masks_list_batch, \
             candidates_token_ids_list_batch, candidates_input_masks_list_batch, labels_batch
 
+# this class transforms data in required training format and inference for text and tokens
+
+
+class InferDataset(Dataset):
```
ypkang: I am not sure what's the purpose of this. Was this not needed before for inference?

jskit/encoders/utils/evaluate.py
```diff
@@ -60,7 +61,13 @@ def get_inference(model, tokenizer, contexts, candidates):
             logits = model(context_token_ids_list_batch,  context_input_masks_list_batch,
                            candidate_token_ids_list_batch, candidate_input_masks_list_batch, mode="eval")
         _, prediction = torch.max(logits, dim=-1)
-        return candidates[prediction]
+        prediction = prediction.tolist()
+        if isinstance(prediction, list):
```
ypkang, line 43: Isn't `prediction` guaranteed to be a list?

jskit/encoders/utils/evaluate.py
```diff
@@ -47,10 +47,11 @@ def get_inference(model, tokenizer, contexts, candidates):
     )
     val_dataloader = DataLoader(
         test_data,
-        batch_size=1,
+        batch_size=2,
```
ypkang: Why is this changed?

ypkang (changes requested):
Good changes overall but I'd like us to tighten this up some more before merging. Left comments and questions.

jskit/encoders/utils/tokenizer.py
AshishMahendra: This is removed

jskit/encoders/bi.py
AshishMahendra: fixed

jskit/encoders/bi.py
AshishMahendra: fixed

jskit/encoders/bi.py
```diff
@@ -11,16 +11,18 @@
 from utils.train import train_model
 from jaseci.actions.live_actions import jaseci_action
 import random
+import json
+import shutil
 
-config = configparser.ConfigParser()
-model, model_name, shared, seed, tokenizer = None, None, None, None, None
+# config = configparser.ConfigParser()
```
ypkang: remove if its not needed anymore.

jskit/encoders/bi.py
```diff
@@ -11,16 +11,18 @@
...
+model, model_name, tokenizer = None, None, None
 save_restart = False
 basepath = None
-device = torch.device("cpu")
+# device = torch.device("cpu")
 # uncomment this if you wish to use GPU to train
 # this is commented out because this causes issues with
 # unittest on machines with GPU
-# device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
```
ypkang, line 21: Does this work for you on your machine? When I run this on my machine which has a GPU but if I want to use the CPU, this fails. Would be good to figure out how to make this work properly.

jskit/encoders/bi.py
```diff
@@ -30,29 +32,23 @@ def set_seed(seed):
...
 def config_setup():
     """
     Loading configurations from utils/config.cfg and initialize tokenizer and model
     """
-    global seed, model, save_restart, tokenizer, device, shared, config, basepath
-    model_name = config['MODEL_PARAMETERS']['MODEL_NAME']
-    shared = config['MODEL_PARAMETERS']['SHARED']
-    seed = int(config['TRAIN_PARAMETERS']['SEED'])
-    basepath = config['TRAIN_PARAMETERS']['BASEPATH']
+    global model, save_restart, tokenizer, shared, model_config, model_save_path
```
ypkang: We should explore if there is a way to avoid global variables.

jskit/encoders/bi.py
```diff
@@ -79,14 +75,18 @@ def config_setup():
         save_restart = False
 
     model.to(device)
-    set_seed(seed)
+    set_seed(12345)
```
ypkang: This seed should probably a configuration parameter.

jskit/encoders/bi.py
```diff
@@ -79,14 +75,18 @@ def config_setup():
...
 
 
 config_setup()
 
-# API for getting the cosine similarity
+
+# to validate data and datatype provided
+def is_list_of(data, dtype):
```
ypkang: We don't need this anymore right?

jskit/encoders/bi.py
```diff
@@ -101,16 +101,48 @@ def cosine_sim(vec_a: List[float], vec_b: List[float], meta):
 
 
 @jaseci_action(act_group=['bi_enc'], allow_remote=True)
-def infer(contexts: List, candidates: List):
+def infer(contexts: Union[List[str], List[List]], candidates: Union[List[str], List[List]], context_type: str, candidate_type: str):
```
ypkang: We should start enforcing code formatting/linting rules on jskit code as well. We can talk about this.

jskit/encoders/bi.py
```diff
@@ -101,16 +101,48 @@ def cosine_sim(vec_a: List[float], vec_b: List[float], meta):
...
-                                        contexts=contexts,
-                                        candidates=candidates)
-    return predicted_candidate
+    predicted_candidates = []
+    if (context_type == "text") and (candidate_type == "text"):
+        predicted_candidates = get_inference(model, tokenizer,
+                                             contexts=contexts,
+                                             candidates=candidates)
+    elif (context_type == "text") and (candidate_type == "embedding"):
+        con_embed = []
```
ypkang: `con_embed` is being overwritten immediately on the next line so this line is not necessary right?

jskit/encoders/bi.py
```diff
@@ -101,16 +101,48 @@ def cosine_sim(vec_a: List[float], vec_b: List[float], meta):
...
+        con_embed = get_context_emb(contexts)
+        for data in con_embed:
+            score_dat = []
+            for lbl in candidates:
+                score_dat.append(cosine_sim(
+                    vec_a=data, vec_b=lbl, meta="string"))
+            predicted_candidates.append(float(np.argmax(score_dat)))
+
+    elif (context_type == "embedding") and (candidate_type == "text"):
+        cand_embed = []
```
ypkang: Same here. `can_embed` is overwritten on line 128 anyways.

jskit/encoders/bi.py
```diff
@@ -101,16 +101,48 @@ def cosine_sim(vec_a: List[float], vec_b: List[float], meta):
...
+        predicted_candidates = []
```
ypkang: Redundant from line 110.

jskit/encoders/bi.py
```diff
@@ -101,16 +101,48 @@ def cosine_sim(vec_a: List[float], vec_b: List[float], meta):
...
+    elif (context_type == "embedding") and (candidate_type == "embedding"):
+        for data in contexts:
+            score_dat = []
+            for lbl in candidates:
+                score_dat.append(cosine_sim(
+                    vec_a=data, vec_b=lbl, meta="string"))
+            predicted_candidates.append(float(np.argmax(score_dat)))
+    else:
+        raise HTTPException(status_code=404, detail=str(
+            "input type not supported"))
```
ypkang: Let's make this error message more informative. Let's also provide what input type are supported and what are being passed in actually.

jskit/encoders/bi.py
```diff
@@ -101,16 +101,48 @@ def cosine_sim(vec_a: List[float], vec_b: List[float], meta):
...
+                                             candidates=candidates)
+    elif (context_type == "text") and (candidate_type == "embedding"):
+        con_embed = []
+        con_embed = get_context_emb(contexts)
+        for data in con_embed:
+            score_dat = []
+            for lbl in candidates:
+                score_dat.append(cosine_sim(
+                    vec_a=data, vec_b=lbl, meta="string"))
+            predicted_candidates.append(float(np.argmax(score_dat)))
```
ypkang: Hmm why is this casted to `float`? Integer should be fine for indices.

jskit/encoders/bi.py
```diff
@@ -101,16 +101,48 @@ def cosine_sim(vec_a: List[float], vec_b: List[float], meta):
...
+                score_dat.append(cosine_sim(
+                    vec_a=data, vec_b=lbl, meta="string"))
+            predicted_candidates.append(candidates[np.argmax(score_dat)])
+    elif (context_type == "embedding") and (candidate_type == "embedding"):
+        for data in contexts:
+            score_dat = []
+            for lbl in candidates:
+                score_dat.append(cosine_sim(
+                    vec_a=data, vec_b=lbl, meta="string"))
+            predicted_candidates.append(float(np.argmax(score_dat)))
```
ypkang: Same here, why cast to float?

jskit/encoders/bi.py
```diff
@@ -101,16 +101,48 @@ def cosine_sim(vec_a: List[float], vec_b: List[float], meta):
...
+        for data in contexts:
+            score_dat = []
+            for lbl in cand_embed:
+                score_dat.append(cosine_sim(
+                    vec_a=data, vec_b=lbl, meta="string"))
+            predicted_candidates.append(candidates[np.argmax(score_dat)])
+    elif (context_type == "embedding") and (candidate_type == "embedding"):
+        for data in contexts:
+            score_dat = []
+            for lbl in candidates:
```
ypkang: there are quite a bit of duplicate code here between these if/elif blocks. I would re-factor it a bit. You can first convert both candidates and context to embeddings according to the input type. And then they can share this for-loop for calculating cos_sim and finding the maximum one.

jskit/encoders/bi.py
```diff
@@ -162,36 +194,67 @@ def get_candidate_emb(candidates: List):
...
 
+
 # API for setting the training and model parameters
+@jaseci_action(act_group=['bi_enc'], allow_remote=True)
+def get_train_config():
+    try:
+        with open("utils/train_config.json", "r") as jsonfile:
+            data = json.load(jsonfile)
+            print("Read successful")
+        print(data)
```
ypkang: Do you intend to keep these prints?

jskit/encoders/bi.py
```diff
@@ -162,36 +194,67 @@ def get_candidate_emb(candidates: List):
...
-        save_restart = True
-    with open("utils/config.cfg", 'w') as configfile:
-        config.write(configfile)
+def set_train_config(training_parameters: Dict = None):
     try:
+        with open("utils/train_config.json", "r") as jsonfile:
+            data = json.load(jsonfile)
+        with open("utils/train_config.json", "w+") as jsonfile:
+            data.update(training_parameters)
+            json.dump(data, jsonfile, indent=4)
```
ypkang: You can just load the jsonfile once with both read and write.

jskit/encoders/bi.py
```diff
@@ -162,36 +194,67 @@ def get_candidate_emb(candidates: List):
...
+@jaseci_action(act_group=['bi_enc'], allow_remote=True)
+def set_model_config(model_parameters: Dict = None):
+    global save_restart
+    try:
+        with open("utils/model_config.json", "r") as jsonfile:
+            data = json.load(jsonfile)
+        with open("utils/model_config.json", "w+") as jsonfile:
+            data.update(model_parameters)
+            json.dump(data, jsonfile, indent=4)
+        save_restart = True
```
ypkang: I am not a big fan of using global variable to corrdinate actions that are going to happen later as a result of the current API call. Let's talk about this one.

jskit/encoders/utils/evaluate.py
```diff
@@ -2,40 +2,31 @@
...
-import configparser
-config = configparser.ConfigParser()
-
-max_history, max_contexts_length, max_candidate_length, device = None, \
-    None, None, None
+import json
 
 
 # inference parameters setup
 def config_setup():
```
ypkang: Let's consolidate the different `config_setup` function from different files into one.

jskit/encoders/utils/model_config.json
```diff
@@ -0,0 +1,5 @@
+{
+    "shared": false,
+    "model_name": "prajjwal1/bert-tiny",
```
ypkang, line 3: Have we evaluated `bert-tiny` performance yet?

ypkang: I am wary of setting a model that we haven't tested performance yet as the default option.

jskit/encoders/utils/tokenizer.py
```diff
@@ -9,19 +9,24 @@ def __init__(self, contexts, candidates, context_transform, candidate_transform,
...
-                for cand, lbl in zip(candidates, labels):
+                group = {
+                    'context': None,
+                    'candidates': [],
+                    'labels': []
+                }
+                for cand in candidates:
                     group['candidates'].append(cand)
-                    group['labels'].append(lbl)
+                    group['labels'].append(0)
```
ypkang, line 24: The label is always 0?

ypkang (changes requested):
This is improving nicely overall. This review is focusing more on the fine details and coding standard.

jskit/encoders/bi.py
ypkang: Actually I will look into this one.

jskit/encoders/utils/tokenizer.py
ypkang: Leave a comment that this is just a placeholder for inference.

jskit/encoders/utils/tokenizer.py
```diff
@@ -168,7 +173,7 @@ def __init__(self, tokenizer, max_len):
 
         self.cls_id = self.tokenizer.convert_tokens_to_ids('[CLS]')
         self.sep_id = self.tokenizer.convert_tokens_to_ids('[SEP]')
-        self.tokenizer.add_tokens(['\n'], special_tokens=True)
+        # self.tokenizer.add_tokens(['\n'], special_tokens=True)
```
ypkang: remove commented out code.

jskit/encoders/utils/train.py
```diff
@@ -142,26 +106,20 @@ def train_model(model_train, tokenizer, contexts, candidates, labels, val=False)
...
-                    print(global_step, tr_loss / nb_tr_steps)
-                    log_wf.write('%d\t%f\n' %
-                                 (global_step, tr_loss / nb_tr_steps))
-
-                log_wf.flush()
-                pass
+                bar.update()
+        print(f"\nEpoch : {epoch+1} \t loss : {tr_loss/nb_tr_steps}\n")
+        log_wf.flush()
+        pass
```
ypkang: rm this.

ypkang (changes requested):

jskit/encoders/bi.py
ypkang: This is still here.

jskit/encoders/bi.py
ypkang: This still.

jskit/encoders/bi.py
```diff
@@ -30,184 +31,249 @@ def set_seed(seed):
...
+
+@ jaseci_action(act_group=['bi_enc'], allow_remote=True)
+def infer(contexts: Union[List[str], List[List]],
+          candidates: Union[List[str], List[List]],
+          context_type: str, candidate_type: str):
     """
     Take list of context, candidate and return nearest candidate to the context
     """
-    global model
+    # global model
```
ypkang: remove commented code please.

jskit/encoders/bi.py
```diff
@@ -30,184 +31,249 @@ def set_seed(seed):
...
+        for data in con_embed:
+            score_dat = []
+            for lbl in cand_embed:
+                score_dat.append(dot_prod(
+                    vec_a=data, vec_b=lbl))
+            if candidate_type == "embedding":
+                predicted_candidates.append(int(np.argmax(score_dat)))
+            else:
+                predicted_candidates.append(candidates[np.argmax(score_dat)])
+        return predicted_candidates
```
ypkang, line 135: This API is looking a lot nicer now! 👍🏼

jskit/encoders/bi.py
```diff
@@ -30,184 +31,249 @@ def set_seed(seed):
...
     """
-    global model, tokenizer, shared, config
+    # global model, tokenizer, model_config
     try:
-        if not model_path.isalnum():
+        if not model_path.replace('_', '').isalnum():
             raise HTTPException(
                 status_code=400,
-                detail='Invalid model name. Only alphanumeric chars allowed.'
+                detail="Invalid model name. Only alphanumeric chars allowed "
```
ypkang: Update this error message to be more accurate -- from the code, it looks like we allow alphanumeric and underscore.

jskit/encoders/bi.py
```diff
@@ -30,184 +31,249 @@ def set_seed(seed):
...
+        if model_config["shared"] is True:
             model.cont_bert.save_pretrained(model_path)
             tokenizer.save_vocabulary(model_path)
-            with open(model_path+"/config.cfg", 'w') as fp:
-                config.write(fp)
         else:
-            cand_bert_path = os.path.join(model_path)+"/cand_bert/"
-            cont_bert_path = os.path.join(model_path)+"/cont_bert/"
+            cand_bert_path = os.path.join(model_path + "/cand_bert")
+            cont_bert_path = os.path.join(model_path + "/cont_bert")
```
ypkang, line 269: should be `os.path.join(model_path, "cand_bert")` right?

jskit/encoders/utils/evaluate.py
```diff
@@ -1,41 +1,18 @@
-import os
 import torch
 from torch.utils.data import DataLoader
 from . import tokenizer as token_util
-import configparser
-config = configparser.ConfigParser()
 
-max_history, max_contexts_length, max_candidate_length, device = None, \
-    None, None, None
+device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
```
ypkang: Currently we are setting this in more than one place in the code. Is that necessary? can we consolidate this too>

jskit/encoders/utils/tokenizer.py
ypkang: Same here.

jskit/encoders/bi.py
AshishMahendra: I tried but doesn't work, somehow

ypkang (approved):

PR #177, opened 2022-03-29
fix syntax highlight for comments

Gim3l:
Fix syntax highlight for comments...only some were showing.


marsninja (approved):
Looks like good progress

PR #179, opened 2022-03-31
Regression testing for JSKit and github action workflow

ypkang:


ypkang:
@AshishMahendra This PR will add jskit regression tests to be part of github action, which runs with every PR. But entity extraction currently has a few failing tests that we need to fix. You can check the failure log on this PR.

AshishMahendra:
> 

Hi @ypkang , i check it's failing , as there is no requirements.txt and it's unable to find jaseci and torch package

PR #212, opened 2022-04-19
[MULTIPART-FORM-HANDLING] : Update request post to support files

amadolid:


marsninja:
Hi @amadolid , This is a very nice contribution!, can you sign the CLA, I'll also do a quick review before approving. Thanks! and Keep it up!!! :-D

jaseci_core/jaseci/actions/standard/file.py
```diff
@@ -40,3 +42,38 @@ def dump_json(fn: str, obj, indent: int = None):
     """Standard built in for dumping json to file from dictionary"""
     with open(fn, 'w') as file:
         json.dump(obj, file, indent=indent)
+
+
+@jaseci_action()
+def build_form_data_from_b64(files: list):
```
marsninja: Are `files` in this case  a list of bytestrings? If so we should probably call this parameter `byte_strs` or `byte_strings`

Or perhaps it seems like to goal of this lib call would be to get a list of decoded filenames from b64 encoded strings, if so perhaps we can keep it as `files` but change the call from `build_form_data_from_b64` to `get_files_from_b64`

Would love to hop on a quick zoom to discuss.

jaseci_core/jaseci/actions/standard/file.py
```diff
@@ -40,3 +42,38 @@ def dump_json(fn: str, obj, indent: int = None):
...
+        res.raise_for_status()
+        with BytesIO() as buffer:
+            for chunk in res.iter_content(chunk_size=8192):
+                buffer.write(chunk)
+            ret = buffer.getvalue()
+            
+    return ret
+
+@jaseci_action()
+def bytes_to_base64(file: bytes, encoding: str = "utf-8"):
```
marsninja: Similar question here based on the scope of use case with this lib call. Lets hop on a quick call :-)

jaseci_core/jaseci/actions/standard/request.py
```diff
@@ -23,16 +23,17 @@ def get(url: str, data: dict, header: dict):
 
 
 @jaseci_action()
-def post(url: str, data: dict, header: dict):
+def post(url: str, data: dict, files: list, header: dict):
```
marsninja: This looks great is great!

marsninja (changes requested):
Also, it would be great to add a few test cases for these addtions. I'd be happy to show you a nice way to add these test cases (mostly cut and paste of other tests). Do join the official Jaseci Open Source slack channel and we can chat more. Do share your email you'd like to add. You can always email me jason@jaseci.org.

jaseci_core/jaseci/actions/standard/file.py
amadolid: Actually, I'm planning to remove this and replace it with **base64_to_bytesio** (just decoding base64 and convert it to BytesIo)

The reason for adding **build_form_data_from_b64** is we need to upload file to a third party service. Their API uses multipart/form-data. I'm not fully sure what other approach to this other than using requests.post with array of tuples for files.

**build_form_data_from_b64** accepts json array
```
[
{
"name": "fileName.ext",
"base64": "base64=="
}
]
```
and will return array of tuples used as form-data 
```
[
('file', (fileName.ext, Bytesio(decoded base64)))
]
```

If I can declare tuples on jac file as same with python I think I can use the **base64_to_bytesio** instead

jaseci_core/jaseci/actions/standard/file.py
amadolid: This is for converting the file's content from the 3rd party service. When someone tries to download their uploaded file,
the 3rd party service's API will return it as bytes reason for converting it back to base64 so I can include it on the json response.

amadolid:
Hello @marsninja,

I have removed the update on **_post_** action. instead, I have added **_multipart_** action.
This is to avoid conflict on existing post usages. **_multipart_** method will be used solely for post call with **_Content-Type: multipart/form-data_**. Header **_Content-Type: multipart/form-data_** will also be included automatically once you have used the files parameter on requests.post

This is also the same approach on java MockMvcRequestBuilders. Multipart approach has different method solely for **_multipart/form-data_**.

--------

The two additional action **_base64_to_bytesio_** and **_file_to_form_data_** will be the replacement for **_build_form_data_from_b64_**

**_build_form_data_from_b64_** this method is solely for converting array of json to array of tuples so we can use it as form-data on multipart request. This requires specifc json structure which is 

```
[
   {
      "name": "fileName.ext",
      "base64": "base64=="
   }
]
```

and not all developer can use this reason for refactoring it to be more dynamic and useful

previously, developer can use **_build_form_data_from_b64_** if they have the same structure as ours. Now they can just manipulate it on jac file itself
ex:

```
file.file_to_form_data(
    "fieldName",
    "fileName.ext",
    file.base64_to_bytesio(
        "base64====="
    ) 
)
```

**_base64_to_bytesio_** will be the converter of base64. We use base64 since we pass it on context. **_file_to_form_data_** will be the converter of any file to be a tuple (act as form-data for multipart request)


Please let me know if you concern or suggestions on the new approach

marsninja:
Hi @amadolid Ping me when you can, I had a chat with @asimsalim He can also bring you up to speed, Looking forward to mergining in your first awesome commit!!!

marsninja (approved):
This is fine for now @amadolid , Lets have a chat and talk through a final set of multiparts :-)

PR #214, opened 2022-04-19
entity-extraction-type2 version 1.0

SatyamRajawat:
Api for entity-extraction-type2 version 1.0


marsninja (approved):
Approving for now, in general we'd want some tests, but since testing is a mess on jaseci kit right now we'll overhaul this for the whole package soon.

PR #236, opened 2022-04-24
T5 summarization

Shawn-Jemmott:
Added T5 Model Summarization to Jaseci-Kit ( small model ) which is about 230MB, and returns data at an average speed of 2.1s. It collects a key of text ( required ), min_length ( optional ), max_length ( optional ).


marsninja (changes requested):
Hey @Shawn-Jemmott can you sign the CLA real quick.

marsninja (changes requested):
Hi @Shawn-Jemmott CAn you move all the various folders into the t5_sum folder and do any cleanup that makes sense like the text_summarization folder seems to have requirements and the summarizaiton has requirements. Have a quick look

ypkang:
This looks good ( +1 for the test!). Could you fix the code formatting issue? Also could you add a section for T5 summarizer to the jaseci kit README? You can use the USE encoder one as an example. So Add an entry to the model directory table at the top and add a section for T5 summarization. Link to the readme here https://github.com/Jaseci-Labs/jaseci/tree/main/jaseci_kit#model-directory

ypkang (approved):
Great job on this PR @Shawn-Jemmott. The readme looks good. Ready to 🚢

PR #247, opened 2022-05-03
ner type 2 training and unittest case

SatyamRajawat:
ner type 2 tarining and unittest case


marsninja (commented):
Lets Keep the *Zone.Identifier files out of repo, perhaps we can add it to gitignore as part of this pr

marsninja:
@AshishMahendra can you review this PR as well?

marsninja (approved):
This set of changes are a solid improvement on prior state and appear to not have any conflicts. Lets iterate here.

PR #249, opened 2022-05-03
Poly-Encoder updated

AshishMahendra:


marsninja (changes requested):
Hi @AshishMahendra , add --max-line-length=88 --extend-ignore=E203 commands to flake8 and black will work just fine. It good to have black as the formatter and flake8 as the checker but it requires these two commands. You can also add these as args for the flake8 config in VSCode properties. Just search flake8 settings and add teh two args.

AshishMahendra:
> --max-line-length=88 --extend-ignore=E203

Sure, I'll try this

marsninja (approved):

PR #258, opened 2022-05-06
Updated Bi encoder for relative import

AshishMahendra:


ypkang (approved):
@AshishMahendra So Jason and I met about this and came up with this solution that doesn't require modifying `sys.path`.  I pushed the changes to this branch. If you run 
```
python3 -m unittest discover . -p "test_*.py"
```
in `jaseci_kit/`, it will find all the test cases and run them.

However, the biencoder test cases seem to be failing currently, which we will need to fix.

PR #262, opened 2022-05-10
Adjust path to the text segmenter

Gorgeous-Patrick:
The path is wrong. Please check Ashish's repository.

The alternative way is to clone the main repo instead of ashish's

### DIFF
--- a/jaseci_kit/jaseci_kit/modules/text_segmenter/text_segmenter.yaml
+++ b/jaseci_kit/jaseci_kit/modules/text_segmenter/text_segmenter.yaml
@@ -18,11 +18,10 @@ metadata:
   namespace: default
 data:
   prod_up: |-
-    git clone https://github.com/AshishMahendra/jaseci.git /jaseci/
+    git clone https://github.com/Jaseci-Labs/jaseci.git /jaseci/
     cd /jaseci/jaseci_core/
-    git checkout text_segmentation_prod
     source install.sh
-    cd /jaseci/jskit/text_segmenter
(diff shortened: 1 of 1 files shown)


jaseci_kit/jaseci_kit/modules/text_segmenter/text_segmenter.yaml
```diff
@@ -22,7 +22,7 @@ data:
     cd /jaseci/jaseci_core/
     git checkout text_segmentation_prod
     source install.sh
-    cd /jaseci/jaseci_kit/jaseci_kit/modules/text_segmenter
+    cd /jaseci/jskit/text_segmenter
```
marsninja: Hi @Gorgeous-Patrick I dont think this path would work with the latest main branch. Take a look.

marsninja (changes requested):

jaseci_kit/jaseci_kit/modules/text_segmenter/text_segmenter.yaml
Gorgeous-Patrick: Well, yes it does not work for the main branch. And the main repo does not have that `text_segmentation_prod` branch actually. 
The other way around is to use the programs and paths in the main branch in the main repo. I just thought that maybe the programs are different (the branch is named a production branch so it looks like a more mature version). @marsninja

ypkang: Not sure why this manifest was pointing to Mahendra's fork instead of the main repo. I have corrected it and it should work now.

PR #263, opened 2022-05-11
[WALKER_API]: Support Query params and raw context payload

amadolid:
Removed some unreachable lines

### DIFF
--- a/jaseci_core/jaseci/api/interface.py
+++ b/jaseci_core/jaseci/api/interface.py
@@ -25,7 +25,14 @@ class interface:
         """
         self._pub_committer = None
 
-    def assimilate_api(api_list, func, cmd_group=None, cli_args=None, url_args=None):
+    def assimilate_api(
+        api_list,
+        func,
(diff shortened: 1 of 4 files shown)


marsninja (commented):
Love how lightweight this PR is. Lets validate this solves what @asimsalim needs and then we can add a test and a comment here or there. Very nice @amadolid

marsninja (changes requested):
Hi @amadolid This is an awesome change!!! Very impressive.  Lets just add a test, check out https://github.com/Jaseci-Labs/jaseci/blob/main/jaseci_serv/jaseci_serv/jac_api/tests/test_jac_api.py  and you can cut and paste a test or something, there are good ones around line 400 ish. Lets just plop a test. We can discuss more in call today.

marsninja (approved):
Accepting this change so we don't let the PR sit too long, was going to have you write a quick test @amadolid but I'll do one and show it to you next week.

PR #278, opened 2022-05-17
test cases updated

AshishMahendra:
## Describe your changes
created test cases for :
1. fasttext
2.  summarization(cl_summer)
3. text segmenter
4. Use Enc
5. Use QA
## Link to related issue
N/A
N/A
N/A
N/A


ypkang (changes requested):
@AshishMahendra as mentioned on the other PR #258 , we figured out a solution that doesn't require modifying sys.path. Let's adopt that and update it for the test cases here.

jaseci_kit/jaseci_kit/modules/fasttext/tests/test_fast_enc.py
```diff
@@ -0,0 +1,43 @@
+import sys
+from unittest import TestCase
+from jaseci.utils.utils import TestCaseHelper
+from fasttext_classifier import serv_actions
+from fastapi.testclient import TestClient
+import os
```
ypkang: let's make sure not import packages that are not being used. I don't think `os` is used in here.

ypkang (changes requested):

AshishMahendra:
@ypkang I have updated and checked the testing for all modules, findings:
1. t5_sum has some error while testing which i didn't deep dive ( it would be helpful is Shawn can have a look)
2. I wasn't able to run all at once as my system ran out of memory  
3. I have updated bi.py and entity extraction for config files import issue

ypkang:
@AshishMahendra some of the biencoder and NER tests are still failing. I updated the test.sh in jaseci_kit. That script will run everything. We should look at these failure together.

ypkang (changes requested):
some of biencoder and ner tests are still failing.

ypkang:
1, confirm that click version misamtch is the issue and come up with somlution to upgrade/downgrade one
2. add a quick fastapi based test for t5_sum

3. fix t5_sum import path issue

ypkang (approved):

PR #285, opened 2022-05-20
[FILE-UPLOAD-CONCEPT]: Initial file upload handling

amadolid:
## Describe your changes
 - supports multipart/form-data (MFD) content-type
 - MFD will have default field `jaseci-context` used for json body (context)
 - additional field will be accessible via _context.files.{{field name}}

## Link to related issue
N/A

New feature, to be added

## Will this impact Jaseci users?
User can now implement multipart/form-data approach. All of their previous implementation will be also supported by just using form-data 'jaseci-context' = {their json body}.
Adding different attribute like 'files' will be accesible thru _context.files["field_name"]
Structure:

_context.files = {
  "jaseci-context": [{
    "name": "body.json",
    "base64": "#####"
  }],
  "files: [{
    "name": "uploadedFile.pdf",
    "base64": "########"
  }]
}

## Anything in this PR should the Jaseci developer/contributor community pay particular attention to?
Current implementation should not be affected

### DIFF
--- a/jaseci_serv/jaseci_serv/jac_api/views.py
+++ b/jaseci_serv/jaseci_serv/jac_api/views.py
@@ -1,3 +1,4 @@
+import json
 from rest_framework.views import APIView
 from knox.auth import TokenAuthentication
 from rest_framework.permissions import IsAdminUser, IsAuthenticated, AllowAny
@@ -8,6 +9,8 @@ from jaseci_serv.base.orm_hook import orm_hook
 from jaseci_serv.base.models import JaseciObject, GlobalVars
 from jaseci_serv.base.models import master as core_master
 from time import time
+from base64 import b64encode
+from io import BytesIO
 
 
 class JResponse(Response):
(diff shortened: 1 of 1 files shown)


marsninja (commented):
Hi @amadolid This looks super interesting, would love to have you walk me through the use case.

amadolid:
Hi @marsninja,

Currently, to work with upload file, we pass base64 on ctx payload. The file will also have bigger byte size if converted to base64. This approach causes error on json parsing if the file is too large (we have tested ~4MB).

To handle this issue, I have implemented multipart/form-data handling. This is to pass the actual file and read~convert it to on python side so that it will automatically be json serializable once passed to jaseci side.

The approach will look like this

POST
Content-Type: multipart/form-data

##Form Data Fields:

**jaseci-content**
anyName.json
`{
"name": "walker_sample",
"ctx": {
"testing": true,
},
"nd": "active:graph",
"snt": "active:sentinel"
}`

**files**
sample.pdf


##On Jaseci side
_context.body
`{
"name": "walker_sample",
"ctx": {
"testing": true,
},
"nd": "active:graph",
"snt": "active:sentinel"
}`

_context.files
`[{
"name": "sample.pdf",
"base64": "#######"
}]`

amadolid:
moved to https://github.com/Jaseci-Labs/jaseci/pull/295

PR #298, opened 2022-06-06
Shyam sso implementation

pathakshyam:
## Describe your changes

Did some modification in User  model added additional fields called **auth_provider**, which help us to identify which auth provider service user associated with also added serializers and views for facebook and google authentication.

## Link to related issue

NA

Its a new feature for social login allows user to login various social sites using auth token.

Yes, it will provide them a flexibility to login to Jaseci server by social login, by this feature users wont have to manage multiple accounts.

Nothing, as of now


ypkang:
@asimsalim should we close this PR? this is going into jaseci-x (if it hasn't already), right?

asimsalim:
@ypkang That is correct. This is now in Jaseci-X . So we can close this PR.

PR #305, opened 2022-06-12
Version substitute in jaseci kit

ashagraw91:
## Describe your changes

Make use of version number for github for jaseci KIT and substitute specif version number

## Link to related issue
https://app.clickup.com/t/2dtqwpf

## Does this introduce a new feature or change the existing feature of Jaseci? If so, please add tests.
yes

Should have no impact

The versioning of the jaseci KIT while deployments


asimsalim (changes requested):
@ashagraw91  Can we create new yaml files as discussed ?
It can be in the format use_enc.prod.yaml 
This is also required for Jaseci.yaml

ypkang:
@asimsalim @ashagraw91 I like this change. Just curious -- what is the original problem/use case that lead to this change? 
And, yea I agree it would be good to keep the option of installing from source.

jaseci_kit/jaseci_kit/modules/use_enc/README.md
```diff
@@ -0,0 +1,19 @@
+# Install Specific version of jaseci KIT model in the POD
+
+To install Specific version of jaseci KIT, Follow below process.
+The use_enc.yaml ConfigMAP has environmnet variable declared as $version , which we will substitute while applying the Yaml 
+
+
+### Export environment variable 
+
+```bash
+export version= {version no}  i.e 1.3.3.16
```
marsninja: @ashagraw91 I'd use a bit more specific env var like "jskit_version" throughout

ashagraw91:
@ypkang this change is because the new implementation requires specific version of jaseci kit to be deployed in a pod, keeping it latest breaks changes if there has been a major update in kit model. the version helps in keeping in the stable model for ai pods

ashagraw91:
@asimsalim I have updated for use_qa and use_enc. jaseci willnot need to change as the pod Ip for both will same for jaseci pod. Also let me know if other yamls also need to be updated?

ypkang (approved):
This looks good to me.

marsninja (approved):
LGTM

marsninja (approved):
LGTM

PR #307, opened 2022-06-14
[MULTIPART-FORMDATA-ENHANCEMENT]: Allow file type field to be access using `has fieldName`

amadolid:
## Describe your changes
 - File type field will now be able to access via `has fieldName`
 - Update file parser to be able to read large file

## Link to related issue
N/A

Nope, tho I have added unit tests for different multipart scenarios

Slightly, file type field will now be accessible via `has fieldName` instead of using `global.info["request_context"]["body"]["fieldName"]`

multipart/form-data approach with files


amadolid:
## Sample Request

image


ctx still requires json string or json file

file type fields like `files` in this example will now be accessible via `has files`

PR #322, opened 2022-07-06
[TRY-ELSE]: Improve exception backtracking when using Try/Else

amadolid:
## Describe your changes
 - Added TryException class to improved Try/Else backtracking
 - Removal of default exception handling on actions trigger interp to enable Try/Else on jaseci actions

## Link to related issue
N/A

Yes, you can now use Try/Else on Jaseci Actions
No, It's supposed to be an improvement for Try/Else backtracking

Yes, Try/Else can now determine the exact line of the error.
Jaseci Actions now throws exception to be handled on Jac side

Jaseci Actions, the removal of exception handling might need changes on current/existing jac projects since exception should be handled "mandatorily" instead of having fix handling

### DIFF
--- a/jaseci_core/jaseci/jac/interpreter/interp.py
+++ b/jaseci_core/jaseci/jac/interpreter/interp.py
@@ -4,6 +4,7 @@ Interpreter for jac code in AST form
 This interpreter should be inhereted from the class that manages state
 referenced through self.
 """
+import logging
 from jaseci.utils.utils import is_jsonable, is_urn, parse_str_token
 from jaseci.element.element import element
 from jaseci.graph.node import node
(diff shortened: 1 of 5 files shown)


jaseci_core/jaseci/jac/interpreter/interp.py
```diff
@@ -401,6 +405,12 @@ def run_report_action(self, jac_ast):
                 self.report_status = self.run_expression(kid[4]).value
             elif kid[2].token_text() == "custom":
                 self.report_custom = self.run_expression(kid[4]).value
+            elif kid[2].token_text() == "error":
+                err = self.run_expression(kid[4]).value
+                self.runtime_errors.append(
```
marsninja, line 409: Very good pr so far @amadolid ! Here I would use a the rt_error function that is inherited from this class: https://github.com/Jaseci-Labs/jaseci/blob/cd8943b5cbe7b27c530fa840b8fdf9287910a918/jaseci_core/jaseci/jac/machine/machine_state.py#L144

You can see it used in other places in this interp.py file.

marsninja (approved):
This LGTM for now, its a nice improvement!

PR #328, opened 2022-07-12
fix: remove unnecessary match regex

Gim3l:
## Describe your changes
Regex no longer necessary; breaks syntax highlighting.


marsninja (approved):
LGTM, thanks @Gim3l

PR #332, opened 2022-07-20
[DEREF-SYNTAX]: Node crosscheck and backward compatibility

amadolid:
## Describe your changes
Added checking if deref syntax argument is already a node
Added backward compatibility for different prefixes (urn. jac, noprefix) 

## Link to related issue
N/A

No

No

The deref approach since it will not check any prefix on request. It will just append jac prefix once uuid is available


marsninja (approved):
HI @amadolid I'll make some tweaks and share a followon PR with you. Basically I'll move teh matching uuid regex to utils, and add a few tests, Accepting this now.  I'm also going to add an error case that if the string is greater than 32 chars we print error. Let me know if this policy is fine with you.

PR #340, opened 2022-07-25
[UNIT-TEST]: Transfer try test to new file to avoid conflict

amadolid:
## Describe your changes
Moved the walkers with try from `zsb.jac` to `try_syntax.jac`. This is to avoid conflicts when other dev added unit test. The try_catch test will determine the exact line on jac file. If the jac file is updated and new lines is added above the initial walkers, it will affects the try_catch test

## Link to related issue
N/A

No

No

New walkers for unit test should be added on zsb.jac or new jac file


marsninja (approved):
This LGTM! Thanks @amadolid

PR #345, opened 2022-07-26
Yolov5

Shawn-Jemmott:
@ypkang


ypkang:
I still see this git submodule at `jaseci_kit/modules/object_detection/yolov5/`. I assume this is the source code for the yolov5 implementation. Do you think we can bring the source code in? Or import it as a python package if it supports that?

Shawn-Jemmott:
@ypkang  what do you mean by brining the source code in

Shawn-Jemmott:
@ypkang I will check to see if there is a python package and implement it but that may take a while since I will have to do some it.

ypkang (approved):

PR #346, opened 2022-07-26
CanoniCAI Documentation

Shawn-Jemmott:
@ypkang  seems there is not any specific format based on all the docs I ran through.


ypkang:
Oh I think we just need an empty directory for now. This will be where you are committing new content to. @Shawn-Jemmott

Shawn-Jemmott:
Great will change that now

ypkang:
@Shawn-Jemmott can you take a look at the branch used to create this PR? Not sure why it is showing 176 files changed.

ypkang:
Since the PR diff is a little messed up at the moment, I am just going to leave my feedback and comments here (as opposed to inline).

https://github.com/Shawn-Jemmott/jaseci/blob/documentation/docs/docs/canonicai/chapter1.md

* This is a great start! 
* For the real example and diagram for the AI capabilities, let's just keep one diagram. So we can just keep the one you made (which is a variation of the first diagram for that AI capability). That's what I meant by real example
* We should be more descriptive with more text in the codelab. For each diagram, there should be a (short) paragraph explaining the diagram. The diagrams are to complement the text and make it easy to understand but it doesn't replace the text description.

PR #352, opened 2022-07-28
[FEATURE-REQUEST]: Simplified Dot

amadolid:
## Describe your changes
Dot mode now have `detailed` param. Default dot mode will now include the label only. Detailed dot mode will include every attributes of the node/edge

## Link to related issue
N/A

Yes, Unit test for `dot` is now updated

Yes, existing graph_get should now have `detailed: true` but if they doesn't use the node/edge attributes it should work fine

Yes, existing graph_get usage


marsninja (approved):
This is a very nice improvement. Thanks @amadolid I'll merge and make some tweaks in a separate PR. (such as updating the docs etc)

PR #353, opened 2022-07-28
Chapter 1

Shawn-Jemmott:
@ypkang


ypkang:
@eldonm @king-mango FYI we are planning to directly add the canoniCAI written codelabs to the docs. These are first versions and we plan to do another iteration once we have a complete version of all the chapters.

PR #358, opened 2022-07-29
canonicai codelab code for chapter 3

ypkang:
Work in progress. Need discussion.

Todos:

- [ ] Context retention across requests
- [ ] Add test cases.

### DIFF
--- a/examples/cai_edu/main.jac
+++ b/examples/cai_edu/main.jac
@@ -12,7 +12,7 @@ walker talk {
         if (!traveled) {
             here::nlu;
             # Entity transition takes precedance over intent transition
-            take -[entity_transition(entities=extract_entities.d::keys)]-> node::cai_state else {
+            take -[entity_transition(entities=extracted_entities.d::keys)]-> node::cai_state else {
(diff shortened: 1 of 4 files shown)


examples/cai_edu/main.jac
```diff
@@ -0,0 +1,68 @@
...
+        state_confirmation -[intent_transition(
+            intent = "yes"
+        )]-> state_confirmed;
+
+        state_confirmation -[intent_transition(
+            intent = "no"
+        )]-> state_collect_info;
+
+        state_collect_info -[intent_transition(
+            intent = "cancel"
```
Shawn-Jemmott, line 55: @ypkang intent needs to be changed from "cancel" to "collect info". A small bug I discovered.

ypkang: Why is that? This is a transition going from collect_info the canceled state so the intent should be `cancel`, no?

Shawn-Jemmott: Lol my fault now I saw that.

PR #364, opened 2022-08-03
Chapter3

Shawn-Jemmott:
@ypkang  initial version of chapter 3


docs/docs/canonicai/chapter3.md
```diff
@@ -0,0 +1,339 @@
...
+    can drive {
+        report 'yes';
+    }
+}
+```
+In this case it will seem that writing it this way will be more effient, however what if each vehicles have more functions like blow_horn, accelerate, breaks, park and etc or they are more vehicles like truck, scooter, and etc you can imagine how big this code base will be and that's why it's important to have inheritance in our code base and in this program we use inheritance to increase the efficiency of the entire application.
+
+So, lets use what's in the code base and see how we implemented it. ```Do not worry about what each nodes and state works, that will be explained in detailed in the next section, just look at how it is implemented.```
+```
+node cai_state {
```
ypkang, line 68: Let's break this down into each individual node ability and explain what they do.

docs/docs/canonicai/chapter3.md
```diff
@@ -0,0 +1,339 @@
...
+node canceled:cai_state {
+    has name = "canceled";
+    can gen_response {
+        visitor.response = "No worries. We look forward to hear from you in the future!";
+    }
+}
+```
+As you can see here, all of the conversational state node have something in common, that's why it is very important to use inheritance, because this code instead of being 200 lines of code it can be over 1000+ lines of code and it could be very hard to manage and read.
+
+You can go to [jaseci docs](https://docs.jaseci.org/) for more information on inheritance. If you are still confused about inheritance and how it works.
```
ypkang: Can we link to the specific page/section that talks about inheritance? If we don't have that in the docs yet, we can mention specific chapter of the bible.

This applies to all the places we want to link people to documentation throughout these codelabs.

docs/docs/canonicai/chapter3.md
```diff
@@ -0,0 +1,339 @@
...
+    # Construct natural language response
+    can nlg {
+        ::gen_response;
+    }
+}
+```
+Above, this is a conversational state node. We use inheritance here, because we will have multiple conversational state nodes like a confirm state, cancel state, collect information state and etc, that have the same functionality but does have different node abilities.
+
+```
+node collect_info:cai_state {
```
ypkang, line 111: Let's break this down into a code snippet + explanation for each node. For the first one, collect_info, you can explain why we are override gen_response and why we are **only** override gen_response. For the other ones, the explanation can be brief.

ypkang: Oh you actually are breaking this down later. What's the purpose of this code snippet here?

docs/docs/canonicai/chapter3.md
```diff
@@ -0,0 +1,339 @@
...
+    state_collect_info  = spawn node::collect_info;
+    state_confirmation = spawn node::confirmation;
+    state_confirmed = spawn node::confirmed;
+    state_canceled = spawn node::canceled;
+}
+```
+
+### connecting the nodes with the two types of edges
+There are two types of edges: intent_transition edge and entity_transition edge and we will use these to connect states (nodes) to each other. 
+
```
ypkang, line 243: Let's use two statements from the code below as examples to illustrates how intent and entity transition can be leveraged. You introduced intent/entity transitions earlier but there are no concrete examples yet.

docs/docs/canonicai/chapter3.md
```diff
@@ -0,0 +1,339 @@
+# Build a conversational AI system in Jaseci
+
+# Chapter 3
+
+### **cai_state node**
+### Two main stages in handling a conversational AI request
+![Two Stages](./images/nlu_nlg.png?raw=true "Title")
```
ypkang, line 7: For the diagram, can we annotate the arrows with what is the input and output of each module? For example, input to intent classificaiton would be the input text and list of candidate intent labels and output would be the predicted intent.

ypkang (changes requested):
This is a very good first version! I left some comments.

ypkang:
@Shawn-Jemmott very first version! Take a look at the comments I left. 

One more thing, could you work with @king-mango to get these codelabs show up in the docs site? I don't think they are currently being picked up.

docs/docs/canonicai/chapter3.md
Shawn-Jemmott: It's like a single page application. It don't have a link to the exact inheritance section.

docs/docs/canonicai/chapter3.md
Shawn-Jemmott: The purpose of the code snippet it to show how we used inheritance in the application. It's to show them there is no big difference between that example vs this code snippet.

docs/docs/canonicai/chapter3.md
Shawn-Jemmott: Like an example? I'm kinda confused.

docs/docs/canonicai/chapter3.md
Shawn-Jemmott: I'm kinda confuse here too like in addition to the code snippet use an example to plug into the code statment one for intent classification and one for entity extraction.

docs/docs/canonicai/chapter3.md
ypkang: Not sure if that's true. I can certainly link to a specific page in the docs, for example, https://docs.jaseci.org/docs/getting-started/JAC-Language-Overview/report.

Now, I am not sure if the current docs have any page on inheritance. If it doesn't, that's a different story.

docs/docs/canonicai/chapter3.md
ypkang: No. Just annotate the arrows with what the input and output is. Like, literally "input text" and "candidate intent labels"

docs/docs/canonicai/chapter3.md
Shawn-Jemmott: Yes now I see it. It was under this link.

https://docs.jaseci.org/docs/Developing_with_JAC/Language_Features/OOP

docs/docs/canonicai/chapter3.md
ypkang: What I meant was, breaking down the large spawn block below into smaller sections and explain how the intent and entity transition works in those examples. 

For instance, something like this,

```
        state_collect_info -[entity_transition(
            entities = ["name", "address"]
        )]-> state_confirmation;
```
"Here we have an entity transiton going from the collect_info state to the confirm state. This entity transition has its entities variable set as "name" and "address", which will be checked later in the talk walker to evaluate if this transition should be triggered or not."

docs/docs/canonicai/chapter3.md
ypkang: Ok.

docs/docs/canonicai/chapter3.md
ypkang: @Shawn-Jemmott what about this one? I don't see the changes corresponding to this comment.

docs/docs/canonicai/chapter3.md
ypkang: Okay this is moving in the right direction. Although, technically what's currently on the diagram is not correct...."candidate labels" are input to intent classification instead of the output.

* intent classification: input is "input text" and "candidate labels", output is "predicted intent".
* entity extraction: input is "input text", output is "extracted entities."

Can you update?

ypkang (changes requested):

docs/docs/canonicai/chapter3.md
Shawn-Jemmott: It's right after that section where we explain collect_info state, confirmation state and etc

ypkang (approved):

ypkang (approved):

PR #385, opened 2022-08-11
Stripe Integration

panikingginoo12:
## Describe your changes
Added action set
- creating and retrieving customer information.
- attaching, retrieving and updating payment method.
- creating, updating, cancelling and retrieving of subscription.
- creating and retrieving of customer invoice.
- creating and retrieving of customer payment.


marsninja:
Hi @panikingginoo12 This implementation as an action set looks like a great idea of how to do it. Ultimately I think you can take the stripe_api out of the master and package it with the action. We can discuss once you're ready. Keep going though, this is a great direction.

panikingginoo12:
Got it @marsninja I will do that. Thank you

ypkang:
@panikingginoo12 @marsninja what are the next steps for this one?

PR #397, opened 2022-08-15
Chapter 6 Code

Shawn-Jemmott:
@ypkang it works, however we have to discuss a bit more on the intent classification section. Currently we have 4 intents "test drive", "yes", "no", "cancel". When we ask a question like "can I test drive the tesla", it will classify to "test drive", and it will prompt for name and address but while passing in the name and address it also classifies the intent of the address and name so if I somehow ask "My name is Yiping", it will extract the slot however it may classify as "yes" and go to confirmation state and throw off the system. Should I add those utterances to test drive intent? However the AI models is in there working like a charm with the training data and etc.


Shawn-Jemmott:
@ypkang I'm currently working to fix this out.

ypkang:
@Shawn-Jemmott I reworked the ai model nodes and walker logic and push to your branch here. check it out.

ypkang (approved):

PR #423, opened 2022-09-02
[GRAPH-GET]: Only include the id when detailed is true

amadolid:
## Describe your changes
graph will now only include the id when detailed is true
## Link to related issue
N/A
No
Maybe? I'm not sure if "id" in dot is used on any 3rd party graph viewer
Dot


marsninja (approved):
LGTM

PR #439, opened 2022-09-06
[ENHANCEMENT]: Modular apps with dedicated configs and hooks isolation

amadolid:


amadolid:
Commit History:
https://github.com/Jaseci-Labs/jaseci/compare/main...amadolid:jaseci:enhancement/async-and-emailer-backup

PR #457, opened 2022-09-15
[REFACTOR]: Imports and classes

amadolid:


marsninja (approved):
Solid refactor as we start the journey to following PEP8 naming conventions @amadolid

PR #461, opened 2022-09-16
added docker support

daynauth:
## Describe your changes
Added dockerfile and bash script for running the code-along example. 

The bash script assumes that you've created a folder within this directory called 'mycode'


marsninja (changes requested):
Hi @daynauth ,

I'd add this directly into the codelab itself with instructions. Like at the beginning, I'd add a section called "Optional: Using Docker"

ypkang:
@daynauth Could you make the update @marsninja suggested and we can merge this in?

PR #468, opened 2022-09-21
Unify master_allusers interface between core and serv

ypkang:
## Describe your changes
Unify master_allusers interface between core and serv


marsninja (approved):
LGTM. Thanks  @ypkang !!

PR #482, opened 2022-09-25
fixed walker run command (missing quotes)

rgupta2718:
The walker run command was originally: walker run talk -ctx {\"question\": \"I want to schedule a test drive\"}
The command does not run as is, instead resulting in the following: Error: Got unexpected extra arguments ("I want to schedule a test drive"}). 
This is fixed by adding outer quotation marks around the brackets (as is consistent with the other -ctx arguments across the tutorial).


marsninja (approved):
LGTM! Thanks @rgupta2718

PR #489, opened 2022-09-26
Fixed discrepencies between main.jac and faq.jac to avoid confusion.

BBBenG23:
## Describe your changes
The file main.jac changed to faq.jac without any notice in the code-a-long. There was also an example which used a file named main.jac that was different than the main.jac the user created. I resolved this issue by changing the initial file we create to be named faq.jac, and changed all of the commands to work with the new name. 

This will hopefully make the code-a-long less confusing


ypkang (changes requested):
@BBBenG23 Thanks for opening up this PR. This is a good change. With this change, we also should update the place where we introduce main.jac again which is at https://github.com/Jaseci-Labs/jaseci/blob/main/examples/CanoniCAI/CCAI_codelab.md#unify-faq--dialogue-code.

Could you update the Note there accordingly and we can merge this one in?

ypkang:
@BBBenG23 Friendly ping :)

BBBenG23:
@ypkang I fixed the note.

ypkang:
> @ypkang I fixed the note.

👍🏼

PR #511, opened 2022-10-07
changed jac run to walker run

Shawn-Jemmott:
@ypkang

### DIFF
--- a/examples/CanoniCAI/CCAI_codelab.md
+++ b/examples/CanoniCAI/CCAI_codelab.md
@@ -635,9 +635,9 @@ jaseci > actions load module jaseci_kit.bi_enc
 We have provided an example training file that contains some starting point training data for the two intents, `test drive` and `order a tesla`.
 
 ```js
-jaseci > jac run bi_enc.jac -walk train -ctx "{\"train_file\": \"clf_train_1.json\"}"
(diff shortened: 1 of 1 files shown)


marsninja (approved):
LGTM! Thanks @Shawn-Jemmott

marsninja:
@ypkang Needs conflict resolution

ypkang:
@Shawn-Jemmott I don't think the changes work here. We can't just simply change `jac run` to `walker run` arbitrarily., The places where you are proposing to change, sentinel register hasn't even been introduced yet. `walker run` will not work. Also, `jsserv` is introduced at the end and most people will be running everything locally in jsctl for most of this codelab so they wouldn't really need to worry about sentinel register or walker run.

A better solution here is to add a quick new section at the end, after `jsserv`, explaining that if they are running jaseci on a remote server and wish to train model there, they can do the following (sentinel register, walker run etc.).

ypkang:
@Shawn-Jemmott could you take a look at my comment and make the change here? If anything is not clear, let me know.

ypkang (changes requested):
See comment above.

ypkang:
Gonna close this for now. Not a super important thing to add now and we can come back to it later.

PR #523, opened 2022-10-12
circular reference fix

daynauth:
## Describe your changes
Refactored ast_builder to remove a bit of circular between itself and the multipass_optimizer.


daynauth:
accidentally turned on debug

PR #568, opened 2022-10-31
Fix a bug where activation email does not send for un-activated user logging in

ypkang:
login when they are not activated yet


marsninja (approved):
Nice @ypkang !!

PR #569, opened 2022-11-01
feat(UI Kit): allow changing serverUrl for graph viewer and auth form

Gim3l:


marsninja (approved):
LGTM! @Gim3l 

Though you should bring up a testing approach and add a simple testing pipeline.

PR #574, opened 2022-11-02
Codelab for contributing and new date action included

king-mango:
## Also v1 of the contributing codelab is added.


marsninja (approved):
Nice work on this piece @king-mango

PR #577, opened 2022-11-03
[WIP] jaseci-infrastructure-terraform-gcp init

itsbharatsaini:


ypkang:
@asimsalim Is this useful to keep around? The change looks small so if it is not something we are actively working on, can we close this for now and reopen later when it is being actively worked?

marsninja (dismissed):
Looks good @itsbharatsaini

PR #606, opened 2022-11-11
[WIP] Extend api to get the total number of nodes, walkers, and edges known…

Gim3l:
… to the sentinel

## Describe your changes
Added these endpoints that I needed for jaseci studio.
- node_total
- node_list
- walker_total
- edges_total

Tests are pending.


marsninja (commented):
Hi @Gim3l , Have you looked at the architype_apis, I think perhaps some of these apis should integrate with architype list, and I imagine perhaps you could add a count, perhaps.

PR #619, opened 2022-11-16
Added missing django contrib app

pathakshyam:
Added missing django contrib app, application wasn't serving static file due to this error


marsninja (approved):
LGTM @pathakshyam

PR #622, opened 2022-11-21
[FOR-REVIEW][JSORC]: Version 2

amadolid:
## Describe your changes
 - Merge Kubernetes, JsOrc to MetaService
 - MetaService will now act as JsOrc
 - Automation of create/delete/patch for kube configs
 
## Link to related issue
N/A

No

No

JsOrc


jaseci_core/jaseci/api/config_api.py
```diff
@@ -47,25 +52,61 @@ def config_set(self, name: str, value: str or dict, do_check: bool = True):
         self._h.save_glob(name, value)
         return [f"Config of '{name}' to '{value}' set!"]
 
+    @Interface.admin_api()
+    def config_yaml(self, name: str, file: list):
```
ypkang: If I am understanding this correctly, this is basically you can use yaml file to update the config. What is the problem you are trying to address here vs. the existing config set API?

Also, for the `file` parameter, does the yaml file need to be already on the filesystem of the jaseci instance?

jaseci_core/jaseci/api/config_api.py
```diff
@@ -47,25 +52,61 @@ def config_set(self, name: str, value: str or dict, do_check: bool = True):
...
+                new_config[kind] = []
+            new_config[kind].append(conf)
+
+        old_config = self._h.get_glob(name)
+        if old_config:
+            old_config = loads(old_config)
+            old_config.pop("__OLD_CONFIG__", None)
+            for kind, confs in old_config.items():
+
+                print(kind)
```
ypkang: Remove debugging prints.

jaseci_core/jaseci/api/config_api.py
ypkang: I think what we really need is to make the config set flexible so it can just update selected fields, as opposed to having to specify the entire JSON every time.

jaseci_core/jaseci/api/config_api.py
```diff
@@ -47,25 +52,61 @@ def config_set(self, name: str, value: str or dict, do_check: bool = True):
...
+
+        to_start = not hook.meta.is_automated()
+
+        service = getattr(hook, name, None)
+
+        if isinstance(service, CommonService):
+            service.reset(hook, to_start)
+            return
+
+        return "Not a valid operation!"
```
ypkang: I think we need to return a JSON for APIs always so let's make these two returns all JSON and following our convention of
```
{
  "success": true/false,
}
```
And for the line 109, i'd add the message to the response and I'd make it more accurate to something like
```
f"{name} is not a valid service. Can not refresh config."
```

jaseci_core/jaseci/svc/common.py
```diff
@@ -125,12 +137,253 @@ def __init__(self):
...
+        }
+        self.delete_apis = {
+            "Service": self.core.delete_namespaced_service,
+            "Deployment": self.api.delete_namespaced_deployment,
+            "ConfigMap": self.core.delete_namespaced_config_map,
+            "ServiceAccount": self.core.delete_namespaced_service_account,
+            "ClusterRole": self.auth.delete_cluster_role,
+            "ClusterRoleBinding": self.auth.delete_cluster_role_binding,
+            "Secret": self.core.delete_namespaced_secret,
+            "PersistentVolumeClaim": (
```
ypkang, line 240: I think we should be thoughtful on the delete behavior of JSORC. Some delete can be destructive and undesirable. I added deletes in my PR too and I didn't include PVC. We should discuss this with @marsninja to get his thoughts.

jaseci_core/jaseci/svc/common.py
```diff
@@ -125,12 +137,253 @@ def __init__(self):
...
+                        and type(old_config_map) is dict
+                        and kind in old_config_map
+                        and name in old_config_map[kind]
+                    ):
+                        old_config_map.get(kind, []).remove(name)
+
+                    for to_be_removed in old_config_map.get(kind, []):
+                        res = self.read(kind, to_be_removed, namespace)
+                        if not isinstance(res, ApiException) and res.metadata:
+                            print(type(res))
```
ypkang: remove debug prints.

jaseci_core/jaseci/svc/common.py
```diff
@@ -209,60 +456,20 @@ def state(self, val: Ss):
...
-        return self.cls._super_master
+    def quiet(self) -> bool:
+        return self.cls._quiet
 
-    @super_master.setter
-    def super_master(self, val):
-        self.cls._super_master = val
+    @quiet.setter
+    def quiet(self, val: bool):
+        pass
```
ypkang, line 492: Is this correct?

jaseci_core/jaseci/svc/mail/config.py
```diff
@@ -1,7 +1,9 @@
+import sys
+
 MAIL_ERR_MSG = "Mail service is disabled or not yet configured!"
 
 MAIL_CONFIG = {
-    "enabled": True,
+    "enabled": False,
```
ypkang, line 4: Is there a reason the mail service shouldn't be enabled by default?

jaseci_core/jaseci/svc/mail/config.py
```diff
@@ -25,3 +27,7 @@
         "resetpass_html_body": "Your Jaseci password reset" "token is: {{token}}",
     },
 }
+
+
+if "python3 -m unittest" in sys.argv:
+    MAIL_CONFIG["enabled"] = True
```
ypkang: Hmmm I am not sure I am a fun of this kind of checking. @marsninja 's call.

jaseci_serv/jaseci_serv/svc/mail/config.py
```diff
@@ -0,0 +1,31 @@
...
+        "{{link}}",
+        "resetpass_subj": "Password Reset for Jaseci Account",
+        "resetpass_body": "Your Jaseci password reset token is: {{token}}",
+        "resetpass_html_body": "Your Jaseci password reset" "token is: {{token}}",
+    },
+    "migrate": False,
+}
+
+if "test" in sys.argv or "test_coverage" in sys.argv:
+    MAIL_CONFIG["backend"] = "locmem"
```
ypkang: Same here. Not sure if it is a good practice to have separate code path for production vs. testing.

scripts/defaults/jaseci.yaml
```diff
@@ -101,7 +101,7 @@ metadata:
 data:
   prod_up: |-
     apt update; apt -y upgrade; apt -y install --no-install-recommends git g++;
-    git clone -b feature/jsorc https://github.com/amadolid/jaseci.git
+    git clone -b refactor/jsorc https://github.com/amadolid/jaseci.git
```
ypkang: We should remember to change this to point to the main jaseci repo right before we merge..

scripts/defaults/prometheus.yaml
```diff
@@ -894,7 +894,7 @@ metadata:
 spec:
   ports:
     - name: http
-      port: 80
+      port: 9090
```
ypkang, line 897: Is this intentional?

ypkang (changes requested):
@amadolid Good PR overall. I left some comments and questions in the PR. Take a look.

The main thing though is that I don't think I fully understand your re-architecture of the services and configs. So let's definitely discuss this over a meeting.

jaseci_core/jaseci/svc/common.py
amadolid: Yes sir, as default for CommonServices `quiet` field can be changed. However, in MetaService, `quiet` should not be updatable and value should always be false

jaseci_core/jaseci/svc/mail/config.py
amadolid: MAIL_CONFIG doesn't have any default credentials. If they have credentials they can just enable it on update

Jaseci instance with existing MAIL_CONFIG will not be affected

scripts/defaults/jaseci.yaml
amadolid: yup, will add it on this upcoming pr
https://github.com/amadolid/jaseci/pull/1

scripts/defaults/prometheus.yaml
amadolid: yes sir, since the deployment uses 9090 instead of 80

amadolid: yes sir, since the deployment uses 9090 instead of 80

jaseci_core/jaseci/svc/mail/config.py
ypkang: @amadolid We talked about this -- instead of having this check in the code, we can just define a different config json in the unit test right?

jaseci_serv/jaseci_serv/svc/mail/config.py
ypkang: Same here @amadolid we are removing this right?

ypkang (changes requested):

ypkang:
@amadolid I am trying to test the new JSORC but I can not get META_CONFIG to work. Let's hop on a call and maybe you can do quick demo.

jaseci_core/jaseci/svc/common.py
ypkang: @amadolid As we discussed, let's remove the ability to delete PVC from JSORC.

jaseci_core/jaseci/svc/common.py
```diff
@@ -125,12 +134,252 @@ def __init__(self):
...
+                        names = old_config_map.get(kind, [])
+                        if name in names:
+                            names.remove(name)
+
+                        if kind == "Service":
+                            pod_name = name
+                        res = self.read(kind, name, namespace)
+                        if hasattr(res, "status") and res.status == 404 and conf:
+                            self.create(kind, name, namespace, conf)
+                        elif res.metadata:
```
ypkang: There is a corner case. If the `kube.read` fails in other ways than `404` not found, it will execute line 348 and it will fail because `res` will be an `APIException`

jaseci_core/jaseci/svc/meta.py
```diff
@@ -23,10 +24,26 @@ def __init__(self, run_svcs=True):
...
+        if self.run_svcs:
+            self.app.build()
+            if self.is_automated():
+                logger.info("JsOrc is automated. Pushing interval check alarm...")
+                self.push_interval(1)
+            else:
+                logger.info("JsOrc is not automated.")
+
+    def push_interval(self, interval):
+        if self.running_interval == 0:
```
ypkang, line 41: I am still not sure I understand the purpose this `running_interval` serves. Let's discuss.

jaseci_serv/jaseci_serv/svc/config.py
```diff
@@ -0,0 +1,7 @@
+import sys
+
+KUBERNETES_CONFIG = {"in_cluster": False, "config": None}
+
+META_CONFIG = {"automation": False, "monitoring": True, "kubernetes": KUBERNETES_CONFIG}
```
ypkang: What is the relationship between the config in jaseci_serv and the config in jaseci_cor? Can we avoid having two very similar `config.py` and `META_CONFIG` in the repo?

jaseci_serv/jaseci_serv/svc/prometheus/config.py
```diff
@@ -0,0 +1,7 @@
+import os
+
+PROMON_CONFIG = {
```
ypkang, line 3: @amadolid Let's brainstorm around this -- I think we should build a set of APIs under jsorc that turn on/off and configure each of the services. I think this config based API interface is not super intuitive and kinda hard to use.

scripts/defaults/jaseci.yaml
```diff
@@ -61,11 +61,11 @@ rules:
       "validatingwebhookconfigurations",
       "volumeattachments"
     ]
-    verbs: ["get", "watch", "list", "update", "patch", "create"]
+    verbs: ["get", "watch", "list", "update", "patch", "create", "delete"]
```
ypkang, line 64: Oh....what is the difference between this yaml and the one in `scripts/jaseci.yaml`?

scripts/jaseci.yaml
```diff
@@ -11,9 +11,61 @@ metadata:
   namespace: default
   name: jaseci-orc
 rules:
-  - apiGroups: ["v1", "apps"]
-    resources: ["pods", "deployments", "services", "deployments.apps"]
-    verbs: ["get", "watch", "list", "update", "patch", "create"]
+  - apiGroups: [
```
ypkang: @amadolid You can ignore the changes in this file for now...what happened was I was using this yaml (not knowing there was another one in `scripts/default`) and none of kubernetes operations were working so I added the following apiGroups and resources definition but i think the one in default is the one we should all use now?

ypkang (changes requested):

jaseci_serv/jaseci_serv/svc/config.py
amadolid: Removed

jaseci_core/jaseci/api/jsorc_api.py
```diff
@@ -34,6 +38,69 @@ def load_yaml(self, files: list, namespace: str = "default"):
 
         return res
 
+    @Interface.admin_api(cli_args=["name"])
+    def config_yaml(self, name: str, file: list, unsafe_paraphrase: str = ""):
```
ypkang: Change this to apply_yaml

jaseci_core/jaseci/api/jsorc_api.py
```diff
@@ -34,6 +38,69 @@ def load_yaml(self, files: list, namespace: str = "default"):
...
+            old_config.pop("__OLD_CONFIG__", None)
+            for kind, confs in old_config.items():
+                names = []
+                for conf in confs:
+                    names.append(conf["metadata"]["name"])
+                old_config[kind] = names
+
+            new_config["__OLD_CONFIG__"] = old_config
+
+            if unsafe_paraphrase == UNSAFE_PARAPHRASE:
```
ypkang, line 76: document how this unsafe_paraphrase should be used.

jaseci_serv/jaseci_serv/svc/mail/config.py
ypkang: remove this.

scripts/jaseci.yaml
ypkang: Yiping to rever the changes in this.

scripts/defaults/jaseci.yaml
```diff
@@ -61,11 +61,11 @@ rules:
       "validatingwebhookconfigurations",
```
ypkang, line 61: change the folder from scripts/default to scripts/jsorc

ypkang (changes requested):

ypkang (approved):

PR #626, opened 2022-11-24
Bart Summarization

chandralegend:
Used facebook's Bart-large-CNN for better summarization. the module provides only one endpoint where you can provide a text, a list of texts, or a URL to summarize.


ypkang (approved):
LGTM. Good work @chandralegend

PR #634, opened 2022-11-28
Jaseci Docker Images and Github Action Automation

ypkang:


marsninja (changes requested):
Is this merge ready @ypkang ? Also, we should write a little doc about how this works, would fit in the "jaseci internals" section or "contributors" or can be in a regular section.

ypkang:
> Is this merge ready @ypkang ? Also, we should write a little doc about how this works, would fit in the "jaseci internals" section or "contributors" or can be in a regular section.

Yeah I can write a little bit about the process and put in the doc hierarchy and then we can merge. This is merge ready.

ypkang:
Actually, there is one more thing I want to add to this PR (creating a dependency between the various github actions jobs we have ). Then it will be merge ready.

ypkang:
@marsninja this is ready for merge. you can squash and merge this one. I had to frequently commit to test out the github actions so there are many commits on this branch.

jaseci_ai_kit/setup.py
```diff
@@ -20,7 +20,7 @@ def get_ver():
         "torch>=1.10.2,<2.0.0",
         "pandas>=1.4.1,<2.0.0",
         "flair==0.10",
-        "numpy>=1.22.1,<2.0.0",
+        "numpy==1.23.5",
```
marsninja: Should we pin these so specifically @ypkang , Increases the likeliness of package conflicts, i like something like >=1.23.0,<1.24 a little better

marsninja (changes requested):
see inline comment

jaseci_ai_kit/setup.py
ypkang: yea we can change it to that. I just needed to upgrade numpy because there is a problem with 1.22.* that was causing the docker build to fail from time to time.

ypkang: updated.

marsninja (approved):
LGTM @ypkang !

Issue #645, opened 2022-12-03
Avoid potential OOM issue with logging api


marsninja:
Current logging api to get logs uses a string buffer that continuously grows. We should create a limited sliding buffer. This link has a partial solution https://bytes.com/topic/python/answers/169145-feature-request-max-size-limit-stringio

I'd replace `raise IOError("allocated buffer space exceeded")` with logic to chop of the the difference between `cursize` and `maxsize` and then reset `cursize` to `maxsize`, then `write of course`

ypkang:
@Gim3l is this resolved or still needs to be done?

PR #648, opened 2022-12-05
replace StringIO buffer with limited sliding buffer for logging

Gim3l:
## Describe your changes
Addresses #645


marsninja (approved):
This looks good, I'll play with it and make any needed changes @Gim3l

Issue #583, opened 2022-11-06
Would be great to have some zlib functions in the standard library


marsninja:
**Is your feature request related to a problem? Please describe.**
It would be very useful if Jaseci had some gzip compatible standard library functions. 

**Describe the solution you'd like**
It would be helpful if there were some actions like `zlib.compress` and `zlib.decompress` that can be used in the Jac programming language. To adhere to Jac's principles these functions should take in base64 strings as input and produce base64 as output. I'd think a good design would be to convert the base64 input into bytes, perform compression, then encode the compressed bytes into base64 again for the return. 

**If this feature request is accepted, would you be willing to work on a PR?**
If time allows.

**Additional context**
I'd imagine this library would piggyback on pythons zlib functions.

apatel66:
Hi, Matthew Wang (@Hextechagon) and I are juniors at the University of Michigan taking a software development class. Our latest assignment is to contribute to an open source project. We would like to contribute to Jaseci so would you be able to assign this issue to Matthew and I? Thanks!

marsninja:
Sure thing!

marsninja:
@Hextechagon reply to this thread if you can. I'm trying to get github to let me assign you but I suspect you have to comment first.

Hextechagon:
Hi, sure thing!

Hextechagon:
Hi @marsninja , can you assign this issue to me too? Thanks!

ypkang:
@Hextechagon assigned it to you. How is this coming along? Feel free to reach out to me or @marsninja if you need anything.

Hextechagon:
Thanks, we're making progress.

PR #649, opened 2022-12-05
Add zlib functionalities

apatel66:
Added some zlib functions to the standard library (compress and decompress). Created associated tests as well.
Closes: https://github.com/Jaseci-Labs/jaseci/issues/583.


marsninja (changes requested):
Nice work @apatel66 , On change needed, the jac/jac_parse files changed is likely due to not running `black` before commiting, by simply typing `black .` it should reformat properly then you can push.

apatel66:
> Nice work @apatel66 , On change needed, the jac/jac_parse files changed is likely due to not running `black` before commiting, by simply typing `black .` it should reformat properly then you can push.

Yep, that seemed to have worked. I only had the black formatter run whenever I saved a file in vscode, so I also added it to my terminal.

marsninja (approved):
LGTM! Thanks @apatel66 !

PR #675, opened 2022-12-14
Multilingual Translation

chandralegend:
* Module that supports multilingual translation


ypkang (approved):
Tested and this works well.

PR #682, opened 2022-12-14
add workflow file for building jaseci studio

Gim3l:

### DIFF
--- a/.github/workflows/build-and-release-studio.yml
+++ b/.github/workflows/build-and-release-studio.yml
@@ -13,8 +13,6 @@ jobs:
     outputs:
       release_id: ${{ steps.create-release.outputs.result }}
 
-    
-    
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


ypkang:
@Gim3l This mostly looks good. There are two things I would like changed if possible.
1. @marsninja and I discussed our tagging strategy and decided to stick with the current process where all packages are synced with the same tag, which includes jaseci-studio. So we will need this github action to use the tag of the repo as the version for the studio binary. Take a look at the github action config for docker images here for how to do that https://github.com/Jaseci-Labs/jaseci/blob/main/.github/workflows/build-and-release.yml#L118-L120
2. The release name. So basically I am imagine this will be a jaseci-wide release which will contain a binary for jaseci studio. So we will need to name the release Jaseci release.

ypkang (changes requested):
see the comment above.

ypkang (approved):

PR #712, opened 2022-12-28
Added switchable male/female TTS

TharukaCkasthuri:


TharukaCkasthuri:
The Jaseci_ai_kit required numpy 1.23, but TTS required numpy==1.21.6; python_version<"3.10"
numpy==1.22.4;python_version=="3.10". So I updated the test kit python version to 3.10.

PR #716, opened 2023-01-04
[MINOR]: Allow to save the actual request body before merging with request url mapping

amadolid:
## Describe your changes
- This is to avoid conflict when django url variable has the same name with the request body. This will save the request body first before merging with django url variable

## Link to related issue
N/A

No

No

Request context


marsninja (approved):
Thanks @amadolid

PR #718, opened 2023-01-05
Prometheus node exporter fix and new Jaseci APIs for prometheus

ypkang:
* Added a set of promethues APIs
* Fix promethues node exporter manifest (node metrics used to not work for cluster in WSL environment)


marsninja (approved):
LGTM!

PR #754, opened 2023-01-11
added tts module with voice conversion support

TharukaCkasthuri:


TharukaCkasthuri:
added the feature in new pr

PR #757, opened 2023-01-11
[ELASTIC-SERVICE]: Initial

amadolid:


marsninja (approved):
Merging in so i might try to get main passing @amadolid , however we should add tests and such in a separate pr.

PR #759, opened 2023-01-12
[MINOR]: Include django fields on master_self detailed true

amadolid:


marsninja (changes requested):
Can you add a tiny test in jsserv for this @amadolid ? Should be super quick.

PR #760, opened 2023-01-12
add github workflow for running tests for jaseci studio

Gim3l:
Closes: #750

### DIFF
--- a/.github/workflows/jaseci-studio-test.yml
+++ b/.github/workflows/jaseci-studio-test.yml
@@ -1,0 +1,26 @@
+name: Jaseci Studio Tests
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: prepare stencil ui components
+        run: yarn setup:ui
+        working-directory: "./jaseci_studio"
+      - name: install dependencies
+        run: yarn install && npx playwright install --with-deps
+        working-directory: "./jaseci_studio"
+      - name: run unit test
+        run: yarn test:unit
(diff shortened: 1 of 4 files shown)


jaseci_studio/__e2e__/logs.spec.ts
```diff
@@ -7,12 +7,12 @@ test("logs page loads", async ({ page }) => {
 
 test("logs messages load", async ({ page }) => {
   await page.addInitScript(() => {
-    window.localStorage.setItem("serverUrl", "http://mysite.com");
+    window.localStorage.setItem("serverUrl", "http://localhost:8005");
```
ypkang, line 10: @Gim3l what do these URL do? Would localhost here work with tests and would it work with production?

ypkang (changes requested):

.github/workflows/jaseci-studio-test.yml
```diff
@@ -0,0 +1,26 @@
+name: Jaseci Studio Tests
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
```
ypkang, line 9: @Gim3l Could you update this to only trigger the tests if there are changes in the `jaseci_studio` folder? You can take a look at the jaseci_ai_kit github action workflow file for how to do that.

jaseci_studio/__e2e__/logs.spec.ts
Gim3l: Those are mock urls used for e2e testing. Yes, the localhost version works with the tests on github, the other link doesn't.

.github/workflows/jaseci-studio-test.yml
Gim3l: Ok, I made the change.

ypkang (approved):

PR #765, opened 2023-01-13
Refactoring AI Kit

chandralegend:


ypkang:
* Does the `test.sh` at the top level of jaseci_ai_kit directory need to be updated?
* When you update the readme, we should organize it hierarchically. So top level readme of jaseci ai kit should just be a table of content into each jac_* and then there should be a readme for each jac_* that describe each module that is similar to what we have today
* Let's add to the top level readme a new section about contributing to jaseci ai kit. Where they should put their module etc.
* Let's change `dump` to be `_inactive`.
*

ypkang (approved):
LGTM. Great work @chandralegend

PR #853, opened 2023-01-26
[JSORC]: Version 3 - Application Context + Inversion of Control + Service Lazy Loading

amadolid:
## **`JsOrc`**
- Owns everything
    - Services
    - Repository (hook / etc)
    - Context (super/master/etc)
- Always running with regeneration (fail check) - db_regen will be included once PR [#862](https://github.com/Jaseci-Labs/jaseci/pull/862) is merged
- can preload services
- All Objects produced by JsOrc is injectable or can be access directly
    - Injection: `@JsOrc.inject(services=[], repository=[], context=[])`
    - Direct Access
        - Service: `JsOrc.svc("serviceName ex: task")`
        - Repository `JsOrc.src("repositoryName ex: hook")`
        - Context
            - New `JsOrc.ctx("contextName ex: master")`
            - Persisting `JsOrc.get("contextName ex: master")`
            - Renew `JsOrc.renew("contextName ex: master")`
            
## **`Services`**
- Lazy Load -  will only initialize if needed unless it was automated
- Automated - will regen (by JsOrc) if not running
- No need to attach on other object like previous approach hook.task / hook.redis / etc
- Simple class tagging via Decorator - `@JsOrc.service(...)`

## **`Repository (hook)`**
- Simple class tagging via Decorator - `@JsOrc.repository(...)`

## **`Context (super_master / master / action_manager)`**
- Simple class tagging via Decorator - `@JsOrc.context(...)`


amadolid:
Documentations to be follow

@ypkang, would like to ask for your review as I have adjusted the ActionOptimizer too

jaseci_core/jaseci/__init__.py
```diff
@@ -9,6 +11,7 @@ def get_ver():
 __version__ = get_ver()
 __creator__ = "Jason Mars and friends"
 __url__ = "https://jaseci.org"
+__all__ = ["JsOrc", "JsOrcSettings"]
```
ypkang: Is this so that we can import JSORC directly from the jaseci package at the top level?

jaseci_core/jaseci/actions/standard/tests/test_mail_lib.py
```diff
@@ -11,22 +11,23 @@ class MailLibTest(CoreTest):
...
+        JsOrc.settings("MAIL_CONFIG")["enabled"] = True
         MailService.connect = MagicMock(return_value=Mock())
         super(MailLibTest, self).__init__(*args, **kwargs)
 
     def setUp(self):
-        super().setUp(True)
+        super().setUp()
 
-    def test_send_mail(self):
+    @JsOrc.inject(services=["mail"])
```
ypkang, line 21: What does inject do?

jaseci_core/jaseci/api/jsorc_api.py
```diff
@@ -18,27 +22,28 @@ class JsOrcApi:
     """
 
     @Interface.admin_api()
-    def load_yaml(self, files: list, namespace: str = "default"):
+    def load_yaml(self, files: list, namespace: str = None):
```
ypkang, line 25: Any reason we don't want namespace to be default to `default`?

jaseci_core/jaseci/api/jsorc_api.py
```diff
@@ -148,25 +135,19 @@ def jsorc_actions_load(self, name: str, mode: str):
...
         """
         Get the current status of an action
         """
-        hook = self._h
-        if hook.meta.run_svcs:
-            status = hook.meta.app.get_actions_status(name)
-            return {"success": True, "action_status": status}
-        else:
-            return {"success": False, "message": "No running JSORC service."}
+        status = JsOrc.get("action_manager", ActionManager).get_actions_status(name)
```
ypkang, line 149: What happens if JSORC is not enabled? We shouldn't return success: True if that's the case right?

jaseci_core/jaseci/jsorc.py
```diff
@@ -0,0 +1,527 @@
...
+    def service(
+        cls,
+        name: str = None,
+        config: str = None,
+        manifest: str = None,
+        priority: int = 0,
+        proxy: bool = False,
+    ):
+        """
+        Save the class in services options
```
ypkang, line 461: I would like to document these concepts better. At this point, I can;'t really understand what is a service, repository or context in JSORC. @amadolid lets chat.

jaseci_core/jaseci/jsorc_settings.py
```diff
@@ -0,0 +1,9500 @@
+import os
```
ypkang, line 1: I am not a big fan of this file. It is massive and going to be hard to update and deal with down the line. How about we separate each config into is JSON file and then load them in here?

jaseci_core/jaseci/svc/kube_svc.py
```diff
@@ -0,0 +1,230 @@
...
+    ApiextensionsV1Api,
+    AdmissionregistrationV1Api,
+)
+from kubernetes.client.rest import ApiException
+
+from jaseci.utils.utils import logger
+
+
+@JsOrc.service(name="kube", config="KUBE_CONFIG")
+class KubeService(CommonService):
```
ypkang: I don't know if kubernetes should be a service. Let's discuss.

ypkang (changes requested):

jaseci_core/jaseci/__init__.py
amadolid: Yup and also JsOrc class would be one of the first classes to be loaded

jaseci_core/jaseci/actions/standard/tests/test_mail_lib.py
amadolid: upon calling that method mail (service) will be injected as keyword argument

jaseci_core/jaseci/api/jsorc_api.py
amadolid: kubeService now holds the namespace (defaults to "default").
If admin wants to override it, that's when they need to include namespace on load_yaml api

jaseci_core/jaseci/api/jsorc_api.py
amadolid: JsOrc now cannot be disabled while action_manager is a context type which should return as demand (.get will reuse same instance). However, if the method call in action_manager has service connection like the kubeService it should raise exception and let jac view handle the exception

will add some adjustment on action_optimizer.kube_create/kube_delete

jaseci_core/jaseci/jsorc.py
amadolid: Yup yup, currently working on the documentation

jaseci_core/jaseci/jsorc_settings.py
amadolid: Yup, i'm planning to have a yaml parser as default for manifest config
JsOrcSettings:
 ELASTIC_MANIFEST = load_yaml("elastic.yaml") // should return the same

jaseci_core/jaseci/svc/kube_svc.py
amadolid: The reason for this is to allow user change kubernetes connection configs (also ping test) on runtime

jaseci_core/jaseci/jsorc_settings.py
amadolid: Added the new load yaml from file

jaseci_core/jaseci/actions/standard/tests/test_mail_lib.py
ypkang: Let's remove inject.

ypkang (approved):
f

PR #862, opened 2023-01-26
[JSORC]: DB Regen

### LINKED ISSUE #833: Bring DB under JSORC's automation
Similar to how JSORC is automatically keeping redis alive right now.

amadolid:
Closes #833

### DIFF
--- a/jaseci_core/jaseci/svc/common.py
+++ b/jaseci_core/jaseci/svc/common.py
@@ -316,10 +316,15 @@ class JsOrc:
 
         self.db_check()
         if self.has_db:
-            for item in self.kubernetes.core.list_namespaced_pod(
-                namespace=self.namespace, label_selector=f"pod={DATABASE['pod']}"
-            ).items:
-                self.kubernetes.core.delete_namespaced_pod(item["name"], self.namespace)
+            try:
+                for item in self.kubernetes.core.list_namespaced_pod(
+                    namespace=self.namespace, label_selector=f"pod={DATABASE['pod']}"
+                ).items:
+                    self.kubernetes.core.delete_namespaced_pod(
+                        item["name"], self.namespace
+                    )
+            except Exception:
+                raise Exception("Force termination to restart the pod!")
 
     ###################################################
     #                     BUILDER                     #
--- a/jaseci_core/jaseci/svc/common.py
+++ b/jaseci_core/jaseci/svc/common.py
@@ -9,10 +9,12 @@ from jaseci.utils.utils import logger
 from jaseci.actions.live_actions import load_action_config
 from jaseci.svc.actions_optimizer.actions_optimizer import ActionsOptimizer
 from .state import ServiceState as Ss
-from .config import META_CONFIG, KUBERNETES_CONFIG
+from .config import META_CONFIG, KUBERNETES_CONFIG, DATABASE as db, POSTGRES_MANIFEST
 
 import time
 import numpy as np
+
+import psycopg2
 
 ###################################################
 #                  UNSAFE PARAMS                  #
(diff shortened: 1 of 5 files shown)


jaseci_core/jaseci/svc/common.py
```diff
@@ -9,11 +9,13 @@
 from jaseci.actions.live_actions import load_action_config
 from jaseci.svc.actions_optimizer.actions_optimizer import ActionsOptimizer
 from .state import ServiceState as Ss
-from .config import META_CONFIG, KUBERNETES_CONFIG
+from .config import META_CONFIG, KUBERNETES_CONFIG, DATABASE as db, POSTGRES_MANIFEST
```
ypkang: Any particular reason you are renaming DATABASE as db here? It is not a big deal but i feel like it doesn't help with readability.

jaseci_core/jaseci/svc/common.py
```diff
@@ -288,19 +290,53 @@ def __init__(self, meta):
...
 
     def build(self):
         try:
-            hook = self.build_context("hook")
-            config = hook.service_glob("META_CONFIG", META_CONFIG)
+            config = META_CONFIG
+            if self.has_db:
+                hook = self.build_context("hook")
+                config = hook.service_glob("META_CONFIG", META_CONFIG)
+                self.prometheus = self.meta.get_service("promon", hook)
```
ypkang, line 339: Why is prometheus hoisted up here?

jaseci_core/jaseci/svc/config.py
```diff
@@ -1,9 +1,138 @@
...
+USER = os.environ.get("POSTGRES_USER", "postgres")
+
+DATABASE = {
+    "enabled": os.environ.get("JSORC_DB_REGEN") == "true",
+    "pod": os.environ.get("JSORC_POD_NAME", "jaseci"),
+    "host": os.environ.get("POSTGRES_HOST", "jaseci-db"),
+    "db": os.environ.get("POSTGRES_DB", "postgres"),
+    "name": USER,
+    "user": USER,
+    "password": os.environ.get("POSTGRES_PASSWORD", "lifelogifyjaseci"),
```
ypkang: Let's not have the username/pw as plain text here. Let's just not have default value.

ypkang: Same with the USER field above.

scripts/jsorc/jaseci.yaml
```diff
@@ -151,12 +151,24 @@ spec:
             - name: POSTGRES_HOST
               value: jaseci-db
 
+            - name: POSTGRES_DB
+              value: postgres
```
ypkang, line 155: Let's call this DBNAME.

ypkang (changes requested):
* I left some comments inline.
* I tested this and it works fine.
* The main thing I'd like to change is the first time jaseci comes up, there are many errors in the logs. I think they are all related to can't access DB which is expected but it looks like something is going wrong when in reality, everything is working as expected. Can we suppress those errors from the logs?

jaseci_core/jaseci/svc/common.py
amadolid: Prometheus is part of JsOrc and pointed to JsOrc's Kubernetes so I just initialized it when JsOrc is automated

ypkang (changes requested):
@amadolid did you see this comment?

The main thing I'd like to change is the first time jaseci comes up, there are many errors in the logs. I think they are all related to can't access DB which is expected but it looks like something is going wrong when in reality, everything is working as expected. Can we suppress those errors from the logs?

amadolid:
> @amadolid did you see this comment?
> 
> The main thing I'd like to change is the first time jaseci comes up, there are many errors in the logs. I think they are all related to can't access DB which is expected but it looks like something is going wrong when in reality, everything is working as expected. Can we suppress those errors from the logs?

ow sorry, I've overlooked this part.
Will post update for this

amadolid:
I have updated some approach so lessen the exceptions. On django related logs, I'm not sure yet if can suppress that without overriding django implementation. I think logging for db connection issue would be needed too as it would be the tracker on current jaseci build up

ypkang (approved):

PR #866, opened 2023-01-28
update multi-turn

andrewye16:
This DOT graph is incorrect and outdated.
1. The current code will have an edge connect to itself (while this one does not)
2. This graph assumes user have created the entity transition, which is only covered in the next part. Hence it will lead to confusion while the edge does not exist.


marsninja (approved):
Looks good to me! Thanks! @andrewye16

PR #874, opened 2023-01-31
docs: update pip commands in Section 2

zwgold:
This PR is a docs fix addressing the pip commands in Section 2. When installing modules and wanting to specify optional dependencies to install, you need the module and brackets surrounded in quotations. This should fix the commands so students in the future are not confused.


ypkang (approved):
LGTM.

Just FYI, the quotation marks are only required for certain shell (e.g. zsh). For bash, the original commands also work, as well as the ones with quotation marks so this is a more general command.

PR #891, opened 2023-02-06
JacTalk inital commit

AshishMahendra:


jaseci_core/jaseci/jsctl/jac_talk_utils.py
```diff
@@ -0,0 +1,130 @@
...
+
+def chat_setup():
+    try:
+        # base bot directory
+        bot_dir = curr_path / "sample_bot"
+        # Create temporary dir
+        base_path = tempfile.mkdtemp()
+
+        # Clone into temporary dir
+        git.Repo.clone_from(
```
ypkang, line 75: Do we have to download the git repository and get the templates there? The JacTalk binary already have access to all the jac files. Maybe it can just load it from file and then dump it out, instead of copying the files.

ypkang (changes requested):

ypkang:
@AshishMahendra take a look at the inline comment.

ypkang:
Not a priority at the moment. Closing for now.

PR #898, opened 2023-02-08
Move studio docs

Gim3l:
Closes: #896


marsninja (approved):
This is ok progress @Gim3l But i dont think this merge will have your studio work integrated into CCAI in official docs. You have to move content into examples/ <CCAI> instead of docs/<CCAI>

PR #909, opened 2023-02-10
specified_tutorial_location

jzhou1402:
When following the Chapter 7* tutorial, users are instructed to add files but not specified where to add them. If users add files in a directory other than examples/CanoniCAI/code, they would run into FileNotFound errors in section 6.3. A lot of this could be avoided by simply letting the users know where the files are in this complicated repository in the first place. I personally spent no less than an hour trying to figure out why my code wasn't finding bi_enc.


marsninja (approved):
LGTM @jzhou1402 Thanks!!

PR #912, opened 2023-02-11
Update commands in section 7.5

YupitoUM:
When the user is working on the tutorial, there is no "main.jac", it should be "tesla_ai.jac" and "tesla_ai.jir"


marsninja (approved):
LGTM Thanks @YupitoUM

PR #916, opened 2023-02-13
Fixing dependency issue in vc_tts

TharukaCkasthuri:


jaseci_ai_kit/install.sh
```diff
@@ -1,15 +1,15 @@
...
+JAC_SPEECH_MODULES=("stt" "vc_tts")
 JAC_MISC_MODULES=("pdf_ext" "translator" "cluster" "ph")
 JAC_VISION_MODULES=("detr" "rftm" "yolos" "dpt")
 
 install_modules() {
 	if [[ $1 == "all" ]]; then
 		echo -e "\033[0;32mInstalling all the modules\033[0m"
 		cd jac_speech
-		pip install -e .[all]
+		pip install .[all]
```
ypkang, line 15: Is there a reason we don't want to install live for just jac_speech here?

ypkang (changes requested):
One comment inline. Also, remember to add some instructions on how to solve numba version conflict issues when encountered in the jaseci_ai_kit top level README.

jaseci_ai_kit/install.sh
TharukaCkasthuri: When we install with -e sign the packages in the ORDEREDREQS list in setup file doesn't get installed.

ypkang: I see. In that case, I have removed `-e` for the other modules as well to keep the behavior of this script consistent.

ypkang (approved):

PR #919, opened 2023-02-13
Prevent access to views that require superuser privilege

Gim3l:
Closes: #918


ypkang:
@Gim3l could you resolve the conflicts here thanks

ypkang (approved):

PR #947, opened 2023-02-20
[JSORC]: Update namespace handling (prefixing); Elastic manifest update

amadolid:
TBD


ypkang:
I think we should have all the relevant jaseci services in the same namespaces, as opposed to each service in its own namespace
. so instead of redis is in the jaseci1-redis name space and db is in the jaseci1-db namespace, they should both be in the same jaseci1 namespace. And when dedidcated is false, we should just use the default namespace.

i did some googling on namespaces best practices and organizing namespace by project is one of the common standard, which applies here pretty well

also I think elastic should default to dedicated=True because it makes sense each application has its own logs

and the namespaces for JSORC-controlled resources should just inherit from the namespace it is currently in.

so a user workflow is something like this

1. kubectl apply -f jaseci.yaml -n zsb_production

2 . this launches jaseci in the zsb_production namespace, and then JSORC launches everything else in the same namespace (redis, DB, elastic)


the yaml file itself would not have any namespace specified

ypkang:
okay there are 3 different piece of feedback i was talking about

 1. let's not have each pod to be in its own namespace. we should group the pods based on each jaseci instance. For example, I want to launch a jaseci stack for Myca, everything should be in the myca  namespace but with the same name as currently (jaseci, jaseci-db, jaseci-redis etc.).

2. when JSORC launches new k8s resources (pods, deployments etc),. it should just use whatever namespace it is currently in. Therefore, we shouldn't need to actually have the namespace configuration in neither code nor the DB. It shouldn't even need to be in jaseci.yaml and the user will specify it when launching it via the -n  argument

3. The last thing is related to this dedicated flag. I understand the reason for this dedicated flag but its behavior will need to change quite a bit in this new design I described in 1 and 2. So for this first version, it is okay if we just leave that out for now. And launch all services in namespaces.

ypkang (approved):
I have tested that the JSORC is able to bring up elastic and with custom namespaces organizations. 🚢

PR #955, opened 2023-02-23
added jac built in feature to get unique items from the list

TharukaCkasthuri:


marsninja (approved):
Very nice addition! Thanks @TharukaCkasthuri

PR #965, opened 2023-02-24
Jaseci Examples Tested and More content from the jaseci Bible was added to Language Basics

RolexAlexander:
Good Morning Dr Mars and Team, I added more examples and content from jaseci Bible to the Language Basics Section. I also tested these examples to ensure that they work.


.gitignore
```diff
@@ -34,8 +34,3 @@ node_modules
 
 # Docs
 mdbook/
-
-# Docs Pathe
-/support/guide/other/cheatsheet.md
-/support/guide/other/classes.md
-/support/guide/other/stdlib.md
```
marsninja, line 41: We have this in here so that this autogenerated content doesn't get committed in a way that continually expands the size of the repo. Basically it should be generated by the gen_mdbooks.py script. I realize the readme links in github wont work, but that ok for now as long as the links work on the docs website.

marsninja (changes requested):

marsninja (approved):

PR #968, opened 2023-02-24
add install command to jaseci_serv setup file

Gim3l:
Closes: #967

### DIFF
--- a/jaseci_serv/setup.py
+++ b/jaseci_serv/setup.py
@@ -1,5 +1,20 @@
+import os
+import subprocess
 from setuptools import setup, find_packages
 from os.path import join
+from setuptools.command.install import install
+
+
+class ExportStudioCommand(install):
+    """Custom install setup to help run shell commands (outside shell) before installation"""
+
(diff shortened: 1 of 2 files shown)


jaseci_serv/setup.py
```diff
@@ -1,5 +1,20 @@
...
 from setuptools import setup, find_packages
 from os.path import join
+from setuptools.command.install import install
+
+
+class ExportStudioCommand(install):
+    """Custom install setup to help run shell commands (outside shell) before installation"""
+
+    def run(self):
+        studio_dir = os.getcwd().replace("jaseci_serv", "jaseci_studio")
```
ypkang: This might causes issue in rare corner cases, where the path contains another jaseci_serv. for example `/home/gimel/jaseci_serv-experimental_repo/jaseci/jaseci_serv`

You can just remove last part of the path and then add on jaseci_studio.

ypkang (changes requested):

ypkang (approved):

PR #969, opened 2023-02-26
Docs Architype Section Completed

RolexAlexander:
Good Afternoon Dr Mars and the Development Team. I completed the architype section of the docs and also added some more info to the language basics sections. Additionally, the plan for this week is to ensure all content from jaseci bible that belongs in the architype section is migrated to the docs along with the respective examples.


docs/docs/jaseci_architype/take.md
```diff
@@ -0,0 +1,264 @@
...
+### Skipping
+
+The idea behind the abstraction of `skip` in the context of a walkers code block is that it tells a walker to halt and abandon any unfinished work on the current node in favor of moving to the next node (or complete computation if no nodes are queued up).
+
+> **Note**
+>
+> Node/edge abilities also support the usage of the skip directive. The skip merely decides not to use the remaining steps of that `ability` itself in this context.
+
+
+Lets change the `init` walker of **Example 3** to demostrate how the `skip` command works;
```
ypkang, line 14: @RolexAlexander Does it make sense to be indexing the examples in the docs like this? `Example 3`. Is there an example 3 in the docs actually? If so, we should make the `Example 3` text a hyper link to it. If not, we should remove it. There are several places here.

ypkang (changes requested):

RolexAlexander:
Good Night noted, I will make the changes.

ypkang (approved):

Issue #670, opened 2022-12-12
Ability to parse parameters to an action module


chandralegend:
Currently, when we load a module through `actions load module jaseci_ai_kit.bart_sum`, the initialization only happens for the `bart-large-cnn` model. 

What If we can parameterize the model initialization, where we can say which model to use. then the JAC user has the ability to change the module on the load stage.

Example Behavior for JSCTL
```bash
jac> actions load module jaseci_ai_kit.bart_sum --ctx {"model":"bart-tiny-cnn"}
```

Example change in Jaseci AI Kit code
```python
def setup(model):
  bart_tokenizer = BartTokenizer.from_pretrained(f"facebook/{model}")
  bart_model = BartForConditionalGeneration.from_pretrained(f"facebook/{model}").to(
      device
  )

setup(model)
```

ypkang:
Duplicate with #730

PR #971, opened 2023-02-27
Ability to Parse Context to `actions load module`

### LINKED ISSUE #730: Pass parameters for action module initilaizaation
Actions load module jaseci_ai_kit.whisper –ctx {“model”:”medium”}

chandralegend:
Closes #730 and #670


jaseci_core/jaseci/api/actions_api.py
```diff
@@ -79,7 +79,7 @@ def actions_load_remote(self, url: str):
         return {"success": success}
 
     @Interface.admin_api(cli_args=["mod"])
-    def actions_load_module(self, mod: str):
+    def actions_load_module(self, mod: str, ctx: dict = {}):
```
ypkang, line 82: This should also be supported with actions load local and remote. The corresponding functions for those two are in this file too.

jaseci_core/jaseci/actions/live_actions.py
```diff
@@ -110,6 +110,13 @@ def load_module_actions(mod, loaded_module=None):
                     del live_actions[i]
 
         mod = importlib.import_module(mod)
+        try:
+            if hasattr(mod, "setup"):
+                mod.setup(**ctx)
+        except Exception:
+            logger.error(
+                f"Cannot run setup for module {mod}. Refer the setup parameters."
```
ypkang: I don't think this flow works for existing modules that don't have a setup().  We should first check if ctx is specified, if not, and only attempt to call setup() if there is a non empty ctx specified.

ypkang: Lets also make this error message more clear. Lets say this
```
Cannot run set up for module {mod}. This could be because the module doesn't have a setup procedure for initialization, or wrong set up parameters are provided.
```

ypkang (changes requested):

jaseci_core/jaseci/actions/live_actions.py
chandralegend: According to the current logic, it checks whether there is a setup function in the module. if so runs the setup module. with the context given. if the context is empty still it will run the setup function but with the default values given in the module allowing the module to do just one intialization, so we can remove the setup call in the module file.

ypkang:
Just keep the setup actions change in the jaseci ai kit actions and we can merge those. @chandralegend going to implement a `action call`-style command in jsctl.

ypkang (changes requested):
See comment above.

ypkang:
Superseded by #987 and #989

PR #978, opened 2023-03-01
Hotfix: jsctl remote calls bug

ypkang:
Also, temporarily suppress the warning messages from initializing redis service and task service because local jsserv logs are unusable because of it.


ypkang:
cc @amadolid

PR #993, opened 2023-03-06
[MINOR]: Manual token validation util

amadolid:
image


marsninja:
Hi @amadolid , Whats the context around this change?

amadolid:
This is for manual validation tool of token inside custom actions.

While this might be helpful, it's a minor tool that can be added also directly to the custom action that will also install jsserv

Closing this PR for now

PR #995, opened 2023-03-07
`action call <action> -ctx <context>` and `action load local/module/remote <module> -ctx <context>`

### LINKED ISSUE #989: Implement action run for jsctl
This will enable user to directly call an action in jsctl. This can be useful in many scenarios. One specific use case is for initializing a jaseci action module.

For load local/module,
* As part of `actions load local/module`, we will 1) load the python file/module and 2) call the `setup()` action for that module using the `ctx` passed in with the `actions load` command. `ctx` can be empty which will run the default setup or it can be specified by the user to run custom setup.

For load remote,
* Because a remote action server is stateful, meaning that the current state (i.e. loaded models etc.) on the server should be retained and is expected to be. Therefore,`actions load remote` will 1) load the action via the remote URL and 2) ONLY call the `setup()` when there is `ctx` specified (i.e. non empty). In this case, if the a user just calls `actions load remote` which will be the common case, they will be using whatever state the server is currently in. Any custom setup will have to be done explicitly by providing a non-empty `ctx` with actions load remote.
* For a remote server, it needs a `setup()` just once after a fresh start. We can take care of that in the yaml file of the pod or the user can directly call it in jsctl.
 

cc @chandralegend

chandralegend:
closes #989

### DIFF
--- a/jaseci_core/jaseci/actions/live_actions.py
+++ b/jaseci_core/jaseci/actions/live_actions.py
@@ -70,7 +70,7 @@ def assimilate_action(func, act_group=None, aliases=list()):
     return func
 
 
-def load_local_actions(file: str):
+def load_local_actions(file: str, ctx: dict = {}):
     """Load all jaseci actions from python file"""
     try:
         name = file.rstrip(".py")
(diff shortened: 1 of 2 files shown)


ypkang (approved):
The changes here look good. I do want to manually test this one before merging.

ypkang (approved):

PR #593, opened 2022-11-09
[SET-OPERATION] supersetof, subsetof and disjointof filter for filter_compare

amadolid:
## Describe your changes
 - issuperset/issubset support on reference filter
 - Added unit test to cover changes

## Link to related issue
N/A

## Does this introduce new feature or change existing feature of Jaseci? If so, please add tests.
Yes, unit test is added

## Will this impact Jaseci users? If so, please write a paragraph describing the impact.
No

## Anything in this PR should the Jaseci developer/contributor community pay particular attention to? If so, please describe.
filtering reference

PR #1003, opened 2023-03-11
Revert "[SET-OPERATION] supersetof, subsetof and disjointof filter for filter_compare"

ypkang:
Reverts Jaseci-Labs/jaseci#593


marsninja:
This is a very interesting language feature i'm curious about. We should discuss the pros/cons of having this in. I've never seen anything like this in any other language but it is elegant. Of course we should finalize the nomenclature. I reason about it as an expansion on the `in` operator which is a set inclusion type op that is precedented in python.

PR #1005, opened 2023-03-13
Revert "Revert "[SET-OPERATION] supersetof, subsetof and disjointof filter for filter_compare""

ypkang:
Reverts Jaseci-Labs/jaseci#1003


marsninja:
I'd be happy to merge this pr if we change the normanclature to `subs`, `sups`, and `mexs` instead of `subsetof`, `supersetof`, and `disjointof`

Also I'd have these implemenations accept both lists and dictionaries (in the case of dictionaries the operator should work on [k, v] equality) 

bonus points for updating +, -, and * to support lists and dictionaries in a set theory type way where + is union, - is difference, and * is intersection (with dups allowed in all cases), of course the comparison ops above should apply on set casts of the lists.

PR #1011, opened 2023-03-16
Remote action call kwargs fix

ypkang:
Fix an issue where using kwargs for remote action call do not use default value set in the action definition


marsninja (changes requested):
Lets add a test based on this test file https://github.com/Jaseci-Labs/jaseci/blob/1fea0c35a96daf6738bad3eee014669e2937d2ff/jaseci_core/jaseci/actions/tests/test_actions.py#L16-L54

You should be able to import and make a call to the gen_remote_func_hook and then call the function with subsetted kwargs for one of the remote actions in the test, then check the request.post that goes out. You may have to use the mock library to override request.post, hit chatgpt for how to do that with "hey, how can I use mock in python to catch request.post so i can check the structure of hte post request going out"

PR #1024, opened 2023-03-23
Include setup() as part of remote action startup procedure

ypkang:


ypkang:
The implementation is ready but I want to add a test.

marsninja (approved):
LGTM @ypkang!

jaseci_core/jaseci/actions/remote_actions.py
```diff
@@ -71,7 +71,7 @@ def gen_api_service(app, func, act_group, aliases, caller_globals):
 
     # Create duplicate funtion for api endpoint and inject in call site globals
     @app.post(f"/{func.__name__}/")
-    def new_func(params: model):
+    def new_func(params: model = model.construct()):
```
ypkang, line 74: @marsninja can you take a quick look at this? The previous test failure was because when we used `model()` it was failing pedantic validation because `model` has required fields which we are not setting (and can't set here). So I am using `construct` to create a model instance without validation here. 

I tested with several remote actions, with and without required parameters and its working.

marsninja: Well if it works it works :-)

PR #1046, opened 2023-04-03
[SOCIAL-AUTH]: Support sign in with using only one of these (code, id_token, access_token)

amadolid:


marsninja (approved):
LGTM @amadolid

PR #1050, opened 2023-04-05
[REFACTOR]: Add elastic log handler when elastic service already running

amadolid:


ypkang:
Talked through the changes with Alexie and this looks good. I will merge once I test it.

PR #1074, opened 2023-04-20
Update webtool get_meta with timeout and headers

ypkang:
Also add `html5lib` parser option for future proof.


ypkang:
FYI @panikingginoo12

PR #1085, opened 2023-04-25
pdf_ext sample url updated

AshishMahendra:


ypkang (approved):
👍🏼 on the quick fix @AshishMahendra

PR #1103, opened 2023-05-02
Huggingface Integration

chandralegend:


ypkang (changes requested):
Two changes:
1. Add an action that supports all HF API endpoints, future proof.
2. Add a test case with mocking requests.post.

chandralegend:
2, Add a test case with mocking requests.post was not successful. Need Help @ypkang

ypkang (approved):

PR #1140, opened 2023-05-16
[OPENAI]: Support 3rd party AI provider

amadolid:
## Changing the setup argument's defaults to openai defaults since it's also getting from environment variables
```python
# openai
api_key = os.environ.get("OPENAI_API_KEY")
api_base = os.environ.get("OPENAI_API_BASE", "https://api.openai.com/v1")
api_type = os.environ.get("OPENAI_API_TYPE", "open_ai")
api_version = (
    "2023-03-15-preview" if api_type in ("azure", "azure_ad", "azuread") else None
)
```

### DIFF
--- a/jaseci_ai_kit/jac_misc/jac_misc/openai/main.py
+++ b/jaseci_ai_kit/jac_misc/jac_misc/openai/main.py
@@ -11,13 +11,17 @@ Requires OpenAI API Key
 
 from jaseci.jsorc.live_actions import jaseci_action
 from jaseci.utils.utils import logger
-import os
 import openai
 from typing import Union
 
 
 @jaseci_action(act_group=["openai"], allow_remote=True)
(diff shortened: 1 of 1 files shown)


ypkang:
@chandralegend can you take a look at this?

chandralegend:
I think this change doesnt not need any change in the code. Needs a small change in the DOCs.
We dont need to provide the customizabilty of the openai base and openai type through the setup. Which make the things more of extra steps. Instead we can add a line saying if you want to change those set the following EnV variables.

amadolid:
Hi @chandralegend,

This approach will still use openai default values which is from environment variables. This changes will just allow dev to customize the configs without adding/changing enviroment vars on the openai instance.
This will run as is as before with just additional option to customize it and allow triggers at any given time.

Thank you!

chandralegend (approved):

PR #1147, opened 2023-05-19
Better elastic configuration for production

ypkang:
This PR changes a number of configurations for the elastic stack to make it more scalable and suitable for production logging usage. Specifically,
* Configure a `data stream` for core and app logs.
* The data stream is configured with an elastic index lifecycle management (ILM) that
  * Automatically rollover to create backing index for every 7 days or when the current index size reaches 1GB in size
  * Automatically delete indices that are older than 14 days
 * The data stream is also configured with a index template with useful mapping rules for ingesting logs
  * `@timestamp` is mapped to `date` field in elastic so logs can be viewed/searched as timeseries data in Kibana
  * text fields (except for `message`) are mapped to `keyword` field so they can be searched.
* Increase the storage size of the PVC for elasticsearch data to 10GB
* Added additional fields to record for `walker_run` requests
  * success
  * walker_name
* Use multiprocess to handle request to log to elastic instead of thread.

### DIFF
--- a/jaseci_core/jaseci/extens/svc/elastic_svc.py
+++ b/jaseci_core/jaseci/extens/svc/elastic_svc.py
@@ -1,5 +1,6 @@
 import re
 from jaseci.jsorc.jsorc import JsOrc
+from jaseci.jsorc.jsorc_settings import JsOrcSettings
 from jaseci.extens.svc.kube_svc import KubeService
 from jaseci.utils.utils import logger, app_logger
 from requests import get, post, put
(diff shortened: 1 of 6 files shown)


ypkang:
cc @amadolid

marsninja (approved):
LGTM though I would have put a test in for the walker run detection in log dict but whatever.

marsninja:
@ypkang tests not passing, also you should beef up the test with new stuff if possible (at least the walker run detection for the log dict stuff)

ypkang:
> @ypkang tests not passing, also you should beef up the test with new stuff if possible (at least the walker run detection for the log dict stuff)

will do

PR #1168, opened 2023-06-22
Add an API that make a sentinel globally accessible

ypkang:


ypkang:
Still need to add test.

ypkang:
> Still need to add test.

Need this for Myca. Going to merge and followup with a test.

PR #1192, opened 2023-07-21
Time Series Forecasting Module

TharukaCkasthuri:


.github/workflows/jac-misc-test.yml
```diff
@@ -49,6 +49,8 @@ jobs:
       run: |
         cd jaseci_ai_kit/jac_misc
         pip install -e .[all]
+        pip install darts==0.24.0
+        pip install jaseci
```
ypkang, line 53: 1. Shouldn't `darts` in the requirements.txt for `forecast`?
2. I don't think we need to `pip install jaseci` here. the jac-misc install already includes that.

TharukaCkasthuri: pip install darts will downgrade a version of a package that is required by Jaseci. So we have to reinstall the Jaseci to fix that. Otherwise, Jaseci throws an error. I have mentioned that in the user docs in darts.

ypkang (approved):

PR #1195, opened 2023-07-25
[TRY-CATCH]: Clean up unreachable codes

amadolid:


jaseci_core/jaseci/jac/interpreter/interp.py
```diff
@@ -221,8 +221,7 @@ def run_for_stmt(self, jac_ast):
                     self._loop_ctrl = None
                     break
                 if loops > self._loop_limit:
-                    self.rt_error("Hit loop limit, breaking...", kid[0])
-                    self._loop_ctrl = "break"
```
amadolid, line 225: changing error message and remove unreachable code
Loop limit should throw exception instead of just breaking the loop. This is to inform the developer that it has already stop the call as we can only process part of their loop. This is to avoid further discrepancies if there's another process involved after the loop.

jaseci_core/jaseci/jac/interpreter/interp.py
```diff
@@ -240,8 +239,7 @@ def run_for_stmt(self, jac_ast):
                         self._loop_ctrl = None
                         break
                     if loops > self._loop_limit:
-                        self.rt_error("Hit loop limit, breaking...", kid[0])
-                        self._loop_ctrl = "break"
```
amadolid, line 244: changing error message and remove unreachable code
Loop limit should throw exception instead of just breaking the loop. This is to inform the developer that it has already stop the call as we can only process part of their loop. This is to avoid further discrepancies if there's another process involved after the loop.

jaseci_core/jaseci/jac/interpreter/interp.py
```diff
@@ -269,8 +267,7 @@ def run_for_stmt(self, jac_ast):
...
```
amadolid, line 273: changing error message and remove unreachable code
Loop limit should throw exception instead of just breaking the loop. This is to inform the developer that it has already stop the call as we can only process part of their loop. This is to avoid further discrepancies if there's another process involved after the loop.

jaseci_core/jaseci/jac/interpreter/interp.py
```diff
@@ -291,8 +288,7 @@ def run_while_stmt(self, jac_ast):
                 self._loop_ctrl = None
                 break
             if loops > self._loop_limit:
-                self.rt_error("Hit loop limit, breaking...", kid[0])
-                self._loop_ctrl = "break"
```
amadolid, line 295: changing error message and remove unreachable code
Loop limit should throw exception instead of just breaking the loop. This is to inform the developer that it has already stop the call as we can only process part of their loop. This is to avoid further discrepancies if there's another process involved after the loop.

jaseci_core/jaseci/jac/interpreter/interp.py
```diff
@@ -1663,7 +1647,6 @@ def run_filter_compare(self, jac_ast, obj):
             return self.pop().value
         else:
             self.rt_error(f"{name} not present in object", kid[0])
-            return False
```
amadolid, line 1666: Unreachable code.
Forcing error on filtering let the developer know that there are still elements that are not yet synced on the latest structure and they still need to do a proper migration

jaseci_core/jaseci/prim/sentinel.py
```diff
@@ -145,7 +145,10 @@ def get_arch_for(self, obj):
         """Returns the architype that matches object"""
         ret = self.arch_ids.get_obj_by_name(name=obj.name, kind=obj.kind)
         if ret is None:
-            self.rt_error(f"Unable to find architype for {obj.name}, {obj.kind}")
+            self.rt_error(
+                f"Unable to find architype for {obj.name}, {obj.kind}",
+                self._cur_jac_ast,
+            )
```
amadolid: Added jac_ast for rt_error

amadolid (commented):
Added some reference per line change

jaseci_core/jaseci/prim/sentinel.py
```diff
@@ -145,7 +145,10 @@ def get_arch_for(self, obj):
         """Returns the architype that matches object"""
         ret = self.arch_ids.get_obj_by_name(name=obj.name, kind=obj.kind)
         if ret is None:
-            self.rt_error(f"Unable to find architype for {obj.name}, {obj.kind}")
+            self.rt_subtle_error(
+                f"Unable to find architype for {obj.name}, {obj.kind}",
+                self._cur_jac_ast,
+            )
```
amadolid, line 151: Adding subtle error. This will not halt the procedure. It will just show that there's an error happen.

This will only occurs when there's an old architype that is not existing anymore.
Syntax that includes this `-->` with old node that's not existing anymore will error out.

ypkang (approved):

PR #1199, opened 2023-07-27
New Module: Generate Embedding (gen_emb)

AshishMahendra:


jaseci_ai_kit/jac_nlp/jac_nlp/gen_emb/gen_emb.py
```diff
@@ -0,0 +1,36 @@
...
+@jaseci_action(act_group=["gen_emb"], allow_remote=True)
+def generate_embedding(texts: Union[List[str], str], selected_module: str):
+    if selected_module not in model_name_map:
+        raise HTTPException(
+            status_code=500,
+            detail=f"Invalid module selected. Please choose from the available options: {', '.join(model_name_map.keys())}",  # noqa
+        )
+
+    module_path, embeddings_function, setup_function = model_name_map[selected_module]
+    module = __import__(module_path, fromlist=[embeddings_function, setup_function])
```
ypkang: We shouldn't import the module as part of the generate_embedding call since it can slow down a lot. Maybe we can move import to a `setup` call?

jaseci_ai_kit/jac_nlp/jac_nlp/gen_emb/gen_emb.py
```diff
@@ -0,0 +1,36 @@
...
+    embeddings_function = getattr(module, embeddings_function)
+    setup_function = getattr(module, setup_function)
+
+    setup_function()  # Setup the selected module
```
ypkang: same here. we should do this in  `setup`

ypkang (changes requested):

jaseci_ai_kit/jac_nlp/jac_nlp/gen_emb/gen_emb.py
AshishMahendra: updated

jaseci_ai_kit/jac_nlp/jac_nlp/gen_emb/gen_emb.py
AshishMahendra: updated

PR #1200, opened 2023-07-28
[HOTFIX]: Bypass node saving when smart building edges

amadolid:
Not fully tested yet but it's a possible fix on current issue on all nodes are being save when traversing

### DIFF
--- a/jaseci_core/jaseci/prim/node.py
+++ b/jaseci_core/jaseci/prim/node.py
@@ -60,7 +60,10 @@ class Node(Element, Anchored):
                 edge.bidirected = v[1] == BI
                 edge.jid = v[2] if len(v) > 2 else uuid.uuid4().urn
                 edge.context = v[3] if len(v) > 3 else {}
-                edge.save()
+                # old `edge.save()` might be confusing
+                # this line doesn't mean it has to be saved on db
+                # it only needs to be available on cache (memory, redis)
+                self._h.commit_obj_to_cache(edge, True)
                 self._fast_edge_ids.add_obj(edge, bypass=True)
 
     def smart_add_edge(self, obj):
--- a/jaseci_core/jaseci/prim/node.py
+++ b/jaseci_core/jaseci/prim/node.py
@@ -61,7 +61,7 @@ class Node(Element, Anchored):
                 edge.jid = v[2] if len(v) > 2 else uuid.uuid4().urn
                 edge.context = v[3] if len(v) > 3 else {}
                 edge.save()
-                self._fast_edge_ids.add_obj(edge)
+                self._fast_edge_ids.add_obj(edge, bypass=True)
 
     def smart_add_edge(self, obj):
         # make sure fast edges built
--- a/jaseci_core/jaseci/prim/node.py
+++ b/jaseci_core/jaseci/prim/node.py
@@ -52,18 +52,22 @@ class Node(Element, Anchored):
         for k in self.fast_edges.keys():
             for v in self.fast_edges[k]:
                 link_order = [v[0], self.jid] if v[1] == FROM else [self.jid, v[0]]
-                edge = Edge(
-                    m_id=self._m_id, h=self._h, kind="edge", name=k, auto_save=False
-                )
-                edge.from_node_id = link_order[0]
-                edge.to_node_id = link_order[1]
-                edge.bidirected = v[1] == BI
-                edge.jid = v[2] if len(v) > 2 else uuid.uuid4().urn
-                edge.context = v[3] if len(v) > 3 else {}
-                # old `edge.save()` might be confusing
-                # this line doesn't mean it has to be saved on db
-                # it only needs to be available on cache (memory, redis)
-                self._h.commit_obj_to_cache(edge, True)
+                edge = self._h.get_obj(self._m_id, v[2])
+                if edge:
+                    v[3] = edge.context
+                else:
+                    edge = Edge(
+                        m_id=self._m_id, h=self._h, kind="edge", name=k, auto_save=False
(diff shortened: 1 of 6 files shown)


ypkang:
@amadolid I am not sure if this actually fixes the issue where reading nodes cause a save. Let's discuss.

jaseci_core/jaseci/prim/node.py
```diff
@@ -52,16 +52,23 @@ def smart_build_fast_edge_ids(self):
...
-                edge.from_node_id = link_order[0]
-                edge.to_node_id = link_order[1]
-                edge.bidirected = v[1] == BI
-                edge.jid = v[2] if len(v) > 2 else uuid.uuid4().urn
-                edge.context = v[3] if len(v) > 3 else {}
-                edge.save()
-                self._fast_edge_ids.add_obj(edge)
+                edge = self._h.get_obj(self._m_id, v[2])
+                if edge:
+                    v[3] = edge.context
```
ypkang, line 57: @amadolid I assume this is to handle cases where a particular fast edge is already built by another node and you are avoiding creating the same Edge object again?

jaseci_serv/jaseci_serv/base/tests/test_orm_hooks.py
```diff
@@ -315,3 +315,186 @@ def test_fast_edges_with_context(self):
...
+        )
+
+        hook.commit()
+        hook.clear_cache()
+
+        test_walker = sent.run_architype("sample3")
+        test_walker.prime(gph)
+        test_walker.run()
+
+        # no a3::node since it was already disconnected
```
ypkang, line 395: Even it is disconnted, `a3` should still be saved right? Because you are modifying a3 directly with `a3: here.val = 2;`

jaseci_serv/jaseci_serv/base/tests/test_orm_hooks.py
```diff
@@ -315,3 +315,186 @@ def test_fast_edges_with_context(self):
...
+
+        hook.commit()
+        hook.clear_cache()
+
+        # read node::a1 without accessing graph::root
+        test_walker = sent.run_architype("sample6")
+        test_walker.prime(a1)
+        reps = test_walker.run()["report"]
+
+        self.assertEqual([8], reps)
```
ypkang, line 500: Should also assert here that `save_obj_list` is empty.

jaseci_serv/jaseci_serv/jac_api/tests/test_jac_api.py
```diff
@@ -1,9 +1,12 @@
 import base64
 import json
 
+from unittest.mock import MagicMock, Mock
```
ypkang, line 4: @marsninja when you look at this, you can ignore the change in this file -- this is for something else and not related to this bug.

jaseci_core/jaseci/utils/id_list.py
```diff
@@ -27,7 +27,9 @@ def __init__(self, parent_obj, auto_save=True, in_list=None):
     def cache_reset(self):
         self.cached_objects = []
 
-    def add_obj(self, obj, push_front=False, allow_dups=False, silent=False):
+    def add_obj(
+        self, obj, push_front=False, allow_dups=False, silent=False, bypass=False
+    ):
```
ypkang, line 32: @amadolid I talked with @marsninja about this PR. He said that if we set `auto_save` as false for the fast_edge_id IdList should have the same effect as what we are trying to do with bypass. Can you test that out and make the change if it works?

jaseci_core/jaseci/prim/node.py
amadolid: Yes sir, tho there's another referencing happen
`v[3] = edge.context`

the last node that got build (`from` or `to`), their fast_edge's context will be used as same context on the created Edge.

```python
# node.py:66
edge.context = v[3] if len(v) > 3 else {}
```

if any change happen on this edge.context it will also be the same value on `v[3]` owner.
```python
# node.py:57
v[3] = edge.context
```

this line will handle the issue. Instead of rebuilding the same Edge we will now try to get it from memory and sync the reference of existing Edge.context to current fast_edge's `v[3]` (context)

jaseci_serv/jaseci_serv/base/tests/test_orm_hooks.py
amadolid: the second trigger of `sample3` will not include a3 as it was already disconnected from first `sample3` trigger

jaseci_serv/jaseci_serv/base/tests/test_orm_hooks.py
amadolid: Added the assert

jaseci_core/jaseci/utils/id_list.py
amadolid: I think, I have implement this before and saw that it might not trigger the saving on other approach reason for just bypassing only the buildup part

we call `smart_add_edge` on edge.connect and this function will try to add to _fast_edge_ids. I'm assuming it supposed to do the saving on parent's node whenever we trigger the smart_add_edge. `auto_save` false will disable this
```python
def smart_add_edge(self, obj):
        # make sure fast edges built
        if not len(self._fast_edge_ids):
            self.smart_build_fast_edge_ids()
        # add new edge
        self._fast_edge_ids.add_obj(obj)
        # then store how needed
        if obj.is_fast():
            self.smart_edge_to_fast_edge(obj)
        else:
            self.edge_ids.add_obj(obj)
```

cc: @marsninja

ypkang (approved):

PR #1229, opened 2023-10-24
[BUGFIX]: Webtool optional headers

amadolid:
MINOR:
 - Use different url for get_meta_need_auth as previous URL ([google docs](https://docs.google.com/presentation/d/1lIYEuzzhZZ9PJaG_u3XgrFXX5Y6xd0zHV-aB2F8bXXU/edit)) doesn't have meta if loaded without javascript


marsninja (approved):
Awesome thanks.

PR #1231, opened 2023-10-24
[UNIT-TEST]: Change dummy api

amadolid:
REASON:
 - Old url https://jsonplaceholder.typicode.com/todos/100 currently returning 403


amadolid:
it's working now again

PR #1272, opened 2024-09-11
trailing comma support implemented

ThakeeNathees:
```py
"""
This test file is to ensure the valid syntax of jac lang.
Add new jac syntax here to test if it compile without any issue.
"""

# Import statement without trailing comma.
import:py from time {
  sleep,
  timezone,
  tzname
}


# Import statement with trailing comma.
import:py from os {
  path,
  getenv,
  getpid,
}


enum WithoutTrailComma {
    FOO = "FOO",
    BAR = "BAR"
}


enum WithTrailComma {
    FOO = "FOO",
    BAR = "BAR",
}


can without_trail_comma(a:int, b:int) {}

can with_trail_comma(
      a:int,
      b:int,
      c:int,
    ) {}


with entry {

    dict_without_trail_comma = { "key" : "value" };
    dict_with_trail_comma = {
        "key" : "val",
    };

    list_without_trail_comma = [ "foo", "bar" ];
    list_with_trail_comma = [
        "foo",
        "bar",
    ];

    set_without_trail_comma = { "foo", "bar" };
    set_with_trail_comma = {
      "foo",
      "bar",
    };

    print("Code compiled and ran successfully!");
}


```


kugesan1105 (changes requested):
The changes look good overall @ThakeeNathees 🎯. However, it would be great if we could extend support for trailing commas to tuples as well:  @marsninja  
```
(1, 2, 3,)
```

kugesan1105:
Hi  @AshishMahendra , the recent grammar change introducing trailing commas is causing the formatter to break, particularly in the FuncCall handling. The IndexError: list index out of range seems to occur when processing the trailing commas. Could you please look into fixing this in the same PR?

```
     foo_instance1 = foo11(val=90);
    # foo_instance1_2 = foo11(val=90,);  # TODO: FORMATTER: trailing comma is not supported

    foo_instance3 = foo11(2, 34, val3=90);
    # foo_instance3_2 = foo11(2,34,val3=90,);  # TODO: FORMATTER: trailing comma is not supported

    foo_instance4 = foo11(2, 34,val3=90, val4=90);
    # foo_instance4_2 = foo11(2, 34,val3=90, val4=90,);   # TODO: FORMATTER: trailing comma is not supported

kugesan1105 (approved):

PR #1287, opened 2024-09-14
Expr typeinfo

ThakeeNathees:


jac/jaclang/compiler/absyntree.py
```diff
@@ -346,6 +346,31 @@ def __init__(self) -> None:
 class Expr(AstNode):
     """Expr node type for Jac Ast."""
 
+    def __init__(self) -> None:
+        """Initialize expression node."""
+        self._sym_type: str = "NoType"
+        self._type_sym_tab: Optional[SymbolTable] = None
+
+    @property
+    def sym_type(self) -> str:
```
mgtm98: Can we change this to expr type

jac/jaclang/compiler/passes/utils/mypy_ast_build.py
```diff
@@ -131,63 +183,45 @@ def __init__(
...
-        return out
-
-    def visit_member_expr(
-        self, e: myfp.MemberExpr, is_lvalue: bool = False
-    ) -> myb.Type:
-        """Type check a member expr."""
-        out = super().visit_member_expr(e, is_lvalue)
-        FuseTypeInfoPass.node_type_hash[e] = out
-        return out
+        # For every expression there, create attach a method on this instance (self) named "enter_expr()"
```
mgtm98: Can we change this part and add the visitor methods manually instead of dynamically attaching them here

mgtm98 (approved):
For me the changes looks good, thakee will add more tests to cover the most of the supported expr. 
An item on me to change FuseTypeInfo to use the new handling (remove expand atom trailer code)

mgtm98 (approved):
thanks thakee, I see we can merge that

jac/jaclang/tests/test_cli.py
```diff
@@ -244,9 +280,9 @@ def test_type_info(self) -> None:
         cli.tool("ir", ["ast", f"{self.fixture_abs_path('type_info.jac')}"])
         sys.stdout = sys.__stdout__
         stdout_value = captured_output.getvalue()
-        self.assertEqual(stdout_value.count("type_info.ServerWrapper"), 7)
+        self.assertEqual(stdout_value.count("type_info.ServerWrapper"), 5)
         self.assertEqual(stdout_value.count("builtins.int"), 3)
-        self.assertEqual(stdout_value.count("builtins.str"), 10)
+        self.assertEqual(stdout_value.count("builtins.str"), 8)
```
marsninja: @mgtm98 @ThakeeNathees is this part ok, looks like we're changing the number of types recongized here?

marsninja (changes requested):
This look ok, except for my one question there on changing the number of type matches that are coming from the one test.

jac/jaclang/tests/test_cli.py
ThakeeNathees: If the values are literal mypy type would be `Literal['foo']`, it changes the typeinfo which I need to update here to make the tests pass.

marsninja: @ThakeeNathees I see, but wouldn't we need to then add a couple more assertEquals/assertIns to validate that Literal['foo'] is present. Basically we should be able to cover the 4 instances that change.

ThakeeNathees: * Note that the expression type `type_info.ServerWrapper` count is reduced by 2, they are of node `ArchRef`.
* Even though archref is marked as expression node, a NameAtom instance is set inside the arcref called `arch_name` which is the expression that holds the expression type info
* And the typeinfo by mypy and the fuse type pass will set the type information to that givent `arch_name`
* Before this pr Archref has the type `ServerWrapper` because `AstSymbolNode` initialize `name_spec` with the `arch_name` of the archtype instance and `expr_type` (expr_typ of expresstion nodes = sym_type of symbol nodes) returns the `sym_type` of that `arch_name`.
* In here expression type is fetched from mypy and set to each expression nodes. The archtype instance and it's attribute arch_name are two different nodes, only the `arch_name` gets the typeinfo.

* This needs to be discussed further.

CC @marsninja

marsninja (approved):

PR #1299, opened 2024-09-18
Test to diagnosis syntax error in impl file

kugesan1105:

### DIFF
--- a/jac/jaclang/langserve/engine.py
+++ b/jac/jaclang/langserve/engine.py
@@ -80,28 +80,23 @@ class JacLangServer(LanguageServer):
                     impl_parent=self.modules[file_path],
                 )
 
-    def quick_check(self, file_path: str) -> tuple[bool, list[lspt.Diagnostic]]:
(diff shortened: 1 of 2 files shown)


jac/jaclang/langserve/engine.py
```diff
@@ -80,23 +80,28 @@ def update_modules(
                     impl_parent=self.modules[file_path],
                 )
 
-    def quick_check(self, file_path: str) -> bool:
+    def quick_check(self, file_path: str) -> tuple[bool, list[lspt.Diagnostic]]:
```
marsninja: Not a fan of returning a tuple here just for the purpose of test case checking. 

can you use the hover possition strategy for checking as in other tests?

marsninja (changes requested):

jac/jaclang/langserve/engine.py
marsninja: @kugesan1105 let me know if that works, the `lsp.Position` type stuff in the hover tests.

PR #1328, opened 2024-09-29
syntax error was implemented

ThakeeNathees:
Description

Now the alerts are printed with the visual location indication for the end user to understand cleanly. Here are the screen shots of the implementation and the old one.

Current:
image

Previous:
image


mgtm98 (commented):
Thanks a lot for your efforts here in this PR. It will really help users to use the lang. I have some questions

1. When I print the ast and there is a syntax error it prints it however it should error out 
2. When I run a code with a syntax error it shows the new error message besides other error messages from python, we need to hide the python error message 
3. If there is a run time error we need to pretty print the jac source too

ThakeeNathees:
@mgtm98 The python errors are result of jac import and should be removed from a different pr imo, it's the same for run time errors and there is a pretty print for run time error implemented which is the first error message before all the python errors. Once we get rid of the python errors it would be cleaner.

mgtm98 (approved):
Had a call with thakee and I see this PR can be merged

marsninja:
Hi Guys @ThakeeNathees @mgtm98 and @kugesan1105 I have a simplifying refactor that should be a quick change. Instead of introducing the file_source at the interface of every token type, lets simply change file_path, to be a src object of type JacSource. JacSource objects include both the file_path and teh file_source. See snips below

This is the JacSource object:

image

so instead of adding a file_source to below, I'd simply change file_path to be orig_src of type JacSource:

image

and then refactor throughout using orig_src for everything, (as a bonus you can change the var name in Module type to be orig_src too for consistency, though not required).

Make sense @ThakeeNathees ?

marsninja (changes requested):
See my comment.

ThakeeNathees:
Suppressed by #1379

PR #1340, opened 2024-10-04
[LOGGER]: Initial Jac-Cloud logger handler

amadolid:


ypkang:
* Let's see if we can use rotating log setup for filebeat as well so we just have one single logging setup
* Put all the rotating log files in a dedicated directory `jac_cloud_logs`

ypkang:
* change to json dump for response and payload
* move filebeat.yaml to a `scripts/` folder
* add documentation on how to use. specifically, the environment variables you can configure.

ypkang:
@amadolid can you add a quick documentation on instructions to connect a jac-cloud instance with an elastic instance (via filebeat)?

ypkang (approved):
Tested and this works well for me. I made some tweaks to the documentation @amadolid FYI.

PR #1344, opened 2024-10-05
Implementing import declarations from python through SymTableBuildPass

mgtm98:
## **Description**
In this pull request, I implemented the ability to import Python declarations (such as classes, variables, and abilities) into Jac. This is done through `SymbolTableBuildPass`, where the entire Python module specified by `ModulePath` is added to the Jac AST. The necessary items from this module are identified and stored in `py_needed_items`. During `SymbolTableBuildPass`, only these selected items are added to the symbol table.
`py_needed_items` is defined as an optional dict, if it's None then the whole module is needed and the new support won't work, if the dict has a value then the specified items in the dict are only added in the symbol table.


mgtm98:
@marsninja Need to check with you the import of assignment nodes.

kugesan1105 (approved):
All good!

marsninja:
@mgtm98 @kugesan1105 can you guys resolve conflicts real quick. I suspect some other merged code introduced it, im guessing would be quick. Then I'll merge.

mgtm98:
@marsninja PR is ready now

PR #1348, opened 2024-10-07
[JAC-CLOUD-MINI]: Initial Mini Version

amadolid:


jac-cloud/jac_cloud/tests/simple_graph_mini.jac
```diff
@@ -0,0 +1,347 @@
+"""Example of simple walker walking nodes."""
```
ypkang, line 1: Is it necessary to have different jac file and openAPI specs for testing mini vs normal jac-cloud? They should be the same right?

jac-cloud/jac_cloud/tests/test_simple_graph_mini.py
```diff
@@ -0,0 +1,416 @@
+"""JacLang Jaseci Unit Test."""
```
ypkang, line 1: Is there anyway we can share the test methods between jac-cloud and jac-mini? e.g., maybe inherit the test class and change the setup logic depending if its mini or cloud?

ypkang (changes requested):

jac-cloud/jac_cloud/tests/simple_graph_mini.jac
amadolid: Mini version only includes walker API generations. Most of the features from jac-cloud will not work

jac-cloud/jac_cloud/tests/test_simple_graph_mini.py
amadolid: Unit Tests also have different checks as jac-mini is very minimal implementations. It also uses different DB

amadolid:
REBASED

ypkang:
@amadolid I tested this and it works. But I would like to understand better the difference in feature sets between jac-mini and full-featured jac-cloud.  And, as a result, see if we can optimize the arrangement of the test file and the openapi specs JSON file to reduce some duplication.

The reason I'd like to share as much as possible between mini and full cloud is for maintainability, even for the testing stuff.

### Comparing `simple_graph_mini.jac` and `simple_graph.jac`

I see the the mini version does not have any `return` statements. Is that not supported with mini?

### Comparing the two openapi specs json (The screenshots below have the full specs on the left and mini specs on the right)

1. the health checking endpoint is different. I assume this can be made the same between the two?
CleanShot 2024-10-11 at 18 17 23
2. Most of the walker APIs difference looks like the following. What is fundamentally the difference in the APIs created via the two modes?
CleanShot 2024-10-11 at 18 19 13
3. Some APIs have error type in their specs (e.g. 422) and those are not present in the mini version. What is the reason there?
CleanShot 2024-10-11 at 18 22 12


Ideally, we should have one base set of test files (jac, specs) that represent the core API functionalities of jac serve. These test files are then used for jac-mini and then augmented to be used for jac-cloud. For example, we can have another specs json files with just the SSO and auth endpoints and then combine the two in the tests. If we keep two set of test files, it is going to get difficult to maintain in the future.

ypkang:
* Create core test jac file and core python test case that can be shared between jac-mini and jac-cloud
* We can then create test jac and python test case for jac-cloud and import from the core files.
* We will also explore using yaml instead of json for the openapi specs so its less lines of code.

ypkang (changes requested):
This so much cleaner! 

One more ask :) -- could we share the same pytest too?

ypkang:
Oh also, need a rebase. @amadolid

amadolid:
> One more ask :) -- could we share the same pytest too?

Currently, this is not possible. Their tests have different behavior. Even tho the use the same method/api name it still have different request (authentication), response (returns support) and db checks (shelves vs mongodb)

ypkang (approved):

PR #1363, opened 2024-10-11
[JAC-CLOUD]: Improved Deserialization

amadolid:
CHANGES:
 - Manually assign fields from datasource based on dataclass fields
 - Value error will be raised
   - if field is required but not existing from db
   - if field has different type from db
 - default will get populated if field has default/factory and not existing from db
 - excess fields will be included but no type checking

OLD:
 - Uses __init__ generated by dataclass


ChrisIsKing:
@amadolid This is good. Can we also consider the case where the node type is not present. I would like for an warning to be thrown in this case instead in addition to defaulting to the base NodeArchitype.

ypkang:
This looks good to me. @ChrisIsKing feel free to merge after the warning is added.

ypkang (approved):
LGTM

PR #1389, opened 2024-10-17
[JAC-CLOUD]: Refactors and enhancements

amadolid:
CHANGES:
 - Align jac-cloud-mini response to jac-cloud
 - use same unit test for jac-cloud and jac-cloud-mini
 - minor improvement on jaclang
    - move asdict from jac-cloud to jaclang
    - move returns from jac-cloud to jaclang


ypkang:
@amadolid I think we might not need all of this PR right? after we merge local mongodb option. Can you update this to just have jac-cloud improvements we still need?

amadolid:
Yup, tho there's a improvement here regarding "returns" but we can discuss it later

PR #1398, opened 2024-10-21
Fix for Caching Issue in jac test

### LINKED ISSUE #1396: Jac test Caching Issue – Changes Not Detected Without jac clean
## **Description**
<br>
It seems that jac test is not recognizing updates to files after they are modified. Even if the file content changes, the results remain the same unless I manually execute jac clean before rerunning the tests. This adds an extra step in the workflow that might be unintended.

### Steps to replicate:

1. Run jac test on a file.
2. Modify the file.
3. Re-run jac test without running jac clean.
4. The test results remain unchanged despite the modifications.

### Expected behavior:
- The jac test command should detect file changes and run the updated tests without needing a manual jac clean after each modification.
<be>

jac_test1

Thamirawaran:
## **Description**
Fixed the caching issue in jac test.

## Changes
1. run_test function in default.py file is updated to avoid caching
2. test case is added in test_cli.py

## Issue
Fixes #1396


kugesan1105:
Good one, @Thamirawaran 👍 
Could you add a test case in test_cli.py to cover this scenario?

kugesan1105 (commented):
Good one, @Thamirawaran ! 👍
 Could you also add a test case in test_cli.py to cover this?

kugesan1105 (changes requested):
Good one, @Thamirawaran ! 👍
Could you also add a test case in test_cli.py to cover this?

marsninja (approved):
Very nice again! @Thamirawaran

PR #1399, opened 2024-10-22
python errors (compile time, runtime) are removed

ThakeeNathees:
## **Description**

for jaclang compile time or run time errors it'll only pretty print the source location of the error occurred without python's errors, would be cleaner for the end user to understand what went wrong.

Screenshot 2024-10-22 162009


jac/jaclang/cli/cli.py
```diff
@@ -89,23 +93,30 @@ def run(
...
-            base_path=base,
-            cachable=cache,
-            override_name="__main__" if main else None,
-        )
-    elif filename.endswith(".jir"):
-        with open(filename, "rb") as f:
-            JacMachine(base).attach_program(
-                JacProgram(mod_bundle=pickle.load(f), bytecode=None)
-            )
+        try:
```
mgtm98, line 96: Can we combine both the if and else and do one try except on both of them

ThakeeNathees: This will specifically catch the exception by the jac_import, leaving other kind of exception to throw as it may needed to debug the jaclang itself.

marsninja (approved):

PR #1401, opened 2024-10-23
[FEATURE]: Elevate Root

amadolid:
Support elevate current root to be system root

### DIFF
--- a/jac/jaclang/plugin/default.py
+++ b/jac/jaclang/plugin/default.py
@@ -55,6 +55,13 @@ class JacAccessValidationImpl:
 
     @staticmethod
     @hookimpl
+    def elevate_root() -> None:
+        """Elevate context root to system_root."""
+        jctx = Jac.get_context()
+        jctx.root = jctx.system_root
+
+    @staticmethod
+    @hookimpl
     def allow_root(
         architype: Architype, root_id: UUID, level: AccessLevel | int | str
     ) -> None:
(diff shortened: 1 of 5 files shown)


ChrisIsKing:
@amadolid Let's say you wanted to elevate to root access by from the python side instead of jac, what would be the interface to do that? Like if I wanted to run `_Jac.spawn_call()` programmatically.

amadolid:
> @amadolid Let's say you wanted to elevate to root access by from the python side instead of jac, what would be the interface to do that? Like if I wanted to run `_Jac.spawn_call()` programmatically.

it should be the same `Jac.elevate_root()` or `_Jac.elevate_root()`

jac/jaclang/plugin/default.py
```diff
@@ -54,6 +54,13 @@
 class JacAccessValidationImpl:
     """Jac Access Validation Implementations."""
 
+    @staticmethod
+    @hookimpl
+    def elavate_root() -> None:
```
ChrisIsKing: Everything looks fine, just that it should be spelt `elevate_root`

ChrisIsKing (changes requested):

jac/jaclang/plugin/default.py
ypkang: @amadolid ping on this.

amadolid:
Rename and rebased

ypkang (approved):

ChrisIsKing (approved):

PR #1407, opened 2024-10-26
semantic errors for uninitialized obj hasvars

ThakeeNathees:
## **Description**

```py

# 1. Test initialization order.
obj Test1 {
    has var1: int;
    has var2: int = 42;
    has var3: int; # <-- This should be syntax error.
}

# 2. Test if all un initialized vars are initialized in the init method.
obj Test2 {
    has var1: int;
    has var2: int;
    has var3: int;
    has var4: int = 42;

    can init() {  # <-- Should be error because "var2", "var3" are not initialized.
        self.var1 = 1;
    }
}

# 3. Test the same of Test2 but the init is defined somewhere else.
obj Test3 {
    has var1: int;
    has var2: int;
    has var3: int;
    has var4: int = 42;

    can init();
}

:obj:Test3:can:init() {
    self.var1 = 1;
}

```

image


ThakeeNathees:
suppressed by: https://github.com/Jaseci-Labs/jaseci/pull/1439

Issue #1405, opened 2024-10-25
[Bug] Syntax Error Bug


kugesan1105:
```
obj Shape {
    has shape_type: ShapeType int; # It should generate syntax error but it doesnot

}
```
![image](https://github.com/user-attachments/assets/db2ec4dd-4d13-43c9-a214-950c4d58198f)

```
obj Shape {
   # has shape_type: ShapeType str;  #will genrate syntax error 
}
```
![image](https://github.com/user-attachments/assets/654ddf09-a42a-486d-8385-d8581e539b16)

PR #1412, opened 2024-10-29
lark file keywords word boundry added

ThakeeNathees:
## **Description**
FIX: #1405


ThakeeNathees:
Not sure why all the tests are failing here, however all are passing in my local. @marsninja @kugesan1105

PR #1414, opened 2024-10-30
[Chore] SemIR is raised into JacProgram

Jayanaka-98:
## **Description**

- The semtable is not available in the jac program.
- No pickel files will be generated.
- To read the sem registry from within the process itself. 
```python
import from jaclang.runtimelib.machine { JacMachine }
semtable = JacMachine.get().jac_program.sem_ir;
```


Jayanaka-98:
All the tests pass on local machine (using jaclang and mtllm from source) but as jac changes are not pushed, mtllm tests fail. Same case for the precommit.

marsninja (approved):

PR #1415, opened 2024-10-31
[WEBSOCKET]: Initial Implementation

amadolid:
# **`FEATURES`**:
- User Notifications
- Channel Notifications
- Client Notifications
- Support Cross Instance Notifications
- Support Running Multiple Worker Uvicorn
- Includes Tester Files
  - runner file for multiple worker uvicorn
  - websocket.jac for sample usage
  
---
# **`WEBSOCKET`**

## Walker Declaration
- walker can be declared as websocket thru __specs__ configurations.
- it can still work along with other http methods however, the only limitation is it doesn't support file (maybe in the future).
```python
walker your_event_name {
    has val: int;
    can enter with `root entry {
        report "Do something!";
    }

    class __specs__ {
        has methods: list = ["websocket"];
    }
}
```

- once connected, you will recieved first event for connection information
- there's two type of connection
    - Authenticated - with valid Authorization Token
    - Non Authenticated
- you may specify your desired channel_id via query param
    - this will recieved notification from specific channel
    - usage: group chat/notification
```python
{
	"type": "connection",
	"data": {
        # your websocket client id
		"client_id": "1730887348:f46d85203c704c099e9f44e948322a20",

        # user's root_id
		"root_id": "n::672b35cec309e5ef8469c372",

        # non authenticated
		# "root_id": "n::000000000000000000000001",

        # user's channel id, random if not specified
		"channel_id": "1730887348:796ad2e9fa3e484ebe01f071c381b7e8"
	}
}
```

### Walker
- this will trigger a normal walker as if it was trigger via rest API
```python
{
    # event type
	"type": "walker",

    # walker's name
	"walker": "your_event_name",

    # if you want to recieve a notification for response
	"response": true,

    # walker's request context
	"context": {
        "val": 1
    }
}
```

### User
- this will send notification to target user's clients
- if target user/s has multiple clients, all of it will get notified
```python
{
    # event type
	"type": "user",

    # target user/s via root_id
    "root_ids": ["n::672b35cec309e5ef8469c372"],

    # data you want to send
	"data": {
        "val": 1
    }
}
```

### Channel
- this will send notification to target channel/s
- all clients that's subcribed to the channel will get notified
```python
{
    # event type
	"type": "channel",

    # target channel_id/s
    "channel_ids": ["anystring"],

    # data you want to send
	"data": {
        "val": 1
    }
}
```

### Client
- this will send notification to target client/s
```python
{
    # event type
	"type": "client",

    # target client_id/s
    "client_ids": ["1730887348:f46d85203c704c099e9f44e948322a20"],

    # data you want to send
	"data": {
        "val": 1
    }
}
```

### **`PREREQUISITE`**
```python
import:py from jac_cloud.plugin {WEBSOCKET_MANAGER as socket}
```

### Self
- this will send notification to current websocket it was from (only valid on websocket walker event)
- if via walker api, nothing will happen
```python
socket.notify_self({"any_field": "for_progress", "progress": "0%", "status": "started"});
```

### User
- this will send notification to target user's clients
- if target user/s has multiple clients, all of it will get notified
```python
socket.notify_users([root], {"any_field": "for_progress", "progress": "0%", "status": "started"});
```

### Channel
- this will send notification to target channel/s
- all clients that's subcribed to the channel will get notified
```python
socket.notify_channels([channel_id], {"any_field": "for_progress", "progress": "0%", "status": "started"});
```

### Client
- this will send notification to target client/s
```python
socket.notify_clients([client_id], {"any_field": "for_progress", "progress": "0%", "status": "started"});
```


amadolid:
BACKUP
https://github.com/amadolid/jaseci/tree/websocket-backup

amadolid:
Will continue to upgrade to support channels using
https://github.com/encode/broadcaster

amadolid:
REBASED

ypkang:
@amadolid I tested this and have some question:

1. How do we set up a server-side walker that listens on a connected web socket and continuously receive data from it? For example, something like this
```
@app.websocket("/ws")
async def websocket_endpoint(websocket: WebSocket):
    await websocket.accept()
    while True:
        data = await websocket.receive_text()
        await websocket.send_text(f"Message text was: {data}")
```

2. Does the notify_user/channel/client accept string or does it have to be json? The example above is just sending a string via `send_text`

ypkang:
Also, I think we are going to need a different type of documentation on this. We need to include end-to-end examples, like what fastAPI has https://fastapi.tiangolo.com/advanced/websockets/#await-for-messages-and-send-messages.

We need maybe 3 common use case:
1. A user establishing a connection with server and continuously exchanging messages
2. Multiple users (unauth) establish connections via a specific channel id and listening on messages. An authenticated user broadcast a message to the channel via restAPI call.
3. A user establish WS connections with multiple clients. Server send messages to one or many clients.

jac-cloud/jac_cloud/tests/jac-cloud-websocket.insomnia
ypkang, line 1: I don't think we should commit the insomnia collection.

jac-cloud/jac_cloud/jaseci/security/__init__.py
```diff
@@ -104,4 +104,24 @@ def authenticate(request: Request) -> None:
     raise HTTPException(status_code=401)
 
 
+def authenticate_websocket(websocket: WebSocket) -> bool:
+    """Authenticate websocket connection."""
+    if (
+        authorization := websocket.headers.get("Authorization")
+    ) and authorization.lower().startswith("bearer"):
+        token = authorization[7:]
```
ypkang, line 188: a nitpick: I'd use `len("bearer")` instead of `7`. A bit more readable.

jac-cloud/jac_cloud/tests/runner.py
```diff
@@ -0,0 +1,44 @@
+"""Sample Runner."""
```
ypkang, line 1: Hmm is this a helper script? do we need to commit this?

ypkang (changes requested):

ypkang:
> > Also, I think we are going to need a different type of documentation on this. We need to include end-to-end examples, like what fastAPI has https://fastapi.tiangolo.com/advanced/websockets/#await-for-messages-and-send-messages.
> > We need maybe 3 common use case:
> > ```
> > 1. A user establishing a connection with server and continuously exchanging messages
> > 
> > 2. Multiple users (unauth) establish connections via a specific channel id and listening on messages. An authenticated user broadcast a message to the channel via restAPI call.
> > 
> > 3. A user establish WS connections with multiple clients. Server send messages to one or many clients.
> > ```
> 
> Agree, tho the sample insomnia export has the complete use case including the request examples. Might need to convert it to docs type

Yeah but the insomnia doesn't include the server side implementation.

jac-cloud/jac_cloud/tests/runner.py
amadolid: This might be good example for deploying using uvicorn

jac-cloud/jac_cloud/tests/jac-cloud-websocket.insomnia
```diff
@@ -0,0 +1,1164 @@
+_type: export
```
ypkang: @amadolid lets not commit this. I think it is a good reference but i don't know if it belongs to the repo. You can upload it somewhere else (e.g. google drive) and add a link to it in the docs.

jac-cloud/setup.py
```diff
@@ -34,6 +34,7 @@
         "ecs-logging~=2.2.0",
         "types-PyYAML~=6.0.12.20240917",
         "montydb~=2.5.3",
+        "broadcaster[redis]~=0.3.1",
```
ypkang, line 37: Does websocket work with fake-redis?

amadolid: No, since it's only in runtime memory

jac-cloud/jac_cloud/tests/jac-cloud-websocket.insomnia
amadolid: Removed

ypkang (approved):

ypkang:
I am okay with merging this @amadolid. I'd like us to do some serious testing with websockets. I know zsb uses it and we are starting to migrate it to jaclang so maybe we can start with the streaming part of the implementation. Let's discuss @amadolid @asimsalim

PR #1420, opened 2024-11-01
Absorb py ast

mgtm98:
## **Description**
In this PR, I enhanced the PyImportPass to be more modular and enabled import_from functionality under an ENV until it's complete. I also added some debugging messages that can be used to debug the pass.


mgtm98:
We concluded that I need to create another branch from this branch and start to work on part2 then after part2 we need to merge the PRs with each other

marsninja (approved):
After my last commit I'm fine to merge, we should follow up as discussed in slack. Good work overall.

PR #1422, opened 2024-11-02
definition and declaration can have different parameter names

ThakeeNathees:
```py

obj SomeClass {
    can method1(some_long_paramenter_name: int) -> None;
    can method2(param1:int, param2:str) -> None;
}

:o:SomeClass:c:method1(short_name:int) -> None {
  print("short_name =", short_name);
}

:o:SomeClass:c:method2(p1:int, p2: str) -> None {
  print("p1 =", p1, ", p2 =", p2);
}

with entry {
  sc = SomeClass();
  sc.method1(42);
  sc.method2(64, "foobar");
}
```


marsninja (approved):
LGTM!

Issue #1400, opened 2024-10-22
Bug:  Multi Dimensional Array Index Slicing Not Supported


kugesan1105:
### Description
When using multidimensional arrays in Jaclang with the numpy library, it seems that slicing with multi-dimensional indices (e.g., x[:, 1:]) results in a syntax error. This syntax should allow for slicing all rows and selecting columns from a specified index to the end, but it is currently not supported.
```
import:py numpy as np ;

with entry {
    x = np.array(
        [[1, 2, 3, 4, 5],
         [6, 7, 8, 9, 10],
         [11, 12, 13, 14, 15],
         [16, 17, 18, 19, 20],
         [21, 22, 23, 24, 25]
        ]
    );    

    # Currently not supported
    # result = x[:, 1:];  # Causes grammar error

    result = x[1:4:2];

    print('Original Array:');
    print(x);
    print('\nSliced Array (all rows, columns 1 to end):');
    print(result);
}
```

#### Expected Behavior
The code should successfully slice the array by selecting all rows and columns starting from the second column.
```
Sliced Array (all rows, columns 1 to end):
[[ 2  3  4  5]
 [ 7  8  9 10]
[12 13 14 15]] 
```

#### Actual Behavior
The code throws a syntax error, indicating that the multi dimensional array slicing syntax x[:, 1:] is not supported.

PR #1436, opened 2024-11-11
multi dim array split syntax support

ThakeeNathees:
## **Description**
FIX: #1400

### DIFF
--- a/jac/jaclang/compiler/absyntree.py
+++ b/jac/jaclang/compiler/absyntree.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import ast as ast3
 import builtins
 import os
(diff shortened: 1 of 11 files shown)


ThakeeNathees:
@kugesan1105 The test is failing as to `numpy` is not available in the test env. Installing might resolve the issue

kugesan1105:
> @kugesan1105 The test is failing as to `numpy` is not available in the test env. Installing might resolve the issue

@ThakeeNathees   To validate the syntax without actually running the code, you can use the `jac_file_to_pass` function to check with the parser.

kugesan1105 (approved):
Good to go from my end!

PR #1473, opened 2024-11-28
[FEATURE]: JID

amadolid:


amadolid:
BACKUP
https://github.com/amadolid/jaseci/tree/feature/jid-backup

jac/jaclang/plugin/tests/fixtures/other_root_access.jac
```diff
@@ -67,15 +66,15 @@ walker allow_other_root_access {
         if self.via_all {
             Jac.unrestrict(here, self.level);
         } else {
-            Jac.allow_root(here, UUID(self.root_id), self.level);
+            Jac.allow_root(here, JacLangJID(self.root_id), self.level);
```
ypkang, line 69: i wonder if we can somehow not having separate handling for jid between jaclang and jac-cloud. It would be cool if the jac code is the same.

jac/jaclang/runtimelib/architype.py
```diff
@@ -3,18 +3,110 @@
...
-from typing import Any, Callable, ClassVar, Optional, TypeVar
+from typing import Any, Callable, ClassVar, Generic, Optional, Type, TypeVar
 from uuid import UUID, uuid4
 
 logger = getLogger(__name__)
 
-TARCH = TypeVar("TARCH", bound="Architype")
-TANCH = TypeVar("TANCH", bound="Anchor")
+
+JID_REGEX = compile(
```
ypkang, line 18: I'd add a docstring here with an example jid.

jac/jaclang/plugin/tests/fixtures/other_root_access.jac
ypkang: for example, we define a plugin method that converts string to the corresponding JID type

jac/jaclang/runtimelib/architype.py
```diff
@@ -60,100 +152,41 @@ class Permission:
...
 
-@dataclass(eq=False, repr=False, kw_only=True)
+@dataclass(eq=False, kw_only=True)
 class Anchor:
     """Object Anchor."""
 
     architype: Architype
-    id: UUID = field(default_factory=uuid4)
-    root: Optional[UUID] = None
+    id: Any = field(default_factory=uuid4)
```
ypkang, line 164: Can we be more specific on type here and avoid `Any`?

jac/jaclang/runtimelib/architype.py
```diff
@@ -163,59 +196,34 @@ def report(self) -> AnchorReport:
...
-        state = super().__getstate__()
-
-        if self.is_populated():
-            state["edges"] = [edge.make_stub() for edge in self.edges]
+    edges: list[JID["EdgeAnchor"]]
 
-        return state
 
-
-@dataclass(eq=False, repr=False, kw_only=True)
```
ypkang: Shouldn't we keep `repr=False`? since we are defining a custom __repr__

jac/jaclang/runtimelib/architype.py
```diff
@@ -163,59 +196,34 @@ def report(self) -> AnchorReport:
...
+@dataclass(eq=False, kw_only=True)
 class EdgeAnchor(Anchor):
     """Edge Anchor."""
 
     architype: EdgeArchitype
-    source: NodeAnchor
-    target: NodeAnchor
+    source: JID["NodeAnchor"]
+    target: JID["NodeAnchor"]
```
ypkang, line 222: Why is this changed?

ypkang (commented):
Is this ready for review and merge @amadolid? I went through it and i think there are a couple of things we need to discuss.

jac/jaclang/runtimelib/architype.py
amadolid: the reference now is only the ID instead of actual NodeAnchor

jac/jaclang/runtimelib/architype.py
amadolid: If I remember correctly, it will be only generated if you haven't specified `__repr__`

jac/jaclang/runtimelib/architype.py
amadolid: Currently, the ID could be anything as it's also technically a read only field

PR #1477, opened 2024-12-03
jaclib

ThakeeNathees:


amadolid:
Hello @ThakeeNathees ,

Would like to know if this also consider the previous PR for decluttering architypes.
The idea is, as much as possible architypes should not holds any default functions/attributes (private ones might be debatable). The reason for it is, architypes are like "frontend" class representation on jac lang. Dev can add any implementation there. Adding hidden default functions/attributes might got overridden by devs as we don't tagged it as reserved keywords.

In my perspective, architypes should be just object holders (models), JacFeature is the processor (services) and API or CLI is the controller

cc: @ypkang , @marsninja

PR #1490, opened 2024-12-16
Add CI/CD Pipeline for jac-splice-orc

AshishMahendra:
## **Description**
This PR introduces a new CI/CD pipeline for jac-splice-orc. It adds a release workflow triggered.


jac-splice-orc/setup.py
```diff
@@ -21,11 +29,17 @@
         "requests",
         "python-dotenv",
         "numpy",
+        "jaclang",
```
ypkang: We should add version requirements range for these packages.  Take a look at the setup.py in jac-cloud as reference.

ypkang (changes requested):

ypkang (approved):

PR #1493, opened 2024-12-20
Supporting getting symbols for Jac object

mgtm98:
## **Description**

This PR adds support for recognizing and linking `Jac built-in` functions like jid and dotgen.
It also enables "`Go to Definition`" functionality — for example, when the user clicks on `dotgen`, it should navigate to its definition in `jac_builtins.py`.
image

### DIFF
--- a/jac/jaclang/langserve/engine.py
+++ b/jac/jaclang/langserve/engine.py
@@ -178,8 +178,9 @@ class JacLangServer(LanguageServer):
         current_line = document.lines[position.line]
         current_pos = position.character
         current_symbol_path = parse_symbol_path(current_line, current_pos)
-        builtin_tab = mod_ir.sym_tab.kid[-1]
-        completion_items = []
+
+        # ['jac_features', 'jac_builtins', 'builtins']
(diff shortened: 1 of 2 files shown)


kugesan1105:
@mgtm98 
From the latest changes, we now have type information for Jac built-ins. However, it currently requires using `_Jac` or `Jac`

For example
```
obj_id = _Jac.jid(x);
obj_id = Jac.jid(x);
  ```
    
image
It would be great if we could support type information for Jac built-ins directly, without needing the _Jac or Jac prefix.

**Review by kugesan1105 — CHANGES_REQUESTED** [review]
@mgtm98 
From the latest changes, we now have type information for Jac built-ins. However, it currently requires using `_Jac` or `Jac`

For example
```
obj_id = _Jac.jid(x);
obj_id = Jac.jid(x);
  ```
    
image
It would be great if we could support type information for Jac built-ins directly, without needing the _Jac or Jac prefix.

kugesan1105 (changes requested):
image
@mgtm98  I think there is a confusion between `dotgen` and `Jac.node_dot`

PR #1505, opened 2025-01-01
Cache builtins

Thamirawaran:
## **Description**
1. Caching Builtins.pyi file 
2. Test case added

### DIFF
--- a/jac/jaclang/compiler/passes/main/import_pass.py
+++ b/jac/jaclang/compiler/passes/main/import_pass.py
@@ -13,7 +13,7 @@ from typing import Optional
 
 
 import jaclang.compiler.absyntree as ast
-from jaclang.compiler.constant import SymbolType
+from jaclang.compiler.constant import Constants, SymbolType
 from jaclang.compiler.passes import Pass
 from jaclang.compiler.passes.main import DefUsePass, SubNodeTabPass, SymTabBuildPass
 from jaclang.compiler.passes.main.sym_tab_build_pass import PyInspectSymTabBuildPass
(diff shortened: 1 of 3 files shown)


jac/jaclang/compiler/passes/main/import_pass.py
```diff
@@ -463,27 +464,38 @@ def __load_builtins(self) -> None:
...
-        file_to_raise = str(
-            pathlib.Path(os.path.dirname(__file__)).parent.parent.parent
-            / "vendor"
-            / "mypy"
-            / "typeshed"
-            / "stdlib"
-            / "builtins.pyi"
+        base_dir = pathlib.Path(os.path.dirname(__file__)).parent.parent.parent
+        jac_gen_dir = (
+            base_dir / "vendor" / "mypy" / "typeshed" / "stdlib" / "__jac__gen__"
```
ThakeeNathees: `__jac_gen__` should be referenced from constants

jac/jaclang/compiler/passes/main/import_pass.py
```diff
@@ -463,27 +464,38 @@ def __load_builtins(self) -> None:
...
-                input_ir=ast.PythonModuleAst(
-                    py_ast.parse(file_source),
-                    orig_src=ast.JacSource(file_source, file_to_raise),
-                ),
-            ).ir
-            mod.parent = self.ir
-            SubNodeTabPass(input_ir=mod, prior=self)
-            SymTabBuildPass(input_ir=mod, prior=self)
-            mod.parent = None
+        mod_file_path = jac_gen_dir / "builtins_mod.jbc"
```
ThakeeNathees: Why the extension of the file is called `.jbc` as it shows it might be a jac byte code file but it's a pickel dump? shouldn't it be a `.pkl` or `.dump` file?

ThakeeNathees (changes requested):
LGTM but there are some nitpicks i've left, @Thamirawaran can you have a look

marsninja:
Hi @Thamirawaran This will likely break when it's in a pypi package since doing a pip install from pypi makes the library dir read only. So on trying to create the dir and pkl file, it will have a read only error. I believe settings.py use home directory. Also is this for perforamance, if so did you measure how much faster it makes things?

PR #1509, opened 2025-01-03
FIX : TYPE_OP Issue

kugesan1105:

### DIFF
--- a/jac/jaclang/plugin/feature.py
+++ b/jac/jaclang/plugin/feature.py
@@ -201,6 +201,34 @@ class JacClassReferences:
     Edge: ClassVar[TypeAlias] = EdgeArchitype
     Walker: ClassVar[TypeAlias] = WalkerArchitype
 
+    @staticmethod
+    def type(type_name: str) -> Type[Architype]:
+        """Get the type of the object."""
+        type_map = {
+        "node": JacNode,
+        "walker": JacWalker,
+        "edge": JacEdge
+        }
+        import inspect
+        import re
+        try:
+            source = inspect.getsource(type(type_name))
+            lines = source.splitlines()
+            decorators = [line.strip() for line in lines if line.strip().startswith("@")]
+            for decorator in decorators:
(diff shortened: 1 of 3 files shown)


jac/jaclang/plugin/feature.py
```diff
@@ -201,6 +201,23 @@ class JacClassReferences:
     Edge: ClassVar[TypeAlias] = EdgeArchitype
     Walker: ClassVar[TypeAlias] = WalkerArchitype
 
+    @staticmethod
+    def type(instance: object) -> Type[Architype] | None:
```
ThakeeNathees: I would say rename this method to `archi_type_of(instance)` for more readability.

jac/jaclang/tests/fixtures/type_op.jac
```diff
@@ -0,0 +1,51 @@
...
+with entry {
+    print(`root); # This prints the type of root which is Root
+    print(`n); # This prints the type of n which is NodeArchitype
+    print(`m); # This prints the type of m which is WalkerArchitype
+    print(`e); # This prints the type of e which is EdgeArchitype
+    print(`o); # This prints the type of o which is Architype
+    `List;
+
+}
+
```
ThakeeNathees, line 20:
```c++
walker test_walker {
    can traverse with `n entry {
        visit [-->];
    }
}
```
Does the above test work? in that case add it. Wonder if a runtime type can be used as type annotation.

ThakeeNathees (changes requested):

jac/jaclang/tests/fixtures/type_op.jac
kugesan1105: Yes, @ThakeeNathees, the test works, and I’ve just added a test case to cover it. Thanks for pointing it out!

jac/jaclang/plugin/feature.py
kugesan1105: done 👍

PR #1511, opened 2025-01-06
Reuse previous build

Thamirawaran:
## **Description**
 Dump ast of each imported files in pkl format and reuse if that certain file or its dependencies aren't modified

### DIFF
--- a/jac/jaclang/compiler/passes/main/import_pass.py
+++ b/jac/jaclang/compiler/passes/main/import_pass.py
@@ -215,9 +215,7 @@ class JacImportPass(Pass):
                 os.makedirs(cache_dir, exist_ok=True)
                 module_name = os.path.splitext(os.path.basename(target))[0]
                 cache_path = os.path.join(cache_dir, f"{module_name}.pkl")
-                mod = self.load_cached_module(
-                     cache_path, os.path.getmtime(target)
-                )
+                mod = self.load_cached_module(cache_path, os.path.getmtime(target))
             else:
                 mod = None
             if not mod:
--- a/jac/jaclang/tests/fixtures/cache_ast.jac
+++ b/jac/jaclang/tests/fixtures/cache_ast.jac
@@ -1,0 +1,7 @@
+import:py cache_pyast;
+import:jac cache_jacast;
+
+with entry{
+    print(cache_jacast.add(1,2));
+    print(cache_pyast.add(1,2));
+}
--- a/jac/jaclang/tests/fixtures/cache_jacast.jac
+++ b/jac/jaclang/tests/fixtures/cache_jacast.jac
@@ -1,0 +1,1 @@
+can add(a:int,b:int)->int{return a + b + 12 ;}
--- a/jac/jaclang/tests/fixtures/cache_pyast.py
+++ b/jac/jaclang/tests/fixtures/cache_pyast.py
@@ -1,0 +1,5 @@
+# flake8: noqa
+
+
+def add(a: int, b: int) -> int:
+    return a + b + 23
--- a/jac/jaclang/tests/fixtures/test_cache_ast.jac
+++ b/jac/jaclang/tests/fixtures/test_cache_ast.jac
@@ -1,7 +1,0 @@
-import:py test_cache_pyast;
-import:jac test_cache_jacast;
-
-with entry{
-    print(test_cache_jacast.add(1,2));
-    print(test_cache_pyast.add(1,2));
-}
--- a/jac/jaclang/tests/fixtures/test_cache_jacast.jac
+++ b/jac/jaclang/tests/fixtures/test_cache_jacast.jac
@@ -1,3 +1,0 @@
-can add(a:int,b:int)->int{
-    return a+b;
-}
--- a/jac/jaclang/tests/fixtures/test_cache_pyast.py
+++ b/jac/jaclang/tests/fixtures/test_cache_pyast.py
@@ -1,6 +1,0 @@
-"""Module for testing simple addition functionality."""
-
-
-def add(a: int, b: int) -> int:
-    """Add two integers."""
-    return a + b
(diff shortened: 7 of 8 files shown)


jac/jaclang/compiler/passes/main/import_pass.py
```diff
@@ -427,15 +472,19 @@ def __import_py_module(
...
-                    input_ir=ast.PythonModuleAst(
-                        py_ast.parse(file_source),
-                        orig_src=ast.JacSource(file_source, file_to_raise),
-                    ),
-                ).ir
-                SubNodeTabPass(input_ir=mod, prior=self)
+            if os.path.exists(cache_path):
+                mod = self.load_cached_module(file_to_raise, cache_path)
+            else:
+                with open(file_to_raise, "r", encoding="utf-8") as f:
```
kugesan1105, line 483: @Thamirawaran  
Add a timestamp check for Python imports to ensure the cached AST is not outdated, similar to the logic used for Jac imports, as it fails to load updated Python imports when a changed file is imported
<br>
image

Thamirawaran: It is already there. Will see what makes error

jac/jaclang/compiler/passes/main/import_pass.py
```diff
@@ -169,28 +170,68 @@ def import_jac_mod_from_dir(self, target: str) -> ast.Module | None:
...
+                pickle.dump(mod, cache_file, protocol=5)
+        except Exception as e:
+            logger.error(f"Failed to cache module {mod.name}: {e}")
+
     def import_jac_mod_from_file(self, target: str) -> ast.Module | None:
         """Import a module from a file."""
         from jaclang.compiler.compile import jac_file_to_pass
         from jaclang.compiler.passes.main import SubNodeTabPass
 
+        cache_dir = os.path.join(os.path.dirname(self.ir.loc.mod_path), "__jac_gen__")
```
ThakeeNathees: `__jac_gen__` is defined in the constants, use it instead of hardcoding, since we may rename the cach dir path in the future and everything should be in sync

jac/jaclang/compiler/passes/main/import_pass.py
```diff
@@ -169,28 +170,68 @@ def import_jac_mod_from_dir(self, target: str) -> ast.Module | None:
...
+        try:
+            with open(cache_path, "wb") as cache_file:
+                pickle.dump(mod, cache_file, protocol=5)
+        except Exception as e:
+            logger.error(f"Failed to cache module {mod.name}: {e}")
+
     def import_jac_mod_from_file(self, target: str) -> ast.Module | None:
         """Import a module from a file."""
         from jaclang.compiler.compile import jac_file_to_pass
         from jaclang.compiler.passes.main import SubNodeTabPass
```
ThakeeNathees: Why importing inside a function? Isn't this go to the top of this file?

jac/jaclang/compiler/passes/main/import_pass.py
```diff
@@ -399,6 +440,10 @@ def __import_py_module(
 
         assert isinstance(self.ir, ast.Module)
 
+        cache_dir = os.path.join(os.path.dirname(self.ir.loc.mod_path), "__jac_gen__")
```
ThakeeNathees: Same here, `__jac_gen__` needs to be referenced from the constants file

jac/jaclang/compiler/passes/main/import_pass.py
```diff
@@ -169,28 +170,68 @@ def import_jac_mod_from_dir(self, target: str) -> ast.Module | None:
...
+    def load_cached_module(
+        self, source_path: str, cache_path: str
+    ) -> ast.Module | None:
+        """Load cached module if up-to-date."""
+        if not os.path.exists(cache_path):
+            return None
+
+        try:
+            source_mtime = os.path.getmtime(source_path)
+            cache_mtime = os.path.getmtime(cache_path)
```
ThakeeNathees: we don't check the `mtime` of the cached file instead of this store the mtime of the source file within the pickle dump and check with the current `mtime` of the source file

jac/jaclang/compiler/passes/main/import_pass.py
```diff
@@ -169,28 +170,68 @@ def import_jac_mod_from_dir(self, target: str) -> ast.Module | None:
...
+                return cached_module
+        except Exception as e:
+            logger.error(f"Failed to load cached module {cache_path}: {e}")
+            return None
+
+    def cache_ast(self, mod: ast.Module, cache_path: str) -> None:
+        """Cache the module."""
+        try:
+            with open(cache_path, "wb") as cache_file:
+                pickle.dump(mod, cache_file, protocol=5)
```
ThakeeNathees: this should be something like:
```py
dump({
  'mtime' : source_mtime,
  'mod' : mod,
})
```

ThakeeNathees (dismissed):

marsninja (changes requested):
Can you make this based on a setting to add to settings.py

Thamirawaran:
> Can you make this based on a setting to add to settings.py

Yeah

jac/jaclang/compiler/passes/main/import_pass.py
Thamirawaran: > Why importing inside a function? Isn't this go to the top of this file?

We can't do this because of circular import

Thamirawaran:
**Current method**
Currently we are dumping python ast of our code in execution format. So that if original code doesn't modified, we can directly execute it. But if it is modified, with my knowledge, we cannot use or modify it. We cannot even retrieve the pyast.

**Build method**
In build method, we are building the ir file by executing all the passes including type check. But it is not efficient to build every time before executing the file.

**Proposed method**
Dumping the ast in pickle file, and retrieve when if a file is not modified. When it comes to imports it is quite useful because if an imported file isn't modified, then we simply load the ast of the file and connect it to the main ast.

There may be some misunderstandings. @marsninja Let me know if there is any. and Could you please guide to make it?

PR #1514, opened 2025-01-14
[CONFIG]: Add custom config merging and Update image version and 

AshishMahendra:
## **Description**
**Summary**:
- Updated the `jac-splice-orc` plugin to support custom configuration merging.
- Enhanced the `ConfigLoader` to handle default, loaded, and custom configurations with `deepmerge`.
- Modified `orc_initialize` to accept a custom config path.
- Updated dependencies to include `deepmerge`.
- Updated image version in `config.json` and aligned the `POD_MANAGER_URL` for local use.


jac-splice-orc/requirements.txt
```diff
@@ -5,3 +5,4 @@ protobuf==4.23.4
 pluggy==1.2.0
 jaclang
 numpy
+deepmerge==2.0
```
ypkang: Let's implement it ourselves without adding this requirement.

ypkang: actually, we shouldn't try to merge the configs. We should just use the config provided by the user OR the default one, as a whole.

ypkang (changes requested):

marsninja (approved):

PR #1519, opened 2025-01-17
[APPROACH #2][ACCESS-CHANGE]: Prioritize sequence access level permission

amadolid:
1. roots access config of target node
2. roots access config of target node's root
3. public (all) access config of target node's root
4. public (all) access config of target node


ypkang (approved):
LGTM

PR #1522, opened 2025-01-25
[Lark-PR-1] lark tokens are grouped

ThakeeNathees:
## **Description**

* Lark reference heading line should starts with `// [Heading]:`. Now we can write our own comments without messing with the reference generator.

* Terminals / Tokens are the lego blocks of a language (ast) so I've grouped them together in a meaningful way.


mgtm98 (approved):
I checked the changes and I don't see any new lines or deleted items just some grouping for the tokens. 
All tests are passing

marsninja (approved):
LGTM

PR #1523, opened 2025-01-26
[Lark-PR-2] Module rule simplified

ThakeeNathees:
## Depends on Lark-PR-1 #1522

* Both STRING and DOC_STRING are now strings in terms of literal values
* `module` rule in lark is simplified.

### DIFF
--- a/jac/jaclang/compiler/jac.lark
+++ b/jac/jaclang/compiler/jac.lark
@@ -1,13 +1,21 @@
 // [Heading]: Base Module structure.
 start: module
 
-module: (doc_tag? element (element_with_doc | element)*)?
-      | doc_tag (element_with_doc (element_with_doc | element)*)?
+// TODO:AST: We should allow the very first top level statement to have
+// docstring but not the module to have one. After allowing that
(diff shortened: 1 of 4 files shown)


jac/jaclang/compiler/jac.lark
```diff
@@ -604,10 +612,11 @@ NOT: "not" // TODO:AST: Rename to KW_NOT
 
 // Literals ---------------------------------------------------------------- //
 
+STRING: /(r?b?|b?r?)"[^"\r\n]*"|(r?b?|b?r?)'[^'\r\n]*'/
```
mgtm98: (r?b?|b?r?)("[^"\r\n]*"|'[^'\r\n]*')

mgtm98 (dismissed):
Other than the comment added before (refactoring a regex) this change looks fine for me

jac/jaclang/compiler/jac.lark
ThakeeNathees: DONE

marsninja (approved):

PR #1524, opened 2025-01-26
[FEATURE]: Custom Access Validation

amadolid:
This is to support complex access validation. Since we can now do anything, dev should be able to implement/integrate access handler from other frameworks/library or at least replicate it.

!!! The only limitation right now is, we only validates it from current root. We don't implement "from" node yet. This means on `__jac_access__` you don't have direct reference of the node you are from. Please let me know if we still need to support that

# Scenario 1 - shared node permission
- access can now depends on attributes
  Ex:
```
node bot {
    has allowed_admins: list[str] = [];
    
    def __jac_access__ {
        current_root = Jac.get_root();

        if jid(current_root) in self.allowed_admins {
            return "WRITE"
        }
        return "NO_ACCESS"
    }
}
```

# Scenario 2 - Organization Group permission
- We can also query directly to DB
  Ex:
```
node bot {
    has allowed_admins: list[str] = [];
    
    def __jac_access__ {
        profile = NodeAnchor.Collection.find_one({"name": "user_profile", root: `root.__jac__.id}).architype;
        if profile.group == "GROUP 1" {
            return "WRITE"
        }
        return "NO_ACCESS"
    }
}
```

# OTHERS
- We can also optionally cache the final access level directly to instance variable
  Ex:
```python
def __jac_access__ {
    if level := getattr(self, "variable_for_your_cache", None):
        return level;
    # do your process here and save to cache
}
```

### DIFF
--- a/docs/docs/communityhub/release_notes.md
+++ b/docs/docs/communityhub/release_notes.md
@@ -6,6 +6,7 @@ This document provides a summary of new features, improvements, and bug fixes in
 
 - **Support Spawning a Walker with List of Nodes and Edges**: Introduced the ability to spawn a walker on a list of nodes and edges. This feature enables initiating traversal across multiple graph elements simultaneously, providing greater flexibility and efficiency in handling complex graph structures.
 - **\_.save(...) should not override root in runtime**: The previous version bypassed access validation because the target archetype root was overridden by the current root, simulating ownership of the archetype.
+- **Support Custom Access Validation**: Introduced the ability to override access validation. `Node`/`Edge` can override `__jac_access__` reserved function (`builtin`) to have a different way of validating access. Either you cross-check it by current attribute, check from db or global vars or by just returning specific access level. [PR#1524](https://github.com/jaseci-labs/jaseci/pull/1524)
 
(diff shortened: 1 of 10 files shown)


marsninja (changes requested):
Should have these be builtin function or functions @amadolid

amadolid:
> Should have these be builtin function or functions @amadolid.

what do you mean by builtin ?

Currently, it's a class level function customization. The trigger is builtin and the `__jac_access__` function is reserved.

marsninja (changes requested):
See previous comment.

amadolid:
Oww, this is not for manual access validation.

It's for overriding automatic access validation per specific archetype.

For example,
I have a node Book with attribute section = "A"

I can set Read Access to that node Book by traversing from current root's graph which ever node has it's section

If it's both "A", I'll return read access else no access

This is for more complex access validation like account, group, organization permissions

amadolid:
I think, scenario 1 & 2 are the most common use case

amadolid:
Littlex will benefit from this because currently, we always manually need to trigger `_.perm_grant(READ)` to public a node

But if we use this approach, we can use a specific field to enable/disable access by just changing its value or even just statically return READ access to make it public

PR #1534, opened 2025-01-30
Thakee lark 3

ThakeeNathees:
## **Description**

`SpecialVarRef` was removed and merged with `Name`

In terms of function and use, there are almost no difference between name and special var ref (almost every where the logic is the same and copied), now they're merged.


jac/jaclang/compiler/absyntree.py
```diff
@@ -4146,11 +4144,20 @@ def __init__(
         NameAtom.__init__(self)
         AstSymbolNode.__init__(
             self,
-            sym_name=value,
+            sym_name=self.py_resolve_name(),
             name_spec=self,
             sym_category=SymbolType.VAR,
         )
 
+    def py_resolve_name(self) -> str:
```
mgtm98, line 4162: Can you please change it to py_resolve_special_name

jac/jaclang/compiler/passes/main/def_use_pass.py
```diff
@@ -160,13 +160,6 @@ def enter_index_slice(self, node: ast.IndexSlice) -> None:
         is_range: bool,
         """
 
-    def enter_special_var_ref(self, node: ast.SpecialVarRef) -> None:
```
mgtm98, line 163: Can we have old SpecialVarRef to have a parent of atom trailer?

mgtm98 (commented):
I checked these changes. There is a comment I added. @marsninja  please check it

jac/jaclang/compiler/absyntree.py
ThakeeNathees: The name `py_resolve_name` is also used in ability, and archrefchain, also this PR is to remove special name, I wouldn't suggest adding **special** in the name.

Also in future (till the AST refactored) this special cases were removed and handled differently, So this function is temporary.

jac/jaclang/compiler/jac.lark
```diff
@@ -1,21 +1,29 @@
...
-element_with_doc: doc_tag element
-
-element: import_stmt
+// TODO:AST: We should allow the very first top level statement to have
+// docstring but not the module to have one. After allowing that
+// the rule for the module would be:
+//
+// module: STRING? toplevel_stmt*
+//
+module: (toplevel_stmt (tl_stmt_with_doc | toplevel_stmt)*)?
```
Thamirawaran: `module: STRING? (STRING? toplevel_stmt)*`
It could replace 
```
module: (toplevel_stmt (tl_stmt_with_doc | toplevel_stmt)*)?
       | STRING        (tl_stmt_with_doc | toplevel_stmt)*
tl_stmt_with_doc: STRING toplevel_stmt
```
So that we can get rid of `tl_stmt_with_doc` node in parser
If you want to keep that node
we can go with
```
module: STRING? tl_stmt_with_doc*
tl_stmt_with_doc: (STRING? toplevel_stmt)
```

Thamirawaran (changes requested):

jac/jaclang/compiler/jac.lark
ThakeeNathees: We can't do that cause there is ambiguity since we have `STRING? STRING?` and the lark reduce doesn't know if the first string should be module dostring or class docstring,

Thamirawaran: Yeah
Let's use this
```
module: STRING? tl_stmt_with_doc
tl_stmt_with_doc: (STRING? toplevel_stmt)*
```
I hope it fixes

PR #1540, opened 2025-01-31
[MINOR]: Additional API Configuration for documentations

amadolid:
```python
walker ask_question {
    class __specs__ {
        has allow_entry: bool = False, tags: list[str] = ["Chat APIs"];
    }
}
```

# `tags` 
- customization of tags to segregate OpenAPI specifications/docs
image

# `entry_type`
- "NODE" for generating api with `/{node}` entry input. "ROOT" for generating api without node entry input (always current root entry). "BOTH" will support both.

## entry_type="NODE"
image
## entry_type="ROOT"
image
## entry_type="BOTH" or not set
image

# Sample Documentations
```python
walker documentation {

    class __specs__ {
        has tags: list[str] | None = ["Documented"];
        has entry_type: EntryType = "root";
        has status_code: int | None = "201";
        has summary: str | None = "Enter summary here!!!";
        has description: str | None = "Enter description here!!!";
        has response_description: str = "Enter response description here!!! Defaults to 'Successful Response'";
        has responses: dict[int | str, dict[str, Any]] | None = {
            200: {
                "description": "Enter response description for http status 200!!!",
                "content": {"application/json": {}},
                "links": {"sample_link": {"description": "Enter link description here!!!"}}
            }
        };
        has deprecated: bool | None = True;
        has name: str | None = "Enter name here!!!";
        has openapi_extra: dict[str, Any] | None = {"addons": True};
    }
}
```
image

### DIFF
--- a/jac-cloud/jac_cloud/plugin/implementation/api.py
+++ b/jac-cloud/jac_cloud/plugin/implementation/api.py
@@ -1,6 +1,7 @@
 """Jaseci API Implementations."""
 
 from dataclasses import Field, MISSING, fields, is_dataclass
+from enum import StrEnum
 from os import getenv
 from re import compile
 from types import NoneType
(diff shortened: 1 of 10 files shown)


jac-cloud/jac_cloud/plugin/implementation/api.py
```diff
@@ -56,6 +56,8 @@ class DefaultSpecs:
     auth: bool = True
     private: bool = False
     webhook: dict | None = None
+    tags: list[str] | None = None
+    allow_entry: bool | None = None
```
ypkang: Two quick ones:
- I'd name this `allow_node_entry` 
- can we default this to True instead of None?

ypkang (changes requested):

jac-cloud/jac_cloud/plugin/implementation/api.py
ypkang: rename this to entry_type and the value should be an enum for ROOT, NODE, BOTH

ypkang (approved):
LGTM

PR #1542, opened 2025-01-31
[ FIX]: Nested Impls

kugesan1105:
## **Description**
image

### DIFF
--- a/jac/jaclang/compiler/parser.py
+++ b/jac/jaclang/compiler/parser.py
@@ -3237,6 +3237,15 @@ class JacParser(Pass):
                     )
                 )
             else:
+                if isinstance(kid[1], ast.ArchRef) and isinstance(
+                    kid[0], ast.ArchRefChain
+                ):
+                    return self.nu(
+                        ast.ArchRefChain(
+                            archs=[*kid[0].archs, kid[1]],
+                            kid=[*kid[0].kid, kid[1]],
+                        )
+                    )
                 raise self.ice()
 
         def abil_to_arch_chain(self, kid: list[ast.AstNode]) -> ast.ArchRefChain:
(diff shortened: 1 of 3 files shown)


jac/jaclang/compiler/parser.py
```diff
@@ -3237,6 +3237,15 @@ def arch_or_ability_chain(self, kid: list[ast.AstNode]) -> ast.ArchRefChain:
                     )
                 )
             else:
+                if isinstance(kid[1], ast.ArchRef) and isinstance(
```
ThakeeNathees: Not in this line but above, in this function we check kid[0] is subnode list, could you explain it? 
``
arch_or_ability_chain? (ability_ref | arch_ref)
```
here is the parsing rule and none of them transformed as a sbunode list. If this is wrong, would also get rid of that if block in this pr.

File: jac/jaclang/tests/fixtures/nested_impls.jac [hunk 2]
```diff
@@ -0,0 +1,55 @@
+obj mainObject;
+
+:obj:mainObject{
+    can greet;
+    obj subObject;
+}
+can foo{
+     obj kk;
+     return kk();
```
ThakeeNathees: I'd like the class, function and variables to clearely state what we're testing, so in this case:

```
obj ObjInsideFunction;
```

instead of `kk`

ThakeeNathees (changes requested):
All good, just left a question in the comments, could you explain the use of subnode list

jac/jaclang/compiler/parser.py
kugesan1105: You're right, 
we don't need to check that. I’ve modified it to check for` len(kid) == 2` instead

ThakeeNathees (approved):
LGTM

marsninja (approved):
LGTM

PR #1543, opened 2025-01-31
Modify Parser Functions for Imports, Architypes, and Enums

MalithaPrabhashana:
## **Description**
Modified So far
import_path
import_items
import_item
architype
architype_decl
enum
enum_decl
enum_stmt

### DIFF
--- a/jac/jaclang/compiler/parser.py
+++ b/jac/jaclang/compiler/parser.py
@@ -688,17 +688,17 @@ class JacParser(Pass):
             enum: decorators? enum_decl
                 | enum_def
             """
-            if isinstance(kid[0], ast.SubNodeList):
-                if isinstance(kid[1], ast.Enum):
-                    kid[1].decorators = kid[0]
-                    kid[1].add_kids_left([kid[0]])
-                    return kid[1]
-                else:
-                    raise self.ice()
-            elif isinstance(kid[0], (ast.Enum, ast.EnumDef)):
-                return kid[0]
+            decorator = self.match(ast.SubNodeList)
+            enum_node: ast.Enum | ast.EnumDef | None = None
+            if decorator is not None:
+                enum_node = self.consume(ast.Enum)
+                enum_node.decorators = decorator
+                enum_node.add_kids_left([decorator])
             else:
-
+                enum_node = self.match(ast.Enum) or self.match(ast.EnumDef)
+            if enum_node is not None:
+                return enum_node
+            else:
                 raise self.ice()
 
         def enum_decl(self, kid: list[ast.AstNode]) -> ast.Enum:
@@ -706,30 +706,22 @@ class JacParser(Pass):
 
             enum_decl: KW_ENUM access_tag? STRING? NAME inherited_archs? (enum_block | SEMI)
             """
-            access = kid[1] if isinstance(kid[1], ast.SubTag) else None
-            semstr = (
-                kid[2]
-                if (access and isinstance(kid[2], ast.String))
-                else kid[1] if isinstance(kid[1], ast.String) else None
+            self.node_idx = 1
+            access = self.match(ast.SubTag)
+            semstr = self.match(ast.String)
+            name = self.consume(ast.Name)
+            inh = self.match(ast.SubNodeList)
+            body = self.match(ast.SubNodeList)
+            if body is None:
+                body, inh = inh, None
+            return ast.Enum(
+                semstr=semstr,
+                name=name,
+                access=access,
+                base_classes=inh,
+                body=body,
+                kid=kid,
             )
-            name = (
-                kid[3]
-                if (access and semstr)
-                else kid[2] if (access or semstr) else kid[1]
-            )
-            inh = kid[-2] if isinstance(kid[-2], ast.SubNodeList) else None
-            body = kid[-1] if isinstance(kid[-1], ast.SubNodeList) else None
-            if isinstance(name, ast.Name):
-                return ast.Enum(
-                    semstr=semstr,
-                    name=name,
-                    access=access,
-                    base_classes=inh,
-                    body=body,
-                    kid=kid,
-                )
-            else:
-                raise self.ice()
 
         def enum_def(self, kid: list[ast.AstNode]) -> ast.EnumDef:
             """Grammar rule.
--- a/jac/jaclang/compiler/parser.py
+++ b/jac/jaclang/compiler/parser.py
@@ -494,7 +494,8 @@ class JacParser(Pass):
             """
             items = [self.consume(ast.ModuleItem)]
             while self.match_token(Tok.COMMA):
-                items.append(self.consume(ast.ModuleItem))
+                if module_item := self.match(ast.ModuleItem):
+                    items.append(module_item)
             ret = ast.SubNodeList[ast.ModuleItem](
                 items=items,
                 delim=Tok.COMMA,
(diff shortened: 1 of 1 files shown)


jac/jaclang/compiler/parser.py
```diff
@@ -510,8 +508,11 @@ def import_item(self, kid: list[ast.AstNode]) -> ast.ModuleItem:
 
             import_item: named_ref (KW_AS NAME)?
             """
-            name = kid[0]
-            alias = kid[2] if len(kid) > 1 else None
+            name = self.consume(ast.Name)
+            if self.match_token(Tok.KW_AS):
+                alias = self.consume(ast.Name)
+            else:
+                alias = None
```
ThakeeNathees: Change it to
```py
alias = self.match_token(Tok.KW_AS) and self.consume(ast.Name)
```

jac/jaclang/compiler/parser.py
```diff
@@ -510,8 +508,11 @@ def import_item(self, kid: list[ast.AstNode]) -> ast.ModuleItem:
...
             if isinstance(name, ast.Name) and (
                 alias is None or isinstance(alias, ast.Name)
             ):
```
ThakeeNathees: Remove this `if` since name is consumed as ast.Name and alias can be either None or ast.Name, so this check is not necessrary.

Also note that the function should never call `self.ice()` since the consume will take care of it.

jac/jaclang/compiler/parser.py
```diff
@@ -532,16 +533,22 @@ def architype(
...
+            if decorators is not None:
+                archspec = self.consume(ast.ArchSpec)
+                archspec.decorators = decorators
+                archspec.add_kids_left([decorators])
+            else:
+                archspec = (
+                    self.match(ast.ArchSpec)
+                    or self.match(ast.ArchDef)
+                    or self.match(ast.Enum)
+                    or self.match(ast.EnumDef)
```
ThakeeNathees: The last line should be 
```py
or self.consume(ast.EnumDef)
```

That way we can ensure that the archspec wil lnever be None, and remove `if archspec is not None:` and just return archspec. Also remove the `self.ice()` this shouldn't be called in these parsing functions.

jac/jaclang/compiler/parser.py
```diff
@@ -532,16 +533,22 @@ def architype(
...
+                )
+            if archspec is not None:
+                return archspec
             else:
                 raise self.ice()
```
ThakeeNathees: Remove these lines

jac/jaclang/compiler/parser.py
```diff
@@ -550,20 +557,17 @@ def architype_decl(self, kid: list[ast.AstNode]) -> ast.ArchSpec:
...
+            arch_type = self.consume(ast.Token)
+            access = self.match(ast.SubTag)
+            semstr = self.match(ast.String)
+            name = self.consume(ast.Name)
+            inh = self.match(ast.SubNodeList)
+            body = self.match(ast.SubNodeList)
+
+            # If there's no inherited arches, then the body is the last element.
+            if body is None:
+                body, inh = inh, None
```
ThakeeNathees: Subnode List will be removed, but for now could you re-write like this.
```py
sub_list1 = self.match(ast.SubNodeList);
sub_list2 = self.match(ast.SubNodeList);
if self.match_token(Tok.SEMI):
  inh, body = sub_list1, None
else:
  body = sub_list2 or sub_list1 # if sub_list2 is None then body is sub_list1
  inh = sub_list2 and sub_list1 # if sub_list2 is None then inh is None.
```

jac/jaclang/compiler/parser.py
```diff
@@ -680,48 +684,40 @@ def enum(self, kid: list[ast.AstNode]) -> ast.Enum | ast.EnumDef:
...
+                enum_node = self.consume(ast.Enum)
+                enum_node.decorators = decorator
+                enum_node.add_kids_left([decorator])
+            else:
+                enum_node = self.match(ast.Enum) or self.match(ast.EnumDef)
+            if enum_node is not None:
+                return enum_node
             else:
-
                 raise self.ice()
```
ThakeeNathees: This should be
```python
if decorator := self.match(ast.SbuNodeList):
    enum_decl = self.consume(ast.Enum)
    enum_decl.decorators = decorator
    enum_decl.add_kids_left([decorator])
    return enum_decl
return self.consume(ast.EnumDef)
```

jac/jaclang/compiler/parser.py
```diff
@@ -680,48 +684,40 @@ def enum(self, kid: list[ast.AstNode]) -> ast.Enum | ast.EnumDef:
...
-            semstr = (
-                kid[2]
-                if (access and isinstance(kid[2], ast.String))
-                else kid[1] if isinstance(kid[1], ast.String) else None
-            )
-            name = (
-                kid[3]
-                if (access and semstr)
-                else kid[2] if (access or semstr) else kid[1]
+            self.node_idx = 1
```
ThakeeNathees: You shouldn't modify `node_idx`, treat it like private and call `self.consume_token(Tok.KW_ENUM)`, that'll internally increment the node index (which will be gone in the future). Also it ensure the first token is keyword "enum".

jac/jaclang/compiler/parser.py
```diff
@@ -680,48 +684,40 @@ def enum(self, kid: list[ast.AstNode]) -> ast.Enum | ast.EnumDef:
...
+            access = self.match(ast.SubTag)
+            semstr = self.match(ast.String)
+            name = self.consume(ast.Name)
+            inh = self.match(ast.SubNodeList)
+            body = self.match(ast.SubNodeList)
+            if body is None:
+                body, inh = inh, None
```
ThakeeNathees: Use the same logic to get explained in the architype_decl above.

jac/jaclang/compiler/parser.py
```diff
@@ -760,51 +756,29 @@ def enum_stmt(self, kid: list[ast.AstNode]) -> ast.EnumBlockStmt:
...
-                    )
-            elif isinstance(kid[0], (ast.PyInlineCode, ast.ModuleCode)):
-                return kid[0]
+            if stmt := (
+                self.match(ast.PyInlineCode)
+                or self.match(ast.ModuleCode)
+                or self.match(ast.Ability)
+            ):
+                return stmt
+            elif name := self.consume(ast.Name):
```
ThakeeNathees: The above if always return so this `elif` is not necessary. remove the elif.

```py
if condition:
  return something()

elif name := consume(...):
  do_something()
```

can be written as:
```py
if condition:
  return something()

name = consume(...):
do_something()
```

jac/jaclang/compiler/parser.py
```diff
@@ -760,51 +756,29 @@ def enum_stmt(self, kid: list[ast.AstNode]) -> ast.EnumBlockStmt:
...
+                semstr = (
+                    self.consume(ast.String) if self.match_token(Tok.COLON) else None
+                )
```
ThakeeNathees: This should be
```py
semstr = self.match_token(Tok.COLON) and self.consume(ast.String)
```

jac/jaclang/compiler/parser.py
```diff
@@ -760,51 +756,29 @@ def enum_stmt(self, kid: list[ast.AstNode]) -> ast.EnumBlockStmt:
...
+                expr = self.consume(ast.Expr) if self.match_token(Tok.EQ) else None
```
ThakeeNathees: This should be

```py
expr = self.match_token(Tok.EQ) and self.consume(ast.Expr)
```

ThakeeNathees (changes requested):
Looks good still, I requested some changes, once fixed ready to merged.

ThakeeNathees (changes requested):
the function parameter should be
```py
def xxx(self, _:None): -> yyy
```

ThakeeNathees (approved):
LGTM

PR #1546, opened 2025-01-31
[Parser-Refactor] Modify Parser Functions for ability, if and try

MalithaPrabhashana:
## **Description**
Modified Parser Functions
- ability
- ability_decl
- abstract_ability
- genai_ability
- event_clause
- if_stmt
- elif_stmt
- else_stmt
- try_stmt
- except_list
- except_def
- finally_stmt

### DIFF
--- a/jac/jaclang/compiler/parser.py
+++ b/jac/jaclang/compiler/parser.py
@@ -807,9 +807,7 @@ class JacParser(Pass):
                 return kid[0]
             raise self.ice()
 
-        def ability(
-            self, _: None
-        ) -> ast.Ability | ast.AbilityDef | ast.FuncCall:
+        def ability(self, _: None) -> ast.Ability | ast.AbilityDef | ast.FuncCall:
             """Grammer rule.
 
             ability: decorators? KW_ASYNC? ability_decl
@@ -1412,8 +1410,8 @@ class JacParser(Pass):
             finally_stmt: KW_FINALLY code_block
             """
             self.consume_token(Tok.KW_FINALLY)
-            body=self.consume(ast.SubNodeList)
-            kid=self.nodes
+            body = self.consume(ast.SubNodeList)
+            kid = self.nodes
             return ast.FinallyStmt(
                 body=body,
                 kid=kid,
--- a/jac/jaclang/compiler/parser.py
+++ b/jac/jaclang/compiler/parser.py
@@ -808,7 +808,7 @@ class JacParser(Pass):
             raise self.ice()
 
         def ability(
-            self, _: list[ast.AstNode]
+            self, _: None
         ) -> ast.Ability | ast.AbilityDef | ast.FuncCall:
             """Grammer rule.
 
@@ -1370,7 +1370,7 @@ class JacParser(Pass):
                 kid=kid,
             )
 
-        def except_list(self, _: list[ast.AstNode]) -> ast.SubNodeList[ast.Except]:
+        def except_list(self, _: None) -> ast.SubNodeList[ast.Except]:
             """Grammar rule.
 
             except_list: except_def+
@@ -1393,7 +1393,7 @@ class JacParser(Pass):
 
             except_def: KW_EXCEPT expression (KW_AS NAME)? code_block
             """
-            name = None
+            name: ast.Name | None = None
             self.consume_token(Tok.KW_EXCEPT)
             ex_type = self.consume(ast.Expr)
             if self.match_token(Tok.KW_AS):
@@ -1411,9 +1411,9 @@ class JacParser(Pass):
 
             finally_stmt: KW_FINALLY code_block
             """
-            kid = self.nodes
             self.consume_token(Tok.KW_FINALLY)
-            body = self.consume(ast.SubNodeList)
+            body=self.consume(ast.SubNodeList)
+            kid=self.nodes
             return ast.FinallyStmt(
                 body=body,
                 kid=kid,
--- a/jac/jaclang/compiler/parser.py
+++ b/jac/jaclang/compiler/parser.py
@@ -808,7 +808,7 @@ class JacParser(Pass):
             raise self.ice()
 
         def ability(
-            self, kid: list[ast.AstNode]
+            self, _: list[ast.AstNode]
         ) -> ast.Ability | ast.AbilityDef | ast.FuncCall:
             """Grammer rule.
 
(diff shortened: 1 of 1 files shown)


jac/jaclang/compiler/parser.py
```diff
@@ -808,29 +808,24 @@ def enum_stmt(self, kid: list[ast.AstNode]) -> ast.EnumBlockStmt:
             raise self.ice()
 
         def ability(
-            self, kid: list[ast.AstNode]
+            self, _: list[ast.AstNode]
```
ThakeeNathees: Mark the type as `None`

jac/jaclang/compiler/parser.py
```diff
@@ -1363,94 +1313,72 @@ def if_stmt(self, kid: list[ast.AstNode]) -> ast.IfStmt:
...
-                    excepts=except_list,
-                    else_body=else_stmt,
-                    finally_body=finally_stmt,
-                    kid=kid,
-                )
-            else:
-                raise self.ice()
 
-        def except_list(self, kid: list[ast.AstNode]) -> ast.SubNodeList[ast.Except]:
+        def except_list(self, _: list[ast.AstNode]) -> ast.SubNodeList[ast.Except]:
```
ThakeeNathees: Same here, the type should be None

jac/jaclang/compiler/parser.py
```diff
@@ -1465,31 +1393,31 @@ def except_def(self, kid: list[ast.AstNode]) -> ast.Except:
...
-            if isinstance(ex_type, ast.Expr) and isinstance(body, ast.SubNodeList):
-                return ast.Except(
-                    ex_type=ex_type,
-                    name=name,
-                    body=body,
-                    kid=kid,
-                )
-            else:
-                raise self.ice()
+            name = None
```
ThakeeNathees: Isn't this `None | ast.Name`

ThakeeNathees: This should be `ast.Name | None`

jac/jaclang/compiler/parser.py
```diff
@@ -1465,31 +1393,31 @@ def except_def(self, kid: list[ast.AstNode]) -> ast.Except:
...
-                    kid=kid,
-                )
-            else:
-                raise self.ice()
+            kid = self.nodes
+            self.consume_token(Tok.KW_FINALLY)
+            body = self.consume(ast.SubNodeList)
+            return ast.FinallyStmt(
+                body=body,
+                kid=kid,
```
ThakeeNathees: Remove the above `kid = self.nodes`, and add `kid=self.nodes` here

ThakeeNathees (changes requested):

jac/jaclang/compiler/parser.py
MalithaPrabhashana: you mean the type of the kid attribute should be None right? instead of list [ast.AstNode]

jac/jaclang/compiler/parser.py
MalithaPrabhashana: to line 1417?

jac/jaclang/compiler/parser.py
```diff
@@ -1363,94 +1311,72 @@ def if_stmt(self, kid: list[ast.AstNode]) -> ast.IfStmt:
...
-                    excepts=except_list,
-                    else_body=else_stmt,
-                    finally_body=finally_stmt,
-                    kid=kid,
-                )
-            else:
-                raise self.ice()
 
-        def except_list(self, kid: list[ast.AstNode]) -> ast.SubNodeList[ast.Except]:
+        def except_list(self, _: None) -> ast.SubNodeList[ast.Except]:
```
ThakeeNathees, line 1376: This should be:

```python
items = [self.consume(ast.Except)]
while ex := self.match(ast.Except):
  items.append(ex)
return ast.Subnodelist[ast.Except]( items=items, ... )
```
1. No need to validate all the kids ast.Except, cause it's in the grammer and if not there will be error when consume or (future) when we leave the function we ensure all the nodes are processed thus any remaining nodes are not valid = error.
2. `self.ice()` shouldn't be called as it's now handled.

jac/jaclang/compiler/parser.py
```diff
@@ -1363,94 +1311,72 @@ def if_stmt(self, kid: list[ast.AstNode]) -> ast.IfStmt:
...
-                        else None
-                    ),
-                    kid=kid,
-                )
-            else:
-                raise self.ice()
+            self.consume_token(Tok.KW_ELIF)
+            condition = self.consume(ast.Expr)
+            body = self.consume(ast.SubNodeList)
+            else_body = self.match(ast.ElseStmt) or self.match(ast.ElseIf)
```
ThakeeNathees, line 1338: This should be:

```python
else_body = self.match(xxx) or self.consume(xxx) # <-- second one is consume
```

jac/jaclang/compiler/parser.py
```diff
@@ -1363,94 +1311,72 @@ def if_stmt(self, kid: list[ast.AstNode]) -> ast.IfStmt:
...
-                        else None
-                    ),
-                    kid=kid,
-                )
-            else:
-                raise self.ice()
+            self.consume_token(Tok.KW_IF)
+            condition = self.consume(ast.Expr)
+            body = self.consume(ast.SubNodeList)
+            else_body = self.match(ast.ElseStmt) or self.match(ast.ElseIf)
```
ThakeeNathees, line 1322: This should be:
```
else_body = self.match(xxx) or self.consume(xxx) # <-- second one is consume
```

jac/jaclang/compiler/parser.py
```diff
@@ -915,112 +896,79 @@ def abstract_ability(self, kid: list[ast.AstNode]) -> ast.Ability:
...
 
             event_clause: KW_WITH expression? (KW_EXIT | KW_ENTRY) (STRING? RETURN_HINT expression)?
             """
-            type_specs = kid[1] if isinstance(kid[1], ast.Expr) else None
-            return_spec = kid[-1] if isinstance(kid[-1], ast.Expr) else None
-            semstr = (
-                kid[-3] if return_spec and isinstance(kid[-3], ast.String) else None
+            return_spec: ast.Expr | None = None
+            semstr: ast.String | None = None
+            self.match_token(Tok.KW_WITH)
```
ThakeeNathees: `consume_token(KW_WITH)`

jac/jaclang/compiler/parser.py
```diff
@@ -915,112 +896,79 @@ def abstract_ability(self, kid: list[ast.AstNode]) -> ast.Ability:
...
+            is_static = self.match_token(Tok.KW_STATIC) is not None
+            self.consume_token(Tok.KW_CAN)
+            access = self.match(ast.SubTag)
+            semstr = self.match(ast.String)
+            name = self.consume(ast.NameAtom)
+            signature = self.match(ast.FuncSignature) or self.consume(
+                ast.EventSignature
+            )
+            self.consume_token(Tok.KW_BY)
+            body = self.consume(ast.FuncCall)
```
ThakeeNathees, line 942: Also consume token `SEMI`, the rule is make sure all the tokens/ rules are processed and thus at the end of this function (future) will be doing an assertion for that. LIKE:

```
parseing the rule.
assert self.nodes is empty. # <-- semicollon is not consumed an error.
```

jac/jaclang/compiler/parser.py
```diff
@@ -915,112 +896,79 @@ def abstract_ability(self, kid: list[ast.AstNode]) -> ast.Ability:
...
+            signature: ast.FuncSignature | ast.EventSignature | None = None
+            is_override = self.match_token(Tok.KW_OVERRIDE) is not None
+            is_static = self.match_token(Tok.KW_STATIC) is not None
+            self.consume_token(Tok.KW_CAN)
+            access = self.match(ast.SubTag)
+            semstr = self.match(ast.String)
+            name = self.consume(ast.NameAtom)
+            signature = self.match(ast.FuncSignature) or self.consume(
+                ast.EventSignature
+            )
```
ThakeeNathees, line 910: Consume tokens abstract and semicollon, (Read bellow comment for the reasoning)

jac/jaclang/compiler/parser.py
```diff
@@ -852,41 +845,29 @@ def ability_decl(self, kid: list[ast.AstNode]) -> ast.Ability:
...
+            body = self.match(ast.SubNodeList)
```
ThakeeNathees: Make sure to consume SEMI if the body is none. Read bellow for the reasoning for processing all valid rules/ tokens.

ThakeeNathees (changes requested):

jac/jaclang/compiler/parser.py
MalithaPrabhashana: If second one is made as self.consume, either Else or Elif will be Mandatory right? So doesn't it give an error only if statement is present?

ThakeeNathees (approved):
LGTM

PR #1548, opened 2025-02-01
[AST-Refactor] AccessNode removed

ThakeeNathees:
## **Description**

This AccessNode is an abstract type of the AST, however, in the new AST (which is being refactored) it'll only be `Expression` and `Statement` abstract class, and we'll be removing everything else.

NOTE: Once all the dependent nodes of `SubTag` are removed, SubTag will also be removed.

### DIFF
--- a/jac/jaclang/compiler/parser.py
+++ b/jac/jaclang/compiler/parser.py
@@ -10,7 +10,7 @@ from typing import Callable, TypeAlias
 
 import jaclang.compiler.absyntree as ast
 from jaclang.compiler import jac_lark as jl  # type: ignore
-from jaclang.compiler.constant import EdgeDir, Tokens as Tok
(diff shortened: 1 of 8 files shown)


jac/jaclang/compiler/parser.py
```diff
@@ -215,7 +215,11 @@ def global_var(self, kid: list[ast.AstNode]) -> ast.GlobalVars:
             global_var: (KW_LET | KW_GLOBAL) access_tag? assignment_list SEMI
             """
             is_frozen = isinstance(kid[0], ast.Token) and kid[0].name == Tok.KW_LET
-            access = kid[1] if isinstance(kid[1], ast.SubTag) else None
+            access = (
```
ThakeeNathees, line 218: NOTE: SubTag will be removed  and the token will be handled properly, for now I'm doing a type ignore since SubTag's tag is a generic type (Not token).

PR #1549, opened 2025-02-03
[SubNode-Remove-3] Removed from Decorators

kugesan1105:

### DIFF
--- a/jac/jaclang/compiler/absyntree.py
+++ b/jac/jaclang/compiler/absyntree.py
@@ -1497,12 +1497,10 @@ class AbilityDef(AstImplOnlyNode):
         body: SubNodeList[CodeBlockStmt],
         kid: Sequence[AstNode],
         doc: Optional[String] = None,
-        decorators: Optional[SubNodeList[Expr]] = None,
         decl_link: Optional[Ability] = None,
     ) -> None:
         """Initialize ability def node."""
         self.signature = signature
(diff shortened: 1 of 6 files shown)


jac/jaclang/compiler/absyntree.py
```diff
@@ -540,9 +540,9 @@ def sem_token(self) -> Optional[tuple[SemTokType, SemTokMod]]:
 class ArchSpec(ElementStmt, CodeBlockStmt, AstSymbolNode, AstDocNode, AstSemStrNode):
     """ArchSpec node type for Jac Ast."""
 
-    def __init__(self, decorators: Optional[SubNodeList[Expr]] = None) -> None:
+    def __init__(self, decorators: Optional[list[Expr]] = None) -> None:
```
ThakeeNathees: Don't make it an optional list, just `list[Expr]` if there is no decorators the list will be empty

ThakeeNathees: That case all the places we're doing
```python
if self.decorators:
  for decor in self.decorators:
    process(decor)
```

will become:
```python
for decor in self.decorators:
  process(decor)
```

ThakeeNathees (changes requested):

ThakeeNathees (changes requested):
can you create pr agains `subnode-list-remove` branch

PR #1573, opened 2025-02-07
Place holder docstrings are removed.

ThakeeNathees:
Moving forward the docstrings **should** be descriptive and explain in detail otherwise could be ommited.

Example:
```python
class Expr(AstNode):
    """Expression is a combination of values, variables operators and functions
    that are evaluated to produce a value.

    1. Literal Expressions.
    2. Binary Operations.
    3. Unary Operations.
    4. Ternary Operations.
    5. Attribute Access.
    6. Subscript.
    7. Call Expression.
    8. List Value.
    9. Dictionary Value.
    10. Set Value.
    11. Generator Expression.
    12. Lambda Expression.
    13. Conditional Expression.
    14. Yield Expression.
    etc.

    An expression can be assigned to a variable, passed to a function, or
    retuurend from a function.

    Examples:
        "hello world"         # literal.
        <expr>(<expr>, ...);  # call.
        <expr>.NAME           # attribute.
        <expr>[<expr>]        # subscript.
        <expr> if <expr> else <expr>  # ternary.
    """
```


marsninja:
Did we get rid of the lint that forces these docstrings? Also this needs to be rebased if we go forward with it @ThakeeNathees

marsninja:
Bring Docstrings flake 8 check back, only remove unneeded docstrings in obvious files, and be surgical with flake8 ignores.

ThakeeNathees:
Hi @marsninja This is rebased however the tests were not triggered. could you have a look at this?

marsninja (approved):

PR #1574, opened 2025-02-07
Fixing a crash due to complex base class expressions not supported

mgtm98:
## **Description**
Fixing a crash due to two unsupported forms of base class. The fix is just reporting that this is not supported. The issue here is that the inheritance pass depends on fuse type info to get the types of the complex base class expressions, fuse type info depends on inheritance class too for atom trailers. so we need some iterative process to run both of them multiple times.


marsninja:
Need a test @mgtm98

marsninja (approved):

PR #1575, opened 2025-02-07
[Parser-Refactor] Modify Parser Functions for New Logics

Chathura-Jayasinghe:
## **Description**

i edited architype_def , arch_type , decorators ,inherited_archs , sub_name , named_ref , special_ref ,enum_def , enum_block , ability_def .

### DIFF
--- a/jac/jaclang/compiler/parser.py
+++ b/jac/jaclang/compiler/parser.py
@@ -564,23 +564,20 @@ class JacParser(Pass):
                 kid=self.cur_nodes,
             )
 
-        def architype_def(self, kid: list[ast.AstNode]) -> ast.ArchDef:
+        def architype_def(self, _: None) -> ast.ArchDef:
             """Grammar rule.
 
             architype_def: abil_to_arch_chain member_block
             """
-            if isinstance(kid[0], ast.ArchRefChain) and isinstance(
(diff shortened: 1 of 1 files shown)


jac/jaclang/compiler/parser.py
```diff
@@ -564,89 +564,77 @@ def architype_decl(self, _: None) -> ast.ArchSpec:
...
-                    kid=kid,
-                )
-            else:
-                raise self.ice()
+            self.consume_token(Tok.DECOR_OP)
+            return ast.SubNodeList[ast.Expr](
+                items=self.consume_many(ast.Expr),
+                delim=Tok.DECOR_OP,
+                kid=self.cur_nodes,
+            )
```
ThakeeNathees: This should be
```
decorators = []
while self.match(Tok.DECOR_OP):
  decorators.append(self.consume(ast.Expr))

return ...
```
In the above it only match the decorator once ie
```py
@ expr expr expr
can foo() {}
```

however the syntax should be like
```python
@exp @expr @expr
can foo() {}
```

ThakeeNathees (changes requested):

kugesan1105:
already merged [PR_LINK](https://github.com/Jaseci-Labs/jaseci/pull/1578)

PR #1583, opened 2025-02-11
SubNodeList removed from AssignCompr

Thamirawaran:


ThakeeNathees (changes requested):
target this pr against `subnode-list-remove`.

Issue #1600, opened 2025-02-17
[Bug] Global Root Reference Limitations


amadolid:
## **Describe the bug**
Current main, uses `root` syntax to reference current root.
With this approach,  current root are set to a global variable that will be available across request if supported.

Ex: `jac run/enter` with initialization of service or `jac serve`

All upcoming request will use the same reference root. Once it's value change it will affect other running requests.

### **To Reproduce**

Let say we have this walker and 2 request happen at the same time

```python
walker testing {
    can enter with entry {
        # first request - root will be root1
        # second request - root will be root2
        print(root); 

        # simulate long process ex: 3 seconds
        
        # first request - root will now be root2 as it's already been updated
        # second request - root will be root2
        report root;
    }
}
```

PR #1603, opened 2025-02-18
Global root removed and root() returns the ctx root.

ThakeeNathees:
## **Description**

In jaclib `root` changed into `root()`

FIX: https://github.com/Jaseci-Labs/jaseci/issues/1600


amadolid (approved):
LGTM

PR #1605, opened 2025-02-19
[SubnodeList-Remove] Removed from Finally statement

Chathura-Jayasinghe:


jac/jaclang/compiler/absyntree.py
```diff
@@ -2055,11 +2055,12 @@ def normalize(self, deep: bool = False) -> bool:
         """Normalize finally statement node."""
         res = True
         if deep:
-            res = self.body.normalize(deep)
+            for stmt in self.body:
+                res = res and stmt.normalize(deep)
```
Thamirawaran, line 2059:
```
for stmt in self.body:
      res = res and stmt.normalize(deep)
```
is enough

Thamirawaran (changes requested):

PR #1606, opened 2025-02-20
SubNodeList removed from AssignCompr

Thamirawaran:


jac/jaclang/compiler/parser.py
```diff
@@ -3169,7 +3169,7 @@ def assign_compr(self, _: None) -> ast.AssignCompr:
             """
             self.consume_token(Tok.LPAREN)
             self.consume_token(Tok.EQ)
-            assigns = self.consume(ast.SubNodeList)
+            assigns = self.consume(ast.SubNodeList).items
```
ThakeeNathees, line 3172: Change the kids to not have subnodelist too.

ThakeeNathees (changes requested):

PR #1609, opened 2025-02-20
Updating the Reference Python Files

MalithaPrabhashana:


jac/examples/reference/global_and_nonlocal_statements.jac
```diff
@@ -1,20 +1,20 @@
 glob x = "Jaclang ";
 
-can foo() -> None {
-    :g: x ;
+can outer_func() -> None {
```
kugesan1105: `()` are optional for functions with no arguments and can be omitted for cleaner code.

marsninja (approved):

PR #1618, opened 2025-02-24
Import pass opt

mgtm98:
## **Description**
Adding `imports ` to module as alist of all imports in the current module
Changed import_pass to use this list


mgtm98:
Opened acros main by mistake

PR #1625, opened 2025-02-28
Enhance JacImporter to include site and user site packages in search path

AshishMahendra:
## **Description**
This PR updates the JacImporter Class in the importer.py file to enhance the search paths for importing Jac modules. The changes ensure that the [JacImporter](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) includes site-specific directories in the search paths, allowing for a more comprehensive module search.

Changes
* Added logic to include directories from site.getsitepackages()
* Added logic to include the user site directory from site.getusersitepackages()

### DIFF
--- a/jac/jaclang/runtimelib/importer.py
+++ b/jac/jaclang/runtimelib/importer.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import importlib
 import importlib.util
 import os
+import site
 import sys
 import types
 from os import getcwd, path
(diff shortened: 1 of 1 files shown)


ChrisIsKing:
@AshishMahendra good start! Here are a few mods:

- The search of sitepackages should be default behavior instead of on the condition that `JACPATH` is present similar to python.
- I'm getting errors thrown from `import_jac_mod_from_file` which I believe is because of the condition where it's searching JACPATH for the sub-imports. Here's how you can replicate this:
    - Install a module that has jac files: `pip install jivas==2.0.0a2` 
    - Run a jac file that uses that module: e.g `include:jac jivas.agent.lib;`

marsninja (changes requested):
This need a rebase @AshishMahendra, but more critically, we need to add test(s). Also @ChrisIsKing if you can, please confirmation this covers your use case, and add a note on what in additon you need.

marsninja (approved):

PR #1628, opened 2025-03-01
A bug in targ argument of jaclib Node.refs() fixed

ThakeeNathees:
## **Description**

A but in the codegen of `Node.refs()` fixed, The argument `target` was always `None`.

```
node a { has val:int=0; }
node b { has val:int=0; }
node c { has val:int=0; }

with entry {

    x = [a(val=i) for i in range(0,3)];
    y = [b(val=i) for i in range(0,3)];
    z = [c(val=i) for i in range(0,3)];

    x ++> y;
    y ++> z;

    print([x-->y-->z]);
    print([x-->y[:1]-->z[:1]]);

}
```


kugesan1105 (approved):
Works as expected. 
good to merge!

marsninja (approved):

PR #1631, opened 2025-03-04
SubNodeList removed from GlobalVars

Thamirawaran:

### DIFF
--- a/jac/jaclang/compiler/absyntree.py
+++ b/jac/jaclang/compiler/absyntree.py
@@ -211,7 +211,6 @@ class AstNode:
     def unparse(self) -> str:
         """Unparse ast node."""
         valid = self.normalize()
-        print(self.kid)
         res = " ".join([i.unparse() for i in self.kid])
         if not valid:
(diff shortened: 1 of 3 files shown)


jac/jaclang/compiler/absyntree.py
```diff
@@ -2693,7 +2698,9 @@ def normalize(self, deep: bool = True) -> bool:
             assign_parent = self.parent
             if self not in [assign_parent.iter, assign_parent.count_by]:
                 new_kid.append(self.gen_token(Tok.SEMI))
```
kugesan1105, line 2700: @Thamirawaran  we can get rid of this

kugesan1105 (changes requested):

jac/jaclang/compiler/absyntree.py
kugesan1105, line 2695: @Thamirawaran  we can get rid of these

kugesan1105 (changes requested):

PR #1642, opened 2025-03-07
Remove SubNodeList from Fstring

Chathura-Jayasinghe:


Thamirawaran:
`f" {{Thami}}"`
This need to be supported

PR #1653, opened 2025-03-12
Jac program based import

mgtm98:
## **Description**

This PR should serve as the base PR for the passes refactoring. It's based on the jac_program branch which showed good improvement on a jac code (from 75 secs to 13 secs) using jac run.

### Changes are
- Add a dict in `JacProgram` to keep track of all imported modules
- Add a list of all the last imported modules to work on 
- Use the last imported modules list in `JacProgram` to do the imports (instead of the while loop on `ImportPass`)
- Change Import pass to only import the modules inside the current module (no run again is used)

**As an output** `JacImportPass` will work as an API that imports the modules in the current module only, to import their dependencies we need to run it again on the imported modules. 

Note: It's not expected to have all the tests passing here as there will be more PRs after this PR.


kugesan1105 (dismissed):
@mgtm98 The annex_impl function is invoked within the same module node, causing duplicate impl nodes to be added to the module. For example, in the main module, both main.impl and main.test nodes are attached twice or more.
image


image

marsninja (approved):

PR #1654, opened 2025-03-14
Annex build refactored

ThakeeNathees:
## **Description**

`get_annex_paths`  function implemented and `annex_impl` method is simplified into the below:

```python
    def annex_impl(self, node: ast.Module) -> None:
        search_paths = self.get_annex_paths(node.loc.mod_path)
        for file_path in search_paths:
            mod = self.import_jac_mod_from_file(file_path)
            if mod:
                if file_path.endswith(".impl.jac"):
                    node.impl_mod.append(mod)
                else:
                    node.test_mod.append(mod)
                node.add_kids_right([mod], pos_update=False)
                mod.parent = node
```


ThakeeNathees:
Suppressed by #1658

PR #1656, opened 2025-03-14
Run jac run passes on modules one by one

mgtm98:
## **Description**

- Stop attaching the sub modules for the imported modules in the main module (no more giant AST)
- Run the `jac run` passes run on the modules one by one
- Comment out exit_import from SymbolTableBuildPass (Should be moved into SymTableLink)


mgtm98:
@kugesan1105, Can we merge this one?

PR #1658, opened 2025-03-17
Include statement inject for optimization

ThakeeNathees:

### DIFF
--- a/jac/jaclang/compiler/passes/main/import_pass.py
+++ b/jac/jaclang/compiler/passes/main/import_pass.py
@@ -65,55 +65,53 @@ class JacImportPass(Pass):
             node.add_kids_right([mod], pos_update=False)
             mod.parent = node
 
+    def get_annex_paths(self, module_path: str) -> list[str]:
+        """Return all the annexable file paths for the given module_path."""
+        search_paths: list[str] = []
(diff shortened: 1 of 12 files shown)


jac/jaclang/compiler/passes/main/import_pass.py
```diff
@@ -40,19 +40,54 @@ def enter_module(self, node: ast.Module) -> None:
...
-                self.enter_module_path(i)
+            import_paths: list[ast.ModulePath] = self.get_all_sub_nodes(
+                node, ast.ModulePath
+            )
+            for import_path in import_paths:
+                import_stmt = Pass.find_parent_of_type(import_path, ast.Import)
+                if not import_stmt:
+                    continue
+                if import_stmt.is_absorb:
+                    self.process_include(import_stmt, import_path)
```
mgtm98, line 51: here we need to make sure it's a jac import not py import

mgtm98:
1. We need to error out if the include node is not inside a module

marsninja:
Is this still relevant? please add description so I can understand the goal here. Plus this needs to be rebased @ThakeeNathees @mgtm98

marsninja:
We decided for the time being to not support this include style impls

PR #1663, opened 2025-03-18
Fix Node Deletion to Remove Connection

kugesan1105:
## **Description**
In the previous implementation, when a node was deleted, it remained attached to the nodes it was already connected to, which caused unintended behavior. This PR addresses that issue by ensuring that deleting a node also removes its connections, fully detaching it from the graph.

### Before
image
When a node was deleted, its connections to other nodes persisted, leading to an inconsistent state.

### After 
image
With this change, deleting a node now also removes all associated connections, ensuring the node is completely detached as intended.

### Codegen
image


marsninja:
I've merged main on this so its up to date. There is a infinite recursion on the deep copy that needs fixing. I'll also do a review @kugesan1105 .

jac/jaclang/compiler/passes/main/pyast_gen_pass.py
```diff
@@ -1871,20 +1872,46 @@ def exit_ctrl_stmt(self, node: ast.CtrlStmt) -> None:
             node.gen.py_ast = [self.sync(ast3.Return(value=None))]
 
     def exit_delete_stmt(self, node: ast.DeleteStmt) -> None:
-        """Sub objects.
+        """Transforms DeleteStmt into Python AST with destroy call and deletion."""
+
+        # Creating a copy of the node to have separate ctx for each delete target
+        # since we are generating both Jac.destroy and Python del statements.
+        node_copy = copy.deepcopy(node)
```
marsninja: That has infinite recursion

jac/jaclang/tests/test_cli.py
```diff
@@ -548,3 +548,14 @@ def test_caching_issue(self) -> None:
...
                 self.assertIn("F", stderr)
         os.remove(test_file)
+
+    def test_node_del(self) -> None:
+        """Test for graph CLI cmd."""
+        captured_output = io.StringIO()
+        sys.stdout = captured_output
+        cli.dot(f"{self.fixture_abs_path('node_del.jac')}")
+        sys.stdout = sys.__stdout__
+        with open("node_del.dot") as f:
```
marsninja: Need to clean up this file (or add a to_screen option in dot command and capture stdio).

marsninja (changes requested):

kugesan1105:
Hey @marsninja, quick thought — 
Currently, **destroy** method only takes one argument. 
Should we consider supporting destroying multiple objects in a single call too?

like `destroy( node_1, node_2)` 
 👀
Right now, syntax like `del [node_1, node1];` is supported, but the semantics do not seem to apply to destroy.

also it is challenging to support this one currently 
```
apple_list = [1, 2, 3, 4, 5];
del apple_list[2];  # Should remove 3
del [apple_list[0], apple_list[1]];  # ideally deletes 1 & 2, but currently throws a runtime error ❌regardless of destroy 
```
image

marsninja (approved):

marsninja (approved):

PR #1681, opened 2025-03-26
Tests fixes (PR1)

Thamirawaran:


jac/jaclang/compiler/compile.py
```diff
@@ -93,7 +93,7 @@ def run_schedule(mod: ast.Module) -> None:
             ast_ret = i(mod, prior=ast_ret)
         ast_ret = target(mod, prior=ast_ret) if target else ast_ret
 
-    for mod in machine.jac_program.modules.values():
+    for _, mod in list(machine.jac_program.modules.items()):
```
mgtm98: Can we change PyImport attach_mod function (call the old implementation in pyimport instead of calling jacImport new implementation)

mgtm98:
```
def attach_mod_to_node(
        self, node: ast.ModulePath | ast.ModuleItem, mod: ast.Module | None
    ) -> None:
        """Attach a module to a node."""
        if mod:
            self.run_again = True
            node.sub_module = mod
            self.annex_impl(mod)
            node.add_kids_right([mod], pos_update=False)
            mod.parent = node
```

mgtm98 (changes requested):

mgtm98:
## Fixes done in this PR

- SymTabLinkPass: Traverse only JAC import nodes 
- JacImportPass: Change `attach_mod_to_node` to use the old implementation (instead of JacImport one)

PR #1690, opened 2025-04-08
VSCE test cases fix  

kugesan1105:

### DIFF
--- a/jac/jaclang/compiler/compile.py
+++ b/jac/jaclang/compiler/compile.py
@@ -57,6 +57,8 @@ def jac_str_to_pass(
     """Convert a Jac file to an AST."""
     from jaclang.runtimelib.machine import JacProgram
 
+    from icecream import ic
+    ic('inside compile',file_path,'')
     if not target:
         target = schedule[-1] if schedule else None
     source = ast.JacSource(jac_str, mod_path=file_path)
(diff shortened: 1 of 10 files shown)


jac/jaclang/compiler/compile.py
```diff
@@ -83,6 +83,10 @@ def jac_str_to_pass(
     if len(ast_ret.errors_had) != 0:
         return ast_ret
 
+    if target == JacImportPass:  # we need a elegant way of doing this
```
mgtm98: Please add a TODO comment so that we can track it after we re-implement the compile.py

jac/jaclang/compiler/passes/main/import_pass.py
```diff
@@ -102,7 +101,7 @@ def annex_impl(self, node: ast.Module) -> None:
             ) and cur_file.endswith(".impl.jac"):
                 mod = self.import_jac_mod_from_file(cur_file)
                 if mod:
-                    node.add_kids_left(mod.kid, parent_update=True)
+                    node.add_kids_left(mod.kid, pos_update=False)
```
mgtm98: we need to add parent_update also

mgtm98 (changes requested):

mgtm98 (approved):

PR #1700, opened 2025-04-11
Adding new performance action to check for perf degradations for jac-…

mgtm98:
…core

## **Description**
Adding new github action to check the performance numbers of the tests


marsninja:
Theres a better approach.

PR #1741, opened 2025-04-22
Hot Fix  [ importpass]

kugesan1105:


kugesan1105:
We should raise an error while we parsing the second.jac rather , before we move into symtablinkpass @marsninja 
image

Issue #1649, opened 2025-03-12
[Bug] List Slicing with Negative Indices


Chathura-Jayasinghe:
## **Describe the bug**
`list[5:-1]`  gives syntax error. 
`list[5: -1]` will work. Because there is space between `:`,`-`
![Image](https://github.com/user-attachments/assets/4fd68b94-63ba-49a4-b858-76c88310f7f7)

this error generated from `Jac version 0.7.31`.

PR #1754, opened 2025-04-24
fix: allow negative indices in list slices by adjusting token priority

afardeen11:
Fixes: #1649
Summary

This PR fixes a bug where list slicing with negative indices (e.g., arr[5:-1]) caused a syntax error in JacLang. Lowered the priority of DARROW_L_P2 token in the JacLang parser to allow the minus (-) to be correctly interpreted in slicing contexts which ensures that expressions like arr[5:-1] are now parsed successfully

The following test file tests both ways of index slicing.
New test file: jaclang/compiler/tests/test_slice_index.py

Added tests:
test_slice_with_space_parses: Validates that slicing with a space (arr[5: -1]) parses correctly.
test_slice_no_space_debug: Ensures arr[5:-1] parses without error, and includes debug output for diagnosing parser errors if they occur.

### DIFF
--- a/jac/jaclang/compiler/jac.lark
+++ b/jac/jaclang/compiler/jac.lark
@@ -648,7 +648,7 @@ CARROW_L_P2: ":<+"
 CARROW_R_P1: "+>:"
 DCARROW_L_P2: ":+"
 DCARROW_R_P1: "+:"
-DARROW_L_P2: ":-"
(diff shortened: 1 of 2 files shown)


Thamirawaran:
This issue is already fixed. We cannot fix this way. Reason both of the symbols are overlapping. if you prioritise one symbol, another one will fail.

PR #1758, opened 2025-04-24
Jaclang jupyter notebook support

ThakeeNathees:
## **Description**
jac

Suppress: #1729


ThakeeNathees:
Hi @marsninja what you think about the file location of `jackernel.py` it's currently under `support` however the pip installed bundle doesn't shipped with that directory, where do you think should we move it to?

PR #1764, opened 2025-04-24
Concurrency Idea 1

Thamirawaran:
## **Description**
**Try this example**
```
node A{
    has val: int;

    can do1 with W entry {
        visit [self -->];
    }
    can do2 with W exit {
    }
}

walker W{}

with entry {
    root ++> (a1:=A(1)) ++> [a2:=A(2), A(3), A(4)];
    a2 ++> A(5);
    W() spatial spawn a1;

}
```
**condition:** You shouldn't change walker attributes during spawn operation

### DIFF
--- a/jac/jaclang/compiler/absyntree.py
+++ b/jac/jaclang/compiler/absyntree.py
@@ -2735,6 +2735,38 @@ class BinaryExpr(Expr):
         return res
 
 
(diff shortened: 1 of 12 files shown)


marsninja:
Closing for now since #1784 covers it.

PR #1768, opened 2025-04-25
Thakee repl

ThakeeNathees:
## **Description**

The bellow command will start the REPL, however it would be better if the jac command itself (without repl) should initiate the shell. This PR is based on the Jupyter notebook support, that needs to be merged first.
```
jac repl
```

image


ThakeeNathees:
After the rebase with the current main the multi line code evaluation isn't working anymore, this needs to be fixed before merged.

CC: @marsninja

marsninja:
We should add tests for the key functionalities @ThakeeNathees .

marsninja:
Lets get a test in there @ThakeeNathees

jac/jaclang/cli/jacrunner.py
```diff
@@ -0,0 +1,205 @@
...
+            try:  # noqa: SIM105
+                self.runner.execute(mod)
+            except Exception:
+                pass
+            self._dump_outputs()
+
+        return self.exit_code
+
+    def _try_compiling(self) -> JacModule | None:
+        return self.runner.compile(f"with entry {{\n\t{self.code};\n}}")
```
marsninja, line 171: Feels a bit hacky with the with entry string wrapping. See comment.

marsninja (changes requested):
Generally speaking the right way to approach this is by updating the grammar lark file and parser to produce a REPL statement node, perhaps it can be a child of module in the AST. This would be an alternative start symbol int he grammar. If you take a look at how the lark python grammar works in their project that shows how its done. The new start production would just go to single statements. Let me know if you need more pointers @ThakeeNathees

ThakeeNathees:
The issue with the Pre-compiled Lark stand alone parser is that it doesn't allow any rule other than `start` to start the parsing and. That leaves us with refinging the start rule with `start: module | repl_statement` how ever this leads to reduce collision. For example:

`node n {}` is either top_level statement (which is parsed by the `module` rule) at the same time it's a valid `statement`.

image

This is also the case for expressions, because the module can also have docstrings that has collision with the string expression causing a reduce collision.

kugesan1105:
@marsninja  This is important but for now lets keep it in the backlog for the moment . once we finish the typecheck to a saturated point we can comeback to this point .

PR #1776, opened 2025-04-27
feat: Integrating CFG into the UniIR

Jayanaka-98:
## **CFG Integration workflow**

- [ ] BasicBlock nodes will be introduced
- [ ] Identify nodes which need to inhereit from BB  node
- [ ] Link the nodes to create control flow.


jac/jaclang/compiler/unitree.py
```diff
@@ -931,7 +942,7 @@ def normalize(self, deep: bool = False) -> bool:
 
 # AST Mid Level Node Types
 # --------------------------
-class Module(AstDocNode, UniScopeNode):
+class Module(AstDocNode, UniScopeNode, UniBasicBlock):
```
marsninja: I get why code statements have UniBB, but trying to understand why a Module would also.

jac/jaclang/compiler/unitree.py
```diff
@@ -1226,7 +1238,7 @@ def normalize(self, deep: bool = False) -> bool:
         return res
 
 
-class Import(ElementStmt, CodeBlockStmt):
+class Import(ElementStmt, CodeBlockStmt, UniBasicBlock):
```
marsninja: Generally, it looks like all yoru current UniBBs are also CodeBlockStmts, if All CodeBlockStmts are UniBB, then I would make Codeblockstmts derive from UniBB and then you wouldnt have to tag it here.

jac/jaclang/compiler/unitree.py
```diff
@@ -1718,6 +1731,7 @@ class Ability(
     AstSemStrNode,
     AstImplNeedingNode,
     UniScopeNode,
+    UniBasicBlock,
```
marsninja: The prior comment is relelvant here too, but for ElementStmts

marsninja: And I just noticed that CodeBlockStmt is here too, maybe we should check if every Element statemetn is a codeblock statement and simplify these (by removing Codeblockstmt from being tagged here if Element statement derives from codeblock statement).

jac/jaclang/compiler/unitree.py
```diff
@@ -1836,7 +1851,7 @@ def normalize(self, deep: bool = False) -> bool:
         return res
 
 
-class AbilityDef(AstImplOnlyNode, UniScopeNode):
+class AbilityDef(AstImplOnlyNode, UniScopeNode, UniBasicBlock):
```
marsninja: Honestly, i feel this shoudl be an element statement looking at it.

marsninja (changes requested):
We need a test case for this too. But this is a nice pr @Jayanaka-98 !!

jac/jaclang/compiler/unitree.py
Jayanaka-98: The rationale was that the module would hold the starting point of the BB, which references to the first BB of the module.

marsninja (changes requested):
Expand test to include abilities and class methods.

marsninja (approved):

PR #1778, opened 2025-04-28
Trying to move to Mypy 1.5

marsninja:


mgtm98:
why  `__init__.py` is removed from vendor?

jac/jaclang/langserve/tests/test_server.py
```diff
@@ -369,7 +369,7 @@ def test_completion(self) -> None:
             (
                 lspt.Position(37, 12),
                 ["self", "add", "subtract", "x", "Colorenum", "Colour1", "red", "r"],
-                160,
+                153,
```
kugesan1105, line 372: @mgtm98  Why does it decrease by 7 symbols

kugesan1105: The following symbols: 

```
help
credits
eval
<>exit
license
quit
copyright
```

are no longer defined as functions but as instances of _sitebuiltins types.
As a result, they're excluded from mypy's builtins.py, reducing the total symbol count from 160 to 153.

Old
------------------------------
image


New
-------------------------------
image

marsninja: sounds good.

PR #1784, opened 2025-04-29
Concurrency PR1

Thamirawaran:
## **Description**
1. Spawn Expr introduced
2. jacgo keyword introduced( Ex: `jacgo spawn` , `jacgo visit` )
3. Example code added
4. implmentation for `jacgo spawn` is done
5. Test case for concurrency spawn is added
6. implementation for `jacgo visit` is done
7. Test case for concurrency visit in progress
8. Joining, Locking mechanism need to be implemented

**Example:**
```
node A{
    has val: int;

    can do1 with W entry {
        visit [self -->];
    }
    can do2 with W exit {
        here.num = self.val;
    }
}

walker W{
    has num: int;

    can do1 with entry {
        print("W Entry action ", here);
    }
    can do2 with exit {
        print("W Exit action ", here);
    }
}

with entry {
    root ++> (a1:=A(1)) ++> [a2:=A(2), a3:=A(3), a4:=A(4)];
    [a2,a3] ++> (a5:=A(5));
    [a4,a5] ++> (a6:=A(6));
    (w1 := W(8)) jacgo spawn a1;
    print("Walker W spawned");

}
```
**Expected output:**
```
W Entry action  A(val=1)
A Entry action  1
Walker W spawned
```
Program exits with `Walker W spawned` output
**Reason:** when main program exits, threads will be killed

### DIFF
--- a/jac/jaclang/compiler/passes/main/pyast_gen_pass.py
+++ b/jac/jaclang/compiler/passes/main/pyast_gen_pass.py
@@ -1740,7 +1740,7 @@ class PyastGenPass(UniPass):
             if node.from_walker
             else ast3.Name(id=Con.HERE.value, ctx=ast3.Load())
         )
-        
+
         visit_call = self.sync(
             ast3.Call(
                 func=self.jaclib_obj("visit"),


Thamirawaran:
> Lets knock these out:
> 
> jacgo visit becomes an expression create a walker with the Queue loaded return that walker object
> 
> spawns are simply non blocking same semantics (i think you have this already)

for this one i need explanation
The reason is with our current implementation, we are not immediately spawn node when we need to visit. What we are doing is we will add to the queue and spawn after that node abilities are executed. If let spawning before compeleting ability execution, it will result recursive calls

marsninja:
> > jacgo visit becomes an expression create a walker with the Queue loaded return that walker object
> 
> for this one i need explanation The reason is with our current implementation, we are not immediately spawn node when we need to visit. What we are doing is we will add to the queue and spawn after that node abilities are executed. If let spawning before compeleting ability execution, it will result recursive calls

What you describe sounds reasonable and is what I was thinking we'd do. Basically for visits, the new parallel walkers should start after all abilities execute.

marsninja:
Notes relevant to design: 

* Its ok if spawns are simple and map closely to thread start semantics (for now) 
* For visit statements, we need a few special data spatial semantics 
  * We take inspiration from classic forks, but at the walker object level for state cloning (as opposed to all virtual memory)
  * Visit statement should generally be expressions, and they return a list of references to bhe visiting objeccts (walkers)
  * Impmortanlty we use the return of these experessions to simulate the fork() return of pid to distinguish forker from forkee
  * Visits will include the calling walker (calling the visit) as 1 of the elements and spawn N-1 new clones for N destinations in the visit
  * The return at the visit statement will be a list of object references, whether parallel or not (w/wo jac go) so we can have unified semantics for visits an d not have 1 be statements and the other be operator (visit vs jacgo visit)
  * the general apporach to IPC will be modeled as leaving state in the graph itself, i.e., inter walker communication through graph topology

marsninja:
Here we should leverage concurrent.futures, we can also make jacgo have walrus semantics to pluck out the futures that we can interact with using awaits something like `(x:=jacgo)` should be allowed. With this we can have enuoght flexibility to realize sync operations on both visits and spawns.

PR #1796, opened 2025-05-05
feat: Async walker 

Thamirawaran:
## **Description**
Async walker introduced
```
import:py time;
node A{
    has val: int;

    can do1 with W entry {
        visit [self -->];
        time.sleep(1);
    }
    can do2 with W exit {
        here.num = self.val;
    }
}

async walker W{
    has num: int;
}

with entry {
    root ++> (a1:=A(1)) ++> [a2:=A(2), a3:=A(3), a4:=A(4)];
    [a2,a3] ++> (a5:=A(5));
    [a4,a5] ++> (a6:=A(6));
    t1 = (w1 := W(8)) spawn a1;
    t2 = (w2 := W(10)) spawn a1;
    await t1;
    k = await t2;
}
```

output:
```
A Entry action  1
It is non blocking
A Entry action  2
A Entry action  3
A Entry action  4
W(num=8)
```


jac/jaclang/runtimelib/machine.py
```diff
@@ -311,6 +313,8 @@ def detach(edge: EdgeAnchor) -> None:
 class JacWalker:
     """Jac Edge Operations."""
 
+    executor = ThreadPoolExecutor()
```
marsninja: Lets have this in the machinestate.py execution context class.

jac/jaclang/compiler/parser.py
```diff
@@ -159,6 +159,17 @@ def _binary_expr_unwind(self, kid: list[uni.UniNode]) -> uni.Expr:
...
+                    and (isinstance(kid[1], uni.Token) and kid[1].name == Tok.KW_SPAWN)
+                    and isinstance(kid[2], uni.Expr)
+                ):
+                    return uni.SpawnExpr(
+                        left=kid[0],
+                        op=kid[1],
+                        right=kid[2],
+                        kid=kid,
+                    )
+                elif (
```
marsninja: Lets do this stuff in the jacgo pr.

jac/jaclang/compiler/jac.lark
```diff
@@ -43,7 +43,7 @@ architype: decorators? architype_decl
          | architype_def
          | enum
 
-architype_decl: arch_type access_tag? STRING? NAME inherited_archs? (member_block | SEMI)
+architype_decl: KW_ASYNC? arch_type access_tag? STRING? NAME inherited_archs? (member_block | SEMI)
```
marsninja: Lets bring this up to the architype production rule, and check for only walkers and node and edges.

jac/jaclang/compiler/passes/main/pyast_gen_pass.py
```diff
@@ -1757,7 +1764,45 @@ def exit_disengage_stmt(self, node: uni.DisengageStmt) -> None:
...
+                                list[ast3.expr],
+                                [
+                                    node.target.gen.py_ast[0],
+                                ],
+                            ),
+                            keywords=[],
+                        )
+                    ),
+                )
+            )
```
marsninja: take a look at the loop. async strategy form issue.

jac/jaclang/runtimelib/machine.py
```diff
@@ -396,92 +400,106 @@ def spawn(op1: Architype, op2: Architype) -> WalkerArchitype:
...
+            for i in warch._jac_exit_funcs_:
+                if not i.trigger:
+                    i.func(warch, current_node)
+                if walker.disengaged:
+                    return warch
+
+            walker.ignores = []
+            return warch
+
+        if hasattr(warch, "is_async") and warch.is_async:
```
marsninja: not sure we need the hasattr part

marsninja (changes requested):

jac/jaclang/runtimelib/machine.py
amadolid: we should also use private fields like \_\_is_async__ to avoid field collision

Thamirawaran: sure

marsninja (approved):

PR #1852, opened 2025-05-13
Prevent LSP error on missing Jac file

kugesan1105:
## **Description**

Adds safe fallback for deep check on missing files.

---

image

---

- Static Semantic Tokens  are not handled properly yet
-

### DIFF
--- a/jac/jaclang/langserve/engine.py
+++ b/jac/jaclang/langserve/engine.py
@@ -9,7 +9,7 @@ from concurrent.futures import ThreadPoolExecutor
 from typing import Callable, Optional
 
 import jaclang.compiler.unitree as uni
-from jaclang.compiler.passes.main import Alert, CompilerMode as CMode
+from jaclang.compiler.passes.main import CompilerMode as CMode
 from jaclang.compiler.program import JacProgram
 from jaclang.compiler.unitree import UniScopeNode
 from jaclang.langserve.sem_manager import SemTokManager
@@ -31,7 +31,7 @@ import lsprotocol.types as lspt
 
 
 class ModuleManager:
-    """Handles Jac module and semantic manager updates."""
+    """Handles Jac module, semantic manager, and alert management."""
 
     def __init__(self, program: JacProgram, sem_managers: dict) -> None:
         """Initialize ModuleManager."""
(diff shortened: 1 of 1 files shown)


jac/jaclang/langserve/engine.py
```diff
@@ -30,103 +30,131 @@
...
+        self.warnings_had = warnings_had
+
+    def clear_for_file(self, file_path_fs: str) -> None:
+        """Remove errors and warnings for a specific file from the lists."""
+        self.errors_had[:] = [
+            e for e in self.errors_had if e.loc.mod_path != file_path_fs
+        ]
+        self.warnings_had[:] = [
+            w for w in self.warnings_had if w.loc.mod_path != file_path_fs
+        ]
```
kugesan1105: @marsninja Now JacLanguageServer is integrated with JacProgram, so we need to clear previous build errors and warnings specific to a file before recompiling it (for the second time). However, we should retain the warnings for other files.

For example, if I have both circle.jac and guessgame.jac open — and they are independent — they will work fine. But if I introduce a syntax error in circle.jac and then fix it, we need to ensure that the build errors caused by the temporary syntax issue are cleared once the file is valid again in the `circle.jac` .

To handle this, I created this class just to add the clear_for_file method.
`clear_for_file` method removes all errors and warnings associated with a specific file path from these lists. This is useful when a file is recompiled or rechecked, so old alerts for that file are cleared before new ones are added.

What’s your review on this?

marsninja: Ah, this make perfect sense. Yes this strikes me as a good approach.

jac/jaclang/langserve/engine.py
```diff
@@ -30,103 +30,131 @@
...
+        file_path = file_path.removeprefix("file://")
+        self.program.mod.hub[file_path] = build
+        self.sem_managers[file_path] = SemTokManager(ir=build)
+        if update_annexed:
+            for p, mod in self.program.mod.hub.items():
+                if p != file_path:
+                    self.sem_managers[p] = SemTokManager(ir=mod)
+
+
+class AlertManager:
```
marsninja: should this be merged into module manager for now since its just 1 function?

jac/jaclang/langserve/engine.py
```diff
@@ -30,103 +30,131 @@
...
+class JacLangServer(JacProgram, LanguageServer):
+    """Jac Language Server, manages JacProgram and LSP."""
 
     def __init__(self) -> None:
-        """Initialize workspace."""
-        super().__init__("jac-lsp", "v0.1")
-        self.modules: dict[str, ModuleInfo] = {}
+        """Initialize JacLangServer."""
+        LanguageServer.__init__(self, "jac-lsp", "v0.1")
+        JacProgram.__init__(self)
```
marsninja, line 69: I wonder whether, we move this init iside the module manager, i also wonder if module manager, i also wonder if you have several "separate" jac programs if we would need separate JacProgram's :-p

jac/jaclang/langserve/engine.py
```diff
@@ -30,103 +30,131 @@
 import lsprotocol.types as lspt
 
 
-class ModuleInfo:
-    """Module IR and Stats."""
+class ModuleManager:
+    """Handles Jac module and semantic manager updates."""
```
marsninja: Since this is so lightweight maybe we dont need a separate class for it, your choice though.

PR #1866, opened 2025-05-15
Markdown of the Torch Dynamo Deep Dive - When to compile

Jayanaka-98:
## **Description**

How Dynamo decides when to compile is added


marsninja:
Very nice start to this doc. Would love to see tables for each key question and justification for completeness and correctness.

PR #1870, opened 2025-05-15
[add] test coverage for jaclang langserver

kugesan1105:
## **Description**
------------------

Adds core tests for JacLanguageServer [VSCE} functionality (open, change, save, etc.)


kugesan1105:
If `pytest-asyncio` is not installed, tests will be skipped.
but it is added in the `test-jaseci.yml`.
It is passing in the local machine, but skipping in github action

marsninja (approved):

PR #1909, opened 2025-05-21
Improve expression line breaking

marsninja:
## Summary
- add helpers to wrap expressions when a group breaks
- apply soft-breaking and parentheses to many expression types
- support long list/dict values with conditional indentation

## Testing
- `python -m pytest jac/jaclang/compiler/passes/tool/tests/test_doc_ir_gen_pass.py -q` *(fails: No module named pytest)*


marsninja:
Crap

PR #1916, opened 2025-05-21
Bringing MTP back to life

Jayanaka-98:
## **Description**
The MTP needs to work with the newest jaclang version. The Implementation of UniiR and the removal of the Semtable has broken it. 

TODO:
- [x] Update the plugin part of mtllm
- [x] Write new UniiR lookup (and  other) algorithms to access the same information as with SemRegistry
- [x] Set the bridging interface to work
- [ ] [Optional] Sem impl implementation (Adding semstrings as impl like files)


Jayanaka-98 (commented):
The codegenning is done, Need to fix the with_llm to work with uniir

PR #1920, opened 2025-05-22
Vsce client test

kugesan1105:


kugesan1105:
@marsninja 
LSP startup errors show up in the VSCode UI during tests, but are missing from automated test results.
Screenshot 2025-05-22 155223
but test are passing 
image

marsninja:
Hi @kugesan1105 , this pr looks good to me, is it the one long running test? we can probably investigate why it takes so long real quick or temporariliy disable that one test and pus now?

PR #1970, opened 2025-05-28
Add support for OOP stuff

mgtm98:
## **Description**

- [x] Add support for enums & classes
- [x] Add support for attribute accessing
- [x] Add support for function calling 
- [x] Add support for automatic constructor creation
- [x] Add support for inheritance
- [ ] Add support for override semantics
- [ ] Add support for overload semantics


mgtm98:
Needed to add support for accessing base class attributes using super
Needed to add support for adding new members using self.<member> in constructor only

PR #1971, opened 2025-05-29
Jac-playground printgraph issue is fixed

MalithaPrabhashana:
Updated the `user_line` method to use `format='json'` instead of `as_json=True` for better consistency with the `printgraph` function's expected parameters.


marsninja (approved):
LGTM, though we need some tests if possible, merging for now.

MalithaPrabhashana:
Yeah will check on the tests

PR #1975, opened 2025-05-30
Mtllm fix dev

ThakeeNathees:


.pre-commit-config.yaml
```diff
@@ -35,4 +35,7 @@ repos:
           - --ignore-missing-imports
           - --install-types
           - --non-interactive
+          - --config-file=mypy.ini
+          - --explicit-package-bases
+          - --exclude=jac-mtllm/scripts
```
ThakeeNathees, line 40: This is to fix `gh_release.py` module is defined in two different places:
```
error: --install-types failed (an error blocked analysis of which types to install)
jac/scripts/gh_release.py: error: Duplicate module named "gh_release" (also at "jac-mtllm/scripts/gh_release.py")
jac/scripts/gh_release.py: note: See [https://mypy.readthedocs.io/en/stable/running_mypy.html#mapping-file-paths-to-modules](vscode-file://vscode-app/c:/Users/Thakee/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) for more info
jac/scripts/gh_release.py: note: Common resolutions include: a) using --exclude to avoid checking one of them, b) adding __init__.py somewhere, c) using --explicit-package-bases or adjusting MYPYPATH
Found 1 error in 1 file (errors prevented further checking)
```

Jayanaka-98 (approved):
The changes seems good. The way _scope is fixed is reasonable and this is ready to merge.

jac-mtllm/mtllm/plugin.py
```diff
@@ -216,7 +209,7 @@ def with_llm(
 
         meaning_out = aott_raise(
             model,
-            informations,
+            [],  # TODO: Collect and pass information here.
```
ThakeeNathees: I removed this cause it was initialized as empty list above (actual initialization is commented) and the type annotation is not compatable with the parameter here. That break mypy. LMK any comments

jac-mtllm/mtllm/utils.py
```diff
@@ -82,23 +81,29 @@ def get_type_annotation(data: Any) -> str:  # noqa: ANN401
...
+#         scope = scope.parent
+#
+#     print(f"Available scopes: {avail_scopes}")
+#
+#     filtered_registry = SemRegistry()
+#     for _scope, sem_info_list in mod_registry.registry.items():
+#         if str(_scope) in avail_scopes:
+#             filtered_registry.registry[_scope] = sem_info_list
+#
+#     return filtered_registry
```
ThakeeNathees, line 106: ## NOTE:
This function is not used anywhere and it was brought from the first version mtllm That was unmaintained for a while. This is now breaking mypy because of the .registry attribute. Uncomment and investigate the old implementation or understand and reimplement **if needed**

jac-mtllm/mtllm/plugin.py
Jayanaka-98: Cant we do this by 
`information = [Information(mod_registry, x[0], x[1]) for x in incl_info]`
bypassing the filtered registry?

marsninja:
Hi @Jayanaka-98 can you take a look at htis, i thought we would be getting rid of some of these unneeded classes like semscope and semregistry. some of the stuff in this pr looks good though.

PR #2040, opened 2025-06-02
Data spatial Async Await support

Thamirawaran:
## **Description**
```
import asyncio;
import from typing {Coroutine}

node MyNode {
    has val:int;
}

edge MyEdge  {
    has value:int;
}

async def fun{
    print("Async function");
}

async walker MyWalker {
    has name:str;

    async can do_print with MyNode entry {
        print("I am here man", here);
        visit [-->];
        await fun();
    }
}

with entry {
    root ++> (node1:= MyNode(5));
    node1 +>:MyEdge(100):+> MyNode(10);
    node1 +>:MyEdge(200):+> MyNode(15);
    walker1 = MyWalker("Mylife");
    async def foo(w:W, a:A)-> None {
        print("Let's start the task");
        x = w spawn a;
        print("It is Coroutine task", isinstance(x, Coroutine));
        await x;
        print("Coroutine task is completed");
    }
    asyncio.run(foo(walker1,node1));
}
```


jac/jaclang/runtimelib/machine.py
```diff
@@ -469,13 +469,26 @@ def spawn_call(
...
+        """Execute abilities."""
+        if inspect.iscoroutinefunction(func):
+            try:
+                asyncio.get_running_loop()
+            except RuntimeError:
+                asyncio.run(func(*args))
+            else:
+                asyncio.ensure_future(func(*args))
+        else:
+            func(*args)
```
marsninja: Why is this type of per call dynamic detection needed? I'm concerned about performance impact of calling this for every ability call. Is this to support the dynamism alexie was asking for?

marsninja (changes requested):
See my note, also we should probably update the relevant md file for jac_ref inclusion, and add a bullet to release_notes.md

jac/jaclang/runtimelib/machine.py
Thamirawaran: Yes. this is for initial implementation. 

yeah I will add relsease notes

marsninja (approved):

Issue #1973, opened 2025-05-29
Semantic token colors lost on file change – VSCE


kugesan1105:
### Description
---------
When editing a file in the VS Code extension (VSCE), semantic token colors are lost after changes are made to the file. This causes the syntax highlighting to disappear or fall back to default styling.

### Steps to Reproduce:
---------

- Open a file with semantic token coloring.
- Make a change (e.g., type a character or whitespace).
- Observe that semantic coloring is lost or reset.

### Expected Behavior:
-----------
Semantic tokens should retain their coloring after file changes.

PR #2045, opened 2025-06-02
[VSCE]: Fix : Semantic token colors retain on file change 

kugesan1105:
Fix for : #1973


marsninja:
hey htis looks good but dont we need some test?

marsninja (approved):

PR #2067, opened 2025-06-05
Testing MTP Examples

udithishanka:


jac/examples/rpg_game/Diagrams/Jac_impl_1_comparison.png
Jayanaka-98, line 1: Need to be replaced with new screen caps, or use a better way to represent. (having the code in two tabs within the code env might be better)

jac/examples/rpg_game/README.md
Jayanaka-98, line 1: The impl 1 to 5 of the rpg_game was removed fro some reason from the repo. Not sure how to change this to fit or bring back the old files (which leads to updating syntax for those files.)

jac/examples/rpg_game/RPG_DSP_journey.md
```diff
@@ -0,0 +1,509 @@
...
+>
+> - Here the fields of this class are defined using ```has``` keyword. This symbolizes that the specified 'obj' 'has' the said variable.
+
+In Jaclang, an ```__init__``` function is not required as the field variables and input parameters(if any) will auto generate and execute an initialization function in the background.
+
+Although an ```__init__``` function is not required by the programmer, there are such cases where some other tasks require to be performed just after ```__init__```. These tasks can be performed in Jaclang with ```postinit``` which will be called after(post) the initialization function.
+
+```python
+    ...
+    can postinit {
```
Jayanaka-98: Needs syntax updating of jac code

Jayanaka-98 (commented):
It is good to bring back the RPG docs, even if they are not linked to the main docs. But as the previous jac impl files of the game no longer exist in the repo this makes things complicated.

Other changes are good.

marsninja:
I generally like the direction but we should keep working on it until we can link it into the actual docs page (in the examples section I'm guessing but up to you guys). Also, we should think through the audience for this, perhaps it should be writing for someone who tries the game, but then wants a code walkthrough, so maybe its framed as a code walkthrough of the game vs a tutorial perhaps. Like an educational journey to deconstruct a dsp implementation and usage of mtllm in practice rather than a "tutorial".

this is wrt the rpg game, also think through the target audience and key message for the other game, perhaps that would be a tutorial (basically im thinking a tutorial is more a document that when you go through it you will have built a complete working app)

jac/examples/rpg_game/README.md
udithishanka: Yes, I removed the old one.

Jayanaka-98 (approved):

marsninja (approved):

PR #2080, opened 2025-06-07
shopping cart example added

Thamirawaran:


jac/examples/shopping_cart/main.impl.jac
daynauth, line 1: Can you add the option to set the sales tax to 0 if the item is a book? What if we wanted to set a tax on an item, optionally? That is, the item is taxed at a flat rate unless we specify what the tax rate should be.

daynauth (changes requested):

daynauth (approved):
Looks Good

Good job with this

PR #2097, opened 2025-06-10
[JacBookFix] Chapter1: minor jac code and mermaid fixes

daynauth:

### DIFF
--- a/docs/docs/jac_book/chapter_1.md
+++ b/docs/docs/jac_book/chapter_1.md
@@ -50,35 +50,41 @@ print(fof_alice)
 
 <div class="code-block">
 ```jac
-# Jac's Object-Spatial approach
- node User {
+# Jac's Data Spatial approach
+node User {
     has name: str;
 }
 
-edge Friendship {
-    has since: str;
-}
+edge Friendship {}
 
 walker FindFriendsOfFriends {
-    has person: User , friends_of_friends: set = set();
+    has person: User;
+    has friends_of_friends: set = set();
+    has friends: list = [];
+
 
     can traverse with User entry {
-        if here != self.person {
+        if here == self.person {
+            self.friends = [->:Friendship:->];
+            self.friends.append(here);
+        }
+
+        if here not in self.friends{
             self.friends_of_friends.add(here);
         }
-        visit [->:Friendship :->];
+        else{
+            visit [->:Friendship :->];
+        }
     }
 }
 
 with entry {
-    alice = User(name="Alice");
-    bob = User(name="Bob");
-    carol = User(name="Carol");
-    dave = User(name="Dave");
+    alice = User(name = "Alice");
 
-    alice +>: Friendship(since="today") :+> bob;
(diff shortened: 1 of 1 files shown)


daynauth:
> Create a list of the improvements that make sense. (What would make the 7 -> 9.5)
1. The first definition of the language doesn't sound very exciting
> Jac is a programming language that extends familiar Python-like syntax with revolutionary concepts from Object-Spatial Programming (OSP). While maintaining the readability and expressiveness that Python developers love, Jac introduces a fundamental paradigm shift in how we think about and structure computation.

Compare this to something like go. 
> The Go programming language is an open source project to make programmers more productive. **Go is **expressive**, **concise**, **clean**, and **efficient**. Its **concurrency** mechanisms make it easy to write programs that get the most out of multicore and networked machines, while its novel type system enables flexible and modular program construction.** Go compiles quickly to machine code yet has the convenience of garbage collection and the power of run-time reflection. It's a fast, statically typed, compiled language that feels like a dynamically typed, interpreted language.

From the first two lines, I know that Go might be a language that I use to gain great system performance while also having a clean, concise, and expressive syntax to work with. 

Why should I use jac? What are one or two killer features of the language that would make me interested enough to drop Python and start using jac? Is it faster? Is it more secure? Can I express complex relationships that are demanded by currently agentic development, that I need a whole new paradigm to represent data and computation? I'm not getting a hook from the first paragraph. 

2. The first example is too long and verbose and would probably turn off most non-developers; it should probably be a simple hello world example. 

```python
with entry {
    print("Hello World");
}
```
followed by
```bash
jac run hello.jac
```

3. Have the quickest installation method in this chapter before chapter 2, so users can get this up and running as fast as possible. 

```bash
pip install jac-lang
```

4. Remove all Python code from this chapter. It's the first chapter; have the readers used to the language before starting with the comparison. In fact, there should be only one chapter dedicated to the comparison between python and jac. The rest of the chapters should be python free. 

5. The layout of the docs is not optimal and makes reading a chore. There are two sidebars that are squeezing the main content in the middle and not allowing the text space to breathe. The two side bars are really distracting and fatigue the reader. Also, the unused white space on both sides of the docs that are not serving a purpose, and just serve as fillers to squeeze the sidebars more into the main content. Consider having the sidebars pushed to the left and right ends of the screen. Maybe even consider removing one or both of the sidebars and have the content stand on its own.

docs/docs/jac_book/chapter_1.md
```diff
@@ -50,35 +50,41 @@ print(fof_alice)
...
-edge Friendship {
-    has since: str;
-}
+edge Friendship {}
 
 walker FindFriendsOfFriends {
-    has person: User , friends_of_friends: set = set();
+    has person: User;
+    has friends_of_friends: set = set();
+    has friends: list = [];
```
marsninja, line 63: I'd call this list ignore

marsninja: I'd call this list ignore, instead of friends

PR #2102, opened 2025-06-11
[DocFix] Jac cloud Docs fix

Thamirawaran:


amadolid:
Add some minor changes.

On other hand, why do we use
```
```jac
```
instead of 
```
```python
```

jac is not supported yet for markdown, it won't do anything yet. While python will still be helpful since we still support python syntax

Thamirawaran:
Let's merge it

amadolid:
> Currently we are supporting for markdown

Got it, approving the PR
LGTM

PR #2106, opened 2025-06-11
MTLLM documentation updated and issues were fixed

ThakeeNathees:

### DIFF
--- a/docs/docs/learn/jac-mtllm/quickstart.md
+++ b/docs/docs/learn/jac-mtllm/quickstart.md
@@ -65,8 +65,8 @@ Following code snippet will be our starting point:
 
 ```jac
 def translate(eng_sentence: str, target_lang: str) -> str {
-    """Normally this would include the translation logic such as calling an API.
-    For the sake of this example, we will return a dummy translated sentence."""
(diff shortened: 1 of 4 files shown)


docs/docs/learn/jac-mtllm/quickstart.md
```diff
@@ -65,8 +65,8 @@ Following code snippet will be our starting point:
 
 ```jac
 def translate(eng_sentence: str, target_lang: str) -> str {
-    """Normally this would include the translation logic such as calling an API.
-    For the sake of this example, we will return a dummy translated sentence."""
+    #* Normally this would include the translation logic such as calling an API.
```
ThakeeNathees, line 68: Since this is not considered as `docstring` and just a string expression in the function servers as a comment, I have change it to comment otherwise the reader would be under the impression of this could be a `docstring`

marsninja (approved):

docs/docs/learn/jac-mtllm/quickstart.md
marsninja: Thisi s great!

PR #1876, opened 2025-05-16
Type hell

mgtm98:
## **Description**
Adding support for type checking in the following cases

1. primatives
2. user defined types
3. assignment
4. function calls

PR #2160, opened 2025-06-17
[Typecheck ] Add Typeshed

kugesan1105:


marsninja:
This looks like the right approach, however I think it will ultimately go into a typeshed folder inside our type system folder. The first mergable version should be with the first typecheck pr. I'm looking at #1876 as hopefully that first pr, but I need a more apples to apples comparison with pyrights archtiecture. Working with @mgtm98 on that. Perhaps next step would be to make this a pr against #1876.

marsninja:
Actually, after seeing other pr, the vendor location should be fine for this. and perhaps it should be merged into #2161 as a base.

kugesan1105:
@marsninja 
Screenshot 2025-07-11 070613

marsninja (approved):

PR #2161, opened 2025-06-17
[Typecheck] Enable Python Import Handling in Jac

kugesan1105:
Converts and embeds Python modules into JacProgram.
It will take around 13 seconds to fully convert . 209 python files.

image


marsninja:
This looks really good, but i think the step with the only pulling in the ptahs based on uses from the defuse pass that @mgtm98 and I recently discussed makes sense. I would also want to understand how pyright accomplishes what our def use pass does to validate we architect things with the best ideas (our current or thiers)

kugesan1105:
yea
Yep, that's the plan from #2193. 
Currently digging into pyright's [ binder ] & how it finds which modules to parse.
and enhance our defuse pass and importpass

kugesan1105:
no more need: as we are following the pyright: typeEvaluator is the true powerhouse

PR #2180, opened 2025-06-19
[DON'T MERGE IT'S A POC] structured output implemented (Ollama)

ThakeeNathees:

### DIFF
--- a/jac-mtllm/mtllm/llms/ollama.py
+++ b/jac-mtllm/mtllm/llms/ollama.py
@@ -1,5 +1,8 @@
 """Ollama client for MTLLM."""
 
+import json
+from pydantic import TypeAdapter
+from typing import Any
 from mtllm.llms.base import BaseLLM
 
 REASON_SUFFIX = """
@@ -57,7 +60,13 @@ class Ollama(BaseLLM):
             k: v for k, v in kwargs.items() if k not in ["model_name", "host"]
         }
 
-    def __infer__(self, meaning_in: str | list[dict], **kwargs: dict) -> str:
+
+    def __infer__(
+        self,
+        meaning_in: str | list[dict],
+        output_type: Any | None = None,
+        **kwargs: dict,
+    ) -> str:
         """Infer a response from the input meaning."""
         assert isinstance(
             meaning_in, str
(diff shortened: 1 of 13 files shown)


jac-mtllm/mtllm/llms/ollama.py
```diff
@@ -85,11 +94,33 @@ def __infer__(self, meaning_in: str | list[dict], **kwargs: dict) -> str:
...
 
+        output_schema = TypeAdapter(output_type).json_schema() if output_type else None
         output = self.client.chat(
             model=model,
             messages=messages,
             options={**self.default_model_params, **model_params},
+            format=output_schema,
         )
+
+        # FIXME:
```
ThakeeNathees, line 105: Since the `__infer__` can only return strings, I'm de serializaing the json to object and serializing to strting and somewhere it's deserialized again to buid the object again and it's super ugly

jac-mtllm/mtllm/plugin.py
```diff
@@ -207,6 +208,24 @@ def with_llm(
         else:
             _tools = []
 
+        # FIXME: I know this is ugly, but the SemInfo, and SemRegistry doesn't
```
ThakeeNathees, line 211: There is no way to get the return type in mtllm, so I'm doing this monkey patching, we need to remvoe the SemRegistry and SemInfo and get actual proper types to make sure the types are in place.

jac/jaclang/compiler/passes/main/pyast_gen_pass.py
```diff
@@ -772,7 +776,7 @@ def exit_ability(self, node: uni.Ability) -> None:
                     ),
                     body=[cast(ast3.stmt, i) for i in body],
                     decorator_list=[cast(ast3.expr, i) for i in decorator_list],
-                    returns=self.sync(ast3.Constant(value=None)),
+                    returns=ast_returns,
```
ThakeeNathees, line 779: This was a *BUG* in the py ast code gen, it currently always generate code as `-> None` this will fix, I should make another PR for this.

jac-mtllm/mtllm/llms/ollama.py
ThakeeNathees: Also note that ReAct is also doing the same, the tool call currently build the object, convert to string and we're building it again. This is not the elegant way to do.

jac-mtllm/mtllm/plugin.py
marsninja: Ah, yes, this may need to have the type checker done first im guessing.

PR #2213, opened 2025-06-24
Update chapter 2: Remove REPL section and clarify its unavailability

AshishMahendra:


Thamirawaran:
@AshishMahendra Ohhh man it is very nice

PR #2223, opened 2025-06-26
[BUGFIX]: Save should not override root owner at runtime

amadolid:


marsninja (approved):

marsninja (changes requested):
Hi @amadolid can you add a bullet to the release_notes.md in the unrelase section on this? Also a test would be nice.

PR #2248, opened 2025-07-01
Enhance jacflash

Thamirawaran:


jac/examples/guess_game/guess_game5.jac
```diff
@@ -0,0 +1,23 @@
...
+    has correct_number: int = random.randint(1, 10);    
+}
+
+## If you want to manually see without cloud
+# with entry {
+#     root spawn GuessGame(3);
+#     root spawn GuessGame(3);
+#     root spawn GuessGame(3);
+#     root spawn GuessGame(3);
+# }
```
marsninja: Instead of commenting out, you can actually do `with entry:__main__ {}` and it will only trigger if you run it with `jac run`

jac/examples/guess_game/guess_game6.jac
```diff
@@ -0,0 +1,31 @@
+"""A Number Guessing Game"""
+
+import random;
+import from mtllm.llms {OpenAI}
+
+glob llm = OpenAI(model_name="gpt-4o",verbose=True );
+
+"""Analyze the player's guess and provide a helpful, encouraging hint about whether they should guess higher or lower. The hint should be creative and fun, without revealing the exact number. Consider how close or far the guess is from the target number and tailor the hint accordingly."""
+def give_hint(guess: int, correct_number: int) -> str by llm();
```
marsninja: It would be good to use the minimal prompt (i wonder what it does if theres actually a good reponse with no string), but perhaps a string like """Provide hint if guess is incorrect""" would produce good results in practice.

marsninja (changes requested):

jac/examples/guess_game/guess_game5.jac
Thamirawaran: `Exception: JaseciContext is not yet available!`

jac/examples/guess_game/guess_game6.jac
Thamirawaran: """Provide a fun hint if guess is incorrect"""

jac/examples/guess_game/guess_game5.jac
Thamirawaran: fixed

Thamirawaran:
Thanks @amadolid 
and @marsninja can you check `cli.py`

Issue #2245, opened 2025-06-30
Jac Cloud set global auth via cli


daynauth:
## **Feature Description**
The current method to set authentication for a walker endpoint is by using a static variable:

```python
walker api_walker{
    obj __specs__ {
        static has auth: bool = False;
    }
}
```

It would be great if we could set this globally via the cli by passing a flag like:
```bash
jac serve book2.jac --host 0.0.0.0 --port 8080 --auth off
```

This would be great for testing and debugging without having to set auth to false for each walker manually.

amadolid:
I don't think testing should affect how authentication works.
If we turn off authentication on endpoint that requires authentication it will conflict to it's use case. Non-authenticated walkers doesn't require a user.

amadolid:
I'm not sure if we really need to support this because every endpoint is different. Even FastAPI doesn't support this as far as I know. Please let me know your thoughts about this.

PR #2255, opened 2025-07-01
[MINOR]: Support Auth switching

amadolid:


marsninja:
Which projects need this? Also, with public/private keyword, I believe that protected can be used to requiering auth. and public being no auth, what do you think @amadolid

marsninja (commented):
see comment.

amadolid:
This is related to @daynauth issue
https://github.com/jaseci-labs/jaseci/issues/2245

This is not required and no current projects need this.

amadolid:
On other hand, public and private is different.

Private walkers are not included in auto endpoint generation and only for internal use ex: `spawn`

public is the normal behavior with auto endpoint generation (default)

marsninja (approved):

PR #2271, opened 2025-07-04
🔑 MTLLM Schema handled properly

ThakeeNathees:
This will generate a simple (subset of json) schema that OpenAI support that also include the semstrings as a descriptsion of other types. (currently semstrings are attached to the calling method, and it's parameters). The schema is generated by pydantic and thus the semstrings not attached.

Also this PR will suport primitive types at the root level of the schema and `dict[k, v]`, `tuple[t, ...]` will be handled properly. This schema generation will always ensure that `additionalProperties` are always `false` which is an OpenAI requirnment.

## Example working test case:

```py
# =============================================================================
# Case: Semstrings attached in schema
# =============================================================================

obj Client {
    has name: str;
    has major: str;
}

sem Client.name = "Should always be 'Alice'";
sem Client.major = "Should always be 'Legal Advisor'";

def generate_client() -> Client by llm();

test semstrings_attached_in_schema {
    client = generate_client();
    assert isinstance(client, Client);
    assert client.name == "Alice";
    assert client.major == "Legal Advisor";
    print(f"Generated Client: {client.name}, Major: {client.major}");
}
```

### DIFF
--- a/jac-mtllm/mtllm/__init__.py
+++ b/jac-mtllm/mtllm/__init__.py
@@ -1,6 +1,6 @@
 """MTLLM Package."""
 
 from mtllm.llm import Model
-from mtllm.types import Image, Video
+from mtllm.types import Image, MockToolCall, Video
 
-__all__ = ["Image", "Model", "Video"]
+__all__ = ["Image", "MockToolCall", "Model", "Video"]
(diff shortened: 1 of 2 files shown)


Jayanaka-98 (approved):
Schema handling is built and tested

Issue #2284, opened 2025-07-07
Validate that associated sems appear in the symbol table debug prints


marsninja:
When you have a sem keyword that is associating a semstr, we need to make sure it appears when printing the symbol tables debug tree.

PR #2289, opened 2025-07-08
🧪 Test added for semstrings in symbol table

ThakeeNathees:
closes: #2284


Jayanaka-98 (approved):
This test effectively checks if semstrings are included in the symbol table

PR #2297, opened 2025-07-09
🌐 webp image support in mtllm for python<3.13

ThakeeNathees:


Jayanaka-98 (approved):
This is a quick change. tests are added, no need of updating docs

PR #2304, opened 2025-07-10
🐍 Python `by` decorator for MTLLM by keyword

ThakeeNathees:
```py
from dataclasses import dataclass
from mtllm import by, Model, Image

llm = Model(model_name="gpt-4o")

@dataclass
class Person:
    name: str
    birth_year: int
    description: str


@by(llm)
def get_person_info(img: Image) -> Person: ...

person = get_person_info('alan-turing.jpg')
print(person)
```


Jayanaka-98 (approved):
Good way of integration MTLLM in library mode. The tests are there. Should thin about the discrepancy between by llm code gen vs python integration.

PR #2305, opened 2025-07-10
🚅 Streaming token implementation for mtllm

ThakeeNathees:
```py

import from mtllm { Model }

glob llm = Model(model_name="gpt-4o-mini");

""" Generate short essay (less than 300 words) about the given topic """
def generate_essay(topic: str) -> str by llm(stream=True);


with entry {
    topic = "The orca whale and it's hunting techniques";
    for tok in generate_essay(topic) {
        print(tok, end='', flush=True);
    }
    print();
}

```

### DIFF
--- a/jac-mtllm/tests/test_mtllm.py
+++ b/jac-mtllm/tests/test_mtllm.py
@@ -86,6 +86,15 @@ class JacLanguageTests(TestCase):
         stdout_value = captured_output.getvalue()
         self.assertIn('The image shows a hot air balloon shaped like a heart', stdout_value)
 
+    def test_streaming_output(self) -> None:
+        """Parse micro jac file."""
+        captured_output = io.StringIO()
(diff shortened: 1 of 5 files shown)


jac-mtllm/tests/test_mtllm.py
Jayanaka-98, line 1: This test does not necessarily check the streaming functionality right, even if `stream=False` in the fixture, this test may still pass right? @ThakeeNathees

Jayanaka-98 (changes requested):

jac-mtllm/tests/test_mtllm.py
ThakeeNathees: I've added an assertion to ensure that it's not the actual string but stream

Jayanaka-98 (approved):
better test

PR #2309, opened 2025-07-10
Updates for new pypi

Thamirawaran:


jac-cloud/jac_cloud/core/memory.py
```diff
@@ -59,6 +59,11 @@ def populate_data(self, nodes: Iterable[NodeArchetype]) -> None:
                     cnodes.add(edge.target)
             self.find(cnodes)
 
+    def all_root(self) -> Generator[Root, None, None]:  # type: ignore[override]
+        """Get all the roots."""
+        for anchor in NodeAnchor.Collection.find({"name": ""}):
+            yield cast(Root, anchor.archetype)
+
```
marsninja: Would this return nodes that are not root nodes that happen to have a blank name field?

jac/examples/littleX/littleX.impl.jac
```diff
@@ -54,7 +54,7 @@ impl Tweet.remove_like {
 impl Tweet.comment {
         current_profile = [root-->(`?Profile)];
         comment_node = current_profile[0] +>:Post():+> Comment(content=visitor.content);
-        _.perm_grant(comment_node[0], level="CONNECT");
+        grant(comment_node[0], level="CONNECT");
```
marsninja: We should do somethign so this isnt a string. I honestly would be fine with something like `Jac.Connect` perhaps, or some kind of builtin enum, or something. I'd leave it to you but not a string for the prime example that is LittleX (i'd be ok with strings working but shouldnt be the main way we show)

marsninja (changes requested):
Couple quick ones...

jac/examples/littleX/littleX.impl.jac
Thamirawaran: Yeap. need to update

jac-cloud/jac_cloud/core/memory.py
Thamirawaran: Nope

Thamirawaran: it finds nodeanchor with "" name. root it the only one

PR #2328, opened 2025-07-15
[FEATURE]: Jac serve hot reload

amadolid:
# Allow hot reload
> jac serve jac_cloud/tests/simple_graph.jac --reload

## Initial Logs
image

## Modified simple_graph.jac
image

## Updated Logs
image


Thamirawaran:
can we preserve authentication settings

amadolid:
> will memory be preserved?

No, it will terminate existing process and spawn a new one.
This is similar to fastapi and uvicorn

amadolid:
> can we preserve authentication settings

If you are referring to token, if you don't have redis and just use the FakeRedis, it will only be available on the same process. If the the process is terminated, it will get cleaned up.

With redis, token will be retained even for multiple replicas/instances.

marsninja (approved):

marsninja:
Add a bit of docs in the relevant spot in the docs.

marsninja (approved):

PR #2397, opened 2025-07-23
[MEMORY]: Dedicated memory commit interface

amadolid:


marsninja (changes requested):
Tests not passing,

Need to use builtins, if you need urgent merge then create issue for improvement and put in progress.

PR #2412, opened 2025-07-24
Github - Read me file updated

RanuriG:
File updates with several changes
[jasecireadme_README.md at main · RanuriG_jasecireadme.pdf](https://github.com/user-attachments/files/21413257/jasecireadme_README.md.at.main.RanuriG_jasecireadme.pdf)


RanuriG:
> This is an improvement to me, made some small tweaks. I think you can merge this one if you'd like, however I think ultimately we'd want to have 1 key code example that folks can run, perhaps 2 or so examples from the tour could be used not sure.

Yes I agree better to add a .gif with the input and output.

Issue #2250, opened 2025-07-01
UniiR generation for python projects for MTLLM plugin usage


Jayanaka-98:
### 🔍 Motivation
To support **library mode** usage of `jaclang`, we need to ensure that any Python module that imports `jaclang` can be correctly analyzed by plugins or tools that rely on the **UniiR (Unified Intermediate Representation)**.

Currently, such tooling might only work when running `.jac` files directly, limiting their use in integrated Python workflows. By building the **UniiR** for regular Python files, we unlock full interoperability and plugin support for Python modules using `jaclang`.

### 💡 Proposal
Implement logic to construct a **UniiR symbol table** when:
- A Python file that imports `jaclang` is converted to jac uniiir
- Plugins or tools request UniiR-based inspection or transformation

This enables:
- 🔌 Seamless plugin compatibility in Python-based projects
- 🧠 Rich introspection and semantic tooling
- 📚 Better integration with editors, linters, and analysis frameworks

PR #2419, opened 2025-07-25
🔠 MTIR implementation for MTLLM

ThakeeNathees:


jac-mtllm/mtllm/mtir.py
```diff
@@ -0,0 +1,194 @@
...
+    @staticmethod
+    def factory(
+        caller: Callable, args: dict[int | str, object], call_params: dict[str, object]
+    ) -> "MTIR":
+        """Create an MTIR instance."""
+        # Prepare the tools for the LLM call.
+        tools = [Tool(func) for func in call_params.get("tools", [])]  # type: ignore
+
+        # Construct the input information from the arguments.
+        param_names = list(inspect.signature(caller).parameters.keys())
```
Jayanaka-98, line 55: MTIR should be created by scraping the UniiR before runtime

Jayanaka-98 (changes requested):
The MT-IR is supposed to be created per by callsite, before runtime, containing static semantics of the code, reconciling all sem keyword definitions gathered by scraping the MT-IR. 
image

This image represents the MT-IR for the by callsite 
image

At runtime, instead of getting information through python internals, assume we can only use the MTIR. We can use python internals for linking code locations to their relevent MTIRs.

**The point of MTIR is that regardless of the language, we can use compiler techniques (static analysis) to identify relevant semantics for by call sites at compile time which will be accessed at runtime to generate the prompts.**

jac-mtllm/mtllm/mtir.py
ThakeeNathees: The parameters can only be constructed at runtime, we can avoid using inspect and hard code the parameter names, which will result in requiring more code from the user:

Currently the library mode is like this (Implicit MTIR construction):
```py
@by(Model("gpt-4o")
def write_an_essay(topic: str) -> str:
     """Generate an essay"""
```

if we require the parameter to be hard coded, it would be something like:

```py
def write_an_essay(topic: str) -> str:
    """Generate an essay"""
    mtir = MTIR(
      params={ "topic" : topic },
      semstr="Generate an essay",
      return_type=str,
    )
    return _.call_llm(Model("gpt-4o"), mtir)
```

let me know if this is an intended trade off?

Jayanaka-98: No,  @by decorator is better, but we need to figure out a way so that when using python mode, we generate the MTIR for even python files and attach it to the by decorator at runtime. (Similar to how we do for jaclang)

Jayanaka-98: #2250 This Issue describes it

ThakeeNathees: So, the generation of MTIR in (pure) python, should be exists somewhere (file or in memory) and the `by` decorator should attach that mtir to the function call?

Are we supposed to configure/override python in a way that it generate information (MTIR) before executing the code?

PR #2445, opened 2025-07-28
CFG Bug Fix for returns & hasvar and if without else

Jayanaka-98:


jac/jaclang/compiler/passes/main/cfg_build_pass.py
```diff
@@ -299,3 +301,21 @@ def printgraph_cfg(self) -> str:
...
+
+    with open(file_name, "r") as f:
+        file_source = f.read()
+
+    ir = (prog := JacProgram()).compile(use_str=file_source, file_path=file_name)
+
+    cfg_pass = CoalesceBBPass(
+        ir_in=ir,
+        prog=prog,
+    )
```
marsninja, line 318: Is this function called anywhere?

marsninja (changes requested):
Needs a test.

jac/jaclang/compiler/passes/main/cfg_build_pass.py
Jayanaka-98: Yes, for a given jac file, the cfg can be genarated as a dot graph

Jayanaka-98: Used for all CFG tests

Jayanaka-98:
Will build teh CFG printer with #2463

PR #2448, opened 2025-07-29
keyword section added

RanuriG:
## **Description**
image
image
image
image
image
image
image

### DIFF
--- a/docs/docs/learn/keywords.md
+++ b/docs/docs/learn/keywords.md
@@ -1,0 +1,146 @@
+# Jac Programming Language Keyword Reference
+
+ Keywords are special reserved words that have a specific meaning and purpose in the language and cannot be used as identifiers for variables, functions, or other objects.
+
+---
+
+## 1. Archetype and Data Structure Keywords
+
+These keywords define the core data and structural elements in Jac, forming the foundation for graph-based and object-oriented programming.
+
+**Core Archetype Keywords**
+
+| Keyword | Description |
+| --- | --- |
+| `obj` | Defines a standard object, similar to a Python class, for holding data and behaviors. |
+| `node` | Represents a vertex or location in a graph, capable of storing data. |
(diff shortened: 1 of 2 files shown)


marsninja (changes requested):
I dont know if this is in the right place in the documentation. Perhapw it should be in teh specificantion section, or you can create a cheatsheets section, or maybe the specification section should be called the cheatsheets section, have a chat with @Jayanaka-98 and @kugesan  to see what makes sense. I'll make @Jayanaka-98 as a reviewer.

docs/docs/learn/keywords.md
```diff
@@ -0,0 +1,145 @@
...
+These keywords facilitate the integration of AI and Large Language Models (LLMs) directly into the language.
+
+
+**AI Integration**
+
+
+| Keyword | Description |
+| --- | --- |
+| `sem` | Associates a natural language "semantic string" with a code element for AI interpretation. |
+| `by llm` | Indicates a function's implementation will be provided by an LLM. |
```
Jayanaka-98: the keyword is `by` llm can be anything. ex: `by qwen_model_that_i_have()`

Jayanaka-98 (changes requested):
I think  this is a must have. 

Lets start with stating that all keywords used in python are also usable in jaclang.

Then introduce keyword sets per each usecase. I suggest that hyperlinks should be added for evry keyword description so that they can visit the docs and see how to use tyhe keyword. The hyperlink must be to jac reference doc, jacbook or to specific docs like mtllm.

docs/docs/learn/keywords.md
Jayanaka-98: `by` is the keyword, not `llm`. Also it is not accurate to say the "implementation will be provided by llm". Rather say "task is deferred to an LLM"

Jayanaka-98 (dismissed):
For each keyword add a hyperlink to the section in the specification.

PR #2483, opened 2025-07-31
MTP hot fix

Thamirawaran:


kugesan1105 (approved):
We need this piece to make the elegant agentic code!
[ also this  is a fix as `parse_response` is taking str as arg]

PR #2494, opened 2025-08-01
Ability execution order 

Thamirawaran:


amadolid:
Minor:
Please avoid using function declaration inside a function if it's a "static process". Merge it on the current function or declare it outside like utility.

ex: spawn's `collect_targets` and `assign`

PR #2508, opened 2025-08-05
included a function for apply_partial_indexes

juzailmlwork:


jac-cloud/jac_cloud/core/archetype.py
```diff
@@ -1063,6 +1067,8 @@ class BaseArchetype:
 
     __jac_classes__: dict[str, type["BaseArchetype"]]
     __jac_hintings__: dict[str, type]
+    __constraints__: Constraints = {}
```
amadolid: It might be preferable to use normal `Class` instead of dict/TypedDict.

One reason is it requires the import of `Constraints` class to be able to use it.

Normal class doesn't need import and can leverage inheritance in case needed.
```
class __constraints__ {
    has unique: list[tuple] = ....;
}
```

PR #2569, opened 2025-08-20
Code block enabling input handling | Getting Inputs from the User

MalithaPrabhashana:
This task is about enabling interactive input() support when running Jac code inside the Pyodide worker. Currently, code execution works, but when input() is called, the Python side blocks and cannot receive the user’s response from the main thread. The goal is to implement a communication mechanism (via postMessage, SharedArrayBuffer, and Atomics) that lets the worker pause for input requests, prompt the user in the browser, and then resume execution with the provided string. This will make the Jac playground interactive for programs requiring user input.

### DIFF
--- a/docs/docs/js/pyodide-worker.js
+++ b/docs/docs/js/pyodide-worker.js
@@ -130,8 +130,6 @@ except Exception as e:
 # Restore original streams
 sys.stdout = original_stdout
 sys.stderr = original_stderr
-
-"Execution completed"
         `);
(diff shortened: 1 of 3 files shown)


MalithaPrabhashana:
This is the way, this is planned

image

MalithaPrabhashana:
Can you try this @kugesan1105 @Thamirawaran @RanuriG

RanuriG:
@marsninja merge ready, do not close the issue however

marsninja (approved):

PR #2576, opened 2025-08-21
Code block accept Inputs - static way

MalithaPrabhashana:

### DIFF
--- a/docs/docs/js/pyodide-worker.js
+++ b/docs/docs/js/pyodide-worker.js
@@ -32,7 +32,7 @@ except Exception as e:
 
 // Worker code
 self.onmessage = async (event) => {
-    const { type, code } = event.data;
+    const { type, code, inputs } = event.data;
 
     if (type === "init") {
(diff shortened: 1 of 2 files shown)


MalithaPrabhashana:
image

MalithaPrabhashana:
Yeah if there is no inputs need to be given to a code, then we can keep this box empty otherwise need to enter the inputs.
- comma separated
- blank separated
- new line separated

Yeah, that issue is there with this approach. Initially they have to exactly enter all the required inputs

RanuriG:
Yeah that's why dynamic inputs are better, as the user can enter them in a new line each time. 
Let's hold this for now without merging check with Thakee on the dynamic inputs. Once it's done let me know, if it cannot be done let's go for this approach with better instructions to the user.

MalithaPrabhashana:
Yeah that'll be great. Will check with Thakee and let you know the update.

Thamirawaran:
image
It needs to pause during the execution. is in't it?
image

MalithaPrabhashana:
Implemented the dynamic input in PR #2569.

closing this PR

PR #2591, opened 2025-08-21
⌨️  Unknown type assignment are allowed

ThakeeNathees:
## **Description**
---------------------------
Since we are in the start of our typechecking journey, strict evaluation of unknown types is temporarily relaxed.
Assignments with unknown types will be treated as valid to avoid unnecessary failures until stricter rules are introduced later.

--------------------------

cehck_page-0001


kugesan1105:
good to merge!!

PR #2595, opened 2025-08-21
🌏 Member Access for typecheck implemented

ThakeeNathees:


ThakeeNathees:
In the above screenshot, we have defined the `Foo` class twice and both has different version of the `bar` (`int` and `Bar` object), thats why line 27 is failing. It should be in a different file. @kugesan

PR #2640, opened 2025-08-27
Infer Language from cmd

Thamirawaran:


kugesan1105 (approved):
Its good to merge!
The changes in `splice_plugin.py` make sense, since it’s installed alongside `jaclang `in our `test-jaseci.yml`
Because this setup includes `jac_import`, it will override Jac’s default `jac_import`, which explains why lng is updated in `plugin/splice_plugin.py`.

image

image

marsninja (approved):

PR #2681, opened 2025-09-01
Refactor AI integration in landing.js: update import statements and s…

udithishanka:
…treamline translation function


RanuriG:
Great I feel like better to add a comment mentioning that notice we didn't define a prompt the fucntion generates it's own function automatically.
What do you think about this @udithishanka

RanuriG (approved):

PR #2813, opened 2025-09-16
If else expr formatting fix

mgtm98:
## **Description**
try to fix the issue in #2803

PR #2850, opened 2025-09-21
Fix calculating current column width in case of indentation blocks

### LINKED ISSUE #2849: [Bug] Issue in calculating remaining width in case of Indent node
## **Bug Details**
In `JacFormatPass` remaining width is not calculated properly in case of an `Indent` node  in the following example
```
Concat([
    Text("("),
    Indent(
        Concat([
            line(),
            Text("aaa")
        ])
    )
])
```
While printing "aaa" the current width should be 2 but with the current implementation, the current width would be 0 as `format_doc_ir` would be called with `width_for_indented_content`, which doesn't take into consideration the current width

mgtm98:
## **Description**
This PR is to fix #2849


mgtm98:
I can't add tests here as the issue started to appear after the formatting fixes done in #2813

PR #2861, opened 2025-09-23
Jac-lang deployment to the new jaseci-cluster

udithishanka:


.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,222 @@
+name: Build and Deploy jac-lang.org Documentation
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - jaclang-deploy
```
AshishMahendra: I think we should only run on the release of Jaclang docs.

.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,222 @@
...
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.12'
+
+      # Cache pip dependencies
+      - name: Cache pip dependencies
+        uses: actions/cache@v3
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('docs/requirements.txt') }}
```
AshishMahendra: Check if we have a requirement file

.github/workflows/deploy-docs.yml
udithishanka: Jason expects to run it on push to the main branch. having to make a separate release is kind of a hassle. currently also, it is set as on push to main branch

.github/workflows/deploy-docs.yml
udithishanka: changed to use pyproject.toml

.github/workflows/deploy-docs.yml
AshishMahendra: Okay, i think we won't need the jaclang-deploy branch in the script

udithishanka: yep. removed. included it earlier to test the deployment before merging.

AshishMahendra (approved):
LGTM

PR #2922, opened 2025-09-29
🏗️ [Refactor] Type evaluator moved to program member

ThakeeNathees:
Type evaluator is currently a singleton (Which is an anti-pattern) that is problematic when there are multiple programs instances are created in a single execution since, it assumes and have a persistant reference to the one program it was created with, this break certain tests when the tests ran parallelly with different programs, so this refactor is necessary to address.

### DIFF
--- a/jac/jaclang/compiler/program.py
+++ b/jac/jaclang/compiler/program.py
@@ -32,6 +32,7 @@ from jaclang.compiler.passes.tool import (
     FuseCommentsPass,
     JacFormatPass,
 )
+from jaclang.compiler.type_system.type_evaluator import TypeEvaluator
 from jaclang.runtimelib.utils import read_file_with_encoding
 from jaclang.settings import settings
 from jaclang.utils.log import logging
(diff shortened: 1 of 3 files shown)


jac/jaclang/compiler/program.py
```diff
@@ -66,6 +67,13 @@ def __init__(self, main_mod: Optional[uni.ProgramModule] = None) -> None:
         self.py_raise_map: dict[str, str] = {}
         self.errors_had: list[Alert] = []
         self.warnings_had: list[Alert] = []
+        self.type_evaluator: TypeEvaluator | None = None
+
+    def get_type_evaluator(self) -> TypeEvaluator:
+        """Return the type evaluator."""
+        if not self.type_evaluator:
+            self.type_evaluator = TypeEvaluator(program=self)
+        return self.type_evaluator
```
ThakeeNathees, line 76: Since the type evaluator will load the builtin module and prefetch some symbols, we don't necessarily have to do that every time we create an instance, instead we lazy load and only run when type check if requested

kugesan1105 (approved):

PR #3041, opened 2025-10-10
LiteLLM Lazy loading

ThakeeNathees:


jac-byllm/byllm/llm.py
```diff
@@ -48,6 +47,8 @@ def __init__(self, model_name: str, **kwargs: object) -> None:
             api_key: API key for the model provider
             **kwargs: Additional configuration options
         """
+        from .llm_connector import LLMConnector
+
```
ThakeeNathees, line 51: `llm_connector` was the bottleneck, importing this inside the function is sufficient.

PR #3103, opened 2025-10-17
Support empty body  match case pattern

Thamirawaran:


marsninja (approved):
Very nice pr @Thamirawaran

PR #3174, opened 2025-10-28
VITE Bundler poc

musab-mah-7:


Thamirawaran (approved):
Kudos @musab-mah-7 :heart:

PR #3264, opened 2025-11-02
New keyword support for onelang

Thamirawaran:


jac/jaclang/compiler/passes/ecmascript/esast_gen_pass.py
```diff
@@ -1545,11 +1548,18 @@ def exit_func_call(self, node: uni.FuncCall) -> None:
                     ),
                     jac_node=node,
                 )
-
-        call_expr = self.sync_loc(
-            es.CallExpression(callee=callee, arguments=args), jac_node=node
-        )
-        node.gen.es_ast = call_expr
+        callee_type = TypeEvaluator(self.prog).get_type_of_expression(node.target)
```
ThakeeNathees: creating type evaluator instance each time is expensive, only construct once and call this

PR #3265, opened 2025-11-02
Create an video for better README.md file to Understand workflow of b…

SupulHeshan:
# byLLM

## **Description**
- Added a video to README.md to understand workflow of byLLM

https://github.com/user-attachments/assets/624ef599-dff0-4728-b842-26911b5c827a


SupulHeshan:
- The reason I added this as a video instead of a GIF is that when this flow is going, anyone can pause the video and see what the popup is.

PR #3333, opened 2025-11-12
Filter Comprehensation type check

ThakeeNathees:
## **Description**

Goto definition works for `SomeNode` inside the filter comprehensation

image


marsninja:
Already merged, W!

PR #3370, opened 2025-11-17
Fixed Minor Typos

TheJoshBrod:


marsninja (changes requested):
Not sure if we're still planning to merge this, if we are we need svg instead of png (or mermaid).

PR #3376, opened 2025-11-17
optimize png files

asottile:
this will technically increase the size of the git repo -- but I will use the optimized files when I rewrite the history


asottile:
ok so while this feels like a good idea -- it adds almost 1.5 minutes to CI.  not quite sure the best approach to this :(

PR #3477, opened 2025-11-25
[Feat][Docs] Implement live reload and improved rebuild handling

Jayanaka-98:
Introduce live reload functionality for the MkDocs site and enhance the rebuild process to prevent 404 errors during rebuilding.


docs/scripts/mkdocs_serve.py
```diff
@@ -140,10 +253,11 @@ def serve_with_watch() -> None:
 
     print("Initial build of MkDocs site...")
     subprocess.run(["mkdocs", "build"], check=True, cwd=root_dir)
+    reload_queue.put_nowait("reload")
 
     # Set up file watcher
     event_handler = DebouncedRebuildHandler(
-        root_dir=root_dir, debounce_seconds=3600, ignore_paths=ignore_paths
+        root_dir=root_dir, debounce_seconds=1, ignore_paths=ignore_paths
```
marsninja, line 258: Isnt this the only thing we really need, this pr feels overly complex.

marsninja (changes requested):

docs/scripts/mkdocs_serve.py
Jayanaka-98: I have added two more features 
1. Auto refresh the pages after a change (without manually hitting refresh)
2. When trying to refresh during a rebuild, a 404 would appear, but the PR added "Rebuild in Progress" screen, so not that blaring. 

Can remove if unnecessary.

marsninja (approved):
I'll approve, but we should add tests for this script and make sure we're not over complicating in general.

PR #3517, opened 2025-11-27
precommit for type checking and formating added

ThakeeNathees:
PR description:
- Summary:
  - Add two tiny repository wrapper scripts and update  so pre-commit hooks run the repository source for  formatting and build checks instead of attempting to pip-install the package inside hook virtualenvs.

- Why:
  - Hook venv pip-installs of local paths (e.g., ) proved fragile in CI and local environments (path resolution, Python/pip version mismatches, editable install differences). Pre-commit attempted to install  and failed in some environments.
  - The wrapper scripts run the package via  so hooks execute the repository code directly without requiring a distribution build or install step.
  - This approach is deterministic across CI and developer machines, faster (no install on every hook run), and easier to debug.

- What changed:
  - Added  (runs  with repo on ).
  - Added  (runs  on staged files; prints failures but exits 0 so the hook is non-blocking).
  - Updated  to use these wrappers as  and  (set to  so pre-commit runs the script directly).
  - Kept  behavior non-blocking to avoid blocking commits for CI-only build problems while still surfacing failures.

- Notes / Alternatives:
  - Long-term, adding a top-level packaging config and exposing  as an installable console script would allow  with . This is a clean solution but requires ensuring builds/install behavior in CI.
  - If you prefer the hooks to fail hard, we can switch  to exit non-zero on failure.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -155,11 +155,17 @@ repos:
   - repo: local
     hooks:
 
+      - id: install-jac
+        name: install jaclang locally
+        entry: pip3 install -e ./jac/
+        language: system
+        pass_filenames: false
+        always_run: true
+
       - id: jac-format
         name: jac format
         entry: jac format
-        language: python
-        additional_dependencies: ['./jac']
+        language: system
         types: [file]
         files: \.jac$
         pass_filenames: true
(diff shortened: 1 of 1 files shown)


scripts/precommit-jac-check
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
```
ThakeeNathees, line 1: Ignore these two files, i was trying to do a wrokaround and it didnt work

PR #3532, opened 2025-11-28
Add installation of 'pluggy' package in Pyodide worker

SupulHeshan:
## **Pyodide Worker**

- Fix: Added pluggy dependency to Pyodide worker for mkdocs.
- Added `pluggy` to project dependencies in pyproject.toml
- Purpose: To enable mkdocs local run with Pyodide pluggy dependency.
- I guess this is better to add if something happens suddenly in the server as well.

### DIFF
--- a/docs/pyproject.toml
+++ b/docs/pyproject.toml
@@ -18,7 +18,8 @@ dependencies = [
     "uvicorn",
     "watchdog",
     "mkdocs-graphviz",
-    "mkdocs-material-extensions"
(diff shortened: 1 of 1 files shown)


SupulHeshan:
## Additional Changes 

- In `guess_game3.jac` & `guess_game3.impl.jac` files
- `attempts` and `correct_number` are initialized directly in the class definition, reducing the need for extra code in the constructor.

Issue #2435, opened 2025-07-28
[Bug] [py2jac] FStrings not correct converted


kugesan1105:
<img width="1812" height="547" alt="Image" src="https://github.com/user-attachments/assets/062856a3-357a-444c-865d-41b70de5a82b" />

PR #3533, opened 2025-11-28
fixed issue 2435, added regression test

kevinjin420:
fixed py2jac triple-quoted f-string conversion, #2435


marsninja (approved):
CLUTCH @kevinjin420

PR #3693, opened 2025-12-10
Add test case for prompt formatter

SupulHeshan:
## **Description**
- Added test case for prompt formatter in byLLM.
- Added those inside existing `test_params_format` test.


jac-byllm/byllm/llm.impl/mockllm.impl.jac
```diff
@@ -6,6 +6,8 @@ impl MockLLM.init(model_name: str, **kwargs: object) -> None {
 
 impl MockLLM.dispatch_no_streaming(mtir: MTIR) -> CompletionResult {
     params = self.make_model_params(mtir);
+    message = """{"role": "assistant", "content": {"text": "Mock response"}}""";
```
MalithaPrabhashana: No need to represent this as a doc string.

SupulHeshan: Sure

PR #3743, opened 2025-12-16
The Big Jac Toml Project

marsninja:


juzailmlwork:
@marsninja one of the main features I like to have is flag some of the configueratuons made mandatory to set it up.so when a new user try to run the application for first time he should be asked to set the values for those configs(example openaikey,mongodb url,api-key etc)

PR #3867, opened 2025-12-30
Fix jac-client assets build for .js and .py files

SahanUday:
This pull request adds support for copying asset files based on project configuration(`jac.toml`), rather than using a fixed set of file types. The main change is the introduction of a new method that reads asset file extensions from the `jac.toml` configuration and copies only those files. 

**Asset processing improvements:**

* Added a new method `copy_configured_assets` to the `AssetProcessor` class, which copies asset files from a source to a destination directory based on file extensions specified in the project configuration (`jac.toml`). [[1]](https://github.com/jaseci-labs/jaseci/compare/main...SahanUday:jac_client-assets?expand=1#diff-5abcbb4daa780e0b033e768fd9573c36de2e4dab03de17c0c983151d00c75db8R26-R32) [[2]](https://github.com/jaseci-labs/jaseci/compare/main...SahanUday:jac_client-assets?expand=1#diff-d790a505416a1267fe27be30333497955fab95e6392f692619438a33bec476a2L1-R48)
* Updated the Vite compiler (`ViteCompiler.copy_root_assets`) to use `copy_configured_assets` for copying assets from the root assets directory to the build assets directory, ensuring only configured asset types are copied.

**Configuration and dependency changes:**

* Imported `get_config` from `jaclang.project.config` in `asset_processor.jac` to support reading asset configuration.

**Server asset handling improvements:**

* Modified the API server's static file handler to load the list of asset extensions from the `jac.toml` , and use these extensions (in addition to a default set) to determine whether a request path should be treated as an asset file.

### DIFF
--- a/jac-client/jac_client/docs/advance/custom-assets-handling.md
+++ b/jac-client/jac_client/docs/advance/custom-assets-handling.md
@@ -4,7 +4,7 @@ Configure custom asset types for your Jac Client build process to support Web Wo
 
 ## Overview
 
-By default, Jac Client handles common asset types like images, fonts, and CSS files. With custom asset configuration, you can extend this to include any file type, such as `.js`, `.py`, `.wasm`, or `.json`. 
+By default, Jac Client handles common asset types like images, fonts, and CSS files. With custom asset configuration, you can extend this to include any file type, such as `.js`, `.py`, `.wasm`, or `.json`.
 
 ## Quick Start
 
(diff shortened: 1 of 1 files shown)


jac-client/jac_client/plugin/src/asset_processor.jac
```diff
@@ -22,4 +23,11 @@ class AssetProcessor {
         dest_dir: Path,
         preserve_structure: bool = True
     ) -> None;
+
+    def copy_configured_assets(
```
musab-mah-7: copy_extended_custom_file_types_in_toml , a simlar shorter name

jac/jaclang/runtimelib/impl/server.impl.jac
```diff
@@ -749,6 +749,12 @@ impl JacAPIServer.create_handler -> type[BaseHTTPRequestHandler] {
                 return;
             }
             is_static_path = path.startswith('/static/');
+            import from jaclang.project.config { get_config }
+            ;
+            config = get_config();
+            client_cfg = config.get_plugin_config("client") if config else None;
+            assets_config = client_cfg.get("assets", {}) if client_cfg else {};
+            extensions = set(assets_config.get("extensions", []));
```
musab-mah-7: custom_file_extensions

musab-mah-7 (changes requested):

musab-mah-7 (approved):

PR #3868, opened 2025-12-30
Support client file without main file

Thamirawaran:
## **Description**
cl file can exits without main file
`component1.cl.jac` file can exists without `component.jac` file


Thamirawaran:
Core idea is that cl files can exits without main file
`component1.cl.jac` file can exists without `component1.jac` file

PR #3875, opened 2025-12-30
Prevent race condition in test_user_isolation test

akindu-k:
This pull request introduces a minor change to the `test_user_isolation` test in `jac_scale/tests/test_serve.py`. A short delay is added after the first user registration.

* Testing reliability:
  * Added a `time.sleep(5)` call after registering the first user in `test_user_isolation` to improve test stability.


ThakeeNathees (changes requested):
the server should be fault tolerant, adding delay in client wont fix the core issue in the server. Jac scale supposed to be stress tested with 100s of concurrent requests per second.

PR #3876, opened 2025-12-30
Fix: Log appropriate errors for disallowed 'pass' keyword in jac

SupulHeshan:
## **Add Error Detection for Unsupported `pass` Keyword**
- This implementation not only for 'pass' keyword but also other keywords only in python not allowed in jac.
- [x] (1) Detects `pass` in functions, if/else blocks, try/except blocks.
- [x] (2) Detects `pass` in obj/class/node definitions.
- [x] (3) Detects `pass` as variable name in assignments and expressions.(also other keywords not allowed in jac but reserved)
- [x] (4) Detects `pass` used as a parameter name.(also other keywords not allowed in jac but reserved)
- [x] (5) Detects `pass` used as a function name. (also other keywords not allowed in jac but reserved)

Examples:
(1)image
(2)image
(3)image
(4)image
(5)image

## **Add Test Cases**

- [x] Test cases for detecting `pass` in functions, if/else blocks, try/except blocks.
- [x] Test cases for detecting `pass` in obj/class/node definitions.

## **Additional Changes**
- Removed unwanted and unsupported `pass` keyword from `cli.impl.jac` file.
- Removed unwanted and unsupported `pass` keyword from `memory.impl.jac` file
- Refactored the `new_keyword_errors.jac` file to have more jactastic.


jac/jaclang/pycore/jac_parser.py
```diff
@@ -231,6 +231,15 @@ def feed_current_token(iparser: jl.InteractiveParser, tok: jl.Token) -> bool:
...
             )
+
+            # Check for unsupported 'pass' keyword
+            if e.token and e.token.value == "pass":
+                self.log_error(
+                    "'pass' is not supported. Jac allows empty blocks instead",
+                    self.error_to_token(e),
+                )
+                return True
+
```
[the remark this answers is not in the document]
SupulHeshan: fixed

SupulHeshan:
> What happens if you run this?
> 
> ```python
> pass = "foo";
> pass;
> ```
Yes there was an issue here. And I came up with a fix.
image

What do you think if these parse like this?

ThakeeNathees:
We need to have a list of python keywords, dont have to worry about those that we also have but other keywords should not be used in any context as we cannot compile them so doing this one by one will not be the correct approach.

currently it's `def _check_pass_in_expr(` and this function is too specific, we almost never want to handle any special cases, this shold be generalized like this:

This should be the logic.

```py
for kw in all_keywords_in_python.exclude(jaclang_keywords):
  if kw.is_used_as_name:
    error("some syntax error")
```

Note that jaclang keywords can be used in python with escape `<>node` because our target is python and it wont cause any issue but not the vice versa

jac/jaclang/pycore/jac_parser.py
```diff
@@ -40,6 +41,87 @@
...
+    "or",
+    "and",
+    "not",
+    "True",
+    "False",
+    "None",
+}
+
+# Python keywords that should NOT be used as names in Jac
+PYTHON_ONLY_KEYWORDS = set(keyword.kwlist) - JAC_KEYWORDS
```
ThakeeNathees: Dont hard code these values, find a way to get these values from the keyword list in some other files.

jac/jaclang/pycore/jac_parser.py
```diff
@@ -231,6 +313,24 @@ def feed_current_token(iparser: jl.InteractiveParser, tok: jl.Token) -> bool:
...
             )
+
+            # Check for unsupported 'pass' keyword in code blocks
+            if (
+                e.token
+                and e.token.value == "pass"
+                and (Tok.RBRACE.name in e.accepts or Tok.SEMI.name in e.accepts)
+            ):
+                self.log_error(
+                    "'pass' is a keyword not allowed in jac",
```
ThakeeNathees: the error message should be something like: (caue pass is not a keyword in jaclang):

Pass is a python keyword and codegen target is python so it's not allowed

jac/jaclang/pycore/jac_parser.py
```diff
@@ -326,6 +426,21 @@ def _node_update(self, node: T) -> T:
...
+            if isinstance(expr, uni.Name) and expr.sym_name in PYTHON_ONLY_KEYWORDS:
+                self.parse_ref.log_error(
+                    f"'{expr.sym_name}' is a keyword not allowed as {context} in jac",
+                    node_override=expr,
+                )
+            elif hasattr(expr, "kid") and expr.kid:
+                # Recursively check all child nodes
+                for child in expr.kid:
+                    if isinstance(child, (uni.Expr, uni.Name)):
+                        self._check_python_keywords_in_expr(child, context)
```
ThakeeNathees: This should be checked in the parser just after a name token is created, definitely not recusively

jac/jaclang/pycore/jac_parser.py
```diff
@@ -1426,6 +1555,26 @@ def statement(self, kid: list[uni.UniNode]) -> uni.CodeBlockStmt:
...
+                        "'pass' is a keyword not allowed in jac",
+                        node_override=kid[0],
+                    )
+
+                    self.parse_ref.log_error(
+                        "If need an empty code block, simply leave it empty.",
+                        node_override=kid[0],
+                    )
+                else:
+                    # Check if Python keywords are used in the expression (ex: print(pass); )
```
ThakeeNathees: Why this logic in 2 different places (duplicates)? cant we have a single common place to have this logic?

ThakeeNathees (changes requested):

jac/jaclang/pycore/jac_parser.py
SupulHeshan: If this added after a name token is created, It will catch  'pass' in everywhere except only in the classes/obj/node those.

then It will catch 'pass' inside the if else/try except/for loop , functions/methods as well as variables, parameters, identifiers.. 
If we give error with solution suggestion like: "If need an empty code block, simply leave it empty" , we cannot be able to separate them. cause this suggestion is not compatible with variables, identifiers, parameters etc..

if we don't need to give suggestion like that, even for the functions, methods,  then it will be fine.

jac/jaclang/pycore/jac_parser.py
SupulHeshan: If we don’t need to provide solution suggestions, this entire code section can be removed.

SupulHeshan: This part to catch the 'pass' keyword inside the functions/methods, if else/try except/for loop

jac/jaclang/pycore/jac_parser.py
SupulHeshan: Changed it by defining a function.

jac/jaclang/pycore/jac_parser.py
SupulHeshan: changed the error message.

jac/jaclang/pycore/jac_parser.py
SupulHeshan: removed this unnecessary function. defined another one and checked after a name token is created.

jac/jaclang/pycore/jac_parser.py
SupulHeshan: Removed this logic as well

kugesan1105 (approved):
good to merge

PR #3882, opened 2025-12-31
Support Custom headers in the response

Thamirawaran:
## **Description**
validated shared buffer array support with @ahzan-dev and @MalithaPrabhashana 
image

### DIFF
--- a/jac-client/jac_client/examples/all-in-one/jac.toml
+++ b/jac-client/jac_client/examples/all-in-one/jac.toml
@@ -18,3 +18,7 @@ antd = "^6.0.0"
 [dependencies.npm.dev]
 "@types/react" = "^18.2.0"
 "@types/react-dom" = "^18.2.0"
+
+[environments.response.headers]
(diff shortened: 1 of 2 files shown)


jac-client/jac_client/plugin/impl/client.impl.jac
```diff
@@ -21,6 +22,13 @@ impl JacClient.send_static_file(
         handler.send_header('Content-Length', str(len(file_content)));
         handler.send_header('Cache-Control', 'public, max-age=3600');
         ResponseBuilder._add_cors_headers(handler);
+        config = get_config();
+        if config
+        and config.environments
+        and "production" in config.environments
+        and config.environments["production"].get("security_headers") {
```
musab-mah-7: this key is not a good choice

let's have like this

[headers.security]

Thamirawaran: sure

musab-mah-7 (approved):

PR #3915, opened 2026-01-02
Fix: Allow all types to be assigned to object parameter

kugesan1105:
### Error Change 1003 ---> 943

------------
Before
----------------
image

After
------------------
Screenshot 2026-01-03 003040


ThakeeNathees (approved):
lgtm

PR #3933, opened 2026-01-05
feat(router): make walker routing prompt support single or multi node indexes

Hirudika4:
This PR updates the visitable node routing logic to use semstrings instead of embedding routing behavior as a prompt. The routing intent is now expressed through a high level semstring with structured inner semstrings.


MalithaPrabhashana (approved):

jac/jaclang/pycore/runtime.py
```diff
@@ -1632,13 +1632,22 @@ def _filter_visitable_by(
             connected_nodes: list[NodeArchetype], descriptions: str = ""
         ) -> list[int]:
             """
-            Determine which connected nodes are visitable using an LLM.
+            You are a ROUTER for an agentic walker.
```
Jayanaka-98: This no loger a docstring but an explicit prompt.

jac/jaclang/pycore/runtime.py
```diff
@@ -1632,13 +1632,22 @@ def _filter_visitable_by(
...
 
-            The input represents structurally reachable nodes. This function applies
-            semantic reasoning to decide which of those nodes a walker is allowed
-            or intended to visit, returning their indexes in priority order.
+            Input: user_input, reachable_nodes[{index,name,description,...}], incl_info{stage/visited/output_mode/constraints,...}.
```
Jayanaka-98: This much detail should be done over semstrings rather than adding everything onto the docstring/top-level semstring. (Should add as inner semstrings)

Jayanaka-98:
Also, the incl_info constraints should not be explained in the prompt due to the general nature of incl_info. (Anything can be passed through incl_info) User may decide to use it differently which may throw off the LLM.

jac/jaclang/pycore/runtime.py
Hirudika4: I've refactored the routing logic to remove the prompt style docstring and express the behaviour using semstring instead.

jac/jaclang/pycore/runtime.py
Hirudika4: So the routing intent is now captured via a semstring with inner semstrings without assuming any application specific context ( incl_info structure or stages)

MalithaPrabhashana:
@Jayanaka-98 Noticed one issue with inner_sem_strings. Currently the inner_sem_strings are passed to the LLM, only if they are defined with the tools. Not with the main by llm call.

Jayanaka-98 (approved):
This is a better approach

PR #3958, opened 2026-01-06
Test with pypi build

Thamirawaran:


conftest.py
```diff
@@ -17,7 +17,7 @@
 from jac.tests.fixtures_list import MICRO_JAC_FILES
 from jaclang.runtimelib.utils import read_file_with_encoding
 
-_JACLANG_DIR = Path(jaclang.__file__).parent
+_JACLANG_DIR = Path(__file__).parent.joinpath("jac", "jaclang")
```
Thamirawaran: This PR aims to run tests with pypi build package(python build). So jaclang lib files(compiler) is in sitepackages. If we try to use jaclang folder that will direct to sitepackages. But pypi build doesn't contains examples folder. To use examples folder we have to specify its path independent of jaclang lib.

PR #4006, opened 2026-01-09
set k8s branch to main of jaseci repo

juzailmlwork:


RandiJay96:
this will cover under #4033

Issue #4000, opened 2026-01-09
Usage of .tsx and .css Files


RandiJay96:
## Usage of `.tsx` and `.css` Files

When a `.tsx` file is used inside `app.jac`, it compiles successfully. However, if the same file is used anywhere other than `app.jac`, the build fails with a compilation error.

### Example

If we have `homePage.jac` (`/src/pages/homePage.jac`) and it imports `Card.tsx` (`/src/components/Card.tsx`), the build fails. As a result, `Card.tsx` does not appear in the `.jac/compiled` folder.

The same issue occurs with `.css` files. For example, if `landing.css` is located in `/src/styles/landing.css` and is imported anywhere other than `app.jac`, the build fails.

In both cases, copying the `.tsx` or `.css` files into the correct location allows the project to build successfully.

### Folder Structure

```text
project-root/
├── .jac/
│   └── compiled/
│       └── (generated files)
└── src/
    ├── app.jac
    ├── pages/
    │   └── homePage.jac
    ├── components/
    │   └── Card.tsx
    └── styles/
        └── landing.css

PR #4022, opened 2026-01-11
Enhance usage of .tsx and .css Files

SahanUday:
## **Description**

**Issue:** #4000 
The issue arises because `manifest.imports` contain import paths without proper file extensions for imports of `.ts, .tsx, .js, .css` files from the custom directory to nested file. This caused paths like like this `/home/user/project/components/Button/tsx` instead of `/home/user/project/components/Button.tsx`
This  cause to nested imports failing to be copied

**Fix:**
If the last segment of the path (the filename) is just an extension like `tsx, ts, js, css, scss`, identified that malformed nested imports and replaced the last segment with the correct file suffix.

**Support for CSS/SCSS files:**

* Added a new method `_copy_css_file` to the `ViteCompiler` class for copying CSS files to the compiled directory.
* Updated the dependency compilation logic to detect `.css` and `.scss` files and invoke the new `_copy_css_file` method, including a debug print statement for visibility.

### DIFF
--- a/jac-client/jac_client/docs/file-system/ts_js_css-imports.md
+++ b/jac-client/jac_client/docs/file-system/ts_js_css-imports.md
@@ -33,7 +33,7 @@ my/
 JAC uses a dot-prefix notation for relative imports, where each dot represents a directory level:
 
 - `.` - Current directory
-- `..` - Parent directory  
+- `..` - Parent directory
 - `...` - Two levels up
 - `....` - Three levels up, and so on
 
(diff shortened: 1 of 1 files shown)


jac-client/jac_client/docs/file-system/ts_js_css-imports.md
```diff
@@ -0,0 +1,232 @@
+# TS, JS and CSS File Imports
+
```
musab-mah-7: Add a warning, here to perefer jac over these ts and stuffs unless necessory, pls check there;s a documentaiton for it

PR #4031, opened 2026-01-12
Add developer guide doc for custom file extensions

SahanUday:

### DIFF
--- a/docs/docs/internals/jac_import_patterns.md
+++ b/docs/docs/internals/jac_import_patterns.md
@@ -246,6 +246,7 @@ custom_extensions = [".js", ".py"]
 ```
 
 **Project Structure:**
+
 ```
 my-jac-project/
 ├── jac.toml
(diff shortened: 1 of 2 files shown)


docs/docs/internals/jac_import_patterns.md
```diff
@@ -205,6 +205,77 @@ function MyComponent() {
 }
 ```
 
+## Custom File Configuration Pattern
+
+Beyond standard module imports, Jac Client supports custom asset file handling through `jac.toml` configuration. This allows you to include specialized file types (`.js`, `.py`, `.wasm`, `.json`, etc.) as static assets in your build.
```
MusabMahmoodh, line 210: js and json should be supported by default right?

SahanUday: No
image

only these are support as default assets type

MusabMahmoodh: so the doc is mis leading, this is for custom asset files, not files

PR #4054, opened 2026-01-13
implement max iterations for ReAct loop in LLM invocation

Hushan-10:
## **Issue**
The max_react_iterations parameter does not reliably control the number of ReAct iterations.
Setting it to a low value (e.g. 1) does not consistently limit execution as expected.

## Solution (Implementation)

Previously, `max_react_iterations` was not enforced, which could cause ReAct tool loops to run indefinitely if the model never triggered `finish_tool`.

This PR adds iteration tracking inside `BaseLLM.invoke()`:

- Resolves `max_react_iterations` by priority:  
- Tracks the ReAct loop iteration count
- When the limit is reached:
  - **Temporarily disables tools**
  - **Appends a final-answer instruction**
  - Executes one final model call (or uses `_stream_final_answer()` for streaming)

This safely exits tool-mode and forces a final response, ensuring the workflow always terminates.

- Added a test case and release note


jac-byllm/byllm/llm.impl/basellm.impl.jac
```diff
@@ -17,8 +17,39 @@ impl BaseLLM.invoke(mtir: MTRuntime) -> object {
...
+            if mtir.stream {
+                return self._stream_final_answer(mtir);
+            }
+            # Non-streaming: disable tools and ask for final answer once
+            final_instruction = Message(
+                role=MessageRole.USER,
+                content="Based on the tool calls and their results above, provide only your final answer. "
+                "Be comprehensive and synthesize all the information gathered."
+                "IMPORTANT:\n"
+                "- Output only the final answer.",
```
MalithaPrabhashana, line 41: will it work without this part?

Hushan-10: > will it work without this part?

Yes, it worked

PR #4061, opened 2026-01-13
Caching automation using react_query

SahanUday:
## **Description**
Integrate React Query into jacSpawn


jac-client/jac_client/plugin/impl/client_runtime.impl.jac
```diff
@@ -70,8 +70,14 @@ impl __jacSpawn(left: str, right: str = "", fields: dict = {}) -> any {
     return await response.json();
 }
 
+cl import from "@tanstack/react-query" { QueryClient, QueryClientProvider, useQuery }
+glob queryClient = Reflect.construct(QueryClient, []);
 impl jacSpawn(left: str, right: str = "", fields: dict = {}) -> any {
-    return __jacSpawn(left, right, fields);
+    queryResult = useQuery({
+        queryKey: ["greet"],
```
MusabMahmoodh: these should be configurable? other wise no use of having keys?

MusabMahmoodh (changes requested):

jac-client/jac_client/plugin/impl/client_runtime.impl.jac
SahanUday: should be configurable

PR #4069, opened 2026-01-14
Fix known file extension importing issue

SahanUday:
## **Description**
Enhance the module resolver to correctly process filenames with common extensions.

- Enhanced the `resolve_module` function in `modresolver.py` to recognize and correctly merge common file extensions (like `.css`, `.js`, `.json`, etc.) when splitting module paths, preventing issues where extensions might be treated as separate path segments.

### DIFF
--- a/jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
+++ b/jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
@@ -299,7 +299,7 @@ class KubernetesTarget(DeploymentTarget) {
             'command': [
                 'sh',
                 '-c',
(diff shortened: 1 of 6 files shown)


jac-scale/jac_scale/tests/test_examples.py
```diff
@@ -410,7 +410,9 @@ def test_all_in_one(self) -> None:
         with JacScaleTestRunner(
             example_file, session_name="custom_test", setup_npm=True
         ) as runner:
-            assert "background-image" in runner.request_raw("GET", "/styles.css")
+            assert "background-image" in runner.request_raw(
```
MusabMahmoodh: this is wrong?

SahanUday: yes, i updated

MalithaPrabhashana:
This looks good to me!

Need to resolve the conflicts @SahanUday

PR #4074, opened 2026-01-14
[WIP] Add streaming support for function execution and walker spawning

mgtm98:
## **Description**
support depends on the fact when a function yields, it returns a generator and the results of the generator is used when we iterate over it
```
def simple_stream() -> Generator {
    for i in range(5) {
        time.sleep(1);
        print(f"Generating item {i}");
        yield f"Item {i}";
    }
}
```
for this example `simple_stream` returns a generator, so what i do is 

1. Check if the function is a generator using `inspect.isgenerator`
2. if yes then yield the items one by one
3. if not then return the values all at once

depending on this I created a new method in `ExecutionManager` called `execute_function_stream` that checks if the return is generator then yield the results one by one, if not then yield the output. 
```
    if inspect.isgenerator(result) {
            for value in result {
                yield JacSerializer.serialize(value);
            }
            Jac.commit();
        } else {
            yield JacSerializer.serialize(result);
            if ctx.reports {
                for report in ctx.reports {
                    yield JacSerializer.serialize(report);
                }
            }
            Jac.commit();
        }
```
in case of `execute_function` if the output is a generator then we get all results all at once and return them


mgtm98:
I need to review this with @marsninja, @musab-mah-7, @Thamirawaran and @kugesan1105 
if you all agree then I will implement this into walkers as well and test the support.

Thamirawaran:
1. How developers can enable the streaming (even for the functions)

mgtm98:
> > @Thamirawaran, by adding "stream=true" as a query parameter
> 
> There is no api call in jac

can you explain more what is the issue?

Thamirawaran:
> > > @Thamirawaran, by adding "stream=true" as a query parameter
> > 
> > 
> > There is no api call in jac
> 
> can you explain more what is the issue?

in jac, we don't do api calls manually

jac/jaclang/pycore/passes/pyast_gen_pass.py
```diff
@@ -243,6 +243,28 @@ def _function_expr_from_def(
...
+        # Get the body of the walker
+        body_nodes: list[uni.UniNode] = []
+        if isinstance(node.body, uni.ImplDef) and isinstance(node.body.body, list):
+            body_nodes = node.body.body  # type: ignore
+        elif isinstance(node.body, list):
+            body_nodes = node.body
+
+        # Check all statements in the body
+        return any(_check_node(stmt) for stmt in body_nodes)
+
```
Thamirawaran, line 267: why we need to make changes in pyastgen_pass

mgtm98: We need to add a marker to indicate that a walker uses streaming. This can be done by adding a decorator to the walker, adding a decorator to the on-entry function, or change function name. The clean solution for me adding a decorator to the walker class

PR #4116, opened 2026-01-16
implement max iterations for ReAct loop in LLM invocation

Hushan-10:
## **Issue**
The max_react_iterations parameter does not reliably control the number of ReAct iterations.
Setting it to a low value (e.g. 1) does not consistently limit execution as expected.

## Solution (Implementation)

Previously, `max_react_iterations` was not enforced, which could cause ReAct tool loops to run indefinitely if the model never triggered `finish_tool`.

This PR adds iteration tracking inside `BaseLLM.invoke()`:

- Resolves `max_react_iterations` by priority:  
- Tracks the ReAct loop iteration count
- When the limit is reached:
  - **Temporarily disables tools**
  - **Appends a final-answer instruction**
  - Executes one final model call (or uses `_stream_final_answer()` for streaming)

This safely exits tool-mode and forces a final response, ensuring the workflow always terminates.

- Added a test case and release note


MalithaPrabhashana (approved):
LGTM!

PR #4170, opened 2026-01-19
File upload support to jac scale

kashmithnisakya:
To create a walker that accepts file uploads, import `UploadFile` from FastAPI and use it as a field type:

```jac
import from fastapi { UploadFile }

walker ProcessDocument {
    has document: UploadFile;
    has description: str = "";

    can process with `root entry {
        print(f"Received: {self.document.filename}");
        print(f"Type: {self.document.content_type}");
        print(f"Size: {self.document.size} bytes");
        
        report {
            "filename": self.document.filename,
            "content_type": self.document.content_type,
            "size": self.document.size,
            "description": self.description
        };
    }
}
```

**API Request (using curl):**

```bash
curl -X POST "http://localhost:8000/walker/ProcessDocument" \
  -H "Authorization: Bearer YOUR_TOKEN" \
  -F "document=@/path/to/file.pdf" \
  -F "description=My document"
```

### DIFF
--- a/jac-scale/jac_scale/jserver/impl/jfast_api.impl.jac
+++ b/jac-scale/jac_scale/jserver/impl/jfast_api.impl.jac
@@ -141,9 +141,7 @@ impl QueryParameterHandler.generate_param_string(param: APIParameter) -> str {
     } elif (param.default is None) {
         return f"{param.name}: Optional[{type_name}] = Query(description='{description}')";
     } else {
-        return f"{param.name}: Optional[{type_name}] = Query({repr(
-            param.default
-        )}, description='{description}')";
+        return f"{param.name}: Optional[{type_name}] = Query({repr(param.default)}, description='{description}')";
     }
 }
 
@@ -156,9 +154,7 @@ impl HeaderParameterHandler.generate_param_string(param: APIParameter) -> str {
     } elif (param.default is None) {
         return f"{param.name}: Optional[{type_name}] = Header(default=None, description='{description}')";
     } else {
-        return f"{param.name}: Optional[{type_name}] = Header({repr(
-            param.default
-        )}, description='{description}')";
+        return f"{param.name}: Optional[{type_name}] = Header({repr(param.default)}, description='{description}')";
     }
 }
 
@@ -328,7 +324,6 @@ impl JFastApiServer._create_endpoint_function(
             file_params.append(param);
         }
     }
-
     # Initialize parameter strings and mapping
     param_strs: list[str] = [];
     param_mapping: dict[(str, str)] = {};
@@ -340,7 +335,6 @@ impl JFastApiServer._create_endpoint_function(
         param_strs.append('request: Request');
         param_mapping['request'] = 'request';
     }
-
     # Create handler map
     has_file_params = len(file_params) > 0;
     handler_map: dict[(ParameterType, ParameterHandler)] = {
@@ -349,7 +343,6 @@ impl JFastApiServer._create_endpoint_function(
         ParameterType.HEADER: HeaderParameterHandler(type_converter=self),
         ParameterType.FILE: FileParameterHandler(type_converter=self)
     };
-
     # Handle body parameters using appropriate handler
     body_model: (type[BaseModel] | None) = None;
     if (len(body_params) >= 1) {
@@ -375,7 +368,6 @@ impl JFastApiServer._create_endpoint_function(
             }
         }
     }
-
     # Process other parameter types using their handlers
     param_groups = [
         (path_params, ParameterType.PATH),


docs/docs/jac-scale/README.md
```diff
@@ -27,6 +27,13 @@
 
 ### 5. Support for public private endpoints with JWT authentication
 
+### 6. File Upload Support
+
+- **Automatic Form-Data Handling**: Walkers with `UploadFile` fields automatically accept multipart/form-data
+- **Single and Multiple Files**: Support for single file
```
MusabMahmoodh: what is meant here?

docs/docs/jac-scale/file-upload.md
```diff
@@ -0,0 +1,120 @@
...
+**API Request (using curl):**
+
+```bash
+curl -X POST "http://localhost:8000/walker/ProcessDocument" \
+  -H "Authorization: Bearer YOUR_TOKEN" \
+  -F "document=@/path/to/file.pdf" \
+  -F "description=My document"
+```
+
+**API Request (using Python requests):**
```
MusabMahmoodh: we can remove this better if we can have and example how to upload this from Jac client instead

docs/docs/jac-scale/file-upload.md
```diff
@@ -0,0 +1,120 @@
...
+```bash
+curl -X POST "http://localhost:8000/walker/CreatePost" \
+  -H "Authorization: Bearer YOUR_TOKEN" \
+  -F "title=My Post" \
+  -F "content=This is the post content" \
+  -F "image=@photo.jpg" \
+  -F "tags=tech,news"
+```
+
+## Saving Uploaded Files
```
MusabMahmoodh: this we have to remove? instead we have to support and infrastrcutre from Jac scale(An  extendable filestorage which we can upload files)

jac-scale/jac_scale/jserver/impl/jfast_api.impl.jac
```diff
@@ -229,41 +232,71 @@ impl JFastApiServer._create_endpoint_function(
...
-            param_type = self._get_python_type(param.data_type);
-            required = param.required;
-            description = param.description;
-            if required {
-                model_fields[param_name] = (
-                    param_type,
-                    Field(..., description=description)
-                );
-            } else {
+        if has_file_params {
```
MusabMahmoodh, line 349: parameter handling logic is getting quite nested with the addition of file support. Have you considered refactoring this to use a handler/strategy pattern instead of nested if-else blocks? 

**Current approach:**
- Nested conditionals based on parameter type and file presence
- Duplicated logic for generating param strings

**Suggested improvement:**
Create dedicated handlers for each parameter scenario: 
```python
handler_map = {
    (ParameterType.BODY, has_files=True): FormParameterHandler,
    (ParameterType. BODY, has_files=False): JSONBodyHandler,
    ParameterType.FILE: FileParameterHandler,
    ParameterType.PATH: PathParameterHandler,
    # etc... 
}

File: jac-scale/jac_scale/jserver/jserver.jac [hunk 5]
```diff
@@ -19,7 +19,8 @@ class ParameterType(StrEnum) {
     has QUERY: str = 'query',
         PATH: str = 'path',
         BODY: str = 'body',
-        HEADER: str = 'header';
+        HEADER: str = 'header',
+        FILE: str = 'file';
```
MusabMahmoodh, line 23: is that type correct?

MusabMahmoodh (changes requested):

docs/docs/jac-scale/file-upload.md
kashmithnisakya: Removed

docs/docs/jac-scale/file-upload.md
kashmithnisakya: Removed

docs/docs/jac-scale/README.md
kashmithnisakya: This was a bit misleading. I’ll update it.

At the moment, we can upload only one file per key.
If we need to upload multiple files, they must be sent using multiple keys, like this:

```
curl --location 'http://0.0.0.0:8000/walker/upload_document_api' \
  --header 'Authorization: Bearer <TOKEN>' \
  --form 'myfile1=@"/Users/kashmithnisakya/Downloads/02.png"' \
  --form 'myfile2=@"/Users/kashmithnisakya/Downloads/04.png"' \
  --form 'description="Gym logo"'
```

jac-scale/jac_scale/jserver/jserver.jac
kashmithnisakya: What do you mean by type? It’s an enum, so we need to get the string value here

PR #4188, opened 2026-01-20
fixed failure of creating svc if not exist

juzailmlwork:
## **Description**
Currently svc is getting created everytime the code is deployed to codebase while deleting the prvious svc and deployment.the new logic is changed to only delete the deployment and keep the svc


udithishanka (approved):
LGTM

PR #4254, opened 2026-01-22
added fix for early exit if pod   is restarted

juzailmlwork:
## **Description**
previously deployment status  was checked by check_deployment_status function which used to connect /docs multiple times till limit is reached.But this get stuck for 7.5 mins before mentioning error in deployment but now we can detect failure of deployment by identifying restarting of containers and throw error message (within 1.5 min) 

this will fasten up the development and more robust in pipeline

things covered in the pr

- [x] auto detection of pod restarting and stopping k8s deployment rather than waiting till deployment
- [x] check_deployment_status now returns one of ["Sucessful","Containers restarted","Timeout reached"]
- [x] removal of dplicate print statments regarding k8s deployment to make the cli clear 
- [x] k8s deploy function now return deployment_details which has information regarding sucess of each sub stage of deployment
- [x] test case improvement to check deployment_details and fix for deployment sucessful flag

### DIFF
--- a/jac-scale/jac_scale/tests/test_deploy_k8s.py
+++ b/jac-scale/jac_scale/tests/test_deploy_k8s.py
@@ -279,8 +279,7 @@ def test_early_exit():
     print(f"Deployment result: {details}")
     assert "health_check_of_deployment" not in details
     assert len(details) == 7
-    assert result.success is True
-    print(f"✓ Deployment successful: {result.message}")
+    assert result.success is False
 
 
 def test_deployment_target_methods():


jac-scale/jac_scale/targets/kubernetes/utils/kubernetes_utils.impl.jac
```diff
@@ -346,6 +346,35 @@ impl cluster_type -> str {
     }
 }
 
+impl pods_restarted(namespace: str, app_name: str) -> bool {
```
musab-mah-7: vague naming

juzailmlwork: changed to check_pods_restarted

jac-scale/jac_scale/targets/kubernetes/utils/kubernetes_utils.impl.jac
```diff
@@ -347,6 +347,36 @@ impl cluster_type -> str {
...
+                console.print(
+                    f"Pod {pod.metadata.name} container "
+                    f"{container.name} restarted "
+                    f"{container.restart_count} times"
+                );
+                return True;
+            }
+        }
+        return False;
+    }
```
udithishanka, line 389: If multiple pods exist, only the first pod with container_statuses is checked. The function returns False after checking the first pod, even if subsequent pods have restarted containers ?
return False; should be outside the for loop ? 

If pods.items is empty, the function implicitly returns None (no explicit return statement), which could cause unexpected behavior when evaluated.

docs/docs/community/release_notes/jac-scale.md
```diff
@@ -86,12 +88,13 @@ First release of **Jac-Scale** - a scalable runtime framework for distributed Ja
...
-- Performance monitoring and auto-scaling
-- YAML-based configuration
-- Username-based user management for authentication
+- Conversion of walker to fastapi endpoints
+- Multi memory hierachy implementation
+- Support for Mongodb (persistance storage) and Redis (cache storage) in k8s
+- Deployment of app code directly to k8s cluster
+- k8s support for local deployment and aws k8s deployment
+- SSO support for google
+-
```
udithishanka: stray hypen

juzailmlwork: stray hypen is removed

jac-scale/jac_scale/targets/kubernetes/utils/kubernetes_utils.impl.jac
juzailmlwork: nice catch the reason this was previously missed was due to there is only one jaseci pod.also now the return is bought out of the for loop

PR #4384, opened 2026-01-30
support horizontal scaling in k8s for jasaeci-pods

juzailmlwork:
## **Description**
added support for horizontal scaling HPA.need to integrate promethius metric server in upcoming roadmap.Current hpa is activated using average cpu usage

### DIFF
--- a/jac-scale/jac_scale/tests/test_deploy_k8s.py
+++ b/jac-scale/jac_scale/tests/test_deploy_k8s.py
@@ -320,6 +320,21 @@ def test_deploy_all_in_one():
     )
     assert redis_service.spec.ports[0].port == 6379
 
+    # Validate code-server Deployment and Service
+    code_server_deploy = apps_v1.read_namespaced_deployment(
+        name=f"{app_name}-code-server", namespace=namespace
+    )
+    assert code_server_deploy.metadata.name == f"{app_name}-code-server"
+    assert code_server_deploy.spec.replicas == 1
+    print(f"✓ Code-server deployment verified: {app_name}-code-server")
+
+    code_server_service = core_v1.read_namespaced_service(
+        name=f"{app_name}-code-server", namespace=namespace
+    )
+    assert code_server_service.spec.type == "ClusterIP"
+    assert code_server_service.spec.ports[0].port == 8080
+    print("✓ Code-server service verified: ClusterIP on port 8080")
+
     # Validate K8s Secret was created with correct data
     secret = core_v1.read_namespaced_secret(
         name=f"{app_name}-secrets", namespace=namespace
@@ -425,6 +440,21 @@ def test_deploy_all_in_one():
             f"{app_name}-redis-service", namespace=namespace
         )
         raise AssertionError("Redis Service should have been deleted")
+    except ApiException as e:
+        assert e.status == 404, f"Expected 404, got {e.status}"
+
+    # Verify code-server cleanup
+    try:
+        apps_v1.read_namespaced_deployment(
+            f"{app_name}-code-server", namespace=namespace
+        )
+        raise AssertionError("Code-server Deployment should have been deleted")
+    except ApiException as e:
+        assert e.status == 404, f"Expected 404, got {e.status}"
+
+    try:
+        core_v1.read_namespaced_service(f"{app_name}-code-server", namespace=namespace)
+        raise AssertionError("Code-server Service should have been deleted")
     except ApiException as e:
         assert e.status == 404, f"Expected 404, got {e.status}"
 


jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
```diff
@@ -650,6 +651,15 @@ class KubernetesTarget(DeploymentTarget) {
...
         service_url = self.get_service_url(app_name);
 
+        # 16. Enable horizontal pod autoscaling
+        create_hpa(
+            namespace=namespace,
+            deployment_name=app_name,
+            min_replicas=1,
+            max_replicas=3,
+            cpu_target=50
+        );
```
udithishanka: `min_replicas=1,   // Hardcoded
    max_replicas=3,   // Hardcoded (also inconsistent with function default of 5)
    cpu_target=50     // Hardcoded
`

can we add to the KubernetesConfig?

Also is HPA always enabled, can the users opt out from it ?

jac-scale/jac_scale/targets/kubernetes/hpa/impl/hpa.impl.jac
```diff
@@ -0,0 +1,46 @@
...
+    deployment_name: str,
+    min_replicas: int = 1,
+    max_replicas: int = 5,
+    cpu_target: int = 50
+) -> None {
+    autoscaling_v2 = AutoscalingV2Api();
+    hpa = V2HorizontalPodAutoscaler(
+        api_version='autoscaling/v2',
+        kind='HorizontalPodAutoscaler',
+        metadata={'name': f"{deployment_name}-hpa"},
```
udithishanka, line 16: should we add namespace as well ?

metadata={'name': f"{deployment_name}-hpa", 'namespace': namespace},

jac-scale/jac_scale/targets/kubernetes/hpa/impl/hpa.impl.jac
```diff
@@ -0,0 +1,46 @@
...
+        )
+    );
+    try {
+        autoscaling_v2.create_namespaced_horizontal_pod_autoscaler(
+            namespace=namespace, body=hpa
+        );
+    } except ApiException as e {
+        if (e.status != 409) {
+            raise ;
+        }
```
udithishanka, line 44: Silent 409 Handling
Silently ignores existing HPA - should log or update

docs/docs/community/release_notes/jac-scale.md
```diff
@@ -4,6 +4,8 @@ This document provides a summary of new features, improvements, and bug fixes in
 
 ## jac-scale 0.1.6 (Unreleased)
 
+-**support horizontal scaling**:  based on average cpu usage k8s pods are horizontally scaled
```
udithishanka: missing space after hypen (-) . this will not render properly as a bullet point otherwise

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
```diff
@@ -650,6 +651,15 @@ class KubernetesTarget(DeploymentTarget) {
         self._wait_for_deployment(app_name, namespace, health_check_path, is_aws);
         service_url = self.get_service_url(app_name);
 
+        # 16. Enable horizontal pod autoscaling
+        create_hpa(
```
udithishanka: should we have any logging for hpa creation ?

docs/docs/community/release_notes/jac-scale.md
juzailmlwork: fixed

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
juzailmlwork: Not needed udith since it is created only first time not for subsequent requests.currently log statements are added for
1. sync pod
2. create deployment
3. url of deployment

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
juzailmlwork: these are moved to kuberneres config now @udithishanka 

currently HPA is always enabled.since mimimum pod is set to 1 .user dont need to worry about opting it and also it only cost 25 mb or ram for hpa

jac-scale/jac_scale/targets/kubernetes/hpa/impl/hpa.impl.jac
juzailmlwork: @udithishanka  currently they are defined at autoscaling_v2.create_namespaced_horizontal_pod_autoscaler(
            namespace=namespace, body=hpa
        );
In a future pr need to add a special metadata called "jac-scale-managed" so we can clearly identify what are the resources created by jac-scale.

jac-scale/jac_scale/targets/kubernetes/hpa/impl/hpa.impl.jac
juzailmlwork: since we are creating only one time the hpa per deployment.expected behaviour is hpa should not be created again.Also since hpa isa  internal step skipped logging statement

udithishanka:
# HPA + RWO PVC: Multi-Node Scaling is Broken

## Summary

PR #4384 adds HPA (Horizontal Pod Autoscaling) support, but **it doesn't work on multi-node clusters**. When HPA scales beyond 1 replica, new pods scheduled on different nodes hit `Multi-Attach error` because all pods share the same RWO (ReadWriteOnce) PVC for application code.

## Live Test (AWS EKS, 8-node cluster)

**Config:**
```toml
[plugins.scale.kubernetes]
min_replicas = 2
max_replicas = 4
cpu_utilization_target = 50
```

**Result:**
```
$ kubectl get pods -n algo-test2 -o wide
NAME                                READY   STATUS     NODE
algo-test2-7cf8b89f88-hjt5g         1/1     Running    ip-192-168-20-98     # PVC attached here
algo-test2-7cf8b89f88-9q29l         0/1     Init:0/3   ip-192-168-102-222   # STUCK

$ kubectl describe pod algo-test2-7cf8b89f88-9q29l
Events:
  Warning  FailedAttachVolume  attachdetach-controller
  Multi-Attach error for volume "pvc-962831ba-..."
  Volume is already used by pod(s) algo-test2-7cf8b89f88-hjt5g
```

HPA requested 2 replicas. Pod 1 runs fine. Pod 2 was scheduled on a different node and is permanently stuck — the RWO volume can only attach to one node at a time.

## Root Cause

All app pods mount `{app_name}-code-pvc` (created with `ReadWriteOnce` access mode) at `/code`. RWO allows multiple pods on the **same node** to share the volume, but **not across nodes**. On multi-node clusters, K8s schedules pods freely across nodes, causing Multi-Attach errors.

This works on single-node dev clusters (minikube/kind) because all pods are on the same node — which is why it wasn't caught.

## The Path Forward

For HPA to work across nodes, the code distribution needs to change. The approach should be **cloud-agnostic** (no hardcoded AWS/GCP/Azure storage classes).

| Approach | Change | Pros | Cons |
|----------|--------|------|------|
| **RWX volume** (EFS/NFS/Filestore) | Change PVC `accessMode` to `ReadWriteMany` | Minimal code change, drop nodeAffinity, true multi-node | User must configure cloud-specific storage class |
| **Bake code into image** (`jac build` → Docker) | No PVC at all | Most production-ready, pods truly independent, immutable deploys | Needs container registry, slower redeploys |
| **Init container download** (S3/GCS/ConfigMap) | Each pod downloads code on startup | No shared storage, pods independent | Startup latency, needs object store |

## Suggested Implementation

Add user-configurable PVC access mode and storage class to `jac.toml`:

```toml
[plugins.scale.kubernetes]
# Default (single replica) — RWO works fine
pvc_access_mode = "ReadWriteOnce"

# For horizontal scaling — user provides their RWX storage class
pvc_access_mode = "ReadWriteMany"
storage_class = "efs-sc"  # AWS: efs-sc, GCP: standard-rwx, Azure: azurefile, Local: nfs
```

**Behavior:**
- `ReadWriteOnce` (default): Current behavior. If `max_replicas > 1`, pin all pods to the PVC node via nodeAffinity (works but limited to one node's capacity).
- `ReadWriteMany`: Create PVC with RWX access mode using the specified storage class. No nodeAffinity needed — pods can scale across nodes freely.
- **Validation**: If `max_replicas > 1` and `pvc_access_mode = "ReadWriteOnce"`, warn that scaling is limited to a single node.

This keeps jac-scale cloud-agnostic — the user configures their own storage class, nothing is hardcoded.

## Impact on Related PRs

- **PR #4546** (zero-downtime rolling deployment): Uses nodeAffinity to pin pods during rolling updates. Compatible with this approach — nodeAffinity is only needed for RWO.
- **PR #4571** (code-sync node-pin fix): Pins the code-sync pod to the PVC's node. Still needed for RWO deployments; not needed for RWX.

## Impact Matrix

| Scenario | RWO (current) | RWO + nodeAffinity | RWX (user-configured) |
|----------|--------------|-------------------|----------------------|
| Single replica | Works | Works | Works |
| HPA multi-replica | **Multi-Attach crash** | All on 1 node | True multi-node |
| Rolling update | **Multi-Attach crash** | Zero-downtime (1 node) | Zero-downtime (multi-node) |
| Node failure | Pod lost | All replicas lost | Partial survival |
| Cloud-agnostic | Yes | Yes | Yes (user provides storage class) |

udithishanka (changes requested):
Found a breaking point of HPA. need to discuss, on how we are moving forward

PR #4397, opened 2026-01-31
Fix crash on dotted directory imports

SandeepaHWP:
fixes an `IsADirectoryError` when compiling dotted imports (e.g. `include impl.imps`) where segments are directories.

Added a check to skip directory paths during import resolution, ensuring the compiler only attempts to read actual files.

**After :**
```
(.venv) hwpsa@DESKTOP-4NIJJH3:~/Learn/jaseci$ jac check '/home/hwpsa/Learn/jaseci/jac/tests/compiler/passes/main/fixtures/base.jac'
✔ Checked 1 '.jac' files: 1 passed, 0 with errors (0 errors, 0 warnings).
```


SandeepaHWP:
fixed with #4497

PR #4446, opened 2026-02-03
fix(cli): skip .impl.jac files in jac check command

ThakeeNathees:
## Summary
.impl.jac files cannot be checked in isolation as they require all supporting impl files and the main file to be present. 

## Changes
- Updates \jac check\ to skip files ending with \.impl.jac\
- Removes \.impl.jac\ entries from \.jacignore\ since they are now properly handled by the check command itself


jac/jaclang/cli/commands/impl/analysis.impl.jac
```diff
@@ -46,15 +46,18 @@ impl check(
...
             def filter_issues(issues: list) -> list {
                 return [
                     i
                     for i in issues
                     if not (
                         i.loc
                         and i.loc.mod_path
-                        and any(p in i.loc.mod_path for p in ignore_list)
+                        and (
+                            i.loc.mod_path.endswith('.impl.jac')
```
kugesan1105: @ThakeeNathees  what if typecheck error is inside  `impl body`, we can't ignore here

ThakeeNathees: That will be handled when we type check the main (not impl file) but a standalone impl file cannot be type checked that's why we ignore it here.

PR #4452, opened 2026-02-04
support returning None from a function that is expected to return int

mgtm98:
## **Description**
Fix an issue when returning a None in case of a function returns an int
```
def a(x: int) -> int | None {
    if x > 0 {
        return x;
    } else {
        return None;
    }
}
```
without the fix jac native generates the following error
```
✖ Error:   /home/mgtm/forks/jaseci/t.na.jac, line 125, col 1: Native compilation failed: LLVM IR parsing error
<string>:17:7: error: value doesn't match function result type 'i64'
  ret i8* null
```


marsninja (changes requested):
It would also be good to add more description of what is going wrong without this fix in place.  also need to add a bullet to release_notes/jaclang.md and the addition to the test ficture should be validated in the jac/tests/compiler/passes/native/test_native_gen_pass.py .

PR #4466, opened 2026-02-05
feat: Update script tags to use type="module" for ESM compatibility across client and server HTML outputs

MusabMahmoodh:
Added `type="module"` to generated `<script>` tags in the client HTML output. The Vite bundler already produces ES module output, but the script tags were missing the module attribute, causing browsers to reject ESM syntax (e.g., `import`/`export`) from newer npm packages. Affects both the server-rendered page and the `jac build --target web` static output.


marsninja:
Lets put a test in place on this one @musab-mah-7 .

PR #4469, opened 2026-02-05
Fix asset loading on desktop

ahzan-dev:
### Problem

Static assets from the `public/` folder weren't loading in desktop target for both dev and production modes:

1. **Dev mode**: Assets like images, fonts, and other static files from `public/` weren't served because `publicDir` was hardcoded to `false`

2. **Production mode**: 
   - Same `publicDir: false` issue prevented assets from being copied to dist
   - Asset URLs used absolute paths (e.g., `/client.abc123.js`) which don't work with Tauri's `file://` protocol, causing white screen

### Solution

Updated `vite_bundler.impl.jac` with the following fixes:

**Dev config (`create_dev_vite_config`):**
- Added dynamic `publicDir` detection - checks if `public/` folder exists and uses it
- Added Vite plugin to proxy root-level asset requests to backend server
- Added `/assets` proxy route

**Build config (`create_vite_config`):**
- Added dynamic `publicDir` detection (was hardcoded to `false`)
- Added `base: "./"` to use relative paths for Tauri file:// protocol compatibility

### Testing

- `jac start main.jac --client desktop --dev` - Assets load correctly on localhost:8000, localhost:5173, and Tauri dev window
- `jac start main.jac --client desktop` - Assets load correctly in production Tauri app

image


jac-client/jac_client/plugin/src/impl/vite_bundler.impl.jac
```diff
@@ -639,9 +654,36 @@ export default defineConfig({{
...
+              const proxyReq = http.get(targetUrl, (proxyRes) => {{
+                res.writeHead(proxyRes.statusCode || 200, proxyRes.headers);
+                proxyRes.pipe(res);
+              }});
+              proxyReq.on('error', () => {{
+                next(); // Fall through to Vite's 404
+              }});
+            }});
+          }} else {{
+            next();
```
ahzan-dev: GitHub Copilot: ## Explanation of the `proxy-root-assets` Plugin

This is a custom Vite middleware plugin that handles static asset requests during **dev mode**. Here's what it does:

### The Problem It Solves

In the Jac desktop dev setup, you have **3 servers running**:
1. **Backend server** (localhost:8000) - Serves the API and static files from `public/`, `assets/`
2. **Vite dev server** (localhost:5173) - Handles JS/TS compilation and HMR
3. **Tauri webview** - Loads from Vite dev server

When your app tries to load an image like `/logo.png`, the request goes to Vite (port 5173), but Vite doesn't know about files in your project's `public/` folder that the backend serves.

### How It Works

```
                    Request: /logo.png
                           │
                           ▼
                  ┌─────────────────┐
                  │   Vite Server   │
                  │  (port 5173)    │
                  └────────┬────────┘
                           │
        ┌──────────────────┼──────────────────┐
        │                  │                  │
        ▼                  ▼                  ▼
   Is it .png,        Is it /static/*    Is it /@vite/*
   .jpg, .svg,        or /api/*?         or node_modules?
   .ico, etc?              │                  │
        │                  │                  │
        ▼                  ▼                  ▼
   YES → Proxy to     Already has        Let Vite handle
   backend:8000       proxy rules        (internal stuff)
```

### Line-by-Line Breakdown

```javascript
server.middlewares.use((req, res, next) => {
```
Adds middleware to intercept every request to Vite dev server.

```javascript
const url = req.url || '';
if (assetExtensions.test(url) && !url.startsWith('/static/') && !url.startsWith('/@')) {
```
Checks if the URL:
- Matches asset extensions (`.png`, `.jpg`, `.svg`, `.ico`, `.woff2`, etc.)
- Is NOT already a `/static/` path (those have their own proxy)
- Is NOT a Vite internal path (`/@vite/`, `/@react-refresh/`)

```javascript
const targetUrl = `http://localhost:{api_port}${url}`;
http.get(targetUrl, (proxyRes) => {
  res.writeHead(proxyRes.statusCode || 200, proxyRes.headers);
  proxyRes.pipe(res);
});
```
Forwards the request to the backend server and streams the response back.

```javascript
proxyReq.on('error', () => {
  next(); // Fall through to Vite's 404
});
```
If the backend doesn't have the file, let Vite handle it (which will 404).

### Example Flow

```
1. Browser requests: /logo.png
2. Vite receives request
3. Plugin checks: Is ".png" an asset extension? YES
4. Plugin checks: Starts with "/static/"? NO
5. Plugin checks: Starts with "/@"? NO
6. Plugin proxies to: http://localhost:8000/logo.png
7. Backend returns the image from public/ folder
8. Plugin streams it back to browser
```

This ensures assets work seamlessly across all three targets (backend, Vite, Tauri) during development.

musab-mah-7:
Architecture Note: Scope defer to desktop target only

The type="module" → defer change is correct for desktop/Tauri (CORS blocks type="module" on file://), but it's currently applied to all targets including web and server-rendered pages where type="module" works fine and is the modern standard.

Since DesktopTarget.build() already delegates to web_target.build() → _generate_index_html(), the clean fix is to pass the script loading strategy as a parameter rather than hardcoding defer globally:


_generate_index_html(..., script_loading="module")  # default for web
_generate_index_html(..., script_loading="defer")   # desktop passes this
Same applies to render_page in client.impl.jac and runtime.py — those are HTTP-served and should stay type="module".

Why it matters: defer treats the bundle as a classic script. Today this works because the bundle is a single file with no top-level ESM syntax. But if Vite ever code-splits or a dynamic import() is added, web builds will break silently. Keeping type="module" for web targets is both correct and future-proof.

Everything else in this PR (publicDir detection, base: "./", asset proxy middleware, /assets route) looks good and follows existing patterns.

PR #4519, opened 2026-02-09
Enhance jac-scale documentation for webhooks and WebSocket support

kashmithnisakya:
This PR adds WebSocket documentation to the MkDocs

### DIFF
--- a/docs/docs/reference/plugins/jac-scale.md
+++ b/docs/docs/reference/plugins/jac-scale.md
@@ -524,18 +524,7 @@ This walker will be accessible at `ws://localhost:8000/ws/EchoMessage`.
 
 #### Authenticated WebSocket Walker
 
-Omit `: pub` to require JWT authentication:
-
-```jac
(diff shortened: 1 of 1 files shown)


docs/docs/reference/plugins/jac-scale.md
```diff
@@ -447,15 +460,140 @@ curl -X POST "http://localhost:8000/webhook/PaymentReceived" \
...
+        report {
+            "room": self.room,
+            "content": self.message,
+            "authenticated": True
+        };
+    }
+}
+```
+
+Only authenticated users can connect and send messages, and all authenticated users receive broadcasts.
```
kugesan1105: @kashmithnisakya 
There are 4 separate code blocks that are nearly identical, differing only in:

`: pub` vs no `: pub`
`broadcast=True` vs default

Can be simplify to 1-2 examples

kashmithnisakya: removed

PR #4523, opened 2026-02-09
Invalidate stale NodeAnchor persistence cache across Jac applications

SupulHeshan:
## **Description**
This PR fixes a runtime persistence issue where stale NodeAnchor references were reused across sequential Jac program executions, causing:
```
Error: NodeAnchor [UUID] is not a valid reference!
````

Previously, running multiple Jac applications with different node/walker archetypes in the same environment would corrupt the persistence layer. Once corrupted, all subsequent node-based Jac programs failed until the cache files were manually deleted.

### Root Cause:

NodeAnchor objects persisted in the cache (.jac/data/*.db) across runs.
When a new Jac application with different archetypes was executed, cached anchors referenced archetypes that no longer existed in the current runtime context.

This caused invalid reference resolution and left the runtime in a permanently broken state.

### Fix

This PR ensures that:
- Stale NodeAnchor references are not reused across incompatible runs.
- Persistence is properly invalidated or isolated when archetypes change.
- The runtime no longer enters a globally broken state due to stale cache.

Now, running multiple Jac applications sequentially works without requiring manual deletion of:
```
.jac/data/anchor_store.db
.jac/data/main.db
````
### Before
- Run vehicle_finder.jac → ✅ Works
- Run math_poem_agents.jac → ❌ Fails
- Run vehicle_finder.jac again → ❌ Still fails
- Delete cache manually → ✅ Works again

### After
- Run vehicle_finder.jac → ✅ Works
- Run math_poem_agents.jac → ✅ Works
- Run vehicle_finder.jac again → ✅ Works
- No manual cache cleanup required

### DIFF
--- a/jac/tests/runtimelib/test_db_naming.jac
+++ b/jac/tests/runtimelib/test_db_naming.jac
@@ -8,7 +8,6 @@ Verifies that:
 """
 
 import os;
-import shutil;
 import from pathlib { Path }
 import from jaclang { JacRuntime as Jac }
 import from jaclang.cli.commands { execution }

### CODE REVIEW

**Comment by SupulHeshan** [comment]
> This issue can be fixed by deriving session/store for per main entry file. that would be elegant.

If we do that, it will create database files for each files that we run right?
like 100 test files = 100 databases,
Every ad-hoc script creates its own database.
its not good right?
what do you think?

File: jac/jaclang/runtimelib/impl/utils.impl.jac [hunk 1]
```diff
@@ -236,3 +236,113 @@ impl sys_path_context(path: str) -> Iterator[None] {
...
+            main_module = sys.modules['__main__'];
+            if main_module?.__file__ and main_module.__file__ {
+                return Path(main_module.__file__).resolve();
+            }
+        }
+    } except Exception {
+    # If any error occurs (e.g., sys.argv[0] doesn't exist), return None
+    }
+    return None;
+}
```
Thamirawaran: `_get_entry_path()` uses `sys.argv[0]` / `__main__.__file__`  
we can have another better way to resolve this

jac/jaclang/runtimelib/impl/utils.impl.jac
```diff
@@ -236,3 +236,113 @@ impl sys_path_context(path: str) -> Iterator[None] {
...
+
+# Ad-hoc directory patterns (check parent directories only, not filename)
+glob ADHOC_DIRS = frozenset(
+         {'test','tests','fixture','fixtures','example','examples','tmp','temp','build'}
+     );
```
Thamirawaran: it is good way to avoid ad hoc files. but it would be nice if we can think about it again

jac/jaclang/runtimelib/impl/utils.impl.jac
SupulHeshan: Resolved using JacRuntime

jac/jaclang/runtimelib/impl/utils.impl.jac
```diff
@@ -236,3 +236,63 @@ impl sys_path_context(path: str) -> Iterator[None] {
...
+        toml_path = file_path.parent / "jac.toml";
+        if toml_path.exists() and toml_path.is_file() {
+            config = JacConfig.load(toml_path);
+            if config.project.name {
+                return config.project.name;
+            }
+        } else {
+            return file_path.stem
+                if file_path.stem not in ['jac', '__main__']
+                else None;
```
Thamirawaran: cfg = get_config(start_path=entry_path.parent, force_discover=True)

jac/jaclang/cli/commands/impl/execution.impl.jac
```diff
@@ -33,8 +33,11 @@ def _discover_config_from_file(filename: str) -> None {
     import from pathlib { Path }
     import from jaclang.project.config { JacConfig, get_config, set_config }
     import from jaclang.project.dependencies { add_venv_to_path }
+    import from jaclang.jac0core.runtime { JacRuntime as Jac }
 
     file_path = Path(filename).resolve();
+    # Store entry file path for runtime utilities (called before _proc_file)
+    Jac.entry_file_path = str(file_path);
```
Thamirawaran: consider already existing `_proc_file()`

jac/jaclang/runtimelib/impl/utils.impl.jac
SupulHeshan: yeah.. it is better to use get_config(). I made the changes

jac/jaclang/runtimelib/impl/utils.impl.jac
SupulHeshan: This is removed...

jac/jaclang/cli/commands/impl/execution.impl.jac
SupulHeshan: Attribute name changed. but the same place.
better to have same place

PR #4556, opened 2026-02-11
grafana-promethius-deployment

juzailmlwork:
## **Description**
grafana dashboard and promethius storage server deployment to track APM of jaseci application per namespace with inbuilt graphs

### DIFF
--- a/jac-scale/jac_scale/tests/test_sso.jac
+++ b/jac-scale/jac_scale/tests/test_sso.jac
@@ -141,7 +141,7 @@ class MockScaleConfig {
         return {"namespace": "default"};
     }
 
-    def get_metrics_config(self) -> dict {
+    def get_monitoring_config(self) -> dict {
         return {"enabled": False};
     }
 }


jac-client/jac_client/examples/all-in-one/jac.toml
```diff
@@ -58,3 +58,11 @@ background_color = "#ffffff"
 # short_name = "App"
 [plugins.scale.secrets]
 TEST_SECRET_KEY = "${TEST_SECRET_KEY}"
+
+[plugins.scale.monitoring]
+enabled = true
+
+[plugins.scale.metrics]
+enabled = true
```
udithishanka: is it necessary to have another plugins.scale config ? can't we get the plugins.scale.metrics's enabled value and decide ?

juzailmlwork: merged both of them under plugins.scale.monitoring

PR #4565, opened 2026-02-11
Improve  HMR server-side reloading 

SahanUday:
## **Description**
This pull request introduces a refactor to the HMR in Jac, improving how `.impl.jac` annex files are handled and optimizing caching to avoid unnecessary recompilations. It also updates the test client and adds tests for HMR behavior with annex files. The changes enhance reliability and developer experience during code reloads.


jac/jaclang/runtimelib/impl/hmr.impl.jac
```diff
@@ -51,6 +51,7 @@ impl HotReloader.on_file_change(event: Any) -> None {
 
 """Process a file change by classifying declarations and taking action."""
 impl HotReloader._process_change(file_path: str) -> None {
+    import from jaclang.jac0core.bccache { discover_base_file }
```
musab-mah-7: If _process_change is called frequently, consider moving to module-level import to avoid repeated import overhead (though Python does cache imports).

PR #4573, opened 2026-02-12
Integration test for byllm

Hushan-10:
This pull request adds a comprehensive integration test suite for the `byllm` multi-agent Jac application, along with all necessary fixtures and mock implementations. The changes enable reliable testing of agent routing, streaming responses, and research analysis, using mock LLMs and tools. The most important changes are grouped below:

**Integration Test Suite Implementation**

* Added `test_integration_byllm.jac`, which implements end-to-end integration tests for the multi-agent Jac application. The tests verify server startup, user registration/login, walker endpoint accessibility, agent routing, and correctness of agent responses, including streaming and research summary outputs.

**Fixture and Mock Setup for Testing**

* Introduced `jac.toml` configuration for the test fixture, specifying project metadata, dependencies, and serving options required to run the Jac application in test mode.
* Added `main.jac` as the mock entry point for integration tests, defining agent nodes (`ConceptAgent`, `MathAgent`, `ResearchAgent`) and the `Supervisor` walker, using `MockLLM` instances and mock tool calls to simulate LLM behavior.
* Created `mock_responses.jac` to preload mock LLM outputs for each agent scenario, ensuring deterministic responses for concept explanation, math solving (with tool calls), and research summary analysis.
* Provided `tools.jac` with mock implementations for mathematical calculation and Python code execution, used by agents during testing.

**Agent Prompt Definitions**

* Defined agent prompt templates in `main.impl.jac` for `ConceptAgent`, `MathAgent`, `CodeAgent`, and `ResearchAgent`, specifying clear instructions and constraints for each agent's task during testing.

image


jac-byllm/tests/test_integration_byllm.py
```diff
@@ -0,0 +1,250 @@
...
+                    assert len(response_data) > 0, "Response should not be empty"
+
+        except HTTPError as exc:
+            error_body = exc.read().decode("utf-8", errors="ignore") if exc.fp else ""
+            pytest.fail(f"ResearchAgent test failed: {exc.code}\n{error_body}")
+
+
+# Run the test if executed directly
+if __name__ == "__main__":
+    pytest.main([__file__, "-v", "-s"])
```
MalithaPrabhashana: Why we need this part? can't we run the test using `pytest test_integration_byllm.py` command?

jac-byllm/tests/test_integration_byllm.py
```diff
@@ -0,0 +1,250 @@
...
+                        "Response should not be empty"
+                    )
+                else:
+                    assert len(response_data) > 0, "Response should not be empty"
+
+        except HTTPError as exc:
+            error_body = exc.read().decode("utf-8", errors="ignore") if exc.fp else ""
+            pytest.fail(f"ConceptAgent test failed: {exc.code}\n{error_body}")
+
+        # Test Case 2: MathAgent query
```
MalithaPrabhashana: Here we have `TestByllmWalkerEndpoints` class, so shall we split the test cases as methods inside the class

```jac
class TestByllmWalkerEndpoints:

    def test_supervisor_routes_to_concept_agent(...)

    def test_supervisor_routes_to_math_agent(...)

    def test_supervisor_routes_to_research_agent(...)
```

jac-byllm/tests/test_integration_byllm.py
```diff
@@ -0,0 +1,250 @@
...
+                    )
+                    assert len(agent_report["response"]) > 0, (
+                        "Response should not be empty"
+                    )
+                else:
+                    assert len(response_data) > 0, "Response should not be empty"
+
+        except HTTPError as exc:
+            error_body = exc.read().decode("utf-8", errors="ignore") if exc.fp else ""
+            pytest.fail(f"ConceptAgent test failed: {exc.code}\n{error_body}")
```
MalithaPrabhashana: This walker endpoint calling logic repeats, so let's create a helper function and use that function to call the walker endpoints.

Example:
```jac
def post_json(port: int, path: str, payload: dict) -> dict:
    req = Request(
        f"http://127.0.0.1:{port}{path}",
        data=json.dumps(payload).encode(),
        headers={"Content-Type": "application/json"},
        method="POST",
    )
    with urlopen(req, timeout=90) as resp:
        assert resp.status == 200
        return json.loads(resp.read().decode())
```

```jac
response = post_json(server_port, "/walker/Supervisor", concept_payload)
```

This is clean and maintainable.

udithishanka (approved):

jac-byllm/tests/fixtures/integration_byllm/main.jac
```diff
@@ -0,0 +1,138 @@
...
+obj Person {
+    has full_name: str,
+        year_of_death: str,
+        personality: str;
+}
+
+node ConceptAgent {
+
+    def explain(topic: str) -> str by llms["llm1"](
+        stream=True
```
udithishanka: remove the stream parameter if we are not testing that.

udithishanka (changes requested):

jac-byllm/tests/fixtures/integration_byllm/main.jac
Hushan-10: fixed

jac-byllm/tests/fixtures/integration_byllm/main.jac
```diff
@@ -0,0 +1,193 @@
...
+
+        if not [root --> (?:ConceptAgent)] { root ++> ConceptAgent(); }
+        if not [root --> (?:MathAgent)] { root ++> MathAgent(); }
+        if not [root --> (?:ResearchAgent)] { root ++> ResearchAgent(); }
+        if not [root --> (?:CodeAgent)] { root ++> CodeAgent(); }
+
+        # Load mock outputs based on query content before routing.
+        # Use re to detect math expressions (digits near operators) to avoid
+        # false positives on hyphens inside words like "fine-tuning".
+        import re;
```
udithishanka: `import re` inside the `can route` ability gets re-executed on every walker invocation. Move it to the top of the file with the other imports.

```suggestion
import re;
```

jac-byllm/tests/fixtures/integration_byllm/tools.jac
```diff
@@ -0,0 +1,18 @@
+"""Safely evaluate a math expression"""
+def calculate(expr: str) -> str {
+    try {
+        return str(eval(expr));
```
udithishanka: `eval(expr)` with unconstrained input is an arbitrary code execution sink. This is a reusable LLM tool — if it ever gets wired to a real model, any string the LLM produces becomes executable. Use `ast.literal_eval` for simple expressions, or restrict to numeric-only input before evaluating.

```suggestion
        import ast;
        return str(ast.literal_eval(expr));
```

jac-byllm/tests/fixtures/integration_byllm/tools.jac
```diff
@@ -0,0 +1,18 @@
...
+    } except Exception as e {
+        return "Error while evaluating expression.";
+    }
+}
+
+"""Run and verify Python code."""
+def run_and_test_python_code(code: str) -> str {
+    try {
```
udithishanka, line 33: Same issue as `eval` above — `exec(code, {}, {})` with LLM-generated code is unsafe even with an empty globals/locals dict. For a test fixture this is fine, but add a comment making it clear this is test-only and should never be used as a template for production tools.

jac-byllm/tests/test_integration_byllm.jac
```diff
@@ -0,0 +1,301 @@
...
+import from urllib.request { Request, urlopen }
+
+
+"""Get a free port by binding to port 0 and releasing it."""
+def get_free_port -> int {
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s {
+        s.bind(("", 0));
+        s.listen(1);
+        port = s.getsockname()[1];
+    }
```
udithishanka, line 32: TOCTOU: the socket is released here before the server process starts. Another process can bind that port in the window between `get_free_port` returning and `Popen` actually binding. In CI with parallel test workers this will cause flaky failures. Safer approach is to pass `--port 0` to the server if jac-scale supports it and read the bound port from stdout, or at minimum set `SO_REUSEPORT` before releasing.

jac-byllm/tests/fixtures/integration_byllm/main.jac
```diff
@@ -0,0 +1,193 @@
...
+
+    "llm3": MockLLM(
+        model_name="mockllm",
+        config={"outputs": [], "verbose": False},
+    ),
+};
+
+# enums
+
+enum Correctness { CORRECT, PARTIAL, WRONG }
```
udithishanka: `Correctness`, `Intent`, `Personality`, `ConfidenceLevel`, `StudyAnswer`, and `Person` are declared but never referenced anywhere in the fixture or tests. Clean these up.

udithishanka (changes requested):
4 issues to address.

jac-byllm/tests/fixtures/integration_byllm/main.jac
Hirudika4: Fixed

jac-byllm/tests/fixtures/integration_byllm/tools.jac
Hirudika4: Replaced unsafe eval(expr) with ast.literal_eval(expr) in calculate() to safely evaluate math expressions without arbitrary code execution.

jac-byllm/tests/fixtures/integration_byllm/tools.jac
Hirudika4: Added prominent comment in run_and_test_python_code() clarifying this is test-only fixture code that should never be used as a production template.

jac-byllm/tests/test_integration_byllm.jac
Hirudika4: Added SO_REUSEPORT socket option before releasing the port. This allows immediate port reuse and prevents the race condition. Tests are now reliable even with parallel test workers.

jac-byllm/tests/fixtures/integration_byllm/main.jac
Hirudika4: Removed unused type definitions from test fixture

jac-byllm/tests/fixtures/integration_byllm/tools.jac
```diff
@@ -0,0 +1,23 @@
+"""Safely evaluate a math expression"""
+def calculate(expr: str) -> str {
+    import ast;
+    try {
+        return str(ast.literal_eval(expr));
```
udithishanka: `ast.literal_eval` only handles literals (strings, numbers, tuples, dicts) — it can't evaluate arithmetic expressions. `ast.literal_eval("(15 + 5) * 2 - (10 / 2)")` raises `ValueError`. So `calculate` always returns the error string for any actual math input.

The test still passes because MockLLM ignores tool results and returns the hardcoded `"35.0"` regardless — meaning this doesn't validate tool execution at all.

Use a safe AST-based evaluator instead:

```suggestion
        import ast;
        import operator;
        _ops = {
            ast.Add: operator.add,
            ast.Sub: operator.sub,
            ast.Mult: operator.mul,
            ast.Div: operator.truediv,
            ast.Pow: operator.pow,
            ast.USub: operator.neg,
        };
        def _eval(node: ast.AST) -> float {
            if isinstance(node, ast.Expression) { return _eval(node.body); }
            if isinstance(node, ast.Constant) and isinstance(node.value, (int, float)) { return node.value; }
            if isinstance(node, ast.BinOp) and type(node.op) in _ops { return _ops[type(node.op)](_eval(node.left), _eval(node.right)); }
            if isinstance(node, ast.UnaryOp) and type(node.op) in _ops { return _ops[type(node.op)](_eval(node.operand)); }
            raise ValueError(f"Unsupported expression: {ast.dump(node)}");
        }
        return str(_eval(ast.parse(expr, mode='eval')));
```

jac-byllm/tests/test_integration_byllm.jac
```diff
@@ -0,0 +1,302 @@
...
+        wait_for_port("127.0.0.1", server_port, timeout=60.0);
+        token = register_user(server_port);
+        test_fn(server, server_port, token);
+    } finally {
+        server.terminate();
+        try {
+            server.wait(timeout=15);
+        } except Exception {
+            server.kill();
+            server.wait(timeout=5);
```
udithishanka: Each of the 5 tests calls `with_byllm_server()` — that's 5 cold `jac start` startups with 60s timeouts, user registration, and teardown. In CI this could easily be 5+ minutes of just startup overhead.

Run a single server for all tests instead. Either refactor `with_byllm_server` to wrap all test callbacks in one invocation, or restructure as a module-level setup/teardown that starts the server once and shares the port+token across tests.

jac-byllm/tests/fixtures/integration_byllm/jac.toml
```diff
@@ -0,0 +1,19 @@
+[project]
+name = "integration_byllm"
+version = "1.0.0"
+description = "Jac client application: integration_byllm"
+entry-point = "main.jac"
+
+[dependencies.npm]
+jac-client-node = "1.0.7"
```
udithishanka: This is a server-side-only integration test but the toml pulls in npm deps (`jac-client-node`, `@jac-client/dev-deps`) and enables `[plugins.client]`. These are unused and will break CI if npm isn't available. Strip down to just `[project]` and the minimal `[serve]` config.

```suggestion

[serve]
```

udithishanka (changes requested):
3 issues to address.

jac-byllm/tests/fixtures/integration_byllm/tools.jac
Hirudika4: Fixed: calculate() now correctly evaluates arithmetic expressions using a safe AST-based evaluator.

jac-byllm/tests/test_integration_byllm.jac
Hirudika4: Fixed: Eliminated redundant server startups — server now starts once for all tests.
Replaced with_byllm_server() callback pattern with a module-level shared server approach:
- setup_byllm_server() — starts the server once and stores _server, _server_port, and _server_token as global state
- teardown_byllm_server() — gracefully shuts down after the last test

jac-byllm/tests/fixtures/integration_byllm/jac.toml
Hirudika4: Fixed: Stripped unnecessary npm dependencies and client plugin from integration test fixture's

marsninja:

PR #4583, opened 2026-02-12
`jacIsloggedIn` modified to check the Auth status 

MalithaPrabhashana:


jac-client/jac_client/tests/test_it.py
```diff
@@ -1604,7 +1604,7 @@ def test_pwa_build_generates_manifest_and_service_worker() -> None:
                 try:
                     print("[DEBUG] Fetching /manifest.json")
                     manifest_bytes = _wait_for_endpoint(
-                        f"http://127.0.0.1:{server_port}/manifest.json",
+                        f"http://127.0.0.1:{server_port}/static/manifest.json",
```
musab-mah-7, line 1607: why this change?

jac-client/jac_client/plugin/impl/client_runtime.impl.jac
```diff
@@ -206,7 +242,21 @@ impl __removeLocalStorage(key: str) -> None {
 """Auth guard component for file-based routing. Renders child routes if
 authenticated, otherwise redirects to the login path."""
 impl AuthGuard(redirect: str = "/login") -> any {
-    if jacIsLoggedIn() {
+    authState = reactUseState("checking");
```
musab-mah-7, line 245: we can use has variable for this right?

jac-client/jac_client/plugin/impl/client_runtime.impl.jac
```diff
@@ -67,9 +67,13 @@ impl __jacSpawn(left: str, right: str = "", fields: dict = {}) -> any {
...
-        error_text = await response.text();
-        walker_name = f"{left}/{right}" if right else left;
-        raise Exception(f"Walker {walker_name} failed: {error_text}") ;
+        if response.status == 401 {
+            __removeLocalStorage("jac_token");
+        } else {
+            walker_name = f"{left}/{right}" if right else left;
+            console.error(f"Walker {walker_name} failed with status {response.status}");
+        }
+        return {};
```
musab-mah-7, line 76: Why silently return {} instead of throwing an error? The caller has no way to know the operation failed.
How should the UI handle a failed walker call if it just gets an empty object back?

jac-client/jac_client/plugin/impl/client_runtime.impl.jac
```diff
@@ -206,7 +242,21 @@ impl __removeLocalStorage(key: str) -> None {
...
+        token = __getLocalStorage("jac_token");
+        if not token or token == "" {
+            authState[1]("unauthenticated");
+            return;
+        }
+        loggedIn = await jacIsLoggedIn();
+        authState[1]("authenticated" if loggedIn else "unauthenticated");
+    }
+    reactUseEffect(lambda -> None { checkAuth();}, []);
+    if authState[0] == "checking" {
```
musab-mah-7, line 256: What does returning None from a component actually render? Is this the intended UX (blank screen)?
Should there be a loading spinner or skeleton that User can provide if not a default jac laoder

jac-client/jac_client/plugin/impl/client_runtime.impl.jac
```diff
@@ -91,13 +95,22 @@ impl __jacCallFunction(function_name: str, args: dict = {}) -> any {
...
-        error_text = await response.text();
-        raise Exception(f"Function {function_name} failed: {error_text}") ;
+        if response.status == 401 {
+            __removeLocalStorage("jac_token");
+        } else {
+            console.error(
+                f"Function {function_name} failed with status {response.status}"
+            );
+        }
+        return None;
```
musab-mah-7, line 105: The error message is logged but the caller just gets None. How can they distinguish "no data" from "error occurred"?
Should this return a Result type or throw?

jac-client/jac_client/plugin/impl/client_runtime.impl.jac
```diff
@@ -181,7 +194,30 @@ impl jacLogout -> None {
 
 impl jacIsLoggedIn -> bool {
     token = __getLocalStorage("jac_token");
-    return token != None and token != "";
+    if not token or token == "" {
+        return False;
+    }
+    base_url = __getApiBaseUrl();
+    response = await fetch(
```
musab-mah-7, line 201: Why does a simple isLoggedIn check hit the network every time? This seems expensive. bascially, for every page chagne you think we need this?, better isntead f using this manully , we can inlcude this on AuthGuard no?

PR #4618, opened 2026-02-16
Add scheduled walkers support to jac-scale

kashmithnisakya:
## **Description**
Adds the ability to schedule walker execution in jac-scale, supporting both static scheduling via `@restspec(schedule=...)` decorator and dynamic scheduling via `REST API` endpoints.

##  Key changes:
  - New `scheduler.jac` and `impl/scheduler.impl.jac` with `ScheduleConfig`, `ScheduleJob`, `ScheduleStore`, and `SchedulerManager` using `APScheduler`
  - REST API endpoints: `POST /schedule/create`, `GET /schedule/list`, `GET /schedule/{id}`, `PUT /schedule/{id}`, `DELETE /schedule/{id}`
  - Three trigger types: `cron` (cron expression), `interval` (recurring), `date` (one-time)
  - Static schedules auto-register from `@restspec(schedule=...)` decorated walkers at server startup
  - Scheduler configuration via `[plugins.scale.scheduler]` in `jac.toml`
  - Integration tests for all schedule endpoints
  - Scheduler guide documentation

### DIFF
--- a/.jacignore
+++ b/.jacignore
@@ -261,6 +261,7 @@ test_restspec.jac
 test_sso.jac
 test_validate_md_files.jac
 test_webhook.jac
+test_scheduling.jac
 top_contributors.jac
 top_voices.jac
 validate_docs_code.jac


jac-scale/jac_scale/scheduler.jac
```diff
@@ -0,0 +1,145 @@
...
+obj SchedulerManager {
+    has _store: ScheduleStore by postinit,
+        _scheduler: Any | None = None,
+        _execution_handler: Any | None = None,
+        _walkers: dict[str, type] by postinit,
+        _is_running: bool = False;
+
+    def postinit -> None;
+    def start -> None;
+    def stop -> None;
```
musab-mah-7, line 136: where are we calling this? it  has to be called on server shut down no?

jac-scale/jac_scale/impl/scheduler.impl.jac
```diff
@@ -0,0 +1,528 @@
...
+impl SchedulerManager._run_async(coro: Any) -> Any {
+    import asyncio;
+    try {
+        loop = asyncio.get_running_loop();
+    } except RuntimeError {
+        loop = None;
+    }
+    if loop and loop.is_running() {
+        future = asyncio.run_coroutine_threadsafe(coro, loop);
+        return future.result(timeout=300);
```
musab-mah-7: this value should be configurable(propalby from toml)

jac-scale/jac_scale/impl/scheduler.impl.jac
```diff
@@ -0,0 +1,528 @@
...
+            if func?.func and func.func?.__globals__ {
+                func_globals = func.func.__globals__;
+                for name in dir(jaclib) {
+                    if not name.startswith('_') and name not in func_globals {
+                        func_globals[name] = getattr(jaclib, name);
+                    }
+                }
+            }
+            try {
+                _ = func.trigger;
```
musab-mah-7, line 503: DeadCode? waht's the effect of this?

jac-scale/jac_scale/impl/serve.impl.jac
```diff
@@ -2489,3 +2513,814 @@ impl JacAPIServer.register_dynamic_websocket_endpoint -> None {
     }
     self.server.app.websocket("/ws/{walker_name}")(dynamic_ws_handler);
 }
+
+# ============================================================================
+# Scheduler Support Implementation
```
musab-mah-7, line 2538: shall we move this to a seprate file? already this file is huge

jac-scale/jac_scale/impl/scheduler.impl.jac
```diff
@@ -0,0 +1,528 @@
...
+        return future.result(timeout=300);
+    } else {
+        return asyncio.run(coro);
+    }
+}
+
+impl SchedulerManager._execute_walker(schedule_id: str) -> None {
+    job = self._store.get(schedule_id);
+    if not job or not self._execution_handler {
+        return;
```
musab-mah-7, line 446: better if we log, cosole this, if nto wil lfail silently

jac-scale/jac_scale/impl/scheduler.impl.jac
```diff
@@ -0,0 +1,528 @@
...
+                self._execution_handler.spawn_walker(
+                    name=job.walker_name,
+                    fields=job.walker_fields,
+                    username=job.username
+                )
+            );
+        }
+        self._store.update_run_stats(schedule_id, success=True);
+    } except Exception as e {
+        logger.error(f"Failed to execute walker {job.walker_name}: {e}");
```
musab-mah-7: better if we have a simple return mechanism that is configurable in toml and defualt upto around 3 times(exponential backoff)

jac-scale/jac_scale/impl/scheduler.impl.jac
```diff
@@ -0,0 +1,528 @@
...
+    }
+}
+
+impl SchedulerManager._execute_system_walker(job: ScheduleJob) -> None {
+    if not self._execution_handler {
+        raise Exception("No execution handler configured") ;
+    }
+    user_manager = self._execution_handler.execution_manager.user_manager;
+    if not user_manager.user_exists(job.username) {
+        user_manager.create_user(job.username, '__no_password__');
```
musab-mah-7: what we are doing here?  cant 'we have this recorded in the user graph it self? this is a security loophole

jac-scale/jac_scale/impl/scheduler.impl.jac
```diff
@@ -0,0 +1,528 @@
...
+                setattr(job, field, datetime.fromisoformat(data[field]));
+            } except Exception { }
+        }
+    }
+    return job;
+}
+
+impl ScheduleStore.postinit -> None {
+    self._schedules = {};
+    self._use_mongodb = self._init_mongodb();
```
musab-mah-7, line 91: If someone deploys with mongodb_uri configured but MongoDB starts 5 seconds after the app, they get in-memory storage forever with no indication. Is that the intended behavior?

jac-scale/jac_scale/scheduler.jac
kashmithnisakya: Yeah, we should add it. I’ve added it.

jac-scale/jac_scale/impl/scheduler.impl.jac
kashmithnisakya: Added to the TOML config.

jac-scale/jac_scale/impl/scheduler.impl.jac
kashmithnisakya: Not dead code. It pre-caches a `@cached_property` at startup so the `trigger` type is resolved in the main thread context rather than failing later in the scheduler's background thread.

jac-scale/jac_scale/impl/serve.impl.jac
kashmithnisakya: Yeah, I have an idea to refactor the server file like this:
```
class Base:
    def __init__(self):
        self.x = 1


class Mixin1:
    def meth1(self):
        print(self.x)


class Mixin2:
    def meth2(self):
        print(self.x)


class Mixin3:
    def meth3(self):
        print(self.x)


class Combined(Base, Mixin1, Mixin2, Mixin3):
    pass
```


We should create a separate PR for this refactor. I suggested it to @marsninja as well, and he approved it.

jac-scale/jac_scale/impl/scheduler.impl.jac
kashmithnisakya: Added

jac-scale/jac_scale/impl/scheduler.impl.jac
kashmithnisakya: `self.user_manager.create_user('__system__', '__no_password__');`
Here we execute it as the system account

jac-scale/jac_scale/impl/scheduler.impl.jac
kashmithnisakya: Added a MongoDB retry mechanism

jac-scale/jac_scale/impl/scheduler.impl.jac
kashmithnisakya: Added

jac-scale/jac_scale/impl/scheduler.impl.jac
musab-mah-7: why we are asisgning it to a variable?

PR #4621, opened 2026-02-16
Fix symbol resolution for types within nested control flow statement inside `impl` blocks.

SandeepaHWP:
This fixes symbol resolution for Language Server Protocol (LSP) features inside nested control flow statements (`if`, `while`, `for`) within `impl` blocks. now go-to-definition, hover, and syntax highlighting work correctly for symbols in these contexts.

**Before and After Comparison**

image


kugesan1105 (approved):
LGTM !

PR #4663, opened 2026-02-18
feat(jac-scale): add resource tagging and jac status command

juzailmlwork:
## Summary

Two related improvements to jac-scale's Kubernetes integration:

1. **Resource tagging** — every Kubernetes resource created by jac-scale now carries a `managed: jac-scale` label, making it easy to identify, audit, and select all cluster resources owned by jac-scale.
2. **`jac status` command** — a new CLI command that queries the live cluster and prints a formatted table showing the health of every component (Jaseci App, Redis, MongoDB, Prometheus, Grafana) along with their service URLs.
3. [Breaking-change] **`health_check_path` configuration** — `health_check_path` can now only be set via `jac.toml` (through `k8s_config.health_check_path`). The previous `K8s_HEALTHCHECK_PATH` environment variable override has been removed; this field is no longer configurable via environment variable.

### Changes

**`abstractions/deployment_target.jac`**
- Added `get_full_status(app_name)` abstract method to the `DeploymentTarget` base class

**`plugin.jac`**
- Registered new `jac status <file_path>` CLI command that creates the deployment target, calls `get_full_status`, and renders the result
- Added `_print_deployment_status()` helper that uses `rich.table.Table` to render a colour-coded status table (● Running, ◑ Degraded, ⟳ Pending, ↺ Restarting, ✗ Failed, ○ Not Deployed) with a pod ready/total count and a Service URLs section
- Fixed `code_folder` resolution to use `pathlib.Path(...).parent` instead of `os.path.dirname(...) or '.'` (in both `_scale_pre_hook` and the deploy command)

**`kubernetes_target.jac`**
- Implemented `get_full_status()`: makes 4 bulk API calls (list deployments, list statefulsets, list pods, read app service) instead of one call per component, then derives status per component from the cached results
- Detects crash-looping pods (containers with `restart_count > 0`) and reports them as `restarting` rather than `pending`
- Reports Grafana URL only when the Grafana deployment is present
- Added `_get_component_status()` helper for single-resource status resolution (used internally)
- Added `managed: jac-scale` label to the app Deployment and app Service metadata
- Changed `health_check_path` to read exclusively from `self.k8s_config.health_check_path` (via `jac.toml`); removed the `os.getenv('K8s_HEALTHCHECK_PATH')` fallback — this field is no longer configurable via environment variable

**`monitoring.jac`**
- Added `managed: jac-scale` label to all Prometheus and Grafana resources (ConfigMaps, Deployment, Services, Secrets)

**`providers/database/kubernetes_mongo.jac`**
- Added `managed: jac-scale` label to MongoDB StatefulSet and Service

**`providers/database/kubernetes_redis.jac`**
- Added `managed: jac-scale` label to Redis Deployment and Service

**`targets/kubernetes/hpa/impl/hpa.impl.jac`**
- Added `managed: jac-scale` label to the HorizontalPodAutoscaler

**`targets/kubernetes/utils/kubernetes_utils.impl.jac`**
- Added `managed: jac-scale` label to PVCs, Secrets, and the code-sync Pod

**`tests/test_deploy_k8s.jac`**
- Added assertion that `get_full_status()` returns a non-null dict with the correct `app_name` after a successful deploy

- Added release note under `jac-scale 0.2.1 (Unreleased)`

### Motivation

Without consistent labels it is difficult to distinguish jac-scale-owned resources from other cluster workloads. The `managed: jac-scale` label enables `kubectl get all -l managed=jac-scale` across a namespace for quick auditing. The `jac status` command surfaces deployment health directly in the terminal without requiring users to run `kubectl` commands manually.

### DIFF
--- a/jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
+++ b/jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
@@ -33,7 +33,7 @@ import from jac_scale.factories.registry_factory { ImageRegistryFactory }
 import from jac_scale.targets.kubernetes.monitoring { MonitoringDeployer }
 import from jac_scale.targets.kubernetes.hpa.hpa { create_hpa }
 import time;
-import from kubernetes { client, config as k8s_config }
+import from kubernetes { client, config as k8s_cluster_config }
 import from kubernetes.client.exceptions { ApiException }
 
 """Kubernetes deployment target implementation."""
@@ -552,7 +552,11 @@ class KubernetesTarget(DeploymentTarget) {
             time.sleep(self.k8s_config.aws_nlb_wait);
             nlb_url = None;
             try {
-                config.load_kube_config();
+                try {
+                    k8s_cluster_config.load_kube_config();
+                } except ConfigException {
+                    k8s_cluster_config.load_incluster_config();
+                }
                 core_v1 = client.CoreV1Api();
                 service_obj = core_v1.read_namespaced_service(
                     f"{app_name}-service", namespace
@@ -652,7 +656,11 @@ class KubernetesTarget(DeploymentTarget) {
             );
             details["validate_resource_limits"] = "Successful";
             # 2. Initialize Kubernetes clients
-            config.load_kube_config();
+            try {
+                k8s_cluster_config.load_kube_config();
+            } except ConfigException {
+                k8s_cluster_config.load_incluster_config();
+            }
             apps_v1 = client.AppsV1Api();
             core_v1 = client.CoreV1Api();
             check_K8s_status();
@@ -1096,7 +1104,11 @@ class KubernetesTarget(DeploymentTarget) {
         }
 
         try {
-            config.load_kube_config();
+            try {
+                k8s_cluster_config.load_kube_config();
+            } except ConfigException {
+                k8s_cluster_config.load_incluster_config();
+            }
             apps_v1 = client.AppsV1Api();
             core_v1 = client.CoreV1Api();
             autoscaling_v2 = client.AutoscalingV2Api();
@@ -1223,7 +1235,11 @@ class KubernetesTarget(DeploymentTarget) {
 
     def get_status(self: KubernetesTarget, app_name: str) -> ResourceStatusInfo {
         try {
-            config.load_kube_config();
+            try {
+                k8s_cluster_config.load_kube_config();
+            } except ConfigException {
+                k8s_cluster_config.load_incluster_config();
+            }
             apps_v1 = client.AppsV1Api();
             core_v1 = client.CoreV1Api();
             namespace = self.k8s_config.namespace;
@@ -1262,7 +1278,11 @@ class KubernetesTarget(DeploymentTarget) {
         }
 
         try {
-            config.load_kube_config();
+            try {
+                k8s_cluster_config.load_kube_config();
+            } except ConfigException {
+                k8s_cluster_config.load_incluster_config();
+            }
             apps_v1 = client.AppsV1Api();
             namespace = self.k8s_config.namespace;
 
@@ -1290,7 +1310,11 @@ class KubernetesTarget(DeploymentTarget) {
 
     def get_service_url(self: KubernetesTarget, app_name: str) -> (str | None) {
         try {
-            config.load_kube_config();
+            try {
+                k8s_cluster_config.load_kube_config();
+            } except ConfigException {
+                k8s_cluster_config.load_incluster_config();
+            }
             core_v1 = client.CoreV1Api();
             namespace = self.k8s_config.namespace;
             service_name = f"{app_name}-service";
@@ -1331,7 +1355,11 @@ class KubernetesTarget(DeploymentTarget) {
     """
     def get_full_status(self: KubernetesTarget, app_name: str) -> dict {
         try {
-            config.load_kube_config();
+            try {
+                k8s_cluster_config.load_kube_config();
+            } except ConfigException {
+                k8s_cluster_config.load_incluster_config();
+            }
             apps_v1 = client.AppsV1Api();
             core_v1 = client.CoreV1Api();
         } except Exception as e {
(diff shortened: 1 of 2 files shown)


juzailmlwork:
image
@kashmithnisakya running locally but failing in pipeline

jac-scale/jac_scale/plugin.jac
```diff
@@ -212,6 +280,59 @@ class JacCmd {
...
+            examples=[("jac status app.jac", "Show deployment status of app"), ],
+            group="deployment",
+            priority=CommandPriority.PLUGIN,
+            source="jac-scale"
+        )
+        def status(file_path: str, target: str = "kubernetes") -> int {
+            if not os.path.exists(file_path) {
+                raise FileNotFoundError(f"File not found: '{file_path}'") ;
+            }
+            code_folder = os.path.dirname(filename) or '.';
```
udithishanka: Bug: `filename` is not defined in this scope — the function parameter is `file_path`. This will crash with `NameError` on every `jac status` invocation.

```suggestion
            code_folder = os.path.dirname(file_path) or '.';
```

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
```diff
@@ -1322,4 +1320,211 @@ class KubernetesTarget(DeploymentTarget) {
...
+
+    Makes 4 bulk API calls instead of one per component:
+      1. list_namespaced_deployment
+      2. list_namespaced_stateful_set
+      3. list_namespaced_pod  (to detect crash-looping pods)
+      4. read_namespaced_service  (app URL)
+    """
+    def get_full_status(self: KubernetesTarget, app_name: str) -> dict {
+        try {
+            k8s_config.load_kube_config();
```
udithishanka: Bug: `k8s_config` here is the jac-scale config object (`self.k8s_config`), not the `kubernetes` module's `config`. Should be `config.load_kube_config()` (from `import from kubernetes { client, config }`).

Also missing the in-cluster fallback. Compare with the existing pattern elsewhere in this file:
```jac
try { config.load_kube_config(); }
except ConfigException { config.load_incluster_config(); }
```

jac-scale/pyproject.toml
```diff
@@ -6,6 +6,8 @@ authors = [{ name = "Jason Mars", email = "jason@mars.ninja" }]
 readme = "README.md"
 requires-python = ">=3.12"
 dependencies = [
+    "jaclang>=0.11.0",
```
udithishanka: Duplicate: `jaclang>=0.11.1` already exists on line 11. This line (`jaclang>=0.11.0`) should be removed — keep only the stricter version constraint.

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
```diff
@@ -1322,4 +1320,211 @@ class KubernetesTarget(DeploymentTarget) {
         }
         return None;
     }
+
+    """Return status dict for a single Deployment or StatefulSet.
+
+    Returns dict with keys: status, ready, total.
+    status values: 'running', 'degraded', 'pending', 'not_deployed', 'unknown'.
+    """
+    def _get_component_status(
```
udithishanka: Dead code: `_get_component_status()` is defined but never called. `get_full_status()` resolves status inline from the bulk-fetched maps. Either remove this method or refactor `get_full_status()` to use it.

jac-scale/jac_scale/plugin.jac
```diff
@@ -33,6 +34,7 @@ def _scale_pre_hook(context: HookContext) -> None {
         if not os.path.exists(filename) {
             raise FileNotFoundError(f"File not found: '{filename}'") ;
         }
+        # code_folder = pathlib.Path(filename).parent;
```
udithishanka: Remove commented-out code (also on line 237).

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
```diff
@@ -706,9 +706,7 @@ class KubernetesTarget(DeploymentTarget) {
             repository_name = self._handle_image_registry(app_name, app_config);
 
             # 6. Get health check path
-            health_check_path = (
-                os.getenv('K8s_HEALTHCHECK_PATH') or self.k8s_config.health_check_path
-            );
+            health_check_path = self.k8s_config.health_check_path;
```
udithishanka, line 717: This removes the `os.getenv('K8s_HEALTHCHECK_PATH')` env var override that existed before. Is this intentional? If so it should be called out in the PR description. If not, this is a regression for anyone relying on that env var.

jac-scale/jac_scale/plugin.jac
juzailmlwork: removed

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
juzailmlwork: this is intentional.I will update the description

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
juzailmlwork: removed the dead code

jac-scale/pyproject.toml
juzailmlwork: removed

jac-scale/jac_scale/plugin.jac
juzailmlwork: fixed

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
juzailmlwork: pr is updated

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
juzailmlwork: k8s_config.load_kube_config(); this is replaced by config.load_kube_config();also removed in-cluster fallback from another place since it is only needed when running jac scale k8s command inside another pod.

PR #4664, opened 2026-02-18
Delete non default namespace for jac destroy

juzailmlwork:
## Summary

Previously, `jac destroy` would remove individual Kubernetes resources (deployments, services, PVCs, etc.) but leave the namespace itself behind when using a non-default namespace. Orphaned namespaces continue to consume cluster resources and can cause confusion. This PR deletes the namespace at the end of the destroy flow and adds a confirmation prompt to prevent accidental data loss.

### Changes

**`plugin.jac`**
- Resolves `app_name` and `namespace` before creating the deployment target so they are available for the confirmation prompt
- Prints a bold red warning showing the namespace about to be deleted
- Requires the user to type `y` to proceed; any other input cancels the operation and exits early

**`kubernetes_target.jac`**
- Imports the new `delete_namespace` utility
- After all resources are torn down, calls `delete_namespace(namespace)` if the namespace is not `default`
- Polls the Kubernetes API (up to 120 s, 2 s interval) waiting for the namespace to fully terminate
- Logs success or a timeout warning via the logger

**`kubernetes_utils.jac`**
- Added declaration for new `delete_namespace()` utility function

**`kubernetes_utils.impl.jac`**
- Implemented `delete_namespace(namespace)`: loads kube config, calls `delete_namespace` on the CoreV1 API, and handles 404 (namespace already gone) gracefully; skips deletion if namespace is `default`

**`tests/test_deploy_k8s.jac`**
- Updated the `deploy all in one` test to handle a 404 when listing PVCs (namespace may already be deleted by the time the assertion runs)
- Added an assertion that the namespace itself returns 404 after destroy completes

- Added release note under `jac-scale 0.2.1 (Unreleased)`

### Motivation

Non-default namespaces created by `jac deploy` should be fully cleaned up by `jac destroy`. Leaving namespaces behind wastes cluster resources and can cause NodePort or naming conflicts on subsequent deployments. The confirmation prompt ensures users do not accidentally destroy a shared namespace.


jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
```diff
@@ -1153,6 +1154,34 @@ class KubernetesTarget(DeploymentTarget) {
...
+                        time.sleep(ns_poll);
+                        ns_elapsed = ns_elapsed + ns_poll;
+                    } except ApiException as e {
+                        if e.status == 404 {
+                            break;
+                        }
+                        raise ;
+                    }
+                }
+                if self.logger {
```
udithishanka, line 1182: Bug: False success log on timeout. After the polling loop, this logs "deleted" even if the 120s timeout was reached without the namespace actually being deleted. Need a boolean flag to differentiate timeout vs actual deletion, and log a warning on timeout instead.

jac-scale/jac_scale/targets/kubernetes/utils/kubernetes_utils.impl.jac
```diff
@@ -487,6 +487,24 @@ impl create_tarball(source_dir: str, tar_path: str) -> None {
     Ensure that a given namespace exists in the Kubernetes cluster.
     If it doesn't exist and is not 'default', it will be created.
 """
+impl delete_namespace(namespace: str) -> None {
```
udithishanka, line 487: Bug: Docstring misplacement. The delete_namespace impl was inserted between the ensure_namespace_exists docstring (lines 486-489) and its impl block. The docstring "Ensure that a given namespace exists..." now incorrectly sits above delete_namespace.

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
```diff
@@ -1153,6 +1154,34 @@ class KubernetesTarget(DeploymentTarget) {
             }
             self._wait_for_deletion(app_name, namespace, apps_v1, core_v1);
 
+            # Delete namespace if not the default one
+            if namespace != 'default' {
```
udithishanka, line 1158: Deleting the entire namespace destroys all resources in it. If other workloads share the namespace, they'll be silently killed. 

Can we do a y/N check with a user confirmation that the user acknowledges that everything will be removed?

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
juzailmlwork: this is fixed  udith

jac-scale/jac_scale/targets/kubernetes/utils/kubernetes_utils.impl.jac
juzailmlwork: this is fixed udith

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
juzailmlwork: currently only other namespace are deleted and not the default namespaces.what do you think?

PR #4686, opened 2026-02-19
check-whether-ports-are-used-already-in-different-namespace

juzailmlwork:
## Summary

Before deploying to a local Kubernetes cluster, validate that the NodePorts requested for the app, Prometheus, and Grafana are not already bound by services in other namespaces. NodePorts are cluster-scoped, so conflicts across namespaces cause silent failures; this PR surfaces them as a clear error before deployment begins.

### Changes

**`kubernetes_utils.jac`**
- Added declaration for new `check_node_ports_available()` utility function

**`kubernetes_utils.impl.jac`**
- Implemented `check_node_ports_available(node_port, prometheus_node_port, grafana_node_port, current_namespace)`
- Lists all services across all namespaces, builds a map of occupied NodePorts (excluding the current namespace, whose services will be replaced)
- If any requested port is already in use, raises a `ValueError` listing each conflict with the owning `namespace/service-name` and instructions to choose different port values
- Moved the `from kubernetes import client` import to module level (was previously inside `cluster_type`)

**`kubernetes_target.jac`**
- Imports `check_node_ports_available` from the utils module
- After determining `cluster_env`, calls `check_node_ports_available` on local (non-AWS) clusters before any deployment work begins
- Prometheus and Grafana ports are only checked when `monitoring_enabled` is `True`

- Added release note for the new NodePort conflict check under `jac-scale 0.2.1 (Unreleased)`

### Motivation

NodePorts are cluster-wide resources. Two services in different namespaces cannot share the same NodePort. Without this check, port conflicts would surface as cryptic Kubernetes API errors mid-deployment. Failing fast with a descriptive message lets users correct their `KubernetesConfig` before any resources are created.

### DIFF
--- a/jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
+++ b/jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
@@ -678,7 +678,6 @@ class KubernetesTarget(DeploymentTarget) {
             service_type = 'LoadBalancer'
             if cluster_env in ('aws', 'gcp', 'azure')
             else 'NodePort';
-            ;
 
             # 4b. On local clusters, verify that the requested NodePorts are not
             #     already bound by a service in another namespace.


jac-scale/jac_scale/targets/kubernetes/utils/kubernetes_utils.impl.jac
```diff
@@ -322,6 +322,63 @@ Detect if the current connected cluster is AWS or local.
...
+) -> None {
+    """Check that the requested NodePorts are not already bound by services
+    in other namespaces.
+
+    NodePorts are cluster-scoped, so two services in different namespaces
+    cannot share the same NodePort value.  Services inside current_namespace
+    are excluded because they will be replaced by this deployment.
+    Raises ValueError listing every conflicting port and the service that owns it.
+    """
+    import from kubernetes { client }
```
udithishanka: put the import at the top of the file

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
```diff
@@ -676,9 +677,29 @@ class KubernetesTarget(DeploymentTarget) {
 
             # 4. Determine cluster type
             cluster_env = cluster_type();
-            is_aws = cluster_env == 'aws';
+            is_aws: bool = cluster_env == 'aws';
             service_type = 'LoadBalancer' if is_aws else 'NodePort';
 
+            # 4b. On local clusters, verify that the requested NodePorts are not
+            #     already bound by a service in another namespace.
+            is_local: bool = not is_aws;
```
udithishanka: aws specifc?? This breaks the jac scale being cloud agnostic .    gcp, azure will be considered as local clusters. logic is wrong.

jac-scale/jac_scale/targets/kubernetes/utils/kubernetes_utils.impl.jac
```diff
@@ -322,6 +322,63 @@ Detect if the current connected cluster is AWS or local.
...
+    # Build a map: port_number -> "namespace/service-name"
+    # Only include services from OTHER namespaces.
+    used_ports: dict[int, str] = {};
+    for svc in all_services.items {
+        if svc.metadata.namespace == current_namespace {
+            continue;
+        }
+        svc_id = f"{svc.metadata.namespace}/{svc.metadata.name}";
+        for port_obj in (svc.spec.ports or []) {
+            if port_obj.node_port {
```
udithishanka: node_port, if its 0, this breaks. use,` if port_obj.node_port is not None {`

jac-scale/jac_scale/targets/kubernetes/utils/kubernetes_utils.impl.jac
juzailmlwork: this is moved to top of the file

jac-scale/jac_scale/targets/kubernetes/utils/kubernetes_utils.impl.jac
juzailmlwork: this is fixed

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
juzailmlwork: for now we are going with aws specific.in upcoming ticket we are going to seperate load balancer layer where a  complete regactoring of cluster detection(aws,gcp,azure) and loadbalancer will be implemented

juzailmlwork: https://github.com/jaseci-labs/jaseci/pull/4758 rather than using is_aws we are going to use a variable called cluster_type and bring logic for each providers inside the relevant functions to adjust accordingly.this makes the code to work with future k8s clusters also

PR #4707, opened 2026-02-20
Patrick li/less accesses

Gorgeous-Patrick:
## Summary

- Avoid unnecessary `populate()` calls during edge/node traversal by passing raw `edge_type`/`node_type` from the compiler through to the runtime hot loop, enabling `issubclass()` checks on stub anchors **before** lazy loading is triggered.

## Problem

When executing `visit [->:Follow:->(?Profile)]`, the runtime previously loaded **all** edges and nodes connected to the current node into memory, then applied type filters (`isinstance`) in Python after each `populate()`. This caused unnecessary database/memory reads for edges and nodes that would be immediately discarded.

## Solution

**4 files changed across 3 layers (data model → compiler → runtime):**

| File | Change |
|---|---|
| `archetype.jac` | Added `edge_type` and `node_type` fields to `ObjectSpatialDestination` |
| `impl/archetype.impl.jac` | Plumbed `edge_type`/`node_type` through `append()`, `edge_out/in/any()`; Updated `make_stub()` to preserve `arch_cls` on stub anchors; Updated `__getstate__` to serialize `arch_cls` |
| `pyast_gen_pass.impl.jac` | Compiler emits `edge_type=<Type>` / `node_type=<Type>` keyword args when `FilterCompr.f_type` is present |
| `runtime.jac` | Added early-exit `issubclass(anchor.arch_cls, destination.edge_type)` check in `get_edges`, `get_edges_with_node`, `edges_to_nodes` — **before** accessing `.source`/`.target`/`.archetype` |

**Key design**: `arch_cls` is set directly on the stub's `__dict__` so it never triggers `__getattr__` → `populate()`. This is a zero-cost annotation on stubs that enables O(1) type rejection.

## Testing

- **1930 passed**, 0 regressions
- 1 pre-existing failure (`test_checker_pass::binary_op` — expects 2 errors, gets 6 — exists on `main`)
- No test modifications needed


Gorgeous-Patrick:
# `arch_cls` Serialization Bug (LessAccesses Branch)

## TL;DR
Commit `b81826b5` added `arch_cls: type(self.archetype)` to `Anchor.__getstate__()`. This caches the class type so the runtime knows which object to populate when it is running commands with a edge / node type filter, like `visit [->:Follow:->];`. However, Python's `pickle` cannot serialize class objects defined in `__main__`, causing **anchors to silently fail to persist/load from SQLite**.

## Verification: `arch_cls` is the Culprit

**main branch** `__getstate__` (works):
```python
return {
    'id': self.id,
    'archetype': unlinked,
    'root': self.root,
    'access': self.access,
    'persistent': self.persistent
}
```

**LessAccesses branch** `__getstate__` (broken):
```python
return {
    'id': self.id,
    'archetype': unlinked,
    'arch_cls': type(self.archetype),  # <-- ADDED, CAUSES FAILURE
    'root': self.root,
    'access': self.access,
    'persistent': self.persistent
}
```

The `archetype` instance itself pickles fine because Jac registers classes before unpickling. But `arch_cls` (the class object itself) fails because pickle stores it as `__main__.ClassName` and `__main__` changes between runs.

marsninja:

PR #4712, opened 2026-02-20
feat(byllm): Add ModelPool for LLM fallback and load-balancing via LiteLLM Router

Hirudika4:
### The Problem

byllm has zero resilience around LLM API calls. When a model hits a rate limit, timeout, or any transient API error — the entire Jac application crashes immediately. There's no retry, no fallback, no graceful degradation. This is a critical gap for production use and even for development workflows.

#### 1. Free-tier and budget-conscious users

Most LLM providers (Google Gemini, OpenAI, Anthropic, Groq, etc.) offer free tiers with strict rate limits — typically 15–60 requests per minute. Students, hobbyists, and developers prototyping with Jac hit these limits constantly. With ModelPool, they can pool multiple free API keys and keep their applications running uninterrupted:

```jac
import from byllm.lib { Model, ModelPool }
import os;

# Pool 3 free Gemini keys — 3x the effective rate limit at zero cost
glob llm = ModelPool(
    models=[
        Model(model_name="gemini/gemini-2.5-flash", api_key=os.getenv("GEMINI_KEY_1")),
        Model(model_name="gemini/gemini-2.5-flash", api_key=os.getenv("GEMINI_KEY_2")),
        Model(model_name="gemini/gemini-2.5-flash", api_key=os.getenv("GEMINI_KEY_3")),
    ],
    strategy="simple-shuffle",
);
```

Instead of "you need a paid API key", users can say "pool a few free keys and you're good to go."

#### 2. Production applications

Real-world AI applications need resilience. ModelPool enables:

- **Multi-provider redundancy:** Primary on Gemini, fallback to OpenAI, last resort on Anthropic
- **API key rotation:** Distribute load across multiple keys from the same provider
- **Zero-downtime:** Applications keep serving users even when one provider is degraded

#### 3. Cost optimization

With `cost-based-routing`, LiteLLM automatically selects the cheapest deployment using its built-in cost-per-token database ([100+ models](https://github.com/BerriAI/litellm/blob/main/model_prices_and_context_window.json)):

```jac
glob llm = ModelPool(
    models=[
        Model(model_name="gpt-4o-mini"),   # $0.15/$0.60 per 1M tokens
        Model(model_name="gpt-4o"),         # $2.50/$10.00 per 1M tokens
    ],
    strategy="cost-based-routing",
);
```

#### 4. Education and workshops

When running Jac workshops with 30+ students all sharing an API key, rate limits are guaranteed. `simple-shuffle` distributes load across multiple keys so no student's code crashes mid-demo.

### The Gap ModelPool Fills

| Scenario | Without ModelPool | With ModelPool |
|---|---|---|
| Rate limit hit | Application crashes | Router retries and falls back to next model |
| API provider outage | Application crashes | Router falls back to alternative provider |
| Free tier quota exhausted | Must wait or pay | Rotates to next free key |
| Production traffic spike | Requests fail | Load distributed across deployments |
| Need cheapest option | Manual research required | LiteLLM picks cheapest automatically |
| Need fastest response | Manual benchmarking | LiteLLM tracks latency and routes accordingly |

---

## Design: In-Process LiteLLM Router

`ModelPool` wraps **LiteLLM's `Router`** directly in-process. No subprocess, no proxy server, no HTTP overhead, no startup delay — the Router handles fallback, load-balancing, retries, and cost/latency tracking entirely within the same Python process.

### Architecture

```
Your Jac Code               ModelPool (BaseLLM)         LiteLLM Router (in-process)
─────────────  by llm()  ──────────────────────────  ──────────────────────────────
  def foo                                              
  by llm()  ────────────► ModelPool.invoke()          litellm.Router
                          ModelPool.model_call_*()  ──► ├─► gemini-2.5-flash  ✓
                                                        ├─► gpt-4o-mini  (fallback)
                                                        └─► claude-sonnet (last resort)
```

### How it works

1. `ModelPool(models=[...], strategy="fallback")` is created
2. `postinit` builds a `litellm.Router` model list from the provided `BaseLLM` models
3. The Router is initialized in-process with the chosen routing strategy
4. All calls are routed through `router.completion()` (or `router.acompletion()` for `cost-based-routing`)
5. The Router handles retries, cooldowns, fallbacks, and deployment selection transparently
6. `shutdown()` releases the Router instance (no process to terminate)

### Fallback Strategy — Provider Redundancy

Try models in order. If the first fails (rate limit, timeout, error), automatically try the next:

```jac
import from byllm.lib { Model, ModelPool }

glob llm = ModelPool(
    models=[
        Model(model_name="gemini/gemini-2.5-flash"),    # try first
        Model(model_name="gpt-4o-mini"),                 # if gemini fails
        Model(model_name="claude-sonnet-4-20250514"),    # last resort
    ],
    strategy="fallback",
);
```

### Simple-Shuffle — API Key Rotation / Load Distribution

Randomly picks a deployment each call. Ideal for distributing load across multiple free-tier keys:

```jac
import from byllm.lib { Model, ModelPool }
import os;

glob llm = ModelPool(
    models=[
        Model(model_name="gemini/gemini-2.5-flash", api_key=os.getenv("KEY_1")),
        Model(model_name="gemini/gemini-2.5-flash", api_key=os.getenv("KEY_2")),
        Model(model_name="gemini/gemini-2.5-flash", api_key=os.getenv("KEY_3")),
    ],
    strategy="simple-shuffle",
);
```

### Cost-Based Routing — Automatic Cheapest Model

LiteLLM checks its built-in cost database and routes every call to the cheapest available deployment:

```jac
import from byllm.lib { Model, ModelPool }

glob llm = ModelPool(
    models=[
        Model(model_name="gpt-4o"),          # $2.50/$10.00 per 1M tokens
        Model(model_name="gpt-4o-mini"),      # $0.15/$0.60 per 1M tokens
        Model(model_name="gpt-3.5-turbo"),    # $0.50/$1.50 per 1M tokens
    ],
    strategy="cost-based-routing",
);
```

> **Note:** `cost-based-routing` uses LiteLLM's async deployment selection (`acompletion`). ModelPool bridges this to the sync call path automatically via `asyncio.run()` — no changes needed in your Jac code.

You can also set custom per-token pricing for models not in LiteLLM's database:

```jac
glob llm = ModelPool(
    models=[
        Model(model_name="my-custom-model", config={
            "input_cost_per_token": 0.000001,
            "output_cost_per_token": 0.000002,
        }),
        Model(model_name="gpt-4o-mini"),
    ],
    strategy="cost-based-routing",
);
```

### Latency-Based Routing — Fastest Model

LiteLLM tracks response times using an EWMA (exponentially weighted moving average) and routes to the historically fastest deployment:

```jac
import from byllm.lib { Model, ModelPool }

glob llm = ModelPool(
    models=[
        Model(model_name="gemini/gemini-2.5-flash"),
        Model(model_name="gpt-4o-mini"),
        Model(model_name="groq/llama3-8b-8192"),
    ],
    strategy="latency-based-routing",
);
```

### Usage-Based Routing — Rate Limit Aware

Routes to the deployment with the lowest current TPM/RPM usage — keeps all deployments under their rate limits:

```jac
import from byllm.lib { Model, ModelPool }

glob llm = ModelPool(
    models=[
        Model(model_name="gemini/gemini-2.5-flash", api_key=os.getenv("KEY_1")),
        Model(model_name="gemini/gemini-2.5-flash", api_key=os.getenv("KEY_2")),
    ],
    strategy="usage-based-routing",
);
```

### Least-Busy — Concurrency Control

Routes to the deployment with the fewest in-flight requests. Best for controlling concurrency under heavy parallel load:

```jac
import from byllm.lib { Model, ModelPool }

glob llm = ModelPool(
    models=[
        Model(model_name="gpt-4o-mini"),
        Model(model_name="gpt-4o-mini", api_key=os.getenv("SECOND_KEY")),
    ],
    strategy="least-busy",
);
```

---

## Strategies Reference

| Strategy | `strategy=` value | How It Picks | Best For |
|---|---|---|---|
| Fallback | `"fallback"` | First model; next on failure | Provider redundancy, ordered priority |
| Simple Shuffle | `"simple-shuffle"` | Random pick each call | API key rotation, even load distribution |
| Cost-Based | `"cost-based-routing"` | Cheapest by LiteLLM cost DB | Cost optimization |
| Latency-Based | `"latency-based-routing"` | Fastest by EWMA response time | Performance-critical apps |
| Usage-Based | `"usage-based-routing"` | Lowest current TPM/RPM usage | Staying within rate limits |
| Least Busy | `"least-busy"` | Fewest concurrent requests | High-concurrency workloads |

---

### Constructor Parameters

| Parameter | Type | Default | Description |
|---|---|---|---|
| `models` | `list[BaseLLM]` | *(required)* | List of `Model` instances to pool |
| `strategy` | `str` | `"fallback"` | Routing strategy (see table above) |
| `num_retries` | `int` | `1` | Retries per deployment before moving on |
| `timeout` | `float` | `60.0` | Per-request timeout in seconds |

### Global Defaults via `jac.toml`

```toml
[plugins.byllm.model_pool]
strategy = "fallback"
num_retries = 1
timeout = 60.0
```

---

### Files Changed

| File | Change |
|---|---|
| `byllm/llm.jac` | `ModelPool` declaration — extends `BaseLLM`, fields for `models`, `strategy`, `num_retries`, `timeout`, `_router` |
| `byllm/llm.impl/modelpool.impl.jac` | **New file** — Full Router-based implementation: `postinit`, `model_call_no_stream`, `model_call_with_stream`, `shutdown` |
| `byllm/llm.impl/model.impl.jac` | `api_key=self.api_key or None` in streaming proxy branch (consistency fix) |
| `byllm/llm.impl/basellm.impl.jac` | API key fallback chain; verbose log shows actual model name from response |
| `byllm/lib.jac` | Export `ModelPool` |
| `jac-byllm/byllm/config_loader.jac` | Added `get_fallback_config()` method declaration |
| `jac-byllm/byllm/impl/config_loader.impl.jac` | Implemented fallback config defaults (`num_retries`, `timeout`) |

---

### 1. In-process LiteLLM Router over subprocess proxy

LiteLLM's `Router` class provides the same fallback, load-balancing, cost-based routing, latency tracking, and retry logic as the proxy — but runs directly in the calling process. This eliminates subprocess management, startup latency, temp file security concerns, and pipe deadlock risk entirely.

### 2. `ModelPool` extends `BaseLLM`

`ModelPool` is a drop-in replacement for `Model` — it overrides `model_call_no_stream` and `model_call_with_stream` to delegate to the Router. The rest of the `BaseLLM` dispatch, prompt formatting, and tool-call logic is reused unchanged. This means `by pool()` works identically to `by llm()`.

### 3. `cost-based-routing` uses `asyncio.run(acompletion())`

LiteLLM's cost-based deployment selection (`async_filter_deployments`) only runs on the async code path. For sync callers, ModelPool detects `strategy="cost-based-routing"` and calls `asyncio.run(router.acompletion(...))` transparently — no changes needed in Jac code.

### 4. `model_info: {"id": ...}` per deployment

LiteLLM's cost/latency/usage strategies require a unique `model_info.id` per deployment to track per-deployment metrics. ModelPool auto-generates these (`pool-model-id-0`, `pool-model-id-1`, ...) so users never need to set them manually.

### 5. Verbose logging shows the actual model served

When `verbose=True`, the log `Calling LLM: gpt-4o-mini` shows the **real model** from the response (e.g., `gpt-4o-mini-2024-07-18`), not the internal router alias (`pool-model`). This is critical for debugging which deployment was actually activated.

---

## Known Limitations

1. **`cost-based-routing` requires models in LiteLLM's cost DB** — If a model is not in [LiteLLM's cost map](https://github.com/BerriAI/litellm/blob/main/model_prices_and_context_window.json), it defaults to `$1/token`. Use `input_cost_per_token` / `output_cost_per_token` in `config` for custom models.

2. **`latency-based-routing` cold start** — On the first few calls, no latency history exists. The Router falls back to random selection until enough data is collected.

3. **`usage-based-routing-v2` requires Redis** — The v2 variant tracks usage across distributed processes via Redis. The default `usage-based-routing` works in-process without Redis.

4. **Streaming with fallback** — If a streaming response fails mid-stream (after chunks have been yielded), the fallback produces a complete new response from scratch. Already-yielded chunks cannot be recovered.

5. **`cost-based-routing` + `asyncio` event loop** — `asyncio.run()` creates a new event loop. If your Jac code already runs inside an async context (e.g., a web framework), use `strategy="usage-based-routing"` instead, which works on the sync path.

---

## Dependencies

| Dependency | Purpose | Required |
|---|---|---|
| `litellm` | In-process Router, cost DB, LLM calls | Yes |
| Standard library: `asyncio` | Bridge async `acompletion` for `cost-based-routing` | Built-in |

---


jac-byllm/byllm/llm.impl/modelpool.impl.jac
```diff
@@ -0,0 +1,243 @@
...
+        [
+            'litellm',
+            '--config',
+            self._config_path,
+            '--host',
+            self.host,
+            '--port',
+            str(self._assigned_port)
+        ],
+        stdout=subprocess.PIPE,
```
MalithaPrabhashana: checkout this page https://docs.litellm.ai/docs/routing
LiteLLM itself supports a routing mechanism. So we'll be able to use this for model pool. Shall we try this?

jac-byllm/byllm/llm.impl/modelpool.impl.jac
```diff
@@ -0,0 +1,153 @@
+"""ModelPool implementation — uses LiteLLM Router for in-process fallback and load-balancing."""
+import from byllm.mtir { MTRuntime }
+import from typing { Generator }
+
+"""Allow pool() call syntax, storing call_params and returning self."""
+impl ModelPool.__call__(**kwargs: object) -> ModelPool {
+    self.call_params = kwargs;
+    return self;
+}
```
MalithaPrabhashana: Why need to override again since we have this in the BaseLLM?

jac-byllm/byllm/llm.impl/modelpool.impl.jac
```diff
@@ -0,0 +1,153 @@
...
+    # Propagate verbose setting from any pool model to BaseLLM.log_info.
+    verbose = any(m.config.get("verbose", False) for m in self.models);
+    if verbose {
+        self.config["verbose"] = True;
+    }
+    logger.info(
+        f"ModelPool: LiteLLM Router initialized "
+        f"with strategy '{self.strategy}' and {len(self.models)} models."
+    );
+}
```
MalithaPrabhashana: The litellm_params building logic (api_key, base_url extraction) is nearly identical in both branches. Only the name differs. Try to get it to single loop

jac-byllm/byllm/llm.impl/modelpool.impl.jac
```diff
@@ -0,0 +1,153 @@
...
+    }
+    # cost-based-routing requires acompletion() — its deployment selection
+    # only runs on the async path (async_filter_deployments).
+    if self.strategy == "cost-based-routing" {
+        import asyncio;
+        response = asyncio.run(self._router.acompletion(**params));
+    } else {
+        response = self._router.completion(**params);
+    }
+    return response.model_dump();
```
MalithaPrabhashana: If we do like this, is it working with tool calling?
don't we need this `return self._router.completion(**params);`

jac-byllm/byllm/llm.impl/modelpool.impl.jac
Hirudika4: Tried with the LiteLLM's routing mechanism. It worked.

jac-byllm/byllm/llm.impl/modelpool.impl.jac
Hirudika4: This override is actually unnecessary. ModelPool already inherits the implementation from BaseLLM, and since jac doesn’t enforce return type annotations at runtime, self will always be the correct ModelPool instance.

I’ve removed the redundant override from both llm.jac and modelpool.impl.jac

jac-byllm/byllm/llm.impl/modelpool.impl.jac
Hirudika4: Refactored postinit to use a single loop, consolidating the litellm_params construction (including api_key and base_url extraction).

The is_fallback flag now controls both the model name assignment and whether fallback_names are accumulated.

The Router instantiation remains separate, as it requires different parameters (fallbacks= vs routing_strategy=) depending on the case.

Also standardized the model_info ID format to pool-model-id-{i} across both strategies, since these are only used as internal unique identifiers for the Router.

jac-byllm/byllm/llm.impl/modelpool.impl.jac
Hirudika4: Yes you are correct . response.model_dump() breaks tool calling.
In dispatch_no_streaming, we rely on attribute access (tool_call.function.name, tool_call.function.arguments), but model_dump() converts everything into plain dicts, causing AttributeError.

The Router already returns a ModelResponse (dict-like), so response.get("choices") works while preserving tool_calls as objects.

Fixed by returning the Router response directly.

jac-byllm/byllm/llm.impl/modelpool.impl.jac
```diff
@@ -0,0 +1,129 @@
...
+impl ModelPool.model_call_no_stream(params: dict) -> dict {
+    if self._router is None {
+        raise RuntimeError("ModelPool router not initialized.");
+    }
+    # cost-based-routing requires acompletion() — its deployment selection
+    # only runs on the async path (async_filter_deployments).
+    response: dict;
+    if self.strategy == "cost-based-routing" {
+        import asyncio;
+        response = asyncio.run(self._router.acompletion(**params));  # type: ignore
```
udithishanka: `asyncio.run()` crashes with `RuntimeError: This event loop is already running` inside FastAPI, Jupyter, or any framework with an existing event loop. LiteLLM Router's sync `router.completion()` already works for **all** strategies including `cost-based-routing` — there's no need for the async path at all.

Same issue on line 120 for streaming.

The entire `model_call_no_stream` and `model_call_with_stream` can each be a single line:

```suggestion
        response = self._router.completion(**params);  # type: ignore
```

And streaming:
```jac
impl ModelPool.model_call_with_stream(params: dict) -> Generator[str, None, None] {
    return self._router.completion(stream=True, **params);
}
```

No strategy branching, no `asyncio.run()`, no `_collect_stream` hack.

jac-byllm/byllm/llm.impl/model.impl.jac
```diff
@@ -79,6 +78,7 @@ impl Model.model_call_no_stream(params: dict) -> dict {
         } elif self.proxy {
             client = OpenAI(base_url=self.api_base, api_key=self.api_key or None);
             response = client.chat.completions.create(**params);
+            response = response.model_dump();
```
udithishanka: `response.model_dump()` converts the OpenAI `ChatCompletion` into nested plain dicts. Downstream in `dispatch_no_streaming`, tool calls are accessed via attribute access (`tool_call.function.name`), which raises `AttributeError` on plain dicts.

This breaks `Model(proxy=True)` with tool calling. The PR author already acknowledged this in the comment thread. This line should be removed — the proxy response works fine without it since `dispatch_no_streaming` uses `.get()` for the outer dict but attribute access for tool_calls.

```suggestion
```

jac-byllm/byllm/impl/config_loader.impl.jac
```diff
@@ -68,6 +69,17 @@ impl JacByllmConfig.get_litellm_config(self: JacByllmConfig) -> dict[str, Any] {
     };
 }
 
+"""Get fallback/resilience configuration from jac.toml."""
+impl JacByllmConfig.get_fallback_config(self: JacByllmConfig) -> dict[str, Any] {
```
udithishanka, line 73: `get_fallback_config()` is dead code — `ModelPool.postinit` never calls it. The `has` field defaults in `llm.jac` hardcode `strategy="fallback"`, `num_retries=1`, `timeout=60.0`, so setting `[plugins.byllm.fallback]` in `jac.toml` does nothing.

Either wire it up in `postinit` (read config, use sentinels like `None` to detect user-supplied vs config defaults) or remove the config section entirely. The docs and release notes claim this works, but it doesn't.

jac-byllm/byllm/llm.impl/modelpool.impl.jac
```diff
@@ -0,0 +1,129 @@
...
+    );
+}
+
+"""Make a direct model call without streaming via the LiteLLM Router.
+
+The router receives the same model alias used at init time ("pool-primary"
+or "pool-model") and handles fallback / load-balancing transparently.
+"""
+impl ModelPool.model_call_no_stream(params: dict) -> dict {
+    if self._router is None {
```
udithishanka: The `_router is None` guard and `_primary_model_name` field are unnecessary. `postinit` always runs before any call (Jac guarantees this), and `_primary_model_name` just duplicates `self.model_name` which is already set in `postinit`.

Also `response: dict;` pre-declaration + the two branches can be replaced with a single `return self._router.completion(**params);`.

Here's a cleaned-up alternative for the entire impl file (~70 lines vs 129):

```jac
"""ModelPool implementation — in-process LiteLLM Router."""
import from typing { Generator }

impl ModelPool.postinit -> None {
    import from litellm { Router }
    if not self.models {
        raise ValueError("ModelPool requires at least one model.");
    }
    is_fallback = self.strategy == "fallback";
    model_list: list = [];
    fallback_names: list = [];
    for (i, model) in enumerate(self.models) {
        name = (
            ("pool-primary" if i == 0 else f"pool-fallback-{i}")
                if is_fallback
                else "pool-model"
        );
        litellm_params: dict = {
            "model": model.model_name,
            "api_key": model.api_key or model.config.get("api_key", "") or None,
        };
        base_url = (
            model.config.get("base_url")
            or model.config.get("host")
            or model.config.get("api_base")
        );
        if base_url {
            litellm_params["api_base"] = base_url;
        }
        model_list.append({
            "model_name": name,
            "litellm_params": litellm_params,
            "model_info": {"id": f"pool-{i}"},
        });
        if is_fallback and i > 0 {
            fallback_names.append(name);
        }
    }
    if is_fallback {
        self._router = Router(
            model_list=model_list,
            fallbacks=[{"pool-primary": fallback_names}],
            num_retries=self.num_retries,
            timeout=self.timeout,
        );
        self.model_name = "pool-primary";
    } else {
        self._router = Router(
            model_list=model_list,
            routing_strategy=self.strategy,
            num_retries=self.num_retries,
            timeout=self.timeout,
        );
        self.model_name = "pool-model";
    }
    if any(m.config.get("verbose", False) for m in self.models) {
        self.config["verbose"] = True;
    }
    logger.info(
        f"ModelPool initialized with {len(self.models)} models, "
        f"strategy='{self.strategy}'"
    );
}

impl ModelPool.model_call_no_stream(params: dict) -> dict {
    return self._router.completion(**params);
}

impl ModelPool.model_call_with_stream(params: dict) -> Generator[str, None, None] {
    return self._router.completion(stream=True, **params);
}
```

No `asyncio.run()`, no `_collect_stream`, no `shutdown`, no `_primary_model_name`, no null guards. `router.completion()` is sync for all strategies.

jac-byllm/byllm/llm.impl/modelpool.impl.jac
Hushan-10: found that router.completion() does not work for "cost-based-routing".  it only has async_get_available_deployments(), no sync equivalent. LiteLLM designed cost-based-routing to be async-only because cost estimation (querying per-token costs, counting tokens for all candidates) is done asynchronously. 

so updated the implementation: all strategies except "cost-based-routing" now use router.completion() as suggested. For "cost-based-routing" only, we use acompletion() but run it through a ThreadPoolExecutor instead of bare asyncio.run(). This avoids the nested event loop crash you flagged ,  since asyncio.run() runs inside a dedicated thread, it always gets a fresh event loop and never conflicts with an existing one in FastAPI or Jupyter

jac-byllm/byllm/llm.impl/model.impl.jac
Hirudika4: Removed response.model_dump() from the proxy path. The OpenAI ChatCompletion object already supports .get() via its dict-like interface, so dispatch_no_streaming works fine without the dump.

jac-byllm/byllm/impl/config_loader.impl.jac
Hirudika4: Yes. get_fallback_config() was dead code. postinit was reading directly from the hardcoded has field defaults, so [plugins.byllm.fallback] in jac.toml had no effect.

Fixed using the sentinel approach you suggested, changed strategy, num_retries, and timeout has field defaults to None in llm.jac. At the top of postinit, we now call _byllm_config.get_fallback_config() and fill in any None fields from the config before any routing logic runs. User-supplied values (non-None) are left untouched, so explicit constructor arguments still always win over jac.toml.

jac-byllm/byllm/llm.impl/modelpool.impl.jac
Hirudika4: Applied all the valid simplifications: removed _router is None guards, removed _primary_model_name (setting self.model_name directly in both branches), removed shutdown, dropped the unused MTRuntime import, and cleaned up the postinit body.

However, we cannot remove the cost-based-routing async branch. After inspecting LiteLLM's source, the sync get_available_deployment() has no handler for "cost-based-routing" at all — it falls to else: deployment = None and raises "No deployments available". The async async_get_available_deployment() is the only path that handles it.

PR #4774, opened 2026-02-24
perf: Pyright-style `affected_symbols` optimization for CFG type narrowing

kugesan1105:
### Problem
- Type checking was **exponentially slow** when files had multiple independent `if` statements. For example, 6 `if y is not None` statements caused type narrowing to explore 2^6 = 64 paths, leading to **>1 minute** type check times.

### Solution: Pyright-style affected_symbols Optimization
- Instead of exploring ALL CFG predecessors during type narrowing, we track which symbols are affected (narrowed or assigned) in each branch. If a symbol isn't in affected_symbols, we skip the backward traversal entirely.

### Key Improvements
1. Performance: 6+ if statements now type-check in ~2s instead of >1 minute
2. Error Detection: `value.bit_length()` in else branch where `value `is `None `now reports:
                  `"<class NoneType> has no attribute bit_length"`
3. Member Access Uses Narrowed Types: After `if isinstance(x, int)`, accessing `x.method()` uses narrowed `int `type

### DIFF
--- a/docs/docs/community/release_notes/jaclang.md
+++ b/docs/docs/community/release_notes/jaclang.md
@@ -4,7 +4,6 @@ This document provides a summary of new features, improvements, and bug fixes in
 
 ## jaclang 0.11.1 (Unreleased)
 
-
 - **Perf: Type Narrowing Optimization**: Fixed exponential slowdown in `jac check` with many `if` statements (~1 min → ~2s). Member access now uses narrowed types and reports errors for invalid attribute access on `None`.
 - **Import Path Alias Resolution**: The module resolver now supports path aliases configured in `[plugins.client.paths]` in `jac.toml`. Aliases like `@components/Button` are resolved to their filesystem paths before standard module lookup, enabling cleaner imports in client-side Jac code.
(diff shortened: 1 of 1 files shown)


kugesan1105:
### Next PR
- `elif narrowing` doesn't work yet due to CFG node identity issues (predecessor nodes during type checking don't match nodes from CFG building)

jac/jaclang/compiler/passes/main/impl/cfg_build_pass.impl.jac
```diff
@@ -197,6 +197,16 @@ impl CFGBuildPass.exit_node(nd: uni.UniNode) -> None {
...
+        # Pyright-style: collect affected symbols for this CFG node.
+        # For IfStmt/ElseIf, collect symbols from isinstance checks in condition.
+        # For assignments/loops, collect assigned symbols.
+        if isinstance(nd, (uni.IfStmt, uni.ElseIf)) {
+            narrowing_syms = self._collect_narrowing_symbols(nd.condition);
+            nd.affected_symbols.update(narrowing_syms);
+        }
+        assigned_syms = self._collect_assigned_symbols(nd);
+        nd.affected_symbols.update(assigned_syms);
+
```
kugesan1105: This is the core of the optimization

Image

PR #4832, opened 2026-02-26
Assignment-based type narrowing

kugesan1105:
## Summary

- Assigning a more specific value to a union-typed variable now narrows the type for subsequent uses (e.g., `a: int | None = None; a = 90; add_one(a)` works without false positives)
- Added guard to prevent widening during narrowing - only applies when assigned type is strictly more specific


kugesan1105:
#4844

PR #4841, opened 2026-02-26
Fix : Support string escape decoding.

SandeepaHWP:
Escape sequences like `\x1b` (hex), `\033` (octal), and `\u001b` (unicode) were not being decoded correctly when used inside f-strings. For example:

```python

f"\x1b[{code}m"   # first char was '\' (ord 92) instead of ESC (ord 27)

```
Previously we are only manually handle `\n`, `\t`, `\r`, and `\\` .

Now we handle the edge cases like  `\x1b`.

### DIFF
--- a/docs/docs/community/release_notes/jac-mcp.md
+++ b/docs/docs/community/release_notes/jac-mcp.md
@@ -2,6 +2,8 @@
 
 ## jac-mcp 0.1.2 (Unreleased)
 
+- 2 Minor refactors/changes
+
 ## jac-mcp 0.1.1 (Latest Release)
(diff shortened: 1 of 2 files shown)


kugesan1105 (approved):
This is adhoc for now

we are aware of this: this `lit_value` shoudl be handled propely handled in **lexer** and **pyastload_pass**

Image

`/home/kuggix/jaseci/jac/jaclang/jac0core/impl/unitree.impl.jac` : 

```jac
 @ property
impl String.lit_value(self: String) -> str
 ```

Image

Issue #4793, opened 2026-02-25
byllm: Expose ReAct intermediate steps via event/callback system


udithishanka:
## Problem

When a `by llm(method="ReAct", tools=[...])` function executes, the ReAct loop produces rich intermediate data at each iteration — LLM reasoning text, tool call names + arguments, tool execution results, and iteration progress. **None of this is currently exposed to the caller.**

The user sees a "Thinking..." spinner, then the final streamed answer. The entire agent thought process is invisible — a black box.

### Current behavior
```
User sends message → "Thinking..." (opaque wait) → Final answer streams in
```

### Desired behavior
```
User sends message →
  💭 "I need to search for Jac walker syntax..."     (thought)
  🔧 Calling search_docs("walker syntax jac")        (tool call)
  📄 Found 3 relevant documents...                   (tool result)
  ✅ Done in 2 iterations
  Final answer streams in                            (streaming chunks)
```

---

### Design

Add an **opt-in** `logging=True` parameter to `by llm()` call_params. When enabled, the generator yields `StreamEvent` objects instead of plain strings — exposing every intermediate step.

**Zero breaking changes.** Existing code without `logging=True` is completely untouched.

```jac
obj StreamEvent {
    has event_type: str,   # "chunk" | "thought" | "tool_call" | "tool_result" | "steps_done"
        data: dict;        # Event-specific payload
}
```

### Event Types

| Event | When | Payload |
|---|---|---|
| `thought` | LLM produces reasoning text | `{"content": "I need to search...", "iteration": 1}` |
| `tool_call` | Before tool execution | `{"tool": "search_docs", "args": {"query": "..."}, "call_id": "...", "iteration": 1}` |
| `tool_result` | After tool execution | `{"tool": "search_docs", "result": "Found 3 docs...", "call_id": "...", "iteration": 1}` |
| `steps_done` | ReAct loop complete | `{"iterations": 2}` |
| `chunk` | Final answer token | `{"content": "Based on"}` |

---

### Tool calling with full visibility

```jac
import from byllm.lib { Model, StreamEvent }

glob llm = Model(model_name="gpt-4o");

"""Get the current weather for a city."""
def get_weather(city: str) -> str {
    return "80°F, Sunny";
}

"""Convert Fahrenheit to Celsius."""
def to_celsius(temp_f: float) -> str {
    return f"{round((temp_f - 32) * 5 / 9, 1)}°C";
}

"""Answer using tools with logging enabled."""
[...]

PR #4857, opened 2026-02-27
Implement thinking process for streaming in LLM invoke

Hushan-10:
image

## **Description**
This pull request introduces a new streaming event system for LLM invocations with logging, enabling richer and more structured streaming output for both tool and non-tool requests. The main changes add a `StreamEvent` type, update the invocation logic to yield structured events during streaming, and ensure all relevant modules are updated to support the new event type.

**Streaming event system and interface changes:**

* Introduced a new `StreamEvent` type in `byllm/types.jac` to represent unified streaming output events, including thoughts, tool calls, tool results, and answer chunks. 
* Updated imports and public interface in `byllm/lib.jac` and `byllm/llm.jac` to include `StreamEvent`, and added the `_invoke_streaming` method to the `BaseLLM` interface. 

**LLM invocation and streaming logic:**

* Modified `BaseLLM.invoke` in `basellm.impl.jac` to detect when logging is enabled with streaming, and route calls to the new `_invoke_streaming` method, which yields `StreamEvent` objects for all intermediate steps and answer chunks.
* Implemented the `_invoke_streaming` method in `basellm.impl.jac` to handle both tool-enabled and plain streaming requests, yielding structured events for thoughts, tool calls, tool results, and answer chunks, and handling iteration limits and final answer streaming. 

**Tool streaming behavior:**

* Changed `_stream_final_answer` to clear all tools (instead of keeping only `finish_tool`) before streaming the final answer, ensuring the LLM produces text output rather than invoking finish_tool again. 

These changes enable more granular and structured streaming output, especially useful for debugging and UI integration when logging is enabled.

### DIFF
--- a/docs/docs/reference/plugins/byllm.md
+++ b/docs/docs/reference/plugins/byllm.md
@@ -701,15 +701,17 @@ With `logging=True`, the user sees the first `tool_call` event after just one LL
 The final `StreamEvent` in every `logging=True` stream is a `usage` event containing aggregated token counts across all LLM calls in that invocation:
 
 ```jac
-for event in my_function("input") {
-    if event.event_type == "usage" {
-        # Aggregated totals across all LLM calls
-        print(event.data["total"]);
-        # e.g. {"prompt_tokens": 1200, "completion_tokens": 85, "total_tokens": 1285}
+with entry {
+    for event in my_function("input") {
+        if event.event_type == "usage" {
+            # Aggregated totals across all LLM calls
+            print(event.data["total"]);
+            # e.g. {"prompt_tokens": 1200, "completion_tokens": 85, "total_tokens": 1285}
 
-        # Per-call breakdown (one dict per LLM call in the ReAct loop)
-        for call_usage in event.data["per_call"] {
-            print(call_usage);
+            # Per-call breakdown (one dict per LLM call in the ReAct loop)
+            for call_usage in event.data["per_call"] {
+                print(call_usage);
+            }
         }
     }
 }


jac-byllm/byllm/llm.impl/basellm.impl.jac
```diff
@@ -268,3 +270,137 @@ impl BaseLLM._stream_final_answer(mt_run: MTRuntime) -> Generator[str, None, Non
...
+                    if token {
+                        yield StreamEvent(event_type="chunk", data={"content": token});
+                    }
+                }
+                mt_run.tools = original_tools;
+                done = True;
+            }
+
+            if not done {
+                resp = self.dispatch_no_streaming(mt_run);
```
udithishanka, line 439: Usage tracking is missing. The existing `invoke()` appends `resp.usage` to `self._usage_history` after every `dispatch_no_streaming()` call. This method calls `dispatch_no_streaming()` but never tracks usage. Any `stream=True, logging=True` call will silently stop accumulating token stats — `get_total_usage()` will be wrong.

```suggestion
                resp = self.dispatch_no_streaming(mt_run);
                self._usage_history.append(resp.usage);
```

jac-byllm/byllm/llm.impl/basellm.impl.jac
```diff
@@ -268,3 +270,137 @@ impl BaseLLM._stream_final_answer(mt_run: MTRuntime) -> Generator[str, None, Non
...
+
+        while not done {
+            iter_count += 1;
+
+            if int(max_iters) > 0 and iter_count > int(max_iters) {
+                yield StreamEvent(
+                    event_type="steps_done",
+                    data={"iterations": iter_count - 1, "reason": "max_iterations"}
+                );
+                # Inline _stream_final_answer to avoid triple-nested generator issue
```
udithishanka: The issue's own design (#4793) reuses `_stream_final_answer()` cleanly:
```
for token in self._stream_final_answer(mt_run) {
    yield StreamEvent(event_type="chunk", data={"content": token});
}
return;
```
That's 3 lines. Instead, this PR inlines the final answer logic **twice** (lines 306-322 and 344-360), ballooning from the issue's ~60 lines to ~137 lines. The "triple-nested generator issue" comment isn't a real issue — wrapping `_stream_final_answer()` in a for loop works fine since `yield from` inside `_stream_final_answer` produces tokens that the for loop consumes one level up.

jac-byllm/byllm/llm.impl/basellm.impl.jac
```diff
@@ -268,3 +270,137 @@ impl BaseLLM._stream_final_answer(mt_run: MTRuntime) -> Generator[str, None, Non
...
+    } else {
+        # ReAct loop with tools
+        max_iters = (
+            mt_run.call_params.get("max_react_iterations")
+            or self.call_params.get("max_react_iterations")
+            or _call_params_config.get("max_react_iterations")
+            or 0
+        );
+        iter_count = 0;
+        done = False;
```
udithishanka: The `done` flag pattern is fragile. The issue's design uses `return` to exit cleanly after finish_tool. With `done = True`, after processing finish_tool at line 360, the `for tool_call in resp.tool_calls` loop continues iterating — any remaining tool calls after finish_tool are silently dropped via the `if not done` guard at line 363. The original `invoke()` uses `return` (line 71-73) which is cleaner and avoids this edge case.

Recommend following the issue's design: use `while True` + `return` instead of `while not done` + `done = True`.

jac-byllm/byllm/llm.impl/basellm.impl.jac
```diff
@@ -255,11 +259,9 @@ impl BaseLLM._stream_final_answer(mt_run: MTRuntime) -> Generator[str, None, Non
         "Be comprehensive and synthesize all the information gathered.",
     );
     mt_run.add_message(final_instruction);
-    # Keep only finish_tool so Anthropic sees a valid tools= param
-    # (conversation history may already contain tool call results)
+    # Clear ALL tools so the LLM is forced to produce text, not call finish_tool again
     original_tools = mt_run.tools;
-    finish_tool_obj = mt_run.get_tool("finish_tool");
-    mt_run.tools = [finish_tool_obj] if finish_tool_obj else [];
+    mt_run.tools = [];
```
udithishanka: This change from `mt_run.tools = [finish_tool_obj] if finish_tool_obj else []` to `mt_run.tools = []` affects ALL streaming-with-tools callers — not just the new logging path. The deleted comment warned that Anthropic needs a valid `tools=` param when conversation history contains `tool_use` blocks. Sending `tools=[]` may break Anthropic models.

This change is also unnecessary if `_invoke_streaming()` reuses `_stream_final_answer()` as the issue's design proposes — then the final answer logic stays in one place and this method doesn't need to change.

jac-byllm/byllm/llm.impl/basellm.impl.jac
```diff
@@ -268,3 +270,137 @@ impl BaseLLM._stream_final_answer(mt_run: MTRuntime) -> Generator[str, None, Non
...
+
+                            # Execute the tool
+                            result_msg = tool_call();
+
+                            # Yield tool_result event after execution
+                            yield StreamEvent(
+                                event_type="tool_result",
+                                data={
+                                    "tool": tool_call.tool.get_name(),
+                                    "result": result_msg.content[:500],
```
udithishanka: Hardcoded 500 char truncation. If a tool returns structured JSON, this silently breaks the data. At minimum make this configurable via `call_params` or document it as a known limitation.

jac-byllm/byllm/llm.impl/basellm.impl.jac
```diff
@@ -13,6 +13,10 @@ Handles both streaming and non-streaming modes. For tool-enabled requests,
 implements a ReAct loop that continues until finish_tool is called.
 """
 impl BaseLLM.invoke(mt_run: MTRuntime) -> object {
+    # If logging is enabled with streaming, yield StreamEvent objects
+    if mt_run.stream and mt_run.call_params.get("logging", False) {
```
udithishanka, line 17: No tests for this feature. 155 additions, 0 test changes. `MockLLM` already supports streaming (`dispatch_streaming` yields random-sized chunks) and tool call simulation (`MockToolCall`). Should be straightforward to test both the no-tools path and the ReAct path with `logging=True`.

jac-byllm/tests/fixtures/logging_stream_react.jac
```diff
@@ -0,0 +1,81 @@
+"""Test logging=True streaming with ReAct tool calls."""
+import from byllm.lib { Model, MockToolCall }
```
udithishanka: Wrong import — `Model` uses litellm under the hood and doesn't understand `config={"outputs": [...]}`. That's `MockLLM`'s mock format. This test will try to call a real API with model name `"mockllm"` (which doesn't exist in litellm) and fail. `logging_stream_no_tools.jac` in the same PR correctly uses `MockLLM`.

```suggestion
import from byllm.lib { MockLLM, MockToolCall }
```

jac-byllm/tests/fixtures/logging_stream_react.jac
```diff
@@ -0,0 +1,81 @@
...
+def get_weather(city: str) -> str {
+    return f"The weather in {city} is sunny, 25°C.";
+}
+
+def finish_tool(final_output: str) -> str {
+    return final_output;
+}
+
+# Create a mock LLM that simulates ReAct: thought -> tool call -> finish
+glob llm = Model(
```
udithishanka:
```suggestion
glob llm = MockLLM(
```

jac-byllm/byllm/llm.impl/basellm.impl.jac
```diff
@@ -292,15 +337,249 @@ impl BaseLLM._stream_final_answer(mt_run: MTRuntime) -> Generator[str, None, Non
...
+                    if token {
+                        yield StreamEvent(event_type="chunk", data={"content": token});
+                    }
+                }
+                break;
+            }
+
+            try {
+                resp = self.dispatch_no_streaming(mt_run);
+            } except OutputConversionError {
```
udithishanka, line 440: Both this `OutputConversionError` recovery block and the no-tool-call recovery block (~line 520) are copy-pasted verbatim from `invoke()`. This is what Jayanaka is flagging — future fixes to `invoke()`'s ReAct recovery (like those in PR #4829) won't automatically apply here. If a new edge case is discovered in the recovery path, it needs to be patched in two places.

The right fix is to extract the shared ReAct iteration logic into a helper — e.g. `_run_react_step(mt_run) -> CompletionResult` — so both `invoke()` and `_invoke_streaming()` call the same recovery code.

jac-byllm/byllm/llm.impl/basellm.impl.jac
```diff
@@ -292,15 +337,249 @@ impl BaseLLM._stream_final_answer(mt_run: MTRuntime) -> Generator[str, None, Non
...
     mt_run.add_message(final_instruction);
     # Keep only finish_tool so Anthropic sees a valid tools= param
-    # (conversation history may already contain tool call results)
+    # (conversation history may already contain tool_use blocks)
     original_tools = mt_run.tools;
     finish_tool_obj = mt_run.get_tool("finish_tool");
     mt_run.tools = [finish_tool_obj] if finish_tool_obj else [];
+    # Force text generation — tool_choice="none" tells the LLM
+    # "tools are defined but do NOT call any, just produce text"
+    mt_run.call_params["tool_choice"] = "none";
```
udithishanka, line 317: If the caller already had a `tool_choice` in `call_params` (e.g. `by llm(tool_choice="auto", stream=True)`), this overwrites it — and then the `finally` at line 353 deletes it entirely without restoring the original. Should save and restore:

```suggestion
    original_tool_choice = mt_run.call_params.pop("tool_choice", None);
    mt_run.call_params["tool_choice"] = "none";
```

Then in the `finally`, after the pop, add:
```
if original_tool_choice is not None {
    mt_run.call_params["tool_choice"] = original_tool_choice;
}
```

udithishanka:
image

@Hushan-10 @Hirudika4 nice work!

let's address the above issues as well.

jac-byllm/byllm/llm.impl/basellm.impl.jac
Hushan-10: Resolved. Now saving the caller's original tool_choice via pop() before overriding with "none", and restoring it in the finally block. Follows the same save-override-restore pattern already used for mt_run.tools.

jac-byllm/byllm/llm.impl/basellm.impl.jac
Hushan-10: added test cases

jac-byllm/byllm/llm.impl/basellm.impl.jac
```diff
@@ -292,15 +306,230 @@ impl BaseLLM._stream_final_answer(mt_run: MTRuntime) -> Generator[str, None, Non
...
+                recovery_resp = self._attempt_recovery(mt_run);
+                if recovery_resp is None {
+                    raise ;
+                }
+                yield from self._yield_recovery_result(
+                    recovery_resp, iter_count, mt_run
+                );
+                break;
+            }
+            self._usage_history.append(resp.usage);
```
udithishanka, line 451: `_usage_history` is an instance-level list on `BaseLLM` that grows with every `_invoke_streaming` call but is never cleared and never exposed to consumers. Since LLM instances are typically long-lived (`glob llm = Model(...)`), this is an unbounded memory leak in production.

Either:
- clear it at the start of each `_invoke_streaming` call,
- emit a `StreamEvent(event_type="usage", ...)` at the end and reset, or
- remove it entirely until there's a consumer for it.

jac-byllm/byllm/llm.impl/basellm.impl.jac
```diff
@@ -292,15 +306,230 @@ impl BaseLLM._stream_final_answer(mt_run: MTRuntime) -> Generator[str, None, Non
...
+tool calls, tool results) and wraps final answer tokens in StreamEvent too.
+Works for both tool and non-tool streaming paths.
+"""
+impl BaseLLM._invoke_streaming(
+    mt_run: MTRuntime
+) -> Generator[StreamEvent, None, None] {
+    # No tools: wrap plain streaming tokens in StreamEvent
+    if len(mt_run.tools) == 0 {
+        for token in self.dispatch_streaming(mt_run) {
+            yield StreamEvent(event_type="chunk", data={"content": token});
```
udithishanka: Every other streaming path in this method filters empty tokens with `if token`, but this one doesn't. `dispatch_streaming` yields `delta.content or ""`, so you'll get a bunch of empty-string `StreamEvent`s here.

```suggestion
        for token in self.dispatch_streaming(mt_run) {
            if token {
                yield StreamEvent(event_type="chunk", data={"content": token});
            }
        }
```

jac-byllm/byllm/llm.impl/basellm.impl.jac
Hushan-10: Resolved. Extracted recovery logic into two helpers: _attempt_recovery() (shared precondition check + re-prompt + re-dispatch) and _yield_recovery_result() (streaming event emission after recovery). Both invoke() and _invoke_streaming() now call the same helpers , any future fix to recovery applies to both paths automatically.

jac-byllm/byllm/llm.impl/basellm.impl.jac
Hushan-10: Fixed. _usage_history is now reset at the start of each _invoke_streaming call. A final StreamEvent(event_type="usage") is emitted with aggregated total and per_call token usage (converted to plain dicts for JSON serialization), and the list is cleared immediately after emission.

jac-byllm/byllm/llm.impl/basellm.impl.jac
Hushan-10: Addressed. Added an if token guard to prevent emitting empty-string StreamEvent chunks in the no-tools path, ensuring consistent behavior with other streaming paths.

jac-byllm/tests/fixtures/logging_stream_react.jac
Hushan-10: fixed

jac-byllm/tests/fixtures/logging_stream_react.jac
Hushan-10: fixed

jac-byllm/byllm/llm.impl/basellm.impl.jac
Hushan-10: add the usage tracking

PR #4860, opened 2026-02-27
Fix type narrowing for walrus operator in if conditions

kugesan1105:
## Summary
- Add walrus operator (`:=`) support to type narrowing in if conditions
- Variables assigned via walrus are now correctly narrowed (e.g., `str | None` → `str` in truthy branch)
- Handles compound conditions like `if (a := x()) and (b := y())`


kugesan1105:
https://github.com/jaseci-labs/jaseci/pull/4909

PR #4908, opened 2026-03-01
Fixed dict.remove and del d[INDEX]

mgtm98:
Bug Summary:
    In Jac native, both `del d[k]` and `d.remove(k)` on a `dict[int, int]` (or
    any dict type) are silently dropped during IR generation — no LLVM IR is
    emitted for the delete operation.  The key remains in the dict after the
    call, and `k in d` still returns True.  The bug affects all dict value types
    (int, float, str, custom obj).  The root cause is in the dict codegen layer
    inside na_ir_gen_pass.impl (likely dicts.impl.jac or stmt.impl.jac) where
    delete/remove dispatch is either missing or produces a no-op.

### DIFF
--- a/docs/docs/community/release_notes/jaclang.md
+++ b/docs/docs/community/release_notes/jaclang.md
@@ -6,6 +6,7 @@ This document provides a summary of new features, improvements, and bug fixes in
 
 - 4 small refactors/changes.
 - **Type System Improvement**: Fixed type narrowing not working correctly inside while loops, for loops with break/continue, and loop else blocks.
+- **Fix: Match-Case & Walrus Type Narrowing**: Variables inside `match`/`case` blocks now correctly narrow to the matched type. Walrus operator (`if (x := get_optional())`) now narrows `x` to exclude `None` in the true branch.
 - **Fix: Formatter Comment Injection for `na {}` Blocks**: Fixed a bug where `jac format` would orphan comments inside `na {}` (native) blocks, dumping them at the end of the file.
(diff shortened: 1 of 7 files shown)


docs/docs/community/release_notes/jaclang.md
```diff
@@ -9,6 +9,8 @@ This document provides a summary of new features, improvements, and bug fixes in
...
 - **Native Primitives: Set Algebra & Dict `setdefault`**: Implemented 6 new native LLVM emitters -- `set.symmetric_difference`, `set.update`, `set.intersection_update`, `set.difference_update`, `set.symmetric_difference_update`, and `dict.setdefault`. Native primitive coverage rises from 47% → 49% implemented (147/299) and 45% → 47% tested (140/299), with SetEmitter at 61% and DictEmitter at 81%.
+- **Native Codegen: `del d[k]` and `d.remove(key)` for Dicts**: Both the `del` keyword (`del d[key]`) and the `.remove(key)` method now correctly remove entries from native dicts. Previously both were silently dropped, leaving the dict unchanged. The fix adds a `DeleteStmt` dispatch in `_codegen_stmt` and a new `emit_remove` emitter in `NativeDictEmitter` that shifts elements and decrements the length in-place. Works for both scalar-value dicts (`dict[int, int]`) and object-pointer-value dicts (`dict[int, T]`).
+- **Fix: `super().__repr__()` in `UniScopeNode`**: Corrected `super.__repr__()` (unbound descriptor) to `super().__repr__()` in `unitree.impl.jac`, fixing a crash when the scope symbol table was formatted (e.g. in error messages for native programs that declare archetypes).
```
kugesan1105: Rewrite this from the user’s POV and avoid exposing internal implementation details; if it’s purely internal, name it as“1 Minor internal refactors.” or increase the counter(1->2)

PR #4924, opened 2026-03-02
feat: replace string arch_cls with frozenset[TypeRef] for inheritance-aware type filtering

Gorgeous-Patrick:
## Summary

- Avoid unnecessary `populate()` calls during edge/node traversal by passing raw `edge_type`/`node_type` from the compiler through to the runtime hot loop, enabling `issubclass()` checks on stub anchors **before** lazy loading is triggered.

## Problem

When executing `visit [->:Follow:->(?Profile)]`, the runtime previously loaded **all** edges and nodes connected to the current node into memory, then applied type filters (`isinstance`) in Python after each `populate()`. This caused unnecessary database/memory reads for edges and nodes that would be immediately discarded.

## Solution

**4 files changed across 3 layers (data model → compiler → runtime):**

| File | Change |
|---|---|
| `archetype.jac` | Added `edge_type` and `node_type` fields to `ObjectSpatialDestination` |
| `impl/archetype.impl.jac` | Plumbed `edge_type`/`node_type` through `append()`, `edge_out/in/any()`; Updated `make_stub()` to preserve `arch_cls` on stub anchors; Updated `__getstate__` to serialize `arch_cls` |
| `pyast_gen_pass.impl.jac` | Compiler emits `edge_type=<Type>` / `node_type=<Type>` keyword args when `FilterCompr.f_type` is present |
| `runtime.jac` | Added early-exit `issubclass(anchor.arch_cls, destination.edge_type)` check in `get_edges`, `get_edges_with_node`, `edges_to_nodes` — **before** accessing `.source`/`.target`/`.archetype` |

**Key design**: `arch_cls` is set directly on the stub's `__dict__` so it never triggers `__getattr__` → `populate()`. This is a zero-cost annotation on stubs that enables O(1) type rejection.

## Testing

- **1930 passed**, 0 regressions
- 1 pre-existing failure (`test_checker_pass::binary_op` — expects 2 errors, gets 6 — exists on `main`)
- No test modifications needed


Gorgeous-Patrick:
## How TypeRef Collision Detection Works

### Problem

A `TypeRef` identifies a type by `(module, qualname)`. If two **distinct** Python classes share the same `module` and `qualname` — e.g., two classes both named `Foo` defined in `__main__` — then a single `TypeRef` is ambiguous: the shortcircuit optimization (checking `TypeRef in arch_cls` instead of calling `isinstance`) could match the wrong class. We need to detect this and fall back to the safe `populate() + isinstance` path.

### Mechanism: Weakref Registry + Collision Set

Two globals in `archetype.jac`:

- **`_TYPEREF_REGISTRY: dict[TypeRef, weakref.ref]`** — maps each `TypeRef` to a `weakref` pointing to the class object that first registered it.
- **`_TYPEREF_COLLISIONS: set[TypeRef]`** — set of `TypeRef`s known to be ambiguous (two or more live classes map to the same one).

### Detection Flow (`build_type_chain`)

When a stub is created (`make_stub`), we call `build_type_chain(cls)` which walks the class's MRO and, for each ancestor class:

1. **Construct** a `TypeRef(module=klass.__module__, qualname=klass.__qualname__)`.
2. **Check known collisions** — if this `TypeRef` is already in `_TYPEREF_COLLISIONS`, return `None` immediately (skip shortcircuit).
3. **Check the registry** — look up the `TypeRef` in `_TYPEREF_REGISTRY`:
   - If **no entry exists**: register it → `_TYPEREF_REGISTRY[ref] = weakref.ref(klass)`.
   - If an **entry exists**, dereference the weakref:
     - If the weakref is **dead** (class was GC'd, e.g., from a previous test run redefining `__main__`): overwrite silently — no collision, the old class no longer exists.
     - If the weakref is **alive** and points to the **same class** (`existing_cls is klass`): no collision — same class seen again (e.g., from a diamond MRO). Continue.
     - If the weakref is **alive** and points to a **different class** (`existing_cls is not klass`): **real collision** — two distinct live classes share the same `TypeRef`. Add to `_TYPEREF_COLLISIONS`, log a warning, return `None`.

4. If the full MRO is traversed without collision, return a `frozenset[TypeRef]` of the chain.

### Fallback Behavior

When `build_type_chain` returns `None`, `make_stub` simply **does not set `arch_cls`** on the stub anchor. At traversal time, the existing shortcircuit logic:

```python
if (_edge_type_ref and (ecls := anchor.__dict__.get('arch_cls')) and not type_ref_in_arch_cls(...)): continue
```

safely skips the optimization because `ecls` is `None` (falsy), causing the code to fall through to the full `populate()` + `isinstance()` path — which is always correct.

### Why Weakrefs?

Using `weakref.ref` instead of strong references means:
- Classes that get garbage-collected (e.g., between test runs that redefine classes under `__main__`) automatically expire in the registry.
- A dead weakref is treated as "slot available" — the new class takes over, no false positive.
- Only **simultaneously live** distinct classes with the same `TypeRef` trigger a collision.

### Test Isolation

Test fixtures call `clear_typeref_registry()` at the start to reset both `_TYPEREF_REGISTRY` and `_TYPEREF_COLLISIONS`, preventing cross-test contamination when multiple `.jac` fixtures define classes under the same `__main__` module in the same process.

marsninja:

PR #4948, opened 2026-03-04
Convert username and password for redis and mongodb to secret when injecting to pod deployment

juzailmlwork:
## Summary

- MongoDB and Redis credentials (`username`/`password`) are now loaded from `jac.toml` (`[plugins.scale.kubernetes]`) instead of being hardcoded
- Credentials are stored as Kubernetes `Secret` resources (`{app}-mongodb-secret`, `{app}-redis-secret`) and injected into pods via `valueFrom.secretKeyRef` — so `kubectl describe pod` shows secret references, not plaintext values
- Redis is started with `--requirepass` sourced from the secret env var; connection strings are updated accordingly
- Secrets are properly deleted on `destroy()`
- All-in-one test updated to verify secrets exist with correct data post-deploy and are gone post-destroy

## Why

Hardcoded credentials in pod specs are a security risk — they appear as plaintext in the Kubernetes API and any tooling that inspects deployments. Using `Secret` + `secretKeyRef` keeps credentials out of pod specs and follows Kubernetes security best practices.


jac-scale/jac_scale/providers/database/kubernetes_redis.jac
```diff
@@ -229,7 +247,31 @@ obj KubernetesRedisProvider(DatabaseProvider) {
                                 'ports': [
                                     {'containerPort': redis_port, 'name': 'redis'}
                                 ],
-                                'command': ['redis-server', '/etc/redis/redis.conf'],
+                                'command': [
+                                    'sh',
+                                    '-c',
+                                    'redis-server /etc/redis/redis.conf --requirepass "$REDIS_PASSWORD"'
```
udithishanka, line 258: this will break the deployment. Redis is now started with `--requirepass "$REDIS_PASSWORD"`, but the readinessProbe on L294 still runs bare `redis-cli ping`. An unauthenticated ping against a password-protected Redis returns `NOAUTH Authentication required` (not PONG), so the probe fails forever and the pod never becomes ready.

```suggestion
                                'command': [
                                    'sh',
                                    '-c',
                                    'redis-server /etc/redis/redis.conf --requirepass "$REDIS_PASSWORD"'
                                ],
```

and update the readinessProbe command to:
```jac
'exec': {'command': ['sh', '-c', 'redis-cli -a "$REDIS_PASSWORD" --no-auth-warning ping']}
```

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
```diff
@@ -449,6 +449,20 @@ obj KubernetesTarget(DeploymentTarget) {
             self.env_list.append(
                 {'name': 'MONGODB_URI', 'value': mongo_result['connection_string']}
```
udithishanka: the connection string has plaintext credentials embedded (`mongodb://admin:password@...`) and it's injected as a plain `value` env var into the app pod. `kubectl describe pod` on the app pod will still show the password in clear text — which is the exact thing this PR is trying to fix.

the same issue applies to `REDIS_URL` below (L548). both connection strings should either be stored in the respective K8s Secrets and referenced via `secretKeyRef`, or assembled at runtime from secret-backed env vars inside the container.

jac-scale/jac_scale/providers/database/kubernetes_redis.jac
```diff
@@ -293,14 +335,15 @@ obj KubernetesRedisProvider(DatabaseProvider) {
             }
         };
 
-        self.connection_string = f"redis://{redis_service_name}:{redis_port}/0";
+        self.connection_string = f"redis://{redis_username}:{redis_password}@{redis_service_name}:{redis_port}/0";
 
         logger.info(
             f"Prepared Redis deployment: {redis_name} "
             f"(connection: {self.connection_string}, config hash: {config_hash})"
```
udithishanka: this logs the full connection string at `info` level, which now includes the password (`redis://admin:password@...`). credentials shouldn't show up in logs.

```suggestion
            f"(connection: redis://{redis_service_name}:{redis_port}/0, config hash: {config_hash})"
```

docs/docs/community/release_notes/jac-scale.md
```diff
@@ -6,6 +6,8 @@ This document provides a summary of new features, improvements, and bug fixes in
 
 - **Automatic Port Fallback**: When starting the server with `jac start`, if the specified port is already in use, the server now automatically finds and uses the next available port instead of crashing with "Address already in use". A warning message displays when using an alternative port. Supports up to 10 port retries with cross-platform compatibility (Linux and Windows).
 - [fix]Fix for internet facing aws load balancer
+- 1 Minor refactor/change.
```
udithishanka: nit: "1 Minor refactor/change." doesn't describe anything. the existing entries in this file are descriptive — this should be too, or just remove this line since L10 already covers it.

juzailmlwork: this is removed

jac-scale/jac_scale/providers/database/kubernetes_redis.jac
juzailmlwork: this is fixed by suggestion

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac

jac-scale/jac_scale/providers/database/kubernetes_redis.jac
juzailmlwork: this is fixed

Issue #4666, opened 2026-02-18
partial update for anchor in presistance storage


juzailmlwork:
Currently memory hierachy fully replace the anchor when it is updated in persistance storage. this will cause a race condition if 2 requests are try to change different parameter of same anchor.so we need to support partial update.for that we may need serialized anchor stored

PR #4961, opened 2026-03-04
feat(jac-scale): Add partial update support for anchors to prevent race conditions in concurrent requests

### LINKED ISSUE #4616: [Bug] Jac Scaling Concerns (Pocketnest)
jaclang: 0.10.0
jac-scale: 0.1.7

Image

As discussed, latest jac architecture uses 3 layer memory. **As of this version scaling will not work.**

# L1 has no invalidation per request (current):
- With this rules, running multiple instances (workers/replicas) will not work.
> Since `--workers` is not supported yet, you may simulate this by running two same instance on different port.

With only this line `root ++> some_node();`, it will introduce memory misalignment across instances.
If all instances already L1 cached root node and one of them update it, other instances will not have the same data on L1. This will happen on every kind of update operations.

### Invalidating L1 per request will fix this issue and will allow scaling.

# L1 & L2 memory leaks
Since we cache every anchors and we don't invalidate cache, we can assume every L3 data will be in L2 and L1.
Most of the time in production, redis and server memory is significantly lower than database size.

With this architecture, L2 and L1 will be flooded by anchors and will always be full or faster to get full. For this reason, here are the consequences:
- redis and server RAM requires to be the same size as the database which you wouldn't prefer because it will significantly cost more. Upgrading will also not stop.
- redis can/might invalidate old data when it gets full to accomodate new data. This means the worst case scenario on this issue https://github.com/jaseci-labs/jaseci/issues/4552 will happen more often.
- Server might hangup/crash/slowdown when memory gets full

# Race Condition (when L1 has invalidation)
> With this we can now scale with multiple instances.

Jac doesn't support atomic operations such as increment, decrement push, pull, etc.
And since jac doesn't support partial update, it tends to update the full anchor.
This includes edges and archetype data.

With only this line `root ++> some_node();`, this will update the anchor with the full serialization of the current copy. This means if this line happen simultaneusly on two seperate request, they will both have the same copy (ex: no connect node yet):
Both request flow will have root node copy without edges yet (empty list).
Both will append one edge connecting to some_nodes (edges length == 1)
Request finished and committed.

The last request committed will be the latest data. root will only have one connected node even you trigger it twice. This issue wiill occur on every simultaneus update request within the same node. This also happen even on different fields.

akindu-k:
## Summary

Adds a `partial_update()` method to the persistent memory interface, enabling field-level updates to anchors without replacing the entire document. This prevents race conditions when multiple concurrent requests modify different fields of the same archetype.

## Problem

When two concurrent requests update different fields on the same node (e.g., one updates `title`, another updates `completed`), the last full-document write wins and silently overwrites the other's changes. 

## Solution

Introduce `partial_update(anchor, fields)` that writes **only the specified fields** to persistent storage, leaving other fields untouched.

### API Usage

```jac
import from jaclang.jac0core.runtime { JacPersistence }

# Update only the title field — other fields remain unchanged in storage
JacPersistence.partial_update(cast(Archetype, task), {"title": "New Title"});
```

### jaclang (core)

| File | Change |
|------|--------|
| `memory.jac` | Added `partial_update()` as an abstract method on `PersistentMemory` and declared it on `SqliteMemory` |
| `memory.impl.jac` | Implemented `SqliteMemory.partial_update` — reads the stored anchor, patches only the specified fields, writes back, and updates the in-memory cache |
| `runtime.jac` | Added `JacPersistence` utility class with `partial_update()` (handles L1/L2/L3 tiered memory, cache invalidation) and `save()` |
| `jaclang.md` | Release notes entry |

### jac-scale (MongoDB backend)

| File | Change |
|------|--------|
| `memory_hierarchy.jac` | Declared `partial_update()` on `MongoBackend` |
| `memory_hierarchy.mongo.impl.jac` | Implemented `MongoBackend.partial_update` using MongoDB `$set` with dot notation for surgical field updates |
| `jac-scale.md` | Release notes entry |

### Test fixtures & tests

| File | Change |
|------|--------|
| `todo_app.jac` | Added `UpdateTaskTitle` and `UpdateTaskCompleted` walkers that use `JacPersistence.partial_update`; added default values for `Task.id` and `Task.title` |
| `test_memory_hierarchy.jac` | Added `"concurrent partial updates to different fields"` integration test — creates a task, concurrently updates title and completed via separate HTTP requests, then verifies both fields are correctly persisted in MongoDB |

### SQLite (`SqliteMemory.partial_update`)
1. Reads the stored anchor from the database
2. Patches only the specified fields on the deserialized archetype
3. Writes the full anchor back (SQLite doesn't support partial column updates)
4. Updates the in-memory cache

### MongoDB (`MongoBackend.partial_update`)
1. Builds a `$set` operation using dot notation (e.g., `data.archetype.title`)
2. Serializes each field value individually
3. Executes `update_one` with `upsert=False` — falls back to full `put` if the document doesn't exist yet

### Cache Invalidation (`JacPersistence.partial_update`)
1. Delegates to the L3 backend's `partial_update`
2. Invalidates the L2 cache entry so subsequent reads fetch fresh data
3. Removes the anchor from the L1 in-memory cache to force a reload

## Testing

- **Integration test**: Spins up a MongoDB container, starts a `jac start` server, creates a task, fires two concurrent HTTP requests updating different fields, then queries MongoDB directly to verify both fields were updated without overwriting each other.

## Closes

#4616 & #4666


amadolid:
Hello @akindu-k ,

While this is a must have feature. I don't think this fixes the initial race condition in memory handler. Instead, it adds more layer of race condition.

Initial Issue (https://github.com/jaseci-labs/jaseci/issues/4616):
- Connecting nodes/edges utilize memory handling. Since the memory handler commits the full data, you will have race condition.
- L1 is not invalidated, this means if you have multiple workers/replicas you will have data discrepancy because they will have different "source of truth L1". Scaling will not work

The issue with this PR:
- Within the same instance
  - Since you invalidated the data in L1, if you have another request that's still running and referencing what you invalidated, their reference will be lost too. The changes they will do during that runtime will not be included in the commit.
- With multiple workers/replicas
  - Since you don't invalidate L1 on other instance, it will still not work. Other instance will still have their own unupdated data.
  - Incase you will support invalidation, the first issue of loosing the reference will also occur
  
You may try to run tests with two instances and simulate simultaneous request to have visibility on the race conditions issues.

PR #4962, opened 2026-03-04
Fix: String literal type checking

kugesan1105:
### Problem
The type checker was reporting false errors on common string operations:
- `text.split(" ")[0] + " suffix"` → "No matching overload for `__add__`"
- `msg = "hello"; msg += " world";` → "Cannot assign Literal to Literal"
- `len(f"hello {name}")` → "Cannot assign LiteralString to Sized"
- Any file with `b"..."` byte strings crashed

### Now supported
- String method chaining with concatenation works correctly
- String variable reassignment and `+=` operations
- Built-in functions like `len()` on f-strings and literals
- Byte string literals typed as `bytes`

### DIFF
--- a/.github/workflows/jac-check.yml
+++ b/.github/workflows/jac-check.yml
@@ -28,6 +28,14 @@ jobs:
           pip install -e ./jac-byllm
           pip install -e ./jac-super
 
+      - name: Run jac format (check formatting)
+        run: |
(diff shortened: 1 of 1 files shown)


jac/jaclang/compiler/type_system/impl/types.impl.jac
```diff
@@ -61,7 +64,12 @@ impl FunctionType._index_of_type_param(
...
-        if (type_param is type_var) {
+        # Compare by name instead of identity (Pyright uses scopeId + name)
+        if type_param.name and type_var.name and type_param.name == type_var.name {
+            return idx;
+        }
+        # Fallback to identity check for backwards compatibility
+        if type_param is type_var {
             return idx;
         }
     }
```
kugesan1105, line 75: Image

Issue #5070, opened 2026-03-10
`streamable-http` transport crashes on startup


SahanUday:
# Description
Running `jac mcp` with `--transport streamable-http` fails immediately with a `TypeError` before the server can start. The `stdio` and `sse` transports are unaffected. The root cause is a wrong keyword argument (`mcp_endpoint`) being passed to `StreamableHTTPServerTransport.__init__()`, which does not accept that parameter.

## Steps to Reproduce

`jac mcp --transport streamable-http --port 3001`

## Expected Behavior
The MCP server starts and listens for Streamable HTTP connections:

<img width="833" height="147" alt="Image" src="https://github.com/user-attachments/assets/e283fdbd-4464-49e1-9c4b-71f8c32b842b" />

## Actual Behavior
The server exits immediately with a pre-hook failure:

<img width="1167" height="53" alt="Image" src="https://github.com/user-attachments/assets/e136ed13-7a50-4bf5-b1a9-d59ebd1dc68e" />

## Root Cause Analysis
`jac_mcp/impl/server.impl.jac`

```
transport_inst = StreamableHTTPServerTransport(
    mcp_endpoint="/mcp", is_json_response_enabled=True
);
```
The argument `mcp_endpoint` does not exist in `StreamableHTTPServerTransport.__init__()`. 

```
def __init__(
    self,
    mcp_session_id: str | None,
    is_json_response_enabled: bool = False,
    event_store: EventStore | None = None,
    security_settings: TransportSecuritySettings | None = None,
    retry_interval: int | None = None,
) -> None:
```

The first positional argument is `mcp_session_id `, a session identifier for tracking individual client connections, not a URL path. The `/mcp` path is correctly used as the Starlette `Mount` prefix on line 228; it does not belong in the transport constructor at all.

PR #5073, opened 2026-03-10
Refactor `streamable HTTP` transport method of `jac-mcp`

SahanUday:
## Description
This pull request refactors the server initialization logic for the "streamable-http" transport to use a new session management approach. The main change is replacing the direct use of `StreamableHTTPServerTransport` with the new `StreamableHTTPSessionManager`, and updating the Starlette application to leverage an async lifespan context for managing session state.

**Refactor to use session manager for streamable HTTP transport:**

* Replaced the import and instantiation of `StreamableHTTPServerTransport` with `StreamableHTTPSessionManager`, introducing session management for the streamable HTTP server.
* Added an async lifespan context using `contextlib.asynccontextmanager` to ensure proper session management lifecycle within the Starlette app.
* Updated the Starlette app initialization to use the new session manager's `handle_request` method and the lifespan context.

### DIFF
--- a/docs/docs/community/release_notes/jac-scale.md
+++ b/docs/docs/community/release_notes/jac-scale.md
@@ -4,6 +4,10 @@ This document provides a summary of new features, improvements, and bug fixes in
 
 ## jac-scale 0.2.5 (Unreleased)
 
+- **NGINX Ingress Controller**: Replaced individual NodePort services with a single NGINX Ingress controller. All services are now ClusterIP, accessible via path-based routing through `ingress_node_port` (default: `30080`): `/` app, `/grafana`, `/cache-dashboard/`, `/db-dashboard`.
+- **Fix: Ingress routes now update correctly on re-deploy**: Switched from `patch` to `replace` for Ingress resources so toggling monitoring or dashboards off actually removes the old routes instead of leaving them in place.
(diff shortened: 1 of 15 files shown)


jac-mcp/tests/test_mcp.jac
```diff
@@ -22,6 +22,13 @@ test "server creation with custom config" {
     assert server.host == "0.0.0.0";
 }
 
+test "server creation with streamable-http transport" {
+    server = JacMcpServer(transport="streamable-http", port=9090, host="0.0.0.0");
+    assert server.transport == "streamable-http";
+    assert server.port == 9090;
+    assert server.host == "0.0.0.0";
+}
```
udithishanka, line 50: this test only checks constructor fields — it would pass on the broken code too since the bug was in `_async_start()`, not in the constructor. replace with a test that exercises the actual initialization path:

```suggestion
test "streamable-http transport initializes correctly" {
    import from mcp.server { Server }
    import from mcp.server.streamable_http_manager { StreamableHTTPSessionManager }
    import from starlette.applications { Starlette }
    import from starlette.routing { Mount }
    import contextlib;

    server = Server("test-mcp");
    session_manager = StreamableHTTPSessionManager(
        app=server, json_response=True, stateless=False
    );

    @contextlib.asynccontextmanager
    async def lifespan(app: object) -> object {
        async with session_manager.run() {
            yield;
        }
    }

    starlette_app = Starlette(
        lifespan=lifespan,
        routes=[Mount("/mcp", app=session_manager.handle_request)]
    );

    assert starlette_app is not None;
    assert session_manager is not None;
}
```

this mirrors the exact init code from `_async_start()` and would have caught the original TypeError from #5070.

SahanUday: updated

PR #5100, opened 2026-03-11
fix: correct ExecutionContext field types to non-optional

akindu-k:
## Summary
- Changed `system_root`, `user_root`, and `entry_node` fields on `ExecutionContext` from `NodeAnchor | None` to `NodeAnchor`, since `postinit` always initializes them to valid anchors.
- Removed the scattered `if jroot` null-guard workarounds in access validation and persistence logic that were only needed because of the incorrect `| None` type annotation.
- Updated release notes for 0.13.1.

## Context
Review feedback pointed out that the previous null checks in `check_access_level` and `save` were workarounds to satisfy the type checker rather than genuine runtime guards. Since `ExecutionContext.postinit` (and `JScaleExecutionContext.postinit`) always sets these fields to `system_root`, the root fix is correcting the type annotation — not adding defensive null checks throughout the codebase.

## Files Changed
- `jac/jaclang/runtimelib/context.jac` — Type annotations fixed
- `docs/docs/community/release_notes/jaclang.md` — Release note updated

### DIFF
--- a/.github/workflows/test-jaseci.yml
+++ b/.github/workflows/test-jaseci.yml
@@ -338,6 +338,11 @@ jobs:
         jac create jac-gpt --use https://raw.githubusercontent.com/jaseci-labs/jacpacks/refs/heads/main/jac-gpt/jac-gpt.jacpack
         cd jac-gpt && jac install
 
+    - name: Install Playwright dependencies
+      run: |
+        pip install playwright
+        playwright install chromium --with-deps
+
     - name: Type check jac-gpt
       working-directory: jac-gpt
       run: jac check main.jac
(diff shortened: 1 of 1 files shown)


jac/jaclang/jac0core/impl/runtime.impl.jac
```diff
@@ -107,8 +107,13 @@ impl JacAccessValidation.check_access_level(
...
-    if ((jroot == jctx.system_root) or (jroot.id == `to.root) or (jroot == `to)) {
-        return AccessLevel.WRITE;
+    if jroot {
+        if ((jroot == jctx.system_root) or (jroot == `to)) {
+            return AccessLevel.WRITE;
+        }
+        if (`to and (jroot.id == `to.root)) {
+            return AccessLevel.WRITE;
+        }
     }
```
kugesan1105: @akindu-k  This feels like a workaround to enable type checking.
It would be better if we improved the type checker to handle this case properly.

akindu-k: image

We can remove these None cases to fix it right

PR #5102, opened 2026-03-11
Support domain name for ingress

juzailmlwork:

### DIFF
--- a/docs/docs/community/release_notes/jac-client.md
+++ b/docs/docs/community/release_notes/jac-client.md
@@ -4,6 +4,8 @@ This document provides a summary of new features, improvements, and bug fixes in
 
 ## jac-client 0.3.6 (Unreleased)
 
+- **Fix: Desktop Target Asset Loading**: Fixed an issue where images and other static assets referenced with `/static/assets/` URLs were not loading in desktop (Tauri) builds. Assets are now correctly copied from `compiled/assets/` to `dist/static/assets/` during the build process, ensuring they are available when Tauri serves the frontend bundle. This fix applies to both `jac build --client desktop` and `jac start --client desktop` commands.
+
 ## jac-client 0.3.5 (Latest Release)
 
 - **Fix: Parser Strictness Compliance**: Moved docstrings before signatures across all test files (`test_cli`, `test_it`, `test_e2e`, `test_helpers`, `test_desktop_api_url`) and backtick-escaped `entry`/`walker` keyword parameters in `client_runtime` to comply with the stricter RD parser.
--- a/jac-client/jac_client/plugin/src/targets/impl/desktop_target.impl.jac
+++ b/jac-client/jac_client/plugin/src/targets/impl/desktop_target.impl.jac
@@ -1540,6 +1540,8 @@ impl DesktopTarget.build(
         os.environ.pop(API_BASE_URL_ENV_VAR, None);
     }
     console.print(f"  ✔ Web bundle built: {web_bundle_path}", style="success");
+    # Step 1.1: Copy assets to dist for desktop builds (Tauri serves from dist directly)
+    _copy_assets_to_dist(project_dir, web_bundle_path.parent);
     # Regenerate main.rs with latest TOML config (base_url may have changed since setup)
     _generate_main_rs(tauri_dir, api_base_url=toml_base_url);
     # Step 1.5: Bundle sidecar (optional - can be skipped if not needed)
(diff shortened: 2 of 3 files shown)


jac-scale/jac_scale/targets/kubernetes/kubernetes_config.jac
```diff
@@ -57,7 +57,10 @@ obj KubernetesConfig(BaseConfig) {
         redis_username: str = 'admin',
         redis_password: str = 'password',
         # Ingress configuration
-        ingress_node_port: int = 30080;
+        ingress_node_port: int = 30080,
+        # Domain / TLS configuration
+        domain: str = "",
+        cert_manager_email: str = "juzailmlwork@gmail.com";
```
udithishanka: This ships your personal email as the default for every user. Anyone who doesn't override `cert_manager_email` in their `jac.toml` will register Let's Encrypt certs under `juzailmlwork@gmail.com` and you'll get their expiry notices. Same issue on L191 in `from_dict`.

```suggestion
        cert_manager_email: str = "";
```

jac-scale/jac_scale/targets/kubernetes/monitoring.jac
```diff
@@ -984,7 +984,9 @@ obj MonitoringDeployer {
                                 'env': [
                                     {
                                         'name': 'GF_SERVER_ROOT_URL',
-                                        'value': '%(protocol)s://%(domain)s/grafana'
+                                        'value': f"https://{self.k8s_config.domain}/grafana"
```
udithishanka: This hardcodes `https://` during the initial deploy, but TLS isn't active yet at this point — the user hasn't run `--enable-tls`. Grafana will generate wrong redirect URLs for the entire HTTP phase.

Should use `http://` here. The `enable_tls` method already patches Grafana to `https://` later, so this just needs to match the initial HTTP state.

```suggestion
                                        'value': f"http://{self.k8s_config.domain}/grafana"
```

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
```diff
@@ -1041,6 +1041,25 @@ obj KubernetesTarget(DeploymentTarget) {
             details["health_check_of_deployment"] = deployment_details;
             service_url = self.get_service_url(app_name);
 
+            # Print CNAME instructions when a custom domain is set on AWS
+            if self.k8s_config.domain and cluster_env == 'aws' {
```
udithishanka: CNAME instructions are AWS-only but domain/TLS support isn't. Users on GCP, Azure, or any cloud with a LoadBalancer also need to know what to point their DNS at, but they'll get nothing here.

At minimum, try to resolve the LB hostname/IP for all cloud envs (the `_get_nlb_hostname` method already falls back to `.ip`), or print a generic "point your domain to your LoadBalancer IP" message for non-AWS clusters.

jac-scale/jac_scale/tests/test_deploy_k8s.jac
```diff
@@ -1378,3 +1380,75 @@ test "redis template file in package" {
...
+        with unittest.mock.patch(
+            "jac_scale.targets.kubernetes.kubernetes_target.client.NetworkingV1Api",
+            return_value=mock_networking_v1
+        ) {
+            assert deployment_target.get_service_url("myapp") == "http://app.example.com";
+        }
+    }
+}
+
+test "enable tls raises when no domain configured" {
```
udithishanka, line 1477: This test passes for the wrong reason. It only mocks `k8s_cluster_config` but not `client.NetworkingV1Api`, `client.CustomObjectsApi`, etc. So `IngressDeployer.enable_tls` blows up with a K8s connection error before it ever hits the domain validation check.

You're testing "can't reach cluster" not "rejects empty domain." Mock the networking/custom clients too so the actual domain-check codepath in `IngressDeployer.enable_tls` runs and raises the `"No domain found"` exception.

jac-scale/jac_scale/targets/kubernetes/kubernetes_config.jac
juzailmlwork: this is fixed and made mandatroy parameter

jac-scale/jac_scale/targets/kubernetes/monitoring.jac
juzailmlwork: this is fixed

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac

jac-scale/jac_scale/tests/test_deploy_k8s.jac
juzailmlwork: this is ixed

PR #5126, opened 2026-03-12
Fix[jac-mcp]: SSE `Transport TypeError: 'NoneType' object is not callable` on Client Disconnect

SahanUday:
## Problem
The `jac mcp --transport sse` server crashes with a `500 Internal Server Error` on every POST to `/messages/`, making the SSE transport completely unusable.

## Root Cause
**`sse.handle_post_message` was incorrectly registered as a Route endpoint:**
`handle_post_message` is a raw ASGI application (`scope, receive, send` signature), not a Starlette endpoint (`request` → `Response` signature). Using `Route(..., endpoint=...)` applied the wrong calling convention via the `request_response` wrapper.

## Fix
- Added `return Response()` after SSE context manager exit to satisfy Starlette's response contract.
- Replaced `Route("/messages/", endpoint=sse.handle_post_message, methods=["POST"])` with `Mount("/messages", app=sse.handle_post_message)` to correctly register the raw ASGI handler.


jac-mcp/tests/test_mcp.jac
```diff
@@ -49,6 +49,42 @@ test "streamable-http transport initializes session manager and app" {
     assert starlette_app.routes[0].path == "/mcp";
 }
 
+test "sse transport uses Mount for messages and handle_sse returns Response" {
+    import from mcp.server { Server }
+    import from mcp.server.sse { SseServerTransport }
+    import from starlette.applications { Starlette }
+    import from starlette.routing { Mount, Route }
```
udithishanka, line 55: this test doesn't actually validate the fix — it reconstructs the app inline with the corrected code and asserts on route types, so it passes regardless of what `_async_start` does in production. a regression in `server.impl.jac` wouldn't be caught here.

let's at least mock `uvicorn.Server.serve`, call `_async_start`, and inspect the resulting Starlette app's routes. that way you're testing the actual production code path, not a copy-paste of it.

docs/docs/community/release_notes/jac-mcp.md
```diff
@@ -2,6 +2,8 @@
 
 ## jac-mcp 0.1.5 (Unreleased)
 
+- **Fix SSE transport menthod issue**
```
udithishanka: typo: "menthod" → "method"

```suggestion
- **Fix SSE transport method issue**
```

jac-mcp/jac_mcp/impl/server.impl.jac
```diff
@@ -189,16 +190,13 @@ async def _async_start(srv: JacMcpServer) -> None {
...
 
             starlette_app = Starlette(
                 routes=[
                     Route("/sse", endpoint=handle_sse),
-                    Route(
-                        "/messages/",
-                        endpoint=sse.handle_post_message,
-                        methods=["POST"]
-                    ),
+                    Mount("/messages", app=sse.handle_post_message),
```
udithishanka: nit: stray blank line inside the routes list. the streamable-http block doesn't have this.

udithishanka (changes requested):
2 things to fix, 1 nit.

jac-mcp/tests/test_mcp.jac
SahanUday: test updated

docs/docs/community/release_notes/jac-mcp.md
SahanUday: fixed

jac-mcp/jac_mcp/impl/server.impl.jac
SahanUday: remove stray blank line

PR #5175, opened 2026-03-16
fix: Remove params replacement that breaks JIR cache serialization

kugesan1105:
The `validate_params_match` function was replacing `FuncSignature.params` items with objects from the impl definition while keeping `kid` unchanged. This caused `params` and `kid` to reference different ParamVar objects.

When JIR serializes a module:
1. `_assign_ids` only walks `kid` tree, assigning IDs to those nodes
2. `_write_field` for `params` calls `_get_node_id(impl_param)`
3. Since impl_param wasn't in `kid`, it has no ID -> returns EXTERNAL_REF
4. Reader skips EXTERNAL_REF -> params becomes empty after cache read

This broke type checking for functions like `traverse_graph` (13 params) which showed 0 params after JIR cache read, causing false E1051 errors.

Root cause: decl_impl_match_pass.impl.jac line 186 replacing params.

### DIFF
--- a/jac/jaclang/jac0core/passes/impl/decl_impl_match_pass.impl.jac
+++ b/jac/jaclang/jac0core/passes/impl/decl_impl_match_pass.impl.jac
@@ -181,8 +181,6 @@ impl DeclImplMatchPass.validate_params_match(
                 name=valid_decl.name_spec.sym_name
             );
         }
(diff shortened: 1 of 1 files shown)


kugesan1105:
https://chatgpt.com/share/69b77fa7-af48-800e-ba98-425a3e67b32a

PR #5198, opened 2026-03-17
 Fix ingress unreachable via IP/NLB when domain is set before --enable-tls

juzailmlwork:
## Problem

When `domain` is set in `jac.toml` and `jac start --scale` runs before `--enable-tls`, the ingress deployed with a `host` restriction, making the app unreachable via IP/NLB since no DNS was pointed yet.

## Fix

Ingress now always deploys with a wildcard rule (no `host`). The domain is stored in the `jac-scale/domain` annotation so `--enable-tls` can apply the host + TLS together in one step.


jac-scale/jac_scale/tests/test_deploy_k8s.jac
```diff
@@ -1521,3 +1522,50 @@ test "enable tls raises when no domain configured" {
...
+    # IP/NLB before TLS is enabled (the original bug: host set here caused nginx
+    # to reject requests arriving via the raw NLB address)
+    rule = created_body["spec"]["rules"][0];
+    assert "host" not in rule , "ingress rule must not set host before TLS is enabled";
+
+    # Domain must be persisted in the annotation so --enable-tls can read it later
+    annotations = created_body["metadata"]["annotations"];
+    assert annotations.get("jac-scale/domain") == "app.example.com" , "domain must be stored in jac-scale/domain annotation for --enable-tls";
+    assert annotations.get("jac-scale/cert-email") == "user@example.com" , "cert_manager_email must be stored in jac-scale/cert-email annotation for --enable-tls";
+}
```
udithishanka, line 1901: The Redis ingress had the same `if self.k8s_config.domain` → host-in-rule bug and it's fixed in this PR too, but this test calls `_deploy_ingress_resource` with `redis_dashboard=False` so the Redis path never runs. A regression there would go undetected.

Add a second call block with `redis_dashboard=True` and assert `"host" not in redis_rule` on the resulting Redis ingress body — same pattern as the main ingress check above.

juzailmlwork: added a new test case for this

jac-scale/jac_scale/tests/test_deploy_k8s.jac
```diff
@@ -1851,3 +1852,90 @@ test "enable tls raises when no domain configured" {
...
+    );
+    ingress_deployer = IngressDeployer(k8s_config, None);
+
+    mock_networking_v1 = unittest.mock.MagicMock();
+    # Simulate ingress not yet existing (first deploy) so create path is taken
+    mock_networking_v1.replace_namespaced_ingress.side_effect = ApiException(status=404);
+
+    mock_core_v1 = unittest.mock.MagicMock();
+    # Simulate basic-auth secret not yet existing
+    mock_core_v1.read_namespaced_secret.side_effect = ApiException(status=404);
```
udithishanka: this mock is never triggered — the code uses `replace_namespaced_secret`, not `read_namespaced_secret`. same issue on L1918 in the Redis test. the comment "Simulate basic-auth secret not yet existing" is misleading.

```suggestion
    mock_core_v1.replace_namespaced_secret.side_effect = ApiException(status=404);
```

jac-scale/jac_scale/tests/test_deploy_k8s.jac
```diff
@@ -1851,3 +1852,90 @@ test "enable tls raises when no domain configured" {
...
+    );
+    ingress_deployer = IngressDeployer(k8s_config, None);
+
+    mock_networking_v1 = unittest.mock.MagicMock();
+    # Simulate ingress not yet existing for both main and Redis ingresses
+    mock_networking_v1.replace_namespaced_ingress.side_effect = ApiException(status=404);
+
+    mock_core_v1 = unittest.mock.MagicMock();
+    # Simulate basic-auth secret not yet existing
+    mock_core_v1.read_namespaced_secret.side_effect = ApiException(status=404);
```
udithishanka: same dead mock as L1874 — `read_namespaced_secret` is never called here.

```suggestion
    mock_core_v1.replace_namespaced_secret.side_effect = ApiException(status=404);
```

jac-scale/jac_scale/tests/test_deploy_k8s.jac
juzailmlwork: addressed the commenrt

jac-scale/jac_scale/tests/test_deploy_k8s.jac
juzailmlwork: addressed the commenrt

PR #5209, opened 2026-03-18
feat: materialize topology index with type-keyed maps and MRO support

Jayanaka-98:
## Summary

Extends the topology index introduced in #5205 with two enhancements that improve query performance on local persistence backends and add support for Jac's archetype inheritance hierarchy.

### 1. Materialized Type-Keyed Maps

The original topology index stores adjacency data as a compact binary blob on root anchors, decoded and re-encoded on every access. This design is elegant for portability, but the decode/encode cycle becomes the dominant cost when L3 is local SQLite (sub-millisecond fetches).

We restructure the in-memory representation from a flat adjacency list (`list[tuple[int, int, int]]` scanned in O(E) per query) to a **type-keyed dict** (`sam[source]["n:TypeName"] -> [target_ids]`), giving O(1) lookup per type filter. The live object is cached on the anchor and only serialized during `__getstate__` for persistence — zero encode/decode overhead on the hot path.

### 2. MRO-Aware Indexing

The original index stores one type name per node and uses exact string equality for matching. This means `[-->(?:BaseContent)]` does not return `PostNode` or `CommentNode` subtypes — a correctness gap for Jac programs using archetype inheritance.

At edge-creation time, we now index the target under **each of its MRO ancestors** (e.g., `PostNode` is indexed under both `"n:PostNode"` and `"n:BaseContent"`). Parent-type queries resolve at zero additional query cost with no post-hoc `isinstance` checks.

### Performance

Benchmarked via `jac start` with L1 cache eviction between iterations (pure query timing, build excluded):

| Scenario | σ | n | main | This PR | Speedup |
|----------|---|---|----------|---------|---------|
| Single-hop | 0.1 | 100 | 0.21 ms | 0.07 ms | **3.2x** |
| Single-hop | 0.1 | 500 | 1.04 ms | 0.20 ms | **5.2x** |
| Single-hop | 0.1 | 1000 | 2.25 ms | 0.38 ms | **6.0x** |
| Inheritance | 0.25 | 100 | N/A | 0.11 ms | **N/A** |
| Inheritance | 0.25 | 500 | N/A | 0.49 ms | **N/A** |
| Inheritance | 0.25 | 1000 | N/A | 0.97 ms | **N/A** |
| Multi-hop | 0.3 | 15x20 | 0.43 ms | 0.40 ms | **1.1x** |
| Edge filter (combined) | — | 100 | 0.25 ms | 0.12 ms | **2.1x** |

> Inheritance queries are new in this PR — the original index (#5205) does not support MRO, so `[-->(?:BaseContent)]` does not return `PostNode` or `CommentNode` subtypes.

Storage overhead: ~10% larger than the original binary format (pickle vs hand-tuned struct).

### Changes

**Modified modules:**
- `runtimelib/topology_index.jac` — Replaced binary blob with type-keyed dict structure
- `runtimelib/impl/topology_index.impl.jac` — O(1) `query_targets()`, MRO fan-out in `add_edge()`, `resolve_chain()` using dict lookups, pickle-based encode/decode (persistence only)
- `runtimelib/topo_utils.jac` — Added `_get_type_mro()` for MRO extraction
- `runtimelib/impl/topo_utils.impl.jac` — MRO-aware edge indexing in `on_edge_created()`, direct UUID fetch in `batch_load_nodes()` (no edge re-walking)
- `jac0core/archetype.jac` — `_topology_index_cache` field on `NodeAnchor`
- `jac0core/impl/archetype.impl.jac` — Cache live index object on anchor; encode only during `__getstate__` for persistence

**Tests:** 22 total (13 unit, 3 integration, 6 e2e including 2 inheritance-specific)

## Test plan

- [x] 13 unit tests for type-keyed map operations: add/remove edge, node type queries, MRO fan-out, combined filters, chain resolution, encode/decode round-trip, cross-root IDs
- [x] 3 integration tests: config parsing, L3 fetch counter, anchor serialization survival
- [x] 6 e2e tests using native Jac graph syntax: exact-type queries, inheritance parent-type queries, index-vs-non-indexed correctness comparison, edge type filtering, null-root handling
- [x] All existing tests pass (no regressions)
- [x] Benchmark correctness verified across all 4 scenarios (single-hop, inheritance, multi-hop, edge filter)


marsninja:

PR #5215, opened 2026-03-18
feat(jac-scale): Add configurable MongoDB PVC storage size with safe resize support

juzailmlwork:
## Summary
- Replaces hardcoded `1Gi` MongoDB PVC storage size with a configurable `mongodb_storage_size` field
- Implements safe resize logic: increases are patched, same size is skipped, decreases are blocked
- Adds `parse_storage_quantity()` and `resize_pvc_if_needed()` utilities with unit tests

## Notes
StorageClass must have `allowVolumeExpansion: true` for live resizes to work.


jac-scale/jac_scale/tests/test_k8s_utils.jac
```diff
@@ -377,3 +378,40 @@ test "check pods restarted selector and return values" {
 
     assert result_stabilised is False , "pod that restarted once but is now running should not be flagged as crashed";
 }
+
+# --- resize_pvc_if_needed ---
+test "resize pvc if needed: increase patches, no-op skips, decrease raises" {
```
udithishanka, line 384: The 404 case is missing — this is the code path hit on every first deploy, when the StatefulSet was just created but the PVC hasn't been provisioned yet. Without it there's no test that the silent-return guard actually works.

```suggestion
test "resize pvc if needed: increase patches, no-op skips, decrease raises, 404 returns silently" {
```

jac-scale/jac_scale/tests/test_k8s_utils.jac
```diff
@@ -377,3 +378,40 @@ test "check pods restarted selector and return values" {
...
+        "10Gi"
+    );
+    raised = False;
+    try {
+        resize_pvc_if_needed(core_v1_shrink, "ns", "mongo-pvc", "1Gi");
+    } except ValueError {
+        raised = True;
+    }
+    assert raised , "Expected ValueError when trying to shrink PVC";
+}
```
udithishanka, line 471: Add a Case 4 to the test above for the PVC-not-found path, and then add a standalone parametrized test for `parse_storage_quantity` itself (compare to how `parse_cpu_quantity` and `parse_memory_quantity` are tested). Right now decimal suffixes (k, M, G), plain byte values, and invalid inputs are completely untested.

```suggestion
    assert raised , "Expected ValueError when trying to shrink PVC";
}

# --- parse_storage_quantity ---
def _test_parse_storage_valid(pair: tuple) {
    (raw, expected) = pair;
    import from jac_scale.targets.kubernetes.utils.kubernetes_utils { parse_storage_quantity };
    assert parse_storage_quantity(raw) == expected , f"parse_storage_quantity({raw!r}) should be {expected}";
}

parametrize(
    "parse storage quantity valid",
    [
        ("1Ki", 1024),
        ("1Mi", 1024 ** 2),
        ("1Gi", 1024 ** 3),
        ("500", 500),
        ("1k", 1000),
        ("1G", 1000 ** 3),
    ],
    _test_parse_storage_valid
);
```

jac-scale/jac_scale/targets/kubernetes/utils/kubernetes_utils.impl.jac
```diff
@@ -317,6 +317,78 @@ impl wait_for_pod_phase(
     );
 }
 
+"""Parse a Kubernetes storage quantity string into bytes (int).
+
+Supports binary suffixes (Ki, Mi, Gi, Ti, Pi, Ei) and decimal suffixes
+(k, M, G, T, P, E) as well as plain byte values.
+"""
+impl parse_storage_quantity(quantity: str) -> int {
```
udithishanka, line 325: `parse_memory_quantity` normalizes input to uppercase before matching, so `"10gi"` → `"10GI"` → matches `'GI'`. This function does case-sensitive matching, so `"10gi"` or `"10GI"` from jac.toml fall through both loops and hit `int("10gi")` which raises `ValueError: invalid literal for int() with base 10: '10gi'` — not helpful to the user. Either add `q = q.strip()` → use `.upper()` for the check (like parse_memory_quantity) or raise a clear `ValueError(f"Unrecognized storage suffix in '{quantity}'")`  at the end instead of letting `int()` fail.

udithishanka (changes requested):
3 issues to address.

jac-scale/jac_scale/tests/test_k8s_utils.jac
juzailmlwork: this  is fixed

jac-scale/jac_scale/targets/kubernetes/utils/kubernetes_utils.impl.jac
juzailmlwork: this is fixed

jac-scale/jac_scale/tests/test_k8s_utils.jac
juzailmlwork: this is updated

jac-scale/jac_scale/targets/kubernetes/utils/kubernetes_utils.impl.jac
```diff
@@ -317,6 +317,88 @@ impl wait_for_pod_phase(
...
+        'k': 1000,
+        'M': 1000 ** 2,
+        'G': 1000 ** 3,
+        'T': 1000 ** 4,
+        'P': 1000 ** 5,
+        'E': 1000 ** 6
+    };
+    for (suffix, multiplier) in suffixes_binary.items() {
+        if q.endswith(suffix) {
+            return int(q[:-len(suffix)]) * multiplier;
```
udithishanka: suffix-only inputs like `"Gi"` or `"Ki"` match here, then `int("")` blows up with `ValueError: invalid literal for int() with base 10: ''` — not the nice message at the bottom. `parse_memory_quantity` already handles this (line 46). Same guard needed here and in the decimal loop below.

```suggestion
            number_part = q[:-len(suffix)].strip();
            if number_part == '' {
                raise ValueError(
                    f"Storage quantity '{quantity}' is missing a numeric component."
                );
            }
            return int(number_part) * multiplier;
```

jac-scale/jac_scale/targets/kubernetes/utils/kubernetes_utils.impl.jac
```diff
@@ -317,6 +317,88 @@ impl wait_for_pod_phase(
...
+        }
+    }
+    for (suffix, multiplier) in suffixes_decimal.items() {
+        if q.endswith(suffix) {
+            return int(q[:-len(suffix)]) * multiplier;
```
udithishanka: same issue in the decimal suffix loop.

```suggestion
            number_part = q[:-len(suffix)].strip();
            if number_part == '' {
                raise ValueError(
                    f"Storage quantity '{quantity}' is missing a numeric component."
                );
            }
            return int(number_part) * multiplier;
```

udithishanka (changes requested):
1 issue to address.

jac-scale/jac_scale/targets/kubernetes/utils/kubernetes_utils.impl.jac
juzailmlwork: addressed

jac-scale/jac_scale/targets/kubernetes/utils/kubernetes_utils.impl.jac
juzailmlwork: addressed

juzailmlwork:
this is verified working at 21/04/2026 all the corner cases

udithishanka (approved):

PR #5234, opened 2026-03-19
Fix: generic functions like `find_parent_of_type` return Unknown instead of the correct type

SandeepaHWP:
## Summary

Fixes type checker returning `Unknown` for generic function calls like `find_parent_of_type[T](typ: type[T]) -> T | None`. This was causing ~20 false type errors in `sym_tab_build_pass.jac` and anywhere else these generic functions were used.

## The Problem

When calling a generic function like:
```jac
parent = nd.find_parent_of_type(uni.Module);  # should be Module | None
```
The type checker was returning `Unknown` instead of `Module | None`. This happened because of three separate issues:

### 1. FunctionType did not track type parameters
`FunctionType` had no `type_params` field. When the type checker built the type for a generic function like `def find_parent_of_type[T](typ: type[T]) -> T | None`, it had nowhere to store the `[T]` part. So at call sites, there was no way to know the function was generic or to solve what `T` should be.

### 2. `T | None` was collapsing to Unknown
The binary expression handler for `|` (used in union types like `T | None`) had an early return: if either side "is any type", return `Unknown`. `TypeVarType.is_any_type()` returns `True` (because a TypeVar can be anything), so `T | None` was short-circuited to `Unknown` before the union type could be formed.

### 3. type_params were lost during cloning
When a `FunctionType` was cloned through `specialize()` or `_apply_to_function()` (which happens when accessing methods through class instances), the clone did not copy `type_params`. So even after fixing issues 1 and 2, the type parameters were dropped before reaching the call site.

## The Fix

- **Added `type_params` field to `FunctionType`** to store function-level TypeVar declarations
- **Extract type params in `get_type_of_ability()`** so generic functions get their `[T]` tracked
- **Added TypeVar binding and substitution in `validate_call_args()`** — when a generic function is called, collect what each TypeVar maps to from the arguments, then substitute into the return type
- **Added `specialize_func_type_vars()` method** on `FunctionType` that takes solved bindings and returns a new FunctionType with TypeVars replaced by concrete types
- **Fixed union formation** — skip the `is_any_type` early return when the operator is `|`, so `T | None` correctly produces a `UnionType` instead of `Unknown`
- **Preserved `type_params` in clones** — both `specialize()` and `_apply_to_function()` now copy `type_params` to the cloned FunctionType
- **Added fallback inference for `type[T]` params** — when the parser produces empty type args for `type[T]` annotations, fall back to matching bare `type` params against Instantiable class arguments

### DIFF
--- a/jac/tests/langserve/test_server.jac
+++ b/jac/tests/langserve/test_server.jac
@@ -740,7 +740,7 @@ test "test_hover_types" {
             passes_main_fixture_abs_path("checker/checker_gap_func_typevar.jac")
         );
         lsp.type_check_file(typevar_file);
-        
+
         typevar_cases: list[tuple[int, int, str, str]] = [
             (37, 4, "ok1", "int"),
             (42, 4, "result1", "Animal"),


jac/jaclang/compiler/type_system/type_evaluator.impl/parameter_type_check.impl.jac
```diff
@@ -689,6 +696,21 @@ impl TypeEvaluator.assign_type_to_type_var(
...
+        and param_type.shared.class_name == "type"
+        and param_type.private
+        and param_type.private.type_args
+        and (arg_type.flags & types.TypeFlags.Instantiable)
+    ) {
+        for pta in param_type.private.type_args {
+            self.assign_type_to_type_var(pta, arg_type.clone_as_instance(), bindings);
+        }
+        return;
+    }
```
kugesan1105: it should return `bool`- True in this case


function signature is -> bool

jac/jaclang/compiler/type_system/type_evaluator.impl/construct_types.impl.jac
```diff
@@ -72,6 +72,12 @@ impl TypeEvaluator.get_type_of_ability(node_: uni.Ability) -> TypeBase {
     if node_.name_spec.type is not None {
         return node_.name_spec.type;
     }
+    func_type_params: list[types.TypeVarType] | None = None;
+    if node_.type_params {
+        func_type_params = [
+            types.TypeVarType(name=tp.sym_name) for tp in node_.type_params
```
kugesan1105: Only name is copied.
 If the user writes `def foo[T(bound=Animal)]`, the bound is silently dropped. 

check `assign_type_to_type_var ` whihc already has the logic to validate both `constraints` and `bounds`

jac/tests/compiler/passes/main/test_checker_gaps.jac
```diff
@@ -519,3 +520,30 @@ test "gap 50 postinit assignment tracking" {
...
+    # identity(42) returns int, assigning to str should error
+    assert "wrong_type" in all_errors ,
+        f"Gap #51: identity[T] not resolving T=int. Errors: {all_errors}";
+
+    # identity("hi") returns str, assigning to int should error
+    assert "wrong_type2" in all_errors ,
+        f"Gap #51: identity[T] not resolving T=str. Errors: {all_errors}";
+
+    # Should have exactly 2 type errors (the two wrong assignments)
+    assert len(program.errors_had) == 2 ,
```
kugesan1105: Also worth adding tests for:

`identity[T]` with a keyword argument call (`identity(val=42)`)
A function with 2 TypeVars (`def swap[T, U](a: T, b: U) -> tuple[U, T]`)

jac/jaclang/compiler/type_system/type_evaluator.impl/parameter_type_check.impl.jac
```diff
@@ -689,6 +696,21 @@ impl TypeEvaluator.assign_type_to_type_var(
         return True;
     }
 
+    if (
+        isinstance(param_type, types.ClassType)
+        and isinstance(arg_type, types.ClassType)
+        and param_type.shared
+        and param_type.shared.class_name == "type"
+        and param_type.private
+        and param_type.private.type_args
```
kugesan1105: `_is_type_subscript`

jac/jaclang/compiler/type_system/type_evaluator.impl/parameter_type_check.impl.jac
SandeepaHWP: chnaged

jac/jaclang/compiler/type_system/type_evaluator.impl/construct_types.impl.jac
SandeepaHWP: fixed

PR #5258, opened 2026-03-20
[Jac Mobile ] : Add mobile target with WebView-based architecture

mahfoos:
## **Description**

- Add mobile build target (`jac build/start --client mobile`) that wraps Jac web applications inside a React Native WebView using Expo SDK 54
  - Implement full mobile lifecycle: `jac setup mobile` (scaffolding), `jac start --client mobile --dev` (dev server with HMR), `jac build
  --client mobile` (production builds via EAS)
  - Add `MobileConfig` system for reading `[mobile]` configuration from `jac.toml`
  - Include documentation on mobile target architecture and self-hosted deployment strategy

  ## What's Included

  ### New Files
  - **`mobile_target.jac` + `mobile_target.impl.jac`** — `MobileTarget` class extending `WebTarget` with React Native wrapping (setup,
  build, dev, start)
  - **`mobile_config.jac` + `mobile_config.impl.jac`** — Configuration loader for `[mobile]` section in `jac.toml` with EAS profile
  generation
  - **`mobile/`** — Example scaffolded Expo project (WebView screen, bundle creator, assets)
  - **`JAC_MOBILE_TARGET.md`** — Full documentation of the mobile target architecture

  ### Modified Files
  - **`cli.jac`** — Extended `build` and `start` commands with `--client mobile`, `--platform android/ios`, `-b
  development/preview/production` options
  - **`register.jac` / `registry.jac`** — Register `MobileTarget` in the target factory
  - **`all-in-one/jac.toml`** — Added `[mobile]` and `[environments.mobile]` configuration sections

  ## Architecture

  Jac Source → ViteBundler → HTML/CSS/JS → create-bundle.js → bundle.ts (self-contained HTML)
                                                                      ↓
                                                       React Native WebView loads bundle

  **Two rendering modes:**
  - **Bundle mode** (default): Loads pre-built HTML from embedded assets (offline capable)
  - **Dev mode (HMR)**: Loads from Vite dev server with live reload


MusabMahmoodh (commented):
Hi @mahfoos, thanks for the big PR. I left line comments below. Few top level things first:

**1. Same problems as PR #5630 which I review earlier.** Duplicate `mobile/` at repo root, same `[mobile]` pollution in root `jac.toml`, same `create-bundle.js` bugs (base64 dance, regex missing `s` flag, `$` escape wrong), same `MobileConfig` that is ~70% copy-paste of `DesktopConfig`. My earlier comments on #5630 apply here also. Can we close #5630 since this PR supersede it, and fix all those points here together?

**2. `mobile_target.impl.jac` is 1800 lines in one file.** Please split — setup, bundle creation, dev mode, EAS build can be separate file under `targets/mobile/` folder (same way `targets/desktop/` already have subfolders).

**3. Docs file in code directory.** `JAC_MOBILE_TARGET.md` (537 lines) and `SELF_HOSTED_DEPLOYMENT.md` (345 lines) are placed inside `plugin/src/targets/`. Please move to `docs/docs/reference/` or similar, not mix with code.

**4. `MobileTarget.start()` just call `self.dev()`.** start and dev usually mean different thing — start is for run the built app, dev is with HMR. Please separate properly, or remove `start` method.

**5. `subprocess.Popen` with `stdout=PIPE, stderr=PIPE` but no drain** in `_start_backend_server`. When pipe buffer fill up (around 64KB on Linux), the backend process will deadlock. Please use `subprocess.DEVNULL` or drain in a background thread.

**6. Breaking API change.** Function `build()` parameter renamed from `target` to `client`. Any existing caller using `build(target="web")` will break.

For me the main blockers are item 1 (all the things from #5630), item 5 (PIPE deadlock), and the dead-code `USE_DEV_SERVER` flag (I comment on that line separately). Rest can clean up in review rounds.

mobile/app/index.tsx
```diff
@@ -0,0 +1,275 @@
+import React, { useState, useEffect } from 'react';
+import { View, Text, StyleSheet, Platform, ActivityIndicator } from 'react-native';
+import { WebView } from 'react-native-webview';
+
+// Dev mode: load from Vite dev server for HMR
+// Production: load from bundled HTML
+const USE_DEV_SERVER = __DEV__ && false; // Set to true for dev server mode
```
MusabMahmoodh: `__DEV__ && false` is always `false`, so this dev server mode never run. Then in `dev()` the message later say "Set USE_DEV_SERVER = true in .jac/mobile/app/index.tsx" meaning user must manually edit a generated TypeScript file just to switch mode. Very bad UX. Please remove the `&& false`, and control mode from `--dev` flag or environment variable instead.

mobile/app/index.tsx
```diff
@@ -0,0 +1,275 @@
...
+const getBackendApiUrl = () => {
+  if (__DEV__) {
+    if (Platform.OS === 'android') {
+      return 'http://10.0.2.2:9000';
+    }
+    // Use local network IP for iOS
+    const localIP = DEV_CONFIG.localIP || '127.0.0.1';
+    return `http://${localIP}:9000`;
+  }
+  return 'https://your-production-url.com';
```
MusabMahmoodh: `your-production-url.com` is a placeholder, this URL does not exist. If someone build production mode without changing this line, the app shipped to the user's phone will try to connect to a URL that never resolve. The `[mobile.features].api_url` in `jac.toml` already exist for this purpose — please wire it through, or fail the build if the value is still placeholder.

jac.toml
```diff
@@ -10,3 +10,18 @@ exclude = [
     "console.impl.jac",
     "cli_boot.jac"
 ]
+
+
+[mobile]
```
MusabMahmoodh: This root `jac.toml` is for the jaclang project itself, only have `[project]` and `[check.lint]` before this PR. Setting `name = "jaclang"` and `bundle_id = "com.example.jaclang"` here mean the jaclang compiler is now declare as a mobile app with placeholder bundle id. This is wrong. Sample config should go only in `jac-client/jac_client/examples/all-in-one/jac.toml`, not the root.

mobile/.gitignore
```diff
@@ -0,0 +1,32 @@
+# Expo
```
MusabMahmoodh: This whole top-level `mobile/` directory is a duplicate of `jac-client/jac_client/examples/all-in-one/mobile/`. Same `app.json`, same `create-bundle.js`, same assets. This version also have `package-lock.json` with 8833 lines committed. There is no precedent for top-level `mobile/` sitting beside `jac-client/`, `jac-coder/` etc. Please delete this whole directory and keep only the one under `examples/all-in-one/`.

docs/docs/community/release_notes/unreleased/jac-client/5258.feature.md
```diff
@@ -0,0 +1 @@
+- **Feat: Mobile Target with WebView-based React Native/Expo Support**: Added complete mobile target implementation (`jac build main.jac --client mobile`) that generates React Native apps using Expo Router. The mobile build extends the web target: it first builds the web bundle (TypeScript + React), then scaffolds an Expo project with WebView integration to render the Jac app. Includes `jac setup mobile` for project scaffolding, `[mobile]` configuration section in `jac.toml` for app metadata (bundle ID, display name, platforms, Expo settings), EAS Build integration for production builds (`-b production`), and dev server support (`jac start main.jac --client mobile --dev`) that runs both the web dev server and Expo development client. Templates include proper navigation setup, asset handling, and platform-specific configuration (iOS bundle identifier, Android package name, adaptive icons, splash screens).
```
MusabMahmoodh, line 1: The filename say `8053` but this is PR 5258, no? Should be `5258.feature.md`. If the towncrier style use issue number, please check issue number also is right. Mismatched number mean release note won't link to the correct PR.

jac-client/jac_client/plugin/src/targets/impl/mobile_target.impl.jac
```diff
@@ -0,0 +1,1800 @@
...
+    }
+
+    # Build using Expo
+    try {
+        if platform == "ios" {
+            console.print(
+                "  Building iOS app (requires Mac with Xcode)...", style="muted"
+            );
+            subprocess.run(
+                ["npx", "expo", "build:ios"], cwd=str(mobile_dir), check=True
```
MusabMahmoodh: `expo build:ios` and `expo build:android` were removed in Expo SDK 46 (early 2022). PR description say SDK 54, so these commands will only print a deprecation error and do nothing, the user see a build "success" that never produce any artifact. Modern Expo use `eas build --platform ios` / `--platform android`. Since this PR already have `_build_with_eas()` function, can we just delete `_build_mobile_app()` entirely and call EAS build only?

jac-client/jac_client/plugin/cli.jac
```diff
@@ -297,6 +314,57 @@ def _handle_build_target(ctx: HookContext) -> None {
...
+    if target_name == "mobile" and platform is None {
+        console.print("\n📱 Mobile Build Configuration", style="bold");
+        console.print("  Platform options:", style="muted");
+        console.print("    • android  - Build for Android devices", style="muted");
+        console.print(
+            "    • ios      - Build for iOS devices (requires Mac)", style="muted"
+        );
+        console.print("    • all      - Build for both platforms", style="muted");
+
+        platform_choice = input(
```
MusabMahmoodh, line 347: `input()` block waiting for keyboard input. In CI or any headless build (GitHub Actions, Docker build, etc.) this will hang forever with no indication what is wrong. Same problem for the EAS login prompt in `mobile_target.impl.jac`. Please skip the prompt when `--platform` and `--build-type` flag is already given on command line, and add a `--non-interactive` flag that fail loudly instead of prompting.

MusabMahmoodh (commented):
Part 2 — more things I found on a deeper pass. Anti-patterns, dead code, weird workaround, code smells. Line comments below.

Summary of what is below:
- WebView is configured with very loose security flags, and `[environments.mobile.response.headers]` also loosen COOP/COEP/CORS — look like a workaround rather than a fix.
- A full function (`_build_mobile_app`) call `expo build:ios/android` that was removed in Expo SDK 46 — it run but do nothing, user think build success.
- Weird workaround: `socket.connect(('8.8.8.8', 80))` just to get local IP (need internet for that).
- `subprocess.Popen(..., stdout=PIPE, stderr=PIPE)` in backend start without drain — buffer fill, process deadlock.
- `if target_name == "mobile"` in `cli.jac` break the polymorphic dispatch that the comment in the same function claim.
- `_check_eas_ready` return structured dict but caller use string match on `error_message` for branch — same pattern I flagged in PR #5604.
- Scope creep: a new `[desktop]` section added to `all-in-one/jac.toml` even though this PR is titled "Add mobile target".
- Small things: `.gitignore` `.jac/` → `.jac/*` subtle change, `10.0.2.2:9000` Android-emulator-only hardcoded, brittle string `replace` to inject backend URL into bundle.

jac-client/jac_client/examples/all-in-one/jac.toml
```diff
@@ -72,3 +69,93 @@ typescript = "^5.3.3"
 [environments.response.headers]
 Cross-Origin-Opener-Policy = "same-origin"
 Cross-Origin-Embedder-Policy = "require-corp"
+
+# Mobile-specific headers (for WebView compatibility)
+[environments.mobile.response.headers]
+Cross-Origin-Opener-Policy = "unsafe-none"
```
MusabMahmoodh, line 86: Here the mobile environment override the strict `same-origin` / `require-corp` COOP/COEP (from `[environments.response.headers]` above on line 70-71) with `unsafe-none`, and also open CORS to `*`. This look like a workaround — probably the dev could not get WebView to load with strict headers, so relax everything instead of fixing the real issue (example: use `credentialless` COEP, or load bundle via blob URL). Please document WHY each line need to be relaxed, or find tighter config that still work for WebView.

jac-client/jac_client/examples/all-in-one/jac.toml
```diff
@@ -72,3 +69,93 @@ typescript = "^5.3.3"
...
+Cross-Origin-Embedder-Policy = "unsafe-none"
+Access-Control-Allow-Origin = "*"
+Access-Control-Allow-Methods = "GET, POST, PUT, DELETE, OPTIONS"
+Access-Control-Allow-Headers = "Content-Type, Authorization"
+
+
+# Desktop target configuration (Tauri)
+[desktop]
```
MusabMahmoodh: PR title is "Add mobile target" but this block add a full `[desktop]` section too. If the example was missing desktop config, please split to separate PR. Mixing scope make it harder to review and bisect later if something break on desktop side.

jac-client/jac_client/examples/all-in-one/.gitignore
```diff
@@ -20,7 +20,7 @@ venv/
 node_modules/
 
 # Jac build artifacts
-.jac/
+.jac/*
```
MusabMahmoodh: Changed from `.jac/` (ignore whole directory) to `.jac/*` (ignore contents but not directory). Why? If the intent is to keep `.jac/` folder tracked as a placeholder, normal pattern is add a `.gitkeep` file inside. If not intentional, please revert. The diff does not explain the reason.

mobile/app/index.tsx
```diff
@@ -0,0 +1,275 @@
...
+// Configuration for URLs
+const getDevServerUrl = () => {
+  const localIP = DEV_CONFIG.localIP || '127.0.0.1';
+  return `http://${localIP}:5173`;
+};
+
+const getBackendApiUrl = () => {
+  if (__DEV__) {
+    if (Platform.OS === 'android') {
+      return 'http://10.0.2.2:9000';
```
MusabMahmoodh: `10.0.2.2:9000` is the Android emulator alias for host machine's localhost. Only work on emulator — on a real Android phone `10.0.2.2` is nothing and the app cannot reach backend. For iOS the code already use `DEV_CONFIG.localIP` (see line 41), but Android is hardcoded to emulator. Asymmetric. Please use `DEV_CONFIG.localIP` for Android also, and fall back to `10.0.2.2` only if IP detection fail.

mobile/app/index.tsx
```diff
@@ -0,0 +1,275 @@
...
+    }
+  }, []);
+
+  const prepareBundle = () => {
+    try {
+      if (!JAC_BUNDLE_HTML) {
+        throw new Error('Bundle not available');
+      }
+      // Inject the backend API URL into the bundle
+      const html = JAC_BUNDLE_HTML.replace(
```
MusabMahmoodh: This `replace` match exact string from `create-bundle.js`:

```
window.JAC_BACKEND_URL = window.JAC_BACKEND_URL || 'http://localhost:9000';
```

If `create-bundle.js` change the format later (add space, change quote style, rename variable), this replace silently fail and production backend URL stay as `localhost:9000`. Very fragile coupling between two files. Better approach: put an explicit marker in the bundle like `%%__JAC_BACKEND_URL__%%` and replace that only. Contract between two file become explicit.

mobile/app/index.tsx
```diff
@@ -0,0 +1,275 @@
...
+          }}
+          style={styles.webview}
+          javaScriptEnabled={true}
+          domStorageEnabled={true}
+          mixedContentMode="always"
+          onLoadEnd={handleLoadEnd}
+          onError={handleError}
+          onMessage={handleMessage}
+          allowFileAccess={true}
+          allowFileAccessFromFileURLs={true}
```
MusabMahmoodh: These three flags together — `allowFileAccess={true}` (line 154), `allowFileAccessFromFileURLs={true}` (line 155), `allowUniversalAccessFromFileURLs={true}` (line 156) — plus `originWhitelist={['*']}` (line 159) and `mixedContentMode="always"` (line 150) open up WebView to cross-origin file access and any origin, including http in a mixed-content page. Combined, they remove the sandbox that normally protect mobile WebView. If the bundled HTML ever load third-party script or user-controlled content, this become a real security hole. Please justify each flag individually or remove the ones not strictly needed.

jac-client/jac_client/plugin/src/targets/impl/mobile_target.impl.jac
```diff
@@ -0,0 +1,1800 @@
...
+    import socket;
+
+    try {
+        # Create a socket to determine the local IP
+        # We connect to an external address (doesn't actually connect)
+        s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM);
+        s.settimeout(0);
+        try {
+            # Connect to a public DNS server to find local IP
+            s.connect(('8.8.8.8', 80));
```
MusabMahmoodh: Reaching out to Google DNS `8.8.8.8:80` just to discover the local network IP is a very round-about trick. If the dev machine has no internet (offline dev, corporate firewall blocking 8.8.8.8, or behind a proxy), the socket throw and fallback is `127.0.0.1` which the phone cannot reach. Please use a local-only method like `socket.gethostbyname(socket.gethostname())`, or iterate interfaces with `netifaces`, or at least try multiple fallbacks.

jac-client/jac_client/plugin/src/targets/impl/mobile_target.impl.jac
```diff
@@ -0,0 +1,1800 @@
...
+        console.print(f"  {result.stdout.strip()}", style="success");
+    } except subprocess.TimeoutExpired {
+        raise RuntimeError("Bundle creation timed out");
+    } except Exception as e {
+        raise RuntimeError(f"Failed to create bundle: {e}");
+    }
+}
+
+"""Build mobile app for production."""
+def _build_mobile_app(mobile_dir: Path, platform: str) {
```
MusabMahmoodh: This whole function is dead code. It call `expo build:ios` / `expo build:android` (lines 1269, 1274 below) — both commands were removed from Expo CLI in SDK 46 (early 2022). PR target SDK 54. On the user's machine, Expo will print a deprecation error to stderr and return 0 doing nothing. But `subprocess.run(check=True)` only check returncode, so user see "Building iOS app..." message and think it work — no artifact produced anywhere. The PR already have `_build_with_eas()` function for modern builds, so please delete this entire `_build_mobile_app` function and call EAS only.

jac-client/jac_client/plugin/src/targets/impl/mobile_target.impl.jac
```diff
@@ -0,0 +1,1800 @@
...
+"""Start backend server."""
+def _start_backend_server(
+    entry_file: Path, project_dir: Path, port: int
+) -> subprocess.Popen {
+    # Start jac serve in background
+    try {
+        process = subprocess.Popen(
+            ["jac", "serve", str(entry_file), "--port", str(port)],
+            cwd=str(project_dir),
+            stdout=subprocess.PIPE,
```
MusabMahmoodh: `Popen` with `stdout=PIPE, stderr=PIPE` but no one read from those pipes. When the backend server write enough log output to fill the OS pipe buffer (about 64KB on Linux, smaller on Windows), the write block and the server process freeze. User see "backend server running" but no response ever come. Either use `subprocess.DEVNULL` for both if output is not needed, or use `PIPE` with a background thread that drain continuously.

jac-client/jac_client/plugin/src/targets/impl/mobile_target.impl.jac
```diff
@@ -0,0 +1,1800 @@
...
+    console.print(f"  Build type: {build_type}", style="muted");
+    console.print(f"  Platform: {platform}", style="muted");
+
+    # Check if EAS CLI is installed
+    eas_status = _check_eas_ready(mobile_dir);
+    if not eas_status['is_ready'] {
+        error_msg = eas_status['error_message'];
+
+        # Handle different error scenarios
+        if "EAS CLI not found" in error_msg {
```
MusabMahmoodh: Using text inside an error message to drive control flow here. If someone later reword the string in `_check_eas_ready`, this `if` branch silently fail and the EAS-login flow stop working with no test catch it. `_check_eas_ready` already return a structured dict, please add an `error_code` field like `{'is_ready': False, 'error_code': 'CLI_NOT_FOUND', 'error_message': "..."}` and switch on `error_code` instead.

jac-client/jac_client/plugin/cli.jac
```diff
@@ -354,7 +422,12 @@ def _handle_build_target(ctx: HookContext) -> None {
         }
 
         # Polymorphic call - each target handles its own build logic
-        bundle_path = target.build(entry_path, project_dir, platform);
+        # Mobile target accepts build_type parameter
+        if target_name == "mobile" {
```
MusabMahmoodh: The comment on line 424 above say "Polymorphic call — each target handles its own build logic", but this `if target_name == "mobile"` branch special-case mobile to pass an extra `build_type` argument. This break the polymorphism — the caller should not know which target it is. Two cleaner options:

1. Add `build_type` (default `None`) to the `build()` signature of all targets. Non-mobile targets simply ignore the argument.
2. Before calling, set `target.build_type = build_type` on the target instance, and have `MobileTarget.build()` read it from `self`.

Either way, the call site stays uniform.

Issue #5263, opened 2026-03-20
[Bug]: Saving even no update happens (Reopen)


amadolid:
### Operating System

Ubuntu

### Versions

Jac 0.12.2 | python 3.12.12 | jac-scale 0.2.6

### Steps to Reproduce

```jac
node Task {
    has id: int,
        title: str,
        completed: bool = False;
}

walker:pub CreateTask {
    has id: int,
        title: str;

    can create with Root entry {
        new_task = here ++> Task(id=self.id, title=self.title);
        report new_task;
    }
}

walker:pub GetAllTasks {
    can read with Root entry {
        tasks = [-->];
        report tasks;
    }
}

walker GetTask {
    has id: int;

    can get with Root entry {
        for task in [-->] {
            if (self.id == task.id) {
                report task;
            }
        }
    }
}

walker:pub CompleteTask {
    has id: int;

    can complete with Root entry {
        for task in [-->] {
            if (self.id == task.id) {
                visit task;
            }
        }
    }

    can mark_complete with Task entry {
        here.completed = True;
        report here;  # here is the visited task
    }
}

walker DeleteTask {
    has id: int;

    can visit_node with Root entry {
        for task in [-->] {
            if (self.id == task.id) {
                visit task;
            }
        }
    }

    can delete with Root entry {
        del here;
        report {"deleted": True, "id": self.id};
    }
}
```

### Expected vs Actual Behavior

No update should happen. Currently, it still updates the docs without any changes.
The added test only checks for "changes" on DB but it doesn't test if it calls the update or not.

I added some logs in redis and mongo `put` impl to visualize the trigger.

<img width="1493" height="549" alt="Image" src="https://github.com/user-attachments/assets/c26894a3-20f8-46a6-8248-9c12eec92071" />
<img width="1043" height="723" alt="Image" src="https://github.com/user-attachments/assets/d3804dff-d550-41de-bacb-fdd9c3c69d3c" />

GetTask still save data
<img width="755" height="270" alt="Image" src="https://github.com/user-attachments/assets/d6f0d74c-27b2-4c33-b384-2a1565c4ec37" />

Another note, since we use `__setattr__` to check for updates, it will not be able to catch update operations for tuple, list, set, dict and custom/nested `obj`.

Including here the concern for full data update in L2/L3 which will cause data discrepancy or race condition.

PR #5278, opened 2026-03-22
fix: prevent unnecessary DB writes on read-only operations (Redis L2 cache path)

akindu-k:
# PR: Hash-based dirty checking for anchor persistence

## Summary

Replaced the `__setattr__`/`is_updated` flag approach with hash-based change detection. Anchors are snapshot-hashed on load and compared at sync time. This fixes two problems in one:

1. **Read-only requests triggered unnecessary database writes** — deserialized anchors inherited `is_updated = True` from the class default, causing every loaded anchor to be re-written
2. **In-place mutations were silently lost** — `list.append()`, `dict[k]=v`, `set.add()`, and nested object mutations bypass `__setattr__`, so the `is_updated` flag was never set

## Related Issue

#5263

## How It Works

```
Load anchor from DB
  → _deserialize_anchor() computes anchor.hash = _compute_hash()

User code runs (any mutations happen in memory)

End of request → sync()
  → _compute_hash() == anchor.hash?  → skip (no I/O)
  → _compute_hash() != anchor.hash?  → put() to MongoDB, update hash
  → hash == 0 (new anchor)?          → always written
```

Single mechanism — no special cases for different mutation types.

## What Mutation Types Are Detected?

| Type | Example | Detected? |
|------|---------|-----------|
| Direct assignment | `task.title = "new"` | Yes |
| list | `.append()`, `.extend()`, `.pop()`, `[i]=x` | Yes |
| dict | `[k]=v`, `.update()`, `.pop()`, `del d[k]` | Yes |
| set | `.add()`, `.discard()`, `.update()` | Yes |
| Nested object | `task.config.timeout = 30` | Yes |
| Nested container | `task.config.items.append(x)` | Yes |

### jaclang (core)

| File | Change |
|------|--------|
| `archetype.impl.jac` | Removed `__setattr__` dirty-tracking logic (kept bare `object.__setattr__` call) |
| `runtime.impl.jac` | Removed explicit `is_updated = True` from `connect()`, `disconnect()`, `remove_edge()` — hash catches edge list changes |
| `serializer.jac` | Added `_compute_hash()` method declaration |
| `serializer.impl.jac` | Implemented `_compute_hash()` (serialize → JSON → `hash()`). Compute baseline hash in `_deserialize_anchor()` |

### jac-scale

| File | Change |
|------|--------|
| `memory_hierarchy.mongo.impl.jac` | `sync()`: replaced two-layer `is_updated` + hash check with hash-only comparison. `put()`: recomputes hash after write. `get()`/`query()`: removed `is_updated = False` workarounds |
| `memory_hierarchy.main.impl.jac` | `commit()`: removed `is_updated = True` on single-anchor path. L2 pending list uses hash-only comparison |
| `memory_hierarchy.redis.impl.jac` | `get()`: removed `is_updated = False` workaround — hash baseline handles it |

### Tests

| File | Change |
|------|--------|
| `fixtures/mutation_app.jac` | New fixture with `Item` node (list, dict, set fields) and walkers for in-place mutations |
| `test_memory_hierarchy.jac` | Replaced `is_updated` unit test with hash-based test covering direct assignment, `list.append`, `dict[k]=v`. Added `opcounters.update` assertion to read-only test. New integration test: in-place mutations persist across server restart |

### Release Notes

| File | Change |
|------|--------|
| `jaclang.md` | Updated entry: hash-based anchor dirty checking |
| `jac-scale.md` | Updated entry: hash-based dirty checking for MongoDB/Redis |

## Test Plan

- [x] `test "no unnecessary saves on read-only operations"` — zero MongoDB `update` commands during read-only requests (server-side `opcounters`)
- [x] `test "hash-based dirty check detects all mutation types"` — unit test: direct assignment, `list.append`, `dict[k]=v` all produce different hashes
- [x] `test "in-place mutations on list and dict are persisted via hash-based dirty check"` — integration test: `list.append` + `dict[k]=v` survive server restart
- [x] Manual verification with MongoDB profiling: zero writes on `GetAllTasks`, correct writes on `CompleteTask`/`AddTag`


akindu-k:
image

PR #5300, opened 2026-03-23
feat(jac-scale): Docker build pipeline with auto-Dockerfile, private registry, and cross-platform support

juzailmlwork:
Adds jac start app.jac --scale --build as a production deployment pathway that builds and pushes a Docker image to DockerHub before deploying to Kubernetes.

What's included:

Auto-generates a Dockerfile and .dockerignore if none exist (jac.local.toml excluded automatically)
Dockerfile installs Bun, runs jac install, and uses layer caching for jac.toml
Validates docker_username / docker_password are set before attempting a build
Creates a kubernetes.io/dockerconfigjson pull secret for private DockerHub repos (cleaned up on jac destroy)
build_platform config option (default: linux/amd64) for cross-platform builds from ARM machines
Sets correct K8s startup command (jac start <file>) when using --build
Tuned health check timing: 10s interval, 60 retries, aws_nlb_wait reduced to 10s
Updated plugin docs and release notes


jac-scale/jac_scale/providers/registry/dockerhub.jac
```diff
@@ -1,50 +1,131 @@
 """DockerHub registry implementation."""
+import docker;
```
udithishanka: this breaks the optional-dep guard. `_optdeps.docker` exists specifically so jac-scale can load without the `docker` SDK installed, and line 6 already aliases `docker_module as docker` on top of this import anyway so the line is also dead code. drop it.

```suggestion
```

jac-scale/jac_scale/providers/registry/dockerhub.jac
```diff
@@ -1,50 +1,131 @@
...
+            build_kwargs: dict = {
+                'path': code_folder,
+                'tag': repository_name,
+                'dockerfile': 'Dockerfile',
+                'rm': True,
+                'buildargs': {'TARGETPLATFORM': self.build_platform}
+            };
+            if self.build_platform {
+                build_kwargs['platform'] = self.build_platform;
+            }
```
udithishanka: two redundancies here:

1. `buildargs: {'TARGETPLATFORM': ...}` on L79 isn't needed — BuildKit already populates `TARGETPLATFORM` automatically when `--platform` is set. passing it as a build-arg doesn't reach stages that don't declare `ARG TARGETPLATFORM`, and for stages that do, BuildKit wins.
2. `if self.build_platform` is dead — `postinit` defaults it to `'linux/amd64'`, so it's never empty.

collapse to just:

```suggestion
            build_kwargs: dict = {
                'path': code_folder,
                'tag': repository_name,
                'dockerfile': 'Dockerfile',
                'rm': True,
                'platform': self.build_platform
            };
```

jac-scale/jac_scale/providers/registry/dockerhub.jac
```diff
@@ -1,50 +1,131 @@
...
+        if not self.docker_username {
+            raise ValueError(
+                "docker_username is not set. Add 'docker_username' under [plugins.scale.kubernetes] in jac.toml."
+            );
+        }
+        if not self.docker_password {
+            raise ValueError(
+                "docker_password is not set. Add 'docker_password' under [plugins.scale.kubernetes] in jac.toml."
+            );
+        }
```
udithishanka, line 33: validating credentials inside `build_image` means the error surfaces partway through `deploy()` — after namespace creation, env loading, etc. move this into `postinit` (or have `_handle_image_registry` check before creating the registry) so users get the error up-front.

jac-scale/jac_scale/tests/test_docker_images.jac
```diff
@@ -0,0 +1,395 @@
...
+        mock_core_v1.create_namespaced_secret.side_effect = capture_create_secret;
+        # Build target and invoke pull-secret creation path directly
+        k8s_config = KubernetesConfig(
+            app_name=app_name,
+            namespace=namespace,
+            docker_username=docker_username,
+            docker_password=docker_password,
+            build_platform="linux/amd64"
+        );
+        # Simulate what deploy() does when app_config.build=True
```
udithishanka: this test re-implements the secret-creation logic inline and then asserts on the re-implementation — it never actually invokes `deploy()` or any production method. if someone broke the real secret body (wrong `type`, wrong auth key, missing namespace), this test would still pass.

extract the secret-body construction in `kubernetes_target.jac:1281-1298` into a standalone method (e.g. `_build_registry_pull_secret(namespace, username, password)`) and call that here, or drive `deploy()` end-to-end with the K8s client fully mocked. same structural issue applies to test 4 (L356).

jac-scale/jac_scale/tests/test_docker_images.jac
```diff
@@ -0,0 +1,395 @@
...
+        sec_ctx = container_cfg.get("securityContext", {});
+        cap_add = sec_ctx.get("capabilities", {}).get("add", []);
+        assert len(cap_add) > 0 , (
+            "needs_root=True must grant runtime capabilities (CHOWN, DAC_OVERRIDE, ...)"
+        );
+        # Verify imagePullSecrets is wired into the pod spec template
+        pod_spec_image_pull_secrets = [{"name": pull_secret_name}];
+        assert pod_spec_image_pull_secrets[0]["name"] == f"{app_name}-registry-secret";
+    }
+}
```
udithishanka, line 281: test 4 has the same problem as test 3: it builds `build_command` and asserts on it inline (L384-388) and only exercises `_build_container_config` for the needs_root path — nothing here proves that `deploy()` actually assembles `imagePullSecrets` into the pod spec. the `pod_spec_image_pull_secrets` assertion at the end is asserting on a literal the test itself constructed. drive `deploy()` or extract the pod-spec build into a method you can call.

docs/docs/reference/plugins/jac-scale.md
```diff
@@ -1682,13 +1682,40 @@ graph TD
...
-**Production mode** requires Docker credentials in `.env`:
+**Production mode** requires Docker credentials in `jac.toml`:
 
-```env
-DOCKER_USERNAME=your-dockerhub-username
-DOCKER_PASSWORD=your-dockerhub-password-or-token
+```toml
+[plugins.scale.kubernetes]
+docker_username = "your-dockerhub-username"
+docker_password = "your-dockerhub-token"
```
udithishanka, line 2499: putting `docker_password` in `jac.toml` is a footgun — `jac.toml` is normally committed. the PR itself auto-adds `jac.local.toml` and `.env` to `.dockerignore`, which tells me the intended location is `jac.local.toml`. point users there (or document env-var interpolation) instead of the committed file.

```suggestion
**Production mode** requires Docker credentials in `jac.local.toml` (not `jac.toml` — it's typically committed to git):

```toml
[plugins.scale.kubernetes]
docker_username = "your-dockerhub-username"
docker_password = "your-dockerhub-token"
```

jac-scale/jac_scale/targets/kubernetes/kubernetes_config.jac
```diff
@@ -31,7 +32,7 @@ obj KubernetesConfig(BaseConfig) {
         pvc_size: str = '5Gi',
         # Timing configuration
         resource_deletion_wait: int = 5,
-        aws_nlb_wait: int = 60,
+        aws_nlb_wait: int = 10,
```
udithishanka: the field default is now `10` but L176 (`from_dict`) still falls back to `60`. a config loaded from toml that doesn't set the key will get `60` via `from_dict` and `10` via direct construction — the two paths disagree. sync the fallback.

jac-scale/jac_scale/targets/kubernetes/kubernetes_config.jac
```diff
@@ -7,6 +7,7 @@ obj KubernetesConfig(BaseConfig) {
     has docker_image_name: str = '',
         docker_username: str = '',
         docker_password: str = '',
+        build_platform: str = 'linux/amd64',
```
udithishanka, line 9: `build_platform` is documented as a `[plugins.scale.kubernetes]` toml key but it's not registered in `plugin_config.jac` (the schema) or `impl/config_loader.impl.jac` (the defaults). add it to both so it round-trips cleanly through the plugin config loader.

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
```diff
@@ -1270,6 +1271,56 @@ obj KubernetesTarget(DeploymentTarget) {
                 }
             }
 
+            # 3c. Create Docker registry pull secret for private DockerHub images
+            pull_secret_name: str | None = None;
+            if app_config.build
+            and self.k8s_config.docker_username
+            and self.k8s_config.docker_password {
+                import json;
+                import base64;
```
udithishanka: nit: move these imports to the top of the file.

docs/docs/community/release_notes/unreleased/jac-scale/5300.feature.md
```diff
@@ -0,0 +1,4 @@
+- **Docker Build Pathway (`--build`)**: `jac start app.jac --scale --build` now builds and pushes a Docker image to DockerHub before deploying. Auto-generates a `Dockerfile` (with Bun, `jac install`, and layer-cached `jac.toml`) and `.dockerignore` (excludes `jac.local.toml`) if not present. Sets the correct K8s startup command automatically.
```
udithishanka: release note says `.dockerignore` excludes `jac.local.toml` — the code also appends `.env` and `.jac/` (dockerhub.jac L45). worth mentioning, especially `.env` since users may have workflows that rely on it being copied into the image.

udithishanka (changes requested):
A handful of things to address: an optional-dep regression, tests that don't cover the real code path, docs that point creds at a git-tracked file, and a couple of config gaps.

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
juzailmlwork: fixed

jac-scale/jac_scale/targets/kubernetes/kubernetes_config.jac
juzailmlwork: this is fixed

docs/docs/community/release_notes/unreleased/jac-scale/5300.feature.md
juzailmlwork: a detailed description is added

jac-scale/jac_scale/targets/kubernetes/kubernetes_config.jac
juzailmlwork: this is synchronized to 10

docs/docs/reference/plugins/jac-scale.md
juzailmlwork: this is updated

jac-scale/jac_scale/providers/registry/dockerhub.jac
juzailmlwork: this is moved up now

jac-scale/jac_scale/providers/registry/dockerhub.jac
juzailmlwork: this is fixed

jac-scale/jac_scale/tests/test_docker_images.jac
juzailmlwork: this is fixed

jac-scale/jac_scale/tests/test_docker_images.jac
juzailmlwork: this is fixed

jac-scale/jac_scale/providers/registry/dockerhub.jac
juzailmlwork: this is resolved

jac-scale/jac_scale/tests/test_docker_images.jac
```diff
@@ -0,0 +1,300 @@
...
+        );
+        assert len(captured_deployments) == 1;
+        pod_spec = captured_deployments[0]["spec"]["template"]["spec"];
+        pull_secrets = pod_spec.get("imagePullSecrets", []);
+        assert len(pull_secrets) == 1 , "Pod spec must contain exactly one imagePullSecret";
+        assert pull_secrets[0]["name"] == pull_secret_name , (
+            f"Expected '{pull_secret_name}', got '{pull_secrets[0]['name']}'"
+        );
+    }
+}
```
udithishanka, line 281: this is only half-fixed. the first half (calling _build_container_config and asserting on command + securityContext) is real, that's fine. the second half at L256-298 still has the exact problem from the original review.

the test builds a deployment_body dict literal at L273-288, hands it to a mock at L289-291, captures it back via side_effect, and asserts on it at L292-298. those final assertions are just checking that python dicts work and that mocks capture call args. nothing actually invokes deploy(), and the real pod-spec assembly at kubernetes_target.jac:1428-1430 (the imagePullSecrets list keyed off pull_secret_name being non-None) is never exercised. if someone breaks that conditional, mistypes the key, drops the wrapping list, or forgets to wire pull_secret_name through, this test still passes.

either extract the deployment-body construction at kubernetes_target.jac:1409-1437 into a _build_deployment_body helper and call it from the test (same pattern as _build_registry_pull_secret in test 3), or drive deploy() end-to-end with AppsV1Api and CoreV1Api mocked and assert on what create_namespaced_deployment was actually called with.

udithishanka (changes requested):
test 4 still half-fixed.

jac-scale/jac_scale/tests/test_docker_images.jac
juzailmlwork: this is fixed in last commit

jac-scale/jac_scale/targets/kubernetes/kubernetes_config.jac
```diff
@@ -40,7 +41,7 @@ obj KubernetesConfig(BaseConfig) {
         mongodb_storage_size: str = '1Gi',
         # Timing configuration
         resource_deletion_wait: int = 5,
-        aws_nlb_wait: int = 60,
+        aws_nlb_wait: int = 10,
```
udithishanka: aws_nlb_wait dropped from 60 to 10. real AWS NLBs typically take 60-90s to assign a hostname, so on first deploy this hits the nlb_url=None branch in _wait_for_deployment, skips check_deployment_status entirely, and the deploy reports success without ever verifying the app is healthy. either revert to 60, or restructure _wait_for_deployment to poll read_namespaced_service until the LB has an ingress or hits a deadline.

docs/docs/community/release_notes/unreleased/jac-scale/5300.feature.md
```diff
@@ -0,0 +1,4 @@
+- **Docker Build Pathway (`--build`)**: `jac start app.jac --scale --build` now builds and pushes a Docker image to DockerHub before deploying. Auto-generates a `Dockerfile` (with Bun, `jac install`, and layer-cached `jac.toml`) and `.dockerignore` if not present. The auto-generated `.dockerignore` excludes `jac.local.toml`, `jac.*.toml`, `.env`, and `.jac/` - note that `.env` is intentionally excluded to prevent secrets from being baked into the image; use `[plugins.scale.secrets]` in `jac.toml` for runtime env vars instead. Sets the correct K8s startup command automatically.
+- **Private DockerHub Support**: Automatically creates and manages a K8s image pull secret for private DockerHub images. Secret is updated on re-deploy and deleted on `jac destroy`.
+- **Cross-Platform Builds**: Added `build_platform` config (default: `linux/amd64`) so ARM machines (Apple Silicon) produce images compatible with x86_64 K8s nodes. Configurable via `[plugins.scale.kubernetes]` in `jac.toml`.
+- **Health Check Tuning**: Interval reduced from 15s to 10s, max retries increased to 60, `aws_nlb_wait` reduced to 10s.
```
udithishanka: the 'interval reduced from 15s to 10s, max retries increased to 60' parts aren't in this PR's diff. those constants live in check_deployment_status (kubernetes_utils.jac), which isn't modified here. drop them from the bullet or include the actual change.

jac-scale/jac_scale/providers/registry/dockerhub.jac
```diff
@@ -21,30 +36,91 @@ obj DockerHubRegistry(ImageRegistry) {
...
         repository_name = f"{self.docker_username}/{image_name}";
 
+        # Generate a Dockerfile from template if one doesn't exist in the code folder
+        dockerfile_path = os.path.join(code_folder, 'Dockerfile');
+        if not os.path.exists(dockerfile_path) {
+            template_path = os.path.join(
+                os.path.dirname(__file__),
+                '../../targets/kubernetes/templates/base.Dockerfile'
+            );
+            shutil.copy(template_path, dockerfile_path);
```
udithishanka: shutil.copy writes a Dockerfile straight into the user's source tree, and the .dockerignore append at L61 does the same. the microservices auto-build pipeline in _image_build.jac already solved this by putting generated artifacts under .jac/Dockerfile.gen so they don't leak into git. worth aligning with that pattern, or at minimum logging clearly that the file is landing in the repo.

jac-scale/jac_scale/providers/registry/dockerhub.jac
```diff
@@ -21,30 +36,91 @@ obj DockerHubRegistry(ImageRegistry) {
...
             raise RuntimeError(f"Failed to build Docker image: {e}") from e;
         }
+        console.print(f"Docker image '{repository_name}' built successfully.");
 
         return repository_name;
     }
 
     def push_image(image_name: str) {
         # Push Docker image to DockerHub
         docker_client = docker.from_env();
```
udithishanka: nit: docker.from_env() also runs at L70. each call sets up a fresh httpx client over the unix socket. cache it as a has field set in postinit.

udithishanka (changes requested):
earlier round looks resolved (helper extractions, postinit creds, docs, build_platform schema). a couple of new ones:

jac-scale/jac_scale/targets/kubernetes/kubernetes_config.jac
juzailmlwork: this is fixed by polling with a deadline of 120s

docs/docs/community/release_notes/unreleased/jac-scale/5300.feature.md
juzailmlwork: this is removed

jac-scale/jac_scale/providers/registry/dockerhub.jac
juzailmlwork: Dockerfile generation: Previously shutil.copy(template, code_folder/Dockerfile)  dirtied the source tree. Now copies to code_folder/.jac/Dockerfile.gen, which is gitignored. The absolute path is passed as dockerfile= so Docker reads it outside the build context, even though .jac/ is excluded by the .dockerignore. This is the same pattern _image_build.jac uses with -f <absolute_path>.

.dockerignore: Previously appended to the user's file on every run —silently mutating their repo. Now writes the file only if it doesn't exist, using open(..., 'w') with the full needed content (including jac.local.toml, .env, .jac/). If the user already has one, it's untouched; the console message is gone (their file is their responsibility).

jac-scale/jac_scale/providers/registry/dockerhub.jac
juzailmlwork: docker.from_env() is now called once in postinit and stored as self._client. Both build_image and push_image use the cached clien

jac-scale/jac_scale/providers/registry/dockerhub.jac
```diff
@@ -1,16 +1,31 @@
 """DockerHub registry implementation."""
 import from jac_scale._optdeps.docker { docker_module as docker, APIError, BuildError }
 import from jac_scale.abstractions.image_registry { ImageRegistry }
+import from jaclang.cli.console { console }
```
udithishanka, line 10: missing imports. the file uses os.path.join/exists/abspath/dirname, os.makedirs and shutil.copy (L43-65), plus the Any annotation on L12, but none of os, shutil, or Any are imported. jac check reports 22 errors against this file on the head sha, and jac run dies with "name 'os' is not defined" the moment build_image is invoked. add:

```suggestion
import os;
import shutil;
import from typing { Any }
import from jac_scale._optdeps.docker { docker_module as docker, APIError, BuildError }
import from jac_scale.abstractions.image_registry { ImageRegistry }
import from jaclang.cli.console { console }
```

jac-scale/jac_scale/tests/test_docker_images.jac
```diff
@@ -0,0 +1,280 @@
+"""Integration tests for Docker image build/push support in jac-scale.
```
udithishanka, line 1: this whole file is never run by ci. .github/workflows/test-jaseci.yml enumerates each pytest invocation (L203-260) and test_docker_images.jac isn't in the list, so the test-scale job passes without exercising any of these tests. that's how the import bug in dockerhub.jac and the failures in tests 1/2/4 below all slipped past. add a step that runs this file (and please confirm it passes locally with the existing dockerhub.jac fixes before pushing).

jac-scale/jac_scale/tests/test_docker_images.jac
```diff
@@ -0,0 +1,280 @@
...
+# Exercises the full DockerHub registry pipeline:
+#   build_image (Dockerfile generation, .dockerignore update, docker build)
+#   → push_image (docker login + push with streaming chunks)
+#   → get_image_url (returns the final qualified name)
+#
+# All Docker daemon calls are mocked; filesystem side-effects are real.
+# ---------------------------------------------------------------------------
+test "end to end build and push produces correct image url and side effects" {
+    with tempfile.TemporaryDirectory() as tmpdir {
+        with unittest.mock.patch("docker.from_env") as mock_docker {
```
udithishanka: this patch target is wrong. production code reaches docker via the optdep import in dockerhub.jac L2 (docker_module aliased as docker) and calls docker.from_env(). patching the python docker package's from_env doesn't intercept that lookup, so once the file is actually run, postinit will still invoke the real docker daemon (or fail on import if docker isn't installed). patch jac_scale._optdeps.docker.docker_module.from_env (or whatever name the optdep stub exposes) instead. same fix needed on L149.

jac-scale/jac_scale/tests/test_docker_images.jac
```diff
@@ -0,0 +1,280 @@
...
+                "build_platform": "linux/amd64"
+            };
+            registry = DockerHubRegistry(config=config);
+            # --- execute the full pipeline ---
+            final_url = registry.build_and_push(tmpdir, image_name="myservice:v1");
+            # Verify final URL
+            assert final_url == "integuser/myservice:v1" , f"Unexpected URL: {final_url}";
+            # Dockerfile must have been generated from template (no Dockerfile existed)
+            dockerfile_path = os.path.join(tmpdir, "Dockerfile");
+            assert os.path.exists(dockerfile_path) , "Dockerfile should have been generated";
```
udithishanka: this assertion is stale. the code now writes to code_folder/.jac/Dockerfile.gen (dockerhub.jac L46-55), not code_folder/Dockerfile. the test still asserts at tmpdir/Dockerfile and will fail the moment it actually runs.

```suggestion
            dockerfile_path = os.path.join(tmpdir, ".jac", "Dockerfile.gen");
            assert os.path.exists(dockerfile_path) , "Dockerfile.gen should have been generated under .jac/";
```

jac-scale/jac_scale/tests/test_docker_images.jac
```diff
@@ -0,0 +1,280 @@
...
+    original = KubernetesConfig(
+        app_name="platform-test",
+        docker_username="platformuser",
+        docker_password="platformpass",
+        build_platform="linux/arm64"
+    );
+    config_dict = original.to_dict();
+    assert config_dict["build_platform"] == "linux/arm64";
+
+    restored = KubernetesConfig.from_dict(config_dict);
```
udithishanka: wrong call signature. KubernetesConfig.from_dict is declared as a static method taking (cls, config), so it needs the class passed explicitly as the first arg. as written this throws TypeError: missing 1 required positional argument: 'config'.

```suggestion
    restored = KubernetesConfig.from_dict(KubernetesConfig, config_dict);
```

jac-scale/jac_scale/tests/test_docker_images.jac
```diff
@@ -0,0 +1,280 @@
...
+        docker_username="user",
+        docker_password="pass"
+    );
+    target = KubernetesTarget(config=k8s_config);
+
+    # --- verify _build_container_config for the build=True path ---
+    probe_configs = {
+        "readiness": {"httpGet": {"path": "/docs", "port": 8000}},
+        "liveness": {"httpGet": {"path": "/docs", "port": 8000}}
+    };
```
udithishanka, line 236: probe_configs is missing the startup key. _build_container_config unconditionally reads probe_configs['startup'] (kubernetes_target.jac:426), so this raises KeyError before any of the assertions run.

```suggestion
    probe_configs = {
        "readiness": {"httpGet": {"path": "/docs", "port": 8000}},
        "liveness": {"httpGet": {"path": "/docs", "port": 8000}},
        "startup": {"httpGet": {"path": "/docs", "port": 8000}}
    };
```

docs/docs/community/release_notes/unreleased/jac-scale/5300.feature.md
```diff
@@ -0,0 +1,4 @@
+- **Docker Build Pathway (`--build`)**: `jac start app.jac --scale --build` now builds and pushes a Docker image to DockerHub before deploying. Auto-generates a `Dockerfile` (with Bun, `jac install`, and layer-cached `jac.toml`) and `.dockerignore` if not present. The auto-generated `.dockerignore` excludes `jac.local.toml`, `jac.*.toml`, `.env`, and `.jac/` - note that `.env` is intentionally excluded to prevent secrets from being baked into the image; use `[plugins.scale.secrets]` in `jac.toml` for runtime env vars instead. Sets the correct K8s startup command automatically.
```
udithishanka: this no longer matches the code. release note says "auto-generates a Dockerfile" (implying top-level in the project), but dockerhub.jac now writes the generated file to .jac/Dockerfile.gen. same drift in the plugin docs at docs/docs/reference/plugins/jac-scale.md:2095 ("If no Dockerfile exists in your project folder, one is automatically generated"). either reword both to mention the .jac/Dockerfile.gen location, or revert the generation path back to the project root if that was the intended behavior.

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
```diff
@@ -1338,12 +1455,19 @@ obj KubernetesTarget(DeploymentTarget) {
...
+            }
             container_config = self._build_container_config(
                 app_name,
                 repository_name,
                 probe_configs,
                 resources_config,
-                secret_ref_name=secret_ref_name
+                command=build_command,
+                secret_ref_name=secret_ref_name,
+                needs_root=app_config.build
```
udithishanka: this contradicts the security comment in _build_container_config. that comment says needs_root exists specifically because the no-build path runs apt-get and pip at container start; pre-built images should be "fully locked down". with --build the image is pre-built (no runtime apt/pip), so it should take the non-root, read-only-rootfs, dropped-caps branch. passing needs_root=app_config.build inverts the intent and grants CHOWN/DAC_OVERRIDE/FOWNER/FSETID/SETUID/SETGID + writable rootfs to a container that doesn't need any of it.

```suggestion
                command=build_command,
                secret_ref_name=secret_ref_name
```

jac-scale/jac_scale/targets/kubernetes/templates/base.Dockerfile
```diff
@@ -14,7 +14,18 @@ ENV PYTHONUNBUFFERED=1 \
...
     rm -rf /var/lib/apt/lists/*
 
-RUN pip install --no-cache-dir jaclang
+RUN pip install --no-cache-dir jaclang jac-scale[all] jac-client jac-super
+
+# Install Bun (required for jac install npm dependencies)
+RUN curl -fsSL https://bun.sh/install | bash
+ENV PATH="/root/.bun/bin:$PATH"
+
+COPY jac.toml* ./
```
udithishanka: COPY with a glob requires at least one match or buildkit errors out ("no source files were specified"). the very next line guards on jac.toml existence with a shell test, implying the file might not exist. if it doesn't, the build dies here before the conditional ever runs. either drop the glob and rely on the conditional, or restructure so an absent jac.toml is a no-op:

```suggestion
COPY . /tmp/ctx/
RUN if [ -f /tmp/ctx/jac.toml ]; then cp /tmp/ctx/jac.toml ./ && jac install; fi
```

or more simply, just defer the jac install to after the full COPY at L31 and drop this layer cache trick until it actually works for the no-jac.toml case.

udithishanka (changes requested):
ci passes but never exercises the new code. a few real issues + docs/code drift.

jac-scale/jac_scale/providers/registry/dockerhub.jac
juzailmlwork: this is fixed

jac-scale/jac_scale/tests/test_docker_images.jac
juzailmlwork: this is fixed

jac-scale/jac_scale/tests/test_docker_images.jac
juzailmlwork: this is fixed

jac-scale/jac_scale/tests/test_docker_images.jac
juzailmlwork: this is fixed

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac

jac-scale/jac_scale/targets/kubernetes/templates/base.Dockerfile
juzailmlwork: this is fixed

jac-scale/jac_scale/tests/test_docker_images.jac
juzailmlwork: this is fixed

jac-scale/jac_scale/tests/test_docker_images.jac
juzailmlwork: a pipelin is added for this

docs/docs/community/release_notes/unreleased/jac-scale/5300.feature.md
juzailmlwork: updated the content to new version

udithishanka (approved):

kugesan1105:
@juzailmlwork 
works fine and the tests pass, but a few things to be sorted before it goes in.

🔴 **The Dockerfile/`.dockerignore` generation duplicates an existing module.** The inline logic in `DockerHubRegistry.build_image` is basically a re-implementation of `microservices/images/build_helpers.jac`, which already has `resolve_dockerfile()` and `ensure_dockerignore()` doing the same write-if-absent thing. It landed in #5948 and is already on this branch's merge-base, so we'd end up with two build paths that don't agree — they already write different `.dockerignore` contents to the same `.jac/Dockerfile.gen` spot, and that gap will only widen. Could we route through `build_helpers` instead? Bonus: it'd actually *shrink* this diff, since ~50 lines of inline logic collapse into a couple of calls.

🟠 **"Cross-platform" is a bit overstated.** Passing a single `platform=` to the docker-py SDK only targets one arch, and the old non-BuildKit builder tends to ignore it anyway — so it's not really multi-arch. Either we lean on `buildx --platform`, or we call it "configurable target platform" so an ARM user doesn't trust the label and ship something broken.

🟠 **`aws_nlb_wait` quietly changed from a 60s one-shot wait to a 10s poll interval.** The polling itself is a nice fix, but reusing the same key changes behavior for anyone who'd tuned it, and `120` is a magic number. Probably cleaner as new keys like `nlb_poll_interval` / `nlb_timeout`.

🟡 **Couple of small ones:** `docker.from_env()` in `postinit` ties construction to a running daemon (even when you just want `get_image_url()`), and `jac-scale[all]` in the template is heavier than the `[deploy,data]` we use elsewhere.

marsninja:

PR #5301, opened 2026-03-23
fix: widen TypeVar to union on conflicting type inference

SandeepaHWP:
## Summary

When the same TypeVar appears in multiple parameters and the arguments have incompatible types, the type checker now **widens to a union** instead of giving up.

### Before
```jac
def merge[T](a: T, b: T) -> list[T] { return [a, b]; }

merge(1, "hello");
# Step 1: T = int (from arg 1)
# Step 2: T = str (from arg "hello"), but int != str → return False → Unknown
```

### After
```jac
merge(1, "hello");
# Step 1: T = int (from arg 1)
# Step 2: T = str (from arg "hello"), but int != str → widen → T = int | str
# Result: list[int | str]
```

### How it works

Three cases when a TypeVar already has a binding and a new argument arrives:

1. **New fits in existing** → keep existing (e.g., `merge(1, True)` → T stays `int` because `bool` is a subtype of `int`)
2. **Existing fits in new** → use the new wider type (e.g., `merge(True, 1)` → T becomes `int`)
3. **Neither fits** → combine into a union (e.g., `merge(1, "hello")` → T becomes `int | str`)

The old code returned `False` in case 3. The fix replaces that with creating a `UnionType`.


SandeepaHWP:
Fixed in #5272

PR #5305, opened 2026-03-23
fix: skip implicit return None error for functions with ellipsis-only bodies

SandeepaHWP:
## Summary

- Ellipsis-only function bodies (`...;`) are now recognized as stubs by the type checker
- The E1004 diagnostic ("may implicitly return None") is suppressed for these stubs, matching Pyright's behavior for Python
- Functions with real statements alongside `...;` still correctly trigger E1004

## Problem

Previously, stub functions like this would incorrectly produce E1004:

```jac
class A {
    def method(self: A, x: "A") -> "A" {
        ...;  # E1004: Function 'method' declared return type A but may implicitly return None
    }
}
```

In Python, Pyright treats `...`-only bodies as stubs via `isSuiteEmpty()` and skips the return type check. Jac was missing this logic.

## Fix

Added `_is_ellipsis_only_body()` to `TypeCheckPass` that checks if a function body contains only `Ellipsis` and `Semi` nodes. This is checked alongside the existing `is_overload` and `is_generator` skip conditions.

**Stub (no error):**
```jac
def greet(name: str) -> str {
    ...;  # ✓ treated as stub, no E1004
}
```

**Real implementation (still errors correctly):**
```jac
def greet(name: str) -> str {
    print(name);
    ...;  # ✗ E1004: has real statements, not a stub
}
```

### DIFF
--- a/jac/jaclang/compiler/passes/main/type_checker_pass.jac
+++ b/jac/jaclang/compiler/passes/main/type_checker_pass.jac
@@ -77,5 +77,4 @@ class TypeCheckPass(UniPass) {
     def _body_always_returns(
         self: TypeCheckPass, body: list[uni.CodeBlockStmt]
     ) -> bool;
-
 }
--- a/jac/jaclang/compiler/type_system/type_evaluator.impl/evaluator_util_methods.impl.jac
+++ b/jac/jaclang/compiler/type_system/type_evaluator.impl/evaluator_util_methods.impl.jac
@@ -1335,9 +1335,7 @@ impl TypeEvaluator._get_outer_scope_type_params(
 }
 
 """Check if a function body contains only ellipsis (stub body)."""
-impl TypeEvaluator._is_ellipsis_only_body(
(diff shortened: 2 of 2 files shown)


jac/jaclang/compiler/passes/main/impl/type_checker_pass.impl.jac
```diff
@@ -661,6 +663,24 @@ impl TypeCheckPass._body_always_returns(
...
+        }
+        if isinstance(stmt, uni.ExprStmt) and isinstance(stmt.expr, uni.Ellipsis) {
+            has_ellipsis = True;
+            continue;
+        }
+        return False;
+    }
+    return has_ellipsis;
+}
+
```
kugesan1105: move this to `type_evaluator`

SandeepaHWP: Done

PR #5327, opened 2026-03-24
Fix RedisInsight cache-dashboard not loading in K8s

### LINKED ISSUE #5324: [bug] redis inight dashboard is not getting connected automatically to jaseci redis deployment
### Description
RedisInsight dashboard fails to connect to the Jaseci Redis deployment in Kubernetes
When deploying a Jac application to Kubernetes with the Redis dashboard enabled, the RedisInsight dashboard at /cache-dashboard loads but cannot connect to the Redis instance. Three issues contribute to this:
Redis authentication mismatch — The Redis server only recognizes a default user, but the dashboard tries to authenticate with a custom username that doesn't exist.
RedisInsight auto-configuration not working — RedisInsight v3 requires explicit EULA acceptance and an encryption key to function in a container environment. Without these, the pre-seeded database connection is never created.
Nginx ingress instability — Uncompressed asset proxying overwhelms the ingress pod, causing health check timeouts and pod restarts.

kashmithnisakya:
## Summary

Fixes the RedisInsight **cache-dashboard not loading** on Kubernetes. The dashboard was unreachable/unusable because it couldn't authenticate to Redis, the RedisInsight container needed manual EULA/keyring setup, and it wasn't served correctly behind the ingress sub-path.

## Changes

- **Redis authentication** (`kubernetes_redis.jac`, `redis.conf.template`): Redis ACL users are defined in `redis.conf` (via a `{redis_acl_rules}` template placeholder) instead of `--requirepass`; username/password come from a Kubernetes Secret. This is what lets RedisInsight actually connect to Redis.
- **RedisInsight auto-config** (`kubernetes_redis.jac`):
  - `RI_ACCEPT_TERMS_AND_CONDITIONS=true` — auto-accepts the EULA so the dashboard is usable immediately
  - `RI_ENCRYPTION_KEY` — explicit key, since the container has no OS keyring
  - `RI_REDIS_HOST`/`RI_REDIS_PORT` pre-seeded; `RI_REDIS_USERNAME`/`RI_REDIS_PASSWORD` from the Redis Secret (never inlined)
  - `RIPROXYPATH=/cache-dashboard` so it serves correctly under the ingress sub-path
- **Ingress** (`ingress.jac`): the RedisInsight ingress (basic-auth + `sub_filter` asset rewrites + `rewrite-target`), server-level gzip via the nginx ConfigMap (`use-gzip: true`), `proxy-buffering: on`, and improved HTML handling so the dashboard loads through the proxy.

## Tests

- `test_deploy_k8s.jac`: compact, cluster-free unit test for the RedisInsight cache-dashboard spec (EULA auto-accept, encryption key, `RIPROXYPATH`, secret-backed credentials, container hardening); the existing integration test also now asserts `RI_REDIS_USERNAME`/`RI_REDIS_PASSWORD` come from the Secret.

## Screenshots

Screen Shot 2026-03-25 at 12 34 17 AM


jac-scale/jac_scale/impl/serve.core.impl.jac
```diff
@@ -134,23 +134,20 @@ impl JacAPIServerCore.postinit -> None {
...
-        metrics_handler = self._create_admin_metrics_handler(
-            self._metrics.get_endpoint_handler()
-        );
+        # Register /metrics endpoint (public)
         self.server.add_endpoint(
             JEndPoint(
                 method=HTTPMethod.GET,
                 path=metrics_endpoint,
-                callback=metrics_handler,
+                callback=self._metrics.get_endpoint_handler(),
```
udithishanka, line 142: this removes auth from `/metrics` entirely — the prior PR explicitly added Bearer+Basic auth so Prometheus could scrape it, and the test in `test_deploy_k8s.jac` was flipped from 403→200 to match. metrics leak request patterns, user counts, endpoint usage, memory/GC stats — not stuff we want anonymous. the PR body doesn't explain why this was removed.

if Prometheus scraping really needs to be anonymous (e.g. in-cluster scrape), gate it on a config flag or restrict to in-cluster source IPs via the ingress — don't default-open it.

jac-scale/jac_scale/admin/impl/admin_portal.impl.jac
```diff
@@ -988,16 +1052,12 @@ impl JacAPIServerAdmin.admin_graph_data(
...
 impl JacAPIServerAdmin.admin_get_metrics(
     Authorization: str | None = None
 ) -> TransportResponse {
     import re;
     import from prometheus_client { generate_latest, REGISTRY }
-    error = self.require_admin(Authorization);
-    if error {
-        return error;
-    }
     if not self._metrics or not self._metrics.is_enabled() {
```
udithishanka, line 1061: `require_admin()` was removed here and the handler at L422 no longer passes the `Authorization` header. a route under `/admin/` that bypasses admin auth is a footgun — either keep the admin check and add a separate `/metrics` route for Prometheus, or rename this out of `/admin/`. right now the docstring says "public endpoint" while the path says "admin", which is a trap for anyone reading the code later.

jac-scale/jac_scale/admin/impl/grafana_manager.impl.jac
```diff
@@ -0,0 +1,484 @@
...
+
+    # Use configured password or auto-generate
+    password = self._monitoring_password;
+    if not password {
+        password = generate_random_password(16);
+        self._monitoring_password = password;
+    }
+
+    # Create user in Grafana with Admin role
+    created = self._create_grafana_user(self._monitoring_user, password, 'Admin');
```
udithishanka, line 231: docstring on L199 and the release notes both say "Viewer role / dashboard-only access", but this hardcodes `'Admin'`. the monitoring user ends up with full Grafana admin, including the ability to change other users' passwords and reset the admin. either make it Viewer as advertised, or update the docs + release note to say Admin and justify why.

```suggestion
    created = self._create_grafana_user(self._monitoring_user, password, 'Viewer');
```

jac-scale/jac_scale/admin/impl/grafana_manager.impl.jac
```diff
@@ -0,0 +1,484 @@
...
+    }
+
+    if success {
+        self._admin_password = new_password;
+        # Persist encrypted password
+        try {
+            import jwt;
+            jwt_cfg = get_scale_config().get_jwt_config();
+            jwt_secret2: str = str(jwt_cfg.get('secret', ''));
+            encrypted = jwt.encode(
```
udithishanka, line 173: `jwt.encode` with HS256 is a signature, not encryption — the payload is base64-encoded and readable by anyone who can read the Mongo doc. calling the stored field `encrypted_password` and the helper `_decrypt_persisted_admin_password` misrepresents what's happening.

if you need to persist the password: use real encryption (e.g. `cryptography.fernet` with a key derived from `jwt_secret`, or AES-GCM). better option: don't persist the admin password at all — only persist a flag that a sync is pending, and read the current password from the admin user record on demand.

jac-scale/jac_scale/impl/serve.core.impl.jac
```diff
@@ -1084,52 +1082,44 @@ impl JacAPIServerCore._setup_scheduler -> None {
...
+    grafana_mgr = GrafanaManager.instance();
+    grafana_mgr.setup(
+        admin_password=admin_cfg.get('default_password', 'changeme'),
+        monitoring_user=monitoring_cfg.get('monitoring_username', 'jac-monitor'),
+        monitoring_password=monitoring_cfg.get('monitoring_password', 'monitoring123'),
+        grafana_url=grafana_url
+    );
+
+    # Wait for Grafana to be ready (up to 60 seconds)
+    for attempt in range(12) {
```
udithishanka, line 1111: this blocks `start()` for up to 60s (12 × 5s) waiting for Grafana when monitoring is enabled. every app pod startup now pays that cost if Grafana isn't ready yet — which is common during cluster-wide rollouts where everything comes up together. run this in a background task (`asyncio.create_task` or a thread) so the app doesn't wait on a dashboard service to boot.

jac-scale/jac_scale/impl/config_loader.impl.jac
```diff
@@ -99,8 +99,9 @@ impl JacScaleConfig.get_default_config(self: JacScaleConfig) -> dict[str, Any] {
                 5.0,
                 10.0
             ],
-            'prometheus_admin_password': 'Adminpassword123',
-            'k8s_metrics_enabled': True
+            'k8s_metrics_enabled': True,
+            'monitoring_username': 'jac-monitor',
+            'monitoring_password': 'monitoring123'
```
udithishanka, line 104: `monitoring123` is the same class of weak default as the `Adminpassword123` this PR is replacing — if the operator doesn't override, it ships to prod. either auto-generate a random password on first deploy (GrafanaManager already has `generate_random_password`) and surface it via `/admin/grafana/credentials`, or fail-fast in `_setup_grafana` when the default is detected outside dev.

udithishanka (changes requested):
6 issues — auth regression on /metrics is the big one.

jac-scale/jac_scale/providers/database/kubernetes_redis.jac
```diff
@@ -94,6 +96,14 @@ obj KubernetesRedisProvider(DatabaseProvider) {
             redis_keyspace_events_line = '';
         }
 
+        # Configure ACL rules — the 'default' user gets the password,
+        # and if a custom username is requested we create an additional ACL user.
+        acl_lines: list[str] = [f"user default on >{redis_password} ~* &* +@all"];
```
udithishanka: the redis password is interpolated into the ACL line here, and then this whole conf string gets written into a regular ConfigMap on L214. before this PR the password only lived in the redis Secret, read at runtime via --requirepass $REDIS_PASSWORD. now anyone with configmaps:get in the namespace can dump the conf and read the admin password. ConfigMaps aren't encrypted at rest by default and tend to have looser RBAC than Secrets in practice.

fix: use redis's aclfile directive in redis.conf pointing to /etc/redis/users.acl, and mount that file from a Secret (the existing {app_name}-redis-secret will do) instead of inlining the ACL into the ConfigMap.

jac-scale/jac_scale/providers/database/kubernetes_redis.jac
```diff
@@ -94,6 +96,14 @@ obj KubernetesRedisProvider(DatabaseProvider) {
...
+        if redis_username != 'default' {
+            acl_lines.append(f"user {redis_username} on >{redis_password} ~* &* +@all");
```
udithishanka: no validation that redis_password or redis_username is free of newlines. a password containing \n would inject extra ACL lines here, e.g. a new unauthenticated user. defense in depth, the aclfile-from-Secret approach from the other comment also closes this since a Secret-mounted file gets written byte-for-byte.

jac-scale/jac_scale/providers/database/kubernetes_redis.jac
```diff
@@ -407,7 +414,11 @@ obj KubernetesRedisProvider(DatabaseProvider) {
             redis_username = k8s_config.get('redis_username', 'admin');
             redis_password = k8s_config.get('redis_password', 'password');
             service_name = f"{self.app_name}-redis-service";
-            self.connection_string = f"redis://{redis_username}:{redis_password}@{service_name}:6379/0";
+            if redis_username == 'default' {
+                self.connection_string = f"redis://:{redis_password}@{service_name}:6379/0";
```
udithishanka: deploy() at L187 builds the connection string as redis://{user}:{pw}@host:6379/0 unconditionally, but this branch drops the username when it's 'default'. same provider, same config, two different URLs depending on which call path hits first. either both paths special-case default, or neither does (redis://default:pw@host is valid and accepted by redis).

udithishanka (changes requested):
3 issues. Password ending up in the ConfigMap is the main one.

udithishanka (approved):

PR #5347, opened 2026-03-26
feat(jac-scale): add configurable uvicorn workers support

kugesan1105:
## Summary
- Add configurable Uvicorn workers for local and Kubernetes deployments
- Support integer values or `"auto"` (resolves to `2 * CPU cores + 1`)
- Configure via `[plugins.scale.server]` in `jac.toml` or CLI (`--workers` / `-w`)
- **Default: 1 worker** (safe for development)

## Configuration

**jac.toml:**
```toml
[plugins.scale.server]
workers = 4        # explicit count
# workers = "auto" # CPU-based: 2 * CPU + 1
```
## CLI override:

```bash
jac start app.jac --workers 4
jac start app.jac -w auto
```
image

-------------------

## verification

-------------------------

**1. local**

image

-------------------------------------

**2. k8s**
image

### DIFF
--- a/jac-scale/jac_scale/jserver/impl/jfast_api.impl.jac
+++ b/jac-scale/jac_scale/jserver/impl/jfast_api.impl.jac
@@ -119,25 +119,21 @@ impl JFastApiServer.run_server(
                     style="warning"
                 );
             }
-            if workers > 1 {
-                # Pre-import to compile and cache before workers spawn
-                import jac_scale.jserver._worker_app_holder;
-                # Set env vars for worker factory to recreate the app
-                import os;
-                os.environ["JAC_WORKER_MODULE"] = module_name;
-                os.environ["JAC_WORKER_BASE_PATH"] = base_path or os.getcwd();
-                # Factory mode: each worker creates its own app instance
-                uvicorn.run(
-                    "jac_scale.jserver._worker_app_holder:create_app",
-                    factory=True,
-                    host=host,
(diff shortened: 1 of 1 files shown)


kugesan1105:
`_worker_app_holder.py`:  Uvicorn with multiple workers requires an import string, not an app object.

### The problem:
```
# This FAILS with workers > 1
uvicorn.run(app, workers=4)  # ❌ Can't serialize app across processes


# This WORKS
uvicorn.run("module:factory", factory=True, workers=4)  # ✅ Import string
```

bcz When workers > 1, uvicorn spawns separate Python processes and each process must:

1. Import a module (via import string)
2. Call a factory function to create its own app instance
we couldnot able to can't pass a Python object between processes - uvicorn needs a "recipe" (import string) to recreate the app in each worker.


```
Main Process                          Worker Processes
     │                                      │
     │  Sets env vars:                      │
     │  JAC_WORKER_MODULE=app.jac           │
     │  JAC_WORKER_BASE_PATH=/path          │
     │                                      │
     └──► uvicorn.run("...:create_app")     │
                    │                       │
                    └──────► spawn ────────►│
                                            │
                              Worker 1: create_app() → FastAPI
                              Worker 2: create_app() → FastAPI
                              Worker 3: create_app() → FastAPI
```

TLDR: It's the standard pattern used by Django, Flask, and every Python framework that supports multiple workers.

PR #5353, opened 2026-03-27
refactor(jac-scale): decouple endpoint registration and use import string for uvicorn

kugesan1105:
## Summary
- Extract `_setup_endpoints(dev)` method from `start()` for reusable endpoint registration
- Add `create_app` factory function for uvicorn import string pattern
- Modify `run_server` to use `uvicorn.run("...:create_app", factory=True, ...)` 
- Update log config to use class path string for `ConsoleLogHandler`

## Problem
The `start()` method tightly coupled endpoint registration with server startup, making it impossible to register endpoints separately for multi-worker mode (where each worker needs its own server instance).

Additionally, `run_server` passed the app object directly to uvicorn, which doesn't work with multi-worker mode (uvicorn can't serialize app objects across processes).

## Why This Solution
The import string pattern is uvicorn's recommended production approach. It enables:
- Multi-worker support via `workers=N` parameter (future PR)
- Proper process isolation - each worker creates its own server instance
- Clean separation between server configuration and endpoint registration

### DIFF
--- a/docs/docs/community/release_notes/jac-client.md
+++ b/docs/docs/community/release_notes/jac-client.md
@@ -2,13 +2,15 @@
 
 This document provides a summary of new features, improvements, and bug fixes in each version of **Jac-Client**. For details on changes that might require updates to your existing code, please refer to the [Breaking Changes](../breaking-changes.md) page.
 
-## jac-client 0.3.9 (Unreleased)
+## jac-client 0.3.10 (Unreleased)
+
+## jac-client 0.3.9 (Latest Release)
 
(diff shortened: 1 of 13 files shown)


jac-scale/jac_scale/impl/serve.core.impl.jac
```diff
@@ -255,6 +264,15 @@ impl JacAPIServerCore.start(
...
     }
+    # Note: Endpoints are NOT registered here. They are registered in create_app()
+    # factory function, which is called by uvicorn. This avoids creating two
+    # JacAPIServer instances with duplicate metrics registration.
+
+    # Set environment variables for worker processes (factory will use these)
+    import os;
+    os.environ["JAC_WORKER_MODULE"] = self.module_name;
+    os.environ["JAC_WORKER_BASE_PATH"] = self.base_path;
+    os.environ["JAC_WORKER_DEV_MODE"] = "1" if dev else "0";
```
Thamirawaran, line 275: why we need to save it to env variables

PR #5354, opened 2026-03-27
fix: resolve 'by postinit' symbol so go-to-definition works and W2001 is suppressed

SandeepaHWP:
## Problem

Any field declared with `by postinit` gets a false warning:

```jac
obj Foo {
    has x: str by postinit;  # W2001: Name 'postinit' may be undefined
    def postinit -> None;
}
```

Clicking on `postinit` in the field declaration also does nothing - go-to-definition is broken.

## Root cause

The parser creates a `Name` node for `postinit` as part of the field declaration, but nothing ever links it to the actual `postinit` ability. The symbol table stores the ability under `"__post_init__"` (the Python name), so a normal name lookup for `"postinit"` silently finds nothing. The name node's `.sym` stays `None`, which triggers the W2001 warning and breaks go-to-definition.

## Fix

At the end of `DeclImplMatchPass.transform` - after all symbols are fully built - look up `"__post_init__"` directly in the archetype's symbol table and assign it to the `Name` node. This also covers `postinit` defined in a separate `.impl.jac` file.

**File changed:** `jac/jaclang/jac0core/passes/impl/decl_impl_match_pass.impl.jac`

## Tests

Both tests were written first and confirmed to **fail** before the fix, then **pass** after.

- `test_decl_impl_match_pass.jac` - checks that the `Name` node's `.sym` resolves to the `postinit` ability.
- `test_static_analysis_pass.jac` - checks that no W2001 warning fires for a `by postinit` field.
- Fixture: `fixtures/static_analysis/by_postinit_sym.jac`

### DIFF
--- a/jac/tests/langserve/test_server.jac
+++ b/jac/tests/langserve/test_server.jac
@@ -475,15 +475,15 @@ test "test_go_to_definition_by_postinit" {
             passes_main_fixture_abs_path("static_analysis/by_postinit_sym.jac")
         );
         lsp.type_check_file(file_uri);
-        # Foo: inline 
+        # Foo: inline
         result = lsp.get_definition(file_uri, lspt.Position(1, 20));
         assert result is not None , "Foo: expected go-to-def for 'postinit', got None";
         assert "by_postinit_sym.jac:2:" in str(result) , f"Foo: expected line 3 (0-indexed 2), got: {result}";
-        # Bar: stub+impl file 
+        # Bar: stub+impl file
         result = lsp.get_definition(file_uri, lspt.Position(8, 20));
         assert result is not None , "Bar: expected go-to-def for 'postinit', got None";
         assert "by_postinit_sym.jac:9:" in str(result) , f"Bar: expected line 10 (0-indexed 9), got: {result}";
-        # Baz: impl block 
+        # Baz: impl block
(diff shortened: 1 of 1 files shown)


jac/jaclang/compiler/passes/main/impl/type_checker_pass.impl.jac
```diff
@@ -681,6 +681,18 @@ impl TypeCheckPass.exit_param_var(self: TypeCheckPass, nd: uni.ParamVar) -> None
 """Set type on class fields (has vars) and validate initializer type."""
 impl TypeCheckPass.exit_has_var(self: TypeCheckPass, nd: uni.HasVar) -> None {
     import from jaclang.compiler.type_system { types }
+    if (nd.defer and nd.kid and isinstance(nd.kid[-1], uni.Name)) {
+        postinit_nd: uni.Name = nd.kid[-1];
+        if postinit_nd.sym is None {
+            arch = nd.parent.parent if isinstance(nd.parent, uni.ArchHas) else None;
+            if isinstance(arch, uni.Archetype) {
```
kugesan1105, line 648: @SandeepaHWP  this will miss the follwing case

main.jac
```jac
obj Person;
```

main.impl.jac
```jac
impl obj{
       has age:int by postinit;
}
```

PR #5387, opened 2026-03-30
feat(serializer): add `ref_mode` for shared and circular object graphs

akindu-k:
## Summary

Adds a `ref_mode` option to `Serializer.serialize()` that emits `{"$ref": "<uuid-hex>", "$type": "<anchor-type>"}` placeholders for any Jac object encountered more than once during serialization, instead of inlining it again. This prevents **data duplication** from shared references and **infinite recursion** on circular graphs.

On the deserialization side, `$ref` placeholders are automatically resolved back to live archetypes (via context memory lookup) or to typed anchor stubs when the context is unavailable.

## Changes

| File | What changed |
|------|-------------|
| `jac/jaclang/runtimelib/serializer.jac` | Added `ref_mode` and `_seen` parameters to `serialize`, `_serialize_value`, `_serialize_attrs`. Added new method declarations: `_is_jac_ref`, `_deserialize_jac_ref`. Imported `WalkerAnchor` and `ObjectAnchor`. |
| `jac/jaclang/runtimelib/impl/serializer.impl.jac` | Implemented the ref-tracking logic in `_serialize_value` (Archetype branch). Implemented `_is_jac_ref` and `_deserialize_jac_ref`. Hooked `$ref` detection into `_deserialize_value`. |
| `jac-scale/jac_scale/tests/test_serializer_social.jac` | Added 14 new tests covering ref_mode serialization, deserialization, circular graphs, mutation stability, and lifecycle. |
| `docs/docs/community/release_notes/jaclang.md` | Added release note entry. |

### Serialization (`ref_mode=True`)

```
Serializer.serialize(obj, ref_mode=True)
│
├── Creates a fresh `_seen: set()` to track visited object UUIDs
│
└── Calls _serialize_value(obj, ..., _seen)
    │
    ├── Primitives / UUID / Enum → returned as-is
    │
    ├── list / tuple → recurse each element, passing _seen through
    │
    ├── dict → recurse each value, passing _seen through
    │
    ├── Permission / Access → recurse nested values with _seen
    │
    ├── Anchor (NodeAnchor, EdgeAnchor) → serialize anchor fields,
    │   recurse archetype/edges with _seen
    │
    ├── ★ Archetype (the core ref_mode logic):
    │   │
    │   ├── 1. Get the object's __jac__ anchor and its UUID
    │   │
    │   ├── 2. If UUID is already in _seen:
    │   │      → Return {"$ref": "<uuid-hex>", "$type": "node|edge|walker|object"}
    │   │      (compact back-reference, stops recursion)
    │   │
    │   ├── 3. If UUID is NOT in _seen:
    │   │      → Add UUID to _seen
    │   │      → Register anchor in context memory via Jac.get_context().mem.put()
    │   │        (enables later $ref resolution during deserialization)
    │   │      → Continue with full inline serialization of attributes
    │   │
    │   └── 4. Serialize all public attributes via _serialize_attrs(..., _seen)
    │
    └── Generic __dict__ objects → recurse with _seen
```

### Deserialization (`$ref` resolution)

```
Serializer._deserialize_value(val)
│
├── If val is a dict with "$ref" key:
│   │
│   └── Calls _deserialize_jac_ref(val)
│       │
│       ├── 1. Extract UUID from "$ref" field
│       │
│       ├── 2. Try context memory lookup:
│       │      Jac.get_context().mem.get(UUID(id_str))
│       │      │
│       │      ├── Found & populated → return anchor.archetype (live object)
│       │      └── Found but not populated → call anchor.populate(), return archetype
│       │
│       └── 3. Fallback (no context or anchor not in memory):
│              Read "$type" field → select correct anchor class:
│                "node"   → NodeAnchor
│                "edge"   → EdgeAnchor
│                "walker" → WalkerAnchor
│                default  → ObjectAnchor
│              → Create a lazy stub via _id_to_stub(anchor_cls, id_str)
│
├── If val is a dict with "__type__" key → Serializer.deserialize(val)
│
└── Otherwise → recurse key/value pairs
```

### Key Design Decisions

1. **First-occurrence-wins**: The first time an object is seen, it is fully inlined with all its attributes. Only subsequent encounters emit a `$ref`. This means the serialized output is always self-describing for at least one occurrence of every object.

2. **`_seen` is a flat `set` of UUIDs**: Simple O(1) lookup. The set is created fresh per `serialize()` call, so there's no cross-call state leakage.

3. **`$type` annotation on `$ref`**: Each ref carries a type hint (`"node"`, `"edge"`, `"walker"`, `"object"`) so the deserializer can create the correct anchor stub class without needing to probe the database.

4. **Context memory registration during serialization**: When an object is first seen, its anchor is `put()` into context memory. This ensures that `_deserialize_jac_ref` can resolve `$ref` back to the live archetype (not just a lazy stub) within the same runtime session.

5. **Graceful fallback**: If context memory is unavailable or the anchor isn't found, deserialization falls back to creating a typed anchor stub. This means `$ref` round-trips work even in disconnected/offline scenarios — the stub just won't have populated fields until it's explicitly loaded.

## Example

```python
# Circular graph
city = City(name="Metropolis")
alice = Person(name="Alice", city=city)
bob = Person(name="Bob", city=city, friend=alice)
alice.friend = bob

# Without ref_mode → infinite recursion (alice→bob→alice→bob→...)
# With ref_mode:
data = Serializer.serialize(alice, ref_mode=True)
```

Produces:

```json
{
  "name": "Alice",
  "city": {"name": "Metropolis"},
  "friend": {
    "name": "Bob",
    "city": {"$ref": "<city-uuid>", "$type": "object"},
    "friend": {"$ref": "<alice-uuid>", "$type": "object"}
  }
}
```

- `city` is fully inlined on first encounter (inside Alice), then `$ref`'d inside Bob.
- `alice` is `$ref`'d inside Bob's `friend` field, breaking the cycle.

image


jac/jaclang/runtimelib/impl/serializer.impl.jac
```diff
@@ -131,6 +180,46 @@ impl Serializer._serialize_value(
     return str(val);
 }
 
+"""Check if a value is a $ref placeholder dict."""
+impl Serializer._is_jac_ref(val: object) -> bool {
+    return isinstance(val, dict) and '$ref' in val;
+}
+
```
Thamirawaran: this is not used anywhere.

Thamirawaran (changes requested):
Looks fine. look at the usage of the code snippet. and remove it if it is unnecessary

Thamirawaran (approved):

PR #5391, opened 2026-03-30
Add email field to user registration endpoint

kashmithnisakya:
## Description:
Adds a required `email` field to `POST /user/register`, validated via pydantic `EmailStr`. Email is stored in MongoDB alongside user data for future use. Also renames the internal `__system__` user to system.

### DIFF
--- a/jac-scale/jac_scale/tests/scale_test_client.jac
+++ b/jac-scale/jac_scale/tests/scale_test_client.jac
@@ -191,9 +191,9 @@ obj ScaleTestClient {
     }
 
     """Register a user and return (username, token) tuple."""
-    def register_user(username: str, password: str) -> tuple[str, str] {
+    def register_user(username: str, password: str, email: str = "test@test.com") -> tuple[str, str] {
         response = self.post(
-            "/user/register", json={"username": username, "password": password}
+            "/user/register", json={"username": username, "password": password, "email": email}
         );
         data = response.json();
         if data.get("ok") and data.get("data") {
--- a/jac-scale/jac_scale/tests/test_admin.jac
+++ b/jac-scale/jac_scale/tests/test_admin.jac
@@ -75,7 +75,7 @@ test "admin login non-admin user rejected" {
 
         client.post(
             "/user/register",
-            json={"username": f"nonmin_{uuid.uuid4().hex[:8]}", "password": "testpass"}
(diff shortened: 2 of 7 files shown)


docs/docs/community/release_notes/jac-scale.md
```diff
@@ -4,6 +4,9 @@ This document provides a summary of new features, improvements, and bug fixes in
 
 ## jac-scale 0.2.12 (Unreleased)
 
+- Added `email` field to `POST /user/register` with `EmailStr` validation.
+- Renamed internal `__system__` user to `system`.
```
MusabMahmoodh, line 8: why is this change???

jac-scale/jac_scale/impl/serve.core.impl.jac
```diff
@@ -439,6 +439,14 @@ impl JacAPIServerCore.register_create_user_endpoint -> None {
                     default=None,
                     description='Password for new user',
                     type=ParameterType.BODY
+                ),
+                APIParameter(
+                    name='email',
+                    data_type='email',
+                    required=True,
```
MusabMahmoodh, line 446: so both usernamae and email is required?? thats not a good XP

MusabMahmoodh (approved):

jac-scale/jac_scale/impl/serve.core.impl.jac
kashmithnisakya: Yeah, we can’t make it optional at the API level.

But I made it optional at the user manager level. If we make email optional, we’ll still need a way to send the password reset token.

docs/docs/community/release_notes/jac-scale.md
kashmithnisakya: All other pre-creation users are added like this (guest, admin), so that’s why I applied the same change here.

jac-scale/jac_scale/impl/serve.core.impl.jac
MusabMahmoodh: username can be optional no??

kashmithnisakya: No, we can’t. We use the username as a unique identifier, so we can’t create multiple users with the same username. However, multiple users can have the same email, so the username can’t be optional.

PR #5405, opened 2026-03-31
fix: resolve JWT auth timing window by moving validation into dedicated middleware

akindu-k:
## Problem

There was a race condition / timing window in the authentication flow:

1. `request_context_middleware` created the Jac context with `user_root = system_root`
2. `call_next(request)` then ran the endpoint, which called `set_user_root()` to switch to the real user
3. Any code that executed between context creation and `set_user_root()` ran as `system_root` — shared by all users
4. Per-endpoint JWT parsing was also duplicated across `create_walker_callback` and `create_function_callback`, with no central rejection point for bad tokens

## Solution

Introduce a dedicated `jwt_validation_middleware` that runs **before** `request_context_middleware`:

1. Extracts the JWT from the `Authorization: Bearer <token>` header
2. Validates the token via `user_manager.validate_jwt_token()`
3. Looks up `root_id` via `user_manager.get_root_id(username)`
4. Sets `request.state.user_root_id` and `request.state.username` before `call_next`
5. Rejects invalid/expired/forged tokens with `401` immediately — before the context is even created

`request_context_middleware` then reads `user_root_id` from `request.state` and calls `ctx.set_user_root()` at context creation time, so the context is **never** in a `system_root` state for an authenticated request.

Starlette middleware ordering is handled correctly: `request_context_middleware` is registered first, then `jwt_validation_middleware` — since Starlette inserts at index 0, JWT ends up outermost and runs first.

### `jfast_api.jac` / `jfast_api.impl.jac`
- Added `_user_manager` field and `set_user_manager()` to `JFastApiServer` (raises `RuntimeError` on `None` to prevent silent misconfiguration)
- Added `jwt_validation_middleware` — handles token extraction, validation, `root_id` lookup, and early 401 rejection
- Updated `request_context_middleware` to read `user_root_id` from `request.state` and set it on the context immediately at creation

### `serve.endpoints.impl.jac`
- Removed duplicated inline JWT parsing from `create_walker_callback` and `create_function_callback`
- Auth check in endpoints is now a single line reading `request.state.username` populated by the middleware

### `user_manager.impl.jac`
- Added a `ContextVar`-based per-request cache for `get_root_id()` to avoid redundant storage lookups when the middleware and endpoint both need the root ID within the same request

### `tests/test_serve.jac`
- Added Group 8: JWT middleware tests covering:
  - Per-user data isolation (two users cannot see each other's nodes)
  - Token rejection: forged secret, expired token, malformed JWT
  - Unauthenticated requests to auth-required endpoints return `401`
  - Anonymous requests to public endpoints still return `200`
 
image


jac-scale/jac_scale/user_manager.jac
```diff
@@ -19,7 +20,10 @@ glob _jwt_config = get_scale_config().get_jwt_config(),
      JWT_ALGORITHM = _jwt_config['algorithm'],
      JWT_EXP_DELTA_DAYS = _jwt_config['exp_delta_days'],
      SSO_HOST = _sso_config['host'],
-     logger = logging.getLogger(__name__);
+     logger = logging.getLogger(__name__),
+     _root_id_request_cache: ContextVar[dict[str, str] | None] = ContextVar(
+         '_jac_scale_root_id_cache', default=None
+     );
```
kugesan1105, line 30: why this is a `dict`? 
per-request there will be only one root_id??

akindu-k: in future in case the admin handler can see multiple users graphs, dict is more suitable

PR #5464, opened 2026-04-07
fix: add explicit UTF-8 encoding to open() calls in compiler and CLI

MusabMahmoodh:
On Windows, Python defaults to cp1252 encoding. Without explicit encoding="utf-8", reading/writing .jac files with non-ASCII characters causes UnicodeDecodeError/UnicodeEncodeError.

Changes:
- cfg_build_pass.impl.jac: encoding="utf-8" on file read
- core.impl.jac (na_ir_gen): encoding="utf-8" on file read
- project.impl.jac: encoding="utf-8" on toml/template/gitignore write
- tools.impl.jac: encoding="utf-8" on config read/write


kugesan1105:
good one!
but cleaner way is to Centralize file I/O , Instead of repeating `encoding="utf-8"` everywhere:
```jac
def write_text_file(path: Path, content: str){
    with open(path, "w", encoding="utf-8") as f{
        f.write(content);
   }
}
```

same as our **read_file_with_encoding** in `jac/jaclang/jac0core/helpers.jac`

PR #5466, opened 2026-04-07
fix: add __init__.py with Jac imports for dirs that have __init__.jac

MusabMahmoodh:
Converts jac_client from namespace package to proper Python package. Dirs with __init__.jac get __init__.py that re-exports symbols. Matches jac_scale's packaging for PyInstaller compatibility.


marsninja:
@MusabMahmoodh  I'd rather fork pyinstaller and then fix it to go through python's support for meta importer than merge this. Lets look into the complexity and perhaps we can make a pr against pyinstaller. Otherwise we can fork.

PR #5473, opened 2026-04-07
fix: resolve streaming fallback infinite loop in ModelPool

Hirudika4:
**Fallback strategy improvements:**

* Added a custom implementation for the "fallback" strategy in `model_call_with_stream` that iterates through models and calls `Router._completion()` with `fallbacks=[]` to prevent a recursive infinite loop in `Router.stream_with_fallbacks()`.
* Updated the method documentation to reflect the workaround for the fallback strategy and clarify the use of `Router._completion()` per model.


jac-byllm/byllm/llm.impl/modelpool.impl.jac
```diff
@@ -123,6 +127,40 @@ impl ModelPool.model_call_with_stream(params: dict) -> Generator[str, None, None
...
+                        messages=messages,
+                        stream=True,
+                        fallbacks=[],
+                        **call_params
+                    );
+                    for chunk in response {
+                        yield chunk;
+                    }
+                    return;
+                } except Exception {
```
udithishanka: Bare `except Exception` with no logging makes production debugging a nightmare. If model A fails silently and model B succeeds, you'd never know A had an issue.

```suggestion
                } except Exception as exc {
                    logger.warning(f"ModelPool streaming fallback: {name} failed — {exc}");
                    if idx >= len(model_names) - 1 {
                        raise;
                    }
```

jac-byllm/byllm/llm.impl/modelpool.impl.jac
```diff
@@ -123,6 +127,40 @@ impl ModelPool.model_call_with_stream(params: dict) -> Generator[str, None, None
...
+                    call_params.pop("model", None);
+                    messages: list = call_params.pop("messages", []);
+                    response = router._completion(
+                        model=name,
+                        messages=messages,
+                        stream=True,
+                        fallbacks=[],
+                        **call_params
+                    );
+                    for chunk in response {
```
udithishanka, line 153: If model A yields 10 chunks then throws mid-stream, this catches the error and falls through to model B which starts fresh. The caller gets half of response A glued to all of response B — corrupted output that's really hard to debug.

Two options: (1) buffer chunks per model and only yield after full success (loses streaming benefit), or (2) if any chunks were already yielded, re-raise instead of falling back. Option 2 is probably the right call here — a partial response + error is better than a silently corrupted one.

jac-byllm/byllm/llm.impl/modelpool.impl.jac
```diff
@@ -123,6 +127,40 @@ impl ModelPool.model_call_with_stream(params: dict) -> Generator[str, None, None
 
         }
     }
+    # Fallback strategy: call Router._completion() per model with fallbacks=[]
+    # to prevent the recursive stream_with_fallbacks infinite loop.
+    if self.strategy == "fallback" {
+        model_names: list = ["pool-primary"];
```
udithishanka: This alias construction (`"pool-primary"` / `"pool-fallback-{i}"`) is duplicated from `postinit` (lines 28-31). If naming changes in one place, the other silently breaks. Store the ordered list during init — e.g. `self._fallback_model_names` — and reference it here.

jac-byllm/byllm/llm.impl/modelpool.impl.jac
```diff
@@ -123,6 +127,40 @@ impl ModelPool.model_call_with_stream(params: dict) -> Generator[str, None, None
...
+                    }
+                    return;
+                } except Exception {
+                    if idx >= len(model_names) - 1 {
+                        raise;
+                    }
+                }
+            }
+        }
+        return _fallback_gen();
```
udithishanka, line 171: No test covers this new streaming fallback path. The existing `model_pool_fallback` fixture only exercises non-streaming (`by pool()` → `dispatch_no_streaming`). Need a test that mocks `_completion()` to fail on the primary model and succeed on the fallback, with `stream=True`, to verify fallback ordering and prevent regression of the infinite loop.

jac-byllm/byllm/llm.impl/modelpool.impl.jac
Hirudika4: Applied your suggestion. logger.warning with the model name and exception message so we can trace which model failed and why without having to reproduce the issue.

jac-byllm/byllm/llm.impl/modelpool.impl.jac
Hirudika4: This is a valid concern.

Went with option 2. Added a yielded flag per model attempt. If any chunk has already been streamed to the caller, we immediately re-raise the error instead of falling back. This guarantees the caller receives either a complete response or a clear error, never a mixed/corrupted output.

If the failure happens before the first chunk (e.g., Gemini quota exhaustion, where the API rejects before streaming starts), yielded remains False, and we safely fall through to the next model.

jac-byllm/byllm/llm.impl/modelpool.impl.jac
Hirudika4: Agreed, the duplication was a maintenance trap. Added _fallback_model_names as a field on ModelPool, populated in postinit directly as the aliases are registered with the Router. model_call_with_stream now reads from it. Single source of truth, naming changes in one place can't silently break the other.

jac-byllm/byllm/llm.impl/modelpool.impl.jac
Hirudika4: Added model_pool_streaming_fallback.jac fixture and test. Mocked router._completion to fail on the primary (before streaming) and return a chunked stream on the fallback. Assertions verify correct ordering, fallback takeover, and clean output (no partial mix). Test completion also guards against the infinite loop regression.

jac-byllm/tests/fixtures/model_pool_streaming_fallback.jac
```diff
@@ -0,0 +1,57 @@
...
+    return types.SimpleNamespace(choices=[choice]);
+}
+
+glob _models_tried: list = [];
+
+def mock_completion(model: str, messages: list, stream: bool,
+                    fallbacks: list, **_kwargs: object) -> object {
+    _models_tried.append(model);
+    if model == "pool-primary" {
+        raise RateLimitError(
```
udithishanka, line 47: This only tests the pre-stream failure path — primary raises before yielding anything, fallback takes over cleanly. But the `yielded` guard in `_fallback_gen` (the most important safety check in this PR) is never exercised.

Need a second test where the mock primary yields a couple of chunks *then* raises. That test should assert the exception propagates instead of silently falling back to model B. Without it, someone could delete the `yielded` check and all tests still pass.

Something like:
```python
def mock_midstream_completion(model, messages, stream, fallbacks, **kw):
    _models_tried.append(model)
    if model == "pool-primary":
        def _failing_gen():
            yield make_chunk("partial")
            raise RateLimitError(...)
        return _failing_gen()
    # fallback should never be reached
    def _gen():
        yield make_chunk("should_not_appear")
    return _gen()
```
Then assert it raises and `"pool-fallback-1"` is NOT in `_models_tried`.

Hirudika4: Added Scenario B to the fixture: mock_midstream_completion yields "partial_chunk" then raises RateLimitError. The entry block collects chunks mid-iteration and catches the exception.

jac-byllm/byllm/llm.impl/modelpool.impl.jac
Jayanaka-98: Nice catch @udithishanka & @Hirudika4

PR #5474, opened 2026-04-07
Integrate MCP client and tool support to Byllm

Thamirawaran:


jac-byllm/byllm/impl/mcp.impl.jac
```diff
@@ -0,0 +1,284 @@
...
+        }
+    }
+}
+
+# Module-level dict mapping McpClient id -> _McpState.
+# This avoids adding internal state fields to the public McpClient obj.
+glob _states: dict = {};
+
+def _get_state(client: McpClient) -> _McpState {
+    cid = id(client);
```
Jayanaka-98: Minor Issue:
`id()` can be reused after a client is garbage-collected, causing a new McpClient to silently inherit stale connection state from a dead one. Use a `uuid` instead.

jac-byllm/byllm/impl/mtir.impl.jac
```diff
@@ -34,14 +34,20 @@ impl MTRuntime.factory(mtir: MTIR) -> MTRuntime {
...
-        tools = [
-            Tool(func, info=ir_info.tools[i] if i < len(ir_info.tools) else None)
-            for (i, func) in enumerate(call_params.get("tools", []))
-        ];  # type: ignore
-    } else {
-        # Python library mode: create tools without MTIR info
-        tools = [Tool(func) for func in call_params.get("tools", [])];  # type: ignore
+    raw_tools = call_params.get("tools", []);
+    tools = [];
+    for (i, func) in enumerate(raw_tools) {
```
Jayanaka-98: `i` is the raw index over all tools including MCP pass-throughs. If MCP tools appear before local functions (e.g. `tools=[*mcp.get_tools(), local_fn]`), `ir_info.tools[i]` is misaligned — local_fn picks up the wrong semantic info. Track a separate counter for local tools only.

jac-byllm/tests/test_byllm.jac
Jayanaka-98, line 1: The `ImportError` --> `McpError` path in `connect()` is untested. A simple test that mocks the import failure would catch any future refactor that breaks the install hint message.

jac-byllm/byllm/impl/mcp.impl.jac
```diff
@@ -0,0 +1,284 @@
...
+        if self.command {
+            transport_type = "stdio";
+        } elif self.url {
+            transport_type = "streamable-http";
+        } else {
+            raise McpError(
+                message="McpClient requires either 'command' (stdio) or 'url' (HTTP)."
+            );
+        }
+    }
```
Jayanaka-98, line 207: `transport="auto"` always picks `streamable-http` for any URL. There's no reliable way to distinguish SSE from HTTP by URL alone. When connection fails, an error message should hint: "If this is an SSE server, pass transport='sse' explicitly." or something like that.

jac-byllm/byllm/impl/mcp.impl.jac
Jayanaka-98, line 1: If two by llm() calls share the same McpClient and fire concurrently, both post to the same ClientSession simultaneously. So responses can get mixed up (not sure if MCP SDK itself is thread safe). Maybe add a `threading.Lock` around `_McpState.run()` to serialize calls, or document that McpClient must not be shared across concurrent walkers.

Jayanaka-98 (changes requested):

Jayanaka-98 (approved):

PR #5483, opened 2026-04-08
feat: microservice mode for local dev mode

MusabMahmoodh:
## Summary

Adds microservice mode to jac-scale: when `[plugins.scale.microservices] enabled = true` is set in `jac.toml`, `jac start` runs each sv-imported module as a separate subprocess behind an API gateway. Inter-service RPC goes through `sv import`, which routes transparently over HTTP with the user's JWT forwarded automatically.

Adds one additive hookspec to jaclang core (`sv_service_call`) so plugins can customize the sv-to-sv RPC transport without patching `sv_client.call`.

## Architecture

- **`sv import`** (core, already on main): compiler-generated HTTP stubs for cross-module function calls
- **`sv_service_call` hookspec** (core, new, additive): transport-layer seam for plugins. Default impl keeps the existing plain-httpx behavior byte-for-byte
- **jac-scale override**: reads the inbound request's `Authorization` header from a `ContextVar` set by the FastAPI middleware, forwards it on every outgoing sv call
- **`ensure_sv_service` hookimpl**: spawns each provider as a subprocess (instead of core's in-process thread) so services have isolated graphs, data directories, and lifecycles — matching K8s pod semantics
- **`MicroserviceGateway`**: FastAPI reverse proxy with path-based routing (`/api/{module}/function/{name}`), built-in route passthrough (`/user/*`, `/walker/*`, etc.), static/SPA serving, and admin UI

## What's in this PR

**jaclang core** (additive only — no behavior changes to existing code):
- `JacAPIServer.sv_service_call` hookspec + default impl (moves the existing `sv_client.call` body into a hookable location)
- `sv_client.call` becomes a thin dispatcher to `JacRuntime.sv_service_call`

**jac-scale** (`jac_scale/microservices/`):
- `service_registry.jac` — prefix-based route table with longest-prefix match
- `process_manager.jac` — subprocess lifecycle with `/healthz` polling and per-service data isolation (`JAC_DATA_DIR`)
- `deployer.jac` / `local_deployer.jac` — `ServiceDeployer` interface + local impl (K8s impl deferred)
- `gateway.jac` — FastAPI middleware split into named handlers (`handle_health`, `handle_proxy`, `handle_admin`, `handle_static`, `handle_builtin_passthrough`)
- `orchestrator.jac` — builds the registry, deploys services, waits for health, starts the gateway
- `setup.jac` — interactive CLI
- `_auth_ctx.jac` — per-request auth `ContextVar` consumed by the `sv_service_call` hookimpl
- `_util.jac` — shared helpers (`pick_free_port`, `resolve_jac_binary`, `wait_for_health`)

**Plugin hooks** (`jac_scale/plugin.jac`):
- `@hookimpl ensure_sv_service` — subprocess-based spawning
- `@hookimpl sv_service_call` — auth-forwarding transport

**CLI:**
- `jac setup microservice` — interactive config, `--add`, `--remove`, `--list`
- `jac scale status / stop / restart / logs / destroy`

**Config** (`[plugins.scale.microservices]`):
- `routes = { module_name = "/api/prefix" }` — gateway URL map. Services are discovered automatically via `sv import`; this table only controls public URL routing

**Example app** (`examples/micr-s-example/`):
- E-commerce with `products_app`, `cart_app`, `orders_app` using `sv import` for inter-service calls + SPA frontend

## Test plan

- [x] 106 tests across 9 suites pass: gateway, deployer, process_manager, orchestrator, registry, setup, sv_auth_forward, microservice, eager_spawn
- [x] New `test_sv_auth_forward.jac` covers the `sv_service_call` override, auth forwarding, and error-envelope handling against `httpx.MockTransport`
- [x] Core `sv_client` tests still pass — the new hookspec's default impl is behavior-preserving
- [ ] Manual E2E: `jac start main.jac` in `examples/micr-s-example/` — gateway + 3 services come up, JWT auth flows client → gateway → service → downstream service
- [ ] `jac scale status/stop/restart/logs` against running services


jac-scale/jac_scale/microservices/impl/gateway.impl.jac
```diff
@@ -0,0 +1,658 @@
...
+            logger.info(f"Passthrough: {request.method} {path} -> {target_url}");
+
+            result = await raw_forward(request.method, target_url, fwd_headers, body);
+            if result is not None {
+                (sc, rh, rb) = result;
+                # If service returned 404, try next service (walker might be elsewhere)
+                if sc == 404 {
+                    continue;
+                }
+                return Response(status_code=sc, headers=rh, content=rb);
```
udithishanka: the "first non-404 wins" rule misroutes when the correct service flaps on transport. caught this live: builder_sv had a one-shot aiohttp ServerDisconnectedError, raw_forward returned None, the loop silently fell through to jac_coder_sv, which 405'd because it doesn't have that walker — and the 405 went back to the user as the final answer.

two issues stacked here:
1. transport error (None) is treated the same as "service didn't have it" — should be a hard error, not a fallthrough
2. 405 means "this service has the path but not this method" — same signal as 404 for fan-out routing, should also `continue`

```suggestion
            result = await raw_forward(request.method, target_url, fwd_headers, body);
            if result is None {
                # Transport error talking to a healthy service — don't silently
                # fall through to a sibling that doesn't own this path.
                import from jac_scale.microservices._errors { service_unavailable }
                return service_unavailable(svc_name, "transport_error");
            }
            (sc, rh, rb) = result;
            # 404/405 from the service mean "I don't own this route" in the
            # fan-out context — try the next service rather than returning.
            if sc in (404, 405) {
                continue;
            }
            return Response(status_code=sc, headers=rh, content=rb);
```

bigger picture: the fan-out strategy itself is fragile. since `_openapi_agg` already pulls each service's openapi.json, you could build a `path -> service` table at gateway startup and skip fan-out entirely for built-ins. happy to follow up on that in a separate PR if you want — for now the above unblocks the misroute.

jac-scale/jac_scale/microservices/impl/http_forward.jac
```diff
@@ -0,0 +1,132 @@
...
+                for (k, v) in resp.headers.items() {
+                    if k.lower() not in _HOP_BY_HOP {
+                        resp_headers[k] = v;
+                    }
+                }
+                return (resp.status, resp_headers, resp_body);
+            }
+        }
+    } except (aiohttp.ClientError, OSError) as e {
+        _logger.error(f"HTTP forward error: {e}");
```
udithishanka: `f"HTTP forward error: {e}"` produces a literal `HTTP forward error: ` (no detail) for `aiohttp.ServerDisconnectedError` because its `__str__` is empty. cost me 30 minutes of log-staring. include the type:

```suggestion
        _logger.error(f"HTTP forward error: {type(e).__name__}: {e}");
```

jac-scale/jac_scale/microservices/impl/gateway.impl.jac
```diff
@@ -0,0 +1,658 @@
...
+
+"""Serve a static file from the client dist directory. Returns None if not found."""
+impl MicroserviceGateway.serve_static(file_path: str) -> Response | None {
+    dist = Path(self.client_dist_dir);
+    target = dist / file_path;
+
+    try {
+        resolved = target.resolve();
+        dist_resolved = dist.resolve();
+        if not str(resolved).startswith(str(dist_resolved)) {
```
udithishanka: `str(resolved).startswith(str(dist_resolved))` matches `/var/app/dist_secrets/x` against `/var/app/dist`. verified locally — `Path("/tmp/app/dist") / "../dist_secrets/secret.txt"` resolves to `/tmp/app/dist_secrets/secret.txt` and the startswith check returns True. `is_relative_to` correctly returns False. same flaw at L129 (`serve_extra_static`) and L221 (`serve_admin_asset`) — applying the same fix there too.

```suggestion
        if not resolved.is_relative_to(dist_resolved) {
```

udithishanka (changes requested):
2 things to fix in the gateway. Hit one of them in production today (the passthrough fallthrough) — easy to reproduce intermittently. Plus a path traversal in the static file serving.

udithishanka (approved):

PR #5499, opened 2026-04-09
Fix PWA install banner not rendering and service worker cache failures

udithishanka:
## Summary
- Fix JS `//` comments in Jac string literals being corrupted by compiler (stripped to single `/`), causing `SyntaxError` in generated PWA scripts
- Replace `cache.addAll()` with resilient individual `cache.add()` + `.catch()` so auth-redirected URLs don't crash service worker installation
- Add missing `beforeinstallprompt` capture and install banner script injection to `render_page()` runtime path

## Changes
- `jac-client/jac_client/plugin/src/targets/impl/pwa_target.impl.jac` — Changed `//` to `/* */` in all generated JS strings; replaced `addAll` with individual caching; removed `'./'` from precache list
- `jac-client/jac_client/plugin/impl/client.impl.jac` — Import and call `_build_install_prompt_capture_script` and `_build_pwa_install_banner_script` in PWA render path

## Test plan
- [ ] Run `jac build --client pwa main.jac` and verify `sw.js` and `index.html` have valid JS (no single-slash comments)
- [ ] Run `jac start --client pwa main.jac` and verify install banner appears after 3 seconds
- [ ] Verify service worker registers without `addAll` cache errors in console
- [ ] Test on app with auth (401 responses should not break SW installation)


marsninja:

PR #5511, opened 2026-04-10
Native fixes 3

mgtm98:
## **Description**
# Native Compiler Fixes — Branch Summary

### Symbol Collision Detection + Abort on Compile/Link Errors

Previously, when two native modules exported the same symbol name, `link_in()` would silently merge them and produce a binary that crashed with SIGSEGV at runtime. Compile and link failures for imported modules were also warnings (W5024/W5025), so `nacompile` would write a corrupt binary without any error exit.

**Fixed by:**
- Tracking a `symbol_origins` dict keyed by exported symbol name; when a collision is detected before `link_in()`, an error (E5024/E5025) is emitted and compilation aborts.
- Promoting import compile/link failures from warnings (W5024/W5025) to **errors** (E5024/E5025) so `nacompile` always exits on failure.
- Renaming `jac_entry` / `__jac_glob_init` DISubprogram debug metadata for each imported module so GDB can correctly resolve symbols across modules.

---

### Incremental Compilation + `--scrub` Option

Every `jac nacompile` previously recompiled all imported `.na.jac` modules from scratch, even when sources had not changed.

**Fixed by:**
- Adding an **incremental compilation** cache under `.jac_ir/`: each source file is SHA-256 hashed; if a matching `<stem>.<hash12>.ir_cache` + `<stem>.<hash12>.ir_meta` (JSON with `clib_paths` and `transitive_deps`) exists, the full recompile is skipped and the cached LLVM IR is reused directly.
- Adding a `--scrub` CLI flag to `jac nacompile` that wipes all `.jac_ir/` cache directories and forces a clean rebuild.

---

### Variable Re-Declaration Type Fix in Assignment Codegen

When a variable was re-declared with a different type (e.g. `x: bool` in one branch vs. `x: SomeStruct` in another), codegen would store through a wrongly-typed alloca pointer, causing miscompilation and runtime crashes.

**Fixed by:**
- Detecting typed re-declarations (`x: T = ...`) where the declared type differs from the existing alloca's `pointee` type, and allocating a fresh alloca of the correct type in that case.
- Skipping the re-allocation for collection types (`list`, `dict`, `set`, `frozenset`) which are handled separately.

---

### Clear `.jac_ir` Cache on `--no-cache`

`jac run --no-cache` only cleared the Python-side `.jac/` cache, leaving stale native IR caches in `.jac_ir/` directories in place.

**Fixed by:**
- Recursively globbing and removing all `.jac_ir/` directories under the current working directory when `--no-cache` is passed, ensuring truly fresh compilation across all sibling modules.


marsninja:
## Review notes from a structural pass over the diff

(Updated after merging main.) The native-compiler fixes here are mostly real, well-targeted work — symbol-collision detection, error promotion, `--scrub`/`--no-cache`, the redecl-type fix, and the native-module-resolution refactor are all going in the right direction. About 70% of the diff is solid.

The remaining concerns below are quality-of-implementation issues. Items 1–4 I'd want addressed before this lands; the rest can be follow-ups.

### Concerns

1. **`jac run --no-cache` rmtrees every `.jac_ir` under `os.getcwd()`** — `execution.impl.jac:103-112` does `glob(os.getcwd() + '/**/.jac_ir', recursive=True)` and `shutil.rmtree`s each match. If a user runs `jac run --no-cache` from a workspace root, it nukes caches in every sibling project. `nacompile --scrub` is correctly scoped to `os.path.dirname(os.path.abspath(filename))` (`nacompile.impl.jac:74`) — the asymmetry between the two paths is the giveaway. `--no-cache` should also be scoped to the file's source tree.

2. **`JAC_VERIFY_IR=1` is dead code under current llvmlite.** `debug.impl.jac:132` calls `llvm.initialize()`. With the llvmlite in this project's `.venv`, that raises `RuntimeError: llvmlite.binding.initialize() is deprecated and will be removed`. The surrounding `try { ... } except Exception` swallows it, so `JAC_VERIFY_IR=1` always prints `[JAC_VERIFY_IR] VERIFICATION FAILED: ...` with the deprecation message — the verifier never runs. Drop the deprecated `llvm.initialize()` call (the sibling `initialize_native_target()` / `initialize_native_asmprinter()` are not deprecated and can stay).

3. **`os.environ.get("JAC_DEBUG_IR", False)` and `JAC_VERIFY_IR` use truthy strings** — `debug.impl.jac:113` and `:129`. `"0"`, `"false"`, `"no"` all evaluate truthy. Use `== "1"` or `.lower() in {"1","true","yes"}`.

4. **Re-declaration alloca swap leaks refcounted values.** `stmt.impl.jac:685-706` replaces the alloca for `x: T = ...; x: U = ...` but doesn't release the old refcounted value. `x: str = "hello"; x: int = 5;` abandons the string ref. No test covers pointer→non-pointer redecl. The hardcoded `{"list","dict","set","frozenset"}` skip set also doesn't match parameterized `list[int]` (which arrives as `IndexExpr`, not `Name`/`BuiltinType`), so `x: list[int] = ...; x: list[str] = ...` falls through to the alloca swap.

5. **Magic linkage ints** — `na_compile_pass.impl.jac:266-267` uses `_INTERNAL_LINKAGE = 8`, `_PRIVATE_LINKAGE = 9`. Pins a hidden contract to llvmlite's `Linkage` enum order. Use `_mfn.linkage.name == "internal"` or import the enum.

6. **`transform()` doesn't bail on `self.errors_had`** before MCJIT engine creation. After symbol-collision detection emits errors, execution continues into `create_mcjit_compiler(mod, target_machine)` at `na_compile_pass.impl.jac:187` on top of garbage IR. The CLI gates at `nacompile.impl.jac:130` and `execution.impl.jac:131` catch it eventually, but `transform()` itself should short-circuit on `self.errors_had` before JIT creation.

7. **Comment at `func.impl.jac:198` is misleading** — it says "Non-:pub abilities get internal linkage" but the code only sets `internal` when `KW_PRIV` is explicit (default + `:pub` both stay external). The `globals.impl.jac` comment is correct ("Default (no access modifier) is public in Jac"). Fix this one.

8. **Extended LLVM IR text-replace pipeline** (`na_compile_pass.impl.jac:489-515`) is fragile. A user string literal `"jac_entry"` baked into IR as a constant gets corrupted; `linkageName: "jac_entry"` in DWARF stays out of sync with the rewritten `name:` field. The right fix is `function.name = ...` on the parsed module before `link_in()` — but understandable if that's a follow-up.

9. **`_DepBinding` class defined inside the cache-hit branch** (`na_compile_pass.impl.jac:388-392`). Cleaner as `SimpleNamespace` or hoisted to module scope.

10. **Symbol-collision check classifies weak/`linkonce_odr` as a hard collision.** Anything not `internal`/`private` triggers E5026. The Jac codegen never emits weak symbols today, but this is a footgun for future C-interop modules where weak symbols are the standard dedupe mechanism.

11. **Removed "stdlib shadow handling" test** — commit calls it redundant but I don't see a replacement exercising stdlib-shadowed module names. The 5511 release note still claims the fix exists. Worth confirming coverage or restoring the test.

### What's clearly right

- `nacompile`/`run` error gates on `prog.errors_had` before writing binary / starting JIT.
- W5024/W5025 → E5024/E5025 promotion + new E5026 with both source files named — good error UX.
- The refactor unifying native module resolution + DWARF helpers (commit `4872705388`) — net –131 LOC, factors out 6 copy-pasted DWARF blocks. Solid.
- Removing no-op `with entry {}` from the transitive-import fixtures — consistent with the stricter symbol rules.
- The redecl-type fix itself (modulo the leak in #4) — narrow and correct for the int/bool case.
- `jac-gdb-helper.jac` null-pointer formatting — sensible UX fix.
- The incremental-compilation cache mechanism — SHA-256 of source → IR text + JSON metadata, with stale-cache cleanup. Sensible design.

marsninja:
## Codebase-fit review: existing patterns this branch should reuse

Separate from the correctness review above — this is about whether the new code matches existing conventions in the project, or reinvents infrastructure that already exists. Most of the integration is good (diagnostic registry, `self.emit`, DWARF helpers, `codeinfo.jac` placement, `uni?.X` guard pattern all match the rest of the codebase). The places where it diverges are below.

### 1. JIR cache already exists — the new `.jac_ir/` cache is a parallel system

[`jac/jaclang/jac0core/jir.jac`](jac/jaclang/jac0core/jir.jac) is the unified cache. Keyed by SHA-256 of source path, with TLV sections including `SEC_LLVM_IR` (zlib-compressed IR text), `SEC_NATIVE_OBJ` (post-opt bitcode), and `SEC_INTEROP` (manifest as JSON). The branch's flat `.jac_ir/<stem>.<hash8>.ir_cache` + separate `.ir_meta` JSON sidecar in `na_compile_pass.impl.jac:344-368` ignores all of this. Even the hash truncation is inconsistent — JIR uses `[:8]` ([jir.jac:292](jac/jaclang/jac0core/jir.jac#L292)), the branch uses `[:12]`. The aligned approach would be: read IR from `prog.compile()`'s already-cached JIR module rather than caching the IR text separately.

### 2. Project-root resolution is already solved

`_get_module_cache_dir()` ([jir.jac:260](jac/jaclang/jac0core/jir.jac#L260)) walks: installed package → global cache; project source → `config.get_cache_dir()`; otherwise → `.jac/cache/` next to source. The branch hard-codes `.jac_ir/` next to source and uses **two different roots for the same operation** — `os.getcwd()` for `--no-cache` ([execution.impl.jac:107](jac/jaclang/cli/commands/impl/execution.impl.jac#L107)), `os.path.dirname(filename)` for `--scrub` ([nacompile.impl.jac:74](jac/jaclang/cli/commands/impl/nacompile.impl.jac#L74)). Should use `config.project_root` from [config.jac:213](jac/jaclang/project/config.jac#L213) (already imported into `execution.impl.jac` via `_discover_config_from_file`).

### 3. `jac clean --cache` already exists

[project.impl.jac:811](jac/jaclang/cli/commands/impl/project.impl.jac#L811) implements project-scoped cache cleaning via `config.get_cache_dir()`. `--scrub` and `--no-cache` reinvent it. If the IR were stored in JIR like everything else, `jac clean --cache` would just work — no new flag, no new glob.

### 4. There's an established `JAC_*` env-var idiom

[runtime.jac:1057](jac/jaclang/jac0core/runtime.jac#L1057) uses `os.environ.get(name, "").strip().lower() in ("1", "true", "yes", "on")`. The same file as the broken `JAC_DEBUG_IR` already uses `bool(os.environ.get("JAC_NATIVE_DEBUG", ""))` correctly at [na_compile_pass.impl.jac:61](jac/jaclang/compiler/passes/native/impl/na_compile_pass.impl.jac#L61). The default-`False` truthy form at [debug.impl.jac:113,129](jac/jaclang/compiler/passes/native/na_ir_gen_pass.impl/debug.impl.jac#L113) is a one-line fix to match. (Side note: 14+ `JAC_*` vars across the codebase justify a `getenv_bool` helper if anyone wants to land one.)

### 5. llvmlite has a `name` setter — use it instead of 6 string `.replace()` calls

llvmlite's `ValueRef.name` is settable. The branch already iterates `imported_mod.functions` for collision detection at [na_compile_pass.impl.jac:532](jac/jaclang/compiler/passes/native/impl/na_compile_pass.impl.jac#L532). Extending that loop with `if func.name == "jac_entry": func.name = entry_rename` replaces the 6 quoted/bare/DI-name `.replace()` lines at [na_compile_pass.impl.jac:489-515](jac/jaclang/compiler/passes/native/impl/na_compile_pass.impl.jac#L489) and stops corrupting unrelated string literals. The same module already uses `engine.get_function_address(...)` — the llvmlite API path is the established norm here, not text munging.

### 6. `_DepBinding` should reuse `InteropBinding`

The class-in-function shim at [na_compile_pass.impl.jac:388-392](jac/jaclang/compiler/passes/native/impl/na_compile_pass.impl.jac#L388) is a duck-typed substitute for `InteropBinding` ([codeinfo.jac:161](jac/jaclang/jac0core/codeinfo.jac#L161)) — which is exactly what the work queue is parameterized on. Using `InteropBinding(name="", source_context=..., source_module=dep_mod_path)` keeps the queue homogeneous and means downstream changes to `InteropBinding` won't silently bypass the cache-hit path. There's no `dataclass`/`NamedTuple`/`SimpleNamespace` precedent in this neighborhood — `obj`/`class` declarations in `codeinfo` are the convention.

### 7. Bare `try { ... } except Exception { }` doesn't match the codebase

7 such blocks in `na_compile_pass.impl.jac` (lines 184, 193, 377, 454, 479) plus dead commented-out `# print(...)` next to them. The codebase pattern when tolerating an error is `except OSError as e { print(f"Error ...{e}", file=sys.stderr); }` ([cli_boot.jac:62](jac/jaclang/jac0core/cli_boot.jac#L62)) or routing via `console.error(...)` ([execution.impl.jac:399](jac/jaclang/cli/commands/impl/execution.impl.jac#L399)). The branch's swallows hide cache corruption and symbol-map write failures silently — should at minimum gate behind `JAC_NATIVE_DEBUG` and route through `console`.

### Patterns this branch establishes (worth keeping)

- **Symbol-origin tracking for link-time collision diagnostics** (`symbol_origins` dict + `_global_origins`). Genuinely new — AST-level dup checks (E0050/E0075/E0076) cover parse-time dups, E2017 covers impl dups, but nothing existed for cross-module link-time symbol collisions. The seed-with-main + check-on-link pattern is sound and worth keeping as the template if/when JS or bytecode codegen grow link-time merging.
- **DWARF helpers** `_attach_di_subprogram` / `_finalize_di_locations` (commit `4872705388`) replace 6 copy-pasted blocks with shared helpers in `debug.impl.jac`. Already used in [objects.impl.jac:461](jac/jaclang/compiler/passes/native/na_ir_gen_pass.impl/objects.impl.jac#L461) — the refactor extends the right path.
- **`resolve_native_module` placed in `codeinfo.jac`** — alongside `NativeModuleInfo`, `NativeFunctionInfo`, `InteropBinding`. Now reused by `InteropAnalysisPass` and `NaIRGenPass`.
- **Field declarations on `CodeGenTarget`/`InteropManifest`** ([codeinfo.jac:244](jac/jaclang/jac0core/codeinfo.jac#L244)) replacing dynamic attribute injection — matches the obj-with-`has` convention.

### Top architectural item

If I had to pick one thing to push back on, it's **#1 + #2 + #3 collapsed**: this branch built a parallel native-IR cache when JIR exists, with project-root resolution that contradicts itself across two flags, and reinvented `jac clean --cache`. Aligning with JIR is non-trivial work, but it's the difference between this becoming the new caching fork and the existing system continuing to be the source of truth.

mgtm98:
More fixes are added 


## 1. Borrowed-Parameter Promotion — Double-Free / Use-After-Free Fix

**Files:** `func.impl.jac`, `objects.impl.jac`, `stmt.impl.jac`, `na_ir_gen_pass.jac`

**Problem:** When a function parameter was reassigned inside the function body, the compiler would release (decrement RC on) the old alloca value — which was still owned by the *caller*. Combined with the caller's own scope-exit release this produced a double-free, and under the Boehm GC (which reuses freed memory) it could manifest as apparent AST cycles or corrupted state.

**Fix:** On the *first* reassignment of a borrowed parameter, a retain is emitted in the function's **entry block**, immediately after the original `store(arg, alloca)` instruction. This converts the alloca from "borrowed" to "owned" at runtime, regardless of whether the reassigning branch is actually executed (e.g., inside a loop that never runs). Subsequent reassignments and scope cleanup then treat the variable as a normal owned local.

New fields on `NaIRGenPass`:
- `_param_store_instr: dict[str, object]` — maps param name to its entry-block store instruction, used to `position_after()` for the retain insertion.
- `_param_promoted: set[str]` — tracks which params have already had the entry retain emitted (avoids duplicate retains on multiple reassignments).

Both fields are reset and saved/restored for each function, method, and no-sig-method codegen scope.

---

## 2. `__rc_safe_free` — `JAC_NO_GC=1` Now Suppresses All Frees

**Files:** `refcount.impl.jac`, `container_helpers.impl.jac`, `dicts.impl.jac`, `gc_debug.impl.jac`

**Problem:** `JAC_NO_GC=1` was intended to disable all retain/release (for benchmarking "with vs. without GC overhead"), but direct `free()` calls in struct/container destructors, array grow, and dict resize bypassed the GC-disabled flag — memory was still freed.

**Fix:** A new `__rc_safe_free(i8* ptr)` function is emitted into each module. It checks `__rc_gc_disabled` and calls `free()` only when RC is active. All compiler-emitted `free()` calls (struct dtors, container dtors, array grow, dict resize, live-list node removal, raw-free path) now go through `__rc_safe_free` instead of `free` directly.

---

## 3. Shared RC Globals — `internal` → `weak` Linkage

**Files:** `refcount.impl.jac`, `gc_debug.impl.jac`

**Problem:** Several RC control globals (`__rc_gc_disabled`, `__rc_debug`, `__rc_ctx_op`, `__rc_ctx_loc`, `__rc_live_head`, `__rc_alloc_id_counter`) had `internal` linkage. After `link_in()` merges imported modules, each module had its *own private copy*. Calling `__rc_gc_disable()` from `jac_entry` only disabled the main module's copy — imported modules still saw `__rc_gc_disabled = 0` and freed memory anyway.

**Fix:** These globals are now emitted with `weak` linkage so the linker merges all per-module copies into a single shared symbol after `link_in()`. Function definitions `__rc_debug_enable`, `__rc_debug_disable` are changed to `weak_odr` (one definition rule — body is identical across modules, linker keeps one).

New constant definitions in `na_compile_pass.impl.jac`:
- `_WEAK_LINKAGE = 5` and `_WEAK_ODR_LINKAGE = 6` — used to skip weak/weak_odr symbols during function/global collision detection (they are safe to define in multiple modules).

---

## 4. `JAC_NO_GC` Check Before Module Init

**Files:** `func.impl.jac`

**Problem:** `__jac_glob_init()` was called before the `JAC_NO_GC` env-var check. Any retain/release emitted during module initialization (e.g., dropping interim string values) would execute with RC still active, defeating the purpose of `JAC_NO_GC=1`.

**Fix:** The `JAC_NO_GC` getenv + `__rc_gc_disable()` call is now emitted *before* `glob_init_fn` and all imported module init calls.

---

## 5. Compiler Fingerprint in IR Cache Keys

**Files:** `na_compile_pass.impl.jac`

**Problem:** The incremental IR cache key was derived only from the source file hash. Changing the compiler implementation (e.g., fixing a codegen bug) would not invalidate existing `.ir_cache` files, so stale binaries from the previous compiler version could be silently reused.

**Fix:** At the start of `_compile_and_link_native_imports`, all `.jac` files under `compiler/passes/native/impl/` and `na_ir_gen_pass.impl/` are hashed together to form an 8-hex-character compiler fingerprint `_compiler_fp`. The cache paths now include this fingerprint:

```
<stem>.<src_hash>.<comp_fp>.ir_cache
<stem>.<src_hash>.<comp_fp>.ir_meta
```

Any change to a compiler implementation file automatically invalidates all cached IR for all sources.

---

## 6. Symbol Map Output for Profiling (`JAC_SYMMAP=1`)

**Files:** `elf_linker.jac`, `elf_linker.impl.jac`

**Change:** Added a `sym_vaddrs: dict[str, int]` field to `ElfLinker`. After `build_executable()` resolves final virtual addresses for all symbols, it populates this map. In `ElfLinker.link()`, when the `JAC_SYMMAP` environment variable is set (non-empty), a `<binary>.symmap` file is written alongside the output binary. Each line has the format:

```
<symbol_name> 0x<vaddr_hex_16>
```

This allows profiling post-processors (e.g., `scripts/cg-annotate`) to map raw Callgrind addresses to real function names.

---

## 7. `__float_bits` / `__bits_float` Compiler Intrinsics

**Files:** `calls.impl.jac`

**Change:** Two new built-in compiler intrinsics are recognized by `_codegen_call`:

- `__float_bits(d: float) -> int` — reinterprets the IEEE-754 bit pattern of a `float64` as a signed `int64` via a single LLVM `bitcast`. Zero memory traffic; no alloca round-trip.
- `__bits_float(b: int) -> float` — the inverse: reinterprets an `int64` bit pattern as `float64`.

These are used by NaN-boxed `JSValue` helpers in jac-bun (`jsv_from_double` / `jsv_to_double`) to eliminate the `memcpy` shim that was previously needed.

---

## 8. RC Debug Context (`_set_rc_ctx` / `_clear_rc_ctx`) on All Allocations

**Files:** `primitives_native.impl.jac`, `primitives_native.jac`, `builtins.impl.jac`, `file_io.impl.jac`, `types.impl.jac`

**Change:** `_set_rc_ctx` / `_clear_rc_ctx` calls have been added around every `rc_alloc_fn` call in primitive string/numeric emitters, builtins, and file I/O. This ensures the RC debug globals (`__rc_ctx_op`, `__rc_ctx_loc`) carry meaningful context for every allocation when RC tracing is enabled.

Affected operations: `to_bytes`, `as_integer_ratio` (int & float), `float_hex`, `complex_new`, `capitalize`, `upper_lower`, `title`, `swapcase`, `replace`, `lstrip`, `rstrip`, `removeprefix`, `removesuffix`, `join`, `chr`, `str_substr`, `str_strip`, `str_split`, `argv`, `str_repeat`, `file_read`.

---

## 9. `loc` Propagation in RC Release Calls

**Files:** `expr.impl.jac`

**Change:** In `_codegen_binary` and `_codegen_compare`, the `loc=""` argument passed to `_emit_rc_release_typed` / `_emit_rc_release_simple` is replaced with `getattr(val, 'jac_loc', '')`. This propagates actual source location information into RC debug traces for binary-operator and `is`/`is not` comparisons.

PR #5546, opened 2026-04-14
feat(server): opt-in walker thread offload via JAC_WALKER_OFFLOAD

asimsalim:
Addresses #5486

## Problem

`ExecutionManager.spawn_walker` is declared async, but the critical section it owns is a mix of synchronous and async work that all runs on the uvicorn main loop:

- `self.user_manager.get_root_id(username)` — sync SQLite auth lookup
- field deserialization loop
- `Jac.get_object(target_node_id)` — sync memory/DB fetch for the target node
- `Jac.spawn(walker, target_node)` — sync walker body execution for sync walkers
- `await spawn_result` — async walker body, awaited back on the main loop
- `await self.report_collector(call_state)` — report drain
- `Jac.commit()` — sync DB write

Any blocking call anywhere in this chain holds the event loop for the full duration and serialises every concurrent request behind it. A narrow fix that only wraps `Jac.spawn` in `asyncio.to_thread` helps the sync-walker case but leaves the other four blocking sections untouched — async walkers in particular still stall because their coroutine is awaited on the main loop at the existing `iscoroutine` branch.

## Approach

When `JAC_WALKER_OFFLOAD=1` is set, dispatch the whole critical section to a sync helper running under `asyncio.to_thread`:

```jac
impl ExecutionManager.spawn_walker(...) -> dict[str, JsonValue] {
    import os;
    if os.environ.get('JAC_WALKER_OFFLOAD') {
        import asyncio;
        return await asyncio.to_thread(
            self._spawn_walker_offloaded, walker_cls, fields, username
        );
    }
    # Default path is byte-identical to the original implementation.
    ...
}
```

`_spawn_walker_offloaded` is a new sync method on `ExecutionManager`. It duplicates the body of `spawn_walker` with one structural change: since it runs on a worker thread with no running loop, it creates its own dedicated `asyncio.new_event_loop()` and uses `loop.run_until_complete(...)` for both the async walker coroutine and the async `report_collector`. Everything else — auth lookup, deserialization, target node resolution, `Jac.spawn`, `Jac.commit`, serialization — runs as sync code on the worker thread, which is exactly what you want for blocking IO.

Default behaviour (env var unset) is unchanged. Every existing deployment keeps the original code path byte-for-byte.

### Why opt-in rather than always-on

Running the critical section on a worker thread changes semantics in ways that can break user code. Known scenarios under offload:

- Sync walkers calling `asyncio.get_running_loop()` / `asyncio.create_task()` directly raise `RuntimeError` because the worker has no running loop during the non-async parts. If a walker body reaches asyncio APIs it needs to be declared `async` so it's inside the worker's `run_until_complete`.
- Async walker coroutines now resolve on the **worker thread's own event loop**, not the uvicorn main loop. Any async resource pinned to the main loop (a shared `aiohttp` session created at startup, an `asyncpg` pool bound to uvicorn's loop) may not be cross-loop safe.
- `contextvars.Context` propagates into `asyncio.to_thread` automatically (standard Python `copy_context` behaviour), so user ContextVars set before spawn are visible. `ContextVar.set` inside the walker body only mutates the worker's copy and does not persist back to the caller's context.
- Thread pool exhaustion is now possible under sustained blocking load since the default executor is bounded (~32 workers). This replaces one throughput problem with a different and much higher ceiling, but it is worth flagging.

Users hitting #5486 opt in per-deployment via `JAC_WALKER_OFFLOAD=1`, everyone else is untouched.

### Compatibility notes

- **Local `jac start` + SQLite**: both jaclang SQLite connections (`UserManager` at `server.impl.jac:26`, storage at `memory.impl.jac:175`) already use `check_same_thread=False`, so the offloaded helper can safely call `get_root_id`, `Jac.get_object`, `Jac.spawn`, and `Jac.commit` from a worker thread. No breakage. The flag is a no-op for concurrency on the local plain `HTTPServer` transport though, because that transport is already single-threaded at the socket accept layer (a ~50-200µs thread-hop per request for no throughput gain). Intended for production FastAPI/uvicorn via jac-scale.
- **jaseci-enterprise**: flag is orthogonal. Enterprise already offloads at a higher layer (whole walker/function endpoint wrapped in a dedicated 50-worker `ThreadPoolExecutor`), so enabling this flag on an enterprise deployment would just produce a nested thread hop. Harmless but wasteful. Not recommended on enterprise.

## Tests done

Verified locally by running the repo source via `PYTHONPATH=jac:jac-scale python -m jaclang.cli.cli start` (no install side-effects) against a walker that exercises all three blocking categories:

```jac
import time;

node Item {
    has idx: int = 0;
    has label: str = "";
}

walker ping {
    has msg: str = "";
    can go with Root entry {
        self.msg = "pong";
        report self.msg;
    }
}

walker sync_sleep {
    has msg: str = "";
    can go with Root entry {
        time.sleep(3);
        self.msg = "sync_done";
        report self.msg;
    }
}

async walker async_sleep_blocking {
    has msg: str = "";
    async can go with Root entry {
        # Intentionally blocking (not asyncio.sleep) inside an async walker.
        time.sleep(3);
        self.msg = "async_done";
        report self.msg;
    }
}

walker commit_heavy {
    has count: int = 500;
    has created: int = 0;
    can go with Root entry {
        i = 0;
        while i < self.count {
            here ++> Item(idx=i, label=f"item-{i}");
            i += 1;
        }
        self.created = i;
        report self.created;
    }
}
```

Procedure: fire the long walker in the background, wait 300ms (or 50ms for `commit_heavy`), fire `/walker/ping` and measure latency while the long walker is still pending. Same server binary, same walker, same JWT, only `JAC_WALKER_OFFLOAD` toggled.

| Scenario | What blocks today | `ping` (default, unchanged) | `ping` (`JAC_WALKER_OFFLOAD=1`) | Long walker (default) | Long walker (offload) |
|---|---|---:|---:|---:|---:|
| A. sync walker with `time.sleep(3)` | `Jac.spawn` body | **2724 ms** (queued) | **7 ms** (concurrent) | 3029 ms | 3010 ms |
| B. async walker with blocking `time.sleep(3)` inside `async can` body | `await spawn_result` at `iscoroutine` branch | **2740 ms** (queued) | **24 ms** (concurrent) | 3020 ms | 3010 ms |
| C. sync walker creating 500 nodes + `Jac.commit` | walker body + `Jac.commit` | **2013 ms** (queued) | **16 ms** (concurrent) | 2030 ms | 5008 ms |

Observations:

**Scenario A** — sync walker body. Default: ping queues behind the 3s `time.sleep`. Offload: ping returns in 7ms while the sync walker still runs to full 3s.

**Scenario B** — async walker body with blocking IO. Default: even though the walker is declared `async walker`, its blocking `time.sleep` inside the `async can` ability stalls the main loop because the coroutine is awaited on uvicorn's loop. Offload: the coroutine resolves on the worker thread's own loop via `loop.run_until_complete`, so the main loop stays free and ping returns in 24ms.

**Scenario C** — `Jac.commit` on a large write. Default: even though the walker's Python work finishes quickly, the 500-node commit on the main loop stalls ping for ~2s. Offload: commit runs on the worker thread, ping returns in 16ms. Note the long walker end-to-end latency went from 2030 → 5008 ms, about 2.5x slower for that one request. That is the cost of moving the entire pipeline to a worker thread — the main loop stays responsive for every other request, but the single slow request pays a worker-loop setup overhead plus GIL contention for 500 Python object creates and sqlite writes. This trade is exactly what the opt-in flag exists for: worth it on a multi-tenant FastAPI server, not worth it for a single-tenant CLI.

Also ran `jac/tests/runtimelib/test_serve.jac`: **10 passed**. The default code path is byte-identical to the original so existing server tests are unaffected.

## Scope

- `jac/jaclang/runtimelib/server.jac`: +3 lines, one new sync method declaration on `ExecutionManager`.
- `jac/jaclang/runtimelib/impl/server.impl.jac`: +~65 / -~10, a 4-line dispatch in `spawn_walker` plus the new `_spawn_walker_offloaded` sync impl.
- Release notes updated to reflect the wider coverage and the compatibility notes.
- No new tests added yet. Happy to add a regression test under `jac/tests/runtimelib/` that exercises all three scenarios with a mocked clock, or to flip the default from off to on in a later release once we have production signal from opt-in users.


asimsalim:
Closing per review feedback.

PR #5554, opened 2026-04-15
feat(jac-scale): Support custom `obj` types as walker & `@restspec` function body parameters

akindu-k:
## Problem

When a walker or `@restspec` function declared a field/parameter typed as a user-defined Jac `obj`, the jac-scale HTTP layer collapsed it to a flat `str` in the Pydantic body model. For example:

```jac
obj UserBody { has name: str, age: int; }

walker:pub CreateUser {
    has user: UserBody;
    can create with Root entry { ... }
}
```

produced a `user: str` field in the OpenAPI schema and handed the walker raw, unparsed data. Callers had to serialize nested objects as strings and handlers had to re-parse them, defeating the point of typed `obj`s at the HTTP boundary. Nested (`UserWithAddress`), list (`list[UserBody]`), and self-referential (`TreeNode`) types were not supported at all.

A second, related problem: any archetype with a mutable default (e.g. `has children: list[TreeNode] = []`) triggered `PydanticJsonSchemaWarning: Default value <factory> is not JSON serializable ...` when generating the OpenAPI schema. Because `pytest.ini` sets `filterwarnings = error`, this escalated into a failing `openapi specs` test.

## Solution

Carry the real Python type object through the parameter pipeline, build nested Pydantic models from Jac archetypes on demand, and reconstruct archetype instances from the validated payload before dispatching to the walker/function handler.

1. **Preserve the real type object.** Added `APIParameter.type_obj: Any = None` alongside the existing stringified `data_type`. `create_walker_parameters` / `create_function_parameters` now populate it via `get_type_hints(cls.init)` / `get_type_hints(func)` at registration time.

2. **Resolve types to Pydantic-compatible types.** New `_resolve_type(type_string, type_obj)` centralises the lookup: primitive fallback → Jac archetype (→ `_build_pydantic_model`) → generic `list`/`dict` (recurse) → `Union`/`Optional` (recurse per arm) → plain Python type.

3. **Build Pydantic models from Jac archetypes.** New `_build_pydantic_model(jac_cls)` mirrors a Jac archetype's `has` fields into a dynamic Pydantic `BaseModel`. To make self-referential types like `TreeNode { has children: list[TreeNode]; }` terminate, it caches a `ForwardRef(model_name)` in `self._models` *before* recursing into field types; after `create_model`, it calls `model.model_rebuild(force=True, _types_namespace=...)` to resolve the forward refs.

4. **Fix the `<factory>` warning.** Within `_build_pydantic_model`, when a parameter default is the dataclass sentinel `_HAS_DEFAULT_FACTORY`, look up the real factory via `jac_cls.__dataclass_fields__[name].default_factory` and emit `Field(default_factory=...)` instead of `Field(default=<sentinel>)`. This produces a valid JSON schema and silences the warning.

5. **Reconstruct archetype instances after validation.** New `_pydantic_to_jac(data, jac_cls)` walks the validated payload and rebuilds real Jac archetype instances (handles `None`, Pydantic model, `list`, `dict`, `Union`/`Optional`, and the archetype case). `_create_endpoint_function` threads a `body_type_map: {param_name: jac_type}` into the generated endpoint wrapper's globals as `_body_types`, and the wrapper calls `_pydantic_to_jac(body_data.<name>, _body_types['<name>'])` for each body param that carries a Jac type. Walkers/functions then see real typed instances, so `self.user.name` and `type(self.user).__name__` work as expected.

6. **Use the resolver for request bodies.** `JSONBodyParameterHandler.generate_body_model` now calls `_resolve_type(param.data_type, param.type_obj)` instead of the string-only `_get_python_type`.


akindu-k:
image
image

Thamirawaran (changes requested):
Only walkers are handled. functions also need to be handled
add test cases for recursive possibilities , custom objects, jac objects in both cases walker, functions

akindu-k:
References: [Forward Annotations](https://pydantic.dev/docs/validation/latest/concepts/forward_annotations/)

Thamirawaran (approved):

akindu-k:
image

PR #5558, opened 2026-04-15
feat(websocket): overhaul auth, scaling, transport & function support

kashmithnisakya:
## Description
Rewrites the WebSocket layer in jac-scale to fix auth flakiness, scale across workers, and support def functions.

- Auth at connect time: query-token validated pre-accept (clean handshake reject on failure), first-frame as fallback, with tagged error reasons `(missing/expired/invalid/unknown_user)` and `{type:"refresh"}` control message for long-lived sessions
- Per-target / per-user / per-anonymous-target connection caps, message size limit, per-send timeout, rate limiter, heartbeat/ping
- Pluggable pub/sub backplane (Memory + Redis) so broadcasts reach every worker; explicit `[plugins.scale.websocket].backplane = 'memory' | 'redis'` selector with auto-detect fallback
- WebSocket support for `def` functions at `/ws/function/{name}` with optional `streaming=True` and `broadcast=True` (via `WebSocketTransport` with pluggable `sender` so streaming output can fan out)
- Walkers at canonical `/ws/walker/{name}` with `/ws/{name}` legacy alias kept for one release
- Real `WebSocketTransport` impl in jac core; unified `TransportResponse` envelope across HTTP and WS
- Prometheus metrics + graceful shutdown hook


jac-scale/jac_scale/impl/config_loader.impl.jac
```diff
@@ -321,6 +328,23 @@ impl JacScaleConfig.get_scheduler_config(self: JacScaleConfig) -> dict[str, Any]
...
+    return {
+        'max_message_bytes': int(ws_config.get('max_message_bytes', 65536)),
+        'max_connections_per_target': int(
+            ws_config.get('max_connections_per_target', 5000)
+        ),
+        'max_connections_per_user': int(ws_config.get('max_connections_per_user', 10)),
+        'target_timeout_seconds': int(ws_config.get('target_timeout_seconds', 30)),
+        'messages_per_second': int(ws_config.get('messages_per_second', 20)),
+        # Auto-detect: Redis URL present → RedisBackplane; else MemoryBackplane.
+        'redis_url': self.get_database_config().get('redis_url')
```
MusabMahmoodh: the WS backplane gets auto-promoted to redis the moment someone sets `redis_url` under `[plugins.scale.database]`. thats a surprising coupling no? the database section is for caching / queue stuff, suddenly all my ws broadcasts start flowing through that same redis too.

also the release note advertises `[plugins.scale.websocket].backplane = 'redis'` but that knob isnt actually wired up anywhere — `build_backplane` in ws_backplane.jac only reads `cfg['redis_url']`.

imo: explicit `[plugins.scale.websocket].backplane = 'memory' | 'redis'` with its own optional `redis_url` (falling back to the db one if not set, but honoring `'memory'` even when a db redis_url exists). less spooky, matches what the release note says.

jac-scale/jac_scale/websocket.jac
```diff
@@ -1,54 +1,306 @@
...
+        if per_target is None {
+            per_target = {};
+            self._by_target[target_key] = per_target;
         }
-        self._active_connections[walker_name].append(websocket);
+        if len(per_target) >= self.max_per_target {
+            rejection = 'per_target_limit';
+        } elif ctx.user_id and self._by_user.get(ctx.user_id, 0) >= self.max_per_user {
+            rejection = 'per_user_limit';
+        } else {
```
MusabMahmoodh, line 144: the per-user cap only kicks in for `ctx.user_id` (authenticated). anonymous / `pub` ws connections only hit `per_target_limit` which defaults to 5000.

so a single misbehaving client can open ~5000 anon sockets to a `pub` ws endpoint and we have a free DoS. is that intentional?

can we add a per-IP cap (or a `max_per_target_anonymous`)? otherwise the docs should explicitly say "don't expose `pub` ws to the internet without a layer-7 proxy doing per-ip limits".

jac-scale/jac_scale/impl/serve.endpoints.impl.jac
```diff
@@ -463,111 +481,566 @@ impl JacAPIServerEndpoints.register_webhook_endpoints -> None {
...
     walker_cls = self.get_walkers()[walker_name];
     restspec = walker_cls.restspec
         if walker_cls?.restspec and walker_cls.restspec
         else None;
     is_broadcast = restspec.broadcast if restspec and restspec?.broadcast else False;
+    user_manager = self.user_manager;
+    execution_manager = self.execution_manager;
     async def websocket_handler(websocket: WebSocket) {
-        await ws_manager.connect(websocket, walker_name);
+        await websocket.accept();
```
MusabMahmoodh: we `accept()` the socket before validating the token. once accept runs the client has seen a successful 101 handshake, and anything after is a post-handshake close — some clients / proxies surface that as a generic disconnect rather than an auth failure.

i get why its structured this way (the auth handshake on line 508 reads JSON from the socket so it needs to be open). but the query-token branch (`_ws_authenticate` line 510-512) doesnt need an open socket. could we short-circuit pre-accept when a query token is present? something like:

  if query token present: validate, reject pre-accept on failure
  else: accept, read first frame as auth

matches the standard ws auth flow + close code 4401 lands more cleanly on most clients.

jac-scale/jac_scale/impl/serve.endpoints.impl.jac
```diff
@@ -463,111 +481,566 @@ impl JacAPIServerEndpoints.register_webhook_endpoints -> None {
...
-                                }
-                            }
-                        );
+                    if status == 'skip' {
                         continue;
                     }
                 }
-
-                # Spawn the walker with received fields
+                ctx.last_pong = time.monotonic();
```
MusabMahmoodh: minor but `ctx.last_pong` is updated on every received frame here, not just on pongs. so the heartbeat thinks a chatty client is alive even if they never pong (which is probably fine for keep-alive purposes).

but the variable name is misleading — its actually "last received frame". if someone reads the heartbeat code expecting `last_pong` to track real ws-level pongs they'll get confused. rename to `last_seen` / `last_recv_at`?

also means a client that holds the socket but only sends app data never gets idle-killed even if the underlying tcp is dead. only real pongs prove the path back is alive.

jac-scale/jac_scale/impl/serve.endpoints.impl.jac
```diff
@@ -463,111 +481,566 @@ impl JacAPIServerEndpoints.register_webhook_endpoints -> None {
...
+    func_obj = self.get_functions()[func_name];
+    restspec = func_obj.restspec if func_obj?.restspec and func_obj.restspec else None;
+    is_broadcast = restspec.broadcast if restspec and restspec?.broadcast else False;
+    is_streaming = restspec.streaming if restspec and restspec?.streaming else False;
+    if is_broadcast and not is_streaming {
+        logger.warning(
+            f"Function WS '{func_name}' has broadcast=True without "
+            f"streaming=True; broadcast ignored for RPC-style function endpoints."
+        );
+    }
```
MusabMahmoodh: this is exactly the kind of misconfiguration that should fail loud at registration time imo. `broadcast=True` + `streaming=False` on a function ws is meaningless — the user typed `broadcast=True` because they wanted broadcasting, the warning gets buried in 1000s of log lines and they ship a feature that silently doesnt broadcast.

raise at endpoint registration instead? something like `raise ValueError(f"function ws '{func_name}': broadcast=True requires streaming=True")`. dev finds out before deploy.

jac-scale/jac_scale/ws_backplane.jac
```diff
@@ -0,0 +1,232 @@
...
+
+    async def subscribe(channel: str, handler: Any) {
+        lock = self._get_lock();
+        await lock.acquire();
+        try {
+            if channel not in self._handlers {
+                self._handlers[channel] = [];
+                await self._pubsub.subscribe(channel);
+            }
+            self._handlers[channel].append(handler);
```
MusabMahmoodh, line 173: small but real race: if `await self._pubsub.subscribe(channel)` raises (transient redis blip), we've already done `self._handlers[channel] = []`. next call to `subscribe(channel, ...)` sees `channel in self._handlers` → True → skips the redis subscribe entirely. local dict says "subscribed" but redis isnt, and the handler never fires.

fix: do `await self._pubsub.subscribe(channel)` first, then `self._handlers[channel] = []` only on success. or roll back the dict insert in an except.

jac-scale/jac_scale/ws_backplane.jac
```diff
@@ -0,0 +1,232 @@
...
+            lock.release();
+        }
+    }
+
+    async def unsubscribe(channel: str) {
+        lock = self._get_lock();
+        await lock.acquire();
+        try {
+            if channel in self._handlers {
+                del self._handlers[channel];
```
MusabMahmoodh: `unsubscribe(channel)` blows away all handlers for a channel rather than removing a specific one. fine for the current single-manager use case, but the interface is asymmetric — `subscribe(channel, handler)` takes a handler, `unsubscribe(channel)` doesnt.

if anything else in the system ever subscribes to the same channel, one component calling unsubscribe takes everyone else down. should the signature be `unsubscribe(channel, handler)` with channel-level redis-unsubscribe only when `len(handlers) == 0`? same applies to MemoryBackplane on line 50.

jac-scale/jac_scale/websocket.jac
```diff
@@ -1,54 +1,306 @@
...
 obj WebSocketConnectionManager {
-    has _active_connections: dict[str, list[WebSocket]] = {};
+    has _by_target: dict[str, dict[str, ConnectionContext]] = {},
+        _by_user: dict[str, int] = {},
+        _lock: (Lock | None) = None,
+        max_per_target: int = 5000,
+        max_per_user: int = 10,
+        send_timeout: float = 5.0,
+        backplane: (BaseBackplane | None) = None,
+        _subscribed: set[str] = set(),
```
MusabMahmoodh, line 60: same mutable-default trap as flagged on the scheduler PR (#5674). `_subscribed: set[str] = set()`, `_by_target: dict = {}`, `_by_user: dict = {}` — all shared across instances by default.

in prod the singleton hides it because there's only one manager, but tests that construct fresh `WebSocketConnectionManager()` instances will inherit whatever state another instance left in there. either init these in `postinit` or make them `None` and lazy-init.

jac-scale/jac_scale/impl/serve.endpoints.impl.jac
```diff
@@ -463,111 +481,566 @@ impl JacAPIServerEndpoints.register_webhook_endpoints -> None {
...
         if protocol == APIProtocol.WEBSOCKET {
             handler = self.create_websocket_handler(walker_name);
+            self.server.app.websocket(f"/ws/walker/{walker_name}")(handler);
+            # Legacy alias; deprecate after clients migrate.
             self.server.app.websocket(f"/ws/{walker_name}")(handler);
-            logger.info(f"Registered WebSocket endpoint: /ws/{walker_name}");
+            logger.info(
+                f"Registered WebSocket walker endpoint: /ws/walker/{walker_name} "
+                f"(+ legacy /ws/{walker_name})"
+            );
```
MusabMahmoodh, line 1244: nit: the new canonical `/ws/walker/{name}` route is registered first but every test in test_serve.jac still uses the legacy `/ws/{name}` form. the comment says "deprecate after one release" but if the deprecation isnt tested on the canonical path, removing the legacy alias later is going to silently break things.

at least one test should hit `/ws/walker/{name}` so we know it actually works before the legacy goes away.

jac-scale/jac_scale/impl/config_loader.impl.jac
kashmithnisakya: You're right that the warning in `MemoryBackplane.start` referenced a `[plugins.scale.websocket].backplane` knob that wasn't actually wired, even though the release note only described the DB auto-detect. Pushed a fix adding an explicit `[plugins.scale.websocket].backplane = 'memory' | 'redis'` selector with optional WS-section `redis_url`; precedence is explicit value, then WS `redis_url`, then DB `redis_url`, so `backplane = 'memory'` now forces in-process even when a DB Redis URL exists, and existing deployments are unchanged.

jac-scale/jac_scale/websocket.jac
kashmithnisakya: valid concern, 5000 anon was too loose. added a separate `max_anonymous_per_target` cap (default 100) that only applies when there's no `user_id`, and wired it through `[plugins.scale.websocket]` so it's tunable per deploy. release note already mentions it for `pub` endpoints.

jac-scale/jac_scale/impl/serve.endpoints.impl.jac
kashmithnisakya: moved the query-token branch ahead of `websocket.accept()`, so an invalid `?token=...` now lands as a pre-handshake reject (HTTP 403 with close code 4401) instead of a post-accept close. first-frame auth still runs post-accept when no query token is supplied, since that path needs an open socket.

jac-scale/jac_scale/impl/serve.endpoints.impl.jac
kashmithnisakya: renamed to `last_seen` since it tracks any received frame, not just pongs.

jac-scale/jac_scale/impl/serve.endpoints.impl.jac
kashmithnisakya: swapped the warning for a `raise ValueError` at function ws registration so the misconfig surfaces at boot rather than in prod.

jac-scale/jac_scale/ws_backplane.jac
kashmithnisakya: swapped the order: `pubsub.subscribe(channel)` runs first, and `_handlers[channel] = []` is only set after it succeeds, so a transient redis failure can't leave a stale entry that makes future callers skip the real subscribe.

jac-scale/jac_scale/ws_backplane.jac
kashmithnisakya: changed both backplanes to `unsubscribe(channel, handler)` and only call the upstream redis/dict drop once the handler list is empty, so a second subscriber can't be taken down by another component's unsubscribe.

jac-scale/jac_scale/websocket.jac
kashmithnisakya: added a postinit on `WebSocketConnectionManager` that overwrites `_by_target`, `_by_user`, `_anon_by_target`, and `_subscribed` with fresh per-instance copies, so a new manager doesn't inherit state from prior instances.

jac-scale/jac_scale/impl/serve.endpoints.impl.jac
kashmithnisakya: flipped the basic echo test to hit `/ws/walker/EchoMessage` so the canonical route has real coverage; the rest of the ws tests still use the legacy `/ws/{name}` alias to keep deprecation tracking.

jac-scale/jac_scale/impl/serve.endpoints.impl.jac
```diff
@@ -602,111 +620,618 @@ impl JacAPIServerEndpoints.register_webhook_endpoints -> None {
...
+        if detail['status'] != 'ok' {
+            await websocket.send_json(
+                _ws_error_frame(
+                    'UNAUTHORIZED', 'Token refresh failed', detail.get('reason')
+                )
+            );
+            await websocket.close(code=WS_CLOSE_UNAUTHORIZED, reason=detail['status']);
+            return 'break';
+        }
+        ctx.user_id = detail.get('user_id');
```
kugesan1105, line 866: **Refresh can switch identity and corrupt the connection counters.**

This accepts *any* valid token and overwrites `ctx.user_id` with it — but it never checks the token belongs to the same user who opened the connection. The connection was counted under the original identity at register time; unregister later decrements whatever `ctx.user_id` has become.

So if A connects and refreshes with B's token, A's `_by_user` count never gets decremented (slow cap exhaustion — A eventually can't connect at all), and B gets a phantom `-1`. On public endpoints it's worse: an anonymous connection that refreshes into a user was counted in `_anon_by_target` but is unregistered via `_by_user`, corrupting both. It also lets a live connection change *who it is* mid-session.

A refresh is only meant to extend your *own* session with a newer token, so one guard fixes both the accounting and the identity-switch:
```jac
if ctx.user_id is not None and detail.get('user_id') != ctx.user_id {
    await websocket.close(code=WS_CLOSE_UNAUTHORIZED, reason='identity_mismatch');
    return 'break';
}
```

jac-scale/jac_scale/impl/user_manager.impl.jac
```diff
@@ -235,22 +235,75 @@ impl JacScaleUserManager.get_jwt_claims(token: str) -> dict | None {
...
+        'status': 'ok',
+        'user_id': user_id,
+        'exp': decoded.get('exp'),
+        'reason': None
+    };
+}
+
+"""Thin wrapper returning just user_id if valid, None otherwise."""
+impl JacScaleUserManager.validate_jwt_token(token: str) -> str | None {
+    result = self.validate_jwt_token_detail(token);
```
kugesan1105: **This quietly adds a DB lookup to every authenticated request, not just WebSockets.**

`validate_jwt_token` is now a wrapper over `validate_jwt_token_detail`, which calls `user_exists()` — up to two storage queries (`user_exists_by_id`, then a `find_user_by_identity` fallback on a miss). The old version was deliberately signature-only, with a comment saying the existence round-trip was "unnecessary and costly." That rationale is gone, and the cost now lands on every caller: the HTTP auth dependency (`utils.jac:87`), static serving, the admin portal, and jfast — the entire authenticated surface, per request.

The `unknown_user` check is genuinely useful at WS *connect* time (long-lived sessions). It just shouldn't be on the shared per-request path. Either keep it WS-connect-only, or cache it with a short TTL — the scheduler already does exactly this (`_user_exists_cache` / `_user_exists_ttl`, `scheduler.impl.jac:287-302`).

jac-scale/jac_scale/impl/serve.endpoints.impl.jac
```diff
@@ -602,111 +620,618 @@ impl JacAPIServerEndpoints.register_webhook_endpoints -> None {
...
-                data = await websocket.receive_json();
-
-                # Handle authentication if required
-                username: (str | None) = None;
-                if requires_auth {
-                    token = data.pop('token', None)
-                    or websocket.query_params.get('token');
-                    if token {
-                        username = self.user_manager.validate_jwt_token(token);
+                if first_data is not None {
```
kugesan1105: **A pure auth handshake frame gets executed as an empty call.**

`_ws_authenticate` pops `token` and hands the rest back as `first_data`. A client that sends `{"token": "..."}` as a clean handshake — the natural reading of first-frame auth — leaves `{}`, and this `is not None` check feeds that empty dict straight into walker spawn / function execution as a phantom invocation.

Guarding on emptiness instead fixes it: `if first_data` skips the bare-auth case while clients that bundle token+args in one frame still work. Same applies to the function handler at line 1132.

kugesan1105 (commented):
A few things on the new WebSocket layer worth fixing before merge — two are WS-scoped, one reaches the whole authenticated HTTP surface. Details inline.

jac-scale/jac_scale/impl/serve.endpoints.impl.jac
kashmithnisakya: good catch, added a guard that closes with 4401 identity_mismatch when the refresh token's user_id doesn't match the connection's, so a refresh can only extend its own session and can't switch identity or corrupt the per-user/anon counters.

jac-scale/jac_scale/impl/user_manager.impl.jac
kashmithnisakya: agreed, the existence check is now gated behind a check_existence flag (default on for the ws connect/refresh path) and the validate_jwt_token http wrapper passes check_existence=false, so the per-request auth surface is signature-only again and no longer round-trips to storage.

jac-scale/jac_scale/impl/serve.endpoints.impl.jac
kashmithnisakya: fixed, both the walker and function loops now guard on truthiness instead of is not None, so a bare {token} handshake leaves an empty dict that gets skipped instead of executed as a phantom spawn/call.

kashmithnisakya:
Superseded by two smaller PRs rebuilt against the current `jac/jaclang/scale/` layout. This branch was written against the old top-level `jac-scale/` package, which has since been reorganized, so the changes have been re-done and split for independent review:

- #7264: feat(scale): WebSocket function endpoints and streaming
- #7265: feat(scale): harden WebSocket auth, connection limits and broadcast (stacked on #7264)

Closing in favor of those two.

PR #5570, opened 2026-04-16
fix: Refactoring Codes 

SandeepaHWP:
## What this PR fixes

Two separate type checker issues in \`type_evaluator.impl.jac\`:

### 1. Wrong \`has\` declaration for \`ProgramModule.hub\` (unitree.jac)

\`ProgramModule\` was assigning \`self.hub\` dynamically inside \`postinit\` without declaring it as a \`has\` field. Jac requires fields initialized in \`postinit\` to be declared explicitly with \`by postinit\`:

\`\`\`jac
has hub: dict[str, Module] by postinit;
\`\`\`

Also fixed a wrong \`dict[(str, Module)]\` syntax in the postinit impl (should be \`dict[str, Module]\`).

### 2. Type narrowing for \`PrefetchedTypes\` fields (type_evaluator.impl.jac)

All fields in \`PrefetchedTypes\` are declared as \`ClassType | None\`. Two issues arose from this:

**E1001** - \`_get_type_from_module\` returns the broad \`TypeBase\`, not \`ClassType | None\`, so assigning its result directly to \`type_var_class\`, \`param_spec_class\`, and \`type_var_tuple_class\` failed. Fixed by narrowing with \`isinstance\` before assignment.

**E1099** - Several call sites accessed \`.shared\`, \`.clone_as_instance()\`, and \`.get_class_symbol()\` on prefetch fields without None guards. Fixed by adding \`and self.prefetch.xxx\` checks before each access.


kugesan1105:
@SandeepaHWP  this should be an error!!
in jac , we are not motivating devs to use `dynamic runtime attributes`.
Every field must be explicitly declared in the `hasvar` with `by postinit`

```jac
obj Foo {
    has msg: str by postinit;
    def postinit {
        self.msg = "hello";
    }
}
```

```py
from dataclasses import dataclass, field

@dataclass
class MyClass:
    name: str
    length: int = field(init=False)
    upper: str = field(init=False)

    def __post_init__(self):
        self.length = len(self.name)
        self.upper = self.name.upper()
```

SandeepaHWP:
@kugesan1105 will cover this

PR #5587, opened 2026-04-16
fix(jsx): support {#* block comments *#} inside JSX body

### LINKED ISSUE #5563: Bug: Component comments are not stripped during render.
### Description

Expected: Comments are ignored and not included in the DOM.

Actual: Comments are rendered, breaking layout/structure.

Developer-Linus:
## What this fixes

Block comments inside JSX bodies didn't work. This:

```jsx
<div>
    <span>before</span>
    {#* this is a JSX comment *#}
    <span>after</span>
</div>
```

…would parse-error or, in earlier iterations, format-error with the comment dumped to the end of the file. Now it parses, formats, and compiles cleanly.

## Approach

The cleanest version of this fix treats a JSX comment as a real thing in the AST instead of patching around the missing one.

- The lexer recognizes `{ #* … *# }` as a single token in JSX-content mode.
- The parser builds a `JsxComment` AST node — sibling to the existing `JsxText` and `JsxExpression`.
- The formatter renders it, the codegens (Python and JS) skip it because a comment compiles to nothing.

Each layer stays in its lane: the lexer decides shape, the parser builds the tree, the formatter renders the tree.

## Why this approach over the obvious alternatives

The first attempt did everything in the formatter — when it found a stranded comment between consumed `{ }` tokens, it would re-emit them with the comment in between. It worked for the happy path but fell over on `if true { #* x *# }` because the formatter had no way to tell a JSX child from a regular block. False positives broke same-line block comments in non-JSX code.

A second alternative was to overload `CommentToken` with optional fields like `lbrace` / `rbrace` so it could double as a JSX child. That conflates two responsibilities into one type — every consumer of `CommentToken` would have to null-check every site. Same dispatch boilerplate, just hidden inside `if`/`else` chains instead of named methods.

Putting the recognition in the lexer (where JSX context is already tracked, since the lexer already handles JSX tags / text / fragments) and naming the AST node for what it is gives us the smallest possible blast radius: a typed thing that flows through the existing pass dispatch like every other JSX child.

## Decisions worth knowing

- **Bare `{}` is a parse error**, like TSX. An earlier draft accepted it incidentally, but there was no use case so we made it explicit.
- **The comment lives on the AST node**, not in the free-comments list (`source.comments`). The formatter's existing comment-injection pass never sees it, so there's no chance of a double render. One pipeline per role, no overlap.
- **No new internal design doc.** An earlier draft added one, but it wasn't wired into the website nav. Architecture rationale belongs in the PR description, not as a permanent file.

## Limitations

- `{ #* a *# #* b *# }` (two comments in one slot) is not supported. The lexer's atomic-scan pattern allows only `{ \s* #* … *# \s* }` — anything else falls back to normal `{` handling, which then tries to parse an expression and errors. Same limitation in every prior iteration. Don't write that.
- Mid-expression comments like `{count #* trailing *#}` do work, but go through the regular comment-injection pipeline (since they're attached to an expression). Output may break across lines.
- `.jir` cache files become stale on first use because the new AST node shifts type indices. One-time recompile, no user action needed.

## Test coverage

- **Formatter**: comment between siblings, as only child, adjacent to expression, no displacement, non-JSX same-line braces unaffected, JSX + non-JSX braces in the same module.
- **Lexer**: the atomic-match cases (with and without surrounding whitespace), the rewind cases (expression slot, comment+expression, two comments, non-JSX context), and the position-invariant after rewind.
- **Codegen**: JS output contains no `null` slots for JSX comments; Python output is exec'd at runtime as a sanity check.
- All three regressions from the original PR draft still pass.

## Diff size note

The PR is bigger than a typical bug fix (~27 files changed). Most of that is unavoidable: a new AST node has to register with each pass that already handles JSX children (formatter doc-IR, unparse, normalize, plus JS / Python codegens), and the JIR registry auto-regenerates with shifted type indices. The actual user-facing change is small — the bulk is mechanical pass dispatch.


docs/docs/community/release_notes/jaclang.md
```diff
@@ -2,6 +2,10 @@
 
 This document provides a summary of new features, improvements, and bug fixes in each version of **Jaclang**. For details on changes that might require updates to your existing code, please refer to the [Breaking Changes](../breaking-changes.md) page.
 
+## jaclang 0.14.1 (Unreleased)
+
+- **Type Checker: Walker `.reports` Attribute Resolution**: Accessing `.reports` on a spawned walker no longer produces a spurious `E1030` error. Walker and node archetypes now inherit from their builtin base types (`Walker`, `Node`) in the type system's MRO, so fields like `reports: list[Any]` resolve through normal inheritance. Users can also declare `has reports: list[MyType]` on a walker for compile-time type checking of `report` statements.
+
```
kugesan1105: u can remove this one

Developer-Linus: I've removed the the text in the screenshot.

kugesan1105:
@Developer-Linus this will fail during jac format; bcz formatter heuristic is too broad and regresses `jac format` on regular blocks.
```jac
def empty_block_with_comment {
    if True { #* same-line comment *# }
}

def empty_block_multi_call {
    if True { #* a *# }
    if True { #* b *# }
}

def empty_dict_set {
    x = { #* nothing *# };
}

with entry {
    empty_block_with_comment();
}
```

kugesan1105:
i am working on this imprvement @Developer-Linus

Developer-Linus:
Noted @kugesan1105. Thank you for the comment.

Thamirawaran (approved):

PR #5590, opened 2026-04-17
Add Parallel tool calling support for byllm

Hushan-10:
This pull request introduces parallel tool calling to the byllm framework, allowing multiple tool calls emitted by the LLM in a single response to be executed concurrently via a shared thread pool. The change is configurable both globally and per-call, and includes intelligent scheduling hints so the LLM can batch independent tools or sequence dependent ones. The implementation ensures consistent behavior for both sequential and parallel modes, updates streaming and batch execution logic, and adds metrics for tool and LLM call durations.

**Parallel tool calling and configuration:**

* Added support for parallel tool calling, allowing multiple tool calls in one LLM response to be executed concurrently using a shared thread pool. This can be enabled globally via `jac.toml` or the `BYLLM_PARALLEL_TOOL_CALLING` environment variable, or per-call with `by llm(parallelize=True)`. The default remains sequential for backward compatibility. 

* Added the `_should_parallelize` function to determine whether parallel execution should be used for a given call, respecting both per-call and global configuration.

**Tool call batching, scheduling, and LLM prompt hints:**

* Updated tool call execution logic in `BaseLLM.invoke` and streaming methods to route all tool calls through a unified `dispatch_batch` function, supporting both parallel and sequential execution. Tool calls are split into batches, and results are collected and added to the message history in a deterministic order. 

* When parallel execution is enabled, tool descriptions are annotated with scheduling metadata, and the system prompt is injected with intelligent scheduling rules. This helps the LLM decide which tools can be batched and which must be executed sequentially, based on dependencies and side effects.

**Streaming and metrics:**

* Enhanced streaming execution to emit timing metrics for both LLM calls and individual tool calls, including event streams for tool call initiation, results, and durations. 

**Testing and mocks:**

* Updated the mock LLM to support batch tool calls, allowing tests to simulate multiple tool calls in one response.
image


jac-byllm/byllm/parallel.jac
```diff
@@ -0,0 +1,259 @@
...
+        }
+    } except Exception as e {
+        _plog.warning(
+            "batch fallback=sequential reason=pool_error size=%d error=%s", n, e
+        );
+        for tc in tool_calls {
+            yield _run_one(tc);
+        }
+        return;
+    }
```
udithishanka: this is the same double-execution bug copilot flagged on the previous HEAD. by the time the outer except fires, some futures have already completed and their results have been yielded — the tools ran. the fallback loop then re-runs the entire batch via `_run_one(tc)`, so non-idempotent tools execute twice. the per-future try/except inside `as_completed` already handles per-tool errors; drop the outer fallback entirely.

```suggestion
    futures: dict = {};
    for tc in tool_calls {
        f = _WORKER_POOL.submit(_run_one, tc);
        futures[f] = tc;
    }
    for f in as_completed(futures) {
        try {
            yield f.result();
        } except Exception as e {
            tc_err = futures[f];
            err_msg = ToolCallResultMsg(
                content=f"Parallel dispatch error: {e}",
                tool_call_id=tc_err.call_id,
                name=tc_err.tool.get_name()
            );
            yield (tc_err, err_msg, 0);
        }
    }

    wall_ms = int((time.time() - t_batch) * 1000);
    _plog.info(
        "batch done=parallel size=%d workers=%d wall_ms=%d", n, workers, wall_ms
    );
```

jac-byllm/byllm/parallel.jac
```diff
@@ -0,0 +1,259 @@
...
+            return f.result(timeout=timeout);
+        } except Exception as e {
+            f.cancel();
+            raise TimeoutError(
+                f"Tool '{tc.tool.get_name()}' timed out after {timeout}s"
+            ) from e;
+        }
+    }
+    return tc();
+}
```
udithishanka, line 123: pool-within-pool deadlock. `dispatch_batch` submits `_run_one` wrappers to `_WORKER_POOL` (L229). each wrapper calls `_call_sync`, which submits `tc` *back to the same pool* (L101) and blocks on `f.result(timeout=...)`. with N tools all using `timeout_sec`, N outer workers occupy every slot waiting on N nested submissions that can't schedule. hard deadlock.

swap to `threading.Thread(...).join(timeout=...)` so the timeout wait doesn't require a pool slot:

```suggestion
def _call_sync(tc: Any, timeout: Any) -> Any {
    if timeout is None {
        return tc();
    }
    import threading;
    result_box: list = [None];
    error_box: list = [None];
    def _worker -> None {
        try {
            result_box[0] = tc();
        } except Exception as e {
            error_box[0] = e;
        }
    }
    t = threading.Thread(target=_worker, daemon=True);
    t.start();
    t.join(timeout=timeout);
    if t.is_alive() {
        raise TimeoutError(
            f"Tool '{tc.tool.get_name()}' timed out after {timeout}s"
        );
    }
    if error_box[0] is not None {
        raise error_box[0];
    }
    return result_box[0];
}
```

jac-byllm/byllm/parallel.jac
```diff
@@ -0,0 +1,259 @@
...
+    if timeout is not None {
+        f = _WORKER_POOL.submit(tc);
+        try {
+            return f.result(timeout=timeout);
+        } except Exception as e {
+            f.cancel();
+            raise TimeoutError(
+                f"Tool '{tc.tool.get_name()}' timed out after {timeout}s"
+            ) from e;
+        }
```
udithishanka, line 111: this catches *every* exception from `f.result(timeout=timeout)` and re-raises as `TimeoutError`. if the tool itself raises ValueError, the user sees `Tool 'X' timed out after 5s` — completely wrong diagnosis. catch only `concurrent.futures.TimeoutError`.

(obsoleted if you adopt the threading suggestion above — noting it separately in case you keep the pool-based version.)

jac-byllm/byllm/parallel.jac
```diff
@@ -0,0 +1,259 @@
...
+        if timeout is not None {
+            return await asyncio.wait_for(coro, timeout=timeout);
+        }
+        return await coro;
+    }
+    result_val = asyncio.run(_run());
+    return ToolCallResultMsg(
+        content=str(result_val), tool_call_id=tc.call_id, name=tc.tool.get_name()
+    );
+}
```
udithishanka, line 139: two things here. (1) `asyncio.run()` raises RuntimeError if the calling thread already has a running event loop — possible in some jac runtime contexts. (2) `asyncio.TimeoutError` from `wait_for` gets caught by the generic `except Exception` in `_run_one` (L145) and reported as `Tool error: ...` instead of the timeout-specific message the sync path produces. inconsistent UX.

min fix: in `_run_one`, handle `TimeoutError` / `asyncio.TimeoutError` before the generic catch and format with the same `timed out after Xs` shape as `_call_sync`.

jac-byllm/byllm/llm.impl/basellm.impl.jac
```diff
@@ -134,22 +134,46 @@ impl BaseLLM.invoke(mt_run: MTRuntime) -> object {
...
+            finish_calls = [
+                tc
+                for tc in resp.tool_calls
+                if tc.is_finish_call()
+            ];
+
+            if non_finish_calls {
+                # Collect results from dispatch_batch generator, then
+                # add messages in input order for deterministic history.
+                import from byllm.parallel { dispatch_batch, _index_by_identity }
```
udithishanka: nit: this re-imports on every ReAct iteration. move to the top of the file with the other imports. same story at L717 in the streaming path.

jac-byllm/byllm/llm.impl/basellm.impl.jac
```diff
@@ -200,6 +224,23 @@ impl BaseLLM.make_model_params(mt_run: MTRuntime) -> dict {
...
+    and len(messages) > 0 {
+        first_msg = messages[0];
+        if isinstance(first_msg, dict) and first_msg.get("role") == "system" {
+            import from byllm.parallel { _PARALLEL_TOOL_HINT }
+            content = first_msg.get("content", "");
+            if isinstance(content, str) and _PARALLEL_TOOL_HINT not in content {
+                first_msg["content"] = content + _PARALLEL_TOOL_HINT;
+            }
+        }
+    }
```
udithishanka, line 260: `first_msg["content"] = content + _PARALLEL_TOOL_HINT` mutates the dict in place. if `mt_run.get_msg_list()` returns a reference (it does — it's the persistent message list), you're mutating shared state from inside a param-building method. the `not in content` guard prevents double-append in the same call, but this still means the hint persists into `mt_run.messages` and leaks across iterations. copy the message before appending:

```suggestion
        if isinstance(first_msg, dict) and first_msg.get("role") == "system" {
            import from byllm.parallel { _PARALLEL_TOOL_HINT }
            content = first_msg.get("content", "");
            if isinstance(content, str) and _PARALLEL_TOOL_HINT not in content {
                messages = list(messages);
                messages[0] = {**first_msg, "content": content + _PARALLEL_TOOL_HINT};
            }
        }
```

jac-byllm/tests/test_parallel.jac
```diff
@@ -0,0 +1,441 @@
...
+    assert "tool_scheduler" not in content;
+    assert "_dispatch_tool_batch" not in content;
+}
+
+test "basellm uses dispatch batch not old api" {
+    src = pathlib.Path(_ROOT_DIR) / "byllm" / "llm.impl" / "basellm.impl.jac";
+    content = src.read_text();
+    assert "dispatch_batch" in content;
+    assert "_dispatch_tool_batch" not in content;
+}
```
udithishanka: tests 12, 14 (second), 15-20 are source-text assertions (`pathlib.read_text()` + `"X" in content`). these pass even if imports are broken or functions are stubbed — they validate string presence in source, not runtime behavior. e.g. test 17 checks `"_serialize" in content` of `tool.impl.jac`, but if `Tool.postinit` never actually reads it, the test still passes.

replace with actual imports + runtime calls. for example, test 17 should construct a tool with `_serialize=True` on the func and assert `Tool(func).serialize == True`. the behavioral tests (1-11, 13, 14-first) are fine — keep those.

udithishanka (changes requested):
3 correctness bugs and a few important issues.

jac-byllm/byllm/parallel.jac
Hushan-10: Fixed. removed the outer try/except fallback. The per-future try/except inside as_completed already handles errors per-tool; the outer block caused double-execution of non-idempotent tools.

jac-byllm/byllm/parallel.jac
Hushan-10: Fixed.  replaced _WORKER_POOL.submit with threading.Thread + join(timeout=...) to avoid pool-within-pool deadlock when all workers have timeout_sec set.

jac-byllm/byllm/parallel.jac
Hushan-10: Already resolved by the threading.Thread fix from comment 2 . _call_sync no longer catches Exception broadly. TimeoutError is only raised when t.is_alive() after join; tool-raised exceptions propagate with their original type.

jac-byllm/byllm/parallel.jac
Hushan-10: Fixed. added except (TimeoutError, asyncio.TimeoutError) as a dedicated handler in _run_one before the generic catch. Both sync and async timeout paths now produce the same "Tool 'X' timed out after Ys" message.

jac-byllm/byllm/llm.impl/basellm.impl.jac
Hushan-10: Fixed. moved all three inline imports to a single top-level import from byllm.parallel in llm.jac. No more per-iteration re-imports.

jac-byllm/byllm/llm.impl/basellm.impl.jac
Hushan-10: Fixed. make_model_params now copies the list and creates a new dict before appending the hint: messages = list(messages); messages[0] = {**first_msg, "content": ...}. Original mt_run.messages is never mutated.

jac-byllm/tests/test_parallel.jac
Hushan-10: Fixed. replaced all 10 source-text assertions with runtime behavioral tests. Tests now import real modules (Tool, MockToolCall, BaseLLM, _WORKER_POOL, _PARALLEL_TOOL_HINT) and exercise actual behavior (construct objects, call functions, verify attributes)

jac-byllm/byllm/llm.impl/basellm.impl.jac
```diff
@@ -134,20 +149,61 @@ impl BaseLLM.invoke(mt_run: MTRuntime) -> object {
...
+                        }
+                        return tool_call.get_output();
+                    } else {
+                        result_msg = tool_call();
+                        last_tool = tool_call.tool.get_name();
+                        last_result = str(result_msg.content)[:500];
+                        mt_run.add_message(result_msg);
+                    }
                 }
             }
```
udithishanka, line 195: two things on this branching.

first, this is ~60 lines of parallel path and ~15 lines of sequential path that diverge in non-trivial ways but both consume the same `resp.tool_calls`. every future fix needs mirroring in both branches (and again at L773+ in the streaming path). can we always go through `dispatch_batch` and have the sequential mode just be a 1-at-a-time degenerate case? one code path, same observable behavior, no drift.

second, the two branches don't produce the same history for `resp.tool_calls = [A, finish, B]`. the sequential branch returns after `finish` (B is skipped, never added to mt_run). the parallel branch runs A + B concurrently, then returns finish (B's result IS added). same LLM response → different `mt_run.messages` depending on the mode flag. rare shape in practice but worth normalizing — either both run B or both skip it.

jac-byllm/byllm/llm.impl/basellm.impl.jac
```diff
@@ -646,45 +863,61 @@ impl BaseLLM._invoke_streaming(
...
+                                    "iteration": iter_count
+                                }
+                            );
+                            mt_run.add_message(result_msg);
+                        }
+                    }
+                    # If we broke out of the for loop due to finish_tool, break the while loop too
+                    if tool_call.is_finish_call() {
+                        break;
                     }
```
udithishanka: same duplication story here — ~90 lines parallel and ~50 lines sequential doing the same job. any streaming-event change (new event type, ordering tweak, truncation rule) has to land in both blocks. if the scheduler unification from the L151 comment happens, this collapses with it.

jac-byllm/tests/test_parallel.jac
```diff
@@ -0,0 +1,469 @@
...
+    assert not hasattr(BaseLLM, 'tool_scheduler');
+    assert not hasattr(BaseLLM, '_dispatch_tool_batch');
+}
+
+test "dispatch batch is callable and importable" {
+    assert callable(dispatch_batch);
+    # Verify it returns a generator for an empty batch
+    gen = dispatch_batch([]);
+    assert list(gen) == [];
+}
```
udithishanka, line 476: no coverage for any of the new gate logic introduced in the last few commits. specifically missing:

- `_should_parallelize` default (parallel OFF unless opted in)
- `BYLLM_PARALLEL_TOOL_CALLING` env var turns it on
- per-call `call_params={"parallelize": True}` overrides the config
- `make_model_params` appends `[PARALLEL-SAFE: ...]` / `[SEQUENTIAL: ...]` to tool descriptions when parallel is enabled
- `finish_tool`'s description is NOT tagged
- the "Intelligent tool scheduling rules" block is injected into the system prompt only when parallel+tools+`parallel_hint` are all true

all of these are user-facing switches; if any of them silently breaks, users just get mysterious slow / fast runs. worth a few focused behavioral tests (MockLLM + a captured `make_model_params` output is enough for the annotation + hint assertions).

udithishanka (changes requested):
2 maintenance issues, missing tests for new features, and 2 red CI checks (jac-check, Contribution Checks / release notes).

jac-byllm/byllm/llm.impl/basellm.impl.jac
Hushan-10: Unified both paths. parallel and sequential now share a single code path through dispatch_batch. The mode flag only controls batch granularity: full batch (parallel) vs one-at-a-time (sequential, inline via size=1). History is normalized by splitting at the first finish_tool, so [A, finish, B] produces identical mt_run.messages in both modes. Same fix applied to the streaming path.

jac-byllm/tests/test_parallel.jac
Hushan-10: Added 10 focused behavioral tests covering all listed gaps: _should_parallelize default/env/per-call gates (4 tests with MockLLM + timing), make_model_params tool description annotations including finish_tool exclusion (2 tests), and scheduling hint injection gating including parallel_hint=False opt-out and serialize tool name listing (4 tests)

jac-byllm/byllm/llm.impl/basellm.impl.jac
Hushan-10: fixed

MalithaPrabhashana:
Update the docs to show how to enable parallel tool calling and add a clean description on this PR @Hushan-10 @Hirudika4

Jayanaka-98 (approved):

udithishanka (approved):

PR #5600, opened 2026-04-17
ci: block new Python files via PR check

SandeepaHWP:
## Summary

Adds a GitHub Actions workflow that fails any PR introducing new `.py` files.

**What it does:**
- Runs on every pull request (open, update, reopen, or label added)
- Checks if any new `.py` files are being added (`--diff-filter=A` catches additions only)
- Fails with a clear error message if any are found

**To skip this check:**
Add the `skip-no-python-check` label to your PR. The check will re-run automatically when the label is added and will pass.

And 3 quick parallel jobs were merged into one for better readability of checks.


kugesan1105:
add a label to ignore this check!
also move this to a part of some existing Jobs

kugesan1105:
combine them to single sub jobs
image

kugesan1105 (approved):
LGTM

PR #5604, opened 2026-04-17
[Feat] Implement per-user LLM rate limiting and update documentation

Jayanaka-98:
## Per-User LLM Rate Limiting and Budget Enforcement

Adds configurable per-user rate limits and spend caps on all `by llm()` calls in jac-scale. When a user exceeds any configured limit, the call is blocked before it reaches the LLM provider.

---

### Motivation

Without per-user limits, a single user can exhaust the entire LLM budget of a deployment, either accidentally (runaway loops) or deliberately. This PR adds hard guardrails at the platform level with no changes required in application code.

---

### What changed

**`jac-byllm`**
- Added `register_pre_invoke_hook` and `register_post_invoke_hook` to `Model.invoke` — synchronous hooks that fire before and after every `by llm()` call. Pre-hooks can raise to block the call.

**`jac-scale`**
- New `LLMRateLimitExceeded(Exception)` class in `llm_telemetry.jac`
- New `JacUserRateLimiter` class: pure logic, no litellm dependency. Checks RPM, RPD, TPM, TPD, daily budget, monthly budget against Redis counters and MongoDB aggregates
- `register_llm_telemetry_endpoints` wires the rate limiter into byllm's pre-invoke hook for blocking, and into litellm's `success_callback` for cost/token tracking
- Daily and monthly budgets survive Redis restarts: on cache miss, rebuilt from MongoDB aggregation and re-cached for 60 seconds
- `JScaleExecutionContext` gains a `username` field, set by the request middleware from the JWT token
- Unauthenticated requests (no JWT, background jobs, scheduled tasks) are pooled under `__anonymous__` and subject to the same limits — no silent bypass
- New `[plugins.scale.llm_limits]` config section in `jac.toml`

**Fixes from review:**
- Redis-down now blocks calls (raises `LLMRateLimitExceeded`) instead of silently skipping limits
- `nx=True` on budget cache `set` prevents a race where a concurrent `record_success` increment gets overwritten
- Daily budget Redis key and MongoDB aggregate both use UTC date (was mixing server local TZ with UTC)
- Exception caught by type, not string matching
- Discovered `log_pre_api_call` is non-blocking in LiteLLM (exceptions are swallowed). Replaced with byllm pre-invoke hook which actually blocks

---

### Configuration

```toml
[plugins.scale.llm_limits]
enabled = true
rpm  = 60
rpd  = 1000
tpm  = 100000
tpd  = 500000
daily_budget_usd   = 5.00
monthly_budget_usd = 50.00
```

All fields optional. Omit any to leave that dimension unlimited. If `enabled = false` or the section is absent, no limiting occurs. Requires `jac-scale[data]` (Redis + MongoDB).

---

### Design notes

- **No application code changes required.** Limits are enforced at the infrastructure layer.
- **Unauthenticated requests are not silently bypassed.** Calls without a username (public endpoints, background jobs) are pooled under `__anonymous__` and rate-limited as a single user.
- **Token limits are best-effort pre-call.** The check reads the accumulated counter before the call. The first call that crosses the threshold goes through.
- **byllm is a test dependency only.** Production code guards all byllm imports with `try/except ImportError`.

---

### Tests

17 tests across four tiers:
- Unit (mock Redis/MongoDB): each limit type blocks, Redis-down behavior, budget cache hit/miss, MongoDB write shape, anonymous pooling
- Integration (real Redis): RPM per-user isolation, TPM accumulation, RPD day-rollover
- Persistence (real Redis + MongoDB): daily budget rebuild after flush, monthly budget multi-day aggregation
- End-to-end hook (real byllm `Model.invoke`): confirms the pre-invoke hook actually blocks a second call when RPM is exceeded


jac-scale/jac_scale/admin/impl/llm_telemetry.impl.jac
```diff
@@ -19,6 +19,179 @@ import json as json_module;
...
+        self.rc.set(cache_key, spent, ex=ttl);
+    }
+    return spent;
+}
+
+"""Check all configured limits for a user before an LLM call. Raises on violation."""
+impl JacUserRateLimiter.check_pre_call(
+    self: JacUserRateLimiter, username: str
+) -> None {
+    if not self.rc {
```
MusabMahmoodh: If Redis is  unavailable at startup or flaky  due to some reasons , every configured limit silently disappears no?. so all of the limits will be skipped?? better if we reject all the LLM calls unitl Redis is back or have some faalbacks to preserve the limtis?

jac-scale/jac_scale/admin/impl/llm_telemetry.impl.jac
```diff
@@ -19,6 +19,179 @@ import json as json_module;
...
+        self.col.aggregate(
+            [
+                {'$match': {'username': username, 'ts': {'$gte': since}}},
+                {'$group': {'_id': None, 'total': {'$sum': '$cost_usd'}}}
+            ]
+        )
+    );
+    spent = float(result[0]['total']) if result else 0.0;
+    if self.rc {
+        self.rc.set(cache_key, spent, ex=ttl);
```
MusabMahmoodh: Same Redis key is used for two things here. In record_success it is a live counter (incrbyfloat). Here on cache miss it is overwritten with the Mongo aggregate total (set). If a record_success call land between our cache-miss read and our set, that increment is lost.

Can we use nx=True on the set? self.rc.set(cache_key, spent, ex=ttl, nx=True). That way the set only happen when the key is still empty. If another call already populate the counter, we leave it alone.

jac-scale/jac_scale/admin/impl/llm_telemetry.impl.jac
```diff
@@ -19,6 +19,179 @@ import json as json_module;
...
+        }
+    }
+    if self.tpd {
+        current = int(self.rc.get(f"llm:tpd:{username}:{ts // 86400}") or 0);
+        if current >= int(self.tpd) {
+            raise Exception("LLM token-per-day limit exceeded. Retry tomorrow.");
+        }
+    }
+    if self.daily_budget {
+        today = str(date.today());
```
MusabMahmoodh: date.today() use the server local timezone, but _get_budget_spent compute the day boundary in UTC via datetime.now(timezone.utc).replace(hour=0, ...). So the Redis key say "today in local time" but the Mongo rebuild sum records from "today in UTC". If the server TZ is not UTC, they point at different days.

Example: server run with TZ=America/New_York. At 8 PM local on April 22 (already 01:00 UTC April 23) the key is llm:budget_daily:alice:2026-04-22 but the Mongo aggregate only include records after 00:00 UTC April 23. Most of alice spend is missed.

jac-scale/jac_scale/admin/impl/llm_telemetry.impl.jac
```diff
@@ -19,6 +19,179 @@ import json as json_module;
...
+            if self.tpd {
+                pipe = self.rc.pipeline();
+                pipe.incrby(f"llm:tpd:{username}:{ts_int // 86400}", tokens);
+                pipe.expire(f"llm:tpd:{username}:{ts_int // 86400}", 172800);
+                pipe.execute();
+            }
+        }
+        if cost_usd > 0 {
+            if self.daily_budget {
+                today = str(date.today());
```
MusabMahmoodh: same issue as line 120, need UTC here also

jac-scale/jac_scale/admin/impl/llm_telemetry.impl.jac
```diff
@@ -672,6 +845,113 @@ impl JacAPIServerLLMTelemetry.register_llm_telemetry_endpoints -> None {
...
+                    kwargs: Any
+                ) {
+                    try {
+                        ctx = JacRuntimeInterface.get_context();
+                        username = getattr(ctx, 'username', None);
+                        if username {
+                            self.rl.check_pre_call(username);
+                        }
+                    } except Exception as e {
+                        if 'limit exceeded' in str(e).lower()
```
MusabMahmoodh: This is bad, matching on exception message text only. If someone reword the message in check_pre_call, the 'limit exceeded' match will silently fail and limiting stop working, no test will catch this.

Also any unrelated error that happen to have "budget" in message (Mongo pool error for example) will be wrongly re-raised as rate limit.

Better to make a proper LLMRateLimitExceeded(Exception) class and catch by type, no? Then no string match needed.

MusabMahmoodh (changes requested):

jac-scale/jac_scale/admin/impl/llm_telemetry.impl.jac
Jayanaka-98: Fixed. Added `nx=True` to `self.rc.set(cache_key, spent, ex=ttl, nx=True)`. The aggregate write only lands when the key is still absent. If record_success is already populated between the cache-miss read and the set, the existing value is left untouched.

jac-scale/jac_scale/admin/impl/llm_telemetry.impl.jac
Jayanaka-98: Fixed, and then some. Introduced LLMRateLimitExceeded(Exception) in llm_telemetry.jac and all limit violations raise it by type.

The original string matching wasn't arbitrary. It was there because log_pre_api_call doesn't propagate exceptions cleanly. LiteLLM wraps every callback invocation in a broad try/except Exception at line 1086 of litellm_logging.py (explicitly marked [Non-Blocking]) and swallows all exceptions after logging them. The string match was an attempt to detect rate limit errors in an outer catch and re-raise from there. But that outer catch is also swallowed, so it never actually blocked anything.

The real fix was to stop using LiteLLM callbacks for blocking entirely. Replaced with byllm's new register_pre_invoke_hook mechanism, which fires synchronously inside Model.invoke before the LLM call and propagates exceptions correctly to the caller. The hook catches except LLMRateLimitExceeded { raise; } by type — no string matching, and blocking actually works end-to-end (verified by a new integration test that calls Model.invoke twice with rpm=1 and asserts the second call raises).

jac-scale/jac_scale/admin/impl/llm_telemetry.impl.jac
Jayanaka-98: Fixed. Replaced date.today() with time.strftime('%Y-%m-%d', time.gmtime(ts)) in check_pre_call, ensuring the Redis key always uses the UTC date, consistent with the Mongo day boundary.

jac-scale/jac_scale/admin/impl/llm_telemetry.impl.jac
Jayanaka-98: Fixed. Same change applied in record_success. Daily budget key now derived from time.strftime('%Y-%m-%d', time.gmtime(ts_int)) instead of date.today().

jac-scale/jac_scale/admin/impl/llm_telemetry.impl.jac
Jayanaka-98: Fixed. check_pre_call now raises `LLMRateLimitExceeded("LLM rate limiting is configured but Redis is unavailable.")` when limits are configured but Redis is None. No silent bypass.

jac-scale/jac_scale/admin/impl/llm_telemetry.impl.jac
```diff
@@ -20,6 +20,185 @@ import json as json_module;
...
+        self.col.aggregate(
+            [
+                {'$match': {'username': username, 'ts': {'$gte': since}}},
+                {'$group': {'_id': None, 'total': {'$sum': '$cost_usd'}}}
+            ]
+        )
+    );
+    spent = float(result[0]['total']) if result else 0.0;
+    if self.rc {
+        self.rc.set(cache_key, spent, ex=ttl, nx=True);
```
MusabMahmoodh, line 69: nx=true fixed the previous race where record_success increment was lost, but introduces the opposite race no? if get returns None and a concurrent record_success runs between our get and our set, the set fails because key now exists, so we lose the entire mongo aggregate. next check_pre_call sees the small live counter (like 0.50) instead of historical spend (5.00) and the undercount sticks until ttl expires which is 25h since record_success extends to 90000s. happens at startup or after redis flush mainly. proper fix probably needs incrbyfloat with the aggregate delta, or a small lua script that does populate-if-empty atomically

jac-scale/jac_scale/admin/impl/llm_telemetry.impl.jac
```diff
@@ -20,6 +20,185 @@ import json as json_module;
...
+    }
+    if self.rpd {
+        self._check_and_incr(
+            f"llm:rpd:{username}:{ts // 86400}", int(self.rpd), 172800
+        );
+    }
+    if self.tpm {
+        _raw = self.rc.get(f"llm:tpm:{username}:{min_bucket}");
+        current = int(_raw.decode() if isinstance(_raw, bytes) else (_raw or 0));
+        if current >= int(self.tpm) {
```
MusabMahmoodh, line 106: this is check-then-act, two concurrent calls can both read current=999 with limit 1000 and both pass, both go to llm, both record_success, ends up exceeding. pr description acknowledges this as best effort but worth a one liner comment here so future readers know the race is intentional and not a bug

jac-scale/jac_scale/admin/impl/llm_telemetry.impl.jac
```diff
@@ -20,6 +20,185 @@ import json as json_module;
 glob logger = logging.getLogger(__name__);
 
 
+"""Atomically increment a Redis counter and raise if the limit is exceeded."""
+impl JacUserRateLimiter._check_and_incr(key: str, limit: int, ttl: int) -> None {
+    count = self.rc.incr(key);
```
MusabMahmoodh, line 25: incr then expire as two seperate calls, if process dies between them the key lives forever with no ttl, very small window but real, easier to wrap in a pipeline (pipe.incr + pipe.expire + pipe.execute) so both commands always go out together

jac-scale/jac_scale/admin/impl/llm_telemetry.impl.jac
```diff
@@ -571,6 +750,108 @@ obj TelemetryStore {
...
+        import sys as _sys;
+        _llm_mod = _sys.modules.get('jac_scale.admin.llm_telemetry');
+        if _llm_mod {
+            setattr(_llm_mod, '_rate_limiter_instance', rl);
+        }
+
+        # Pre-call: byllm hook that blocks before the LLM call fires
+        def _pre_invoke(model_name: Any, mt_run: Any) {
+            ctx = JacRuntimeInterface.get_context();
+            username = ctx?.username or '__anonymous__';
```
MusabMahmoodh, line 811: all unauthenticated requests pool under __anonymous__ which means any single bad actor hitting a public walker can exhaust the rate limit for ALL other anonymous users, basically a free dos vector for public endpoints. pr description calls this intentional but atleast worth a doc warning, ideally fallback to per ip rate limit for anonymous so one bad ip doesnt starve everyone

jac-byllm/byllm/llm.impl/model.impl.jac
```diff
@@ -44,6 +54,11 @@ impl Model.invoke(mt_run: MTRuntime) -> object {
     _ctx_token = _current_invocation_id.set(mt_run._invocation_id);
     _start = time.time();
 
+    # Run pre-invoke hooks — any hook may raise to block the call
+    for hook in _pre_invoke_hooks {
```
MusabMahmoodh, line 58: this fires for every llm call not per user request, so an agent with tool calls (like 5 tool roundtrips for one user prompt) burns 5 against the rpm budget for that one user request, worth documenting in the rate limit config section so users dont set rpm based on user request count and get suprised when their agents get blocked after 2 prompts

jac-scale/jac_scale/admin/impl/llm_telemetry.impl.jac
```diff
@@ -672,6 +953,9 @@ impl JacAPIServerLLMTelemetry.register_llm_telemetry_endpoints -> None {
         logger.warning(f"LLM telemetry: failed to register litellm callback: {e}");
     }
 
+    # --- Register LLM per-user rate limiting ---
+    self._init_rate_limiting(store);
```
MusabMahmoodh, line 957: rate limiting init is bundled inside register_llm_telemetry_endpoints, so if someone wants rate limiting but not the admin telemetry endpoints exposed they cant get one without the other, worth splitting these into two seperate setup methods so they can be enabled independently

MusabMahmoodh (commented):
follow up review after the previous round of fixes, the nx=true / utc / exception type changes all look good. found 6 new things worth flagging, main one is a different race that nx=true introduces in the cache miss path (top comment), rest is anonymous pooling dos, per llm call vs per request semantics, and a couple of coupling/race nits.

marsninja:

PR #5630, opened 2026-04-20
Adds mobile configuration infrastructure to support mobile app builds.

mahfoos:

### DIFF
--- a/jac-client/jac_client/plugin/src/impl/mobile_config.impl.jac
+++ b/jac-client/jac_client/plugin/src/impl/mobile_config.impl.jac
@@ -1,0 +1,308 @@
+"""Implementation of Mobile target configuration loader."""
+import from pathlib { Path }
+import from typing { Any }
+import from jaclang.project.config { JacConfig, get_config }
+import from jaclang.project.plugin_config { deep_merge }
+import from jaclang.cli.console { console }
+
+"""Initialize MobileConfig with project directory."""
+impl MobileConfig.init(self: MobileConfig, project_dir: Path | None = None) -> None {
+    if project_dir is None {
+        self.project_dir = Path.cwd();
+    } else {
+        self.project_dir = project_dir;
+    }
+    self._config = None;
+    self._jac_config = None;
+}
+
+"""Get default configuration structure for mobile target."""
+impl MobileConfig.get_default_config(self: MobileConfig) -> dict[str, Any] {
+    # Try to get project name/version from JacConfig for defaults
+    project_name = "my-jac-app";
+    project_version = "1.0.0";
+    try {
+        jac_config = self._get_jac_config();
+        if jac_config and jac_config.project {
+            project_name = jac_config.project.name or project_name;
+            project_version = jac_config.project.version or project_version;
+        }
+    } except Exception {
+        console.warning("Failed to load JacConfig, using fallback defaults");
+    }
+    # Generate bundle identifier from project name
+    bundle_id = _generate_bundle_id(project_name);
+    slug = project_name.lower().replace(" ", "-");
+    scheme = project_name.lower().replace(" ", "").replace("-", "").replace("_", "");
+
+    return {
+        'name': project_name,
+        'bundle_id': bundle_id,
+        'version': project_version,
+        'display_name': project_name,
+        'slug': slug,
+        'scheme': scheme,
+        'platforms': {'ios': True, 'android': True},
+        'features': {'use_sidecar': False, 'api_url': None},
+        'expo': {
+            'orientation': 'portrait',
+            'userInterfaceStyle': 'light',
+            'icon': './assets/icon.png',
+            'splash': {
+                'image': './assets/splash.png',
+                'resizeMode': 'contain',
+                'backgroundColor': '#ffffff'
+            },
+            'ios': {
+                'supportsTablet': True
+            },
+            'android': {
+                'adaptiveIcon': {
+                    'foregroundImage': './assets/adaptive-icon.png',
+                    'backgroundColor': '#ffffff'
+                }
+            },
+            'web': {
+                'favicon': './assets/favicon.png'
+            },
+            'plugins': ['expo-router']
+        }
+    };
+}
+
+"""Get or load the core JacConfig instance."""
+impl MobileConfig._get_jac_config(self: MobileConfig) -> JacConfig {
+    if self._jac_config is None {
+        self._jac_config = get_config();
+        if self._jac_config is None {
+            raise RuntimeError(
+                "No jac.toml found. Run 'jac create' to create a project."
+            ) ;
+        }
+    }
+    return self._jac_config;
+}
+
+"""Load configuration from jac.toml, merging with defaults."""
+impl MobileConfig.load(self: MobileConfig) -> dict[str, Any] {
+    if self._config is not None {
+        return self._config;
+    }
+    default_config = self.get_default_config();
+    # Try to get [mobile] section from jac.toml
+    mobile_config = {};
+    try {
+        jac_config = self._get_jac_config();
+        # Get [mobile] section from raw TOML data
+        if jac_config and jac_config._raw_data {
+            raw_data = jac_config._raw_data;
+            mobile_config = raw_data.get('mobile', {});
+        }
+    } except Exception {
+        console.warning("Failed to load JacConfig, using fallback defaults");
+    }
+    # Deep merge defaults with user configuration
+    if mobile_config {
+        self._config = deep_merge(default_config, mobile_config);
+    } else {
+        self._config = default_config;
+    }
+    return self._config;
+}
+
+"""Get mobile configuration (name, bundle_id, version, display_name)."""
+impl MobileConfig.get_mobile_config(self: MobileConfig) -> dict[str, Any] {
+    config = self.load();
+    return {
+        'name': config.get('name', ''),
+        'bundle_id': config.get('bundle_id', ''),
+        'version': config.get('version', '1.0.0'),
+        'display_name': config.get('display_name', '')
+    };
+}
+
+"""Get platforms configuration."""
+impl MobileConfig.get_platforms_config(self: MobileConfig) -> dict[str, Any] {
+    config = self.load();
+    return config.get('platforms', {});
+}
+
+"""Get features configuration."""
+impl MobileConfig.get_features_config(self: MobileConfig) -> dict[str, Any] {
+    config = self.load();
+    return config.get('features', {});
+}
+
+"""Get app.json configuration for Expo from TOML."""
+impl MobileConfig.get_app_json_config(self: MobileConfig) -> dict[str, Any] {
+    config = self.load();
+    expo_config = config.get('expo', {});
+
+    # Get basic info
+    name = config.get('display_name', config.get('name', 'my-jac-app'));
+    slug = config.get('slug', name.lower().replace(' ', '-'));
+    scheme = config.get('scheme', name.lower().replace(' ', '').replace('-', '').replace('_', ''));
+    version = config.get('version', '1.0.0');
+    bundle_id = config.get('bundle_id', 'com.example.myapp');
+
+    # Build app.json structure
+    app_json = {
+        'expo': {
+            'name': name,
+            'slug': slug,
+            'scheme': scheme,
+            'version': version,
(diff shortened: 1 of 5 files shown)


jac.toml
```diff
@@ -10,3 +10,18 @@ exclude = [
     "console.impl.jac",
     "cli_boot.jac"
 ]
+
+
+[mobile]
```
MusabMahmoodh, line 15: why this is on the repo root?

mobile/assets/jac-app/create-bundle.js
```diff
@@ -0,0 +1,76 @@
+const fs = require('fs');
```
MusabMahmoodh, line 1: This file is committed in two places. The exact same 76 line file is also at `jac-client/jac_client/examples/all-in-one/mobile/assets/jac-app/create-bundle.js`. Repo root has no `mobile/` directory before this PR, and no precedent for a top-level `mobile/` sitting next to `jac-client/`, `jac-coder/` etc. I think this top-level copy is the mistake. Two copies will drift the moment someone edit one of them. Please remove this file and keep only the one under `examples/all-in-one/`.

mobile/assets/jac-app/create-bundle.js
```diff
@@ -0,0 +1,76 @@
...
+const path = require('path');
+
+// Read the built files
+const cssPath = path.join(__dirname, 'styles.css');
+const jsPath = path.join(__dirname, 'client.js');
+const htmlPath = path.join(__dirname, 'index.html');
+
+if (!fs.existsSync(cssPath) || !fs.existsSync(jsPath) || !fs.existsSync(htmlPath)) {
+    console.error('❌ Missing required files. Please build the web app first.');
```
MusabMahmoodh, line 10: `❌` emoji is inconsistent with repo style. Other places use `jaclang.cli.console` (`console.error/warning` with color), no emoji in messages.

mobile/assets/jac-app/create-bundle.js
```diff
@@ -0,0 +1,76 @@
...
+    process.exit(1);
+}
+
+const css = fs.readFileSync(cssPath, 'utf8');
+const js = fs.readFileSync(jsPath, 'utf8');
+const sourceHtml = fs.readFileSync(htmlPath, 'utf8');
+
+// Extract __jac_init__ config from original HTML
+const jacInitMatch = sourceHtml.match(/<script id="__jac_init__" type="application\/json">(.*?)<\/script>/);
```
MusabMahmoodh, line 19: No `s` flag on the regex. `.*?` does not cross a newline by default. If the `__jac_init__` JSON is pretty-printed across multiple lines (very common), the match fail silently and `jacInitConfig` fall back to the hardcoded `{"module": "main", "function": "app", ...}`. The generated app then run with the wrong entry point and no error is shown. Add the `s` flag, or use `[\s\S]*?`.

mobile/assets/jac-app/create-bundle.js
```diff
@@ -0,0 +1,76 @@
...
+    <div id="root"></div>
+
+    <!-- Jac initialization config -->
+    <script id="__jac_init__" type="application/json">
+    ${jacInitConfig}
+    </script>
+
+    <!-- Backend API URL - will be injected by mobile app -->
+    <script>
+        window.JAC_BACKEND_URL = window.JAC_BACKEND_URL || 'http://localhost:9000';
```
MusabMahmoodh, line 41: `localhost` on a mobile device is the device itself, not your dev machine. This bundle will ship with a backend URL that never work. The PR already add `[mobile.features].api_url` in `jac.toml`, that is what should flow through here. Today the config and this script never meet.

mobile/assets/jac-app/create-bundle.js
```diff
@@ -0,0 +1,76 @@
...
+    </script>
+
+    <!-- Jac client bundle as blob URL -->
+    <script>
+    (function() {
+        const jsCode = atob('${Buffer.from(js).toString('base64')}');
```
MusabMahmoodh, line 47: Why base64, then `atob`, then Blob, then `createObjectURL`, when the goal is a single standalone HTML? Just inline the JS as `<script type="module">${js}</script>`. Base64 pay a 33% size tax and on mobile WebView this is also a memory cost. If CSP is the reason, blob URL does not fix CSP. If it is not CSP, I do not see the benefit.

mobile/assets/jac-app/create-bundle.js
```diff
@@ -0,0 +1,76 @@
...
+</body>
+</html>`;
+
+// Write standalone HTML file
+fs.writeFileSync(path.join(__dirname, 'bundle.html'), html);
+console.log('✓ Created bundle.html');
+
+// Also create TypeScript export
+const tsContent = `// Auto-generated standalone Jac app bundle
+export const JAC_BUNDLE_HTML = \`${html.replace(/`/g, '\\`').replace(/\$/g, '\\$')}\`;
```
MusabMahmoodh, line 71: The escape for `${` inside a template literal is `\${`, not `\$`. This replace every `$` in the HTML with `\$`, including any `${...}` that exist inside the bundled client JS (common in template-literal-heavy code). And the backtick escape is applied first, so `\${` sequence get further mangled when a backtick happen to come right before a `$`. Right order is: escape backslash first, then backtick, then the `${` sequence specifically (replace `/\${/g` with `'\\${'`).

MusabMahmoodh (commented):
Three big points first:

**1. No `MobileTarget`.** `DesktopConfig` is wired up because `DesktopTarget` exist at `jac-client/jac_client/plugin/src/targets/desktop_target.jac` and it is registered in `targets/register.jac`. In this PR the `MobileConfig` class is added but nothing call it. No `MobileTarget`, no entry in `register.jac`, no `targets/mobile/` folder. So `.validate()`, `.get_eas_config()`, `.get_app_json_config()` are dead code until a target is landed. The title say "infrastructure to support mobile app builds" but the build path is not in this PR. Please either include `MobileTarget` here, or mark this as scaffold only and hold the merge.

**2. `MobileConfig` is ~70% copy-paste of `DesktopConfig`.** Open `impl/mobile_config.impl.jac` and `impl/desktop_config.impl.jac` side by side: `init`, `_get_jac_config`, `load`, `get_platforms_config`, `get_features_config`, `validate` skeleton, and `_generate_identifier` vs `_generate_bundle_id` are almost identical. Please pull a `TargetConfigBase` out (or reuse the existing `PluginConfigBase` in `jaclang.project.plugin_config`) that own the common template. `DesktopConfig`/`MobileConfig` then only override `section_name` and `get_default_config`. If we skip this, every new target after this one is one more 200 line clone.

**3. PR description is empty.** The body is just `## **Description**` with no text. Please fill in what this PR does, how to verify it, and screenshot if any.

jac-client/jac_client/plugin/src/mobile_config.jac
```diff
@@ -0,0 +1,32 @@
...
+import from pathlib { Path }
+import from typing { Any }
+import from jaclang.project.config { JacConfig, get_config }
+
+"""Mobile target configuration loader.
+
+Provides access to mobile/React Native configuration
+from the [mobile] section of jac.toml.
+"""
+class MobileConfig {
```
MusabMahmoodh, line 17: This whole class is ~70% the same code as `DesktopConfig`. Please extract a `TargetConfigBase` (or reuse `PluginConfigBase` from `jaclang.project.plugin_config`) that own the `init`/`_get_jac_config`/`load`/`validate` template. `DesktopConfig` and `MobileConfig` then only override `section_name` and `get_default_config`. Otherwise each new target (TV, watch, whatever) is one more 200 line clone.

jac-client/jac_client/plugin/src/impl/mobile_config.impl.jac
```diff
@@ -0,0 +1,308 @@
...
+    # Try to get project name/version from JacConfig for defaults
+    project_name = "my-jac-app";
+    project_version = "1.0.0";
+    try {
+        jac_config = self._get_jac_config();
+        if jac_config and jac_config.project {
+            project_name = jac_config.project.name or project_name;
+            project_version = jac_config.project.version or project_version;
+        }
+    } except Exception {
```
MusabMahmoodh, line 30: `except Exception` swallow everything with no `e` detail, and the warning text here is identical to the one inside `load()` at line 101. If a user has a malformed `[mobile]` section or a typo, they see the same "Failed to load JacConfig" warning two times and no clue what actually fail. Please narrow the exception type, or at least log the real exception message.

jac-client/jac_client/plugin/src/impl/mobile_config.impl.jac
```diff
@@ -0,0 +1,308 @@
...
+        console.warning("Failed to load JacConfig, using fallback defaults");
+    }
+    # Generate bundle identifier from project name
+    bundle_id = _generate_bundle_id(project_name);
+    slug = project_name.lower().replace(" ", "-");
+    scheme = project_name.lower().replace(" ", "").replace("-", "").replace("_", "");
```
MusabMahmoodh, line 36: This same chain `name.lower().replace(" ", "").replace("-", "").replace("_", "")` appear three times in this file: here, inside `_generate_bundle_id` around line 290, and again in `get_app_json_config` around line 144. Extract one helper `_slugify_scheme` and call it from all three places.

jac-client/jac_client/plugin/src/impl/mobile_config.impl.jac
```diff
@@ -0,0 +1,308 @@
...
+    if self._config is not None {
+        return self._config;
+    }
+    default_config = self.get_default_config();
+    # Try to get [mobile] section from jac.toml
+    mobile_config = {};
+    try {
+        jac_config = self._get_jac_config();
+        # Get [mobile] section from raw TOML data
+        if jac_config and jac_config._raw_data {
```
MusabMahmoodh, line 97: Reaching into `jac_config._raw_data` is using a private attribute (underscore prefix). `DesktopConfig` already do this so this PR is not the first offender, but adding a second caller make the private attribute feel like real API. `JacConfig` expose `get_plugin_config(name)` for proper plugin sections. If a non-plugin top level section really need special access, please add one real accessor on `JacConfig` (for example `get_section(name)`) and have both `DesktopConfig` and `MobileConfig` use it. Don't open-code `_raw_data` in every target.

jac-client/jac_client/plugin/src/impl/mobile_config.impl.jac
```diff
@@ -0,0 +1,308 @@
...
+            'ios': {
+                'simulator': False,
+                'buildConfiguration': 'Release'
+            }
+        }
+    };
+
+    # Merge user-defined profiles with defaults
+    profiles = eas_config.get('profiles', {});
+    merged_profiles = {**default_profiles};
```
MusabMahmoodh, line 239: `{**default_profiles}` only copy the top level dict. The inner `development`/`preview`/`production` dicts are still shared reference with `default_profiles`. Right now you are saved because `default_profiles` is built from a fresh literal every call, but the pattern invite a bug the moment someone hoist that literal to module scope or cache it. Use `copy.deepcopy(default_profiles)`, or build `merged_profiles` by deep-merging into an empty dict.

jac-client/jac_client/plugin/src/impl/mobile_config.impl.jac
```diff
@@ -0,0 +1,308 @@
...
+"""Validate bundle identifier format."""
+def _is_valid_bundle_id(bundle_id: str) -> bool {
+    # Basic validation: should be reverse domain notation
+    # Example: com.example.myapp
+    parts: list[str] = bundle_id.split('.');
+    if len(parts) < 2 {
+        return False;
+    }
+    for p in parts {
+        if not p.isidentifier() {
```
MusabMahmoodh, line 303: `str.isidentifier()` is weaker than the regex `DesktopConfig._is_valid_identifier` use (`^[a-z0-9]+(\.[a-z0-9]+)+$`). `isidentifier()` accept underscore, uppercase, and even some Unicode letters. But Apple bundle id rule is alphanumeric + hyphen only, no underscore, and should be lowercase. So the mobile validator (where bundle id is actually a mobile concept) is weaker than the desktop one. Please reuse the regex from `DesktopConfig`, or extend it to also accept hyphen.

MusabMahmoodh (changes requested):

PR #5641, opened 2026-04-20
fix(parser): detect JS arrow function syntax early, emit E0025 with actionable   hint, and prevent cascade errors

Developer-Linus:
## Why this matters

A JS/TS developer hears "Jac has JSX" and writes:

```jsx
<button onClick={e => doThing(e)}>Click</button>
```

It fails — Jac uses `lambda`, not `=>`. The question is **how badly it fails**.

### Before

```
✖ error[E0001]: Expected '}', got '>'
✖ error[E0005]: Unexpected token '>'
✖ error[E0030]: Unexpected semicolon at module level
✖ error[E0005]: Unexpected token '}'
```

Four cascade errors. None mentions arrow functions. None points at `=>`. The lexer split `=>` into `=` then `>`; the lone `>` confused the JSX tag-context parser; chaos followed.

### After

```
✖ error[E0025]: JavaScript arrow function syntax ('=>') is not supported in Jac
  --> file.jac:2:31
    2 |     return <button onClick={e => doThing(e)}>Click</button>;
      |                               ^^
help: Use Jac lambda syntax instead:
    lambda e: Type { ... }
    lambda -> None { ... }
```

One error. Right location. Tells the user what to write instead.

## What it does

| Piece | Where | Effect |
| --- | --- | --- |
| Detect `=>` atomically in the lexer | `lexer.impl.jac::scan_operator` | Prevents `>` from leaking into JSX tag context — the root cause of the cascade |
| Emit diagnostic **E0025** with `lambda` hint | `diagnostics.jac` | Single, accurate error message |
| Drain the RHS after `=>` in the parser | `parser.impl.jac::parse_expression` | Stops downstream parser confusion (no "Missing `}`" cascade); brace-counter for block bodies, expression discard for simple bodies |
| Show `help:` line in CLI/exception output | `transform.impl.jac::Alert.__str__` | Hint is visible everywhere errors are printed |
| Suppress spurious W2001 on the recovery placeholder | `static_analysis_pass.impl.jac` | Pre-existing `isidentifier()` filter handles the `=>` Name placeholder |
| Diagnostics parser handles structured **and** legacy formats | `jac-client/.../diagnostics.impl.jac` | IDE error display stays clean for both `error[E0025]:` and `, line X, col Y:` outputs, with help text stripped from the message |

## False-positive analysis

The detection is narrow by construction.

| Case | Triggers E0025? | Why |
| --- | --- | --- |
| `a >= b`, `a <= b`, `a == b`, `a != b` | No | Different chars / different order; matched before `=>` in `scan_operator` |
| `a > b` (whitespace between `=` and `>`) | No | `match_string("=>")` requires literal byte-adjacency |
| `:=` (walrus), `->` (return type), `-->` / `<--` (graph arrows) | No | Don't start with `=`; matched as their own multi-char tokens |
| `"this has => inside"` (string literal) | No | Strings are scanned as a unit; never reach `scan_operator` |
| `#* comment has => *#` (block comment) | No | Comments are scanned as a unit |
| `<button onClick={e => fn(e)}>` (JSX attribute slot) | **Yes** | Real bug |
| `<div>{e => fn(e)}</div>` (JSX child slot) | **Yes** | Real bug |
| `glob f = x => y;` (module level) | **Yes** | Real bug |
| `f"{x => y}"` (f-string expression slot) | **Yes** | Slot is normal Jac — real bug |

### Why it can't false-positive

1. **Literal adjacency.** The lexer uses `match_string("=>")`. Only matches when `=` and `>` are byte-adjacent in source. `a = b > c` (whitespace/identifier between them) never matches.
2. **Mode isolation.** The check lives in `scan_operator` (NORMAL-mode operator scanner). String, comment, and JSX-tag-attribute scanners all consume their content separately. `=>` inside `"..."`, `#* ... *#`, or `<div x=...>` never reaches the operator path.
3. **Order in `scan_operator`.** Multi-char operators (`==`, `>=`, `<=`, `!=`, `->`, `-->`, `<--`, `>>=`, etc.) are matched before the `=>` check, so they're never split.

## Test coverage

**Parser tests** (`jac/tests/compiler/test_parser.jac`):
- `outside jsx: one E0025 at => with help hint` — proves lexer atomic match (col_start==12, not 13), correct code, help visibility, no E0004 duplicate.
- `jsx expression slot: one E0025, no cascade` — drain works in JSX attribute context.
- `block body drains braces cleanly` — brace-counter path; JSX continues parsing.
- `stray RBRACE does not over-consume jsx slot` — drain's else-branch doesn't eat past slot closer.
- `gte operator not confused` — `>=` regression guard.
- `lambda syntax unaffected` — replacement syntax still parses.

**Static analysis** (`test_static_analysis_pass.jac`): W2001 is suppressed for the synthesized `=>` Name placeholder.

**Diagnostics parser** (`jac-client/.../test_diagnostics_parser.jac`): both error formats parsed correctly with line/col extracted and help text stripped from message.

## Known follow-ups (not blocking)

These are architectural concerns from initial review that apply to the whole parser recovery convention, not just JS arrows. They belong in a separate cleanup PR:

1. **`TokenKind.ERROR` overload** — the parser's drain logic reacts to any ERROR token. A dedicated `JS_ARROW` kind would make the contract explicit.
2. **Chained `e => f => g`** — current drain handles only one `=>` per call; chained arrows produce extra cascade. 
3. **Grammar spec cosmetic leak** — recovery branches show up in published `jac.spec` rules. Hiding via grammar-extractor `ignore` registration is the right shape.

## Verification

- All 11 CI checks green on this branch.
- Hand-verification probes in `play/`:
  - `arrow_func_typo.jac` — canonical case → 1 error with help.
  - `e0025_false_positive_probes.jac` — 10 lookalikes → compiles clean.
  - `e0025_should_fire.jac` — `f"{a => b}"` (f-string slot) → fires E0025 correctly.


RagavRida:
The `Contribution Checks` failure here is the new fragment-format rule (`scripts/check-release-notes.sh`), which was tightened sometime between when this PR last passed CI on Apr 21 and now. Each fragment must start with `- ` so it's a literal markdown bullet — `**` alone trips the `^[^[:space:]-]` check.

One-line fix on each file (just prepend `- `):

```diff
--- a/docs/docs/community/release_notes/unreleased/jac-client/5641.bugfix.md
+++ b/docs/docs/community/release_notes/unreleased/jac-client/5641.bugfix.md
-**Diagnostics Parser: Structured Error Format and Help Text Support**: ...
+- **Diagnostics Parser: Structured Error Format and Help Text Support**: ...
```

```diff
--- a/docs/docs/community/release_notes/unreleased/jaclang/5641.bugfix.md
+++ b/docs/docs/community/release_notes/unreleased/jaclang/5641.bugfix.md
-**Parser: Arrow Function Syntax Detected Early with Actionable Hint**: ...
+- **Parser: Arrow Function Syntax Detected Early with Actionable Hint**: ...
```

That should turn the last red check green and unblock merge.

PR #5655, opened 2026-04-21
Adopt identity/credential auth payload in jaclang and jac-client

kashmithnisakya:
## **Description**

Aligns jaclang's `/user/login` and `/user/register` with the identity/credential payload shape already used by jac-scale, so a single client payload works against either server.

- **jaclang server** (`./jac`): `/user/register` and `/user/login` now require the new `{identity(ies), credential}` envelope. Hard cutover - the legacy `{username, password}` payload now returns 400. Only `identity.type == "username"` is accepted (SQLite user store has no email column yet).
- **jac-client + jaclang client runtime**: `jacLogin` / `jacSignup` overloaded to accept either `(username, password)` strings (backward-compatible) or full `(identity, credential)` dicts. Internally both send the new wire payload.
- New `auth_models.jac` with `IdentityType` / `CredentialType` enums and `LoginRequest` / `RegisterRequest` validators.


kugesan1105:
1.  Normalization methods needed? Make sure they are robust
image

2. Remove redundant code related to identity/credential from jac-scale
3. In the same PR, add test case that work with `jac-client` and `jac-scale`

Thamirawaran (approved):
Merging this pr as it is urgent. But make sure to address comments in the next pr

PR #5661, opened 2026-04-21
fix: add /assets proxy to Vite dev server configuration

### LINKED ISSUE #5660: Inconsistent Development Mode Behavior (Ports 8000 vs 8001)
### Description

**Description**
In development mode (jac start --dev), there is a significant inconsistency between the Vite Dev Server (port 8000) and the Jac API Server (port 8001) regarding static asset loading and Hot Module Replacement (HMR).

**Reproduction Steps**
Create a new full-stack project: jac create myapp --use client.
Add an image to the assets/ directory.
Display the image in 
main.jac: ``.
Start the server: jac start --dev.
Access http://localhost:8000/ and http://localhost:8001/.
**Expected Behavior**
Static assets should load on both ports.
Hot-reloading should eventually sync changes to both ports (or consistently use one for development).
Reloading/Rebuilding the client bundle should be performant.
**Actual Behavior**
Port 8000: Loads the app and supports HMR, but fails to load static assets from the assets/ directory (returns 500/404).
Port 8001: Loads static assets correctly, but does not support HMR. Requires a manual restart to see changes.
Performance: Client bundle building takes ~60+ seconds even for minor changes.
Environment
OS: Linux
Jac Version: 0.14.0 (Dev Server)

kiptuidenis:
## **Description**
This PR fixes an inconsistency in Jaseci's development mode where static assets (e.g., images) fail to load when accessing the application via the Vite dev server (port 8000), while they load correctly on the API server (port 8001).

**Problem**
In development mode (jac start --dev), the Vite dev server manages the frontend. While it was configured to proxy API requests and some static routes to the Jac backend, the /assets directory was missing from the proxy configuration. This resulted in 404 errors for any images or static files referenced via /assets/....

**Proposed Changes**
Modified `jac-client/jac_client/plugin/src/impl/vite_bundler.impl.jac`
 to include a proxy rule for the /assets path in the Vite dev server configuration.
Added a release note fragment following the new contributor process.
**Verification Results**
Created a test application using jac create myapp --use client.
Verified that static images (e.g., in assets/logo.png) now load correctly on both the Vite server (port 8000) and the Jac API server.


kugesan1105:
@kiptuidenis  it would be better if we can add a test for this one

PR #5662, opened 2026-04-21
feat(jac-scale): partial anchor updates — skip full $set when only archetype fields changed

akindu-k:
## Problem

Every anchor save performed a full `$set` replacing the entire MongoDB document, even when a walker changed a single field. A one-field mutation triggered `Serializer.serialize()` on the whole archetype and rewrote every key in the stored document.

## Solution

Four layers of implementation, plus a combined dirty-detection strategy that covers both direct assignments and in-place mutations.

### Layer 1 — Dirty-field tracking on `Archetype` (`archetype.jac`, `archetype.impl.jac`)

- `Archetype.__setattr__` lazily populates `__jac_dirty_fields__: set[str]` on the instance when a non-underscore field is mutated after initialization.
- `Archetype.postinit` sets `__jac_initialized__ = True` after dataclass `__init__` completes so construction-time assignments are not tracked.
- `Serializer._deserialize_archetype` sets `__jac_initialized__ = True` after field population so DB-loaded nodes also get dirty tracking on subsequent mutations.
- Both flags live in `__dict__` (not dataclass fields) and are invisible to `_serialize_attrs` (double-underscore prefix).

### Layer 2 — Selective serialization (`serializer.jac`, `serializer.impl.jac`)

- New static method `Serializer.serialize_fields(anchor, field_names)` returns only the requested archetype fields as a serialized dict.
- Intersects `field_names` with `_get_field_types(cls)` (introduced in #5561) to guard against invalid or private names before touching the object.
- Passes a shared `_seen` set to every `_serialize_value` call so circular field values emit `$ref` placeholders instead of recursing infinitely.

### Layer 3 — Partial MongoDB write (`memory_hierarchy.jac`, `memory_hierarchy.mongo.impl.jac`)

- New `MongoBackend.put_partial(anchor, field_names)` builds `data.archetype.<field>` dot-path `$set` entries and updates only those keys in MongoDB, leaving `type`, `arch_module`, `fingerprint`, `edges`, `access`, and all other outer-document fields untouched.
- Recomputes `anchor.hash` and clears `__jac_dirty_fields__` after success so the next sync sees a clean state.

### Layer 4 — Combined dirty detection and smart routing in `sync()` (`memory_hierarchy.mongo.impl.jac`)

`__setattr__` alone cannot detect in-place mutations to mutable fields. The solution combines two mechanisms:

**Fast path — `__setattr__` tracking (zero cost):**
Catches direct assignments (`node.x = v`). `__jac_dirty_fields__` is non-empty → `put_partial` is called immediately.

**Fallback path — per-field hash comparison (paid only when needed):**
When the anchor-level hash changed but `__jac_dirty_fields__` is empty (in-place mutation), three new helpers derive the dirty set by serializing each field and comparing against a stored snapshot:

- `_field_hash(val)` — JSON-serializes a field value and hashes it; handles dict, list, and scalar uniformly.
- `_snapshot_field_hashes(anchor)` — serializes all known fields and stores `{field: hash}` as `__jac_field_hashes__` on the archetype instance after every save and load. This is the baseline for comparison.
- `_derive_dirty_from_field_hashes(anchor)` — reserializes all fields, diffs against the snapshot, returns the set of changed field names.

The fallback only pays the serialization cost when the coarse anchor-level hash signals a change without a `__setattr__` signal — i.e., exactly the in-place mutation case.

**Full routing table in `sync()`:**

| Condition | Action |
|---|---|
| `NodeAnchor` with edge delta | `_put_node_atomic()` — atomic edge merge |
| `__jac_dirty_fields__` non-empty and `hash != 0` | `put_partial()` — write only dirty fields |
| `hash != 0`, `__jac_dirty_fields__` empty, field hashes differ | `put_partial()` — derived dirty set from field-level diff |
| `hash == 0` (first insert) or no dirty fields found | `_write_to_db()` — full document upsert |

## Data flow

```
walker mutates node.my_list.append(42)          walker mutates node.value = 42
    │                                                │
    │ (no __setattr__ signal)                        ▼ Archetype.__setattr__
    │                                            __jac_dirty_fields__ = {"value"}
    ▼                                                │
sync(): anchor hash changed, dirty set empty    sync(): dirty set non-empty
    │                                                │
    ▼ _derive_dirty_from_field_hashes()              ▼
  field hash diff → {"my_list"}               put_partial({"value"})
    │                                                │
    ▼ put_partial({"my_list"})                       ▼
  $set {"data.archetype.my_list": [...]}      $set {"data.archetype.value": 42}
```

## Foundation PRs actually used

| PR | What it contributed | How we use it |
|---|---|---|
| **#5387** `ref_mode` | `_serialize_value` gained `_seen` cycle guard and `ref_mode` | `serialize_fields` and `_snapshot_field_hashes` pass a shared `_seen` set to every `_serialize_value` call |
| **#5561** persistence layer | `_get_field_types(cls)` — resolves dataclass field names + type hints | `serialize_fields` and `_snapshot_field_hashes` intersect against `_get_field_types` output to block invalid/private keys |

## Files changed

| File | Change |
|------|--------|
| `jac/jaclang/jac0core/archetype.jac` | Add `postinit` declaration to `Archetype` |
| `jac/jaclang/jac0core/impl/archetype.impl.jac` | `__setattr__` dirty tracking; `postinit` sets `__jac_initialized__` |
| `jac/jaclang/runtimelib/serializer.jac` | Declare `serialize_fields` static method |
| `jac/jaclang/runtimelib/impl/serializer.impl.jac` | Implement `serialize_fields`; set `__jac_initialized__` after deserialization |
| `jac-scale/jac_scale/memory_hierarchy.jac` | Declare `put_partial` and `_write_to_db` on `MongoBackend` |
| `jac-scale/jac_scale/impl/memory_hierarchy.mongo.impl.jac` | `_field_hash`, `_snapshot_field_hashes`, `_derive_dirty_from_field_hashes` helpers; `put_partial` impl; `_write_to_db` impl; `sync` three-way routing; snapshot on load/save |


akindu-k:
image

PR #5666, opened 2026-04-22
test(auth): add e2e roundtrip fixture for jacSignup/jacLogin

kashmithnisakya:
## **Description**

- Adds `auth-calling-forms` example under `jac-client/examples/`, demonstrating `jacLogin`/`jacSignup` called with both the bare-string form and the new identity/credential (dict/list) form.
- Adds a Playwright e2e test that drives the example through all four call combinations (signup-old, login-old, signup-new, login-new).
- CI re-runs the test once with `jac-scale` installed, so both the jaclang and jac-scale backends are validated in one job.


kugesan1105:
@kashmithnisakya  merge the main

kugesan1105:
**Context**: client e2e tests previously only validated `jac-client` ↔ `jaclang-core`. The auth flows had no coverage against `jac-scale`, so drift between the two backends could slip through. This PR adds an e2e that exercises both backends in CI, closing that compatibility gap.

PR #5670, opened 2026-04-22
feat: PyInstaller hook for .jac auto-discovery (obsoletes manual __init__.py scaffolding)

### LINKED ISSUE #5606: PyInstaller `FrozenImporter` bypasses Jac's meta path finder — breaks namespace-package imports in frozen desktop builds
## Summary

When Jac packages (`jac_client`, `jac_scale`, etc.) are frozen into a desktop sidecar with PyInstaller, submodules declared only via `__init__.jac` (no `__init__.py`) fail to import at runtime with `ModuleNotFoundError` — even though every `.jac` source and compiled artifact is physically present in the bundle.

The workaround used today is to ship a stub `__init__.py` in every directory that has an `__init__.jac`. This works, is already how `jac_scale` is packaged, and is what PR #5466 proposes for `jac_client`. This issue tracks the **real fix**: teaching PyInstaller's import machinery to cooperate with Jac's custom meta path finder so the shim files are no longer needed.

## Root Cause

Jac installs a custom meta path finder at interpreter startup that:

1. Treats any directory containing `__init__.jac` as an importable Python package (PEP 420 namespace package semantics, extended to Jac sources).
2. Compiles `.jac` modules on demand and injects the resulting module objects into `sys.modules`.

In a normal CPython process this works because `sys.meta_path` is walked in order and Jac's finder gets a turn before the default `PathFinder`.

Inside a PyInstaller-frozen binary, the picture changes. PyInstaller installs its own `FrozenImporter` (`PyInstaller.loader.pyimod02_importers.PyiFrozenImporter`) at the **front** of `sys.meta_path` and is architected to resolve everything from its embedded archive — it does not consult downstream finders for packages it considers missing. The consequences:

- Directories without `__init__.py` are not recognised as packages by `FrozenImporter`, so `import jac_client.plugin.src.targets.desktop.sidecar.main` fails at the first missing `__init__.py` — even if the Jac sources and the compiled `.py` files are bundled alongside.
- Jac's meta path finder never gets a turn, because `FrozenImporter` raises `ImportError` on unknown names rather than returning `None` for every miss the way `PathFinder` does.

The workaround (adding empty or re-exporting `__init__.py` files) sidesteps this by making each Jac package a **regular Python package** that `FrozenImporter` recognises directly, without needing the custom finder.

## Current State

- **`jac_scale`** already ships with `__init__.py` files in every submodule directory. This has been the de-facto packaging pattern since it was first made PyInstaller-freezeable.
- **PR #5466** proposes applying the same pattern to `jac_client`, unblocking Windows desktop builds (Tauri + PyInstaller sidecar). Review feedback on #5466 raised the valid architectural concern that the real fix should live in PyInstaller, not in our tree.
- **Desktop build pipeline** (PRs #5576–5581, plus the desktop tutorial in #5594) is blocked or degraded without #5466 on Windows, because the frozen sidecar crashes at startup on the first `jac_client.*` import.

## Why the Workaround Is Acceptable Short-Term

1. `jac_scale` already does this, so merging #5466 **restores consistency** rather than introducing a new precedent.
2. The shim files are trivial and have zero runtime behaviour beyond package registration.
3. When the upstream fix lands, removing them is a single mechanical PR.

## Path Forward

Two non-exclusive options for the real fix:

### Option A: Upstream PR to PyInstaller

Modify `PyiFrozenImporter.find_spec` (and/or `find_module` on legacy Python) so that, when it cannot resolve a fully-qualified name from the embedded archive, it returns `None` rather than silently owning the import. This would allow Jac's meta path finder — which PyInstaller already leaves installed on `sys.meta_path` at user-code execution time — to take over for Jac submodules.

A smaller variant: make `FrozenImporter` treat a directory as a package when any file of a configured extension (e.g. `.jac`) is present, not just `__init__.py`. This is less general but closer to what Jac specifically needs.

Risks / complexity:

- PyInstaller's maintainers are cautious about changes to the frozen import path because it's load-bearing for every frozen app in the wild. A PR would need strong test coverage and a clear explanation of the non-regression path.
- Estimated effort: 2–4 focused weeks minimum to get to a mergeable PR, plus uncertain review latency.

### Option B: Soft Fork

Fork PyInstaller, apply the patch as a series of commits, and pin Jaseci's desktop build tooling to the fork. Periodically rebase on upstream.

Risks / complexity:

- Indefinite maintenance tax — every PyInstaller release requires a rebase.
- Downstream Jac users building their own desktop apps need to install the fork, not standard PyInstaller. That's a significant UX regression.

### Recommended Sequence

1. Merge #5466 now using the `__init__.py` shim pattern, to unblock Windows desktop builds and keep `jac_client` consistent with `jac_scale`.
2. Prototype Option A locally against the current PyInstaller main branch. Measure the diff size and test impact.
3. If the prototype is clean, open an upstream PR against `pyinstaller/pyinstaller`.
4. If the upstream PR stalls or is rejected, re-evaluate Option B.
5. When the upstream fix lands and is released, remove the `__init__.py` shims from `jac_client` and `jac_scale` in a single follow-up PR.

## Acceptance Criteria

Done looks like one of:

- PyInstaller upstream accepts a PR that allows custom meta path finders to resolve frozen-binary imports for directories without `__init__.py`. Jac packages no longer require `__init__.py` shims. Shim files are removed from `jac_client` and `jac_scale`.
- A narrowly-scoped soft fork is published, Jaseci tooling points at it, and the removal PR is ready to land once upstream catches up.

## References

- PR #5466 — `fix: add __init__.py with Jac imports for dirs that have __init__.jac`
- PRs #5576, #5577, #5578, #5579, #5580, #5581 — Windows desktop build fixes that depend on the frozen sidecar working correctly
- PR #5594 — Desktop build tutorial (documents the current CI flow)
- PyInstaller source: `PyInstaller/loader/pyimod02_importers.py` (`PyiFrozenImporter.find_spec`)
- PEP 451 (ModuleSpec type, `find_spec` protocol)
- PEP 420 (namespace packages)

kugesan1105:
## Summary

Teaches PyInstaller (and every other ``importlib``-based tool) to recognize Jac packages, so freezing a Jac/jaclang app no longer requires sprinkling empty ``__init__.py`` files next to every ``__init__.jac``. This obsoletes the manual scaffolding approach of #5466.

The core intervention is a **one-line extension to the existing Python hijack** in [`_jac_finder`](jac/_jac_finder.py): register ``.jac`` in ``importlib.machinery.SOURCE_SUFFIXES`` and install a ``FileFinder`` path hook for it. That single change cascades to every tool that goes through ``importlib`` — PyInstaller's analyzer, setuptools ``find_packages``, Nuitka, cx_Freeze, IDE indexers — with zero per-tool adapters.

### The problem, in one paragraph

PyInstaller's module graph uses ``pkgutil.get_importer(dir) → FileFinder``, not ``sys.meta_path`` (deliberately — meta-path finders can load from memory/network, which a bundler can't capture). ``FileFinder`` only looks at suffixes in ``importlib.machinery.SOURCE_SUFFIXES``, which is hardcoded to ``['.py']``. So ``myapp/__init__.jac`` is invisible, ``from myapp.core.greeter import greet`` resolves to ``missing module``, and the whole package gets excluded from the bundle. PR #5466 worked around this by scaffolding empty ``__init__.py`` markers by hand. This PR fixes the underlying cause.

## Architecture

Three layers, structurally different from each other:

| Layer | Files | Lines | Role |
|---|---|---|---|
| Python hijack extension | [`_jac_finder.py`](jac/_jac_finder.py), [`jaclang/__init__.py`](jac/jaclang/__init__.py) | 107 | One-time ``importlib`` intervention; serves every tool |
| Sovereign ``jaclang.packaging`` API | [`packaging/`](jac/jaclang/packaging/) | 172 | First-class Jac package discovery — ``find_packages``, ``iter_jaclang_data_files``. Future ``jac build --target=binary`` will consume the same module. |
| Thin PyInstaller adapter | [`_pyinstaller/`](jac/jaclang/_pyinstaller/) | 67 | Translation only. ``hook-jaclang.py`` is 30 lines of visibly-delegation code. ``rthooks.dat`` + ``pyi_rth_jac_finder.py`` are the minimum PyInstaller protocol surface. |

Everything in the adapter delegates to ``jaclang.packaging``. No business logic lives in a PyInstaller-shaped directory. When the native ``jac build`` pipeline lands, the adapter becomes a compatibility layer, not dead code.

## Key design choices

- **Path hook is gated on ``sys.frozen``.** Installed at every non-frozen Python startup; skipped at frozen-app runtime. The loader's empty-source fallback is correct at build time (analyzer just needs to see the file) but would actively shadow ``JacMetaImporter`` during jaclang's own internal bootstrap at runtime.
- **``_JacLazyFinder`` moves to ``sys.meta_path`` position 0.** Ensures it wins against PyInstaller's frozen-app PYZ finder, which would otherwise claim ``.jac`` modules via compiled empty stubs.
- **``import jaclang`` now triggers ``_jac_finder.install()``.** Covers PEP 660 editable installs where ``jaclang.pth`` doesn't get wired up. For wheel installs (production), ``.pth`` still fires at Python startup as designed.
- **Rthook is keyed on ``_jac_finder`` (not ``jaclang``).** Clearer dependency — the rthook's job is to run ``_jac_finder.install()`` at frozen startup. ``_jac_finder`` is explicitly added to ``hiddenimports`` so the rthook is guaranteed to fire.

## Tests

**18 tests in [`tests/packaging/`](jac/tests/packaging/)** — 15 unit + 3 integration.

- **[`test_discovery.py`](jac/tests/packaging/test_discovery.py)** (15 unit, ~50 ms)
  - ``TestFindPackages`` (6) — correctness across edge cases: hidden/dunder dirs, dedupe, Python-only packages, missing dirs
  - ``TestIterSources`` (4) — dotted module names, hidden-subdir skip, non-``.jac`` file rejection
  - ``TestJacSource`` (1) — dataclass immutability
  - ``TestIterJaclangDataFiles`` (3) — scope, tuple shape, presence of bootstrap-critical ``modresolver.jac``
- **[`test_pyinstaller_integration.py`](jac/tests/packaging/test_pyinstaller_integration.py)** (3 integration, ~60 s)
  - ``test_frozen_app_runs_jac_only_package`` — the acceptance test. Materializes a Jac-only package (5 ``.jac`` files, **zero ``__init__.py``**), builds with PyInstaller, runs the binary from ``cwd=/`` to prevent source-tree escape, asserts expected output. If this passes, PR #5466's scaffolding is no longer needed.
  - ``test_no_init_py_was_added`` — fixture invariant guard.
  - ``test_pyinstaller_cli_resolvable`` — environment self-check.

**CI wired up:** [`test-jaseci.yml`](.github/workflows/test-jaseci.yml) now ``pip install``s PyInstaller in the ``test-core-runtime`` job so the acceptance test actually executes — without this, ``pytest.importorskip`` would cause it to skip silently in CI.

## Test plan

- [ ] CI green on ``test-core-runtime`` job (all 18 tests, including the 60 s frozen-build integration test)
- [ ] ``pytest jac/tests/packaging -q`` → 18 passed locally
- [ ] Manually verify a local PyInstaller build of a Jac-only package works end-to-end without ``__init__.py`` scaffolding
- [ ] No regression in existing test suites (``test-core-compiler``, ``test-core-runtime`` other tests)
- [ ] Follow-up (separate PR against ``jac-client``): delete the manual ``__init__.py`` scaffolding added by #5466 and confirm the desktop sidecar builds using this hook

## Obsoletes

Closes / supersedes #5466 once the follow-up jac-client cleanup PR merges.

## Commit history

6 commits on the branch, narrating the evolution from initial hook → architectural refactor into sovereign ``jaclang.packaging`` → acceptance-test coverage → final cleanup. Can be squashed at merge time, or kept for the audit trail — reviewer's call.


marsninja:

PR #5575, opened 2026-04-16
fix(runtime): allow bidirectional traversal of undirected edges

kiptuidenis:
## **Description**
This PR fixes an issue where bidirectional (undirected) edges could not be traversed from the `target` node when using directional operators (e.g., `-->` or ` ->:edge:->`).

Root Cause
The runtime traversal logic (`get_edges`, `get_edges_with_node`, and `edges_to_nodes`) strictly validated OUT traversal against the edge source and IN traversal against the edge target, ignoring the `is_undirected` flag on the `EdgeAnchor`.

Solution
Introduced an `effective_dir` check: if an edge is undirected, the traversal direction is treated as `ANY`, effectively allowing traversal from either side regardless of internal source/target identity.

Testing
Verified locally with a reproduction script involving both typed `<+:edge:+>` and generic `<++>` bidirectional edges. All directional queries from the reverse side now return the expected nodes.

PR #5672, opened 2026-04-22
Docs/fix undirected edges wording

kiptuidenis:
## **Description**
This PR corrects and enhances the Jac documentation regarding edge directionality, specifically focusing on the behavior of the `<++>` (undirected connection) operator.

**Core Changes**
Wording Standardization: Replaced inaccurate descriptions of `<++>` (e.g., "creates edges both ways" or "two edges") with the correct terminology: single undirected edge. This aligns the documentation with the actual behavior of the runtime post-#5575.

### DIFF
--- a/docs/docs/quick-guide/syntax-cheatsheet.md
+++ b/docs/docs/quick-guide/syntax-cheatsheet.md
@@ -744,7 +744,7 @@ with entry {
     root ++> a;             # Connect root -> a
     a ++> b;                # Connect a -> b
     c <++ a;                # Connect a -> c (backward syntax)
-    a <++> b;               # Bidirectional a <-> b
+    a <++> b;               # Undirected a <-> b
 
     # --- Typed connections (with edge data) ---
     a +>: Friendship(since=2020) :+> b;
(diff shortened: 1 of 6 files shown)


drPod (approved):
LGTM, this is cleaner than my PR's rewrite-of-§-Edges.3 approach and covers `appendices.md`, `library-mode.md`, and `coding_primer.md` which I'd missed.

One minor suggestion, non-blocking: in `osp.md` you could add a one-line pointer to `[<-->]` as the query-side direction-agnostic filter, since the rest of the doc already uses it that way in Walkers § 3 and Data Spatial Queries § 1 - makes the "why two operators plus a filter" story coherent. But happy for this to land as-is and follow up separately if you'd rather keep this PR minimal.

Closing #5665 as superseded by this one.

Thamirawaran:
As we can traverse bidirectionally through bidirectional edge, it should be bi directional. why we need to name it as undirected?
Let me clarify
Initially there was a bug in creation of bidirectional edges. This issue is fixed in #5575. and PR naming has conflict. But anyway it is bidirectional edge. therefore this issue is currently outdated. And make sure in all of our documentation we keep the naming as bidirectional edge.

PR #5674, opened 2026-04-22
scheduler hardening - auth, correctness, and ops

kashmithnisakya:
## **Description**

- Push pagination + filters (`limit`, `offset`, `trigger`, `created_by`) into the MongoDB query instead of fetching everything and slicing
- Atomic `record_run` via `$inc` + `$set` (one round-trip)
- Cache `user_exists` with a 30s positive TTL so high-frequency jobs stop hitting the DB every tick
- Reject cron triggers when APScheduler isn't installed (was a silent skip)
- Extract `_utc_now_iso()` helper for the duplicated UTC timestamp
- Add cross-user ownership test


jac-scale/jac_scale/scheduler/impl/scheduler.impl.jac
```diff
@@ -0,0 +1,731 @@
...
+                    f"'{item_name}' has DYNAMIC trigger — managed by APScheduler"
+                );
+                continue;
+            }
+            logger.debug(f"Registered static {label}: {item_name}");
+            base_scheduler.register(
+                name=item_name,
+                target=_make_target(item_name, is_walker),
+                spec=spec,
+                is_walker=False
```
MusabMahmoodh: this looks like a regression to me. the outer loop is `(walkers, True), (functions, False)` so `is_walker` already has the right value here, but its getting passed as a hardcoded `False`.

execution itself still works because `_make_target(item_name, is_walker)` on line 409 captures the right loop var, but anywhere the base `Scheduler` reads `is_walker` (telemetry, dispatch, logs) its now wrong for static walkers. think it should just be `is_walker=is_walker` like before the refactor?

jac-scale/jac_scale/impl/serve.scheduler.impl.jac
```diff
@@ -1,39 +1,115 @@
...
+    err_name = type(e).__name__;
+    storage_markers = (
+        'Mongo',
+        'ServerSelection',
+        'NetworkTimeout',
+        'ConnectionFailure',
+        'AutoReconnect',
+        'OperationFailure',
+        'WriteError'
+    );
```
MusabMahmoodh: sniffing the exception class *name* as a string for storage classification is brittle. if pymongo renames or wraps a class this silently falls through to 500. and any third party exception that happens to have `OperationFailure` in its name maps to 503.

also `EncryptionError`, `InvalidOperation` etc dont contain any of these substrings so they end up as 500 even though theyre clearly storage layer.

usual approach is `isinstance(e, pymongo.errors.PyMongoError)` no? pymongo is already a runtime dep when mongo is configured so importing it here should be fine.

jac-scale/jac_scale/scheduler/impl/scheduler.impl.jac
```diff
@@ -0,0 +1,731 @@
...
+        'job_id': job_id,
+        'name': name,
+        'created_by': created_by,
+        'is_walker': is_walker,
+        'created_at': now_str,
+        'status': 'active'
+    };
+    job_data.update(spec.to_dict());
+
+    with self._lock {
```
MusabMahmoodh: the RLock is held across `self._store.insert(...)` on the next line which is a network call to mongo. if mongo is slow every other /jobs request and every aps worker tick blocks behind it.

also its a per-process lock so it doesnt help across replicas anyway. if we want safe concurrent CRUD a conditional `update_one(filter={'job_id': X, 'status': 'active'}, ...)` at the mongo layer is way more useful than a process-local mutex giving the illusion of safety.

if the lock has to stay can it at least not span the I/O? same pattern in `update_job` (566), `delete_job` (589), and the orphan transition (706).

jac-scale/jac_scale/impl/serve.scheduler.impl.jac
```diff
@@ -273,99 +350,65 @@ impl JacAPIServerScheduler.list_jobs_callback -> Callable[..., TransportResponse
...
             scheduler = JacScaleScheduler.instance();
-            updated = scheduler.update_job(
-                job_id=job_id,
-                trigger=trigger,
-                interval=float(interval) if interval is not None else None,
-                cron=cron_expr,
-                date=date_str
-            );
+            existing = scheduler.get_job(job_id);
+            access_err = _ensure_authorized(existing, user_id, self.user_manager);
```
MusabMahmoodh: this is `get_job` then `_ensure_authorized` then a separate `update_job` later. two concurrent owner requests can step on each other since the read and write arent atomic, and the per-process `_lock` doesnt cover cross-replica.

clean fix: push `created_by` (or admin override) into the mongo filter on the actual mutating call. then existence check + auth check + write becomes one atomic op, and unauthorized writes become zero-row.

same applies to the DELETE callback at line 427-432.

jac-scale/jac_scale/impl/serve.scheduler.impl.jac
```diff
@@ -1,39 +1,115 @@
...
+    );
+    for marker in storage_markers {
+        if marker in err_name {
+            return TransportResponse.fail(
+                code='STORAGE_UNAVAILABLE',
+                message='Job store temporarily unavailable',
+                meta=Meta(extra={'http_status': 503})
+            );
+        }
+    }
```
MusabMahmoodh: the 500 path on line 38 calls `logger.error(...)` but the 503 path here returns silently. for ops, storage outages are exactly what should show up on dashboards/alerts. should be the other way around imo.

a `logger.error(f"storage failure on /jobs: {type(e).__name__}: {e}")` before the return would be enough.

jac-scale/jac_scale/scheduler/scheduler.jac
```diff
@@ -0,0 +1,146 @@
...
+    static def from_dict(d: dict) -> TriggerSpec;
+    def to_dict -> dict;
+    def apscheduler_kwargs -> dict;
+}
+
+"""Jac-Scale scheduler that bridges APScheduler and the jaclang base Scheduler."""
+obj JacScaleScheduler {
+    has _apscheduler: Any | None = None,
+        _system_user_id: str = '__system__',
+        _store: JobStore = MemoryJobStore(),
```
MusabMahmoodh: `_user_exists_cache` on line 110 got the postinit reset because of mutable-default-shared-across-instances. `_store: JobStore = MemoryJobStore()` has the same trap. singleton hides it in prod but tests that construct fresh `JacScaleScheduler()` instances will inherit whatever state another instance left in there.

same fix: reset `_store` in postinit, or make it `None` and lazy-init inside `setup()`.

jac-scale/jac_scale/scheduler/scheduler.jac
```diff
@@ -0,0 +1,146 @@
...
+        _server: Any | None = None,
+        _lock: threading.RLock = threading.RLock(),
+        # Positive cache: user_id -> expiry epoch seconds. Negative results are
+        # not cached (user was just deleted → transition immediately).
+        _user_exists_cache: dict[str, float] = {},
+        _user_exists_ttl: float = 30.0;
```
MusabMahmoodh, line 107: 30s positive cache means a 1s-interval job belonging to a user who was just deleted can fire ~30 times before flipping to orphaned. depending on what the job does (billing, side effects, sending email) thats not great.

can `_user_exists_ttl` at least be configurable? would be cleaner to invalidate from the user-deletion path but that adds a backref dependency.

jac-scale/jac_scale/scheduler/impl/scheduler.impl.jac
```diff
@@ -0,0 +1,731 @@
...
+                job_id=persisted_id,
+                name=name,
+                spec=spec,
+                user_id=created_by,
+                is_walker=is_walker
+            );
+        } except Exception {
+            self._store.delete(persisted_id);
+            raise;
+        }
```
MusabMahmoodh, line 596: question: when does `_schedule_apscheduler_job` actually fail post-validation? `TriggerSpec.validate` already runs `CronTrigger.from_crontab` upfront so the only real post-validation failures left are aps-internal bugs.

also the rollback `self._store.delete(persisted_id)` can itself fail (transient mongo blip etc) and if it does we silently leak a record + raise the original error. orphan-detection on next restart picks it up but its weird state.

might be cleaner to drop the rollback and rely on the upfront validation? or at minimum a comment explaining what scenario this is actually catching.

jac-scale/jac_scale/scheduler/scheduler_models.jac
```diff
@@ -0,0 +1,88 @@
...
+            None,
+            description='5-field cron expression interpreted in UTC. '
+            'Required when trigger="cron".'
+        ),
+        date: str | None = Field(
+            None,
+            description='ISO-8601 datetime (UTC) to fire once. '
+            'Required when trigger="date".'
+        ),
+        model_config: ConfigDict = ConfigDict(
```
MusabMahmoodh, line 38: nit: `extra` policy isnt set explicitly. default is "ignore" today but pydantic has flipped this between major versions before. setting `extra='ignore'` (or 'forbid' if we want stricter) keeps it stable across upgrades. same for `UpdateJobRequest`.

jac-scale/jac_scale/scheduler/impl/scheduler.impl.jac
```diff
@@ -0,0 +1,731 @@
...
+            raise ValueError("interval must be > 0 seconds");
+        }
+    } elif trigger == 'cron' {
+        if not cron {
+            raise ValueError("cron expression required for trigger='cron'");
+        }
+        # Structural + range validation via APScheduler's parser. Catches things
+        # like '99 99 99 99 99' that pass a naive 5-field check. If APScheduler
+        # is not installed, cron triggers are unsupported — don't silently accept.
+        _ct: Any = CronTrigger;
```
MusabMahmoodh, line 31: minor: `TriggerSpec` is a request-validation object but `validate()` now reaches into apscheduler. so anything that wants to validate a cron string in a context without aps installed (a unit test, a one-off CLI tool) hits "install jac-scale[scheduler]".

not blocking but could pull the cron-structural check into a small helper so `TriggerSpec` stays pure?

jac-scale/jac_scale/scheduler/impl/scheduler.impl.jac
kashmithnisakya: Regression from the refactor. Fixed, now passes `is_walker=is_walker` through so the base scheduler reports the right type for static walkers.

jac-scale/jac_scale/impl/serve.scheduler.impl.jac
kashmithnisakya: Switched to `isinstance(e, PyMongoError)` so any pymongo error maps to 503, including `EncryptionError` and `InvalidOperation` that the substring check missed. Added `PyMongoError` to the `_optdeps/pymongo` shim with a sentinel fallback so the check is safe when pymongo isn't installed.

jac-scale/jac_scale/scheduler/impl/scheduler.impl.jac
kashmithnisakya: Dropped the per-process RLock from create/update/delete/orphan paths and pushed the auth predicate into the mongo filter via a new `where` arg on `JobStore.update`/`delete`. APScheduler is internally thread-safe so the lock was protecting a network call without giving cross-replica safety. Now `update_job(acting_as=user_id)` and `delete_job(acting_as=user_id)` issue a single `update_one`/`delete_one` with `{job_id, created_by}` in the filter, so unauthorized or racing writes return zero rows and surface as 404. Orphan transition uses `where={status: active}` for idempotent cross-replica writes. Fixes #3 and #4 together.

jac-scale/jac_scale/impl/serve.scheduler.impl.jac
kashmithnisakya: Fixed together with #3. The endpoint now computes `acting_as = None if is_admin else user_id` and passes it to `scheduler.update_job`/`delete_job`, which issue a single `update_one`/`delete_one` with `created_by` in the filter. The existence check, auth check, and write are now one atomic mongo op — unauthorized writes match zero rows and the caller gets 404. Same pattern applied to the DELETE callback. The cross-user ownership test still passes (returns NOT_FOUND for both PUT and DELETE).

jac-scale/jac_scale/impl/serve.scheduler.impl.jac
kashmithnisakya: Added `logger.error("Storage failure on /jobs: ...")` to the 503 path so storage outages now surface in dashboards/alerts.

jac-scale/jac_scale/scheduler/scheduler.jac
kashmithnisakya: Switched `_store: JobStore = MemoryJobStore()` to `_store: JobStore | None = None` and routed every read through a new `_require_store() -> JobStore` helper that raises if `setup()` was never called. No class-level mutable default to share across instances anymore, and pre-setup access fails loudly instead of silently leaking state. All call sites updated; `MemoryJobStore` is still the runtime fallback inside `setup()` when there is no `mongodb_uri` or the connection fails.

jac-scale/jac_scale/scheduler/scheduler.jac
kashmithnisakya: Made `_user_exists_ttl` configurable via `[scheduler].user_exists_ttl` in jac.toml, default still 30.0. Operators with side-effecting jobs can lower it (e.g. `user_exists_ttl = 5.0`); deployments with high-frequency jobs against a slow user DB can raise it. Skipping the user-deletion-path invalidation for now since it adds the back-reference dependency you flagged - happy to do that as a follow-up if needed.

jac-scale/jac_scale/scheduler/impl/scheduler.impl.jac
kashmithnisakya: Wrapped the rollback `store.delete(persisted_id)` in its own try/except. If the rollback itself fails (mongo blip during cleanup), we now log the leaked `job_id` and let the original APScheduler error propagate — orphan-detection on next restart cleans up the leaked row. Without this, the rollback exception was silently replacing the original one, hiding the real root cause from the caller and from logs. Also added a comment block explaining what post-validation scenarios the outer rollback is actually catching (APS shutdown race, internal APS errors) since `TriggerSpec.validate` already covers cron/date format errors upfront — kept the rollback because it still prevents retry-duplicates in those rare cases.

jac-scale/jac_scale/scheduler/scheduler_models.jac
kashmithnisakya: Pinned `extra="forbid"` on both `CreateJobRequest` and `UpdateJobRequest` model_config. Went with `forbid` over `ignore` so client typos like `intervall: 60` surface as a 422 at the API boundary instead of being silently dropped. Behavior is now explicit and stable across pydantic upgrades.

jac-scale/jac_scale/scheduler/impl/scheduler.impl.jac
kashmithnisakya: Agreed on the layering concern but skipping the change for now. The literal extract-to-helper would be cosmetic since the helper would still reach into APScheduler - it does not actually unblock the unit-test / CLI-tool case you described. The real fix needs either a pure-python cron parser (e.g. croniter) or dropping the deep validation entirely, both of which have trade-offs that are not worth taking on without a concrete consumer that wants to validate cron without APScheduler.

jac-scale/jac_scale/scheduler/impl/scheduler.impl.jac
```diff
@@ -0,0 +1,796 @@
...
+        name = job_data.get('name', '');
+        user_id = job_data.get('created_by', self._system_user_id);
+        is_walker = job_data.get('is_walker', True);
+        if not name {
+            continue;
+        }
+        if user_id != self._system_user_id and not self._user_is_live(user_id) {
+            try {
+                self._require_store().update(
+                    job_id, {'status': 'orphaned', 'orphaned_at': _utc_now_iso()}
```
MusabMahmoodh: here the orphan update is uncondtional but the same transition inside `_schedule_apscheduler_job.job_func` around line 771 is doing `where={'status': 'active'}` which is the safe one no? when two replicas are starting up togther both will write the same row. its idempotent now becuase status is just a string, but if anyone reuse status field later this will break silently. better to keep both paths consistant and pass the where filter here also.

jac-scale/jac_scale/scheduler/impl/scheduler.impl.jac
```diff
@@ -0,0 +1,796 @@
...
+APScheduler.add_job is internally thread-safe, so no per-process lock is taken.
+"""
+impl JacScaleScheduler.update_job(
+    job_id: str, spec: TriggerSpec, acting_as: str | None = None
+) -> (dict | None) {
+    existing = self.get_job(job_id);
+    if existing is None {
+        return None;
+    }
+    if acting_as is not None and existing.get('created_by') != acting_as {
```
MusabMahmoodh: small thing only - this `acting_as != created_by` check is alredy enforced by the atomic `update(..., where=where)` few lines below. the docstring itself says "predicate is evaluated atomicaly by the store", so this app-level branch is bit redudant. you still need `existing` for name/user_id/is_walker so cant remove the whole block, but the ownership early return can go - atomic write returns 0 modified and we return None same as now.

jac-scale/jac_scale/scheduler/impl/scheduler.impl.jac
```diff
@@ -0,0 +1,796 @@
...
+    where = {'created_by': acting_as} if acting_as is not None else None;
+    # DB first: if deletion fails, APScheduler still fires the job — safe.
+    # If we removed APScheduler first and the DB delete failed, the job
+    # would resurrect on next restart via register_dynamic_tasks_from_db.
+    result = self._require_store().delete(job_id, where=where);
+    if result and self._apscheduler is not None {
+        try {
+            self._apscheduler.remove_job(job_id);
+        } except Exception as e {
+            logger.debug(f"Could not remove APScheduler job {job_id}: {e}");
```
MusabMahmoodh: this except is loging at `debug` level only. if `remove_job` fails after DB delete succeded, APScheduler will keep firing the job every tick but `record_run` will silently fail because DB row is gone. operator wont see anything because debug is filtered out in prod. atleast bump to `logger.warning` no? otherwise this is a silent zomby.

jac-scale/jac_scale/impl/serve.scheduler.impl.jac
```diff
@@ -1,39 +1,105 @@
...
         );
     }
-    if trigger == 'interval' and interval is None {
-        return TransportResponse.fail(
-            code='INVALID_REQUEST',
-            message='Field "interval" (seconds) is required for trigger="interval"',
-            meta=Meta(extra={'http_status': 400})
-        );
+    return None;
+}
```
MusabMahmoodh: nit - `_storage_failure_response` is called from exactly one place which is `_err_response`. its just returning a TransportResponse or None. why two functions for one caller? bit over enginered no? can inline the isinstance check directly inside `_err_response` and drop this helper.

jac-scale/jac_scale/scheduler/scheduler.jac
```diff
@@ -0,0 +1,150 @@
...
+    ) -> tuple[list[dict], int];
+
+    def find_active_one(job_id: str) -> dict | None;
+    def update(job_id: str, updates: dict, where: dict | None = None) -> bool;
+    def delete(job_id: str, where: dict | None = None) -> bool;
+    def record_run(job_id: str, status: str, error: str | None = None) -> None;
+}
+
+obj MemoryJobStore(JobStore) {
+    has _jobs: dict[str, dict] = {};
```
MusabMahmoodh: same trap that was flaged earlier for `_store` is still here for `_jobs` (MemoryJobStore) and `_user_exists_cache` on line 108. postinit is reseting them so practically it works, but the latent issue is still there. if someone later refactor and skips postinit this will silently share state accross instances. better to be consistant with how you fixed `_store` - make them `None` and lazy init in postinit, or drop the postinit defense everywhere. mixing both approachs in same file is confusing for the next person.

docs/docs/community/release_notes/unreleased/jac-scale/5674.refactor.md
```diff
@@ -0,0 +1 @@
+- **Refactor: Scheduler hardening**: Dynamic jobs now run as the creating user (stable `user_id`) and static jobs continue to run as `__system__`; system-role accounts are no longer loginable over HTTP; `/jobs` endpoints enforce per-job ownership on GET/PUT/DELETE and now support `limit`/`offset`/`trigger`/`created_by` pagination; jobs are persisted before scheduling with MongoDB authoritative when configured; APScheduler mutations are serialized with an `RLock`; cron expressions are validated via `CronTrigger.from_crontab`; thread-pool size, misfire grace, and shutdown drain timeout are configurable under `[plugins.scale.scheduler]`; scheduled jobs record `last_run_at`, `last_status`, `last_error`, and `run_count`; graceful shutdown drains in-flight tasks on SIGTERM.
```
MusabMahmoodh: this is bit hard to read in a changelog - one paragraph stuffing every bulet point in one sentance. can we break it into small list? someting like:

```
- Scheduler: dynamic jobs now run as creating user (was: system)
- Scheduler: /jobs supports limit/offset/trigger/created_by pagination
- Scheduler: graceful drain on SIGTERM, configurable shutdown_timeout
- Scheduler: cron expressions validated via CronTrigger.from_crontab
- Auth: system-role accounts not loginable over HTTP
- ...
```

much easier to scan later when debugging or backporting.

MusabMahmoodh:
small note on testing - the 5 integration tests cover happy CRUD and ownership which is great work. but the PR title itself is "scheduler hardening" no? these edges have no coverage:

- 503 storage failure path (mongo down -> `_err_response` mapping)
- user_exists TTL cache expiry behaviour (positive cache expires after configured ttl)
- cron validation failure (the new `from_crontab` check inside `TriggerSpec.validate`)
- orphan transition (active -> orphaned in `register_dynamic_tasks_from_db`)
- admin override path (`acting_as=None` actualy bypasses the predicate, no test asserts admin can edit others)

not blocking the merge, but the hardening claim becomes more solid with atleast few of these. should be easy to add with mocks no?

PR #5676, opened 2026-04-23
perf(jac-scale): share testcontainers across `test_memory_hierarchy` tests to cut suite runtime

akindu-k:
## Problem

`test_memory_hierarchy` doubled in runtime (5 min → 10 min) because every test spun up fresh MongoDB and Redis Docker containers and tore them down again on exit.

With 9 container-using tests, the suite was paying:
- ~9 × `MongoDbContainer.start()` + `MongoDbContainer.stop()`
- ~5 × `RedisContainer.start()` + `RedisContainer.stop()`

Each Docker container start/stop cycle takes several seconds, and the teardown also included a `time.sleep(0.5)` + `gc.collect()` per test to let the container clean up. None of that time was spent testing anything — it was pure infrastructure overhead.

## Solution

Start **one** MongoDB container and **one** Redis container for the entire test module, and reset state between tests by dropping `jac_db` (via pymongo) and calling `redis.flushall()` instead of stopping and restarting containers.

### What changed in `test_memory_hierarchy.jac`

| Before | After |
|--------|-------|
| Each test calls `MongoDbContainer("mongo:latest").start()` | `_get_mongo()` lazily starts one shared container on first call |
| Each test calls `RedisContainer("redis:latest").start()` | `_get_redis()` lazily starts one shared container on first call |
| `finally` blocks call `.stop()`, `time.sleep(0.5)`, `gc.collect()` | `finally` blocks only terminate the jac server and clean up SQLite data files |
| Per-test DB cleanup: loop over `list_database_names()` + `drop_database()` | `_reset_mongo()` called at the **start** of each test |
| Per-test Redis cleanup: container stop implicitly flushes | `_reset_redis()` calls `flushall()` at the start of each test |
| `mongo_client.close()` in two tests | Removed — shared client must not be closed mid-suite |

### New helpers

```
_shared: dict            module-level mutable dict holding container/client refs
_get_mongo() -> tuple    lazy-start Mongo container; returns (MongoClient, uri)
_get_redis() -> tuple    lazy-start Redis container; returns (redis.Redis, url)
_reset_mongo() -> None   drop all non-system databases (resets jac_db)
_reset_redis() -> None   flushall() on the shared Redis client
```

Containers are registered with `atexit` so they are stopped cleanly when the test process exits — no manual teardown needed.

### Tests that need no Redis

The tests `"data persists after server restart for mongo db"`, `"sequential requests"`, `"batch_get reduces MongoDB queries"`, and `"concurrent requests do not lose edges"` start the jac server without a Redis URL. After the refactor these tests explicitly `del os.environ["REDIS_URL"]` before starting the server so a previous test's env variable cannot bleed in.

### Tests that need neither backend

`"data persists after server restart for sqlite"` and `"hash-based dirty check detects all mutation types"` are unchanged — they manage their own env vars via `env_backup` and touch no containers.

## Expected impact

Eliminating ~14 container start/stop cycles should bring the suite back to its pre-regression runtime.


akindu-k:
image

Thamirawaran (approved):

Thamirawaran:
This is awesome :fire: This is very cool. memory hierarchy  testing time reduced from 11 to 1min. Good work @akindu-k

PR #5691, opened 2026-04-25
ci: harden build-standalone.yml against expression injection

RagavRida:
## Summary

`.github/workflows/build-standalone.yml` inlines several `${{ ... }}` expressions directly into shell `run:` blocks. Because GitHub substitutes those expressions *before* bash parses the line, any value containing shell metacharacters (`$()`, backticks, `;`, etc.) would execute as code in the runner with the job's `GH_TOKEN` (which has `contents: write` — enough to push to releases and tags).

The values that flow into shell here include:
- `github.event.release.tag_name` (release event)
- `inputs.jaseci_version`, `inputs.release_tag` (workflow_dispatch / workflow_call)
- `needs.resolve-versions.outputs.{jaseci,jaclang}_version`
- `env.asset_name` (derived from the above, propagated via `GITHUB_ENV`)
- `needs.build-{jaseci,jaclang}.result` in the summary step

This PR moves each of those into an `env:` block on the step that uses it, and references them as `\"\$VAR\"` in the shell body — the pattern GitHub recommends in their [security hardening guide](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable).

`matrix.platform` is left inline because it comes from a hardcoded `include:` list and cannot be influenced by an attacker. `with:` parameters and step `if:` conditions are unaffected.

No behavior change — same commands, same outputs, same artifact and release-upload paths.

## Test plan

- [ ] Trigger via `workflow_dispatch` with a known good `jaseci_version` and confirm both `build-jaclang` and `build-jaseci` matrix jobs build, checksum, and either upload to a release tag or upload as artifacts.
- [ ] Trigger via `workflow_call` with `jaseci_version` + `release_tag` and confirm assets land on the target release.
- [ ] Confirm the summary step still renders the version table on `$GITHUB_STEP_SUMMARY`.
- [ ] (Optional) Push a benign tag like `v0.0.0-test\$(echo hi)` against a sandbox repo with the **old** workflow to demonstrate the prior injection, then re-run with the patched workflow to confirm the value is now passed safely as a literal env var.

### DIFF
--- a/.github/workflows/build-standalone.yml
+++ b/.github/workflows/build-standalone.yml
@@ -43,12 +43,16 @@ jobs:
 
       - name: Resolve jaclang version from jaseci dependency
         id: versions
+        env:
+          EVENT_NAME: ${{ github.event_name }}
+          RELEASE_TAG: ${{ github.event.release.tag_name }}
+          INPUT_JASECI_VERSION: ${{ inputs.jaseci_version }}
         run: |
-          if [ "${{ github.event_name }}" = "release" ]; then
-            JASECI_VERSION="${{ github.event.release.tag_name }}"
+          if [ "$EVENT_NAME" = "release" ]; then
+            JASECI_VERSION="$RELEASE_TAG"
             JASECI_VERSION="${JASECI_VERSION#v}"
           else
-            JASECI_VERSION="${{ inputs.jaseci_version }}"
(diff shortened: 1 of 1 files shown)


marsninja:

PR #5692, opened 2026-04-25
install.sh: fail closed on missing checksum or hashing tool

RagavRida:
## Summary

The `--standalone` install path in `scripts/install.sh` silently skipped binary verification in two scenarios:

1. **No hashing tool present** — when neither `sha256sum` nor `shasum` was on the host, the code set `actual=\"\$expected\"`, so the equality check at the next line trivially passed. Only a `warn` was printed; the install otherwise looked successful.
2. **`.sha256` file 404 / unreachable** — the `if curl -fsSL -o ... 2>/dev/null` wrapper meant any download failure dropped through to a `warn` and the binary was installed anyway.

Together these two paths mean a standalone install could end up running an unverified binary on a normal-looking successful run. This is the kind of thing the verification step exists to prevent.

## Changes

- Both \"can't verify\" paths now `exit 1` instead of silently continuing.
- New `--no-verify-checksum` flag for users who genuinely need to opt out (air-gapped rollouts, releases without checksums, etc.). Logged as a `warn` so it's visible.
- Help text updated.

Happy path is unchanged: when both the `.sha256` file and a hashing tool are available, the verification logic is identical to before.

## Test plan

- [ ] `bash -n scripts/install.sh` passes (done locally).
- [ ] On a host with `sha256sum`: run `bash scripts/install.sh --standalone` against a known-good release and confirm \"Checksum verified.\" still prints.
- [ ] Simulate missing `.sha256`: temporarily point `checksum_url` at a 404 — confirm the script now exits with the new error message.
- [ ] Simulate missing tool: run with `PATH` stripped of `sha256sum`/`shasum` — confirm the script exits with the new error.
- [ ] `--no-verify-checksum`: confirm the install still proceeds (with a warning) when the flag is passed.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -30,6 +30,7 @@ CORE_ONLY=false
 STANDALONE=false
 VERSION=""
 UNINSTALL=false
+NO_VERIFY_CHECKSUM=false
 
 # --- Colors and output helpers ---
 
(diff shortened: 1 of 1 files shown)


marsninja:

PR #5694, opened 2026-04-25
fix(pytest_plugin): surface Jac test-collection import errors

RagavRida:
## Summary

`pytest_plugin.JacFile.collect` wraps the test-module import in `try: ... except Exception: return []`. Any failure during `importlib.import_module(qualified_name)` — a compile error, a runtime error in an entry block, an ImportError on a sibling module — caused the file to silently drop out of collection with no diagnostic. Pytest reported \"no tests collected\" and a real regression looked identical to \"this file just has no test cases.\"

This swallowing made CI worse, not better: a broken `.jac` test file or a broken module it imports passes through cleanly until someone notices the expected test count dropped.

## Change

Remove the broad `except`. Pytest's collector framework already turns exceptions raised from `Collector.collect()` into a `CollectReport(failed=True)` rendered as `ERROR <file>` in the test report, with the original traceback intact. That's the behavior we want: loud, attributed, traceable.

Filtering already happens upstream in `pytest_collect_file` (`.impl.jac`, anything under `fixtures/`, anything that doesn't match `test_*.jac` / `*.test.jac` is rejected before `JacFile` is ever created). So the loud-failure behavior only affects files pytest had already decided were Jac test suites.

## Test plan

- [ ] Existing test suite passes (no behavior change for files that import cleanly).
- [ ] Introduce a deliberate compile error in a `test_*.jac` fixture and confirm pytest now reports `ERROR <file>` with the traceback, instead of silent zero-collection.
- [ ] Confirm `*.test.jac` annexes whose `discover_base_file` hits a benign exception still continue (the inner `try/except Exception: pass` around `discover_base_file` at lines 208–215 is unchanged — that one is intentional graceful degradation, not error swallowing).

## Release notes

`docs/docs/community/release_notes/unreleased/jaclang/5694.bugfix.md` is included. (Filename predicted as 5694; will rename if the actual PR number differs.)


RagavRida:
CI on this PR went red on `test-core-runtime` and `test-client`. Both failures are pre-existing bugs that the silent-swallow was hiding - exactly the value this PR is trying to deliver. Diagnosis below, with companion PRs.

### `test-client` — `ImportError: No bytecode found for .../test_desktop_build.jac`

Root cause: five tests in `jac-client/jac_client/tests/test_desktop_build.jac` have a `\"\"\"...\"\"\"` docstring as the first statement *inside* the `test { ... }` body. Reproduced locally with a clean cache:

```
error[E0002]: Missing ';'  (lines 10:5, 23:5, 38:5, 51:5, 66:5)
warning[W0060]: Docstrings in Jac go before the declaration, not inside the body.
```

Compile fails to emit `py_bytecode`, `meta_importer.exec_module` raises ImportError, pytest collection blows up. Before this PR the error was swallowed and pytest just reported "no tests collected" for the file.

Companion PR: **#5720** - moves the docstrings to the W0060-recommended position (matching the pattern in `test_cli.jac`). After it, the file produces bytecode and pytest collects 5 items. Two of those then fail at runtime because `DesktopConfig` has no `client_only` field - a separate bug that has been hidden in main for as long as the silent-swallow has been in place. I left that out of #5720 since it's a real product question (was the feature dropped, or were the assertions written against a planned-but-never-shipped design?).

### `test-core-runtime` — `ModuleNotFoundError: No module named 'numpy'`

Root cause: `jac/examples/littleX/littleX.test.jac` imports `numpy` and `sklearn`, but the `test-core-runtime` job only installs `jac`, `pytest`, `pytest-xdist`, `watchdog`. The example test was being collected, failed at import, and the failure was silently dropped.

Companion PR: **#5721** - adds `--ignore=jac/examples/littleX` to that one pytest invocation. The other `.test.jac` under `jac/examples/` (`manual_code/circle_pure.test.jac`) collects fine without extra deps, so the ignore is narrowly scoped. If littleX is supposed to have CI coverage somewhere, the right home is a separate job that installs the ML deps - happy to follow up if you'd prefer that route instead.

### Suggested merge order

1. #5720 (test_desktop_build docstrings) - small, clearly correct.
2. #5721 (CI ignore) - one-line workflow change.
3. Rebase this PR. CI should go green.

The two unmasked failures are exactly the kind of compile-regression-hiding-in-CI scenario the release-note fragment describes; if anything, they're stronger evidence for landing this change than for reverting it.

marsninja:

PR #5695, opened 2026-04-25
fix(jac0): handle empty switch/match bodies and misordered default arms

RagavRida:
## Summary

`jaclang/jac0.py` (the bootstrap Jac→Python transpiler) emitted invalid Python in three latent edge cases of `_emit_match` and `_emit_switch`:

1. **`match X {}`** — emitted `match X:` followed by zero `case` arms, which is a Python `SyntaxError`.
2. **`switch X {}`** — emitted nothing at all. The whole switch (including any side effects of evaluating the subject) silently disappeared.
3. **`switch X { default: ...; case 1: ...; }`** and **`switch X { default: ...; }`** — emitted `else:` with no preceding `if`/`elif`, again a Python `SyntaxError`. The bug also bites if `default:` appears anywhere ahead of value arms in source order.

These are latent because jac0 only handles bootstrap files (the production compiler is written in Jac itself), and the bootstrap files in `jac0core/` happen not to use any of these patterns today. But the codegen producing invalid Python silently is the kind of foot-gun that's much better fixed before someone steps on it.

## Change

In `_emit_match` and `_emit_switch`:

- Empty body → emit the subject as a bare expression statement, so any side effects of evaluating it still run, and emit nothing else.
- Split the `default:` arm out of `node.cases` and always place it at the end of the generated `if`/`elif`/`else` chain, regardless of where it appeared in source order.
- If only a `default:` arm is present, emit the subject (for side effects) followed by the default body — no naked `else:`.

The happy path (one or more `case` arms, optionally followed by a single `default:`) is unchanged.

## Tests

`jac/tests/compiler/test_jac0.jac` gains four new tests, each compiling a small Jac snippet with `compile_jac(...)` and asserting the output parses with `ast.parse(result)`:

- `jac0 empty match body emits valid Python`
- `jac0 empty switch body emits valid Python`
- `jac0 switch with only default emits valid Python`
- `jac0 switch with default-first reorders default to end`

All five tests in `test_jac0.jac` pass locally:

```
jac/tests/compiler/test_jac0.jac::jac0 fullbody with impl does not duplicate PASSED
jac/tests/compiler/test_jac0.jac::jac0 empty match body emits valid Python PASSED
jac/tests/compiler/test_jac0.jac::jac0 empty switch body emits valid Python PASSED
jac/tests/compiler/test_jac0.jac::jac0 switch with only default emits valid Python PASSED
jac/tests/compiler/test_jac0.jac::jac0 switch with default-first reorders default to end PASSED
```

Pre-commit hooks (incl. `jac-format`, `ruff`, `mypy`, `markdownlint-cli2`, `Validate release note fragment filenames`) all pass on the changed files.

## Release notes

`docs/docs/community/release_notes/unreleased/jaclang/5695.bugfix.md` is included.


marsninja:

PR #5698, opened 2026-04-25
fix(jac-scale): @restspec endpoints crash on unknown query params

### LINKED ISSUE #4661: Bug: Unknown query parameters crash function and walker endpoints
## Description

### Project structure:
```
my-project/
├── jac.toml
└── main.jac
```

### `jac.toml`
```toml
[project]
name = "query-param-test"
version = "1.0.0"
entry-point = "main.jac"
```

### `main.jac`
```jac

}
```

### Run and test:
```bash
```

```bash

# → 500 Internal Server Error
```

## Verified on

## Expected Behavior

## Impact

## Suggested Fix

RagavRida:
## Summary

- `@restspec` walker and function endpoints currently 500 when a request includes any query parameter that isn't in the declared signature. Browsers (`?v=...` cache busters), CDNs, and frontend frameworks all add such params routinely, which makes these endpoints brittle in production.
- Fix in `create_walker_callback` / `create_function_callback`: precompute the set of accepted parameter names from `walker_cls.init` / the function signature once at registration. At call time, drop kwargs that aren't in that set unless the callable explicitly declares `**kwargs`. Matches FastAPI's default behavior of silently ignoring undeclared query params.
- Adds two regression tests against existing fixtures (`GetWalkerWithParams`, `get_func_with_params`) hit with `?v=&_t=` extras.

Closes #4661

## Reproducer (before this PR)

```bash
# → 200 OK

# → 500 Internal Server Error
```

After this PR both return 200.

## Notes

- Filter happens at the `serve.endpoints.impl.jac` layer (close to the user callable) rather than at the lower `jfast_api.impl.jac` query-param forwarding code, to avoid changing semantics for any other consumers of `_create_endpoint_function` that may rely on the existing `**kwargs` forwarding for non-`@restspec` paths.
- If a user's function/walker declares its own `**kwargs`, every query param is still forwarded — preserves intentional pass-through.
- `signature(...)` is wrapped in `try/except`; non-introspectable callables fall back to permissive behavior so we don't regress for any edge case.

## Test plan

- [x] Two new regression tests added to `jac-scale/jac_scale/tests/test_restspec.jac`:
  - `walker tolerates unknown query params`
  - `function tolerates unknown query params`
- [x] CI `Run RestSpec tests` job (`pytest -x jac-scale/jac_scale/tests/test_restspec.jac`)
- [ ] CI full backend suite (`test_serve.jac`, `test_admin.jac`, etc.) to confirm no regression


marsninja:

PR #5699, opened 2026-04-25
fix(jaclang): read-only walker clobbers concurrent writes in local mode

### LINKED ISSUE #5696: Anchor field mutations silently revert to defaults across walker invocations (local mode)
# Anchor field mutations silently revert to defaults across walker invocations (LOCAL MODE ONLY — MongoDB-backed production appears unaffected)

> **⚠️ Scope update (post-investigation):** This bug reproduces deterministically in jac-scale **local mode** (no MongoDB / no Redis — `jac start main.jac` with no flags). After deploying the same code to a MongoDB-backed environment, the symptom does **not** manifest — fields persist correctly across concurrent walker traffic. The original jacBuilder production-shape symptoms we attributed here turned out to be a **different** bug (dangling edge references in `_anchors` after class renames), which we fixed separately via DB-side cleanup.
>
> So **production with MongoDB is not affected.** This issue downgrades from "blocks all multi-pod deployments" to "local-dev correctness gap that misleads app authors during development."

> **Root cause (local mode):** jac-scale commits the entire node snapshot at walker exit, **even for read-only walkers**, causing concurrent writes to be silently clobbered when a long-running read walker exits with a stale snapshot. 30-line deterministic reproducer in [the comment below](#issuecomment-4318416554).

## Summary

In **local mode** (no MongoDB backing), field mutations made on a graph node inside one walker are silently overwritten by a later read-only walker's stale-snapshot writeback at exit. Same Python process, no error, no log — just default values surfacing.

Originally suspected to be the cause of "GitHub connection state spontaneously reverts" in [jacBuilder](https://github.com/jaseci-labs/jacBuilder) production. Investigation showed that production symptom was a **separate** issue (orphan edge anchors + dangling refs in `node.data.edges[]` arrays after a `jac_coder.nodes.ProjectMemory` → `jac_coder.core.nodes.ProjectMemory` class rename). Production with MongoDB serves concurrent walker traffic correctly.

## Why this still matters (even if production is fine)

App authors **develop locally**. The local-mode behavior is the one they iterate against. If local-mode commits behave fundamentally differently from MongoDB-backed commits, every app author writes their code on a substrate that silently drops writes under any concurrency, which:

1. **Hides real-but-rarer production bugs** (e.g. our class-rename dangling-ref issue) until deploy, because local "works" and prod "doesn't"
2. **Burns days of debugging** when the symptom does surface — every team will hit this once and lose time figuring out it's local-mode-only
3. **Misleads app authors into wrong fixes** — we initially attributed jacBuilder's prod symptom to this bug, and it's only after DB inspection that we found the real prod cause. If local mode behaved like prod, we'd have caught the real bug faster.

The local-mode reproducer is still valuable as a forcing function: either fix the local-mode commit semantics to match MongoDB-backed mode (preferred), or document the divergence prominently so app authors know what they can and can't trust about local development.

This is no longer "all in-progress jacBuilder statelessness PRs are blocked." Those PRs' graph-node-storage approach **does** work in production. But the local-mode discrepancy remains a real correctness gap.

User-visible symptom: "I just clicked Connect / Save / Submit, the system told me it succeeded, but two minutes later it acts as if I never did." No exception, no log line, just default field values surfacing.

### Setup

- Branch: [`feat/stateless-graph-persistence-phase-a`](https://github.com/jaseci-labs/jacBuilder/tree/feat/stateless-graph-persistence-phase-a) (commit `81ecadb`)
- `jac start main.jac` (no `--scale` flag — pure local-dev mode)
- Single Python process, no MongoDB, no Redis
- Deps: jaclang 0.14.0, jac-scale 0.2.14, jac-client 0.3.12, byllm 0.6.4
- macOS 25.3.0, Python 3.13

### Flow

1. Sign in via GitHub SSO (creates `UserProfile` node on user_root)
2. Create a project from a template
3. From inside the project: Source Control → GitHub → Connect GitHub
4. Complete OAuth on github.com
5. Callback to `/auth/github-connect` triggers `github_ops("exchange_code")` walker which sets three fields on `UserProfile`:
   ```jac
   profile.github_access_token = encrypt_token(access_token);
   profile.github_username = github_username;
   profile.github_connected = True;
   ```
6. UI shows GitHub as connected; `github_ops("status")` and `me` walkers correctly return `github_connected: True, github_username: "udithishanka"`.
7. Click "Publish to GitHub" → triggers `github_ops("push_new")` walker, which reads the same fields.

### Captured debug log (smoking gun)

We instrumented `exchange_code` and `push_new` with `print` statements that dump the raw field state. Same Python process, same JWT, same `user_id=ace045bf-279d-4f4f-9154-b94af0cd14a7`:

```
T1  [github_ops/exchange_code DEBUG] WROTE  user=ace045bf-... access_token_len_input=40 encrypted_len=40 encrypted_starts='gho_gwoV'
T1  [github_ops/exchange_code DEBUG] READBACK profile.github_access_token len=40 starts='gho_gwoV'
T1  github_ops "status" → {github_connected: True,  github_username: "udithishanka"}   ✓
T2  /walker/me            → {github_connected: True,  github_username: "udithishanka"}   ✓
T3  /walker/me            → {github_connected: True,  github_username: "udithishanka"}   ✓
... 8 unrelated walker calls (ide_file_ops, version_ops, git_ops, ai_chat) — none touch UserProfile fields ...
T4  [github_ops/push_new DEBUG] github_connected=False  github_username=''  github_access_token_len=0   💥
T4  github_ops "push_new" → "GitHub not connected"
```

The `READBACK` line at T1 confirms the mutation reached the in-walker view. Three subsequent reads (different walker invocations, different HTTP requests) all see the persisted value. `push_new` reads the **same `profile.user_id`**, but every field has reverted to its `has` default. No code path between T1 and T4 ever writes a default value to these fields.

### What this jacBuilder symptom turned out to be (post-investigation update)

After deploying the same jacBuilder branch to a MongoDB-backed cluster and re-running the same flow, **the symptom did not reproduce** — `github_connected` stayed `True` after exchange_code, and `push_new` saw the persisted token. The production path with MongoDB is fine.

The original jacBuilder symptom was caused by a **separate** issue:

- A previous deploy had `jac_coder.nodes.ProjectMemory` anchors in `_anchors` (old class path).
- A later deploy moved the class to `jac_coder.core.nodes.ProjectMemory`.
- jac-scale's `MongoBackend` saw the now-unresolvable `__type__` and moved those 4 anchors to `_anchors_quarantine` — but **did not** clean up edge anchors pointing at them, nor `data.edges[]` arrays in source nodes referencing those edge IDs.
- Walker traversal `[-->(?:UserProfile)]` from root would hit the dangling edge IDs and `ValueError: NodeAnchor [...] is not a valid reference!` → 500.
- Fix was DB-side: delete the 4 quarantined docs + 4 orphan edge anchors + pull 7 dangling edge IDs from `node.data.edges[]` arrays.

That's a **separate jac-scale gap** — quarantining an unloadable anchor should also sweep edges that reference it (or at least make traversal lenient about dangling refs). Will file as a follow-up issue. **It is not the read-only-walker stale-snapshot bug this issue is about** — the local-mode reproducer below isolates a different mechanism that does not require any class-rename / quarantine to trigger.

## What's needed in jac-scale

The local-mode commit semantics need to match the MongoDB-backed semantics. Either:

1. **Make local mode use the same commit path as MongoDB-backed mode** — whatever MongoDB-backed mode does that avoids the stale-snapshot writeback (per-field updates? optimistic version check? read-only detection?), use the same logic in local mode. This is the highest-value fix because it eliminates the dev-vs-prod behavioral divergence that misled us.
2. **Or, if (1) is infeasible, document the divergence prominently.** Local mode should print a one-line warning at startup along the lines of "running without persistent backend; concurrent walker semantics differ from production. See <doc>." Future app authors then know what they're getting.

If (1) is being scoped, the candidate mechanisms (in order of "smallest patch that fixes it"):

- **Skip commit for read-only walkers** — if a walker never mutated any field, don't write anything back. Smallest change, fixes the exact reproducer below.
- **Field-level dirty tracking** — only commit fields the walker explicitly assigned to. Reading `p.x` shouldn't cause `p.x` to be written back at exit.
- **Optimistic-concurrency version check** — at commit time, abort (or merge) if the node was modified by another walker since this walker's snapshot was taken.

**Whatever the chosen fix:** please ensure the failure mode becomes **loud** rather than silent. Today, an app author committing a field, having it survive several reads, then seeing it spontaneously revert in local-dev is the worst-possible debugging experience — there's no exception, no log, no version conflict surfaced. Even just a debug-level log of "walker X committed field Y over a value committed by walker Z" would have saved us several days.

## Repro environment

- OS: macOS 25.3.0 (Darwin)
- Python: 3.13
- jaclang: 0.14.0 (editable from `udith/jaseci/jac`)
- jac-scale: 0.2.14 (editable)
- jac-client: 0.3.12
- byllm: 0.6.4
- jacBuilder: branch `feat/stateless-graph-persistence-phase-a` at commit `81ecadb`

## Related tracking

- [jaseci-labs/jacBuilder#226](https://github.com/jaseci-labs/jacBuilder/issues/226) — multi-pod statelessness audit (consumer side); root cause here may invalidate several of the proposed mitigations there.
- [jaseci-labs/jaseci#5628](https://github.com/jaseci-labs/jaseci/issues/5628) — app-author requirements for multi-pod jac-scale; this bug is upstream of much of what's discussed there.
- [jaseci-labs/jacBuilder#220](https://github.com/jaseci-labs/jacBuilder/pull/220) — the PR that exposed this; `_github_oauth_states` dict was removed under the assumption that storing on `UserProfile` graph node would persist correctly. It does — until any other walker reads UserProfile in parallel.

RagavRida:
## Summary

In jac-scale local mode (`SqliteMemory` persistence, no MongoDB / no Redis), a long-running read-only walker commits its stale snapshot at exit and silently overwrites writes from concurrent walkers. The `MongoBackend.sync` path already gates writeback on a hash-based dirty check; `SqliteMemory.sync` did not. This brings the SQLite path to parity.

Closes #5696.

## What was happening

```
T1  walker A writes  {value: "A=10", counter: 1}    -> disk has A=10
T2  walker B (slow read, 1s sleep) loads snapshot of A=10, never mutates
T3  walker C writes  {value: "B=11", counter: 2}    -> disk has B=11
T4  walker B exits and commits its A=10 snapshot    -> disk REVERTS to A=10
```

`SqliteMemory.sync` was reading the now-current DB state, comparing it to the in-memory snapshot, seeing them differ (B=11 vs A=10), and writing the stale snapshot back over the newer value. No exception, no log, just silently lost writes.

## The asymmetry

`MongoBackend.sync` has had this gate since landed:

```jac
current_hash = Serializer._compute_hash(anchor);
if current_hash == anchor.hash {
    continue;   # walker didn't mutate; skip writeback
}
```

`anchor.hash` is set by `Serializer.deserialize` at load time, so if the walker only read the anchor, its in-memory state still hashes to the loaded value and the writeback is skipped, no matter what concurrent writers did to the DB in between.

`SqliteMemory.sync` had no such gate. Every persistent populated anchor in `__mem__` fell into the merge logic regardless of whether the walker had mutated it.

## The fix

Two changes to `jac/jaclang/runtimelib/impl/memory.impl.jac`:

1. **Hash gate at top of per-anchor loop** in `SqliteMemory.sync` — mirrors `MongoBackend.sync`.
2. **Refresh `anchor.hash` after a successful merge-write** so subsequent syncs in the same context don't redundantly re-write. Also mirrors `MongoBackend`.

## Test

Adds `test "read-only walker snapshot does not clobber concurrent writes (#5696)"` to `jac/tests/runtimelib/test_persistence_migration.jac` that drives the exact A/B/C sequence from the issue using three `SqliteMemory` instances on the same DB file.

- Without the fix: test fails with `AssertionError: Read-only walker B clobbered C's write. Expected name='B=11', got 'A=10'`.
- With the fix: test passes.

## Why this matters even though prod (MongoDB) is fine

App authors develop locally. A divergence between local and prod commit semantics means correctness gaps surface only at deploy. The issue author lost days chasing a production symptom they thought was this bug, and only realized it was a separate issue (dangling edge anchors after a class rename) after DB inspection. With local-mode behaving like MongoDB-backed mode, that misattribution doesn't happen.

## Test plan

- [x] New regression test added and verified fail-without / pass-with fix
- [x] Full `jac/tests/runtimelib/` suite passes locally (164 tests, no regressions)
- [ ] CI test-jaseci.yml workflow
- [ ] CI jac-scale `test_memory_hierarchy.jac` (the existing "no unnecessary saves on read-only operations" MongoDB test should already be passing; the SQLite analogue is now covered by the new persistence test)


marsninja:

PR #5700, opened 2026-04-25
fix(type-check): false positive E1031 on classmethod calls returning Self

### LINKED ISSUE #5688: False positive Type Check Errors
### Description

1. `Self` return type from stdlib classmethods isn't resolved (e.g. `datetime.now(...).isoformat()`)
```jac
import from datetime { datetime, timezone }

with entry {
    now = datetime.now(timezone.utc).isoformat();
    print(now);
}
```

RagavRida:
## Summary

`datetime.now(timezone.utc).isoformat()` (and any user-defined `@classmethod ... -> Self` chained access) raised:

```
error[E1031]: Cannot access attribute "isoformat" for type "Self"
```

The call expression typed as the raw `Self` TypeVar instead of the receiver class, then every chained attribute fell into the unbounded-TypeVar branch in atom-trailer evaluation and emitted E1031.

Closes #5688.

## Root cause

`_collect_type_var_bindings` strips the implicit `cls` / `self` parameter for bound method / classmethod calls (lines 1170-1180), since Python auto-binds it. But that strip leaves no source from which to solve `Self`: the `cls` param's annotation (`type[Self]`) is gone before the binding loop runs, and the user doesn't pass `cls` as a positional argument.

For instance methods, `specialize_member_type` already substituted Self at attribute-access time. But classmethod calls via `Class.method(...)` ended up in `validate_call_args` with `return_type` still as the raw `Self` TypeVar.

## Fix

Pre-seed `bindings["Self"]` from the receiver expression when the first parameter will be stripped. Receiver class type comes from `expr.target.target` on `AtomTrailer` attribute access. Bound classmethod becomes `bindings["Self"] = ReceiverClass.clone_as_instance()`.

Carefully gated:

- **Only** when the first param will actually be stripped (`is_self`, or `is_cls` AND `ClassMethod` flag). The staticmethod path `object.__new__(self.__class__)` keeps `cls` as a real positional and continues to bind Self via `assign_type_to_type_var` on `type[Self]`. That path is covered by the existing `checker_self_type_attr_access.jac` fixture and is unaffected.
- Skipped for unbound `Class.method(instance, ...)` and constructor calls; those have their own binding paths.

## Cases verified

| Case | Before | After |
|---|---|---|
| `datetime.now(timezone.utc).isoformat()` (issue's repro) | E1031 | ok |
| `Box.make().kick()` (user `@classmethod -> Self`) | E1031 | ok |
| `Builder().again().total()` (instance method `-> Self`) | ok (already worked) | ok |
| `obj: datetime = datetime.now(...); obj.isoformat()` (with annotation) | ok (already worked) | ok |
| `object.__new__(self.__class__)` (staticmethod with `cls: type[Self]`) | ok | ok |

## Tests

- New fixture `checker_self_classmethod_return.jac` covering stdlib classmethod, user `@classmethod` with explicit `cls: type[Self]`, and instance method.
- New `test "self_return_resolves_via_receiver_5688"` in `test_checker_pass.jac`. Verified fail-without (4 errors) / pass-with (0 errors).
- Full `jac/tests/compiler/passes/main/test_checker_pass.jac` (178 tests) and `jac/tests/language/` (220 tests) green.

## Test plan

- [x] New regression test added and verified fail-without / pass-with fix
- [x] All `self*` tests in `test_checker_pass.jac` (9) pass
- [x] Full checker_pass suite (178 tests) passes
- [x] Full language suite (220 tests) passes
- [ ] CI test-jaseci.yml workflow


kugesan1105:
Welcome and thanks for the contribution! @RagavRida  🎉 Nice catch on the direct-call case.
but One gap I hit testing on your branch:
```jac
import from datetime { datetime, timezone }

with entry {
    datetime.now(timezone.utc).isoformat();   # ✓ fixed
    f = datetime.now;
    x= f();                        # still leaks Self
    x.isoformat();            # E1031: Cannot access attribute isoformat for type Self
    #^^^^^^^^^^^^^

}
```
The seed lives at call-time, so it only fires when the `AtomTrailer `is the direct call target. Indirection (variable assignment, callback, IDE hover) still sees `() -> Self`. Might be worth moving the binding up to member-access in `type_evaluator.impl.jac`:`631 `so the receiver context travels with the function.

image

Issue #5673, opened 2026-04-22
No API for authenticated apps to retrieve the current user's identities (username/email/SSO details)


ahzan-dev:
# No API for authenticated apps to retrieve the current user's identities (username/email/SSO details)

## Summary

After the identity+credential auth refactor, there is **no supported way** for a Jac app built on `jac-scale` to retrieve identity information (username, email, SSO provider) for the authenticated caller. JWTs carry only `user_id`, the SSO callback redirect exposes only the token, and the HTTP surface has no `/user/me` (or equivalent) endpoint. Inside a walker, `here.__jac__.id` gives the user's root, but `UserManager` lacks a `root_id → user_id` reverse lookup.

The net effect: after Google/GitHub/Apple SSO, applications have no way to show the user's name, email, or avatar, and fall back to displaying `"User"` or an empty string. This isn't a cosmetic bug — it's a gap in the public auth API that blocks any app doing SSO.

## Environment

- `jac-scale` main (commit `f19e6eb859`)
- `jaclang` main
- Tested locally with SQLite identity storage
- Affects any `jac start --scale` deployment, including Kubernetes
- JWT claims confirmed in `jac-scale/jac_scale/impl/user_manager.impl.jac:139-155`

### 1. JWT payload has no identity info

Post–identity+credential refactor, JWTs contain only:
```json
{ "user_id": "<uuid>", "exp": ..., "iat": ..., "role": "user" }
```
Source: `JacScaleUserManager.create_jwt_token()` — `user_manager.impl.jac:139-155`.
Documented in `docs/community/breaking-changes.md:71-85`.

Clients cannot decode the JWT and recover any human-readable identifier.

### 2. SSO callback redirect carries only the token

`_create_sso_callback_wrapper` (`serve.auth.impl.jac:190-214`) redirects to:
```
{frontend_url}?token={token}
```
No `username`, `email`, `name`, or `provider` is passed to the frontend, even though the backend has all of that at the moment of redirect (it's in `SSOUserInfo` from the OAuth provider — see `sso/google.jac:70`, `sso/github.jac:70`).

### 3. No `/user/me` (or equivalent) endpoint

The only auth-related HTTP endpoints exposed by jac-scale are:
- `POST /user/register`
- `POST /user/login`
- `POST /user/refresh-token`
- `PUT  /user/password`
- `GET  /sso/{platform}/{operation}`
- `GET/POST /sso/{platform}/callback`
- `/api-key/*`

[...]

PR #5708, opened 2026-04-27
feat(auth): add optional user profile and /user/me endpoint

kashmithnisakya:
## Summary

Adds optional, free-form user profile fields (firstname, lastname, address, postcode, ...) to `POST /user/register`, a new authenticated `GET /user/me` endpoint, and SSO profile population from Google/GitHub/Apple OAuth callbacks. The `jac-client` `jacSignup` helper accepts an optional `profile` arg.

## What's new

- **Optional `profile` on register.** Devs can attach arbitrary `{string|number|boolean}` fields when creating a user. Profile is optional, omit it for the same behavior as before.
- **`GET /user/me` endpoint.** Bearer-auth; returns the authenticated user's identities, role, status, profile, and timestamps. Credentials, password hashes, and SSO `external_id` are never returned.
- **SSO profile population.** Google, GitHub, and Apple SSO callbacks now write `display_name`, `first_name`, `last_name`, `picture` into `profile.sso.<platform>` on the user record. Refreshed on every SSO login so values stay current without overwriting developer-set fields outside the `sso` namespace.
- **`jac-client` `jacSignup(identities, credential, profile?)`.** Forwards `profile` to the register call when provided. Returns `{success, user_id}` on success; the caller calls `jacLogin` to start a session (matches the server contract that `/user/register` does not issue tokens).
- **Storage.** Profile is stored as a `profile` sub-document on the existing user record. Mongo handles arbitrary fields natively; SQLite gets a new `profile` JSON column. No new collections.

## Validation (security limits)

Profile dicts are bounded in `auth_models.jac::validate_profile`:

| Limit | Value |
|---|---|
| Max keys | 20 |
| Max key length | 64 |
| Max value length | 1024 chars |
| Max total size | 8192 bytes (JSON) |
| Allowed value types | `str`, `int`, `float`, `bool` |
| Key pattern | `^[a-zA-Z][a-zA-Z0-9_]{0,63}$` |

Key pattern blocks MongoDB operators (`$where`), dot-path injection, JS prototype pollution (`__proto__`), and homoglyph/control-char attacks. Profile is always stored under the `profile` sub-key, never spread into the user-doc root, so collisions with `role`, `user_id`, etc. cannot escalate privileges.

## DB user document structure

```jsonc
{
  "user_id": "550e8400-e29b-41d4-a716-446655440000",
  "status": "active",
  "role": "user",
  "identities": [
    {
      "type": "email",
      "value_raw": "user@example.com",
      "value_normalized": "user@example.com",
      "verified": false,
      "is_recovery": true
    },
    {
      "type": "sso",
      "provider": "google",
      "external_id": "<google-numeric-id>",
      "verified": true,
      "linked_at": "2025-01-15T10:30:00.000000+00:00"
    }
  ],
  "credentials": [
    { "type": "password", "password_hash": "<bcrypt-hash>" }
  ],
  "root_id": "<32-hex-chars>",
  "profile": {
    "firstname": "Alice",
    "lastname": "Doe",
    "sso": {
      "google": {
        "display_name": "Alice Doe",
        "first_name": "Alice",
        "last_name": "Doe",
        "picture": "<google-cdn-picture-url>"
      }
    }
  },
  "created_at": "2025-01-15T10:30:00.000000+00:00",
  "updated_at": "2025-01-15T10:30:00.000000+00:00"
}
```

The `profile.sso` namespace is server-managed (written by the SSO callback). Developer-set fields live as flat keys at the top level of `profile`.

### `POST /user/register` (extended)

Request:
```json
{
  "identities": [{"type": "username", "value": "myuser"}],
  "credential":  {"type": "password", "password": "<password>"},
  "profile":     {"firstname": "Alice", "postcode": "00000"}
}
```

Response (201):
```json
{ "ok": true, "data": { "user_id": "550e8400-e29b-41d4-a716-446655440000", "message": "User registered successfully" }, "meta": { ... } }
```

Errors: `400 VALIDATION_ERROR` if profile fails the limits above, otherwise unchanged behavior.

### `GET /user/me` (new)

Auth: `Authorization: Bearer <token>`

Response (200):
```json
{
  "ok": true,
  "data": {
    "user_id": "550e8400-e29b-41d4-a716-446655440000",
    "role": "user",
    "status": "active",
    "identities": [
      {
        "type": "email",
        "value": "user@example.com",
        "verified": false,
        "is_recovery": true
      },
      {
        "type": "sso",
        "provider": "google",
        "verified": true,
        "is_recovery": false
      }
    ],
    "profile": {
      "firstname": "Alice",
      "lastname": "Doe",
      "sso": {
        "google": {
          "display_name": "Alice Doe",
          "first_name": "Alice",
          "last_name": "Doe",
          "picture": "<google-cdn-picture-url>"
        }
      }
    },
    "created_at": "2025-01-15T10:30:00.000000+00:00",
    "updated_at": "2025-01-15T10:30:00.000000+00:00"
  },
  "meta": { "extra": { "http_status": 200 } }
}
```

Internal fields stripped from the response: `credentials`, `password_hash`, `value_normalized`, `external_id`, `root_id`. SSO identities expose `provider` instead of `value`.

Errors: `401 UNAUTHORIZED` (missing/expired token), `404 NOT_FOUND` (user deleted).

### `GET /sso/{platform}/callback` (extended)

Same response shape as before. New side effect: writes `profile.sso.<platform>` from the OAuth response, both on first signup and on every subsequent login.

## SSO architecture changes

- `SSOUserInfo` extended with `display_name`, `first_name`, `last_name`, `picture` (all optional).
- `SSOUserInfo.profile_fields()` method returns the populated optional fields as a flat dict, leaving namespacing to the caller.
- `SSOUserInfo.platform` field and `SSOProvider.get_platform_name()` abstract method removed, nothing in production read them. The platform identifier flows from the URL path through `sso_callback(platform)` to `link_sso_account(..., platform, ...)` to identity storage's `provider` field as a single source of truth.

## Test plan

- [x] 46/46 SSO tests pass, including new coverage for the existing-user profile refresh and the Apple later-login no-data case.
- [x] `jac check` passes for all modified files (no new errors beyond pre-existing mixin-resolution noise).
- [x] Manual smoke: register with profile, login, `GET /user/me` returns the profile.
- [x] Manual smoke: SSO sign-in via Google + GitHub, both linked to the same email, `GET /user/me` returns `profile.sso.google` and `profile.sso.github` populated, identities array shows both providers.
- [x] Full server integration tests run against MongoDB once CI infra is available.

## Notes

- SQLite (local-dev backend) gets the column added on table create. Existing local `.jac/data/users.db` files need to be deleted after pulling this branch. Mongo handles arbitrary fields natively.


ahzan-dev:
Thanks for picking this up — `GET /user/me` covers the core ask from #5673 and unblocks SSO apps showing the user's email/identity. 🙏

Two notes from integrating against this branch:

### 1. SDK description says auto-login, but code doesn't auto-login

The PR description states *"After successful register the client now auto-calls `jacLogin` so the caller ends up authenticated"*, but in `client_runtime.impl.jac` `jacSignup` returns `{success: True, user_id}` without calling `jacLogin` or storing a token:

```jac
if response.ok {
    user_id = None;
    try {
        data = JSON.parse(await response.text());
        if data["data"] and data["data"]["user_id"] {
            user_id = data["data"]["user_id"];
        }
    } except Exception { }
    return {"success": True, "user_id": user_id};
}
```

For our use case this is actually the behavior we want (signup → explicit login), so just flagging the description mismatch — either drop the auto-login claim from the description, or add the `jacLogin` call. Either is fine for us.

### 2. SSO path doesn't populate the new `profile` field — `/user/me.profile` is always `{}` for SSO users

This is the bigger one and would be great to address (in this PR or a follow-up). The infrastructure you added (`profile` column, `validate_profile`, `create_user(..., profile=...)`) is solid, but the SSO callback doesn't actually use it.

Trace:

| Stage | What's available | What's kept |
|---|---|---|
| Google OAuth response | `email, sub, name, given_name, family_name, picture, locale` | — |
| `SSOUserInfo` (`sso/provider.jac:12-18`) | Only `email, external_id, platform, display_name` | **5 fields dropped** |
| `sso_callback()` calls `create_user` (`impl/user_manager.impl.jac:329`) | Could pass `profile={...}` | **No `profile` arg passed** → stored as `{}` |

Even `display_name` (which **is** captured in `SSOUserInfo` at `sso/google.jac:70`) is then thrown away because `sso_callback()` calls:
```jac
self.create_user(
    identities=[{'type': 'email', 'value': email, 'verified': True}],
    credential={'type': 'password', 'password': generate_random_password()}
    # no profile=... here
)
```

Re-login doesn't backfill either: `link_sso_account()` (`:311-326`) only attaches the SSO identity and never refreshes `profile` from fresh provider data.

**Smallest fix** would be ~3 spots:
1. Extend `SSOUserInfo` with `picture`, `given_name`, `family_name`.
2. Each provider (`sso/google.jac:66-71`, `sso/github.jac:64-72`, `sso/apple.jac`) reads those from the fastapi_sso response (which already returns them).
3. In `sso_callback()`, build a profile dict from `SSOUserInfo`, validate via `validate_profile`, pass to `create_user(..., profile=...)`. Optionally namespace under `profile.sso = {...}` to avoid collision with developer-set profile fields.

Plus a `test_sso.jac` assertion that `GET /user/me → data.profile != {}` after a Google login.

Happy to file this as a follow-up issue if you'd prefer to keep this PR scoped to the endpoint + register changes.

kashmithnisakya:
Thanks for the careful review.

**1. Auto-login description mismatch.** You're right, the description is stale. We dropped auto-login during review (commit `2c95c7154`) because it hides a second request, breaks future verification flows, and fights the server contract that `/user/register` doesn't issue tokens. Explicit `jacSignup` then `jacLogin` is the intended behavior. I'll update the PR description.

**2. SSO profile population.** Good catch. The infrastructure (column, validator, `create_user(..., profile=...)`) was built so SSO can populate `profile`, but the SSO path isn't wired through yet.

I'd prefer to keep this PR scoped to register + `/user/me` and do the SSO profile work as a dedicated follow-up PR, since it touches all three providers, `SSOUserInfo`, the callback, and `link_sso_account`, and needs its own tests. I'll open it right after this one merges. Sound good?

musab-mah-7 (approved):

kugesan1105:
Nice work on this! 🎉 A few small follow-up ideas — none are blockers:

- Profile validation could ride on Pydantic's `AfterValidator` so FastAPI returns 422 automatically. That drops the try/except in the register handler, and `/user/me` could use a typed `MeResponse` model instead of hand-building the response dict.
- `authenticate_or_fail` already exists in `utils.jac` and the scheduler uses it — the auth endpoints could pick it up to drop the repeated bearer-token + 401 boilerplate.
- `jacSignup` currently swallows JSON parse errors silently in the success path — returning `success: False` with the error would be safer.

Would be great to see these tackled in a follow-up PR! 🙌

kugesan1105:
also @kashmithnisakya 
- No `PATCH /user/me` for editing profile after registration — `update_user_fields` already exists in storage, would just need an endpoint and to reuse validate_profile. Worth adding so users can update firstname/lastname/etc. without going around the API.
- Small edge case — `sso `isn't reserved as a key in `validate_profile`, so if a dev registers with` profile = {"sso": "anything"}`, the validator accepts it. Next SSO login then crashes trying to merge into a non-dict, locking the user out. Worth reserving `sso `as a server-managed key in the validator.

Issue #5711, opened 2026-04-27
refactor(byllm): reduce bloat in parallel tool-calling PR #5590


marsninja:
PR #5590 introduced parallel tool calling for byllm. The core dispatch engine is well-designed, but the implementation has several areas of unnecessary complexity that should be cleaned up:

## 1. Duplicated dispatch logic in invoke vs streaming paths
The finish-tool split → batch → `dispatch_batch` → reorder pattern is repeated nearly identically in both `BaseLLM.invoke` and `BaseLLM._invoke_streaming` (~90 lines duplicated). Should be extracted into a shared helper.

## 2. Over-engineered LLM prompt injection (~50 lines in `make_model_params`)
The "intelligent scheduling" system mutates every tool description with `[PARALLEL-SAFE]`/`[SEQUENTIAL]` tags AND injects multi-line scheduling rules into the system prompt. LLMs already understand how to batch independent tool calls — this prompt stuffing burns tokens on every request for marginal benefit. Consider removing or making it opt-in only (currently opt-out via `parallel_hint=False`).

## 3. Thread-in-thread timeout in `_call_sync`
`_call_sync` spawns a `threading.Thread` for timeout handling, but it's already running inside a `ThreadPoolExecutor` worker. `concurrent.futures` supports `future.result(timeout=...)` natively — the manual thread + `result_box`/`error_box` pattern adds unnecessary complexity.

## 4. `asyncio.run()` per async tool call
Each async tool creates a fresh event loop via `asyncio.run()` inside the thread pool. Multiple concurrent async tools could share a loop instead.

## 5. Timing metrics should be a separate concern
`llm_timing` and `tool_timing` StreamEvents were added inline in the streaming loop. These are useful but orthogonal to parallel dispatch — should be split into a separate PR or at minimum not interleaved with the dispatch refactor.

## 6. Brittle source-grep tests
Several of the 777 lines of tests are string searches on source code (e.g., verifying `dispatch_batch` appears in basellm source). These test code existence rather than behavior and will break on any refactor.

## Suggested action
~30-40% of the added lines could be cut or deferred without losing the parallel dispatch feature. The core `parallel.jac` module and config plumbing are solid — the cleanup is mostly in the integration points.

Ref: #5590

RagavRida:
[...]

PR #5713, opened 2026-04-27
refactor(byllm): simplify _call_sync timeout via Future.result()

RagavRida:
## Summary

Addresses item 3 of #5711.

`_call_sync` runs inside `_WORKER_POOL` (a `ThreadPoolExecutor`) yet spawned its own `threading.Thread` with manual `result_box`/`error_box` slots to capture the tool result and any raised exception. `concurrent.futures.Future` already provides exactly those semantics.

This PR replaces the manual plumbing with a one-shot `ThreadPoolExecutor` + `Future.result(timeout=...)`. `shutdown(wait=False)` preserves the prior `daemon=True` leak behavior: on timeout the in-flight tool keeps running in the background and the outer pool worker is freed immediately.

## Behavior parity

Smoke-tested four paths against the new implementation:

- `timeout=None` returns the tool's value inline
- Tool finishes within timeout returns the value
- Tool exceeds timeout raises `TimeoutError` with the same `Tool '{name}' timed out after {t}s` message format
- Tool exceptions propagate with original type and message

The 40 existing tests in `jac-byllm/tests/test_parallel.jac` pass unchanged.

## Test plan

- [x] `pytest -x jac-byllm/tests/test_parallel.jac` (40 passed locally)
- [x] Manual smoke test for the four `_call_sync` paths above
- [ ] CI green on `test-jaseci` / `Run package tests` job


marsninja:

PR #5714, opened 2026-04-27
refactor(byllm): extract _split_finish_tc and _batches_for helpers

RagavRida:
## Summary

Addresses item 1 of #5711.

`BaseLLM.invoke` and `BaseLLM._invoke_streaming` both had the same finish_tool split loop and the same `batches = [non_finish] if parallel else [[tc] for tc in non_finish]` construction inline. This PR pulls both into module-level helpers next to `_should_parallelize`:

- `_split_finish_tc(tool_calls)` returns `(non_finish_calls, finish_tc)`
- `_batches_for(non_finish_calls, mt_run)` returns the list of batches to dispatch

Both call sites now read as:

```jac
(non_finish_calls, finish_tc) = _split_finish_tc(resp.tool_calls);
if non_finish_calls {
    batches = _batches_for(non_finish_calls, mt_run);
    ...
}
```

Net: -3 lines, but more importantly the two paths now share one definition for the split and grouping rules. The streaming path keeps its `tool_call` / `tool_result` / `tool_timing` event yields and `max_result_len` truncation, since those are genuinely streaming-only.

## Test plan

- [x] `pytest -x jac-byllm/tests/` (155 passed locally)
- [ ] CI green on `test-jaseci` / `Run package tests` job

## Notes

The original issue described "~90 lines duplicated"; the actual exact duplication is the ~15 lines extracted here. The middle dispatch loop differs between paths because streaming yields events that invoke does not. Happy to do a deeper unification (single generator that both paths consume) if that is preferred over the helper-extraction approach taken here.

### DIFF
--- a/docs/docs/community/release_notes/unreleased/byllm/5714.refactor.md
+++ b/docs/docs/community/release_notes/unreleased/byllm/5714.refactor.md
@@ -1,0 +1,1 @@
+- **Refactor: extract finish_tool split and batch grouping helpers**: Pull duplicated `non_finish_calls`/`finish_tc` split and `batches` construction out of `BaseLLM.invoke` and `BaseLLM._invoke_streaming` into `_split_finish_tc` and `_batches_for`. No behavior change. Addresses item 1 of #5711.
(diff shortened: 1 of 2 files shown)


marsninja:

PR #5718, opened 2026-04-27
surpress-health-check-logs

juzailmlwork:
## feat(jac-scale): configurable health-check log suppression

Health-check and documentation endpoints (`/`, `/docs`, `/openapi.json`, `/health`, `/healthz`, `/healthz/ready`, `/healthz/live`) generate constant access log noise in production — especially in Kubernetes where liveness and readiness probes hit these paths every few seconds. This PR adds an opt-in server config flag to suppress those entries.

## What changed

- Added `HealthCheckFilter` — a `logging.Filter` subclass that drops uvicorn access log records matching the above paths.
- Added `suppress_health_check_logs` under `[plugins.scale.server]` in `jac.toml`. Defaults to `false` (all logs shown). Set to `true` to enable suppression.
- Filter is only attached to the console log handler when the flag is enabled; no behaviour change for existing deployments.

## Usage

```toml
[plugins.scale.server]
suppress_health_check_logs = true
` ``

### CODE REVIEW

File: jac-scale/jac_scale/jserver/impl/jfast_api.impl.jac [hunk 1]
```diff
@@ -65,6 +65,21 @@ class ConsoleLogHandler(logging.Handler) {
     }
 }
 
+"""Filter that suppresses noisy health-check access log entries."""
+class HealthCheckFilter(logging.Filter) {
+    _SUPPRESSED = {'"GET /docs HTTP','"GET / HTTP','"GET /openapi.json HTTP','"GET /health HTTP'};
```
udithishanka: this misses the most spammy endpoint. jac-scale's liveness endpoint is /healthz, not /health (registered in serve.core.impl.jac around L356, and K8s probes in targets/kubernetes/monitoring.jac and ingress.jac all hit /healthz). the substring "GET /health HTTP" doesn't match "GET /healthz HTTP" because of the trailing z, your own test even asserts /healthz is NOT suppressed. so K8s probes (every 5-10s per pod) will still flood pod logs and the PR misses its stated goal.

```suggestion
    _SUPPRESSED = {'"GET /docs HTTP','"GET / HTTP','"GET /openapi.json HTTP','"GET /health HTTP','"GET /healthz HTTP'};
```

and update the corresponding allowed-list entries in test_serve.jac so /healthz moves to the suppressed set.

jac-scale/jac_scale/tests/test_serve.jac
```diff
@@ -1960,6 +1976,134 @@ test "server startup - port fallback integration" {
...
+# HEALTH CHECK FILTER UNIT TESTS
+# =============================================================================
+"""Test HealthCheckFilter suppresses noisy health-check access log entries."""
+test "health check filter suppresses health endpoint log records" {
+    import logging;
+
+    # Inline the filter class so we test the exact logic without needing
+    # to export it from the impl module.
+    class _HCF(logging.Filter) {
+        _SUPPRESSED = {'"GET /docs HTTP','"GET / HTTP','"GET /openapi.json HTTP','"GET /health HTTP'};
```
udithishanka: the test defines _HCF as an inline copy of HealthCheckFilter instead of importing the real class, so any future change to _SUPPRESSED in the impl file (e.g. someone removes /openapi.json) won't fail these tests. the comment justifies this with "without needing to export it from the impl module" but HealthCheckFilter is already at module scope in jfast_api.impl.jac and is importable as is. drop the inline _HCF in both tests and import the real one:

```
import from jac_scale.jserver.impl.jfast_api { HealthCheckFilter }
```

then use HealthCheckFilter() directly. otherwise these tests aren't actually validating the implementation.

udithishanka (changes requested):
2 issues to address.

jac-scale/jac_scale/jserver/impl/jfast_api.impl.jac
juzailmlwork: this is fixed and now instead of using /docs as the health check endpoint this pr is using /healthz as the endpoint

jac-scale/jac_scale/tests/test_serve.jac
juzailmlwork: this is fixed

udithishanka (approved):

udithishanka (approved):

PR #5720, opened 2026-04-27
fix(jac-client): move test_desktop_build.jac docstrings before declarations

RagavRida:
## Problem

`jac-client/jac_client/tests/test_desktop_build.jac` has five tests, each with a `\"\"\"...\"\"\"` docstring as the first statement *inside* the `test { ... }` body:

\`\`\`jac
test \"client_only config is supported\" {
    \"\"\"Verify DesktopConfig supports client_only field.\"\"\"
    import jaclang;
    ...
}
\`\`\`

The Jac parser rejects this. Compiling the file (clean cache) emits:

\`\`\`
error[E0002]: Missing ';'
  --> .../test_desktop_build.jac:10:5, 23:5, 38:5, 51:5, 66:5
warning[W0060]: Docstrings in Jac go before the declaration, not inside the body.
                Move this string before the enclosing test/can/obj/etc.
\`\`\`

With those errors the compile result has no `py_bytecode`, so `meta_importer.exec_module` raises:

\`\`\`
ImportError: No bytecode found for .../test_desktop_build.jac
\`\`\`

This was previously masked because `pytest_plugin.JacFile.collect` swallowed every exception and returned `[]` (a silent zero-collection — see #5694, which removes the swallow). On the `test-client` and `test-core-runtime` jobs the file just reported "no tests collected" and the CI ran green.

## Fix

Move each docstring above the `test \"...\"` declaration, exactly where W0060 says they belong. This matches the established pattern in `jac-client/jac_client/tests/test_cli.jac` and the other Jac test suites in the repo.

After this change:

\`\`\`
$ python -m pytest jac-client/jac_client/tests/test_desktop_build.jac --collect-only
collected 5 items
\`\`\`

The file now produces bytecode and pytest collects all five tests.

## Note on remaining failures (out of scope for this PR)

Once collection works, two of the five tests fail at runtime - they assert that `DesktopConfig` has a `client_only` field that is **not** present in the current default config. That's a separate concern (either the feature was never implemented or the assertions were stale); this PR is strictly about restoring collection so that gap is visible. Three tests pass: \`is_client_only method exists\`, \`JAC_BUILD env var set during build\`, and \`client_only skips sidecar bundling\`.

## Test plan

- [x] Clean `.jac` cache + `python -c \"...compiler.compile(...)\"` reports `errors: 0` and `bytecode: True` on the changed file.
- [x] \`pytest --collect-only\` collects 5 items (was 0 with collection error after #5694).


marsninja:

PR #5721, opened 2026-04-27
ci: skip jac/examples/littleX in test-core-runtime job

RagavRida:
## Problem

`jac/examples/littleX/littleX.test.jac` imports modules that are not installed in the `test-core-runtime` CI job:

```jac
import datetime;
import numpy;
import from sklearn.feature_extraction.text { TfidfVectorizer }
import from sklearn.metrics.pairwise { cosine_similarity }
```

But the job only installs:

```yaml
pip install -e jac
pip install pytest
pip install pytest-xdist
pip install watchdog
```

So `import numpy` raises `ModuleNotFoundError` at collection time. Until #5694, that error was silently swallowed by `pytest_plugin.JacFile.collect` and the file just reported "no tests collected." Once #5694 surfaces the import error, the job goes red.

## Fix

Add `--ignore=jac/examples/littleX` to the `test-core-runtime` pytest invocation:

```diff
-      run: pytest -x jac -n auto --ignore=jac/tests/compiler
+      run: pytest -x jac -n auto --ignore=jac/tests/compiler --ignore=jac/examples/littleX
```

The other `.test.jac` file under `jac/examples/` (`manual_code/circle_pure.test.jac`) collects and runs fine without extra deps - verified locally - so the ignore is narrowly scoped to littleX.

## Out of scope

If littleX is supposed to have CI coverage somewhere, the right home is a job that installs `numpy` + `scikit-learn`. That's a maintainer decision and a separate change. This PR just stops the runtime job from crashing on a file it can never have run.

## Companion PRs

- #5694 (the change that surfaces this and other previously-hidden collection errors)
- #5720 (the test_desktop_build.jac docstring fix - same root cause: silent collection-time failure)

## Test plan

- [x] `git diff` is one line in `.github/workflows/test-jaseci.yml`.
- [ ] CI green on this PR's own runs (this PR doesn't depend on #5694; it's safe regardless of merge order).


marsninja:

PR #5722, opened 2026-04-27
feat(byllm): auto-compaction for long ReAct loops

Jayanaka-98:
## Summary

Agents running many tool-calling iterations eventually exhaust the model's context window, causing the call to fail. This PR adds automatic compaction: after each ReAct iteration byLLM checks `prompt_tokens / ctx_window` and, when the ratio exceeds a configurable threshold, summarises old tool-call rounds into a single message so the loop can continue indefinitely.

Key additions:

- **Threshold-based compaction** - fires proactively before the limit is hit
- **Emergency fallback** - catches `ContextWindowExceededError` from the provider, compacts, and retries once
- **Consecutive-compaction guard** - raises `CompactionNotEffectiveError` if compaction fires twice in a row with no reduction, preventing infinite loops
- **`on_compaction` hook** - drop-in replacement for the built-in summarisation; receives `(messages, keep_recent)` and returns the compacted list
- **`compaction_model`** - use a separate (cheaper) model for summarisation; inherits `api_key` and `base_url` from the active model
- **`ctx_window` on `BaseLLM` / `ModelPool`** - explicit override for models not in LiteLLM's registry; `ModelPool` takes `min()` across members when unset
- **`[plugins.byllm.compaction]` config section** - all knobs configurable globally in `jac.toml`, overridable per-call or per-object
- **`MockLLM` tuple format** - `(payload, usage_dict)` entries in `outputs` inject `prompt_tokens` for deterministic compaction testing without a real API call

Default behaviour is enabled at 80% threshold. Existing code is unaffected - set `compaction_enabled=False` per-call or `enabled = false` in `jac.toml` to opt out.


jac-byllm/byllm/llm.impl/basellm.impl.jac
```diff
@@ -126,8 +384,27 @@ impl BaseLLM.invoke(mt_run: MTRuntime) -> object {
...
+            if just_compacted {
+                raise CompactionNotEffectiveError(
+                    "Context window exceeded again after emergency compaction - "
+                    "provide a custom on_compaction hook, increase ctx_window, "
+                    "or switch to a model with a larger context window."
+                );
+            }
+            logger.warning(
+                "byLLM compaction: context window exceeded, triggering emergency compaction."
+            );
```
MalithaPrabhashana, line 401: This block looks the same as lines `1031–1052`. Maybe pull it into a helper so we don't have to update both if something changes?

jac-byllm/byllm/llm.impl/basellm.impl.jac
```diff
@@ -60,6 +60,257 @@ def _apply_prompt_caching(messages: list, tools: list | None) {
...
+        } else {
+            model_for_compaction = _copy_for_compaction(self);
+        }
+        compacted = _default_compact(msg_dicts, keep_recent, model_for_compaction);
+    }
+    mt_run.messages = compacted;
+    removed = original_len - len(compacted);
+    logger.info(
+        f"byLLM compaction: reduced message list from {original_len} to {len(
```
MalithaPrabhashana, line 305: Right now if `_compact_messages` doesn't actually shrink anything (when summarization itself fails), it returns silently and the caller assumes it worked, right? Could it raise `CompactionNotEffectiveError` directly when `removed == 0`?

Jayanaka-98: Done, with two changes working together. First, removed the `try/except` in `_default_compact`, because if the summarization call fails, the original exception propagates rather than silently returning the unchanged list (which would hide things like RateLimitError behind a misleading compaction error). Second, added a `removed == 0 and original_len >= 4` guard in _compact_messages that raises CompactionNotEffectiveError for the other case, where compaction runs but returns the same length (e.g. a hook that's a no-op). The >= 4 fence avoids false positives when the history is still too short to have anything to compact.

jac-byllm/byllm/llm.impl/basellm.impl.jac
Jayanaka-98: Good catch, pulled it into `_resolve_compaction_params(mt_run)` which returns a dict with all five compaction params. Both `invoke()` and `_invoke_streaming()` now call that instead of repeating the setup block.

MalithaPrabhashana: `_resolve_compaction_params` was already shared between both methods (called at line 345 and 964) before this PR, right? The duplicate I was pointing at is the emergency catch block itself (lines 386-403 in `invoke()` mirroring lines 1035-1052 in `_invoke_streaming()`), and the threshold-check block (lines 432-451 vs 1076-1095). Those are the ~50 lines of trigger logic still copy-pasted between the two methods. Could those go into a helper too?

PR #5719, opened 2026-04-27
refactor(byllm): make parallel scheduling hints opt-in

RagavRida:
## Summary

Addresses **item 2 of #5711** (the deslop audit of #5590).

The maintainer's audit on the parallel tool-calling cleanup:

> **2. Over-engineered LLM prompt injection (~50 lines in `make_model_params`)**
> The "intelligent scheduling" system mutates every tool description with `[PARALLEL-SAFE]`/`[SEQUENTIAL]` tags AND injects multi-line scheduling rules into the system prompt. LLMs already understand how to batch independent tool calls — this prompt stuffing burns tokens on every request for marginal benefit. **Consider removing or making it opt-in only (currently opt-out via `parallel_hint=False`).**

This PR takes the second option: flip the default so the hint is opt-in.

## Change

`jac-byllm/byllm/llm.impl/basellm.impl.jac` — one substantive line:

```diff
-    and mt_run.call_params.get("parallel_hint", True)
+    and mt_run.call_params.get("parallel_hint", False)
```

Comment block above it rewritten to describe the new default + reason. Runtime behavior (`mark_serialize`, `dispatch_batch` parallel/sequential gating, `_should_parallelize`) is unchanged — the hint only ever changed how the LLM was *asked* to plan, never how the runtime actually dispatched.

## Tests

`jac-byllm/tests/test_parallel.jac` — three tests that exercise the hint path now set `parallel_hint=True` explicitly (4-line diff total). The existing test `"parallel_hint=False suppresses hint even when parallel is on"` continues to pass on the new default. Full `test_parallel.jac` suite: **40/40 passed locally.** Full `jac-byllm` suite: 154 passed, 1 skipped.

## Docs

`docs/docs/reference/plugins/byllm.md` — section renamed `Intelligent Scheduling` → `Scheduling Hints (opt-in)`, `parallel_hint` row in the configuration table flipped to `False` default with the new opt-in instruction, and a stale comment in the example removed (the `[PARALLEL-SAFE]`/`[SEQUENTIAL]` tags don't appear by default anymore, so the comment "Intelligent LLMs won't mix..." was misleading).

## Note for users opting in

Anyone currently relying on the hint just needs to add `parallel_hint=True` to their `by llm()` call. The behavior of that path is byte-identical to the previous default. The release fragment calls this out.

## Test plan

- [x] `pytest jac-byllm/tests/test_parallel.jac` — 40 passed
- [x] `pytest jac-byllm` — 154 passed, 1 skipped
[...]

PR #5723, opened 2026-04-27
refactor(byllm): let McpClient run concurrent call_tool requests in parallel

RagavRida:
## Summary

byllm's parallel tool-calling (#5590) is real for in-process Python tools, but for tools served by an MCP server it has been silently serializing. The cause was a per-client lock in `_McpState.run` that wrapped the whole `future.result()` wait, so even though `dispatch_batch` spawned N worker threads, each thread's `client.call_tool(...)` queued behind the others on the lock.

## Demonstration

Before this change, on `upstream/main`, the new test reports:

```
AssertionError: 5 concurrent call_tool requests took 2.52s
(expected < 1.5s); the per-client lock has regressed and is
serializing requests again.
```

Five calls × 0.5s sleep = 2.5s. Perfectly serial.

After this change, the same test passes at ~0.5s wall-time. The MCP `ClientSession` was already multiplexing requests by JSON-RPC id - that is how async MCP clients are designed - byllm's lock was just defeating it. I sanity-checked this directly against the upstream SDK with a 5-coroutine `asyncio.gather` probe (one `ClientSession`, one loop): 5 × 0.5s = 0.508s wall.

## Change

Move the lock from `_McpState.run` to `McpClient.connect` / `McpClient.close`. State transitions still serialize (concurrent first-use `call_tool`s racing past the `if state.connected: return;` guard no longer double-spawn the loop and subprocess), and `call_tool` posts its coroutine and waits on the future lock-free.

Net behavior change is four lines in `mcp.impl.jac`; the rest of the diff is the indentation of `connect`/`close` bodies inside the new `with state._lock` block.

## New test

`jac-byllm/tests/fixtures/mcp_slow_server.py` is a 40-line stdio MCP server with one tool that sleeps 0.5s. The new test in `test_byllm.jac`:

```jac
test \"McpClient runs concurrent call_tool requests in parallel\"
```

fires 5 concurrent `call_tool` requests via threads and asserts wall-time stays under 1.5s. The threshold is generous for CI noise but still fails loudly (~2.5s) if the lock regresses.

## Test plan

- [x] New test `McpClient runs concurrent call_tool requests in parallel` passes locally (~0.5s wall).
- [x] Same test fails on unmodified `mcp.impl.jac` (~2.5s wall) - i.e. it is a real regression detector, not a tautology.
- [x] `pytest jac-byllm` - **155 passed, 1 skipped** (the +1 vs main is the new concurrency test).
- [ ] CI green.

## Why this matters

The common shape for MCP usage is *one server, many tools* (e.g. `jac-mcp` itself hosts a Jac source-of-truth tool surface). `dispatch_batch` was already willing to fan those calls out concurrently; this PR is what lets them actually finish concurrently.

## Related

- #5590 (the parallel tool-calling feature this PR makes real for MCP-served tools)
- #5719 / #5713 / #5714 (siblings under #5711 deslop audit; same flavor: cut redundant defensive wrappers around async machinery)


marsninja:

PR #5731, opened 2026-04-28
support loki-logs

juzailmlwork:
## Add optional Loki + Promtail log aggregation to the monitoring stack

Extends jac-scale's Kubernetes monitoring stack with an opt-in log aggregation pipeline using Grafana Loki and Promtail.

### What's new

- **`loki_enabled` config flag** — disabled by default; enable via `jac.toml` under `[plugins.scale.kubernetes]`
- **Loki deployment** — single-process log store (`grafana/loki:3.4.2`) on port 3100 (ClusterIP). Filesystem/TSDB storage backed by `emptyDir` (ephemeral; suitable for dev/staging). Runs hardened: non-root, read-only root filesystem, no capabilities
- **Promtail DaemonSet** — deployed on every node (tolerates `NoSchedule`), tails `/var/log/pods` via Kubernetes SD, labels streams by `namespace`/`pod`/`container`, and pushes to Loki
- **Grafana integration** — Loki added as a datasource (`uid: loki`) and a new **Pod Logs** dashboard provisioned automatically with a log-volume timeseries panel and a live log viewer
- **Full teardown** — `destroy()` removes all Loki and Promtail resources including ClusterRole and ClusterRoleBinding

### How to enable

```toml
[plugins.scale.kubernetes]
loki_enabled = true

### CODE REVIEW

File: jac-scale/jac_scale/tests/test_deploy_k8s.jac [hunk 1]
```diff
@@ -224,6 +224,7 @@ test "deploy all in one" {
     target_config["ingress_node_port"] = 30080;
     target_config["redis_dashboard"] = True;
     target_config["mongodb_dashboard"] = True;
+    target_config["loki_enabled"] = True;
```
kashmithnisakya, line 227: No test covers the disabled path. Could we add a case asserting Loki, Promtail, the Grafana Loki datasource, and `pod-logs.json` are not created when `loki_enabled` is false?

jac-scale/jac_scale/targets/kubernetes/monitoring.jac
```diff
@@ -1196,6 +1231,454 @@ obj MonitoringDeployer {
...
+                        'securityContext': {
+                            'seccompProfile': {'type': 'RuntimeDefault'}
+                        },
+                        'containers': [
+                            {
+                                'name': 'promtail',
+                                'image': 'grafana/promtail:3.4.2',
+                                'securityContext': {
+                                    'runAsUser': 0,
+                                    'allowPrivilegeEscalation': False,
```
kashmithnisakya, line 1615: Promtail is the only container in this file without `readOnlyRootFilesystem: True` (even node-exporter sets it). It only writes to the `positions` emptyDir, so this is safe to enable. Suggest adding it alongside `allowPrivilegeEscalation`.

jac-scale/jac_scale/targets/kubernetes/monitoring.jac
```diff
@@ -1196,6 +1231,454 @@ obj MonitoringDeployer {
...
+            'metadata': {
+                'name': cluster_role_name,
+                'labels': {'app': promtail_name, 'managed': 'jac-scale'}
+            },
+            'rules': [
+                {
+                    'apiGroups': [''],
+                    'resources': [
+                        'nodes',
+                        'nodes/proxy',
```
kashmithnisakya: With `role: pod` SD, Promtail only needs `pods` (and optionally `namespaces`). `nodes/proxy`, `services`, `endpoints` look unused by this scrape config. Intentional for parity with the upstream chart, or can we trim to least privilege?

jac-scale/jac_scale/tests/test_deploy_k8s.jac
juzailmlwork: if we have it by default those resources are not created.but adding a test case will increase the pipeline time which is already 19 minutes.but once k8s test case are broken into smaller pipeline we can add one

jac-scale/jac_scale/targets/kubernetes/monitoring.jac
juzailmlwork: this enhancement is implemented

jac-scale/jac_scale/targets/kubernetes/monitoring.jac
juzailmlwork: This is fixed good catch

udithishanka:
closing this pr. pr #6155 has the this prs commits + microservices support.

PR #5741, opened 2026-04-29
Add identity management, send-verification, password reset, and pluggable Emailer

kashmithnisakya:
## Summary

Adds five new `/user/*` endpoints (identity attachment, verification mail dispatch + consumption, password-reset flow) plus a pluggable `Emailer` abstraction so any SMTP / SendGrid / Mailgun backend can be plugged in via `jac.toml`.

`/user/add-identity` is now purely additive (it never sends mail); `/user/send-verification` dispatches the verification email and is retryable. This split makes register-with-email work end to end, allows resending verifications, and gives admins a way to attach identities silently.

Reset and verification tokens are SHA256-hashed at rest, single-use, and TTL-bounded; persisted in MongoDB when configured (with a TTL index) or in-memory otherwise. `/user/send-verification` revokes prior outstanding verify tokens for the user before issuing a new one.

---

## Configuration (`jac.toml`)

```toml
[plugins.scale.emailer]
provider = "smtp"                      # 'smtp', a registered name, or 'pkg.module:ClassName'
from_address = "noreply@example.com"

[plugins.scale.emailer.smtp]
host = "smtp.example.com"
port = 587
username = "apikey"
use_tls = true
# password = "..."  (or set EMAILER_SMTP_PASSWORD env var)

[plugins.scale.auth]
reset_token_ttl_seconds = 1800
verify_token_ttl_seconds = 86400
reset_url_template  = "https://app.example.com/reset?token={token}"
verify_url_template = "https://app.example.com/verify?token={token}"
# Rate limits (per-worker token buckets):
forgot_password_rate_per_hour   = 3      # keyed on recipient email
forgot_password_burst           = 3
send_verification_rate_per_hour = 5      # keyed on authenticated user_id
send_verification_burst         = 5
```

---

## Flow

```
register (email allowed)
   └─► /user/login → Bearer
         └─► /user/send-verification {email}    ← retryable
              └─► (email) /user/verify-identity {token}
                   └─► identity.verified = true
                        └─► /user/forgot-password → /user/reset-password
```

---

### `POST /user/add-identity`

Attach an identity (email or username) to the authenticated user. **Never sends mail** — email identities are stored as `verified=false`; use `/user/send-verification` to dispatch the email. **Requires Bearer token.**

**Request:**

```json
{
  "identity": {"type": "email", "value": "alice@example.com"},
  "is_recovery": true
}
```

**Response (`200 OK`):**

```json
{
  "ok": true,
  "data": {"status": "added", "verified": false},
  "meta": {"extra": {"http_status": 200}}
}
```

**Errors:** `401 UNAUTHORIZED`, `409 IDENTITY_TAKEN`, `404 NOT_FOUND`.

---

### `POST /user/send-verification`

Issue a verification token for an email identity on the current user and deliver it via the configured emailer. Idempotent: returns `already_verified` for verified identities. Revokes any prior outstanding verify tokens on retry. **Requires Bearer token.**

**Request:**

```json
{"identity": {"type": "email", "value": "alice@example.com"}}
```

**Response (`202 Accepted`):**

```json
{
  "ok": true,
  "data": {"status": "pending_verification", "email_sent": true},
  "meta": {"extra": {"http_status": 202}}
}
```

**Response when identity already verified (`200 OK`):**

```json
{
  "ok": true,
  "data": {"status": "already_verified"},
  "meta": {"extra": {"http_status": 200}}
}
```

**Errors:** `400 VALIDATION_ERROR` (non-email identity), `401 UNAUTHORIZED`, `404 NOT_FOUND` (identity not on this user), `429 RATE_LIMITED` (per-user budget; response carries `retry_after_seconds`), `503 EMAIL_DISABLED`.

---

### `POST /user/verify-identity`

Consume a verification token (from the email link) and flip the matching identity to `verified=true`. **No auth required** (the token is the credential).

**Request:** `{"token": "Hk7sQpL2x9aRnZ8Dm4e3vT7g..."}`

**Response (`200 OK`):** `{"data": {"status": "verified", "identity": "alice@example.com"}}`

**Errors:** `400 INVALID_TOKEN`, `404 NOT_FOUND` (identity gone or already verified by another call).

---

### `POST /user/forgot-password`

Issue a one-time reset token to the user's verified recovery email. **Always returns `200`** to avoid leaking account existence. SSO users (whose email is verified at sign-in) get a reset link as a fallback path.

**Request:** `{"identity": {"type": "email", "value": "alice@example.com"}}`

**Response (`200 OK`, regardless of whether the account exists):** `{"data": {"status": "ok", "message": "If that account exists, a reset link has been sent."}}`

---

### `POST /user/reset-password`

Consume a reset token and set a new password. Other outstanding reset tokens for the same user are revoked on success.

**Request:** `{"token": "...", "new_password": "newSecret123"}`

**Response (`200 OK`):** `{"data": {"status": "password_reset"}}`

**Errors:** `400 INVALID_TOKEN`, `404 NOT_FOUND` (user deleted between token issue and consume).

---

## Custom emailer (Python or Jac)

Subclass `jac_scale.abstractions.emailer.Emailer`, then point `provider` at it:

```toml
[plugins.scale.emailer]
provider = "myapp.email:SendGridEmailer"
```

The factory imports the class via `importlib` at startup and validates that it subclasses `Emailer`. **No jac-scale code changes required.**

---

## Examples

Two ready-to-run example projects under `jac-scale/examples/`:

- [`emailer-smtp/`](https://github.com/jaseci-labs/jaseci/tree/feat/user-add-identity-and-password-reset/jac-scale/examples/emailer-smtp) — built-in SMTP provider, no custom code
- [`emailer-sendgrid/`](https://github.com/jaseci-labs/jaseci/tree/feat/user-add-identity-and-password-reset/jac-scale/examples/emailer-sendgrid) — custom `SendGridEmailer` subclass wired via `provider = "emailer:SendGridEmailer"`

---

## Review follow-ups

Reviewer feedback addressed in-tree after the initial approval. Each row is a self-contained commit.

| # | Reviewer note | Fix |
|---|---|---|
| 1 | TOCTOU race in `/user/add-identity` (kugesan1105, MusabMahmoodh) | [`215cf2cf9`](https://github.com/jaseci-labs/jaseci/pull/5741/commits/215cf2cf9) - storage-layer uniqueness (mongo unique sparse index, sqlite PK + txn rollback), handler maps `IdentityTakenError` to 409 |
| 2 | TOCTOU `peek_token + _delete` in `consume_token` | [`104b00281`](https://github.com/jaseci-labs/jaseci/pull/5741/commits/104b00281) - `_delete`'s return is the single-use arbiter; mongo `delete_one` atomic per doc, in-memory uses `dict.pop` |
| 3 | Per-process token-store singleton silently breaks under multi-worker Uvicorn | [`98f8d7d75`](https://github.com/jaseci-labs/jaseci/pull/5741/commits/98f8d7d75) - loud startup warning whenever the store falls back to in-memory |
| 4 | Manager reached into storage primitives | [`d5f69f426`](https://github.com/jaseci-labs/jaseci/pull/5741/commits/d5f69f426) - `reset_password` lives in base `IdentityStorage`; manager is now a thin delegate |
| 5 | `reset_password_by_user_id` return not checked | [`a1efdd9c3`](https://github.com/jaseci-labs/jaseci/pull/5741/commits/a1efdd9c3) - storage gates on `user_exists_by_id`; handler returns 404 on user-gone |
| 6 | Mongo db name inconsistency | [`90fc95938`](https://github.com/jaseci-labs/jaseci/pull/5741/commits/90fc95938) - aligned to `jac_db`, index creation routed through `_get_mongo_collection` |
| 7 | Backend semantic divergence on `_set_identity_verified` | [`f96d1cae2`](https://github.com/jaseci-labs/jaseci/pull/5741/commits/f96d1cae2) - Mongo `$elemMatch` on `verified != True`, matches SQLite contract |
| 8 | `emailer_factory` try/except only wrapped import | [`d90947e26`](https://github.com/jaseci-labs/jaseci/pull/5741/commits/d90947e26) - shared `_instantiate` helper wraps the `cls(...)` call too |
| 9 | No rate limiting on outbound-email endpoints | [`582928be3`](https://github.com/jaseci-labs/jaseci/pull/5741/commits/582928be3) - `forgot-password` keyed on recipient email (silent reject preserves existence guarantee), `send-verification` keyed on user_id (429 + `retry_after_seconds`) |
| 10 | No audit log for security flows | [`6bf5c846c`](https://github.com/jaseci-labs/jaseci/pull/5741/commits/6bf5c846c) - structured events on dedicated `jac_scale.audit` logger covering `password_reset.requested/completed/failed` and `identity_verification.completed/failed` |
| 11 | Em-dashes in new code | [`1e5c09ecf`](https://github.com/jaseci-labs/jaseci/pull/5741/commits/1e5c09ecf) - replaced with regular punctuation |


kugesan1105:
### Follow-up PR @kashmithnisakya 
1. TOCTOU race in add-identity - Spotted a small race in `add-identity — the is_identity_taken()` check and the actual insert aren't atomic, so two near-simultaneous requests can both pass the check and the second one ends up as a 500 from the Mongo unique index instead of a clean 409. Should be a tiny fix: drop the precheck, wrap the insert in try/except DuplicateKeyError, map to 409.

kugesan1105 (approved):

MusabMahmoodh (commented):
A few things worth flagging before merge — specifics inline.

One general concern not tied to a line: `/user/forgot-password` and `/user/send-verification` both trigger outbound email and have no rate-limiting. The existing microservices rate-limit config doesn't apply to these routes. Without throttling (per-IP for forgot-password, per-user for send-verification) this is a straightforward abuse vector.

Also: there's no audit log for password-reset or verification-consume events — currently just `logger.debug`. For security-sensitive flows it'd be worth emitting structured audit events. Not blocking, but worth considering.

jac-scale/jac_scale/impl/auth_tokens.impl.jac
```diff
@@ -0,0 +1,188 @@
+"""Auth token store implementation: SHA256 hashing, TTL, single-use, Mongo+memory."""
+
+impl AuthTokenStore.postinit -> None {
+    auth_cfg = get_scale_config().get_auth_config();
+    self._collection = auth_cfg.get('token_collection', 'auth_tokens');
+    self._in_memory = {};
+    db_cfg = get_scale_config().get_database_config();
+    if db_cfg.get('mongodb_uri') {
+        try {
+            self._db = Db(client=get_mongo_client(), db_name='jac_scale');
```
MusabMahmoodh: The `jac_scale` database name is inconsistent with the rest of the codebase, which uses `jac_db` (see `user_manager.impl.jac:23` and the default in `db.jac:132`). Operators backing up `jac_db` would silently miss the `auth_tokens` collection.

Also, lines 13-16 reach into `self._db.client['jac_scale'][self._collection]` directly, bypassing the `Db` abstraction that was just constructed. If `db_name` ever changes those raw accesses will drift out of sync.

jac-scale/jac_scale/impl/auth_tokens.impl.jac
```diff
@@ -0,0 +1,188 @@
...
+}
+
+impl AuthTokenStore.consume_token(token: str, purpose: str) -> (dict[str, Any] | None) {
+    record = self.peek_token(token, purpose);
+    if record is None {
+        return None;
+    }
+    self._delete(record.get('token_hash', ''));
+    return record;
+}
```
MusabMahmoodh, line 104: There's a race here between `peek_token` and `_delete`: two concurrent requests can both peek the same record and proceed to reset the password before either delete lands. The window is small but the impact (silent double password change / verification) is high for a security-sensitive flow.

For the Mongo path, `find_one_and_delete` is atomic in a single op. For the in-memory path, a lock around the peek+delete pair would close it.

jac-scale/jac_scale/auth_tokens.jac
```diff
@@ -0,0 +1,59 @@
...
+import hashlib;
+import logging;
+import secrets;
+import from datetime { UTC, datetime, timedelta }
+import from typing { Any }
+import from uuid { uuid4 }
+import from jac_scale.config_loader { get_scale_config }
+import from jac_scale.db { Db, get_mongo_client }
+
+glob _auth_token_store: (AuthTokenStore | None) = None,
```
MusabMahmoodh, line 24: This singleton is per-process. Without Mongo configured and with multiple Uvicorn workers, worker A's tokens are invisible to worker B and the verification/reset flow fails intermittently with `INVALID_TOKEN`.

Suggest either logging a loud startup warning when `mongodb_uri` is unset, or gating the emailer features on Mongo being available. Silent breakage under load is the worst failure mode to debug in prod.

jac-scale/jac_scale/impl/user_manager.impl.jac
```diff
@@ -129,6 +129,50 @@ impl JacScaleUserManager.update_password(
...
+            cred.pop('algorithm', None);
+            password_set = True;
+        }
+    }
+    if not password_set {
+        credentials.append({'type': 'password', 'password_hash': new_hash});
+    }
+    storage._update_credentials(user_id, credentials);
+    return {'user_id': user_id, 'message': 'Password updated successfully'};
+}
```
MusabMahmoodh, line 142: This reaches into `storage._get_credentials` and `storage._update_credentials` directly. Those underscore methods are documented as "Abstract storage primitives" in `identity_storage.jac:47-51` and are meant to be called only by the base `IdentityStorage` class.

`update_password` already lives in the base correctly. Suggest adding a parallel `reset_password(user_id, new_password)` to the base and calling that from the manager. Otherwise any future storage backend has to expose its internals to the manager layer.

jac-scale/jac_scale/impl/serve.identity.impl.jac
```diff
@@ -0,0 +1,416 @@
...
+                message='Reset token is invalid',
+                meta=Meta(extra={'http_status': 400})
+            );
+        }
+        self.user_manager.reset_password_by_user_id(user_id, body_data.new_password);
+        # Drop any other outstanding reset tokens for this user.
+        AuthTokenStore.instance().revoke_user_tokens(user_id, 'reset');
+        return TransportResponse.success(
+            data={'status': 'password_reset'}, meta=Meta(extra={'http_status': 200})
+        );
```
MusabMahmoodh: The return value of `reset_password_by_user_id` isn't checked. If the user is deleted between token issue and consumption (the 30-minute TTL leaves plenty of room), `_get_credentials` returns `[]`, a fresh credential gets appended, and `_update_credentials` runs `UPDATE ... WHERE user_id=?` touching 0 rows. The handler still returns `password_reset` success, and the token is already consumed at this point.

The caller believes their password was changed when nothing actually happened. Suggest returning a 404 in this case.

jac-scale/jac_scale/impl/serve.identity.impl.jac
```diff
@@ -0,0 +1,416 @@
...
+            );
+        }
+        # Reject if any user already owns this identity.
+        if self.user_manager.is_identity_taken(normalized) {
+            return TransportResponse.fail(
+                code='IDENTITY_TAKEN',
+                message=f"This {id_type} is already registered",
+                meta=Meta(extra={'http_status': 409})
+            );
+        }
```
MusabMahmoodh, line 156: TOCTOU between `is_identity_taken` at line 99 and `add_identity` later: two concurrent requests can both pass the check.

On SQLite: `identity_lookups` is PK'd on `value_normalized` so the insert silently `OR IGNORE`s, but the losing request still appends a duplicate to the user's `identities` JSON column. `find_user_by_identity` will only ever resolve to the winning user, leaving the loser with a stale identity that's unreachable through lookup.

On Mongo: the unique sparse index raises `DuplicateKeyError`, which isn't caught here, so the client gets a 500 instead of a clean 409.

Suggest letting storage uniqueness do the work and catching the duplicate-key exception to convert to 409. The handler check can remain as a fast path.

jac-scale/jac_scale/impl/identity_storage.mongo.impl.jac
```diff
@@ -100,6 +100,22 @@ impl MongoIdentityStorage._add_identity(
...
+    result = self.users_col.update_one(
+        {'user_id': user_id, 'identities.value_normalized': value_normalized},
+        {
+            '$set': {
+                'identities.$.verified': True,
+                'updated_at': datetime.now(UTC).isoformat()
+            }
+        }
+    );
+    return result.matched_count > 0;
```
MusabMahmoodh, line 133: Backend inconsistency: this returns `matched_count > 0`, which is `True` even when the identity was already verified (no state change). The SQLite version at `identity_storage.sqlite.impl.jac:278-285` returns `False` in that case.

The `/user/verify-identity` handler reads this bool, so the same call behaves differently depending on backend. Suggest aligning both to return `True` only when the verified flag actually changed (pick one semantic and apply it across backends).

jac-scale/jac_scale/factories/emailer_factory.jac
```diff
@@ -0,0 +1,102 @@
...
+        cls = _import_dotted(provider);
+    } except (ImportError, AttributeError, TypeError, ValueError) as e {
+        logger.error(
+            "Failed to load emailer provider '%s': %s. Email features disabled.",
+            provider,
+            e
+        );
+        return None;
+    }
+    return cls(config=config, from_address=from_address, enabled=enabled);
```
MusabMahmoodh: The `try/except` only catches errors from `_import_dotted`. If the resolved class's constructor itself raises — e.g. the SendGrid example's `postinit` when `SENDGRID_API_KEY` is malformed, or any user subclass doing validation in `postinit` — the exception bubbles up and crashes the server at startup.

The docstring says "Returns None when no provider is configured (caller should treat this as email features disabled)", but actual behavior is mixed: sometimes `None`, sometimes a hard crash. Either wrap the `cls(...)` call as well, or document that constructor errors are intentionally loud (which may be the right call for misconfig, but should be explicit).

jac-scale/jac_scale/impl/serve.identity.impl.jac
```diff
@@ -0,0 +1,416 @@
...
+            purpose='reset',
+            identity_value=recipient_normalized or '',
+            ttl_seconds=ttl,
+            url_template=url_template
+        );
+        body_text = _build_link_email(
+            'reset your password',
+            link,
+            ttl,
+            extra_footer=' — your password will remain unchanged'
```
MusabMahmoodh: Nit: 12 em-dashes (—) across the new code (this line, plus docstrings in `auth_tokens.jac`, `emailer_factory.jac`, `emailers.impl.jac`, etc.). Not blocking, but worth replacing with regular hyphens for consistency with the rest of the codebase.

jac-scale/jac_scale/impl/auth_tokens.impl.jac
kashmithnisakya: fixed in 104b00281 - using _delete's return as the single-use arbiter: mongo delete_one is atomic per doc, in-memory switched to dict.pop. losing-race consumers now return None and surface as INVALID_TOKEN instead of double-consuming.

kashmithnisakya:
@kugesan1105 
fixed the toctou in add-identity in 215cf2cf9 - dropped the precheck as source of truth, both backends now raise IdentityTakenError on uniqueness collision and the handler maps to 409.

jac-scale/jac_scale/auth_tokens.jac
kashmithnisakya: fixed in 98f8d7d75, AuthTokenStore now emits a loud startup warning whenever it falls back to the in-memory path so the multi-worker failure mode is visible at deploy time.

jac-scale/jac_scale/impl/auth_tokens.impl.jac
kashmithnisakya: fixed in 90fc95938, switched to 'jac_db' and routed index creation through _get_mongo_collection so the abstraction is no longer bypassed.

jac-scale/jac_scale/impl/serve.identity.impl.jac
kashmithnisakya: fixed in 215cf2cf9, dropped the precheck as source of truth, both backends now raise IdentityTakenError on uniqueness collision (mongo unique sparse index, sqlite PK + txn rollback) and the handler maps to 409.

jac-scale/jac_scale/impl/user_manager.impl.jac
kashmithnisakya: fixed in d5f69f426, added reset_password to the IdentityStorage base alongside update_password and turned the manager method into a thin delegate, so future backends do not have to re-expose _get_credentials or _update_credentials.

jac-scale/jac_scale/impl/serve.identity.impl.jac
kashmithnisakya: fixed in a1efdd9c3, reset_password now gates on user_exists_by_id and returns an error dict when the user is gone, handler reads that and returns 404 instead of a false password_reset success.

jac-scale/jac_scale/impl/identity_storage.mongo.impl.jac
kashmithnisakya: fixed in f96d1cae2, tightened the Mongo filter with $elemMatch on verified != True so the update only fires on an actual state change, matched_count now mirrors SQLite's contract.

jac-scale/jac_scale/factories/emailer_factory.jac
kashmithnisakya: fixed in d90947e26, factored a shared _instantiate helper that wraps the cls(...) call too, so all three resolution paths (smtp, registered, dotted-import) log the failure and return None instead of crashing startup.

kashmithnisakya:
fixed in 582928be3, wired the existing InProcessRateLimiter into both handlers. forgot-password is keyed on the normalized recipient email so rotating IPs cannot bomb a victim, defaults to 3/hour, still returns the 200 envelope on rejection to preserve the existence-leak guarantee. send-verification is keyed on the authenticated user_id, defaults to 5/hour, returns 429 with retry_after_seconds. Both budgets are tunable under [plugins.scale.auth] in jac.toml. Buckets are per-worker, so the effective ceiling scales with worker count; a Redis-backed limiter for cross-worker correctness can follow separately.

kashmithnisakya:
audit log added in 6bf5c846c. routes structured events to the dedicated jac_scale.audit logger so ops can ship them to file/syslog/ELK via standard logging config. emits cover password_reset.requested / completed / failed and identity_verification.completed / failed with user_id, identity_value (normalized only), source_ip (honors X-Forwarded-For), and a reason on failures. no raw passwords, tokens, or auth headers ever land in the audit stream.

jac-scale/jac_scale/impl/serve.identity.impl.jac
kashmithnisakya: fixed in 1e5c09ecf, em-dashes replaced with regular punctuation across the new identity and emailer modules. the only user-visible string (forgot-password email footer) now reads naturally with a comma.

MusabMahmoodh (approved):

PR #5744, opened 2026-04-29
[byllm]: add `conversation=<list>` kwarg for stateful multi-turn `by llm()` calls

MalithaPrabhashana:
## Summary

Adds a `conversation=<list>` kwarg to `by llm(...)`. byllm reads the list as prior message context, runs the ReAct loop, and writes the persistable history back into the same list - preserving the full structured turn (user message, assistant tool-calls, `tool_call_id`-linked tool results) with byllm-nativefidelity.

byllm itself stays stateless: the list is caller-owned. Callers (agents, sessions) decide where it lives - graph node, plain global, dict on disk, and byllm just mutates it on each call.

## Usage

```jac
import from byllm.lib { Model }

glob llm = Model(model_name="claude-sonnet-4-6");
glob history: list = [];   # caller owns this

def ask(message: str) -> str by llm(
    tools=[get_weather],
    conversation=history    # ← new kwarg
);

with entry {
    ask("What's the weather in Paris?");
    # `history` now contains: user → assistant(tool_calls=[get_weather]) →
    #                         tool(result, tool_call_id=...) → assistant(final)
    ask("And tomorrow?");
    # Turn 2 sees turn 1; history grows in place.
}
```

The list may contain Message instances or plain `dicts` (e.g. loaded from disk in a previous session) - `byllm` reconstructs via `Message.from_dict` on the way in and writes back in whatever shape it received.


jac-byllm/byllm/llm.impl/basellm.impl.jac
```diff
@@ -83,6 +83,16 @@ impl BaseLLM.invoke(mt_run: MTRuntime) -> object {
...
+    } finally {
+        mt_run.write_back_conversation();
+    }
+}
+
+
+impl BaseLLM._invoke_react_loop(mt_run: MTRuntime) -> object {
     max_iters = (
         mt_run.call_params.get("max_react_iterations")
         or self.call_params.get("max_react_iterations")
```
Jayanaka-98, line 119: - `write_back_conversation()` fires before the stream is

When stream=True with tools, `_invoke_react_loop` returns a generator from `_stream_final_answer`. Python's `finally` fires immediately when `invoke()` returns that object, before any tokens are yielded. The final assistant answer hasn't been appended yet (that happens at line 578-582), so history is permanently missing it.

jac-byllm/byllm/impl/mtir.impl.jac
```diff
@@ -143,26 +143,45 @@ impl MTRuntime.factory(mtir: MTIR) -> MTRuntime {
...
+    # Conversational mode: `by llm(conversation=<list>)` splices the
+    # caller's prior conversation between SYSTEM and USER. Entries can be
+    # Message instances or dicts (from a previous turn's persisted form).
+    conv_arg = call_params.get("conversation");
+    conv_msgs: list[MessageType] = [];
+    conv_was_dicts: bool = False;
+    conv_list_ref: object = None;
+    if isinstance(conv_arg, list) {
+        conv_list_ref = conv_arg;
+        conv_was_dicts = (not conv_arg) or isinstance(conv_arg[0], dict);
```
Jayanaka-98: - Design issue: empty list silently defaults to dict mode

A developer writing `glob history: list[Message] = []` gets dicts back. Their type annotation lies in `history[0].role` breaks at runtime with no warning.

The empty-list case should not guess a format. Either default to `list[Message]` when the list is empty, or add an explicit `conversation_format: "message" | "dict"` kwarg so intent is declared upfront.

It is the type annotation that can mislead a dev. If we can build a way to capture the type annotation and make the decision, it would be ideal, but complex. So picking a non confusing behavior such as using only dict would be okay.

Jayanaka-98 (changes requested):

Jayanaka-98 (approved):

PR #5745, opened 2026-04-29
fix(parser): docstring before a declaration inside a function body produces a warning instead of attaching

SandeepaHWP:
## What was wrong

Two related docstring issues were fixed:

**Case 1** - A docstring before a declaration inside a function body was treated as a plain string expression instead of attaching to the declaration. This triggered a `W0060` warning.

```jac
def test_func() -> None {
    '''this is MyResult'''      # W0060 warning - treated as expression
    obj MyResult {              # docstring lost
        has value: str = "";
    }
}
```

**Case 2** - A docstring before `with entry` inside an archetype body was silently dropped. The parser consumed the string but threw it away instead of attaching it to the `with entry` block.

```jac
obj MyResult {
    '''this docstring was dropped'''
    with entry {
        has value: str = "";
    }
}
```

## Why they happened

**Case 1** - Jac uses different parsers depending on context. Module level and archetype bodies use `parse_element_stmt`, which already had a string lookahead to detect docstrings. Function bodies use `parse_statement`, which had no such check, so every string was treated as a bare expression.

**Case 2** - Inside `parse_archetype_member`, every STRING token is consumed upfront into a `doc` variable. Every branch then attaches it to the parsed node - except the `with entry` branch, which hardcoded `doc=None` in the `ModuleCode` constructor, silently dropping it.

## The fixes

**Case 1** - Added a string-followed-by-declaration-keyword lookahead to `parse_statement`. When a string is followed by `def`, `can`, `obj`, `node`, `edge`, `walker`, `class`, `async`, `override`, `static`, or `@`, it routes to `parse_docstring_target` to attach correctly.

**Case 2** - In the `with entry` branch of `parse_archetype_member`, attach the already-consumed `doc` to the `ModuleCode` node the same way every other branch does.

## Why `with entry` is excluded from Case 1's fix

Inside a function body, `with` always starts a context manager statement (`with open(...) as f`), never a `with entry` block. So a string before `with` inside a function body is just a bare expression. `KW_WITH` is intentionally left out of the lookahead list.


SandeepaHWP:
## Design choices for the `jac.spec` change

There are two ways to express this fix in the grammar, and they reflect different priorities.

---

**Option A - Reuse `docstring_target`**

The spec change is a single line:

```
statement ::=
    ...
    | docstring_target
```

This reuses the existing `docstring_target` rule rather than introducing something new.
The rule itself doesn't change, we're just making it reachable from a statement context.

The problem is that `docstring_target` in the spec only covers a specific subset of targets:

```
docstring_target ::=
    STRING ("test" | "enum" | type_alias | global_var | "impl" | module_code | cl/sv/na)?
```

`archetype` and `ability` are not listed there. They are handled in the implementation
via `_dispatch_decorated()` inside `parse_docstring_target()`, but the spec never says so.

That gap already existed before this PR. Adding `| docstring_target` to `statement` inherits
that incomplete picture: a reader of the spec would have no idea that docstrings before
`obj`, `node`, `walker`, `def`, `can` etc. are also valid in a statement context.

---

**Option B - Introduce `decl_docstring`**

A new non-terminal is added that explicitly lists everything valid in this position:

```
statement ::=
    ...
    | decl_docstring

decl_docstring ::=
    STRING (
        "impl" impl_def
        | ("obj" | "node" | "edge" | "walker") archetype
        | "class" archetype
        | ("def" | "can" | "async" | "static" | "override" | "class") ability
        | ("@" atomic_chain)* (impl_def | archetype | ability)
    )
```

This makes the spec accurately reflect what the parser actually does.
It also surfaces the distinction between statement context and module/archetype context:
`with entry` is intentionally absent here because inside a function body `with` starts
a context manager, not a `with entry` block.

---

PR #5753, opened 2026-04-29
 fix(jac-scale): `pub` endpoints no longer return 401 on invalid/expired bearer tokens

### LINKED ISSUE #5752: [Bug-jac-scale] Invalid bearer token returns 401 on public endpoints
## Summary

When a request to a `pub` endpoint includes an `Authorization: Bearer <token>` header with an invalid or expired token, jac-scale returns 401. Public endpoints should ignore invalid tokens, not reject them.

This breaks browsers that have a stale `jac_token` left in localStorage from a previous session — public endpoints that work in a fresh tab fail in the stale tab until the user manually clears storage.

## Reproduction

`main.jac`:

```jac
def:pub greet(name: str) -> dict {
    return {"message": f"Hello, {name}!"};
}
```

Start with jac-scale:

```bash
jac start main.jac
```

Request the endpoint with an invalid bearer token:

```bash
curl -X POST http://localhost:8000/function/greet \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer not-a-valid-token" \
  -d '{"name":"world"}'
```

## Actual

```json
{"ok":false,"error":{"code":"UNAUTHORIZED","message":"Unauthorized"}}
```

HTTP 401.

## Expected

```json
{"ok":true,"data":{"result":{"message":"Hello, world!"}}}
```

HTTP 200, same as when no `Authorization` header is sent. The endpoint is `pub` — invalid tokens should be ignored.

Walker endpoints (`/walker/<name>`) marked `pub` exhibit the same behavior.

akindu-k:
## Problem

Public (`pub`) walker and function endpoints returned `HTTP 401` when a request included an invalid or expired `Authorization: Bearer` token, even though no authentication was required to access them.

This broke real-world browser applications where a stale token cached in `localStorage` would cause every call to a public API to fail — until the user manually cleared their browser storage.

**Affected endpoints:** `/walker/<name>` and `/function/<name>` marked `pub`.

## Root Cause

`jwt_validation_middleware` in `jac_scale/jserver/impl/jfast_api.impl.jac` ran a global early-return block:

```python
if token_attempted and not username:
    return JSONResponse(status_code=401, ...)
```

This fired for **every** request carrying a bearer token that failed validation — before the request ever reached an endpoint handler. The middleware had no visibility into whether the target endpoint required authentication, so public endpoints were caught by the same rejection path as protected ones.

## Fix

Removed the early-return block from `jwt_validation_middleware`. The middleware now only extracts and validates the token to populate `request.state.username` (`None` if invalid or absent), then passes the request along.

Access control enforcement remains entirely at the endpoint level, where it already worked correctly:

```python
# In create_walker_callback / create_function_callback
if requires_auth and not username:
    return TransportResponse.fail(code='UNAUTHORIZED', ...)
```

- **Protected endpoints** — still return `401` when no valid token is present.
- **Public endpoints** — now proceed as guest (`__guest__`) when an invalid token is sent, matching the behaviour when no token is sent at all.

## Changes

| File | Change |
|------|--------|
| `jac-scale/jac_scale/jserver/impl/jfast_api.impl.jac` | Removed `token_attempted` guard and early-return `401` block from `jwt_validation_middleware` |
| `jac-scale/jac_scale/tests/test_serve.jac` | Corrected JWT middleware test: bad tokens on `pub` endpoints now assert `200`, not `401`; added clarifying comment |
| `docs/docs/community/release_notes/unreleased/jac-scale/5752.bugfix.md` | Release note |

## Testing

The existing test `"jwt middleware: per-user isolation and token rejection"` covered this scenario but had the assertion backwards (encoding the buggy behaviour). It now asserts the correct expectation:

- Forged-secret token → `200` on `PublicInfo`
- Expired token → `200` on `PublicInfo`
- Malformed token → `200` on `PublicInfo`
- No token → `200` on `PublicInfo` (unchanged)
- Any bad token → `401` on `CreateTask` (protected, unchanged)

image

Fixes #5752.


jac-scale/jac_scale/tests/test_serve.jac
```diff
@@ -2115,9 +2115,11 @@ test "jwt middleware: per-user isolation and token rejection" {
         r = client.post("/walker/CreateTask", json={"title": "x"});
         assert r.status_code == 401 , f"no token on auth endpoint: expected 401, got {r.status_code}";
 
+        # Bad tokens on public (pub) endpoints must be ignored — the request
+        # should succeed as guest rather than returning 401.
         for (hdrs, label) in bad_token_cases {
             r = client.post("/walker/PublicInfo", json={}, headers=hdrs);
-            assert r.status_code == 401 , f"{label} on public endpoint: expected 401, got {r.status_code}";
+            assert r.status_code == 200 , f"{label} on public endpoint: expected 200 (guest), got {r.status_code}";
```
musab-mah-7: we should have a test for pub defs as well no?

musab-mah-7 (dismissed):

jac-scale/jac_scale/tests/test_serve.jac
akindu-k: added the assertion for public def as well

kugesan1105:
@akindu-k  think about guest user and personalization !!

kugesan1105 (approved):

marsninja (approved):

PR #5754, opened 2026-04-30
chore: extend jac-check format enforcement to test files

akindu-k:
## What changed

The `jac-check` CI workflow previously excluded `tests/` directories from the `jac format --check --lintfix` step. This PR removes that exclusion so test files are held to the same formatting standard as the rest of the codebase.

All existing test files have been reformatted in bulk to bring them into compliance before enforcement kicks in.

## Files changed

- **`.github/workflows/jac-check.yml`** — removed the `! -path "*/tests/*"` exclusion from the `find` command in the format check step.
- **65 `.jac` test files** across `jac/`, `jac-scale/`, `jac-client/`, `jac-byllm/`, `jac-mcp/`, and `docs/` — reformatted to pass the linter.

## What this means for other devs

From this PR onwards, **CI will fail if any `.jac` test file is not correctly formatted**. Before pushing, run:

```bash
jac format --lintfix <your_test_file.jac>
```

Or to format all test files at once:

```bash
find . -type f -name "*.jac" -path "*/tests/*" \
  ! -path "*/fixtures/*" \
  ! -name "*_err.jac" \
  ! -name "*_syntax_err.jac" \
  -print0 | xargs -0 --no-run-if-empty -I{} jac format --lintfix {}
```

If you have a local branch with unformatted test files, run the script above before rebasing onto `main` to avoid CI failures.

## Bug caught by the formatter

Extending format enforcement to test files immediately surfaced a real bug that had been silently passing in CI for weeks.

CI run where the failure was first observed: https://github.com/jaseci-labs/jaseci/actions/runs/25148029840/job/73711983420?pr=5754

In `jac-client/jac_client/tests/test_desktop_build.jac`, the `test_client_only_config_is_supported` test asserts that `client_only` is present in the dict returned by `get_default_config()`. It isn't — `get_default_config()` in `desktop_config.impl.jac` never included that key, even though `is_client_only()` and the test were both added in PR #5494.

**Why it wasn't caught on main:** the `test-client` CI job runs `pytest -x jac-client -n auto`. The `-x` flag stops the entire run at the first failure. That job also drives E2E/Playwright tests which depend on infrastructure (browser, running sidecar) that isn't available in the standard CI environment. When PR #5494 landed, those E2E tests were likely already failing first, causing pytest to exit before `test_desktop_build.jac` was ever reached. The broken implementation was never caught.

**How this PR exposed it:** formatting enforcement now applies to test files. The bulk format run touched `test_desktop_build.jac`, which caused the file to be re-processed by the Jac compiler during the `jac format --check` step. More importantly, once the file was part of the staged diff, the full `pytest -x jac-client` run on this PR's CI reached the test — and this time no earlier failure masked it.

**The underlying bug:** `get_default_config()` is missing `'client_only': False` in its returned dict. The test was written correctly; the implementation was incomplete. It needs to be fixed separately.

## Script used to bulk-format

```bash
find . -type f -name "*.jac" -path "*/tests/*" \
  ! -path "*/fixtures/*" \
  ! -name "*_err.jac" \
  ! -name "*_syntax_err.jac" \
  | while read -r f; do
    echo "Formatting: $f"
    jac format --lintfix "$f"
  done
```


akindu-k:
Bugs fixed in the following PRs.

1. https://github.com/jaseci-labs/jaseci/pull/5759 
2. https://github.com/jaseci-labs/jaseci/pull/5756

PR #5755, opened 2026-04-30
Feat: Prometheus StatefulSet with Persistent Storage

juzailmlwork:
Summary
Converts Prometheus from a Deployment (with emptyDir) to a StatefulSet backed by a 2Gi PersistentVolumeClaim
Moves service creation before the StatefulSet so the required serviceName reference is satisfied on first deploy
Explicitly deletes the prometheus-data-{name}-0 PVC in destroy() since Kubernetes does not cascade-delete StatefulSet PVCs automatically
Why
Prometheus uses an on-disk TSDB to store time-series data. With a Deployment + emptyDir, every pod restart (node eviction, OOM kill, rolling update) wiped all historical metrics — dashboards lost their history and alert evaluation windows reset to zero. Switching to a StatefulSet with a PVC means the data directory is mounted from a persistent volume and survives restarts.

The explicit PVC deletion on destroy is equally important: Kubernetes intentionally leaves StatefulSet PVCs behind as a safety net, so without the extra delete call the 2Gi volume would be orphaned and continue billing after the stack is torn down.


jac-scale/jac_scale/targets/kubernetes/monitoring.jac
```diff
@@ -184,52 +218,31 @@ obj MonitoringDeployer {
...
                         ]
                     }
-                }
+                },
+                'volumeClaimTemplates': [
+                    {
+                        'metadata': {'name': 'prometheus-data'},
+                        'spec': {
+                            'accessModes': ['ReadWriteOnce'],
+                            'resources': {'requests': {'storage': '2Gi'}}
```
udithishanka: storage size hardcoded to 2Gi. mongo has the same shape and surfaces mongodb_storage_size as a config option in kubernetes_config.jac (default 1Gi). add prometheus_storage_size the same way so users with longer retention windows aren't stuck on 2Gi forever.

jac-scale/jac_scale/targets/kubernetes/monitoring.jac
```diff
@@ -184,52 +218,31 @@ obj MonitoringDeployer {
...
-            'spec': {
-                'type': 'ClusterIP',
-                'selector': {'app': prometheus_name},
-                'ports': [{'protocol': 'TCP', 'port': 9090, 'targetPort': 9090}]
-            }
-        };
-        try {
-            core_v1.read_namespaced_service(
-                name=f"{prometheus_name}-service", namespace=namespace
+            apps_v1.patch_namespaced_stateful_set(
```
udithishanka, line 276: patching an existing StatefulSet rejects changes to immutable fields (storage size, serviceName, selector) with 422, not 404. so once 2Gi is set, changing it in a future deploy errors out uncaught. mongo handles PVC growth with resize_pvc_if_needed at kubernetes_target.jac:587. no analogous call here, so PVCs can never grow without manual kubectl. the same pattern would fix both.

jac-scale/jac_scale/targets/kubernetes/monitoring.jac
```diff
@@ -1222,11 +1235,22 @@ obj MonitoringDeployer {
...
             prometheus_name,
             namespace,
-            'Deployment'
+            'StatefulSet'
         );
+        # Explicitly delete the PVC created by the StatefulSet volumeClaimTemplate.
+        # Kubernetes does not cascade-delete PVCs when a StatefulSet is deleted.
+        try {
+            core_v1.delete_namespaced_persistent_volume_claim(
+                name=f"prometheus-data-{prometheus_name}-0", namespace=namespace
```
udithishanka: prometheus-data-{name}-0 only matches replicas=1 with one volumeClaimTemplate. if anyone ever bumps replicas, replicas 1..N orphan silently. cleanest fix is to set persistentVolumeClaimRetentionPolicy whenDeleted Delete on the StatefulSet spec (K8s 1.27+, both jaseci-cluster and ninja-cluster run newer) and drop the explicit delete entirely. alt is to list PVCs by label selector and loop.

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
```diff
@@ -1619,7 +1619,7 @@ obj KubernetesTarget(DeploymentTarget) {
             if self.k8s_config.monitoring_enabled {
                 prometheus_name = f"{app_name}-prometheus";
                 try {
-                    apps_v1.read_namespaced_deployment(
+                    apps_v1.read_namespaced_stateful_set(
```
udithishanka, line 1650: existing users upgrading from the Deployment version: this returns 404 so resources_exist stays false, the new StatefulSet gets created, but the old Deployment named prometheus_name is never deleted. both pods stay alive with matching app labels and the Service routes to both, doubling scrapes. needs an explicit delete_namespaced_deployment of prometheus_name early in deploy if one exists.

jac-scale/jac_scale/tests/test_deploy_k8s.jac
```diff
@@ -468,16 +483,23 @@ test "deploy all in one" {
...
-    prometheus_container = prometheus_deploy.spec.template.spec.containers[0];
+    assert prometheus_sts.metadata.name == prometheus_name;
+    assert prometheus_sts.spec.replicas == 1;
+    prometheus_container = prometheus_sts.spec.template.spec.containers[0];
     assert prometheus_container.image == "prom/prometheus:latest";
-    print(f"Prometheus Deployment '{prometheus_name}' exists");
+    assert prometheus_sts.spec.volume_claim_templates is not None;
+    assert len(prometheus_sts.spec.volume_claim_templates) == 1;
+    pvc_tpl = prometheus_sts.spec.volume_claim_templates[0];
+    assert pvc_tpl.metadata.name == "prometheus-data";
```
udithishanka, line 522: tests verify the PVC exists and gets cleaned up but never prove data actually survives a restart. the whole point of this feature is data persistence across pod death. add a step that scrapes some metrics, kills the prometheus pod, waits for the new pod, and verifies historical metrics are still there. otherwise the test passes even if the PVC mount ends up broken.

docs/docs/community/release_notes/unreleased/jac-scale/5755.feature.md
```diff
@@ -0,0 +1 @@
+- **Feat: Prometheus StatefulSet with Persistent Storage**: Converts Prometheus from a Deployment to a StatefulSet backed by an 2Gi PVC so TSDB data survives pod restarts, and explicitly deletes the PVC on destroy to prevent orphaned volumes.
```
udithishanka: typo: 'an 2Gi' should be 'a 2Gi'.

```suggestion
- **Feat: Prometheus StatefulSet with Persistent Storage**: Converts Prometheus from a Deployment to a StatefulSet backed by a 2Gi PVC so TSDB data survives pod restarts, and explicitly deletes the PVC on destroy to prevent orphaned volumes.
```

udithishanka (changes requested):
6 things to look at. main blockers are the missing migration from the old Deployment, the hardcoded storage size, and the brittle PVC cleanup.

jac-scale/jac_scale/targets/kubernetes/monitoring.jac
juzailmlwork: this is fixed similar to mongodb

jac-scale/jac_scale/targets/kubernetes/monitoring.jac
juzailmlwork: this is fixed similar to mongodb

jac-scale/jac_scale/targets/kubernetes/monitoring.jac
juzailmlwork: this is fixed

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
juzailmlwork: this is fixed

jac-scale/jac_scale/tests/test_deploy_k8s.jac
juzailmlwork: Implementing this will increase the time of test cases

juzailmlwork: This is supposed behavior of k8s which is auto managed by it

docs/docs/community/release_notes/unreleased/jac-scale/5755.feature.md
juzailmlwork: this is fixed and a better description is given

jac-scale/jac_scale/targets/kubernetes/monitoring.jac
```diff
@@ -184,57 +238,47 @@ obj MonitoringDeployer {
...
+                ]
             }
         };
         # Always patch so the config-hash annotation propagates and triggers a rollout
         # when the scrape config changes. Fall back to create on first deploy.
+        # Storage size and serviceName are immutable StatefulSet fields — never patch them.
         try {
-            apps_v1.patch_namespaced_deployment(
-                name=prometheus_name, namespace=namespace, body=prometheus_deployment
+            apps_v1.patch_namespaced_stateful_set(
```
udithishanka, line 276: the comment on L262 says storage size and serviceName are immutable StatefulSet fields, never patch them, but the very next call sends the full StatefulSet body (which includes volumeClaimTemplates with the new prometheus_storage_size) to patch_namespaced_stateful_set. when an operator bumps prometheus_storage_size on a redeploy, K8s rejects the patch with 422 because volumeClaimTemplates is immutable. the except only catches 404 and re-raises, so the deploy crashes before resize_pvc_if_needed at L276 ever runs.

cleanest fix is to follow mongo's pattern at kubernetes_target.jac:574-594 which doesn't patch the StatefulSet on subsequent deploys, only creates on 404, and relies on resize_pvc_if_needed to grow the PVC. the Prometheus path needs the patch only to propagate the config-hash annotation for a rollout, so the alternative is to build a narrower patch body for the existing case (metadata, spec.template, spec.replicas, spec.updateStrategy only) and keep the full body just for the 404 create branch.

udithishanka (changes requested):
1 issue: the StatefulSet patch will 422 on storage-size changes.

jac-scale/jac_scale/targets/kubernetes/monitoring.jac
juzailmlwork: this is fixed

udithishanka (approved):

marsninja:

PR #5757, opened 2026-04-30
fix(desktop): correct PyInstaller dependency bundling + frozen-sidecar data_path

### LINKED ISSUE #5686: jac-client desktop: `byllm not bundled` in frozen sidecar → client crashes with `TypeError: t.append is not a function`
# `byllm not bundled` in frozen sidecar → client crashes with `TypeError: t.append is not a function`

## Summary

Desktop builds on `main` (e.g. the `jac-client/examples/all-in-one` example) ship a PyInstaller-frozen sidecar that does **not** include `byllm`. The frozen sidecar's plugin registration logs `byllm not bundled` and continues, but the jac-client web runtime then crashes in the webview with `TypeError: t.append is not a function`, surfaced through `ErrorBoundary`.

Two coupled defects:

1. **Packaging regression** — the PyInstaller spec used by the desktop target on `main` is not collecting `byllm` into the frozen sidecar.
2. **Client defensive-coding gap** — when a registered plugin is absent, the jac-client web runtime calls `.append(...)` on an undefined/wrong-shape payload instead of no-op'ing. The crash takes down the whole webview via `ErrorBoundary`.

Fixing (1) eliminates the symptom for all supported builds; (2) should still be fixed as defense in depth so lightweight/OSS-only variants without `byllm` don't crash.

## Environment

- Repro project: `jac-client/jac_client/examples/all-in-one` ([jac.toml](jac-client/jac_client/examples/all-in-one/jac.toml))
- OS: Windows (WebView2 / Tauri)
- Branch: `main`
- Not reproducible on `wip/pyinstaller-jac-hook` (PR #5670, "PyInstaller hook for .jac auto-discovery") — that hook happens to collect `byllm` correctly, which masks defect (1). Defect (2) remains latent regardless.

## Logs

Sidecar (stderr):

```
[sidecar] Registered client plugin
[sidecar] byllm not bundled: No module named 'byllm'
  File "D:\all-in-one-new\binaries\jac-sidecar\_internal\jac_client\plugin\src\targets\desktop\sidecar\main.py", line 83, in _register_frozen_plugins
    mod = importlib.import_module(module_path)
  File "importlib\__init__.py", line 90, in import_module
  ...
ModuleNotFoundError: No module named 'byllm'
INFO - Admin portal endpoints registered at /admin
INFO - LLM telemetry: litellm not available, skipping litellm callback
INFO - LLM telemetry: byllm not available, skipping agent callback
INFO - LLM telemetry endpoints registered at /admin/llm/telemetry/{summary,traces,...}

  JAC DEV SERVER v0.13.3
  >  Local:   http://localhost:58276/
  🚀 Server ready
```

Client (webview, relayed to server via `POST /cl/__error__`):

```
ERROR - [Client] t.append is not a function
  Stack: TypeError: t.append is not a function
    at h  (http://tauri.localhost/client.Bg7Nmad7.js:98:53578)
    at DN (http://tauri.localhost/client.Bg7Nmad7.js:208:48881)
    at Uc (http://tauri.localhost/client.Bg7Nmad7.js:30:17258)
    at $v (http://tauri.localhost/client.Bg7Nmad7.js:32:44471)
    at hv (http://tauri.localhost/client.Bg7Nmad7.js:32:40095)
    at oE (http://tauri.localhost/client.Bg7Nmad7.js:32:40023)
    at Zl (http://tauri.localhost/client.Bg7Nmad7.js:32:39876)
    at fv (http://tauri.localhost/client.Bg7Nmad7.js:32:34903)
    at M  (http://tauri.localhost/client.Bg7Nmad7.js:17:1601)
    at MessagePort.re (http://tauri.localhost/client.Bg7Nmad7.js:17:1971)
  URL: http://tauri.localhost/
  Source: ErrorBoundary
```

A second variant surfaces via `window.onerror` with the same `h`/`t.append` frame, stack through `yC` → fiber commit.

## Root cause

[jac-client/jac_client/plugin/src/targets/desktop/sidecar/main.py:68-98](jac-client/jac_client/plugin/src/targets/desktop/sidecar/main.py#L68-L98) explicitly tries to register three plugins from frozen app:

```python
plugins = [
    ("jac_scale.plugin",        "JacCmd",    "scale"),
    ("jac_client.plugin.client","JacClient", "client"),
    ("byllm.plugin",            "JacRuntime","byllm"),
]
```

On `main`, the PyInstaller spec for the desktop sidecar does not collect `byllm` submodules/data into the frozen app. `importlib.import_module("byllm.plugin")` raises `ModuleNotFoundError`, `_register_frozen_plugins` logs and continues. The sidecar boots without byllm's `JacRuntime`.

The jac-client web runtime then initializes against a sidecar missing the byllm plugin shape. A message received over the `MessagePort` between webview and sidecar triggers a state update; during commit the client does `t.append(...)` on a payload that is `undefined` (or the wrong container because byllm's stub never registered). This is the `TypeError`.

Stack signature confirms:

- `MessagePort.re` — IPC receive handler between webview and sidecar.
- `M` — runtime dispatcher.
- `fv`/`Zl`/`oE`/`hv`/`$v`/`Uc`/`ud` — React/Preact fiber commit flow.
- `DN`/`yC`/`h` — user-component/runtime frame where `.append` is called.

`.append` in JS exists on `FormData`, `URLSearchParams`, `Headers`, and DOM `Element`. Any of these is plausible as the expected type; the actual value is `undefined`/plain-object from the missing plugin payload.

## Reproduction

1. Check out `main`.
2. Build and run `jac-client/jac_client/examples/all-in-one` as a Windows desktop (PyInstaller sidecar).
3. Launch the packaged `.exe`.
4. Sidecar stderr shows `[sidecar] byllm not bundled`. Webview shows `TypeError: t.append is not a function` in `ErrorBoundary`. Both also forwarded to `/cl/__error__`.

Negative control: on `wip/pyinstaller-jac-hook` (PR #5670) the same build has byllm bundled, sidecar logs `[sidecar] Registered byllm plugin`, client does not crash.

## Proposed fix

1. **Primary — packaging:** Ensure the desktop PyInstaller spec on `main` collects `byllm` (submodules + `.jac` data files). PR #5670 ("PyInstaller hook for .jac auto-discovery") achieves this as a side effect and is the lowest-risk fix; either merge that or add an explicit `collect_all('byllm')` / `collect_submodules('byllm')` + data-file collection to the desktop spec.
2. **Secondary — client defense in depth:** Guard the `t.append(...)` site in the jac-client web runtime so a missing plugin payload is a soft no-op rather than an `ErrorBoundary` crash. The Jac source that lowers into the `h(...)` frame at `client.Bg7Nmad7.js:98:53578` should check that its target is defined and of the expected shape (`FormData`/`URLSearchParams`/`Headers`/`Element`) before calling `.append`.

Defect (2) should be fixed even after (1) lands, so lightweight builds that legitimately ship without `byllm` don't crash the webview.

## Related

- PR #5670 — PyInstaller hook for `.jac` auto-discovery (unblocks this as a side effect).
- #5467 — multi-mode sidecar with frozen plugin registration (introduced the explicit-registration path that exposes this).
- #5493 / #5494 — earlier sidecar/desktop build fixes in the same area.

kugesan1105:
## The problem

If you `jac build --client desktop` and install the resulting `.deb` system-wide, the app opens, you see the static landing page, then it gets stuck on "Connecting to server…". The DevTools console shows a `TypeError` or `SyntaxError` because the webview never got a real API URL to call.

Two unrelated things go wrong in the sidecar's startup chain, and either one is enough to break the app:

**1. Some PyPI dependencies silently don't make it into the bundle.**

jac-client's PyInstaller spec converts a package's distribution name to its import name using `dep_name.replace('-', '_')`. That works for `jac-scale` → `jac_scale`, but breaks the moment a PyPI package's wheel name doesn't match its top-level Python module (a normal Python convention). When the spec asks PyInstaller to bundle a name that doesn't exist as a module, PyInstaller quietly finds nothing and skips it. The package never ships. The sidecar tries to `import` it at runtime, fails, logs `not bundled`, and any plugin that depended on it doesn't register.

Plus, `collect_submodules` (which the spec used) only finds `.py` files. Any `.jac` source files shipped inside a Python package were dropped on the floor regardless.

**2. jac-scale's identity store tries to write into a read-only system path.**

`JacScaleUserManager.postinit` hard-codes `base_path/.jac/data` for the SQLite store and calls `os.makedirs(...)`. For a system-installed `.deb`, `base_path` is `/usr/lib/<app>/jac/`, which the user has no write permission to. `os.makedirs` raises `PermissionError`. That happens inside the server class constructor, so the server never binds a port. Tauri waits for `JAC_SIDECAR_PORT=` on stdout, times out, and the webview ends up with no backend URL.

End result: app opens, page loads, but nothing actually works because the API died on boot.

## What I looked at

**For the bundling miss**, four options:

- Hard-code a translation table per offender. Scales poorly; every new package that breaks the heuristic needs a code change.
- Force users to declare module names in `jac.toml` themselves. Boilerplate, and the info is already on disk.
- Read `top_level.txt` from the installed wheel's metadata via `importlib.metadata`. This is the canonical PyPA lookup. Every wheel ships it.
- Rely on PR #5670's PyInstaller hook to follow real imports. Complementary; it helps PyInstaller see `.jac` files, but doesn't fix the distname-vs-modname confusion in jac-client's `[dependencies]` plumbing.

**For the read-only crash**, three options:

- Hardcode a user-dir path. Breaks dev/source-tree runs where the current behavior is correct.
- Honor `JAC_DATA_PATH` env var. The sidecar entry point already sets this to the platform-appropriate user data dir; UserManager was just ignoring it.
- Move the `makedirs` to lazy-init. Same crash, just deferred.

## What I picked, and why

**Bundling:** use `importlib.metadata.distribution(name).read_text('top_level.txt')`. It's the canonical answer, zero maintenance, and `top_level.txt` is universally shipped. The old dash-to-underscore heuristic is kept only as a fallback for cases where metadata is genuinely missing (rare). Also added `collect_data_files(name)` so `.jac` source files inside Python packages ship alongside `.py`.

**Data path:** make `JacScaleUserManager` read `JAC_DATA_PATH` from env before falling back to `self.base_path`. One line. The sidecar already sets this env var; UserManager just needs to look. Dev runs are unaffected (they don't set the var).

## How it works in code

`jac-client/.../desktop_target.impl.jac` (+29 / −7): added the `importlib.metadata` import, added `_resolve_top_level(dist_name)` helper, and switched the loop body to iterate resolved names + collect both submodules and data files.

`jac-scale/.../user_manager.impl.jac` (+6 / −1): one-line change in the SQLite branch of `postinit`:

```python
base_for_data = os.environ.get('JAC_DATA_PATH', self.base_path);
data_dir = os.path.join(base_for_data, '.jac', 'data');
```

## How I verified

- Built the `.deb` in CI, installed it on WSL2 Ubuntu via `sudo dpkg -i`.
- Launched the app: sidecar logs print `JAC_SIDECAR_PORT=<port>` followed by `🚀 Server ready`.
- Webview connects to the injected URL; clicked through tools in the in-app MCP playground; the `validate_jac` tool round-trips through the built-in jac-mcp server and renders results.
- Inspected the bundle: packages that were previously missing now show up under `_internal/<modname>/`, with their `.jac` files alongside the `.py` ones.

## Not fixed by this PR (separate follow-ups)

- `JacMetaImporter` loses `__file__` for some namespace-package descendants at frozen runtime. Manifests as `ImportError ... (unknown location)` for one plugin. Non-blocking (the registration loop catches it and the app boots), but worth a separate fix. Closer to PR #5670's domain.
- A downstream desktop app's launcher code assumes a system Python on PATH (returns None in frozen sidecars). Belongs in the downstream app's repo, not here.

## Related

Issue #5686 (`byllm not bundled` in frozen sidecar). The byllm-specific symptom there is a special case of the distname-vs-modname class fixed in this PR.


kugesan1105:
Follow up PR @musab-mah-7 
- Integration test for Desktop App building

PR #5759, opened 2026-04-30
fix(jac-client): correct assert string and docstring placement in `test_desktop_build.jac`

akindu-k:
## What changed

Fixed two issues in [jac-client/jac_client/tests/test_desktop_build.jac](jac-client/jac_client/tests/test_desktop_build.jac):

1. **Wrong assert string** in `client_only requires base_url` — the test was checking for a substring that never existed in the implementation.
2. **Misplaced docstrings** — docstrings were placed inside test bodies instead of immediately before the `test` keyword, causing the file to fail compilation. Because the file never compiled, none of its tests ever ran.

### Misplaced docstrings (compile failure)

Jac requires docstrings to be placed *before* the `test` block, not inside it. All five tests in the file had their docstrings as the first statement inside the body:

```jac
# before (invalid — docstring inside body)
test "client_only config is supported" {
    """Verify DesktopConfig supports client_only field."""
    ...
}
```

```jac
# after (valid — docstring before test keyword)
"""Verify DesktopConfig supports client_only field."""
test "client_only config is supported" {
    ...
}
```

Because the file failed to compile, CI silently skipped all tests in it.

### Wrong assert string (logical failure)

The `client_only requires base_url` test asserted:

```jac
assert "client_only = true requires" in content, ...
```

But the actual error message emitted by `desktop_target.impl.jac` (introduced in PR #5494) is:

```
"client_only mode requires [plugins.client.api] base_url in jac.toml. "
```

The substring `"client_only = true requires"` has never existed. The fix updates the assertion to match the real message:

```diff
- assert "client_only = true requires" in content,
+ assert "client_only mode requires" in content,
      "client_only mode should validate base_url is set";
```

## Why it wasn't caught on main

The `test-client` CI job ran `pytest -x jac-client -n auto`. The `-x` flag stops at the first failure, and with parallel workers (`-n auto`) distribution is non-deterministic. An earlier failure on another test consistently caused xdist to shut down before this file was reached, masking the compile error on every main CI run.

It was first surfaced by PR #5757, which removed `-x` from the `test-client` job so the full suite runs to completion.

## Files changed

- [jac-client/jac_client/tests/test_desktop_build.jac](jac-client/jac_client/tests/test_desktop_build.jac) — move docstrings before `test` blocks; fix assert substring
- [docs/docs/community/release_notes/unreleased/jac-client/5759.bugfix.md](docs/docs/community/release_notes/unreleased/jac-client/5759.bugfix.md) — release note


kugesan1105:
@musab-mah-7 do we really need this test file : jaseci/jac-client/jac_client/tests/test_desktop_build.jac
Shall we remove this ?

-----------------------------------


image

PR #5765, opened 2026-04-30
refactor(jac-native/rc): factor helpers, gate debug machinery, replace id()-based ownership (alt for #5736)

marsninja:
## Summary

Architectural cleanup on top of #5736 (mgtm98's RC-GC fixes). Same branch name (`jac-native_RC-GC_fixes`) but on marsninja's fork, with these refactors stacked on top. All five tests in `test_native_gc.jac` continue to pass at every step. Diff vs the base of #5736: +509 / -432 across 17 files.

### Changes

- **`_emit_set_type_tag(b, ptr, name)` helper** — replaces 8+ inline tag-stamp blocks across `builtins/objects/tuples/types/refcount/container_helpers`.
- **`_rc_call_with_ctx(fn, args, op, loc)`** — collapses ~12 `_set_rc_ctx` + call + `_clear_rc_ctx` triplets into one call.
- **`_emit_loop_iter_cleanup` now wired into all 5 `_codegen_for` variants** (range, dict-items, dict, string, list) so for-loops match while-loops.
- **`HDR_*`, `LIVE_NODE_*`, `HDR_MAGIC` named constants** in `na_ir_gen_pass.jac` replace the magic offsets (`-32/-24/-16/-8` and `+0/+8/+16/+24/+32`) scattered across `refcount.impl.jac` and `gc_debug.impl.jac`.
- **`_emit_magic_check(b, ptr)` helper** — de-duplicates the inline magic-check sequence in `__rc_retain` and `__rc_release_simple`.
- **Phase R2: `_rc_debug_codegen` flag** (read from `JAC_RC_DEBUG_CODEGEN` env var, default off) gates `_emit_live_node_insert`, `_emit_live_list_remove`, `_emit_rc_live_dump`, and the live-list walk in `__rc_set_type`. Production binaries no longer pay `malloc(LIVE_NODE_SIZE)` per Jac allocation — observable as nogc RSS dropping ~50% in `retain_release` / `chess` fixtures.
- **`rc_owned_ids: set[int]` replaced with `is_owned` attribute on `ir.Value`**, propagated through `_coerce_type` via a thin wrapper around `_coerce_type_inner`. Eliminates the `_pre_coerce_owned` workarounds and the fragility of Python `id()` reuse.

### Observed RSS impact (Phase R2)

| Test | nogc RSS before | nogc RSS after | Recovered |
|---|---|---|---|
| perf_alloc_storm | 11264 KB | 7168 KB | 4 MB |
| perf_retain_release | 109568 KB | 63488 KB | 46 MB |
| perf_deep_call_chain | 110592 KB | 63488 KB | 47 MB |
| chess | 498688 KB | 346112 KB | 152 MB |

That overhead was being paid unconditionally on every binary; now only emitted when `JAC_RC_DEBUG_CODEGEN=1` is set at compile time.

## Test plan

- [x] `jac test jac/tests/compiler/passes/native/test_native_gc.jac` — 5/5 pass after each individual change
- [x] Pre-commit `jac-format` passes


marsninja:
Re-targeting: pushing commit directly to mgtm98:jac-native_RC-GC_fixes so it lands in #5736.

mgtm98:
Nice change, It was on my list too :)

PR #5766, opened 2026-04-30
refactor(auth): address review feedback on user profile endpoint

kashmithnisakya:
## Description

Addresses review feedback on #5708. Bundles bug fixes, refactors, and one new endpoint, plus a hardening pass and a CI regression fix.

## Changes

**Bug fixes**
- Reserve `sso` as a server-managed profile key. A dev who registered with `profile = {"sso": ...}` would crash later SSO logins and lock the user out.
- `jacSignup` now surfaces JSON parse errors instead of silently returning `{success: True}` on a malformed 200 body.
- SSO callback defensively coerces `profile.sso` to `{}` when it isn't a dict, protecting legacy users from before the reserved-key check.

**Refactors**
- Profile validation moves to a Pydantic `AfterValidator` so invalid input returns 422 automatically; drops the hand-rolled try/except in `register_handler`.
- `/user/me`, `/user/password`, and the API key endpoints reuse `authenticate_or_fail`, removing duplicated bearer-token + 401 boilerplate.
- `GET /user/me` returns a typed `MeResponse` (with `exclude_none=True` preserving the original wire shape) for a stable OpenAPI schema.
- New public `JacScaleUserManager.update_profile(...)` so the PATCH handler doesn't reach into private storage.

**New endpoint**
- `PATCH /user/me` — partial profile update. Reuses `validate_profile` (reserved keys still rejected) and shallow-merges into the existing profile so `profile.sso.*` survives.


kugesan1105:
hey @kashmithnisakya  all good but its better to do these in this pr
1. if there are test for new public PATCH /user/me + changed GET response shape 
2. adding a `Feature` release fragment: for the  PATCH /user/me is a new endpoint

Issue #5768, opened 2026-04-30
type checker error summary of pyast_gen_pass


SandeepaHWP:
## Overview

Running `jac check jac/jaclang/jac0core/passes/pyast_gen_pass.jac` (post `jac purge` on updated main).

**Total errors: 76** (down from 92 — 16 fixed by `CheckNodeIsinstanceCallResult` update)

---

<img width="857" height="602" alt="Image" src="https://github.com/user-attachments/assets/ad5a5bd8-98e0-466b-b865-74a5e2276e2e" />

## False Positives (54 errors, 71%) — Type Checker Gaps

These are not code bugs. No fix needed.

| # | Category | Count | Codes |
|---|---|---|---|
| 1 | `isinstance()` narrowing failures — checker doesn't narrow `ast3.AST` to subtype after check, flags `.lineno`, `.col_offset`, `.end_lineno`, `.jac_link`, `.id`, `.elts`, `.ctx`, `.attr`, `.left`, `.ops` | 16 | E1030, E1031 |
| 2 | AST subtype attrs post-isinstance — same root, wider set: `.ctx`, `.func`, `.args`, `.slice`, `.value`, `.keys`, `.patterns`, `.rest`, `.comparators`, `.keyword` | 19 | E1030, E1031 |
| 3 | Union type safe access ignored — `.parent`, `.sym_name`, `.name` on unions where None guards already exist (~lines 214, 222, 262, 286, 582) | 5 | E1099 | #5794 #5795 #5797 #5798  
| 4 | Property setter not recognized — `@gen.setter` treated as function not property descriptor (`unitree.jac:58`) | 2 | E1031 |
| 5 | Unknown type in loops/comprehensions — checker reports `<Unknown>` inside complex comprehensions, flags `.arg`, `.value` downstream | 6 | E1032, E1053 |
| 6 | Enum `__members__` not recognised — `Tokens.__members__` is a valid Python enum runtime attribute | 2 | E1030 |
| 7 | Parameter narrowing after guards — type narrowed by surrounding guards but not propagated through call site | 4 | E1053, E1099 |

---

## Real Errors (22 errors, 29%) — Need Fixing

- [ ] **impl.jac:37** — `E1053`: `PyJsxProcessor(self)` — `self` typed as `PyastGenPass` but `pass_ref` param type doesn't accept it. Verify and align the param type in `PyJsxProcessor`.

[...]

PR #5769, opened 2026-04-30
fix(pyast_gen_pass): resolve 22 real type-checker errors from #5768

RagavRida:
## Summary

Fixes every "Real Errors" item triaged in #5768. Takes `jac check pyast_gen_pass.jac` from **76 → 53 errors** — the 53 leftovers are all in the false-positive categories the issue itself says "no fix needed" (post-`isinstance` narrowing on `ast3.AST` subtypes, `Tokens.__members__`, union-with-None already guarded, comprehension-element `Unknown`).

Closes the code side of #5768. The 6 remaining `list[Module | list[Module] | None]` → `list[Module | list[Module] | None]` mismatches (where parameter and argument types are **literally identical** in the error message) are a Pyright-port invariance evaluator bug; that fix belongs in `type_evaluator.impl/parameter_type_check.impl.jac` and is out of scope here.

## Audit verdict

I reproduced the 76 errors on current `main` (`b5297aa2e`) and went through the issue's "Real Errors" claims one by one. A few of them turned out to be checker false positives the issue mis-categorized — e.g. `:741` `ast3.TypeVar(...)` (this is the **AST-node class** added in Python 3.12, not `typing.TypeVar`, and its `name` kwarg correctly accepts a runtime `str`), and `:197` `binding` (already None-guarded with `continue` two lines above). Those still triggered checker errors though, so I added narrow workarounds — `getattr(ast3, 'TypeVar')` for the former, `cast(InteropBinding, binding)` for the latter — that silence the FP without changing runtime behavior.

The genuine code bugs in the list:
- **`assertIsNotInstance` branch (~:1958)** — copy-paste typo: `isinstance(nd.condition, uni.UnaryExpr)` was written twice and `isinstance(nd.condition.operand, uni.UnaryExpr)` came **after** the FuncCall check, narrowing operand to UnaryExpr right before passing it to a function that wants FuncCall.
- **`check_node_isinstance_call`** — `len(nd.params)` would crash at runtime when `params` is `None` (`Sequence[…] | None`).
- **`exit_func_signature`** — `[*nd.posonly_params, *nd.params]` would crash spreading `None` when `params` is unset.
- **`exit_node` / `resolve_stmt_block`** — `else nd.body if not isinstance(nd.body, uni.Expr)` let `ImplDef` instances fall through into a `Sequence[CodeBlockStmt]` parameter.
- **`gen_llm_body` / `exit_ability`** — `nd.name_ref` is `NameAtom | None` but passed unconditionally to `_py_name(AstSymbolNode)`.
- **`pyinline_sync` / `sync`** — `i.jac_link: list[ast3.AST] = [jac_node]` and `i.jac_link: ast3.AST = [self.cur_node]` had wrong annotations (`jac_link` holds `UniNode`, not `ast3.AST`).

Everything else (narrowing through subscripts, member-access, comprehensions, generic substitution) is the Pyright port not propagating narrowing — fixed at the call site with hoists or `cast()`.

## Files changed

- `jac/jaclang/jac0core/passes/impl/pyast_gen_pass.impl.jac` — all behavioral fixes
- `jac/jaclang/jac0core/passes/pyast_gen_pass.jac` — added `Any` to `from typing` imports for the `cast(Any, self)` at the `PyJsxProcessor(self)` call site (workaround for the TYPE_CHECKING duplicate-symbol artifact in `jsx_processor.jac`)

## Test plan

- [x] `jac check jac/jaclang/jac0core/passes/pyast_gen_pass.jac` — 76 → 53 errors
- [x] `jac test jac/tests/compiler/passes/main/test_pyast_gen_pass.jac` — 10/10 pass
- [x] `jac test jac/tests/compiler/passes/main/test_pyast_build_pass.jac` — 5/5 pass
- [ ] CI green (mypy, ruff, jac-format already pass locally)

## Follow-up

The 6 surviving list-invariance errors trace to the type evaluator rejecting `list[X]` → `list[X]` assignments (same type both sides). Happy to open a separate PR against `parameter_type_check.impl.jac` if maintainers want it bundled.

### DIFF
--- a/jac/jaclang/jac0core/passes/impl/pyast_gen_pass.impl.jac
+++ b/jac/jaclang/jac0core/passes/impl/pyast_gen_pass.impl.jac
@@ -34,7 +34,10 @@ impl PyastGenPass.postinit -> None {
             jac_node=self.ir_in
         )
     ];
-    self.jsx_processor = PyJsxProcessor(self);
+    # cast bypasses TYPE_CHECKING duplicate-symbol artifact: jsx_processor.jac
+    # imports PyastGenPass under TYPE_CHECKING to break the circular import,
+    # so the checker sees the parameter type and self's type as distinct symbols.
+    self.jsx_processor = PyJsxProcessor(cast(Any, self));
     # Call parent postinit to run the transformation
     super.postinit();
 }
@@ -192,9 +195,10 @@ impl PyastGenPass._generate_sv_to_sv_stubs(nd: uni.Import) -> bool {
         if binding is None {
             continue;
         }
+        bound_binding = cast(InteropBinding, binding);
         parts.append(
             self._gen_py_sv_func_stub(
-                local_name, item_name, module_name, binding, boundary_types
+                local_name, item_name, module_name, bound_binding, boundary_types
             )
         );
     }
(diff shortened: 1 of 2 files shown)


marsninja:

PR #5772, opened 2026-04-30
[feat] Add SSE streaming primitives to jac-client (jacSpawnStream, useJacStream)

Jayanaka-98:
## Summary

`jac-scale` already auto-wraps any walker / `def:pub` that returns a Python generator as `text/event-stream`, but `jac-client` had no consumer for it. The only client-side primitive was `jacSpawn`, which buffers the full JSON response. This PR adds the missing client-side streaming layer so apps can render tokens, progress events, and structured agent steps as the server yields them.

## What's added

- **`jacSpawnStream(walker, nodeId, fields, onEvent)`**: async, fires `onEvent` per SSE frame; returns when the server emits `event: end`, throws on `event: error`.
- **`jacCallFunctionStream(name, args, onEvent)`**: same shape for `def:pub` functions.
- **`useJacStream(walker, nodeId, fields)`**: React-style hook returning `{events, isStreaming, error, start(extra_fields), reset()}`. `events` is the raw array of yielded objects, so callers can map tokens or render structured steps directly.
- **Internal helpers** `__doWalkerFetchStream` / `__doFuncFetchStream` handle the `fetch` -> `ReadableStream` -> SSE-frame-parser pipeline.

The bundler is updated to inject the new symbols into the auto-generated `@jac/runtime` import so user `cl{}` blocks can call them without an explicit import.

## How streaming happens

The server side is identical for both modes: yield a Python generator from a walker (or `def:pub`) and `jac-scale` wraps it as `text/event-stream` automatically. The difference is purely in the **shape of the yielded dict**, and how the client renders it.

### Mode 1: simple token streaming

Useful for LLM token-by-token rendering. Server yields `{"token": "..."}` chunks; client joins them.

**Server** (`main.jac`):

```jac
def _generate_reply(prompt: str) -> Any {
    for word in ["Hello", " ", "world", "!"] {
        yield {"token": word};
    }
}

walker:pub chat {
    has prompt: str = "";
    can run with Root entry {
        report _generate_reply(self.prompt);
    }
}
```

**Client** (`cl{}` block):

```jac
cl {
    def:pub app -> JsxElement {
        stream = useJacStream("chat");
        text = [e.token for e in stream.events].join("");
        return
            <div>
                <button onClick={lambda { stream.start({"prompt": "hi"}); }}>Send</button>
                <p>{text}</p>
                <p>{("..." if stream.isStreaming else "")}</p>
            </div>;
    }
}
```

As tokens arrive, `stream.events` grows and React re-renders the joined text.

### Mode 2: structured event streaming

Useful for agent traces (tool calls, reasoning steps, progress). Server yields tagged event objects; client renders them based on `type`.

**Server** (`main.jac`):

```jac
def _agent_run(task: str) -> Any {
    yield {"type": "thinking",  "detail": "reading files"};
    yield {"type": "tool_call", "detail": "grep main.jac"};
    yield {"type": "tool_done", "detail": "grep main.jac", "duration_ms": 42};
    yield {"type": "done",      "detail": "task complete"};
}

walker:pub agent {
    has task: str = "";
    can run with Root entry {
        report _agent_run(self.task);
    }
}
```

**Client**:

```jac
cl {
    def:pub app -> JsxElement {
        stream = useJacStream("agent");
        return
            <div>
                <button onClick={lambda { stream.start({"task": "summarize"}); }}>
                    Run
                </button>
                <ul>
                    {[
                        <li key={i} className={f"event-{e.type}"}>
                            <strong>{e.type}</strong>: {e.detail}
                        </li>
                        for (i, e) in enumerate(stream.events)
                    ]}
                </ul>
                {(<p>error: {stream.error}</p> if stream.error else None)}
            </div>;
    }
}
```

The same hook handles both modes. `stream.events` is the raw array of yielded objects, so the rendering layer decides whether to join, group, or filter them. No server-side protocol changes are needed to switch between modes.

### Lower-level callback API

For non-React contexts, `jacSpawnStream` exposes the per-frame callback directly:

```jac
async def fetch_logs() -> None {
    await jacSpawnStream(
        "log_tail",
        nodeId="",
        fields={"lines": 100},
        onEvent=lambda(evt: dict) -> None { print(evt); }
    );
}
```

## Why

Streaming is the dominant UX for agentic apps (token streaming, tool-call traces, progress updates). The server side was already in place; this closes the loop on the client.

## Tests

`test_streaming.jac` adds 11 tests across three layers:

1. **Server-side SSE** (httpx, no browser): content-type, frame count, frame payloads, mid-stream errors, plain JSON walkers, `def:pub` functions
2. **Client bundle compilation** (`jac build`): compiled `client_runtime.js` contains `useJacStream`, `jacSpawnStream`, `jacCallFunctionStream`, `getReader`, and the SSE Accept header
3. **End-to-end browser** (Playwright + Chromium): `useJacStream` drives a real walker, streamed events render into the DOM; token-accumulation pattern joins streamed tokens correctly

Two event-shape tests also confirm both server conventions above work without code changes: per-token chunks (`{"token": "..."}`) and structured agent events (`{"type": "tool_call", "detail": "..."}`).


MusabMahmoodh (commented):
reviewed end to end against the sse wrapper in sse.jac and existing __doWalkerFetch baseline, overall its solid work, server contract matches properly, bundler injection is correct, tests covers all 3 layers nicely. main thing is the abort/cancellation gap, rest is polish.

jac-client/jac_client/plugin/impl/client_runtime.impl.jac
```diff
@@ -1012,6 +1012,272 @@ impl JacForm(props: dict) -> JsxElement {
...
+    _sState = reactUseState(False);
+    _errState = reactUseState(None);
+    curEvents = _eState[0];
+    setEvents = _eState[1];
+    curIsStreaming = _sState[0];
+    setIsStreaming = _sState[1];
+    curError = _errState[0];
+    setError = _errState[1];
+
+    def start(extra_fields: dict = {}) {
```
MusabMahmoodh, line 1237: no AbortController so two issues, double click on send will keep both streams alive and mix events into same array since onEvent captured setEvents from outer scope, also if component unmounts mid stream setState fires on unmounted comp, need abort on reset/new start plus isMounted ref

jac-client/jac_client/plugin/impl/client_runtime.impl.jac
```diff
@@ -1012,6 +1012,272 @@ impl JacForm(props: dict) -> JsxElement {
...
+        ).catch(
+            lambda err: Any {
+                setStFn.call(None, False);
+                msg = err.message if err and err.message else String(err);
+                setErrFn.call(None, msg);
+            }
+        );
+    }
+
+    def reset {
```
MusabMahmoodh, line 1263: reset only clears the react states but fetch loop is still running, next frame will repopulate events and flip isStreaming back, same fix as the abort issue above

jac-client/jac_client/plugin/impl/client_runtime.impl.jac
```diff
@@ -1012,6 +1012,272 @@ impl JacForm(props: dict) -> JsxElement {
...
+        return;
+    }
+    reader = response.body.getReader();
+    decoder = Reflect.construct(TextDecoder, []);
+    buffer = "";
+    _NL = String.fromCharCode(10);  # "\n" compiles to literal \n in Jac client code
+    _NLNL = String.fromCharCode(10) + String.fromCharCode(10);
+    while True {
+        chunk = await reader.read();
+        if chunk.done {
```
MusabMahmoodh, line 1062: if reader returns done=true without seeing event:end function returns normally so isStreaming goes false but error stays null, doc claims error is set on connection drop so either fix here or remove that from doc

jac-client/jac_client/plugin/impl/client_runtime.impl.jac
```diff
@@ -1012,6 +1012,272 @@ impl JacForm(props: dict) -> JsxElement {
...
+    base_url = __getApiBaseUrl();
+    headers = {"Content-Type": "application/json", "Accept": "text/event-stream"};
+    if token {
+        headers["Authorization"] = f"Bearer {token}";
+    }
+    response = await fetch(
+        f"{base_url}/function/{function_name}",
+        {"method": "POST", "headers": headers, "body": JSON.stringify(args)}
+    );
+    if not response.ok {
```
MusabMahmoodh, line 1128: no 401 handling here, walker stream version reloads on 401, intially thought its a bug but __doFuncFetch also doesnt have it so atleast consistent, worth aligning all 4 fetch impls later

jac-client/jac_client/plugin/impl/client_runtime.impl.jac
```diff
@@ -1012,6 +1012,272 @@ impl JacForm(props: dict) -> JsxElement {
...
+            return;
+        }
+        error_text = await response.text();
+        walker_label = `walker if not nodeId else f"{`walker}/{nodeId}";
+        raise Exception(f"Walker {walker_label} stream failed: {error_text}");
+    }
+    # If the server returned plain JSON (walker didn't yield), call onEvent once
+    # with the full payload so the caller gets something useful.
+    ct = response.headers.get("content-type") or "";
+    if not ct.includes("text/event-stream") {
```
MusabMahmoodh, line 1047: if walker didnt yield server returns json and onEvent fires once with full payload, nice fallback but doc says events is raw yielded objects, useJacStream user with non generator walker will get one event with whole report which is suprising, document it or restrict to sse only

jac-client/jac_client/plugin/impl/client_runtime.impl.jac
```diff
@@ -1012,6 +1012,272 @@ impl JacForm(props: dict) -> JsxElement {
...
+        data = payload["data"] if payload["data"] else {};
+        if onEvent {
+            onEvent(data);
+        }
+        return;
+    }
+    reader = response.body.getReader();
+    decoder = Reflect.construct(TextDecoder, []);
+    buffer = "";
+    _NL = String.fromCharCode(10);  # "\n" compiles to literal \n in Jac client code
```
MusabMahmoodh, line 1058: this workaround is fine and the comment explains it well, but worth a followup ticket to fix the jac to js escape issue so future sse like code doesnt need this dance

jac-client/jac_client/plugin/impl/client_runtime.impl.jac
```diff
@@ -1012,6 +1012,272 @@ impl JacForm(props: dict) -> JsxElement {
...
+                continue;
+            }
+            if evt and onEvent {
+                onEvent(evt);
+            }
+        }
+    }
+}
+
+impl __doFuncFetchStream(function_name: str, args: dict, onEvent: Any) -> None {
```
MusabMahmoodh, line 1117: like 70 lines of identical sse frame parsing duplicated from __doWalkerFetchStream, mirrors the existing walker/func split so its consistent, but a _consumeSSEFrames(response, label, onEvent) helper would cut it in half

jac-client/jac_client/plugin/impl/client_runtime.impl.jac
```diff
@@ -1012,6 +1012,272 @@ impl JacForm(props: dict) -> JsxElement {
...
+                continue;
+            }
+            if evt and onEvent {
+                onEvent(evt);
+            }
+        }
+    }
+}
+
+impl jacSpawnStream(
```
MusabMahmoodh, line 1208: bypassing __cachedEndpointCall is correct since you dont want to cache sse, but a streaming walker registered as writer wont trigger reader invalidaton, edge case but a one liner comment saying its intentional would help future maintainers

docs/docs/reference/plugins/jac-client.md
```diff
@@ -694,6 +694,99 @@ def:pub LiveData() -> JsxElement {
...
+}
+```
+
+`useJacStream(walker, nodeId="", fields={})` returns:
+
+| Field | Type | Description |
+|----|----|----|
+| `events` | list | Raw array of yielded dicts; component re-renders as it grows. |
+| `isStreaming` | bool | True between `start()` and the server's `event: end`. |
+| `error` | str \| None | Set on `event: error` or connection drop. |
```
MusabMahmoodh, line 766: doc says error is set on connection drop but the impl doesnt actually do that (see comment on the chunk.done branch), either remove this or handle it in impl

marsninja:

PR #5773, opened 2026-05-01
fix(jac-scale): silently drop unknown query parameters instead of crashing

### LINKED ISSUE #4661: Bug: Unknown query parameters crash function and walker endpoints

akindu-k:
## Problem

Endpoints decorated with `@restspec` would crash with HTTP 500 when the request URL contained query parameters not declared in the walker or function signature.

```
GET /items?category=books&v=1708012345
→ TypeError: get_items() got an unexpected keyword argument 'v'
```

This is a common real-world scenario — CDNs, proxies, and browser cache-busting mechanisms routinely append extra parameters that the application has no knowledge of and should not be required to handle.

## Root Cause

`jfast_api.impl.jac` contains an `extra_query_params_code` block that, when the endpoint callback accepts `**kwargs`, loops over all raw request query parameters and injects any not explicitly declared into `callback_args`. Walker and function callbacks always use `**kwargs` as a catch-all for FastAPI parameter injection, so this block was always active — silently forwarding unknown params (like `v`) into the underlying Jac walker/function that doesn't declare them, causing the crash.

Removing `extra_query_params_code` from `jfast_api.impl.jac` entirely would fix walkers and functions but would break other legitimate consumers of `add_endpoint` — specifically `render_page_callback` in `serve.static.impl.jac`, which has `def callback(page_name: str, **kwargs)` and genuinely needs all extra query parameters forwarded to the frontend page renderer.

## Fix

The filtering is applied in `serve.endpoints.impl.jac` inside `create_walker_callback` and `create_function_callback` only, leaving `jfast_api.impl.jac` and all other `add_endpoint` callers untouched.

At endpoint registration time, `inspect.signature()` is called on the actual walker/function to build an allowed-key set, which is captured in the callback closure (zero per-request overhead). Before passing `kwargs` to `spawn_walker` / `execute_function`, unknown keys are filtered out.

Three cases are handled:

| Situation | Behaviour |
|---|---|
| Normal walker/function (no `**kwargs`) | Unknown query params silently dropped |
| Walker/function that declares its own `**kwargs` | All query params forwarded — intentional pass-through preserved |
| Introspection fails (non-introspectable callable) | Permissive fallback — no regression |

## Changes

- `jac-scale/jac_scale/impl/serve.endpoints.impl.jac` — add registration-time signature introspection and request-time `kwargs` filter in `create_walker_callback` and `create_function_callback`
- `jac-scale/jac_scale/tests/fixtures/restspec_fixtures.jac` — add `get_items` function and `SearchWalker` walker fixtures for regression coverage
- `jac-scale/jac_scale/tests/test_restspec.jac` — add test *"unknown query params are silently dropped"* covering both function and walker endpoints
- `docs/docs/community/release_notes/unreleased/jac-scale/5773.bugfix.md` — release note fragment

## Test Plan

- [x] New test `unknown query params are silently dropped` passes (function + walker endpoints both return 200 with declared params intact)
- [x] All 18 existing `test_restspec.jac` tests continue to pass

Closes #4661


akindu-k:
image

akindu-k:
Will be addressed in #6615

Issue #5523, opened 2026-04-11
jac format: multi-statement lambda bodies mash statements together with no whitespace


marsninja:
## Bug

`jac format` on a lambda with multiple body statements concatenates the statements with no whitespace, producing broken output like `{ return undefined;  }interval = setInterval(...);return lambda { clearInterval(interval);}; }`.

This is a **pre-existing bug** in `exit_lambda_expr` — it predates the JSX/lambda formatter fixes in #5519. Reproducible on `main` as well as on the branch.

## Repro

```jac
def main -> None {
    useEffect(
        lambda -> any {
            if not autoRefresh { return undefined; }
            interval = setInterval(lambda { loadMetrics(); }, 3000);
            return lambda { clearInterval(interval); };
        },
        [autoRefresh]
    );
}
```

### Expected (something like)

```jac
def main -> None {
    useEffect(
        lambda -> any {
            if not autoRefresh {
                return undefined;
            }
            interval = setInterval(lambda { loadMetrics(); }, 3000);
            return lambda { clearInterval(interval); };
        },
        [autoRefresh]
    );
}
```

### Actual (what `jac format` produces today)

```jac
def main -> None {
    useEffect(
        lambda -> any { if not autoRefresh { return undefined;  }interval = setInterval(
            lambda { loadMetrics();}, 3000
        );return lambda { clearInterval(interval);}; },
        [autoRefresh]
    );
}
```

Notice `}interval` and `);return` and `;};` all collapsed together with no separator.

## Real-world hit

Surfaced while formatting `jac-scale/jac_scale/admin/ui/pages/admin/monitoring/MetricsPage.cl.jac` for the downstream reformat pass in #5519. Line 42 reads `);return lambda { clearInterval(interval);}; },` after format, which is obviously wrong.

## Root cause

`exit_lambda_expr` in `jac/jaclang/compiler/passes/tool/impl/doc_ir_gen_pass.impl.jac` iterates `nd.kid` and handles tokens, single-expression bodies, and the function signature explicitly, but when the body is a `Sequence[CodeBlockStmt]` (multi-statement lambda body), each statement kid falls through to:

```jac
} else {
    parts.append(i.gen.doc_ir);
}
```

No `hard_line()` separators are inserted between body statements and no indent wraps the body, so the statements run together. Every other statement-body node in the formatter (`exit_if_stmt`, `exit_ability`, function bodies, etc.) uses the pattern:

```jac
[...]

PR #5780, opened 2026-05-01
fix(jac-format): fix whitespace bugs in lambda and if inline blocks

### LINKED ISSUE #5779: jac format: remaining whitespace bugs in lambda/if inline blocks (follow-up to #5523)
## Context

This is a follow-up to #5523. The original mashing bug (statements concatenated with no whitespace like \`}interval\` and \`);return\`) **has been fixed** — multi-statement lambda bodies now correctly stay as separate lines. However, running \`jac format\` on the repro case still produces incorrect whitespace inside inline blocks.

## Repro

Input (already clean, hand-written):

```jac
def main -> None {
    useEffect(
        lambda -> any {
            if not autoRefresh { return undefined; }
            interval = setInterval(lambda { loadMetrics(); }, 3000);
            return lambda { clearInterval(interval); };
        },
        [autoRefresh]
    );
}
```

Run:

```
jac format test.jac
```

## Actual output

```jac
def main -> None {
    useEffect(
        lambda -> any {
            if not autoRefresh { return undefined;  }
            interval = setInterval(lambda { loadMetrics();}, 3000);
            return lambda { clearInterval(interval);};
        },
        [autoRefresh]
    );
}
```

## Expected output

```jac
def main -> None {
    useEffect(
        lambda -> any {
            if not autoRefresh {
                return undefined;
            }
            interval = setInterval(lambda { loadMetrics(); }, 3000);
            return lambda { clearInterval(interval); };
        },
        [autoRefresh]
    );
}
```

### 1. `if` block inside multi-statement lambda body not expanded to multi-line

| Input | After format |
|---|---|
| `if not autoRefresh { return undefined; }` | `if not autoRefresh { return undefined;  }` |

An `if` block that lives inside a multi-statement lambda body should be expanded to multi-line form (as shown in the expected output above). Instead it stays inline and an extra spurious space is inserted before `}`.

### 2. Trailing space before `}` stripped from inline lambda bodies

| Input | After format |
|---|---|
| `lambda { loadMetrics(); }` | `lambda { loadMetrics();}` |
| `return lambda { clearInterval(interval); };` | `return lambda { clearInterval(interval);}` |

The space between the last `;` and the closing `}` of an inline single-statement lambda body is removed.

### 3. Formatter is not idempotent

Running `jac format` on already-formatted code changes it. A correct formatter should produce no diff when run on its own output.

## What was fixed in #5523

- Multi-statement lambda body statements are no longer mashed onto one line
- `}interval`, `);return`, `;};` collapsed output is gone
- Statements inside a multi-statement lambda body are kept as separate lines

## Related

- #5523 — original mashing bug (fixed)

SandeepaHWP:
## Summary

Fixes three whitespace formatting bugs in `exit_lambda_expr` and `exit_if_stmt` inside `doc_ir_gen_pass.impl.jac`. All three caused `jac format` to produce incorrect output and to not be idempotent (running it on already-formatted code would still change it).

Closes #5779

### 1. `if` block inside a multi-statement lambda body stayed inline

An `if` block directly inside a lambda with multiple body statements was incorrectly kept on one line instead of being expanded. The `flat_capable` check in `exit_if_stmt` only checked that the parent was a `LambdaExpr`, but not whether that lambda had a single or multiple body statements.

**Before:**
```jac
lambda -> any {
    if not autoRefresh { return undefined; }
    interval = setInterval(lambda { loadMetrics(); }, 3000);
    return lambda { clearInterval(interval); };
}
```

**After:**
```jac
lambda -> any {
    if not autoRefresh {
        return undefined;
    }
    interval = setInterval(lambda { loadMetrics(); }, 3000);
    return lambda { clearInterval(interval); };
}
```

### 2. Double space before `}` in flat-capable `if` blocks

The body statement (e.g. `return undefined;`) already ends with a trailing space from `exit_return_stmt`. The `self.line()` before `}` added another space in flat mode, producing `{ return undefined;  }`. Fixed by using `if_break(hard_line, Text(""))` so flat mode emits nothing and the body's existing trailing space provides the one separator.

### 3. Missing space before `}` in inline single-statement lambda bodies

`lambda { loadMetrics(); }` was formatted as `lambda { loadMetrics();}`. The RBRACE path in `exit_lambda_expr` for single-statement bodies had no preceding space. Fixed by adding `self.space()` before the closing brace.

## Test

Input (clean, hand-written):
```jac
def main -> None {
    useEffect(
        lambda -> any {
            if not autoRefresh { return undefined; }
            interval = setInterval(lambda { loadMetrics(); }, 3000);
            return lambda { clearInterval(interval); };
        },
        [autoRefresh]
    );
}
```

After fix, `jac format` produces:
```jac
def main -> None {
    useEffect(
        lambda -> any {
            if not autoRefresh {
                return undefined;
            }
            interval = setInterval(lambda { loadMetrics(); }, 3000);
            return lambda { clearInterval(interval); };
        },
        [autoRefresh]
    );
}
```

Running `jac format` a second time produces no changes (exit 0) - the formatter is now idempotent for this case.

## Related

- #5779 - issue tracking these bugs
- #5523 - original mashing bug (already fixed before this PR)


SandeepaHWP:
@RandiJay96  @kugesan1105  can someone add `skip-release-note-check` label to this PR

jac-byllm/byllm/llm.impl/model.impl.jac
```diff
@@ -87,9 +87,7 @@ impl Model.invoke(mt_run: MTRuntime) -> object {
                     _current_invocation_id.reset(ctx_token);
                 }
             }
-            return _wrap_generator(
-                result, mt_run, self.model_name, _start, _ctx_token
-            );
+            return _wrap_generator(result, mt_run, self.model_name, _start, _ctx_token);
```
kugesan1105, line 90: why is this formatted in a single line; previously it was properly formatted

jac-client/jac_client/examples/all-in-one/components/TransactionForm.jac
```diff
@@ -61,7 +61,7 @@ cl {
                             className={("toggle-btn active")
                                 if txType == "expense"
                                 else ("toggle-btn")}
-                            onClick={lambda -> any { txType = "expense";}}
+                            onClick={lambda -> any { txType = "expense"; }}
```
kugesan1105, line 64: @marsninja is this the expected formatted code? to have a white space before the `}`

jac-byllm/byllm/llm.impl/model.impl.jac
SandeepaHWP: This line is exactly 88 characters - right at the formatter's `MAX_LINE_LENGTH = 88` limit. 

The old formatter had a bug where every return statement was leaking a phantom trailing space into the width calculation. This line is exactly 88 chars, but the bug made it look like 89, so it got broken across lines unnecessarily. 

This fixes that bug aslo, so the line now correctly stays on one line.

PR #5783, opened 2026-05-01
fix(ci): replace `MicroK8s` with `kind` for k8s tests

### LINKED ISSUE #5781: ci: test-scale-k8s broken due to flaky snap store connectivity on GitHub Actions runners
## Bug

`test-scale-k8s` fails intermittently on PRs with:

```
sudo: microk8s: command not found
```

## What is actually happening

The real error was always hidden by `set +e` in the install step:

```yaml
set +e
sudo snap install microk8s --classic  # fails silently
sleep 20
set -e
```

The actual failure is:

```
error: unable to contact snap store
```

`set +e` silences the snap install failure so execution continues. MicroK8s was never installed, so the next step `sudo microk8s status --wait-ready` fails with `command not found`. The k8s tests were **never reliably running** - the cluster was never being created.

## Root cause

GitHub Actions runners have intermittent and unreliable connectivity to the snap store (`snapcraft.io`). `snap` is not designed for ephemeral CI environments. This is a fundamental mismatch - MicroK8s is built for long-running production/edge deployments (IoT, on-prem, Raspberry Pi), not for short-lived CI runners that spin up and tear down per job.

SandeepaHWP:
## Summary

Replaces MicroK8s with `kind` (Kubernetes IN Docker) in the `test-scale-k8s` CI job.

## Why

MicroK8s uses `snap` which has flaky connectivity to the snap store on GitHub Actions runners - `sudo snap install microk8s` fails intermittently with `unable to contact snap store`, causing `test-scale-k8s` to fail on PRs that have nothing wrong. The original `set +e` was hiding this failure silently, making the tests never actually run reliably.

`kind` is Docker-based. Docker is always available on GitHub Actions runners. No snap, no snap store, no network flakiness. It is also the tool Kubernetes itself uses for its own CI.

## Changes

**Only the CI workflow changes - zero changes to jac-scale source code or tests.**

The jac-scale codebase was already cluster-agnostic:
- All k8s interaction uses standard Python kubernetes client + `kubectl`
- Kubeconfig loaded via standard `load_kube_config()` which reads `~/.kube/config`
- No MicroK8s-specific CLI calls anywhere in the source

Replaced 3 steps (Install MicroK8s + Initialize MicroK8s + Configure kubectl) with 2 steps:

- **Install kind and kubectl** - curl download, no snap needed
- **Create kind cluster** - kind automatically writes `~/.kube/config`, provides a default storage class for PVCs, and the cluster config maps NodePort 30080 to the host for ingress tests

## Related

- Closes #5781

### DIFF
--- a/.github/workflows/test-jaseci.yml
+++ b/.github/workflows/test-jaseci.yml
@@ -274,18 +274,9 @@ jobs:
           pip install pytest-xdist
           pip install pytest-asyncio
 
-      - name: Install kind and kubectl
+      - name: Write kind config
         run: |
-          curl -Lo ./kind https://kind.sigs.k8s.io/dl/v0.23.0/kind-linux-amd64
-          chmod +x ./kind
-          sudo mv ./kind /usr/local/bin/kind
-          curl -LO "https://dl.k8s.io/release/$(curl -Ls https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
-          chmod +x kubectl
-          sudo mv kubectl /usr/local/bin/kubectl
-
-      - name: Create kind cluster
-        run: |
-          cat <<EOF | kind create cluster --name jaseci-test --config=-
+          cat <<EOF > /tmp/kind-config.yaml
           kind: Cluster
           apiVersion: kind.x-k8s.io/v1alpha4
           nodes:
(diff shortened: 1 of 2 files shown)


kugesan1105 (approved):
Nice fix! 🙌 The scale-k8s tests have been **flaky** for a while because MicroK8s relies on the snap store, and GitHub runners can't always reach it — a known pain point with snap in CI.
Moving to **Kind** makes a lot of sense since it's the standard for Kubernetes testing in CI: no snap dependency, faster, and way more reliable

PR #5794, opened 2026-05-02
fix(type-checker): support constant integer subscripts as narrowing keys

SandeepaHWP:
## Summary

Previously, the type checker couldn't narrow types through subscript
expressions. Writing `if items[0].start is not None` didn't actually
narrow `items[0].start` inside the block — the checker silently dropped
the guard and kept reporting false-positive E1099 errors.

This PR fixes three related gaps:
- `items[0]`, `items[-1]`, `config["host"]` — constant subscripts now produce stable narrowing keys
- `items[-1].value` — the narrowed type is now actually applied when the base is a subscript expression, not just a plain variable name
- `pyast_gen_pass` — `_source_context_token()` was being called twice in a condition; now cached into a local variable

## Root cause

The narrowing system works by matching a string key from the guard (e.g. `"items[0].start"`) against the same key at the usage site. Both `get_expr_key` (CFG pass) and `create_key_for_reference` (type evaluator) simply returned `None` for any subscript expression, so the key was never built and narrowing was silently skipped. On top of that, even when the key could be built, the type evaluator only applied narrowing when the base was a plain name like `x` — subscript bases like `items[-1]` were ignored.

## Fix

Extracted a shared `int_subscript_suffix` helper that both functions use to build consistent keys. Added an `elif` branch in the AtomTrailer evaluator to apply narrowing for non-Name bases. Removed an unnecessary `is_attr` guard that was blocking subscript expressions from being recognized as narrowing targets.

## Errors fixed

Error 1:

image

Error 2:

image


kugesan1105 (commented):
Nice fix — the gating logic is right (range/multi-slice/non-`Int` correctly excluded, recursion handles `m[0][0]`), and I confirmed variable-index `items[i]` correctly does **not** narrow. Couple of things before merge though:

**1. Duplication is the real concern.** The same 16-line block now lives in both `symbol_utils.get_expr_key` and `TypeEvaluator.create_key_for_reference`. The latter doesn't even use `self` — it's purely structural. Every future extension (string keys, `items[-1]`, named constants) will need synchronized edits in two places, and they will drift. Could we collapse them — e.g. have `create_key_for_reference` delegate to `get_expr_key` and only add the walrus branch on top? Worth doing now while the diff is small.

**2. Coverage feels narrow.** Only positive `Int` literals. `items[-1]` (parses as `UnaryExpr`), `config["host"]` (TypedDict-style), and `items[FIRST]` all silently miss. Not blocking, but the next bug report is going to be one of these.

**3. Tests are happy-path only.** Could you add negatives — variable index doesn't narrow, range slice doesn't narrow? Cheap insurance against regressions.

kugesan1105:
```jac
"""Narrow proof: subscript narrowing leaks past mutation."""

obj Item { has value: str = ""; }
obj Other { has code: int = 0; }

# Baseline: union access without narrowing should error - .value not on Other
def baseline(items: list[Item | Other]) -> str {
    return items[0].value;  # expected: ERROR (Item|Other has no .value)
}

# Soundness probe: assign Other to items[0] AFTER narrowing.
# Narrowing should be invalidated; .value should error like baseline.
def stale_narrow(items: list[Item | Other]) -> str {
    if isinstance(items[0], Item) {
        items[0] = Other();
        x = items[0];
        return x.value;  # expected: ERROR if narrowing properly invalidated
    }
    return "";
}

# Same soundness probe with dict.
def stale_dict(config: dict[str, Item | Other]) -> str {
    if isinstance(config["host"], Item) {
        config["host"] = Other();
        return config["host"].value;  # expected: ERROR
    }
    return "";
}
```

PR #5795, opened 2026-05-02
fix(type-checker): apply isinstance narrowing to non-Name AtomTrailer base targets

SandeepaHWP:
## Summary

- `isinstance(c.pet, Cat)` did not narrow `c.pet` inside the if body — `c.pet.meow` still reported an error
- The `case uni.AtomTrailer()` handler only called `get_narrowed_type` when `expr.target` was a simple `Name`
- For complex targets like `c.pet` (another `AtomTrailer`), `get_narrowed_type_for_expr` was never called, so the narrowed type was silently discarded

## Root cause

In `_get_type_of_expression_core`, the AtomTrailer handler had only one narrowing branch for `Name` targets. Non-Name targets (attribute chains, subscripts) went through `get_type_of_expression` to get their base type but received no narrowing, even when CFG flow guaranteed a narrowed type was available.

## Fix

Added an `elif` branch after the Name narrowing check that calls `get_narrowed_type_for_expr` for any non-Name target and applies the returned narrowed type to `base_type`.

Error in `pyast_gen_pass.impl.jac`

image


SandeepaHWP:
This is fixed in #5794

PR #5797, opened 2026-05-02
fix(type-checker): handle union target type in cast() special-form

SandeepaHWP:
## Summary

- `cast()` with a union target type like `cast((FuncDef | AsyncFuncDef), nd)` was returning class types instead of instances for the union members.
- `is_class_instance()` returned `False` for these members, so any attribute access on the cast result was incorrectly rejected with E1099.
- Fix: when the cast target evaluates to a `UnionType`, apply `clone_as_instance()` to each instantiable `ClassType` member before returning, so the result is a proper instance union.

## Files changed

- `jac/jaclang/compiler/type_system/type_evaluator.impl/parameter_type_check.impl.jac` - add `UnionType` branch in cast special-form handler
- `jac/tests/compiler/passes/main/fixtures/checker/checker_bug_cast_union_instance.jac` - minimum reproduction fixture
- `jac/tests/compiler/passes/main/test_checker_bugs.jac` - regression test `bug_5768_cast_union_instance`

Error is `pyast_gen_pass.impl.jac`

image


jac/tests/compiler/passes/main/test_checker_bugs.jac
```diff
@@ -608,3 +608,12 @@ test "bug_5768_while_loop_narrowing" {
         f"Got errors: {[str(e) for e in program.errors_had]}"
     );
 }
+
+test "bug_5768_cast_union_instance" {
```
kugesan1105: rename the test

kugesan1105 (changes requested):
Replace the bespoke `ClassType`/`UnionType`/`fallthrough `cascade with a single call to `self._convert_to_instance(cast_target_type)`. Same fix, fewer lines, future-proof against nested types.

jac/tests/compiler/passes/main/test_checker_bugs.jac
SandeepaHWP: done

SandeepaHWP:
> Replace the bespoke `ClassType`/`UnionType`/`fallthrough `cascade with a single call to `self._convert_to_instance(cast_target_type)`. Same fix, fewer lines, future-proof against nested types.

Done

kugesan1105 (approved):

PR #5805, opened 2026-05-03
fix(format): lambda/if inline-block whitespace bugs (#5779)

### LINKED ISSUE #5779: jac format: remaining whitespace bugs in lambda/if inline blocks (follow-up to #5523)

RagavRida:
Closes #5779.

## Summary

Three follow-up regressions to the #5523 lambda-mashing fix:

1. **`if` block inside multi-stmt lambda body stays inline** — the `flat_capable` shortcut in `exit_if_stmt` fired whenever the parent was a `LambdaExpr`, even when the lambda had several body statements. It now requires the parent lambda's body to be exactly the single if-stmt; otherwise the if falls through to the hard-line layout.
2. **Stray double space before `}` in compact `if` form** — return stmts emit a trailing space; the if-stmt's flat path's `line()` (which renders as a space) compounded with it. Fixed by adding a `trim_trailing_whitespace` helper and stripping any trailing whitespace from the body doc before appending the closing line+`}`.
3. **Inline single-stmt lambda lost space before `}`** (`lambda { stmt; }` -> `lambda { stmt;}`) — the body-stmt branch in `exit_lambda_expr` didn't add a trailing space and the RBRACE branch suppressed one. Now: trim trailing whitespace from `parts`, then add exactly one space before `}` when the body is non-empty. Empty `lambda { }` is preserved.

After this, `jac format` is also idempotent on the corrected output.

## Side effect on prior test

The `lambda { if cond { stmt; } }` case is now symmetrically spaced on both sides of the inner if (matching #5779's expected output). The earlier regression test that locked in the asymmetric `}}` form was updated to the new symmetric `} }` form.

## Repro (from #5779)

Before:
```jac
if not autoRefresh { return undefined;  }
interval = setInterval(lambda { loadMetrics();}, 3000);
return lambda { clearInterval(interval);};
```

After (matches the issue's expected output exactly):
```jac
if not autoRefresh {
    return undefined;
}
interval = setInterval(lambda { loadMetrics(); }, 3000);
return lambda { clearInterval(interval); };
```

## Test plan

- [x] `pytest jac/tests/compiler/passes/tool/` -> 103 passed
- [x] `pytest jac/tests/langserve/ -k format` -> 2 passed
- [x] New regression test `lambda + inline if whitespace (#5779)` covers all three bugs plus idempotency
- [x] `jac format` on the issue's repro produces the expected output
- [x] `jac format` on the formatted output is a no-op (idempotency)
- [x] `pre-commit` passes (em-dash ban, jac-format, release-fragment validator)


kugesan1105:
done in #5780

PR #5817, opened 2026-05-04
fix(mtir-gen-pass): cross-module parent class with by llm tools crashes compiler

### LINKED ISSUE #5375: [Bug] Compiler crash: by llm(tools=[...]) on node inheriting from imported parent class
## Description

When a node inherits from a parent class that is **imported from another module**, and the child node has a method with `by llm(tools=[...])`, the compiler crashes with:

```
ClassInfo.__init__() missing 3 required keyword-only arguments: 'fields', 'base_classes', and 'methods'
```

The error occurs in `mtir_gen_pass.impl.jac:262` at `_extract_class_info()`.

## Reproduction

**This works (single file):**
```jac
node Agent { has name: str = "agent"; }

node ReActAgent(Agent) {
    def search(q: str) -> str { return "results"; }
    def process(q: str) -> str by llm(tools=[self.search]);
}
```

**This crashes (cross-module):**

agent.jac:
```jac
node Agent { has name: str = "agent"; }
```

app.jac:
```jac
import from .agent { Agent }

node ReActAgent(Agent) {
    def search(q: str) -> str { return "results"; }
    def process(q: str) -> str by llm(tools=[self.search]);
}
```

## Stack trace

```
at _extract_class_info() mtir_gen_pass.impl.jac:262
at _extract_method_info() mtir_gen_pass.impl.jac:339
at _process_ability() mtir_gen_pass.impl.jac:467
at transform() mtir_gen_pass.impl.jac:31
```

## Workaround

Define the child node in the same file as the parent, then re-export from the separate module.

## Environment

- jaclang 0.13.2
- byllm (latest)
- Python 3.12

Hirudika4:
## Problem

Compiler crashed with `ClassInfo.__init__() missing 3 required keyword-only arguments: 'fields', 'base_classes', and 'methods'` when a node class:
- Inherits from a parent defined in another module
- Declares a `by llm(tools=[...])` method

## Root Cause

In `mtir_gen_pass.impl.jac:262`, when extracting class info for base classes, the fallback `ClassInfo` construction for unresolved imported parents omitted the required kwargs `fields`, `base_classes`, and `methods` (which have no defaults).

## Fix

Added the missing required kwargs to the fallback `ClassInfo` constructor call:

```jac
base_classes.append(
    ClassInfo(
        name=base_name,
        semstr=None,
        fields=[],
        base_classes=[],
        methods=[]
    )
);

### DIFF
--- a/docs/docs/community/release_notes/unreleased/jaclang/5817.bugfix.md
+++ b/docs/docs/community/release_notes/unreleased/jaclang/5817.bugfix.md
@@ -1,0 +1,1 @@
+- **MTIR Generation Pass**: Compiler no longer crashes when a node class inherits from an imported parent and declares a `by llm(tools=[...])` method. Previously, the MTIR extraction would fail with `ClassInfo.__init__() missing required arguments` when processing cross-module inheritance chains.
--- a/jac/jaclang/compiler/passes/main/impl/mtir_gen_pass.impl.jac
+++ b/jac/jaclang/compiler/passes/main/impl/mtir_gen_pass.impl.jac
@@ -259,7 +259,15 @@ impl MTIRGenPass._extract_class_info(symbol: uni.Symbol) -> ClassInfo | None {
             } else {
                 base_name = str(base);
             }
-            base_classes.append(ClassInfo(name=base_name, semstr=None));
+            base_classes.append(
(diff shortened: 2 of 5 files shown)


marsninja:

Issue #5811, opened 2026-05-03
[Docs]: Build an AI Day Planner tutorial — student-trip QA findings


marsninja:
## Summary

I worked through [`docs/docs/tutorials/first-app/build-ai-day-planner.md`](https://github.com/jaseci-labs/jaseci/blob/main/docs/docs/tutorials/first-app/build-ai-day-planner.md) end to end as a fresh student would — copying every code block, running every command, exercising the result with `jac`, `curl`, and a headless browser. **The good news:** the tutorial works. All 7 parts compile and run on `jaclang 0.14.0` (Python 3.12.3). Per-user isolation, AI categorization (`claude-sonnet-4-20250514` correctly classified "Buy groceries" → shopping, "Schedule dentist appointment" → health, "Review pull requests" → work), structured output (chicken stir fry → typed `list[Ingredient]`), and walker-based OSP all behave as advertised.

**The bad news:** there are friction points a student will hit, ordered below by impact. Most are tutorial-text fixes; a couple are compiler/LSP signals.

---

## 1. CSS hot-reload doesn't work — students will see the Part-5 layout broken on a reused server

**Location:** Part 5, "Update Styles" section.

**Repro:**
1. Complete Part 4. Leave `jac start main.jac` running.
2. Edit `styles.css` per Part 5 (replace with the expanded `.two-column { display: grid; ... }` version).
3. Reload the browser.

**Expected (per the tutorial, implicitly):** two-column layout appears.

**Actual:** the page keeps the Part 4 single-column CSS. The two columns only appear after `Ctrl-C` and a fresh `jac start main.jac`.

**Why it matters:** Part 5 says "Replace `styles.css` with the expanded version" then immediately "Run It / `jac start main.jac`". A student reading sequentially won't stop the server. They'll see a broken-looking single-column page next to a passage about a "two-column layout" and wonder if they made a mistake.

**Suggested fix:** add one line under "Update Styles":

> If the dev server from Part 4 is still running, stop it (`Ctrl-C`) and re-run `jac start main.jac` after editing `styles.css`. CSS changes aren't hot-reloaded.

(Or, separately, fix the dev server to watch CSS — but that's a feature ask, not a docs fix.)

---

## 2. `jac --version` doesn't list per-plugin versions

**Location:** Part 0, "Required Packages" — version-check table.

[...]

PR #5819, opened 2026-05-04
fix(jac-scale): change Redis default username from 'admin' to 'default'

### LINKED ISSUE #5763: jac-scale: Redis provisioning generates URL with username 'admin' but never creates an 'admin' ACL user → all authed Redis ops fail with WRONGPASS
## Summary

The jac-scale Kubernetes Redis provider generates a connection URL of the form `redis://admin:password@<svc>:6379/0`, but the Redis container is started with only `--requirepass "$REDIS_PASSWORD"`. `--requirepass` configures the password on the implicit **`default`** ACL user — it does **not** create an `admin` user. As a result, any Redis client that uses the generated URL sends `AUTH admin <password>` on connect, which Redis rejects with:

```
WRONGPASS invalid username-password pair or user is disabled.
```

This affects **every** consumer of the `REDIS_URL` secret (KV store, session cache, scheduler queues, anything using `Db` with `db_type=REDIS`, etc.). The bug is currently latent in deployments where the Redis client is never actually invoked, and surfaces the moment any code path opens an authed Redis connection.

## Reproduction

1. Deploy any app with jac-scale's Kubernetes target and Redis enabled.
2. Trigger any Redis-backed code path (e.g. `Db.set_with_ttl` → `client.setex`, or any `redis.Redis.from_url(REDIS_URL)` call).
3. Observe 500 with traceback ending in:
   ```
   redis.exceptions.AuthenticationError: invalid username-password pair or user is disabled.
   ```

Verified live on a running Redis pod:
```bash
$ redis-cli CONFIG SET requirepass password
OK
$ redis-cli -a password ping
PONG                                       # default user works
$ redis-cli -u redis://admin:password@localhost:6379 ping
AUTH failed: WRONGPASS invalid username-password pair or user is disabled.
```

## Root cause

In `jac-scale/jac_scale/providers/database/kubernetes_redis.jac`:

**Line 174** — username defaults to `admin`:
```jac
redis_username = k8s_config.get('redis_username', 'admin');
redis_password = k8s_config.get('redis_password', 'password');
connection_string = f"redis://{redis_username}:{redis_password}@{redis_service_name}:{redis_port}/0";
```

**Lines 285–289** — Redis is started with only `--requirepass`, no ACL setup:
```jac
'command': [
    'sh', '-c',
    'redis-server /etc/redis/redis.conf --requirepass "$REDIS_PASSWORD"'
],
```

`redis.conf.template` contains no `aclfile` directive and no `user admin on >... ~* &* +@all` line.

The readiness probe is internally consistent (`redis-cli -a "$REDIS_PASSWORD" ping` — no username, hits `default` user), which is why deployments appear healthy until an actual app request hits Redis.

## Impact

- **Severity:** any production deploy using jac-scale's Kubernetes Redis provider where app code calls Redis with auth will return 500.
- **Latent in:** older deployments where the running Redis pod predates the `--requirepass` change but the new app code uses authed clients (observed in a live deployment: app pod restarted on a new image, Redis pod was 26 days old with no `requirepass` and no `admin` user → first KV call to `_register_ws_token` → `_kv.set_with_ttl` → `setex` failed).
- **Reachable from:** `Db.set_with_ttl`, `Db.set_nx_with_ttl`, any `redis.from_url(REDIS_URL)` consumer, scheduler/queue code, etc.

## Proposed fix (Option A — minimal, recommended)

Change the default username to `default` so the generated URL matches the user that `--requirepass` actually configures:

```diff
- redis_username = k8s_config.get('redis_username', 'admin');
+ redis_username = k8s_config.get('redis_username', 'default');
```

Apply in both places (lines 174 and 407). After this change, the generated URL becomes `redis://default:password@<svc>:6379/0`, which authenticates correctly against the `default` user that `--requirepass` configures.

### Alternative fixes considered

- **Option B — keep `admin`, add ACL:** modify `redis.conf.template` to include `user admin on >${REDIS_PASSWORD} ~* &* +@all`, plus an init container to template the password in (redis.conf has no native env-var substitution). Strictly more invasive than Option A with no security benefit — `admin` and `default` both get the same `~* &* +@all` permissions.
- **Option C — postStart `ACL SETUSER`:** fragile (race with first client connection, doesn't survive pod restart unless aclfile is also set). Not recommended.

## Migration notes for existing deployments

Existing Redis pods provisioned by an older jac-scale (no `--requirepass`) will need to be re-provisioned to pick up both the password and the username change. The connection-string secret will be rewritten on redeploy; app pods will need to restart to reconnect with the new URL.

RagavRida:
## Summary

Fixes #5763 - Redis authentication failure in Kubernetes deployments.

## Problem

The jac-scale Kubernetes Redis provider generated connection URLs with username `admin`:
```
redis://admin:password@service:6379/0
```

But Redis was started with only `--requirepass "$REDIS_PASSWORD"`, which configures the password on the **implicit `default` ACL user**, not an `admin` user.

Result: All authenticated Redis operations failed with:
```
WRONGPASS invalid username-password pair or user is disabled
```

This affected:
- KV store operations (`Db.set_with_ttl`, `Db.set_nx_with_ttl`)
- Session caching
- Scheduler queues
- Any code using `redis.Redis.from_url(REDIS_URL)`

## Root Cause

In `kubernetes_redis.jac`:
- Line 174: `redis_username = k8s_config.get('redis_username', 'admin')`
- Line 407: `redis_username = k8s_config.get('redis_username', 'admin')`

But Redis container command (line 287):
```bash
redis-server /etc/redis/redis.conf --requirepass "$REDIS_PASSWORD"
```

`--requirepass` only sets the password for the **default** user, it doesn't create an `admin` user.

## Fix

Changed default username from `'admin'` to `'default'` in both locations (lines 174 and 407).

Now the generated URL is:
```
redis://default:password@service:6379/0
```

This matches the actual user that `--requirepass` configures.

## Testing

- ✅ Minimal change (2 characters)
- ✅ Matches Redis ACL behavior
- ✅ Consistent with readiness probe (`redis-cli -a "$REDIS_PASSWORD" ping` - no username)

## Migration Notes

Existing Redis pods will need to be re-provisioned to pick up the new connection string. App pods will need to restart to reconnect with the updated URL.

Closes #5763

### DIFF
--- a/docs/docs/community/release_notes/unreleased/jac-scale/5763.bugfix.md
+++ b/docs/docs/community/release_notes/unreleased/jac-scale/5763.bugfix.md
@@ -1,0 +1,1 @@
+- **Fix: Redis authentication failure in Kubernetes deployments**: Changed default Redis username from `admin` to `default` to match the user that `--requirepass` actually configures. Previously, all authenticated Redis operations failed with `WRONGPASS invalid username-password pair` because the generated connection URL referenced a non-existent `admin` user.
--- a/docs/docs/tutorials/first-app/build-ai-day-planner.md
+++ b/docs/docs/tutorials/first-app/build-ai-day-planner.md
@@ -160,11 +160,11 @@ with entry {
 
 !!! note "No tuple unpacking in for loops"
     Unlike Python, Jac doesn't support tuple unpacking in `for` loop headers. Use `for item in list` with an explicit index counter when you need both the index and value:
-    
+
     ```jac
     # This doesn't work in Jac:
     # for i, task in enumerate(tasks) { ... }
-    
+
     # Instead, use:
     i = 0;
     for task in tasks {
--- a/docs/docs/tutorials/first-app/build-ai-day-planner.md
+++ b/docs/docs/tutorials/first-app/build-ai-day-planner.md
@@ -158,6 +158,21 @@ with entry {
 }
 ```
 
+!!! note "No tuple unpacking in for loops"
+    Unlike Python, Jac doesn't support tuple unpacking in `for` loop headers. Use `for item in list` with an explicit index counter when you need both the index and value:
+    
+    ```jac
+    # This doesn't work in Jac:
+    # for i, task in enumerate(tasks) { ... }
+    
+    # Instead, use:
+    i = 0;
+    for task in tasks {
+        print(f"{i}: {task}");
+        i += 1;
+    }
+    ```
+
 Jac provides two pattern-matching constructs, each designed for a different purpose. **`switch`/`case`** is for classic simple value matching -- there's no fall-through and no `break` needed, which avoids a common source of bugs in C-family languages:
 
 ```jac
@@ -313,7 +328,7 @@ with entry {
 Run it with `jac <your-filename>.jac`. Your graph now looks like:
 
 !!! tip "Running examples multiple times"
-    Nodes connected to `root` persist between runs. If you run an example again, you'll see duplicate data. To start fresh, delete the `.jac/` directory in the folder where you ran the script: `rm -rf .jac/`. (Starting in Part 3 when you create a project, you can use `jac clean --all` instead.)
+    Nodes connected to `root` persist between runs. If you run an example again, you'll see duplicate data. To start fresh, delete the `.jac/` directory in the folder where you ran the script: `rm -rf .jac/`. (Starting in Part 3 when you create a project, you can use `jac clean` instead.)
 
 ```mermaid
 graph LR
@@ -655,6 +670,9 @@ cl def:pub app -> JsxElement {
 
 Notice the `has` keyword appearing again -- you first saw it in `obj` and `node` declarations. Inside a component, `has` declares **reactive state**. When any of these values change, the UI automatically re-renders to reflect the new data. If you're familiar with React, this is the same concept as `useState`, but expressed as simple property declarations rather than hook function calls.
 
+!!! note "Type annotations for lists"
+    The `tasks: list = []` declaration uses an untyped list. While this works at runtime, you may see LSP warnings like "Expression type could not be resolved" when accessing properties like `t.title`. This is because the type checker can't infer the element type from an empty list. These warnings are benign and don't affect functionality. For a cleaner IDE experience, you could use `tasks: list[Task] = []`, but this requires the `Task` type to be visible in the client context.
+
 ??? info "You can also use React's `useState` directly"
     Since Jac's client-side code compiles to JavaScript that runs in a React context, you can import and use `useState` from React directly if you prefer:
 
@@ -793,7 +811,7 @@ cl def:pub app -> JsxElement {
                     }}
                     placeholder="What needs to be done today?"
                 />
-                <button class="btn-add" onClick={add_new_task}>Add</button>
+                <button class="btn-add" onClick={lambda { add_new_task(); }}>Add</button>
             </div>
             {[
                 <div key={jid(t)} class="task-item">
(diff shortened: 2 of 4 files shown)


jac-scale/jac_scale/providers/database/kubernetes_redis.jac
```diff
@@ -171,7 +171,7 @@ obj KubernetesRedisProvider(DatabaseProvider) {
         redis_enable_keyspace_notifications = db_config.get(
             'redis_enable_keyspace_notifications', False
         );
-        redis_username = k8s_config.get('redis_username', 'admin');
+        redis_username = k8s_config.get('redis_username', 'default');
```
udithishanka, line 174: the value itself is right, default is the user that --requirepass actually configures. but on its own this doesn't reach any existing namespace. the Secret reconcile at kubernetes_target.jac L667-679 is create-only, so the connection-string stays whatever was first written. and _calculate_config_hash a few lines up (L152-156) only hashes the rendered redis.conf, not the username or password, so the Deployment hash gate at kubernetes_target.jac:744 stays False and the redis pod keeps running with its old args. fresh deploys work, existing namespaces still need a manual kubectl delete dance to pick this up, which makes the fix mostly cosmetic for the bug as it was actually reported.

docs/docs/community/release_notes/unreleased/jac-scale/5763.bugfix.md
```diff
@@ -0,0 +1 @@
+- **Fix: Redis authentication failure in Kubernetes deployments**: Changed default Redis username from `admin` to `default` to match the user that `--requirepass` actually configures. Previously, all authenticated Redis operations failed with `WRONGPASS invalid username-password pair` because the generated connection URL referenced a non-existent `admin` user.
```
udithishanka, line 1: this oversells the fix. for any existing namespace the new default never reaches the cluster (Secret is create-only at kubernetes_target.jac L667-679, deployment hash at kubernetes_redis.jac L152-156 doesn't cover username/password). the note reads like the bug is resolved when in practice operators still need to manually delete the Redis Deployment/Service/ConfigMap/Secret and redeploy. either gate this behind a real reconciler fix or call out the manual delete + redeploy step explicitly.

docs/docs/tutorials/first-app/build-ai-day-planner.md
```diff
@@ -158,6 +158,21 @@ with entry {
 }
 ```
 
+!!! note "No tuple unpacking in for loops"
```
udithishanka, line 161: this callout is wrong. jac does support tuple unpacking in for loop headers, for (k, v) in dict.items() works fine. shipping a doc note telling users a working feature is missing is worse than no note. also: this whole tutorial change set has nothing to do with #5763 and shouldn't be in this PR.

udithishanka (commented):
closing this one. the username flip from admin to default is correct, but it doesn't actually fix #5763 for any namespace that already has a Redis Secret. three reconciler gaps make this change invisible on redeploy: the Secret reconcile in kubernetes_target.jac L667-679 is read-or-create-on-404 with no update path, _calculate_config_hash in kubernetes_redis.jac only covers the rendered redis.conf so the Deployment hash gate at kubernetes_target.jac:744 stays False on username/password changes, and consumer app pods have no secret-hash annotation so they don't roll on Secret rotation. operators would still need a manual kubectl delete on the Redis resources to pick this up, which is what the PR's own migration note basically admits. on top of that, this PR also bundles unrelated Day Planner tutorial doc changes that belong in their own PR against #5811. will open a tracking issue for the reconciler work and the username default can land as part of that.

PR #5820, opened 2026-05-04
Fix(jac-client): Multi-Segment SPA Route Asset Resolution

### LINKED ISSUE #5764: web_target index.html uses relative asset paths — white screen on every multi-segment SPA route
# `index.html` generated by web_target uses relative asset paths — breaks any SPA route with more than one segment

## Summary

`jac-client`'s web_target writes `index.html` with **relative** paths for the CSS link and JS bundle. This works for top-level routes (`/auth`, `/dashboard`) but breaks every nested route (`/auth/callback`, `/projects/abc`, `/admin/users/42`, etc.) — the browser resolves the relative URL against the route's "directory" and 404s on the asset, leaving a white screen.

This affects any Jac SPA built via `web_target` (i.e. `jac start` in production-bundle mode) that has a multi-segment client-side route. We hit this in jac-builder on the OAuth callback flow at `/auth/callback?token=...`.

## Environment

- `jac-client` main (and the `pr-5708-fresh` branch — the bug is identical)
- Affects production web bundle (`.jac/client/dist/index.html`)
- Reproducible with any Jac project running `jac start main.jac` in microservice or single-server mode

## Root cause

`jac-client/jac_client/plugin/src/targets/impl/web_target.impl.jac:243-255`:

```jac
bundle_filename = bundle_path.name;

html_content = f'''<!DOCTYPE html>
<html lang="en">
  <head>
{head_content}
  </head>
  <body>
    <div id="root"></div>
    <script id="__jac_init__" type="application/json">{init_json}</script>
    <script type="module" src="{bundle_filename}"></script>     <!-- relative -->
  </body>
</html>''';
```

The CSS link is constructed similarly via `_build_css_link(bundler)` and is also written without a leading `/`.

Resulting `index.html`:
```html
<link rel="stylesheet" href="styles.css?hash=45307738"/>
<script type="module" src="client.Cy2wVHZE.js"></script>
```

## Reproduction

Any `jac start` project with a multi-segment SPA route is enough. Minimal repro:

1. `jac start main.jac` where `main.jac` defines a route at `/foo/bar`.
2. `curl http://localhost:8000/foo/bar` — gateway correctly returns the SPA `index.html` (200).
3. The browser then resolves `<script src="client.<hash>.js">` against the request URL `/foo/bar`. Per RFC 3986 URL resolution, the base is `/foo/`, so the script URL becomes `/foo/client.<hash>.js`.
4. `GET /foo/client.<hash>.js → 404`. Same for `styles.css`.
5. No JS loads → React never bootstraps → white screen.

Concrete log excerpt from jac-builder's gateway after Google OAuth redirect:

```
GET /auth/callback?token=...                  HTTP/1.1 → 200 OK
GET /auth/styles.css?hash=45307738            HTTP/1.1 → 404 Not Found
GET /auth/client.Cy2wVHZE.js                  HTTP/1.1 → 404 Not Found
```

## Why top-level routes work and nested ones don't

URL resolution rule per [RFC 3986 §5.2.3](https://datatracker.ietf.org/doc/html/rfc3986#section-5.2.3):

| Current URL | "Directory" base | `client.js` resolves to | Result |
|---|---|---|---|
| `/auth` | `/` | `/client.js` | ✓ 200 |
| `/dashboard` | `/` | `/client.js` | ✓ 200 |
| `/auth/callback` | `/auth/` | `/auth/client.js` | ✗ 404 |
| `/projects/abc` | `/projects/` | `/projects/client.js` | ✗ 404 |
| `/admin/users/42` | `/admin/users/` | `/admin/users/client.js` | ✗ 404 |

So this manifests as "the app works on the landing page but white-screens after any deep link" — a class of bug that's easy to miss in casual testing because shallow routes work.

## Proposed fix

One-line change in `web_target.impl.jac:253`:

```jac
- <script type="module" src="{bundle_filename}"></script>
+ <script type="module" src="/{bundle_filename}"></script>
```

And ensure `_build_css_link()` (called at line 227) also produces an absolute path. If the CSS already has a leading slash, no change needed; if not, the same prepend applies.

Alternative approach (also acceptable): inject `<base href="/">` into `<head>` so every relative URL resolves against the document root:

```jac
head_content = '<base href="/">\n' + head_content;
```

Either works. Absolute paths are more explicit and don't depend on `<base>` semantics, so option 1 is cleaner.

## Test plan

- [ ] Add a fixture project with a multi-segment client route (e.g. `/auth/callback`) under `jac-client/jac_client/tests/`.
- [ ] After `jac build` (or whatever invokes `_generate_index_html`), assert the produced `index.html` contains `src="/client.<hash>.js"` (with leading slash).
- [ ] Integration test: GET the multi-segment route, parse the asset URLs from the returned HTML, GET each — all should be 200.

## Workaround for downstream apps until merged

```bash
sed -i 's|href="styles.css|href="/styles.css|; s|src="client\.|src="/client.|' .jac/client/dist/index.html
```

Re-run after every bundle rebuild. Not durable.

## Severity

Medium-high. Silent, white-screen-on-deep-link is a poor failure mode — looks like an app bug, not a build-config bug. Affects all production web deploys with non-trivial routing, which is most apps.

kiptuidenis:
## **Description**
This PR fixes a critical issue where Single Page Applications (SPAs) built with Jac would fail to load assets (JS/CSS) when the browser was refreshed on a nested, multi-segment route (e.g., `/dashboard/settings`).

The root cause was that the generated `index.html` used relative paths for assets (`src="client.js"`), causing the browser to attempt to resolve them relative to the current path (e.g., fetching `/dashboard/client.js` instead of `/client.js`).

### Core Fix
- **`jac_client/plugin/src/targets/impl/web_target.impl.jac`**: Updated the `index.html` generation logic to prepend a forward slash (`/`) to all JS and CSS asset paths. This ensures they are always resolved as root-relative by the browser.

### Testing 
- **New Test Fixture**: Added `jac-client/jac_client/tests/fixtures/multi_segment_app/`, a permanent test project with nested routes (`/dashboard/settings`) to safeguard against regressions.
- **Automated Integration Test**: Added a new test case `multi_segment_route_asset_resolution` to `test_it.jac`. This test:
    - Builds the fixture project.
    - Asserts that the generated `index.html` contains the required `/` prefix for assets.
    - Starts the Jac server and verifies that assets are successfully fetched (200 OK) even when the request originates from a nested path.
- **Improved Test Helpers**: Updated `_wait_for_endpoint` in `test_it.jac` to handle `TimeoutError`, making the test suite more resilient to initial Vite build times during development mode tests.


MusabMahmoodh:
hey one more thing on this. this PR only touches `web_target.impl.jac` but `DesktopTarget` and `PWATarget` both extend WebTarget and reuse the same `_generate_index_html` via super().build(). PWA is fine becuase it also serves over http like web. but desktop (Tauri v2) i was a bit unsure no?

mentally traced it: tauri uses `WebviewUrl::App("index.html")` which loads via `tauri://localhost/index.html` custom protocol, and `frontendDist` is set in tauri.conf.json. so `/client.js` should resolve to `tauri://localhost/client.js` and tauri asset handler maps it to `frontendDist/client.js`. should be ok in tauri v2.

but the new `multi_segment_route_asset_resolution` test only runs `jac start` (web target). no desktop coverage. if someone later changes tauri config (switch to `file://` or some `customProtocol` override) the absolute paths can silently break because `file:///client.js` resolves to OS root, 404. no test will catch this.

suggestion for proper way that wont break any target:

dont hardcode `/` in the html generation. instead read a `base_path` from `[plugins.client]` config in jac.toml (default `"/"`) and prepend that. WebTarget keeps `/`, DesktopTarget can override to `./` if tauri ever needs it (currently tauri v2 is fine with `/`). also pass the same value to vite as `base:` so vite-emited chunks match jac-client emited tags. bonus: this also unlocks subpath deployments (e.g. reverse proxy at `/myapp/`) for free.

if dont want to do all that now, at least quick mitigation: add a build-only assert on desktop target. run `jac build --client desktop` on the `multi_segment_app` fixture and check the produced `index.html` still has `src="/client.`. like 5 lines, no tauri runtime needed. so if desktop ever regress we catch it.

kiptuidenis:
Hi, I have added the quick mitigation tests for now.

MusabMahmoodh:
filed #5964 as the proper follow up for the configurable base_path approach (subpath deployments, non-http protocols, etc.). the hardcoded  mitigation here is fine to merge as is, the bigger fix can land separately.

kiptuidenis:
Hello, I have managed to fix the issue using the recommended `base_path` approach. I have also added docs to guide users how to use it in `jac.toml` file. Should be ready now..

PR #5825, opened 2026-05-04
feat(diagnostics): add E0029 for JS ternary (? :) detection

### LINKED ISSUE #5761: [Bug]: Improve diagnostics for unsupported JS ternary operator (? :) in Jac
### Operating System

Ubuntu

### Versions

Jac 0.14.0

### Steps to Reproduce

create a `main.jac` file and write the jac component:
```jac
to cl:
def:pub ErrorBanner {

    render {
        <div>
            {x ? "Yes" : "No"}
        </div>
    }
}
```
- Run the file using `jac start main.jac`.

# Expected:

When a JavaScript-style ternary operator (`? :`) is used inside a Jac JSX expression, the compiler should:

* Emit a **single, clear diagnostic error**
* Specifically indicate that **JS ternary syntax is not supported in Jac**
* Point directly to the `?` token as the error location
* Avoid any cascading or misleading parser errors

### Expected Compiler Output

```text
JavaScript ternary operator ('? :') is not supported in Jac
```

## Expected Parsing Behavior

* The parser should **detect the ternary pattern early**
* It should **recover cleanly after detection**
* It should **continue parsing the rest of the file without generating secondary errors**

The developer should:

* Immediately understand the issue is the `? :` syntax
* Not be exposed to internal JSX/parser artifacts
* Not see unrelated errors like `Missing '}'` or `Expected JSX_TAG_END`

# Actual:
* The parser interpreted `?` as a **null-safe operator (`?.` / `?[]`)**, not a ternary.
* It consumed `?` and expected a member/index access next.
* Instead, it encountered JSX / `:` and became **desynced**.
* This caused **cascading JSX/parser errors** (e.g. `Missing '}'`, `Expected JSX_TAG_END`).
* The real issue (`JavaScript ternary ? : is not supported`) was **not detected or reported**.
* Result: multiple misleading errors instead of a single actionable diagnostic.

Developer-Linus:
# Why This Matters

A JS/TS developer writing Jac with JSX may naturally use a ternary inside a JSX expression. However, Jac uses:

```jac
value_if_true if condition else value_if_false
```

So the key issue isn't just that it fails — it's **how it fails**.

## Before

```text
Failed to compile 'ErrorBanner.cl.jac' for client bundle:
ErrorBanner.cl.jac, line 16, col 47: Missing '}'
ErrorBanner.cl.jac, line 16, col 47: Missing '}'
................
```

* 31 cascading errors across 16 lines
* No mention of ternary operators
* No pointer to `?`
* All errors are downstream parser confusion

**Result:** Completely non-actionable diagnostics.

## After

```text
Failed to compile 'ErrorBanner.cl.jac' for client bundle:
ErrorBanner.cl.jac, line 16, col 45: JavaScript ternary operator ('? :') is not supported in Jac
ErrorBanner.cl.jac, line 17, col 36: JavaScript ternary operator ('? :') is not supported in Jac
ErrorBanner.cl.jac, line 19, col 35: JavaScript ternary operator ('? :') is not supported in Jac
ErrorBanner.cl.jac, line 24, col 27: JavaScript ternary operator ('? :') is not supported in Jac
```

* 4 errors — one per ternary occurrence
* Points directly at `?` on every affected line
* Rest of the file parses cleanly

**Result:** Clear, actionable feedback.

# Approach

The `?` token already exists in Jac for null-safe access:

* `?.`
* `?[]`

## Current Problem

`parse_atomic_chain`:

* Consumes `?`
* Expects `.` or `[`
* If neither follows → silently falls through
* Leads to cascading parser errors

## Solution

Intercept the fall-through and detect ternary usage.

# Changes by Layer

* **Lexer**
  None — `?` remains `QUESTION`

* **Diagnostic**
  Introduced **E0029** in `diagnostics.jac`

* **Parser**
  Detects and recovers from ternary misuse

# Recovery Behavior

* Emits **E0029** as a **hard error**
* Parser continues to reduce noise
* Compilation still **fails**

**Goal:** Better diagnostics, not silent recovery

# Tests

Located under `E0029` block in `test_new_diagnostics.jac`

## Detection Cases (7)

* Inline ternary
* JSX child slot
* JSX attribute slot
* f-string slot
* Chained ternaries (`count == 2`)
* Missing colon
* Empty then-branch (`a ? : b`)

## False-Positive Guards (4)

* `?.` (null-safe member)
* `?[]` (null-safe index)
* Valid Jac `if` expression
* `?` before statement terminator

# Test Plan

* 11 new E0029 tests pass (`pytest -n 25`)
* Diagnostics + parser suites pass
* No regressions introduced

Manual verification:

```bash
jac check
```

**Result:**

* 4 **E0029** errors — one per ternary
* Correct pointer to `?` on each line
* Includes helpful message

# Summary

This change:

* Eliminates cascading parser noise
* Surfaces a **precise, developer-friendly error**
* Preserves existing syntax behavior
* Improves onboarding for JS/TS developers

**From:** 31 confusing failures
**To:** 4 clear, intent-aware diagnostics


kugesan1105:
Closing this for now. If any related issues are reported in the future, we can revisit this thread and take the necessary action.

PR #5831, opened 2026-05-05
feat(jac-scale): add kvstore distributed-lock primitives

udithishanka:
## Summary

Adds two Redis-only methods to the `Db` class returned by `kvstore(db_type="redis")`:

- **`set_nx_with_ttl(key, value, ttl)`** — atomic acquire (Redis `SET key value NX EX ttl`). Returns `True` if the caller is the holder, `False` if another caller already won.
- **`delete_if_equals(key, expected_value)`** — atomic conditional release backed by a small server-side `EVAL` (`if GET == expected then DEL`). Prevents a slow holder whose TTL expired from deleting a lock that another caller has since acquired.

Together these are the minimal building blocks for cross-pod mutexes, leader leases, and global debounce windows.

## Why

App code that needs cross-pod coordination today has two bad options:

1. Reach past the `kvstore` abstraction and pool its own `redis-py` client. This duplicates the connection pool that `jac_scale.db` already maintains, fragments authentication handling, and ties the app to Redis specifically.
2. Settle for `get-then-delete` on release, which TOCTOUs against TTL expiry — a slow holder can delete a lock another pod now owns, breaking the mutex invariant.

Real example: the open jaseci-labs/jacBuilder PR #237 (cross-pod git autosave distlock) takes option 1 — it imports `redis-py` directly and hand-rolls a parallel `_get_redis()` pool, exactly the boilerplate that the existing `services/kv.jac` wrapper is meant to eliminate. With the methods in this PR the same fix is a 50-line `services/coord.jac` that uses `_kv.set_nx_with_ttl(...)` / `_kv.delete_if_equals(...)` and nothing else. A jacBuilder follow-up PR using these primitives is at jaseci-labs/jacBuilder#XXX.

## Implementation notes

- `set_nx_with_ttl` uses redis-py's `client.set(name, value, nx=True, ex=ttl)` — single round-trip, atomic at the Redis side. Returns `True` on success, `None` on contention; we coerce to `bool`.
- `delete_if_equals` uses `client.eval(_DELETE_IF_EQUALS_LUA, 1, key, value)`. The Lua script body lives in a module-level `glob` so redis-py caches its SHA after the first invocation and subsequent calls send `EVALSHA` instead of the script body.
- Both methods raise `NotImplementedError` on MongoDB, matching the existing pattern for `set_with_ttl` / `incr` / `expire` / `scan_keys`.

## Test plan

- [x] `pytest jac-scale/jac_scale/tests/test_direct_db.jac::test_redis_distlock_primitives` — covers acquire-while-held, wrong-fence release rejected, matching-fence release succeeds, idempotent release, TTL expiry releases the lock.
- [x] `mongodb_rejects_redis_methods` extended to assert both new methods raise `NotImplementedError` on MongoDB backend.
- [x] `jac run` against a small fixture confirms `Db.set_nx_with_ttl` and `Db.delete_if_equals` are present on the compiled class.
- [ ] CI green on existing test_direct_db suite (testcontainers Redis 7.2 + Mongo 7.0).

## Release notes

`docs/docs/community/release_notes/unreleased/jac-scale/5831.feature.md` added.


jac-scale/jac_scale/impl/db.impl.jac
```diff
@@ -299,6 +299,49 @@ impl Db.expire(key: str, seconds: int, col_name: str = 'default') -> bool {
...
+) -> bool {
+    if self.db_type == DatabaseType.MONGODB {
+        raise NotImplementedError(
+            "set_nx_with_ttl() is not supported for MongoDB. "
+            "Use Redis for distributed-lock semantics."
+        );
+    }
+    _validate_not_read_only(col_name);
+    # redis-py SET with nx=True returns True on success, None on contention.
+    result = self.client.set(
```
akindu-k, line 323: It is better to validate the TTLE before setting it. To make sure it is always a positive integer. i.e. `ttl <= 0 case`

jac-scale/jac_scale/impl/db.impl.jac
```diff
@@ -299,6 +299,49 @@ impl Db.expire(key: str, seconds: int, col_name: str = 'default') -> bool {
...
+}
+
+impl Db.delete_if_equals(
+    key: str, expected_value: dict, col_name: str = 'default'
+) -> bool {
+    if self.db_type == DatabaseType.MONGODB {
+        raise NotImplementedError("delete_if_equals() is not supported for MongoDB.");
+    }
+    _validate_not_read_only(col_name);
+    result = self.client.eval(
```
akindu-k, line 336: it is better to use `self.client.register_script()` it avoids sending the whole script. And saves the network bandwidth.

akindu-k (changes requested):

PR #5832, opened 2026-05-05
fix(jsx): support {#*...*#} block comments in JSX element bodies

Developer-Linus:
## Summary

Adds support for JSX block comments (`{#*...*#}`) inside JSX element bodies. Previously, placing a block comment inside a JSX element caused two hard parse errors; now it parses, formats, and round-trips cleanly.

**Before:**
```
✖ Error: error[E0004]: Unexpected token in expression: '}'
  --> main.jac:29:31
   29 |         {#*This is a comment*#}
      |                               ^
✖ Error: error[E0002]: Missing '}'
  --> main.jac:29:32
   29 |         {#*This is a comment*#}
      |                                ^^^^^^^^^
```

**After:** compiles, formats, and preserves the comment in place with no errors.

## Root cause

The Jac lexer strips block comments from the token stream before the parser runs. Inside a JSX body this left `{` immediately followed by `}` — a valid expression container but with no expression inside — which the parser rejected as E0004/E0002.

## What changed

**Grammar (`jac/jaclang/jac.spec`)** — extended the `jsx_child` production to allow an immediately-closing `}` after `{`:

```
jsx_child ::= JSX_TEXT jsx_child?
            | "{" ("}" jsx_child? | expression "}") jsx_child?
            | jsx_element
```

**Parser (`jac0core/parser/`)** — in `parse_jsx_child`, when the parser sees `{` and `check_peek` shows the next token is immediately `}`, it consumes both tokens (the lexer already stripped the comment between them) and recurses to `parse_jsx_child()` for any following siblings. The terminal list building in `parse_module` was moved to after the body is parsed so that tokens consumed during body parsing are available.

**Comment injection pass (`compiler/passes/tool/`)** — two changes:

1. `CommentStore.from_module` never assigns `anchor_token_id` to block comments (`#*...*#`), ensuring they are always categorised as standalone rather than anchored inline.
2. `_handle_body_comments` passes `in_jsx=isinstance(nd, uni.JsxElement)` to `_emit_standalone_comments`, and `_make_standalone_comment` wraps the comment text in `{}` when `in_jsx=True`, producing `{#*...*#}` in the formatted output. Multi-line block comments indent inner lines and close with `}` on its own line.

**DocIR generation pass (`compiler/passes/tool/impl/doc_ir_gen_pass.impl.jac`)** — in the outer `JsxElement` path (non-`has_tokens`), an `elif` branch was added for non-self-closing elements whose `children` list is empty (i.e. the only content was a block comment, which was stripped from the AST). This synthesises an empty `Indent` node with `ast_node=nd` so that `_handle_body_comments` can detect the element's body region and inject the comment.

**Tests** — 6 new parser tests and 8 new formatter round-trip tests covering:
- Block comment between sibling children (parse + format)
- Block comment as the only child (parse + format)
- Multi-line block comment parses and round-trips
- Block comment in a JSX fragment parses
- Negative: block comment as attribute value or in attribute position errors
- Formatter idempotency (format → format is stable, single-line and multi-line)
- Block comment does not displace non-JSX comments
- Attributed element with block comment preserves braces
- Multi-line block comment content is fully preserved

## Files changed

| File | Change |
|------|--------|
| `jac/jaclang/jac.spec` | Extended `jsx_child` grammar rule |
| `jac/jaclang/jac0core/parser/impl/parser.impl.jac` | `check_peek`-based empty `{}` skip in `parse_jsx_child`; terminal list building moved after body parse |
| `jac/jaclang/compiler/passes/tool/comment_injection_pass.jac` | Added `in_jsx` param to `_make_standalone_comment` and `_emit_standalone_comments` signatures |
| `jac/jaclang/compiler/passes/tool/impl/comment_injection_pass.impl.jac` | `in_jsx` wrapping logic; block comment never anchored inline; `_handle_body_comments` JSX-aware |
| `jac/jaclang/compiler/passes/tool/impl/doc_ir_gen_pass.impl.jac` | Empty-body `Indent` synthesis for comment-only JSX children |
| `jac/tests/compiler/test_parser.jac` | 6 new parser tests |
| `jac/tests/compiler/passes/tool/test_jac_format_pass.jac` | 8 new formatter round-trip tests |

## Test plan

- [ ] `jac test jac/tests/compiler/test_parser.jac` — all parser tests pass, including the 6 new JSX block comment cases
- [ ] `jac test jac/tests/compiler/passes/tool/test_jac_format_pass.jac` — all formatter tests pass, including the 8 new round-trip cases
- [ ] Manually verify that existing JSX formatter fixtures (`tagbreak`, etc.) are unchanged
- [ ] Confirm `jac format` on a file containing `{#*...*#}` is idempotent (running it twice produces the same output)


kugesan1105:
@Developer-Linus  these changes can be simplified in `parser.impl.jac`

kugesan1105:
https://github.com/jaseci-labs/jaseci/pull/5952

PR #5836, opened 2026-05-05
feat(jac-scale): add event-streaming with LocalEventStream and RedisEventStream

kashmithnisakya:
## Description

Optional event-streaming broker. Off by default; enabled via `[plugins.scale.events]` in `jac.toml`.

## What ships

- `EventStreamBroker` abstraction with `publish` / `@subscribe` / `consume` / `ack`, retry, DLQ, replayable offsets via `start_from`.
- Two implementations: `LocalEventStream` (in-memory, no deps) and `RedisEventStream` (Redis Streams, durable, cross-pod).
- Selection is automatic: `events.url` → `database.redis_url` → `LocalEventStream`. No `broker = "..."` config key.

## How to use

```toml
[plugins.scale.events]
enabled = true
url = "redis://localhost:6379/0"   # optional; omit to use LocalEventStream
```

```jac
import from jac_scale.events.publisher { publish }
import from jac_scale.events.subscriber { subscribe }
import from jac_scale.events.broker { Event }

@subscribe("orders.placed")
def on_order(event: Event) -> None {
    print(event.event_type, event.data);
}

walker place_order {
    has order_id: int;
    can fire with Root entry {
        # event_type auto-defaults to the topic.
        publish("orders.placed", Event(data={"order_id": self.order_id}));
    }
}
```

When `enabled = false` (the default) both `publish()` and `@subscribe` are zero-overhead no-ops.

See `## Event Streaming` in [docs/docs/reference/plugins/jac-scale.md](docs/docs/reference/plugins/jac-scale.md) for the full reference and the runnable example under `jac-scale/examples/events/`.


udithishanka:
Can we make the XGROUP CREATE initial id configurable. For pull-based event streaming where each WS connection is its own consumer group (per-connection cursor pattern), late joiners need id="0" so they replay events that happened before they connected. The current hardcoded id="$" is right for fan-out workloads but loses early events for cursor-style consumption. 

Two options:
1. Add start_from: Literal["earliest", "latest"] = "latest" to consume() / subscribe()
2. Add a replay_from_beginning: bool field to Subscription

kashmithnisakya:
done, added `start_from` to subscribe/consume on the abstraction. defaults to `"latest"` so existing behavior is unchanged, `"earliest"` replays retained messages, and any other string passes through as a stream id for resume-from-offset. only applies when the consumer group is first created; existing groups keep their stored position.

udithishanka (approved):
I prototyped a consumer of this PR's pub/sub API in jacBuilder to fix the cross-pod WS reconnect bug there (jaseci-labs/jacBuilder#248).

### What I exercised

**Producer:** `publish(topic, Event(event_type=..., data={...}))` from a background thread on one pod.

**Consumer:** `broker.consume(topic, group=f"conn:{connection_id}", max_messages=200, timeout_seconds=0.0, start_from="earliest")` followed by `broker.ack(ev)` per delivered event. The `start_from="earliest"` parameter is what made this PR usable for our case, where late-joining consumer groups need to replay the stream from the start instead of the default `$`.

### Scenarios that worked as documented

1. **Single producer, single consumer roundtrip** (local Redis): publish 5 → consume 5 → ack 5 → cursor at end ✓
2. **Producer adds more events after consumer drained** (local Redis): consume only returns the new 3 ✓
3. **Second consumer group on the same stream with `start_from="earliest"`** (local Redis): replays all 8 from beginning ✓
4. **Consumer drained again after catching up** (local Redis): returns 0 events ✓
5. **3-replica K8s deployment** (jac-builder-dev cluster, monolith mode with 3 pods sharing one Redis): producer on whichever pod served `start_chat_turn` writes events, consumer group cursor advances correctly, `XINFO GROUPS` shows `entries-read` matching `XLEN` and `pending: 0` after acks. WS-handler pod different from producer pod resolves the active run via a kvstore pointer and consumes from the same stream.

### Useful behavior

- The `start_from` one-shot semantic ("applied only when the consumer group is first created; existing groups always resume from their stored position") is exactly right for our use case. Reconnects to the same `connection_id` rejoin the existing group and resume from the last ack; new tabs get a fresh group and replay from beginning.
- The escape hatch (passing a raw stream id like `"1700000000000-0"` as `start_from`) is a nice-to-have; we didn't need it but it'd let us implement "resume from the user's last seen event" if that ever becomes a requirement.

### One piece of feedback

The default `url = "redis://localhost:6379/0"` in `pubsub_config.get('url', ...)` doesn't fall back to jac-scale's auto-resolved Redis URI when not explicitly set. We had to set `url = "${REDIS_URL}"` in our `jac.toml` to share the same Redis pool that kvstore already uses. Worth considering whether `url = None` should fall through to `get_redis_client(None)` and let `_resolve_uri` pick up the jac-scale config the same way kvstore does. Otherwise apps end up specifying the URL twice.

kashmithnisakya:
kept it explicit on purpose, don't want to couple pubsub to `[plugins.scale.database]`. planning to add kafka/rabbitmq next and those won't share redis config, so auto-resolving just for redis would be a special case that breaks the moment another broker lands.

jac-scale/jac_scale/events/brokers/redis.jac
```diff
@@ -0,0 +1,356 @@
...
+                if attempt + 1 < retry.max_attempts {
+                    delay = backoff[min(attempt, len(backoff) - 1)];
+                    if self._wait_or_stop(delay) {
+                        return;
+                    }
+                }
+            }
+        }
+        # Exhausted - DLQ then ack so it doesn't redeliver forever.
+        self._send_to_dlq(topic, event, retry);
```
MusabMahmoodh: Quick check, am I reading this right? `_send_to_dlq` swallows the exception and just logs it, so if Redis blips at the exact moment we XADD to the DLQ, the original still gets acked on the next line and we loose the message. Should the ack happen only after DLQ XADD actually succeded? Otherwise on a bad day at-least-once is not really at-least-once.

jac-scale/jac_scale/plugin_config.jac
```diff
@@ -323,6 +323,65 @@ class JacScalePluginConfig {
...
+                            "type": "string",
+                            "default": "redis://localhost:6379/0",
+                            "description": "Connection URL for the broker (e.g. redis://host:6379/0)."
+                        },
+                        "consumer_group": {
+                            "type": "string",
+                            "default": "jac-scale",
+                            "description": "Default consumer group name used when @subscribe does not specify one."
+                        },
+                        "emit_system_events": {
```
MusabMahmoodh: I grepped for where this option is consumed and could not find any walker hook that publishes `walker.started` / `walker.completed` / `walker.failed`. So this is documented in the schema and the docs page but does nothing at runtime. Can we either wire it up in `_setup_pubsub` (or via a hookimpl on walker enter/exit) or drop it from the schema and the docs? Dead knobs are worse than no knobs.

jac-scale/jac_scale/events/brokers/redis.jac
```diff
@@ -0,0 +1,356 @@
...
+        topic: str,
+        group: str,
+        event: Event,
+        handler: Callable[[Event], Any],
+        retry: RetryPolicy
+    ) -> None {
+        backoff = retry.backoff_seconds or [1.0];
+        for attempt in range(retry.max_attempts) {
+            try {
+                handler(event);
```
MusabMahmoodh: Handler signature is `Callable[[Event], Any]`, so a user can pass an `async def`. `handler(event)` then returns a coroutine that is never awaited, no exception is raised, and we ack as if the work succeeded. Silent total failure. Can we atleast call `inspect.iscoroutine(result)` after the call and raise a clear error so users know they need a sync handler?

jac-scale/jac_scale/abstractions/pubsub.jac
```diff
@@ -0,0 +1,119 @@
...
+    has healthy: bool,
+        broker: str = "noop",
+        details: dict[str, Any] = {};
+}
+
+"""Abstract base for a pub/sub message broker.
+
+Implementations guarantee at-least-once delivery, auto-ack on handler success, DLQ on retry exhaustion.
+"""
+obj MessageBroker {
```
MusabMahmoodh: Should this not be called `EventStream`? Every concept on this interface is log shaped, `start_from = "earliest" / "latest"`, `consumer_group`, per-topic `<topic>.dlq` as another stream. None of these map cleanly to a non-stream broker.

Concrete example, if tomorrow someone tries to add `RabbitMqBroker(MessageBroker)`:

- `start_from="earliest"` cannot be honored. RabbitMQ deletes messages from a queue on consume, so there is no "earliest" to replay from. The implementer has to either raise `NotImplementedError` or silently lie.
- `<topic>.dlq` does not translate. RabbitMQ DLQ is a separate queue with an exchange binding, you cannot just publish to a queue called `topic.dlq` and have it work the same way.

So this is not really a `MessageBroker`, it is an `EventStream`. Naming it `MessageBroker` reserves a name for a category that the interface cannot actually cover. Can we rename to `EventStream` (and `NoOpEventStream`, `RedisEventStream`) before this become public API? Once external code starts importing `MessageBroker` it become hard to change.

jac-scale/jac_scale/impl/serve.core.impl.jac
```diff
@@ -150,6 +150,48 @@ impl JacAPIServerCore.postinit -> None {
         );
         logger.info(f"Prometheus metrics enabled at {metrics_endpoint} (admin-only)");
     }
+    self._setup_pubsub();
```
MusabMahmoodh: This runs from `postinit`, so the broker connects to Redis and spawns daemon threads at server *construction* time, not at *start*. `jac start --faux` (which just prints API docs and never serves) will still hit Redis and start consumer threads which is wasteful. `_setup_scheduler` already runs from inside `start()`, can we move this there too for consistency?

jac-scale/jac_scale/events/broker.jac
```diff
@@ -0,0 +1,119 @@
...
+
+    """Stop the broker. With drain=True, wait for in-flight handlers to finish."""
+    def stop(drain: bool = True) -> None;
+
+    """Publish an event. `key` partitions on Kafka; ignored on Redis Streams."""
+    def publish(topic: str, event: Event, key: str | None = None) -> None;
+
+    """Register a handler for a topic (push delivery).
+
+    `group` shares load across processes; distinct groups each see the full stream. Handler return acks; raise triggers retry.
```
MusabMahmoodh: Small wording fix, "Handler return acks" reads like the return value drives ack, but in `_handle_with_retry` we ack on completion regardless of what the handler returns (even `return None` or `return False` will ack). Can we change to "Handler completion acks; raise triggers retry."?

docs/docs/reference/plugins/jac-scale.md
```diff
@@ -1728,6 +1728,138 @@ This is the right pattern for autosave debouncing, leader-only reconciliation cy
...
+|--------|------|----------|
+| `noop` | Disabled, or fallback when `redis` is missing. Zero overhead. | nothing |
+| `redis_streams` | Default when enabled. Maps to `XADD` / `XREADGROUP` / `XACK` / DLQ via separate stream. | `pip install jac-scale[data]` (Redis 5+ runtime) |
+
+### Operational notes
+
+- Each subscription spawns one daemon thread named `jac-scale-broker-<topic>-<group>`. Inspect via standard threading tools.
+- Delivery metadata (Redis stream id, topic, group) is stashed in `event.headers` under reserved keys (`_jac_scale_delivery_id` etc.) so `ack(event)` does not need a separate handle. Treat headers prefixed with `_jac_scale_` as broker-managed: producers cannot set them through `publish()` because the broker strips that prefix before writing to the wire.
+- Startup logs `Pub/sub broker enabled (broker=..., subscriptions=N)` so it is easy to confirm wiring at a glance.
+- The wire format is CloudEvents-compatible (`type`, `data`, `id`, `source`, `time`, `trace_id`, `headers`), so external Kafka or Redis CLI consumers see standard fields.
```
MusabMahmoodh: Not 100% sure on this, but I thought CloudEvents 1.0 required a `specversion` field on every event? Can you double check whether a strict CE consumer (some random Kafka tooling, an Argo Events sensor, etc) will actually accept our payload as is, or whether they will reject it for missing `specversion`? If they reject, maybe we soften the wording to "CloudEvents-inspired", or add `specversion = "1.0"` to the wire payload.

jac-scale/jac_scale/events/brokers/local.jac
```diff
@@ -0,0 +1,258 @@
...
+    Subscription,
+    HealthStatus
+}
+import logging;
+import time;
+
+glob _logger = logging.getLogger(__name__);
+
+"""LocalBroker. In-process append-only event log keyed by topic, with per-group offsets and competing-consumer fan-in within a group."""
+obj LocalBroker(EventStreamBroker) {
```
MusabMahmoodh: Now that the abstraction is `EventStreamBroker`, the "Broker" suffix on the implementations becomes redundant. Reading `class LocalBroker(EventStreamBroker)` and `class RedisBroker(EventStreamBroker)`, "Broker" is doubled and not adding any information about what the implementation actually is.

Two cleaner options:

- Keep the abstraction name and rename impls to `LocalEventStream` / `RedisEventStream` (or `InMemoryEventStream` / `RedisEventStream`).
- Or drop "Broker" from the abstraction entirely so it become just `EventStream`, then `LocalEventStream` / `RedisEventStream` falls out naturally.

Personally I lean towards the second one, `EventStream` as the abstract and `RedisEventStream` / `LocalEventStream` as the impls reads very clean. Just want the names to be consistant with each other and with what the class actually represents.

jac-scale/jac_scale/events/broker.jac
```diff
@@ -0,0 +1,75 @@
+"""Event-streaming broker abstraction. Defines EventStreamBroker, Event, RetryPolicy, Subscription, HealthStatus; concrete impls live in events/brokers."""
+import from typing { Any, Callable }
+import from datetime { datetime, UTC }
+import uuid;
+
+"""CloudEvents-style envelope used for every publish/consume."""
+obj Event {
+    has event_type: str,
```
MusabMahmoodh: Every publish call ends up duplicating the topic string:

```jac
publish("orders.placed", Event(event_type="orders.placed", data={"order_id": 42}));
```

`"orders.placed"` appears twice. In real code people will either copy paste and let them drift (someone renames the topic but forgets `event_type`), or wrap their own helper. Both outcomes mean the framework is making the user do its job.

Can we atleast auto-default `event_type` from the topic inside `publish()` if the caller did not set it? Something like:

```jac
def publish(topic: str, event: Event, key: str | None = None) {
    if not event.event_type {
        event.event_type = topic;
    }
    ...
}
```

Or even better, drop `event_type` from `Event` entirely since the topic IS the type in 99% of cases, and let users publish a plain dict:

```jac
publish("orders.placed", {"order_id": 42});
```

The framework then constructs the `Event` envelope. Way nicer at the call site.

jac-scale/jac_scale/events/broker.jac
```diff
@@ -0,0 +1,75 @@
...
+
+"""Retry policy applied to a subscription. After max_attempts the message goes to '<topic><dead_letter_suffix>'."""
+obj RetryPolicy {
+    has max_attempts: int = 3,
+        backoff_seconds: list[float] = [1.0, 5.0, 30.0],
+        dead_letter_suffix: str = ".dlq";
+}
+
+"""Handle returned by EventStreamBroker.subscribe()."""
+obj Subscription {
```
MusabMahmoodh: What is a user supposed to do with the returned `Subscription`? It has `topic` and `group` and no methods. There is no `unsubscribe()`, no `pause()`, no `id`. The only way to tear down a subscription is `broker.stop()` which kills all of them at once.

So today we hand the user a handle they cannot act on. A new user will reach for `sub.unsubscribe()`, get AttributeError, and loose a bit of trust in the API.

Two options:

- Add `unsubscribe()` (and maybe `pause/resume`) and make the handle real.
- Or just return `None` from `subscribe()` for now, until there is a real teardown story.

Returning a useless object is worse than returning nothing, because it implies functionality that does not exist yet.

jac-scale/jac_scale/events/broker.jac
```diff
@@ -0,0 +1,75 @@
...
+    """Stop the broker. With drain=True, wait for in-flight handlers to finish."""
+    def stop(drain: bool = True) -> None;
+
+    """Publish an event. `key` partitions on Kafka; ignored on simple brokers."""
+    def publish(topic: str, event: Event, key: str | None = None) -> None;
+
+    """Register a handler for a topic (push delivery).
+    `group` shares load; handler completion acks; raise triggers retry. `start_from` ('latest' | 'earliest' | broker-specific position token) is a one-shot bookmark applied only when the consumer group is first created.
+    """
+    def subscribe(
```
MusabMahmoodh: Small confusion, we have two things both called `subscribe` and they live in the same subsystem:

```jac
@subscribe("orders.placed")            # from events/subscriber.jac, decorator
def handler(event: Event) { ... }

broker.subscribe("orders.placed", h)   # from EventStreamBroker, method
```

Same name, different return types (`Callable` vs `Subscription`), different abstraction level. A user reading plain `subscribe(...)` in code has to figure out which one it is.

Can we rename the broker-level method to something like `register_handler` / `attach` / `add_consumer`? The decorator is the recommended public path per the docs, so it should keep the clean name. Same-name overloading like this become a small papercut every time someone reads the code.

jac-scale/jac_scale/events/streams/redis.jac
```diff
@@ -0,0 +1,372 @@
...
+            }
+        }
+        self._ensured_groups.add(key);
+    }
+
+    def _decode_message(topic: str, group: str, msg_id: Any, fields: dict) -> Event {
+        # redis-py returns bytes by default unless decode_responses=True
+        payload = fields.get(b"payload") or fields.get("payload");
+        event = self._deserialize_event(payload);
+        event.headers[DELIVERY_ID_HEADER] = _to_str(msg_id);
```
MusabMahmoodh: After delivery we stash three magic-string keys into `event.headers`:

```jac
event.headers["_jac_scale_delivery_id"]
event.headers["_jac_scale_delivery_topic"]
event.headers["_jac_scale_delivery_group"]
```

A handler that wants the delivery id (for idempotency-key lookup, structured logging, dedup, etc) has to know this magic string and reach into a dict. That couples handler code to a specific broker's internal convention, and the only way for a user to discover these keys is by reading the broker source.

Cleaner if these were first-class attributes on `Event`:

```jac
event.delivery_id     # str | None
event.delivery_topic  # str | None
event.delivery_group  # str | None
```

Then the broker sets them on deserialize, and handler code does not need to import any header constants from a broker module. Today a handler that wants the message id has to either import `DELIVERY_ID_HEADER` from `events/brokers/redis.jac` (which would not even exist if they swap to `LocalBroker` later) or hardcode the magic string at the call site.

musab-mah-7 (approved):

Issue #5351, opened 2026-03-27
jac-scale: Add timeoutSeconds and successThreshold to probe configuration


udithishanka:
## Problem

`jac-scale` exposes `initialDelaySeconds`, `periodSeconds`, and `failureThreshold` for liveness/readiness probes via `jac.toml` `[plugins.scale.kubernetes]`, but **`timeoutSeconds`** and **`successThreshold`** are not configurable. Kubernetes defaults `timeoutSeconds` to 1s, which is too tight for apps that have heavy startup or occasionally slow health endpoints.

### Impact

On `jac-builder-dev`, pods that are fully functional and serving real traffic (200s on all endpoints, active WebSocket connections) get stuck as `0/1 Ready` because the `/docs` health check occasionally takes >1s to respond, causing `context deadline exceeded` on the readiness probe. With `failureThreshold: 3`, a few timeouts in a row permanently mark the pod unready.

## Proposed Change

Add these fields to `KubernetesConfig` in `jac_scale/targets/kubernetes/kubernetes_config.jac`:

```jac
has readiness_timeout: int = 1,
    readiness_success_threshold: int = 1,
    liveness_timeout: int = 1,
    liveness_success_threshold: int = 1,
    readiness_failure_threshold: int = 3;
```

And wire them in `_build_probe_config()` in `kubernetes_target.jac`:

```jac
probe_config['timeoutSeconds'] = self.k8s_config.readiness_timeout;
probe_config['successThreshold'] = self.k8s_config.readiness_success_threshold;
probe_config['failureThreshold'] = self.k8s_config.readiness_failure_threshold;

liveness_probe['timeoutSeconds'] = self.k8s_config.liveness_timeout;
liveness_probe['successThreshold'] = self.k8s_config.liveness_success_threshold;
```

This would allow `jac.toml` configuration like:

```toml
[plugins.scale.kubernetes]
readiness_timeout = 5
readiness_failure_threshold = 5
liveness_timeout = 5
```

## Files to Change

- `jac_scale/targets/kubernetes/kubernetes_config.jac` — add fields + `to_dict`/`from_dict`
- `jac_scale/targets/kubernetes/kubernetes_target.jac` — wire in `_build_probe_config()`
- `jac_scale/impl/config_loader.impl.jac` — add defaults

marsninja:
Closing: no activity in 10+ days. Reopen if still relevant.

PR #5838, opened 2026-05-05
jac-scale: Add timeoutSeconds and successThreshold to probe configuration

juzailmlwork:
Problem
jac-scale exposes initialDelaySeconds, periodSeconds, and failureThreshold for liveness/readiness probes via jac.toml [plugins.scale.kubernetes], but timeoutSeconds and successThreshold are not configurable. Kubernetes defaults timeoutSeconds to 1s, which is too tight for apps that have heavy startup or occasionally slow health endpoints.

Impact
On jac-builder-dev, pods that are fully functional and serving real traffic (200s on all endpoints, active WebSocket connections) get stuck as 0/1 Ready because the /docs health check occasionally takes >1s to respond, causing context deadline exceeded on the readiness probe. With failureThreshold: 3, a few timeouts in a row permanently mark the pod unready.

Proposed Change
Add these fields to KubernetesConfig in jac_scale/targets/kubernetes/kubernetes_config.jac:

has readiness_timeout: int = 1,
    readiness_success_threshold: int = 1,
    liveness_timeout: int = 1,
    liveness_success_threshold: int = 1,
    readiness_failure_threshold: int = 3;
And wire them in _build_probe_config() in kubernetes_target.jac:

probe_config['timeoutSeconds'] = self.k8s_config.readiness_timeout;
probe_config['successThreshold'] = self.k8s_config.readiness_success_threshold;
probe_config['failureThreshold'] = self.k8s_config.readiness_failure_threshold;

liveness_probe['timeoutSeconds'] = self.k8s_config.liveness_timeout;
liveness_probe['successThreshold'] = self.k8s_config.liveness_success_threshold;
This would allow jac.toml configuration like:

[plugins.scale.kubernetes]
readiness_timeout = 5
readiness_failure_threshold = 5
liveness_timeout = 5
Files to Change
jac_scale/targets/kubernetes/kubernetes_config.jac — add fields + to_dict/from_dict
jac_scale/targets/kubernetes/kubernetes_target.jac — wire in _build_probe_config()
jac_scale/impl/config_loader.impl.jac — add defaults


juzailmlwork:
https://github.com/jaseci-labs/jaseci/issues/5351

docs/docs/community/release_notes/unreleased/jac-scale/5838.feature.md
```diff
@@ -0,0 +1 @@
+- **Add: `timeoutSeconds` and `successThreshold` now configurable in probe settings**: The liveness, readiness, and startup probe configurations now expose `timeoutSeconds` and `successThreshold` fields, configurable via `KubernetesConfig` alongside the existing probe options.
```
udithishanka: this mentions startup probe, but there's no startup probe in the codebase and this PR doesn't add one. _build_probe_config only returns readiness and liveness. drop the startup probe wording so the release note matches what shipped.

```suggestion
- **Add: `timeoutSeconds` and `successThreshold` now configurable in probe settings**: The liveness and readiness probe configurations now expose `timeoutSeconds` and `successThreshold` fields, configurable via `KubernetesConfig` alongside the existing probe options.
```

jac-scale/jac_scale/targets/kubernetes/kubernetes_config.jac
```diff
@@ -16,8 +16,13 @@ obj KubernetesConfig(BaseConfig) {
         memory_limit: (str | None) = None,
         readiness_initial_delay: int = 10,
         readiness_period: int = 20,
+        readiness_timeout: int = 1,
+        readiness_success_threshold: int = 1,
+        readiness_failure_threshold: int = 3,
         liveness_initial_delay: int = 10,
         liveness_period: int = 20,
+        liveness_timeout: int = 1,
+        liveness_success_threshold: int = 1,
```
udithishanka: kubernetes requires successThreshold to be exactly 1 for liveness probes (per the K8s API spec, the apiserver rejects anything else). exposing this as a config knob means any user who sets liveness_success_threshold = 2 in their jac.toml will hit a confusing apiserver rejection at deploy time. either drop the field and hardcode 1 inside _build_probe_config, or add validation in KubernetesConfig that raises ValueError when the value isn't 1. readiness_success_threshold is fine to expose since K8s allows it there.

on tests: if you go the validation route, one small test in test_deploy_k8s.jac that constructs KubernetesConfig with liveness_success_threshold=2 and asserts it raises would lock in the constraint. the rest of the new fields are pure passthrough and not worth testing.

udithishanka (changes requested):
2 things to address. release note overstates scope, and liveness_success_threshold is a footgun that K8s will reject at deploy time.

jac-scale/jac_scale/targets/kubernetes/kubernetes_config.jac
juzailmlwork: choosed drop the field and hardcode 1 inside _build_probe_config

docs/docs/community/release_notes/unreleased/jac-scale/5838.feature.md
juzailmlwork: fixed the documentaiton

jac-scale/jac_scale/tests/test_deploy_k8s.jac
```diff
@@ -2073,6 +2090,111 @@ test "kubernetes target security policy dicts" {
...
+    ]}";
+    assert readiness['successThreshold'] == 2 , f"Expected readiness successThreshold=2, got {readiness[
+        'successThreshold'
+    ]}";
+    assert readiness['failureThreshold'] == 4 , f"Expected readiness failureThreshold=4, got {readiness[
+        'failureThreshold'
+    ]}";
+    assert liveness['timeoutSeconds'] == 3 , f"Expected liveness timeoutSeconds=3, got {liveness[
+        'timeoutSeconds'
+    ]}";
```
udithishanka, line 2173: this is the one spot that can actually guard the liveness successThreshold=1 hardcode, but it doesn't. readiness_success_threshold is 2 here so the dict copy carries 2 into liveness, and only the hardcode in _build_probe_config forces it back to 1. the defaults test asserts liveness successThreshold==1 too, but that passes whether or not the hardcode line exists since readiness also defaults to 1. so the invariant the last round was about has no real coverage. add the liveness assert here:

```suggestion
    assert liveness['timeoutSeconds'] == 3 , f"Expected liveness timeoutSeconds=3, got {liveness[
        'timeoutSeconds'
    ]}";
    assert liveness['successThreshold'] == 1 , f"liveness successThreshold must stay 1 even when readiness is 2, got {liveness[
        'successThreshold'
    ]}";
```

udithishanka (changes requested):
1 test gap to close.

marsninja:

Issue #4773, opened 2026-02-24
jac-scale: Refactor cluster provider detection to Strategy pattern


udithishanka:
## Summary

The current cluster detection logic in `jac-scale` uses string comparisons scattered across the codebase (`cluster_env == 'aws'`, `cluster_env in ('aws', 'gcp', 'azure')`). This is fragile — adding a new cloud provider (e.g., DigitalOcean, bare metal) requires hunting down every conditional across multiple files.

## Problem

In `kubernetes_target.jac`, `monitoring.jac`, and related files, provider-specific behavior is determined inline:

```jac
service_type = 'LoadBalancer' if cluster_env in ('aws', 'gcp', 'azure') else 'NodePort';
if cluster_env == 'local' { check_node_ports_available(...); }
if cluster_env == 'aws' { ... }
```

Every new provider means touching multiple files and every existing conditional.

## Proposed Solution — Strategy Pattern

Define a `ClusterProvider` base class with provider-specific subclasses:

```jac
obj ClusterProvider {
    def service_type(self) -> str;
    def validate_ports(self, node_port: int, ..., namespace: str) -> None;
    def get_service_url(self, ...) -> str;
}

obj AWSProvider(ClusterProvider) { ... }
obj GCPProvider(ClusterProvider) { ... }
obj AzureProvider(ClusterProvider) { ... }
obj LocalProvider(ClusterProvider) { ... }
```

`cluster_type()` returns a `ClusterProvider` instance. The deploy method becomes:

```jac
provider = cluster_type();
service_type = provider.service_type();
provider.validate_ports(node_port=..., current_namespace=namespace);
```

## Benefits

- Adding a new provider = adding one new class, zero changes to deploy logic
- No scattered `if cluster_env == 'x'` checks
- Each provider's behavior is encapsulated and independently testable
- Type-safe — invalid provider strings are impossible at the call site

## Context

Tracked as a follow-up to PR #4686 (NodePort conflict detection), which introduced `cluster_env in ('aws', 'gcp', 'azure')` as an acceptable interim step.

PR #5840, opened 2026-05-05
jac-scale: Refactor cluster provider detection to Strategy pattern

juzailmlwork:
Summary
The current cluster detection logic in jac-scale uses string comparisons scattered across the codebase (cluster_env == 'aws', cluster_env in ('aws', 'gcp', 'azure')). This is fragile — adding a new cloud provider (e.g., DigitalOcean, bare metal) requires hunting down every conditional across multiple files.

Problem
In kubernetes_target.jac, monitoring.jac, and related files, provider-specific behavior is determined inline:

service_type = 'LoadBalancer' if cluster_env in ('aws', 'gcp', 'azure') else 'NodePort';
if cluster_env == 'local' { check_node_ports_available(...); }
if cluster_env == 'aws' { ... }
Every new provider means touching multiple files and every existing conditional.

Proposed Solution — Strategy Pattern
Define a ClusterProvider base class with provider-specific subclasses:

obj ClusterProvider {
    def service_type(self) -> str;
    def validate_ports(self, node_port: int, ..., namespace: str) -> None;
    def get_service_url(self, ...) -> str;
}

obj AWSProvider(ClusterProvider) { ... }
obj GCPProvider(ClusterProvider) { ... }
obj AzureProvider(ClusterProvider) { ... }
obj LocalProvider(ClusterProvider) { ... }
cluster_type() returns a ClusterProvider instance. The deploy method becomes:

provider = cluster_type();
service_type = provider.service_type();
provider.validate_ports(node_port=..., current_namespace=namespace);
Benefits
Adding a new provider = adding one new class, zero changes to deploy logic
No scattered if cluster_env == 'x' checks
Each provider's behavior is encapsulated and independently testable
Type-safe — invalid provider strings are impossible at the call site


juzailmlwork:
https://github.com/jaseci-labs/jaseci/issues/4773

jac-scale/jac_scale/targets/kubernetes/utils/kubernetes_utils.impl.jac
```diff
@@ -531,6 +536,44 @@ impl cluster_type -> str {
...
+                )
+            ) {
+                return AWSProvider();
+            }
+        }
+        return LocalProvider();
+    } except Exception {
+        return LocalProvider();
+    }
+}
```
udithishanka, line 500: get_cluster_provider duplicates the detection logic from cluster_type sitting right above it. after this PR cluster_type has no callers in jac-scale anymore (only test_k8s_utils.jac still references it). if EKS labels or provider-id conventions ever change, two functions need updating. either delete cluster_type entirely and migrate its test, or implement get_cluster_provider in terms of cluster_type so there's one source of truth for AWS detection.

jac-scale/jac_scale/targets/kubernetes/cluster_provider.jac
```diff
@@ -0,0 +1,183 @@
...
+        return True;
+    }
+}
+
+obj LocalProvider(ClusterProvider) {
+    def service_type -> str {
+        return 'NodePort';
+    }
+
+    def validate_ports(node_port: int, current_namespace: str) {
```
udithishanka, line 108: this is a verbatim copy of check_node_ports_available in kubernetes_utils.impl.jac:460, which is no longer imported anywhere after this PR. delete the standalone function, or have LocalProvider.validate_ports call it. duplicating the body here means two places to update if the conflict-detection logic ever changes.

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
```diff
@@ -2163,17 +2147,12 @@ obj KubernetesTarget(DeploymentTarget) {
...
-                # Resolve the ingress controller's LoadBalancer endpoint
-                endpoint = self._get_lb_endpoint(app_name);
-                if endpoint {
-                    return f"http://{endpoint}";
-                }
-            } else {
-                return f"http://localhost:{self.k8s_config.ingress_node_port}";
+            provider = get_cluster_provider();
+            url = provider.service_url(
+                self._get_lb_endpoint(app_name), self.k8s_config.ingress_node_port
```
udithishanka: this hits the K8s API (read_namespaced_service) on every get_service_url call, even on local clusters where LocalProvider.service_url ignores the lb_endpoint argument. the old code branched on cluster_env == 'aws' and skipped the lookup on local. easy fix is to gate it on provider.needs_nlb_wait(), or pass a callable so AWSProvider resolves the endpoint lazily.

jac-scale/jac_scale/targets/kubernetes/cluster_provider.jac
```diff
@@ -0,0 +1,183 @@
...
+import from jac_scale._optdeps.kubernetes { client }
+import from jac_scale.abstractions.logger { Logger }
+
+"""Abstract base class for cloud-provider-specific deployment behaviour.
+
+Subclass for each new provider (e.g. GCP, DigitalOcean). Return the
+appropriate instance from get_cluster_provider(). Callers never branch on
+provider type — all conditional logic lives in the subclass.
+"""
+obj ClusterProvider {
```
udithishanka, line 11: no tests added for any of this. the strategy classes and get_cluster_provider are easy to test in isolation, and test_k8s_utils.jac already mocks client.CoreV1Api so the pattern is right there. would want to see at minimum: service_type / service_url / prometheus_scrape_port / needs_nlb_wait return values per provider, LocalProvider.validate_ports raising on conflict and AWSProvider.validate_ports being a no-op, and get_cluster_provider returning the right instance for fake AWS vs local nodes.

udithishanka (changes requested):
4 things to address. solid refactor, but cluster_type and check_node_ports_available are now duplicates of the new code, get_service_url makes an unnecessary K8s API call on local clusters, and there's no test coverage for the new providers.

jac-scale/jac_scale/targets/kubernetes/utils/kubernetes_utils.impl.jac
juzailmlwork: this is fixed in commented called "addressed first 3 comments"

jac-scale/jac_scale/targets/kubernetes/cluster_provider.jac
juzailmlwork: this is fixed in commented called "addressed first 3 comments"

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
juzailmlwork: this is fixed in commented called "addressed first 3 comments"

jac-scale/jac_scale/targets/kubernetes/cluster_provider.jac
juzailmlwork: Test case is added

udithishanka (approved):

PR #5846, opened 2026-05-06
feat(jac-scale): K-track v1 — Kubernetes microservice deployment plane

MusabMahmoodh:
# K-track v1 — Kubernetes microservice deployment plane

## TL;DR

`jac start --scale` now deploys a Jac microservice app (gateway + N
peer services) to Kubernetes without the user writing Dockerfiles,
pushing images, authoring manifests, or wiring peer URLs. Single
source-of-truth in `jac.toml`. Same source runs as monolith or as
microservices — flip one config flag.

Validated end-to-end on **both** local Kubernetes (k3d on EC2) and
real cloud Kubernetes (EKS), against `examples/micr-s-example`
fixtures (CI) and **jacBuilder** (jaseci-labs/jacBuilder, real
third-party app) including SSO login, walker calls, sandbox spawn,
real LLM chat over sv-to-sv RPC, and a shared workspace mounted
across pods over EFS.

### Auto-build pipeline (P0.5–P0.13, P0.15)

Cluster-type detection, Dockerfile resolution chain (user override →
dev-checkout → shipped → embedded fallback), FE bundle build before
docker, content-addressed image tags so rebuilds trigger automatic
rolling updates, post-build assertion that `dist/index.html` made it
into the image, `jac install` inside the image to resolve user
`[dependencies]`, honest install log distinguishing dev-checkout vs
PyPI source.

### Microservice mode in K8s (K1–K10)

Each route in `[plugins.scale.microservices.routes]` becomes its own
Deployment + ClusterIP Service + HPA + PodDisruptionBudget; gateway
fronts everything with `sessionAffinity: ClientIP`. Built-in
endpoint passthrough (`/user`, `/walker`, `/function`, `/ws`, `/docs`,
etc.). Standardized error envelope across every failure path. Drain
on SIGTERM with `Retry-After: 2` for rolling deploys. Aggregated
`/openapi.json` + Swagger UI across services. Prometheus metrics on
`/metrics`. `X-Trace-Id` propagation through every sv-to-sv hop.
Per-service RPC timeout via `[microservices.services.<name>].rpc_timeout`.
HTTP/SSE/WebSocket all proxy through gateway transparently.

### `[plugins.scale.kubernetes]` config flow (P0.20)

Resource limits and probe cadences from the config block actually
reach the manifest. Priority chain: per-service > kubernetes block >
P0.8b Burstable QoS floor. Previously the kubernetes-block values
were silently dropped; jacBuilder hit OOMKilled in production
because of this.

### Cross-service shared volumes (P0.25)

`[[plugins.scale.microservices.shared_volumes]]` array-of-tables.
PVC mode (cloud K8s, RWX storage classes — EFS, Filestore, Azure
Files) + hostPath mode (k3d/kind dev). Apps that share filesystem
state across pods (jacBuilder workspace) declare once; every named
service gets the mount.

### K8s Secrets injection (P4.1)

`[plugins.scale.secrets]` flows through jaclang core's
`${ENV_VAR}` interpolation into a managed `<app>-secrets` K8s Secret
mounted on every pod via `envFrom: secretRef`. Applied before
Deployments so pods don't crash-loop with "secret not found".

### Pre-bound ServiceAccount (P0.30)

`[plugins.scale.kubernetes].service_account_name` injects
`serviceAccountName` on every microservice + gateway pod spec. Apps
that hit the cluster API at runtime (sandbox-spawning,
operator-style controllers) point at a pre-bound SA. K-track does
not create the SA — auto-create + bind is in the post-PR roadmap
(P4.3).

### sv-to-sv peer URL injection (P0.32)

Mirrors `LocalDeployer.process_manager`'s pattern: every microservice
+ gateway pod gets `JAC_SV_<PEER>_URL=http://<peer>-service.<ns>.svc.cluster.local:<port>`
auto-injected for every other peer in the routes table. Without this,
`sv import` silently fell back to local Python imports when peer
source was bundled in the image — making the microservice split
theatre at runtime. Empirically: `jac_coder_sv` pod served zero
non-healthz traffic before this fix; after, real `LiteLLM completion()`
calls land on the dedicated coder pod end-to-end.

### EFS UID-mismatch workaround

`git config --system --add safe.directory '*'` baked into all three
shipped Dockerfile sources (Dockerfile.microservice,
Dockerfile.microservice.ci, embedded fallback in `_image_build.jac`).
EFS CSI access points enforce a non-root POSIX UID; without this,
in-pod `git` against a shared volume tripped CVE-2022-24765
dubious-ownership and aborted. Proper fix via securityContext
config (P0.31) is post-PR.

## Tests

| Suite | New tests | Total |
|-------|-----------|-------|
| `test_k8s_microservice_target.jac` | +41 (P0.20, P0.25, P4.1, P0.30, P0.32) | 123 |
| `test_image_build.jac` | +9 (P0.10–P0.13, P0.15) | 33 |
| `test_gateway.jac` | +2 (P0.23 SPA root) | 64 |

`examples/micr-s-example/test_e2e.sh` (CI) covers the full local-mode
microservice stack with 54 e2e checks. K-track-specific manifest
generation is fully unit-tested; behavior on a real cluster is
covered by `scripts/k8s_microservice_real_e2e.sh` against minikube/k3d.

## Verification on real cloud K8s

Validated on EKS (`jaseci-cluster`, `us-east-2`) with **jacBuilder**:

- Public URL via ALB ingress → gateway → routed peer services
- SSO login (Google + GitHub) flowing through P4.1 secrets
- Walker calls + microservice routing
- Sandbox spawn via P0.30 pre-bound SA
- **Real LLM chat with sv-to-sv RPC**: `jac_coder_sv` pod served
  `LiteLLM completion()` calls (proves P0.32 + microservice split is
  actually crossing pod boundaries)
- Shared workspace `/root/.jac-ide` mounted across `builder_sv` +
  `jac_coder_sv` over EFS PVC
- **Promise A — Independent HPA scaling**: `jac-coder-sv` 1→3→1 on
  its own CPU pressure cycle; `builder-sv` 3→2→1→2→3 on a different
  cycle, different times. Each service has its own HPA + own
  `min/max/cpu_target`.
- **Promise C — Blast-radius isolation**: `kubectl delete pod -l
  app=jac-coder-sv` mid-flight; `builder-sv` (3 replicas) and
  gateway stayed `1/1 Ready` throughout (zero restarts); follow-up
  chat returned 200 in 0.64s — routing recovered seamlessly through
  surviving replicas while the new pod warmed up.

## Migration / breaking changes

None. K-track is opt-in via `[plugins.scale.microservices].enabled = true`
in `jac.toml`. Apps using the existing monolith K8s target (`--target
kubernetes` without microservices enabled) keep working unchanged.

The probe-cadence defaults in `KubernetesConfig` were tightened
(P0.20 — readiness/liveness `initial_delay` + `period` reduced to
match the K-track drain semantics). Apps with explicit
`[plugins.scale.kubernetes].liveness_initial_delay` etc. continue to
override; only apps relying on defaults see the change. The new
defaults are tested and verified end-to-end (jacBuilder runs
healthy on them).

## Deferred to follow-ups (post-PR)

Logged in `ROADMAP.md`, not blocking this PR:

- **P4.2** NetworkPolicies between services
- **P4.3** Auto-create + bind RBAC for sandbox-spawning apps (today
  users provide a pre-bound SA via `service_account_name`)
- **P0.31** `securityContext` config (`fs_group` / `run_as_user`) —
  proper UID-match instead of today's safe.directory bypass
- **P0.16** Per-service `pip_dependencies` for real microservice
  isolation (heavy LLM libs only on the coder service, etc.)
- **P0.17** `jac scale doctor` — aggregated diagnostics
- **P0.18** Post-apply health window in `jac start --scale`
- **P0.19** `jac scale init-cluster` — k3d + metrics-server bootstrap
- **P0.21** Factory-driven `config_section` dispatch (let external
  plugins register new target types without `plugin.jac` edits)
- **P0.24** Default-on Ingress
- **Service-lifecycle CLI**: `jac scale deploy <service>` (push one
  service without rebuilding all), `jac scale rollback`,
  `jac scale smoke-test`, `--dry-run`
- **Manifest naming prefix `<app>-`** so multiple K-track apps
  coexist in one namespace
- **No-registry runtime-clone mode** for parity with parent target's
  `--experimental` mode
- **Walker support for sv-import HTTP stub** (sv-import currently
  generates dispatch stubs for `def:pub` only)
- **PyPI publish** — jac-scale is ~170 commits ahead of last release
- **jaclang upstream PR** — silent-swallow of missing env vars in
  `_interpolate_recursive`

## Reviewer notes

- The branch is **commit-by-commit** (not squashed). Each P0.X has
  its own self-contained commit + tests; reviewers can read in
  chronological order to track the architecture evolution.
- The ~30 commits cluster naturally: K1–K10 (microservice manifest
  shapes), P0.5–P0.13 (auto-build pipeline), P0.15–P0.32 (config
  flow + sv-to-sv routing + shared volumes + secrets + SA + safe
  directory).
- All changes are isolated to `jac-scale/`. No jaclang core changes.
- The ECMAScript codegen path (used by `.cl.jac` client modules) is
  untouched.

## Use it

Minimal `jac.toml` for a microservice K8s deploy:

```toml
[plugins.scale.kubernetes]
app_name = "myapp"
namespace = "default"
memory_limit = "4Gi"
service_account_name = "myapp-sa"   # P0.30 — pre-bound SA
image_registry = "${ECR_REGISTRY}"   # required for remote clusters

[plugins.scale.microservices]
enabled = true

[plugins.scale.microservices.routes]
orders_sv = "/api/orders"
billing_sv = "/api/billing"

[plugins.scale.microservices.client]
entry = "main.jac"

[[plugins.scale.microservices.shared_volumes]]   # P0.25
name = "workspace"
mount_path = "/data"
services = ["orders_sv", "billing_sv"]
size = "10Gi"
access_mode = "ReadWriteMany"
storage_class = "efs-sc"

[plugins.scale.secrets]                          # P4.1
OPENAI_API_KEY = "${OPENAI_API_KEY}"
```

Then:

```bash
jac start --scale main.jac
```

That's it. The user never touches Docker or kubectl.


jac-scale/jac_scale/plugin.jac
```diff
@@ -723,6 +798,44 @@ class JacScalePlugin {
...
+                        }
+                    } except Exception {
+                        namespace = "default";
+                    }
+                }
+                k8s_name = k8s_safe_name(module_name);
+                dns_url = (
+                    f"http://{k8s_name}-service.{namespace}"
+                    f".svc.cluster.local:{port}"
+                );
```
udithishanka, line 791: this hand-builds the peer DNS URL with an f-string and re-derives namespace via POD_NAMESPACE env then in-cluster file fallback then 'default'. The same logic already exists in `k8s_deployer.build_k8s_registry` (k8s_deployer.jac:149-172) and `_util.k8s_service_dns_url` (_util.jac:83). Three copies of the URL format and two copies of the namespace lookup, so the DNS suffix or `<svc>-service` convention can drift across files silently. Pull a single-service helper out of build_k8s_registry and call it from both sites.

jac-scale/jac_scale/microservices/k8s_target.jac
```diff
@@ -0,0 +1,1878 @@
...
+                    "app": "jac-scale-user-secrets",
+                    "managed": "jac-scale",
+                    "jac-scale.role": "user-secrets"
+                }
+            },
+            "type": "Opaque",
+            # stringData (vs data) so we don't have to base64-encode
+            # values on the client - K8s accepts plaintext here and
+            # encodes server-side. Identical to parent's create_k8s_secret.
+            "stringData": dict(self.secrets)
```
udithishanka: no value coercion here. `self.secrets` comes from `get_secrets_config()` which passes the TOML loader's output through, so a user who writes `MY_PORT = 5432` under `[plugins.scale.secrets]` ends up with a non-string value in stringData. K8s rejects that with 422 mid-apply, leaving a half-applied bundle (Deployments + Services already created, Secret missing).

```suggestion
            "stringData": {k: str(v) for (k, v) in self.secrets.items()}
```

jac-scale/jac_scale/microservices/k8s_target.jac
```diff
@@ -0,0 +1,1878 @@
...
+    PVCs are left as-is. Users who need to resize/rebind delete the
+    PVC manually (or use `jac start --scale --destroy`).
+    """
+    def _apply_or_replace_pvc(core_v1: Any, namespace: str, manifest: dict[str, Any]) {
+        name: str = manifest["metadata"]["name"];
+        try {
+            core_v1.read_namespaced_persistent_volume_claim(
+                name=name, namespace=namespace
+            );
+            # Already exists - skip the patch. See docstring above.
```
udithishanka, line 455: skip is correct (most PVC fields are immutable post-bind) but it's silent. User who edits `size` or `storage_class` in jac.toml expecting it to take effect gets no signal that the change was ignored. Add a `self.logger.info(...)` here naming the PVC and saying delete-and-redeploy is needed to resize, so the no-op is at least visible in deploy logs.

udithishanka (changes requested):
3 things to fix.

jac-scale/jac_scale/plugin.jac
MusabMahmoodh: Good catch. Refactored in 7764ed2a110d41b1a9ec1b72a045fc4f47f8ae07:

- Extracted `in_cluster_namespace()` into `_util.jac` — single canonical impl (POD_NAMESPACE env → in-cluster secret file → "default").
- Refactored both `plugin.jac:ensure_sv_service` and `k8s_deployer.build_k8s_registry` to call `_util.k8s_service_dns_url()` + `_util.in_cluster_namespace()`.
- Three copies of the URL format / two copies of the namespace lookup → one canonical impl, two thin callers. DNS suffix / `<svc>-service` convention now lives in exactly one place.

jac-scale/jac_scale/microservices/k8s_target.jac
MusabMahmoodh: Applied the suggestion verbatim in 7764ed2a110d41b1a9ec1b72a045fc4f47f8ae07, plus a comment block calling out the failure mode (TOML int/float/bool → 422 → half-applied bundle → crash-loop on "secret not found") so the next reader sees why the coercion is load-bearing. `${ENV_VAR}` paths always returned strings so we hadn't tripped it empirically, but a user literally writing `MAX_RETRIES = 3` under `[plugins.scale.secrets]` would have.

jac-scale/jac_scale/microservices/k8s_target.jac
MusabMahmoodh: Added in 7764ed2a110d41b1a9ec1b72a045fc4f47f8ae07:

```
self.logger.info(
    f"PVC '{name}' already exists; skipping patch (PVCs are mostly "
    f"immutable post-bind). To apply size / storage-class changes, "
    f"delete the PVC manually first or run \`jac start --scale --destroy\`."
);
```

No-op is now at least visible in deploy logs. Guarded on `self.logger` since some unit-test paths construct the target without a logger.

jac-scale/jac_scale/microservices/orchestrator.jac
```diff
@@ -137,12 +178,10 @@ def start_microservice_mode(ms_config: dict[str, Any]) -> None {
     }
 
     # 4. Print startup banner
-    gateway_port: int = int(ms_config.get("gateway_port", 8000));
-    gateway_host: str = str(ms_config.get("gateway_host", "0.0.0.0"));
+    gateway_port: int = int(ms_config.get("gateway_port", DEFAULT_GATEWAY_PORT));
```
udithishanka, line 188: the 20s health-wait at process_manager.impl.jac:208 is too tight.

First-run jac compile of the service file plus plugin init (warmpool prefill, byllm import, sqlite create_all, MCP register) easily crosses 30-40s on the first pod start before bytecode is cached.

Hit this on jacBuilder's dev pod: both builder_sv and jac_coder_sv got marked UNHEALTHY at startup even though they came up healthy ~25s later. From the pod logs:

```
WARNING - builder_sv did not become healthy within 20s; proceeding
WARNING - builder_sv is UNHEALTHY (3 consecutive failures)
... 15 consecutive failures ...
WARNING - Proceeding with unhealthy services...
```

meanwhile direct curl :18032/healthz from inside the pod returned 200 throughout.

make the timeout configurable per-service, same shape as rpc_timeout. in process_manager.impl.jac:208:

```
# instead of: timeout=20.0
startup_timeout = float(getattr(self, "config", {}).get(name, {}).get("startup_timeout", 60.0));
ok = wait_for_health(entry.url, lambda : proc.poll() is None, timeout=startup_timeout);
```

and read [plugins.scale.microservices.services.<name>].startup_timeout the same way rpc_timeout is already plumbed. bumping the default to 60s would also have saved jacBuilder from this hit.

jac-scale/jac_scale/microservices/orchestrator.jac
```diff
@@ -137,12 +178,10 @@ def start_microservice_mode(ms_config: dict[str, Any]) -> None {
...
+    gateway_host: str = str(ms_config.get("gateway_host", DEFAULT_GATEWAY_HOST));
```
udithishanka, line 189: no recovery from initial UNHEALTHY state.

once the wait loop above (lines 148-178) exits with services marked UNHEALTHY, there's no recovery path. check_all_health is only called from this loop and from LocalDeployer.status (on-demand, never periodic). the gateway's handle_proxy (gateway.impl.jac:33) and handle_builtin_passthrough (gateway.impl.jac:412) both short-circuit on status not in (HEALTHY, STARTING), so a service that crossed the 20s wait_for_health bar by a few seconds gets locked out for the entire process lifetime.

symptom on jacBuilder's dev pod: services answer 200 to direct :18032/healthz continuously, but every gateway-routed walker returns 404 "Not found on any service" (unprefixed) or 503 "Service unhealthy" (prefixed). only kubectl rollout restart recovers it.

cleanest fix is a small background thread that calls pm.check_all_health() every N seconds and lets check_all_health flip UNHEALTHY back to HEALTHY when the probe recovers. it already does the right thing on success: resets _failure_counts[name] = 0 and sets status HEALTHY, so the recovery is one extra goroutine away. alternative is a lazy on-demand probe in handle_proxy / handle_builtin_passthrough, on a non-HEALTHY entry do a single inline healthz check before short-circuiting (mirrors the trust-but-verify of _mark_all_healthy_k8s for the K8s path), but that adds first-request latency. either way the gateway shouldn't permanently brick itself on a 20-second startup race.

udithishanka (changes requested):
2 things to fix. Gateway health monitor bricks itself when a service is slow to start at boot. Hit this on jacBuilder's dev pod just now: services came up healthy ~25s in, but the gateway permanently marked them UNHEALTHY at the 20s wait, and every walker call after that returned 404/503 for the whole pod lifetime. Only `kubectl rollout restart` recovers.

Both issues anchored on the new `gateway_port` line for visibility; the actual lines that need fixing are noted in the comment bodies (`process_manager.impl.jac:208` and the wait loop above this point at `orchestrator.jac:148-178`).

udithishanka (approved):

Issue #5669, opened 2026-04-22
topology_index updates on root anchor are silently dropped by hash-gated sync (regression from #5663)


udithishanka:
## Summary

After PR #5663 (defer `SqliteMemory`/`MongoBackend` `put()` to `sync()`), updates to `NodeAnchor.topology_index_data` are silently lost because `Serializer._compute_hash` explicitly excludes `topology_index_data` from the hash. The deferred-sync path uses hash comparison to decide which anchors to flush, so a root anchor whose only in-memory change is a topology-index refresh is treated as unchanged and never written back to storage.

End-user symptom: graph queries that go through `plan_query()` (any typed filter like `[source -->(?:SomeNodeType)]`) return empty results as soon as a node of a type that wasn't in the index at walker-entry time is added in a subsequent walker, because `plan_query` resolves against the stale on-disk index.

## Reproduction (observed on jac-builder-dev running jac-scale 0.2.14 / jaclang 0.14.0 from main)

1. New user registers → pod creates root + `UserProfile`. Root's `topology_index` ends up containing `{Root, UserProfile, GenericEdge}` and this state is persisted (presumably because at this point `edges` on the root changed, so hash changed for another reason and sync wrote the full anchor, including the index blob).
2. User creates a Project → walker runs `profile ++> Project(...)`. `build_edge` calls `on_edge_created`, which calls `root.set_topology_index(idx)` with the Project/edge added.
3. Walker commits. `MongoBackend.sync` iterates L1. For the root anchor:
   - `_compute_hash(root)` serializes the root then `data.pop('topology_index_data', None)` (see [`serializer.impl.jac:702`](https://github.com/jaseci-labs/jaseci/blob/main/jac/jaclang/runtimelib/impl/serializer.impl.jac#L702)).
   - Root's `edges` list is unchanged this walker (edges were added to `UserProfile`, not root).
   - Archetype scalar fields on root are unchanged.
   - Resulting hash == `anchor.hash` → `continue` → root is **not** written.
4. Project node + edge are new anchors without a load snapshot, so their sync goes through `_write_to_db` and lands correctly.
5. MongoDB state: `Project` node present, edge present, `UserProfile.edges` updated via `_put_node_atomic`, **but `root.topology_index_data` still missing Project**.
6. Any later walker that runs `[-->(?:Project)]` calls `plan_query` → loads the stale index from root → returns empty set → runtime reports zero results even though the edges and targets exist.

[...]

PR #5855, opened 2026-05-07
fix(runtimelib): prevent silent topology index drops in deferred sync

akindu-k:
## Problem

After the deferred sync implementation (PR #5663), updates to `NodeAnchor.topology_index_data` were silently dropped and never persisted to the database.

The root cause: `Serializer._compute_hash()` deliberately excludes `topology_index_data` from the hash computation, it was treated as a derived cache to avoid serialization overhead on every sync. Both `SqliteMemory.sync` and `MongoBackend.sync` use hash comparison to decide whether an anchor needs writing. When a walker **only** updates the topology index, no edge mutations, no archetype field changes, the hash before and after is identical, so sync skips the write entirely and the index update is lost.

### Reproduction scenario

1. **Walker 1**: creates root + `UserProfile` node; index persists because the new edges also change the anchor hash.
2. **Walker 2**: adds a `Project` node and calls `root.set_topology_index(updated_idx)`. This is the **only** change on the root anchor.
3. During sync, root's hash matches its pre-change value (topology excluded) → write skipped → `Project` type never added to the persisted index.
4. Subsequent typed queries like `[root -->(?:Project)]` return empty results even though the node exists in the DB.

---

## Fix: Explicit `is_topology_dirty` flag

Preserves the hash-exclusion performance intent while making the dirty signal explicit.

### `NodeAnchor`: new field (`archetype.jac`)

Added `is_topology_dirty: bool = False` to `NodeAnchor`. The field is naturally transient, the serializer only writes `topology_index_data` explicitly and never touches `is_topology_dirty`, so it defaults to `False` on every deserialization and never round-trips through SQLite or MongoDB.

### `set_topology_index`: content-aware dirty marking (`archetype.impl.jac`)

```jac
impl NodeAnchor.set_topology_index(index: Any) -> None {
    new_blob = index.encode();
    if new_blob != self.topology_index_data {
        self.topology_index_data = new_blob;
        self.is_topology_dirty = True;
    }
}
```

Only raises the flag when the encoded content actually changes, preventing false positives from no-op calls.

### `SqliteMemory.sync`: topology merge in stored-anchor path (`memory.impl.jac`)

Added a check before the `if changed` write block in the existing-anchor branch:

```jac
if isinstance(anchor, NodeAnchor) and getattr(anchor, 'is_topology_dirty', False) {
    stored_anchor.topology_index_data = anchor.topology_index_data;
    changed = True;
}
```

The flag is reset to `False` after every successful write, both the existing-anchor merge path and the new-insert path.

### `MongoBackend.sync`: bypass hash early-out and force full write on dirty topology (`memory_hierarchy.mongo.impl.jac`)

Two changes:

**1. Early-out guard**: passes through when `is_topology_dirty` even if hash matches:

```jac
topology_dirty = isinstance(anchor, NodeAnchor) and getattr(anchor, 'is_topology_dirty', False);
if current_hash == anchor.hash and not topology_dirty {
    continue;
}
```

**2. Skip `put_partial` when topology is dirty**: `topology_index_data` lives at `data.topology_index_data` in the MongoDB document, outside `data.archetype.*` where `put_partial` writes. Falling through to `_write_to_db` (full upsert) ensures it is always written:

```jac
if anchor.hash != 0 and not topology_dirty {
    ...
    if dirty {
        self.put_partial(anchor, dirty);
        continue;
    }
}
# Full upsert — first insert, no field changes, or topology is dirty
self._write_to_db(anchor);
if isinstance(anchor, NodeAnchor) {
    anchor.is_topology_dirty = False;
}
```

The flag is also reset in the `_put_node_atomic` path (atomic edge-merge), since that path serializes the full anchor including topology.

---

## Files Changed

| File | Change |
|------|--------|
| `jac/jaclang/jac0core/archetype.jac` | Add `is_topology_dirty: bool = False` to `NodeAnchor` |
| `jac/jaclang/jac0core/impl/archetype.impl.jac` | Update `set_topology_index` to set dirty flag on content change |
| `jac/jaclang/runtimelib/impl/memory.impl.jac` | SQLite sync: merge topology when dirty, reset flag after write |
| `jac-scale/jac_scale/impl/memory_hierarchy.mongo.impl.jac` | MongoDB sync: bypass hash early-out, skip `put_partial`, reset flag |
| `jac/tests/runtimelib/test_topology_integration.jac` | 3 new regression tests |

---

## Tests Added

All in `jac/tests/runtimelib/test_topology_integration.jac`:

- **`is_topology_dirty set on content change cleared on sync`** — verifies the flag is raised by `set_topology_index` and cleared to `False` after `sync()`.
- **`is_topology_dirty not set when content unchanged`** — verifies that calling `set_topology_index` with identical content does not re-raise the flag.
- **`topology-only update persisted in sqlite after hash unchanged`** — the exact regression scenario: Request 1 creates root + `UserProfile`, Request 2 adds `Project` as the *only* change (hash stays the same), Request 3 loads from DB and asserts `Project` is present in the persisted index.

All 276 existing tests continue to pass.

## Related Issue
#5669


akindu-k:
resolved by PR https://github.com/jaseci-labs/jaseci/pull/6388

PR #5859, opened 2026-05-07
feat(compiler)!: strict Any gradual semantics in Jac modules

marsninja:
## Summary

In `.jac` files, `Any` can no longer be silently assigned to a declared non-`Any`, non-`object` destination. The dispatch axis is the file flavor of the AST node where the assignment happens (via `Module.is_raised_from_py`), not where the `Any` came from — so a leak that crosses the Python → Jac boundary gets caught at the Jac side, where the user is in control. Python modules (`.py`/`.pyi`) keep PEP 484 permissive semantics.

This is Layer 2 of the gradual-typing tightening. Layer 1 — opt-in typed walker `has reports: list[T]` — is in #5857; together they make `report` channels and consumer code share a single typed contract end-to-end.

## Why

`Walker.reports: list[Any]` and other `Any`-typed runtime APIs were silently widening typed locals: `tasks: list[Task]` could become `list[Any]` after a comprehension that mixed in a reported value, and the type checker accepted it. Stricter-than-Python typing was a stated goal; PEP 484's bidirectional `Any` compatibility was the load-bearing wall keeping Jac as permissive as Python. This PR replaces the Any short-circuit with a file-flavor-aware rule.

## Semantics

- `Any → Any`, `Any → object`, `Any → TypeVar`: **always allowed** (these dests legitimately accept anything).
- `Any → declared concrete T` (Task, list[Task], MyProtocol, …): **error** in `.jac` modules.
- `T → Any`: always allowed (the dest opted into permissive output).
- Inferred destinations (no annotation): no strict check — the user's escape valve. `data = py_call();` keeps `data: any` and propagates only as far as the user lets it.
- Recurses into containers, callables, and unions: `list[any] → list[Task]` errors via element-wise recursion.

## Coverage

The check fires at every "destination has a declared type" site:

| Site | File | Notes |
|---|---|---|
| Annotated assignment | `exit_assignment` | When LHS is annotated, not on inferred re-assignment |
| Has-var initializer | `exit_has_var` | Always declared |
| Function arg | `parameter_type_check.validate_arg_types` | Determined by the param's declared type |
| Return | `exit_return_stmt` | When function has a declared return type |
| Yield | `exit_yield_expr` | Vs. expected yield type |
| Edge connection assign | `operations.jac` | `a ++>:Edge:val=expr` against typed field |

## Migration

For code that hits the new error, two paths:

1. **Fix the source** (preferred): type the runtime API or library boundary so it doesn't return `Any`. For walker reports, see #5857. For Python interop, write/improve `.pyi` stubs.
2. **Use the escape valve**: drop the annotation (`x = py_call()`) or annotate the destination as `any` explicitly (`x: any = py_call()`). This documents the boundary clearly.

## Test plan

- [x] New fixture `checker_strict_any_jac.jac` exercises annotated assignment, container nesting, function arg, and return — exactly 4 strict-Any errors expected.
- [x] New test `strict_any_in_jac_modules` asserts the 4 errors and no spurious ones.
- [x] Two pre-existing fixtures updated to match the new semantics:
  - `checker_jac_any_keyword_annotation.jac` switched `from_any: str = passthrough(1)` to inferred destination.
  - `checker_list_indexing.jac` uses `list[list[int]]` instead of bare `list`.
- [x] Full suites green: 185 checker + 17 checker_gaps + 88 language + 455 (typed_interop / micro_suite / checker_shortcomings) = **745 tests**.
- [x] `jac/examples/day_planner/walkers/components/TasksPanel.cl.jac` now correctly errors on the original motivating case (`Cannot assign list[Any | Task] to list[Task]`).

## Sequencing note

This PR depends on #5857 for clean migration: without typed walker `has reports`, every walker-using example will fire strict errors with no easy fix beyond annotating destinations as `any`. Recommend landing #5857 first, then this one.


RagavRida:
@marsninja — looks like #5857 just merged! 🎉 The branch still has conflicts with `main` that are blocking CI. Once you rebase/merge `main` into this branch, the checks should be able to run cleanly.

PR #5865, opened 2026-05-07
fix(bundle): rename `root` local var to avoid Jac keyword conflict

SupulHeshan:
## **Description**
`root` is a reserved keyword in Jac (it refers to the graph root node). Four `impl` files in the bundle pipeline were using it as a plain local variable, which caused `jac bundle` to fail at import time with "No bytecode found" because the compiler produced no bytecode for the affected modules.

### Changes

- Renamed root → `proj_root` in `builder.impl.jac`, `metadata.impl.jac`, `sources.impl.jac`, and `_collect_from_dir's` parameter
- Fixed `set_config` signature (`JacConfig` → `JacConfig | None`) to match its actual usage and the `_config` global's declared type
- Added `tests/publish/test_bundle_cli.jac` - four CLI-level tests for `jac bundle` covering the success path, custom output dir.


kugesan1105 (changes requested):
add a `cli test` calling `jac bundle`

SupulHeshan:
> add a `cli test` calling `jac bundle`

yes. I refactored the test cases by spawning subprocess run.

kugesan1105 (approved):

PR #5883, opened 2026-05-09
fix(jac-scale): scalability fixes — connection pools, Redis batch/TTL, JWT, MongoDB CAS

### LINKED ISSUE #5874: arch(jac-scale): master scalability audit — Blocking IO, Memory Leak, Race Conditions, Async Support
## Overview

This is a full architectural audit of the `jac-scale` serving layer on `main`, organized into four themes. Each finding includes the file:line, the observed behaviour, and the concrete impact.

---

# 1. Blocking IO

Every route callback in jac-scale is `async def`, running on uvicorn's single event loop thread. All IO drivers are synchronous. Any blocking call inside an `async def` function pins the event loop for its duration — no other coroutine, connection, or health-check can run until it returns.

### 1a. Two new TCP connections opened per request (availability probes)

**`memory_hierarchy.main.impl.jac:13,23` | `memory_hierarchy.mongo.impl.jac:28–43` | `memory_hierarchy.redis.impl.jac:24–40`**

`request_context_middleware` creates a fresh `ScaleTieredMemory` for every request. Its `postinit` unconditionally calls `RedisBackend.is_available()` and `MongoBackend.is_available()`, each of which opens a brand-new client, sends a blocking `PING`, then discards the client:

```python
# MongoBackend.is_available()
client = MongoClient(self.mongo_url)                # TCP handshake — blocking
client.admin.command('ping', maxTimeMS=100)          # network call — blocking
client.close()

# RedisBackend.is_available()
client = redis.from_url(self.redis_url)             # TCP handshake — blocking
client.ping()                                       # network call — blocking
client.close()
```

At 200 RPS: 400 new TCP connections opened and closed per second before any user code runs.

### 1b. System root fetched from MongoDB on every request

**`context.impl.jac:14` | `memory_hierarchy.mongo.impl.jac:59`**

`JScaleExecutionContext.init` constructs `ScaleTieredMemory()` with an empty `__mem__ = {}`. The next line reads the system root:

```python
system_root_anchor = self.mem.get(UUID(Con.SUPER_ROOT_UUID))
# L1 is empty → falls through to:
db_obj = self.collection.find_one({'_id': str(_id)})   # MongoDB, every request
```

The system root is written once at boot and never mutated. This is a guaranteed MongoDB `find_one` per request for a value that never changes.

### 1c. Full per-request blocking IO chain (authenticated walker, 1 dirty anchor)

All of the following are synchronous calls inside `async def` functions on the event loop:

| # | Operation | Location |
|---|-----------|----------|
| 1 | Redis `PING` (availability probe) | `memory_hierarchy.redis.impl.jac:28` |
| 2 | MongoDB `ping` (availability probe) | `memory_hierarchy.mongo.impl.jac:35` |
| 3 | MongoDB `find_one` — system root | `memory_hierarchy.mongo.impl.jac:59` |
| 4 | MongoDB `find_one` — `user_exists` during JWT validation | `user_manager.impl.jac:175` |
| 5 | MongoDB `find_one` — `get_root_id` | `server.impl.jac:264` |
| 6 | MongoDB `find_one` — `set_user_root → mem.get` | `context.impl.jac:48` |
| 7 | MongoDB `update_one` — `Jac.commit()` inside `spawn_walker` | `server.impl.jac:306` |
| 8 | MongoDB `update_one` + Redis `SET` — `ctx.mem.commit()` in middleware | `jfast_api.impl.jac:868` |

Minimum 7 blocking network calls per request. At 5ms average each: the event loop is blocked for ≥35ms per request, capping throughput at roughly 28 req/s before any walker code runs.

### 1d. `MongoBackend.sync()` issues N individual `update_one` calls; `bulk_write` exists but is never called

**`memory_hierarchy.mongo.impl.jac` — `sync()` vs `bulk_put()`**

The commit path iterates L1 and writes one anchor at a time:

```python
# sync() — N sequential MongoDB round-trips
for (_, anchor) in list(self._mem_ref.items()):
    if current_hash == anchor.hash:
        continue
    self.put(anchor)    # → collection.update_one(...)
```

`bulk_put()` already exists in the same file and uses `collection.bulk_write(ops)` for a single round-trip. `sync()` never calls it. A walker writing 10 nodes: 10 × `update_one` (~50ms) vs 1 × `bulk_write` (~5ms).

### 1e. `RedisBackend.batch_get` issues N individual `GET` calls; no `MGET` pipeline

**`memory_hierarchy.redis.impl.jac:167–174`**

```python
for id in ids:
    if (anchor := self.get(id)):   # one round-trip per ID
        result[id] = anchor
```

`redis_client.mget([...])` would fetch all N keys in a single round-trip. At N=20 nodes: 20× the latency of the batch path.

### 1f. byllm uses the synchronous OpenAI client and `requests.post`

**`jac-byllm/byllm/llm.impl/model.impl.jac:80,118,149`**

```python
response = client.chat.completions.create(**params)   # sync openai client — blocks
# ...
response = requests.post(...)                          # sync requests — blocks
```

An LLM call typically takes 0.5–30 seconds. When called from any walker via `by_llm`, the event loop is pinned for the full LLM latency. At 10 concurrent users each triggering an LLM call: requests queue serially — user 10 waits 10× the LLM latency before their response even starts.

### 1g. Double commit per request — dirty anchors serialised and written twice

**`server.impl.jac:306` | `jfast_api.impl.jac:868`**

`spawn_walker` calls `Jac.commit()` (line 306). The middleware then calls `ctx.mem.commit()` again after the handler returns (line 868). In the clean case the second pass is a no-op (hash match), but `sync()` still iterates the entire L1 dict and recomputes hashes for every anchor — O(graph-size) CPU work on the event loop, twice per request.

---

# 2. Memory Leaks

### 2a. L2 Redis: default TTL is 0 (no expiry) — cache grows without bound

**`memory_hierarchy.redis.impl.jac:71–79`**

```python
ttl = db_config.get('redis_default_ttl', 0)
if ttl > 0:
    self.redis_client.setex(key, ttl, json.dumps(data))
else:
    self.redis_client.set(key, json.dumps(data))    # ← no TTL
```

With the default config every anchor ever loaded or written is cached in Redis permanently. As the graph grows, Redis memory grows monotonically. There is no eviction policy set by the application. Redis will either OOM or begin evicting based on server-side `maxmemory-policy` (default: `noeviction` — which returns errors instead of evicting).

### 2b. Redis connection pool is unbounded

**`memory_hierarchy.redis.impl.jac:10` | `memory_hierarchy.redis.impl.jac:28`**

```python
self.redis_client = redis.from_url(self.redis_url)   # no max_connections
```

`redis.from_url` with no `max_connections` creates an unbounded pool. Under high concurrency the pool grows without limit, eventually exhausting file descriptors. Each discarded `is_available()` client (Finding 1a) also creates a pool and never reuses it.

### 2c. Scheduled-task execution context is never freed

**`scheduler.impl.jac:338–363` | `jac0core/impl/runtime.impl.jac:909–912`**

HTTP requests use `request_context_middleware` which creates a fresh `JScaleExecutionContext` per request and frees it at request end via `clear_request_context → ctx.close() → mem.close() → __mem__.clear()`. Scheduled jobs bypass this path entirely.

`_execute_task` calls `spawn_walker` which calls `Jac.get_context()`. Since no request context is set, this falls through to `JacRuntime.exec_ctx` — the global process singleton. The singleton's `__mem__` dict accumulates every anchor loaded or created by every scheduled job, forever. There is no cleanup, no close, no eviction.

---

# 3. Race Conditions

### 3a. Read-modify-write with full-document replace — no optimistic locking

**`memory_hierarchy.mongo.impl.jac:78–85` | `jac/jaclang/runtimelib/impl/memory.impl.jac`**

Each request gets a private L1 `__mem__` dict. At commit time the full serialized anchor (including its complete edge list) is written unconditionally:

```python
# MongoDB path
self.collection.update_one(
    {'_id': str(_id)},
    {'$set': {'data': data, 'type': type(anchor).__name__}},
    upsert=True          # ← unconditional full-document replace
)

# SQLite path
INSERT OR REPLACE INTO anchors (id, data) VALUES (?, ?)   -- unconditional replace
```

There is no version field, no conditional write, and no conflict detection at either layer.

**Classic race:**

```
Request A: loads Root → edges = [e1, e2]
Request B: loads Root → edges = [e1, e2]   ← same stale snapshot

Request A: Root ++> NodeX  → edges = [e1, e2, NodeX]
Request B: Root.updated_at = now

Request A commits → DB has [e1, e2, NodeX]  ✓
Request B commits → DB has [e1, e2]          ✗  NodeX silently erased
```

This affects every shared node touched by concurrent requests. The shared `Root` and `UserProfile` nodes are the most common victims because every walker for the same user traverses them.

### 3b. Scheduled jobs share the global execution context — concurrent jobs race on the same L1

**`scheduler.impl.jac:338` | `jac0core/impl/runtime.impl.jac:909`**

As noted in 2c, scheduled jobs all share `JacRuntime.exec_ctx`. Two scheduled jobs running concurrently (APScheduler's ThreadPoolExecutor has 10 workers by default) read from and write to the same `__mem__` dict with no locking. This is the same read-modify-write race as 3a, but between scheduled jobs rather than HTTP requests, and with no isolation boundary at all.

### 3c. Scheduler posts coroutines onto the uvicorn event loop — no priority separation

**`scheduler.impl.jac:356–358`**

```python
loop = asyncio.get_event_loop()
if loop.is_running():
    asyncio.ensure_future(_run())    # fires on uvicorn's loop
```

Scheduled walkers compete with incoming HTTP requests on the same event loop. A long-running scheduled job (e.g., a batch processing walker) can starve HTTP request handlers entirely. In Python 3.12, `asyncio.get_event_loop()` from a non-async thread raises `DeprecationWarning` and may not return the uvicorn loop, causing `asyncio.run(_run())` to fire instead — behaviour is Python-version-dependent.

### 3d. JWT validation hits MongoDB on every request — no token cache

**`user_manager.impl.jac:173–175`**

```python
decoded = jwt.decode(token, JWT_SECRET, algorithms=[JWT_ALGORITHM])
username = decoded['username']
return username if self.user_exists(username) else None   # DB call every request
```

A JWT is cryptographically self-validating. The `user_exists()` call adds a MongoDB `find_one` to every authenticated request. At 500 RPS this is 500 auth queries/second. There is no cache; every token re-validates from scratch on every call.

---

# 4. Path to Proper Async Support

The root cause of section 1 is that all IO drivers are synchronous while all route handlers are `async def`. The correct fix is end-to-end async IO. The required driver replacements:

| Current (sync) | Async replacement | Notes |
|---|---|---|
| `pymongo.MongoClient` | `motor.motor_asyncio.AsyncIOMotorClient` | Drop-in for most operations |
| `redis.Redis` (redis-py sync) | `redis.asyncio.Redis` (redis-py ≥ 4.2) | Same package, async submodule |
| `sqlite3` | `aiosqlite` | Wraps sqlite3 in a thread |
| `openai.OpenAI` (sync) | `openai.AsyncOpenAI` | Same SDK, async variant |
| `requests.post` (byllm HTTP path) | `httpx.AsyncClient` | Maintained, async-native |

With async drivers in place:
- All `find_one`, `update_one`, `bulk_write`, `get`, `set` calls become `await`-able
- Route handlers stay `async def` and compose correctly
- No thread offloading needed; the event loop thread is never blocked
- `spawn_walker` and `execute_function` remain `async def`; all current `await` call sites are correct

Intermediate mitigation (without driver changes): wrap each `spawn_walker` call in `await asyncio.to_thread(...)` so blocking IO runs in a thread pool worker and the event loop stays free. This requires the thread pool to be sized appropriately and does not fix the race conditions.

---

## Summary

| Theme | Finding | Severity |
|-------|---------|----------|
| Blocking IO | 2 TCP pings per request (availability probes) | Critical |
| Blocking IO | System root fetched from MongoDB every request | Critical |
| Blocking IO | 7+ blocking calls per request on event loop | Critical |
| Blocking IO | `sync()` N × `update_one` instead of `bulk_write` | High |
| Blocking IO | `batch_get` N × Redis GET instead of MGET | High |
| Blocking IO | byllm uses sync OpenAI client + `requests` | Critical |
| Blocking IO | Double commit per request | Medium |
| Memory Leak | L2 Redis: no TTL, unbounded growth | High |
| Memory Leak | Redis connection pool: no `max_connections` | Medium |
| Memory Leak | Scheduled-task global context never freed | High |
| Race Condition | Full-document replace with no optimistic locking | Critical |
| Race Condition | Scheduled jobs share global L1 context | Critical |
| Race Condition | Scheduled jobs post onto uvicorn event loop | High |
| Race Condition | JWT re-validates from MongoDB every request | High |
| Async Support | All drivers are synchronous | Root cause |

_Audit performed on `main` branch, 2026-05-08._

asimsalim:
Closes #5874

## Summary

Seven targeted fixes for the scalability issues identified in the audit, plus proof tests for each.

- **Connection pooling**: `MongoClient` and Redis clients are now process-level singletons in `_process_cache`; `is_available()` result is also cached so it doesn't re-probe on every request
- **Commit dispatch**: moved `commit()` from `ExecutionManager` teardown to `ExecutionHandler` so each request owns its write phase
- **Scheduler isolation**: each scheduled job gets its own `ExecutionContext` lifecycle instead of sharing one across jobs
- **Batch anchor loading**: `batch_get()` uses `MGET` for Redis (single round-trip) and `$in` query for MongoDB; `ScaleTieredMemory` prefetches a request's full anchor set from L2/L3 before walker dispatch
- **Redis TTL**: default TTL set to 3600s and `max_connections` capped, so Redis keys have bounded lifetime instead of growing forever
- **JWT hot path**: `validate_jwt_token` no longer calls `user_exists()` (a MongoDB `find_one`) on every authenticated request — JWT signature + expiry is sufficient; `get_root_id` gains a 300s process-level LRU cache
- **MongoDB CAS**: `_hash` field stored alongside each anchor document; writes use compare-and-swap (`{_id, _hash}` filter) to detect concurrent modifications instead of last-writer-wins; `bulk_write(ordered=False)` prevents a single conflict from aborting a batch

**Bug fixed during testing**: `anchor.hash` defaults to `0` (int), not `None`. The initial CAS guard used `if old_hash is not None`, which treated every brand-new anchor as existing, fired a `{'_hash': 0}` filter that matched nothing, and silently dropped the first MongoDB write — causing 500s on all walker calls. Fixed to `if old_hash` (truthy), consistent with `bulk_put`.

## Proof tests (new)

Each fix has a corresponding test in `jac-scale/jac_scale/tests/`:

| Test file | What it proves |
|-----------|----------------|
| `test_connection_overhead.jac` | Connection pool reuse; system root fetch overhead |
| `test_redis_batch_overhead.jac` | MGET replaces N individual GETs (ratio ≈ 1.0 after fix) |
| `test_redis_memory_leak.jac` | All Redis keys have TTL after fix |
| `test_jwt_validation_overhead.jac` | JWT validation no longer grows MongoDB query count |
| `test_persistence_race.jac` | MongoDB CAS prevents last-writer-wins data loss |
| `test_async_io_blocking.jac` | Async IO blocking characterisation |
| `test_bulk_write_overhead.jac` | Bulk write overhead characterisation |

Tests written as "bug documentation" fail once the fix is applied (the assertion proves the bug was present); fix-validation variants pass.

## Known remaining items

- **System root caching**: `test_system_root_fetched_from_mongodb_on_every_request_cold_l1` still shows 2.0 queries/request. System root needs a process-level singleton in `_process_cache` (deferred — touches context init path).
- **SQLite CAS**: `test_sqlite_backend_concurrent_writes_to_same_root_must_not_lose_edges` — SQLite `INSERT OR REPLACE` still overwrites concurrent writes.
- **Async IO drivers**: all MongoDB/Redis IO is still synchronous (blocking the event loop). Full fix requires Motor + aioredis; tracked separately.


asimsalim:
Superseded by 5 focused individual PRs — each a single, reviewable fix:

- #5899 — MongoDB process-level connection pool + `is_available` stale-cache fix
- #5900 — Redis process-level pool + MGET `batch_get` + TTL default 3600s
- #5901 — `ScaleTieredMemory.batch_get` full L1 → L2 → L3 read-through
- #5902 — Remove redundant `user_exists()` DB call from JWT validation
- #5903 — Isolated `ExecutionContext` per scheduled job

Closing this in favour of the split PRs.

PR #5888, opened 2026-05-09
Preserve TLS state on Ingress redeploys in jac-scale

### LINKED ISSUE #5834: jac-scale: redeploy wipes TLS configuration from Ingress, breaking HTTPS
## Summary

Running `jac start <file> --scale` on an app that already has TLS enabled silently wipes the TLS configuration from its Ingress, leaving the app reachable on HTTP only and serving the controller's default self-signed cert on HTTPS (browsers see `ERR_CERT_AUTHORITY_INVALID`). The cert-manager `Certificate` and TLS secret remain intact, but they are no longer referenced by the Ingress.

This was hit in production: two apps (`maizemind-juncture/jaseci-ingress`, `scoochai/jaseci-ingress`) lost their TLS spec after a redeploy. Both Ingresses were at generation 5/6 with valid `Ready=True` certs in the namespace but no `spec.tls` block and no host on the rules.

## Root cause

`jac-scale/jac_scale/targets/kubernetes/ingress.jac` has two separate code paths for the Ingress:

1. `_deploy_ingress_resource` (called on every `jac start --scale` run) builds the Ingress body with **no host and no `tls` block** by design (line 528: `# Always deploy with a wildcard rule (no host).`), then writes it via:
   ```jac
   networking_v1.replace_namespaced_ingress(
       name=ingress_name, namespace=namespace, body=main_ingress
   );
   ```
   `replace_namespaced_ingress` is a full-object PUT, not a strategic merge.

2. `enable_tls` (called once via `--enable-tls`) patches in:
   - `cert-manager.io/issuer: letsencrypt-prod` annotation
   - `ssl-redirect` / `force-ssl-redirect` annotations
   - `spec.rules[*].host = domain`
   - `spec.tls = [{hosts: [domain], secretName: ...}]`

Because `_deploy_ingress_resource` overwrites the entire body, every subsequent `jac start --scale` run after `--enable-tls` strips out the host, the TLS spec, and the TLS-related annotations. The `jac-scale/domain` and `jac-scale/cert-email` annotations are preserved (they are re-applied unconditionally), so the metadata looks half-correct, masking the issue at a glance.

## Reproduction

```bash
# 1. Initial deploy
jac start main.jac --scale --target kubernetes

# 2. Enable TLS (cert issued, ingress patched, HTTPS works)
jac start main.jac --scale --target kubernetes --enable-tls

# 3. Make any change and redeploy (image bump, HPA tweak, anything)
jac start main.jac --scale --target kubernetes

# Result: HTTPS to the domain returns ERR_CERT_AUTHORITY_INVALID
# kubectl get ingress -n <ns> -o yaml shows no spec.tls and no host on rules
```

## Impact

- Production users hit certificate errors with no clear signal in `jac start --scale` output.
- Cert-manager `Certificate` resources stay `Ready=True` and the TLS secret is still present, so the issue is invisible from the cert-manager side.
- Recovery requires either a manual `kubectl patch` to re-add the TLS spec (what we did to unblock prod) or re-running `--enable-tls` after every redeploy (fragile).

## Proposed fix

Make `_deploy_ingress_resource` preserve TLS state when an existing Ingress already has it. Two reasonable options:

**Option A (preferred):** Read the existing Ingress before building the new body. If `spec.tls` is non-empty or rules have a `host`, copy those fields and the TLS-related annotations (`cert-manager.io/issuer`, `ssl-redirect`, `force-ssl-redirect`) into the new body before calling `replace_namespaced_ingress`.

**Option B:** Switch from `replace_namespaced_ingress` to a strategic-merge patch that only updates the fields jac-scale owns (annotations it sets, rules paths, ingress class), leaving TLS spec and host alone.

Option A keeps the current single-step write and minimises behavioural change. The same fix should apply to the RedisInsight Ingress branch (line 693), which has the identical pattern.

## Workaround

For affected ingresses, patch the TLS spec back manually:

\`\`\`bash
kubectl patch ingress <name> -n <ns> --type=merge -p '{
  "metadata":{"annotations":{
    "cert-manager.io/issuer":"letsencrypt-prod",
    "nginx.ingress.kubernetes.io/ssl-redirect":"true",
    "nginx.ingress.kubernetes.io/force-ssl-redirect":"true"
  }},
  "spec":{
    "rules":[{"host":"<domain>","http":{...existing paths...}}],
    "tls":[{"hosts":["<domain>"],"secretName":"<existing-secret>"}]
  }
}'
\`\`\`

udithishanka:
## Summary

Fixes #5834.

`_deploy_ingress_resource` was rebuilding the Ingress body with no host and no `spec.tls`, then writing it via `replace_namespaced_ingress` (a full-object PUT). Every `jac start --scale` run after `--enable-tls` silently wiped the TLS spec, host, and cert-manager / ssl-redirect annotations, leaving the app on HTTP only and serving the controller's default self-signed cert on HTTPS (browsers saw `ERR_CERT_AUTHORITY_INVALID`).

Hit in production: `maizemind-juncture/jaseci-ingress` and `scoochai/jaseci-ingress` both lost their TLS spec after a redeploy. cert-manager `Certificate` and TLS secret were intact, just no longer referenced by the Ingress.

## Fix

New `_preserve_tls_state` helper: read the live Ingress before each `replace_namespaced_ingress` call. If the live object has `spec.tls`, copy:

- `spec.tls`
- `host` from the existing rules onto every rule we're writing
- `cert-manager.io/issuer`, `nginx.ingress.kubernetes.io/ssl-redirect`, `nginx.ingress.kubernetes.io/force-ssl-redirect` annotations

When the live Ingress has no TLS (or doesn't exist yet) the helper returns the body unchanged, so the existing "wildcard rule until `--enable-tls`" first-deploy path is preserved.

Applied to both the main Ingress and the RedisInsight Ingress write paths.

## Test plan

- [x] Manual: production `scoochai/jaseci-ingress` was patched out-of-band to restore TLS; follow-up redeploys with this fix in place no longer overwrite that state
- [ ] Add unit coverage for `_preserve_tls_state` (404, no-tls, with-tls cases)

### DIFF
--- a/docs/docs/community/release_notes/unreleased/jac-scale/5888.bugfix.md
+++ b/docs/docs/community/release_notes/unreleased/jac-scale/5888.bugfix.md
@@ -1,0 +1,1 @@
+- **Fix: Preserve TLS state on Ingress redeploys**: `jac start --scale` no longer wipes the host, `spec.tls`, and cert-manager / ssl-redirect annotations from an existing Ingress, so HTTPS keeps working after a redeploy.
--- a/jac-scale/jac_scale/targets/kubernetes/ingress.jac
+++ b/jac-scale/jac_scale/targets/kubernetes/ingress.jac
@@ -461,6 +461,79 @@ obj IngressDeployer {
         }
     }
 
+    """Copy TLS state (spec.tls, rule hosts, TLS-related annotations) from a live
+    Ingress into a freshly built body before a full-object replace.
+
+    `replace_namespaced_ingress` is a PUT that overwrites everything jac-scale
+    does not re-include in the body. enable_tls writes spec.tls, sets host on
+    every rule, and adds cert-manager / ssl-redirect annotations; without this
+    helper, every subsequent `jac start --scale` strips them. See issue #5834.
+
+    Returns new_body unchanged if the Ingress doesn't exist yet, or if it has
+    no TLS configured (so the original "wildcard rule until enable_tls" path
+    is preserved on first deploy).
+    """
+    def _preserve_tls_state(
+        ingress_name: str, namespace: str, new_body: dict, networking_v1: Any
+    ) -> dict {
+        try {
+            existing = networking_v1.read_namespaced_ingress(
+                name=ingress_name, namespace=namespace
+            );
+        } except ApiException as e {
+            if e.status == 404 {
+                return new_body;
+            }
+            raise;
+        }
+
+        api_client = client.ApiClient();
+        existing_dict = api_client.sanitize_for_serialization(existing);
+        existing_spec = existing_dict.get('spec', {}) or {};
+        existing_tls = existing_spec.get('tls', []) or [];
+
+        if not existing_tls {
+            return new_body;
+        }
+
+        existing_host = '';
+        for rule in (existing_spec.get('rules', []) or []) {
+            host = rule.get('host', '') or '';
+            if host {
+                existing_host = host;
+                break;
+            }
+        }
+
+        new_body['spec']['tls'] = existing_tls;
+        if existing_host {
+            for rule in (new_body['spec'].get('rules', []) or []) {
+                rule['host'] = existing_host;
+            }
+        }
+
+        existing_ann = (existing_dict.get('metadata', {}) or {}).get('annotations', {})
+        or {};
+        new_ann = new_body['metadata'].get('annotations', {}) or {};
+        for k in [
+            'cert-manager.io/issuer',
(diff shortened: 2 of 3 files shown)


juzailmlwork:
@udithishanka the pr looks good but in project deployment I think we have 4 projects who are using a domain.plz fix them also by manual intervention

juzailmlwork:
@udithishanka Also before merging this make sure you test it with a new project with maybe a subdomain of jaseci

PR #5917, opened 2026-05-11
feat(jac-scale): MongoDB server-side type/field query pushdown with indexes

asimsalim:
## Summary

- `MongoBackend.query()` gains an optional `mongo_filter` dict pushed to `collection.find()` before the Python post-filter, eliminating full collection scans on filtered queries
- New `query_by_type(arch_type_name, mongo_filter)` pushes both the archetype type and optional field predicates (e.g. `{'age': {'\': 28}}`) to MongoDB; field names auto-prefixed to `data.archetype.*`
- `get_roots()` replaces the `query()` full scan + Python `isinstance` loop with `collection.find({'arch_type': 'Root', 'type': 'NodeAnchor'})` directly against the index
- `ensure_indexes()` (called in `postinit`, idempotent) creates a compound `(arch_type, type)` index and an `updated_at` descending index so all type-based queries use IXSCAN instead of COLLSCAN

## Test plan

- [ ] `test_mongo_query_pushdown.jac` — 4 tests using testcontainers MongoDB:
  1. `ensure_indexes` creates `arch_type_1_type_1` compound index; `explain()` confirms IXSCAN
  2. `query_by_type('_QUser')` returns only `_QUser` anchors, not `_QProduct`
  3. `query_by_type('_QUser', {'age': {'$lt': 23}})` returns only matching subset
  4. `get_roots` returns only Root anchors; `explain()` confirms IXSCAN 

---
**Closing:** Superseded by #6073 which delivered the QueryPlan IR + capability-driven planner. The `ensure_indexes` / compound-index work from this PR was preserved verbatim in #6073. The `query_by_type` API is replaced by `execute_plan(QueryPlan(...))` — PR #5920 has been updated to use the new IR.

PR #5920, opened 2026-05-11
feat(compiler): FilterSpec storage pushdown for [?:Type, field op const] filters

asimsalim:
## Why

Edge traversal filters like \`[?:User, age < 28]\` currently compile to a Python lambda evaluated against every node after loading from storage - a full scan disguised as graph traversal.

The existing \`[?:Type, conditions]\` syntax already carries everything needed for a server-side query: the type name and simple field comparisons. This PR teaches the compiler to extract those predicates at compile time and route them through the **QueryPlan IR** (#6073), so \`node --> [?:User, age < 28]\` becomes a single \`execute_plan()\` call on the backend rather than load-all + filter-in-Python.

Supersedes the \`query_by_type\` duck-type approach from #5917 (now closed). Builds directly on the \`capabilities()\` / \`execute_plan()\` contract introduced in #6073.

### Compiler (\`jac0core\`)
- Emits \`FilterSpec(type_name, predicates, func)\` instead of a plain lambda when all predicates in a \`[?:Type, field op const]\` filter are simple field-op-constant comparisons. Predicate dict uses MongoDB-style \`$op\` notation (\`{'age': {'$lt': 28}}\`) which every backend translates natively.
- Type-only filters (\`[?:User]\`) keep the plain lambda unchanged - the \`TopologyIndex\` fast-path is unaffected.

### Runtime (\`refs()\` in \`jac0core\`)
- Detects a \`FilterSpec\` on a single-hop traversal.
- Checks \`mem.capabilities()\` for \`'field_pushdown'\` or \`'type_pushdown'\` (capability-declared, not duck-typed).
- Builds \`QueryPlan(node_type_final, field_predicates, id_in, slc)\`:
  - \`id_in\`: neighbour IDs from origin-anchor edge lists, pushed when \`'id_in'\` is declared.
  - \`slc\`: pushed to plan when \`'slice'\` is declared; applied Python-side otherwise.
- Calls \`mem.execute_plan(plan)\`; if it returns \`None\` (backend opts out), falls through to the fused-lambda / topology-index path.
- Python lambda in \`FilterSpec\` is still applied as a safety post-filter on every returned anchor.

### Backends
- \`MongoBackend\` (via #6073) already declares \`{'type_pushdown', 'field_pushdown', 'id_in', 'slice'}\` and implements \`execute_plan\` - the fast path activates automatically.
- Future backends (Postgres jsonb GIN, SQLite) just declare capabilities and implement \`execute_plan\` - no further runtime changes needed.

## Relationship to #6073

#6073 delivered the QueryPlan IR + capability planner. This PR wires the compiler's FilterSpec output into that IR so predicate filters from Jac syntax reach the backend in one round-trip.

## Test plan

- [x] \`test_filterspec_emitted_for_predicate_filters_not_for_type_only\` - compiler emits FilterSpec for predicate filters, plain lambda for type-only; 12/12 pyast gen pass tests green
- [x] \`jac check\` + \`jac format --check --lintfix\` clean on all modified files
- [ ] End-to-end: \`node --> [?:User, age < 28]\` against a real MongoBackend (testcontainers) - follow-up
- [ ] Verify \`execute_plan\` opts-out path (None return) falls through correctly to topology-index


asimsalim:
Converting to draft pending @marsninja's plugin hook spec for OSP filter rewriting.

The FilterSpec emitted by the compiler currently carries MongoDB-specific `$op` predicates. Jason's proposal is a backend-agnostic `FilterPredicate` that each backend plugin can translate natively (MongoDB → `collection.find()`, SQLite → `WHERE` clause), with a Python-lambda fallback for backends that don't implement the hook.

Will rebase/rewrite the compiler emission once the hook API is specced out.

asimsalim:
Closing in favour of #6079 — same FilterSpec compiler changes but correctly routes pushdown through `runtimelib/planner.impl.jac` (`_build_plan_from_path`) instead of `runtime.impl.jac`. No jac-core changes in the new PR.

PR #5922, opened 2026-05-11
feat(mobile): native Android CI/CD pipeline (migrated from EAS)

mahfoos:
## Summary

 ## What "Mobile Jac" is now                                                                                                                                                                                
                                                                                                                                                                                                             
  A Jac mobile app is a **React Native + Expo** shell wrapping a **WebView** that loads the Jac web bundle. The native shell is identical across projects; only the embedded web app changes.
  
  - **Native compile on the runner.** Pipeline installs JDK 17, Android SDK, Node 20, Yarn, jaclang, jac-client. Runs the same gradle command you'd run locally.                                             
  - **Two signing modes, auto-selected.** Real keystore from `ANDROID_KEYSTORE_BASE64` + 3 password/alias secrets when present; throwaway keystore generated by `keytool` otherwise. Both produce installable
   APKs; only the real keystore is Play-Store-distributable.                                                                                                                                                 
  - **Backend URL injection.** Resolved by `_resolve_production_api_url(project_dir)`: `JAC_MOBILE_API_URL` env var → `[mobile.features].api_url` in `jac.toml` → empty. Written to                          
  `.jac/mobile/dev-config.json`, Metro statically bakes it into the APK's JS bundle, WebView reads it at startup, web app sees `window.JAC_BACKEND_URL` set correctly.
  - **Three local dev workflows.** `jac start --client mobile --dev` (HMR via Vite + backend + Expo Go), `jac start --client mobile` (bundled mode, no HMR), or full local APK build via gradle. All three   
  documented in `JAC_MOBILE.md`.              
                                                                                                                                                                                                             
  - Replace EAS Cloud–based mobile build with a native Android pipeline that compiles APKs on `ubuntu-latest` via Gradle, matching the desktop pipeline's "build on the runner" model.                     
  - Auto-keystore fallback: pipeline produces an installable, release-mode APK with zero secrets configured. Drop in four `ANDROID_*` secrets to upgrade to a stable Play-Store-ready signing identity.
  - Bake the backend URL into the APK at build time via `JAC_MOBILE_API_URL` env var or `[mobile.features].api_url` in `jac.toml`. CI exposes it as a `workflow_dispatch` input.
  - Remove all EAS code paths from the `jac-client` plugin.


mahfoos:
Here the builded APK : https://github.com/mahfoos/jaseci/actions/runs/25654098416

jac-client/jac_client/examples/all-in-one/mobile/app/index.tsx
```diff
@@ -0,0 +1,281 @@
...
+    }
+  }, []);
+
+  const prepareBundle = () => {
+    try {
+      if (!JAC_BUNDLE_HTML) {
+        throw new Error('Bundle not available');
+      }
+      // Inject the backend API URL into the bundle
+      const html = JAC_BUNDLE_HTML.replace(
```
MusabMahmoodh: hi, i think we have probelm here. this file is doing the replace with the literal string `window.JAC_BACKEND_URL = window.JAC_BACKEND_URL || 'http://localhost:9000';` no? but in `setup_helpers.jac` the generator is writting the new placeholder `%%__JAC_BACKEND_URL__%%`. So if someone is regenarating via `jac setup mobile` the two contracts wont match and backend url injection silently fails.

i tried to mentaly follow the flow and it confused me alot. can we pick one approach and use it everywhere? otherwise next person reading this will get same headach.

jac-client/jac_client/examples/all-in-one/mobile/assets/jac-app/bundle.ts
```diff
@@ -0,0 +1,43 @@
+// Auto-generated standalone Jac app bundle
```
MusabMahmoodh: this file is 700+kb of minified bundle no? isnt it auto-generated by `create-bundle.js` itself. same for `bundle.html`, `client.js`, `styles.css` - they all get rebuild during `jac build`.

i think these should go to `.gitignore`, otherwise every web bundle change is creating a huge diff on this example folder which is not really usefull for review.

jac-client/jac_client/plugin/src/targets/mobile/dev_helpers.jac
```diff
@@ -0,0 +1,157 @@
...
+
+"""Read api.base_url from jac.toml via JacClientConfig."""
+def _get_toml_api_base_url(project_dir: Path) -> str {
+    import from jac_client.plugin.src.config_loader { JacClientConfig }
+    config_loader = JacClientConfig(project_dir);
+    return config_loader.get_api_config().get('base_url', '');
+}
+
+"""Get local network IP using only local-only OS calls (no external network)."""
+def _get_local_ip -> str {
```
MusabMahmoodh, line 39: i checked this and on ubuntu/debian dev machines `/etc/hosts` typically maps hostname to `127.0.1.1`. this fails the `not ip.startswith('127.')` check and falls through to return `'127.0.0.1'` at the bottom. mobile device on same wifi cant reach loopback so dev workflow is broken for most linux folks.

the older udp-connect trick `socket.connect(('8.8.8.8', 80))` doesnt actually send any packet, kernel only picks the routable source ip - so the "no external network" worry is not really valid no? can we use that as primary method and keep getaddrinfo as fallback?

.github/workflows/build-mobile.yml
```diff
@@ -0,0 +1,244 @@
...
+          #   2. Switch the release buildType to use signingConfigs.release
+          #      instead of signingConfigs.debug (Expo's default).
+          python3 - <<'PYEOF'
+          from pathlib import Path
+          import re
+
+          p = Path('app/build.gradle')
+          t = p.read_text()
+
+          if 'MYAPP_RELEASE_STORE_FILE' in t:
```
MusabMahmoodh: this regex patching is little scary - if expo template changes thier `build.gradle` structure tommorow, the second `re.sub` falls back to the `NOTE` message and we silently sign the release apk with the *debug* keystore. that is bad surprise to find out later.

can the fallback be `raise SystemExit(...)` so signing-config drift fails the build loudy? or even better, write our own `signing.gradle` file fresh instead of patching upstream layout at all.

.github/workflows/build-mobile.yml
```diff
@@ -0,0 +1,244 @@
...
+
+      - name: Set up Node.js 20
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+
+      - name: Install Yarn
+        run: npm install -g yarn
+
+      - name: Install Python dependencies (jaclang + jac-client)
```
MusabMahmoodh, line 98: no caching on python/yarn/gradle here. each run does full `pip install` of jaclang+jac-client (editable), full `yarn install`, full gradle download. with `timeout-minutes: 60` this is bit wastefull.

`actions/setup-python` supports `cache: pip`, `setup-node` supports `cache: yarn`, `setup-java` supports `cache: gradle`. easy win, can probably half the build time.

jac-client/jac_client/plugin/src/targets/impl/mobile_target.impl.jac
```diff
@@ -0,0 +1,402 @@
...
+            "   Dependencies & bundles will be in .jac/mobile/ during build/dev",
+            style="muted"
+        );
+        console.print("\nNext steps:", style="bold");
+        console.print(
+            "  jac build main.jac --client mobile -p android -b development",
+            style="muted"
+        );
+        console.print("  jac start main.jac --client mobile --dev", style="muted");
+    } finally {
```
MusabMahmoodh: this `finally` block doesnt actually do anything usefull no? if the try body's `mkdir` succeded, dir is already exist. if it failed, the function has already raised before getting to here. can we just remove it? right now it just adds noise.

same pattern is repeated in the `dev()` and `start()` methods below also.

jac-client/jac_client/plugin/cli.jac
```diff
@@ -297,6 +322,67 @@ def _handle_build_target(ctx: HookContext) -> None {
...
+
+        console.print("\n📱 Mobile Build Configuration", style="bold");
+        console.print("  Platform options:", style="muted");
+        console.print("    • android  - Build for Android devices", style="muted");
+        console.print(
+            "    • ios      - Build for iOS devices (requires Mac)", style="muted"
+        );
+        console.print("    • all      - Build for both platforms", style="muted");
+
+        platform_choice = input(
```
MusabMahmoodh, line 391: cli accepts `--platform ios|all` and the interactive prompt offers same, but i couldnt find any ios-specific logic in `MobileTarget.build` - everything is android (gradle, keystore, `expo prebuild --platform android`).

readme says ios not implemented yet which is fine, but the cli should match - maybe gate the choices to android-only untill ios is ready? otherwise user picks `ios` and gets a confusing silent android build.

jac-client/jac_client/examples/all-in-one/mobile/assets/jac-app/create-bundle.js
```diff
@@ -0,0 +1,76 @@
...
+
+    <!-- Backend API URL - will be injected by mobile app -->
+    <script>
+        window.JAC_BACKEND_URL = window.JAC_BACKEND_URL || 'http://localhost:9000';
+    </script>
+
+    <!-- Jac client bundle as blob URL -->
+    <script>
+    (function() {
+        const jsCode = atob('${Buffer.from(js).toString('base64')}');
```
MusabMahmoodh: the comment says "satisfy CSP" but webview is loading from `html:` source - there is no CSP enforced unless page itself sets one (and it doesnt). the base64 + blob + `URL.createObjectURL` is adding ~33% size and extra latency for nothing. can we just inline the js directly in `<script type="module">${js}</script>` like we already do for the `<style>`?

jac-client/jac_client/plugin/src/targets/mobile/dev_helpers.jac
```diff
@@ -0,0 +1,157 @@
...
+        bundler = ViteBundler(
+            project_dir=project_dir, output_dir=None, minify=False, config_path=None
+        );
+        dev_config_path = bundler.create_vite_config(
+            entry_file,
+            api_base_url_override=f"http://localhost:{api_port}",
+            is_dev=True,
+            api_port=api_port
+        );
+        process = bundler.start_dev_server(port=5173, suppress_initial_output=True);
```
MusabMahmoodh: Vite port 5173 hardcoded here, backend port 9000 hardcoded in the webview path (`index.tsx`), but `MobileTarget.dev()` and `start()` both default `api_port: int = 8000`. so the default `jac start --client mobile --dev` will start backend on 8000 but the webview is hardcoded to look at 9000 - they wont talk to eachother unless user explicitly passes `--port 9000`.

can we atleast share one constant for the prod backend port, and make the default consistant?

jac-client/jac_client/plugin/src/targets/mobile/setup_helpers.jac
```diff
@@ -0,0 +1,572 @@
...
+        raise RuntimeError("yarn install timed out after 5 minutes");
+    } except FileNotFoundError {
+        raise RuntimeError("yarn not found. Install with: npm install -g yarn");
+    } except Exception as e {
+        raise RuntimeError(f"yarn install failed: {e}");
+    }
+}
+
+"""Generate placeholder PNG icons for Expo (icon, splash, adaptive-icon, favicon)."""
+def _generate_placeholder_icons(mobile_dir: Path) {
```
MusabMahmoodh, line 524: the fallback branch (no PIL) writes one giant `raw_data` as `b"\x00" + (b"\x4A\x90\xE2" * width)` and then multiplies by `height` inside `zlib.compress(...)` - but PNG format requires a *filter byte per scanline*, not just one at the start of the whole image. so the resulting png will not decode on any spec-compliant decoder, only the leftmost column of pixels (if any) renders.

can we either require pillow as a dep or vendor a small static 1x1 placeholder png in the package?

MusabMahmoodh (changes requested):
thanks for the work on this @mahfoos , the architecture is solid and `JAC_MOBILE.md` is realy thorough. left a bunch of inline comments. the big ones for me are :

- the committed example (`index.tsx` + `create-bundle.js`) is using the *old* `window.JAC_BACKEND_URL = ... || 'http://localhost:9000'` string-replace contract, but the generators in `setup_helpers.jac` are emiting the new `%%__JAC_BACKEND_URL__%%` placeholder. these two will drift and silently break the backend url injection.
- `bundle.ts` (700+kb) and friends are autogenerated build artifacts and shouldnt be commited.
- 3500 lines added with 0 tests, can we add atleast a few unit tests for `MobileConfig` and `_resolve_production_api_url` precedence?

the rest are smaller, see inline. happy to chat if anything is unclear.

MusabMahmoodh:
@mahfoos thanks for the quick turnaround on the resolved items 🙏 . before raising for re-review can we please go through this checklist - there are a few from the earlier review that didnt make it in, plus a bunch of new things that came up after a closer look at the design. trying to capture everthing in one place so we dont go back and forth.

[`pre-commit.ci - pr` is still red](https://results.pre-commit.ci/run/github/422711680/1778621017.cJEqLI8jQp2zSLMTVQAldA) - i think on the em-dash and the binary PNG items below.

### 🚫 CI / standards blockers

- [ ] **Em-dashes (`—`) in `.github/JAC_MOBILE.md`** trip the repo's `ban-em-dashes` hook (`types: [markdown]`). quick fix: `sed -i 's/—/-/g' .github/JAC_MOBILE.md`.
- [ ] **Binary PNGs cant be commited** without explicit allowlist - this PR adds 6 of them: `mobile/assets/{icon,splash,adaptive-icon,favicon}.png` and `pwa_icons/pwa-{192,512}x{192,512}.png`. the repo's `check-no-binary-files` hook only excludes a hand-maintained list and none of these are in it. since `_generate_placeholder_icons` already creates them at `jac setup mobile` time, kindly delete them from git instead of adding to the allowlist.
- [ ] **`build_type` / `-b` flag is dead code** - `MobileTarget.build` line 157, `WebTarget.build` line 111, `DesktopTarget.build`, `PWATarget.build` all accept `build_type: str = "production"` but **none of them reference it in the body**. the CLI registers `-b development|preview|production` with an interactive prompt, the CI workflow passes `-b preview`, and it has no effect anywhere - native gradle always runs `assembleRelease`. since this PR removes EAS, the EAS `--profile` concept doesnt apply either. please **drop the flag, drop the interactive prompt, drop the CI input**. one less thing to maintain.

### 🧪 Need before merge ready

- [ ] **Zero tests for 3500+ LOC.** atleast a few unit tests would be very helpfull:
  - `MobileConfig.get_app_json_config()` produces a valid expo config from a synthetic `[mobile]` table
  - `MobileConfig._is_valid_bundle_id` / `_generate_bundle_id` round-trip
  - `_resolve_production_api_url` env > toml > empty precedence
- [ ] **iOS still silently accepted but unimplemented** - cli accepts `--platform ios|all` and the interactive prompt offers same, but no iOS-specific code anywhere. user picks `ios`, gets a confusing silent android build. either gate the choices to android-only untill iOS lands or fail-fast with a clear message.

### 🏛️ Architecture concerns

these are not just nits, they shape how the next target (or the next person reading this code) will integrate. flagging them now becouse fixing them after merge means a breaking refactor.

- [ ] **Hardcoded `if target_name == "desktop" / elif target_name == "mobile"` branches in `cli.jac`** (`_handle_setup_target`, `_handle_start_target`) check for target-specific directories. this is the dispatcher knowing target internals - that knowledge belongs in the target class. proper fix: add a `target.setup_dir: Path` field or `target.verify_setup() -> bool` method, then the dispatcher is `if target.requires_setup and not target.verify_setup(): error`. polymorphic. **this PR makes the smell worse by adding a second target to the chain** - the next target makes it three.
- [ ] **`TargetConfig` refactor is half-done.** the new base class only covers `DesktopConfig` + `MobileConfig`. `WebTarget` and `PWATarget` are still on the old `JacClientConfig` / direct-load path. we now have **two parallel config-loading systems** in the same plugin. either pull web/pwa under `TargetConfig` too, or roll back the base class and just duplicate the small amount of shared code between desktop+mobile.
- [ ] **`MobileTarget(WebTarget)` is the wrong is-a relationship.** PWA is-a web target (it ships HTML+JS+manifest+SW). Mobile is **not** a web target - it *consumes* a web bundle and produces an entirely different artifact (APK). the inheritance forces awkward calls like `super.build(entry_file, project_dir, platform)` where `platform="android"` gets passed to `WebTarget.build` which has no idea what "android" means and silently ignores it ([mobile_target.impl.jac:171](jac-client/jac_client/plugin/src/targets/impl/mobile_target.impl.jac#L171)). composition is cleaner: `MobileTarget` *has-a* `WebTarget` and calls its `build()` explicitly with web-relevant args only.
- [ ] **Generated source files (TSX/JS) embedded as multi-line heredoc strings in `setup_helpers.jac`.** `_generate_webview_screen`, `_generate_bundle_script`, `_generate_layout_file` each hold 50-280 lines of TSX/JS as triple-quoted strings inside a `.jac` file. no syntax check, no formatter, no LSP, no diff readability, **and this is exactly why the committed example drifted from the generator** (issue #2). proper fix: extract to `plugin/src/targets/mobile/_templates/{index.tsx,_layout.tsx,create-bundle.js}` and `shutil.copy2` them. two lines.

### 👃 Code smells

- [ ] **`MobileTarget.setup()` uses `for (fn, label) in [(lambda: gen_a(...), "a.json"), ...] { try { fn() } except Exception as e { console.warning(...) } }`** - silently swallows errors from each generator. if `_generate_app_json_from_toml` fails, you get a yellow warning and setup continues, leaving a `mobile/` directory missing `app.json` that the user then has to debug. atleast aggregate failures and raise at the end, or fail-fast on first error.
- [ ] **`MobileConfig.validate()` defined but never called** anywhere in the codebase. dead method. either wire it into `MobileTarget.setup` / `MobileTarget.build` or remove it. same for the pre-existing `DesktopConfig.validate()`.
- [ ] **Defaults drift between `MobileConfig.get_default_config()` and `_add_mobile_config()`.** the in-memory default includes `expo.plugins: ['expo-router']` but the TOML written to `jac.toml` has no `expo.plugins` key. user who never edits jac.toml gets `expo.plugins` working; user who *does* edit it can loose the default. one source of truth please.
- [ ] **`_resolve_production_api_url` reaches into `features` dict directly** (`config_data.get('features', {}).get('api_url', '')`) instead of using the `MobileConfig.get_features_config()` accessor that was added for this purpose. two paths to the same data == easy place for skew.
- [ ] **Interactive `input()` prompts inside `_handle_build_target` hook** mix UI with the plugin-dispatch flow. hooks are supposed to be data-in / decision-out. if a user pipes `jac build --client mobile` from a script and forgets `-n`, the hook blocks on stdin forever. atleast `sys.stdin.isatty()` guard with a clear error message.

### 🧹 Cleanup / redundancy

- [ ] **Do we realy need TypeScript at all for the mobile shell?** `app/_layout.tsx` is 5 lines, `app/index.tsx` has no type annotations doing meaningful work, and `bundle.ts` is literally just `export const JAC_BUNDLE_HTML = \`...\`` - that is plain JS. dropping `tsconfig.json`, `@types/react`, `typescript` devDep removes a whole toolchain for very little. atleast `bundle.ts` → `bundle.js`, no question.
- [ ] **`mobile/.gitignore` has two sources of truth** - one committed at `examples/all-in-one/mobile/.gitignore`, and one written by `_generate_gitignore` in `setup_helpers.jac`. one of them please.
- [ ] **Stale backslash escape in committed `create-bundle.js`** - generator in `setup_helpers.jac` correctly escapes `\\` → `\\\\` before backticks and `$`, but the committed `examples/all-in-one/mobile/assets/jac-app/create-bundle.js` only escapes backticks and `$`. any web bundle containing a `\` will break the TS template literal in the committed version. please sync.
- [ ] **Base64 + Blob URL is unneccessary** - `create-bundle.js` does `atob('${Buffer.from(js).toString('base64')}')` + Blob + `URL.createObjectURL`. comment says "satisfy CSP" but the webview loads from `html:` source with no enforced CSP. base64 inflates the bundle by ~33% for no gain. just inline as `<script type="module">${js}</script>` like we already do for the `<style>`.
- [ ] **No `yarn.lock` committed** for the example app. every install can drift. either commit the lockfile (and switch CI to `yarn install --frozen-lockfile`) or document why we are accepting drift.

### 🩹 Workarounds (shortcuts that should be done properly)

- [ ] **`_create_mobile_bundle` shells out to `node create-bundle.js`** for a script that only does `fs.readFileSync` / `fs.writeFileSync` / regex match + template literal - operations python can do natively. spawning a Node subprocess (with 30s timeout, error handling, stdout capture, etc.) just to read 3 files and write 2 files is a workaround for "we wrote the bundler in JS". port it to a jac/python function in `bundle_helpers.jac` and delete the node dependency from this step entirely.
- [ ] **Subprocess cleanup uses `proc.terminate()` without process-group teardown.** expo and vite both spawn child processes (Metro bundler, watchman, esbuild worker). Ctrl+C terminates the immediate child but the grandchildren survive as orphans. on a dev's machine this means stale processes piling up across sessions. use `subprocess.Popen(..., start_new_session=True)` + `os.killpg(os.getpgid(proc.pid), signal.SIGTERM)` for a clean teardown.
- [ ] **CLI prompt offers `--platform all`** when only android is implemented, so "all" == "android". confusing UX that masks the missing iOS work. either remove `all` from the choices untill iOS lands or make `all` fail-fast with "iOS not implemented yet, use android".
- [ ] **`super.build(entry_file, project_dir, platform)` passes `"android"` / `"ios"` strings into `WebTarget.build`** which has no concept of mobile platforms (its `platform` choices are windows/macos/linux/all for desktop). WebTarget silently ignores it, so it "works" - but the type contract is being lied to. either document `platform` as target-specific values per-subclass, or split into `desktop_platform` / `mobile_platform` kwargs.

### 📦 Workflow hygiene

- [ ] **Remove `push: branches: [feature/mobile-cicd-pipeline]`** from `build-mobile.yml` - it was added for testing this branch. before merge it should be `pull_request` and `workflow_dispatch` only, otherwise nothing builds on `main`.
- [ ] **Cache keys hash `mobile/package.json`** instead of a lockfile. once `yarn.lock` is committed, switch to hashing it for accurate invalidation. comment in the workflow already acknowledges this.
- [ ] **PR description still references "EAS Cloud-based mobile build"** as present tense. since EAS is removed, kindly reword to past tense / "previously used EAS, now native".

### 🤔 Smaller items (can be follow-ups)

- [ ] **`_generate_placeholder_icons` shells out to `python3 -c`** even though the parent is already python (jaclang). kindly do the work inline.
- [ ] **`expo ~54.0.0`, `react-native 0.81.5`, `react-native-webview 13.15.0`** tightly pinned. add a note in `JAC_MOBILE.md` about which Expo SDK we are tracking and the expected bump cadence.

---

happy to chat through any of these if the framing isnt clear. the deeper concerns under **Architecture** (`TargetConfig` half-refactor, `MobileTarget(WebTarget)` inheritance, embedded TSX heredocs, hardcoded target-name branches in `cli.jac`) are the ones i'd realy like alignment on before more code lands on top of them - everthing else is mostly mechanical cleanup.

mahfoos:
@MusabMahmoodh I couldn’t check the checkbox for the above-mentioned items. All the changes are completed and pushed. Please check.

PR #5923, opened 2026-05-11
fix(jac-scale): make microservice startup timeout configurable + add background recovery monitor

udithishanka:
## Summary

The local microservice orchestrator hardcoded a 20s startup wait in `wait_for_health` (`process_manager.impl.jac:208`), then permanently marked services UNHEALTHY when they missed the bar. The gateway short-circuits routed calls to UNHEALTHY services, so a single slow boot bricked the gateway for the lifetime of the process:

- `/walker/*` returned 404 "Not found on any service"
- `/api/<svc>/*` returned 503 "Service '...' is unhealthy"

Even though the underlying service was answering `/healthz` on its port a few seconds later.

Hit this in production on jac-builder-dev: every dashboard request 404'd after a routine pod restart, and only a full `kubectl rollout restart` cleared it.

## Fix

Two parts, both needed (either alone is insufficient).

**1. Configurable per-service `startup_timeout`.** New `[plugins.scale.microservices.services.NAME].startup_timeout` and global `[plugins.scale.microservices].startup_timeout`, mirroring the existing `http_forward_timeout` precedence (per-service > global > `_DEFAULT_STARTUP_TIMEOUT=60s`). Apps with slow boots (warmpool prefill, byllm import, first-run jac compile) raise the bar via jac.toml.

**2. Background recovery monitor.** `ServiceProcessManager.start_health_monitor` spawns a daemon thread that periodically re-runs `check_all_health` so an UNHEALTHY entry flips back to HEALTHY when its `/healthz` starts responding. `check_all_health` already did the right thing on success (resets `_failure_counts`, sets HEALTHY); the monitor just keeps it running.

`check_all_health` is now lock-protected so the monitor and ad-hoc callers (`LocalDeployer.status`, admin endpoints) can't race on `_failure_counts`. Orchestrator starts the monitor right after the initial wait loop and stops it on shutdown.

## Tests

- `startup_timeout` default + precedence (mirrors `http_forward_timeout` tests)
- `check_all_health` flips UNHEALTHY back to HEALTHY on recovery
- `start_health_monitor` is idempotent, spawns a daemon thread, cleanly stops
- Background monitor actually drives `check_all_health` on its loop

## Backwards compatibility

No behavior change for apps with services that come up under 60s on their first boot (the new default). Apps that were silently relying on the 20s bar to mark something unhealthy will see it stay healthy a bit longer; can pin back to 20s via `startup_timeout = 20.0` if needed.

## Test plan

- [ ] CI: jac-check + test-scale + test-scale-k8s all green
- [ ] Deploy to jac-builder-dev with this branch pinned in `jac.toml`, confirm gateway routing recovers from slow startup


udithishanka:
this is already addressed by @musab-mah-7

PR #5929, opened 2026-05-11
feat(jac-scale): replace Motor with PyMongo native AsyncMongoClient

asimsalim:
## Summary

- Motor (`motor.motor_asyncio`) reached end-of-life May 2026; MongoDB now ships first-class async support directly in PyMongo >= 4.9 via `AsyncMongoClient`
- `MongoBackend.aget` and `acommit` now import `AsyncMongoClient` from the existing `_optdeps/pymongo` shim instead of the motor shim — same bulk_write logic, no behaviour change
- A process-level singleton (`_process_cache['async_mongo_client']`) replaces the per-call client allocation Motor used, matching the existing pattern for the sync `MongoClient`
- `motor.jac` shim is left in place for backward compatibility but is no longer imported by the backend

**Why Motor is EOL:** MongoDB announced Motor's end-of-life (last supported release May 2026) in favour of PyMongo's own native async support (`AsyncMongoClient` added in PyMongo 4.9). Continuing to depend on Motor would require pinning an unmaintained library.

## Test plan

- [ ] Verify existing MongoDB integration tests pass with `AsyncMongoClient`
- [ ] Confirm no `motor` package is required at import time (graceful fallback if unavailable)
- [ ] Check `_process_cache['async_mongo_client']` singleton is reused across requests (no per-request client leak)


musab-mah-7:
@asimsalim we can get rid of the motor references no?
1.  jac-scale/jac_scale/_optdeps/motor.jac: this file is oprpahned, we can get rid of this

2. jac-scale/jac_scale/memory_hierarchy.jac#L150

image

3.  tests are sti lrefenecing/naming with motor: jac-scale/jac_scale/tests/test_async_drivers.jac

4.  let's keep the rleease notes clean? no need of motor stuffs since they are unreleased anyway on main

5. 
image

6.

PR #5932, opened 2026-05-11
fix(jac-scale): forward anchor arg from acommit() to commit()

### LINKED ISSUE #5931: jac-scale: ScaleTieredMemory.acommit() silently ignores the anchor parameter
## Summary

`ScaleTieredMemory.acommit(anchor)` accepts an `anchor` parameter (matching the parent interface) but never forwards it to `commit()` — it always runs a full L1 flush regardless of what is passed.

## Location

`jac-scale/jac_scale/impl/memory_hierarchy.main.impl.jac`

```jac
impl ScaleTieredMemory.acommit(anchor: (Anchor | None) = None) -> None {
    import asyncio;
    if self._committed {
        return;
    }
    await asyncio.to_thread(self.commit);  // anchor is silently dropped here
    self._committed = True;
}
```

## Why it matters

The signature implies a targeted single-anchor commit is possible. If a caller passes a specific anchor expecting only that anchor to be flushed:

1. A full L1 commit runs instead (silent correctness surprise).
2. `_committed` is set to `True` — the middleware safety-net `acommit()` will short-circuit on the next call, believing a full commit already happened (which it did, but not intentionally).

## Current impact

**None in practice.** No call site in jac-scale passes an anchor to `acommit()`. All callers use `acommit()` with no argument. This is a latent footgun for anyone reading the signature and trying to use it as a partial-commit API in the future.

## Fix options

1. **Remove the parameter** — if partial async commits are never needed, drop `anchor` from `acommit()` to match actual behaviour.
2. **Forward the parameter** — pass `anchor` through to `asyncio.to_thread(self.commit, anchor)` so the partial-commit path works as the signature implies, and only set `_committed = True` when `anchor is None` (full commit).

Option 1 is simpler and safer. Option 2 requires deciding whether a partial commit should count as "committed" for the purposes of the flag.

## Related

Discovered during review of PR #5925 (fix for issue 1g — eliminate redundant sync() pass).

akindu-k:
## Problem

`ScaleTieredMemory.acommit(anchor)` accepted an `anchor` parameter but silently dropped it:

```jac
# BEFORE
impl ScaleTieredMemory.acommit(anchor: (Anchor | None) = None) -> None {
    import asyncio;
    await asyncio.to_thread(self.commit);   // anchor never passed
}
```

`commit()` always received `None` regardless of what the caller passed. This violates the contract of the base `Memory.acommit()` interface, which correctly does `asyncio.to_thread(self.commit, anchor)`.

## Fix

Forward the argument:

```jac
# AFTER
impl ScaleTieredMemory.acommit(anchor: (Anchor | None) = None) -> None {
    import asyncio;
    await asyncio.to_thread(self.commit, anchor);
}
```

## Notes

`ScaleTieredMemory.commit(anchor)` intentionally routes all writes through `l3.sync()` (merge-aware full scan) and does not act on the `anchor` argument — this is documented in the existing comment at the top of that implementation and is out of scope here. This PR fixes the forwarding gap in `acommit()`.

## Test

Added to `test_memory_hierarchy.jac` (CI pipeline):

```
test "acommit forwards anchor argument to commit()"
```

Patches `commit()` to capture its argument. Asserts:
1. `acommit()` → `commit()` receives `None`
2. `acommit(sentinel)` → `commit()` receives the same `sentinel` object

No testcontainers or subprocess required.

## Related

Closes #5931. Discovered during review of #5925.


kashmithnisakya:
Hey @akindu-k , one observation for a follow-up: `anchor` is still dropped one level deeper. `ScaleTieredMemory.commit(anchor)` ignores it and runs a full `L1` scan + `l3.sync()`, and `MongoBackend.commit(anchor)` is a no-op. So callers like:


```
Jac.commit(root_anchor)     // server.impl.jac:73
Jac.commit(user_root)       // identity_storage.impl.jac:86
```
still trigger a full bulk sync instead of a single-anchor write.

Suggest landing this as-is and opening a follow-up that:

1. Add a `_sync_one(anchor)` helper in MongoBackend for single-anchor writes.
2. Make `MongoBackend.commit(anchor)` use it when an anchor is passed.
3. Make `ScaleTieredMemory.commit(anchor)` write just that anchor instead of scanning all of L1.

Could you please address these in a follow-up PR? Just flagging that #5931 isn't fully closed by this one.

kashmithnisakya (approved):

akindu-k:
> Hey @akindu-k , one observation for a follow-up: `anchor` is still dropped one level deeper. `ScaleTieredMemory.commit(anchor)` ignores it and runs a full `L1` scan + `l3.sync()`, and `MongoBackend.commit(anchor)` is a no-op. So callers like:
> 
> ```
> Jac.commit(root_anchor)     // server.impl.jac:73
> Jac.commit(user_root)       // identity_storage.impl.jac:86
> ```
> 
> still trigger a full bulk sync instead of a single-anchor write.
> 
> Suggest landing this as-is and opening a follow-up that:
> 
> 1. Add a `_sync_one(anchor)` helper in MongoBackend for single-anchor writes.
> 2. Make `MongoBackend.commit(anchor)` use it when an anchor is passed.
> 3. Make `ScaleTieredMemory.commit(anchor)` write just that anchor instead of scanning all of L1.
> 
> Could you please address these in a follow-up PR? Just flagging that #5931 isn't fully closed by this one.

hey @kashmithnisakya!

You're right, `ScaleTieredMemory.commit(anchor)` still does a full L1 scan and l3.sync() regardless of whether a specific anchor is passed, so the partial-commit optimization isn't actually wired through to the storage layer yet.

Will open a follow-up PR that!

PR #5937, opened 2026-05-11
fix(jac-scale): forward anchor parameter in ScaleTieredMemory.acommit()

### LINKED ISSUE #5931: jac-scale: ScaleTieredMemory.acommit() silently ignores the anchor parameter

RagavRida:
## Summary

Fixes #5931 — `ScaleTieredMemory.acommit(anchor)` silently ignores the `anchor` parameter.

## Problem

`acommit(anchor)` accepted an `anchor` parameter (matching the parent `Memory.acommit` interface) but never forwarded it to `commit()` — it always ran a full L1 flush regardless:

```jac
await asyncio.to_thread(self.commit);  // anchor silently dropped
self._committed = True;               // unconditional — blocks future full commits
```

## Fix

```jac
await asyncio.to_thread(self.commit, anchor);  // anchor forwarded
if anchor is None {
    self._committed = True;                     // only on full commits
}
```

### Behavioral changes

| Scenario | Before | After |
|---|---|---|
| `acommit()` (full, first) | Full L1 flush, `_committed=True` | Same |
| `acommit()` (full, repeat) | Short-circuited | Same |
| `acommit(specific_anchor)` | Full L1 flush, `_committed=True` | Only that anchor, `_committed` unchanged |
| `acommit(anchor)` then `acommit()` | Second call short-circuited | Second call runs (correct) |

## Current impact

**None** — no existing call site passes an anchor. This is a latent correctness fix that makes the API match its signature.

## Files changed

- `jac-scale/jac_scale/impl/memory_hierarchy.main.impl.jac` — fixed `acommit` impl
- `docs/.../jac-scale/5931.bugfix.md` — release note fragment

## Related

Discovered during review of PR #5925.


MichaelAshley2000:
This issue is already being actively addressed in #5932

PR #5944, opened 2026-05-12
fix(byllm): replace sync LLM clients with async equivalents to unblock event loop

akindu-k:
## Problem

`Model.model_call_no_stream`, `model_call_with_stream`, and `model_call_http` used the synchronous `openai.OpenAI` client and `requests.post`. Any walker using `by llm()` pinned the uvicorn event loop for the full LLM round-trip (typically 0.5–30 s).

For async walkers (`__jac_async__ = True`) this was a hard block — no other requests, health checks, or coroutines could run until the LLM responded. For sync walkers the call already ran in a thread pool via `asyncio.to_thread` in `spawn_walker_sync`, but 10 concurrent LLM calls still exhausted thread workers for up to 30 s each.

This is finding **1f** from the jac-scale scalability audit (#5874).

---

### Compiler — emit `await acall_llm(...)` for async abilities

`pyast_gen_pass` now checks `nd.is_async` and emits `await Jac.acall_llm(model, mt_run)` instead of `Jac.call_llm(model, mt_run)` when the enclosing ability is declared `async`. Sync abilities are unchanged.

| File | Change |
|---|---|
| `pyast_gen_pass.jac` | `_invoke_llm_call` signature gains `is_async: bool = False`; return type widens to `ast3.expr` |
| `pyast_gen_pass.impl.jac` | When `is_async`, wraps the `call` in `ast3.Await`; selects `acall_llm` hook name |
| `gen_llm_body` | Passes `is_async=nd.is_async` to `_invoke_llm_call` |

### Runtime hook interface

| File | Change |
|---|---|
| `runtime.jac` | Adds `static async def acall_llm(model, mt_run)` to `JacByLLM` |
| `jaclib.jac` | Registers `acall_llm` in `__all__` and the hook dispatch list |
| `runtime.impl.jac` | Base stub: `await asyncio.sleep(0)` + random-value fallback (mirrors `call_llm`; ensures `inspect.iscoroutinefunction` returns `True` even without byllm installed) |
| `generate_plugin_helpers` | `make_proxy` extended to produce an `async_proxy` wrapper when `inspect.iscoroutinefunction(method)` is true |

### byllm — native async call paths

| File | Change |
|---|---|
| `llm.jac` | Adds `AsyncOpenAI` import; `ainvoke`, `adispatch_no_streaming`, `adispatch_streaming`, `model_call_no_stream_async`, `model_call_with_stream_async` to `BaseLLM`; same + `model_call_http_async`, `_async_proxy_client` to `Model` |
| `basellm.impl.jac` | `ainvoke` routes non-streaming/streaming to native async dispatchers; ReAct loop falls back to `asyncio.to_thread(self.invoke, mt_run)`. `adispatch_no_streaming` mirrors sync dispatcher but awaits `model_call_no_stream_async`. `adispatch_streaming` async-iterates `model_call_with_stream_async`. `model_call_no_stream_async` fallback: `asyncio.to_thread(self.model_call_no_stream, params)`. `model_call_with_stream_async` fallback: drives sync generator step-by-step via `asyncio.to_thread(next, gen)` — avoids blocking the event loop per chunk |
| `model.impl.jac` | `model_call_no_stream_async`: `litellm.acompletion` / cached `AsyncOpenAI` / `model_call_http_async`. `model_call_with_stream_async`: same, `stream=True`; adds missing `except Exception` catch-all matching `model_call_no_stream_async` and the sync path. `model_call_http_async`: `httpx.AsyncClient`. `model_call_http` (sync): `requests` replaced with `httpx.Client`. `_async_proxy_client`: lazily created once per `Model` instance and reused to avoid opening a new connection pool per request |
| `plugin.jac` | Adds `acall_llm` `@hookimpl` → `await model.ainvoke(mt_run)` |
| `pyproject.toml` | Adds `httpx>=0.27.0` |

---

## How it works

**Async walkers** — the compiler emits `await acall_llm(model, mt_run)`. At runtime this calls `await model.ainvoke(mt_run)` which uses `litellm.acompletion`, `AsyncOpenAI`, or `httpx.AsyncClient` natively. The event loop is free for the entire LLM round-trip.

**Sync walkers** — unchanged. They run inside `asyncio.to_thread` via `spawn_walker_sync` and call `model.invoke()` directly.

**ReAct / tool-call loops** — the ReAct iteration logic is synchronous state machinery; `ainvoke` offloads the full `self.invoke(mt_run)` call to a thread pool via `asyncio.to_thread`. The event loop is free; the ReAct loop blocks only the thread.

**Custom `BaseLLM` subclasses** — if they don't override `model_call_with_stream_async`, the base fallback drives their sync generator step-by-step in the thread pool (`asyncio.to_thread(next, gen)` per chunk) so the event loop stays free between chunks.

---

## Closes

Partially addresses #5874 (finding 1f).


asimsalim:
## Code Review

### What's good

- **`ainvoke` is the right interim approach** — `asyncio.to_thread(self.invoke, mt_run)` correctly frees the event loop without rewriting the entire react loop. Clean single-line implementation.
- **Compiler codegen change is correct** — `_invoke_llm_call(is_async=nd.is_async)` emitting `ast3.Await(value=call)` for async walker abilities is the right place to make this change.
- **`generate_plugin_helpers` proxy fix** — detecting `iscoroutinefunction(method)` on the spec and generating `async_proxy` with the `iscoroutine(result)` guard is a solid generic fix for async hookimpls.
- **Tests cover the new async methods** — `model_call_no_stream_async`, `model_call_with_stream_async`, and `model_call_http_async` all have test coverage.

---

### Issues

**1. Dead code — async model methods are never called**

`model_call_no_stream_async`, `model_call_with_stream_async` and `model_call_http_async` are full implementations using `AsyncOpenAI` / `litellm.acompletion` / `httpx.AsyncClient` — but `ainvoke` routes through `asyncio.to_thread(self.invoke, ...)` which calls the **sync** path. These methods are unreachable until the full async dispatch chain is connected.

The PR description acknowledges this ("requires the compiler to emit `await Jac.acall_llm(...)`"), but that change IS in this PR (via `pyast_gen_pass`). The codegen now emits `await Jac.acall_llm(...)` for async walkers, but `acall_llm` → `ainvoke` → `asyncio.to_thread(invoke)` → sync methods. The native async methods (`model_call_no_stream_async` etc.) are never on the call path.

Either wire them up (replace `asyncio.to_thread` with a proper `adispatch_no_streaming` path), or hold them for a follow-up PR and don't add them here.

**2. `acall_llm` stub is not `async`**

In `runtime.impl.jac`:
```
impl JacByLLM.acall_llm(model: LLMModel, mt_run: MTRuntime) -> Any {
```
The declaration in `runtime.jac` is `static async def acall_llm`. The stub should match. Without `async`, when byllm is not installed the stub returns a plain value through the `async_proxy`. The proxy guards with `if inspect.iscoroutine(result)` which handles it, but the asymmetry is confusing and fragile.

**3. `model_call_with_stream_async` missing `http_client` branch**

`model_call_no_stream_async` correctly handles three cases — `http_client`, `proxy`, and litellm. `model_call_with_stream_async` only handles `proxy` and litellm. Anyone with `http_client=True` + streaming on an async walker will silently fall through to litellm instead of the custom endpoint:

```
# model_call_no_stream_async — correct
if self.http_client {
    response = await self.model_call_http_async(...);
} elif self.proxy { ... }

# model_call_with_stream_async — missing http_client branch
if self.proxy { ... }   # ← http_client case skipped
```

**4. `httpx` not added to `pyproject.toml`**

The PR description says "Add `httpx>=0.27.0` as explicit dependency" but there's no `pyproject.toml` change in the diff. If `httpx` is only available transitively the install will be fragile. Needs to be added explicitly.

**5. `requests` not removed after being replaced**

`model_call_http` now uses `httpx.Client` in the sync path, but `requests` likely remains in `pyproject.toml`. If `httpx` fully replaces it, `requests` should be dropped.

**6. No test for the codegen path**

Tests cover the async model methods and the `acall_llm` proxy signature, but there's no test that:
- Compiles an `async walker` with `by llm()`
- Verifies the generated Python contains `await Jac.acall_llm(...)` (not `Jac.call_llm(...)`)
- Runs the walker end-to-end against a mocked `acall_llm`

The codegen change in `pyast_gen_pass` is the most load-bearing part of this PR and is currently untested.

---

### Summary

The `ainvoke` → `asyncio.to_thread` approach is correct and the compiler wiring is right. Blockers before merge: fix the `acall_llm` stub to be `async`, add the missing `http_client` branch to `model_call_with_stream_async`, add `httpx` to `pyproject.toml`, and add at least one codegen test. The dead async model methods should either be connected or deferred to a follow-up PR.

PR #5953, opened 2026-05-12
fix(parser): Emit Clear Error for `glob` used other than Module scope

### LINKED ISSUE #5849: [Better Error Msg]:  Defining glob keyword inside a function
### Operating System

Ubuntu

### Versions

Jac 0.14.0 

### Steps to Reproduce

1. Create a jac file `test.jac` and paste the code below
2. 
```jac
with entry {
  def foo() -> None {
      glob x: int = 5;
  }
}
```
3. Run the jac check with `jac check test.jac`

Expected: Proper error message saying glob keyword can not declare inside a function

Actual: 
```
✖ Error: error[E0002]: Missing ';'
  --> test2.jac:3:12
    1 | with entry {
    2 |   def foo() -> None {
    3 |       glob x: int = 5;
      |            ^
    4 |   }
    5 | }
```

SandeepaHWP:
Closes #5849

## What was the problem?

When you write `glob x: int = 5` inside a function body, the compiler was giving a confusing error:

```
error[E0002]: Missing ';'
```

This message has nothing to do with the actual mistake. On top of that, `jac run` would silently execute the code as if it were valid, producing wrong output instead of stopping.

## What was actually happening?

The parser had no handler for the `glob` keyword inside function bodies. It fell through to expression parsing, treated `glob` as a variable name, and then failed midway through — hence the misleading `Missing ';'` error.

## What was changed?

- Added a new diagnostic `E0063` with the message: `'glob' is only valid at module level`
- Added an early check in `parse_statement()` — when `glob` is seen inside a function, it now emits `E0063` and recovers cleanly
- Added tests for both the error case and the valid module-level case

## Before vs After

**Before:**
```
error[E0002]: Missing ';'
  --> test.jac:3:12
```

**After:**
```
error[E0063]: 'glob' is only valid at module level
  --> test.jac:3:7
    3 |         glob x: int = 5;
      |         ^^^^
```

**Jac Run and Check Results**

Screenshot 2026-05-12 183032

--------

Screenshot 2026-05-12 183043


RagavRida:
The CI failure is in the `grammar matches spec` test — this is expected when the parser grammar changes.

### Root Cause

The test at `jac/tests/compiler/passes/tool/test_grammar_extract_pass.jac:304` ("grammar matches spec") works by:
1. Extracting grammar rules from the live parser code
2. Comparing them against the saved snapshot in `jac.spec`
3. Failing if any rules were added, removed, or changed

Since this PR adds a new early check for `glob` in `parse_statement()`, the extracted grammar now differs from the saved `jac.spec` file.

### Fix

Regenerate the spec file after your parser changes:

```bash
jac grammar -o jac/jaclang/jac.spec
```

Then commit the updated `jac.spec`. The test will pass.

The test error message itself tells you this (though it gets swallowed by the assertion format):
> `If intentional: jac grammar -o <path>`

PR #5965, opened 2026-05-14
perf(auth): Two-layer JWT cache + credential-epoch token revocation

akindu-k:
## Problem

JWT validation had two production-grade gaps that made it unsuitable for high-throughput deployments.

### 1. No token decode cache

Every authenticated request called `jwt.decode()` regardless of whether the same token had just been validated milliseconds earlier. At 500 RPS, that is 500 cryptographic HMAC-SHA256 verifications per second — pure repeated CPU work with no reuse.

```jac
# Before — called on every single request
decoded = jwt.decode(token, JWT_SECRET, algorithms=[JWT_ALGORITHM]);
return decoded.get('user_id') or None;
```

### 2. No token revocation

Once a JWT was issued it remained valid until its `exp` claim elapsed — regardless of what happened to the account. Concretely:

- A user deletes their account → their token stays valid for up to 7 days (the default `exp_delta_days`)
- A user changes their password after a suspected compromise → the attacker's stolen token keeps working
- An admin force-logs out a user → no mechanism existed to honour that

There was no `tokens_valid_from` boundary, no revocation list, and no call site that invalidated tokens on security-sensitive mutations.

---

## Solution

A three-layer validation strategy in `validate_jwt_token` and a new `invalidate_user_tokens` method that is called automatically on password change and account deletion.

### Layer 1 — Process-local decode cache (hot path)

```
SHA-256(raw_token) → (user_id, monotonic_expiry)   TTL = 30 s   max = 50 000 entries
```

On a cache hit the method returns `user_id` immediately — no `jwt.decode()`, no DB access. A reverse index (`user_id → set[token_hash]`) allows eager eviction of all cached entries for a user the moment their tokens are revoked, so revocation in the same worker is instant rather than waiting for TTL expiry.

### Layer 2 — Cryptographic verification (`jwt.decode`)

On a cache miss, `jwt.decode()` validates the signature and `exp` claim. Still no DB access.

### Layer 3 — Credential-epoch revocation check

Each user document may carry a `tokens_valid_from` Unix timestamp. Any token whose `iat` (issued-at) predates that timestamp is rejected. This covers the entire token history for that user in a single comparison.

The per-user timestamp is cached in-process for 60 seconds, so even at high request rates the DB is consulted at most once per user per window.

```
user_id → (tokens_valid_from_ts, monotonic_expiry)   TTL = 60 s
```

If the user no longer exists (deleted account), the cache stores `float('inf')` as the boundary so subsequent requests are rejected without further DB lookups for the remainder of the TTL.

### `invalidate_user_tokens(user_id)`

New public method that:

1. Writes `tokens_valid_from = time.time()` to the user document so other workers and process restarts pick it up
2. Eagerly evicts all decode-cache entries for that user via the reverse index
3. Evicts the revocation-cache entry so the next validation immediately reads the new boundary

Called automatically in:
- `update_password` — on success, so a compromised credential cannot extend an existing session
- `delete_user` — after the record is removed, so the deleted user cannot keep authenticating

### `jti` claim on new tokens

`create_jwt_token` now includes a `jti` (JWT ID) UUID in every issued token. This follows RFC 7519 best practice and lays the groundwork for a server-side allowlist or per-token revocation list if tighter guarantees are needed in the future.

#### Refer: #5976 
---

## Performance model

| Scenario | Before | After |
|---|---|---|
| Same token seen twice within 30 s | 2× `jwt.decode()` | 1× decode + 1× dict lookup |
| 500 RPS, 100 active users | 500 decodes/s | ≤ 100 decodes/s (TTL misses only) |
| DB reads for revocation check | 0 (no revocation at all) | ≤ 1 per user per 60 s |
| Password changed, old token used | Accepted until `exp` | Rejected within 60 s (next revocation-cache miss) |
| User deleted, old token used | Accepted until `exp` | Rejected immediately in same worker; within 60 s in other workers |

---

## Multi-process behaviour

The process-local caches are **not** shared across uvicorn/gunicorn workers. Revocation propagates to other workers on their next revocation-cache miss, which happens within `_REVOCATION_CACHE_TTL` (default 60 s). The `tokens_valid_from` value in the DB is the durable source of truth; the in-process caches are read-through layers on top of it.

For sub-second cross-worker revocation, `_REVOCATION_CACHE_TTL` can be reduced, or the revocation timestamp can be centralised in Redis using the same pattern as the existing `RedisBackend`.

---

## Files changed

| File | Change |
|---|---|
| `jac_scale/user_manager.jac` | Added `time`, `hashlib` imports; added cache globals (`_decoded_token_cache`, `_revocation_ts_cache`, `_user_token_hashes`) and TTL/size constants; added `invalidate_user_tokens` to the `JacScaleUserManager` interface |
| `jac_scale/impl/user_manager.impl.jac` | Rewrote `validate_jwt_token` with three-layer logic; added `invalidate_user_tokens` impl; updated `update_password` to invalidate on success; updated `delete_user` to invalidate before removal; added `jti` claim to `create_jwt_token` |

---

## Testing checklist

- [x] Existing auth tests pass unchanged (no interface breakage)
- [x] Token issued before `invalidate_user_tokens` is rejected on next validation
- [x] Token issued after `invalidate_user_tokens` is accepted
- [x] Password change immediately invalidates old tokens in the same worker
- [x] Deleted user's token is rejected immediately in the same worker
- [x] Cache eviction under high volume (> 50 000 unique tokens) does not panic
- [x] `tokens_valid_from` field is absent on legacy user documents — validated as `0.0` (all tokens valid, backward-compatible)


akindu-k:
# Backward Compatibility — PR #5965 

Existing tokens issued before this PR are fully unaffected. Here is why, layer by layer.

## L1 — Decode Cache

No impact. The cache is empty on startup. Existing tokens get a cache miss and fall through to L2 — same as any first-time token.

## L2 — Cryptographic Verification

No impact. `jwt.decode()` only checks the signature and `exp` claim. It does not care whether `jti` is present or absent in the payload. Old tokens without `jti` decode fine.

## L3 — Revocation Check

This is the layer that could have broken things, but does not. Existing users have no `tokens_valid_from` in their document — it was never set before this PR. So when L3 reads the user doc:

```jac
_raw_vf = user_doc.get('tokens_valid_from');
valid_from_ts = 0.0 if _raw_vf is None else float(_raw_vf);
```

`_raw_vf` is `None`, so `valid_from_ts` becomes `0.0`. The boundary check then evaluates as:

```
iat < 0.0  →  false
```

`iat` is always a large positive Unix timestamp (e.g. `1747123456.0`), so every existing token passes this check unconditionally.

This behaviour is explicitly covered by a test:

> `tokens_valid_from absent on legacy user document is backward-compatible`

## SQLite Schema Migration

For existing SQLite databases, `ensure_indexes()` runs an `ALTER TABLE` on startup to add the `tokens_valid_from` column. If the column already exists, the error is caught and ignored. Existing rows get `NULL` for that column, which is exactly the `None` case handled above.

## When Does an Existing Token Stop Working?

Only in these cases — all of which are intentional:

- The user changes their password → `invalidate_user_tokens()` is called, boundary is set to now
- The user's account is deleted → same
- The token's own `exp` claim expires naturally

Projects already running jac-scale will see zero disruption on upgrade. Tokens keep working, users stay logged in, nothing needs to be reissued.

Issue #2, opened 2021-10-29
Apply MIT License to all code


marsninja:
We need to apply the MIT Liscense to the entire codebase using the necessary best practices.

PR #5977, opened 2026-05-15
fix(jaclang): clear error message when jac.toml is malformed

MusabMahmoodh:
## Summary

A TOML syntax error in `jac.toml` produced N copies of a misleading "Install the missing package" warning instead of telling the user to fix their config. This PR makes the plugin loader recognise `TOMLDecodeError`, print one actionable red error, and exit with code 2.

## Before

```
$ jac start main.jac
warning: plugin 'jac_scale' failed to load -- TOMLDecodeError: Unterminated
string (at line 2, column 14). Install the missing package, or disable this
plugin via JAC_DISABLED_PLUGINS / [plugins].disabled in jac.toml.
warning: plugin 'jac_cloud' failed to load -- TOMLDecodeError: ...
warning: plugin 'jac_splice_orc' failed to load -- TOMLDecodeError: ...
... (repeats for every installed plugin)
```

The "install the missing package" hint is wrong - the package is fine, the TOML file is broken - and it repeats once per plugin.

## After

```
$ jac start main.jac
ERROR: jac.toml is malformed: Unterminated string (at line 2, column 14)
  Fix the syntax error in jac.toml and re-run.
  Common causes: missing closing quote, unbalanced brackets, trailing
  comma, missing `=`.

$ echo $?
2
```

One message, points at the actual problem, exits non-zero so CI / wrapper scripts notice.

## Root cause

`jac/jaclang/jac0core/helpers.jac` catches every plugin-load exception with the same generic "missing dependency" hint. Plugins that read `jac.toml` eagerly at import time (jac-scale's webhook config, etc.) all surface the same `TOMLDecodeError`, so the user sees N identical wrong hints.

## Fix

Per-plugin `except Exception` now dispatches by type:

| Exception                 | Behaviour                                              |
|---------------------------|--------------------------------------------------------|
| `tomllib.TOMLDecodeError` | Print actionable red error, `sys.exit(2)`              |
| `ModuleNotFoundError`     | Existing soft-fail warning (other plugins still load)  |
| anything else             | Existing soft-fail warning                             |

`sys.exit(2)` is deliberate: a malformed `jac.toml` fails every plugin's `load()` anyway, so there's no useful work to continue with. Exit 2 is the conventional "user config error, fix and re-run" code.

## Test plan

- [x] `jac test jac/tests/runtimelib/test_plugin_loader_toml_error.jac` - subprocess test asserts exit 2, stderr contains "jac.toml" / "malformed" / "fix", and does NOT contain "Install the missing package"
- [x] Manual repro: broken `jac.toml` -> new red error, exit 2
- [x] Manual: valid `jac.toml` -> `jac --version` and `jac start` work normally, no warnings
- [x] Manual: plugin with genuinely missing dep -> existing soft-fail warning still appears (not regressed)

## Scope

Two files: the loader and one new test. No behaviour change for the `ModuleNotFoundError` path or for successful loads.


kugesan1105:
@MusabMahmoodh

few things from testing this locally:

## 1. this makes `import jaclang` refuse to start instead of warn

right now if jac.toml is broken anywhere up the directory tree, anything that imports jaclang dies. python scripts, notebooks, vscode extensions, even `jac format` (which you'd want to use to fix the file). that's a much bigger change than "clearer error message".

python doesn't refuse to start when pyproject.toml is broken. node doesn't refuse when package.json is broken. i think we should do the same:

- `import jaclang` / `get_disabled_plugins()` should just log a warning ("jac.toml is malformed, ignoring") and continue with `disabled_list = []`.
- `jac start` / `jac run` / `jac build` etc. are where the user is actually asking jac to use the project config, so hard-exit with the red banner there.

the `_exit_on_malformed_jac_toml` helper is fine, just move the call out of library bootstrap into the CLI commands. fixes #2 below as a side effect.

## 2. fatal traceback during site-init

since `sys.exit(2)` runs during `import jaclang`, and `jac/sitecustomize.py` imports jaclang during python's site-init, the SystemExit triggers `Fatal Python error: init_import_site`. you get the red banner *and* a full traceback, exit code 1 instead of 2.

some people don't see this because their conda env ships its own `sitecustomize.py` that shadows the repo's, so they get the clean output. depends on `sys.path` ordering. fixing #1 makes this go away.

## 3. the test fails on dev setups

the test asserts `returncode == 2` but on any machine where `/path/to/jac/` is on `sys.path` (i.e. normal contributor setup) the subprocess hits the site-init thing and returns 1 with a traceback. CI probably hits this too. needs `python -I` or scrubbed `PYTHONPATH`, or just retarget at the CLI path after #1.

image

## 4. only one caller is covered

`read_toml_with_encoding` has ~7 callers (config, eject, template loader, etc.) and only `get_disabled_plugins` gets the friendly error. same broken jac.toml will show the red banner in one path and a raw `TOMLDecodeError` in another. better to push it into `read_toml_with_encoding` itself with a flag.

MusabMahmoodh:
@kugesan1105 — all four addressed in `cc9f44460`. Thanks, this was a real find — the site-init path made the original fix worse than the bug it was solving.

**Architecture now:**

```
read_toml_with_encoding(path)               # the one chokepoint
  TOMLDecodeError -> raise MalformedJacTomlError(file_path, original)

Library callers (import time)               # warn + continue
  get_disabled_plugins -> _warn_malformed_jac_toml -> return []

CLI callers (jac <command>)                 # banner + exit 2
  _load_project_config -> _exit_on_malformed_jac_toml -> sys.exit(2)
```

**Per-point:**

1. `sys.exit(2)` removed from library bootstrap. `import jaclang`, `get_disabled_plugins`, and the autonative check now log a one-line stderr warning and continue with empty config. Notebooks, scripts, and `jac format` keep working on a broken jac.toml.

2. Side-effect fix: no `sys.exit` during `import jaclang` means no `Fatal Python error: init_import_site` from `sitecustomize.py`.

3. Test retargeted at the CLI path. Two tests now:
   - `library: import jaclang with malformed jac.toml succeeds with warning` — asserts rc=0, stderr contains the warning.
   - `cli: jac --version with malformed jac.toml exits 2 with banner` — asserts rc=2, stderr contains the red banner.
   PYTHONPATH and user-site are scrubbed in the subprocess env so dev machines and CI behave the same.

4. `MalformedJacTomlError` lives inside `read_toml_with_encoding` itself, so the 7+ callers (config, eject, template loader, project, etc.) all get the typed error consistently. Library-vs-CLI handling becomes a `try/except MalformedJacTomlError` decision at each site.

**Behavior now:**

```
$ python -c "import jaclang; print('ok')"   # broken jac.toml present
warning: jac.toml is malformed (/path/jac.toml): Illegal character '\n' (at line 2, column 15); ignoring [plugins].disabled. Run a `jac` CLI command for the actionable error.
warning: plugin 'jac-scale:scale' failed to load -- jac.toml parse failure (see warning above).
ok
$ echo $?
0

$ jac --version   # same broken jac.toml
warning: jac.toml is malformed ... (above)
ERROR: jac.toml is malformed: Illegal character '\n' (at line 2, column 15)
  File: /path/jac.toml
  Fix the syntax error in jac.toml and re-run.
  Common causes: missing closing quote, unbalanced brackets, trailing comma, missing `=`.
$ echo $?
2
```

The "Install the missing package" hint is also no longer repeated for every plugin that fails on the same parse error — `MalformedJacTomlError` gets its own concise "see warning above" message in the per-plugin loop.

PR #6003, opened 2026-05-17
 fix(persistence): cascade-quarantine edges when their node is quarantined

### LINKED ISSUE #5987: Schema drift quarantine path leaves dangling inbound edges (NodeAnchor not a valid reference)
## Summary

Renaming or removing a node `archetype` between deploys leaves dangling edge anchors in the database after the runtime quarantines the unloadable target nodes. The source node's `data.edges` array still references the now-quarantined targets, and the corresponding `EdgeAnchor` docs remain in `_anchors`. Subsequent walker traversals that dereference these edges hit `NodeAnchor [<uuid>] is not a valid reference` in production.

This is distinct from #5446 / #5644. PR #5644 fixed the concurrent-edge-add race on the source's `data.edges` list and is doing its job — a separate add/delete concurrency MRE (200 seed items, 200 parallel add/delete pairs, then a 10-thread chaos phase) produces zero orphans against jaseci@main + MongoDB 7.

The orphan-edge state in this issue is produced by a different path: the **quarantine-on-deserialize-failure** flow in `MongoBackend` correctly moves an unloadable node anchor to `_anchors_quarantine`, but does not also clean the inbound `EdgeAnchor` docs nor strip those edge ids from every other node's `data.edges`. Result: corruption that matches what we cleaned up tonight on `jac-builder` prod — 9 orphan edges (7 with `source exists, target missing`, all sources being `user_root`).

## What I verified locally

Tested against the head of jaseci main with the user-fork installed in editable mode (paths under `udith/jaseci`). MongoDB 7 in a local container.

### Step 1 — v1 schema

`app.jac`:
```jac
node Foo {
    has x: int = 0;
}

walker AttachFoo {
    has x: int = 0;
    can run with Root entry {
        f = Foo(x=self.x);
        here ++> f;
        report {"id": str(f.__jac__.id)};
    }
}
```

`jac.toml`:
```toml
[project]
name = "schema-drift"
version = "0.1.0"
entry-point = "app.jac"
```

```bash
docker run -d --name mre-mongo -p 27018:27017 mongo:7
export MONGODB_URI=mongodb://localhost:27018/jac_db
jac start app.jac --port 8002 &

# Register, login, get JWT
curl -s -X POST http://localhost:8002/user/register \
    -H "Content-Type: application/json" \
    -d '{"identities":[{"type":"username","value":"u1"}],"credential":{"type":"password","password":"P!"}}'
TOKEN=$(curl -s -X POST http://localhost:8002/user/login \
    -H "Content-Type: application/json" \
    -d '{"identity":{"type":"username","value":"u1"},"credential":{"type":"password","password":"P!"}}' \
    | python3 -c 'import json,sys;print(json.load(sys.stdin)["data"]["token"])')

# Attach 3 Foo nodes
for i in 1 2 3; do
    curl -s -X POST http://localhost:8002/walker/AttachFoo \
        -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
        -d "{\"x\": $i}"
done
```

State after step 1: `_anchors` has root + user_root + UserProfile + 3 Foo nodes + 3 generic edges from user_root to each Foo.

### Step 2 — v2 schema: `Foo` is removed

Stop the server, replace `app.jac` so `Foo` is no longer defined anywhere:

```jac
node OtherType {
    has label: str = "";
}

walker MeStyle {
    can run with Root entry {
        # Same traversal shape as jac-builder's `me` walker:
        # profiles = [here-->(?:UserProfile)];
        profiles = [here-->(?:OtherType)];
        report {"count": len(profiles)};
    }
}
```

Restart against the same database. Sign in as the same user, run the walker.

### Step 3 — what happens

The server log shows the runtime correctly recognising the unloadable archetype:

```
WARNING - Refused to deserialize unregistered class: app.Foo
WARNING - Serializer: archetype reconstruction failed for app.Foo; dropping anchor (caller should quarantine).
WARNING - MongoBackend: class app.Foo unresolvable for anchor 25db7a7d-34cb-4f23-9944-62b6636f8ccc. Moving to quarantine.
```

This is the right behaviour for the missing-class case. But the move is one-sided. Inspect the post-traversal state:

```bash
docker exec mre-mongo mongosh jac_db --quiet --eval '
const a = db.getCollection("_anchors");
const q = db.getCollection("_anchors_quarantine");
const nodeIds = new Set(a.distinct("_id", {type:"NodeAnchor"}));
print("Live NodeAnchors:  " + nodeIds.size);
print("Live EdgeAnchors:  " + a.countDocuments({type:"EdgeAnchor"}));
print("Quarantined:       " + q.countDocuments({}));
a.find({type:"NodeAnchor", "data.edges.0": {$exists:true}}).forEach(n => {
    const eids = (n.data && n.data.edges) || [];
    print("node " + n._id.slice(0,8) + " (" + (n.arch_type||"?") + ") has " + eids.length + " edges:");
    eids.forEach(eid => {
        const e = a.findOne({_id: eid});
        if (!e) { print("    edge " + eid.slice(0,8) + " — anchor missing from _anchors"); return; }
        const tgt = e.data.target;
        const live = nodeIds.has(tgt);
        const inQ = q.findOne({_id: tgt}) !== null;
        print("    edge " + eid.slice(0,8) + " -> tgt " + (tgt||"").slice(0,8)
            + " tgt_in_anchors=" + live + " tgt_in_quarantine=" + inQ);
    });
});'
```

The same run produces this (matches the prod orphan pattern):

```
Live NodeAnchors:  3
Live EdgeAnchors:  3
Quarantined:       3
node <root.id>... edges: 3
    edge <e1>... -> tgt <foo1>... tgt_in_anchors=false tgt_in_quarantine=true
    edge <e2>... -> tgt <foo2>... tgt_in_anchors=false tgt_in_quarantine=true
    edge <e3>... -> tgt <foo3>... tgt_in_anchors=false tgt_in_quarantine=true
```

Three edge anchors remain in `_anchors` with `data.target` pointing into `_anchors_quarantine`. The source `data.edges` array still contains all three ids.

## How this surfaces in production

Production `jac-builder` (https://jac-builder.jaseci.org) accumulated 9 such orphan edges over a multi-month window — 7 of them with `source exists, target missing`, all sources being `user_root`. Users on those affected roots see `walker/me` fail with:

```
NodeAnchor [<uuid>] is not a valid reference!
  File "/app/services/ideServer.jac", line 559, in handle_action
    profiles = [-->(?:UserProfile)];
```

In our local jaseci@main reproduction the same dangling state exists, but the typed traversal returns `count: 0` without raising. The state is verified identical to prod; the user-facing exception did not manifest in my local runtime version. I'm noting this honestly — I can't claim to reproduce the exception itself, only the underlying corruption that prod's exception is consistent with. Differences in jaclang version, walker context, or the exact load order may be required to trigger the throw. If the maintainer wants help isolating which code path turns the silent-skip into an exception, I can keep digging.

## What is *not* the cause (so the fix doesn't go the wrong direction)

- **Not** a delete-cascade race. A single walker that does `f = Foo(); here ++> f; del f;` cleans up perfectly — verified: ending state has 0 orphan edges and 0 dangling ids in `data.edges`.
- **Not** the concurrent-edge-add race fixed by #5644. The add race remained zero across 200 parallel add/delete cycles + a 10-thread chaos phase on the same Mongo backend. #5644 is doing exactly what it claims.

## Suggested fix shape

In the `MongoBackend` (and the equivalent SQLite path) — wherever the "Moving to quarantine" log line fires — extend the same transaction so it also:

1. Finds every `EdgeAnchor` in `_anchors` where `data.source == quarantined_id OR data.target == quarantined_id`.
2. Moves those edge anchors to a parallel `_edges_quarantine` collection (or to `_anchors_quarantine`).
3. Runs `$pullAll` on every live `_anchors` doc's `data.edges` to strip the moved edge ids.

All three in a single Mongo `with_transaction()` block (SQLite: single `BEGIN IMMEDIATE`).

The cost is one indexed scan over `_anchors` per quarantine event. Adding an index on `data.source` and `data.target` would keep this sub-ms. Without this cascade, every renamed/removed node permanently corrupts its source's `data.edges` until manually scrubbed.

## Operator workaround (current)

`jac-builder` uses a 4-layer cleanup script:
1. Quarantine orphan edges from `_anchors` to a backup collection.
2. `$pullAll` stale edge ids from source nodes' `data.edges`.
3. `DEL` the affected keys from the Redis hot cache.
4. Rolling-restart the app pods.

The script is at `jaseci-labs/jacBuilder` as `orphan-edge-quarantine` — happy to upstream a generic version of layers 1+2 if useful.

## Related

- #5446 — graph persistence race (closed by #5644). Distinct from this issue.
- #5644 — fix(persistence): prevent concurrent walker edge loss via delta-merge atomic writes (merged 2026-04-21). Working correctly; this issue is a different code path.

akindu-k:
## Problem

When a `NodeAnchor`'s archetype becomes unresolvable after schema drift (e.g. `app.Foo` is removed between deploys), the runtime correctly moves that node to `_anchors_quarantine`. However the move was **one-sided**:

- The `EdgeAnchor` documents whose `data.source` or `data.target` referenced the quarantined node stayed in `_anchors`.
- The source node's `data.edges` array still listed those dead edge IDs.

This left the database in a permanently corrupt state — every traversal touching the source node would encounter edges pointing into quarantine with no live target.

A second silent corruption existed independently: when a field's stored value could not be coerced to the current type annotation (e.g. a `list` stored in an `int` field, or a nested object whose class was removed), the anchor was returned to the caller with a corrupt field value instead of being quarantined.

**Pre-fix state after loading three `Foo` nodes on a v2 server:**

```
Live NodeAnchors:  3
Live EdgeAnchors:  3       ← orphaned edges still in _anchors
Quarantined:       3

node <root> has 3 edges:
    edge <e1> -> tgt <foo1>  tgt_in_anchors=false  tgt_in_quarantine=true  ← BROKEN
    edge <e2> -> tgt <foo2>  tgt_in_anchors=false  tgt_in_quarantine=true  ← BROKEN
    edge <e3> -> tgt <foo3>  tgt_in_anchors=false  tgt_in_quarantine=true  ← BROKEN
```

### 1. Cascade quarantine on node removal

Extended the quarantine path in both backends so that quarantining a `NodeAnchor` cascades to its connected edges.

**MongoDB — `_mongo_quarantine`**

After the node is moved to `_anchors_quarantine` and deleted from `_anchors`:

1. Query `_anchors` for all `EdgeAnchor` docs where `data.source == id` **or** `data.target == id`.
2. Write all cascade-quarantine docs in one `bulk_write` call (a single round-trip, not N `update_one` calls).
3. `delete_many` those edges from `_anchors`.
4. `$pullAll` the orphaned edge IDs from every remaining live `NodeAnchor`'s `data.edges`.

**SQLite — `_quarantine`**

Parallel logic using `json_extract(data, '$.source')` / `json_extract(data, '$.target')` for the edge lookup, Python-level JSON read-modify-write to strip edge IDs from live node rows.

- All cascade writes are wrapped in a `SAVEPOINT cascade_q` so a mid-cascade failure rolls back only the cascade, not the already-written node quarantine.
- The per-edge `DELETE` is batched into a single `DELETE … WHERE id IN (…)` instead of N round-trips.
- All operations (node quarantine + cascade) share a single `conn.commit()`.

### 2. Edge re-link on recovery

When `recover-all` / `recover-one` restores a cascade-quarantined `EdgeAnchor`, the connected node(s)' `data.edges` were already stripped during the cascade. Fixed by re-adding the edge ID on recovery:

- **Directed edges**: edge ID is re-added to the source node's `data.edges` only.
- **Undirected edges** (`is_undirected=True`): edge ID is re-added to **both** the source and target node's `data.edges`, restoring full bidirectional graph connectivity.
- **MongoDB**: `$addToSet {'data.edges': edge_id}` on each applicable node after write-back.
- **SQLite**: JSON read-modify-write to append the edge ID if absent, for each applicable node.

This fully restores graph connectivity after a recover cycle.

### 3. Quarantine when nested class is unresolvable on load

`_deserialize_archetype` already called `coerce()` against field type annotations but silently kept the bad value when coercion failed. This is now split into two paths:

- **Hard corruption — raise `TypeError`:** when `_deserialize_value` returns `None` for a non-Optional field. This only happens when a nested object's class is unresolvable (removed between deploys). The `TypeError` is caught by the existing `except` in `_deserialize_archetype`, which returns `None` to `_deserialize_anchor`, triggering the standard quarantine + cascade path with no additional code path needed.
- **Serialisation limitation — warn and keep:** all other coercion misses (e.g. `set` round-tripping as `str` due to JSON limitations) log a `WARNING` and keep the stored value, preserving existing behaviour. These are not corruption — `coerce` simply has no handler for that round-trip form.

## Verification

Reproduced the exact scenario from the issue report against a live `mongo:7` container.

### Step 1 — Seed v1 state

Directly inserted into `_anchors`:
- 1 root `NodeAnchor` with `data.edges: [e1, e2, e3]`
- 3 `Foo` `NodeAnchor` docs (archetype `app.Foo`, not registered in v2)
- 3 `EdgeAnchor` docs: root → foo1, root → foo2, root → foo3

```
Live NodeAnchors:  4
Live EdgeAnchors:  3
Quarantined:       0
Root data.edges:   ['e1', 'e2', 'e3']
```

### Step 2 — Simulate v2 (load Foo nodes via patched MongoBackend)

```
WARNING - MongoBackend: class app.Foo unresolvable for anchor <foo1>. Moving to quarantine.
INFO    - MongoBackend: cascade-quarantined 1 edge(s) connected to quarantined node <foo1>.
WARNING - MongoBackend: class app.Foo unresolvable for anchor <foo2>. Moving to quarantine.
INFO    - MongoBackend: cascade-quarantined 1 edge(s) connected to quarantined node <foo2>.
WARNING - MongoBackend: class app.Foo unresolvable for anchor <foo3>. Moving to quarantine.
INFO    - MongoBackend: cascade-quarantined 1 edge(s) connected to quarantined node <foo3>.
```

### Step 3 — Post-fix state (matches expected outcome from issue)

```
Live NodeAnchors:  1       ← only root remains
Live EdgeAnchors:  0       ← no orphaned edges in _anchors
Quarantined:       6       ← 3 Foo nodes + 3 cascade-quarantined edges

Root data.edges:   []      ← stripped clean
PASS: root data.edges is empty — no dangling edge references
```

No `EdgeAnchor` with a quarantined target remains in `_anchors`. No live `NodeAnchor` retains a reference to a quarantined edge.

## Tests

**New** — `test_mongo_layer123.jac`: "MongoBackend cascade-quarantines edges when their target node is quarantined"

Seeds the exact v1→v2 schema-drift layout from the issue, triggers quarantine via `backend.get()`, then asserts:
- The `Foo` node is absent from `_anchors` and present in `_anchors_quarantine`.
- Each edge is absent from `_anchors` and present in `_anchors_quarantine` with `"cascade quarantine"` in its error.
- The source node's `data.edges` is empty.

**New** — `test_mongo_layer123.jac`: "MongoBackend recover re-links undirected edge to both endpoints"

Seeds two live nodes and a quarantined `GenericEdge` with `is_undirected=True`, calls `recover_one`, then asserts both `source.data.edges` and `target.data.edges` contain the recovered edge ID.

**New** — `test_persistence_migration.jac`: "unresolvable nested class in field is quarantined, not silently loaded as None"

Seeds two rows against a `_TypedProfile` node (`score: int`): one valid row and one with a nested `{ __type__: RemovedClass }` in `score`. `_deserialize_value` returns `None` for the unresolvable class; `coerce(None, int)` raises `TypeError`. Asserts the valid row loads and the bad row is quarantined and absent from the live table.

**Updated** — `test_db_cli.jac`:
- `"db recover-all"`: quarantine count `2`→`4` and recovery string `"2 of 2"`→`"4 of 4"` (cascade now also quarantines the 2 connecting `GenericEdge` anchors; all 4 are fully recovered after alias add). Added a direct DB assertion via `json_each` that the root node's `data.edges` has exactly 2 entries after recovery, confirming re-link correctness beyond just counts.
- `"quarantine list and show"`: filter `LIMIT 1` to `NodeAnchor` rows — previously relied on lexicographic UUID ordering to return a node row, now that edge rows are mixed in the ordering is no longer guaranteed.

## Files changed

| File | Change |
|---|---|
| `jac-scale/jac_scale/impl/memory_hierarchy.mongo.impl.jac` | Cascade uses `bulk_write` + clearer variable names (`orphaned_edges`, `orphan_edge_ids`); re-link in `_mongo_attempt_recover` handles undirected edges |
| `jac/jaclang/runtimelib/impl/memory.impl.jac` | Cascade uses `SAVEPOINT` + batched `DELETE IN`; re-link in `_sqlite_attempt_recover` handles undirected edges |
| `jac/jaclang/runtimelib/impl/serializer.impl.jac` | `_deserialize_archetype` raises `TypeError` only when `_deserialize_value` returns `None` for a non-Optional field; other coercion misses warn and keep the value |
| `jac-scale/jac_scale/tests/test_mongo_layer123.jac` | New cascade integration test + new undirected-edge re-link recovery test |
| `jac/tests/runtimelib/test_persistence_migration.jac` | New incompatible field type quarantine test |
| `jac/tests/runtimelib/test_db_cli.jac` | Updated counts, `LIMIT 1` filter, and post-recovery `data.edges` DB assertion |

## Related Issue
#5987


akindu-k:
## Case 1: Class removed between deploys

image

image

foo2 + e2 remain live until walker traverses e2 => same process repeats

akindu-k:
## Case 2: field type changed

### Step 01
image

### Step 02
image

### Step 03
image

kugesan1105:
1. add test for Case 2: field type changed

jac-scale/jac_scale/impl/memory_hierarchy.mongo.impl.jac
```diff
@@ -152,13 +159,56 @@ def _mongo_quarantine(
...
+                    'quarantined_at': now_ts,
                     'from_format_version': from_format_version
                 }
             },
             upsert=True
         );
         backend.collection.delete_one({'_id': id_str});
+        # Cascade: when a NodeAnchor is quarantined, orphan its edges too.
+        if anchor_type == 'NodeAnchor' {
+            orphaned = list(
```
kugesan1105: change to `orphaned_edges`

kugesan1105:
Two follow-ups (not blockers):
@akindu-k 
1. **`recover_all` ordering** — [`memory.impl.jac:1438`](jac/jaclang/runtimelib/impl/memory.impl.jac#L1438) / [`memory_hierarchy.mongo.impl.jac:926`](jac-scale/jac_scale/impl/memory_hierarchy.mongo.impl.jac#L926) iterate in undefined order. If an edge recovers before its node, the re-link no-ops and `data.edges` stays empty. Process nodes first.

2. **Silent skip on missing endpoint** — when the endpoint isn't live, re-link skips with no log. Operator has no signal recovery was partial. Add a `logger.warning`.

jac-scale/jac_scale/impl/memory_hierarchy.mongo.impl.jac
akindu-k: image


### changed `orphaned` to `orphaned_edges` & `orphaned_ids` to `orphaned_edge_ids`

kugesan1105:
No more 
`"NodeAnchor [xxxxxx.xxxxxxxx.xxxxxxxxxx] is not a validreference!"` issue
`"EdgeAnchor [xxxxxx.xxxxxxxx.xxxxxxxxxx] is not a validreference!"` issue

Good job @akindu-k 👏

akindu-k:
> Two follow-ups (not blockers): @akindu-k
> 
> 1. **`recover_all` ordering** — [`memory.impl.jac:1438`](jac/jaclang/runtimelib/impl/memory.impl.jac#L1438) / [`memory_hierarchy.mongo.impl.jac:926`](jac-scale/jac_scale/impl/memory_hierarchy.mongo.impl.jac#L926) iterate in undefined order. If an edge recovers before its node, the re-link no-ops and `data.edges` stays empty. Process nodes first.
> 2. **Silent skip on missing endpoint** — when the endpoint isn't live, re-link skips with no log. Operator has no signal recovery was partial. Add a `logger.warning`.

Addressed in #6035

PR #6012, opened 2026-05-18
Add shared NGINX ingress support to jac-scale to reduce per-app controller overhead

juzailmlwork:
### Problem

jac-scale previously deployed a dedicated NGINX ingress controller per application. In clusters
that already run a shared `ingress-nginx` controller (e.g. installed via Helm), this creates
unnecessary resource duplication — a separate controller Deployment, NodePort/NLB Service,
IngressClass, ClusterRole, and ClusterRoleBinding for every app.

### Solution

Added `shared_ingress` and `shared_ingress_class` config options to `KubernetesConfig`. When
`shared_ingress = true`, jac-scale skips deploying the dedicated controller entirely and attaches
the app's Ingress routing rules to the pre-existing shared controller.

```toml
[plugins.scale.kubernetes]
shared_ingress = true
domain = "myapp.example.com"   # required in shared mode

# shared_ingress_class = "nginx"  # default

### CODE REVIEW

File: jac-scale/jac_scale/tests/test_deploy_k8s.jac [hunk 1]
```diff
@@ -251,7 +251,7 @@ test "deploy all in one" {
     print(f"Deployment result details: {details}");
     print(f"Deployment status: {result.message}");
     assert result.success is True;
-    assert len(details) == 13;
+    assert len(details) == 14;
```
udithishanka, line 255: no new tests for shared_ingress, only the existing len(details) bump. the flag branches across _wait_for_nginx_ready, _deploy_ingress_resource, start_controller, finish_ingress, _install_cert_manager, destroy, validate_ports, and _wait_for_deployment, none of it is covered. at minimum need: shared_ingress=True skips _deploy_nginx_controller, host is set immediately on the rule (mirror of the test at L2009), destroy doesn't touch controller resources, missing domain raises early at deploy time, NodePort validation is skipped.

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
```diff
@@ -1175,6 +1175,50 @@ obj KubernetesTarget(DeploymentTarget) {
...
+                            f"Health check against '{domain_url}{path}' did not succeed. "
+                            "This is expected if the A record has not been added yet or "
+                            "DNS has not propagated. The application pods are healthy. "
+                            f"Verify DNS: add an A record for '{self.k8s_config.domain}' "
+                            "pointing to the shared ingress controller's node IP."
+                        );
+                    }
+                }
+            }
+            return domain_details;
```
udithishanka, line 1225: this returns success-looking output even when the health check failed. check_deployment_status can return (False, "Timeout reached") and you only log a warning, then return domain_details to the caller, which stuffs it into details["health_check_of_deployment"]. result is deploy reports success and the details dict shows "Timeout reached" right next to it. either return an explicit marker like "Skipped: DNS not propagated yet" or thread the success bool back so callers can tell warning from failure.

jac-scale/jac_scale/targets/kubernetes/ingress.jac
```diff
@@ -691,11 +722,16 @@ obj IngressDeployer {
         core_v1: any,
         networking_v1: any
     ) {
+        if self.k8s_config.shared_ingress {
+            ingress_class = self.k8s_config.shared_ingress_class;
+        } else {
+            ingress_class = f"{namespace}-{app_name}-nginx";
```
udithishanka: this same if/else lives at L867-871 in _install_cert_manager. extract a _resolve_ingress_class(app_name, namespace) helper on IngressDeployer and call it from both sites.

udithishanka (changes requested):
2 things to address, plus dedup nit. also: PR title 'shared ingress' is too vague. something like 'Add shared NGINX ingress support to jac-scale to reduce per-app controller overhead' conveys the actual value.

jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac
juzailmlwork: this is fixed with better message

jac-scale/jac_scale/targets/kubernetes/ingress.jac
juzailmlwork: The logic is now in exactly one place. If shared_ingress_class ever gets a different default or the naming scheme changes, there's one edit to make.thanks for noting it

jac-scale/jac_scale/tests/test_deploy_k8s.jac
juzailmlwork: 3 new test cases are added covering corner cases plz check

marsninja:

PR #6016, opened 2026-05-18
fix(byllm): bind streaming generator to its own contextvars.Context (fixes cross-Context Token reset crash)

### LINKED ISSUE #6015: byllm: streaming generator crashes with 'Token was created in a different Context' when driven from a thread-pool worker (breaks jac-scale SSE)
## Summary

byllm's sync streaming path crashes when the resulting generator is driven from a thread-pool worker (which is exactly how `jac-scale`'s SSE microservice consumes it). The crash is a `ValueError: Token was created in a different Context` raised when `_wrap_generator`'s `finally` tries to reset the `byllm_invocation_id` ContextVar token.

Net effect on a running app: streaming endpoints look alive (handshake + headers succeed) but no chunks ever reach the client. UI shows empty placeholders, server logs show `ERROR - sse_frame_stream: producer raised`.

## Repro

Any `by llm()` call invoked in streaming mode (`mt_run.stream == True`), served over SSE through `jac-scale`'s microservice gateway. Concrete repro currently in production on `jac-gpt-test`: open the Jac GPT UI, send a chat message, observe nothing streams.

## Stack trace from a live pod

```
ERROR - sse_frame_stream: producer raised
  File "/app/jaseci/jac-scale/jac_scale/microservices/_sse.jac", line 89, in sse_frame_stream
    item = await loop.run_in_executor(None, _next_or_stop, gen);
  File "/usr/local/lib/python3.12/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/app/jaseci/jac-scale/jac_scale/microservices/_sse.jac", line 52, in _next_or_stop
    return next(gen);
  File "/app/services/jacServer.jac", line 11, in stream_chat_response
    ...
  File "/app/jaseci/jac-byllm/byllm/llm.jac", line 104, in _wrap_generator
    def _invoke_streaming(mt_run: MTRuntime) -> Generator[StreamEvent, None, None];
ValueError: <Token var=<ContextVar name='byllm_invocation_id' default=None at 0x7faf61f2bb00> at 0x7fad5c69dc40> was created in a different Context
```

## Root cause

`Model.invoke` in `jac-byllm/byllm/llm.impl/model.impl.jac` does:

```jac
_ctx_token = _current_invocation_id.set(mt_run._invocation_id);
...
if isinstance(result, _types_mod.GeneratorType) {
    def _wrap_generator(gen, mt, model_name, start, ctx_token) -> object {
        try {
            yield from gen;
            ...
        } finally {
            ...
            _current_invocation_id.reset(ctx_token);   # <-- bug
        }
    }
    return _wrap_generator(result, mt_run, self.model_name, _start, _ctx_token);
}
```

The token is created in the request handler's asyncio Context. `jac-scale`'s `sse_frame_stream` drives the generator via `loop.run_in_executor(None, _next_or_stop, gen)`, which runs `next(gen)` in a thread-pool worker that has its own `contextvars.Context`. When the generator's `finally` runs in that worker context, `_current_invocation_id.reset(ctx_token)` fails because tokens are scoped to the Context where `set()` was called.

This is fundamentally incompatible with the fact that the returned generator is part of byllm's public surface: callers are free to iterate it anywhere, including thread-pool executors. Token-based reset only works if the same coroutine that called `set()` calls `reset()`, in the same Context.

## When it was introduced

- `cfe7c3e32` (2026-04-17, #5478 "Telemetry for nested byllm calls") introduced the `byllm_invocation_id` ContextVar and the `_wrap_generator` cleanup pattern. The bug landed with this commit.
- `9fc5e9929` (2026-05-17, #5944 "fix(byllm): replace sync LLM clients with async equivalents") added a parallel async dispatch path (`ainvoke` / `acall_llm` / `adispatch_streaming`) but did NOT touch sync `Model.invoke`, so the bug remains in the path that `jac-scale` SSE actually uses today.

## Why the async PR didn't fix this

`jac-scale`'s microservice gateway still calls the sync `call_llm` hook (which routes to sync `Model.invoke`), not the new `acall_llm`. So adding async clients underneath doesn't help: requests still hit the sync wrapper with the broken token reset.

## Suggested fixes

Either side alone closes the symptom. Doing the byllm fix is cleaner because it heals every consumer, not just SSE.

**Option A (preferred), in `jac-byllm/byllm/llm.impl/model.impl.jac` `_wrap_generator`:**

Don't rely on token-based reset across the generator boundary. Capture the parent value up front and `set()` it back in the finally:

```jac
_parent = _current_invocation_id.get(None);
_ctx_token = _current_invocation_id.set(mt_run._invocation_id);
...
finally {
    ...
    try {
        _current_invocation_id.reset(ctx_token);
    } except ValueError {
        # Generator was driven from a different Context (e.g. thread-pool
        # executor). Best-effort: clear in current context.
        _current_invocation_id.set(_parent);
    }
}
```

**Option B, in `jac-scale/jac_scale/microservices/_sse.jac` `sse_frame_stream`:**

Capture the request Context once and run each executor task inside that Context, so the worker thread sees the same `contextvars.Context` as the producer:

```jac
import contextvars;
...
ctx = contextvars.copy_context();
while True {
    item = await loop.run_in_executor(None, ctx.run, _next_or_stop, gen);
    ...
}
```

## Impact

Every `jac-scale` deployment with a `by llm()` streaming endpoint is affected. Confirmed live on `jac-gpt-test` on jaseci-cluster (us-east-2). Severity: feature-breaking, no client-visible workaround. The user just sees an empty response.

## Environment

- `jac-byllm` upstream `main` (deployment runs with `--experimental` so it pulls latest)
- `jac-scale` upstream `main`
- python 3.12
- Cluster: jaseci-cluster (jac-gpt-test namespace), but reproducible anywhere SSE drives a sync byllm streaming generator

udithishanka:
## Summary

Fix the byllm streaming bug that broke every `jac-scale` SSE endpoint backed by `by llm(stream=True)`. The user-visible symptom: chat UIs sent a message, saw the streaming envelope arrive, then nothing. Live reproduction was on `jac-gpt-test` (jaseci-cluster).

Closes #6015.

## Root cause

`Model.invoke()` did:

```jac
_ctx_token = _current_invocation_id.set(mt_run._invocation_id);
...
return _wrap_generator(result, mt_run, self.model_name, _start, _ctx_token);
```

and the wrapper's `finally` called `_current_invocation_id.reset(_ctx_token)`.

The token is bound to the asyncio Context where `set()` was called (the request handler's). `jac-scale`'s `sse_frame_stream` drives the generator via `loop.run_in_executor(None, _next_or_stop, gen)`, so each `next(gen)` runs in a thread-pool worker with its own `contextvars.Context`. When the wrapper's `finally` runs in that worker context, `reset(_ctx_token)` raises `ValueError: Token was created in a different Context`, the SSE producer dies, and the stream collapses to zero chunks.

This is not a `jac-scale`-only bug. Token-based reset is intrinsically incompatible with returning a generator across Context boundaries, which is part of `Model.invoke`'s public surface. Any consumer that iterates the result from a different Context hits the same crash.

## Fix

Replace the token-reset pattern with a Context-bound iterator:

1. Reset the caller's token immediately on the streaming branch, so the caller's Context is restored to exactly what it was before `invoke()` returned.
2. Capture a fresh `contextvars.copy_context()` and set `byllm_invocation_id` on the captured Context.
3. Drive each iteration via `ctx.run(next, gen)`, so nested `by llm()` calls during streaming still see the right parent invocation id for telemetry.
4. The captured Context is local to the wrapper and dies with the generator. No token is ever reset across Context boundaries.

Net result: streaming generators are now safe to consume from a thread-pool executor, a different coroutine, or any other Context-foreign driver, with no behavioral change for in-Context consumers.

## Tests

New fixture `streaming_cross_context.jac` drives a MockLLM streaming result from two foreign-Context positions, mirroring exactly how `jac-scale` consumes the stream:

- `contextvars.copy_context().run(list, gen)`: directly exercises the original bug.
- `asyncio.run(...)` + `loop.run_in_executor(None, _next_or_stop, gen)`: the literal `jac-scale` SSE pattern.

Pre-fix, either path raises `ValueError: Token was created in a different Context`. Post-fix, both consume cleanly.

New test in `test_byllm.jac`: "streaming generator iterates safely from a different Context".

```
$ python -m pytest jac-byllm/tests/test_byllm.jac -k streaming
============================== 5 passed ===============================
```

## Notes

This is a byllm-side fix that heals every consumer (SSE, custom executor drivers, anything else that may iterate from a foreign Context). A defense-in-depth follow-up on the `jac-scale` side (`contextvars.copy_context()` around `run_in_executor`) is still reasonable but no longer required to keep streaming working.


marsninja:

PR #6021, opened 2026-05-19
 fix(persistence): `_put_node_atomic` clobbers scalars via shallow `$mergeObjects`

### LINKED ISSUE #5997: [jac-scale] _put_node_atomic clobbers archetype scalars with stale snapshot on every concurrent commit
# `_put_node_atomic` clobbers archetype scalars with stale snapshot when any edge change triggers the atomic-merge path

## Summary

`MongoBackend._put_node_atomic` (introduced in PR #5644 to atomically merge edge changes from concurrent walkers) writes the **entire** non-edge serialized archetype back to Mongo on every commit that goes through this path. Because `$mergeObjects` does a shallow merge on the `data` field, the `data.archetype` subdocument is fully replaced by the committing walker's snapshot version. Any scalar a sibling walker committed in between this walker's load and commit is silently overwritten back to the stale value.

The path is taken whenever `anchor.edge_delta()` returns non-None, which is true for **any** NodeAnchor that was loaded from the database (even if the walker added or removed zero edges). So in practice this path is hit on almost every commit to an existing node, not just edge mutations.

PR #5644 advertised "atomic edge-merge preserves edge changes from sibling walkers". It does protect the `data.edges` array (the `$setUnion`/`$setDifference` pipeline is correct). It does not protect any other archetype field.

Severity: high. The bug fires when a walker's L1 cache is stale relative to mongo. That happens deterministically in any multi-pod / multi-process jac-scale deployment, because each process has its own L1 cache and there is no cross-process L1 invalidation. Within a single process, `ExecutionContext` forks share the same Memory backend (see `jac/jaclang/runtimelib/impl/server.impl.jac:260-275`), so this race does not fire in a single-process test.

## Repro

Two reproducers below. The first is a 70-line pipeline-level proof that runs in 2 seconds against a vanilla `mongo:7.0` container and isolates the bug to the exact aggregation pipeline used by `_put_node_atomic`. The second is a full end-to-end test through the jac runtime in the existing `test_persistence_race.jac` style.

### Repro 1: pipeline-level proof (verified, run in 2s against `mongo:7.0`)

```python
"""Pipeline-level repro of jac-scale _put_node_atomic clobber bug.

Demonstrates that the exact aggregation pipeline used by
MongoBackend._put_node_atomic (impl/memory_hierarchy.mongo.impl.jac:510-538)
overwrites archetype scalars even when only ONE field changed.

Bypasses the jac runtime entirely to isolate the bug at the
MongoDB layer: shallow $mergeObjects on the `data` dict replaces
`data.archetype` wholesale.
"""

from pymongo import MongoClient

client = MongoClient("mongodb://localhost:27018/")
db = client["repro_db"]
coll = db["_anchors"]
coll.drop()


def put_node_atomic(_id, data_no_edges, edges_added, edges_removed):
    """Exact pipeline copied from jac-scale impl/memory_hierarchy.mongo.impl.jac:510."""
    coll.update_one(
        {"_id": _id},
        [
            {
                "$set": {
                    "data": {
                        "$mergeObjects": [
                            "$data",
                            data_no_edges,
                            {
                                "edges": {
                                    "$setUnion": [
                                        {
                                            "$setDifference": [
                                                {"$ifNull": ["$data.edges", []]},
                                                edges_removed,
                                            ]
                                        },
                                        edges_added,
                                    ]
                                }
                            },
                        ]
                    },
                    "type": "NodeAnchor",
                }
            }
        ],
        upsert=True,
    )


_id = "counter-node-1"
coll.insert_one(
    {
        "_id": _id,
        "type": "NodeAnchor",
        "data": {
            "archetype": {"a": 0, "b": 0, "name": "counter"},
            "edges": ["edge-existing-1"],
        },
    }
)

# Step 1: Walker A loads snapshot, sets a=42, commits (no edge change).
walker_a_snapshot = {"archetype": {"a": 42, "b": 0, "name": "counter"}}
put_node_atomic(_id, walker_a_snapshot, edges_added=[], edges_removed=[])
assert coll.find_one({"_id": _id})["data"]["archetype"]["a"] == 42

# Step 2: Walker B was loaded BEFORE A committed. B's snapshot still has a=0.
# B sets b=99 and adds an edge.
walker_b_snapshot = {"archetype": {"a": 0, "b": 99, "name": "counter"}}
put_node_atomic(
    _id,
    walker_b_snapshot,
    edges_added=["edge-b-new"],
    edges_removed=[],
)

after = coll.find_one({"_id": _id})["data"]
print(after)
assert after["archetype"]["a"] == 42, f"A's write clobbered: a={after['archetype']['a']}"
assert after["archetype"]["b"] == 99
```

Run:
```bash
docker run --rm -d --name mongo-repro -p 27018:27017 mongo:7.0
python repro_pipeline_level.py
```

Verified output:
```
{'archetype': {'a': 0, 'b': 99, 'name': 'counter'}, 'edges': ['edge-b-new', 'edge-existing-1']}
AssertionError: A's write clobbered: a=0
```

Note the edges are merged correctly (both `edge-existing-1` and `edge-b-new` are present), confirming PR #5644's edge logic is sound. The archetype scalar `a` was reverted from 42 back to 0 by B's stale snapshot, even though B never touched field `a`.

### Repro 2: end-to-end through the jac runtime (cross-process)

The bug fires when walker L1 caches are stale relative to mongo. Within a single process, concurrent `ExecutionContext` forks share the same Memory backend via `_parent` (see `jac/jaclang/runtimelib/impl/server.impl.jac:260-275`), so a single-process two-thread test does NOT trigger this. The realistic trigger is multi-pod, where each pod has its own L1 cache. The end-to-end repro therefore runs **two separate `jac start` processes** against the same Mongo+Redis (mirroring a multi-pod deployment).

Walker A on pod 1 writes scalar `a`. Walker B on pod 2 (which loaded the node before A's commit lands in its local L1) writes scalar `b` and adds one edge. Expected: both persist. Actual: `a` is reverted to the stale value pod 2 had cached.

#### `fixtures/clobber_app.jac`

```jac
"""Repro fixture: two walkers race on the same Counter node.
SetA writes counter.a. SetBAndAddEdge writes counter.b AND adds an Item child.
After both commit, both .a and .b should be set. On current main, .a is lost."""

node Item {
    has name: str = "";
}

node Counter {
    has a: int = 0;
    has b: int = 0;
}

walker:pub Init {
    can create with Root entry {
        existing = [-->(?:Counter)];
        if not existing {
            here ++> Counter();
        }
        report {"ok": True};
    }
}

walker:pub SetA {
    has value: int = 0;

    can write with Root entry {
        counters = [-->(?:Counter)];
        counters[0].a = self.value;
        report {"ok": True, "a": counters[0].a};
    }
}

walker:pub SetBAndAddEdge {
    has value: int = 0;
    has item_name: str = "x";

    can write with Root entry {
        counters = [-->(?:Counter)];
        counters[0].b = self.value;
        here ++> Item(name=self.item_name);
        report {"ok": True, "b": counters[0].b};
    }
}

walker:pub ReadCounter {
    can read with Root entry {
        counters = [-->(?:Counter)];
        c = counters[0];
        report {"a": c.a, "b": c.b};
    }
}
```

#### `test_scalar_clobber_cross_process.jac`

```jac
"""Cross-process proof: two jac-start processes (simulating two pods)
share the same Mongo+Redis. Pod 1's walker loads the Counter and caches
it. Pod 2 commits a=42 to mongo. Pod 1's L1 cache still has a=0.
Pod 1 then writes b=99 + adds an edge, which on commit serializes its
stale archetype and clobbers a back to 0 via _put_node_atomic's shallow
$mergeObjects."""

import os;
import socket;
import subprocess;
import sys;
import time;
import from pathlib { Path }
import from typing { Any, cast }

import requests;
import from testcontainers.mongodb { MongoDbContainer }
import from testcontainers.redis { RedisContainer }

glob FIXTURES_DIR = Path(__file__).parent / "fixtures",
     JAC_FILE = FIXTURES_DIR / "clobber_app.jac";

def get_free_port -> int {
    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s {
        s.bind(("", 0));
        return s.getsockname()[1];
    }
}

def extract(json_response: Any) -> Any {
    if isinstance(json_response, list) and len(json_response) == 2 {
        json_response = json_response[1];
    }
    if isinstance(json_response, dict) and "data" in json_response {
        return json_response["data"];
    }
    return json_response;
}

def register_and_login(base_url: str, username: str) -> str {
    res = requests.post(
        f"{base_url}/user/register",
        json={
            "identities": [{"type": "username", "value": username}],
            "credential": {"type": "password", "password": "pass123"}
        },
        timeout=5
    );
    assert res.status_code == 201;
    login = requests.post(
        f"{base_url}/user/login",
        json={
            "identity": {"type": "username", "value": username},
            "credential": {"type": "password", "password": "pass123"}
        },
        timeout=5
    );
    return cast(dict, extract(login.json()))["token"];
}

def login(base_url: str, username: str) -> str {
    login = requests.post(
        f"{base_url}/user/login",
        json={
            "identity": {"type": "username", "value": username},
            "credential": {"type": "password", "password": "pass123"}
        },
        timeout=5
    );
    return cast(dict, extract(login.json()))["token"];
}

def call_walker(
    base_url: str, walker_name: str, token: str, payload: dict | None = None
) -> Any {
    res = requests.post(
        f"{base_url}/walker/{walker_name}",
        json=payload or {},
        headers={"Authorization": f"Bearer {token}"},
        timeout=15
    );
    return res.json();
}

def start_pod(fixtures_dir: Path, jac_file: Path, port: int, env_extra: dict) -> subprocess.Popen {
    import shutil;
    jac_executable = shutil.which("jac") or str(Path(sys.executable).parent / "jac");
    cmd = [str(jac_executable), "start", str(jac_file.name), "--port", str(port)];
    env = os.environ.copy();
    env.update(env_extra);
    server = subprocess.Popen(
        cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE,
        text=True, cwd=str(fixtures_dir), env=env
    );
    base_url = f"http://localhost:{port}";
    for _ in range(30) {
        try {
            r = requests.get(f"{base_url}/healthz", timeout=1);
            if r.status_code == 200 {
                return server;
            }
        } except Exception {
            time.sleep(1);
        }
    }
    raise RuntimeError("pod failed to start");
}

test "cross-process: scalar write from pod with stale L1 clobbers other pod's write" {
    with (
        MongoDbContainer("mongo:7.0") as mongo,
        RedisContainer("redis:7-alpine") as redis_c
    ) {
        env_shared = {
            "JAC_PERSISTENCE": "mongo",
            "JAC_MONGO_HOST": mongo.get_connection_url(),
            "JAC_REDIS_HOST": redis_c.get_container_host_ip(),
            "JAC_REDIS_PORT": str(redis_c.get_exposed_port(6379)),
        };

        port1 = get_free_port();
        port2 = get_free_port();
        url1 = f"http://localhost:{port1}";
        url2 = f"http://localhost:{port2}";

        pod1 = start_pod(FIXTURES_DIR, JAC_FILE, port1, env_shared);
        pod2 = start_pod(FIXTURES_DIR, JAC_FILE, port2, env_shared);

        try {
            user = f"user_{int(time.time())}";
            token1 = register_and_login(url1, user);
            token2 = login(url2, user);

            # 1. Init the Counter via pod1.
            call_walker(url1, "Init", token1);

            # 2. Warm pod2's L1 cache by reading the Counter via pod2.
            #    Pod2 now has Counter{a:0, b:0} in its in-memory _mem_ref.
            call_walker(url2, "ReadCounter", token2);

            # 3. Pod1 writes a=42 to mongo. Pod2's L1 cache is NOT invalidated.
            call_walker(url1, "SetA", token1, {"value": 42});

            # 4. Pod2 writes b=99 AND adds an edge. Pod2's in-memory anchor
            #    still has the stale a=0. On commit, _put_node_atomic
            #    serializes pod2's full archetype {a:0, b:99} and ships it
            #    through $mergeObjects, wholesale-replacing mongo's archetype.
            call_walker(url2, "SetBAndAddEdge", token2, {"value": 99, "item_name": "x"});

            # 5. Read final state via a fresh process or by clearing pod1's
            #    cache. Easiest: read via pod1 after a brief wait (its cache
            #    will be evicted on next miss), or we can read directly.
            time.sleep(0.5);
            final = extract(call_walker(url1, "ReadCounter", token1))["reports"][0];

            assert final["a"] == 42, f"a clobbered by pod2 stale snapshot: got {final['a']}";
            assert final["b"] == 99, f"b lost: got {final['b']}";
        } finally {
            for pod in [pod1, pod2] {
                pod.terminate();
                try { pod.wait(timeout=5); } except Exception { pod.kill(); }
            }
        }
    }
}
```

Run with: `pytest jac-scale/jac_scale/tests/test_scalar_clobber_cross_process.jac`.

Caveat: this end-to-end test is illustrative of the production trigger but I did not run it as part of the verification (Docker spinning of testcontainers, two `jac start` processes, etc.). Repro 1 is the authoritative proof of the bug at the layer where it actually lives.

## Expected vs Actual

Expected on a correct implementation:

```
{"a": 42, "b": 99}
```

Actual on `main` (jac-scale 0.2.x, jaclang 0.12.x), verified via Repro 1:

```
{"a": 0, "b": 99}     # A's write was clobbered by B's stale snapshot
```

Note: both walkers go through `_put_node_atomic` because `edge_delta()` returns non-None for any NodeAnchor that was loaded from the DB (i.e., has an `_initial_edge_ids` snapshot), even if the walker did not change any edges. So whichever walker commits second always wins for all archetype fields.

## Root cause

In `jac-scale/jac_scale/impl/memory_hierarchy.mongo.impl.jac:493-544`:

```jac
impl MongoBackend._put_node_atomic(
    anchor: NodeAnchor, edges_added: list[str], edges_removed: list[str]
) -> None {
    ...
    data: dict[str, object] = dict(Serializer.serialize(anchor, include_type=True));
    data_no_edges = {
        k: v for (k, v) in data.items() if k != 'edges'
    };
    self.collection.update_one(
        {'_id': _id},
        [
            {
                '$set': {
                    'data': {
                        '$mergeObjects': [
                            '$data',
                            data_no_edges,      # <-- entire archetype, including stale fields
                            {
                                'edges': {
                                    '$setUnion': [
                                        {
                                            '$setDifference': [
                                                {'$ifNull': ['$data.edges', []]},
                                                edges_removed
                                            ]
                                        },
                                        edges_added
                                    ]
                                }
                            }
                        ]
                    },
                    ...
                }
            }
        ],
        upsert=True
    );
    ...
}
```

Two compounding issues:

1. **`$mergeObjects` is shallow.** It merges the top-level keys of `data`, so `data.archetype` from `data_no_edges` (which is the walker's snapshot of every archetype field) wholesale-replaces `data.archetype` from `$data` (the current Mongo state). Any field the sibling walker just committed is lost.

2. **The atomic path is taken on every node-anchor commit, not just edge mutations.** The dispatcher in `MongoBackend.sync` (line 458-465) calls this whenever `anchor.edge_delta()` returns non-None, and `edge_delta()` returns non-None for any node that was loaded from the DB (i.e., `_initial_edge_ids` snapshot exists), regardless of whether edges actually changed. So a walker that only modifies a scalar still goes through this path and clobbers sibling scalars.

The dispatcher in `sync` does have a fast-path that uses `put_partial` to write only dirty archetype fields, but it is unreachable for any node that was loaded from the DB because `_put_node_atomic` is checked first and `delta is not None` is always true for loaded nodes.

## Suggested fix

Make `_put_node_atomic` use the dirty-field approach in the same pipeline as the edges merge, instead of writing the whole archetype:

```jac
impl MongoBackend._put_node_atomic(
    anchor: NodeAnchor, edges_added: list[str], edges_removed: list[str]
) -> None {
    if self.client is None {
        return;
    }
    _id = str(to_uuid(anchor.id));
    try {
        # Compute the dirty archetype field set, same as put_partial does.
        arch = anchor.archetype if anchor.is_populated() else None;
        setattr_dirty: set[str] = (
            arch.__dict__.get('__jac_dirty_fields__') or set()
        ).copy() if arch is not None else set();
        derived = _derive_dirty_from_field_hashes(anchor);
        dirty = setattr_dirty | derived;

        partial_data = (
            Serializer.serialize_fields(anchor, dirty) if dirty else {}
        );

        # Top-level NodeAnchor fields (type, access, etc.) that aren't archetype.
        full = dict(Serializer.serialize(anchor, include_type=True));
        non_arch_top = {
            k: v for (k, v) in full.items()
            if k not in ('edges', 'archetype')
        };

        set_doc: dict[str, object] = {};
            set_doc[f"data.{k}"] = v;
        }
        for (k, v) in partial_data.items() {
            set_doc[f"data.archetype.{k}"] = v;
        }
        set_doc["type"] = "NodeAnchor";
        set_doc["data.edges"] = {
            '$setUnion': [
                {
                    '$setDifference': [
                        {'$ifNull': ['$data.edges', []]},
                        edges_removed
                    ]
                },
                edges_added
            ]
        };

        self.collection.update_one(
            {'_id': _id},
            [{'$set': set_doc}],
            upsert=True
        );
        anchor.hash = Serializer._compute_hash(anchor);
        if arch is not None {
            arch.__dict__.pop('__jac_dirty_fields__', None);
        }
        _snapshot_field_hashes(anchor);
    } except Exception as e {
        logger.error(f"MongoDB _put_node_atomic failed: {e}");
    }
}
```

Key change: build the update doc as per-field `$set` with `data.archetype.<dirty_field>` keys (matching `put_partial`'s safe path), and combine it with the existing atomic edges pipeline in the same `update_one` call. This preserves the atomicity guarantee from PR #5644 for edges while extending the same isolation to scalars. Non-dirty archetype fields are never touched in Mongo, so concurrent commits from sibling walkers cannot clobber each other's scalar writes.

Note: this also obsoletes the separate `put_partial` path in `sync` for loaded NodeAnchors. The atomic version handles both cases (with or without edge deltas) correctly.

## Real-world impact (jacBuilder)

This bug surfaces in the jac-builder web IDE (https://github.com/jaseci-labs/jacBuilder) as "chat history disappears mid-stream" and "AI processing spinner stuck on after the agent finishes". Reproduction in production: open the AI chat panel, send a long-running build request, navigate around the IDE while the agent runs. The `preview_control:start` and `ide_file_ops:write` walkers commit edge changes on the same `Project` node concurrently with `ai_chat:persist_done`, which writes the scalar `ai_processing=False` plus a new `JacCoderMessage` edge. The `persist_done` archetype write loses to whichever sibling commits last: `ai_processing` reverts to True (spinner stuck), `last_commit` and `updated_at` revert (file list out of sync). Filed downstream: https://github.com/jaseci-labs/jacBuilder/issues/337 (related).

## Environment

- Reproduced on: jac-scale main as of 2026-05-16, jaclang main as of 2026-05-16
- Backend: `MongoBackend` with real MongoDB 7.0 and Redis 7
- PR that introduced the atomic path: #5644
- Related issues: #4616, #5446, #5451, #5475 (all reported variants of this race; #5644 partially addressed it for edges only)

### LINKED ISSUE #6022: fix(persistence): sync() routes scalar-only commits through _put_node_atomic instead of put_partial
## Summary

`MongoBackend.sync()` dispatches to `_put_node_atomic` for **every** DB-loaded `NodeAnchor`, even when no edges changed. This means the simpler, well-tested `put_partial` path is effectively unreachable for any node that was previously saved to the database.

## Root Cause

`anchor.edge_delta()` returns non-None (a tuple of two sets) for any node that was loaded from the database, because loading snapshots `_initial_edge_ids`. When no edges changed, it returns `(empty_set, empty_set)` — not `None`. The dispatch only checks `is not None`:

```jac
delta = anchor.edge_delta();
if delta is not None {          // true even when added=[], removed=[]
    self._put_node_atomic(...); // always taken for DB-loaded nodes
    continue;                   // put_partial never reached
}
```

So a walker that only modifies scalars (no edge mutations) still goes through `_put_node_atomic` — a heavier aggregation pipeline update — and bypasses `put_partial` entirely.

## Correct Dispatch

```jac
delta = anchor.edge_delta();
if delta is not None {
    added = [str(eid) for eid in delta[0]];
    removed = [str(eid) for eid in delta[1]];
    if added or removed {           // only route atomic when edges actually changed
        self._put_node_atomic(anchor, added, removed);
        continue;
    }
    // no edge changes — fall through to put_partial for scalar-only writes
}
```

This restores the intended routing:

| Situation | Correct path |
|-----------|-------------|
| New node (never saved) | `_write_to_db` — full upsert |
| DB-loaded, scalars only changed | `put_partial` — simple `$set` |
| DB-loaded, edges changed (with or without scalar changes) | `_put_node_atomic` — atomic pipeline |

## Relationship to #5997

PR #6021 fixes the scalar clobber inside `_put_node_atomic` (the `$mergeObjects` → per-field dot-notation fix). That fix is still necessary — when edges AND scalars change together, `_put_node_atomic` is the correct path and must use per-field dot-notation. This issue is a separate dispatch-level bug: scalar-only commits should never reach `_put_node_atomic` in the first place.

## Impact

- Every scalar-only commit on a DB-loaded node runs an aggregation pipeline update unnecessarily instead of a plain `$set`.
- `put_partial` is dead code for the most common walker pattern (read node, mutate a field, no edge changes).

## File

`jac-scale/jac_scale/impl/memory_hierarchy.mongo.impl.jac` — `MongoBackend.sync` (the dispatch block around line 506–515).

akindu-k:
# fix(persistence): `_put_node_atomic` clobbers scalars via shallow `$mergeObjects`

## Problem

`MongoBackend._put_node_atomic` (introduced in #5644) serialized the **entire** archetype and merged it into Mongo via `$mergeObjects`. Because `$mergeObjects` is a shallow merge, `data.archetype` was replaced wholesale on every commit that went through this path.

In a multi-pod deployment each pod has its own L1 cache with no cross-pod invalidation. When Walker B commits after Walker A — even if B never touched the fields A wrote — B's stale snapshot overwrites A's values:

```
t=0   Pod 1 loads Counter {a:0, b:0}
t=0   Pod 2 loads Counter {a:0, b:0}         ← stale cache, never invalidated
t=1   Pod 1 writes a=42 → Mongo: {a:42, b:0}
t=2   Pod 2 writes b=99 → Mongo: {a:0,  b:99} ← A's write silently lost
```

The edge merge logic from #5644 (`$setUnion`/`$setDifference`) was correct and is untouched. Only archetype scalars were broken.

**Severity:** high — fires deterministically under any multi-pod load where two walkers touch the same node concurrently, which is the normal operating condition for jac-scale.

Closes #5997. Closes #6022.

## Root cause

Two compounding issues:

**1. `$mergeObjects` is shallow.**
It merges top-level keys of `data`, so `data.archetype` from the committing walker's snapshot replaces `data.archetype` in Mongo entirely — any field a sibling committed in between is reverted.

**2. The atomic path was taken on every DB-loaded node commit, not just edge mutations.**
`anchor.edge_delta()` returns non-None for any node loaded from the database (loading always snapshots `_initial_edge_ids`). When no edges changed it returns `(empty_set, empty_set)` — not `None`. The old dispatch only checked `is not None`, so `_put_node_atomic` was always reached and the safe `put_partial` path was effectively dead code for existing nodes.

## Fix

Two changes in `memory_hierarchy.mongo.impl.jac`:

### 1. `_put_node_atomic` — per-field dot-notation instead of full archetype blob

Replace the full-archetype `$mergeObjects` with per-field `data.archetype.<field>` dot-notation keys, writing only the dirty field set — the same approach `put_partial` uses — combined into the same aggregation pipeline `update_one` that handles the edge delta atomically.

Dirty field computation uses the same union already present in `sync()`:
- `__jac_dirty_fields__` — `__setattr__`-tracked direct assignments
- `_derive_dirty_from_field_hashes` — hash-derived in-place mutations

```
Before (broken)                          After (fixed)
───────────────────────────────────────  ────────────────────────────────────────
$set: {                                  $set: {
  data: $mergeObjects([                    type: "NodeAnchor",
    $data,                                 updated_at: <now>,
    { archetype: {a:0, b:99, ...} },  ←   data.access: <access>,
    { edges: $setUnion(...) }              data.archetype.b: 99,      ← dirty only
  ])                                       data.edges: $setUnion(...)
}                                        }
      ↑ replaces archetype wholesale            ↑ a is never touched
```

Additional fixes in `_put_node_atomic`:
- `__jac_dirty_fields__` is now cleared after a successful write (previously only `put_partial` cleared it, leaving the flag stale).
- `updated_at` is now stamped on the document (previously missing from this path).

### 2. `sync()` dispatch — only route to `_put_node_atomic` when edges actually changed

```jac
# Before
if delta is not None {                    # true for ALL DB-loaded nodes
    self._put_node_atomic(...);
    continue;
}

# After
if delta is not None {
    added = [...]; removed = [...];
    if added or removed {                 # only when edges actually changed
        self._put_node_atomic(...);
        continue;
    }
    # scalar-only commits fall through to put_partial
}
```

The routing is now correct:

| Node state | Path |
|---|---|
| New node (never saved) | `_write_to_db` — full upsert |
| DB-loaded, edges changed (± scalars) | `_put_node_atomic` — atomic pipeline |
| DB-loaded, scalars only changed | `put_partial` — simple `$set` |

## Tests

Two cross-process regression tests in `test_scalar_clobber_cross_process.jac`, each spinning up two real `jac start` pods sharing a Mongo + Redis testcontainer. This exercises the full code path — dirty-field computation, serialization, and the MongoDB pipeline — rather than re-implementing the pipeline in test code.

**Test 1 — `_put_node_atomic` dirty-only writes (issue #5997)**

Reproduces the stale-snapshot clobber on the atomic edge-mutation path:

1. Pod1 creates `Counter {a:0, b:0}` and warms its own L1.
2. Pod2 reads `Counter` — now has stale `{a:0, b:0}` in its L1.
3. Pod1 commits `a=42` to Mongo. Pod2's L1 is not invalidated.
4. Pod2 commits `b=99` and adds an edge (dirty set: `{b}` only) — triggers `_put_node_atomic`.
5. Assert `a=42`, `b=99` — pod1's write must not be clobbered.

**Test 2 — dispatch guard (issue #6022)**

Verifies that scalar-only commits no longer reach `_put_node_atomic`:

1–3. Same setup as Test 1.
4. Pod2 commits `b=99` with **no edge change** (dirty set: `{b}` only) — must route to `put_partial`.
5. Assert `a=42`, `b=99` — the stale archetype snapshot must not overwrite pod1's write.

Both tests fail against the pre-fix code and pass with the fix.

## Files changed

| File | Change |
|------|--------|
| `jac-scale/jac_scale/impl/memory_hierarchy.mongo.impl.jac` | Rewrote `_put_node_atomic` (per-field dot-notation); fixed `sync()` dispatch guard |
| `jac-scale/jac_scale/tests/test_scalar_clobber_cross_process.jac` | Cross-process regression tests for #5997 and #6022 (two real pods, real Mongo + Redis) |
| `jac-scale/jac_scale/tests/fixtures/clobber_app.jac` | Fixture app for the cross-process tests |
| `.github/workflows/test-jaseci.yml` | Registered new test file in CI |
| `docs/docs/community/release_notes/unreleased/jac-scale/6021.bugfix.md` | Release note fragment |


akindu-k:
## Problem

image

akindu-k:
## Follow-up gaps (non-blocking)

### 1. `non_arch_top` stale-snapshot risk
`_put_node_atomic` always writes `data.access` (and any other top-level non-archetype data
fields) from the committing pod's snapshot, even if those fields are unchanged. Under concurrent
access-control mutations this is the same class of clobber that this PR fixes for archetype
scalars. Pre-existing and not worsened here, but should be extended to use the same dirty-only
dot-notation approach.

**Fix:** compute a dirty set for top-level fields (mirroring the archetype dirty logic) and
write only `data.<field>` keys that have actually changed.

#### Issue #6027 
---

### 2. `acommit` async path still does full-document writes
`MongoBackend.acommit` uses `_anchor_to_doc` → `UpdateOne $set` — a full document replace.
The same shallow-merge clobber risk exists on the async code path; the fix in `_put_node_atomic`
and `sync()` does not cover it.

**Fix:** give `acommit` the same per-field dot-notation + edge-delta pipeline that
`_put_node_atomic` now uses, or route it through `sync()`.

#### Issue #6028

PR #6023, opened 2026-05-19
fix(scale): webhook security hardening

kashmithnisakya:
## Summary

Security hardening for the jac-scale webhook flow. Tightens how api keys are issued, signed, scoped, and validated, and makes the webhook callback safe against signature forgery, replay, mass assignment, traceback leaks, oversized bodies, and stale config.

## What changed

- **Per-key signing secret.** `signing_secret` is generated on key creation, returned **once**, and stored alongside the key. HMAC verification uses the signing secret, not the api key.
- **Replay protection.** Requests must include `X-Webhook-Timestamp`; signatures cover `f"{timestamp}." + body`. Timestamps outside `replay_tolerance_seconds` are rejected.
- **Scoped api keys.** `allowed_walkers` on `/api-key/create` restricts a key to specific webhook walkers. Empty/omitted means any walker.
- **Mass-assignment filter.** The webhook callback only forwards body fields that are declared on the target walker.
- **Stable user identity.** Keys are keyed by `user_id` (stable) rather than `username` (mutable/optional). JWT `sub` carries the user id.
- **Fail-closed lookup.** A key whose record is missing or revoked is rejected, even if the JWT is otherwise valid.
- **Audit log.** Every webhook outcome (accept or reject) is logged once with walker, client IP, and status.
- **Generic error responses.** Walker failures return a generic 500 to the caller; full traceback is logged server-side only.
- **Body size + rate limit.** Configurable `max_body_bytes` (413) and per-key `rate_limit_per_minute` (429); both default-on/safe.
- **L2 guard.** `validate_jwt_token` rejects tokens whose `type` is `api_key`, so an api key can never authenticate as a user session.
- **Config at call time.** JWT secret and webhook config are read on each call, so rotation does not require a restart.

## Example

See [`jac-scale/examples/webhook-receiver/`](jac-scale/examples/webhook-receiver/) for a minimal signed-webhook flow (`app.jac`, `jac.toml`).

## Tests

`jac-scale/jac_scale/tests/test_webhook.jac` covers:
- requires api key / invalid api key / revoked api key (fail-closed)
- scoped api key (allowed walker accepted, others 403)
- signed payload with `signing_secret` and timestamp
- stale timestamp rejected (replay window)
- unknown body field dropped (mass-assignment filter)
- api key rejected as user-session Bearer token

Each test runs both in-process (file-backed) and with MongoDB via testcontainers.

## Docs

`jac-scale/jac_scale/docs/webhook-guide.md` updated for the new signing flow, `allowed_walkers`, replay window, and config table.


kugesan1105:
```
API Key 
   ↓
Signature 
   ↓
Timestamp 
   ↓
Allowed Walker 
   ↓
Payload Validation 
   ↓
Rate/Size Limits 
   ↓
✅ ACCEPTED
```

kugesan1105:
@kashmithnisakya  
-  **investigate abt execution timeout** :an externally-triggerable endpoint(slow or hung walker code) runs unbounded work on a shared, finite resource pool
- duplicate handling : (eg: charging twice)
    consider Idempotency Key (Event ID) + Timestamp / Expiry Window

PR #6025, opened 2026-05-19
feat(client): polymorphic asset base resolution on ClientTarget & runtime subpath injection

kiptuidenis:
## Summary
This Pull Request introduces asset resolution architecture for Jac Web and Desktop applications. It enables applications to serve assets correctly under arbitrary configurations, ranging from standard local servers and offline desktop webviews to multi-segment routes behind dynamic reverse proxies (e.g., Nginx, Cloudflare, Traefik), using the same build artifact.

---

### **1: Configurable Static Base Path**
* **Goal**: Enable offline SPA compilations to use a customized prefix for JS bundles, CSS sheets, and manifests during build time.
* **Key Additions**:
  * Added the `static_base_path` configuration key under the `[plugins.client]` section in `jac.toml` (fully integrated into `JacClientConfig`).
  * Updated the offline builder to read `static_base_path` and feed it into `vite.config.js` (`base: ...`), ensuring static builds align with custom subpaths.
  * Implemented an offline-build fallback that defaults compile-time resolutions to root `/` if no custom path is configured.

### **2: Polymorphic Client Target Asset Resolvers**
* **Goal**: Abstract the path resolution rules into independent targets to enforce clean architectural boundaries.
* **Key Additions**:
  * Created polymorphic handlers (`WebTarget`, `PWATarget`, and `DesktopTarget`) that inherit from a unified `ClientTarget` abstract base class.
  * **`WebTarget` & `PWATarget`**: Read and normalize ingress forwarding headers (such as `X-Forwarded-Prefix` or `x-forwarded-prefix`) to compute the target base path dynamically.
  * **`DesktopTarget`**: Bypasses proxy-routing headers and strictly resolves base paths to root `/` to maintain the offline security context of Tauri v2 webviews.

### **3: Runtime Server Integration & Request Context Propagation**
* **Goal**: Intercept incoming request headers at the server level, dynamically inject the `<base>` tag, and resolve assets on-the-fly.
* **Key Additions**:
  * Extended core signatures (`Jac.render_page` and `ModuleIntrospector.render_page` in `jaclang`) to support an optional `request_ctx` dictionary argument.
  * Updated the request handler (`JacRequestHandler._render_page` in `server.impl.jac`) to capture incoming HTTP headers and propagate them down the rendering pipeline.
  * Programmed `render_page` to dynamically resolve `base_prefix` using the active target and prepend it to all static asset links while injecting the `<base href="...">` tag into the HTML `<head>`.
  * **Backwards Compatibility**: Ensured that standard local/CI renders continue to resolve default assets to `/static/...` exactly as before, leaving existing test suites fully functional.

---

## Verification & Stability
* **Polymorphic Unit Tests**: Validated Web, Desktop, and PWA targets under different header contexts (empty, case-insensitive headers, subpaths, ignored scenarios).
* **HTML E2E Verification**: Added assertions confirming proper `<base href>` tag injection and dynamically prefixed script path outputs.
* **Regression-Free**: Verified that existing SPA and multi-segment build tests continue to pass.


jac-client/jac_client/plugin/impl/client.impl.jac
```diff
@@ -132,25 +146,33 @@ impl JacClient.render_page(
     if css_link {
         head_content += f"\n            {css_link}";
     }
+
+    # Inject base href tag if resolved_base is present
+    if resolved_base {
+        head_content = f'<base href="{resolved_base}">\n            {head_content}';
```
MusabMahmoodh: this is the XSS sink. `resolved_base` comes straight from the request header (`X-Forwarded-Prefix`) via `WebTarget.resolve_asset_base` and goes into html with no escaping or validation no?

attacker sends:
```
X-Forwarded-Prefix: "><script>alert(document.cookie)</script><base x="
```
and gets reflected XSS. or `X-Forwarded-Prefix: //evil.com/` and **every** asset (`{base_prefix}static/client.js`, `manifest.json`, `sw.js`) loads from `evil.com` - full client-side takeover including the service worker which persists across reloads even after the attack page is gone.

minimum fix:
1. strict format validation regex like `^/[a-zA-Z0-9_\-/]*$`, reject anything else
2. `html.escape` on the way out before injection
3. opt-in flag `[plugins.client] trust_forwarded_headers = true` defaulting to **false** so the secure path is the default
4. add test cases with malicious payloads - current tests only cover happy path

also worth noting seperately - the `<base href>` tag has wide side effects, it re-resolves *every* relative url in the document (fetch() calls in client js, `<a href>`, `<form action>` etc). apps that assume root-relative urls will silently get rewritten. worth a line in `breaking-changes.md` becouse this affects users who never touched `static_base_path` themselves but happen to be behind a proxy.

jac-client/jac_client/plugin/impl/client.impl.jac
```diff
@@ -119,11 +120,24 @@ impl JacClient.render_page(
     } else {
         bundle_hash = introspector.ensure_bundle();
     }
+
+    # Resolve dynamic asset base path via active ClientTarget
+    import from jac_client.plugin.src.targets.registry { get_target_factory }
+    factory = get_target_factory();
+    target = factory.get_default(base_path);
```
MusabMahmoodh, line 127: if `factory.get_default(base_path)` returns None (no targets registered yet, some test contexts, plugin load order issue) the next line `target.resolve_asset_base(headers)` TypeErrors. atleast a defensive check + fallback to `"/"` so `render_page` doesnt crash on a degenerate state.

jac-client/jac_client/plugin/impl/client.impl.jac
```diff
@@ -162,7 +184,7 @@ impl JacClient.render_page(
     };
     init_json = json.dumps(init_data).replace('</', '<\\/');
     init_script = f'<script id="__jac_init__" type="application/json">{init_json}</script>';
-    page = f'<!DOCTYPE html><html lang="en"><head>{head_content}{pwa_head}</head><body><div id="root"></div>{init_script}{pwa_body}<script type="module" src="/static/client.js?hash={bundle_hash}"></script></body></html>';
+    page = f'<!DOCTYPE html><html lang="en"><head>{head_content}{pwa_head}</head><body><div id="root"></div>{init_script}{pwa_body}<script type="module" src="{base_prefix}static/client.js?hash={bundle_hash}"></script></body></html>';
```
MusabMahmoodh, line 199: nit but worth fixing: `pwa_head` above is an f-string but `pwa_body` here is plain string + `.replace('{base_prefix}', base_prefix)`. inconsistent and the `.replace` is fragile - it substitutes *every* occurance of `{base_prefix}` literal, including if it ever showed up in user-controlled data passed through here somehow. just use an f-string like the surrounding code.

jac-client/jac_client/plugin/src/targets/impl/web_target.impl.jac
```diff
@@ -247,19 +252,30 @@ def _generate_index_html(
...
-    <script type="module" src="{base_path}{bundle_filename}"></script>
+    <script type="module" src="{src_path}"></script>
   </body>
 </html>''';
     index_html = dist_dir / 'index.html';
     index_html.write_text(html_content, encoding='utf-8');
 }
+
+"""Resolve asset base path dynamically based on runtime context headers."""
+impl WebTarget.resolve_asset_base(headers: dict[str, str]) -> str {
```
MusabMahmoodh, line 273: this is the source side of the XSS i flagged in `client.impl.jac:152` - whatever comes in `x-forwarded-prefix` gets returned as-is (after trailing-slash handling) and ends up in `<base href="...">` with no escaping.

needs validation here: reject anything not matching `^/[a-zA-Z0-9_\-/]*$` and html-escape on the consumer side. plus the trust-model opt-in becouse defaulting to "trust whatever the header says" is the wrong direction - direct-to-internet deployments and misconfigured proxies are exposed.

jac-client/jac_client/plugin/src/targets/impl/web_target.impl.jac
```diff
@@ -247,19 +252,30 @@ def _generate_index_html(
...
+    prefix = headers.get("x-forwarded-prefix")
```
MusabMahmoodh: double normalization - `serve.static.impl.jac` already lowercases all headers before passing them in (`{k.lower(): v for k,v in request.headers.items()}`), so the case-insensitive fallback `or headers.get("X-Forwarded-Prefix")` here is dead code no? pick one place to normalize - i'd say at the boundary (server) and trust lowercase keys here. otherwise both spots have to stay in sync forever.

jac-client/jac_client/plugin/src/targets/impl/web_target.impl.jac
```diff
@@ -227,10 +224,18 @@ def _generate_index_html(
     import json;
     import from jac_client.plugin.client { HeaderBuilder }
 
+    static_base = bundler.config_loader.get_static_base_path();
```
MusabMahmoodh, line 227: what is the precedence between `static_base_path` (build-time, prefixes via vite's `base:`) and `X-Forwarded-Prefix` (runtime, prefixes via `<base href>`)? if both are set, doesnt this double-prefix? would be good to either document the precedence rule clearly or detect the conflict and `console.warning` so users dont silently end up with broken asset paths in a hybrid deployment.

jac-client/jac_client/plugin/src/targets/impl/web_target.impl.jac
```diff
@@ -247,19 +252,30 @@ def _generate_index_html(
...
     index_html = dist_dir / 'index.html';
     index_html.write_text(html_content, encoding='utf-8');
 }
+
+"""Resolve asset base path dynamically based on runtime context headers."""
+impl WebTarget.resolve_asset_base(headers: dict[str, str]) -> str {
+    prefix = headers.get("x-forwarded-prefix")
+    or headers.get("X-Forwarded-Prefix")
+    or "/";
+    if prefix and not prefix.endswith('/') {
```
MusabMahmoodh: nit: `prefix` is always truthy here becouse you just did `prefix = headers.get(...) or "/"` two lines above. the `if prefix and ...` is just `if not prefix.endswith('/')`.

jac/jaclang/runtimelib/impl/server.impl.jac
```diff
@@ -1585,8 +1588,9 @@ impl JacAPIServer.create_handler -> type[BaseHTTPRequestHandler] {
                 }
             }
             try {
+                headers = dict(self.headers) if self?.headers and self.headers else {};
```
MusabMahmoodh: nit: `self?.headers and self.headers` - the safe-nav already returns None or the value, so the `and self.headers` is redundant. just `if self?.headers` is enough.

MusabMahmoodh (changes requested):
nice work on this @kiptuidenis - the polymorphic `resolve_asset_base` is exactly the right pattern, much cleaner than hardcoded target-name branches. but i found one real security issue (XSS via the forwarded-prefix header) that needs fixing before merge, plus a few smaller things worth a look. inline.

couple things i liked: real polymorphism (no `if target_name == "..."` branches anywhere), minimal and backward-compatible jaclang core changes, breaking change documented, and the live server test that actually hits a real http endpoint with crafted headers is a proper integration test - those are rare in the recent PRs.

the XSS one is a P0 - default-deployed jac-scale (no proxy, or misconfigured proxy) is exposed to reflected XSS + service worker takeover via attacker-controlled `X-Forwarded-Prefix`. needs strict validation + html escape + opt-in flag with secure default. happy to chat through the mitigation if anything is unclear.

marsninja:

PR #6043, opened 2026-05-20
feat(bundle): add --precompile flag to automate multi-version bytecode precompilation

SupulHeshan:
## **Description**

Adds a `--precompile` flag to `jac bundle` that automatically discovers Python versions on PATH, spins up an isolated `venv` for each, and compiles `.jac` → `.jir` bytecode - eliminating the need to run the precompile script manually or as a separate CI job.

### New flag
- `jac bundle --precompile` scans for python3.11–python3.14 via `shutil.which`, creates a clean `venv` per version, installs the pinned `jaclang` version, and runs `precompilation` before packaging the wheel

### Script relocation
- Moved `precompile_bytecode.jac` from `jac/scripts/` into `jac/utils/` so it sits alongside other project-level utilities

## flow
mermaid-diagram (2)


jac/jaclang/cli/commands/impl/project.impl.jac
```diff
@@ -1123,6 +1180,49 @@ impl bundle(output: str = "dist") -> int {
...
+        import sys;
+        _jac_path = sys.modules.get("jaclang")?.__path__;
+        _jac_root: Path | None = Path(str(_jac_path[0])) if _jac_path else None;
+        if _jac_root is None {
+            console.error("Cannot locate jaclang package root for precompile script.");
+            return 1;
+        }
+        script_path = str(_jac_root.parent / "scripts" / "precompile_bytecode.jac");
+        if not Path(script_path).exists() {
+            console.error(f"Precompile script not found: {script_path}");
```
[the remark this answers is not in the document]
SupulHeshan: The `importlib.resources` approach is the "proper" API, but it returns a Traversable (not a real path string), and its own fallback still uses the wrong .parent path anyway.
for now the simple fix is, 
we can keep this `_jac_root.parent`, and have to install `jac` as editable-only in the `github` workflow.

then,
The cleaner long-term fix to remove this editable-only constraint is to either add the script inside `jaclang/scripts/` (so it ships with the package) or expose a proper `jac precompile` CLI command. Both avoid the path hack entirely.

jac/jaclang/cli/commands/impl/project.impl.jac
```diff
@@ -1123,6 +1180,49 @@ impl bundle(output: str = "dist") -> int {
...
+            return 1;
+        }
+
+        project_root_str = str(config.project_root);
+        pythons = _find_precompile_pythons();
+
+        if not pythons {
+            console.warning(
+                "No Python versions with jaclang found on PATH — skipping precompilation."
```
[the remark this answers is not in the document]
SupulHeshan: fixed

jac/jaclang/cli/commands/impl/project.impl.jac
```diff
@@ -1096,11 +1096,68 @@ impl jacpack(action: str = "", path: str = "", output: str = "") -> int {
...
+        }
+        if subprocess.run(
+            [venv_pip, "install", f"jaclang=={jac_version}", "--quiet"],
+            capture_output=True
+        ).returncode != 0 {
+            return False;
+        }
+        if dep_specs {
+            if subprocess.run(
+                [venv_pip, "install"] + dep_specs + ["--quiet"], capture_output=True
```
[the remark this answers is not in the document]
SupulHeshan: fixed

jac/jaclang/cli/commands/impl/project.impl.jac
```diff
@@ -1096,11 +1096,68 @@ impl jacpack(action: str = "", path: str = "", output: str = "") -> int {
...
+def _find_precompile_pythons -> list[tuple[str, str]] {
+    import shutil;
+    found: list[tuple[str, str]] = [];
+    for minor in range(11, 15) {
+        exe: str | None = shutil.which(f"python3.{minor}");
+        if not exe {
+            continue;
+        }
+        found.append((exe, f"cpython-3{minor}"));
+    }
```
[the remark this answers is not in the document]
SupulHeshan: changed docs

jac/jaclang/cli/commands/project.jac
```diff
@@ -377,13 +377,24 @@ def jacpack(action: str = "", path: str = "", output: str = "") -> int;
             help="Output directory for the built .whl file",
             short="o"
         ),
+        Arg.create(
+            "precompile",
+            typ=bool,
+            default=False,
+            help="Precompile .jac → .jir for every python3.X on PATH that has jaclang installed, then bundle",
```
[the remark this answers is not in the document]
SupulHeshan: already changed

jac/jaclang/cli/commands/impl/project.impl.jac
```diff
@@ -1096,11 +1096,86 @@ impl jacpack(action: str = "", path: str = "", output: str = "") -> int {
...
+            deps_install = subprocess.run(
+                [venv_python, "-m", "pip", "install"] + dep_specs + ["--quiet"],
+                capture_output=True
+            );
+            if deps_install.returncode != 0 {
+                return f"pip install project deps failed:\n{deps_install.stderr.decode(
+                    errors='replace'
+                )}";
+            }
+        }
```
kugesan1105, line 1159: no need of `pip install jaclang`

SupulHeshan:
image

This is the new release workflow in the `.yml` file.
Here, we need to use `pip install jaclang` in order to use the `jac` command inside `jac bundle --precompile`.
Since `jaclang` gets installed only for the final Python 3.14 setup, the other two Python versions do not have `jaclang` available. Because of that, we have to install `jaclang` again within the `jac bundle --precompile` command.

As a result, `jaclang` gets installed twice for Python 3.14.


**To prevent that one:**
image

We can install `jaclang` like this and mention it in the `release.yml` file.
Then, we can remove the code snippet that installs `jaclang` inside the implementation of `jac bundle --precompile`.

jac/jaclang/cli/commands/impl/project.impl.jac
SupulHeshan: we need jaclang installation because we have to install jaclang for each python versions..
thats why this installation here.

kugesan1105:
@SupulHeshan 
1. docs are missing
2. change all the current github release action workflow to use `jac bundle --precompile`
     `replaces the manual precompile CI step`
3. move `scripts/precompile_bytecode.jac` to `jac/jaclang/utils`

jac/jaclang/cli/commands/impl/project.impl.jac
```diff
@@ -1096,11 +1100,130 @@ impl jacpack(action: str = "", path: str = "", output: str = "") -> int {
...
+    }
+    if failed_versions {
+        console.warning(
+            "Precompilation incomplete — "
+            f"{len(failed_versions)}/{len(pythons)} version(s) failed: "
+            f"{', '.join(failed_versions)}"
+        );
+    }
+    return 0;
```
kugesan1105, line 1224: `_precompile_all_pythons `returns 0 even when every version fails — it only `console`.`warnings`. The bundle then proceeds and writes a wheel. So a release pipeline can produce a wheel that promises precompiled bytecode but contains none, with exit code 0.

jac/jaclang/cli/commands/impl/project.impl.jac
```diff
@@ -1096,11 +1100,130 @@ impl jacpack(action: str = "", path: str = "", output: str = "") -> int {
     return 1;
 }
 
-impl bundle(output: str = "dist") -> int {
-    import from pathlib { Path }
+glob _PRECOMPILE_MINOR_RANGE: range = range(11, 15);
```
kugesan1105: we don't need `python 3.11`

jac/jaclang/cli/commands/impl/project.impl.jac
SupulHeshan: yeah..
The behaviour is now:
image

jac/jaclang/cli/commands/impl/project.impl.jac
SupulHeshan: update the range to range(12, 15).

PR #6079, opened 2026-05-21
feat(compiler+planner+jac-scale): FilterSpec storage pushdown for [?:Type, field op const] filters

asimsalim:
## What this PR introduces

This PR introduces **FilterSpec** and **FieldPredicate** — the compiler and runtime plumbing that lets `[?:Type, field op const]` traversal filters push down to storage backends as native queries instead of scanning the full graph.

PR #6073 (merged) introduced the QueryPlan IR, the planner, and `MongoBackend.execute_plan`. It wired up the execution path and left `QueryPlan.field_predicates: dict = {}` as a placeholder. This PR fills that placeholder end-to-end: from what the compiler emits, through the planner, to what MongoBackend sends to MongoDB.

---

### 1. `jac0core/filter_spec.jac` (new file)

Defines two types:

- **`FieldPredicate(op, value)`** — one constraint on one archetype field. `op` is a backend-neutral lowercase name: `eq`, `ne`, `lt`, `lte`, `gt`, `gte`, `in`, `nin`. Multiple predicates on the same field are ANDed.
- **`FilterSpec(type_name, predicates, func)`** — emitted by the compiler instead of a plain lambda when a `[?:Type, field op const]` filter has extractable simple predicates. `func` is always present as a fallback lambda.

These live in `jac0core` (not `runtimelib`) to avoid a circular import: `runtimelib/query_plan.jac` → `jac0core/archetype.jac` → `jaclib.jac` → `query_plan.jac`.

### 2. `jac0core/passes/impl/pyast_gen_pass.impl.jac`

Adds `_build_predicates_ast`. When `exit_filter_compr` processes a `[?:Type, field op const]` filter and every comparison is a simple (field op constant) expression, it compiles to:

```python
FilterSpec(type_name='User', predicates={'age': [FieldPredicate(op='lt', value=28)]}, func=lambda n: ...)
```

Type-only filters (`[?:User]`) keep the plain lambda — the TopologyIndex fast-path is unaffected. Anything complex (method calls, nested expressions) also falls back to a plain lambda so behaviour is unchanged.

### 3. `jac0core/jaclib.jac`

Exports `FilterSpec` and `FieldPredicate` so compiled modules can reference them at runtime.

### 4. `runtimelib/query_plan.jac`

Types `field_predicates` as `dict[str, list[FieldPredicate]]` (was `dict = {}`). Updates the docstring to reflect backend-neutral ops instead of MongoDB-flavored operators.

### 5. `runtimelib/planner.impl.jac`

`_build_plan_from_path` now detects a `FilterSpec` on the destination node and extracts its `predicates` into `QueryPlan.field_predicates`. This detection is needed because `FilterSpec` is not a lambda — the existing `_filter_type_name` closure introspection would raise on it — and `type_name` must be read directly from the object.

### 6. `jac-scale MongoBackend._plan_to_mongo_filter`

Translates `FieldPredicate(op, value)` → `{"$" + p.op: p.value}` at the backend boundary. This is the only place MongoDB-specific syntax appears. A future Postgres adapter adds its own ~2-line translation without any MongoDB knowledge.

---

## Why backend-neutral from the start

The predicate format is baked into the compiled `.pyc` by the compiler. Emitting MongoDB-flavored `{'$lt': 28}` dicts would mean every future backend (Postgres, DynamoDB, etc.) has to parse and re-interpret MongoDB syntax. `FieldPredicate` with plain op names is trivially translatable at each backend's own boundary.

---

## Test plan

- [ ] `python3.12 -m pytest jac-scale/jac_scale/tests/test_topology_slice_pushdown.jac -v` — all 3 e2e pushdown tests pass with real testcontainers (field-predicate filter pushes to MongoDB; slice bounds fetches; Redis cache promotions)
- [ ] `python3.12 -m pytest jac/tests/compiler/passes/main/test_pyast_gen_pass.jac -k FilterSpec -v` — compiler emits `FilterSpec` with `FieldPredicate(op='lt', ...)` for predicate filters; plain lambda for type-only
- [ ] `python3.12 -m pytest jac-scale/jac_scale/tests/test_topology_scale.jac -v` — serializer + sqlite round-trip tests pass


marsninja:
## Follow-up: unify FilterSpec coverage + close predicate-extraction gaps

Pushed `43e6a0db1`. This is a hardening pass over the FilterSpec pushdown work. It came out of a review that found the abstraction was only partially applied and that several predicate forms either silently disabled pushdown or, worse, were dropped from the generated lambda entirely. Each fix below ships with a test that fails on the prior commit and passes now.

### 1. Every typed filter now carries a `FilterSpec` (structural, not bytecode)

Previously only `[?:Type, field op const]` emitted a carrier; type-only `[?:Type]` stayed a bare lambda, so the planner still had to recover the archetype type via `inspect.getclosurevars` + `dis` bytecode scanning. That fragile path was the *fallback for the simpler case*, which is backwards.

Now **every** typed filter lowers to a `FilterSpec` (predicates empty when not pushable). The planner gained `_filter_info()`, which reads `type_name` / `predicates` straight off the carrier. The getclosurevars/dis inspection is demoted to a fallback used only for filters that genuinely can't carry structure: condition-only `[?cond]` (no type) and bare instance/list filters. The dead `try/except FilterSpec import` and the inline `isinstance(..., _FS)` branches in `_build_plan_from_path` are gone.

### 2. Predicate extraction coverage

The original extractor required `field` on the left and a bare `Constant` on the right. The following all silently fell back to a per-node lambda (no pushdown), and `FieldPredicate` advertised `in`/`nin` that the compiler never actually emitted:

- **`in` / `not in`** against literal collections -> `op='in'` / `op='nin'`
- **negative literals** (`age < -1`) -> the `UnaryOp(USub, ...)` is now folded, was previously rejected as "not a Constant"
- **field on the right** (`28 > age`) -> operator flips to `age lt 28`
- **chained comparisons** (`20 <= age < 50`) -> expands into two ANDed predicates

The last two were also a **latent correctness bug independent of pushdown**: because the old lambda builder required the field on the left, `28 > age` and `20 <= age < 50` were dropped from the *generated lambda* too, so those filters matched everything. `exit_filter_compr` is rebuilt around a single `_analyze_compare` normalization that drives both the lambda body and the predicate dict from one analysis, so they can't diverge again.

### 3. Mixed-strategy `post_filter` bug

`_execute_mixed` applied `plan.post_filter(anchor)` while the pushdown path correctly used `anchor.archetype`. The node-filter callable expects the archetype, so any mixed-strategy query that retained a `post_filter` would fail its `isinstance(i, T)` check and silently drop every row. Now it passes `anchor.archetype` (guarded for `None`), matching `_execute_pushdown`.

### Tests

- `test_pyast_gen_pass.jac` + fixture: emission for in/nin/negative/const-on-left/chained, plus the empty-predicate type-only carrier.
- `test_topology_e2e.jac`: runtime correctness pinning chained and constant-on-left filtering (these were the silent-drop repros).
- `test_planner_mixed.jac` (new): a focused `_execute_mixed` contract test asserting `post_filter` receives the archetype.
- `test_topology_slice_pushdown.jac`: mongo pushdown proof for `in` + negative-literal operators (counts via `execute_plan_fetch_count`, `fetch_count == 0`).

Local runs: full `tests/runtimelib/` (306), `tests/language/` (91), `tests/compiler/passes/main/` + formatter pass (516/57 skipped) all green.

### Notes / not in scope

- The bytecode helpers (`_filter_type_name` / `_filter_has_predicates`) are intentionally retained for non-carrier filters; deleting them outright would require also routing the `:EdgeType:weight>=5:` edge-ref syntax through a carrier, which is a separate change.
- While testing I hit a stale **project-local** `.jac/cache` that served pre-change codegen even after clearing `~/.cache/jac/jir` (content-hash invalidation didn't fire on an append-only edit). Worth a look as a separate issue; not addressed here.

marsninja:

PR #6084, opened 2026-05-21
fix-tls-configueration

### LINKED ISSUE #5834: jac-scale: redeploy wipes TLS configuration from Ingress, breaking HTTPS

juzailmlwork:
## Summary

`jac start --scale` silently wiped TLS configuration from the app Ingress on every redeployment after `--enable-tls` had been run. Users would see `ERR_CERT_AUTHORITY_INVALID` in browsers while cert-manager reported the certificate as `Ready=True`, making the root cause very hard to spot.

**Root cause**: `_deploy_ingress_resource` called `replace_namespaced_ingress` (a full object PUT) each time. The new body was built with no `spec.tls`, no rule host, and none of the TLS annotations (`cert-manager.io/issuer`, `ssl-redirect`, `force-ssl-redirect`) — so every redeployment undid what `--enable-tls` had patched in. The `jac-scale/domain` and `jac-scale/cert-email` annotations were preserved (written unconditionally from config), which made the metadata look half-correct and masked the issue.

## Changes

- **`_deploy_ingress_resource` — main Ingress**: Before building the new Ingress body, read the existing Ingress with `read_namespaced_ingress` and deserialize it via `sanitize_for_serialization`. If `spec.tls` is non-empty or `rules[0].host` is set, copy the TLS list, host, and the three TLS annotations into the new body before calling `replace_namespaced_ingress`. First-deploy behaviour (Ingress does not exist → 404 caught silently) is unchanged.

- **`_deploy_ingress_resource` — RedisInsight Ingress**: Same fix applied to the `redis_ingress_name` branch, which had the identical pattern.

- **`enable_tls`**: No changes — it remains the sole path that writes TLS state for the first time.

## Reproduction (before fix)

```bash
# 1. Initial deploy
jac start main.jac --scale --target kubernetes

# 2. Enable TLS — cert issued, HTTPS works
jac start main.jac --scale --target kubernetes --enable-tls

# 3. Any subsequent redeploy
jac start main.jac --scale --target kubernetes
# Result: ERR_CERT_AUTHORITY_INVALID
# kubectl get ingress -n <ns> -o yaml → no spec.tls, no host on rules

### CODE REVIEW

**Review by udithishanka — APPROVED** [review]

**Comment by udithishanka** [comment]
verified this on jaseci-cluster (us-east-2) against two throwaway namespaces, one with jac-scale from upstream/main (dc58f8776) and one with jac-scale from this PR (344ce348d). exercised _deploy_ingress_resource directly so the verification didn't spin up the NGINX controller, NLB, MongoDB, Redis, or any app pods. only the Ingress object itself plus a placeholder IngressClass per namespace.

flow in each namespace was: first deploy with no TLS, then patch the resulting Ingress with the same fields enable_tls writes (spec.tls, rule host on each rule, cert-manager.io/issuer, ssl-redirect, force-ssl-redirect annotations), then a second deploy. read the Ingress back and compared.

bug reproduced cleanly on main. after the second deploy, spec.tls and the rule host were gone, and the three TLS annotations were stripped. only jac-scale/domain and jac-scale/cert-email remained, which matches the issue report exactly.

```
$ kubectl get ingress -n verify-pr-6084-buggy testapp-ingress -o jsonpath='{.spec.tls}'
(empty)
$ kubectl get ingress -n verify-pr-6084-buggy testapp-ingress -o jsonpath='{.spec.rules[0].host}'
(empty)
$ kubectl get ingress -n verify-pr-6084-buggy testapp-ingress -o jsonpath='{.metadata.annotations.cert-manager\.io/issuer}'
(empty)
```

PR branch preserves the TLS state on the second deploy.

```
$ kubectl get ingress -n verify-pr-6084-fixed testapp-ingress -o jsonpath='{.spec.tls}'
[{"hosts":["test.example.com"],"secretName":"testapp-tls"}]
$ kubectl get ingress -n verify-pr-6084-fixed testapp-ingress -o jsonpath='{.spec.rules[0].host}'
test.example.com
$ kubectl get ingress -n verify-pr-6084-fixed testapp-ingress -o jsonpath='{.metadata.annotations.cert-manager\.io/issuer}'
letsencrypt-prod
```

verdict lines from the test script:

```
[main / buggy]
spec.tls preserved: False
rule host preserved: False
cert-manager.io/issuer annotation: None
ssl-redirect annotation: None

[PR / fixed]
spec.tls preserved: True
rule host preserved: True
cert-manager.io/issuer annotation: letsencrypt-prod
ssl-redirect annotation: true
```

also reran the same flow with redis_dashboard=True on the PR branch to cover the matching fix on the testapp-redis-insight-ingress branch. that Ingress also kept spec.tls, the rule host, and the three TLS annotations after the second deploy.

first-deploy behaviour is unchanged. read_namespaced_ingress 404s on a brand-new namespace, the except catches it, and the code falls through to create_namespaced_ingress like before.

namespaces and per-namespace IngressClasses were deleted after the run. no existing apps in the cluster were touched.

kugesan1105:
feels like workaround that adds maintenance debt and leaves a same-class latent bug one annotation away.

- `Preferred:` Convert `_deploy_ingress_resource` to `patch_namespaced_ingress` for the fields `jac-scale` owns, so it never sends `spec.tls`/`host` and there's nothing to preserve. Smallest, most robust, matches `enable_tls`. Delete most of the **+12**2 lines.

PR #6089, opened 2026-05-21
proposal: jac code semantic CLI + agent discoverability

marsninja:
## Status

**Proposal / tracking PR.** No code changes yet -- this PR exists to socialize the roadmap and gather feedback before implementation. The commit is intentionally empty.

## Motivation

External AI coding agents working in a Jac codebase fall back to `grep` because that's the only structural-lookup tool they can reach. Jac already has something far better: the compiler-backed `CodeIntelligence` API at [`jac/jaclang/compiler/code_intel.jac`](jac/jaclang/compiler/code_intel.jac), which answers *"where is this symbol defined, every place it's used, what's its type, which walkers traverse it?"* -- the questions grep can't answer.

Today that API has exactly **one** consumer: the **`jac ai` byLLM agent** ([`jac/jaclang/cli/ai_agent.jac:518-617`](jac/jaclang/cli/ai_agent.jac#L518-L617)), via the semantic tools `find_symbol`, `find_uses`, `project_map`, `find_walkers`, `context_slice`. Only the internal agent sees them. The shared `AgentRequest` contract that hands the agent a `CodeIntelligence` handle already lives in [`jac/jaclang/cli/agent_api.jac`](jac/jaclang/cli/agent_api.jac) -- so there is already a seam to build on.

There is also an **MCP server** ([`jac-mcp`](jac-mcp/jac_mcp/)), excellent for IDEs but heavyweight for terminal-only agents. Worth flagging: the MCP server does **not** route through `CodeIntelligence` -- it carries its own parallel [`compiler_bridge.impl.jac`](jac-mcp/jac_mcp/impl/compiler_bridge.impl.jac). So the compiler-query logic is *already duplicated* across two implementations. That makes the "one implementation, not many" discipline below more urgent than it first looks: a naive `jac code` would be the third copy.

**Gap:** there is no terminal-friendly way for an external coding agent to ask "what is `myProfile`'s type and where is it written from?" without re-implementing Jac's compiler. The result: agents reach for grep, miss renames/shadowing/cross-file effects, and produce worse edits than they could.

This proposal closes that gap with three coordinated changes: a `jac code` CLI command group, an agent-onboarding guide, and selective JSON-by-default output.

### 1. New `jac code` command group (thin CLI shim over `CodeIntelligence`)

One subcommand per `code_intel` query. JSON-by-default, schema-stable, exit codes that mean something.

| Command | Wraps | Use case |
|---|---|---|
| `jac code symbol <name>` | `ci.lookup()` + uses summary | "What is this thing and where is it used?" |
| `jac code uses <name>` | `ci.references()` | "Show me every read/write of this symbol" |
| `jac code map` | `ci.archetypes()` | Project structural overview (nodes/walkers/edges/objs) |
| `jac code walkers <node-type>` | `ci.walkers_visiting()` | "Which walkers traverse this node?" |
| `jac code slice <name>` | `ci.slice_for()` | Typed neighbourhood for prompt assembly |
| `jac code diag [path]` | `ci.diagnostics()` | Structured errors/warnings (better than parsing `jac check` text) |
| `jac code where <file>:<line>:<col>` | LSP-style hover | Symbol-at-position lookup |

**Why a sibling group, not under `jac tool`:** `jac tool` is internal AST plumbing (and `jac tool list_tools` currently errors with `Ast tool list_tools not found` -- worth fixing while you're nearby). `jac code` reads as "ask the compiler about my code," which is what agents need. Call it grep's structural successor in the help text.

**Implementation cost is small.** Each subcommand is roughly `CodeIntelligence(cwd).load(); print(json.dumps(...))`. The hard part (the API) already exists. One real subtlety: the five tools in `ai_agent.jac` currently return `render()`-formatted *strings*, not structured data, so the extraction is not a pure copy-paste -- the shared module must return JSON-friendly dicts and the byLLM agent then renders *from* those dicts. The underlying records (`SymbolInfo`, `SourceRef`, ...) already expose their fields directly, so the data is all there. Lift the five tool bodies out of `ai_agent.jac` into a shared module both the CLI and the byLLM agent import, alongside the existing `agent_api.jac` seam, so there is one implementation, not three.

### 2. `jac guide` becomes the agent onboarding surface

Add one new guide whose **audience is the external agent, not the human**:

- `jac-agent-tools` (working name) -- frontmatter: *"Read this FIRST if you are an AI coding assistant. Lists every CLI probe that lets you understand Jac code precisely instead of falling back to grep or text search."*
- Body in second person ("you, the agent"): when to use `jac code symbol` over grep, when `jac code slice` replaces "read whole file," what `jac code diag --json` returns, how `jac guide --json` indexes the rest. Target ~150 lines.

Then surface it in `jac guide`'s idle output, which today lists 21 topics with no hint that one is *the* entry point:

```
$ jac guide
> AI agents: start with `jac guide jac-agent-tools`
> Humans: start with `jac guide jac-core-cheatsheet`

Jac reference guides
  ...
```

That two-line discovery hint is the cheapest, highest-leverage change in the proposal. An agent finds it the first time it runs `jac guide` and never needs to be told again.

### 3. JSON output, selectively

Default JSON for *structured* commands, keep text for prose:

| Command | Default | Reason |
|---|---|---|
| `jac code *` | **JSON** | structured records, agents will parse |
| `jac check`, `jac lint` | **JSON when stdout is not a TTY** | agents pipe, humans run interactively |
| `jac guide <name>` | markdown (unchanged) | prose body, agents want it raw |
| `jac guide` (list) | already supports `--json` | no change |
| `jac config show` | already supports `--output json` | no change |

The TTY-sniffing pattern matters: `jac check . > errors.json` should just work without `--json`. The `_emit` helper in [`guide.impl.jac`](jac/jaclang/cli/commands/impl/guide.impl.jac#L16) is the right starting point -- today it's an unconditional *verbatim* stdout writer that deliberately bypasses Rich (which corrupts machine output); the TTY-sniffing is net-new but small, layered on that same "verbatim for machines" instinct.

For `jac code`, also include a `--text` flag that uses the `render()` methods already on `SymbolInfo`/`SourceRef`/etc. Free human-readable mode at zero extra cost.

## Sample interaction the agent gets after this lands

```
$ jac code symbol myProfile --json
{"definitions":[
  {"name":"myProfile","kind":"field","file":"frontend.cl.jac","line":49,
   "type_str":"ProfileView | None","container":"app"}],
 "uses":[
  {"file":"frontend.impl.jac","line":96,"snippet":"myProfile = None;","role":"write"},
  {"file":"frontend.impl.jac","line":7,"snippet":"myProfile = result.reports[0] if ...","role":"write"},
  {"file":"ProfileTab.cl.jac","line":7,"snippet":"myProfile: ProfileView | None,","role":"param"},
  ...]}
```

One call replaces three greps and a file read for *"does this field actually need `| None`?"*

### Phase 1: minimum viable, agent-changing (1 weekend)

- [ ] Extract the five semantic-tool bodies from `ai_agent.jac` into a shared module (e.g. `jac/jaclang/cli/code_queries.jac`) returning plain JSON-friendly dicts -- *not* the `render()`-strings they return today. Position it alongside the existing `agent_api.jac` contract.
- [ ] Bake `"schema_version": 1` into every JSON record from the first commit (promoted from Phase 2 -- agents pin to field names the moment this ships, so versioning cannot be retrofitted cleanly).
- [ ] Wire `jac code symbol|uses|map|slice|diag` with `--json` (default) and `--text` flags. Project-root autodiscovery (walk up to `jac.toml` like `jac check` already does).
- [ ] Write the `jac-agent-tools` guide (`jac/jaclang/cli/skills/jac-agent-tools.md`).
- [ ] Add the two-line discovery hint to `jac guide`'s idle output.
- [ ] Update `ai_agent.jac` to render from the shared module so internal and external surfaces stay in lockstep.

**Acceptance:** an external agent in a fresh Jac repo can run `jac guide`, follow the prompt, read `jac-agent-tools`, and answer the `myProfile` question in one `jac code symbol` call.

### Phase 2: rough edges (a few days)

- [ ] `jac code walkers <node-type>` and `jac code where <file>:<line>:<col>`.
- [ ] TTY-aware JSON for `jac check` and `jac lint` (auto-JSON when stdout is not a TTY).
- [ ] Stable exit codes: 0 = ok, 1 = error, 2 = not-found, 3 = compile-failed-but-partial-data.
- [ ] `jac code --schema` (or schema section in the guide) documenting record shapes so agents do not treat field names as load-bearing.

### Phase 3: parity and polish (later)

- [ ] Mirror MCP `jac://` resources as `jac code resource <uri>` so terminal-only agents do not need MCP plumbing for the common 80%. (Bonus: a path toward retiring the MCP `compiler_bridge` duplication onto the shared `CodeIntelligence` core.)
- [ ] Fix `jac tool list_tools` (currently errors -- noticed while exploring).
- [ ] Optional: a `--watch` mode for `jac code diag` that reuses `CodeIntelligence.refresh_and_diff` for live agent feedback loops.

## Non-goals

- No change to `jac ai`'s UX. The internal agent keeps working exactly as today, just on the shared implementation.
- No new dependency. `CodeIntelligence` already ships with the compiler.
- Not trying to replace MCP. MCP stays the right answer for IDEs; `jac code` is the answer for terminals and shell-driven agents.

## Open questions

1. **Naming.** `jac code` vs `jac probe` vs `jac sem` vs folding into `jac tool`. Lean: `jac code`. Reads naturally, parallels `jac check`/`jac lint`.
2. **Should `jac code diag` recompile or read cached diagnostics?** Recompile is honest but slower; cached is fast but can be stale -- and stale diagnostics mislead an agent worse than no tool at all. Suggest: recompile by default, `--cached` to skip.
3. **Schema versioning** is now folded into Phase 1 (`"schema_version": 1` in every record). Remaining question is only *how* to evolve it -- additive-only, or a documented deprecation window?
4. **Should the MCP `compiler_bridge` eventually fold onto `CodeIntelligence`?** Out of scope here, but this proposal makes the shared core that would make it possible. Worth a separate tracking issue.

## Test plan

- [ ] Unit tests for each `jac code` subcommand against the `examples/littleX/fullstack` fixture (known good shapes).
- [ ] Round-trip test: feed `jac code` output back into a small script that asserts the same answers a hand-rolled grep would produce on a controlled fixture.
- [ ] Manual: drive an external coding agent through a sequence of *"find every place X is written to"* style tasks; confirm it picks `jac code` over `grep` after reading the new agent guide.
- [ ] Snapshot test on `jac guide` idle output to lock the discovery hint in.


marsninja:
## Deprecation / consolidation ledger

One thing this proposal is light on: what we *retire* once it lands. The work is overwhelmingly additive, which is the right call -- there isn't much user-facing surface that should be deleted. But the real payoff of building the shared `CodeIntelligence` core is **consolidation, not deletion**. Three things become genuinely redundant:

### 1. The MCP `compiler_bridge` -- deprecate the implementation, keep the server

This is the strongest cleanup target. The MCP server ([`jac-mcp/jac_mcp/impl/tools.impl.jac`](jac-mcp/jac_mcp/impl/tools.impl.jac)) exposes ~20 tools (`get_ast`, `validate_jac`, `check_syntax`, `lint_jac`, `explain_error`, ...) through its **own** parallel [`compiler_bridge.impl.jac`](jac-mcp/jac_mcp/impl/compiler_bridge.impl.jac) -- *not* `CodeIntelligence`. So once `jac code` ships, compiler-query logic exists in **three** places (`code_intel`, the `ai_agent.jac` tool bodies, and the MCP bridge), and they can silently diverge in correctness.

Don't remove the MCP server -- it stays the right answer for IDEs. Deprecate its private bridge by re-pointing those tools at the shared core. This is the one place the proposal actually *retires* code instead of adding it. Suggest promoting this from the Phase-3 bullet / open question #4 into a named follow-up issue.

### 2. `jac tool list_tools` -- remove it, don't fix it

The current Phase-3 note says "fix while you're nearby." I'd push the other way. `jac tool` is AST-dev plumbing ([`lang_tools.jac`](jac/jaclang/utils/lang_tools.jac) -- `ir` / `py_uni_nodes` / `autodoc`), and `list_tools` errors today (`Ast tool list_tools not found`) only because the help text in [`tools.jac:23`](jac/jaclang/cli/commands/tools.jac#L23) advertises a method the AST tool class doesn't have.

For an *agent*, the thing `jac tool` would offer is exactly what `jac code map` now provides, better. So:
- Remove the broken `("jac tool list_tools", ...)` row from the help.
- Demote `jac tool` out of the agent-facing help surface so agents funnel to `jac code`.

Fixing `list_tools` spends effort making a command discoverable that we've just made obsolete.

### 3. Two "what should an agent read" catalogs -- pick one source of truth

After this lands, `jac guide` becomes *the* agent-onboarding surface. But the MCP server already has its own discovery layer -- `understand_jac_and_jaseci`, `search_docs`, `get_resource`, and a `jac://` resource catalog in [`resources.impl.jac`](jac-mcp/jac_mcp/impl/resources.impl.jac). That's a second, independently maintained answer to "which docs matter." Rather than delete it, generate the MCP resource catalog from the same guide registry `jac guide` uses, so the onboarding story doesn't fork into two drifting copies.

### Explicitly *not* removal candidates

- **The five `find_*` tools in `ai_agent.jac`** -- refactored to render from the shared module, not removed (already covered above).
- **`jac code symbol` vs `jac code uses`** -- `symbol` gives a uses *summary*, `uses` gives the full read/write/param list. Distinct; keep both.
- **The `render()` methods / `--text` mode** -- stay; they're the free human-readable path.

**Bottom line:** the genuine wins from doing all this work are collapsing three compiler-query implementations onto one core and two doc-discovery catalogs onto one. The only thing worth outright *removing* is the broken `jac tool list_tools` advertisement.

Issue #6082, opened 2026-05-21
views: pass-level extensions to the JSX slot model (auto-fragment, except, static hoisting, statement-level bodies, …)


marsninja:
Tracking issue for a set of pass-level extensions to the JSX slot model
shipped in #6042 and the `awaiting` clause from #6063. All items here
share a common design constraint: **no new declarators, no new sigils**,
and where possible no parser changes — the philosophy from #6042's
"lower `view` in a pass, not the parser" stays the rule.

Overlaps with #6055 are cross-referenced rather than duplicated:
- Refs (`Ref[T]` + `has` → `useRef`) → see #6055 "Refs" section
- Scoped styles → see #6055 "Scoped styles" section
- `sv`/`na` lowering for `awaiting` → see #6055 "Boundaries" section

This issue covers the items NOT yet captured in #6055.

### 1. Loop auto-fragment + key derivation for `for` slots

`ViewLowerPass` should detect `for` slots whose body emits multiple JSX
statements (or a single keyless JSX element) and synthesize one
`<React.Fragment key={…}>…</>` per iteration.

Key derivation priority:
1. Lifted explicit `key=` on any one child element (strip from child, hoist
   to the synthesized fragment).
2. `jid(loop_var)` when the loop var is typed as a Jac `obj`/`node`.
3. The loop counter for `IterForStmt` (`for i = 0 to n by 1`).
4. Otherwise, **W2021**: "for-loop slot iteration has no derivable key —
   annotate one child with `key=`."

This expresses TSRX-style "key it.id on the loop header" semantics from a
plain `for it in items { … }` source, with no grammar change. Mirrors the
`JacAwaiting` auto-wrap pattern.

### 2. Diagnostic gap: keyless for-loop in a slot

Independent of (1) — `_check_body` at `view_lower_pass.jac:119` today only
calls `_emits_keyless_jsx` on `WhileStmt`, never on `InForStmt`/`IterForStmt`.
The for-loop is the dominant keyless footgun. Extend the check, either
reusing W2019 (rename "while" → "loop") or as a new W2021 if (1) lands
first.

### 3. `except` arms integrated with `try { } awaiting { }` slot lowering

`_lower_one` at `view_lower_pass.jac:394-396` already iterates `s.excepts`,
but the `<JacAwaiting>` wrapper doesn't route them. On the `cl` target,
when a slot `try` has both `awaiting` and one or more `except` clauses:

[...]

PR #6097, opened 2026-05-22
test(views): executable spec for view-lowering JSX slot extensions (#6082 items 1, 3, 5, 7)

marsninja:
**Deliberately RED tracker.** This is a failing-tests-only branch — the executable spec for the four `view_lower_pass.jac` items in #6082 that share the slot-lowering recursion. Each test codifies the contract the implementing PR must satisfy. Split out of #6088 so that CI (`--exitfirst`) stays green on the implementation branch while the spec remains a visible tracker.

## Coverage map

| Test  | #6082 item                                            |
| ----- | ----------------------------------------------------- |
| 1.1   | item 1 priority 1 — lifted explicit `key=`            |
| 1.2   | item 1 priority 2 — `jid(loop_var)` for typed obj     |
| 1.3   | item 1 priority 3 — counter for IterForStmt           |
| 2     | item 3 — try/awaiting + except auto-wrap              |
| 4     | item 5 — static JSX subtree hoisting                  |
| 6     | item 7 — statement-level JSX in def -> JsxElement     |

## Why this grouping

All four items live in `view_lower_pass.jac` and share the `_check_body` / `_lower_body` / `_lower_one` / `_lower_try_with_awaiting` recursion. The issue itself notes "Tier 1 rows 1+2+3+4 bundle naturally — same `_check_body`/`_lower_one` recursion", and item 7 (statement-level JSX) is "the same machinery as a slot body, just at function scope" — a second entry point into the already-parameterized lowering. They review as one surface.

## How to use this PR

- When the implementation lands, each test should turn green; relocate the green tests into `test_slot_extensions.jac` (where the already-shipped #6085 items' positive tests live) and drop them from this file.
- When all six pass, this tracker can be closed.
- rec 4 carries an inline POST-IMPLEMENTATION TIGHTENING note — the `__jac_static_<N>` const-naming assertion is loose where the issue leaves the shape open; tighten it when the implementer commits to a concrete scheme.

## Companions

- #6085 — the four #6082 items already shipped (`{name}` shorthand, E2024, W2021).
- Sibling red-spec trackers split from #6088: Ref/useRef (item 8), scoped `.style.css` styles (item 10), and the streaming items (sv-awaiting + walker progressive frames, gated on design).

### DIFF
--- a/jac/tests/compiler/passes/ecmascript/fixtures/slot_extensions/loop_auto_fragment.cl.jac
+++ b/jac/tests/compiler/passes/ecmascript/fixtures/slot_extensions/loop_auto_fragment.cl.jac
@@ -1,0 +1,35 @@
+"""Rec 1 — for-loop slot whose body emits multiple keyless JSX siblings.
+
+ViewLowerPass should detect this shape and wrap each iteration in a
+synthesized `<React.Fragment key={jid(tweet)}>...</>` so that one
+iteration is one keyed unit in the React diff, instead of three keyless
+siblings that all share the same diff bucket.
+"""
+
+obj Tweet {
+    has author: str,
+        body: str;
+}
+
+def:pub TweetRow(tweet: Tweet) -> JsxElement {
+    return
+        <div class="row">
+            <h3>{tweet.author}</h3>
+            <p>{tweet.body}</p>
+            <button>like</button>
+        </div>;
+}
+
+"""A list of tweets where each iteration emits three top-level keyless
+JSX statements. Without auto-fragmenting, React diffs them as one flat
+keyless list and loses iteration identity on reorder."""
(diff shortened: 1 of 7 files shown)


marsninja:
Closing in favor of splitting this bundle up. Rec 2 (try/awaiting + except → `JacClientErrorBoundary`) is pulled out as its own PR #6109 — it fixes a silent-drop correctness gap and additionally needs a new `JacClientErrorBoundary` runtime component, so it doesn't share the auto-fragment / hoisting / statement-body work the rest of #6082 bundles. The remaining items (1, 5, 7) will be reworked separately (rec 1.2's key-derivation primitive needs settling — `jid()` doesn't exist; `ln()` is the language idiom — and rec 6 is parser-blocked, not a pure view-lowering change).

PR #6105, opened 2026-05-22
 fix(jac-scale): strip empty dicts in `_put_node_atomic` to avoid MongoDB error 40180

### LINKED ISSUE #6076: [Bug]: MongoDB `_put_node_atomic` fails with "empty object" error on all anchor updates
### Operating System

Ubuntu

### Versions

Jac 0.15.3 (latest packages that came with the release of Jac 0.15.3)

## Summary

`MongoBackend._put_node_atomic()` uses aggregation pipeline syntax for updates (`[{'$set': set_doc}]`), which rejects empty embedded documents. Since every `NodeAnchor` serializes `access.roots.anchors` as an empty dict `{}` (the default from `Access.anchors: dict = field(default_factory=dict)`), **every graph write to MongoDB silently fails** with:

```
ERROR - MongoDB _put_node_atomic failed: Invalid $set :: caused by :: an empty object is not a valid value. Found empty object at path data.access.roots.anchors, full error: {'index': 0, 'code': 40180, 'errmsg': '...'}
```

This means **no graph mutations persist** — new edges, node field updates, and node creations are lost on pod restart or when a different pod reads from MongoDB.

## Environment

- **Jaseci version**: `jaclang>=0.14` with `jac-scale`
- **MongoDB**: 6.x+ (EKS deployment)
- **Deployment**: 3 pods behind a gateway (microservices mode)

## Root Cause

`jac_scale/impl/memory_hierarchy.mongo.impl.jac`, method `_put_node_atomic` (line ~545)

```python
# Line 569-574: Serializes ALL non-archetype fields including access
full = dict(Serializer.serialize(anchor, include_type=True));
non_arch_top = {
    k: v
    for (k, v) in full.items()
    if k not in ('edges', 'archetype')
};

# Line 580-586: Sets them directly into $set document
set_doc: dict[str, object] = { ... };
    set_doc[f"data.{k}"] = v;   # <-- includes access.roots.anchors = {}
}

# Line 597: Uses aggregation pipeline syntax ([...])
self.collection.update_one({'_id': _id}, [{'$set': set_doc}], upsert=True);
```

The `access` field serializes as:
```json
{
  "access": {
    "all": 0,
    "roots": {
      "anchors": {}
    }
  }
}
```

The empty `anchors: {}` triggers MongoDB error code `40180` (`EmptyObjectInUpdate`) because **aggregation pipeline updates reject empty embedded documents** in `$set` operations.

## Impact

1. **All graph writes silently fail** — the error is caught and logged but the write is lost
2. **New nodes created in-memory are never persisted** — on pod restart or cross-pod reads, the data is missing
3. **Old user accounts are particularly affected** — they have existing UserProfile nodes that need edge updates (adding new Projects), but the UserProfile update fails because of the empty anchors field
4. **New accounts may appear to work** if they hit the same pod that created them (in-memory graph), but the data is not durable

## Reproduction

1. Start a jac application with jac-scale MongoDB backend and 2+ pods
2. Create a user account (creates UserProfile node in graph)
3. Make any graph mutation from that user (e.g., create a child node)
4. Observe the MongoDB error in logs: `MongoDB _put_node_atomic failed: ... empty object at path data.access.roots.anchors`
5. Send the same request to a different pod — the mutation will be missing (node/edge not found)

### Minimal Reproduction (single pod)

```jac
# In any walker that creates a graph node:
walker test_repro {
    can start with Root entry {
        # This triggers _put_node_atomic on the root anchor
        root ++> SomeNode();
        # Check MongoDB logs — the root anchor update will fail
        # with "empty object at path data.access.roots.anchors"
    }
}
```

### Verify with MongoDB query

```python
from pymongo import MongoClient
client = MongoClient("mongodb://...")
db = client["jac_db"]
# Every anchor has access.roots.anchors = {} (empty dict)
for doc in db["_anchors"].find().limit(3):
    anchors = doc.get("data", {}).get("access", {}).get("roots", {}).get("anchors", "MISSING")
    print(f"id={doc['_id'][:20]} anchors={anchors}")
    # Output: id=00000000-0000-0000-0 anchors={}
```

## Suggested Fix

Option A — Strip empty dicts before including in `$set`:

```python
def _strip_empty_dicts(d: dict) -> dict {
    return {k: v for (k, v) in d.items() if not (isinstance(v, dict) and len(v) == 0)};
}
```

Apply to `non_arch_top` before building `set_doc`:

```python
non_arch_top = {
    k: _strip_empty_dicts(v) if isinstance(v, dict) else v
    for (k, v) in full.items()
    if k not in ('edges', 'archetype')
};
```

Option B — Skip the `access` field entirely in `$set` and only write it when it has meaningful content:

```python
    if k == 'access' {
        # Flatten and skip empty nested dicts
        v = _clean_access(v);
    }
    if v is not None {
        set_doc[f"data.{k}"] = v;
    }
}
```

Option C — Use regular update syntax (not aggregation pipeline) which tolerates empty objects:

```python
# Change from:
self.collection.update_one({'_id': _id}, [{'$set': set_doc}], upsert=True);
# To:
self.collection.update_one({'_id': _id}, {'$set': set_doc}, upsert=True);
```

**Note**: Option C may require restructuring the `data.edges` field since it currently uses aggregation expressions (`$setUnion`, `$setDifference`, `$ifNull`).

## Workaround

A temporary fix is to clean the MongoDB `_anchors` collection by removing the empty `access.roots.anchors` fields:

```python
db["_anchors"].update_many(
    {"data.access.roots.anchors": {}},
    [{"$unset": "data.access.roots.anchors"}]
)
```

This must be re-applied after every graph write that re-introduces the empty field.

_No response_

akindu-k:
## Problem

Every graph write on a MongoDB-backed jac-scale deployment was **silently failing** with:

```
ERROR - MongoDB _put_node_atomic failed: Invalid $set :: caused by :: an empty object is not a valid value.
Found empty object at path data.access.roots.anchors
```

This meant new edges, node field updates, and node creations were never durably persisted. Data appeared correct on the pod that created it (in-memory L1 cache), but was missing on pod restart or when any other pod read from MongoDB.

### Why it happens

`_put_node_atomic` uses MongoDB's **aggregation-pipeline** update syntax — `[{'$set': set_doc}]` — in order to express the atomic edge-list merge with `$setUnion` / `$setDifference`. MongoDB 6.x introduced a strict validation rule for this syntax: it rejects **any embedded empty document** anywhere in the `$set` payload (error code 40180, `EmptyObjectInUpdate`).

The `Permission` / `Access` types that every `NodeAnchor` carries serialize to:

```json
"access": {
  "all": -1,
  "roots": {
    "anchors": {}
  }
}
```

`anchors` is `dict = field(default_factory=dict)` — always `{}` by default unless explicit access has been granted. So every anchor update hit this path and every write failed.

The regular (non-pipeline) `$set` used by `_write_to_db` and `put_partial` does **not** have this restriction, which is why initial inserts and field-only updates appeared to work.

---

## Fix

**File:** `jac-scale/jac_scale/impl/memory_hierarchy.mongo.impl.jac`

Added a recursive helper `_strip_empty_dicts(v)` that walks a serialized value and removes any empty-dict leaves before they enter the aggregation-pipeline `$set` document:

```
access.roots.anchors = {}        →  stripped (empty leaf removed, parent cleaned up too)
access.roots.anchors = {"uuid-1": 2}  →  preserved unchanged
```

Applied it to the `non_arch_top` values inside `_put_node_atomic` before building `set_doc`:

```diff
 non_arch_top = {
-    k: v
+    k: _strip_empty_dicts(v)
     for (k, v) in full.items()
     if k not in ('edges', 'archetype')
 }
```

**Why this is safe:** The `Serializer` restores all `field(default_factory=dict)` fields to `{}` on deserialization regardless of whether the key is stored or absent in the document. Stripping an empty dict from storage is lossless — the value is semantically identical to the key not existing.

**Why not switch to non-pipeline syntax:** The aggregation-pipeline syntax is required specifically for the `$setUnion` / `$setDifference` expression that makes the edge-list merge atomic. Switching to regular `$set` would lose that atomicity guarantee.

---

## Changes

| File | What changed |
|---|---|
| `jac-scale/jac_scale/impl/memory_hierarchy.mongo.impl.jac` | Added `_strip_empty_dicts()` helper; applied it in `_put_node_atomic` when building `non_arch_top` |
| `jac-scale/jac_scale/tests/test_mongo_layer123.jac` | Added 2 regression tests (see below) |

---

## Tests

Two new integration tests using a real `MongoDbContainer`:

**Test 1 — direct trigger reproduction:**
Creates a `NodeAnchor` with the default empty `access.roots.anchors`, calls `_put_node_atomic` directly, then asserts:
- The document was actually written to MongoDB (before the fix it was silently dropped)
- The added edge appears in `data.edges`
- The stored document contains **no empty-dict values** anywhere in `data`

**Test 2 — cross-pod round-trip:**
Mirrors the real production flow — `_write_to_db` seeds the full document (first-insert path), then `_put_node_atomic` applies an edge delta, then a **fresh `MongoBackend` with a cold L1 cache** (simulating a second pod) reads the anchor back and asserts the archetype fields survive intact.

Both new tests pass. 
---

## Checklist

- [x] Root cause identified and confirmed via reproduction
- [x] Fix implemented in `_put_node_atomic` only (no other write paths affected)
- [x] Fix is lossless — deserialization restores stripped empty dicts via `default_factory`
- [x] Atomic edge-merge semantics preserved (still uses aggregation pipeline)
- [x] Two regression tests added and passing
- [x] No regressions in existing tests

Closes #6076.


akindu-k:
# Live Verification: Fix for MongoDB Error 40180

## What was broken

`_put_node_atomic` uses MongoDB's aggregation-pipeline `$set` syntax (required for atomic `$setUnion`/`$setDifference` edge merging). That syntax rejects any empty embedded document with error 40180. `Permission.roots.anchors` always serialises as `{}` by default — so **every edge write through this path failed silently**. The error was caught and swallowed internally, the API returned 200, but the edge delta was never written to MongoDB.

## The fix

`_strip_empty_dicts()` recursively removes empty-dict leaves from the serialised node payload before it enters the pipeline `$set`. `anchors: {}` → stripped → MongoDB accepts the write.

---

## Live before/after test

### Setup
```
MongoDB 7.x via Docker · jac-scale v0.2.19 · repro walker: root ++> SomeNode()
```

### Before (main branch)

Server logs contained:
```
ERROR - MongoDB _put_node_atomic failed: Invalid $set :: caused by ::
an empty object is not a valid value. Found empty object at path data.access.roots.anchors
```

Edge mutations were silently discarded. `_write_to_db` (first insert, regular `$set`) still ran for new nodes, masking the failure.

### After (fix branch)

**Server logs:** zero `_put_node_atomic failed` / error 40180 entries.

**MongoDB after two walker runs + server restart (cold L1 cache):**

| type | edges_count | note |
|---|---|---|
| Root | 2 | one edge per run, both persisted ✓ |
| GenericEdge | 0 | written correctly ✓ |
| SomeNode | 1 | present across restart ✓ |

**`anchors: {}` still appears in documents written by `_write_to_db`** — that is expected and harmless. Regular `$set` (non-pipeline) has no issue with empty dicts. The Serializer reconstructs the empty dict on load whether the key is present or absent, so runtime behaviour is identical.

### The decisive check

The data survived a full server restart. After restart the L1 in-memory cache is empty; the server must reload everything from MongoDB. Both `SomeNode` and the edge connecting it to root were present — confirming `_put_node_atomic` wrote the edge delta to MongoDB, not just to the in-process cache.

MichaelAshley2000 (approved):

kugesan1105:
**Follow-up (not a blocker):** @akindu-k

`_strip_empty_dicts` runs on `non_arch_top` but not `partial_data` — both land in the same pipeline `$set` (memory_hierarchy.mongo.impl.jac:773), so an empty-dict archetype field (`has settings: dict = {}`) writes `data.archetype.settings = {}` and re-triggers 40180. Strip `set_doc` at the choke point instead, so no source can sneak one back in.

akindu-k:
> **Follow-up (not a blocker):** @akindu-k
> 
> `_strip_empty_dicts` runs on `non_arch_top` but not `partial_data` — both land in the same pipeline `$set` (memory_hierarchy.mongo.impl.jac:773), so an empty-dict archetype field (`has settings: dict = {}`) writes `data.archetype.settings = {}` and re-triggers 40180. Strip `set_doc` at the choke point instead, so no source can sneak one back in.

Relevant Issue: #6167

PR #6110, opened 2026-05-22
Fix: false E1002/E1053 errors when Union[] uses string forward references

ThiruvarankanM:
### Problem

When a type alias is defined using `Union[]` with type names written as strings
(forward references), the type checker misreads those strings as plain text values
instead of the actual types they refer to. This causes a flood of false errors on
valid, correctly written code.

Errors seen:
```
error[E1002]: Cannot return Concat, expected Literal["Doc"] | Literal["Text"] | ...
error[E1053]: Cannot assign Text to parameter of type Literal["Doc"] | Literal["Text"] | ...
```

~1,440 false errors in a single file — all on code that is correct.

### Root Cause

In `doc_ir.jac`, `DocType` is defined like this:

```jac
glob DocType = Union[
    ('Doc', 'Text', 'Line', 'Group', 'Indent', 'Concat', 'IfBreak', 'Align')
];
```

The strings `'Doc'`, `'Text'` etc. are **forward references** — they name types defined
later in the same file. This is valid and well-supported in Jac for regular type
annotations like `has x: 'Doc'` and `-> 'Doc'`.

However, the type checker had a blind spot: when it saw these same strings inside
`Union[(...)]`, it treated them as plain text values (e.g. the literal word "Doc")
instead of looking them up as type names. So `DocType` resolved to
`Literal["Doc"] | Literal["Text"] | ...` — a union of text strings — instead of
`Doc | Text | ...` — a union of actual types.

Every method that correctly returned a `Doc` object was then flagged as returning
the wrong thing, because the checker expected the word "Doc", not a `Doc` object.

### Evidence

```
Before fix: Union[('Doc', 'Text')] → Literal["Doc"] | Literal["Text"]   ← wrong
After fix:  Union[('Doc', 'Text')] → Doc | Text                          ← correct
```

| File                     | Before   | After | Fixed  |
|--------------------------|----------|-------|--------|
| doc_ir_gen_pass.impl.jac | ~1,440   | 18    | ~1,422 |

### Fix

The type checker already had a function that resolves string names to real types —
used correctly in every other annotation context. It was simply never called inside
the `Union[]` handler. The fix wires it in: before treating a string as a text value,
first check if it is a known type name and resolve it.

### Result

Before fix:
```
jac check jac/jaclang/compiler/passes/tool/impl/doc_ir_gen_pass.impl.jac
→ ~1,440 errors
```

After fix:
```
jac check jac/jaclang/compiler/passes/tool/impl/doc_ir_gen_pass.impl.jac
→ 18 errors  (real issues only)
```

Screenshot 2026-05-22 at 12 15 07


kugesan1105:
### Follow-up PR
- The fixture only covers `Union[('Foo','Bar')]`. It should also assert `Optional['Foo']`, single-element `Union['Foo']`, mixed `Union['Foo', Bar`], and the negative case (`Union[('red','green')]` of true literals stays Literal). Those extra cases are exactly what expose the gap.
```jac
"""Optional['Foo'] / single-arg Union['Foo'] should resolve the string as a type name, not Literal["Foo"]."""

import from typing { Union, Optional }

obj Foo {}

# Single string argument (no tuple) takes the non-TupleVal branch of the
# Union/Optional special-form handler, which currently lacks forward-ref
# resolution — so the string collapses to Literal["Foo"].
glob OptType = Optional['Foo'],
     UniType = Union['Foo'];

def make_opt -> OptType {
    return Foo();  # Wrong result: Cannot return Foo, expected Literal["Foo"] | NoneType
}

def make_uni -> UniType {
    return Foo();  # Wrong result: Cannot return Foo, expected Literal["Foo"]
}

```

PR #6130, opened 2026-05-23
fix(jir-cache): verify source_content_hash on read to catch git-restored stale JIRs

asimsalim:
## Summary

- `is_module_cache_valid` relied solely on mtime to decide if a cached JIR was still valid. git operations (checkout, stash-pop) can restore files with timestamps **older** than the cached JIR, causing the mtime check to pass even though the source content had changed.
- This is the stale-cache bug noted in #6079: *"stale project-local `.jac/cache` that served pre-change codegen even after clearing `~/.cache/jac/jir`"*.
- The JIR header has always stored `source_content_hash` (CRC32 of head source) but it was never checked on read — only written.

## What changed

**`jac/jaclang/jac0core/jir.jac` — `is_module_cache_valid`**

Added a content-hash fallback that runs only when the mtime fast-path says "valid":
1. mtime check unchanged — still the primary signal, no extra I/O on a normal cache miss
2. After all mtime checks pass, compute `CRC32(source file bytes)` and compare to `header.source_content_hash`
3. Guard `!= 0` skips the check for JIRs written by old compilers where source_hash was zero

**`jac/jaclang/jac0core/impl/jir_passes.impl.jac` — `JirWriter.write_module`**

Changed write-side hash from `CRC32(module.source.code.encode())` to `CRC32(Path(file_path).read_bytes())` so read and write hash the same bytes. Falls back to the text-encode path when `module.source.file_path` is not set (in-memory compilation).

**`jac/tests/compiler/test_importer.jac`**

New regression test: compile a file, write JIR, change file content, restore mtime to the old value via `os.utime`, assert `is_module_cache_valid` returns `False`. Directly pins the scenario that was silently serving stale bytecode.

## Test plan

- [x] `python3.12 -m pytest jac/tests/compiler/test_importer.jac -k "cache" -v` — new test + existing cache round-trip both pass
- [x] `python3.12 -m pytest jac/tests/compiler/test_importer.jac jac/tests/project/ -q` — 268 passed
- [x] Pre-existing `test_jir_cache_invalidated_on_version_mismatch` failure in `test_server.jac` confirmed pre-existing on main

Closes the stale-cache issue noted in #6079.


marsninja:

PR #6131, opened 2026-05-23
feat(native): route lexer through native pathway + fix five native runtime bugs

marsninja:
## Summary

Makes the bootstrapped Jac **lexer** runnable as native-compiled code (when the full compiler is precompiled/accelerated), and fixes the native-runtime bugs that blocked it. The native lexer now drives `parser.parse()` to ASTs **identical** to the Python lexer — including f-strings, nested f-strings, and the Python↔native interleave. The parser itself remains out of scope (a follow-up: it must be split at the unitree boundary).

Native routing is **opt-in** behind `JAC_NATIVE_LEXER=1`; the default path is unchanged pure-Python.

## Lexer decoupling
- The lexer no longer constructs `unitree` `Alert`s inline — it accumulates marshal-safe `LexDiag` / `CommentData` records (removing the lexer's coupling to the Python `unitree` module).
- New free `tokenize_full() -> (tokens, comments, diags, error_count)`. `parse()` lexes through it and **replays diagnostics into `prog`** on the Python side.

## Native runtime fixes
1. **Return-tuple refcount rooting.** Returning `obj.field` (a borrowed list pointer) inside a tuple freed the field at the owning local's scope-exit. Borrowed pointer elements packed into a returned tuple are now retained.
2. **Global init in the acceleration path.** `install_native_wrappers` never ran `__jac_glob_init*`, so native module globals (e.g. `KEYWORDS`) were uninitialized → segfault. Added `init_native_globals()`.
3. **`list[T]` negative indexing.** Subscript raised `IndexError` for any index < 0 instead of Python's `idx += len` (crashed native f-string lexing via `mode_stack[-1]`). Now normalizes and bounds-checks both directions.
4. **Enum-typed function returns.** `lookup_keyword(str) -> TokenKind | None` was marshalled as a raw int by the native wrapper (enums were mapped for struct fields, not function returns), so the hot-swapped function broke the Python lexer path. Added `ret_enum_type` to the function layout; `marshal_value` now maps ordinal → enum member and the `-1` sentinel → `None`.
5. **List methods on object-field / complex targets.** Only `append` was dispatched for complex targets; `self.stack.pop()` was a silent no-op, leaving the lexer's `mode_stack` un-popped and breaking native f-string lexing after the expression `}`. Complex-target list methods now route through the same dispatch as simple-name targets.

## Tests
- `runtime_errors.na.jac`: corrected negative-index error test (`[-10]`) + `test_list_negative_valid` (`[-1] == 30`).
- `list_field_methods.na.jac` + "list method on object field (pop)".
- "native tokenize matches python lexer for fstrings".
- 377 native tests pass; the Python parser/formatter/bootstrap path is unchanged and green.


marsninja:
## Benchmark + end-to-end wiring analysis for the native lexer

Ran the native lexer through both a micro-benchmark and the real `install_native_wrappers` → `parse()` path to characterize whether routing the lexer through native space actually pays off, and what's still needed to make it usable end-to-end.

### Performance

| | Built-in test (1.1 KB) | Corpus (872 KB, 4 real `.impl.jac` files) |
|---|---|---|
| Python `Lexer.tokenize()` | 128 ms | 913 ms |
| Native (LLVM via ctypes) | 26 ms | 179 ms |
| **Raw tokenize speedup** | **4.99x** | **5.11x** |

Raw tokenization is a consistent **~5x** win (marshalling overhead included). Two caveats on the end-to-end impact:

- **Lexing is ~35% of `parse()`** (913 ms of 2,599 ms on the corpus), and a smaller slice of a full compile — so realistic end-to-end gains are well under 5x.
- The ~1.5s native compile runs async and only amortizes over heavy lexing. A one-shot `jac run hello.jac` finishes before the swap completes and never benefits — the win is for long-lived hosts (server, LSP, watch, batch builds).

### Correctness through the real `parse()` path

Going through `install_native_wrappers` (which passes the module namespace so `build_enum_maps` resolves the Python `TokenKind` class):

| | Native vs Python |
|---|---|
| Token kinds → `TokenKind` enum members | ✅ correct (ordinal→member remap works) |
| Token values, valid code | ✅ identical AST |
| Comments (secondary list) | ✅ preserved, strings intact |
| **Lex diagnostics + `error_count`** | ❌ native returns **0 errors / empty diags** for illegal input |

Note: an earlier `'int' object has no attribute 'value'` crash I hit was a **benchmark artifact** — calling `build_enum_maps(layout, {})` with an empty namespace falls back to raw ordinals. Through the real wrapper path, kind remap, token values, and comments all work.

The one real correctness blocker: feed `tokenize_full` a source with illegal chars (`\x00 @ \``) and native returns `error_count=0, diags=[]` while Python reports 2 `E0105` errors — **malformed input is silently accepted**. `Lexer.error()` does `self.error_count += 1` and `self.diags.append(...)` ([lexer.impl.jac:180-192](https://github.com/jaseci-labs/jaseci/blob/feat/native-lexer-integration/jac/jaclang/jac0core/parser/impl/lexer.impl.jac#L180-L192)); neither survives native execution. This looks like the same object-field / list-field-mutation class as the bug fixed in this PR, not a marshalling-out problem.

### Suggested plan to finish the wiring

1. **Fix the native diagnostic path (blocker).** Trace whether the illegal-char branch reaches `error()`, then whether `self.error_count += 1` and `self.diags.append(...)` persist on the instance. `self.diags.append` is the prime suspect.
2. **Add error-case coverage.** Every test in `test_native_lexer.jac` only checks *valid* sources and compares `.value` — which is why this slipped through. Assert native `tokenize_full` matches the Python `Lexer` on `diags`/`error_count` for malformed input.
3. **Verify secondary-list string rooting under GC pressure** (the `parser.jac:626-631` caveat) once diags are populated.
4. **Route everything through `install_native_wrappers`** so enum maps are always built from a namespace containing `TokenKind`; never wrap with an empty namespace. Then the `JAC_NATIVE_LEXER` env gate can become the default.
5. **Decide activation separately from correctness:** enable autonative by default only for long-lived hosts and/or cache the compiled engine to disk (AOT) so the ~1.5s isn't paid per process; keep the Python fallback for one-shot CLI.

Net: marshalling-*in* (kinds, values, comments) already works through the real path. The remaining gate is one functional bug in the native error path plus an operational decision about *when* the swap is worth engaging.

PR #6155, opened 2026-05-25
feat(jac-scale): Loki + Alloy log pipeline for K8s deploys + microservice wire-in (M-14.a)

MusabMahmoodh:
## Summary

Adds a centralised log collection stack (Loki + Grafana Alloy) to jac-scale's Kubernetes target so deploys can pipe pod logs to Loki without leaving the cluster — works for both the monolith all-in-one target and the microservice K-track target.

Standalone PR off `main` (no upstream dependency). Supersedes the Promtail approach in #5731 — Alloy supersedes Promtail because Promtail went EOL on 2026-03-02. Credit to @juzailmlwork for the original Loki + Grafana dashboard groundwork (Co-Authored-By trailer on the foundation commit).

### 1. `feat(jac-scale): Loki + Grafana Alloy log pipeline for K8s deploys`
- Loki StatefulSet + Service (filesystem-backed, single-binary)
- Grafana Alloy v1.6.0 DaemonSet, River-syntax config tailing `/var/log/pods/*` via `discovery.kubernetes` + `loki.source.file` → Loki
- `kubernetes_config.loki_enabled` flag (default False) keeps the pipeline opt-in
- `--storage.path=/tmp/alloy` sidesteps an Alloy v1.6 remotecfg quirk where `mkdir` under a mounted emptyDir fails with EACCES
- Reference docs

### 2. `feat(jac-scale): wire microservice K-track target into log pipeline (M-14.a)`
- `[plugins.scale.microservices.logs].enabled = true` opt-in
- `MicroserviceK8sTarget._deploy_observability()` reuses the same `MonitoringDeployer` after `apply_manifests()`
- Symmetric `MonitoringDeployer.destroy()` in `destroy()` so `jac stop --scale` tears down

### 3. `test(jac-scale): coverage for centralised logs (M-14.a)`
- `test_deploy_k8s`: Alloy DaemonSet manifest is produced when `loki_enabled=True`
- `test_config_loader`: round-trip + default cases for the new `logs` block
- `tests/fixtures/k8s_e2e` + `scripts/k8s_microservice_real_e2e.sh`: minimal microservice app + driver script for the real-cluster Ingress → gateway → service → Loki e2e

## Test plan

- `jac test test_config_loader.jac` — 5 tests pass (covers new `logs` block round-trip + default-empty branch)
- `jac test test_deploy_k8s.jac` — 11/13 pass; the 2 failures (`test_deploy_all_in_one`, `test_early_exit`) require a real k8s API and fail identically on `main` without my changes
- Validated end-to-end on real EKS (2026-05-20): 13/13 monitoring pods Ready, Loki returns query results for the namespace's streams via the real-app e2e workflow

## Out of scope (follow-up PRs)

- M-14.b structured JSON log emission (separate PR)
- A-05a in-admin Logs UI (separate PR, depends on this)
- A-05c admin JSON endpoints on gateway (#6153, independent, already opened)
- K8S_APP_NAME pod env injection (#6152, independent, already opened)


jac-scale/jac_scale/targets/kubernetes/microservice/target.jac
```diff
@@ -164,6 +184,126 @@ obj KubernetesMicroserviceTarget(KubernetesTarget) {
...
+    the deploy - the operator will see the FailedCreate event anyway
+    and can label the namespace by hand."""
+    def _ensure_privileged_namespace(core_v1: any, namespace: str) -> None {
+        try {
+            ns = core_v1.read_namespace(name=namespace);
+            labels: dict = dict(getattr(ns.metadata, "labels", {}) or {});
+            current: str = labels.get("pod-security.kubernetes.io/enforce", "");
+            if current == "privileged" {
+                return;
+            }
```
udithishanka, line 278: this silently downgrades a stricter namespace. if the operator has set enforce=restricted or baseline for compliance, this patches it down to privileged and every pod in the namespace loses that protection, not just the hostPath DaemonSets. the doc comment a few lines up says 'we overwrite only when missing or set to something stricter' which is exactly the unsafe direction.

safer to only touch the label when it's missing entirely, leave anything else alone, and log a clear warning so the operator can opt in by hand.

```suggestion
            current: str = labels.get("pod-security.kubernetes.io/enforce", "");
            if current {
                if self.logger {
                    self.logger.warn(
                        f"namespace '{namespace}' has pod-security.kubernetes.io/enforce={current}; "
                        f"hostPath DaemonSets (node-exporter, Alloy) may be rejected. "
                        f"Relabel manually with `kubectl label namespace {namespace} "
                        f"pod-security.kubernetes.io/enforce=privileged --overwrite` if you want to enable them."
                    );
                }
                return;
            }
```

udithishanka (changes requested):
1 thing to address.

jac-scale/jac_scale/targets/kubernetes/microservice/target.jac
MusabMahmoodh: Good catch - my comment said the opposite of what the code did. Pushed 440535834: only stamp when the label is missing entirely; any existing value (restricted / baseline / privileged / custom) is left alone and the operator gets a warn with the manual relabel command. No more silent downgrade.

udithishanka (commented):
1 more.

jac-scale/jac_scale/targets/kubernetes/microservice/target.jac
```diff
@@ -153,6 +157,22 @@ obj KubernetesMicroserviceTarget(KubernetesTarget) {
...
         }
 
+        # Observability stack (M-14.a): opt-in centralised logs via
+        # `[plugins.scale.microservices.logs].enabled`. Surfaces a failure
+        # in `details` but does not fail the deploy - the app is healthy
+        # without logs; the operator can re-run once Loki settles.
+        try {
+            self._deploy_observability(app_name);
+        } except Exception as e {
+            bundle["observability_error"] = str(e);
```
udithishanka: broken observability stack here returns success=True with the default 'Applied N service Deployment+Service pair(s)' message. the error lives in bundle['observability_error'] which is buried inside DeploymentResult.details, so a user watching stdout assumes everything worked even when Loki is CrashLooping, the grafana/alloy:v1.6.0 image pull fails in a private network, or the alloy clusterrole create gets RBAC-denied.

fine to keep success=True since this is explicitly non-fatal, but the failure needs to surface in the message field too. track the exception in a local and append it to the return message below.

MusabMahmoodh: Addressed in 61e91d8ea — the wrapping try/except now tracks the exception in a local and the success message gets a 'NOTE: observability stack deploy failed - `<error>`. The app is running but pod logs are not being shipped to Loki. Re-run `jac start --scale` once the underlying issue is fixed.' appended when set. Kept success=True per your call since this is explicitly non-fatal; just the message field needed to surface it for users watching stdout.

PR #6163, opened 2026-05-25
feat(code-intel): cross-run structural index cache and lazy load

marsninja:
## Summary

Speeds up `CodeIntelligence` (and the `jac ai` agent) cold start by persisting and reusing each module's parse-only structural index across runs, and by deferring the expensive type-checked pipeline until it's actually needed.

### Cross-run structural index cache
- Each module's flat structural index — archetypes (with fields, member signatures, base classes) and walker entry points, no cross-module pointers — is persisted to a `.symidx.jir` sibling of the module's bytecode JIR, carrying a single `SEC_SYMINDEX` TLV section.
- A later run serves `project_map` and structural queries straight from the on-disk index without re-parsing.
- Freshness is keyed on the source's content hash plus its spliced `.impl.jac`/`.test.jac` annexes (header `source_content_hash`), so no transitive-dependency tracking is needed and only edited modules are rebuilt.
- A schema version guards the payload shape; the JIR header already invalidates on jaclang/Python version change.

### Lazy mode
- New `lazy` flag on `CodeIntelligence`: `load` builds only the cheap, cache-backed structural index and defers the full type-checked pipeline until the first symbol/diagnostic query (`_ensure_full`).
- A session that only reads the structural map (`project_map`, walkers) never pays semantic-analysis cost; a warm cache makes startup nearly free.
- The `jac ai` agent now loads lazily, and the live view reports how many modules were served from cache vs re-parsed.

## Test plan
- New tests in `test_code_intel.jac`:
  - structural index is persisted and served across loads without re-parsing
  - cache invalidates only the edited module
  - lazy mode serves the map from cache and type-checks only on demand


marsninja:
## Benchmark: what actually drives the speedup

Measured `project_map` (the `archetypes()` query) cost per launch on a synthetic project of **80 modules / 560 archetypes** (each node with 6 fields + 4 methods), 3 reps, timing load + query in a warmed interpreter so the numbers isolate `CodeIntelligence` work rather than constant Python/compiler import overhead:

| Scenario | median | vs before |
|---|---:|---:|
| full mode (before) — full semantic pipeline | 5893 ms | 1× |
| lazy cold (after, first-ever launch) — parse-only + writes index | 43 ms | **137× faster** |
| lazy warm (after, later launch) — served from on-disk index | 43 ms | **137× faster** |
| lazy full upgrade (when a *type* query is used) | 5974 ms | ~1× (deferred, not saved) |

Warm-cache load confirmed serving from disk: `recompiled=0, cached=80/80`.

### Honest decomposition

The ~137× win comes from **lazy mode deferring the type-check pipeline**, not primarily from the disk cache. Breaking down the ~43 ms structural floor:

- discovery (`rglob`): **0.6 ms**
- cache read + JSON deserialize of all 80 modules: **37 ms**
- a fresh parse of all 80 modules: **~37 ms** (essentially the same)

So on parse-cheap Jac source the cross-run `SEC_SYMINDEX` cache is roughly **perf-neutral vs just re-parsing** (parse ≈ deserialize); its value is persistence/correctness, and it pulls ahead only when parsing is the bottleneck (very large/numerous files). The dominant lever is skipping `SemanticAnalysis + TypeCheck`.

### Tradeoff

A session that needs a type query (`find_symbol`, `find_uses`, diagnostics, or any write → `write_feedback`) pays the full pipeline once, on demand (~5.3 s here) — no worse than before, just deferred. So:

- `jac code map` / `jac code walkers` (one-shot, never type-check): full ~137× win every run.
- `jac ai` agent: instant `project_map` / orientation (~40 ms instead of ~6 s), paying type-check only when the first symbol query or write happens.

### Note for reviewers

Because the on-disk cache is ~neutral vs re-parse on typical source, most of the speedup is already there from **lazy mode alone**. The `SEC_SYMINDEX` cache earns its keep on large codebases and implements the issue as written — but if we want to keep surface area minimal, lazy mode is doing the heavy lifting and the cache could be dropped or deferred to a follow-up. Open to either direction.

PR #6183, opened 2026-05-26
# fix(serializer): recover non-Optional fields stored as None instead of quarantining (#6181)

### LINKED ISSUE #6181: Persistence: `None` on non-Optional archetype field misreported as 'class unresolvable', cascade-quarantines whole subgraph
## Summary

Persisted anchors get cascade-quarantined with a misleading "class unresolvable" error when a non-Optional archetype field has a stored value of `None`. The class is in fact importable and other live anchors of the same archetype work. The bug is two-part:

1. `Serializer._deserialize_archetype` raises `TypeError("... stored value is None but field type ... does not allow None")` when a stored field value is `None` and the declared type isn't Optional.
2. The caller in `SqliteMemory` (and the Mongo backend) catches that and quarantines the anchor with reason `class <module>.<Class> unresolvable`. Same error text, two unrelated causes.

The cascade-quarantine introduced in `#6003` then drags every connected node and edge into `_anchors_quarantine`, so one stale anchor can park thousands of unrelated docs and break unrelated walker flows.

Hit this on jac-builder-dev today: one old `MainAgent` (`jac_coder.core.nodes.MainAgent`) anchor had `_conv: null` in its persisted archetype. `_conv` is declared on the class as `has _conv: list[dict] = []` (non-Optional). On reload the deserializer raised TypeError, the anchor was quarantined as "class unresolvable", and 1773 connected docs were cascade-quarantined with it. Walker calls into the system-root MainAgent then surfaced as:

```
NodeAnchor [<uuid>] is not a valid reference!
```

## Where the bug lives

`jac/jaclang/runtimelib/impl/serializer.impl.jac`, `_deserialize_archetype`, lines ~678-695:

```jac
(coerced, ok) = Serializer.coerce(deser_v, target);
if not ok {
    origin = get_origin(target);
    args = get_args(target);
    is_optional = (
        (origin is Union or origin is UnionType) and NoneType in args
    );
    if deser_v is None and not is_optional {
        raise TypeError(
            f"{module_name}.{type_name}.{k}: stored value is None "
            f"but field type {target} does not allow None — "
            "nested class likely removed"
        );
    }
    ...
}
```

The raise propagates to the outer try, which returns `None`. The caller in `jac/jaclang/runtimelib/impl/memory.impl.jac` line ~717-739 interprets a `None` return as "class unresolvable" and quarantines with that as the error string:

```jac
if anchor is None {
    # Either the row couldn't deserialize at all, or the archetype
    # class was unresolvable (even after alias lookup). Quarantine
    # rather than drop ...
    _quarantine(..., f"class {arch_module}.{arch_type} unresolvable", ...);
}
```

So a per-field coercion failure is reported as a class-level import failure. Operators chase the wrong thing, and the cascade-quarantine drags the rest of the subgraph in.

## Evidence (jac-builder-dev)

Quarantined `MainAgent` archetype on disk:
```json
{
  "__type__": "MainAgent",
  "__module__": "jac_coder.core.nodes",
  "_conv": null
}
```

Live `MainAgent` in the same DB, same module, written by the same service:
```json
{
  "__type__": "MainAgent",
  "__module__": "jac_coder.core.nodes"
}
```

The `MainAgent` class declares `has _conv: list[dict] = [];` (non-Optional). The persisted `null` triggers the TypeError path.

Quarantine record:
```
error: class jac_coder.core.nodes.MainAgent unresolvable
quarantined_at: 2026-05-26T09:16:28Z
fingerprint: a4c86515b978fe75
```

Of 1774 docs in `_anchors_quarantine`, exactly 1 has the `_conv: null` field. The other 1773 carry `cascade quarantine: connected node <uuid> was quarantined`, which #6003 did correctly given the bogus first-quarantine decision.

## Minimal reproducible example

This uses jaclang's `SqliteMemory` so there's no Mongo dependency.

```jac
"""repro_conv_quarantine.jac, run with `jac run repro_conv_quarantine.jac`.

Demonstrates that a non-Optional field with a stored value of `None`
causes the runtime to quarantine an otherwise-valid anchor as
'class unresolvable'.
"""
import json;
import sqlite3;
import tempfile;
import from pathlib { Path }
import from jaclang.runtimelib.memory { SqliteMemory }

node Foo {
    has bar: list[dict] = [];
}

with entry {
    db_path = Path(tempfile.mkdtemp()) / "repro.db";

    # 1. Persist a Foo node normally.
    mem = SqliteMemory(str(db_path));
    foo = Foo();
    mem.put(foo.__jac__);
    mem.sync();
    foo_id = str(foo.__jac__.id);
    mem.close();

    # 2. Mutate the stored archetype dict so `bar` is null instead of [].
    #    This models what a legacy schema (or a buggy past write path) left in Mongo.
    conn = sqlite3.connect(str(db_path));
    cur = conn.cursor();
    cur.execute("SELECT data FROM anchors WHERE id = ?", (foo_id,));
    row = cur.fetchone();
    data = json.loads(row[0]);
    data["archetype"]["bar"] = None;
    cur.execute("UPDATE anchors SET data = ? WHERE id = ?", (json.dumps(data), foo_id));
    conn.commit();
    conn.close();

    # 3. Reopen and try to load. The anchor will be quarantined.
    mem2 = SqliteMemory(str(db_path));
    result = mem2.get(foo_id);

    print("result:", result);                              # None
    conn = sqlite3.connect(str(db_path));
    cur = conn.cursor();
    cur.execute("SELECT error FROM anchors_quarantine WHERE id = ?", (foo_id,));
    err = cur.fetchone();
    print("quarantine error:", err[0] if err else "none");
    # Expected: 'class __main__.Foo unresolvable'
    # Actual root cause: bar deserialization saw None on list[dict].
}
```

Expected (what most forward-compatible deserializers would do): `bar` falls back to the class default `[]`, the anchor loads, no quarantine.

Actual: `result is None`, anchor moved to `anchors_quarantine` with `error="class __main__.Foo unresolvable"`. The class is plainly resolvable; it was just defined on the line above.

## Why the recent fixes don't cover this

- `#5446` (persistence race): unrelated. This isn't a concurrent-write race.
- `#6003` (cascade-quarantine on node removal): works correctly here. The cascade is doing what it's told. Bug is upstream of the cascade decision.
- `#6035` (recover_all node-before-edge order): only matters if `recover_all` ran. Nothing triggers it automatically, and even if it did, the underlying deserializer would re-quarantine the same doc on every retry.
- `#6105` (empty $set dict on Mongo writes): write path, not read.

## Suggested fix

Pick one, ideally both:

1. In `_deserialize_archetype`, when coercion fails on `None` for a non-Optional field, fall back to the class default value (or to a logged `None` setattr) instead of raising. The current behaviour drops the entire anchor over what is, in practice, "a removed-default-of-empty-collection". Most existing collection-typed fields are exactly this shape: `has things: list[X] = []`.

2. Distinguish the quarantine reasons. When `_get_class` returns `None`, the reason is genuinely "class unresolvable". When it returns a class but `_deserialize_archetype` later raises, the reason is "field coercion failure on `<field>: <reason>`". Operators chasing schema drift today land on imports and module paths and miss the real cause. The two failure modes deserve different error strings, and arguably different quarantine policies (the second is auto-recoverable on next read once the field type is widened or the stored value rewritten).

Option 1 is the bigger win. Most occurrences I have seen on jac-builder-dev fit the "stored None on non-Optional collection" pattern, and "drop into quarantine" is too punitive a response. Option 2 is a smaller PR with no behaviour change beyond clearer ops triage.

## Workaround

Strip the offending field in `_anchors_quarantine`, move the doc back to `_anchors`, restore one connecting edge, push that edge id back onto the parent node's `data.edges`, bounce the pods. Operationally heavy and not safe under traffic.

akindu-k:
## Problem

When a persisted anchor had a `None` stored for a non-Optional field (e.g. `has _conv: list[dict] = []` stored as `null` in the DB), `_deserialize_archetype` raised a `TypeError`. That exception was caught by the outer `try/except`, the archetype was dropped, and `SqliteMemory.get()` / `MongoBackend.get()` received `None` back.

This triggered the quarantine path with the misleading label **"class X unresolvable"** — the class was perfectly importable, it was just one field value that was bad. Worse, the quarantine cascade introduced in #6003 then quarantined every `EdgeAnchor` connected to that node, breaking unrelated workflows. In the reported case: **1,773 documents quarantined over a single null field**.

The same code path was hit when a field stored a nested object whose class had been removed (`_deserialize_value` returns `None` for unresolvable types), making the problem broader than just JSON `null`.

### Minimal repro

```jac
obj Foo(Root) {
    has bar: list[dict] = [];
}
# … store Foo normally, then SET bar = null in the DB, reload →
# anchor goes to anchors_quarantine as "class __main__.Foo unresolvable"
# even though Foo is defined right above.
```

---

### Fix 1 — graceful field recovery (`serializer.impl.jac`, `serializer.jac`)

Replaced the `raise TypeError` with a class-default lookup. When `deser_v is None` and the field is non-Optional, the runtime now walks `dataclasses.fields(cls)` to find the field's `default_factory` or `default` and restores that value, then logs a `WARNING` and continues. The anchor loads normally.

```
# Before
TypeError: __main__.Foo.bar: stored value is None but field type list[dict]
           does not allow None — nested class likely removed
→ anchor dropped → quarantined as "class Foo unresolvable"
→ cascade-quarantine of 1,773 connected documents

# After
WARNING  Serializer: __main__.Foo.bar: stored None on non-Optional field
         list[dict]; restored class default.
→ anchor loads, bar = []
→ no quarantine, no cascade
```

Covers both failure modes:
- **Stored JSON `null`** — a past write bug or schema migration left the field as `null`.
- **Removed nested class** — `_deserialize_value` returns `None` when a nested object's `__type__` is no longer registered.

If a field has `None` stored and **no class default exists** (e.g. `has name: str` with no `= ...`), recovery is impossible. The anchor returns `None` and is quarantined with an accurate reason (see Fix 2).

`MISSING` added to the `dataclasses` import in `serializer.jac`.

### Fix 2 — accurate quarantine reason (`memory.impl.jac`, `memory_hierarchy.mongo.impl.jac`)

When `anchor is None` after `Serializer.deserialize()`, both the SQLite and Mongo backends now call `_get_class` to distinguish the two failure modes before writing to the quarantine table:

| Failure | Old reason | New reason |
|---|---|---|
| Class not in registry | `class X unresolvable` | `class X unresolvable` *(unchanged)* |
| Class found, field failed | `class X unresolvable` | `archetype field deserialization failed: X` |

Operators chasing schema drift now immediately know whether to look at module paths / `@archetype_alias` (class-unresolvable) or at field types and stored values (field failure). With Fix 1 in place, Fix 2 applies only to the remaining hard case: a non-Optional field with no class default.

---

## Files changed

| File | Change |
|---|---|
| `jac/jaclang/runtimelib/serializer.jac` | Add `MISSING` to `dataclasses` import |
| `jac/jaclang/runtimelib/impl/serializer.impl.jac` | Replace `raise TypeError` with default-value recovery loop |
| `jac/jaclang/runtimelib/impl/memory.impl.jac` | Distinguish "class unresolvable" from "field deserialization failed" in quarantine reason (SQLite) |
| `jac-scale/jac_scale/impl/memory_hierarchy.mongo.impl.jac` | Same quarantine-reason fix applied to MongoBackend |
| `jac/tests/runtimelib/test_persistence_migration.jac` | Update existing test + add no-default quarantine test (SQLite) |
| `jac-scale/jac_scale/tests/test_mongo_layer123.jac` | Add field-recovery and quarantine-reason tests (Mongo) |
| `docs/docs/community/release_notes/unreleased/jaclang/6183.bugfix.md` | jaclang release note |
| `docs/docs/community/release_notes/unreleased/jac-scale/6183.bugfix.md` | jac-scale release note |

---

## Testing

All tests pass:

```
jac run repro_conv_quarantine.jac

WARNING - Serializer: __main__.Foo.bar: stored None on non-Optional field
          list[dict]; restored class default.
anchor loaded: True
bar value: []
quarantined: none
ALL ASSERTIONS PASSED
```

**SQLite (`test_persistence_migration`)** — 15 tests pass, including:
- `"unresolvable nested class in field falls back to default, anchor still loads"` — anchor with bad field **loads**, field reset to class default, row **absent** from `anchors_quarantine`.
- `"non-Optional field with no default stores null quarantines only that anchor, no cascade"` — anchor with no-default null field quarantines with reason `"archetype field deserialization failed"`, neighbour anchor untouched.

**Mongo (`test_mongo_layer123`)** — 2 new tests:
- `"MongoBackend recovers null field to class default and loads anchor normally"` — null field on a node with a default loads cleanly, not quarantined.
- `"MongoBackend quarantine reason identifies field failure not class unresolvable"` — no-default null field quarantines with `"archetype field deserialization failed"` reason.

**Coercion (`test_layer3_coercion_alias`)** — 15 tests pass, no regressions.

---

## Checklist

- [x] Root cause identified and confirmed via repro
- [x] Fix 1: non-Optional `None` falls back to class default, anchor loads
- [x] Fix 2: quarantine reason distinguishes class-unresolvable from field failure
- [x] Fix 2 applied to both SQLite (`memory.impl.jac`) and Mongo (`memory_hierarchy.mongo.impl.jac`)
- [x] Cascade quarantine no longer fires for recoverable field issues
- [x] Existing tests updated to reflect correct behaviour
- [x] New SQLite test: no-default field quarantines only that anchor, no cascade
- [x] New Mongo tests: field recovery and quarantine reason
- [x] Release notes added for jaclang and jac-scale
- [x] All persistence and coercion tests pass in CI

Closes #6181.


akindu-k:
## repro_conv_quarantine.jac

```jac

"""repro_conv_quarantine.jac — verify fix for GitHub issue #6181.

A non-Optional field with a stored None must not quarantine the anchor.
The anchor should load with the field restored to its class default.

Run with: jac run repro_conv_quarantine.jac
"""
import json;
import uuid;
import from tempfile { TemporaryDirectory }
import from jaclang.jac0core.archetype { NodeAnchor, Root }
import from jaclang.runtimelib.memory { SqliteMemory }
import from jaclang.runtimelib.serializer { Serializer }

# Use Root base so the archetype is registered and _get_class can resolve it.
obj Foo(Root) {
    has bar: list[dict] = [];
}

with entry {
    with TemporaryDirectory() as tmpdir {
        import os;
        import sqlite3;
        db_path = os.path.join(tmpdir, "repro.db");

        # 1. Set up schema and build a valid serialized row with bar=null.
        mem = SqliteMemory(path=db_path);
        mem._ensure_connection();

        foo_anchor = NodeAnchor(archetype=Foo(), edges=[]);
        foo_anchor.persistent = True;
        foo_id = foo_anchor.id;

        # Serialize normally then corrupt bar to null.
        data = dict(Serializer.serialize(foo_anchor, include_type=True));
        data["archetype"]["bar"] = None;
        data_json = json.dumps(data);

        fp = Foo?.__jac_fingerprint__;
        mem.__conn__.execute(
            """
            INSERT INTO anchors
                (id, type, arch_module, arch_type, fingerprint, data,
                 format_version, updated_at)
            VALUES (?, 'NodeAnchor', ?, 'Foo', ?, ?, 1, '2026-01-01T00:00:00')
            """,
            (str(foo_id), Foo.__module__, fp, data_json)
        );
        mem.__conn__.commit();
        mem.close();

        # 2. Reload — must NOT quarantine; bar must fall back to [].
        mem2 = SqliteMemory(path=db_path);
        result = mem2.get(foo_id);

        cur = mem2.__conn__.execute(
            "SELECT error FROM anchors_quarantine WHERE id = ?", (str(foo_id),)
        );
        err_row = cur.fetchone();
        mem2.close();

        print("anchor loaded:", result is not None);
        print("bar value:", result.archetype.bar if result else "N/A");
        print("quarantined:", err_row[0] if err_row else "none");

        assert result is not None, "FAIL: anchor was quarantined instead of loaded";
        assert result.archetype.bar == [], f"FAIL: bar={result.archetype.bar!r}, want []";
        assert err_row is None, f"FAIL: anchor in quarantine: {err_row[0] if err_row else '?'}";
        print("ALL ASSERTIONS PASSED");
    }
}

```

## Terminal Output

image

jac/jaclang/runtimelib/impl/serializer.impl.jac
```diff
@@ -682,11 +682,27 @@ impl Serializer._deserialize_archetype(
...
+                            }
+                            break;
+                        }
+                    }
+                    logger.warning(
+                        f"Serializer: {module_name}.{type_name}.{k}: stored None "
+                        f"on non-Optional field {target}; restored class default."
                     );
+                    setattr(arch, k, recovered);
+                    continue;
```
udithishanka, line 712: edge case: if the field has neither default_factory nor default (both MISSING), recovered stays at its initial None and you setattr None on a non-Optional field. that's exactly the corrupt state the fix is trying to prevent, just made silent instead of loud. operationally worse than quarantine because nothing flags it. when no default is available, fall back to returning None so the caller quarantines.

```suggestion
                    recovered: object = None;
                    has_default = False;
                    for fd in fields(cls) {
                        if fd.name == k {
                            if fd.default_factory is not MISSING {
                                recovered = fd.default_factory();
                                has_default = True;
                            } elif fd.default is not MISSING {
                                recovered = fd.default;
                                has_default = True;
                            }
                            break;
                        }
                    }
                    if not has_default {
                        logger.warning(
                            f"Serializer: {module_name}.{type_name}.{k}: stored None "
                            f"on non-Optional field {target} and no class default "
                            "available; cannot recover."
                        );
                        return None;
                    }
                    logger.warning(
                        f"Serializer: {module_name}.{type_name}.{k}: stored None "
                        f"on non-Optional field {target}; restored class default."
                    );
                    setattr(arch, k, recovered);
                    continue;
```

jac/tests/runtimelib/test_persistence_migration.jac
```diff
@@ -411,14 +411,18 @@ test "unresolvable nested class in field is quarantined, not silently loaded as
         good = mem.get(uuid.UUID(good_id));
         assert good is not None , "valid row must load";
         assert cast(_TypedProfile, good.archetype).score == 42;
-        # Bad row must be quarantined, not returned.
-        assert mem.get(uuid.UUID(bad_nested_id)) is None , "unresolvable nested class must quarantine";
+        # Row with unresolvable nested class must LOAD (not quarantine) with score
+        # reset to the declared default (0), not None and not quarantined.
+        recovered = mem.get(uuid.UUID(bad_nested_id));
+        assert recovered is not None , "anchor with bad field must load after fix";
+        assert cast(_TypedProfile, recovered.archetype).score == 0 , (
```
udithishanka: nit: PR body and release note both list 'stored JSON null' and 'removed nested class' as the two cases covered. the test only exercises the second one (via the RemovedClass __type__ dict). both go through the same deser_v is None branch today so they're equivalent, but a direct null assertion would lock in the contract and protect against future changes in _deserialize_value semantics. worth adding a third row inserted with score=None and parallel assertions: not in q_ids, score restored to 0.

udithishanka (changes requested):
1 thing to fix, plus 2 small follow-ups.

one other note that doesn't fit inline since the file isn't in the PR: jac-scale's MongoBackend.get in jac_scale/impl/memory_hierarchy.mongo.impl.jac around L382 still hardcodes the misleading 'class unresolvable' message for the same anchor-is-None branch. Fix 2 only landed in SqliteMemory. With Fix 1 in place that path is mostly unreachable, but if anything else ever returns None from deserialize, prod ops on MongoBackend get the wrong message. Same monorepo, worth syncing.

udithishanka (approved):

PR #6186, opened 2026-05-26
feat(jac-scale): admin Deployments page backed by Prometheus API

kashmithnisakya:
## Description

Adds a **Deployment** section to the jac-scale admin portal:

- **Workloads** - every Deployment / StatefulSet / DaemonSet / Pod in the namespace, with health status, ready/desired replicas, live CPU & memory, restarts and age. Filter by kind.
- **Usage** - per-workload CPU & memory history graphs with a 1h / 6h / 24h / 7d range selector.
- **Summary cards** - pods running/pending/failed, ready nodes, traffic (req/s), error rate, p95 latency.

All data comes from the in-cluster Prometheus the monitoring stack already provisions, no new infrastructure. Also fixes app-metrics scraping (correct port + Basic Auth) and adds cAdvisor scraping so per-pod CPU/RAM is available.

https://github.com/user-attachments/assets/560c5c06-c30e-4837-ba97-063da1e67072


MusabMahmoodh (commented):
Solid PR overall - graceful-degradation path is clean, the threadpool-offload + DRY helpers from the prior review round read well, tests cover the helpers thoroughly. Four things worth a look inline, two are real and two are sanity checks.

jac-scale/jac_scale/admin/impl/workloads.impl.jac
```diff
@@ -0,0 +1,693 @@
...
+    ] {
+        rows = prom_query(_scoped(spec[0], ns_filter)) or [];
+        for r in rows {
+            nm = r.get("metric", {}).get(spec[1], "");
+            if nm {
+                names.append(nm);
+            }
+        }
+    }
+    if range_minutes <= 60 {
```
MusabMahmoodh, line 649: `range_minutes` is taken from query params with no upper cap. An admin passing `range_minutes=10000000` falls through to the `else` branch (step=1h) and asks Prometheus for ~19 years of buckets per pod across two range queries. Even though Prometheus will eventually refuse or truncate, the request still chews compute. Cap at 10080 (the 7d max the UI exposes) at the top of this handler? Same for the `<= 0` case, which gives `start == now` and a zero-width window.

jac-scale/jac_scale/targets/kubernetes/monitoring.jac
```diff
@@ -53,6 +56,9 @@ obj MonitoringDeployer {
             "    static_configs:\n"
             f"      - targets: ['{app_name}-service:{service_port}']\n"
             "    metrics_path: '/metrics'\n"
+            "    basic_auth:\n"
+            f"      username: '{self.k8s_config.admin_username}'\n"
+            f"      password: '{self.k8s_config.prometheus_admin_password}'\n"
```
MusabMahmoodh: `prometheus_admin_password` is being interpolated straight into the Prometheus ConfigMap. Anyone with `configmaps:get` in the namespace can read the admin password from there. Prometheus supports `basic_auth.password_file:` pointing at a file mounted from a Secret, and jac-scale already provisions an admin Secret, so this is mostly mount + path change here. Default password is `Adminpassword123` so the practical risk is lower for fresh deploys, but anyone who set a real one is now exposing it via the ConfigMap.

jac-scale/jac_scale/targets/kubernetes/monitoring.jac
```diff
@@ -65,6 +71,30 @@ obj MonitoringDeployer {
...
+            # cAdvisor (per-container CPU/memory) is exposed by the kubelet, not a
+            # standalone exporter. Scrape it through the API-server node proxy so we
+            # don't need host-network access. Requires the prometheus ServiceAccount
+            # to have nodes/proxy RBAC (granted below) and its token mounted.
+            scrape_config = scrape_config + (
+                "  - job_name: 'kubernetes-cadvisor'\n"
+                "    scheme: https\n"
+                "    tls_config:\n"
+                "      ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt\n"
+                "      insecure_skip_verify: true\n"
```
MusabMahmoodh, line 84: Sanity check: the `tls_config` block on this scrape sets both `ca_file: .../ca.crt` and `insecure_skip_verify: true`. The skip overrides the ca_file, so the ca_file is effectively dead. Did you intend to keep the ca_file as documentation for the right path, or was the insecure flag left in from debugging? If the API-server cert SAN doesn't include `kubernetes.default.svc`, drop the ca_file with a comment; if it does, drop the insecure flag.

jac-scale/jac_scale/tests/test_admin_workloads.jac
```diff
@@ -0,0 +1,177 @@
...
+        assert response.status_code == 403;
+    } finally {
+        client.close();
+    }
+}
+
+test "workload detail requires admin" {
+    client = make_client();
+    try {
+        response = client.get("/admin/workloads/jac-app");
```
MusabMahmoodh: Question rather than assertion since CI is green so I'm probably missing something: commit `e8652c8` removed `admin_workload_detail` and the per-workload-timeseries endpoint. The four registered paths are now `/admin/workloads`, `/admin/workloads/list`, `/admin/workloads/pods`, `/admin/workloads/timeseries`. So `client.get("/admin/workloads/jac-app")` here and `client.get("/admin/workloads/jac-app/timeseries")` at L61 should hit unregistered routes and get 404, not 403. What routes them to 403? If nothing, these two tests can come out.

jac-scale/jac_scale/admin/impl/workloads.impl.jac
kashmithnisakya: Fixed

jac-scale/jac_scale/targets/kubernetes/monitoring.jac
kashmithnisakya: Fixed

jac-scale/jac_scale/tests/test_admin_workloads.jac
kashmithnisakya: Fixed

jac-scale/jac_scale/targets/kubernetes/monitoring.jac
kashmithnisakya: fixed

musab-mah-7 (approved):

PR #6194, opened 2026-05-26
fix(jac-scale): invalidate Redis L2 cache on cascade quarantine (#6184)

### LINKED ISSUE #6184: jac-scale: cascade-quarantine in MongoBackend doesn't invalidate Redis L2, leaving stale `data.edges` references that survive pod restarts
## Summary

`MongoBackend._mongo_quarantine` in `jac-scale` mutates MongoDB during cascade-quarantine but never invalidates the L2 Redis cache. The Redis entries for the affected nodes are left holding the pre-cascade state, including edge ids that no longer exist in Mongo. Pods read cache-first, see the stale `data.edges` array, try to load the now-deleted edge anchors, and surface:

```
EdgeAnchor [<uuid>] is not a valid reference!
```

Pod restart does not fix it because Redis is the shared L2 across all replicas; a fresh pod still reads the stale L2 entries.

This is what is blocking jac-builder-dev right now (every user, old and new). Issue #6181 is the upstream cause (the quarantine decision itself); this issue is what makes the resulting damage persist forever in the cache layer.

## Where the bug lives

`jac-scale/jac_scale/impl/memory_hierarchy.mongo.impl.jac`, `_mongo_quarantine`, lines 196-273.

Three Mongo write paths in this function, none of them invalidate Redis:

```jac
backend.collection.delete_one({'_id': id_str});                                       # line 226
backend.collection.delete_many({'_id': {'$in': orphan_edge_ids}});                     # line 259
backend.collection.update_many(                                                        # lines 260-263
    {'type': 'NodeAnchor', 'data.edges': {'$in': orphan_edge_ids}},
    {'$pullAll': {'data.edges': orphan_edge_ids}}
);
```

The `update_many` on line 260-263 is the operationally painful one. It strips the dead edge ids from every live NodeAnchor's `data.edges`. Those NodeAnchors are commonly hot anchors (root, system root) that Redis is caching. After the call, Mongo and Redis diverge:

- Mongo: `data.edges` has the dead ids removed.
- Redis: `data.edges` still contains the dead ids.

Reads through `TieredMemory.get` hit the L2 cache before the L3 Mongo backend (this is the entire point of the cache), so every subsequent traversal that walks those edges fails.

The `RedisBackend` already has a `delete(id)` method (`memory_hierarchy.redis.impl.jac` line 96). `_mongo_quarantine` just doesn't call it.

## Evidence (jac-builder-dev)

System root in Mongo, after cascade-quarantine ran:

```
$ kubectl -n jac-builder-dev exec jac-builder-dev-mongodb-0 -- mongosh ... --eval '
db = db.getSiblingDB("jac_db");
const r = db._anchors.findOne({_id: "9866fc2a-2dda-41bc-916c-7037237183e8"});
print("Mongo edges count:", r.data.edges.length);
print("Has dead edge?", r.data.edges.includes("7a1d285c-a682-47c3-a7a8-89953c348397"));
'
Mongo edges count: 189
Has dead edge? false
```

Same root in Redis, same moment:

```
$ kubectl -n jac-builder-dev exec <redis-pod> -- redis-cli -a ... \
    EXISTS "anchor:9866fc2a-2dda-41bc-916c-7037237183e8"
1

$ kubectl -n jac-builder-dev exec <redis-pod> -- redis-cli -a ... \
    GET "anchor:9866fc2a-2dda-41bc-916c-7037237183e8" \
    | grep -c "7a1d285c-a682-47c3-a7a8-89953c348397"
1

$ kubectl -n jac-builder-dev exec <redis-pod> -- redis-cli -a ... \
    EXISTS "anchor:7a1d285c-a682-47c3-a7a8-89953c348397"
0
```

Mongo says the system root has 189 clean edges, no reference to `7a1d285c`. Redis still has the pre-cascade copy of the same root with `7a1d285c` in its edges array. The edge anchor itself doesn't exist in either store (cascade deleted it from Mongo, was never in Redis to begin with because reads hadn't hit it before). Result: every traversal that touches the system root fails.

## Minimal reproducible example

Self-contained, uses docker for Mongo + Redis.

`docker-compose.yml`:
```yaml
services:
  mongo:
    image: mongo:7
    ports: ["27017:27017"]
  redis:
    image: redis:7
    ports: ["6379:6379"]
```

`repro_cascade_redis.jac`:
```jac
"""Repro for Redis L2 cache not invalidated on cascade-quarantine.

Run mongo + redis via docker-compose, then:
    MONGODB_URI=mongodb://localhost:27017 \
    REDIS_URL=redis://localhost:6379/0 \
    jac run repro_cascade_redis.jac
"""
import json;
import from uuid { uuid4 }
import from jac_scale.memory_hierarchy { ScaleTieredMemory }
import from jaclang.runtimelib.archetype { NodeAnchor, EdgeAnchor }

# A minimal node type
node Foo { }
node Bar { }
edge Link { }

with entry {
    mem = ScaleTieredMemory();
    mem.postinit();

    # 1. Build root --Link--> child, persist all the way through Mongo + Redis.
    foo = Foo();
    bar = Bar();
    foo ++>:Link:++> bar;          # creates edge anchor + adds id to foo.__jac__.data.edges

    mem.put(foo.__jac__);
    mem.put(bar.__jac__);
    for e in foo.__jac__.data.edges {
        mem.put(e);
    }
    mem.commit();

    foo_id = str(foo.__jac__.id);
    bar_id = str(bar.__jac__.id);

    # 2. Warm the Redis cache for foo by reading it back. After this, Redis has
    #    anchor:<foo_id> with data.edges = [<link_id>].
    mem.__mem__.clear();           # drop L1 to force a real fetch
    refetched_foo = mem.get(foo_id);
    print("foo edges (cached path):", refetched_foo.data.edges);

    # 3. Trigger cascade-quarantine by quarantining bar.
    #    Calling _mongo_quarantine directly bypasses any pre-quarantine cache
    #    side effects and exercises exactly the code path under test.
    import from jac_scale.impl.memory_hierarchy_mongo_impl { _mongo_quarantine }
    bar_doc = mem.l3.collection.find_one({'_id': bar_id});
    _mongo_quarantine(
        backend=mem.l3,
        id_str=bar_id,
        anchor_type='NodeAnchor',
        arch_module=bar_doc['arch_module'],
        arch_type=bar_doc['arch_type'],
        fingerprint=bar_doc.get('fingerprint'),
        data=bar_doc['data'],
        error='manual repro'
    );

    # 4. Verify the divergence.
    foo_in_mongo = mem.l3.collection.find_one({'_id': foo_id});
    foo_in_redis_raw = mem.l2.redis_client.get(f"anchor:{foo_id}");
    foo_in_redis = json.loads(foo_in_redis_raw);

    print("\nMongo foo.data.edges:", foo_in_mongo['data']['edges']);    # []
    print("Redis foo.data.edges:", foo_in_redis['data']['edges']);     # still has the link id

    # 5. Read through TieredMemory the way pods do (cache-first).
    mem.__mem__.clear();
    foo_via_tiered = mem.get(foo_id);
    stale_edge_id = foo_via_tiered.data.edges[0] if foo_via_tiered.data.edges else None;
    print("\nTieredMemory.get returned foo with edges:", foo_via_tiered.data.edges);
    print("Now trying to load that edge id:", stale_edge_id);

    edge_anchor = mem.get(stale_edge_id);
    print("Resolved edge anchor:", edge_anchor);   # None -> caller raises 'not a valid reference'
}
```

Expected (correct cascade): Redis is invalidated, the second `mem.get(foo_id)` returns `data.edges = []`, no dangling reference.

Actual: Redis is not invalidated, `mem.get(foo_id)` returns `data.edges = [<link_id>]`, `mem.get(stale_edge_id)` returns `None`, and any walker traversal raises `EdgeAnchor [<link_id>] is not a valid reference`. The condition persists across pod restarts because Redis is shared.

## Production timeline (jac-builder-dev)

1. ~old date: `MainAgent` node `0ad08997-...` written with `_conv: null` (separate issue #6181 caused this to deserialize-fail later).
2. Earlier deploy at `2026-05-26T09:16:28Z`: deserialize fails, `_mongo_quarantine` quarantines the node, cascades 1 Root->MainAgent edge + 1773 connected docs. Mongo is updated, Redis is not.
3. Pods restarted at `09:43Z`. Fresh pods read the system root from Redis (cache hit), get the stale `data.edges` list with the dead edge id.
4. Every subsequent `create_session` call traverses the system root, hits the dead edge, fails with `EdgeAnchor [7a1d285c-a682-47c3-a7a8-89953c348397] is not a valid reference`.
5. Hand-fixing Mongo (which the existing skill does) is not enough. Without flushing Redis, the symptom comes right back the moment a pod re-warms the cache.

## Suggested fix

In `_mongo_quarantine`, after each Mongo mutation, invalidate Redis. Concretely:

```jac
# After: backend.collection.delete_one({'_id': id_str});
backend.l2.delete(to_uuid(id_str));

# After: backend.collection.delete_many({'_id': {'$in': orphan_edge_ids}});
for eid in orphan_edge_ids {
    backend.l2.delete(to_uuid(eid));
}

# After: backend.collection.update_many(... '$pullAll' ...);
# Find the live NodeAnchors that were just patched (the ones whose data.edges
# contained any orphan_edge_ids before the update). They need their Redis
# entries dropped so the next read repopulates from Mongo.
affected = backend.collection.find(
    {'type': 'NodeAnchor', 'data.edges': {'$in': orphan_edge_ids}}
);  # captured BEFORE update_many
for doc in affected {
    backend.l2.delete(to_uuid(doc['_id']));
}
```

The `affected` capture has to happen before the `update_many`, since after the update those edges are gone and the query returns nothing. Same for `delete_many`: capture the orphan edge ids before they are deleted (the function already does this on line 238).

The `backend` here doesn't currently hold a reference to the L2 backend; the function will need that wired through, either via `backend.l2` (if we make MongoBackend aware of its sibling) or by passing the `ScaleTieredMemory` instance instead of just the L3 backend. The cleanest shape is to move `_mongo_quarantine` onto `ScaleTieredMemory` so it has access to both tiers.

## Why the recent fixes don't cover this

- `#5446` (persistence race): different layer; concurrent writes, not stale cache.
- `#6003` (cascade-quarantine): introduced the cascade behaviour but only on the Mongo side. The cache-invalidation half is missing.
- `#6035` (recover_all order): doesn't apply unless `recover_all` is invoked, and even if it were, the corrupt Redis copy would still be served on the next read.
- `#6105` (empty `$set` dict): write path on `_put_node_atomic`, unrelated.

## Workaround

Manually flush the affected Redis keys (the live NodeAnchors that had their `data.edges` patched). For one root:

```bash
kubectl -n <ns> exec <redis-pod> -- redis-cli -a "$REDIS_PW" --no-auth-warning \
    DEL "anchor:<root-uuid>"
```

Then bounce pods to drop L1 caches.

To find all affected anchors after the fact, scan every `anchor:*` key in Redis and check whether any of its `data.edges` ids are missing from Mongo `_anchors`. Inefficient but reliable.

akindu-k:
## Problem

When `_mongo_quarantine` cascade-quarantines a `NodeAnchor` it correctly:

1. Moves the bad node to `_anchors_quarantine` in MongoDB
2. Cascade-quarantines all connected `EdgeAnchor`s
3. Strips the orphaned edge IDs from every live node's `data.edges` in MongoDB

It did **not** invalidate those documents in the Redis L2 cache. Because Redis is shared across pods and outlives individual requests, any pod that had previously cached a live node with a reference to one of those edges would continue serving the stale entry — even across restarts. On the next walker traversal the pod would call `mem.get(stale_edge_id)` and receive `None`, raising:

```
EdgeAnchor [<id>] is not a valid reference
```

## Root cause

`_mongo_quarantine` only had access to `MongoBackend` (L3). It held no reference to `RedisBackend` (L2), so it could not call `l2.delete()` after each Mongo mutation.

## Fix

**`MongoBackend`** gains a `_l2_ref` field and a `set_cache_ref(cache)` method, mirroring the existing `_mem_ref` / `set_memory_ref` pattern used for L1.

**`ScaleTieredMemory.postinit`** wires the reference immediately after L2 is initialised:

```
mongo_backend.set_cache_ref(self.l2);
```

**`_mongo_quarantine`** gains an optional `l2` parameter. Three invalidations are added, all guarded by `if l2`:

| After this Mongo write | Redis keys invalidated |
|---|---|
| `delete_one` (quarantined node) | `l2.delete(node_id)` |
| `delete_many` (orphaned edges) | `l2.delete(edge_id)` for each |
| `update_many` (`$pullAll` on live nodes) | `l2.delete(node_id)` for each affected node |

The affected-node query is executed **before** `update_many` so the edge IDs are still present in `data.edges` and the `$in` filter returns results.

All four `_mongo_quarantine` call sites inside `MongoBackend` (two in `.get`, two in `._load_anchor`) are updated to pass `l2=self._l2_ref`.

## Verification

A self-contained repro is included at `repro/repro_cascade_redis.jac` (requires `docker compose up` from `repro/docker-compose.yml`).

```bash
MONGODB_URI=mongodb://localhost:27017 \
REDIS_URL=redis://localhost:6379/0 \
jac run repro/repro_cascade_redis.jac
```

Expected output (with the fix):

```
Mongo  source.data.edges : []
Redis  source.data.edges : not cached (invalidated)
TieredMemory source.data.edges : []

ALL ASSERTIONS PASSED -- Redis is correctly invalidated on cascade quarantine.
```

Without the fix, `Redis source.data.edges` still holds the stale edge ID and the final tiered read returns a dangling reference.

## Files changed

| File | Change |
|---|---|
| `jac-scale/jac_scale/memory_hierarchy.jac` | Add `_l2_ref` field and `set_cache_ref` method to `MongoBackend` |
| `jac-scale/jac_scale/impl/memory_hierarchy.mongo.impl.jac` | Add `l2` param to `_mongo_quarantine`, implement `set_cache_ref`, update all four call sites |
| `jac-scale/jac_scale/impl/memory_hierarchy.main.impl.jac` | Wire `set_cache_ref(self.l2)` in `ScaleTieredMemory.postinit` |
| `repro/repro_cascade_redis.jac` | Self-contained repro / regression test |
| `repro/docker-compose.yml` | Mongo + Redis containers for local verification |

---

Closes #6184


udithishanka (approved):

kugesan1105:
### Follow-up PR @akindu-k 

- remove these any type 
image

image

akindu-k:
> ### Follow-up PR @akindu-k
> * remove these any type
> 
> image image

noted @kugesan1105. created an issue for it (#6208). Will address this in a separate PR.

PR #6212, opened 2026-05-27
fix(jaclang): lower type(x) is <builtin> to valid JS typeof checks (#6149)

### LINKED ISSUE #6149: [Bug] `type(x) is str` / `is not dict` compiles to `typeof x === str` (bare identifiers) in jac2js
### Description

Valid Jac runtime type checks using `type(x) is <typename>` / `type(x) is not <typename>` compile to JavaScript `typeof` comparisons, but the right-hand side is emitted as a **bare identifier** (`str`, `dict`, `list`) instead of the corresponding JavaScript string literal (`"string"`, `"object"`, etc.).

This only affects the **JavaScript** backend (`jac jac2js`, client `.cl.jac` bundles). Python (`jac run` / `jac2py`) is correct.

### Environment

- Jac **0.15.3**
- jac-client **0.3.17**
- Platform: Linux x86_64

### Steps to reproduce

**1. Minimal server snippet** (`main.jac`):

```jac
with entry {
    x = 42;
    if type(x) is str {
        print("is string");
    }
    if type(x) is not dict {
        print("not dict");
    }
}
```

- `jac run main.jac` → prints `not dict` (correct)
- Transpile to JS → emits `typeof x === str` / `typeof x !== dict` (bug)

**2. Client snippet** (`client.cl.jac`):

```jac
cl {
    def:pub TypeRepro() -> JsxElement {
        has msg: str = "";

        can with entry {
            x = 42;
            if type(x) is str {
                msg = "x is str";
            } elif type(x) is not dict {
                msg = "x is not dict";
            }
        }

        return <div>{msg}</div>;
    }
}
```

```bash
jac jac2js client.cl.jac
```

### Actual output (buggy JS)

```javascript
let x = 42;
if ((typeof x === str)) {
  console.log("is string");
}
if ((typeof x !== dict)) {
  console.log("not dict");
}
```

Client codegen is the same pattern:

```javascript
if ((typeof x === str)) {
  setMsg("x is str");
} else if ((typeof x !== dict)) {
  setMsg("x is not dict");
}
```

### Expected output

```javascript
if (typeof x === "string") { ... }
if (typeof x !== "object") { ... }   // or a stricter dict check if intended
```

For `list`, `typeof [] === "object"` in JS, so the lowering likely needs `Array.isArray(x)` rather than a bare `list` identifier.

### Runtime impact

In strict mode / ES modules (all `.cl.jac` browser bundles), `str` and `dict` are undeclared globals → **`ReferenceError`** when the branch executes:

```
ReferenceError: str is not defined
ReferenceError: dict is not defined
```

`jac check` passes; failure is **runtime-only** in JS targets.

### Suggested mapping (Jac → JS)

| Jac | Suggested JavaScript |
|-----|----------------------|
| `type(x) is str` | `typeof x === "string"` |
| `type(x) is not str` | `typeof x !== "string"` |
| `type(x) is dict` | `typeof x === "object" && x !== null && !Array.isArray(x)` (or project convention) |
| `type(x) is list` | `Array.isArray(x)` |
| `type(x) is int` / `float` | `typeof x === "number"` (with optional `Number.isInteger` for int) |
| `type(x) is bool` | `typeof x === "boolean"` |

### Related issues (same codegen area, different root cause)

- #5930 (closed): sibling `if`/`else` `let` scope → `ReferenceError`
- #5559 (open): same pattern across `elif` branches

### Search note

Searched open/closed issues in `Jaseci-Labs/jaseci` for `typeof str`, `type(x) is str`, `jac2js type is` — no existing report found for this specific lowering bug.

chess10kp:
## Summary

Fixes [#6149](https://github.com/jaseci-labs/jaseci/issues/6149): `type(x) is str` / `type(x) is not dict` (and similar) were lowered to `typeof x === str` with bare identifiers, causing `ReferenceError` in strict ES modules (all `.cl.jac` browser bundles). The jac2js backend now maps Jac builtin typenames to valid JavaScript checks (`typeof`, `Array.isArray`, etc.).

Python (`jac run` / jac2py) behavior is unchanged.

## JS semantics (approximations, not Python-identical)

These mappings follow JavaScript runtime checks, not byte-for-byte Python `type()` identity:

| Jac | JavaScript lowering | Note |
|-----|---------------------|------|
| `str` | `typeof x === "string"` | |
| `dict` | `typeof x === "object" && x !== null && !Array.isArray(x)` | Does not exclude `Date`, `Map`, etc. |
| `list` / `tuple` | `Array.isArray(x)` | No tuple arity |
| `int` / `float` | `Number.isInteger` split | `1.0` is float in JS |
| `i8`…`f64` | `typeof x === "number"` | No bounded-range checks |

User-defined types (`type(x) is MyClass`) are out of scope and still use the generic `is` lowering.

## Test plan

- [x] `jac test tests/compiler/passes/ecmascript/test_js_generation.jac` (67 tests)
- [x] `jac test tests/compiler/passes/ecmascript/test_esast_gen_pass.jac` (26 tests)
- [x] Manual repro: upstream emits `typeof x === str` + `ReferenceError`; fix emits quoted typeof / object checks

Fixes jaseci-labs/jaseci#6149


marsninja:
## Review: works as a bugfix, but the mechanism isn't the right long-term shape

The behavior is correct for the tested cases and the original bug (emitting `typeof x === str`, referencing an undeclared global) is genuinely fixed. But the implementation bypasses the ES AST in a way that I'd treat as tech debt rather than the durable design.

**Core issue — it smuggles raw JS through `es.Identifier`.** Every other branch in this pass builds structured nodes (`es.BinaryExpression`, `es.LogicalExpression`, `es.CallExpression`, `es.UnaryExpression`, `es.MemberExpression`, `es.Literal`). This fix instead unparses the subject to a JS source string, does a textual `template.replace("{0}", subject_js)`, and wraps the compound result in `es.Identifier(name=check_js)` — whose unparser is literally `return nd.name`. So an `Identifier` node secretly holds an expression like `typeof x === "object" && x !== null && !Array.isArray(x)`. Consequences:

1. **Opaque to downstream consumers** — any minifier / optimization / source-map / transform pass over the ES tree sees an identifier it can't reason about, and the unparser can't apply precedence/parenthesization because there's no structure.
2. **Premature mid-pass re-unparsing** — `_expr_to_js_string` spins up a fresh `JSCodeGenerator()` and runs full codegen on the subject during the gen pass, relying on `expr.gen.es_ast` already being populated. Duplicates the codegen entry point and is ordering-fragile.
3. **Source locations collapse** — the whole compound check is `sync_loc`'d onto one synthetic `Identifier`, so source maps point the entire expression at a fake node.
4. **Unsafe textual substitution** — `template.replace("{0}", subject_js)` is a blind string replace; if the subject's generated JS ever contains the literal `{0}` it corrupts.
5. **Double-evaluation baked in textually** — templates like `typeof {0} === "number" && Number.isInteger({0})` inline the subject twice, so `type(sideEffect()) is int` runs it twice.

**Proper version:** keep it in the AST. Reuse the subject's already-generated `es.Expression` (`subject_expr.gen.es_ast`) and build real nodes — `BinaryExpression("===", UnaryExpression("typeof", subj), Literal("string"))`, `LogicalExpression("&&", ...)`, `CallExpression(MemberExpression(Identifier("Array"), Identifier("isArray")), [subj])`, `BinaryExpression("instanceof", subj, Identifier("Set"))`, negation via `UnaryExpression("!", ...)`. A small builder dispatched on type name (returning a node, not a string) replaces both `TYPE_IS_JS_CHECKS` and `_expr_to_js_string`, restoring precedence handling, source maps, and downstream-pass compatibility, and matching the rest of the pass.

**Semantic gaps worth noting regardless of approach** (inherent JS/Python impedance, but the dict bakes in lossy choices): `int`/`float` via `Number.isInteger` is wrong for whole-valued floats (`type(3.0) is int` → true in JS); `list` and `tuple` both map to `Array.isArray` and are indistinguishable.

I'll follow up with the structured-node rewrite.

Issue #6046, opened 2026-05-20
Add KEDA support to jac-scale with a pluggable Autoscaler abstraction


udithishanka:
## Summary

Add [KEDA](https://keda.sh/) as a pluggable autoscaler engine for `jac-scale`, alongside the existing CPU-only Horizontal Pod Autoscaler path. Introduce a clean `Autoscaler` abstraction so HPA, KEDA, and future engines can be swapped via a single `jac.toml` field.

## Motivation

`jac-scale` currently hard-codes `autoscaling/v2` HPAs to CPU utilization in two places:

- Monolith path: [`jac-scale/jac_scale/targets/kubernetes/hpa/hpa.jac`](../tree/main/jac-scale/jac_scale/targets/kubernetes/hpa/hpa.jac) called from `kubernetes_target.deploy()` ([line ~1473](../tree/main/jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac#L1473)).
- Microservice path: `_build_hpa_manifest` in [`microservice/manifest_builder.jac`](../tree/main/jac-scale/jac_scale/targets/kubernetes/microservice/manifest_builder.jac#L553) applied at [`microservice/target.jac:412`](../tree/main/jac-scale/jac_scale/targets/kubernetes/microservice/target.jac#L412).

CPU is rarely the right signal for the workloads `jac-scale` users actually run:

- LLM workers — scale on Redis queue depth, not CPU.
- HTTP services — scale on request rate / latency from Prometheus.
- Batch / event processors — scale on Kafka lag, RabbitMQ queue, SQS depth.
- Cost-sensitive demos — scale to zero when idle.

KEDA solves all of the above with a single CRD (`ScaledObject`), runs identically on EKS / GKE / AKS / k3s / kind, and reuses the Prometheus + Redis already deployed inside every `jac-scale` namespace.

## Proposed design

Three-layer abstraction mirroring how `DatabaseProvider` / `DatabaseProviderFactory` already work in the codebase.

### Layer 1 — `Autoscaler` interface

New file `jac-scale/jac_scale/abstractions/autoscaler.jac`:

```jac
class Autoscaler {
    has config: BaseConfig, logger: (Logger | None) = None;
    def build_manifests(spec: AutoscalerSpec) -> dict[str, list[dict]];
    def apply(spec: AutoscalerSpec) -> None;
    def destroy(target_name: str, namespace: str) -> None;
    def preflight() -> list[str];  # returns warnings, never raises
}
[...]

Issue #6220, opened 2026-05-27
feat(jac-scale): add KEDA autoscaler engine and extra_triggers config (Phase 2 of #6046)


MusabMahmoodh:
Phase 2 of #6046. Depends on Phase 1 (#6219).

Adds `KEDAAutoscaler` implementation on top of the `Autoscaler` abstraction landed in Phase 1. Introduces `extra_triggers` config and `ScaledObject` / `TriggerAuthentication` CR management via `CustomObjectsApi` (same pattern as cert-manager in `targets/kubernetes/ingress.jac` around line 1075).

## Scope
- `jac-scale/jac_scale/abstractions/autoscalers/keda_autoscaler.jac` (built-in)
- Register `"keda"` in `AutoscalerFactory`
- New `jac.toml` keys on `KubernetesConfig`: `idle_replicas`, `autoscaler_polling_interval`, `autoscaler_cooldown`, `autoscaler_initial_cooldown`, `extra_triggers`
- Microservice sibling `[[plugins.scale.microservices.services.<name>.triggers]]` array
- `preflight()` warn-and-skip when KEDA CRD is missing (result cached per cluster)
- `MockKubernetesClient` or testcontainers-based harness for unit tests of CR creation

## Open questions to resolve before merge
- Precedence rule when `autoscaler_engine = "keda"` AND a service still has an `[...]hpa` sub-block (see review comment on #6046)
- Whether a non-CPU trigger with `engine = "hpa"` should auto-promote, hard-error, or be opt-in (see review comment on #6046)
- `TriggerAuth` deduplication across services sharing the same secret

## Acceptance
- [ ] `KEDAAutoscaler` emits `ScaledObject` (and optional `TriggerAuthentication`) via `CustomObjectsApi`
- [ ] Missing-CRD path logs install one-liner and continues with static `min_replicas`
- [ ] `--destroy` cleans up `ScaledObject` + auth resources
- [ ] Documented precedence rule for `hpa` vs `triggers` block
- [ ] End-to-end test against kind / k3d: deploy with KEDA + a Prometheus trigger, assert `ScaledObject` exists and pods scale on synthetic load
- [ ] Release note entry under `docs/docs/community/release_notes/jac-scale.md`

See #6046 for full design rationale.

Developer-Linus:
[...]

PR #6228, opened 2026-05-28
refactor(jac-scale): introduce Autoscaler abstraction over existing HPA paths (Phase 1 of Issue #6046)

Developer-Linus:
## Summary

Introduces the `Autoscaler` interface, `AutoscalerSpec` / `Trigger` models, `HPAAutoscaler` engine, and `AutoscalerFactory`, then routes both the monolith and microservice HPA deploy paths through the factory. Default engine is `"hpa"`. Existing users require no `jac.toml` changes and see zero behaviour difference.

## New Files

| File | Purpose |
|---|---|
| `abstractions/autoscaler.jac` | `Autoscaler` base class with `apply()` and `destroy()` declarations; `AutoscalerSpec` and `Trigger` data models |
| `abstractions/autoscalers/hpa_autoscaler.jac` | `HPAAutoscaler` — declarations |
| `abstractions/autoscalers/impl/hpa_autoscaler.impl.jac` | `HPAAutoscaler` — implementations; wraps the existing HPA logic. HPA manifests gain two new metadata fields (`namespace` and `app` label) and the update verb changes from `replace_*` (PUT) to `read_*` + `patch_*` (GET + PATCH). Functionally equivalent; operators diffing HPAs after upgrade will see these additions. |
| `factories/autoscaler_factory.jac` | `AutoscalerFactory` — registry-first factory with `register()` for external plugin extensibility and `create()` for engine resolution |

## Modified Files

| File | Change |
|---|---|
| `kubernetes_config.jac` | Adds `autoscaler_engine: str = "hpa"` with `to_dict` / `from_dict` support |
| `kubernetes_target.jac` | Replaces direct `create_hpa()` call with `AutoscalerFactory.create() + autoscaler.apply()` |
| `microservice/manifest_builder.jac` | Delegates `_build_hpa_manifest()` body to `HPAAutoscaler._build_manifests()` |
| `microservice/target.jac` | Replaces `_apply_or_replace_hpa()` loop with `AutoscalerFactory.create() + autoscaler.apply()` |
| `tests/test_factories.jac` | Adds `AutoscalerFactory` and `HPAAutoscaler` snapshot tests |
| `tests/test_deploy_k8s.jac` | Migrates HPA fixtures to call through `AutoscalerFactory` |

## Removed Files

| File | Reason |
|---|---|
| `hpa/hpa.jac` | Superseded by `HPAAutoscaler` |
| `hpa/impl/hpa.impl.jac` | Superseded by `HPAAutoscaler` |

## Test Plan

- [ ] `jac test jac_scale/tests/test_factories.jac` — factory and snapshot tests pass
- [ ] `jac test jac_scale/tests/test_deploy_k8s.jac` — all existing deploy tests pass unchanged
- [ ] `autoscaler_engine` defaults to `"hpa"` with no `jac.toml` changes required
- [ ] `AutoscalerFactory.register()` callable from an external plugin
- [ ] No `create_hpa` or `_apply_or_replace_hpa` references remain in active code paths


jac-scale/jac_scale/abstractions/autoscalers/impl/hpa_autoscaler.impl.jac
```diff
@@ -0,0 +1,82 @@
+"""Build HPA manifest dict; strips '-deployment' suffix for the HPA name, keeps it for scaleTargetRef."""
+impl HPAAutoscaler._build_manifests{
+    hpa_name = f"{spec.target_name.removesuffix('-deployment')}-hpa";
```
MusabMahmoodh: **Leaky abstraction.** The caller in microservice path constructs `target_name=f"{k8s_name}-deployment"` only so this line can strip the suffix back off (and same on line 14 for the `app` label). Two different naming conventions are being papered over inside the engine.

Cleaner shape: split `AutoscalerSpec.target_name` into `scale_target_name` (the Deployment to scale, used in `scaleTargetRef.name`) and an optional `autoscaler_name` (override for the HPA's own name, defaults to `f"{scale_target_name}-hpa"`). No string-stripping needed, and a future engine that names resources differently does not have to copy this hack.

jac-scale/jac_scale/abstractions/autoscalers/impl/hpa_autoscaler.impl.jac
```diff
@@ -0,0 +1,82 @@
...
+        "apiVersion": "autoscaling/v2",
+        "kind": "HorizontalPodAutoscaler",
+        "metadata": {
+            "name": hpa_name,
+            "namespace": spec.namespace,
+            "labels": {
+                "app": spec.target_name.removesuffix('-deployment'),
+                "managed": "jac-scale"
+            }
+        },
```
MusabMahmoodh, line 18: **Behaviour change vs the old `create_hpa`, not byte-identical as the PR claims.** Old code produced `metadata: {name, labels: {managed: 'jac-scale'}}`. This produces `metadata: {name, namespace, labels: {app, managed: 'jac-scale'}}`. Two new fields silently added.

Also the apply path on line 56-69 switched from `replace_namespaced_horizontal_pod_autoscaler` (PUT) to `read + patch_namespaced_horizontal_pod_autoscaler` (GET + PATCH). Different semantics: PUT resets fields not in body to defaults, PATCH preserves them. Unifying on patch is probably the right call (matches what the microservice path was doing), but it is a real semantic change.

Neither is wrong, but please call them out in the PR description and the release note instead of saying "byte-identical".

jac-scale/jac_scale/abstractions/autoscalers/impl/hpa_autoscaler.impl.jac
```diff
@@ -0,0 +1,82 @@
...
+        }
+    }
+}
+
+"""Delete the HPA resource; no-op if already absent."""
+impl HPAAutoscaler.destroy{
+    autoscaling_v2 = AutoscalingV2Api();
+    delete_if_exists(
+        autoscaling_v2.delete_namespaced_horizontal_pod_autoscaler,
+        f"{target_name}-hpa",
```
MusabMahmoodh: **Latent bug.** Destroy always computes `f"{target_name}-hpa"`, with no `removesuffix('-deployment')` mirror of `_build_manifests`. If destroy ever gets wired through the abstraction the same way apply is (and it should, see summary comment), the microservice caller will pass `target_name=f"{k8s_name}-deployment"` and this will try to delete `{k8s_name}-deployment-hpa`, which does not exist.

Will not bite today because destroy is not yet routed through `HPAAutoscaler`, but it is a footgun waiting to fire. Fix together with the suffix-strip cleanup suggested on line 3.

jac-scale/jac_scale/abstractions/autoscaler.jac
```diff
@@ -0,0 +1,37 @@
+"""Base autoscaler abstraction for jac-scale."""
+
+"""A single scaling trigger with engine-specific metadata."""
+obj Trigger {
+    has type: str,
+        metadata: dict[str, str] = {},
```
MusabMahmoodh: **`dict[str, str]` is too narrow for Phase 2.** KEDA scaler metadata values include ints, lists, and nested dicts (see `prometheus` and `kafka` triggers in the KEDA docs). Phase 2 will have to widen this to `dict[str, any]`, which breaks any external plugin that has already written against the Phase 1 interface.

Better to make it `dict[str, any]` now even though HPA only uses strings. The current restriction does not buy any safety we use, and the type change later is a breaking interface shift.

jac-scale/jac_scale/abstractions/autoscaler.jac
```diff
@@ -0,0 +1,37 @@
...
+    def apply(self: Autoscaler, spec: AutoscalerSpec) -> None;
+
+    """Remove the autoscaler resource from the cluster.
+
+    Args:
+        target_name: Name of the deployment the autoscaler targets
+        namespace: Kubernetes namespace
+    """
+    def destroy(self: Autoscaler, target_name: str, namespace: str) -> None;
+}
```
MusabMahmoodh, line 83: Two interface concerns.

**(a) Abstract methods with no body.** Calling `Autoscaler().apply(spec)` directly will either silently return `None` or error in an unhelpful way, depending on Jac semantics. Please either raise `NotImplementedError` in the base, or mark properly abstract, so a misconfigured plugin fails loud at the right place.

**(b) `preflight()` not added yet.** Phase 2 (#6220) needs it. Adding it to the base class later means any external plugin written against the Phase 1 interface will not satisfy the new contract. Better to add an empty default `preflight() -> list[str] { return []; }` now so the interface is stable from Phase 1 onwards.

Also, mutable defaults on `AutoscalerSpec.triggers = []` (line 17) and `Trigger.metadata = {}` (line 6) are worth confirming for Jac `obj` semantics. In Python dataclasses these would be shared across instances.

jac-scale/jac_scale/factories/autoscaler_factory.jac
```diff
@@ -0,0 +1,79 @@
...
+        ValueError: If engine is not supported
+    """
+    static def create(
+        engine: (str | None), config: dict[str, any], logger: (Logger | None) = None
+    ) -> Autoscaler {
+        # Check plugin registry first
+        if engine in AutoscalerFactory._registry {
+            factory = AutoscalerFactory._registry[engine];
+            return factory(config, logger);
+        }
```
MusabMahmoodh, line 67: **The `config: dict[str, any]` parameter is dead in Phase 1.** Both real call sites pass `{}`. The built-in `HPAAutoscaler()` constructor ignores it. The `AutoscalerFactoryFn` type alias requires plugins to accept it.

Three options, pick one:
1. Drop the parameter from Phase 1 and add it back in Phase 2 when KEDA actually needs it.
2. Pass `self.k8s_config.to_dict()` from both call sites so plugins get real config from day one.
3. Document explicitly that the parameter is reserved for future engines and that built-ins must ignore it.

Leaving a half-defined contract risks locking in a shape that does not fit KEDA later.

jac-scale/jac_scale/targets/kubernetes/microservice/target.jac
```diff
@@ -479,12 +480,43 @@ obj KubernetesMicroserviceTarget(KubernetesTarget) {
             }
         }
 
-        for (svc_name, hpa_manifest) in bundle.get("hpas", {}).items() {
-            self._apply_or_replace_hpa(autoscaling_v2, namespace, hpa_manifest);
+        for svc_name in bundle.get("hpas", {}).keys() {
+            k8s_name = k8s_safe_name(svc_name);
+            cfg = {};
+            try {
+                import from jac_scale.config_loader { get_scale_config }
```
MusabMahmoodh: `get_scale_config` is already imported at the top of this file (line 34). Drop the inner import.

jac-scale/jac_scale/targets/kubernetes/microservice/target.jac
```diff
@@ -479,12 +480,43 @@ obj KubernetesMicroserviceTarget(KubernetesTarget) {
...
+                cfg = dict(
+                    get_scale_config().get_microservices_config().get(
+                        "services", {}
+                    ).get(
+                        svc_name, {}
+                    ).get(
+                        "hpa", {}
+                    )
+                );
+            } except Exception { }
```
MusabMahmoodh: **Silent `except Exception { }` swallows real config errors.** Any error reading the per-service config (typo in `jac.toml`, missing key, malformed structure) now silently produces `cfg = {}`, and the autoscaler runs with the hardcoded defaults `min=1, max=3, cpu_target=70`. A user mistyping `cpu_targt` would get 70% scaling silently instead of an error.

Please catch the specific exceptions you actually expect (KeyError, AttributeError), log them via `self.logger.warn`, and let the rest propagate. The bare swallow is a regression in observability for the autoscaling path.

jac-scale/jac_scale/targets/kubernetes/microservice/target.jac
```diff
@@ -479,12 +480,43 @@ obj KubernetesMicroserviceTarget(KubernetesTarget) {
...
+                        Trigger(
+                            type="cpu",
+                            metadata={
+                                "averageUtilization": str(cfg.get("cpu_target", 70))
+                            }
+                        )
+                    ]
+                )
+            );
             if self.logger {
```
MusabMahmoodh, line 516: **Logger threaded through the factory but never reaches the built-in.** `AutoscalerFactory.create(..., logger=self.logger)` is called here, but the factory's built-in branch returns `HPAAutoscaler()` with no logger argument. So if any internal step of the engine ever needs to log (preflight failure, retry, fallback in Phase 2), it cannot.

Fix: pass the logger into the built-in too (`return HPAAutoscaler(logger=logger)`), and add a `has logger: Logger | None = None` field on the engine. Otherwise the logger parameter on `create()` is just decoration.

Also: the `f"{k8s_name}-deployment"` literal is repeated in this block and on lines 511 and 519. Worth a helper in `manifest_builder` so the microservice naming convention has a single owner.

jac-scale/jac_scale/targets/kubernetes/microservice/manifest_builder.jac
```diff
@@ -564,36 +566,19 @@ obj MicroserviceManifestBuilder {
...
+            min_replicas=int(cfg.get("min", 1)),
+            max_replicas=int(cfg.get("max", 3)),
+            triggers=[
+                Trigger(
+                    type="cpu",
+                    metadata={"averageUtilization": str(cfg.get("cpu_target", 70))}
+                )
+            ]
+        );
+        return HPAAutoscaler()._build_manifests(spec);
```
MusabMahmoodh: **Dead computation, hardcoded engine, private method called from outside.**

Three issues stacked here:

1. The manifest this returns is added to `bundle["hpas"][svc_name]` (via the caller), but `microservice/target.jac:483` only uses `bundle.get("hpas", {}).keys()` and rebuilds the spec from scratch. The manifest is never read.
2. Hardcoding `HPAAutoscaler()` defeats the abstraction. When `autoscaler_engine = "keda"` lands in Phase 2, this still emits HPA shape into the bundle for no reason.
3. `_build_manifests` is underscore-prefixed (private). Calling it from another module is a layering violation.

Recommend: drop the `hpas` entry from `bundle` entirely (target.jac now owns autoscaling end to end), or invert it so target.jac consumes a single `bundle["autoscalers"]` entry built by the factory once. Right now there are two sources of truth for the same scaling config (`_service_config` here vs `get_scale_config()` in target.jac) and they could drift.

MusabMahmoodh (commented):
**1. Destroy path is not migrated through the abstraction.** `kubernetes_target.jac:1765-1771` and `microservice/target.jac:686` still call `client.AutoscalingV2Api()` and `delete_namespaced_horizontal_pod_autoscaler` directly. This is exactly the "no half-migrated state" the Phase 1 issue said to avoid, and it will leak `ScaledObject` resources the moment Phase 2 (#6220) lands KEDA. Please route destroy through `AutoscalerFactory.create(engine).destroy(...)` in this same PR.

**2. The PR description says "byte-identical manifest output" but it is not.** New monolith HPAs gain an `app` label and a `namespace` in metadata, and the apply path switched from `replace_*` (PUT) to `read + patch_*` (GET + PATCH). Operationally these are fine, even improvements, but they are real behaviour changes. Please update the description and the release note so operators are not surprised when they diff their HPAs after upgrade.

Happy to pair on any of these.

jac-scale/jac_scale/tests/test_factories.jac
```diff
@@ -191,6 +194,92 @@ test "deployment target factory custom target overrides builtin" {
...
+
+    def override_factory(config: dict, logger: object = None) -> object {
+        return mock_autoscaler;
+    }
+
+    AutoscalerFactory.register("hpa", override_factory);
+    result = AutoscalerFactory.create("hpa", {});
+    assert result is mock_autoscaler;
+    del AutoscalerFactory._registry["hpa"];
+}
```
MusabMahmoodh, line 248: **Registry cleanup is not exception-safe.** `del AutoscalerFactory._registry["..."]` runs only on the happy path. If an assertion before it fails, the registry stays dirty and the next test runs against polluted state.

Wrap in `try / finally`, or expose `AutoscalerFactory.unregister(name)` so tests can use a fixture-style teardown. Same concern applies to the `DeploymentTargetFactory` and `ImageRegistryFactory` cleanup blocks above (lines 160, 179, 194, 297) but those are pre-existing.

jac-scale/jac_scale/targets/kubernetes/kubernetes_config.jac
```diff
@@ -29,6 +29,7 @@ obj KubernetesConfig(BaseConfig) {
         min_replicas: int = 1,
         max_replicas: int = 3,
         cpu_utilization_target: int = 50,
+        autoscaler_engine: str = "hpa",
```
MusabMahmoodh, line 33: **Three different CPU target defaults floating around now.**

- `KubernetesConfig.cpu_utilization_target = 50` (line 31, monolith default)
- `cfg.get("cpu_target", 70)` in microservice/target.jac:511 and manifest_builder.jac:577 (microservice default)
- `metadata.get("averageUtilization", "50")` fallback in hpa_autoscaler.impl.jac:4 (engine default)

Three defaults, two values (50 and 70). The engine fallback is dead in practice (callers always populate the trigger) but the drift between monolith=50 and microservice=70 is a real surprise for anyone reading the code. Pick one default, expose it as a module-level constant, reference it everywhere.

Developer-Linus:
> **1. Destroy path is not migrated through the abstraction.** `kubernetes_target.jac:1765-1771` and `microservice/target.jac:686` still call `client.AutoscalingV2Api()` and `delete_namespaced_horizontal_pod_autoscaler` directly. This is exactly the "no half-migrated state" the Phase 1 issue said to avoid, and it will leak `ScaledObject` resources the moment Phase 2 (#6220) lands KEDA. Please route destroy through `AutoscalerFactory.create(engine).destroy(...)` in this same PR.

# Clarification — microservice/target.jac:686 destroy path

`kubernetes_target.jac:1765–1771` is done — `_destroy_application()` has a single `app_name` so it maps cleanly to `AutoscalerFactory.create(engine).destroy(app_name, namespace)`.

`microservice/target.jac:686` is a different shape. The microservice `destroy()` tears down everything in the namespace at once using a label selector:

```jac
autoscaling_v2.delete_collection_namespaced_horizontal_pod_autoscaler(
    namespace=namespace, label_selector=ms_selector
);
```

The current `Autoscaler.destroy(target_name, namespace)` interface takes a single named resource. We cannot route the collection delete through it by iterating config service names — that approach leaves orphaned HPAs on the cluster for any service that was removed from config but not yet torn down. That is a regression against current behaviour.

The only way to route this through the abstraction without introducing orphans is to add a second method to the interface:

```
teardown(namespace: str, label_selector: str) -> None
```

`HPAAutoscaler` would implement it with `delete_collection_namespaced_horizontal_pod_autoscaler`. Phase 2 `KEDAAutoscaler` would implement it with the equivalent ScaledObject collection delete.

**Question:** Is adding `teardown(namespace, label_selector)` to the `Autoscaler` interface the intended fix, or is there a different approach you had in mind?

jac-scale/jac_scale/abstractions/autoscalers/impl/hpa_autoscaler.impl.jac
Developer-Linus: I've used the cleaner shape suggested.

jac-scale/jac_scale/abstractions/autoscalers/impl/hpa_autoscaler.impl.jac
Developer-Linus: Resolved. The PR description's "New Files" table explicitly documents both changes in the hpa_autoscaler.impl.jac row: HPA manifests gain namespace and app label
  in metadata, and the update verb changes from replace_* (PUT) to read_* + patch_* (GET + PATCH). The release note repeats the same callout and labels
  both changes non-breaking. The "byte-identical" claim was removed from the PR description and release note in commit c702a93df.

jac-scale/jac_scale/abstractions/autoscaler.jac
Developer-Linus: Resolved. Changed Trigger.metadata from dict[str, str] to dict[str, any] in autoscaler.jac. HPA is unaffected since strings satisfy any, and Phase 2 KEDA metadata (ints, lists, nested dicts) will work without a breaking interface change.

jac-scale/jac_scale/abstractions/autoscalers/impl/hpa_autoscaler.impl.jac
Developer-Linus: Resolved. Changed destroy's parameter from `target_name` to `autoscaler_name` across the base class, `HPAAutoscaler` declaration, and its implementation. The call site in `kubernetes_target.jac` now passes `f"{app_name}-hpa"` explicitly, mirroring how `apply()` uses `spec.autoscaler_name` directly — no suffix-stripping heuristics in the engine.

jac-scale/jac_scale/abstractions/autoscaler.jac
Developer-Linus: > Also, mutable defaults on `AutoscalerSpec.triggers = []` (line 17) and `Trigger.metadata = {}` (line 6) are worth confirming for Jac `obj` semantics. In Python dataclasses these would be shared across instances.

The Python mutable-default is a valid concern. I investigated the Jac compiler source and found conflicting evidence — the architecture docs claim `has fields with = []` are automatically converted to `field(default_factory=...)`, but the Jac internals (`jac0core/codeinfo.jac, native_marshal.jac`) use explicit `field(default_factory=list)` and `field(default_factory=dict)` throughout. I could not confirm automatic conversion from the compiler code itself.

Developer-Linus: I resolved comment a and b:
 `Comment (a)` — Abstract methods with no body
 Added raise NotImplementedError(...) bodies to apply() and destroy() on the Autoscaler base class directly in autoscaler.jac.
`Comment (b)` — preflight() missing from the interface
 Added `def preflight(self: Autoscaler) -> list[str] { return []; }` to the `Autoscaler` base class. Returns an empty list by
  default so all existing engines and external plugins work unchanged. Phase 2 KEDAAutoscaler will override it with the CRD presence check.

jac-scale/jac_scale/factories/autoscaler_factory.jac
Developer-Linus: Resolved. 
Replaced all three `AutoscalerFactory.create(engine, {}, logger)` call sites in `kubernetes_target.jac` and `microservice/target.jac` with `AutoscalerFactory.create(engine, self.k8s_config.to_dict(), logger)`. Plugins now receive real Kubernetes config from day one instead of an empty dict, establishing a meaningful contract before KEDA needs it in Phase 2.

jac-scale/jac_scale/targets/kubernetes/microservice/target.jac
Developer-Linus: Resolved. Removed redundant import.

jac-scale/jac_scale/targets/kubernetes/microservice/target.jac
Developer-Linus: Resolved.
Replaced the bare except `Exception { }` with `except (KeyError, AttributeError) as e` — the only realistic failure modes when navigating a missing or malformed key in `jac.toml`. The caught error is logged via `self.logger.warn` so the user knows their per-service HPA config was ignored and defaults are being used. All other exceptions now propagate instead of being swallowed.

jac-scale/jac_scale/targets/kubernetes/microservice/target.jac
Developer-Linus: Resolved. 
### Logger Propagation

Added `has logger: (Logger | None) = None` to `HPAAutoscaler` and updated `AutoscalerFactory.create()` to pass the logger through when constructing the autoscaler:

```jac
HPAAutoscaler(logger=logger)
```

Previously, `create()` accepted a logger but discarded it, leaving the autoscaler without access to logging. The logger is now available to the engine for:

* Preflight validation failures
* Retry and recovery logging
* Phase 2 KEDA fallback and diagnostic messages

### Deployment Name Duplication

The concern about duplicating `f"{k8s_name}-deployment"` has already been addressed by the earlier `scale_target_name` / `autoscaler_name` separation.

The deployment name literal now appears only once in the current implementation, so no additional refactoring is required.

jac-scale/jac_scale/targets/kubernetes/microservice/manifest_builder.jac
Developer-Linus: Resolved.
### Engine-Agnostic Autoscaler Bundle

Replaced `bundle["hpas"]` with `bundle["autoscalers"]` throughout the microservice deployment flow.

* **manifest_builder.jac**

  * Removed HPA-specific imports and manifest generation logic.
  * Replaced `_build_hpa_manifest()` with `_get_autoscaler_config()`, which returns a plain autoscaler configuration dictionary.
  * Eliminated engine-specific knowledge from the builder.

* **target.jac**

  * Applies autoscalers directly from `bundle["autoscalers"]`.
  * Removes the redundant `get_scale_config()` lookup and avoids maintaining multiple sources of truth.

* **plan.jac**

  * Updated planning logic to consume the new autoscaler configuration shape.
  * Removed autoscalers from manifest flattening since they are configuration data rather than generated manifests.

* **test_plan.jac**

  * Updated fixtures to use the new flat configuration format:

    ```json
    {
      "min": 2,
      "max": 10,
      "cpu_target": 70
    }
    ```

This refactor makes the bundle engine-agnostic, allowing future autoscaler engines such as KEDA to plug in without requiring changes to the builder, planner, or rendering pipeline.

jac-scale/jac_scale/tests/test_factories.jac
Developer-Linus: Resolved.
### Exception-Safe Registry Cleanup

Wrapped both registry-mutating tests in `try/finally` blocks to ensure `del AutoscalerFactory._registry[...]` always executes, regardless of whether a test assertion succeeds or fails.

Previously, registry cleanup occurred only on the happy path. A failing assertion could leave custom factory registrations behind, causing state leakage and potentially affecting subsequent tests.

jac-scale/jac_scale/targets/kubernetes/kubernetes_config.jac
Developer-Linus: Resolved.
### Unified CPU Utilization Default

Defined `DEFAULT_CPU_UTILIZATION_TARGET: int = 50` in `autoscaler.jac` as the single authoritative source for the default CPU utilization target.

Updated all autoscaler-related code paths to reference this constant, replacing previously inconsistent inline fallbacks:

* `manifest_builder.jac` — `70` → `DEFAULT_CPU_UTILIZATION_TARGET`
* `target.jac` — `70` → `DEFAULT_CPU_UTILIZATION_TARGET`
* `hpa_autoscaler.impl.jac` — `"50"` → `DEFAULT_CPU_UTILIZATION_TARGET`

`KubernetesConfig.cpu_utilization_target` retains its direct `= 50` field default, with a comment referencing the shared constant. This is necessary because typed field defaults cannot reference module-level globals directly.

jac-scale/jac_scale/abstractions/autoscaler.jac
MusabMahmoodh: Tested empirically against the actual Jac compiler rather than going by the docs vs internals contradiction.

Minimal repro:

```jac
obj Box { has items: list[int] = []; }

with entry {
    a = Box(); b = Box();
    a.items.append(1);
    print(a.items, b.items, a.items is b.items);
}
```

Output: `[1] [] False`. Two `Box()` instances get two different lists.

Generated Python (via `jac tool ir py`):

```python
class Box(Obj):
    items: list[int] = field(factory=lambda: [])
```

So the Jac front-end **does** auto-wrap mutable literal defaults in `field(factory=lambda: ...)` for `obj` declarations. The architecture docs are right; the contradicting `field(default_factory=list)` calls in `jac0core` are hand-rolled low-level code that constructs the same AST nodes the compiler emits anyway. The two layers are not in conflict, they are just different abstraction levels.

Short answer: `has triggers: list[Trigger] = []` and `has metadata: dict[str, any] = {}` are safe as written. No action needed.

Same precedent in `abstractions/models/deploy_context.jac:6` (`has env_list: list[dict[str, any]] = []`) and `abstractions/emailer.jac:18` (`has config: dict[str, Any] = {}`), both running fine in production today.

MusabMahmoodh:
Your reading of the shape difference is correct. Iterating config service names is not safe; it leaks orphans for any service that was removed from `jac.toml` between two deploys. Adding a collection-delete to the interface is the right call. Two refinements to the proposal.

### 1. Naming and shape

`teardown(namespace, label_selector)` reads ambiguously next to `destroy(autoscaler_name, namespace)`. Either of these is clearer:

**Option A — two named methods (recommended):**

```jac
def destroy(autoscaler_name: str, namespace: str) -> None;
def destroy_collection(namespace: str, label_selector: str) -> None;
```

Matches the k8s API verbs (`delete_*` vs `delete_collection_*`), each method has one responsibility, no kwargs validation needed in implementations.

**Option B — single method, two modes:**

```jac
def destroy(
    namespace: str, *,
    autoscaler_name: (str | None) = None,
    label_selector: (str | None) = None
) -> None;
```

Raises if neither or both are provided. Smaller interface but pushes validation into every implementation.

I would go with Option A. Option B is fine if you prefer the smaller surface.

### 2. Cross-engine label contract is the load-bearing part

This is the more important of the two. For the collection delete to work, every engine's `apply()` MUST add the same label set (today `managed=jac-scale`) on the resources it creates. `HPAAutoscaler` does this. The Phase 2 `KEDAAutoscaler` MUST do the same on the `ScaledObject` and on any `TriggerAuthentication` it creates, otherwise the bulk delete will silently miss them and orphan KEDA resources on the cluster.

Please make this contract explicit in this PR, either:

- A docstring on the `Autoscaler` base class: "Implementations MUST add `managed=jac-scale` to every resource they create, so that `destroy_collection(namespace, 'managed=jac-scale')` is sufficient to tear them all down.", OR
- A helper on the base class:
  ```jac
  static def _managed_labels() -> dict[str, str] { return {"managed": "jac-scale"}; }
  ```
  that engines mix into their resource metadata. Then the label set has a single owner and a future Phase 2 contributor cannot forget.

Without this written down somewhere, the next contributor adding a third engine has no way to know the collection delete depends on a specific label being present, and Phase 2 may regress this silently.

---

### On the other comments

Walked through the rest. The resolutions land cleanly — the bundle rename to `autoscalers` carrying engine-agnostic cfg (not pre-built HPA manifests) is especially nice, the silent `except Exception` disappeared on its own as a side effect of that refactor, and `to_dict()` plumbing through the factory is good.

Two very small things I noticed while reviewing:

1. `hpa_autoscaler.impl.jac` line 10 still has a hardcoded `else 50` for the empty-triggers fallback. Now that `DEFAULT_CPU_UTILIZATION_TARGET` is imported in the same file (line 5), that branch should reference the constant instead, otherwise the unification is one place short. Will add a small inline comment for this.

2. `hpa_autoscaler.impl.jac:17`: the `app` label is now derived by `hpa_name.removesuffix('-hpa')`. This is the same shape as the old `-deployment` strip, just with a different suffix. Functionally fine for now, but cleaner would be `"app": spec.scale_target_name` directly, since the scaled target's name is the natural value for the `app` label regardless of how the HPA itself is named. Non-blocking.

Neither blocks merge. Just flagging in case you want to tidy them while the destroy work is open.

jac-scale/jac_scale/abstractions/autoscalers/impl/hpa_autoscaler.impl.jac
```diff
@@ -0,0 +1,83 @@
+"""Build HPA manifest dict from spec fields; no string manipulation on names."""
+impl HPAAutoscaler._build_manifests{
+    hpa_name = spec.autoscaler_name or f"{spec.scale_target_name}-hpa";
+    cpu_target = int(
+        spec.triggers[0].metadata.get(
+            "averageUtilization", str(DEFAULT_CPU_UTILIZATION_TARGET)
+        )
+    )
+        if spec.triggers
+        else 50;
```
MusabMahmoodh: **Small slip from the CPU-default unification.** This branch still has a hardcoded `50`. Should be `else DEFAULT_CPU_UTILIZATION_TARGET` to match the constant imported on line 5 of this file. Otherwise the unification has one site short and the two will silently drift if someone changes the constant later.

Note that in practice this branch is dead today (callers always pass at least one trigger), but it is a smell next to the clean unification you just did.

Developer-Linus: Addressed — else 50 replaced with else DEFAULT_CPU_UTILIZATION_TARGET to keep the constant as the single source of truth for
  both fallback paths.

Developer-Linus:
> Your reading of the shape difference is correct. Iterating config service names is not safe; it leaks orphans for any service that was removed from `jac.toml` between two deploys. Adding a collection-delete to the interface is the right call. Two refinements to the proposal.
> 
> ### 1. Naming and shape
> `teardown(namespace, label_selector)` reads ambiguously next to `destroy(autoscaler_name, namespace)`. Either of these is clearer:
> 
> **Option A — two named methods (recommended):**
> 
> ```jac
> def destroy(autoscaler_name: str, namespace: str) -> None;
> def destroy_collection(namespace: str, label_selector: str) -> None;
> ```
> 
> Matches the k8s API verbs (`delete_*` vs `delete_collection_*`), each method has one responsibility, no kwargs validation needed in implementations.
> 
> **Option B — single method, two modes:**
> 
> ```jac
> def destroy(
>     namespace: str, *,
>     autoscaler_name: (str | None) = None,
>     label_selector: (str | None) = None
> ) -> None;
> ```
> 
> Raises if neither or both are provided. Smaller interface but pushes validation into every implementation.
> 
> I would go with Option A. Option B is fine if you prefer the smaller surface.
> 
> ### 2. Cross-engine label contract is the load-bearing part
> This is the more important of the two. For the collection delete to work, every engine's `apply()` MUST add the same label set (today `managed=jac-scale`) on the resources it creates. `HPAAutoscaler` does this. The Phase 2 `KEDAAutoscaler` MUST do the same on the `ScaledObject` and on any `TriggerAuthentication` it creates, otherwise the bulk delete will silently miss them and orphan KEDA resources on the cluster.
> 
> Please make this contract explicit in this PR, either:
> 
> * A docstring on the `Autoscaler` base class: "Implementations MUST add `managed=jac-scale` to every resource they create, so that `destroy_collection(namespace, 'managed=jac-scale')` is sufficient to tear them all down.", OR
> * A helper on the base class:
>   ```jac
>   static def _managed_labels() -> dict[str, str] { return {"managed": "jac-scale"}; }
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   that engines mix into their resource metadata. Then the label set has a single owner and a future Phase 2 contributor cannot forget.
> 
> Without this written down somewhere, the next contributor adding a third engine has no way to know the collection delete depends on a specific label being present, and Phase 2 may regress this silently.
> 
> ### On the other comments
> Walked through the rest. The resolutions land cleanly — the bundle rename to `autoscalers` carrying engine-agnostic cfg (not pre-built HPA manifests) is especially nice, the silent `except Exception` disappeared on its own as a side effect of that refactor, and `to_dict()` plumbing through the factory is good.
> 
> Two very small things I noticed while reviewing:
> 
> 1. `hpa_autoscaler.impl.jac` line 10 still has a hardcoded `else 50` for the empty-triggers fallback. Now that `DEFAULT_CPU_UTILIZATION_TARGET` is imported in the same file (line 5), that branch should reference the constant instead, otherwise the unification is one place short. Will add a small inline comment for this.
> 2. `hpa_autoscaler.impl.jac:17`: the `app` label is now derived by `hpa_name.removesuffix('-hpa')`. This is the same shape as the old `-deployment` strip, just with a different suffix. Functionally fine for now, but cleaner would be `"app": spec.scale_target_name` directly, since the scaled target's name is the natural value for the `app` label regardless of how the HPA itself is named. Non-blocking.
> 
> Neither blocks merge. Just flagging in case you want to tidy them while the destroy work is open.

### Implementation Notes

Went with **Option A** (`destroy()` / `destroy_collection()`) to align with Kubernetes API terminology and resource lifecycle operations.

Added `_managed_labels()` as a static helper on the base `Autoscaler` class so the `managed=jac-scale` label contract has a single owner. Every autoscaler engine now applies these labels through the shared helper during `apply()`, making the contract explicit and ensuring future engines (Phase 2 and beyond) cannot accidentally diverge.

Also addressed the two follow-up items:

1. Replaced the hardcoded `50` in the empty-triggers fallback path with `DEFAULT_CPU_UTILIZATION_TARGET`, making the constant the single source of truth for CPU utilization defaults.
2. Switched the application label to use `spec.scale_target_name` directly rather than deriving it by stripping the `-hpa` suffix from the generated HPA name.

Issue #6234, opened 2026-05-28
[Bug]: `jac check` falsely reports `Enum.__members__` as missing attribute (E1030)


kugesan1105:
### Operating System

Ubuntu (WSL2)

### Versions

Jac main (`bd44513b88`) | Python 3.12 | jaclang from source

### Steps to Reproduce

1. Create a `.jac` file:

   ```jac
   """Minimal repro: jac check falsely flags Enum.__members__ as missing."""

   import from enum { StrEnum }


   class Color(StrEnum) {
       with entry {
           RED = "red";
           GREEN = "green";
           BLUE = "blue";
       }
   }

   with entry {
       name = "RED";
       # `__members__` is a standard Enum attribute, present on every Enum
       # subclass at runtime. Used here for the textbook "is this string a
       # valid member name?" check.
       is_member = name in Color.__members__;
       print(is_member);
       if is_member {
           print(Color.__members__[name]);
       }
   }
   ```

2. Run `jac check enum_members_repro.jac`:

   ```
   ✖ Error: error[E1030]: Type "Color" has no attribute "__members__"
        --> .../enum_members_repro.jac:18:30
      16 |     # subclass at runtime. Used here for the textbook "is this string a
      17 |     # valid member name?" check.
      18 |     is_member = name in Color.__members__;
         |                              ^^^^^^^^^^^
   ```

3. Run `jac run enum_members_repro.jac`:

   ```
   True
   red
   ```

### Expected vs Actual Behavior

**Expected:** `jac check` accepts `Color.__members__` and reports no error for this file. `__members__` is part of the standard Python `enum` API — every `Enum` / `IntEnum` / `StrEnum` subclass has it at runtime, injected by `EnumMeta` at class creation time. `Color.__members__` is the textbook idiomatic way to check whether a string is a valid enum member name without raising.

**Actual:** `jac check` emits `E1030: Type "Color" has no attribute "__members__"` on every access — while `jac run` executes the same code correctly (prints `True` then `red`). So the type checker disagrees with the runtime on standard Python semantics.

### Why this matters — real-world impact

[...]

PR #6259, opened 2026-05-29
fix: resolve false E1030 errors on metaclass attributes like Enum.__members__

ThiruvarankanM:
### Background: What is a Metaclass?

Before getting into the bug, let's understand what a metaclass is — because
this entire fix is about teaching the type checker about metaclasses.

In Python, every class is itself an object. And every object has a type:

```
"hello"   → instance of  →  str
[1, 2]    → instance of  →  list
Color.RED → instance of  →  Color
Color     → instance of  →  ???   ← this is the metaclass
```

For normal classes, that "???" is `type` (Python's default builder).
For enum classes, it's `EnumMeta` — a special builder that adds enum powers.

```python
class Enum(metaclass=EnumMeta):  # use EnumMeta instead of type
    ...
```

This means: when Python builds the `Enum` class object, use `EnumMeta`
as the constructor. The result: `Enum` (and every enum class) gets extra
attributes from `EnumMeta` — `__members__`, `__iter__`, `__len__`,
`__getitem__`.

### Base Class vs Metaclass — The Key Difference

Base classes provide attributes for **instances**:

```
class Enum:
    name = ...        # available on Color.RED.name
    value = ...       # available on Color.RED.value
```

Metaclasses provide attributes for **the class itself**:

```
class EnumMeta(type):
    __members__ = ...  # available on Color.__members__
    __iter__   = ...   # available on iter(Color)
    __len__    = ...   # available on len(Color)
```

Simple rule:
- `Color.RED.something` → looks in Color's MRO (inheritance chain)
- `Color.something`     → looks in Color's MRO, then EnumMeta

---

### The Problem

Accessing `__members__` on any enum class raised a false type error on
valid, correctly written code.

```jac
enum Color { RED = 1, GREEN = 2, BLUE = 3 }

with entry {
    exists = "RED" in Color.__members__;       # E1030 ← false error
    red    = Color.__members__["RED"];         # E1030 ← false error
}
```

Errors seen:
```
error[E1030]: Attribute '__members__' not found in type 'Color'
```

All of this code is valid Jac. The type checker was wrong.

---

### Why It Was Failing

The Jac type checker's `_lookup_class_member` function only walked the
class MRO (inheritance chain):

```
Color.__members__
  → check Color   → not found
  → check Enum    → not found
  → check object  → not found
  → ERROR: E1030
```

But `__members__` doesn't live anywhere in the MRO — it lives on
`EnumMeta` (the metaclass). The type checker never knew to check the
metaclass, so it raised a false error.

### The Deeper Reason — Where the Info Was Lost

In Python's `enum.pyi` stub file:

```python
class Enum(metaclass=EnumMeta):
    ...
```

When Python's AST parses this, it splits the class arguments:
```
nd.bases    = []                          ← positional bases
nd.keywords = [keyword("metaclass",       ← keyword arguments
                Name("EnumMeta"))]
```

Our stub loader (`pyast_load_pass`) only read `nd.bases` and completely
ignored `nd.keywords`. So `metaclass=EnumMeta` was silently dropped —
the type system had zero knowledge that `EnumMeta` was `Enum`'s metaclass.

There was no way to fix the lookup without first capturing this information.

---

### The Fix

The metaclass info had to be carried through every layer:
stub file → AST → type system → attribute lookup.

1. **`unitree.jac`** — Added a `metaclass_expr` field to the `Archetype`
   AST node so there's a place to store the metaclass.

2. **`jir_registry.jac`** — Registered the new field in Jac's binary
   cache (JIR) so it survives serialization.

3. **`pyast_load_pass.impl.jac`** — Extracted `metaclass=` from
   `nd.keywords` and added it to `kid` so the symbol resolution pass
   visits and binds the name. Without `kid`, the metaclass name stays
   unresolved.

4. **`types.jac`** — Added a `metaclass` field to `ClassDetailsShared`
   so every class type can hold its metaclass.

5. **`construct_types.impl.jac`** — For stub classes like `Enum`,
   resolved the metaclass expression to a real `ClassType`. For
   subclasses like `Color(Enum)`, inherited the metaclass from base
   classes — so `Color` automatically gets `EnumMeta` without writing
   it explicitly.

6. **`type_utils.jac`** — Added a `from_metaclass` flag to `ClassMember`
   so lookups can signal whether an attribute came from the class MRO or
   the metaclass.

7. **`evaluator_util_methods.impl.jac`** — Replaced the old enum-specific
   hardcode with a general metaclass fallback in `_lookup_class_member`:
   walk the class MRO first, then fall back to the metaclass MRO. This
   is the core of the fix.

8. **`type_evaluator.impl.jac`** — `__members__` is a `@property` on
   `EnumMeta`, so its raw type is the property object itself. Switched
   the property-unwrapping guard from `is_enum_class()` to
   `from_metaclass` so we return the getter's actual return type
   (`MappingProxyType`) instead of the property object.

9. **`type_evaluator.jac`** — Removed the now-unused
   `enum_meta_class` prefetch since the general metaclass system
   handles it.

---

### End-to-End Flow

```
enum.pyi: class Enum(metaclass=EnumMeta)
         ↓
extract metaclass= → store on AST → resolve to EnumMeta ClassType
         ↓
Color(Enum) inherits → Color.shared.metaclass = EnumMeta
         ↓
Color.__members__ → not in class MRO → check metaclass MRO → found
         ↓
unwrap @property → MappingProxyType[str, Color] ✓
```

---

### Result

```
Before:  Color.__members__          → E1030 (false error)
After:   Color.__members__          → MappingProxyType[str, Color] ✓
         "RED" in Color.__members__ → bool ✓
         Color.__members__["RED"]   → Color ✓
```

- 0 false errors on all `Enum.__members__` usage patterns
- All 268 existing checker tests pass
- New regression test added with 5 cases (direct access, `in`, subscript,
  multiple enums)

---

### Impact

- Fixes false E1030 errors on `Enum.__members__` across the codebase
  (including 6 sites in `esast_gen_pass.impl.jac`)
- The fix is **general** — works for any class with any metaclass
  (`EnumMeta`, `ABCMeta`, custom metaclasses), not just enums
- Resolves the existing `# TODO: Handle metaclass member lookups`
- No behavior change for code that doesn't use metaclass attributes —
  the fallback only runs when the class MRO lookup misses

This PR fixes issue #6234


jac/jaclang/jac0core/unitree.jac
```diff
@@ -565,7 +565,8 @@ obj Archetype(ContextAwareNode, ArchSpec, AstAccessNode, ArchBlockStmt, AstImplN
     has name: Name,
         arch_type: Token,
         base_classes: (Sequence[Expr] | None),
-        type_params: (Sequence[TypeParam] | None) = None;
+        type_params: (Sequence[TypeParam] | None) = None,
+        metaclass_expr: (Expr | None) = None;
```
kugesan1105: is this really necessary to have it as a `field` ???
Can’t we derive it from the base class, maybe as a `property `of `archetype `or just a `utility function?`

jac/jaclang/compiler/passes/main/impl/pyast_load_pass.impl.jac
```diff
@@ -2715,6 +2715,20 @@ impl PyastBuildPass.proc_class_def(nd: py_ast.ClassDef) -> (uni.Archetype | uni.
...
+    class_keywords: list[uni.KWPair] = [];
+    for kw in nd.keywords {
+        if kw.arg is None {
+            continue;  # skip **expansion form (no `arg` name)
+        }
+        converted_kw = self.proc_keyword(kw);
+        if isinstance(converted_kw, uni.KWPair) {
+            class_keywords.append(converted_kw);
+        }
+    }
```
kugesan1105, line 2731: should be fixed in language level

currently the follwing code is a syntax error

```jac
class Enum(metaclass=EnumMeta){
}
```

Image

PR #6260, opened 2026-05-29
fix(type-checker): resolve false-positive E1091/E1092 for @contextmanager methods

SandeepaHWP:
## Summary

- `@contextmanager`-decorated methods were producing false-positive **E1091** (no `__enter__`) and **E1092** (no `__exit__`) on every `with` block that used them
- Root cause: `get_type_of_ability` read the declared `-> None` return type literally and never applied the `@contextmanager` runtime transformation
- Fix: detect `@contextmanager` in `get_type_of_ability` and override the return type to `AbstractContextManager` (lazy-loaded from the vendored contextlib stub) — the `with`-statement checker then finds `__enter__`/`__exit__` naturally with no special-casing at the call site
- Also fixes the stored-variable edge case: `ctx = lock.acquire(); with ctx { }` — `ctx` is now correctly typed as `AbstractContextManager` rather than `NoneType`
- Eliminates **40 false-positive errors** from `langserve/rwlock.jac` usage in `langserve/engine.jac`

## Files changed

| File | Change |
|---|---|
| `compiler/type_system/properties.jac` + `impl` | Add `is_contextmanager_decorator_type` detector (mirrors `is_cached_property_decorator_type`) |
| `type_evaluator.jac` | Add `contextlib_module` field, `_CONTEXTLIB_STUB_FILE_PATH`, `_get_contextmanager_return_type` declaration |
| `type_evaluator.impl/type_evaluator.impl.jac` | Initialize `_CONTEXTLIB_STUB_FILE_PATH` in postinit |
| `type_evaluator.impl/construct_types.impl.jac` | Apply `@contextmanager` return type override in `get_type_of_ability` |
| `type_evaluator.impl/evaluator_util_methods.impl.jac` | Implement `_get_contextmanager_return_type` |
| `tests/compiler/passes/main/test_checker_bugs.jac` | 3 regression tests (`#CTXMGR-1`) |
| `tests/compiler/passes/main/fixtures/checker/` | 3 fixture files for the tests |
| `tests/compiler/fixtures/` | Minimal reproduction files |


kugesan1105:
Closing in favor of #6292 (merged). This PR was a targeted, static name-checking workaround (matching the `contextmanager`/`asynccontextmanager` decorator by name). #6292 is the proper fix: it infers TypeVars through function-typed arguments and applies decorators generically (Pyright-style), so context managers are recognized with no per-decorator name special-casing.

PR #6294, opened 2026-05-31
fix(mongo-backend): persist empty-dict archetype fields via $literal wrap (#6167)

### LINKED ISSUE #6167: [bug] _strip_empty_dicts not applied to partial_data, allowing empty-dict archetype fields to re-trigger 40180
## Summary

`_strip_empty_dicts` is applied to `non_arch_top` but **not** to `partial_data`. Both paths converge into the same `$set` in the MongoDB pipeline (`memory_hierarchy.mongo.impl.jac:773`), so an archetype field declared as `has settings: dict = {}` can write `data.archetype.settings = {}` into the document and re-trigger the bug fixed in #40180.

## Steps to reproduce

1. Define an archetype with a default empty-dict field, e.g. `has settings: dict = {}`.
2. Persist the node/edge so it goes through the `partial_data` path.
3. Observe that `data.archetype.settings: {}` is written to MongoDB via `$set`.
4. This empty dict re-triggers the same condition that #40180 was meant to fix.

## Root cause

`_strip_empty_dicts` is only called on `non_arch_top` before the `$set` is assembled. `partial_data` bypasses the strip, so empty dicts survive into the final `$set` document.

**Relevant location:** `memory_hierarchy.mongo.impl.jac:773`

## Suggested fix

Strip `set_doc` at the choke point — after both `non_arch_top` and `partial_data` are merged into `set_doc` and before it is handed to `$set` — so no upstream source can sneak an empty dict back in.

## Impact

Any archetype with a `has field: dict = {}` default will silently re-introduce the 40180 regression on every save through the `partial_data` path.

---
> Follow-up from PR review by @akindu-k — marked as non-blocking but worth tracking.

akindu-k:
## Problem

`_put_node_atomic` updates a `NodeAnchor` using an **aggregation-pipeline** `$set` (`find_one_and_update(..., [{'$set': set_doc}], ...)` — note the `[ ... ]` list, required for the atomic `$setUnion`/`$setDifference` edge-merge expression).

A pipeline `$set` **evaluates every field value as an aggregation expression**. That has two failure modes for serialized archetype data:

1. **Empty embedded documents.** On **MongoDB older than 6.1.0**, a bare `{}` is rejected with error **40180** (`Location40180`, "an empty object is not a valid value"). The restriction was lifted in **6.1.0** by [`SERVER-54046`](https://jira.mongodb.org/browse/SERVER-54046), with no 6.0.x backport. This matters because jac-scale provisions **`mongo:6.0`** on Kubernetes (`providers/database/kubernetes_mongo.jac`).
2. **`$`-prefixed keys.** A dict whose keys start with `$` is misread as an operator — e.g. the `{'$ref': ..., '$type': ...}` the `Serializer` emits for a repeated anchor reference.

`set_doc` is assembled from two sources:

| Source | Target path | Handled before #6076? |
|--------|-------------|--------------------------|
| `non_arch_top` | `data.<k>` (e.g. `access.roots.anchors`) | ✅ yes — stripped in #6076 |
| `partial_data` | `data.archetype.<k>` (dirty archetype fields) | ❌ no |

#6076 handled only the `non_arch_top` half. The `partial_data` half went into the `$set` raw, so any archetype field that serialized to an empty dict:

```jac
node MyNode { has settings: dict = {}; }
```

reached the pipeline `$set` as `data.archetype.settings = {}` and raised 40180. `_put_node_atomic` swallows the exception in its `try/except`, so **every such partial save was silently lost** on mongo < 6.1.0 — the node stopped persisting updates after its first full write.

## Why stripping is the wrong fix

The obvious fix is to strip the empty dict out before writing. That avoids 40180 but loses information: a deliberate "clear this field back to `{}`" becomes a no-op, leaving the previously stored value in place. A version-gated strip (strip only below 6.1.0) has the same flaw on exactly the version we ship.

So stripping trades a silent write loss for a silent clear loss. We want the empty dict to actually persist. Stripping also does nothing for the `$`-prefixed-key failure mode.

## Fix

Adopt **one uniform rule** for the pipeline `$set`: `data.edges` is the only expression; every serialized data value is wrapped in `{'$literal': v}` and stored verbatim.

```jac
# Build a flat $set document. This is an aggregation-pipeline $set, so
# every value is evaluated as an expression — which rejects bare empty
# dicts (error 40180) and misreads `$`-prefixed keys (e.g. a Serializer
# `$ref`/`$type`). One rule keeps it safe: `data.edges` is the only
# expression; every serialized data value is wrapped in `$literal` and
# stored verbatim. (`type`/`updated_at` are hardcoded literals.)
    set_doc[f"data.{k}"] = {'$literal': v};
}
for (k, v) in partial_data.items() {
    set_doc[f"data.archetype.{k}"] = {'$literal': v};
}
set_doc['data.edges'] = {'$setUnion': [...]};   // the only expression
```

`$literal` is a standard aggregation operator (since MongoDB 2.6) that returns its argument verbatim without evaluating it. Wrapping the **whole value** of both source groups — archetype fields *and* non-archetype top-level fields (`access`) — neutralizes **both** failure modes at once (empty dicts *and* `$`-prefixed keys) with a single mental model. It is safe because every wrapped value comes from serialization and is pure data that must never be evaluated, and each field is assigned as a whole value (no dot-path sub-updates within a field). `type`/`updated_at` are hardcoded literal scalars and need no wrap.

This replaces the earlier split (wrap archetype, strip non-archetype) and lets the `_strip_empty_dicts` helper be **deleted** — one fewer concept, less code, and the `access` field is now also robust against `$`-keys. Two incidental cleanups in the same function: the now-unread `arch` local (dead since the collect-pass refactor) is removed, and the stale `data.archetype.<field> — only dirty scalar fields` comment is corrected (dict fields are written there too).

**Trade-off:** stripping omitted default empty-dict leaves (e.g. `access.roots.anchors = {}`) and the loader rebuilt them; `$literal` stores them. That is a few bytes per document with **no runtime/deserialization difference** (`_deserialize_permission` rebuilds defaults regardless). One pre-existing #6076 test asserted on this stored shape and was updated to assert the real invariant ("the write succeeds and access is persisted") instead of "no empty dicts stored."

Result: an empty, cleared-to-`{}`, or `$`-keyed dict field — archetype or access — now persists identically on all MongoDB versions, with no server-version detection and no data migration.

## What is deliberately left unchanged

**`put_partial`** uses a **classic operator** `$set` (`update_one(..., {'$set': update})`, a plain dict, not a pipeline), which accepts empty embedded documents on every MongoDB version. The 40180 restriction does not apply there, so it stores `{}` as-is and must not be touched. A comment documents this.

**`data.edges`** stays an aggregation expression (`$setUnion`/`$setDifference`) — it is the whole reason the update is a pipeline, and it must be evaluated, not stored literally.

## Note on the merge with `main`

`main` refactored dirty detection into `jaclang.runtimelib.changeset` and made the runtime's **collect pass the single owner of dirty detection**, passing the dirty `fields` set into `_put_node_atomic`. This branch adopts that architecture — the backend trusts `fields` rather than re-deriving dirtiness — and narrows its own contribution to the empty-dict/`$`-key *encoding* fix. The atomic-path tests were updated to call `_put_node_atomic` with the new `fields` argument.

## Tests

Two regression tests in `test_mongo_layer123.jac`, one per failure mode. Each is included only because it **fails without the fix**:

- **Clear-to-`{}` on `mongo:6.0` via `$literal` (the #6167 regression):** seed `settings={"keep": 1}`, clear to `{}`, run the atomic path on the version that raises 40180. A non-`None` document proves the pipeline `$set` did not raise, and `settings == {}` proves the clear persisted.
- **`$`-prefixed keys round-trip on the atomic path:** stores `settings = {"$ref": ..., "$type": "node"}` (the shape the `Serializer` emits for a repeated anchor reference) and asserts it round-trips verbatim — pinning the second failure mode the whole-value wrap covers. Runs on `mongo:latest` since `$ref` is misread as an operator on every version.

A dedicated `mongo:6.0` container is required for the first test because the shared `mongo:latest` helper accepts bare `{}` and cannot reproduce 40180. A `mongo:latest` clear-to-`{}` test was deliberately omitted (6.1+ accepts bare `{}`, so it would pass with or without the fix), as was a `put_partial` test (that path uses a classic `$set` this PR does not change).

The pre-existing **#6076 test** was updated: it asserted the stored document contained *no* empty dicts (a proxy for the old strip approach). Since the uniform `$literal` rule now legitimately stores `access.roots.anchors = {}`, that assertion was flipped to the real invariant — the atomic write succeeds and `access` is persisted despite the default empty dict.

## Reproduction matrix

| Image | Server | Pipeline `$set` with bare empty dict | Behavior with fix |
|-------|--------|-------------------------------|----------|
| `mongo:6.0` | 6.0.x | ❌ raises 40180 | `$literal` wrap stores `{}` |
| `mongo:7.0` | 7.0.x | ✅ OK | `$literal` wrap stores `{}` |
| `mongo:latest` | 8.x | ✅ OK | `$literal` wrap stores `{}` |

Closes #6167.


akindu-k:
Refer: https://jira.mongodb.org/browse/SERVER-54046

jac-scale/jac_scale/impl/memory_hierarchy.mongo.impl.jac
```diff
@@ -925,13 +970,15 @@ impl MongoBackend._put_node_atomic(
...
         dirty = setattr_dirty | hash_dirty;
 
-        # Serialize only the dirty archetype fields.
-        partial_data = Serializer.serialize_fields(anchor, dirty) if dirty else {};
+        serialized: dict = Serializer.serialize_fields(anchor, dirty) if dirty else {};
+        partial_data: dict[str, object] = (
+            cast(dict, _strip_empty_dicts(serialized))
+                if self._strip_pipeline_empty_dicts
+                else serialized
+        );
```
udithishanka: there's a simpler fix that also closes the data-loss trade-off: instead of version-gating a strip, recursively wrap empty-dict leaves in a $literal expression before they go into the pipeline $set. verified this on a real mongo 6.0.28 container: a raw empty dict is rejected with 40180, but {'$literal': {}} is accepted and stores a real empty object, and nested leaves work too ({'a': {'$literal': {}}, 'b': 2} stores {'a': {}, 'b': 2}).

it matters here because jac-scale provisions mongo:6.0 on k8s, so the documented trade-off (a deliberate clear-to-{} can't be expressed on < 6.1) silently loses writes on the default production deployment, not some EOL edge case. with the literal wrap the behavior is identical on every server version: no version sniffing, no process-cache key, no postinit field, no declaration change, and clear-to-{} persists everywhere. roughly:

```jac
def _wrap_empty_dicts_literal(v: object) -> object {
    if isinstance(v, dict) {
        if not v {
            return {'$literal': {}};
        }
        return {k: _wrap_empty_dicts_literal(vv) for (k, vv) in v.items()};
    }
    if isinstance(v, list) {
        return [_wrap_empty_dicts_literal(item) for item in v];
    }
    return v;
}
```

applied unconditionally to partial_data on this pipeline path only (put_partial's classic $set must stay raw, as you already have it). the ~40 lines of version-detection machinery collapse into this one transform, and the mongo:latest test and the 6.0 test can then assert the exact same clear-to-{} behavior.

udithishanka (changes requested):
One design suggestion that also closes the clear-to-empty data loss on the provisioned mongo:6.0.

akindu-k:
image

PR #6296, opened 2026-05-31
fix(testing): scope JacAPIServer config to base_path; eliminate set_config() global mutation in SPA tests

### LINKED ISSUE #6169: flaky: "get root endpoint" test intermittently fails under pytest-xdist parallel workers
## Summary

`jac/tests/runtimelib/test_testing.jac::get root endpoint` fails non-deterministically when the suite runs with `-n auto` (pytest-xdist). The assertion `assert response.ok` at line 46 fails roughly 1 in 20 runs with no code change between passing and failing runs.

## Observed failure

```
______________________________ get root endpoint _______________________________
[gw1] linux -- Python 3.12.13
  jac/tests/runtimelib/test_testing.jac:46 in test_get_root_endpoint
    assert response.ok;
E   AssertionError:
```

Seen on CI run [26417727074](https://github.com/jaseci-labs/jaseci/actions/runs/26417727074) (branch `fix/recover-all-ordering-and-missing-endpoint-warning`, 19/20 runs passed with identical code) and independently on the `main` branch.

## Root cause hypothesis

`JacTestClient` executes requests in-process via a `MockHandler` that invokes `do_GET()` on the server's handler class. The handler and its introspector share process-level state (`Jac.loaded_modules`, the module hub, `ModuleIntrospector._walkers`) across xdist worker processes. When a parallel worker leaves that state dirty — e.g. a partially initialised or torn-down module — a concurrent worker's `GET /` can return a non-200 response.

`25d624008` ("move app identity onto ExecutionContext to isolate in-process clients") addressed some of this isolation, but the flakiness persists, so at least one shared-state access path remains.

## Repro notes

- Only observed under `-n auto` (multi-worker). The test is likely clean when run in isolation or with `-n0`.
- Worker `gw1` is consistently the failing worker, suggesting it picks up a module cache populated by another worker.

## Suggested fix

Ensure `JacTestClient` gets a fully isolated view of `Jac.loaded_modules` / the module hub for each client instance, or add a per-test lock around module load/unload so workers cannot observe each other's intermediate state.

Alternatively, mark the test `@pytest.mark.xdist_group("testing_client")` so all `test_testing.jac` tests run on the same worker and cannot race with unrelated tests.

akindu-k:
## Problem

`jac/tests/runtimelib/test_testing.jac::get root endpoint` was failing intermittently (~1 in 20 runs) under `pytest-xdist -n auto` with no code changes between passing and failing runs.

### Root cause

`make_spa_client()` in `test_serve_client.jac` configured the server by mutating the **process-level global config singleton**:

```python
set_config(JacConfig(serve=ServeConfig(base_route_app="client_page")))
```

Each SPA test reset it afterwards via `set_config(JacConfig())` — but that cleanup only ran if **all assertions passed**. If any assertion in a SPA test failed before the cleanup line (e.g. `assert "<!DOCTYPE html>" in html` failing due to a transient rendering issue), the dirty config stayed in memory.

When `get root endpoint` ran next on the **same xdist worker process**, `do_GET` called `get_config()`, received the stale config with `base_route_app="client_page"`, and routed `GET /` to `_render_page("client_page")`. `serve_api.jac` has no client bundle, so `render_page` raised `RuntimeError` → the handler returned **500** → `assert response.ok` failed with `AssertionError`.

Because `-x` stops after the first failure, only `get root endpoint` appeared in CI logs — the SPA test whose cleanup failed was swallowed.

## Fix

**Two-part change:**

### 1. `JacAPIServer` owns its config

Added `_app_config: (JacConfig | None) = None` to `JacAPIServer`. In `postinit`, it is populated by calling `JacConfig.discover(start_path=Path(self.base_path))` — the same discovery the real production code path uses when reading a project's `jac.toml`.

`do_GET` now reads `config = server._app_config` instead of `config = get_config()`. Each server instance carries its own isolated config snapshot taken at construction time; mutations to the process-level singleton by other in-process clients cannot affect routing.

### 2. `make_spa_client` writes a real `jac.toml`

Instead of calling `set_config()` on the global singleton, `make_spa_client` now writes a minimal `jac.toml` into the fixture tmpdir:

```toml
[serve]
base_route_app = "client_page"
```

`JacAPIServer.postinit` discovers it from `base_path` — exactly the same path a real SPA app would take. All `set_config(JacConfig())` cleanup calls in the SPA tests are removed; they are no longer needed.

## CI runs that showed the flakiness
- https://github.com/jaseci-labs/jaseci/actions/runs/26417727074 (original flaky run)
- https://github.com/jaseci-labs/jaseci/actions/runs/26466744981 (PR #6194, same failure)
- https://github.com/jaseci-labs/jaseci/actions/runs/26707768830 (PR #6294, same failure)

Closes #6169


akindu-k:
closing this as it is addressed by #6318.

Issue #6266, opened 2026-05-29
Concurrent walker check-then-create can produce duplicate child nodes


udithishanka:
## Summary

Concurrent walker invocations that do "look up a node, create it if missing" against the same parent root can both pass the existence check and both create a child node. The persistence layer has no way to express "there should be exactly one of these per parent", silently swallows the `DuplicateKeyError` if an operator adds a unique index by hand, and returns 200 to the walker while the second write was dropped. This is the node-identity analog of #5446 (closed by #5644 for the edge-merge race). Same diagnosis, same fix shape: push atomicity into the backend, not into app code.

This issue specifies the full surface needed to close the bug class, with concrete file paths, line numbers, and code shapes per change. It is intended to be implementable without further design conversation.

## Reproduction

`repro.jac`:

```jac
import from uuid { uuid4 }

node UserProfile {
    has user_id: str = "",
        tier: str = "free";
}

walker me {
    can handle with Root entry {
        profiles = [-->(?:UserProfile)];
        if profiles and len(profiles) > 0 {
            report {"existing": profiles[0].user_id};
        } else {
            new = here ++> UserProfile(user_id=str(uuid4()));
            report {"created": new[0].user_id if isinstance(new, list) else new.user_id};
        }
    }
}
```

Runner:

```bash
jac start repro.jac &
sleep 2
TOKEN=$(curl -s -X POST http://localhost:8000/user/login \
  -d '{"email":"a@a.com","password":"x"}' | jq -r .token)
for i in 1 2 3 4; do
  curl -s -X POST http://localhost:8000/walker/me \
    -H "Authorization: Bearer $TOKEN" &
done
wait
mongosh --quiet --eval 'db.archetype.countDocuments({"arch_type":"UserProfile"})'
```

Observed: `N > 1`. Expected: `1`. Same root accumulates N edges pointing at N distinct `UserProfile` documents with N distinct `user_id` values. All N walker responses return 200 with no signal that anything went wrong.

The real walker this surfaced on is `jacBuilder/services/ideServer.jac:609-687`. Each duplicate triggers downstream side effects (signup-bonus grant, billing customer creation, ws-token registration) per duplicate.

## How persistence actually works today

Understanding why this race is structural requires the layered model. All file references are inside the `jaseci-labs/jaseci` repo unless noted.

[...]

Issue #5446, opened 2026-04-03
Graph persistence race condition: concurrent walker commits cause node/edge loss


ahzan-dev:
## Summary

Concurrent walker invocations that modify the same graph node cause **silent data loss** due to a last-writer-wins race condition in the persistence layer. The entire node document (including its edge list) is replaced unconditionally at commit time, so the last walker to commit overwrites changes made by earlier concurrent walkers.

## Reproduction

1. Start a Jac application with `jac start` (or `jac start --scale`)
2. Create a walker that adds an edge to the user root: `profile ++> MyNode(...)`
3. Immediately fire 10+ concurrent walker requests that read/mutate the same `profile` node (e.g., updating a `last_accessed` timestamp)
4. After all requests complete, query `[profile -->(?:MyNode)]`
5. **Result**: The newly created `MyNode` edge is missing — it was overwritten by a concurrent walker's stale snapshot

### Concrete example (Jac Builder IDE)

```
# User creates a project (adds edge to profile node)
POST /walker/project_ops  {action: "create", name: "my-project"}
# → success, Project node created, edge added to profile

# Frontend navigates to IDE, fires ~12 concurrent requests:
POST /walker/project_ops  {action: "open", project_id: "prj_123"}   # mutates profile.updated_at
POST /walker/git_ops      {action: "status", project_id: "prj_123"}
POST /walker/ide_file_ops {action: "setup", project_id: "prj_123"}
POST /walker/ai_chat      {action: "load_history", ...}
POST /walker/git_ops      {action: "branches", ...}
# ... 7 more concurrent requests

# User navigates back to dashboard
POST /walker/project_ops  {action: "list"}
# → Project is GONE. Only previously-existing projects remain.
```

This is 100% reproducible. The more concurrent requests, the higher the chance of data loss.

## Root Cause

### The read-modify-write cycle (no isolation)

Each HTTP request gets its own `ExecutionContext` with a private L1 memory dict (`__mem__`). At request end, `commit()` flushes changed anchors to the persistent store. **There is no concurrency control at any level.**

### The race

```
Request A (creates edge)              Request B (updates timestamp)
────────────────────────              ──────────────────────────────
1. Load profile from DB
   edges = [existing_proj]
                                      2. Load profile from DB
                                         edges = [existing_proj]  ← SAME snapshot

3. profile ++> NewProject()
   L1 edges = [existing_proj, new_proj]

4. commit() → WRITE to DB
[...]

PR #6297, opened 2026-05-31
fix(jac-scale): invalidate L2 Redis cache after edge-delta atomic write (#6295)

### LINKED ISSUE #6295: jac-scale L2 Redis cache serves stale/empty NodeAnchor.data.edges (edge-delta merge is L3-only, no cache invalidation)
## Summary

jac-scale's L2 Redis anchor cache (`RedisBackend`, enabled by default via `ScaleTieredMemory.use_cache=True`) diverges from L3 (Mongo) on `NodeAnchor.data.edges`. Concurrent/incremental edge writes are merged **server-side in Mongo** by `_put_node_atomic` (`$setUnion`/`$setDifference`, the #5644 fix), but that merged result is never reflected into the in-memory anchor or into L2. The L2 write serializes the **in-memory** `data.edges` (the local writer's un-merged view). So after an edge write, Redis holds a stale or empty edge list while Mongo holds the correct one, and reads served from L2 silently return wrong results. There is also no cross-pod invalidation, so the stale entry persists and is continually re-written by normal traffic.

The #5644 edge-delta-merge fix was applied to the L3 path only. L2 was left with the original stale-snapshot / last-writer-wins behavior. This is the L2 analog of #5446/#5644.

## Mechanism (source-grounded)

- `ScaleTieredMemory.commit()` (`jac-scale/jac_scale/impl/memory_hierarchy.main.impl.jac`): builds `l2_pending` from dirty anchors, runs `self.l3.sync()` (which performs the edge-delta merge into Mongo), then for each dirty anchor calls `self.l2.put(mem_anchor)` using the **in-memory** anchor, not the merged result.
- `RedisBackend.put` (`jac_scale/jac_scale/impl/memory_hierarchy.redis.impl.jac:70`): `Serializer.serialize(anchor)` serializes the in-memory `data.edges` verbatim into Redis.
- `MongoBackend._put_node_atomic` (`jac-scale/jac_scale/impl/memory_hierarchy.mongo.impl.jac:764`): the only place edges are merged, via a server-side `$setUnion($setDifference(data.edges, removed), added)` pipeline. The merged value exists only in Mongo; the in-memory anchor's `data.edges` is never updated to match.
- An L2 ref already exists (`MongoBackend.set_cache_ref`, `:305`) and `_put_node_atomic` has invalidation hooks for quarantine (`:244`), but the normal commit path's `self.l2.put()` re-writes the stale in-memory snapshot and clobbers any invalidation.

Net: L2 and L3 disagree on `data.edges` for any node with incremental/concurrent edge writes. No cross-pod cache invalidation means a stale L2/L1 entry survives indefinitely.

## Reproduction / production evidence

Observed on jac-builder (jac-scale 0.2.21 from `main`, MongoDB + Redis, 10 replicas):

- A user `Root` in Mongo: `data.edges = [<edge -> UserProfile>]` (valid, traversable).
- The same root in Redis: `anchor:<root-id>` cached with `data.edges = []`.
- Pods serving from L2 read the empty root, so `[-->(?:UserProfile)]` returns nothing. The app's `walker me` then mints a new `UserProfile`, and every profile-gated walker returns "User profile not found."
- `DEL anchor:<root-id>` in Redis does not stick: the next commit re-writes the stale in-memory snapshot. No invalidation propagates the correct state to other pods.
- It compounds #6266: the empty L2 read drives the check-then-create branch, so each page load / retry mints another duplicate profile (observed 2 dups per login, ~16s apart, repeatedly).

### Verified minimal reproduction (run against jac-scale 0.2.21 from `main`)

A single edge write is enough. Create a fresh user and call `me` once (its body does `here ++> UserProfile(...)`, one edge write on the root), then compare the root anchor's edge set in L3 (Mongo) vs L2 (Redis):

```
POST /user/register  {"identities":[{"type":"username","value":U}],"credential":{"type":"password","password":PW}}   -> 201
POST /user/login     {"identity":{"type":"username","value":U},"credential":{"type":"password","password":PW}}        -> {"data":{"token":JWT}}
POST /walker/me      (Authorization: Bearer JWT)   # creates UserProfile, appends edge root -> profile

# inspect the same root anchor in each layer:
mongo:  db._anchors.findOne({_id:<root>}).data.edges.length   => 1     # L3 has the edge (and 1 UserProfile node exists)
redis:  GET anchor:<root>  ->  json .data.edges.length        => 0     # L2 cached the root with NO edges
```

Observed result: **L3 `data.edges` = 1, L2 `data.edges` = 0** for the same anchor, immediately after one edge write. A standalone unit test would be: with `ScaleTieredMemory` + Mongo + Redis, do `root ++> child; Jac.commit();` then assert `l2.get(root.id).data.edges == l3.get(root.id).data.edges` (currently fails: L2 is empty, L3 has the edge).

## Impact

- Any `[-->(?:Type)]` / edge traversal can silently return a stale or empty edge set after edge writes, on any jac-scale app using L2 caching (the default). All responses are 200; there is no error signal.
- Breaks find-or-create and edge-gated logic; turns #6266 into a self-sustaining duplicate-creation loop in production.
- No supported mitigation: `use_cache` is hardcoded `True` with no env/`jac.toml` override; turning it off requires a code change and forfeits the cache entirely.

## Root cause

The edge-delta merge is L3-only. The L2 write is not edge-delta-aware (serializes un-merged in-memory edges), and there is no cross-pod invalidation on edge mutation. So L2, and downstream L1 on sibling pods, holds a stale edge set indefinitely.

## Proposed fix (robust, not a toggle)

The cache must never serve an un-merged edge set. Two coordinated changes plus optional hardening:

1. **Make the edge-write path invalidate L2 instead of snapshotting it.** In `_put_node_atomic` (and any path that mutates `data.edges`), after the server-side `$setUnion` merge, delete the L2 entry for that `NodeAnchor` (via the existing `_l2_ref`) rather than relying on `commit()` to `put()` the in-memory copy. Correspondingly, `commit()` must **not** `self.l2.put()` an anchor whose write went through the edge-delta path. The next read then misses L2 and re-hydrates the authoritative merged edges from L3.
   - Higher-cost alternative that keeps the cache warm: after the merge, read the merged `data.edges` back from Mongo and write that authoritative value to L2 (and to the in-memory anchor), so subsequent in-process reads are also correct.

2. **Cross-pod invalidation on edge mutation.** Publish an anchor-invalidation over a Redis pub/sub channel so every pod drops the mutated anchor from its in-process L1. Without this, a sibling pod's stale L1 re-`put()`s the stale snapshot into L2 and re-poisons it even after invalidation. This is the cross-pod-invalidation gap called out as out-of-scope in #6266; it has to be in-scope for the cache to be correct under multi-replica deployment. (Maintainers to confirm whether jac-scale already has an anchor invalidation channel; jacBuilder uses this pub/sub pattern app-side for its sandbox cache, so the primitive is proven, the question is wiring it to anchor edge writes.)

3. **Defense in depth (optional).** Version/hash-stamp anchors so a read can detect an L2 entry that is older than L3 and self-repair, instead of trusting L2 unconditionally. This also guards against direct-to-Mongo writes (migrations, manual ops) that bypass the cache layer.

## Not the fix

`use_cache=False` (or not provisioning Redis) disables the L2 anchor cache and removes the symptom, but also removes the cross-pod cache and is only a stopgap. There is currently no config switch for it regardless. The defect is the cache write being un-merged + un-invalidated; that is what should be fixed.

## Related

- #5644: edge-delta merge, applied to L3 only. This issue is the L2 analog.
- #6266: concurrent check-then-create duplicate child nodes. Amplified by this cache bug (empty L2 read triggers the create branch).
- #5446: original edge-loss persistence race.

akindu-k:
## The Problem

jac-scale's L2 Redis anchor cache diverged from L3 (MongoDB) on `NodeAnchor.data.edges` after any edge write. The symptom in production: `[-->(?:UserProfile)]` traversal returned nothing despite the edge existing in MongoDB, causing every edge-gated walker to silently fail.

### Root cause

The edge-delta merge introduced in #5644 was applied to the L3 path only. The L2 write was never made edge-delta-aware.

`ScaleTieredMemory.commit()` worked like this:

```
1. Capture l2_pending = [dirty anchors in L1]   ← snapshot taken BEFORE sync
2. self.l3.sync()                                ← _put_node_atomic runs here
3. for anchor in l2_pending:
       self.l2.put(anchor)                       ← writes IN-MEMORY snapshot to Redis
```

`MongoBackend._put_node_atomic` performed the correct server-side merge in MongoDB using `$setUnion`/`$setDifference`, but that merged result only ever existed in MongoDB. The in-memory anchor's `edges` list was never updated to match. So step 3 serialised the stale, un-merged in-memory edge list straight back into Redis.

**Net result:** After one edge write —

| Layer | `data.edges` |
|---|---|
| L3 (MongoDB) | `[<edge → UserProfile>]` ✅ |
| L2 (Redis) | `[]` ❌ |

Any pod serving requests from L2 got an empty edge list. No error was raised. The only signal was behavioural: edge traversal returned nothing, find-or-create logic minted duplicate nodes (#6266), and `DEL anchor:<id>` in Redis had no lasting effect because the next commit re-wrote the stale snapshot.

---

## The Fix

Three coordinated changes across the memory hierarchy.

### 1. `_put_node_atomic` writes-and-reads-back in one round-trip, then invalidates L2

The write and the authoritative read-back are now a single `find_one_and_update`:

```python
merged_doc = self.collection.find_one_and_update(
    {'_id': _id},
    [{'$set': set_doc}],                 # same $setUnion/$setDifference merge
    projection={'data.edges': 1},
    upsert=True,
    return_document=ReturnDocument.AFTER # post-merge image, from the primary
)
```

Because `find_one_and_update` is a write, it always runs on the primary and returns the document *after* the merge — so there is no second query and no replica-lag window where a secondary could return pre-write state. With that authoritative edge set the method then:

- **Updates `anchor._initial_edge_ids`** to the merged `frozenset` — this repairs future `edge_delta()` computations on the same pod so they use the post-merge baseline instead of the pre-commit snapshot.
- **Filters `anchor.edges`** to drop any `EdgeAnchor` stubs the server-side merge did not keep (handles concurrent removes from sibling pods).
- **Invalidates the Redis key** via `self._l2_ref.invalidate(anchor.id)` — the in-memory snapshot is the un-merged delta, so the next reader must fall through to MongoDB and receive the authoritative edge list. (The read-back and field-hash snapshotting are each wrapped in their own `try` so a failure there cannot skip the invalidation.)
- **Records `anchor.id` in `self._atomic_written`** so `commit()` knows to skip the stale `l2.put()` call for this anchor.

### 2. `MongoBackend.sync()` returns the set of atomically-written IDs

`sync()` resets `_atomic_written = set()` at the start of each cycle, accumulates IDs as `_put_node_atomic` runs, and **returns** the set. The set is now a private implementation detail of the backend, surfaced through a clean return-value contract rather than a public field that callers reach into.

### 3. `ScaleTieredMemory.commit()` skips `l2.put()` via the `sync()` return value

```jac
atomic_ids = self.l3.sync() or set();
for mem_anchor in l2_pending {
    if mem_anchor.id in atomic_ids {
        continue;   # already invalidated — do not re-poison L2
    }
    self.l2.put(mem_anchor);
}
```

No `isinstance` backend type-sniffing and no reach into a backend-private attribute. Backends that don't use the atomic path (e.g. `SqliteMemory`) return `None`, so `or set()` yields an empty set and every pending anchor is cached exactly as before. This is what prevents the original bug — the stale in-memory snapshot is never written back for atomically-merged anchors.

---

## Files Changed

| File | Change |
|---|---|
| `jac_scale/memory_hierarchy.jac` | Add `_atomic_written: set by postinit` to `MongoBackend`; change the `sync` declaration to `-> set` |
| `jac_scale/impl/memory_hierarchy.mongo.impl.jac` | Init `_atomic_written` in `postinit`; reset and return it from `sync()`; in `_put_node_atomic`, collapse the write + read-back into one `find_one_and_update`, repair the in-memory anchor, invalidate L2, and track `_atomic_written` |
| `jac_scale/impl/memory_hierarchy.main.impl.jac` | `commit()` consumes `sync()`'s returned ID set to skip re-writing atomically-merged anchors to L2 |
| `jac_scale/tests/test_memory_hierarchy.jac` | Regression test: seeds MongoDB with an empty-edges root, loads it into L2, adds an edge, commits, and asserts L2 is invalidated (not re-written with `edges=[]`) while L3 holds the correct merged edge list; updates the Redis key-count assertion in "Redis write cycle via commit" (parent key is now invalidated, not re-written) |

---

## What Was Not Changed

- The server-side `$setUnion`/`$setDifference` merge expression itself is unchanged — it was already correct.
- The SQLite fallback path is unaffected (`sync()` returns `None` there, handled by `or set()`).
- No configuration flags or breaking API changes.

---

## Scope note: cross-pod L1 invalidation deferred

This PR fixes staleness on the **writing pod** (its L1 and the shared L2). In a multi-pod deployment, sibling pods still hold their own in-memory (L1) copies and are not yet notified to evict them. Broadcasting invalidations across pods (Redis pub/sub + a per-process listener thread) is a distinct, larger feature with its own correctness concerns (self-message suppression, listener-thread lifecycle on disconnect, object-leak avoidance, stable L1 registry keying). It is tracked separately in **#6313** and will land in its own PR. 

---

## Testing

`pytest jac-scale/jac_scale/tests/test_memory_hierarchy.jac` — **21 passed**.

The regression test directly reproduces the scenario from the issue report:

1. Insert a root node with `data.edges = []` into MongoDB.
2. Trigger a tiered read — Redis is now populated with the empty-edges root.
3. Append a new `EdgeAnchor` stub to the in-memory anchor (simulating `root ++> child`).
4. Call `commit()`.
5. Assert MongoDB has the edge (`l3_edges` contains `edge_id`).
6. Assert Redis does **not** hold the stale `edges = []` snapshot — the key is either absent (invalidated) or carries the merged edge list.
7. Assert `anchor._initial_edge_ids` was updated to the merged frozenset so future `edge_delta()` calls compute against the correct baseline.

The test uses the shared-container pattern (`_get_mongo` / `_get_redis` / `_reset_*`) consistent with the rest of `test_memory_hierarchy.jac`.

---

## Related

- #6295 — this issue
- #6313 — cross-pod L1 invalidation (follow-up, descoped from this PR)
- #5644 — edge-delta merge (L3-only; this PR is the L2 analogue)
- #6266 — concurrent duplicate-child creation (amplified by this cache bug)
- #5446 — original edge-loss persistence race

Closes #6295.


akindu-k:
# Follow-up: Cross-pod L1 cache invalidation

**Tracking issue:** [jaseci-labs/jaseci#6313](https://github.com/jaseci-labs/jaseci/issues/6313)

The L2 stale-edge fix in PR #6297 fully resolves the single-process bug (#6295): after an edge-delta atomic write, MongoDB's merged edge set is read back in one `find_one_and_update` round-trip and the L2 (Redis) entry is **invalidated** rather than re-written with the un-merged in-memory snapshot.

That fix covers L1 and L2 on the **writing pod only**. In a multi-pod deployment, sibling pods still hold their own stale in-memory (L1) copies. The cross-pod broadcast that evicts those copies (Redis pub/sub channel + a per-process listener thread) was prototyped, then **descoped from #6297** because it is a distinct, larger feature with its own correctness concerns (self-message suppression, listener-thread lifecycle on disconnect, object-leak avoidance, stable L1 registry keying).

It will be implemented in a separate PR under issue **#6313**.

akindu-k:
## L1 / L2 / L3 cache states — before vs. after the fix (#6295)

**Scenario:** a node `root` starts with `data.edges = []` (cached in every layer), then one edge is added — `root ++> child` — followed by `commit()`.

| Layer | What it is | Scope |
|-------|------------|-------|
| **L1** | in-process dict (`__mem__`) | per request, per pod |
| **L2** | Redis (`RedisBackend`) | shared across all pods |
| **L3** | MongoDB (`MongoBackend`) | source of truth |

---

### ❌ Before the fix

After `root ++> child; commit();` on the **writing pod**:

| Layer | `data.edges` | How it got there |
|-------|--------------|------------------|
| **L3** (Mongo) | `[edge→child]` ✅ | `$setUnion`/`$setDifference` merged it server-side |
| **L2** (Redis) | `[]` ❌ | `commit()` wrote the **stale in-memory snapshot** back |
| **L1** (writing pod) | `[edge→child]` ⚠️ | correct *this request*, but `_initial_edge_ids` baseline never updated |
| **L1** (sibling pods) | `[]` ❌ | never told to evict; re-`put()`s the stale value back into L2 |

**Symptom:** Mongo has the edge, Redis says there are none. Any read served from L2 returns an empty edge list → `[-->(?:Type)]` finds nothing → no error, just wrong results. Manually deleting the Redis key doesn't stick — the next commit re-writes the stale snapshot.

```mermaid
flowchart LR
    L1w["L1 writing pod<br/>edges = [child] ⚠️"]
    L2["L2 Redis<br/>edges = [] ❌"]
    L3["L3 Mongo<br/>edges = [child] ✅"]
    L1s["L1 sibling pod<br/>edges = [] ❌"]

    L3 -. "merge stays trapped in Mongo" .- L2
    L1s -- "re-poisons" --> L2
    classDef bad fill:#fecaca,stroke:#b91c1c,color:#000;
    classDef warn fill:#fde68a,stroke:#b45309,color:#000;
    class L2,L1s bad;
    class L1w warn;
```

---

### ✅ After the fix

After `root ++> child; commit();` on the **writing pod**:

| Layer | `data.edges` | How it got there |
|-------|--------------|------------------|
| **L3** (Mongo) | `[edge→child]` ✅ | same server-side merge (unchanged) |
| **L2** (Redis) | **key deleted** ✅ | `_put_node_atomic` invalidates it; `commit()` skips re-write |
| **L1** (writing pod) | `[edge→child]` ✅ | repaired from merged read-back; `_initial_edge_ids` updated too |
| **L1** (sibling pods) | `[]` ⚠️ | **still stale** — cross-pod eviction deferred to #6313 |

**Next read of `root`:**

```
L1 miss → L2 miss (key gone) → L3 hit → returns [edge→child] → re-caches L2
```

So L2 self-heals from the source of truth instead of serving a stale snapshot.

```mermaid
flowchart LR
    L1w["L1 writing pod<br/>edges = [child] ✅<br/>baseline repaired"]
    L2["L2 Redis<br/>KEY DELETED ✅"]
    L3["L3 Mongo<br/>edges = [child] ✅"]
    L1s["L1 sibling pod<br/>edges = [] ⚠️ (→ #6313)"]

    L1w -- "invalidate" --> L2
    L2 == "next read re-hydrates" ==> L3
    classDef good fill:#bbf7d0,stroke:#15803d,color:#000;
    classDef warn fill:#fde68a,stroke:#b45309,color:#000;
    class L1w,L2,L3 good;
    class L1s warn;
```

---

### What specifically changed per layer

| Layer | Before | After |
|-------|--------|-------|
| **L3** | merged edges (correct) | merged edges (correct) — **unchanged** |
| **L2** | overwritten with stale `[]` | **invalidated** (deleted); re-hydrated on next read |
| **L1 (writer)** | edges right, `_initial_edge_ids` stale | edges **and** `_initial_edge_ids` repaired from Mongo |
| **L1 (siblings)** | stale, re-poisons L2 | stale — **out of scope**, tracked in #6313 |

> **Scope note:** This PR makes the **writing pod** and the **shared L2** consistent. Sibling pods' in-process L1 copies are not yet evicted across pods; that requires a pub/sub broadcast and is tracked separately in **#6313**. Until then the safety net is that L2 is correct, so any pod that misses its L1 pulls the right value.

---

### Execution flow after the fix

**Write path** — `root ++> child; commit();`

```
commit()
  └─ l3.sync()
       └─ _put_node_atomic(anchor)
            1. find_one_and_update(merge $setUnion/$setDifference, return AFTER)  → merged edges
            2. invalidate L2 key  +  record id in _atomic_written      ◄── done first, right after write
            3. repair L1: anchor.edges + _initial_edge_ids = merged set
       └─ returns atomic_ids
  └─ for each dirty anchor:
       if id in atomic_ids → SKIP l2.put()   (already invalidated — don't re-poison)
       else                → l2.put()         (normal write-through, unchanged)
```

**Read path** — next read of `root` re-hydrates the cache from the source of truth:

```
L1 miss → L2 miss (key was deleted) → L3 hit (merged edges) → populate L2 + L1 → return [child]
```

```mermaid
flowchart TD
    A["commit()"] --> B["l3.sync() → _put_node_atomic"]
    B --> C["1. find_one_and_update (merge, return AFTER) → merged edges"]
    C --> D["2. invalidate L2 key + record id"]
    D --> E["3. repair L1: edges + _initial_edge_ids"]
    E --> F{"dirty anchor in atomic_ids?"}
    F -- "yes (edge write)" --> G["SKIP l2.put — already invalidated"]
    F -- "no (field write)" --> H["l2.put — write-through"]
    G --> I["later READ: L1 miss → L2 miss → L3 hit → re-cache"]
    H --> I
    classDef fix fill:#bbf7d0,stroke:#15803d,color:#000;
    class C,D,E,G fix;
```

Issue #6146, opened 2026-05-24
Native: design + implement Object-Spatial Programming (walker/node/edge/visit/spawn) runtime


marsninja:
# Native + cross-backend: one portable Object-Spatial runtime, used by every codespace

## Goal

**One source-of-truth OSP runtime** — a portable `osp_kernel` written in the
Jac portable subset — that **every backend (sv/cl/na) lowers `spawn` / `visit` /
`disengage` / `report` into**, with backend-specific persistence, permission,
and identity pushed behind a single `OspHost` seam. No per-backend
reimplementation; no behavioral drift; cross-backend equivalence enforced by
tests that run the same source program through every backend and compare
results.

The architecture (kernel + `OspHost` seam, `(walker_tag, loc_tag) → fn_ptr`
static dispatch tables, kernel structs vs sv-only anchor sidecar) and the
testing strategy live in the design comments below.

## Current state (as of main `8819f2a0b`; Phase 1 landed in #6286, Phase 2 in #6302)

**Landed:**

- `jac/jaclang/jac0core/osp_kernel.jac` — portable kernel: `osp_spawn` /
  `osp_visit` / `osp_disengage` / `osp_report` + `Slot` / `WalkerDesc` /
  `NodeDesc` / `WalkScope` / `OspRuntime`.
- `jac/jaclang/jac0core/osp_kernel_sv.jac` — sv `OspRuntime` factory
  (`make_sv_runtime`, descriptor builder over live archetypes).
- #6158 — compiler emits each ability's `.trigger` (drops `inspect.signature`);
  sv legacy engine reads compiler-stored triggers directly.
- #6263 — module-level `obj` / `node` / `edge` / `walker` declarations shared
  across all codespaces; sv / cl / na codegen materialize the shared types
  referenced by their own code.
- **#6286 — Phase 1: sv now lowers `spawn` / `visit` / `disengage` / `report`
  to the kernel** via `make_sv_runtime`, and the legacy **sync** walker engine
  (`_execute_entries` / `_execute_exits` / `_visit_node_recursive`) is deleted.
  The kernel was extended to sv parity (DFS recursion, post-order exits,
  `WalkScope.path`, `osp_visit` `insert_loc`, opt-in visit-once via
  `scope.ignores`, multi-`starts` spawn). Full sv suite green through the
  kernel. (The **async** walker still uses the legacy CallState/`OspHost` path
  via `osp_spawn_async`; async kernel migration is future work.)
[...]

PR #6302, opened 2026-05-31
test(osp): cross-backend OSP equivalence fixtures (Phase 2 of #6146)

marsninja:
## Summary

Phase 2 of the OSP kernel roadmap in #6146: the systematic cross-backend
Object-Spatial equivalence fixtures. One source program per behavior, run
through every backend that supports it and compared against the canonical
**sv (py)** reference.

Each test does two things:
1. **Pins the canonical sv semantics** with a concrete `run_py_fixture`
   assertion, so the kernel's behavior is a hard regression guard even while
   only one backend runs the fixture.
2. **Enforces cross-backend equivalence** via `run_with_both`, with `require=`
   naming the backends that must match today. As na/cl OSP lowering lands, a
   fixture's `na {}` / `cl {}` entry variant is added and its backend moves into
   `require=`, turning "support landed" into an enforced regression.

Reuses the existing harness (`xbackend_equiv.jac`); no harness changes.

## Fixtures

| Fixture | Behavior pinned | Backends |
|---|---|---|
| `osp_visit_order` | chain `[1,2,3,4]`, diamond `[1,2,3,4,4]` (double-visit), fan-out, guarded cycle `[1,2,3]` | py |
| `osp_exit_ordering` | entry-before-children, exit-after `[1,2,3,-3,-2,-1]` | py |
| `osp_subtype_dispatch` | `with Base entry` on `Sub` + tuple trigger | py |
| `osp_disengage` | early-exit partial set `[1,2,3]` | py |
| `osp_report` | reports survive spawn boundary `[10,20,30]` | py |
| `osp_edge_typed` | `[->:Road:->]` follows only the typed edge `[1,2,4]` | py |
| `osp_multi_walker` | independent walker state | **py + na** |
| `osp_nested_spawn` | ability spawns a second walker | **py + na** |

All 9 tests (incl. the existing `osp_acc`) pass. `test_prim_equivalence`
(shares the harness) unaffected.

## Backend status surfaced by these fixtures

Only the entry-only fixtures carry `require=["na"]`. The traversal subset is
py-only today because:

- **na** (inline spawn stopgap): entry-only, EXACT-type dispatch. `visit` /
  edges / `report` / `disengage` are E5090; **exit abilities compile but do not
  fire**; **subtype/tuple dispatch silently misses** (`with Base entry` on a
  `Sub` returns 0). These map to Phase 4.
- **cl**: OSP entirely unwired (`spawn` lowers to an unbound `__jacSpawn`),
  skipped on every fixture. That is Phase 3.

When that lowering lands, each fixture gains its `na {}` / `cl {}` variant and a
`require=` flip — no other churn.

## Notes for the roadmap

Two items in the issue's *Current state / Gated* section are now inaccurate and
worth correcting:
- Exit abilities are **not** rejected with E5090 on na; they compile and
  silently no-op (a silent wrong answer, not a hard gate).
- The kernel is no longer "BFS, visit-once" — post-#6286 it is DFS with opt-in
  visit-once, so it **revisits** the shared sink (diamond `[1,2,3,4,4]`),
  matching legacy sv. The diamond divergence listed as pending appears resolved.

One behavior not in the plan: the kernel's single shared `next` queue makes a
parent's exit fire **after** its later siblings (`osp_exit_ordering` pins
`[1,2,3,-3,-2,-1]`, not `[1,2,-2,3,-3,-1]`) — siblings nest under the first
rather than being independent subtrees. Pinned here so any future
"independent siblings" reconciliation is a visible, intentional change.

### DIFF
--- a/jac/jaclang/compiler/tests/fixtures/osp_disengage.jac
+++ b/jac/jaclang/compiler/tests/fixtures/osp_disengage.jac
@@ -1,0 +1,32 @@
+"""Cross-backend OSP equivalence fixture: `disengage` early-exit.
+
+The walker steps along a chain `1 -> 2 -> 3 -> 4 -> 5`, recording each value,
+and `disengage`s at `3`. `disengage` halts the walk after the in-flight ability
+returns, so the partial visit-set `[1, 2, 3]` is captured and nodes 4/5 are
+never entered.
+
+Backend status: `disengage` is E5090 on na and unwired on cl (also needs
+visit/edges), so py-only today.
+"""
+
+node N { has val: int; }
+
+walker Stop {
+    has seen: list = [];
+    can step with N entry {
+        self.seen.append(here.val);
+        if here.val == 3 {
+            disengage;
+        } else {
+            visit [-->];
+        }
+    }
+}
+
+def disengage_py() -> dict[str, any] {
+    n1 = N(val=1); n2 = N(val=2); n3 = N(val=3); n4 = N(val=4); n5 = N(val=5);
+    n1 ++> n2; n2 ++> n3; n3 ++> n4; n4 ++> n5;
+    w = Stop();
+    w spawn n1;
+    return {"seen": w.seen};
+}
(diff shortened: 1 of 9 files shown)


marsninja:
Heads-up for reviewers on `osp_exit_ordering`: it pins the **current** kernel order `[1, 2, 3, -3, -2, -1]` (a parent's exit fires after its later siblings, due to the kernel's single shared `next` queue). Per the #6146 Phase 2.5 decision, the canonical semantic is **independent sibling subtrees** `[1, 2, -2, 3, -3, -1]`, so that kernel change + this fixture's expected value (and docstring) will update together as the "visible, intentional change" the docstring references. Merging this fixture as-is is correct — it locks today's behavior so the Phase 2.5 reconciliation shows up as a diff here.

PR #6303, opened 2026-05-31
fix(type-checker): apply the descriptor protocol on member access

kugesan1105:
## What

When an attribute's type is a descriptor (a class defining `__get__`), instance access now resolves to `__get__`'s return type instead of the descriptor object.

The accessing instance selects the instance-access `__get__` overload (by assignability to its `obj` parameter) over the class-access one (`obj: None`), and TypeVars are bound from the descriptor instance — so e.g. a `Descriptor[int]` member resolves to `int`. Builtin `property` / `cached_property` keep their existing dedicated handling and are unaffected.

## Why

Member access previously returned the descriptor object itself for any descriptor other than the builtin `property`/`cached_property` special cases. Accessing such a member (common with library descriptors that implement `__get__`) therefore produced spurious type errors (e.g. an operator not supported on the descriptor object). This generalizes member access to the descriptor protocol.

This mirrors Pyright's `applyDescriptorAccessMethod`: on member access, if the member defines `__get__`, evaluate it and use its return. (Pyright builds a synthetic call through `validateCallArgs`; here the matching overload is chosen directly via assignability, since Jac's call machinery is AST-node-keyed — consistent with how existing magic-method resolution already works.)

## Tests

- New `bug_desc_1_descriptor_get_unwrapped`: a descriptor member resolves to `__get__`'s instance-access return type.
- Builtin `@property`/`@cached_property` unchanged; full checker (`test_checker_bugs`, 42) and language (93) suites pass with no regressions.

### DIFF
--- a/docs/docs/community/release_notes/unreleased/jaclang/6300.bugfix.md
+++ b/docs/docs/community/release_notes/unreleased/jaclang/6300.bugfix.md
@@ -1,0 +1,1 @@
+- Accessing an attribute whose type is a descriptor (a class defining `__get__`) now resolves to the value `__get__` returns instead of the descriptor object, so using that attribute no longer raises spurious type errors.
--- a/jac/jaclang/compiler/type_system/type_evaluator.impl/evaluator_util_methods.impl.jac
+++ b/jac/jaclang/compiler/type_system/type_evaluator.impl/evaluator_util_methods.impl.jac
@@ -882,6 +882,51 @@ impl TypeEvaluator.get_getter_type_from_property(
     return func_type.return_type;
 }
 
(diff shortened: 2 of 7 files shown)


kugesan1105:
Converting to draft. This fix is correct and regression-clean, but on its own it only handles descriptor *instances* typed directly as members. The common real-world case (a method decorated by a descriptor-producing decorator, e.g. `@util.memoized_property def rowcount(self) -> int`) produces the descriptor via a decorator, which this PR does not yet reach. Holding until the decorator-recognition layer (plus retaining a generic descriptor's type parameter through inheritance) exists, so the change can land demonstrably fixing real code rather than only its own fixture.

marsninja:

PR #6304, opened 2026-05-31
fix(type-checker): statically evaluate TYPE_CHECKING when loading Python modules

kugesan1105:
## What

The type checker now statically evaluates `if TYPE_CHECKING:` as taken and `if not TYPE_CHECKING:` as skipped when loading Python modules, per PEP 484. The condition evaluator recognizes bare `TYPE_CHECKING`, `typing.TYPE_CHECKING`, and their negation, reusing the existing `evaluate_version_condition` / `_flatten_version_conditionals` pipeline that already handles `sys.version_info` guards.

## Why

Libraries routinely declare types and aliases only for type checkers behind `if TYPE_CHECKING:` (and provide a runtime fallback under `if not TYPE_CHECKING:`). Jac ignored the guard and took the runtime branch, so those names resolved to `Unknown` (or the wrong runtime type). This mirrors how Pyright treats `TYPE_CHECKING` (`staticExpressions.ts`).

**Resolves Gap #49:** types imported under `if TYPE_CHECKING:` now resolve in annotations. The `checker_gap_fwd_ref.jac` fixture previously could not resolve a `TYPE_CHECKING`-guarded `Path`; it now resolves correctly, leaving only the two intended `str`/`int` mismatch errors.

## Tests

- `checker_gap_fwd_ref.jac` (Gap #49 fixture) resolves the guarded import; the gap test now reflects the fixed behavior.
- No regressions: `test_checker_bugs` (41) and `test_language` (93) pass.

## Out of scope

This does not fully resolve SQLAlchemy-style descriptor attributes (e.g. `CursorResult.rowcount`), which additionally require retaining a generic descriptor's type parameter through inheritance and recognizing descriptor-producing decorators. Those are tracked as separate follow-ups.

### DIFF
--- a/docs/docs/community/release_notes/unreleased/jaclang/6304.bugfix.md
+++ b/docs/docs/community/release_notes/unreleased/jaclang/6304.bugfix.md
@@ -1,0 +1,1 @@
+- The type checker now evaluates `if TYPE_CHECKING:` blocks as taken (and `if not TYPE_CHECKING:` as skipped), so types and aliases that libraries declare only for type checkers resolve correctly. Imports guarded by `if TYPE_CHECKING:` are now usable in annotations instead of resolving to `Unknown`.
--- a/jac/jaclang/jac0core/helpers.jac
+++ b/jac/jaclang/jac0core/helpers.jac
@@ -666,6 +666,24 @@ def _is_version_info_access(expr: 'uni.Expr') -> bool {
     );
 }
 
+"""Check if expr references `TYPE_CHECKING` (bare or `typing.TYPE_CHECKING`)."""
+def _is_type_checking_ref(expr: 'uni.Expr') -> bool {
+    import jaclang.jac0core.unitree as uni;
+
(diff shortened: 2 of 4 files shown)


kugesan1105:
Pushed a follow-up commit fixing a regression the first commit surfaced: a static-guard `if` inside a **function body** was being flattened (scope dropped), which made the guard appear to narrow locals used in its body, resolving them to `Unknown`. Static-guard flattening is now restricted to module/class scope (where it promotes stub declarations); inside a function the guard keeps a normal scoped block, so control-flow typing of locals is unaffected. Verified: the `redundant_typing_imports` static-analysis fixture passes again, Gap 49 and STUB-VGUARD still resolve, and checker/language/sym-tab suites are green.

Noted in-code as pragmatic: the clean generalization is to handle static guards uniformly across SymTabBuildPass (promote declarations only) and CFGBuildPass (model the dead branch as unreachable, Pyright-style), which removes the function-body special case. Tracked as a follow-up.

PR #6307, opened 2026-06-01
fix(jac2js): add .get() to poly dispatch for any-typed variables

### LINKED ISSUE #6306: jac2js: .get(key, default) on any-typed variables compiles to raw JS .get() call
# jac2js compiles `.get(key, default)` on `any`-typed variables to raw JS `.get()` call

## Summary
When a variable is typed `any` and the Jac source calls `.get("key", default)` on it, the jac2js compiler emits a raw JavaScript `.get()` method call. Plain JS objects (e.g. from `JSON.parse`) do not have a `.get()` method, so this throws `TypeError: x.get is not a function` at runtime.

The compiler already handles this correctly for variables typed `dict` (it emits bracket access `target[key] ?? default`). It should do the same for `any`-typed variables.

### Jac source (`repro.cl.jac`)
```jac
cl {
    def test_dict_get(envelope: dict[str, str]) -> str {
        pipeline = envelope.get("pipeline", "default");
        return pipeline;
    }

    def test_any_get(envelope: any) -> str {
        pipeline = envelope.get("pipeline", "default");
        return pipeline;
    }
}
```

### Generated JavaScript (from `jac2js`)
```js
// test_dict_get — CORRECT ✅
function test_dict_get(envelope) {
  const pipeline = envelope["pipeline"] ?? "default";
  return pipeline;
}

// test_any_get — BUG ❌
function test_any_get(envelope) {
  const pipeline = envelope.get("pipeline", "default");
  return pipeline;
}
```

## Expected Behavior
`test_any_get` should emit safe bracket access, just like `test_dict_get`:
```js
const pipeline = envelope["pipeline"] ?? "default";
```

## Actual Behavior
`test_any_get` emits `envelope.get("pipeline", "default")`, which throws at runtime when `envelope` is a plain JS object.

## Root Cause
In `jac/jaclang/compiler/passes/ecmascript/impl/esast_gen_pass.impl.jac`, the `exit_func_call` codegen path does type-aware dispatch:

1. If the receiver's static type is a known primitive (`dict`), it routes `.get()` to `ESDictEmitter.emit_get`, which safely emits bracket access.
2. If the static type is `any` (or unknown), it falls through to the **default codegen path**, which emits a raw `.get()` method call — incorrectly assuming the runtime value is a `Map` or has a `.get()` method.

The compiler already solves the same problem for `"key" in obj` checks on `any`-typed variables by routing them through `_jac.poly.contains`. It should apply the same transformation for `.get()`.

## Workaround
Until fixed, avoid `.get()` on `any`-typed variables in `.cl.jac` files and use bracket access with `in` checks instead:
```jac
pipeline = envelope["pipeline"] if "pipeline" in envelope else "default";
```

## Environment
- Branch: `upstream/main` (latest as of this report)
- Compiler path: `jac/jaclang/compiler/passes/ecmascript/`

chess10kp:
## Problem
Calling .get(key, default) on variables with static type 'any' in Jac-to-JS compilation emitted a raw JS .get() call, which fails at runtime when the receiver is a plain object (Jac dict) because plain JS objects do not have a .get() method.

## Solution
Add 'get' to POLYMORPHIC_METHODS so the emitter routes untyped .get() calls through _jac.poly.get().

The poly.get runtime helper handles four cases in order:
1. **Map** – delegates to Map.prototype.get
2. **Custom objects with .get()** – calls the object's own get method (before bracket notation, matching the pattern used by poly.append, poly.pop, etc.)
3. **Plain objects** – uses bracket notation with nullish-coalescing default fallback
4. **Primitives / null / undefined** – returns the default value

## Test coverage
- poly_get_dispatch.cl.jac – exercises plain-object path via an any-typed dict
- poly_get_custom_obj.cl.jac – exercises custom-object fallback via an obj with a def get() method

## Verification
All 108 ecmascript tests pass.

Fixes #6306


marsninja:
Pushed a follow-up (7eddd7152) that generalizes this beyond `.get()`.

Adding `.get()` to the `POLYMORPHIC_METHODS` allowlist fixes that one method, but the same bug stays open for every other container/string method (`.keys()`, `.values()`, `.update()`, `.add()`, `.insert()`, etc.) on an any-typed receiver, and the allowlist has to be grown by hand each time. This reworks it into the long-term shape:

- **Eligibility is derived, not hand-maintained.** The poly-eligible method set is now computed from the primitive emitter registry (`emit_*`, excluding operators), so it can't drift from the emitters and new primitive methods are covered automatically. The hardcoded set is gone.
- **One runtime dispatcher.** Codegen emits `_jac.poly.call(obj, "name", ...args)` for any eligible method on an unprovable receiver. The dispatcher resolves the receiver's runtime type and applies the matching op (reusing the shared `_jac.<type>` helpers, so the typed and untyped paths share one implementation), forwards to a same-named method on user-defined objects, or raises a catchable `AttributeError` instead of silently returning a wrong value.
- **`dict.get` correctness.** It now uses a membership check, so a present-but-null key returns the stored value rather than the default (`?? default_` wrongly substituted the default for a stored `None`).

The typed-emitter hot path is untouched. Existing poly tests are updated to the new `call` shape, and a new `poly_dispatch_complete` fixture covers dict/list/set ops, user-object forwarding, the null-vs-default fix, and the AttributeError fallback (run in Node). Full ecmascript suite (113) and cross-backend primitive-equivalence (39) pass.

marsninja:
Pushed `7c8018c8a` (rebased onto the latest branch) fixing the `test-client` CI failure this introduced.

**Root cause:** broadening poly-eligibility to the full emitter set meant `Object.keys(...)` in the client runtime routed through `_jac.poly.call(Object, "keys", ...)` — `Object` is an untyped reference, so the method call is unprovable at compile time. At runtime `Object` is the JS global (a function), and the dispatcher only handled string/number/array/Set/Map/bytes/plain-object, so it raised `'function' object has no attribute 'keys'` during render (this also tripped the `undefined JSX child` test, which renders the same page).

**Fix:** the dispatcher now forwards to the receiver's own method for any function or non-plain object that carries it (matching the direct `obj.method(...)` a typed call would emit); genuine primitives still get Python-semantic dispatch. Added an `Object.keys` regression case to the `poly_dispatch_complete` fixture.

Verified locally: the two failing e2e tests (`signup form submission`, `undefined JSX child does not crash render`) now pass, and the full ecmascript JS-generation suite stays green.

PR #6309, opened 2026-06-01
Support nginix and non-NGINX shared ingress controllers (ALB/Traefik/GKE) in jac-scale

udithishanka:
## Problem

#6012 added `shared_ingress` so jac-scale can attach to a pre-existing shared controller instead of deploying a dedicated NGINX controller (and NLB) per app. But it is NGINX-specific: it hardcodes `nginx.ingress.kubernetes.io/*` annotations and gives no way to pass controller-specific settings. Controllers that are driven entirely by annotations, most notably the AWS Load Balancer Controller (ALB), cannot be used. There is no way to set the ALB IngressGroup (to share a single ALB across apps), the ACM certificate ARN (for HTTPS), or scheme / target-type. Setting `shared_ingress_class = "alb"` today would provision a separate, ungrouped, HTTP-only ALB per app.

## Solution

Two cloud-agnostic options on `KubernetesConfig`:

- `shared_ingress_annotations: dict[str, str]` merged onto the Ingress metadata, with caller values taking precedence. This is how non-nginx controllers are configured. NGINX-specific tuning (rate limiting, session affinity, proxy timeouts) is now emitted only when `shared_ingress_class == "nginx"`, so a non-nginx Ingress stays clean.
- `shared_ingress_tls: bool` for controllers that terminate TLS out-of-band (e.g. ALB via an ACM certificate annotation, which leaves no `spec.tls`), so `get_service_url` reports `https`.

No cloud-specific values live in jac-scale; the caller supplies them. Example (ALB):

```toml
[plugins.scale.kubernetes]
shared_ingress = true
shared_ingress_class = "alb"
shared_ingress_tls = true
domain = "myapp.example.com"

[plugins.scale.kubernetes.shared_ingress_annotations]
"alb.ingress.kubernetes.io/group.name" = "shared-alb"
"alb.ingress.kubernetes.io/scheme" = "internet-facing"
"alb.ingress.kubernetes.io/certificate-arn" = "arn:aws:acm:...:certificate/..."
```

## Notes

- Builds on #6012 (`shared_ingress`). That PR should merge first; this branch includes its commits and reduces to just this delta once #6012 lands.
- Unit test added (mock-based, no live cluster): asserts caller annotations are merged, `ingressClassName` is the shared class, and nginx annotations do not leak onto a non-nginx Ingress.
- Backward compatible: dedicated mode and shared-nginx mode are unchanged.


udithishanka:
**Fixed the shared-ingress DNS/endpoint bug (thanks @kashmithnisakya)**

Confirmed root cause: `_get_lb_endpoint()` hardcoded `{app}-ingress-nginx-service`, which only exists in dedicated nginx mode. In shared-ingress mode that Service is never created, so the lookup 404'd, returned `None`, and the DNS-instruction block was silently skipped (and the 404 leaked into the debug log).

Fix in `0f460e6da`: in shared mode, read the LB hostname from the Ingress object's `.status.loadBalancer.ingress[].hostname` (published by the controller: AWS LB Controller / nginx / etc) instead of the nonexistent Service. Dedicated mode is unchanged.

kashmithnisakya:
`Reviewed functionality-wise` - works fine. Verified the ALB shared-ingress path end-to-end (deploy, ingress joins the shared ALB, https URL resolves).

`Note`: couldn't test the shared-nginx path since we don't have a shared nginx controller in the cluster.

`Approving`

kashmithnisakya (approved):

PR #6310, opened 2026-06-01
fix(type-checker): resolve false E1111 for in-operator on any-typed values

ThiruvarankanM:
### Problem

Code like this triggered a false error:

```jac
def check(x: any) -> bool {
    return "key" in x;        # ← false error here
}
```

The error:
```
error[E1111]: Right operand of "in" must support "__contains__"; "Any" does not
```

The code is valid — per PEP 484, `Any` supports all operations. Both mypy
and Pyright accept this code without any error. Only Jac rejected it.

### Why It Was Failing

There are two helpers named `is_any_type` — one on `TypeBase` (incomplete)
and one on `TypeEvaluator` (complete). The `in`/`not in` handler used the
`TypeBase` one, which doesn't recognize `x: any` declarations, so the check
fell through and wrongly emitted E1111.

### Fix

Two-line swap in `type_evaluator.impl.jac` — call the correct
`is_any_type` helper (the one on `TypeEvaluator`, not the one on
`TypeBase`):

### Result

**Codebase-wide:**
```
Removed:        141 errors        (E1111: -129)
```

### Impact

- Fixes valid PEP-484 code that was being rejected
- Removes 141 false errors across the codebase with one trivial change
- No behavior change for non-Any types or other comparison operators


kugesan1105:
#### this is the expected error, 

image


#### Devs should write more type strict code
image

PR #6356, opened 2026-06-02
Builtin cg optimizations

mgtm98:
## 1. RC Header Layout (refcount.impl.jac, na_ir_gen_pass.jac)

The heap allocation header was expanded from 16 bytes to **32 bytes**:

```
ptr - 32  alloc_id    i64  monotonic counter; upper 32 bits = HDR_MAGIC (0xAC1D)
ptr - 24  type_tag    i8*  pointer to static type-name string ("str", "raw", …)
ptr - 16  refcount    i64  RC_SENTINEL (0x7FFFFFFFFFFFFFFF) for static strings
ptr -  8  dtor_fn_ptr i64  repurposed as cached byte-length for strings (see §2)
ptr +  0  user data
```

New constants exported from `na_ir_gen_pass.jac`:

| Constant | Value | Meaning |
|---|---|---|
| `RC_SENTINEL` | `0x7FFFFFFFFFFFFFFF` | Static/sentinel RC — retain/release are no-ops |
| `HDR_ALLOC_ID_OFF` | `-32` | Offset of alloc_id field |
| `HDR_TAG_OFF` | `-24` | Offset of type_tag field |
| `HDR_RC_OFF` | `-16` | Offset of refcount field |
| `HDR_DTOR_OFF` | `-8` | Offset of dtor/cached-length field |
| `HDR_TOTAL` | `32` | Total header size in bytes |
| `HDR_MAGIC` | `44061` (0xAC1D) | Magic tag stored in upper 32 bits of alloc_id |
| `HDR_MAGIC_SHIFT_BASE` | `HDR_MAGIC << 32` | Pre-shifted constant for faster comparison |

`__rc_alloc` now:
1. Allocates `size + 32` bytes.
2. Writes a globally unique `alloc_id` (atomic fetch-and-increment of `__rc_alloc_id_counter`) at `ptr-32`, encoding `HDR_MAGIC` in the upper 32 bits.
3. Writes the type-tag pointer at `ptr-24`.
4. Writes RC=1 at `ptr-16`.
5. Writes 0 at `ptr-8` (dtor slot; overwritten by `_store_str_len` for strings).

Retain and release both do a magic-check (upper 32 bits of `ptr-32 == HDR_MAGIC`) before touching the RC, so C strings and ctypes pointers passed across the FFI boundary are silently ignored.

---

### Problem
Every `len(s)` call on a Jac string previously fell through to a `strlen` call,
making it O(N) in the string length. For a 10,000-char string accessed in a tight loop
this dominated the profile.

### Solution: `_store_str_len` + repurposed dtor slot

The `dtor_fn_ptr` field at `ptr-8` is repurposed as a **cached byte-length** for
strings. `__rc_release_simple` skips function-pointer dispatch when the type-tag is
"str", so the field is safe to reuse.

**`_store_str_len(str_ptr, byte_len)`** — writes `byte_len` (i64) to `str_ptr - 8`.
Called at every string creation site:

- `_make_global_string` — stores `len(text)` in the compile-time struct literal.
- `_codegen_fstring` — stores `snprintf` return value (bytes written).
- `_emit_binary_op` for `+` (string concat) — stores `len(left) + len(right)`.
- `chr()` builtin — stores 1.
- `_codegen_string_index` (pre-char-cache) — stored 1 per char copy.

**`_emit_jac_str_len(str_ptr)`** — O(1) length load with strlen fallback:

1. Null → 0.
2. `ptr-16 == RC_SENTINEL` (static string) → load `ptr-8`.
3. Upper 32 bits of `ptr-32 == HDR_MAGIC` (heap Jac string) → load `ptr-8` if non-zero, else strlen fallback.
4. Else → strlen (raw C string from FFI).

String concat (`+`) and all format helpers now call `_emit_jac_str_len` instead of
`strlen`, making their length computation O(1).

---

## 3. Global String Layout (builtins.impl.jac)

`_make_global_string(text)` creates a compile-time LLVM constant of type
`{ i64, i64, [N x i8] }`:

```
field 0  i64  RC_SENTINEL           → retain/release are no-ops
field 1  i64  byte_len              → O(1) len() via ptr-8
field 2  [N x i8]  null-terminated data
```

The returned `i8*` points to `field2[0]`. This means:
- `ptr-16` = RC_SENTINEL → `_emit_jac_str_len` takes the sentinel branch.
- `ptr-8` = pre-filled byte length → no strlen needed.

---

### Problem
`s[i]` previously called `__rc_alloc(2)` to create a fresh 1-byte string on every
character access. In tight loops (UTF-8 decode, string scanning) this produced
millions of heap allocations per second.

### Solution: `@__jac_chars` global + `_get_char_cache_gv` + `_codegen_string_index`

**`_get_char_cache_gv()`** builds (once per module) a compile-time global:

```llvm
@__jac_chars = private constant [256 x { i64, i64, [2 x i8] }] [
  { i64 RC_SENTINEL, i64 1, [2 x i8] [i8 0, i8 0] },  ; entry 0 = '\0'
  { i64 RC_SENTINEL, i64 1, [2 x i8] [i8 1, i8 0] },  ; entry 1 = '\x01'
  …
  { i64 RC_SENTINEL, i64 1, [2 x i8] [i8 255, i8 0] } ; entry 255
]
```

Each entry has the same layout as a static Jac string: RC_SENTINEL at `ptr-16`,
byte_len=1 at `ptr-8`, and the character byte at `ptr+0`.

**`_codegen_string_index(target_val, index_val)`** replaces the old alloc-per-char path:

```llvm
; Load byte at target_val[index_val]
%char.ptr = getelementptr i8, i8* %target_val, i64 %index_val
%char.val = load i8, i8* %char.ptr
%char.idx = zext i8 %char.val to i64
; Return pointer to cache entry's data byte — no allocation
%char.cached = getelementptr [256 x {i64,i64,[2xi8]}], …, i32 0, i64 %char.idx, i32 2, i32 0
```

The returned pointer is RC-sentinel: `__rc_retain` and `__rc_release_simple` are
no-ops on it. Callers must not call `_mark_owned` on it.

**Impact:** Eliminates `__rc_alloc` and `free` calls for every `s[i]` operation.
Benchmark: 15 s → 10 s (33 % speedup on the vite bundler workload).

---

## 5. snprintf Format Helper (builtins.impl.jac, na_ir_gen_pass.jac)

**`_get_snprintf_fmt(fmt)`** — cached global for `snprintf` format strings (no
trailing newline). Stored under cache key `"snprintf:" + fmt` so it doesn't
collide with printf format strings. Used by f-strings, `str(x)`, `repr(x)`, and
numeric-to-string conversions.

---

## 6. String Concat and f-string (types.impl.jac, builtins.impl.jac)

- **String concat (`+`)**: switched from `strlen` to `_emit_jac_str_len` for both
  operands; calls `_store_str_len(buf, total)` after `strcat` so subsequent
  `len()` calls are O(1). Also calls `_emit_set_type_tag(buf, "str")`.
- **f-string**: `snprintf` return value (bytes written) is captured and stored via
  `_store_str_len`, making `len(fstr)` O(1).
- **Multi-string concat** (`_codegen_multistring`): releases owned intermediate
  temporaries after each pair is concatenated.

---

### Dict literal construction

When emitting a `{k: v, …}` dict literal:
1. `__dict_new_*()` is called via `_rc_call_with_ctx` (for RC debug attribution).
2. The result is marked owned (`_mark_owned`) immediately — one mark at the start
   before any `set` calls, and one at the end to cover the literal as a whole.
3. After each `__dict_set_KV` call, any owned key or value temporary is released
   (the set operation retains them, so the caller's reference is no longer needed).

### Dict assignment (`__dict_set_KV`)

`_codegen_dict_set` saves the original value before any type coercion, then after
calling `__dict_set_KV`, releases the original if it was an owned temporary.

### Dict destructor forward-declaration

`__rc_release_dict_K_V` is forward-declared before `__dict_new_*` so the new
function can store its address as the `dtor_fn_ptr` field in the allocation header.
This allows `__rc_release_simple` to call the correct element-release chain when a
dict is freed.

---

## Files Changed

| File | What changed |
|---|---|
| `na_ir_gen_pass.jac` | New constants (`HDR_*`, `RC_SENTINEL`, `HDR_MAGIC*`); new method declarations (`_store_str_len`, `_emit_jac_str_len`, `_get_char_cache_gv`, `_get_snprintf_fmt`, RC ownership helpers) |
| `na_ir_gen_pass.impl/refcount.impl.jac` | 32-byte header in `__rc_alloc`; alloc_id counter; magic-check in retain/release; `_emit_magic_check` helper |
| `na_ir_gen_pass.impl/builtins.impl.jac` | `_store_str_len`; `_emit_jac_str_len`; `_get_char_cache_gv`; `_codegen_string_index` (char cache); `_make_global_string` (sentinel header + cached length); `_codegen_fstring` (length caching); `_get_snprintf_fmt` |
| `na_ir_gen_pass.impl/types.impl.jac` | String concat: `_emit_jac_str_len` + `_store_str_len` + `_mark_owned`; tuple boxing ownership |
| `na_ir_gen_pass.impl/dicts.impl.jac` | Dict literal RC ownership; `_codegen_dict_set` owned-temp release; dtor forward-declaration |
| `na_ir_gen_pass.impl/primitives_native.impl.jac` | `chr()` calls `_store_str_len(buf, 1)`; `str(x)` / `repr(x)` call `_store_str_len` with snprintf return |


marsninja:
Closing as superseded. The RC header redesign at the core of this PR (32-byte header, alloc_id/HDR_MAGIC, RC_SENTINEL) landed on main via #5736, and the remaining optimization work has been split into the reviewable pr/1–pr/8 stack (#7846, #7847, #7849, #7850, #7851, #7852, #7853, #7854), which is current against main and covered by tests. Thanks for the underlying work — the stack is the right vehicle for landing it.

PR #6365, opened 2026-06-02
fix(jac-desktop): run desktop dev-mode tests in CI via WebKit + Xvfb, with has_display fallback (#6364)

### LINKED ISSUE #6364: Flaky test: `desktop dev mode starts vite server` times out in headless CI (missing has_display guard)
## Summary

The jac-desktop integration test `desktop dev mode starts vite server` ([`jac-desktop/jac_desktop/tests/test_it_desktop.jac:319`](https://github.com/jaseci-labs/jaseci/blob/main/jac-desktop/jac_desktop/tests/test_it_desktop.jac#L319)) intermittently fails in CI with an opaque timeout:

```
TimeoutError: Timed out waiting for 127.0.0.1:5173
  _helpers.jac:94 in wait_for_port
```

## Root cause

The failure is **not** random — it's a deterministic headless-CI dependency that surfaces as flakiness due to a teardown race.

`jac start main.jac --client desktop --dev` does two things:
1. Starts the Vite dev server (this succeeds — `✔ Vite dev server ready` on port 5173).
2. Launches the PyTauri GUI shell (`python app.py`).

In headless CI the GUI shell crashes on import because the WebKit system library is absent:

```
File ".../src-pytauri/app.py", line 21, in <module>
    from pytauri import Commands
  ...
ImportError: libwebkit2gtk-4.1.so.0: cannot open shared object file: No such file or directory
```

When the shell dies, the dev orchestrator tears down **all** dev processes, including Vite:

```
Stopping dev processes...
✔ Dev processes stopped
```

So by the time `wait_for_port("127.0.0.1", 5173)` polls, Vite has already been torn down → timeout. Whether it presents as pass/fail depends on timing, hence the "flaky" appearance.

## Why the sibling test doesn't hit this

The equivalent jac-client test `desktop dev mode starts` ([`jac-client/jac_client/tests/test_it_desktop.jac:571`](https://github.com/jaseci-labs/jaseci/blob/main/jac-client/jac_client/tests/test_it_desktop.jac#L571)) guards with:

```jac
if not has_display() {
    pytest.skip("No display available for GUI tests");
}
```

`has_display()` returns `False` under CI with no `DISPLAY`/`WAYLAND_DISPLAY`, so the GUI dev test is skipped. The jac-desktop copy is **missing this guard**.

## Proposed fix

1. Add the `has_display()` skip guard to the jac-desktop `desktop dev mode starts vite server` test, matching its jac-client sibling — the full `--dev` flow launches a WebKit GUI shell and genuinely requires a display.
2. Harden `wait_for_port`'s failure path in this test to detect an early dev-process exit and skip/fail with the captured reason (mirroring the sidecar test at `test_it_desktop.jac:277`) so future environment regressions produce an actionable message instead of an opaque `TimeoutError`.

## Reproduction

Headless Linux without `libwebkit2gtk-4.1.so.0` installed:

```
pytest -x jac-client jac-desktop -n auto
```

akindu-k:
## Summary

Fixes the intermittently-failing jac-desktop integration test `desktop dev mode starts vite server`, which timed out in headless CI:

```
TimeoutError: Timed out waiting for 127.0.0.1:5173
```

This PR makes three related changes:

1. **Provisions the desktop GUI runtime in CI** (`libwebkit2gtk-4.1-0` + `xvfb`) and runs the client/desktop suite under `xvfb-run`, so the desktop dev-mode tests **actually execute** in CI instead of crashing.
2. **Adds a `has_display()` guard** to the test as a graceful fallback for environments where no display is available (local headless runs, or any runner without the GUI runtime).
3. **Fixes a latent port-mismatch bug** (exposed once the tests actually ran) by pinning each desktop dev test to its own free port.

Fixes #6364.

## Root cause

The failure looked flaky but was a deterministic headless dependency surfacing as a teardown race.

`jac start main.jac --client desktop --dev` does two things:
1. Starts the Vite dev server — this **succeeds**.
2. Launches the PyTauri GUI shell (`python app.py`).

In headless CI the GUI shell crashed on import because the WebKit system library was absent:

```
File ".../src-pytauri/app.py", line 21, in <module>
    from pytauri import Commands
  ...
ImportError: libwebkit2gtk-4.1.so.0: cannot open shared object file: No such file or directory
```

Crucially, the dev orchestrator binds Vite's lifetime to the GUI shell — it runs `py_proc.wait()` and then tears down **all** dev processes (Vite included) in `finally`. So when the shell crashed, Vite was torn down with it, and the test's `wait_for_port("127.0.0.1", 5173)` timed out. Whether it presented as pass/fail depended on timing — hence the "flaky" appearance.

Two missing pieces caused the shell crash in CI:
- The **WebKit library** (`libwebkit2gtk-4.1.so.0`) that PyTauri links against was not installed.
- No **display** was available for the GUI window to open on.

### 1. Make the tests run for real in CI

Install the GUI runtime and run the suite under a virtual display:

```yaml
- name: Install desktop GUI runtime (WebKit + Xvfb)
  run: |
    sudo apt-get update
    sudo apt-get install -y libwebkit2gtk-4.1-0 xvfb

- name: Run all client tests (including E2E in parallel)
  run: xvfb-run -a pytest -x jac-client jac-desktop -n auto
```

With the WebKit library present, `import pytauri` succeeds; with `xvfb-run` providing `DISPLAY`, the PyTauri window can open. The desktop dev-mode tests now exercise the full flow instead of failing.

### 2. Keep a graceful fallback

Add the same `has_display()` guard the jac-client sibling test (`desktop dev mode starts`) already uses:

```jac
if not has_display() {
    pytest.skip("No display available for GUI tests");
}
```

This is **not** what makes the test run in CI — the GUI runtime above is. The guard is the safety net: in any environment without a display (local headless dev, or a runner missing the GUI deps) the test skips cleanly instead of crashing. Once `xvfb-run` sets `DISPLAY`, `has_display()` returns `True` and the guard is a no-op, so the two changes stack rather than conflict.

### 3. Fix the port-mismatch bug exposed by running for real

Once the tests actually executed in CI (rather than skipping), the jac-client sibling test `desktop dev mode starts` failed — it had a **latent bug** that was hidden for as long as the test was display-skipped: it polled port `5173`, but `jac start --dev` binds Vite to the dev command's default port (`8000`), so the wait could never succeed.

`DEFAULT_VITE_PORT` (5173) is effectively dead code in `PyTauriDesktopTarget.dev` — it is only used when `port <= 0`, while the parameter defaults to `8000`.

The fix pins each desktop dev test to an explicit **free** port and passes it through `--port`, so the test always polls exactly where Vite is bound:

```jac
vite_port = get_free_port();
dev_process = Popen(
    [*jac_cmd, "start", "main.jac", "--client", "desktop", "--dev", "--port", str(vite_port)],
    cwd=project_dir, env=env,
);
wait_for_port("127.0.0.1", vite_port, timeout=90.0);
```

Using a per-test **free** port (rather than a fixed `5173`) also prevents the jac-client and jac-desktop dev tests from colliding on the same port under `pytest -n auto`. `get_free_port()` was already imported and used in both test files, so this matches existing patterns.

Production code is intentionally **left unchanged** — the dead `DEFAULT_VITE_PORT` default in `PyTauriDesktopTarget.dev` is a separate concern and tracked separately.

## Validation

A temporary CI step ran both desktop dev tests **10 times in a row** under WebKit + Xvfb:

- **10/10 iterations passed** (20/20 individual test executions).
- Consistent runtime (28.6s–29.2s, ~0.6s spread) — no timeouts, no port races, no GUI-launch failures.

This confirms the GUI-under-Xvfb path is deterministic and not flaky. The full `test-client` job went from `182 passed, 1 skipped (1 failed)` to **`185 passed`** with the desktop GUI tests now included. The temporary probe step has been removed; the desktop dev tests run once per CI run as part of the normal suite.

## Changed files

- `.github/workflows/test-jaseci.yml` — install `libwebkit2gtk-4.1-0` + `xvfb`; run the client/desktop suite under `xvfb-run`.
- `jac-desktop/jac_desktop/tests/test_it_desktop.jac` — add `has_display()` guard; use a free Vite port.
- `jac-client/jac_client/tests/test_it_desktop.jac` — pin the Vite dev server to a free port and pass `--port` (fixes the 5173-vs-8000 mismatch).


akindu-k:
occurrence: https://github.com/jaseci-labs/jaseci/actions/runs/26852878117/job/79188951951?pr=6294

akindu-k:
## Flakiness probe results — desktop dev tests run 10× under WebKit + Xvfb ✅

To validate that running the desktop GUI dev-mode tests in CI (under the new `libwebkit2gtk-4.1-0` + `xvfb` setup) is reliable and not flaky, I added a temporary CI step that runs both desktop dev tests **10 times in a row** and reports a pass/fail tally.

**Tests exercised each iteration:**
- `jac-client/jac_client/tests/test_it_desktop.jac::desktop dev mode starts`
- `jac-desktop/jac_desktop/tests/test_it_desktop.jac::desktop dev mode starts vite server`

**CI run:** https://github.com/jaseci-labs/jaseci/actions/runs/26866034739/job/79229807857

### Result: 10/10 passed (20/20 individual test executions)

| Iteration | Result | Duration |
|:---:|:---:|:---:|
| 1 | ✅ 2 passed | 29.04s |
| 2 | ✅ 2 passed | 29.01s |
| 3 | ✅ 2 passed | 28.80s |
| 4 | ✅ 2 passed | 28.67s |
| 5 | ✅ 2 passed | 28.88s |
| 6 | ✅ 2 passed | 28.95s |
| 7 | ✅ 2 passed | 29.21s |
| 8 | ✅ 2 passed | 29.10s |
| 9 | ✅ 2 passed | 28.88s |
| 10 | ✅ 2 passed | 28.60s |

**Every iteration passed, with remarkably consistent runtime (28.6s–29.2s, ~0.6s spread).** No timeouts, no port races, no GUI-launch failures across 10 consecutive runs.

### Conclusion

The desktop dev-mode tests are **stable under WebKit + Xvfb in CI** — no evidence of flakiness, and the tight, consistent timing indicates the GUI launch + Vite startup path is deterministic in this environment. This confirms it's safe to rely on these tests running for real in CI rather than being skipped.

PR #6378, opened 2026-06-03
Type the native IR-gen pass against llvmlite (stub + container types)

SandeepaHWP:
The native LLVM IR-gen pass is written against `llvmlite.ir`, which ships no type information. With nothing to go on, the type checker fell back to llvmlite's unannotated source and inferred `object` (or "missing attribute") for the core API the pass leans on — `Value.type`, `PointerType.pointee`, `Type.as_pointer()`, and every `IRBuilder` instruction builder. That produced well over a thousand false positives across the pass.

This PR makes the pass type-check against llvmlite, in two complementary pieces.

## 1. A hand-written `llvmlite.ir` stub + resolver wiring

- **Resolver:** module resolution now searches a jac-owned `jaclang/vendor/stubs` directory ahead of importlib, so hand-written stubs for untyped third-party packages take precedence over their unannotated source. (`vendor/typeshed` is a git submodule and can't host local stubs, hence a separate jac-owned dir.)
- **Stub:** models the type / value / instruction hierarchy and the `IRBuilder`. Builder methods take `Any` operands and return the common `Value` base — matching how the pass uses them, and avoiding spurious clashes when one local holds results of different instruction kinds. `phi`, `alloca`, and the block accessors return their concrete types where callers use members those classes provide (`add_incoming`, the `AllocaInstr` field type, `Block.function`, …).
- **Pass spots exposed by precise typing:** a handful of places relied on llvmlite being untyped — reading a subclass-only attribute (`.width` / `.name`) off a base `Type`, or reusing one local for two unrelated types. Those are narrowed / retyped with no change in behavior.

## 2. Type the pass's LLVM type containers as `ir.Type`

The pass tracked its struct types in `object`-typed fields — the `struct_types` / `list_types` / `dict_types` / `set_types` / `tuple_types` registries as `dict[str, object]`, and the `complex_struct_type` / `range_struct_type` / `slice_struct_type` / `jacval_type` singletons as `object`. These only ever hold `ir.Type` (struct) values, so they're now typed precisely (`dict[str, ir.Type]` and `ir.IdentifiedStructType`). Pure annotation change; it depends on the stub above for the precise `.as_pointer()` / `.set_body()` members.

## Impact

- Pass-wide: **~3595 → ~2080 checker errors (~1500 cleared)**. Every file in the pass improves; none regresses (verified with a shift-invariant per-file count comparison).
- The native fixtures still compile, JIT, and execute with correct results.
- Added a type-check test (`jac/tests/compiler/test_llvmlite_stub.jac`) that exercises the stubbed API surface — including the `dict[str, ir.Type]` container pattern — and fails (~7 errors) without the stub.

The stub is intentionally partial — it covers the API surface the native pass uses today and is meant to be extended as the pass touches more of llvmlite.


SandeepaHWP:
Closing since this is not priority one for now

PR #6381, opened 2026-06-03
Type the native pass's LLVM type containers as ir.Type

SandeepaHWP:
## What this does

The native IR-gen pass kept its LLVM struct types in `object`-typed fields:

- per-kind registries — `struct_types`, `list_types`, `dict_types`, `set_types`, `tuple_types` — as `dict[str, object]`
- singleton struct handles — `complex_struct_type`, `range_struct_type`, `slice_struct_type`, `jacval_type` — as `object`

So every lookup came back as `object`, and the routine `.as_pointer()` / `.set_body()` calls on them were flagged as attribute errors by the type checker.

## Why `ir.Type` / `ir.IdentifiedStructType` is correct

These fields only ever hold llvmlite types:

- the registries hold `ir.LiteralStructType(...)` and `context.get_identified_type(...)` results — both subclasses of `ir.Type` — and the only member accessed on a lookup is `.as_pointer()` (on `ir.Type`). So `dict[str, ir.Type]` is the tightest accurate annotation.
- the singletons are always built via `context.get_identified_type(...)` → `IdentifiedStructType`, and are used with `.as_pointer()` **and** `.set_body()`, so `ir.IdentifiedStructType` is the right type.

No widening, no over-narrowing, no new abstractions — just the real types these fields have always held.

## Impact

- Pure annotation change; **no runtime effect**. The native fixtures still compile, JIT, and execute correctly.
- Clears ~110 checker errors across the native pass with **no regressions** (verified per-file).

## Note

The payoff is fullest on top of the llvmlite stub (#6378): that is what gives `ir.Type` / `ir.IdentifiedStructType` their precise members. This annotation is correct on its own and composes with that change.


SandeepaHWP:
Folded into #6378 — the container-typing change depends on the llvmlite stub there, and it's only 9 lines, so it's cleaner as one PR than two dependent ones.

PR #6428, opened 2026-06-05
feat(jac-scale): add yappi profiler admin endpoints

Gorgeous-Patrick:
## Summary
- Add `/admin/profiler/start`, `/admin/profiler/stop`, `/admin/profiler/status` endpoints for on-demand CPU/wall profiling of the running Jac server via yappi
- Stop endpoint saves both `.prof` (pstat-compatible) and `.txt` (human-readable) files, and returns a top-30 summary inline
- Wired into both monolith `JacAPIServer` and `MicroserviceGateway`
- yappi is an optional dependency — endpoints return a clear error if not installed

## Test plan
- [ ] Start server with yappi installed, hit `/admin/profiler/start`, run a walker, hit `/admin/profiler/stop`, verify `.prof` and `.txt` files are written
- [ ] Verify `/admin/profiler/status` returns `running: true/false` correctly
- [ ] Verify all three endpoints return 403 without admin auth
- [ ] Verify graceful error when yappi is not installed


udithishanka (commented):
A few architectural things worth addressing beyond the bot's notes.

The big one: start and stop will land on different pods. jac-scale is horizontally scaled and yappi state is per process. POST /admin/profiler/start hits one arbitrary replica behind the LB and starts yappi there, then the later POST /admin/profiler/stop almost certainly hits a different replica and returns "Profiler is not running", while the pod that's actually profiling never gets stopped. Same failure inside a single pod if uvicorn runs more than one worker. In microservice mode it's worse, since the gateway mixin only profiles the gateway's own routing process and never the builder-sv / jac-coder-sv pods doing the real work. As written this only really works in single-replica, single-worker dev, and that caveat isn't called out anywhere in the PR.

Second, the profile artifacts get written to the pod's ephemeral filesystem. stop saves the .prof and .txt to output_dir (default cwd or JAC_PROFILE_DIR), which on K8s is the container's writable layer, gone on the next restart or redeploy and unreachable without a kubectl exec/cp. The existing admin modules (logs, traces) deliberately never write local files, they proxy Loki/Tempo and return data over HTTP. The .prof is the genuinely useful artifact here and right now it's stranded on disk. Either stream it back in the response or push it to object storage rather than the local pod.

Third, there's no auto-stop or max-duration guard. Start an open-ended global profile on a live pod and forget to stop it, and that pod runs degraded (yappi wall profiling adds per-call overhead across all threads) until it restarts. A max-duration auto-stop would make this safe to run against anything serving real traffic.

fwiw jac-scale already deploys Prometheus + Grafana + Loki + Tempo per namespace, so for the function-level CPU answer this is reaching for, Pyroscope wired into the existing MonitoringDeployer would solve all three of these structurally (works across replicas, survives restarts, no forgotten-profiler degradation).

Gorgeous-Patrick:
@udithishanka My plan is that:
1. I will stop working on this branch.
2. I will change the endpoint so that the profiler can be enabled by a query param `?profile=XXX` to enable profiling.
3. This would require some better authentication method (right?) Because we do not want to leak the prof info to public.
4. We can stream the profiling data back to user in the HTTP response. 
What do you think? Is that a better idea?

PR #6432, opened 2026-06-06
feat(native): structural match, nested closures, isinstance/type, container & complex fixes

marsninja:
## Summary

Expands the `.na.jac` native (LLVM) backend's language coverage and fixes two real codegen bugs. Everything is test-first: each improvement has a behavioral test that failed before the change.

### Pattern matching (§ structural `match`)
Rewrites `_codegen_match` from a single-`i1` model to a **short-circuiting control-flow matcher** (`_emit_match_test`) and adds structural patterns:
- Sequence `case [a, b]`, star `case [first, *mid, last]` (star binds a fresh sub-list)
- Mapping `case {"k": v}`
- Class/archetype `case Point(x=0, y=ay)` (positional via field order + keyword), inheritance-aware
- Element/key/field access is gated behind its length/contains/type guard, so an out-of-shape subject never indexes out of bounds.

### Nested-`def` closures
Nested functions may now capture enclosing-function locals — lifted to hidden trailing params and injected **by name** at the call site, so a capturing nested function can also recurse.

### Introspection
- `isinstance(x, T)` — inheritance-aware via the runtime type tag for archetypes (statically folded when the declared type already settles it); static LLVM-type test for primitives/containers.
- `type(x)` — returns the type-name string (native has no runtime type objects); added a native-only `type -> str` stub.

### Diagnostics
New `E5091` for builtins that can **never** lower in a native binary (`eval`/`exec`/`compile`/`globals`/`locals`/`vars`/`dir`/`getattr`/`setattr`/`hasattr`/`delattr`/`__import__`) — distinct from the "not yet" `E5090`. This also **fixes a latent silent dropped-call**: `getattr`/`vars`/`dir` are registry builtins that previously fell through with no diagnostic at all.

### Bug fixes
- **`dict.pop`** emitted a malformed PHI — the not-found path's runtime-raise splits the block, so the recorded predecessor was wrong. Now records the real one.
- **`complex`** was broken end to end: the `complex` annotation fell through to `i64`, so the boxed pointer was `ptrtoint`'d and `.real`/`.imag` read `0.0`. Now maps `complex -> JacComplex*` and lowers `.real`/`.imag`; construction, `.real`/`.imag`, and `+ - *` all work.

### Docs
Refreshes `jac-native.md`: the `math` stdlib is now lowered (the old "`import math; math.pi` silently breaks" note was stale), unsupported builtins fail loud rather than silent, and the newly supported features are documented.

## Testing
New fixtures + behavioral tests for every item above, plus retargeted guardrails (structural-pattern and nested-closure guardrails replaced with behavioral tests; new `E5091` guardrails). Full native suite green:
- `test_native_gen_pass` + `test_native_feature_guardrail` + `test_native_iterators`: **426 passed, 3 skipped**
- `test_native_lexer` + `test_wasm_linker`: **26 passed**
- Regular (non-native) compile + type-check unaffected.

## Out of scope (deliberately left as honest, loud gaps)
`eval`/`exec`/… (now `E5091`), generators/`yield`, BigInt/Decimal, and `bytes(...)` construction (`E5090`) — none silently miscompile. `async`/`await`/`flow`/`wait` still run sequentially (a pre-existing no-op), noted for a follow-up.


marsninja:
Deep review notes from this branch review:

I would not call this production-ready yet. The direction is useful, and the selected tests pass, but there are ownership and semantic-design risks that should be fixed before this becomes a long-term native-backend foundation.

**Findings**

- **High: Match captures of pointer values are stored as owned locals without a retain.** `_bind_match_capture` stores `subj` directly into `local_vars`, and scope cleanup later releases all pointer locals. For `case s:` on a `str`/object/list parameter, or `[s]` from `list[str]`, this releases a borrowed alias and can double-free or prematurely free caller/container-owned data. Capture binding should follow normal assignment ownership rules: retain borrowed pointer captures, or explicitly track borrowed aliases so cleanup skips them.

- **High: `match` subject temporaries are not registered for cleanup.** `_codegen_match` evaluates the subject once and stores it in `match.subj`, but that alloca is not in `local_vars`. If the subject expression returns an owned pointer, such as `match make_list()`, the owned subject can leak when no capture takes ownership. This needs explicit cleanup or ownership transfer handling.

- **High: `isinstance(x, SomeObj)` can read an RC type tag from arbitrary pointer-typed non-object values.** `_emit_isinstance_check` only returns false for non-pointer scalars. A `str`, raw C pointer, list, dict, or other pointer can flow to `_read_type_tag`, which blindly reads header metadata at an offset. That is unsafe unless the pointer is known RC-managed and object-like. For production, reject or statically return false for non-struct pointer types before reading headers.

- **Medium: Closure capture discovery is too syntactic for production semantics.** `_find_free_vars` walks every `Name` node and captures any enclosing local. It does not distinguish reads from assignment targets, local declarations, nested scopes, type annotations, attribute/member names, or shadowing. This can over-capture and miscompile cases where an inner function declares or assigns a local with the same name as an outer local. This should use symbol/scope information rather than raw AST-name walking.

- **Medium: The closure implementation is name-keyed rather than symbol/function-keyed.** `lambda_captures` is indexed by simple function name and injected by simple callee name. The save/restore reduces leakage, but long term this is fragile around shadowing, duplicate nested names, function-valued locals, and richer scope behavior. Prefer associating captures with the resolved callee symbol/LLVM function.

- **Medium: The new `type(x)` is intentionally not Python-like, but the API name invites Python expectations.** It returns a static string, not a runtime type object, and for base-typed object values it reports the declared type rather than necessarily the concrete allocation tag. That can be acceptable for native, but docs/tests should make this limitation explicit and include a base-typed-subclass case.

**What looks sound**

The overall lowering shape for structural match is reasonable: guard before indexing, short-circuit into fail blocks, and bind captures before guards. The `dict.pop` PHI fix is also a clear localized correction. The E5091 split for impossible builtins is a good long-term improvement because it separates “native can never do this” from “not implemented yet”.

**Verification run locally**

```bash
uv run --with pytest --with pytest-xdist --with pytest-cov pytest tests/compiler/passes/native/test_native_gen_pass.jac -k 'match or closure or isinstance or type or methods'
uv run --with pytest --with pytest-xdist --with pytest-cov pytest tests/compiler/passes/native/test_native_feature_guardrail.jac -k 'guardrail'
```

Results: `51 passed` for the selected native behavior tests, and `13 passed` for guardrails.

Bottom line: promising feature branch, but I would ask for revisions before production merge. Fix pointer ownership for match captures and subject temps first; then harden `isinstance` pointer classification and replace closure free-var detection with scope-aware analysis.

PR #6448, opened 2026-06-06
Fix bytes + str type checking

### LINKED ISSUE #6446: Type checker accepts `bytes + str` (should error , regression from #6429)
### Description

The type checker accepts `bytes + str`, which is a real type error (`b"x" + "y"` raises `TypeError` in Python). The expression is silently typed as `bytes` instead of being rejected. This is a soundness regression: code that fails at runtime now passes the checker.

### Reproduction

```jac
def bytes_plus_str(b: bytes, s: str) -> bytes {
    return b + s;          # accepted today; should be a type error
}
```

`jac check` reports no error for `b + s`.

### The asymmetry that proves it is a bug

Only the `bytes`-on-the-left direction is broken. The reverse is still correctly rejected:

```jac
def str_plus_bytes(b: bytes, s: str) -> str {
    return s + b;          # correctly errors: E1055 No matching overload for __add__
}
```

`a + b` resolves through the left operand's magic method (`type(a).__add__(b)`):

| Expression | Resolves via | Result |
|---|---|---|
| `b + s` (bytes + str) | `bytes.__add__(str)` | wrongly accepted, typed `bytes` |
| `s + b` (str + bytes) | `str.__add__(bytes)` | correctly rejected (E1055) |
| `b + c` (bytes + bytes) | `bytes.__add__(bytes)` | accepted (correct) |

The return-type annotation is irrelevant to the bug. With `-> str`, `b + s` reports only `E1002: Cannot return bytes, expected str` and still no `E1055`, confirming the `__add__` call itself succeeds and yields `bytes`.

### Impact

- False negatives: real `bytes + str` type errors are no longer caught.

### Expected behaviour

`bytes + str` should be rejected with `E1055`, matching the existing `str + bytes` behaviour and Python semantics.

RagavRida:
## Summary
- Resolve chained imported symbols so stdlib aliases like `_typeshed.ReadableBuffer` reach the real `collections.abc.Buffer` protocol.
- Allow star-import re-export resolution for already-resolved `.pyi` paths such as `collections/abc.pyi -> _collections_abc.pyi`.
- Add a regression for #6446 proving `bytes + str` and `str + bytes` reject while `bytes + bytes` remains valid.

Fixes #6446

## Tests
- `PYTHONPATH=/Users/raghavendramachikatla/jaseci/jac python3 -m jaclang test jac/tests/compiler/passes/main/test_checker_bugs.jac --test_name bug_6446_bytes_str_add_rejected --verbose`
- `PYTHONPATH=/Users/raghavendramachikatla/jaseci/jac python3 -m jaclang test jac/tests/compiler/passes/main/test_checker_bugs.jac --verbose`
- `PYTHONPATH=/Users/raghavendramachikatla/jaseci/jac python3 -m jaclang test jac/tests/compiler/passes/main/test_sym_tab_build_pass.jac --verbose`
- `git diff --check`

### DIFF
--- a/docs/docs/community/release_notes/unreleased/jaclang/6446.bugfix.md
+++ b/docs/docs/community/release_notes/unreleased/jaclang/6446.bugfix.md
@@ -1,0 +1,1 @@
+- **Fix: Reject mixed bytes and string addition**: Resolve chained imported type aliases in stdlib stubs so `bytes.__add__` checks `Buffer` instead of accepting any argument.
--- a/jac/jaclang/compiler/type_system/type_evaluator.impl/evaluator_util_methods.impl.jac
+++ b/jac/jaclang/compiler/type_system/type_evaluator.impl/evaluator_util_methods.impl.jac
@@ -827,25 +827,70 @@ impl TypeEvaluator._set_symbol_to_expr(expr: uni.Expr, sym: uni.Symbol) -> TypeB
 
 """Resolve the imported symbols to the actual symbol."""
 impl TypeEvaluator.resolve_imported_symbols(sym: uni.Symbol) -> uni.Symbol {
-    if (
-        isinstance(sym.decl, uni.Name) and sym.decl.find_parent_of_type(uni.ModuleItem)
+    resolved = sym;
+    seen: set[int] = set();
+    while (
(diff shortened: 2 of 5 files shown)


marsninja:

PR #6452, opened 2026-06-06
fix(type-checker): resolve version-guarded names through star re-exports

### LINKED ISSUE #6446: Type checker accepts `bytes + str` (should error , regression from #6429)

SandeepaHWP:
### Summary

The type checker accepted `bytes + str` (and `bytes + int`, `bytes + list`), which are real type errors. The root cause is that version-guarded names re-exported through `from M import *` failed to resolve, which made `bytes.__add__`'s parameter type collapse to `Unknown`.

Fixes #6446.

### Root cause

`bytes.__add__(value: ReadableBuffer)` and `ReadableBuffer = Buffer`. `Buffer` is declared in `_collections_abc` under `if sys.version_info >= (3, 12):` and re-exported by `collections.abc` via `from _collections_abc import *`.

Currently version-guarded declarations live in their own `IfStmt` child scope instead of being folded into the module's `names_in_scope`. The star re-export path (`_resolve_star_import_symbol`) only followed relative star imports and skipped absolute single-segment sources, so it never descended into `_collections_abc`'s guard branches. `Buffer` therefore failed to resolve (`W1101`), `ReadableBuffer` became `Unknown`, and the `bytes.__add__` overload accepted any operand.

### Fix

`_resolve_star_import_symbol` now resolves absolute single-segment star sources via the singular path resolver. The loaded source module is looked up through `collect_member_syms`, which descends the live version-guard branches, so guarded re-exported names (e.g. `Buffer`) resolve.

### Result

| Expression | Before | After |
|---|---|---|
| `bytes + str` | accepted (wrong) | E1055 (correct) |
| `bytes + int` | accepted (wrong) | E1055 (correct) |
| `bytes + bytes` | accepted | accepted |
| `bytes + bytearray` | accepted | accepted |
| `str + bytes` | E1055 | E1055 |

Previously :
image

image


kugesan1105:
change the test files to .jac

Issue #6373, opened 2026-06-03
jac-desktop release fails: pytauri-wheel has no Python 3.14 wheel, source build needs missing GTK/GLib system libs


kugesan1105:
## Summary

The first `jac-desktop` PyPI release (via `release-desktop.yml`) **fails during `jac bundle --precompile`**. Python 3.12 and 3.13 precompile fine, but **3.14 fails** and aborts the whole release before `twine upload`. Nothing publishes.

Failed run: Publish/Release `jac-desktop` (job "Build and publish"), exits in ~1m32s.

## Root cause

`jac bundle --precompile` precompiles bytecode for every Python on PATH (3.12, 3.13, 3.14 are all set up). For each version it installs jac-desktop's deps, including **`pytauri-wheel`** (the Tauri/Rust desktop runtime).

- **3.12 / 3.13** → `pytauri-wheel` has prebuilt wheels → install succeeds → ✅
- **3.14** → **`pytauri-wheel` ships no cp314 wheel** (PyPI has cp39–cp313 only) → pip falls back to building it **from Rust source** → the Tauri/WebKitGTK crates need system GUI dev libraries via `pkg-config`, which are **not installed on the runner**:

```
error: failed to run custom build command for `gobject-sys v0.18.0`
error: failed to run custom build command for `glib-sys v0.18.1`
  pkg-config exited with status code 1
  The system library `glib-2.0`   required by crate `glib-sys`   was not found.
  The system library `gobject-2.0` required by crate `gobject-sys` was not found.
  (PKG_CONFIG_PATH only contained /opt/hostedtoolcache/Python/3.14.2/x64/lib/pkgconfig)
```

The precompiler treats **any** failed version as fatal (`Precompilation incomplete — 1/3 version(s) failed`), so the single 3.14 failure kills the release even though 3.12/3.13 succeeded.

### Why only jac-desktop
Other plugins (jaclang, jac-client, byllm, jac-scale, jac-super, jac-mcp) are `py3-none-any` with pure-Python deps that have 3.14 wheels, so their 3.14 precompile never triggers a source build. jac-desktop is the only package with a native Rust dependency (`pytauri-wheel`) that lacks a 3.14 wheel.

It is **not** OOM and **not** a timeout — it fails fast at the `pkg-config` lookup for missing system libs.

### Option A — Drop Python 3.14 from jac-desktop precompile (recommended)
[...]

PR #6462, opened 2026-06-06
fix(release): build & publish jac-desktop in one combined release (sibling-plugin deps)

kugesan1105:
## Problem

`jac bundle --precompile` builds a throwaway venv and `pip install`s the package's declared deps to generate bytecode. The only escape hatch for a dependency whose new version isn't on PyPI yet is `JAC_PRECOMPILE_LOCAL_INSTALL` — which covers **jaclang only**.

So when a plugin depends on a **sibling plugin** whose new version is being published in the *same* release, precompile tries to pull it from PyPI before it exists:

```
ERROR: Could not find a version that satisfies the requirement jac-client>=0.3.21
  (from versions: ... 0.3.20)
✖ Error: Precompilation incomplete — 2/2 version(s) failed
```

This is exactly what broke the combined release of jaclang + jac-client + jac-desktop (jac-desktop pins `jac-client>=0.3.21`). We worked around it by publishing in **three separate runs** so each sibling was already on PyPI before the next needed it.

### Scope: this is a public tooling gap, not just our pipeline
This is **not** a user-install problem — by the time anyone runs `pip install jac-desktop`, all packages are on PyPI. It bites only at **build/precompile time**, and it affects **any third party** building two interdependent Jac packages together (their `foo-core` + `foo-plugin`), since the only local-install hook was hardcoded to jaclang.

## Fix

Add **`JAC_PRECOMPILE_LOCAL_DEPS`**: a comma-separated list of local Jac-package source dirs. Inside the precompile venv they're installed **before** the PyPI deps, so the sibling's version pin is already satisfied locally and pip never queries PyPI for it.

- Jac plugins have no `pyproject.toml` (jac.toml is the source of truth), so they're installed with **`jac install -e`**, not `pip install`.
- **Unset by default → zero behavior change** for every package that doesn't have a sibling Jac dep.
- The workflow passes the sibling's repo dir via the existing `extra_build_deps` matrix field (jac-desktop's is already `jac-client`).

With this, jac-desktop builds in the same combined release as jac-client — no more 3-separate-runs workaround, and no extra build tier needed.

## Testing

- **Backward compat:** `JAC_PRECOMPILE_LOCAL_DEPS` unset → the new loop is a no-op; existing `JAC_PRECOMPILE_LOCAL_INSTALL` (jaclang) path unchanged.
- **End-to-end (manual):** bumped local jac-client to an unpublished `0.99.0`, `jac install -e`'d it into a fresh venv → `pip install "jac-client>=0.99.0"` then resolves locally (`Requirement already satisfied ... 0.99.0`) instead of failing with `Could not find a version`. Confirmed the exact failing step now succeeds.
- Type-checker: 0 new errors (41 → 41). pre-commit (incl. jac-format) passes.

## Follow-up (not in this PR)
- A dedicated automated test with a two-package fixture (plugin depending on a local sibling) would lock this in; left out here to keep the change focused.
- Publish-time ordering: jac-desktop still depends on jac-client at install time, so for the *publish* step it should come after jac-client (tier ordering). This PR fixes the **build** blocker; publish tiering can be a separate change.

Refs #6373


kugesan1105:
## Follow-up: release docs need updating (separate PR)

This PR changes the release mechanics but intentionally does **not** touch the docs, to keep it focused. The release documentation is now out of date and should be refreshed in a follow-up PR. Known gaps in `CONTRIBUTING.md` ("Release Flow (for maintainers)", ~L148) and wherever else release flow is described:

- **Tier system** — explain the dependency tiers and that publishing is ordered by tier (jaclang → plugins → plugins-that-depend-on-plugins → jaseci meta), now **4 tiers** after jac-desktop:
  - Tier 1: jaclang
  - Tier 2: plugins depending only on jaclang (jac-client, byllm, jac-scale, jac-super, jac-mcp)
  - Tier 3: plugins depending on a tier-2 plugin (jac-desktop → jac-client)
  - Tier 4: jaseci meta-package (depends on everything)
- **Precompilation** — what `jac bundle --precompile` does (per-Python-version bytecode baked into the wheel), why it spins up venvs, and the local-install escape hatches:
  - `JAC_PRECOMPILE_LOCAL_INSTALL` (jaclang from local source)
  - `JAC_PRECOMPILE_LOCAL_DEPS` (sibling plugins from local source — added in this PR)
- **Publishing** — the build-once / publish-by-tier flow, `wait_for_pypi.jac` gating between tiers, `skip-existing: true` for idempotent re-runs.
- **Manual re-publish** — the `workflow_dispatch` trigger on Publish Release (added in #6358) for re-firing a release after a partial failure.
- **Adding a new package** — checklist: register in `scripts/release_utils.jac` (`PACKAGES` + `INTERNAL_DEPS`), pick the correct tier, set `extra_build_deps` if it depends on a sibling plugin, add release-notes path.

cc: this addresses the "is the contributor guide still accurate?" question — it isn't, post-jac-desktop. Tracking as a doc-only follow-up.

PR #6478, opened 2026-06-07
fix(compiler): fix real type errors in type_evaluator, parameter_type_check, and evaluator_util_methods

SandeepaHWP:
## What this does

The Jac type checker type-checks its own source, and it was flagging a handful of genuine type errors in its own code. This PR fixes those so the checker's own files come back clean. There is no behaviour change for users - it is about the checker being honest about its own types.

All of the changes live in the type system:

- `type_evaluator.jac` / `type_evaluator.impl.jac`
- `parameter_type_check.impl.jac`
- `evaluator_util_methods.impl.jac`

## The fixes

- **`validate_arg_types` return type**: the function returns a `bool`, but its declaration said `-> None`. The declaration now matches the implementation.

- **`validate_call_args` could return `None`**: it returned `prefetch.int_class`, which is optional, where a `TypeBase` is required. It now falls back to `UnknownType` when that value is missing.

- **`match_args_to_params` and `get_type_of_magic_method_call` argument types**: both only accepted `list[Expr]`, but every caller passes `expr.params`, which is `Sequence[Expr | KWPair] | None`. The parameter type is widened to match, and the loop over it is guarded against `None`.

- **`len(expr.params)` without a None check**: the `typing.cast` special case called `len(expr.params)` and indexed `[0]` even though `params` can be `None`. It now checks `expr.params` first.

- **`Any` leaking into a `Symbol` parameter**: a few helpers (`get_type_of_member`, `_get_method_with_overloads`, `_find_overload_by_arg_count`, and a `found_member` local) were typed as `any` / untyped, so `member.symbol` came through as `Any`. They are now typed properly (`ClassMember`, `list[Symbol]`), so the values flowing into `get_type_of_symbol` carry a real `Symbol` type.

## Notes

- No behaviour change: these only correct type annotations and add a couple of null guards.
- Any remaining self-check findings in these files are pre-existing and unrelated, and are out of scope for this PR.


SandeepaHWP:
Folded into #6810, which now carries these type-checker fixes alongside the jac0core codegen-pass fixes. Closing in favor of #6810.

PR #6480, opened 2026-06-08
fix(types): keep overload set for imported @overload functions

kugesan1105:
## Summary

Resolving `import from mod { fn }` collapsed an overloaded function to its **first** signature: `_set_module_item_symbol_and_type` set the import item's type from the primary symbol only and never gathered the sibling `@overload`s.

A call matching a *later* overload then failed. Real-world repro (SQLAlchemy):

```jac
import from sqlalchemy { create_engine }
glob eng = create_engine("sqlite:///:memory:");  # E1050: missing required parameter 'future'
```

`create_engine`'s first `@overload` makes `future: Literal[True]` required, so the checker rejected the single-arg call instead of falling through to the `**kwargs` overload.

## Fix

In `_set_module_item_symbol_and_type`, when the imported symbol is a function, gather its sibling overloads from the symbol's source scope (`collect_member_syms(sym.parent_tab, ...)`) and build an `OverloadedType` — mirroring the existing module-attribute path (e.g. `math.floor`).

## Test

`imported_overloaded_function_keeps_overloads` compiles a two-file fixture and asserts the imported function resolves to `OverloadedType` (2 overloads). Red before the fix (bare `FunctionType`), green after.

Earlier broader regression (298 checker/import tests) passed with this change.

## Note

This fixes the `E1050` false positive in the SQLAlchemy repro. A second, larger false positive in the same file — `result.rowcount > 0` (E1110), where `@util.memoized_property` isn't unwrapped via the descriptor `__get__` protocol — is a separate, tracked follow-up.

### DIFF
--- a/docs/docs/community/release_notes/unreleased/jaclang/6480.bugfix.md
+++ b/docs/docs/community/release_notes/unreleased/jaclang/6480.bugfix.md
@@ -1,0 +1,1 @@
+- **Fix: imported overloaded functions keep all overloads**: Calling an `@overload`-ed function imported with `import from mod { fn }` (e.g. SQLAlchemy's `create_engine`) no longer reports a spurious missing-parameter error for calls that match a later overload.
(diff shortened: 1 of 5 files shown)


kugesan1105:
Closing: this targeted the import-item path but does not actually fix the create_engine E1050 false positive (verified: error appears with and without it). The call site collects overloads at the re-export scope, not the definition site. Replaced by a corrected fix that follows the re-export chain to the defining symbol.

PR #6483, opened 2026-06-08
chore(jac-desktop): roll version back to 0.1.1 so the release lands 0.2.0

kugesan1105:
## Why

jac-desktop's `jac.toml` was set to **0.2.0** (the Jac-native rewrite), but **0.2.0 was never published to PyPI** — only `0.1.0` and `0.1.1` are live. If we run a `patch` release from here, jac-desktop bumps `0.2.0 → 0.2.1` and **0.2.0 is skipped on PyPI entirely**.

0.2.0 is the milestone release (full PyTauri → Jac-native rewrite) and should ship under that exact number.

## Change

Roll `jac-desktop/jac.toml` version back to **0.1.1** (the last published version). Then in the upcoming release, give jac-desktop a **minor** bump:

```
0.1.1 --minor--> 0.2.0   ✅ (validated: 0.2.0 is free on PyPI)
0.1.1 --patch--> 0.1.2   (what we're avoiding)
```

So jac-desktop publishes the rewrite as **0.2.0** while the other 7 packages take their normal `patch`.

## No other changes needed
- **Release notes** already have `## jac-desktop 0.2.0 (Latest Release)` → `0.1.1` → `0.1.0`, which stays correct (the release tooling won't touch notes since there are no unreleased fragments).
- **jaseci-package** already pins `jac-desktop>=0.1.1` (it was never moved to 0.2.0 either), satisfied by the upcoming 0.2.0.

## After this merges
Run **Create Release PR** with: jac-desktop = **minor**, everything else = **patch**.

### DIFF
--- a/docs/docs/community/release_notes/jac-desktop.md
+++ b/docs/docs/community/release_notes/jac-desktop.md
@@ -1,27 +1,6 @@
 # jac-desktop Release Notes
 
-## jac-desktop 0.2.0 (Latest Release)
-
-### Breaking: PyTauri/PyInstaller replaced by a Jac-native desktop host
-
-The desktop target is now **Jac-native**: `jac build --client desktop` builds your
-`cl` bundle and compiles a single native host (`jac nacompile`) that embeds the OS
-webview to render it - no Rust toolchain, no PyInstaller sidecar, no separate
-process. Output is one self-contained binary under `.jac/client/desktop/`.
-
-- **Removed**: the PyTauri shell, the PyInstaller sidecar, the `jac desktop plugin`
-  CLI (Tauri plugin catalog), and the `pytauri-wheel`/`anyio`/`pyinstaller`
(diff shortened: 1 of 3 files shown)


kugesan1105:
Also moved the **0.2.0 release notes** out of the main file into a fragment (`unreleased/jac-desktop/6455.breaking.md`):

- The 0.2.0 notes were hand-written directly into `jac-desktop.md` as a `(Latest Release)` section, bypassing the fragment workflow. Since 0.2.0 isn't published and the version is rolled back to 0.1.1, that section was premature.
- The main file's latest section is reset to the actually-published **0.1.1**.
- When **Create Release PR** runs with `jac-desktop=minor`, the assembler folds the fragment into a proper `## jac-desktop 0.2.0 (Latest Release)` / `### Breaking Changes` section.

Verified with `collect_release_notes.jac --package jac-desktop --version 0.2.0 --dry-run` — all original content is preserved and renders correctly.

Issue #6484, opened 2026-06-08
Desktop SSO: pluggable auth transport, loopback state nonce, push handoff


marsninja:
## Context

PR #6485 wires the desktop app's SSO login to the loopback OAuth broker
(system browser, RFC 8252). The *transport* is the correct long-term mechanism
and should stay. This issue tracks turning the MVP-grade *integration* around it
into the proper long-term design.

## 1. Make the auth transport pluggable (layering)

Today the shared client runtime (`@jac/runtime`, used by every browser app)
hardcodes one native host's private protocol: `jacSsoLogin` sniffs
`window.__JAC_DESKTOP__` and knows the literal routes `/__jac/oauth/start` and
`/__jac/session`. That is upside-down -- the generic web runtime should not know
how a specific desktop shell brokers tokens.

Proposed: an abstract auth-transport seam. The runtime calls a transport
interface (e.g. `ssoLogin(platform, operation)`); the default is the web
redirect. The desktop host registers its broker-backed implementation at startup
(it already runs a document-start bootstrap -- the natural injection point).
A second native shell (Tauri, Qt/CEF, mobile) can then implement the same
interface without the web runtime hardcoding anyone's contract.

Acceptance:
- `@jac/runtime` contains no `/__jac/*` literals or `__JAC_DESKTOP__` branch.
- The desktop host registers the transport; web behavior is unchanged.
- The `/__jac/*` contract is documented as the desktop transport's interface.

## 2. Add a state nonce to the loopback callback (security)

The loopback callback has no state/nonce binding: `_oauth_callback` stores
whatever `?token=` arrives and `POST /session` sets any token, on a
deterministic, guessable port. Any *local* process can inject a session token
(login-CSRF / token-fixation). The server validates the redirect is a safe
loopback, but that only stops redirect-to-evil.com, not a forged local callback.

Proposed: `oauth/start` mints a random `state`, keeps it in memory, and the
callback is rejected unless it echoes the matching value (single-use). Bind the
session poll to the same handle so a token can't be read by an unrelated start.

Acceptance:
- A callback with a missing/mismatched `state` is rejected and stores nothing.
- Tests cover the reject path (extend `test_oauth_broker`).

## 3. Replace polling with push (quality)

`__jacPollBrokerSession` polls `/__jac/session` every 1s for up to 5 min because
the webview and the broker have no event channel. Polling localhost is cheap and
robust, but the proper handoff is event-driven: the broker pushes into the
[...]

PR #6485, opened 2026-06-08
feat(desktop): wire frontend SSO login to the system-browser broker

marsninja:
## Summary

The native desktop shell shipped a loopback OAuth broker (#6468) that opens the
user's **system browser** (RFC 8252), but no frontend code called it. Clicking
"Sign in" logged in *inside* the embedded webview, so users had to enter
credentials from scratch and could not reuse their real browser session. This
connects the login button to that broker.

The fix lives in the shared client runtime, so any Jac client app (not just this
one example) gets the right behavior:

- **Desktop shell** (`window.__JAC_DESKTOP__`): `jacSsoLogin()` calls the broker
  `/__jac/oauth/start`, which opens the system browser, then polls
  `/__jac/session` for the token.
- **Web**: redirects to the server's `/sso/<platform>/<operation>` endpoint, as
  before.

## What changed

- **runtime** (`client_runtime`): add `jacSsoLogin()` and `jacSetToken()`; factor
  the duplicated cache-clear out of `jacLogin`/`jacLogout` into `__jacClearCache()`.
- **token key**: the OAuth callback page wrote `auth_token` but the runtime reads
  `jac_token`, so logins silently did not take. The callback now uses
  `jacSetToken()` -> canonical `jac_token`.
- **broker** (`oauth_broker`): add an `operation` (login|register) param so both
  example buttons work through the system browser. Backward-compatible (defaults
  to login).
- **example** (google-auth): complete it so the button is genuinely wired --
  `lib/auth` (AuthProvider/AuthContext delegating to `jacSsoLogin`), the route
  guard, and `styles.css`. Re-include the example's source `lib/` dir, which the
  repo-root Python `.gitignore` was swallowing (likely why this example shipped
  half-broken).

## Test plan

- New runtime SSO cl test (`test_sso_login.cl.jac`, runs under bun): desktop
  broker branch stores `jac_token`, web branch redirects, key reconciliation. 3/3.
- Broker tests: 8/8 (6 original + 2 for the new `operation` param).
- littleX full-stack (live backend) stays green, confirming the
  `jacLogin`/`jacLogout` refactor is behavior-preserving.
- `jac2js` compiles the runtime and all example files cleanly.

Not covered (needs a real OAuth provider + vite/npm SPA build): full
click-through e2e. The logic is locked in by the bun test.

## Follow-up

This is a correct, standard MVP. The proper long-term design (pluggable auth
transport so the web runtime stops hardcoding the desktop broker, a loopback
state nonce, and event-driven token handoff instead of polling) is tracked in
#6484.


jac/jaclang/runtimelib/impl/client_runtime.impl.jac
```diff
@@ -450,32 +450,103 @@ impl jacLogin(
...
+            f"{broker}/oauth/start?platform={encodeURIComponent(platform)}" + f"&operation={encodeURIComponent(
+                operation
+            )}" + f"&api_base={encodeURIComponent(base)}"
+        );
+        started = await fetch(start_url);
+        if not started.ok {
+            console.warn("[jac-client] oauth/start failed:", started.status);
+            return False;
+        }
+        token = await __jacPollBrokerSession(broker, 300000, 1000);
```
MusabMahmoodh, line 555: Nothing clears the broker session before we start polling, and `jacLogout` does not `DELETE /session` either. So if a token is already on disk (a prior login, or a logout that left it behind), the first poll returns the *old* token ~1s in, before the browser round-trip finishes - re-login as a different user silently keeps the old one. Could we clear the session before `oauth/start` and on logout?

jac-desktop/jac_desktop/native/oauth_broker.jac
```diff
@@ -210,8 +218,11 @@ def make_handler(
             q = parse_qs(urlparse(self.path).query);
             platform = (q.get("platform", [""])[0] or "").strip();
             api_base = (q.get("api_base", [""])[0] or "").strip();
+            operation = (q.get("operation", ["login"])[0] or "login").strip();
```
MusabMahmoodh, line 374: Tracked in #6484, but worth weighing for this PR since it makes the flow live: the callback stores any `?token=` with no state binding, and because it is a GET navigation it is reachable cross-origin (not just from a local process) by any page open in the system browser. The state nonce probably should not land *after* this goes live.

jac-desktop/jac_desktop/native/oauth_broker.jac
```diff
@@ -210,8 +218,11 @@ def make_handler(
...
             try {
-                url = build_sso_login_url(api_base, platform, self._callback_url());
+                url = build_sso_login_url(
+                    api_base, platform, self._callback_url(), operation
```
MusabMahmoodh: Separate from the nonce: we trust the `Host` header here with no allow-list, so a DNS-rebinding page could become same-origin and read `/session` to exfil the token. A `Host` check (127.0.0.1/localhost) closes it cheaply.

PR #6514, opened 2026-06-08
fix(interop): relative sv import drops endpoints in dual-role cl→sv modules (#6507)

### LINKED ISSUE #6507: sv-to-sv: a cl->sv endpoint module can't also be a server->server consumer (endpoints silently vanish)
## Summary

A server `.jac` module that is exposed to the browser as a **cl→sv endpoint provider** (its `walker:pub` / `def:pub` symbols are imported by a `.cl.jac` via `sv import from .X`) **cannot also be a server→server consumer** (do its own `sv import from .provider`). The moment such a module gains an outgoing `sv import`, it silently fails to load server-side and **all of its endpoints disappear** — no compile error, no server log.

This blocks the natural pattern of calling a microservice from the same module that already serves the browser — e.g. a guestbook backend that hands its live signature count to a separate `analytics` service.

## Reproduction

A single-`jac start --dev` cl app (the common fullstack shape):

- `main.jac` — `cl { import from .frontend { app } }`
- `frontend.cl.jac` — `sv import from .guestbook { sign_guestbook, get_guestbook }`
- `guestbook.jac` — `walker:pub sign_guestbook {...}`, `walker:pub get_guestbook {...}` → all register ✅
- `analytics.jac` — a provider (`walker:pub Analyze` or `def:pub analyze`)

Add **one line** to `guestbook.jac`:

```jac
sv import from .analytics { Analyze, AnalyticsReport }
```

Restart → 
- `guestbook.jac`'s walkers are **gone** from `/openapi.json` (40 → 34 walker routes; `POST /walker/get_guestbook` → 405).
- **No compile error, no server error.**
- `littleX`'s walkers (a sibling cl→sv module with no outgoing `sv import`) still register fine.

**Contrast — serving the same consumer _directly_ works:** `jac start consumer.jac` auto-spawns the provider (`jac start provider.jac --port <free> --no_client`) and routes the call. Verified end-to-end with distinct PIDs across the boundary, for both walker and `def:pub` providers.

## Root cause (traced)

`sv import from .X` is lowered by `PyastGenPass._generate_sv_to_sv_stubs()` (`jac/jaclang/jac0core/passes/impl/pyast_gen_pass.impl.jac:182`). That generator only emits the HTTP-stub form (plus the `_declare_consumer_provider(__name__, 'X')` edge) **if** `manifest.server_exports_to_server` already contains a binding with `source_module == "X"`. If it does not, it returns `False` and the import **falls through to a plain `from .X import ...`** (`pyast_gen_pass.impl.jac:~1158`) — a normal Python import of a module that is meant to be a *separate service*. That import raises at the consumer module's load time → the module fails to import → its endpoints are silently dropped.

The binding **is** present when the consumer is compiled as the **served/main** module (`BoundaryAnalysisPass.enter_import` → `_register_sv_import` runs over it; manifest populated) — which is why serving the consumer directly works. It is **not** present in the **cl→sv load path**, where the module is pulled in only as an endpoint provider for the browser; its own outgoing `sv import` is not boundary-analyzed into the manifest used at its pyast/load time, so stubs are not generated and the plain-import fallthrough fires.

Relevant code:
- `BoundaryAnalysisPass.enter_import` / `_register_sv_import` — `jac/jaclang/jac0core/passes/impl/boundary_analysis_pass.impl.jac:84`
- `InteropManifest.server_exports_to_server` getter — `jac/jaclang/jac0core/impl/codeinfo.impl.jac:68`
- `_generate_sv_to_sv_stubs` + plain-import fallthrough — `jac/jaclang/jac0core/passes/impl/pyast_gen_pass.impl.jac:182` / `:1158`
- serve-time load of `server_exports_to_client` modules — `jac/jaclang/runtimelib/impl/server.impl.jac`

## Impact

- A common, natural pattern is impossible: call a microservice from a module that also serves the browser. The only workaround is to move the consumer into the served `main.jac` (with circular-import friction against the cl entry) — the kind of "deployment topology inferred from an import keyword" coupling worth avoiding.
- The failure is **silent** — endpoints just vanish, no error. This is a debugging trap.

## Proposed fix

1. **Generate sv-to-sv consumer stubs for cl→sv endpoint modules too.** Ensure a module's *own* outgoing `sv import`s are boundary-analyzed into the manifest used when that module is compiled/loaded in the cl→sv path, so `_generate_sv_to_sv_stubs` fires and the import becomes the HTTP-stub + lazy-spawn form rather than a plain import.
2. **Fail loud.** A server module that fails to load its endpoints must raise (or at minimum warn) instead of silently dropping its walkers — independent of (1).

## Acceptance criteria

- A cl→sv-loaded module (`guestbook.jac`) with an outgoing `sv import from .analytics` keeps **all** of its own endpoints registered, **and** calling one of them that invokes `analyze(...)` lazily spawns `analytics` as a separate process and routes the call (distinct PIDs).
- A regression test covering the dual-role path (the existing `jac-scale/jac_scale/tests/test_eager_spawn.jac` only exercises in-process `TestClient`s, not real subprocess spawning nor the cl→sv dual-role case).
- Module-load failure surfaces as an explicit error.

## Notes / context

Found while building a microservices demo for `this_is_jac`. The sv-to-sv mechanism itself is solid (the runtime literally shells out `subprocess.Popen([jac, "start", provider, "--port", <free>, "--no_client"])` via `ensure_sv_service`, k8s mode resolves a DNS URL instead) — this is specifically about the **dual-role compile/load path** in the cl-app serve mode.

marsninja:
## Summary

Fixes #6507. A server `.jac` module reached by the browser as a **cl→sv endpoint provider** that also does its own **server→server** `sv import from .X` silently lost **all** of its own endpoints — no compile error, no server log. This blocks the natural pattern of calling a microservice from the same module that serves the browser.

## Root cause

The bug is a provider-key mismatch between the two passes that produce and consume the sv-to-sv interop binding, triggered specifically by the **relative** import form (`sv import from .X`):

- `BoundaryAnalysisPass._register_sv_import` recorded `InteropBinding.source_module` using `ModulePath.dot_path_str`, which **keeps** the leading relative dots → `.X`.
- `PyastGenPass._generate_sv_to_sv_stubs` looked the binding up by the **bare** name → `X`.

`.X != X`, so `_generate_sv_to_sv_stubs` found no binding, returned `False`, and the import fell through to a plain `from .X import ...`. That plain relative import crashes the module at load time in the cl→sv load path, and the module's walkers/functions vanish with it.

Every pre-existing sv fixture used **absolute** names (`sv import from inventory_service`), so `dot_path_str` already returned the bare name and the relative form was never exercised — the bug stayed hidden.

## Fix

1. **One canonical provider key.** Add `ModulePath.base_module_name` — the bare dotted module name with no leading relative-level dots — as a getter parallel to `dot_path_str`. Both `_register_sv_import` (producer) and `_generate_sv_to_sv_stubs` (consumer) now read the key from this single source, so they can't drift again. This is the same form the runtime already uses (`JAC_SV_<MOD>_URL`, `ensure_sv_service`, `_declare_consumer_provider`).
2. **Fail loud.** `ModuleIntrospector._collect_cross_boundary_endpoints` now logs a `jaclang.serve` warning when a cross-boundary endpoint's provider module fails to import, instead of silently dropping the endpoint (the issue's "debugging trap").

## Tests

All three were confirmed to **fail before** the fix and pass after:

- `test_typed_interop.jac` — relative `sv import` must lower to the HTTP client stub, not a plain Python import (the precise root-cause regression).
- `test_serve_client.jac` — the cl→sv dual-role module keeps and serves its own walker endpoints when it is also an sv consumer (the reported symptom: `/walkers` was empty).
- `test_serve_client.jac` — a crashing cross-boundary provider module is logged, not silently dropped (the fail-loud behavior).

New fixtures: `sv_relative_*` (compiler), `dualrole_*` and `failload_*` (serve).

### Regression sweep

- interop + serve-client: 112 passed (6 Vite-skips)
- jac-scale sv-to-sv (microservice + eager_spawn): 18 passed
- pyast-gen + JS-gen: 85 passed
- native gen pass: 394 passed

## Known limitation / follow-up

`base_module_name` discards the relative level, so it is only unambiguous together with the consumer's `base_path` — correct for same-directory relative imports (the common case), but a deep relative import (`sv import from ..pkg.mod`) still resolves the bare name against the wrong directory. This is a property of the broader "provider identity = re-derived name string" model, not of this fix. A follow-up RFC proposes promoting the already-resolved provider path to the canonical binding key across compiler + runtime, which would close that gap permanently.


marsninja:
Follow-up RFC for the deeper `provider identity = resolved path` refactor (deep relative `sv import` + eliminating name re-derivation): #6515

Issue #6049, opened 2026-05-20
NA / central TypeEvaluator homogenization: design notes and follow-ups


marsninja:
## Context

PR #PR_PLACEHOLDER homogenizes the type analysis between NaIRGenPass and the central `TypeEvaluator`: NA now consumes `Expr.type` via a new `_lower_type(TypeBase) -> ir.Type` instead of running a parallel ~330-line AST-walking type interpreter. Native fixtures pass 342/342 and the legacy fallback rate dropped from 2025 to 60 across the 80 native fixtures (~97% reduction).

This issue captures the **design decisions** that were locked in to land that refactor safely, and the **scoped follow-ups** that were deliberately *not* done so the refactor stayed focused.

### 1. Implicit native run is inference-only (diagnostic suppression)

`JacCompiler.compile()` now auto-runs the type-check schedule for any `.na.jac` module (and any `native_compat` `.jac`) before codegen, but it **snapshots `prog.errors_had` / `prog.warnings_had` around that schedule and truncates back** when the run was implicit (i.e. `options.type_check=False`).

Rationale: the central `TypeEvaluator` is **stricter than the native dialect**. Running it as enforcement broke 69 of 342 native tests with errors like:

```
Cannot assign float to f64
Type "Piece" has no attribute "color"
Cannot access attribute "get_piece" for type "Board | NoneType"
No matching overload found for method "__add__"
```

These are not bugs in this refactor; they reflect type-checker gaps for the native dialect (see follow-ups below). The original native build path was explicitly designed with `type_check=False` (see [`nacompile.impl.jac`](https://github.com/jaseci-labs/jaseci/blob/main/jac/jaclang/cli/commands/impl/nacompile.impl.jac), comment "native compilation has its own type discipline"). The refactor preserves that contract: native builds keep their own discipline, while gaining the type checker's *inference*. Explicit `jac check` (`options.type_check=True`) continues to surface all diagnostics.

### 2. Re-entrancy guard on the implicit native run

A `_native_typecheck_running` global in `compiler.jac` prevents unbounded recursion: type-checking a native module requires compiling the type-system modules themselves, which in turn would re-fire the implicit branch. The guard is scoped to the implicit branch only — explicit `options.type_check=True` is honored on every module.

### 3. `backend_type_tag` is preserved, not deleted

[...]

PR #6525, opened 2026-06-09
fix(native): coerce sysconfig libpy to str for strict-Any param check

kugesan1105:
## Problem
https://github.com/jaseci-labs/jaseci/actions/runs/27140011566/job/80102699258
The release build failed in `build (jac-desktop, …)` during `jac bundle --precompile`:

```
✖ Error: error[E1053]: Cannot assign Any | str to parameter 'libpy' of type str
  --> jac-desktop/jac_desktop/targets/impl/native_desktop_target.impl.jac:215:9
  213 |     libpy = sysconfig.get_config_var("INSTSONAME") or "libpython3.12.so.1.0";
```

## Root cause

`sysconfig.get_config_var(name: str)` is typed `-> Any` in the stdlib stub, so
`get_config_var("INSTSONAME") or "libpython3.12.so.1.0"` types as `Any | str`.
Passing that to `_generate_host_source`'s `libpy: str` parameter is rejected under
the **strict-Any gradual semantics** (#5859): hand-written Jac modules are *strict*
call sites, so an `Any` leaking across the Python boundary into a concretely-typed
parameter is flagged `E1053`.

This surfaced now because all four conditions coincide only here:
1. `get_config_var` returns `Any` (Python-boundary leak).
2. The call site is a strict Jac module (not Python-derived).
3. `_generate_host_source` declares `libpy: str` concretely.
4. The target generates `host.na.jac` and runs the native type-checker, whose
   diagnostics are **fatal** (#6049) — unlike ordinary `.jac` precompile errors,
   which are caught and counted. jac-desktop is the only plugin with native code.

## Fix

Wrap the expression in `str(...)`, collapsing the boundary `Any` to `str` (a
runtime no-op when it is already a string).

## Verification

In the build's resolution context (sibling deps resolvable, mirroring CI's
`JAC_PRECOMPILE_LOCAL_DEPS`):

| | result |
|---|---|
| before | `1 error` — `E1053 … parameter 'libpy'` at 215:9 (exactly CI's error) |
| after  | `1 passed`, 0 errors |

Full `jac bundle --precompile` of jac-desktop: exit 1 → exit 0.

### DIFF
--- a/jac-desktop/jac_desktop/targets/impl/native_desktop_target.impl.jac
+++ b/jac-desktop/jac_desktop/targets/impl/native_desktop_target.impl.jac
@@ -210,7 +210,7 @@ impl NativeDesktopTarget.build(
     (out_dir / "oauth_broker.py").write_text(broker_res.stdout);
 
     # 5. Generate + compile the native host into the output dir.
-    libpy = sysconfig.get_config_var("INSTSONAME") or "libpython3.12.so.1.0";
+    libpy = str(sysconfig.get_config_var("INSTSONAME") or "libpython3.12.so.1.0");
     host_src = _generate_host_source(
         libpy, win.title or app_name, win.width, win.height, str(app_name)
     );


kugesan1105:
@SandeepaHWP  [Contribution Checks / Contribution Checks (pull_request)](https://github.com/jaseci-labs/jaseci/actions/runs/27177701672/job/80230058556?pr=6525) is passing without release notes addition

PR #6533, opened 2026-06-09
Fix/byllm text answer on chunk

Hushan-10:
## Problem

When a tool-enabled agent with a `-> str` (or `-> None`) return finishes by **writing plain text instead of calling the hidden finish tool**, byLLM delivered that final answer **only** as `thought` events and never emitted it on the `chunk` (answer) channel.

As a result, consumers that read only the answer channel received an **empty** response — the answer was effectively dropped. The behaviour was also asymmetric: structured returns were protected (they re-prompt through finish-tool recovery), but `str`/`None` returns silently accepted the loose text with no answer chunk.

## Root cause

In `_invoke_streaming`, the model's text streams live as `thought` events during the ReAct loop. When a response has no tool calls and recovery is gated off for `str`/`None`, the branch closed out with `steps_done` and **no `chunk`** — so the answer never reached the answer channel.

## Fix

In the no-tool-calls branch, when the model stopped by writing plain text, byLLM now delivers that terminal text on the `chunk` channel.

It reuses the text the model already produced — **no re-prompt, no regeneration, zero extra LLM calls**.

Structured returns and the existing finish-tool path are unchanged.

## Tests

* New deterministic `MockLLM` regression test (`logging_stream_str_skip_finish`): a `-> str` agent whose model skips the finish tool and returns plain text now asserts the answer arrives on the `chunk` channel. Verified to **fail before** the change and **pass after**.
* Full byLLM suite passes; existing streaming / recovery / finish-tool tests are unaffected.

## Before / After

**Scenario:** a tool-enabled `-> str` agent where the model answers in plain text instead of calling the finish tool.

**Before the fix** — the reply only appeared as `thought` events; the `chunk` (answer) channel was empty:

```python
chunk_text = ""
for event in answer("What is the weather in London?"):
    if event.event_type == "chunk":
        chunk_text += event.data.get("content", "")

print(chunk_text)  # ""  ← empty, answer was silently dropped
```

**After the fix** — the terminal text is delivered on the `chunk` channel:

```python
chunk_text = ""
for event in answer("What is the weather in London?"):
    if event.event_type == "chunk":
        chunk_text += event.data.get("content", "")

print(chunk_text)  # "It's sunny in London, about 25°C."
```

No re-prompt and no extra LLM call — the text the model already produced is reused.


MalithaPrabhashana (approved):
LGTM!

PR #6543, opened 2026-06-10
perf(plugin): remove inspect from hook dispatch (codegen facades + single-impl devirtualization)

Jayanaka-98:
## Summary

Every `Jac.*` runtime call (`check_write_access`, `edges_to_nodes`, `get_edges`,
`spawn_call`, `store`, ...) dispatches through a generated proxy in
`generate_plugin_helpers`. That proxy reconstructs its arguments with
`inspect.Signature.bind_partial` + `apply_defaults` on *every* call -- ~73% of a
~1250 ns/call dispatch cost. With both phases, a single-impl hook call drops to
~36 ns (**34x** on a pure-dispatch microbenchmark); a dispatch-bound graph
traversal gets **~2.6x** end to end.

This is core jaclang, so it applies to `jac run`, dev `jac start`, and `--scale`
alike; `--scale` just makes it most visible under per-request fan-out.

Two layers, both gated behind ablation flags (default on):

- **Phase 1 (`JAC_HOOK_CODEGEN`)**: replace the reflective `*args/**kwargs` proxy
  with a real-signature function generated per hook, so CPython binds the
  arguments natively instead of via `inspect`. Behavior-identical; signatures
  with `*args`/`**kwargs`/positional-only keep the reflective path.
- **Phase 2 (`JAC_HOOK_SEAL`)**: at the first `ExecutionContext`, "seal" the
  plugin manager and devirtualize single-impl, simple-signature hooks (75 of 101
  core hooks) to a direct call with no `HookCaller` dispatch. Multi-impl hooks
  stay on the Phase-1 dynamic dispatcher.

Mirrors the approach in #6158 ("drop inspect from dispatch", for OSP ability
triggers); builds on the lightweight plugin system from #4749. cc @marsninja

## Where the seal happens (and why)

`seal()` is called from `ExecutionContext.postinit` (idempotent, lazy), not at
import and not from the CLI commands. This is the universal runtime seam every
Python-runtime launcher passes through (`jac run`/`start`, library import,
desktop/mobile hosts), and it fires only at run time, after all plugins are
loaded and after compilation, so the plugin set is stable.

Under `--scale`, `JScaleExecutionContext.init` overrides core
`ExecutionContext.postinit` (a Jac subclass `init` replaces the parent
`postinit`), so the seal would not fire there. We call `seal()` at the top of
that override too -- idempotent, so the first request seals and the rest no-op.
jac-scale's own multi-impl hooks (e.g. its storage backend) stay on the dynamic
dispatcher.

Two rules make it safe against the compiler's `without_plugins` churn:
- single-impl devirtualization only applies to **core-only** hooks; `without_plugins`
  only unregisters *external* plugins, so it never touches them.
- `register` re-specializes affected hooks; `unregister` deliberately does **not**.
  Multi-impl (external) hooks stay on the dynamic dispatcher, which follows the
  live impl set as plugins drop out and rejoin.

Native/JS emit targets bypass the Python hook system and would need compile-time
devirtualization instead (out of scope here).

## Correctness

This started as a real regression: an earlier version sealed at import and
re-specialized on every register/unregister, which transiently devirtualized
external-plugin hooks (e.g. byllm's `call_llm`/`get_mtir`) to their core stubs
during `without_plugins`, breaking structured output. The current design (seam
at `postinit`, no reseal-on-unregister) fixes it; byllm's suite passes 91/91.

`tests/runtimelib/test_hook_dispatch.jac` (13 tests) guards the lifecycle on a
fresh `PluginManager` + synthetic facade, plus a codegen-vs-reflective
equivalence test built from the real `generate_plugin_helpers`. Every test is
**mutation-verified**: I injected 8 distinct bugs (devirtualize multi-impl, drop
idempotency, drop reseal-on-register, reintroduce reseal-on-unregister, break
arg-forwarding, never devirtualize, ignore each flag) and confirmed each is
caught by exactly the relevant test(s), with no silent passes.

**Validated under `--scale`**: deployed on a live minikube cluster. The seal
fires on the first request, `spawn_call` / `check_write_access` devirtualize,
jac-scale's storage hook (2 impls) correctly stays on the dynamic dispatcher,
and register/login flows work end to end.

## Benchmarks

Reproducible ablation harness in `jac/benchmarks/`
(`python benchmarks/hook_dispatch_eval.py`). Min of 5, checksum-asserted
constant across configs, timing the read-only hot loop only. The micro row is
a separate `timeit` of one no-op single-impl hook call, isolating raw dispatch:

| workload | baseline | phase1 | phase1+2 | speedup |
|---|--:|--:|--:|---|
| pure-dispatch micro (single-impl hook) | 1253 ns/call | ~330 ns/call | 36 ns/call | P1 3.8x, P1+2 34x |
| `wl_traverse_untyped` `[n-->]` | 2858 us | 1438 us | 1089 us | P1 1.99x, P1+2 2.62x |
| `wl_traverse_typed` `[n-->[?:N]]` (low fanout) | 66806 us | 66407 us | 66033 us | P1 1.01x, P1+2 1.01x |

Untyped neighbor traversal isolates hook dispatch (no planner): **Phase 1 ~2x**,
**Phase 2 adds ~32%** for **2.62x**. The typed row is the *low-fanout* case
(few records per query) and is ~1x across configs because it is dominated
(~98%) by the planner's per-call filter-lambda disassembly
(`inspect.getclosurevars` + `dis`) -- a separate optimization (planner
memoization), not hook dispatch. **High-fanout typed queries** (hundreds of
records routed through the devirtualized hooks per query) measure **~2.2x**:
fanout amortizes the planner's fixed per-query cost, so hook dispatch dominates
again. The harness surfaces both regimes honestly.

Net: the win scales with hook-call density per unit of other work.
Dispatch-bound traversal and high-fanout queries benefit 2x+; thin I/O- or
framework-bound request paths see little, as expected.

## Flags

| flag | default | off behavior |
|---|---|---|
| `JAC_HOOK_CODEGEN` | on | reflective proxy (baseline) |
| `JAC_HOOK_SEAL` | on | Phase-1 dynamic dispatch, no devirtualization |

## Files

- `jac0core/runtime.jac` - Phase-1 codegen in `make_proxy` + flag; no import-time seal
- `jac0core/plugin.jac` - `_build_devirt`, `_is_simple_sig`, PluginManager seal fields
- `jac0core/impl/plugin.impl.jac` - `seal`/`_specialize_hook`, reseal-on-register, seal flag
- `runtimelib/impl/context.impl.jac` - `seal()` at the runtime seam
- `jac-scale/.../impl/context.impl.jac` - `seal()` at the `--scale` seam (init override)
- `tests/runtimelib/test_hook_dispatch.jac` - 13 mutation-verified tests
- `benchmarks/` - ablation harness, workloads, README
- `release_notes/unreleased/{jaclang,jac-scale}/6543.feature.md` - fragments

## Follow-ups (separate PRs)

- Planner filter-lambda memoization (the dominant cost in low-fanout typed
  traversal; bigger win for type-filtered queries).
- Compile-time devirtualization for native/JS emit targets.


marsninja:

PR #6547, opened 2026-06-10
Issue per-domain TLS certs on a shared nginx ingress

udithishanka:
## What

In `shared_ingress` mode the Ingress jac-scale builds carried only `ingressClassName` and the host rule, never a `spec.tls` block. cert-manager's ingress-shim issues a certificate only when an Ingress has both a `cert-manager.io/cluster-issuer` (or `issuer`) annotation and a `spec.tls` entry, so a shared nginx controller could route a domain but never get HTTPS for it.

This adds `spec.tls = [{hosts: [domain], secretName: "{app_name}-tls"}]` to the shared-ingress Ingress when `shared_ingress_tls` is set and the class is nginx. The issuer still comes from the caller via `shared_ingress_annotations` (e.g. `cert-manager.io/cluster-issuer: letsencrypt-prod`).

## Why

It lets a single shared nginx ingress front many domains, each getting its own per-domain Let's Encrypt cert. This is the building block for attaching customer custom domains (including bare/apex domains via A records to the shared ingress's stable IPs) to one shared controller instead of one load balancer per app.

## Scope / safety

Gated on the nginx class. Non-nginx shared controllers (AWS ALB + ACM) terminate TLS at the load balancer and must not carry a `spec.tls` block, so they are unaffected. The legacy per-app path (`shared_ingress=False`) is also unaffected; it still adds TLS via the existing `enable_tls` flow. No behavior change unless `shared_ingress` + `shared_ingress_tls` + nginx class + a domain are all set.

## Test plan

Deploy an app in shared-ingress mode against a shared nginx controller with `shared_ingress_tls=True` and a `cert-manager.io/cluster-issuer` annotation; confirm the Ingress gains a `spec.tls` block and cert-manager issues the `{app_name}-tls` secret. Confirm an ALB shared-ingress deploy still produces no `spec.tls`.


udithishanka:
Addressed the review feedback:

- **Missing cert-manager annotation guard (P2):** already handled, the `spec.tls` block is gated on `cert_manager_managed` (the Ingress must carry `cert-manager.io/cluster-issuer` or `cert-manager.io/issuer` in `shared_ingress_annotations`). It is never added without an issuer annotation, so cert-manager always has a `Certificate` to act on.
- **Stale `spec.tls` after TLS is disabled (P1):** the else branch now sets `main_spec['tls'] = None`, so the strategic-merge patch removes a stale TLS block on a redeploy that turns cert-manager TLS off (or switches to a non-cert-manager class).
- **Skip-reachability bypassed pod readiness (P1):** the early return now skips only the external DNS/HTTP probe. It still calls `check_pods_restarted` and raises on a crash-looping / bad-image deploy, so a broken rollout no longer reports success.

CI is green.

udithishanka:
Good catch on the `tls = None` clobbering the dedicated path. Scoped both the add and the null-cleanup inside `if self.k8s_config.shared_ingress`, so a non-shared (dedicated) deploy never sets the `tls` key and enable_tls's cert is left untouched. `test_ingress_tls_preservation.jac` (all 6 cases) passes again.

Issue #6542, opened 2026-06-10
One Owner Per Analysis: Centralizing Jac Compiler Semantics on the UniTree


marsninja:
## Mission

Every semantic fact about a Jac program is computed **exactly once**, by the
centralized analysis pipeline, and recorded **on the unitree** (or in a
registry hanging off it). The three backend pathways - **sv** (Python
ast/bytecode), **cl** (ECMAScript), **na** (LLVM native) - become pure
*consumers*: they read annotations and emit target code; they never resolve,
infer, classify, scan, or re-derive.

Today the native backend carries a complete shadow compiler inside
`NaIRGenPass` (a stringly-typed shadow type system, shadow scoping, a shadow
class/layout model, its own closure analysis, its own OSP semantics, its own
ownership analysis), and the ES backend carries a smaller one (its own scope
stack, type-driven emission heuristics, OSP pre-scans, Ref-detection by AST
shape). This plan relocates each of those responsibilities into the central
system in **self-contained phases**: each phase moves one whole
responsibility, deletes the duplicate in the same change, and leaves exactly
one owner - **no fallbacks, no transition shims that outlive the phase**.

## Execution status (this branch)

Work executed so far on `docs/analysis-centralization-plan` (validated per
slice against the full native suite, 398 passed / 1 skipped):

- **Phase 1 (type authority) - largely done.** `TypeBase.mangle()` is the
  canonical type key; checked-type helper seams (`_type_key_of`,
  `_enum_name_of`, `_tuple_key_of`, `_set_spec_of`, `_dict_key_of`,
  `_list_spec_of`, `_struct_name_of`, `_elem_spec_of`) replace AST sniffing in
  native codegen. Deleted shadow dicts: `var_list_elem_type`, `var_dict_type`,
  `var_set_elem_type`, `var_tuple_type`, `var_enum_type`, `var_complex`,
  `var_range`, `var_slice`, `type_var_map`, `_last_type_node_hint`,
  `var_type_node`. Deliberately kept: `_list_type_hint` /
  `_dict_val_type_hint` (blocked: the checker does not bidirectionally type
  empty collection literals - `x: list[str] = []` leaves the ListVal typed as
  bare `list`), `var_list_elem_struct` (Phase 7 ownership), `field_type_node`
  (remaining readers are clib foreign-ABI lowerings, Phase 8).
- **Phase 2 (symbol/scope authority) - native consumers migrated** to
  `Symbol.storage`.
[...]

PR #6559, opened 2026-06-10
Analysis centralization: checker-stamped callee decl, native AST-signature cache deleted (Phase 4)

marsninja:
Continues #6542 (One Owner Per Analysis), finishing the Phase 4 item: *full call-resolution facts so native's `method_ast_sigs` AST-signature cache can be deleted*.

## What moved

**Central (checker):** `TypeCheckPass.exit_func_call` now stamps `FuncCall.callee_decl` - the resolved callee `Ability` - right after stamping `call_kind`, via the new `symbol_utils.ability_of_symbol` (single owner of the symbol-to-ability mapping; ES's private `_ability_from_symbol` can converge on it in a later slice).

**Native (consumer):** default-arg filling for free functions and methods reads the resolved signature off the call node through one helper (`_callee_sig_of`) instead of looking up the name-keyed `method_ast_sigs` dict. Pipe-forward calls synthesized after checking carry no stamp; the helper recovers the declaration through the target's stored `.sym` - a stamped-fact read, miroring how `call_kind` is classified on the spot for the same nodes.

**Interop layout:** the cache's two remaining readers (tuple-element struct hints, enum-return name recovery in `build_interop_layout`) become declaration-time facts (`func_ret_tuple_hints`, `func_ret_ann_name`), recorded while the Ability node is in hand.

## What was deleted

- `method_ast_sigs` (the native backend's name-keyed AST-signature cache), both writer sites, and all four reader sites' AST dissection.

## JIR discipline

`callee_decl` is a `by postinit` field: like `Expr.type` it is excluded from `__init__` and therefore never serialized (registry regenerated - no diff, no `FORMAT_VERSION` bump). `TypeInferencePass` restamps it on JIR load through the same path that restores `Expr.type`.

## Validation

- Backend purity ratchet test (no new forbidden patterns; `.sym` consumption is the blessed pattern)
- Cross-backend equivalence: `test_prim_equivalence`, `test_osp_equivalence`, `test_shared_types_equivalence` all pass
- `test_native_lexer` passes (exercises the relocated tuple-struct-hint path end to end: `tuple[list[Token], list[LexerError]]` marshalling, native/Python output equality)
- Targeted smokes: free-function/method calls with omitted trailing defaults, inherited-method defaults, pipe-forward desugar with defaults, enum returns across the interop boundary (output byte-identical to main)
- Behavioral note: the checker-resolved decl respects user shadowing where the old name-keyed dict lookup could not.


marsninja:
This PR has grown beyond the original Phase 4 slice - it now carries three self-contained slices of #6542:

**1. Phase 4 (original scope): checker-stamped callee decl, `method_ast_sigs` deleted.** See PR description.

**2. Phase 5 follow-up: one owner for OSP trigger-name derivation** (`c67f5da`). `Ability.event_trigger_type_names` (unitree) is now the single derivation of trigger type names from the checker-resolved `event_triggers`; the native backend's private copy is deleted and ES OSP descriptors consume the central fact with trigger-AST fallback. Fixes union triggers (`with A | B entry`) in cl, which the AST flattener lowered to an untyped fires-everywhere slot. Also deletes write-only `_osp_arch_names` state from the ES pass.

**3. Phases 1/6: bidirectional typing of empty collection literals, native hint threading deleted** (`cbccccb`). The evaluator now adopts the expected type for empty `[]`/`{}` literals (assignments, returns, and `has` field defaults all pass the declared type as expected), closing the "checker does not bidirectionally type empty collection literals" debt recorded in the issue. That unblocked deleting the native backend's entire empty-collection emission-context machinery: `_list_type_hint` / `_dict_val_type_hint` threading (assignments, returns, field defaults, comprehensions, for-loop collections, with save/restore at every nesting seam), the empty-dict/empty-set special cases in `_codegen_assignment`, and the orphaned `_get_list_elem_from_type_node`. List/dict/set literals now read their own checked type. Net -138 lines; purity ratchet for `stmt.impl.jac` `type_tag.tag` tightened 14 -> 6.

Validated per slice with targeted smokes (defaults incl. pipe-forward and inherited methods, empty-literal matrix across all nine contexts, OSP equivalence suite, purity ratchet); full suites left to CI.

marsninja:
CI caught a real regression from the empty-literal slice: the chess fixture segfaulted because the inner comprehension in `[[None for _ in range(n)] for _ in range(n)]` (under `list[list[Piece | None]]`) was stamped bottom-up as `list[None]`, whose element lowers to the scalar i64 None-sentinel - native built `List.i64` of `-1`s where it previously built `List.ptr` of nulls.

Fixed in `ba526cfd7` at the central level: `ListCompr` typing now participates in bidirectional inference (expected element type flows into the output expression, so nested comprehensions adopt the annotation transitively), plus a native guard restoring null-pointer semantics for bare-`None` outputs with no expected type. Chess plays to checkmate again via both `jac run` and the nacompiled binary with GC; slice smokes and purity ratchet stay green.

PR #6565, opened 2026-06-10
feat(planner): cross-root chain resolution for multi-hop queries

Gorgeous-Patrick:
## Summary
- Add `resolve_chain_cross_root` to `TopologyIndex`: iterative hop-by-hop resolver that crosses user boundaries by loading foreign roots' topology indices
- Wire into `plan_chain` so multi-hop queries stay on the fast path (topology index → materialize) instead of falling back to the edge-walk path
- Gated by `[run] cross_root_resolve = true` in jac.toml (default: false)

## Problem
When a multi-hop query like `[->:Follow:->[?:Profile]-->[?:Tweet]]` crosses user boundaries, the planner detects foreign nodes at intermediate hops and bails (`bail_on_intermediate_foreign`). The entire query falls back to `edges_to_nodes`, which triggers thousands of unnecessary edge anchor fetches from MongoDB.

In a benchmark with 5000 users (200 tweets each, 50 follows each), `load_feed` without cross-root resolution fetches ~10,200 Post edges and ~8,450 Follow edges from L3 — none of which are needed on the fast path.

## How it works
`resolve_chain_cross_root` walks the chain hop by hop:
1. Resolve the current hop within the current root's index
2. Classify results into local vs foreign nodes (`get_cross_root_ids`)
3. For each foreign root, load its topology index and continue the next hop there
4. Union all results

The planner uses this for multi-hop chains when `cross_root_resolve` is enabled, keeping the query on the index-only path across root boundaries.

## Security note
The resolver decodes foreign users' topology indices, exposing graph structure (node types, connectivity) internally to the planner. Actual data remains access-controlled — `_materialize_ids` filters results by permission before returning to walker code. Gated off by default.

## Test plan
- [ ] Run existing test suite
- [ ] Verify default (`cross_root_resolve = false`) produces identical behavior to main
- [ ] Enable `cross_root_resolve = true` and verify multi-hop cross-user queries return correct results
- [ ] Benchmark with LittleX5 dataset to measure edge fetch reduction

### DIFF
--- a/docs/docs/community/release_notes/unreleased/jaclang/6565.feature.md
+++ b/docs/docs/community/release_notes/unreleased/jaclang/6565.feature.md
@@ -1,0 +1,1 @@
+- **Feat: Cross-root chain resolution**: Add `resolve_chain_cross_root` to TopologyIndex so multi-hop queries crossing user boundaries stay on the fast path (topology index → materialize) instead of falling back to edge-walk. Gated by `[run] cross_root_resolve = true` in jac.toml.
--- a/jac/jaclang/project/config.jac
+++ b/jac/jaclang/project/config.jac
@@ -49,6 +49,7 @@ obj RunConfig {
         cache: bool = True,
         autonative: bool = False,
         topology_index: bool = True,
+        cross_root_resolve: bool = False,
         diagnostics: str = "error";  # "error" | "all" | "none"
 }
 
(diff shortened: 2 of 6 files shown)


Gorgeous-Patrick:
@greptileai Please re-evaluate.

marsninja:

Issue #6571, opened 2026-06-10
native: allocate clib vtable structs without RC header (flat C layout)


chess10kp:
## Summary

C vtable structs (e.g. CEF's `cef_life_span_handler_t`, `cef_app_t`) are plain C structs — they have no reference-count header. When Jac allocates a clib struct instance today, the default path uses `_rc_call_with_ctx` which prepends an RC header at `ptr - 8`. When C code receives a pointer to such a struct, it reads from the wrong offset and the vtable is garbled.

Additionally, C vtable structs are typically **large** (>16 bytes on SysV AMD64), making them MEMORY-class aggregates that must be passed by pointer (byval), not in registers.

## Expected behavior

When a clib struct is detected as a vtable (has `Callable`/fn-ptr fields), the compiler should:

1. **Skip RC allocation** — allocate via `_entry_alloca(_clib_c_type(name), ...)` so the pointer C sees points directly at field 0 (no RC header)
2. **Use C-laid-out type** — `_clib_c_type(name)` returns the LLVM struct type matching the C memory layout (packed, no Jac metadata fields)
3. **Static storage** — entry-block alloca so the vtable address is stable for the program lifetime (vtables must outlive any single function scope)
4. **MEMORY-class passing** — for structs >16 bytes on SysV AMD64, pass by pointer with `byval` attribute (existing `_codegen_clib_call` handles this, but allocation must cooperate)

## Proposed implementation (from spike)

In `objects.impl.jac` → `_codegen_instantiation`:

```jac
if type_name in self.clib_vtable_structs {
    ptr = self._clib_alloc_vtable(type_name);  # entry alloca, no RC
    # field stores use _clib_coerce_callback for fn-ptr slots
    # skip _emit_rc_retain for pointer fields in vtable structs
}
```

New method `_clib_alloc_vtable(name)` in `clib_vtable.impl.jac`:

```jac
impl NaIRGenPass._clib_alloc_vtable(name: str) -> ir.Value {
    ct = self._clib_c_type(name);
    stype = self.struct_types[name];
    sptype = stype.as_pointer();
    self._clib_vtable_counter += 1;
    buf = self._entry_alloca(ct, f"clib.vt.{name}.{self._clib_vtable_counter}");
    return self.builder.bitcast(buf, sptype, name="clib.vt.ptr");
}
```

## Real-world sizes (CEF 119, Linux x86_64)

| Struct | Size (bytes) |
|--------|-------------|
[...]

PR #6575, opened 2026-06-11
feat(native): C→Jac callback trampolines for clib struct fn-ptr fields

### LINKED ISSUE #6570: native: C→Jac callback trampolines for clib struct fn-ptr fields
## Summary

When a clib struct has `Callable` fields (vtable slots), assigning a Jac `def` to that field must produce a **C-ABI trampoline** — a thin LLVM function with the C calling convention that coerces arguments and calls into the Jac function. Currently there is no trampoline emission, so a Jac function pointer stored in a clib struct has the wrong ABI when C code calls it back.

## Expected behavior

```jac
import from "libcef.so" {
    obj cef_life_span_handler_t {
        has size: int = 0,
            on_before_close: Callable[[int, int], None] = 0;
    }
}

def _on_before_close(self: int, browser: int) -> None {
    g_called = 1;
}

h = cef_life_span_handler_t(size=56, on_before_close=_on_before_close);
# C code calls h.on_before_close(h, browser) — must go through a trampoline
```

The compiler should:
1. Detect that `_on_before_close` (a Jac `def`) is being stored in a `Callable` field of a clib struct
2. Emit a wrapper LLVM function (`_on_before_close.__clibcb.0`) with the **C function signature**
3. Store the trampoline address (not the raw Jac fn address) in the struct field
4. Cache trampolines to avoid duplicates per `(jac_fn, c_fn_ty)` pair

## Proposed implementation (from spike)

New file `na_ir_gen_pass.impl/clib_vtable.impl.jac` (~110 lines) with:

- `_clib_struct_has_fn_ptr(name)` — detect vtable structs by scanning for `PointerType(FunctionType)` fields
- `_clib_detect_vtable_structs()` — run after clib struct registration (pass 1.5)
- `_clib_coerce_callback(val, c_fn_ty)` — `int 0` → null ptr, Jac `def` → trampoline, passthrough otherwise
- `_emit_clib_c_trampoline(jac_fn, c_fn_ty)` — create LLVM wrapper with C signature, coerce args/return, call Jac fn
- Cache in `clib_trampoline_funcs: dict[str, ir.Function]`

Naming convention: `{jac_fn.name}.__clibcb.{n}` — grep for `__clibcb` in LLVM IR tests.

## Use case

CEF vtable callbacks (`on_before_close`, `on_context_created`, etc.) are invoked by the CEF C runtime. Without trampolines, the Jac function is called with the wrong calling convention and crashes.

Related: #6569 (Callable = 0 defaults — prerequisite for declaring the fields as Callable in the first place).

marsninja:
## Summary

Fixes #6570. **Stacked on #6574** (its commit is included here; rebases cleanly once #6574 merges).

A Jac `def` stored into a `Callable` field of a clib struct is later invoked **by the C runtime** with the C signature from the slot's annotation — which need not match the Jac function's native signature (Jac `int` params are i64; a CEF/libuv slot may declare `i32`, `f64`, ...). Previously the raw function address was bitcast into the slot, handing C a pointer with the wrong ABI.

## What changed

- New `na_ir_gen_pass.impl/clib_vtable.impl.jac`:
  - `_clib_detect_vtable_structs()` (pass 1.75 of clib import processing) records clib structs with fn-ptr fields in `clib_vtable_structs`.
  - `_clib_coerce_callback(val, c_fn_ty)`: a matching Jac def stores directly; a mismatched def goes through a trampoline; `int 0` lowers to a NULL C fn ptr.
  - `_emit_clib_c_trampoline(jac_fn, c_fn_ty)`: emits an internal LLVM function named `{fn}.__clibcb.{n}` with the slot's exact C function type — sign-extends/truncates integer args between C and Jac widths (zext is wrong for signed C ints), calls the Jac function, coerces the result back. Arity differences are tolerated (extra C args dropped, missing Jac args zero/null-filled). Cached per `(jac_fn, c_fn_type)` pair in `clib_trampoline_funcs`.
- Constructor kwargs, field defaults, and field re-assignment stores into vtable fn-ptr slots all route through `_clib_coerce_callback`; those slots are code addresses, so the RC retain/release discipline is skipped for them.

## Tests

- IR tests: trampoline emitted with the C signature (`define internal i32 @"_on_event.__clibcb.0"(i32 ...)`), exactly one definition per (fn, type) pair.
- Execution test (skipped if no `gcc`): builds a real C shared library whose exported function invokes the stored callback through the struct. Covers i32→i64 sign extension with **negative arguments** (a raw bitcast store passes the positive-args case by ABI accident; negatives prove real marshalling), i64→i32 return narrowing, and the NULL-slot path.

Targeted runs: the 3 new tests plus the `fnptr`/`clib`/`callable`/`struct` native suites (37 tests) pass.

## Stack

Part 2 of 3 for deleting `cef_shim.c` in jac-desktop: #6574 (`Callable = 0`) → this PR → #6571 (flat vtable allocation).


marsninja:
Superseded: #6576 was merged directly and, being stacked on this branch, its squash carried the full trampoline implementation (clib_vtable.impl.jac, store-site hooks, tests, and the 6575 release fragment) into main at 95476780a. This branch now contains only pre-#6576 versions of the same files, so there is nothing left to merge. Closing; #6570 is fully implemented in main.

PR #6578, opened 2026-06-11
docs: drop em dash from 6574 release fragment

marsninja:
The `ban-em-dashes` pre-commit hook rejects em dashes in markdown, and pre-commit.ci runs hooks against **all files** — so this fragment (merged in #6574 just before the punctuation fix landed on its branch) currently fails the `pre-commit.ci - pr` check for every open PR (e.g. #6575, #6576).

One-character fix: replace the em dash with a comma.


marsninja:
Superseded: the #6576 squash (95476780a) included this em-dash fix for the 6574 release fragment, so main is already clean and pre-commit.ci is unblocked.

PR #6612, opened 2026-06-12
feat(jac-scale): Ops Console - Events (V-05)

MusabMahmoodh:
## What
A new **Operations -> Events** admin page: recent namespace events Warning-first, with a pressure summary that counts and highlights the high-signal failure reasons (OOMKilled, FailedScheduling / Insufficient*, FailedBinding / unbound PVC, ImagePullBackOff, Evicted).

## How
Namespace-scoped, read-only; requires the opt-in read-only RBAC (`[plugins.scale.kubernetes].ops_console = true`); "unavailable" off-cluster. Pure reshape + pressure-classification helpers separate from the K8s call. Registered on both servers.

## Test
Unit tests over pressure-reason matching + Warning-first sorting + registration.


jac-scale-test-triage.md
```diff
@@ -0,0 +1,129 @@
+# jac-scale CI test triage (PR #6850)
```
Thamirawaran, line 1: lets remove this unrelated file

jac-scale/jac_scale/admin/impl/events.impl.jac
```diff
@@ -0,0 +1,172 @@
...
+         "backoff",
+         "evicted",
+         "failedmount",
+         "unhealthy",
+         "nodenotready"
+     ];
+
+
+"""'' for None, else str(v)."""
+def _s(v: any) -> str {
```
Thamirawaran, line 41:
```
def _s(v: str | None) -> str {
```

Thamirawaran: and do we actually need this function. 
can we do it with ternary
`"" if v is None else v`
fine if it is needed but have a good name for function

Thamirawaran (changes requested):

PR #6615, opened 2026-06-12
fix(jac-scale): ignore undeclared query params on restspec endpoints (closes #4661)

### LINKED ISSUE #4661: Bug: Unknown query parameters crash function and walker endpoints

akindu-k:
## The problem

Walker and function endpoints (`@restspec`) crashed with a **500 error** whenever a request URL carried a query parameter the endpoint did not declare.

This happens constantly in the real world: browsers add cache-busting params like `?v=1708012345`, and CDNs/proxies/analytics tools append their own tracking params. None of these have anything to do with the endpoint, and a normal web framework simply ignores them. Here they brought the whole request down.

Example that used to fail:

```
GET /function/get_func_with_params?name=Bob&age=40&v=1708012345&t=abc
```

The function only declares `name` and `age`, but `v` and `t` were passed in too, which is like calling `get_func_with_params(name="Bob", age=40, v="...", t="abc")`. The function has no `v` or `t` argument, so it raised `got an unexpected keyword argument 'v'` and returned a 500.

## Why it happened

The generic route builder that turns walkers/functions into URLs builds a handler that accepts `**kwargs` (it has to be generic across every walker and function). Because the handler accepts anything, the builder had a deliberate rule: *forward **every** query parameter from the URL into the handler, including ones nobody declared.*

Walkers and functions then passed that dict straight into the walker's `.init` / the function's signature, which rejected the undeclared params and crashed.

## How it was fixed

Rather than patching each callback to clean up the mess after the fact, the decision was moved to the one place that already knows which parameters an endpoint declared: the route builder.

- Added a `forward_undeclared_query_params` flag to `JEndPoint`, defaulting to `True` so **all existing behavior is preserved**.
- The route builder now honours the flag: when it is `False`, undeclared query params are dropped before the callback is ever called.
- Walker and function endpoints set the flag to `False`, so unknown query params are ignored (matching FastAPI's native behavior). Endpoints that legitimately want arbitrary query params (e.g. static page rendering) keep the default and are untouched.

The walker/function callbacks themselves are unchanged. There is a single source of truth for what each endpoint accepts, and the fix applies uniformly to every endpoint type.

## Tests

Added regression tests asserting that GET function and walker endpoints return 200 with the correct data when extra cache-busting query params are present. Both fail without the fix (reproducing `got an unexpected keyword argument 'v'`) and pass with it. Full restspec suite (19 tests) passes.

Closes #4661


akindu-k:
## Investigation: which `**kwargs` endpoints actually need the forwarding?

Following up on the design of the opt-out flag, I audited every endpoint whose callback accepts `**kwargs` (the only callbacks the undeclared-query-param forwarding can affect). This confirms why a per-endpoint flag is the right shape rather than flipping the default globally.

| Endpoint | What it does with `kwargs` | Needs forwarding? | Flag |
|---|---|---|---|
| **Walker** | Forwards to `walker.init` (strict signature) → **crashed** | No | `False` (fixed) |
| **Function** | Forwards to the function (strict signature) → **crashed** | No | `False` (fixed) |
| **Webhook** (`_process`) | **Ignores `kwargs` entirely** — reads everything from the request body and headers | No | `True` (default), harmless |
| **Static page** (`render_page_callback`) | **Passes `kwargs` into `render_page(...)` by design** | **Yes** | `True` (default), required |
| **Static base route** (`render_base_route_callback`) | Same — passes `kwargs` into `render_page(...)` | **Yes** | `True` (default), required |

### Takeaways

1. **Static page rendering is a genuine consumer** of undeclared query params. The callbacks are built to accept arbitrary params (`?theme=dark&lang=fr&...`) and feed them all to the renderer; the docstrings say so explicitly ("Render a page by name with all query parameters"). Flipping the default to `False` globally would break this.
2. **Webhooks are unaffected** either way. `_process` never reads `kwargs`; it parses the request body and filters it against the walker's fields itself, so undeclared query params are already dropped there. Leaving the flag at `True` is harmless.
3. **Walkers and functions were the only endpoints that actually broke**, and they are exactly the two that opt out with `forward_undeclared_query_params=False`.

This validates the per-endpoint opt-out approach: a single global "ignore unknown params" rule would have been incorrect because static pages depend on the opposite behavior. The only optional tweak would be setting the webhook endpoints to `False` for consistency, which is purely cosmetic since their behavior would not change.

PR #6622, opened 2026-06-12
fix(byllm): reset ContextVar Token before yielding streaming generator

Hirudika4:
## Problem

When `by llm(stream=True)` was used with jac-scale's SSE endpoint, the server
crashed with:

    ValueError: Token was created in a different Context

jac-scale drives streaming generators on a thread-pool worker via
`loop.run_in_executor`. That worker runs in a **foreign** `contextvars.Context`
— different from the one that called `Model.invoke()`. The old `_wrap_generator`
held a `ContextVar` Token and called `_current_invocation_id.reset(token)` in
its `finally` block, which ran on that foreign worker thread. Python's
`ContextVar.reset()` requires the Token to be reset in the same Context that
created it — cross-Context reset raises `ValueError`.

## Fix

- **`Model.invoke()`**: on the streaming branch, reset the Token immediately
  (in the original Context, before returning), then delegate to
  `_stream_with_telemetry`.

- **`_stream_with_telemetry()`**: replaces `_wrap_generator`. Uses
  `contextvars.copy_context()` to capture a private Context snapshot, pins
  `_current_invocation_id` in it via `ctx.run(set, ...)`, then drives each
  `next()` call via `ctx.run(next, gen)`. No Token is ever reset across
  Contexts. The generator is now safe to pull from any thread.


marsninja:

PR #6632, opened 2026-06-12
fix(type-checker): eliminate three false-positive classes (class→Callable, recursive aliases, init/__init__ + stub TypeAlias)

kugesan1105:
## Summary

Running `jac check` over real-world Jac code (entry point: `jaclang/runtimelib/impl/server.impl.jac`, 135 errors / 229 warnings) surfaced several classes of **false positives** in the type checker - places where the *inferred type was factually wrong*, so the diagnostics pointed at phantom problems. Each was reduced to a minimal reproducer, root-caused against the corresponding pyright behavior, and fixed at the source, with a regression fixture + test per fix.

**Scope note:** this PR deliberately does **not** relax strict-Any semantics in any way. `Any` still cannot flow into any declared non-`Any` destination (annotated assignments included); `typing.cast()` remains the only interop escape. An earlier revision of this branch experimented with pyright-style "annotated declarations absorb Any" and related fallbacks; that direction was rejected in review and fully reverted - Jaclang keeps its maximally strict gradual semantics.

### 1. Class objects are assignable to `Callable` (constructor conversion)
`jac/jaclang/compiler/type_system/type_evaluator.impl/type_evaluator.impl.jac` - `assign_type`, FunctionType-destination block

`json.dumps(data, default=str)` errored with *"Cannot assign str to parameter 'default' of type Callable[[Any], Any] | NoneType"*. The FunctionType-destination rules handled function sources, overloads, and instances with `__call__`, but had no branch for an instantiable **class** source. A class is callable via its constructor, so this mirrors pyright's `createFunctionFromConstructor`: synthesize a callable from the constructor (reusing the existing `_get_constructor_for_inference`) whose return type is the class instance, then run normal callable assignability. A deliberate mismatch (`Callable[[int], int] = str`) still errors via the covariant return check.

### 2. Recursive / forward-ref type aliases no longer degrade to `Literal["Name"]`
`type_evaluator.impl.jac` (generic specialization), `evaluator_util_methods.impl.jac` (`_try_resolve_forward_ref`)

```jac
glob JsonValue: TypeAlias = None | str | int | ... | list['JsonValue'] | dict[(str, 'JsonValue')];
```

Two stacked bugs made every `dict[str, JsonValue]` site in the serve runtime fail with messages like *"Cannot assign ... to dict[str, ... | list[Literal["JsonValue"]] | dict[str, Literal["JsonValue"]]]"*:

- Quoted forward refs in **generic subscript arguments** (`list['Json']`) were typed as the string literal `Literal["Json"]` - the forward-ref path only fired when the string's parent was a `SubTag`/`FuncSignature`. The class-subscript path now routes each type argument through `_type_from_union_annotation_arg`, the same helper the `Union[...]`/`Optional[...]` path already uses.
- The **recursive knot** (`'Json'` inside Json's own RHS) hit the symbol-resolution cycle breaker, got `Unknown` back, and fell through to the literal branch. It is now truncated gradually at depth 1 with `Any` (`list[Any]` / `dict[str, Any]` members), keeping assignments into the alias sound. A pyright-style `TypeAliasPlaceholder` with coinductive assignability is noted as a future refinement for full recursion-depth precision.

### 3. `cls.init` resolves to `__init__`; qualified `TypeAlias` stub form recognized
`type_evaluator.impl.jac`

- `inspect.signature(walker_cls.init)` errored with *"type[WalkerArchetype] has no attribute init"* - but Jac's `init` **is** `__init__` at the Python level (`PyastGenPass._py_name` maps the `KW_INIT` token, so the access works at runtime). Attribute expressions now fall back `init -> __init__`, scoped to explicit attribute access on class / `type[T]` receivers so constructor synthesis (which probes `init`/`__init__` itself, in order) is unaffected.
- `_is_type_alias_annotation` only matched the bare `TypeAlias` name; typeshed stubs use the qualified `typing_extensions.TypeAlias` form (e.g. `_get_type_hints_obj_allowed_types` in `typing.pyi`). The alias annotation check now accepts both, fixing `get_type_hints(func)` reporting *"Cannot assign Callable[..., Any] to parameter 'obj' of type _SpecialForm"*.

## Tests

- `checker_recursive_type_alias.jac` + `recursive_type_alias_forward_refs` - recursive alias assignment/returns with nested payloads, plus forward refs in generic subscripts outside the alias: 0 errors expected.
- `checker_class_as_callable.jac` + `class_object_assignable_to_callable` - `json.dumps(default=str)`, `Callable`-typed annotations and parameters accepting classes; the deliberate return-type mismatch still errors (exactly 1 expected error).
- `strict_any_in_jac_modules` and its fixture are untouched relative to `main` (4 strict-Any errors), pinning that strict-Any behavior is unchanged by this PR.

Validation: `test_checker_pass.jac` plus `test_checker_bugs/gaps/production/shortcomings` all green; pre-commit (jac-format, markdownlint, release-notes validator) clean.

## Risk notes

- Fix 1 uses the first meaningful constructor overload (same limitation as `_get_constructor_for_inference` elsewhere); a later-overload match could in principle be missed. The display string for a class in this diagnostic still prints `str` rather than `type[str]` - display-layer cleanup left for a follow-up.
- Fix 2's depth-1 `Any` truncation trades recursion-depth precision for soundness; noted follow-up above.
- Fix 3's `init -> __init__` mapping is fallback-only (fires only when no `init` member exists), so Jac classes that declare an explicit `init` ability are unaffected.


jac/tests/compiler/passes/main/fixtures/checker/checker_bug_typevar_leak_any_arg.jac
```diff
@@ -0,0 +1,36 @@
...
+        }
+        return "";
+    }
+}
+
+with entry {
+    m: dict[(str, Any)] = {};
+    unannotated = list(m.get('exports', []));
+    for n in unannotated {
+        print(n.lower());
```
kugesan1105: `n.lower` should be an error bcz we are expecting devs to decalre the unannotated with proper types

marsninja:

Issue #6619, opened 2026-06-12
Make referential integrity first-class: graph ChangeSet (unit of work), backend apply() contract, read-path dangling-ref healing, and jac db fsck


marsninja:
## Summary

#6537 (create-path crash atomicity) and #6587 (delete-path gap + read-path intolerance) are not two bugs — they are two symptoms of the same missing abstraction. The referential-integrity invariant ("never write a reference before its referent exists; never delete a referent while a reference remains") is currently owned by **no layer**: #6537 hand-encoded one direction of it inside one backend's flush loop, and #6587 shows the other direction violated in core, for every backend.

This issue proposes making the invariant first-class, in five pieces:

1. **`ChangeSet`** — a typed unit-of-work in `jaclang/runtimelib`, owning write-staging order once, in shared code.
2. **`PersistentMemory.apply(changeset)`** — the new backend contract. Mongo implements it as staged ordered writes (upgrading to a real transaction when a replica set is available), SQLite as one transaction. #6537's buckets and the inline `l3.delete` both dissolve into this.
3. **Abort-on-error: a failed request discards its changeset.** Today a walker that raises mid-execution still persists everything it did before the raise (see finding below) — and `del` does inline I/O, so even abandoning the request cannot undo it. Once all writes are intents, "walker errored → discard instead of apply" is a one-line policy that gives every walker rollback-by-exception at request granularity.
4. **`resolve_ref()`** — a read-path reference-fault boundary: traversal skip-and-quarantines dangling references instead of raising, reusing the existing quarantine machinery. This is the only piece that heals data corrupted **before** any write fix shipped (the production incident in #6587 was minted a week before #6537 merged).
5. **`jac db fsck`** — an integrity scanner/collector that completes the "orphans are harmless garbage" model #6537 implicitly committed to.

Explicit developer-facing `atomic { }` blocks (group N connects/deletes into one all-or-nothing unit *within* a request) are deliberately **deferred** — see the scope comment below this issue for the tier analysis.

Fixes the delete side of #6488 / #6587 structurally rather than by adding more backend-local buckets.

## Why the current architecture keeps minting this bug class

All findings verified on current `main`.

[...]

Issue #6587, opened 2026-06-11
Node delete is not crash-atomic: immediate edge/node delete_one vs deferred data.edges update leaves dangling refs that 500 every traversal (delete-side of #6488)


udithishanka:
## Summary

#6537 made node **creates** crash-atomic, but the **delete path still is not**, and there is **no read-path tolerance** for the dangling references that already exist in production databases. Either one keeps producing user-facing `is not a valid reference` 500s. This issue covers both, because fixing only the delete ordering still leaves every pre-existing dangler (minted before #6537 was deployed, or by future unknown gaps) permanently breaking its user.

Deterministic reproducer below, verified on current `main` (`9e815a3`): one SIGKILL inside the delete window leaves durable Mongo corruption, and after restart every unfiltered traversal (`[-->]`, `visit [-->]`) on that root returns:

```
500 EXECUTION_ERROR: EdgeAnchor [0356e3a1-8122-4efe-a006-ff8fe203e735] is not a valid reference!
```

## Production impact (real incident, 2026-06-11)

On a production jacBuilder-derived deployment, a user's login dashboard froze permanently: `/walker/me` and `billing_ops` returned 500 on every call with `NodeAnchor [805397da-...] is not a valid reference!` (24 hits in 30 minutes across pods). Forensics:

- The user's `Root.data.edges` and two `EdgeAnchor` docs were intact, but both edge **targets** did not exist anywhere in `_anchors`.
- The orphan edges' `updated_at` was `2026-06-02T14:46:58Z` and `2026-06-02T14:55:29Z`, i.e. minted by the **create-path gap a week before #6537 merged**, then sat dormant until the user logged in and every traversal of his root started throwing.
- A full scan found **19 dangling edges across 8 user roots** in that one database. Each of those users hits the same permanent freeze on login.

Two takeaways drive this issue:

1. Write-order fixes do not repair data. Every production DB that ever ran pre-#6537 code (or experienced the delete gap below) carries mints that will keep detonating indefinitely. Only a read-path change ends the user-facing failures.
2. The delete side can keep minting new ones even on current main, as the reproducer shows.

## The exact gap on the delete path (current main, `9e815a3`)

`JacBasics.destroy` for a node (`jac/jaclang/jac0core/impl/runtime.impl.jac:1386`) does:

1. For each incident edge: destroy the edge, which calls `mem.delete(edge.id)`.
2. Then `mem.delete(node.id)` for the node itself.

[...]

PR #6649, opened 2026-06-13
feat: read-path dangling-ref healing, jac db fsck, get_persistent_memory hookspec (#6619)

marsninja:
Implements the three pieces of #6619 deferred from the ChangeSet PR (#6633): read-path dangling-reference healing (piece 4), `jac db fsck` (piece 5), and the `get_persistent_memory` hookspec (phase 2b). Closes the read-path half of #6587.

## Piece 4 — read-path healing + `DANGLING_REF`

`get_edges` / `get_edges_with_node` / `edges_to_nodes` now **resolve-or-skip** a dangling reference instead of raising `ValueError: ... is not a valid reference` on every read shape — the exact #6587 production incident (a node citing an edge document deleted before the ChangeSet delete-ordering shipped).

- New `resolve_ref(stub, referrer)` (`jaclang/jac0core/runtime.jac`): returns the populated anchor, or skip-and-heals a genuine dangler — logs, files it into the **existing** quarantine store under a new `DANGLING_REF` reason code (so it surfaces through `jac db quarantine list/show`), and prunes the stale citation.
- The pruning is staged as a normal `EDGE_LIST_DELTA` intent through the merged ChangeSet, so the repair flushes via `commit() -> apply()` like any edge mutation — a **read-only** request (the corrupted dashboard load) self-heals on commit, not just in memory.
- A new `is_quarantined()` gate distinguishes a **permanent** dangler (referent absent everywhere → heal) from a **recoverable** one (a schema-drift / class-missing quarantine that `jac db recover` can still restore → citations left intact). Without this gate the healer would prune citations the V1→V2 recovery flow depends on.
- Direct attribute access on a stale handle still raises (programmer error, not a storage state) — only graph traversal heals.

## Piece 5 — `jac db fsck [repair]`

A new `jac db` action that completes the orphans-are-collectable model the ChangeSet implicitly committed to.

- Read-only by default (a safe monitoring probe): scans for dangling refs (node→missing-edge, edge→missing-node), orphan edges, and orphan nodes, and prints a summary.
- `jac db fsck repair`: prunes dangling citations (filing each under `DANGLING_REF`) and collects orphans, all inside one `BEGIN IMMEDIATE` transaction so the repair is itself crash-atomic. The hand-written forensics from #6587's incident become one command.

## Piece 2b — `get_persistent_memory` hookspec

Because every Jac runtime hook is `firstresult`, two plugins implementing `create_j_context` conflict rather than compose. This additive hookspec shrinks a DB backend to a `PersistentMemory` subclass plus one hookimpl.

- `TieredMemory.postinit` and `ScaleTieredMemory.postinit` consult it before falling back to `SqliteMemory` / `MongoBackend`. Core default returns `None`, so absent a plugin nothing changes; a returned backend wins and composes with both core bare-serve and jac-scale.

## Backend contract

`PersistentMemory` gains `quarantine_dangling`, `fsck`, and `is_quarantined`, implemented in **both** `SqliteMemory` and `MongoBackend` (mirroring the cascade-quarantine pattern already in each). SQLite paths are execution-tested; the Mongo paths mirror them and are structurally validated (no Mongo on this runner).

## Tests

Written first, failing on `main` for the right reasons:

- `test_dangling_ref_heal.jac`: a `jac enter` subprocess against a DB corrupted out of band (the #6587 shape) — `visit [-->]` no longer crashes, the stale citation is pruned and persisted, and the dangler is filed under `DANGLING_REF`. On `main` the walker raises `is not a valid reference`.
- `test_fsck.jac`: `jac db fsck` detects the dangler read-only (no mutation), `fsck repair` prunes it + files `DANGLING_REF`, and a re-scan is clean. On `main` the action does not exist.
- `test_get_persistent_memory_hook.jac`: the hookspec exists and `TieredMemory.postinit` installs a hook-provided L3, else falls back to `SqliteMemory`.

Regression: the read-path heal is gated by `is_quarantined` so the schema-drift recovery flow (`test_db_cli`) is preserved; verified green alongside `test_changeset`, `test_topology_integration`, `test_persistence_migration`, `test_topology_index`, `test_layer3_e2e`, and core `test_abort_on_error`.

Refs #6619, #6587. Builds on #6633.


marsninja:
Thanks for the review — the ghost-endpoint self-poisoning is a real bug and is now fixed (commit dbe6f49).

**Root cause (confirmed):** the heal gate returned `True` for *any* quarantine row, including the `DANGLING_REF` forensic row `resolve_ref` had just filed. In the edge-cites-deleted-node case the loop then saw that fresh row, treated the genuine dangler as recoverable, and skipped pruning the dead edge.

**Fixes:**
- Renamed `is_quarantined` → `is_recoverable_quarantine` and it now **excludes `DANGLING_REF`** rows (SQLite `error NOT LIKE 'DANGLING_REF%'`, Mongo `reason_code != 'DANGLING_REF'`). The healer is no longer poisoned by its own records, and a genuine dangler keeps being pruned each pass until the repair commits.
- The ghost-endpoint branch in all three traversal helpers now checks **both** missing endpoints (it previously evaluated only one) and prunes if *either* is genuinely gone.
- `is_recoverable_quarantine` is now part of the **`PersistentMemory` abstract surface**, so a third-party backend can't silently fall back to the `Memory` default of `False` (which would prune recoverable citations).
- `jac db fsck` summary counts distinct referrers across both dangling directions and the misleading "roots/node-only" wording is gone.

**New coverage:** added a ghost-endpoint test (edge whose target node was deleted) to `test_dangling_ref_heal` — it fails without the gate fix and passes with it. The schema-drift recovery flow (`test_db_cli`) stays green.

Also pushed docs (commit 1c0c0f3): `jac db fsck` in the CLI reference, the `get_persistent_memory` hook + the new referential-integrity methods in Recipe 7, and a "Dangling references and read-path healing" section in the persistence guide.

PR #6691, opened 2026-06-14
fix(type-checker): widen loop-reassigned variables on the loop exit edge

SandeepaHWP:
## Summary

The type checker reports a false error when a variable is reassigned inside a loop and then used after the loop. This PR fixes the flow analysis so that a variable's type after a loop correctly reflects the assignments made in the loop body.

## The problem

```jac
def first_or_zero(items: list[int]) -> int {
    fb: int | None = None;
    for it in items {
        fb = 5;
    }
    if fb is not None {
        return fb;   # E1002: Cannot return int | None, expected int
    }
    return 0;
}
```

The `if fb is not None` guard should narrow `fb` to `int`, so `return fb` is valid. Instead the checker raised a false `int | None` versus `int` error. Hovering `fb` after the loop showed `NoneType`, as if the loop body never ran. Both `for` and `while` loops were affected.

## Root cause

The flow analysis resolves a variable's type by walking backward over the control flow graph. A loop has this shape:

Two facts combine to cause the bug:

1. The statement after a loop does not connect to the loop body. It connects to the loop header (the loop exits from the header when the iterator is exhausted). So the post-loop type has to come through the header.
2. When the evaluator computes the header's own type, it deliberately drops the body back-edge so that pre-loop narrowing survives into the body (this is intentional and is needed for cases like a variable narrowed by `isinstance` before the loop that should stay narrowed inside it).

As a side effect of step 2, an assignment made in the body (`fb = 5`) never reached the post-loop join. The variable stayed at its pre-loop type (`NoneType`), the `is not None` guard then had nothing to narrow, and the return raised a false error.

## The fix

When the backward walk reaches a loop header from a node that is outside the loop (the exit edge), the post-loop type is now computed as the union of:

- the pre-loop (entry) type, because a loop body may run zero times, and
- every type the body assigns into the variable.

For the example above this yields `union(NoneType, int)` which is `int | None`, so the `is not None` guard narrows it to `int` and the error disappears.

### Finding the body's writes

The body's assignments are found through the variable's own symbol, not a syntactic scan. The helper iterates the symbol's `defn` and `uses` lists, keeps only `Store`-context name nodes that are inside the loop, and reads the assigned type from each. This has two important properties:

- **Scope-sound.** A nested `def`/lambda with a same-named local is a different symbol, so its writes are not counted. A blind AST sweep over the loop subtree would wrongly pull those in.
- **Precise.** When the body assigns only `int` into an `int | str | None` variable, the result is `int | None` (entry plus the assigned type), not the full declared type. A following `is not None` then narrows to `int`.

Only a plain `name = value` write yields a usable type. Augmented assignment (`+=`), walrus (`:=`), destructuring, and `for`-targets cannot be read cleanly, so the helper falls back to the declared type for that variable, which is always sound. When the body never writes the variable, the entry type is returned unchanged, so existing pre-loop narrowing through a loop is preserved.

The assigned type is kept only when it is a strict subtype of the declared type, the same rule used on normal control flow edges.


SandeepaHWP:
Dont need anymore

PR #6695, opened 2026-06-15
feat(jaclang): kind-aware `jac run` dispatch from jac.toml

marsninja:
## Summary

Makes `jac run` (with no file argument) project-aware: it resolves the project *kind* from `jac.toml` and does the natural primary action for that kind — **execute**, **serve**, or **build** — instead of requiring an explicit file. Builds on the project-root/entry-point work in #6694.

Today the `jac.toml` `entry-point` is largely decorative and there is no single "run my project" command: `jac run` needs an explicit file and `jac start` hardcodes `main.jac`. This change makes the project's declared kind load-bearing.

## Behavior

A bare `jac run` resolves `[project] kind` (or infers it from the entry-point's codespace) and dispatches:

| Kind | Action | Delegates to |
|---|---|---|
| `cli`, `native-app` | execute | `run` |
| `native-binary` | build | `nacompile -o dist/<name>` |
| `shared-library` | build | `nacompile --shared` |
| `api-service` / `microservices` | serve | `start [--no-client]` |
| `fullstack` | serve | `start --dev` (needs jac-client) |
| `wasm` | serve | `start` — emits `/static/main.wasm` (needs jac-client) |
| `desktop` / `mobile` | serve/build | needs jac-desktop / jac-client |
| `pypi-package` / `npm-package` | build | `bundle --target …` |

`jac run --show` prints the resolved plan (kind, action, equivalent primitive command) without executing. Explicit `jac run <file>` is unchanged.

## Changes

- `[project] kind` added to `ProjectConfig` + jac.toml parsing.
- New `jaclang/project/kinds.jac`: `resolve_entry` (with stale-entry stem fallback), `resolve_project_kind` (explicit kind, else inferred), `plan_project_run` → `RunPlan` for all 12 kinds.
- `jac run` no-file dispatch + `--show`, delegating to the existing primitives (`run`/`start`/`nacompile`/`bundle`). No primitive logic duplicated.
- Docs/skills: `[project] kind` (config reference + jac-config skill), CLI reference `jac run`, quick-guide + skill project-kinds.

## Tests (`tests/language/test_run_dispatch.jac`, 24)

- A planner test for every one of the 12 kinds (action/verb/flags).
- Kind inference, entry-point resolution (incl. stale-entry fallback), kind validation.
- `--show` prints without executing; unknown kind errors clearly.
- Real execution: `cli` executes, `native-app` native-compiles + runs, `native-binary` builds a runnable `dist/<name>`.

Serve and plugin kinds (api-service/fullstack/desktop/mobile) are tested at the dispatch level (plan + delegation); the serve path was additionally smoke-tested by hand against a live server (`GET /openapi.json` → 200).

## Follow-up

Validation surfaced a pre-existing bug (filed separately): `proc_file` does not strip the compound `.sv.jac` / `.cl.jac` suffixes, so `jac run`/`jac start` on a `.sv.jac` entry fail with `No module named 'main.sv'`. That affects the `api-service` serve path for `.sv.jac` entries and is tracked in its own issue.


marsninja:
Follow-up bug surfaced during validation is tracked in #6696 (proc_file does not strip `.sv.jac`/`.cl.jac` compound suffixes).

marsninja:
Thanks @greptile-apps — reviewed all three (addressed in 2af4aef):

1. **`_has_client` third branch (P1)** — partially disagree with the exact suggestion. `config.plugins` is keyed by the *short* plugin name (the whole codebase reads the client plugin as `get_plugin_config('client')` → `[plugins.client]`), so `"client" in config.plugins` is correct; `"jac-client"` would never match. However, investigating it surfaced a real adjacent bug you didn't flag: the **`enabled`** branch (`"jac-client" in config.plugins_config.enabled`) is the dead one — enabled entries are `package:plugin` identifiers like `jac-client:cli` / `jac-client:*`, never the bare `jac-client`. Fixed that branch to match by package prefix, reordered the signals, and added inference tests for all three (dependency, `[plugins.client]`, `enabled`).

2. **`wasm` bypasses the plugin guard (P2)** — partially disagree with treating `wasm` like `desktop`. `desktop` is blocked because its `--client desktop` flag can't be forwarded through the core `start()` function at all; `wasm`/`fullstack` are valid core `start` invocations (the client plugin hooks in during `start`), so blocking them would break the working (plugin-installed) case. But your underlying point about error quality is right — fixed it properly with an up-front `_plugin_available()` check applied uniformly to fullstack/wasm/desktop/mobile, so a missing plugin now fails cleanly instead of deep inside `start`.

3. **Absolute `entry-point` discards `project_root` (P2)** — agree it's an edge case; the behavior is actually intended (an absolute entry-point should be searched at its own location, relative ones stay anchored under the root). Documented it in `resolve_entry`.

marsninja:
Second review round — the bot re-posted the original three (now addressed/answered) and added two new P1s, both of which were real and are fixed in 789671d:

**New, valid — fixed:**
- **`--profile` overlay dropped in project-mode dispatch (P1)** — correct. `_run_project`'s `force_discover` allocated a fresh `JacConfig` and lost the profile the framework had applied. Now preserves `active_profile` and re-applies the overlay after re-discovery, mirroring `_discover_config_from_file`.
- **`[serve] port` ignored (P1)** — correct (I'd flagged this gap myself). The planner now carries a non-default `[serve] port` in the serve plan flags (so it also shows up in `jac run --show`), and `_execute_plan` forwards it to `start`. Validated live: `[serve] port = 8137` serves on 8137, not 8000.

**Re-posts of the first round (positions unchanged):**
- **`_has_client` `"client" in config.plugins` (P1)** — still disagree with the suggested `"jac-client"`. The plugins dict is keyed by the short section name; the codebase reads the client plugin as `get_plugin_config('client')` (`[plugins.client]`), so `"client"` is correct and `"jac-client"` would never match. (The actually-dead branch was `enabled`, already fixed to match the `jac-client:` package prefix.)
- **`wasm` bypasses the plugin guard (P2)** — now handled by the up-front `_plugin_available` check added in the previous commit: a missing required plugin fails cleanly for fullstack/wasm/desktop/mobile alike, while a present plugin still lets `wasm`/`fullstack` flow through `start`.
- **Absolute `entry-point` discards `project_root` (P2)** — documented; the behavior is intended (absolute entry-points resolve at their own location).

marsninja:
Re: `wasm` plan omits `dev=True` — the **code is correct**; the PR description was imprecise (now fixed).

Per the canonical project-kinds spec, the run verbs differ: full-stack is `jac start --dev`, but wasm is plain `jac start`. The quick-guide is explicit that plain `jac start` "compiles the `na` block to `/static/main.wasm` as part of the client build" — `--dev` only adds hot-reload, it is **not** required to emit the WASM. So `wasm` intentionally dispatches `start` (no `--dev`), and `fullstack` dispatches `start --dev`.

I'd grouped them in the description table under `start [--dev]`, which is what triggered this; I've split that row so it matches the code and the spec. No code change.

(If the team would rather `jac run` use `--dev` for *both* client kinds for a consistent local-dev experience, that's a one-liner — but I've kept it faithful to the documented per-kind verbs.)

Issue #6687, opened 2026-06-14
Native: jac run (in-process JIT) segfaults on heavy by-value-struct FFI rendering that the nacompile standalone binary runs fine — AOT/JIT divergence


marsninja:
## Summary

`jac run <file>.na.jac` JIT-executes a native module in-process (compiles to native IR, then calls `jac_entry` through the MCJIT `native_engine` via ctypes — see `cli/commands/impl/execution.impl.jac:119`). For most programs this matches the standalone `jac nacompile` binary, but on a heavy **by-value-struct C-FFI** workload the two diverge: the AOT binary runs correctly while the in-process JIT path **segfaults**.

Concretely, the [quakepassion](https://github.com/marsninja/quakepassion) native engine (branch `na-conversion`) renders the arena and passes its smoke + tests when built with `jac nacompile main.na.jac -o qp && ./qp`, but `jac run main.na.jac` segfaults after a few frames:

```
$ jac run main.na.jac
INFO - Builtins module is already bound, skipping.        # x6 (per-module native setup)
...
Segmentation fault (core dumped)
```

Env: jaclang 0.16.3, jaseci `main @ d002645bc`, Python 3.12.3, x86_64-linux, raylib 6.0.

## Bisection — JIT works for everything except heavy by-value-struct render FFI

All run via `jac run` (in-process JIT). The same programs all run correctly as `jac nacompile` standalone binaries.

| via `jac run` (in-process JIT) | result |
|---|---|
| minimal raylib: `InitWindow` + `ClearBackground(Color)` (one 4-byte by-value struct), 5 frames | ✅ works (screenshot written) |
| minimal OSP: a `walker spawn node` + entry ability (no FFI) | ✅ works |
| full multi-module OSP loop over the real world graph, **headless** (no render FFI) | ✅ works |
| minimal 3D: `BeginMode3D(Camera3D)` (44-byte MEMORY struct, byval) + one `DrawCube(Vector3,…,Color)`, 5 frames | ✅ works |
| **full windowed render of the real arena** — `Renderer.draw_area` looping `DrawCube`+`DrawCubeWires` built from each `Surface`'s `Vec3`/`list[int]` fields, ~70+ by-value-struct FFI calls/frame, **no OSP at all** | ❌ **segfault** |

[...]

Issue #6449, opened 2026-06-06
Native: tuple with an `any`-typed element doesn't round-trip through a list (follow-up to #6435 #2)


marsninja:
## Native: `tuple` with an `any`-typed element doesn't round-trip through a list (the `any` value is lost → 0)

Follow-up to #6435 (issue #2, list[tuple] round-trip). #6438 fixed the **concrete** case (`tuple[int, int]`), but the variant where a tuple element is **`any` holding a boxed value** still loses that element when the tuple is stored in a list and unpacked — it comes back as `0`.

### Minimal repro
```jac
na {
    def helper(loc: any) -> dict[str, any] {
        s: list = [(loc, 0)];
        (cur, phase) = s.pop();
        return {"cur": cur, "phase": phase};   # got cur=0, expected cur=555
    }
    def t() -> dict[str, any] {
        return helper(555);
    }
}
```
Run `t` via the native ctypes harness → `{"cur": 0, "phase": 0}` (expected `{"cur": 555, "phase": 0}`).

By contrast, the concrete form round-trips fine (already fixed in #6438):
```jac
s: list = [(100, 7)];
(a, p) = s.pop();        # a=100, p=7  ✓
```

So the miss is specifically: the tuple's first element is **`any`** (an already-boxed `JacVal`), not a concrete scalar. Boxing/unboxing the tuple appears to drop/zero the `any` member rather than storing its `JacVal` payload through.

### Why it matters (blocks #6146 Phase 4 native OSP)
This is exactly the portable OSP kernel's iterative-DFS pattern in `osp_kernel.jac`:
```jac
stack: list = [(loc, 0)];          # loc is `any` (the current location handle)
while stack {
    (cur, phase) = stack.pop();
    ... rt.tag_of(cur) ...          # cur comes back null -> segfault in tag_of
}
```
With the native OSP lowering in place, `walker spawn node` reaches the kernel, but `tag_of` receives the lost (null) location and segfaults (`__osp_tag_of` crash confirmed by JIT address match, called from `_visit_recursive`). Fixing the `tuple[any,…]` round-trip should unblock the entry-spawn path end-to-end.

### Likely area
The tuple boxing/unboxing path in the native backend — `na_ir_gen_pass.impl/tuples.impl.jac` (`_box_value_tuple` and the tuple element load/store), where an `any`/`JacVal` member needs its 16-byte tagged value stored and reloaded intact (rather than treated like a scalar slot).

Issue #6404, opened 2026-06-05
Native: Python-congruent standard library roadmap (sv↔na parity)


marsninja:
## Summary

Roadmap for a Python-congruent standard library on the native (`na { }` → LLVM) backend. **"Syntax congruent with the sv pathway"** means the *same* `import X` + `X.func(...)` Jac source compiles and runs on both **sv** (CPython bytecode) and **na** (LLVM), so the native side must mirror CPython's surface. The **Congruence** column flags where the *same source* can still diverge in behavior and must be designed against.

Context: this is the umbrella roadmap behind #6353 ("no native math intrinsics") and #6403 (iterator protocol, a prerequisite for `itertools`/lazy iteration).

### Two implementation mechanisms (partition the whole list)

- **A — compiler intrinsic namespace**: libc/libm/syscall-backed; recognized in codegen (like the existing `open` intercept), lowered via `_get_or_declare_extern`, typed by a native-context-gated `.pyi` stub (like `na_builtins.pyi`). Needs per-module compiler work.
- **B — shippable `.na.jac` library**: pure Jac on primitives, compiled/linked like user code. **Zero compiler changes**, automatically Python-faithful. Most of the list lives here.
- **H** = hybrid (thin A core + B wrapper) · **L** = language-level in Jac already · **S** = shim/no-op.

### Congruence legend
🟢 full (API + behavior; floats within ULP) · 🟡 documented subset · 🟠 behavioral caveat to design against · 🔴 non-goal (don't implement; document as unsupported on native).

### Recommended v1 core
`math`, `time`, `sys`, `os`, `os.path`, `random`, `json` — turns native from "raylib demo with hand-rolled `jcos`" into "you can write real CLI / game / numeric programs."

---

## Baseline (already exists)
| Surface | Mech | Congruence | Notes |
|---|---|---|---|
| builtins (`len`/`print`/`abs`/`round`/`min`/`max`/`sum`/`sorted`/…) | A | 🟢 | implemented |
| `open` + file objects | A | 🟠→🟢 | exists; **fix `open`-returns-`None` → raise `FileNotFoundError`** to be congruent |

## Numeric / math — v1
| Module | Mech | Native lowering | Congruence | Notes |
|---|---|---|---|---|
| `math` | A | libm + `@llvm.*` | 🟢 | ULP-level only |
| `cmath` | A | libm complex | 🟢 | native has `complex` |
[...]

Issue #6366, opened 2026-06-03
Native: pure-Python ELF linker produces a segfaulting aarch64 binary


marsninja:
## Summary

The pure-Python ELF linker advertises aarch64 support (`ArchConfig.from_e_machine(EM_AARCH64)` in `elf_linker.impl.jac`), but a `jac nacompile` aarch64 binary does not run: it `SIGSEGV`s at process entry, even for a trivial program with no C-FFI. The x86_64 path is solid; the aarch64 path appears to have never been exercised end-to-end (no ARM CI/hardware).

This is independent of the C-FFI ABI work in #6361 / #6363. The ABI *codegen* for aarch64 is correct (verified by emitting the same object and linking it with the system cross-linker, then running under qemu, bit-identical to gcc); the problem is purely in the in-house linker's aarch64 output.

## How it was found

Reproduced via cross-compilation on the Phase 2 branch (which adds a `JAC_NATIVE_TARGET` override). On a native aarch64 host the same path would run unconditionally.

```bash
# any .na.jac, no FFI needed
printf 'def add(a: int, b: int) -> int { return a + b; }\nwith entry { x = add(2, 3); }\n' > t.na.jac
JAC_NATIVE_TARGET=aarch64-unknown-linux-gnu jac nacompile t.na.jac -o t_a64
file t_a64        # ELF 64-bit LSB executable, ARM aarch64  (linker produced it)
qemu-aarch64 ./t_a64
# qemu: uncaught target signal 11 (Segmentation fault)
```

## Diagnosis so far

1. **`PT_INTERP` was emitted empty when cross-compiling.** The interpreter is chosen by `os.path.exists(path)` over `arch.interp_paths` on the *build* host (`elf_linker.impl.jac`, "Build interp"). When the target loader is not present on the build host (always true when cross-compiling), the loop matches nothing and `interp_data` stays `b"\x00"`, so the loader path is the empty string and qemu reports `Could not open ''`.
   - The Phase 2 PR includes a small fix for this: fall back to the canonical `interp_paths[0]` for the target arch when none exist locally. After that fix the interpreter is correct (`/lib/ld-linux-aarch64.so.1`).

[...]

PR #6697, opened 2026-06-15
feat(native): lower isinstance on `any` + box `bool` distinctly (#6421)

### LINKED ISSUE #6421: Native: box scalars into any/JacVal on assignment/call + lower isinstance-on-any (Mechanism-B / json prerequisite)
## Summary

The native (`na {}` -> LLVM) backend cannot box a concrete scalar into an `any`/`JacVal` slot on ordinary assignment/call, and cannot lower `isinstance(x, T)` when `x` is `any`. Together these block any code that is **generic over `any`** — most importantly the recursive `dumps`/`loads` of `json`, and the entire **Mechanism-B pure-Jac standard-library tier** (`collections`, `statistics`, `functools`, …) in the stdlib roadmap (#6404).

This surfaced while implementing the native stdlib v1 (#6404 / #6418): `json` is the one v1 module that is fundamentally recursive over heterogeneous `any`, and it is blocked here.

## Gap 1 — no scalar -> `any`/`JacVal` boxing on assignment/call

A bare assignment of a concrete value into an `any` variable fails to compile:

```jac
# any_box.na.jac
with entry {
    y: any = 42;     # boom
}
```
```
✖ Error: Jac compilation failed: cannot store i64 to %"JacVal"*: mismatching types
TypeError: cannot store i64 to %"JacVal"*: mismatching types
```

The same happens when passing a concrete value to a function with an `any` parameter (the call site doesn't box the argument):

```jac
def tag(x: any) -> str { return "got"; }
with entry { print(tag(42)); }   # Type of #1 arg mismatch: %"JacVal" != i64
```

Note the backend **already** knows how to box: `_box_to_jacval` (dicts.impl.jac) boxes `int`/`float`/`bool`/pointer values into a `JacVal {i8 tag, i64 payload}` when building a `dict[*, any]` literal. What's missing is invoking that coercion from the **general** value-coercion path (`_coerce_type` / assignment / call-argument lowering) whenever the target type is `any`/`JacVal` and the source is a concrete scalar or pointer. Symmetrically, reading an `any` back into a concrete-typed slot needs the inverse unbox (tag-checked).

## Gap 2 — `isinstance(x, T)` does not lower for an `any`/`JacVal` operand

Runtime type-dispatch on an `any` value is exactly what a generic serializer needs:

```jac
def describe(x: any) -> str {
    if isinstance(x, bool) { return "bool"; }
    if isinstance(x, int)  { return "int"; }
    if isinstance(x, float){ return "float"; }
    if isinstance(x, str)  { return "str"; }
    if isinstance(x, list) { return "list"; }
    if isinstance(x, dict) { return "dict"; }
    return "other";
}
```

`isinstance` against an `any`/`JacVal` operand should lower to a check on the `JacVal` tag byte. The tag scheme already exists in `_box_to_jacval` (0 = int, 1 = float, 2 = str, 3 = list, 4 = dict, plus the other pointer kinds), so the lowering is a tag-load + compare; no new runtime metadata is needed.

## Why it matters

- **Unblocks `json` (#6404 v1).** With `any` boxing + `isinstance` lowering, `json.dumps`/`loads` become a ~150-line pure-Jac `json.na.jac` (recurse, dispatch on tag, build strings) instead of a large hand-written LLVM serializer.
- **Unblocks the Mechanism-B library tier.** The roadmap's premise is that most of the stdlib ships as pure `.na.jac` on primitives; nearly all of it is generic over `any` and needs this.
- It's the higher-leverage of the two `json` prerequisites (the other being float `repr` shortest-round-trip parity, tracked separately).

## Suggested scope

1. Extend the value-coercion path so `concrete -> any` boxes via the existing `_box_to_jacval`, and `any -> concrete` emits a tag-checked unbox. Cover three sites: local/var assignment, function call arguments, and `return`.
2. Lower `isinstance(any_val, T)` to a `JacVal` tag comparison for the primitive/container tags.
3. Add a tri-backend equivalence fixture (`prim_any.jac`) exercising box/unbox round-trips and `isinstance` dispatch across sv/na.

## Acceptance

```jac
def tag(x: any) -> str {
    if isinstance(x, int)  { return "int"; }
    if isinstance(x, str)  { return "str"; }
    if isinstance(x, list) { return "list"; }
    return "other";
}
with entry {
    y: any = 42;            # boxes
    print(tag(y));          # "int"
    print(tag("hi"));       # "str"
    print(tag([1, 2, 3]));  # "list"
    z: int = y;             # tag-checked unbox -> 42
    print(z);
}
```
compiles and runs on `nacompile` with output matching CPython.

Related: #6404 (Python-congruent native stdlib roadmap), #6418 (stdlib v1).

marsninja:
## Summary

Closes #6421. The native (`na {}` → LLVM) backend could already box scalars into an `any`/`JacVal` (#6449) but could **not** lower `isinstance(x, T)` on an `any` operand, which blocks any code that is *generic over `any`* — the Mechanism-B pure-Jac standard-library tier (`json`, `collections`, `statistics`, …) in the native stdlib roadmap (#6404). This PR lowers `isinstance`-on-`any`, gives `bool` a distinct box tag for CPython-faithful dispatch, and lets an explicit `any` unbox into a concrete slot in native context.

## What changed

- **`isinstance` lowering** (`calls.impl.jac`): `isinstance(x, T)` lowers to a `JacVal` tag comparison for `int`, `bool`, `float`, `str`, `list`, `dict`, `set`, and tuples of those (`isinstance(x, (int, str))`). It reuses the existing `_box_to_jacval` scheme — box once (a no-op for an already-boxed `any`; a constant the optimizer folds for a concrete value), then compare the tag byte. An unsupported type operand (e.g. a user archetype) raises a structured `E5090` rather than miscompiling.
- **Distinct `bool` tag** (`dicts.impl.jac`): `bool` now boxes under its own tag (7) instead of sharing `int`'s tag 0, so `isinstance` can tell them apart, matching CPython's `bool ⊂ int`: `isinstance(True, bool)` and `isinstance(True, int)` are both true, `isinstance(5, bool)` is false. `_unbox_from_jacval` keys off the *target* type, so no unbox change is needed; the cross-backend harness learns tag 7 → Python `bool` (so a boxed `bool` round-trips as `True`/`False`).
- **`any` → concrete in native context** (`type_evaluator.impl.jac`): assigning an explicit `any` into a concrete-typed slot (`z: int = a`) is allowed in native code (the `JacVal` unboxes at the coercion seam), restoring gradual typing congruent with the sv pathway. `_in_native_context()` now also recognizes `na {}` blocks via the canonical `UniNode.in_native_context()` walk, not only `.na.jac` files.
- **Test** (`prim_any.jac` + `test_prim_equivalence.jac`): a tri-backend equivalence fixture (sv↔na enforced via `require=["na"]`) covering box/unbox round-trips, `isinstance` dispatch across all tag types, and the `bool`/`int` subtype congruence.

## Acceptance

The issue's acceptance program compiles and runs on `nacompile` (as a `na {}` block / `.na.jac`), output matching CPython:

```
int
str
list
42
```

## Testing

- `prim_any` tri-backend equivalence: **passes** (py and na produce identical dicts, including `is_int_bool=0`, `is_bool_int=1`, `unbox=42`).
- `test_prim_equivalence`: only the 2 pre-existing, platform-specific failures remain (`prim_time` monotonic, `prim_os` isdir on the macOS `/tmp` symlink); the new `any dispatch` test passes and the `bool` tag change adds no regressions.
- `test_checker_pass`: **225 passed** — the type-checker relaxation is gated to native context and does not affect sv strictness.
- `test_native_gen_pass`: the same 2 environment-specific failures (#6687 clib transitive-import JIT, #6366 aarch64 standalone) are present on clean `main` too — confirmed unrelated.

## Notes / scope

- The auto-promoted plain-`.jac` path (`jac nacompile foo.jac`, experimental) still rejects an explicit `z: int = a` at top-level because type-checking runs before the promotion decision; the genuine native contexts (`.na.jac` and `na {}`) work fully. `isinstance` works in all three.
- Unblocks the `json` prerequisite in #6404; float `repr` shortest-round-trip remains the other tracked prerequisite.


marsninja:
Thanks @greptile — both catches are valid; I reproduced each and pushed fixes in 7896718.

**P1 (mixed-support tuple) — agreed, real silent bug, fixed.** Confirmed: `isinstance(f, (int, Foo))` for a `Foo` instance returned `0` where CPython returns `True`, with no diagnostic — and worse, the single-name `isinstance(f, Foo)` *does* error, so the tuple form was strictly more dangerous. `_emit_isinstance` now tracks unresolved names and raises the structured `E5090` when *any* name in the operand has no native tag, consistent with the single-name path. Added a native-gen regression test (`na_isinstance_mixed_tuple.na.jac`).

**P2 (over-broad `any -> concrete`) — agreed on the risk, fixed at a different layer than suggested.** I hardened the codegen boundary rather than gating the type checker, for two reasons:

1. The type checker can't cleanly separate "representable" from "not": `complex` is pointer-lowered (`JacComplex*`) just like a user `obj`, so at the type level they look the same — encoding the exact LLVM-representable set in the checker would duplicate codegen's lowering knowledge and drift. It would also make `na` stricter than `sv` at type-check time (sv permits the gradual assignment), breaking the congruence this PR is for.
2. The LLVM type *does* carry the distinction, so `_unbox_from_jacval` now emits `E5090` for a target it has no arm for (by-value aggregates) instead of returning raw payload bits. Verified the supported set is unaffected — scalars and all pointer-shaped types (user obj/node, `list`/`dict`/`set`) still round-trip, and the existing any/tuple equivalence fixtures stay green.

**One residual I want on the record (not fixed here):** a *tag-mismatch* unbox — e.g. boxing a `float` into `any` then reading it as a pointer-lowered type (`c: complex = a`) — still produces garbage via the `inttoptr` arm, because the tag is only known at runtime for a true `any`. Catching that needs a runtime tag-checked unbox (a panic-on-bad-cast, à la Go type assertions), which #6421 notes but this PR doesn't implement. I'd rather scope that as a follow-up than half-implement it. Flag if you'd prefer it blocks here.

PR #6715, opened 2026-06-15
fix(native): reject unresolved clib/native type names instead of silent i64 (#6676)

### LINKED ISSUE #6676: Native: clib import decls bypass the type checker, so unstamped C-FFI types silently default to signed i64
## Summary

`import from "lib.so" { def ...; }` declarations (clib / C-FFI) are skipped by
the type checker: `TypeCheckPass` short-circuits on nodes carrying `clib_decls`
(`type_checker_pass.impl.jac:549`, `if nd.clib_decls { return; }`). Every
annotation inside a clib block therefore reaches native codegen with
`.type == None`.

`NaIRGenPass._resolve_jac_type` then falls back: it tries
`_resolve_primitive_name` (a hardcoded name→LLVM map for `int / i32 / f64 /
str / Callable / …`), and when that misses it **silently returns `i64`**
(`na_ir_gen_pass.impl/types.impl.jac`). There is no diagnostic above a
`logging.debug` breadcrumb.

## Why it matters

Any C type whose spelling isn't in the small primitive table — `uint32_t`, a
`short`, a typedef'd handle, an unrecognized struct — is silently treated as a
64-bit **signed** integer at the FFI ABI boundary. No error, no warning above
debug level. Combined with the unconditionally-signed integer comparison path
(filed as a gap addendum on #6353), an unsigned value can be both mis-widened
and mis-compared against zero with no feedback.

This is the shared root cause behind several gaps already catalogued in #6353
(notably its #2 int-return boxing, #3 false `E1051` under a clib import, and #7
struct/int lowering collision): codegen is flying blind on width and signedness
because the typed tree was never populated for these nodes.

## Proper fix

Type-check clib import declarations so `.type` is populated for their
parameter/field annotations. Then:

- `_resolve_primitive_name` and the `i64` catch-all in `_resolve_jac_type`
  become unreachable and can be deleted;
- a genuinely unresolved type can ICE honestly (the typed-tree contract the
  `_lower_type` docstring already claims to enforce);
- signedness/width information becomes available to the comparison and
  ABI-coercion paths.

This is the unitree-as-single-semantic-representation direction tracked in
#6406 / #6417 / #6542 — clib decls are currently the largest hole in that
contract.

## Context

Surfaced while reviewing the CEF desktop target (#6572), whose entire `*.na.jac`
binding is clib imports and therefore runs entirely on this fallback path. As
part of that review the `_lower_type` docstring in
`na_ir_gen_pass.impl/types.impl.jac` was corrected to describe the actual
(silent-`i64`) behavior instead of an aspirational ICE contract, and the
removed debug breadcrumb was restored, both pointing here.

chess10kp:
## Summary

- Reject undefined type names in C-library import declarations and native-context annotations (`.na.jac` / `na {}`) at `jac check` time with E2018, instead of only emitting W2001.
- Replace the native codegen silent `i64` fallback for unlowerable types with honest diagnostics: E2018 for unresolved type names, E5090 for genuine native capability gaps.
- Add guardrail tests covering clib decls, native defs, and a negative control for known types.

Fixes #6676.

## Test plan

- [x] `jac test jac/tests/compiler/passes/native/test_native_feature_guardrail.jac` — new clib/native type resolution cases pass
- [x] Existing native guardrail suite still passes
- [ ] CI green on upstream


chess10kp:
**Follow-up tracked separately → #6739**

While landing the field-type naming fix here, I found the real root cause is a
**front-end gap**, not codegen:

`jac check` (codegen off) does **not** reject an unresolved type name in an
object field declaration in native code:

```jac
obj Handle { has raw: uint32_t; }   # in a .na.jac file
```

It only emits `W2001` (warning), whereas the equivalent clib/native-def
*signature* position is a hard `E2018`. So `_emit_unlowerable_type` in this PR
is currently the **sole** catcher for bad field types — the "near-zero
reachability" assumption from earlier in the review was wrong; the bug was live,
not defensive. (Greptile's Comment 1 was more right than I credited it.)

**Why:** `StaticAnalysisPass._native_type_position` raises `E2018` for unresolved
type names only in clib decls or `name_nd.in_native_context()`. But
`UniNode.in_native_context()` (unitree.impl.jac) only recognizes `NativeBlock`
and native-context `Ability`s — an `obj` is neither, so field type tags never
qualify and the name degrades to `W2001`.

This PR keeps the **short-term** codegen backstop (it names the type honestly now, and the new guardrail locks it in). The **principled** fix — extending native-context detection to object field declarations so `jac check` alone raises `E2018` — is tracked in **#6739** as its own change on a separate branch

PR #6719, opened 2026-06-15
fix(compiler): union loop-body assignments into post-loop type narrowing

kugesan1105:
## Problem

A variable read **after** a loop collapsed to its pre-loop type, dropping every assignment made inside the loop body:

```jac
def t1(items: list[int]) {
    fb: int | str | None = None;
    for it in items {
        if it > 5 { fb = 5; } else { fb = "s"; }
    }
    out = fb;   // hovered as `None` — should be `int | str | None`
}
```

Hovering `out` (or `fb`) on that line reported `NoneType`.

## Root cause

Jac's CFG collapses the loop header and the post-loop join into a **single node** (the `for`/`while`/range-`for` statement): the body tails back-edge into the header, and the statement after the loop takes the header as its sole predecessor. The flow-narrowing walk skipped body back-edges at any loop header (correct for narrowing a `while x:` condition), which also dropped the body's assignments (`fb = 5`, `fb = "s"`) when the post-loop read resolved through that same node — leaving only the pre-loop `fb = None`.

## Fix

At a loop header reached **as a predecessor** (the post-loop / in-body read), compute the type as a **single-pass union** of the pre-loop edge with every body back-edge, so in-loop assignments survive the join (`_loop_join_type`). The body back-edge that recurses back into the same header reads a pre-loop **seed** from `_flow_loop_pending` (a body assignment is a barrier that overwrites the seed anyway), which terminates the walk without iteration.

A read in the loop's **iterable/condition** instead uses pre-loop-only narrowing (back-edges skipped), so guard narrowing on `while x:` and `for x in y` is preserved.

The narrowing walk re-evaluates back-edge assignment RHSs against the seed; diagnostics raised in that window are inference artifacts (the seed is not the real flow state for that read), so they are dropped via a scoped `_narrowing_walk_depth` guard. The main checker walk re-types every such node against its actual flow state and emits any genuine error there — so real loop-body errors are never swallowed (locked by `bug_loop_body_diag_not_swallowed`).

Result: `out` now types as `int | str | None`.

## Scope

This is a **targeted fix**: a single union pass, not a fixed point. It keeps Jac's collapsed loop-header/post-loop-join CFG node rather than adopting Pyright's two-node topology, to limit blast radius across CFG consumers (`jac tool ir cfg`, static analysis, capability checks). Iterating past one pass only changes the result for a monotone recurrence (`x = x.next`), which is out of scope and tracked in #6723.

## Follow-ups

- #6720 — `while`-loop condition narrowing does not reach the loop body (related gap, same subsystem)
- #6722 — give loops a distinct post-loop join CFG node; would let the prebody/seed handling here be simplified further
- #6723 — adopt the incomplete-type model engine-wide, incl. monotone-recurrence narrowing (sequenced after #6722)

## Tests

- LSP **hover test** (`test_postloop_narrowing_hover`) reproducing the exact user scenario: hovering a variable read after a `for`/`while` loop shows the union of the body assignments and the pre-loop type, asserting the precise hovered type.
- `bug_loop_body_diag_not_swallowed` — a genuine error inside a loop body (`.val` on `Node | None`) is still reported despite the narrowing walk re-evaluating that RHS against the seed.
- Regression guards stay green, incl. the cyclic-reassignment `while cur { cur = cur.prev }` case and guard-narrowed reads inside loop bodies.

Suites verified: `test_server` langserve (30), `test_checker_bugs` (45), `test_checker_pass` (225), `test_checker_gaps` (21), `test_cfg_build_pass` (17), `test_checker_production` (10), `test_checker_shortcomings` (9).


kugesan1105:
Scope note: this is a targeted fix. It ports Pyright's incomplete-type fixed point into the evaluator but keeps Jac's collapsed loop-header/post-loop-join CFG node (rather than Pyright's two-node topology), to limit blast radius. Tracked follow-ups:

- #6720 — `while`-loop condition narrowing doesn't reach the loop body (related gap, same subsystem)
- #6722 — give loops a distinct post-loop join CFG node (would let this PR's prebody/join + staging + diag-suppression workarounds be removed)
- #6723 — adopt the incomplete-type model engine-wide (sequenced after #6722)

kugesan1105:
Thanks for the review — addressed all three in f619fe1:

1. **String-based convergence** — replaced `str(new_result) == prev_repr` with a structural `_narrowed_types_equiv` (mutual `assign_type`), so union member reordering across iterations can't mask a fixed point. No longer coupled to `__str__`.

2. **Durable-cache commit bypasses `_flow_cache_put`** — kept the direct durable write (it's intentional: by then depth is decremented, so routing through `_flow_cache_put` would mis-stage an inner loop's final value where the outer loop's per-attempt rollback could discard it) and added a comment explaining exactly that, so the asymmetry isn't a surprise to future changes.

3. **No nested-loop coverage** — added a `for`-inside-`for` case to the fixture and a targeted assertion that both loop levels' assignments (`int` and the inner `str` literal) survive the outer post-loop join, exercising the `outer_staged` staging rollback.

All checker suites still green.

kugesan1105:
Reworked from the fixed-point engine to a **single-pass join** (commit 17c3e68), per review feedback that the iteration was generality the bug doesn't exercise.

**What changed:**
- `_get_loop_header_type` (iterate-to-convergence + staged cache + depth-gated diagnostic kill-switch) → `_loop_join_type`: one `_merge_flow_preds` pass unioning the pre-loop edge with all body back-edges, terminating the cycle via a pre-loop seed in `_flow_loop_pending`.
- **Removed:** `_flow_loop_staging`, `_flow_loop_depth`, `_narrowed_types_equiv`, `_flow_cache_has/get/put`, and the global `emit_diag` suppression. Net -84 lines, 3 fewer mutable fields.

**On the diagnostic concern (the important one):** I verified by instrumentation that the narrowing walk *does* raise diagnostics when it re-evaluates a back-edge assignment's RHS against the pre-loop seed (2 such on the `while cur { cur = cur.prev }` fixture, 0 from the main walk). These are inference artifacts of the seed, not real findings. They're now dropped via a **scoped** `_narrowing_walk_depth` guard set only around the walk (not a global loop-depth side-channel). Crucially, a **genuine** loop-body error is *not* swallowed: the main checker walk re-types every node against real flow state and emits it there. Locked by a new regression test `bug_loop_body_diag_not_swallowed` (asserts `cur.nxt.val` on `Node | None` inside a loop still reports exactly 1 error).

The monotone-recurrence case (`x = x.next` precision) that a fixed point would handle is left to #6723, where it belongs on #6722's topology.

All suites green: checker_bugs 45, langserve 30, checker_gaps 21, cfg_build 17, production 10, shortcomings 9.

marsninja:

PR #6729, opened 2026-06-16
feat(persistence): enforce declared archetype uniqueness to stop concurrent duplicate nodes (#6266)

### LINKED ISSUE #6266: Concurrent walker check-then-create can produce duplicate child nodes

akindu-k:
## The problem (#6266)

When two requests run at the same time and both do "look up a node, create it if it's missing" against the same root, they can **both** pass the existence check and **both** create the node. The result is duplicate child nodes that were supposed to be unique, and every request still returns `200 OK` with no sign anything went wrong.

This happens because nothing enforces "there should only be one of these" at the one place that can settle a race: the database. Writes are staged in memory during the request and flushed at the end, each keyed by a fresh random id, so two concurrent creates simply become two different documents. In production this fires on ordinary concurrent logins (a single login can fan out into several `me` calls across replicas), and it compounds: each duplicate re-triggers downstream side effects, and the duplicated walker can itself create a second duplicated archetype.

## How we fixed it

Uniqueness is a property of the archetype, not of any one caller, so we push enforcement into the backend, the only correct serialization point under concurrency (the same direction taken by the earlier edge-merge fix #5644).

- **Declare uniqueness once.** `register_unique_index(NodeType, fields=[...], scope=...)` declares that an archetype is unique, either `global` (across the whole collection) or `root` (one per owning user root). `fields=[]` means "at most one of this archetype per scope" (the find-or-create singleton case). Root scope reuses the node's existing `data.root`, so no new denormalized field is needed.
- **Enforce it in MongoDB.** Each registered constraint becomes a unique partial index, so the database itself rejects the second write. Duplicates become physically impossible regardless of how many requests or replicas race.
- **Turn the rejection into a clear, typed signal.** The duplicate-key error is translated into a typed `UniqueViolation` instead of a generic failure, and surfaced to the caller as **HTTP 409 Conflict** with the offending archetype, fields, and scope, rather than a misleading `200`.
- **No dangling references on the failure path.** The change set already skips dependents when a write fails in the delete direction; we complete the symmetric create direction, so a failed node create automatically skips its new edge and the edge-list entry that would point at it. Nothing half-written is left behind.

The loser of the race now gets a clear 409 and can retry (the retry reads the winner's node). Making both callers transparently converge on the same node (idempotent find-or-create) is a natural follow-up that needs synchronous-write semantics; it is intentionally out of scope here so this can land incrementally.

Also out of scope for this PR (each its own follow-up): SQLite-backend parity, a declarative `unique` field syntax, a `find_or_create_node` primitive, and a startup gate that refuses to deploy when existing data already violates a newly declared constraint.

## Tests

Added `jac-scale/jac_scale/tests/test_mongo_unique.jac` (real MongoDB via testcontainers), covering:
- the unique index is created from the registry,
- a duplicate root-scoped create raises `UniqueViolation` and only one document persists,
- a global field-scoped conflict carries the offending field and value,
- `apply()` records the violation instead of inserting a duplicate,
- `commit()` raises the violation when two colliding creates land in one unit of work (the exact path that becomes a 409).

Existing memory-hierarchy, change set, and Mongo layer tests pass unchanged.

Closes #6266


akindu-k:
## How to use it + end-to-end verification

### Usage

Uniqueness is opt-in: declare it once per archetype at import, and every code path that creates that node is covered automatically. Using the issue's own `repro.jac`, the only change needed is one registration line:

```jac
import from uuid { uuid4 }
import from jaclang.runtimelib.index_registry { register_unique_index }

node UserProfile {
    has user_id: str = "",
        tier: str = "free";
}

# Declare the constraint once. scope="root" => at most one UserProfile per
# owning user root. Use fields=["email"] for "unique on a field", and
# scope="global" for collection-wide uniqueness.
register_unique_index(UserProfile, scope="root");

walker me {
    can handle with Root entry {
        profiles = [-->(?:UserProfile)];
        if profiles and len(profiles) > 0 {
            report {"existing": profiles[0].user_id};
        } else {
            new = here ++> UserProfile(user_id=str(uuid4()));
            report {"created": new[0].user_id if isinstance(new, list) else new.user_id};
        }
    }
}
```

With that one line, `jac start` builds the MongoDB unique index at boot and concurrent `/walker/me` calls converge to a single profile: the winner returns 200, the losers get HTTP 409 (`unique_violation`) instead of silently creating duplicates.

Other forms:

```jac
register_unique_index(User, fields=["email"], scope="global");  # one email across everyone
register_unique_index(UserCostSettings, scope="root");          # one settings node per root
```

### Verification

I reproduced the issue's exact scenario (4 concurrent requests racing check-then-create against the same root) against a real MongoDB, driving the genuine `ScaleTieredMemory` + commit path with a thread barrier so all four stage their create before any flushes.

**Before (no constraint declared) — the bug:**
```
request 1: committed
request 2: committed
request 3: committed
request 4: committed
UserProfile docs for this root in Mongo: 4      <- N > 1
```

**After (register_unique_index(UserProfile, scope="root")):**
```
request 1: committed
request 2: 409 UniqueViolation: only one UserProfile allowed per root
request 3: 409 UniqueViolation: only one UserProfile allowed per root
request 4: 409 UniqueViolation: only one UserProfile allowed per root
UserProfile docs for this root in Mongo: 1      <- exactly 1
```

The full chain runs for real: registry -> Mongo unique partial index on `(data.root, arch_type)` -> DB rejects the duplicate writes -> typed `UniqueViolation` -> `commit` raises it -> HTTP 409.

The startup guard also verified against live Mongo: with pre-existing duplicates present, the backend refuses to boot rather than come up unprotected:
```
IndexMigrationRequired: unique index uniq_root_UserProfile not enforced
(existing duplicates block the unique index; run a dedupe migration, then redeploy): E11000 duplicate key ...
```

Automated coverage for all of the above lives in `jac-scale/jac_scale/tests/test_mongo_layer123.jac` (six new `unique:` tests, real MongoDB via testcontainers).

akindu-k:
## Execution flow

How a request travels through the system, and where each piece of this PR engages.

### 0. At startup (once per backend connect)

```
MongoBackend.postinit
  -> ensure_indexes
       -> _ensure_unique_indexes(ASCENDING)
            for each spec in get_unique_indexes():     # the registry
              build key:  scope=="root" -> [data.root, arch_type, <fields...>]
                          scope=="global" -> [arch_type, <fields...>]
              collection.create_index(..., unique=True, partialFilterExpression={arch_type})
              on failure -> raise IndexMigrationRequired   # postinit re-raises -> app won't boot
```

After this, the database itself holds the uniqueness rule. Nothing in the hot path has to "remember" to check it.

### 1. During the walker (the check-then-create the user wrote)

```
walker me:
  profiles = [-->(?:UserProfile)]      # read: empty for every racer (writes are deferred)
  here ++> UserProfile(...)            # build node + edge anchors, stage them in L1 memory
```

Two concurrent requests both read empty, both stage a create. Nothing has hit Mongo yet. This is the race window, and it is unchanged by this PR.

### 2. End of request: flush (where the race is settled)

```
spawn_walker_async / _sync         (server.impl.jac)
  -> _finalize_call_response
       -> Jac.acommit()
            -> ScaleTieredMemory.commit()        (the production override)
                 -> _collect_into(changeset)     # turn staged L1 into typed WriteIntents
                 -> l3.apply(changeset)           # <-- the DB write happens here
                 -> if (uv := apply_report.primary_violation()) { raise uv }
```

`apply()` walks the change set in dependency order and writes each intent:

```
MongoBackend.apply
  for intent in changeset.staged():
      if intent.depends_on & poisoned: skip   # see step 4
      try: _apply_one(intent)                  # _write_to_db / _put_node_atomic / put_partial
      except UniqueViolation as uv:            # raised by the write path below
          poisoned.add(intent.anchor.id)
          apply_report.violations.append(uv)   # collected, not thrown
```

Inside the write path, MongoDB rejects the second insert because of the unique index:

```
_write_to_db (and _put_node_atomic / put_partial)
  try: collection.update_one(..., upsert=True)
  except DuplicateKeyError as dke:
      raise _translate_duplicate(dke, anchor)  # raw Mongo error -> typed UniqueViolation
```

So: winner's write succeeds; loser's write hits the index, raises `DuplicateKeyError`, which becomes a `UniqueViolation`, which `apply()` records, which `commit()` re-raises.

### 3. Surfacing it as HTTP 409

```
commit() raises UniqueViolation
  -> propagates out of acommit() / _finalize_call_response
       -> spawn_walker_*/execute_function_* except UniqueViolation:
            ctx.mem.abort()                  # drop the loser's half-built work
            return {error_code:'unique_violation', http_status:409, detail: uv.to_dict()}
  -> serve.endpoints: TransportResponse.fail(http_status=result['http_status'])  # 409 on the wire
```

The winner returns 200 with its profile; each loser returns 409 with the typed body.

### 4. Why no dangling edge is left behind

`here ++> UserProfile` stages three intents: create the node (U), create the edge (E: root->U), and add E to root's edge list. If U's create is rejected, E and the edge-list-add must not happen, or the graph would point at a missing node.

This is handled by the change set's dependency wiring (extended to the create direction in this PR):

```
record_create(edge E):          E.depends_on = {root.id, U.id}
record_edge_delta(root, +E):    root_delta.depends_on = {E.id}

apply():  U create fails -> poison {U}
          E create:        depends_on {U} hits poison -> skipped, poison {E}
          root edge-add:   depends_on {E} hits poison -> skipped
```

Net result for the loser: no node, no edge, no edge-list entry. Nothing dangling. The winner's full node+edge+list-entry lands normally.

### One-line summary of the path

`walker ++>` stages in L1 -> end-of-request `commit` -> `apply` writes to Mongo -> unique index rejects the duplicate -> `DuplicateKeyError` -> `UniqueViolation` -> recorded by `apply`, re-raised by `commit` -> caught at the request boundary -> **HTTP 409**, with the loser's partial work poisoned away so the graph stays consistent.

Issue #6728, opened 2026-06-16
jac-scale microservice deploy unusable for code-injection (jacBuilder) model: no init-tarball path, namespace falls back to default, single-pod localhost/DNS mismatch


udithishanka:
## Summary

jac-scale's microservice mode (`[plugins.scale.microservices]`, `sv import`-discovered topology) works locally and for apps deployed with a pre-built image, but it cannot be used by a code-injection deploy model (the one jacBuilder/jachammer uses for every user app: base image + init-container `wget code.tar.gz` from a code-server, no per-app image build). Three gaps block this. They were found while making a deployed `sv import` app (the `this_is_jac` showcase on jachammer) perform its cross-process `sv->sv` call.

Observed failure (deployed app, single pod, microservices enabled, no `--scale`):

```
Proxy: POST /api/guestbook/function/guestbook_stats -> http://guestbook-service.default.svc.cluster.local:8000/...
HTTP stream-forward error: Cannot connect to host guestbook-service.default.svc.cluster.local:8000 [Name or service not known]
```

## G1 - No code-injection deploy path for the microservice target

The monolith `KubernetesTarget` injects code at runtime: an init container `wget`s `code.tar.gz` from a `{app}-code-server` and untars it into an emptyDir (`jac-scale/jac_scale/targets/kubernetes/kubernetes_target.jac:412-428`), on a `python:3.12-slim` base with runtime `jac install`. No per-app image build.

`KubernetesMicroserviceTarget` has no equivalent. It forces `auto_build=True` (`jac-scale/jac_scale/plugin.jac:179`) and requires either a local Docker build (`jac-scale/jac_scale/microservices/images/auto_build.jac`) or a pre-configured `docker_image_name` / per-service `image_tag` (`targets/kubernetes/microservice/target.jac:62-71`). There is no code-server + init-tarball strategy.

Impact: any platform that deploys via code-injection (jacBuilder/jachammer) cannot deploy a microservice app without standing up per-app image builds.

Suggested fix: add a code-injection deploy strategy to the microservice manifest builder so each sibling Deployment uses a base image + the same `wget code.tar.gz` init pattern + `JAC_SV_NAME=<svc>`, selected when no image is configured.

## G2 - In-cluster namespace falls back to `default`; `POD_NAMESPACE` not injected

[...]

PR #6730, opened 2026-06-16
feat(jac-scale): code-injection deploy for the microservice fleet (G1 of #6728)

udithishanka:
Implements **G1** of #6728: let the K8s microservice fleet deploy via **code-injection** (no per-app image build), so code-sync platforms like jacBuilder/jachammer can deploy `sv import` apps.

## Problem

The `kubernetes-microservice` target only deploys from a built/pushed image (`auto_build` via `docker build`, or a pre-set `docker_image_name`). The monolith `KubernetesTarget`, by contrast, is Dockerless: an init container `wget`s `code.tar.gz` from a `{app}-code-server` into an emptyDir, on a base `python_image` with runtime `jac install`. jacBuilder uses that monolith path for every user app.

Because the microservice target has no code-injection path, a deployed `sv import` app (e.g. the `this_is_jac` showcase) cannot run its cross-process call: the fleet was never provisioned, so `analytics.analyze` resolves a hostname that doesn't exist (`[Errno -2] Name or service not known`).

## Change

Add a code-injection mode, gated by `[plugins.scale.microservices].code_inject = true`. When set, each sibling **and** the gateway:
- boot the base `python_image` (no built image),
- run an init container that fetches the code from a code-server into a shared `/app` volume,
- run an entrypoint of `pip install … && jac install && <JAC_SV_NAME dispatch>` (`jac scale gateway` for the gateway, `jac start <svc>.jac` for siblings).

This produces **real per-sibling Deployments + ClusterIP Services**, so the gateway's `<svc>-service.<ns>.svc.cluster.local` DNS resolves with zero image build. It reuses the monolith target's `_build_volumes_config` (code-server + init + volumes) and `_build_runtime_setup_command` (now takes an optional `start_command`), fanned out per sibling.

### Files
- `abstractions/models/deploy_context.jac` — `DeployContext` gains `volumes` / `volume_mounts` / `image_override` / `entrypoint_command`.
- `targets/kubernetes/kubernetes_target.jac` — `_build_runtime_setup_command` takes an optional `start_command` (default unchanged).
- `targets/kubernetes/microservice/manifest_builder.jac` — pod spec honors the new `DeployContext` overrides (image, entrypoint, volumes/mounts).
- `targets/kubernetes/microservice/target.jac` — `deploy()` code-injection branch + `_code_inject_enabled()`.

## Tests

`tests/test_pod_env.jac` adds two cases (both pass): code-injection yields the base image + code-server init + `/app` volume + dispatch entrypoint; and the default (no code-injection context) keeps the built-image + `JAC_SV_NAME` entrypoint. No existing tests regress.

## Scope / follow-ups

This is G1 only. G2 (`POD_NAMESPACE` injection / namespace fallback) and the jacBuilder consumer changes (use the microservice target, point the shared ingress at `gateway-service`, N-deployment status tracking) are tracked in #6728 and jaseci-labs/jacBuilder#537.


marsninja:

PR #6740, opened 2026-06-16
feat(graphmend): add --graphmend flag and graph-break detection

savini98:
## Summary

First slice of **GraphMend** — a compiler technique that eliminates PyTorch 2
FX **graph breaks** by analyzing and rewriting break-causing source constructs
before bytecode generation, so TorchDynamo can capture larger, uninterrupted FX
graphs.

To keep review tractable, GraphMend is being landed as a series of small,
independent PRs. **This PR is the analysis foundation only** — it adds the
opt-in flag and the graph-break *detection* pass. **No code is transformed yet**;
the three transformation passes (trap lowering, predicated control flow,
deferred side effects) that consume these tags land in follow-up PRs.

## What this PR adds

- **`CompileOptions.graphmend`** flag (default `False`).
- **`GraphBreakDetectPass`** — locates `@torch.compile` entry points (decorated
  functions/`nn.Module`s, or call-site `torch.compile(...)` wraps) and tags the
  fixable graph-break candidates inside them via use-def data-dependence
  analysis over the symbol table:
  - `val_guard` — data-dependent `if not C: raise` guards
  - `dyn_ctrl_fl` — data-dependent `if/else`
  - `side_effect` — `print` / logger calls
- **`torch_attr_table`** — the data-dependent torch attributes + side-effect call
  names that drive detection.
- Scheduled in `get_py_code_gen` just before `PyastGenPass`.

## Impact / safety

- **Zero behavior change when `--graphmend` is off.** The pass `prune`s
  immediately unless the flag is set *and* a `torch.compile` entry point exists,
  so default compilation is untouched.
- Detection-only: it tags nodes; nothing is rewritten in this PR.

## Testing

- `test_graphmend_detect` + fixtures verify each tag (`val_guard` /
  `dyn_ctrl_fl` / `side_effect`) and that static control flow (e.g. `hasattr`)
  is **not** tagged.
- Full `jac/tests/compiler` suite passes (the only local failures are
  `node`/ELF platform tests unrelated to this change).

## Notes

- The pass follows the codebase convention: declaration (`graph_break_detect_pass.jac`)
  + implementation annex (`impl/graph_break_detect_pass.impl.jac`).
- No new `.py` files (namespace package, matching sibling pass dirs).

## Roadmap (follow-up PRs)

2. Predicated trap lowering (validation guards → `torch._assert_async`)
3. Deferred side effects (`print`/logger → buffered epilogue flush)
4. Predicated data-dependent control flow (`torch.where` / `torch.cond`)
5. Real-model enablement (CLI `--graphmend-scope` + scoped import + caching)
6. Evaluation harness


savini98:
@greptile-apps reevaluate please?

savini98:
This PR is intentionally **detection-only**: it attaches inert `_gm_break_kind` markers and rewrites nothing.

**Receiver-type / over-tagging** (`_expr_uses_dynamic_attr`, `LOG_METHODS`): matching on the method name without a receiver-type guard is deliberate. GraphMend's core job is catching `x.sum()` / `x.max()` on a **tensor parameter**, where no static type info exists — restricting to `receiver == torch` would miss exactly the cases the technique targets. Since nothing is rewritten here, an over-tag is inert; the transformation passes that consume these markers re-check legality (tensor-valued branches, receiver types, etc.) **before any rewrite, and they land in follow-up PRs**.

**Type annotations** (`cur`, `_is_validation_guard`): `ElseIf` is a subclass of `IfStmt` in the UniIR (`issubclass(uni.ElseIf, uni.IfStmt) == True`), so `cur: IfStmt | None` and `_is_validation_guard(nd: IfStmt)` accept `ElseIf` values without a diagnostic. The remaining type-checker notes are the inert `_gm_break_kind` analysis markers (consumed by the future passes) and pre-existing `get_all_sub_nodes` stub typing — neither introduced by this PR.

**elif detection**: fixed — `enter_if_stmt` now walks the full `else_body` chain from the leading `if`, so a data-dependent `elif` under a static `if` is tagged (and a guard-style `elif` is still classified `val_guard`). Added tests for the elif chain and the `@torch.compile`-decorated `nn.Module` Archetype path.

savini98:
**`_var_assignment` last-definition-only** — acknowledged, deferring to a follow-up PR rather than fixing in this slice. Rationale:

- It's a detection *false-negative* with **no correctness impact** (a missed optimization, never wrong output), and only on a narrow pattern.
- For sequential reassignment (`m = Model(); m = wrap(m); torch.compile(m)`), last-definition is actually the *correct* reaching definition — `m` holds `wrap(m)`, not the bare `Model`. The only genuine miss is a *branch-assigned* compile target (`if c { m = ModelA() } else { m = ModelB() }; torch.compile(m)`), which is rare in real PyTorch code.
- Catching both arms requires `_resolve_wrapped_entry` to return *multiple* entries — a signature change that ripples into `_collect_entry_points`. That's better motivated and tested once the transformation passes (PR 2+) define the multi-entry consumption, rather than speculatively here in the detection-only slice.

Tracking it for the transform PRs. The related keyword-arg entry-point miss (`torch.compile(model=...)`) is fixed in this PR.

savini98:
Addressed the one new item and the rest are settled:

- **`except ImportError` fallback** (compiler.jac): fixed — the GraphMend import inside the fallback is now guarded by its own `try/except`, so a GraphMend import failure degrades to `[PyastGenPass, PyBytecodeGenPass]` rather than breaking Python-only compilation.
- **Multi-statement guard misclassification**: already resolved — `_is_validation_guard` keys on the *last* statement being a `raise` (not a single-statement body), so `if not C { cleanup(); raise }` stays `val_guard` (`guard_multi_stmt.jac` covers it).
- **elif-raise handling**: resolved — a `raise` inside an `elif` is left untagged (it's neither a standalone guard nor value-predicable); `elif_guard.jac` covers it.
- **Receiver-type on attribute calls**: intentional. Matching `x.sum()` on tensor parameters is GraphMend's core capability and no static type info exists; PR1 tags are inert and the transformation passes re-check tensor/receiver legality before any rewrite. This is detection-only — nothing is transformed in this PR.

The remaining classification nuances apply only once the transformation passes (PR 2+) consume the tags, and will be validated there.

savini98:
@greptile-apps can you recheck?

savini98:
@greptile-apps can you recheck?

savini98:
### Re: type-annotation mismatch (`_is_validation_guard` / `_body_ends_in_raise` / `cur`)

This isn't a real type error — `uni.ElseIf` **is a subclass of** `uni.IfStmt`, so passing `ElseIf` nodes to helpers annotated `uni.IfStmt` is ordinary subtyping, not a mismatch.

**Verified directly:**

```python
issubclass(uni.ElseIf, uni.IfStmt)   # True

marsninja:

Issue #6752, opened 2026-06-17
Native: type `None`/null contextually (null-of-expected-pointer-type) instead of defaulting to `i8*`


marsninja:
## Summary

A language/backend **design improvement** surfaced while fixing #6751 (`(bytes | None) == bytes` crashes/yields no value). The native backend gives most reference types a distinct typed pointer (`bytes` -> `%"jacbytes"*`, list/dict/set/objects -> their own struct pointers) and **dispatches on the LLVM pointee type** as a proxy for "what Jac kind is this value." The one value that breaks this is `None`/null: a `None` literal lowers to a blanket `i8*` null (`na_ir_gen_pass.impl/expr.impl.jac`, `isinstance(nd, uni.Null) -> Constant(i8*, None)`) regardless of the static type it stands in for.

That mismatch is the root of a whole class of `T | None` bugs: when a `None`-valued operand of static type `bytes | None` (or any `T | None`) meets a typed-pointer operand, the LLVM pointer types don't agree, so the type-driven dispatch is skipped and either nothing emits a value (silent) or an `icmp`/deref hits the wrong type. #6751 is the `bytes` instance; the same shape can recur for any nullable reference type.

## Proposal

Lower `None`/null **contextually**: as the null of the expected pointer type, driven by the checker's static type, falling back to `i8*` only when the expression is genuinely untyped.

- `None` in a `bytes | None` slot -> null `%"jacbytes"*`
- `None` in a `list[T] | None` slot -> null list-struct pointer
- etc.

The `is` / `is not` handlers already approximate this at the comparison site (they `bitcast` one operand to the other's pointer type before `icmp` — which is exactly why `x is None` works while `x == None` does not). Generalizing that one move at the point where `None`/`uni.Null` is lowered (using the checked type) makes nullable values uniformly "a possibly-null typed pointer," so:

- type-driven dispatch (`_is_bytes_value`, list/dict/set detection, etc.) sees the right type even when the value is `None`;
- `==`/`!=`/ordering/`in`/member-access against a nullable just need a single runtime null guard, instead of per-type special cases;
- the `T | None` comparison bug class is retired rather than patched per-type.

## Why a separate issue

[...]

Issue #6749, opened 2026-06-16
Native: bytes `%jacbytes*` migration is partial — repeat/in/iter/bool/str/dict-key/ordering still assume i8* (follow-up to #6745)


marsninja:
## Summary

Follow-up to #6743 → #6745. #6745 correctly fixed the two reported defects (`bytes + bytes` concat crash, `bytes([ints])` wrong length) — both verified fixed cold (`b"x"+b"y"` → 2, `bytes([65,0,66])` → 3). But auditing the **full** bytes-operation surface on the native backend shows the `i8*` → length-aware `%"jacbytes"*` migration (#6727) is **systematically partial**: seven more bytes operations still assume the old `i8*` representation or were never implemented for the struct. They share one root cause and are best closed together.

Environment: jaclang 0.16.4, jaseci `main @ 5e05470db`, Python 3.12.3, x86_64-linux. Each row verified `na` vs the `@sv`/CPython baseline.

## Root cause (one systemic gap)

`#6727` made `bytes` a length-aware struct pointer `%"jacbytes"*` (`{ i64 len, [N x i8] data }`) instead of a NUL-terminated `i8*`, but only migrated the operations its tests exercised (`len`, index, slice, `==`/`!=`, encode/decode, `int.to/from_bytes`, `struct`). `#6745` added concat + the `bytes()`/`bytearray()` ctors. **Every other bytes operation was left on the old assumption**, and each fails in one of three shapes:

- **hard `i8*` match → skipped → crash / fall-through**: repeat `*`, ordering `< > <= >=`
- **"any pointer" match → wrong `str` primitive applied**: `in` (→`strstr`), truthiness (→null test), `str()` (→bitcast-to-`i8*`), dict-key (→pointer identity)
- **no branch at all → compile failure**: `for x in bytes` iteration

All paths below are under `jac/jaclang/compiler/passes/native/`.

## The seven gaps

| # | Op | `@sv` | `na` | Site | Failure |
|---|---|---|---|---|---|
| 1 | `b"ab" * 3` (repeat) | `6` | crash | `na_ir_gen_pass.impl/types.impl.jac:906` (STAR_MUL matches `i8*` only) | codegen `TypeError: i64 != %"jacbytes"*` |
| 2 | `b"bc" in b"abcd"` (membership) | `1` | crash | `na_ir_gen_pass.impl/expr.impl.jac:817` (calls `strstr` on any two pointers) | codegen `i8* != %"jacbytes"*` |
[...]

PR #6753, opened 2026-06-17
fix(native): null-guard `bytes` comparison for `(bytes | None)` operands (#6751)

### LINKED ISSUE #6751: Native: `(bytes | None) == bytes` segfaults / yields no value when the operand is None (follow-up to #6749)
## Summary

Follow-up to #6749 (which completed the `bytes` `%"jacbytes"*` op migration). One more case in the same family: **comparing a `bytes | None` value to `bytes` when the value is `None`** is broken on the native backend. It works on `@sv` (`None == b"x"` → `False`), so it's an sv↔na congruence defect — and it crashes.

Two symptoms depending on how the `None` is produced:
- a **branchy** `bytes | None` function (returns `bytes` on one path) → the `None` is a null `%"jacbytes"*` → **SIGSEGV**
- a **constant-`None`** function → the comparison produces **no value** (silent empty output)

Both diverge from `@sv`. `is None` / `is not None` work correctly, and `bytes == bytes` (non-null) works — this is specifically the null/`None` operand of a union.

Environment: jaclang 0.16.4, jaseci `main @ 947e6dbd3` (has #6727/#6745/#6749), Python 3.12.3, x86_64-linux.

## Repros

```jac
// bytes_none_eq.na.jac  — branchy union → SIGSEGV
def get(h: bool) -> bytes | None {
    if h { return b"AB"; }
    return None;
}
with entry {
    z = get(False);                 // value None, static type bytes|None
    print(1 if z == b"AB" else 0);  // @sv: 0  | na: SIGSEGV
}
```

```jac
// bytes_none_eq2.na.jac  — constant-None → silent no output
def g() -> bytes | None { return None; }
with entry {
    z = g();
    print(1 if z == b"AB" else 0);  // @sv: 0  | na: prints nothing
    print(1 if z != b"AB" else 0);  // @sv: 1  | na: prints nothing
}
```

| repro | @sv | na |
|---|---|---|
| `get(False) == b"AB"` (branchy) | `0` | **SIGSEGV** |
| `g() == b"AB"` (const-None) | `0` | empty |
| `g() != b"AB"` (const-None) | `1` | empty |
| `g() is None` (control) | `True` | `True` ✓ |

Affects `==` and `!=`, both operand orders.

## Root cause

The pointer-comparison branch in `_emit_binary_op` (`jac/jaclang/compiler/passes/native/na_ir_gen_pass.impl/types.impl.jac`, the `if self._is_bytes_value(left) and self._is_bytes_value(right)` block added by #6749) dispatches to `_emit_bytes_eq` whenever both operands are typed `%"jacbytes"*`. `_emit_bytes_eq` (`na_ir_gen_pass.impl/bytes.impl.jac:217`) then immediately dereferences both operands with **no null guard**:

```jac
llen = self._bytes_len_val(left);    // bytes.impl.jac:224 — loads i64 len field
rlen = self._bytes_len_val(right);   // :225
```

When an operand's static type is `bytes | None` and its runtime value is `None` (a null `%"jacbytes"*`), `_bytes_len_val` loads from a null pointer → SIGSEGV. (`_emit_bytes_cmp` for ordering has the same exposure.) In the constant-`None` variant the value is typed as `None` rather than `%"jacbytes"*`, so `_is_bytes_value` is false, the bytes branch is skipped, and nothing downstream produces a value for the comparison → silent empty.

## Suggested fix

Null-guard the bytes comparison so a `None`/null operand follows CPython identity-ish semantics instead of dereferencing:
- in `_emit_bytes_eq` (and `_emit_bytes_cmp`): if either operand may be null (union `bytes|None`) test for null first — `==`: result is `lnull && rnull` when either is null (i.e. equal only if both `None`); `!=`: the negation. Only call `_bytes_len_val`/`_bytes_data_ptr` on the non-null path.
- ensure the comparison always yields an `i1` value even when one side is statically `None`-typed (the const-`None` empty-output case), so `==`/`!=` against `None` lower to a constant `False`/`True`.

This mirrors how `None == <obj>` is handled for other reference types. Repros saved at `~/jac-na-osp-repros/bytes_none_eq{,2}.{na.jac,jac}`. Happy to test a fix.

marsninja:
## Summary

Closes #6751. Follow-up to #6749/#6750. Comparing a `bytes | None` value to `bytes` when the value is `None` was broken on the native backend (it works on `@sv`: `None == b"x"` is `False`):

| repro | `@sv` | na (before) | na (after) |
|---|---|---|---|
| `get(False) == b"AB"` (branchy union) | `0` | **SIGSEGV** | `0` |
| `g() == b"AB"` (constant `None`) | `0` | empty (no output) | `0` |
| `g() != b"AB"` (constant `None`) | `1` | empty | `1` |

## Root cause

Both `None` paths already lower the `bytes | None` union to a **null `%"jacbytes"*`** (the union-return path coerces `None` to the member pointer type), so both comparison operands are `%"jacbytes"*` and the #6749 bytes-comparison branch fires correctly. The bug was narrower than it looked: `_emit_bytes_eq` / `_emit_bytes_cmp` loaded the `len`/`data` fields of both operands with **no null guard**. Dereferencing the null operand:
- in a branchy union → runtime **SIGSEGV**;
- in a constant-`None` function → a null load is UB that the optimizer deletes, taking the comparison and its `print` with it → **silent empty output**.

(Confirmed by instrumenting the dispatch: both operands report `%"jacbytes"*` in both repros — it is not a typing mismatch.)

## Fix

Null-guard the two comparison helpers (`bytes.impl.jac`); a `None`/null operand follows CPython identity semantics instead of dereferencing:

- **`_emit_bytes_eq`**: test both operands for null first. Equal iff **both** null (`None == None` → `True`, `None == b"x"` → `False`); the `len`/`data` load happens only on the both-non-null path. (Also switched the data check to `memcmp` over `min(len)` so the value path needs no inner branch.)
- **`_emit_bytes_cmp`** (ordering `< > <= >=`): a `None` operand **raises `TypeError`** (CPython parity: ordering is unsupported between `bytes` and `NoneType`), which also avoids the deref.

This is purely a missing null guard — no typing/dispatch change. The broader "a `None` literal still defaults to `i8*` in genuinely-untyped contexts" design point is tracked separately in #6752.

## Tests

Adds a `bytes vs None comparison is null-safe (#6751)` group to `test_native_bytes_struct.jac`: branchy + constant `None`, both operand orders, both-`None` (equal), and non-null operands still comparing by value. Full file 52/52 green; each repro verified cold via `jac nacompile` (and ordering-vs-`None` now aborts via `TypeError` instead of segfaulting).


marsninja:
@greptileai recheck this

PR #6765, opened 2026-06-17
fix(jac-scale): default gateway min_replicas to 2 to prevent 503 outage on pod restart

### LINKED ISSUE #6762: Gateway defaults to min_replicas=1, causing ~5 minute 503 outage on pod restart
## Problem

Users experience intermittent 503 (Service Unavailable) errors that start suddenly and persist for ~5 minutes before disappearing. The errors resolve once the gateway pod is back up.

## Root Cause

The gateway defaults to `min=1` replica. When the single gateway pod dies, there is nothing to serve traffic. A replacement pod is scheduled immediately, but the readiness probe blocks it from receiving traffic for 5 minutes:

```toml
# jac.toml
readiness_initial_delay = 300
```

This creates a guaranteed ~5 minute outage window every time the gateway pod restarts.

The default comes from `manifest_builder.jac:619`:

```jac
"min": int(hpa.get("min", 1)),
```

`_service_config("__gateway__")` reads from `[plugins.scale.microservices.services.__gateway__]` in `jac.toml`. Since no such block exists in the app config, `hpa` resolves to `{}` and `min` falls back to the hardcoded `1`.

The global `min_replicas = 3` under `[plugins.scale.kubernetes]` does not reach this code path — it is used for monolith mode only.

**Result:** every service has explicit redundancy except the gateway — the single entry point for all external traffic.

| Service | min | max | Source |
|---|---|---|---|
| `builder_sv` | 3 | 5 | explicit in `jac.toml` |
| `jac_coder_sv` | 2 | 5 | explicit in `jac.toml` |
| **gateway** | **1** | **3** | hardcoded default — no entry in `jac.toml` |

## Impact

- All external traffic is down for ~5 minutes every time the gateway pod restarts
- 503 errors affect all users simultaneously regardless of which backend service they are using
- The gateway is a hard SPOF: backend services (`builder_sv`, `jac_coder_sv`) have redundancy; the gateway does not

### Option 1: `jac.toml` (per-app)

Add to `jac.toml`:

```toml
[plugins.scale.microservices.services.__gateway__.hpa]
min = 2
```

**Pros:** No code change required. Immediate.  
**Cons:** Per-app only. Every new jac-scale app still deploys with `min=1` by default. Relies on developers knowing to add this block — the problem recurs on any new deployment that omits it.

---

### Option 2: `manifest_builder.jac` (framework default) ✅ Recommended

**File:** `jac-scale/jac_scale/targets/kubernetes/microservice/manifest_builder.jac`  
**Function:** `_get_autoscaler_config` (line 619)

Change:
```jac
"min": int(hpa.get("min", 1)),
```

To:
```jac
default_min = 2 if svc_name == GATEWAY_NAME else 1;
"min": int(hpa.get("min", default_min)),
```

**Pros:** Framework-level fix. Every jac-scale app gets `min=2` for the gateway automatically. Developers can still override to `min=1` via `jac.toml` if needed. Backend services keep `min=1` default — no resource impact on dev environments.  
**Cons:** Requires a code change and release.

## Recommendation

**Option 2.**

The root cause is that the framework treats the gateway identically to backend services, defaulting `min=1` for both. The gateway is architecturally different — it is the only component all external traffic passes through. The correct fix is at the framework level so the safe default is enforced automatically. Option 1 fixes one app; Option 2 fixes all apps permanently and eliminates the dependency on developer knowledge.

## Outcome After Fix

With `min=2`:

- One gateway pod dies → surviving pod absorbs all traffic immediately → **no 503s**
- Replacement pod initializes (300s readiness delay) → temporary reduced redundancy, not downtime
- Both pods dying within 5 minutes of each other is the only remaining risk — unlikely in normal operation

Developer-Linus:
Closes #6762 

## Problem

Users experienced intermittent 503 (Service Unavailable) errors that started suddenly and persisted for ~5 minutes before disappearing. The errors resolved once the gateway pod was back up.

The root cause: `_get_autoscaler_config` in `manifest_builder.jac` defaulted `min=1` for all services uniformly, including the gateway. With a single gateway pod, any pod restart left no replica to serve traffic during the `readiness_initial_delay` window (300s in production). This created a guaranteed ~5 minute outage every time the gateway pod restarted.

Backend services (`builder_sv`, `jac_coder_sv`) had explicit `min` values in `jac.toml` that provided redundancy. The gateway had no such entry and silently fell through to the hardcoded `1`.

### `manifest_builder.jac`

Changed the default `min` in `_get_autoscaler_config` to be gateway-aware:

```jac
# before
"min": int(hpa.get("min", 1)),

# after
"min": int(hpa.get("min", 2 if svc_name == GATEWAY_NAME else 1)),
```

The gateway is the sole entry point for all external traffic. Defaulting to `min=2` ensures one pod dying never takes it fully down — the surviving pod absorbs traffic immediately while the replacement initialises. Backend services keep `min=1` as their default, preserving existing behaviour and dev-environment resource usage.

Per-app override is still fully respected: adding `[plugins.scale.microservices.services.__gateway__.hpa] min = N` in `jac.toml` takes precedence.

The pattern follows the existing convention in `manifest_builder.jac` where the gateway receives different defaults from backend services (e.g. `"2Gi" if is_gateway else "1Gi"` for memory limit).

### `test_pod_env.jac`

Added three tests to the `MicroserviceManifestBuilder` unit test file:

- **Gateway defaults to `min=2`** when no `__gateway__` config is present
- **Regular service defaults to `min=1`** — existing behaviour unchanged
- **Explicit `min` override is respected** for the gateway — `jac.toml` still wins

## Why not fix it in `jac.toml`?

A per-app `jac.toml` fix would only address one deployment. Every new jac-scale app would still deploy with `min=1` for the gateway by default, and the problem would recur the moment someone deploys without the `__gateway__` block. The framework fix makes the safe default universal.

## Behaviour after this change

| Scenario | Before | After |
|---|---|---|
| Gateway pod restarts | ~5 min 503 outage | Surviving pod absorbs traffic, no outage |
| Rolling deploy | Zero-downtime (unchanged) | Zero-downtime (unchanged) |
| KEDA scale-down | Can reach 1 replica | Bounded at 2 replicas |
| Explicit `min=1` override in `jac.toml` | N/A | Respected — override still works |


Developer-Linus:
Already addressed.

Issue #6027, opened 2026-05-19
fix(persistence): _put_node_atomic writes stale non_arch_top fields (access clobber gap)


akindu-k:
## Background

PR #6021 fixed the `_put_node_atomic` scalar-clobber bug by replacing the shallow `$mergeObjects` archetype write with per-field dot-notation writes for only the dirty archetype fields.

## Gap

`_put_node_atomic` still writes all top-level non-archetype data fields (e.g. `data.access`) unconditionally from the committing pod's snapshot:

\`\`\`jac
full = dict(Serializer.serialize(anchor, include_type=True));
non_arch_top = {k: v for (k, v) in full.items() if k not in ('edges', 'archetype')};
...
for (k, v) in non_arch_top.items() {
    set_doc[f"data.{k}"] = v;
}
\`\`\`

Under concurrent access-control mutations on the same node across pods, this is the same class of stale-snapshot clobber that #6021 fixed for archetype scalars. If Pod1 updates `access` and Pod2 then triggers `_put_node_atomic` (edge change) with a stale snapshot, Pod1's access change is silently reverted.

## Severity

Low in practice — access-control mutations are rare and unlikely to race with edge mutations — but the code path is incorrect by the same logic that motivated #6021.

## Fix

Extend the dirty-field approach to top-level data fields: compute which non-archetype top-level fields have actually changed (using hash comparison or `__jac_dirty_fields__`) and write only `data.<field>` keys that are dirty, rather than all of them unconditionally.

marsninja:
Closing: no activity in 10+ days. Reopen if still relevant.

PR #6771, opened 2026-06-18
feat(persistence): optimistic-concurrency walker transactions to stop duplicate child nodes (#6266)

marsninja:
## Problem (#6266)

Two concurrent walker invocations that do "look up a node, create it if missing" against the same root can both pass the existence check and both create a child node. The result is duplicate child nodes that were meant to be unique, and every request still returns 200 with no sign anything went wrong. This is the node-identity analog of the edge-merge race (#5446, closed by #5644); same diagnosis, different primitive.

## Approach: a walker is a transaction over its root

Instead of pushing a uniqueness constraint into the database, this makes the runtime itself detect the race at commit and converge. There is no DB-level unique index and no app-declared constraint.

- **OCC version-CAS.** `NodeAnchor` carries a `version`, round-tripped by the serializer. `MongoBackend._put_node_atomic` CASes on it when flushing edge-list changes (`{_id, data.version: captured}` + `$inc`, `upsert=False`, with an absent-doc fallback that preserves pre-OCC behavior). A concurrent committer that bumped the version makes the second commit raise `WriteConflict` instead of silently writing a duplicate. #5644's `$setUnion` edge-merge stays inside the CAS.
- **Walker replay-to-converge.** The request boundary (`spawn_walker_sync` / `spawn_walker_async`) wraps spawn+commit in a bounded retry: on `WriteConflict` it aborts the unit of work, clears deferred effects, reloads the root, and replays the walker. The loser re-reads the now-current graph and takes the find branch, so find-or-create converges on the winner's node. No 409 in the common case; a typed 409 is returned only on retry exhaustion.
- **`on_commit` ambient builtin.** Registers deferred external effects that run only after a successful commit and are discarded on abort/replay, so a retried walker does not double-fire side effects (signup bonus, ws-token, etc.). Callable with no import.
- **Conflict plumbing.** `ApplyReport.conflicts` / `primary_conflict`; `apply()` records the conflict and poisons dependents; `commit()` (core `TieredMemory` and `ScaleTieredMemory`) raises it. Create-direction `depends_on` wiring was added to the ChangeSet so a failed node create cannot mint a dangling edge (an integrity gap that existed independently of this race).

## Why not a DB unique index / declaration

An earlier draft (PR #6729) used a Mongo unique partial index plus `register_unique_index(...)`. That works, but it (a) imports a storage concept into the language, (b) needs a per-archetype declaration, and (c) returns 409 to the loser. The OCC approach keeps uniqueness as "a matter of the implementation checking": the runtime serializes on the root's version and converges the loser by replay. All of that index/registry machinery was removed.

## Tests

`jac-scale/jac_scale/tests/test_occ_replay.jac` (real MongoDB via testcontainers, skipped without Docker):

- the version-CAS detects a concurrent root mutation (`WriteConflict` with the expected/actual versions; the stale edge does not land);
- two pods racing find-or-create converge to exactly one linked child (`['created', 'found']`), with one accepted orphan.

Existing `test_mongo_layer123.jac` (17 tests, exercises `_put_node_atomic` / `apply` / `commit` / serializer) passes unchanged.

See the first review comment for the known gaps and follow-ups.


marsninja:
## Known gaps and follow-ups

These are deliberate scope boundaries, not accidents. Each is safe in the sense that it never produces a dangling reference or silent data loss; the cost is extra retries, a collectable orphan, or a path not yet covered. Listed with why it is acceptable for this change and what the follow-up looks like.

### 1. Coarse root-version granularity (extra retries on a contended root)
The CAS is on the root's `version`, which bumps on any edge-list change to that root. So two walkers adding *different* children to the same root both capture the same version, and one will replay even though there is no real duplicate. This is not a correctness loss: the loser replays, re-reads, and both writes land (so #5644's "no lost edge" guarantee is preserved), and the edge-merge `$setUnion` still runs inside the CAS. The cost is that concurrent writes to one root serialize-and-retry instead of merging lock-free.

Why acceptable: in user-root OSP, write concurrency on a single root is low, and the one case that does fan out (a single login firing several `me` calls) is exactly where serializing is *correct*, since that is the duplicate we are preventing.

Where it bites: a high-fan-out shared root (for example an "org" root that many users write to) could see retry amplification, and at the extreme exhaust the retry cap (currently 5, no backoff/jitter) and return a 409.

Follow-up: finer-grained conflict detection. Either a per-subtree / per-edge-type version, or precise read-set OCC (conflict only if the specific traversal the walker read, for example `[-->(?:UserProfile)]`, changed), which needs read-set tracking but removes the false conflicts.

### 2. Accepted orphan on a lost race
The ChangeSet stages writes in order: `NODE_CREATE` (stage 0) before `EDGE_LIST_DELTA` (stage 2, the root CAS). So the loser's new child doc is written *before* its root-CAS fails. That child is then left unreferenced: no edge in `root.data.edges` points at it. It is invisible to traversal and collectable by `jac db fsck`, and it is never a dangling reference, so it is consistent with the ChangeSet invariant ("a truncated flush leaves at worst an orphan, never a dangling reference"). The test asserts this explicitly (1 linked child + 1 orphan).

Why acceptable: the prod symptom of #6266 was N *linked* duplicate profiles, each driving duplicate side effects. After this change there is exactly one linked child; the orphan carries no edges, so it triggers no traversal and no per-duplicate side effects.

Follow-up: a cheap root-version precheck at the start of `apply()` (read the root version before staging writes) so the conflict is detected before the orphan node is written; or stage a compensating `NODE_DELETE` for the just-written orphan on the abort path.

### 3. Function endpoints are not covered by replay
The OCC engine is backend-wide (any `commit` that hits a stale root CAS raises `WriteConflict`), but the replay loop was added only to the walker spawn paths (`spawn_walker_sync` / `spawn_walker_async`). A find-or-create inside a plain `def` endpoint (`execute_function_sync` / `execute_function_async`) currently lets `WriteConflict` fall through the generic handler and surface as a 500 with a traceback, rather than converging or returning a typed 409.

Why acceptable: the documented #6266 case is a walker (the `me` walker in the report). Walkers are where check-then-create lives in practice.

Follow-up: factor the abort + reload + replay loop into a shared helper and apply it to the function paths too (or at minimum translate `WriteConflict` to a typed 409 there, matching the walker exhaustion path).

### 4. on_commit is in-process and post-commit only (no durable outbox)
`on_commit(cb)` registers a callback that runs on the request's process right after the commit flushes, and is discarded on abort/replay. That gives "fire once, only if the transaction commits", which is what stops a retried walker from double-firing side effects. But it is at-most-once: if the process dies between the commit landing and the callback running, the effect is lost. It also runs before the HTTP response envelope is built, so response-coupled effects (those whose result must appear in the response) are not addressed by deferral.

Why acceptable: it is the surface and the correct in-process first step, and it already removes the "duplicate side effects per duplicate node" problem for the converge path.

Follow-up: the durable transactional-outbox design. Persist effect intents in the same commit (as graph state), and have a system/drainer walker execute them at-least-once with runtime-provided idempotency keys, so effects survive a crash. Response-coupled effects need either an idempotent in-body call keyed off a stable id, or a saga-style split.

### 5. system_root is process-shared
The OCC version lives on the node anchor. User roots in jac-scale are loaded request-locally (a fresh `ScaleTieredMemory` per request via `create_j_context`), so each request captures its own version and the CAS is correct both cross-pod and within-pod. The system / shared root (unauthenticated requests), however, is cached process-wide (`_process_cache['system_root']`), so its `version` lives on a single anchor object that concurrent in-process requests mutate. OCC for writes against the shared root is therefore not reliable within one process.

Why acceptable: the #6266 prod case is per-user roots. Writes to the shared root are rare.

Follow-up: stop sharing the mutable version on the cached system_root (load it request-locally too, or keep the cached copy read-only and re-read the version for the CAS).

### 6. Mongo backend only (SQLite/in-memory parity, and write scope)
The version-CAS is implemented in `MongoBackend._put_node_atomic` only. The SQLite and in-memory backends have no CAS, so on local dev (SQLite) the race is not detected and duplicates can still occur locally while prod (Mongo) converges. This is dev/prod skew to be aware of. Separately, the CAS guards only the edge-list/topology path; scalar field updates (`put_partial`) and full writes (`_write_to_db`) remain last-writer-wins, which is fine for the node-identity race (a topology change) but means field-level OCC is a separate item.

Follow-up: add a version check to `SqliteMemory` (a `BEGIN IMMEDIATE` transaction plus a version guard) for parity; consider field-level OCC if scalar lost-update races (for example #6027) are folded in.

marsninja:
## Update: SQLite-backend OCC parity (addresses gap #6, first half)

The dev/prod skew called out in gap #6 (the version-CAS was Mongo-only, so the local/SQLite backend did not detect the duplicate race) is now closed in core `SqliteMemory`:

- `SqliteMemory._merge_write`: on an edge-list (topology) change to a node whose stored version moved since this request read it, raise `WriteConflict` instead of merging a duplicate child; advance the version with the edge-list change and re-baseline the in-memory anchor. This sits inside the existing `BEGIN IMMEDIATE` transaction, so it composes with the cross-process serialization already there.
- `SqliteMemory.apply`: record `WriteConflict` on `ApplyReport.conflicts` (poison dependents, let `commit()` raise it), mirroring `MongoBackend.apply`.

So the same "walker is a transaction over its root, loser converges via replay" behavior now holds on the default local backend, with no Docker required to test it.

New test: `jac/tests/runtimelib/test_occ_sqlite.jac` (two `SqliteMemory` instances on one db file, no testcontainers): the version-CAS detects a stale-write conflict, and a racing find-or-create converges to exactly one linked child.

Still open from gap #6: the in-memory (`VolatileMemory`) backend has no CAS (it is a single-process cache, so the cross-pod race does not apply there), and field-level OCC (scalar last-writer-wins via `put_partial` / `_write_to_db`) remains a separate item. The other gaps (1-5) are unchanged.

marsninja:
## Update: read-gated OCC (addresses gap #1, coarse root-version)

The node-version CAS is now **read-gated**: a commit CASes on a node's version only when this request actually *read* a traversal from that node. This removes the false-conflict surface in gap #1 and restores #5644's lock-free merge for blind appends.

How it works:
- `edges_to_nodes` snapshots the version of every node a request reads an out-traversal from, into request-scoped `ExecutionContext.read_versions`.
- The collect pass copies that onto the edge-list `WriteIntent.cas_version` (None if the node was never read).
- `MongoBackend._put_node_atomic` / `SqliteMemory._merge_write` CAS on the version only when `cas_version` is set; otherwise they just merge + bump. The version is always advanced, so a lock-free append still invalidates a concurrent reader's CAS.

Net effect on gap #1:
- **Blind append now merges lock-free again** — `here ++> Item()` with no preceding `[-->...]` read takes no dependency, so the org-root "many writers, no find-or-create" case stops conflicting entirely. This was the cited "where it bites."
- **Check-then-create still converges** — a walker that reads `[-->(?:X)]` then creates takes a dependency, so a concurrent winner makes it conflict and replay (find branch), exactly as before.

Residual (now the only part of gap #1 left): two *different* find-or-creates on the **same** node still conflict, because node granularity can't tell the read types apart (e.g. a `UserProfile` find-or-create and a `Settings` find-or-create racing on root). It is safe (the loser replays and succeeds), and rarer than the blind-append contention this fixes.

Why not full per-(node, child-type) precision: the traversal's type filter (`(?:UserProfile)`) is compiled to an **opaque `Callable`** on `ObjectSpatialDestination` — there is no structured type to key a per-type version off without a **compiler change** to thread filter types through codegen. That is the natural follow-up to eliminate the residual; flagging it rather than hand-rolling fragile closure introspection.

Tests: read-gated detection + converge **plus a new "blind append merges lock-free" case** for both Mongo (testcontainers) and SQLite (no Docker). Existing mongo layer tests pass (a stale 4-arg `_put_node_atomic` mock was updated to the new `cas_version` signature).

Gap #1 is now closed except for the same-node cross-type residual, which needs the compiler change noted above.

marsninja:
## Update: function endpoints now converge too (addresses gap #3)

Previously the OCC engine detected conflicts on `def`-based endpoints (the loser's write was aborted, so no duplicate persisted) but only the *walker* paths had the replay loop, so a function endpoint that lost a check-then-create race surfaced the `WriteConflict` as a generic 500 instead of converging.

`execute_function_sync` / `execute_function_async` now have the same replay as `spawn_walker_sync` / `spawn_walker_async`: on `WriteConflict` they abort the unit of work, clear deferred (`on_commit`) effects and read-versions, reload the root, and re-run the function. Wire args are deserialized once before the retry loop so a replay doesn't double-deserialize. Retry exhaustion returns a typed 409 (`error_code: write_conflict`).

So both walkers and functions now behave identically: a lost find-or-create converges on the winner instead of erroring. Same caveat as walkers — a replayed function re-executes from scratch, so non-graph side effects should be deferred via `on_commit` (which is cleared on each replay) to stay exactly-once.

marsninja:
## Update: configurable on-conflict policy + docs

**Configurable policy ([serve] in jac.toml).** The retry-vs-fail choice and retry tuning are now operator-configurable, read once per request via the cached `get_config()` and threaded into all four execution paths' replay loops (replacing the hardcoded cap of 5):

```toml
[serve]
on_conflict = "retry"        # "retry" = abort + replay (converge); "fail" = immediate 409
conflict_max_attempts = 5
conflict_backoff_ms = 0      # linear backoff (ms x attempt)
```

Any config-read failure falls back to retry/5/0 (the prior behavior). Per the earlier discussion, this is the app-wide *default* + operational tuning; the per-operation semantic override (a decorator, where replay-safety is actually known) remains the natural follow-up.

**Docs.** Added/updated mkdocs reference pages for everything in this branch:
- `reference/persistence.md` -- new "Concurrent writes: check-then-create and convergence" section: the node-version CAS, read-gating, replay convergence, blind-append lock-free behavior, `on_commit` for replay-safe side effects, the `[serve]` config, and scope/granularity.
- `reference/config/index.md` -- the three new `[serve]` keys.
- `reference/language/osp.md` -- `on_commit` added to the built-in graph functions table, with a cross-link.
- `reference/plugins/jac-scale.md` -- note that cross-pod L1 invalidation is post-commit and the check-then-create race is closed by OCC convergence (cross-link).

`mkdocs build` is clean (no new broken-link warnings).

marsninja:
## Review fixes (commit 4dffbeae4)

Agreed with all five review findings; applied the long-term fixes and added the missing release-note fragments (the CI red was two things: the recursion below, plus the missing release notes).

**P0 -- `_occ_policy` infinite recursion (the CI failures).** Correct and entirely my bug: a `replace_all` that rewired the call sites also rewrote the *one* `max_attempts = 5;` inside `_occ_policy`'s body into a self-call. Restored the function body. This is what was 500'ing every walker/function request (`test-scale`, `test-client`, `test-core-runtime`). `test_serve.jac` (31 tests) now passes.

**P1 -- CAS filter missed legacy documents.** Correct and important for upgrades. A doc written before this feature has no `data.version`, and Mongo equality won't match a missing field, so `{data.version: 0}` 409'd every request on a pre-existing node. Fix: when the request expects v0, match `{$or: [{data.version: 0}, {data.version: {$exists: false}}]}` (treat absent == 0). Added a regression test that strips `data.version` and asserts the read-gated write succeeds and bumps to 1.

**P2 -- TOCTOU in the first-write fallback.** Correct. Removed the read-then-upsert fallback entirely: on a version miss the CAS now always raises `WriteConflict` and the request boundary replays against the now-current graph. An `EDGE_LIST_DELTA` only targets a node the request already loaded, so a genuinely-absent doc is itself a stale read worth converging on -- no upsert race, and no partial (archetype-less) doc minted.

**P2 -- `on_commit` dropped in CLI mode.** Correct. `ExecutionContext.close()` now drains pending effects after `mem.close()`'s commit backstop, so `jac run` / library flows that never call `commit()` explicitly still fire their effects (no-op if an earlier commit already drained; request forks return before this, and are covered by `acommit`). Verified: an effect registered with no explicit commit now fires on exit.

**P1 -- `req_token` leak.** Correct. `_deserialize_wire_args` and `_occ_policy` now run *inside* the outer `try` in all four execution paths, so the `finally` always calls `reset_request_context`.

All OCC + `on_commit` suites are green locally; the new legacy-doc test covers the P1 case.

marsninja:
@greptileai recheck

marsninja:
@greptileai recheck

marsninja:
@greptileai recheck

PR #6772, opened 2026-06-18
feat(jac-scale): --no-image microservice deploy (copy local source, no Docker build/push)

MusabMahmoodh:
## Summary

Adds **`jac start <entry>.jac --scale --no-image`**: deploy a microservice app to Kubernetes **without building or pushing a Docker image**. Brings microservice mode to parity with the monolith's source-clone path for registry-less clusters (roadmap N-22).

This is a development / preview convenience. Two warnings are printed at deploy time (use a real image for production; run `jac setup microservice` to configure for production).

## How it works

Instead of building an image:
- the local source tree is `tar+gz`'d into a **ConfigMap** (`binaryData`, ~1 MiB cap with a fail-fast guard),
- a **bootstrap initContainer** extracts it into a shared `emptyDir` and `pip install`s the **pinned** `jaclang` + `jac-scale` releases (matched to the developer's local versions) into `/app/.pydeps`,
- each service + the gateway run on the generic **`python:3.12-slim`** base, with `PYTHONPATH`/`PATH` pointed at the installed deps and `jac start <svc>.jac` as the entrypoint.

MongoDB/Redis are still provisioned so services share state. The source ConfigMap is applied before Deployments and torn down on `--destroy`.

## What changed

- `microservices/source_bundle.jac` (new): source packing, ConfigMap builder + size guard, version pinning, bootstrap initContainer.
- `targets/kubernetes/microservice/manifest_builder.jac`: no-image pod specs (base image, source/app volumes, env, initContainer) + source ConfigMap in the bundle.
- `targets/kubernetes/microservice/target.jac`: bypass the image pipeline in no-image mode, provision DBs, apply/destroy the source ConfigMap.
- `plugin.jac`: `--no-image` flag + the two production warnings.
- `abstractions/config/app_config.jac`: `no_image` field.

## Scope / follow-up

- The default image-build path and all existing deploys are **unchanged** (no-image is strictly opt-in).
- Production-grade in-cluster image builds (kaniko is deprecated; **buildkit / buildah** candidates) are tracked as the next roadmap item (N-22b).

## Testing

- 8 new unit/integration tests (source packing + exclusions, ConfigMap `binaryData` + size guard, pinned specs, bootstrap initContainer, no-image pod spec, full `generate_manifests` bundle). Regression suite across microservice/K8s manifest tests: **53 passed**.
- Real-cluster e2e (minikube) is pending - not runnable in this environment.

Builds on #6769 (zero-config local mode) but is independent; either can merge first.


MusabMahmoodh:
Superseded by #6776, which combines --no-image with the zero-config auto-detect (local + --scale) so the whole feature reviews as one.

PR #6775, opened 2026-06-18
feat(jac-scale): zero-config local microservice mode (auto-detect, no Docker images)

MusabMahmoodh:
## Summary

Makes **local microservice mode "just work"** like the old monolith DX. `jac start main.jac` on an app that splits itself with `sv import` now auto-detects its services and runs them locally as subprocesses - **no `jac setup`, no Docker images** - and prints a one-line banner pointing to `jac setup microservice` for customization.

This is Part 1 (local) of the zero-config effort. The remote no-image `--scale` deploy is a separate follow-up PR and is intentionally untouched here.

## What changed

- **Service discovery** (`microservices/discovery.jac`): `discover_sv_services()` scans the entry file's directory for real `sv import from <module>` / `to sv:` markers. Comments and docstrings are stripped first, so the example apps' docstrings (which literally contain `sv import from ...`) are never miscounted. Returns default `/api/<name>` routes via the same `derive_prefix` `jac setup` uses.
- **Tri-state `enabled`** (`impl/config_loader.impl.jac`): `get_microservices_config()` now returns `enabled = None` when the key is unset (the `False` default was removed from `get_default_config` so `load()`'s deep-merge no longer erases the distinction). `None` = auto-detect, `False` = explicit opt-out, `True` = explicit opt-in.
- **Auto-enter gate** (`plugin.jac`): `_scale_pre_hook` enters local microservice mode when `enabled` is unset and services are discovered, fills default routes from discovery when none are declared, and prints the auto-detect banner. The pure `should_enter_local_ms()` gate is unit-tested. `sv` sibling children and `--scale` never take this path.
- **Routes optional** (`microservices/orchestrator/orchestrator.jac`): `start_local` falls back to discovery instead of erroring when no routes are declared.

## Behavior

- **Auto-detect app, no config:** spawns each service as a no-Docker subprocess, gateway on `:8000`, sv-to-sv routing works, banner prints.
- **Plain monolith (no `sv import`):** unchanged - never enters microservice mode (verified with an app whose docstring even mentions `sv import`).
- **`enabled = false`:** explicit opt-out is honored.
- **`--scale`:** unchanged; still requires explicit configuration.

## Testing

- 15 new unit tests (discovery scanner, false-positive guard, tri-state gate) + 4 config-loader tri-state tests.
- Full affected suite: **98 passed**.
- Manually verified `jac start main.jac` on the e-commerce example with `[plugins.scale.microservices]` removed: auto-detected products/cart/orders, spawned them, gateway up, routing reached the services.


MusabMahmoodh:
Superseded by #6776, which folds this (local zero-config) together with --scale auto-detect and --no-image into one coherent zero-config feature.

PR #6776, opened 2026-06-18
feat(jac-scale): zero-config microservice mode (local + --scale + --no-image)

MusabMahmoodh:
## Summary

Makes microservice mode **zero-config end to end** - the same `sv import` app runs with no `[plugins.scale.microservices]` block both **locally** (`jac start main.jac`) and **on Kubernetes** (`jac start main.jac --scale`), and can deploy **without building a Docker image** (`--scale --no-image`).

Supersedes #6775 (local zero-config) and #6772 (no-image); combined here so the auto-detect gate, route propagation, and no-image deploy land as one coherent feature.

## What it does

- **Auto-detection** (`discovery.jac`): scans the entry file's directory for real `sv import from <module>` / `to sv:` markers (comments + docstrings stripped, so docs that mention `sv import` never miscount) and derives default `/api/<name>` routes.
- **Tri-state `enabled`**: `true` = explicit opt-in, `false` = opt-out (stays monolith), unset = auto-detect. `jac scale status/stop/restart/logs/destroy` also work in auto-detected mode.
- **Local** (`jac start main.jac`): auto-detected services spawn as subprocesses behind the gateway on `:8000`, with an info banner.
- **`--scale`**: auto-detect selects the `kubernetes-microservice` target; discovered routes are injected into the deploy config and **propagated to every pod via `JAC_SV_ROUTES`** (the config loader reads it when its own routes are empty), so the in-cluster gateway + `get_sv_registry` resolve prefixes/peers without a routes table in the deployed `jac.toml`.
- **`--no-image`**: deploy without docker build/push - source tar+gz'd into a ConfigMap (~1 MiB guard), a bootstrap initContainer extracts it and `pip install`s pinned `jaclang`/`jac-scale`, pods run `python:3.12-slim`. Two production warnings; DBs still provisioned. Cluster-side build (buildkit/buildah) is a tracked follow-up.

## Review history

Addresses all Greptile findings from the prior PRs: the two P1s (tri-state `jac scale` guard, docstring-close heuristic) and the three P2s (explicit `enabled` identity check on the `--scale` path, `warning` not `info` on the no-services branch, removed the dead `str(root_dir)` guard).

## Testing

- Unit/integration: discovery scanner + false-positive guard, tri-state gate, config-loader tri-state + `JAC_SV_ROUTES` env fallback, manifest `JAC_SV_ROUTES` injection, no-image source packing/ConfigMap/pod-spec/bundle. Full sweep across microservice/K8s suites: **88 passed**.
- Apps with no `sv import` stay monoliths (verified live for local; covered by the gate tests).
- Real-cluster (minikube) `--scale` e2e is **pending** - not runnable in this environment.


MusabMahmoodh:
Closing to force a fresh Greptile analysis - the 3/5 review predates the pushed fixes (secret-file exclusion in pack_source, docstring-close-with-code, deduped scale arg). Reopened as a new PR from the same branch.

PR #6782, opened 2026-06-18
fix(compiler): accept subscripted/generic types in lambda param annotations

### LINKED ISSUE #6773: Lambda parameter annotation rejects subscripted/generic types (`lambda d: list[int] : ...`)
## Summary

Jac's lambda syntax supports an explicit parameter type annotation (`lambda <name>: <type> : <body>` — a Jac extension; Python lambdas have no param annotations). But the grammar for that annotation slot only accepts a **bare type name**, not a **subscripted/generic type**. Any `X[...]` there is a parse error.

```jac
with entry {
    f1 = lambda d: dict : d["n"];             # OK — bare type
    f2 = lambda d: list[int] : d[0];          # E0002 "Missing ';'" + E0004 "Unexpected token"
    f3 = lambda d: dict[str, int] : d["n"];   # same parse failure
}
```

It is **not** the comma — single-arg `list[int]` fails too. The `[` is where the lambda-param-annotation production breaks; it appears to use an atomic-type rule rather than the full type-expression rule used by `def` params and `has`/variable annotations.

## Expected

A lambda parameter annotation should accept the same type-expression grammar as everywhere else — including subscripted generics (`list[int]`, `dict[str, int]`, `Optional[X]`, etc.), consistent with `def f(d: dict[str, int])` and `has x: dict[str, int]`.

## Severity / context

- Jac-specific feature gap (no Python parity issue — Python doesn't allow annotated lambda params at all).
- Workaround: drop the annotation. Bidirectional inference now infers the param from the expected callable (e.g. `sorted(rows, key=lambda d: d["n"])` infers `d: dict[str, int]`), so the explicit annotation is usually unnecessary. A bare `lambda d: dict : ...` also parses.

## Repro

`lambda d: list[int] : d[0]` — fails to parse on `main`. Bare `lambda d: dict : ...` parses fine.

## Likely fix area

Parser/grammar: the lambda rule's parameter-type production. Swap the atomic-type production for the general type-expression production (the one `def` params already use).

## Acceptance

- `lambda d: list[int] : d[0]` and `lambda d: dict[str, int] : d["n"]` parse without error.
- Existing bare-type lambdas still parse.
- No regression in parser/checker test suites.

kugesan1105:
Fixes #6773.

## Problem

Jac's lambda syntax supports an explicit parameter type annotation (`lambda <name>: <type> : <body>`), but the annotation slot only accepted a **bare type name**. Any subscripted/generic type (`X[...]`) there was a parse error:

```jac
f1 = lambda d: dict : d["n"];            # OK
f2 = lambda d: list[int] : d[0];         # E0002 "Missing ';'" + E0004
f3 = lambda d: dict[str, int] : d["n"];  # same parse failure
```

This was inconsistent with `def f(d: dict[str, int])` and `has x: dict[str, int]`, which accept the full type-expression grammar.

## Root cause

`parse_lambda_param` disambiguated the param-annotation colon from the lambda body colon with a **fixed two-token lookahead**: it required `peek()` to be a type-name token *and* `peek(2)` to be a delimiter (`,` `=` `:` `->` `{`). For `list[int]`, the token after `list` is `[` — not a delimiter — so the annotation was never recognized, the colon fell through to the body, and the trailing body colon was left dangling.

## Fix

Replace the lookahead with **parse-and-backtrack**: consume the colon, parse the candidate type with `parse_bitwise_or` (the precedence rung that covers names, dotted paths, subscripts like `list[int]`/`dict[str, int]`, and `|` unions, but stops below comparison/arithmetic), then accept it as the annotation only if the parser lands on a real delimiter — otherwise reset `pos` so the colon belongs to the body.

This keeps non-annotation colons working: `lambda d: d[0]`, `lambda s: len(s) > 1`, and `lambda x: -x` still parse with the colon as the body separator.

## Acceptance (from the issue)

- [x] `lambda d: list[int] : d[0]` and `lambda d: dict[str, int] : d["n"]` parse without error
- [x] Existing bare-type lambdas still parse
- [x] No regression in parser/checker test suites

## Tests

Extended `tests/language/fixtures/lambda_arg_annotation.jac` with subscripted, generic, and union annotations, plus matching `jac2py` assertions in `tests/language/test_cli.jac` (annotations are stripped in the Python output; bodies preserved). Verified the repro cases parse, check, and run; confirmed identical pre-existing (unrelated) failures before/after — zero regressions.

### DIFF
--- a/jac/jaclang/jac0core/parser/impl/parser.impl.jac
+++ b/jac/jaclang/jac0core/parser/impl/parser.impl.jac
@@ -3121,41 +3121,30 @@ impl Parser.parse_lambda_param -> ParamVar | None {
         return None;
     }
     type_tag: SubTag | None = None;
-    # Heuristic to distinguish type annotation from lambda body colon:
-    # Type annotation COLON is followed by a type and then COMMA, EQ, COLON, RETURN_HINT, or LBRACE
+    # An un-parenthesized lambda reuses `:` for both the param annotation and
+    # the body separator: `name : <type> : <body>`. Unlike `def`/`has` params
+    # (whose colon is unambiguous), disambiguate by parsing the candidate type
+    # with parse_bitwise_or — the precedence rung that covers type syntax
+    # (names, dotted, subscripts like `list[int]`/`dict[str, int]`, `|` unions)
+    # but stops below comparison and other operators. A real type lands on a
+    # param/body delimiter (`:` `,` `=` `->` `{`); a body like `len(s) > 1,`
+    # stops at `>` and `-x` at `-`, so the trailing-token check rejects them and
+    # the colon falls through to the body.
     if self.check(TokenKind.COLON) {
-        next_tok = self.peek();
-        is_type_annotation = False;
-        if next_tok.kind in [
-            TokenKind.NAME,
-            TokenKind.TYP_STRING,
-            TokenKind.TYP_INT,
-            TokenKind.TYP_FLOAT,
-            TokenKind.TYP_LIST,
-            TokenKind.TYP_TUPLE,
-            TokenKind.TYP_SET,
-            TokenKind.TYP_DICT,
-            TokenKind.TYP_BOOL,
-            TokenKind.TYP_BYTES,
-            TokenKind.TYP_ANY,
-            TokenKind.TYP_TYPE
-        ] {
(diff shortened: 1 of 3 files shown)


marsninja:

PR #6783, opened 2026-06-18
feat(client): client-only (`kind = "client"`) apps — portable build + minimal serve

marsninja:
## What

A `kind = "client"` Jac project runs entirely in the browser with no backend. `jac build` and `jac start` auto-detect such projects and route them to a new `static` build target — **no `--client static` flag required**. An explicit `--client <target>` always overrides the auto-detection.

> Naming: an earlier revision of this PR used `kind = "wasm"`. That named a *technology*, and a pure `cl` app compiles to JavaScript (not WebAssembly), so it was renamed to `kind = "client"` (deployment shape) with build target `static`. `na`→wasm compute stays an automatic composition detail (emitted into the dist, served at `/static/<stem>.wasm`), not a separate kind. Clean break — no alias.

### `jac build` → portable, self-contained dist
After the normal Vite build, the generated `index.html` has its JS bundle and CSS **inlined**, so the page opens directly from disk (`file://`) — e.g. email the single HTML file or drop it on a USB stick. This is necessary because browsers CORS-block external `<script type="module" src=...>` over `file://`; a dist that references its bundle by URL renders blank when double-clicked. Warns when code-splitting leaves chunks the inlined page would still need to fetch.

### `jac start` → minimal static server
Because there is no backend, the `static` target skips the full API server (no walkers, auth, database, or scheduler) and serves the dist with a tiny stdlib `http.server`. It maps the conventional `/static/<name>.wasm` mount onto the dist root, so an `na`→wasm module fetched client-side is still served correctly.

Dev mode (`jac start --dev`) is inherited from the web target (Vite + HMR, no API).

## Design
- New `StaticTarget(WebTarget)` mirrors the existing `PWATarget` pattern; reuses the WebTarget Vite pipeline via `super().build()`.
- Auto-detection reuses the existing project-kind resolver (`resolve_project_kind`) and the target registry/factory; `kind = "client"` → `static` target.
- `inline_dist_assets()` and `serve_static_dir()` are small, independently testable units.
- Added the `client` kind to the canonical kind table; scaffold template, create blurbs, and docs updated.

## Testing
`test_static_target.jac`:
- inline transform produces a self-contained `index.html` (JS + CSS inlined, sourcemap comment stripped, init payload preserved); no-op without `index.html`
- `static` target registered/resolvable via the factory
- `kind = "client"` detected from `jac.toml`
- minimal static server maps `/static/<x>` → dist root
- e2e: a `kind=client` project builds a dist that renders over `file://` in headless Chromium (skips when playwright/build toolchain absent)

Manually verified end-to-end: `jac build` (no flag) on a `kind=client` project emits an inlined `index.html` that renders in headless Chromium over `file://`, and `jac start` serves it via the minimal server with `/static/` mapping.

## Docs
- jac-client reference: new "Client-only apps" section + command-table/option updates.
- project-kinds quick guide: portable-build / minimal-serve / `file://` behavior for the `client` kind.


marsninja:
Thanks @greptile-apps — addressed in 75a8fb3:

- **P1 (wasm MIME)** — Agreed. `_DistRequestHandler` now overrides `guess_type` to return `application/wasm` for `.wasm`, so `WebAssembly.instantiateStreaming` works regardless of the platform MIME database. The static-server test now asserts the `Content-Type` header, not just the bytes.
- **P2 (CSS `</style>` escaping)** — Agreed, clear oversight. `_guard_closing_tag` is now applied to the inlined CSS as well. The inline test feeds CSS containing a literal `</style>` and asserts only one real closing tag survives.
- **P3 (explicit `--client web`)** — Partially agree: the docstring was overclaiming, so I corrected it (and the docs). I did **not** add an escape hatch, because argparse bakes `default="web"` into the namespace, so the hook genuinely cannot distinguish an explicit `--client web` from the default without reworking the shared arg-default mechanism that other commands/plugins rely on — disproportionate for a niche debug case. An explicit non-web `--client` (e.g. `--client pwa`) still overrides routing.

Also fixed the CI failure (unrelated to the above): the core `test_run_dispatch.jac` still asserted a serve plan for `kind = "wasm"`, which I'd renamed to `client`.

Issue #6535, opened 2026-06-09
feat(jac-scale): per-service TriggerAuthentication isolation for KEDA autoscaler (#6534 follow-up)


Developer-Linus:
### Feature Description

Three known gaps in `KEDAAutoscaler._apply_trigger_auth` were deliberately deferred from PR #6534 (Phase 2 KEDA engine). This issue tracks the follow-up work.

**Gap 1 — Per-service isolation**
`TriggerAuthentication` resource names are currently derived as `{trigger.name or trigger.type}-trigger-auth`. Two services that define a trigger with the same name (e.g. both have a `redis` trigger) will reference and mutate the same `TriggerAuthentication` resource. The second deploy silently patches the resource with the second service's credentials, potentially breaking the first service's auth.

**Gap 2 — Within-spec collision**
Two unnamed triggers of the same type in a single `AutoscalerSpec` (e.g. two `redis` triggers without a `name` field) derive the same `TriggerAuthentication` name. The second one overwrites the first during `_apply_trigger_auth`, so only the last trigger's auth survives.

**Gap 3 — `destroy()` cleanup gap**
`KEDAAutoscaler.destroy(autoscaler_name, namespace)` deletes the `ScaledObject` by name but has no trigger list, so it cannot look up and delete the associated `TriggerAuthentication` resources. Orphaned `TriggerAuthentication` objects accumulate across deploys and reference stale secrets. `destroy_collection` (label-based) does clean them up correctly — only the single-resource path is affected.

**Location:** `jac-scale/jac_scale/abstractions/autoscalers/impl/keda_autoscaler.impl.jac` → `_apply_trigger_auth` and `destroy`

**Acceptance criteria**
- `TriggerAuthentication` names are scoped per-service so two services with the same trigger name get independent resources
- Within-spec collision is resolved (e.g. by including a service-name segment in the resource name)
- `destroy()` cleans up `TriggerAuthentication` resources associated with the deleted `ScaledObject`

### Examples & References

**Current (problematic) naming:**
```
service-a  redis trigger  →  redis-trigger-auth
service-b  redis trigger  →  redis-trigger-auth   ← same name, overwrites service-a's auth
```

**Expected naming after fix:**
```
service-a  redis trigger  →  service-a-redis-trigger-auth
service-b  redis trigger  →  service-b-redis-trigger-auth   ← independent
```

**TOML config that triggers all three gaps today:**
```toml
# Two services, both with a redis trigger named "cache"
[...]

PR #6801, opened 2026-06-19
fix(jac-scale): scope TriggerAuthentication names per-service to prevent cross-service and within-spec collisions (part 1 of issue #6535)

Developer-Linus:
## Problem

`KEDAAutoscaler._apply_trigger_auth` and `_build_trigger` both derived `TriggerAuthentication` resource names using:

```
f"{trigger.name or trigger.type}-trigger-auth"
```

This formula had no awareness of which service the trigger belonged to, or where it sat within the trigger list. Two bugs followed:

**Gap 1 — Cross-service collision (microservices mode)**
Two services defining a trigger with the same name (e.g. both with a `redis` trigger named `cache`) resolved to the same `TriggerAuthentication` resource name. The second service's deploy silently patched the first service's auth resource with the wrong credentials. KEDA continued using the stale secret reference for the first service's trigger with no visible error.

**Gap 2 — Within-spec collision (monolith and microservices)**
Two unnamed triggers of the same type within a single `AutoscalerSpec` (e.g. two `redis` triggers without a `name` field) derived the same resource name. The second `_apply_trigger_auth` call overwrote the first trigger's auth, leaving it with the wrong credentials.

Both bugs produced no cluster-side errors — the `ScaledObject` remained healthy while its trigger authentication was silently broken.

## Solution

Introduced a single private `_trigger_auth_name` helper that both `_build_trigger` and `_apply_trigger_auth` now call, ensuring the two sides can never silently diverge:

```
key        = f"{app_name}/{trigger.name or f'{trigger.type}-{trigger_index}'}"
short_hash = hashlib.sha256(key.encode()).hexdigest()[:8]
auth_name  = f"{app_name[:20].rstrip('-')}-{short_hash}-ta"
```

- **`app_name`** (`spec.app_name or spec.scale_target_name`) scopes the name to the service, fixing Gap 1
- **`trigger_index`** (position in `spec.triggers`) differentiates unnamed triggers of the same type, fixing Gap 2
- **SHA-256 short hash** keeps names well within Kubernetes' 63-character resource name limit regardless of service or trigger name length
- The `app_name[:20].rstrip('-')` prefix preserves human readability and avoids double-hyphens when the service name is truncated at a hyphen boundary

### Before / after

| Scenario | Before | After |
|---|---|---|
| `order-service`, trigger named `cache` | `cache-trigger-auth` | `order-service-daa02e20-ta` |
| `payment-service`, trigger named `cache` | `cache-trigger-auth` ← collision | `payment-service-53abb760-ta` ✓ |
| `order-service`, unnamed `redis` at index 0 | `redis-trigger-auth` | `order-service-425f6d48-ta` |
| `order-service`, unnamed `redis` at index 1 | `redis-trigger-auth` ← collision | `order-service-c3a94c81-ta` ✓ |

### `keda_autoscaler.jac`
- Added `import hashlib;`
- Added `_trigger_auth_name(app_name, trigger, trigger_index)` signature
- Updated `_build_trigger` and `_apply_trigger_auth` signatures to accept `app_name` and `trigger_index`

### `keda_autoscaler.impl.jac`
- Implemented `_trigger_auth_name` — single source of truth for the naming formula, with `.rstrip('-')` after truncation to prevent double-hyphens when the service name is cut at a hyphen boundary
- `_build_trigger` and `_apply_trigger_auth` both delegate to `_trigger_auth_name`
- Updated `apply()` and `_build_manifests()` loops to use `enumerate` and pass `app_name` and trigger index
- Removed Gap 1 and Gap 2 items from the `TODO(#6535)` comment; Gap 3 (destroy cleanup) remains

### `test_keda_autoscaler.jac`
- Updated all existing `_build_trigger` call sites to pass the new required parameters
- Updated the `authenticationRef` name assertion to reflect the new hashed format
- Added 3 new tests:
  - `"two services with the same trigger name get independent TriggerAuthentication names"` — Gap 1
  - `"two unnamed triggers of the same type in one spec get independent TriggerAuthentication names"` — Gap 2
  - `"TriggerAuthentication name written to cluster matches authenticationRef name in ScaledObject"` — verifies `_build_trigger` and `_apply_trigger_auth` stay in sync end-to-end; guards the index access with `assert len(create_calls) == 2` for a meaningful failure message if a regression reduces the call count

## Out of scope

Gap 3 (`destroy()` cleanup of orphaned `TriggerAuthentication` resources) is tracked in the follow-up PR stacked on this one.

## Related

- Closes Gap 1 and Gap 2 from issue #6535
- Follow-up to PR #6534 (Phase 2 KEDA autoscaler engine)


jac/jaclang/scale/deploy/autoscale/keda_autoscaler.impl.jac
```diff
@@ -1,3 +1,10 @@
+"""Derive a short, unique TriggerAuthentication resource name scoped to the service."""
+impl KEDAAutoscaler._trigger_auth_name{
+    key = f"{app_name}/{trigger.name or f'{trigger.type}-{trigger_index}'}";
+    short_hash = hashlib.sha256(key.encode()).hexdigest()[:8];
+    return f"{app_name[:20].rstrip('-')}-{short_hash}-ta";
+}
```
MusabMahmoodh, line 9: Two small things to consider here, both non-blocking.

**Empty `app_name` edge case**: `""[:20].rstrip('-')` produces `""`, so an empty `app_name` gives `"-{hash}-ta"`, which starts with a hyphen and is an invalid k8s resource name. Today `app_name = spec.app_name or spec.scale_target_name` so empty is unlikely, but no guard. A simple `assert app_name` (or fallback to `"autoscaler"`) would make this safe.

**Explicit duplicate name within a spec is still silently shared**: when both triggers set `name = "cache"`, the hash key collides because `trigger_index` is only used when `trigger.name` is unset. This was true before the patch too so not a regression, but worth either including the index unconditionally (`f"{app_name}/{trigger.name or trigger.type}-{trigger_index}"`), or catching at config-load with a clear duplicate-name error.

jac/jaclang/scale/deploy/autoscale/keda_autoscaler.impl.jac
```diff
@@ -80,8 +87,9 @@ impl KEDAAutoscaler._build_trigger{
 impl KEDAAutoscaler._build_manifests{
     scaled_obj_name = spec.autoscaler_name or f"{spec.scale_target_name}-scaledobject";
     keda_triggers: list[dict[str, any]] = [];
-    for trigger in spec.triggers {
-        keda_triggers.append(self._build_trigger(trigger));
+    app_name = spec.app_name or spec.scale_target_name;
+    for (i, trigger) in enumerate(spec.triggers) {
```
MusabMahmoodh, line 105: Same `spec.app_name or spec.scale_target_name` expression also appears at line ~233 in `apply()`. Cheap to extract as a property on `AutoscalerSpec` (e.g. `effective_app_name`) so the two call sites cannot drift if the rule ever changes. Pure cleanup, fine to defer.

MusabMahmoodh (approved):
Looks good. The single-helper approach is exactly the right shape, and the third test asserting both sides stay in sync is the part that gives me confidence this cannot silently break again. Approving.

Two non-blocking nits noted inline. Also flagging two small housekeeping items:

1. PR title says `part 1 of issue #6525` but the actual tracker is `#6535` (body is correct). Worth fixing in the title.
2. Mergeable state is CONFLICTING. Please rebase on main before merging.

Developer-Linus:
> Looks good. The single-helper approach is exactly the right shape, and the third test asserting both sides stay in sync is the part that gives me confidence this cannot silently break again. Approving.
> 
> Two non-blocking nits noted inline. Also flagging two small housekeeping items:
> 
> 1. PR title says `part 1 of issue #6525` but the actual tracker is `#6535` (body is correct). Worth fixing in the title.
> 2. Mergeable state is CONFLICTING. Please rebase on main before merging.

I have addressed the review comments in this PR:
Added `_validate_triggers` to catch duplicate named triggers within a spec before any cluster write.

Two triggers sharing the same name produce identical `TriggerAuthentication` resource names. Previously, this allowed the second `apply()` call to silently overwrite the first trigger’s credentials with no visible error.

The new validation raises a `ValueError` naming both the duplicate trigger and the affected service. It is called at the top of `apply()`, before `preflight()`, so misconfigured specs fail fast before touching the cluster.

As part of this change, `app_name` is now hoisted to the top of `apply()`, removing the duplicated `spec.app_name or spec.scale_target_name` expression and ensuring the HPA engine-switch deletion uses the same resolved value.

A focused test covers the duplicate-trigger scenario. The branch has also been rebased onto `main`.

Issue #6809, opened 2026-06-19
Typechecker false positive errors


SandeepaHWP:
## Summary

Running `jac check` on the self-hosted compiler passes (`jac/jaclang/jac0core/passes`) reports many type errors that are false positives: the code is correct and runs fine, but the checker flags it due to narrowing gaps, incomplete stubs, and cascades from unresolved types.

Below is one minimal, self-contained reproduction per case. The offending line is marked inline with the exact error code and message the checker emits. None of these represent a runtime bug.

(Two genuinely real errors were found in the same sweep and are excluded here because they were real, not false positives.)

## Validation

Each case was validated as a false positive by converting the repro to equivalent Python and confirming it has no errors. This shows the code is correct and the Jac checker is flagging it spuriously.

That said, passing as Python does not mean every one of these should be "fixed to behave exactly like Python". The Jac type checker is allowed to have its own rules. While fixing each case, if it turns out the right behavior should differ from Python, then do that instead. The Python check is evidence that the current behavior is a false positive, not a mandate to copy Python in every detail.

---

### 1. Same-name classes compared by identity

```jac
# alert.jac
obj Alert { has msg: str; }
obj Program { has errors_had: list[Alert] = []; }
```
```jac
# repro.jac
import from alert { Program }

obj Alert { has msg: str; }

obj Pass {
    has prog: Program;

    def route(alrt: Alert) {
        self.prog.errors_had.append(alrt);  # error[E1053]: Cannot assign Alert to parameter 'object' of type Alert
    }
}
```
There are two different classes both named `Alert` (one local, one reached through the imported `Program`). The checker compares them by identity, they differ, and the message prints both as the bare name `Alert`, so it reads as "cannot assign Alert to Alert".

---

### 2. Declaration typevar not in scope inside the impl block

```jac
obj Box[T] {
    has item: T;
    def flatten(items: list[T | None]) -> list[T];
}

impl Box.flatten(items: list[T | None]) -> list[T] {   # error[E1065]: Type variable "T" has no meaning in this context
    out: list[T] = [];                                  # error[E1065]: Type variable "T" has no meaning in this context
    for it in items {
        if it is not None {
            out.append(it);
        }
    }
    return out;
}
```
The `impl` signature references the type variable bound by the declaration, but the checker does not carry the declaration's typevars into the impl scope. Generics are erased at runtime, so this has no runtime effect.

---

### 3. Return type rejected (Any return and isinstance mis-narrow)

```jac
[...]

PR #6810, opened 2026-06-19
Fix real type errors in the compiler's type-checker and codegen passes

SandeepaHWP:
## What this does

Fixes the genuinely-real type errors found in the compiler's own source while running `jac check`. This consolidates two sets of real fixes into one PR (the type-checker fixes previously in #6478, plus the jac0core codegen-pass fixes). Most reports from the wider sweep are checker false positives (tracked in #6809) and are left unchanged.

### Type checker (`compiler/type_system`)
- `validate_arg_types` was declared `-> None` but returns a bool; the declaration now matches the implementation.
- `validate_call_args` could return the optional `prefetch.int_class`; it now falls back to `UnknownType`.
- `match_args_to_params` and `get_type_of_magic_method_call` only accepted `list[Expr]`, but callers pass `expr.params` (`Sequence[Expr | KWPair] | None`); the parameter type is widened and a None-guard added.
- The `typing.cast` special-case called `len(expr.params)` without a None check; it now guards `expr.params` first.
- `get_type_of_member`, `_get_method_with_overloads`, `_find_overload_by_arg_count`, and the `found_member` local were typed `any`/untyped, leaking `Any` into a `Symbol` parameter; they are now typed as `ClassMember` / `list[Symbol]`.

### Codegen passes (`jac0core/passes`)
- `_py_name`: `Ability.name_ref` is `NameAtom | None` and three call sites pass it unguarded; `_py_name` dereferences `nd.sym_name`, so `None` would crash at runtime. Widened the parameter to `AstSymbolNode | None` and added a `None` guard that raises an internal compiler error.
- `_collect_boundary_type`: `arch_kind` is `str | None` and `_ak in ("node", "walker")` does not narrow away `None`, so `kind` was inferred `str | None` for a `str` field. Added an explicit `is not None` check.
- Replaced two exception-swallowing `pass;` statements (not valid Jac, E0010) with the empty statement `;`, matching the existing convention in the codebase. Behavior unchanged.

## Notes

- No behavior change beyond the None handling, which previously would have crashed.
- Supersedes #6478 (same type-checker fixes, now folded in here).
- Remaining `jac check` errors on these files are false positives; see #6809.


jac/jaclang/compiler/type_system/type_evaluator.impl/parameter_type_check.impl.jac
```diff
@@ -81,7 +81,8 @@ impl TypeEvaluator.validate_call_args(expr: uni.FuncCall) -> TypeBase {
     if isinstance(caller_type, types.ClassType)
     and caller_type.is_builtin("auto")
     and (not expr.params or len(expr.params) == 0) {
-        return self.prefetch.int_class if self.prefetch else types.UnknownType();
+        return (self.prefetch.int_class if self.prefetch else None)
+        or types.UnknownType();
```
kugesan1105, line 85: why something liek this; is this  a false positive error?

jac/jaclang/jac0core/passes/impl/boundary_analysis_pass.impl.jac
```diff
@@ -752,7 +752,7 @@ impl BoundaryAnalysisPass._collect_boundary_type(
         # Edge archetypes intentionally fall back to "obj" here -- this
         # manifest only distinguishes node/walker shapes.
         _ak = arch_node.arch_kind;
-        kind = _ak if _ak in ("node", "walker") else "obj";
+        kind = _ak if (_ak is not None and _ak in ("node", "walker")) else "obj";
```
kugesan1105, line 755: this should be a false positive error

RandiJay96:
covers in #7249

jac/jaclang/compiler/type_system/type_evaluator.impl/parameter_type_check.impl.jac
SandeepaHWP: Not a false positive — verified with a minimal repro (mirroring this exact shape, using the real `types.ClassType`/`types.TypeBase`/`types.UnknownType()`):

```jac
obj Prefetch {
    has int_class: types.ClassType | None = None;
}
obj Evaluator {
    has prefetch: Prefetch by postinit;
    def postinit -> None { self.prefetch = Prefetch(); }
    def get_int_type() -> types.TypeBase {
        return self.prefetch.int_class if self.prefetch else types.UnknownType();
    }
}
```

`jac check` reports: `Cannot return ClassType | NoneType | UnknownType, expected TypeBase (E1002)` — same error class as here.

`int_class` is declared `types.ClassType | None` (populated via `get_builtin_class("int")`, which can legitimately return `None` if the builtins stub lookup fails to resolve). The original guard checked `self.prefetch` — which is always truthy, since it's a `by postinit` field guaranteed to be set — so it never actually protected against `int_class` itself being `None`. That's a real logic gap, not a checker limitation: if you narrow on the *correct* value directly (`self.prefetch.int_class is not None`), the checker accepts it cleanly, proving its narrowing logic works fine here — the original code was just checking the wrong condition.

The fix wraps it in `or` rather than a plain ternary because the checker has an explicit rule for that: for every non-final operand of an `or` chain, it strips `NoneType` out before unioning types (see `type_evaluator.impl.jac`, the `is_or` branch in the `BoolOp` case) — mirroring Python's own `x or default` idiom. A plain ternary doesn't get this treatment, since the condition and the value are unrelated to the checker. So `(self.prefetch.int_class if self.prefetch else None) or types.UnknownType()` correctly narrows away the `None`, while the original line couldn't.

jac/jaclang/jac0core/passes/impl/boundary_analysis_pass.impl.jac
SandeepaHWP: Agreed — this one is a genuine checker false positive, confirmed with a minimal repro:

```jac
def pick_kind(ak: str | None) -> str {
    result: str = ak if ak in ("node", "walker") else "obj";
    return result;
}
```

`jac check` reports `error[E1001]: Cannot assign str | NoneType to str` on that line even though it's safe at runtime — `None in ("node", "walker")` always evaluates to `False`, so `ak`/`_ak` can never actually be `None` in the true branch.

Root cause: `TypeEvaluator.classify_condition` (`type_evaluator.impl.jac`) only recognizes `isinstance`, `callable`, `hasattr`, `x is/== None`, `x is not/!= None`, single-literal `==`/`!=`, bare truthiness, and walrus as narrowing forms. There's no case for the `in`-against-a-tuple-of-literals pattern, so it returns `None` for this condition and no narrowing is ever applied — the checker just falls back to `_ak`'s original declared type in both branches.

The `_ak is not None and ...` guard only worked because `is not None` on its own *is* a recognized narrowing form, and the checker narrows sequentially through `and`-chains — so it papers over the gap rather than fixing anything. It doesn't belong in this PR (scoped to genuine fixes), so I've reverted it and filed the checker limitation separately: jaseci-labs/jaseci#7256 (includes the repro, root cause, and a suggested fix — a new `MEMBERSHIP` `NarrowingKind`).

PR #6827, opened 2026-06-20
fix(nacompile): propagate clib_symbol_lib transitively and fix Mach-O dylib ABI violations (#6821)

### LINKED ISSUE #6821: nacompile --shared: C-ABI surface not usable via ctypes on native host (Win WinError 127 import; macOS dlsym)
## Summary

A `nacompile --shared` library is not usable via `ctypes` on a **native host**, for either Windows or macOS, even though it builds and (on macOS) its export trie verifiably contains every symbol. The library is the correct object format and loads far enough to be opened — the failure is in **runtime symbol resolution**.

Distinct from #6818 (native-host *executable* emit producing ELF): here the shared object is correct, links, and loads.

## Windows — `WinError 127` on load

Minimal shape — a shared lib that imports a Win32 function and exports one symbol:

```jac
# probe.na.jac
import from "kernel32.dll" {
    def GetStdHandle(std: i32) -> i64;
}
def ping :pub -> i64 { return GetStdHandle(-11); }
```

```
jac nacompile probe.na.jac --shared --target windows -o probe.dll   # builds, COFF, exports `ping`
```

```python
import ctypes
ctypes.CDLL("probe.dll")
# [WinError 127] The specified procedure could not be found  (raised by _dlopen)
```

`WinError 127` on `LoadLibrary` means the DLL's own **import table** references a procedure the named dependency DLL does not export. The imports emitted (e.g. from `kernel32.dll` / `msvcrt.dll`) include one in a form the loader can't resolve — a wrong DLL/name, or a malformed import descriptor in the import directory.

## macOS — `dlsym` symbol not found despite a complete export trie

Minimal shape:

```jac
# probe.na.jac
def ping :pub -> i64 { return 42; }
```

The built `libprobe.dylib` loads, codesigns, and its export trie contains `_ping` (verifiable by walking `LC_DYLD_INFO` / `LC_DYLD_EXPORTS_TRIE`). But the first `dlsym` lookup fails:

```python
import ctypes
lib = ctypes.CDLL("libprobe.dylib")  # ok
lib["ping"]    # or lib["_ping"]
# dlsym(0x..., _ping): symbol not found
```

The export trie stores Mach-O symbols with a leading underscore (`_ping`), but `dlsym` on Darwin takes the bare C name (`ping`) and adds the underscore internally. Whether the caller should pass the underscore, and whether the NA runtime's `dlsym` shim matches stock libdyld semantics, needs to be settled. Either way, a freshly built dylib with a verifiably complete trie cannot have its symbols resolved through the standard `ctypes` path.

## Related

#6818 (native-host executable emit) and #6822 (entry-module compile cascade).

chess10kp:
## Summary

Fixes two `ctypes.CDLL` symbol-resolution failures that caused the TUI shared-library host gate to fail on macOS and Windows.

### macOS — bare symbol names in `test_host.jac`

`nacompile --shared` builds the Mach-O export trie using LLVM IR function names (no `_` prefix). Python ctypes on macOS retries `dlsym` with `_`+name when the bare lookup returns NULL, so `lib["tui_apply_frame"]` correctly resolves `_tui_apply_frame` in the trie. The previous `lib["_" + "tui_apply_frame"]` caused ctypes to retry `__tui_apply_frame` (double underscore), which is never found.

### macOS — Mach-O dylib ABI violations (`macho_linker.impl.jac`)

Two load-command errors caused `dyld` on macOS 15 to silently suppress the dylib's exports:
- `LC_LOAD_DYLINKER` is only valid in `MH_EXECUTE`; removed from `build_shared_object`
- `CS_EXECSEG_MAIN_BINARY` in `execSegFlags` must be `0` for a dylib; `_build_code_signature` now accepts an `exec_seg_flags` parameter (default keeps existing behaviour for executables)

### macOS — Darwin termios constants (`libc_tty.darwin.na.jac`)

Four XNU constants were wrong, causing `test_tty_darwin.jac` to fail:
- `CSIZE`/`CS8`: `0x3000` → `0x300`
- `PARENB`: `0x100` → `0x1000`
- `IXON`: `0x400` → `0x200` (`0x400` is `IXOFF` on Darwin)

### Windows — `clib_symbol_lib` not propagated across module boundaries (`na_compile_pass.impl.jac`)

`_compile_and_link_native_imports` propagated `_clib_paths` (the list of DLL paths) from imported sub-modules to the root module but not `_clib_symbol_lib` (the symbol→DLL map). The PE linker saw an empty map and bucketed all Win32 API symbols into `msvcrt.dll`; `LoadLibrary` found it but `GetConsoleMode`/`WaitForSingleObject`/`WriteFile` are not in `msvcrt` → WinError 127. Fix: propagate `dep_clib_symbol_lib` alongside `dep_clib_paths` and persist it in the `.ir_meta` cache so cache-hit paths also carry the mapping.

### Windows — missing `tty_x_pixels`/`tty_y_pixels` stubs (`console.win32.na.jac`)

`terminal_image.na.jac` imports both symbols from `libc_tty`; the Win32 backend was missing them. Added 0-returning stubs (Windows Console API has no pixel-dimension query); callers already guard on `> 0` before dividing.

## Test plan

- [ ] macOS CI: `test-macos-native` — CDLL host gate (load + frame round-trip) restored and passing
- [ ] macOS CI: `test-macos-native` — PTY termios constant probe passing with corrected Darwin constants
- [ ] Windows CI: `test-windows-native` — CDLL host gate restored and passing (WinError 127 resolved)
- [ ] Windows CI: `test-windows-native` — Win32 console constant + VT gate passing
- [ ] Linux CI: unaffected; existing gates continue to pass

Closes #6821


chess10kp:
Closing — branch has diverged history from the fork; reopening as a clean cherry-pick off upstream/main.

PR #6830, opened 2026-06-20
Per-project plugin provisioning; drop global jaseci meta-package

### LINKED ISSUE #6824: Simplify dependency/project model: pip install jaclang only, plugins provisioned per-project from jac.toml
## Summary

Dependency/project organization is currently messy for developers: jaclang **and** every plugin a project might use (`byllm`, `jac-client`, `jac-scale`, …) must be installed into the *same* environment the `jac` CLI runs from. This proposes a clean simplification: **a developer runs `pip install jaclang` and nothing else.** All plugins are declared per-project in `jac.toml`, provisioned into the project-local `.jac/venv`, and loaded from there. `jac create` becomes kind-aware and writes the right plugin dependency by default. The `jaseci` meta-package and the now-redundant global-install plumbing are deleted.

Most of the machinery already exists (`.jac/venv`, `DependencyInstaller`, `jac install/add/remove`, `KIND_SPECS`, `[dependencies]`/`[plugins]` in `jac.toml`). The work is **flipping the default and deleting redundant paths**, not a from-scratch rebuild.

## Current state

Plugins are discovered by scanning `importlib.metadata.entry_points(group="jac")` against the one global env (`helpers.jac:554-649`, triggered at `__init__.py:35-39`). This forces: ambiguous "what goes where," a `jaseci` meta-package that bundles everything globally, an `install.sh` that enumerates `--with byllm --with jac-client …`, and multiple install narratives in the docs.

## Two findings that anchor the approach

1. **Latent bug = the enabler.** Plugins already in `.jac/venv` have importable *modules* but their pluggy hooks are **never registered**: the entry-point scan at `__init__.py:35-39` runs at `import jaclang` time, *before* the venv is added to `sys.path` by `add_venv_to_path` (`cli.impl.jac:201-207`). Re-running discovery after the venv is on the path makes project plugins work — additively, no deletion required.

2. **`jac create` writes the wrong thing.** A `fullstack` project lists `jac-client` only under build metadata (`[jacpack.plugins]`/`[plugins.client]`), never under `[dependencies]` (`project.impl.jac:106-184`). So `jac install` installs nothing and `jac run` fails. `KIND_SPECS` (`kinds.jac:24-37`) already maps kind→plugin; it just needs threading into rendered `[dependencies]`.

## Decision: hermeticity model

**Target = re-exec launcher (Model A), pipx/uv-tool style.** Global `jac` becomes a thin pure-Python launcher; when a `jac.toml` is found it ensures `.jac/venv` (jaclang + declared plugins) and `os.execv`s the real CLI inside that venv.

Why over sys.path-injection (Model B):
- **One `jaclang` guaranteed by construction** — the global interpreter is *replaced*, not augmented, so there is never a second `jaclang` module / second `plugin_manager`. Model B must actively keep jaclang out of the venv; one transitive `jaclang>=X` pin violates that and silently breaks every hook.
- **True per-project jaclang versions.** Model B forces the global jaclang on every project and cannot honor a plugin's jaclang pin.
- **Honest realization of the goal.** Model B's `--system-site-packages` leaks the global env into the project — the opposite of "nothing but jaclang globally."

We reach Model A in milestones so value lands early and tests stay green.

---

## Milestone 1 — Project-venv plugins load + kind-aware create (additive, ships the core value)

Additive: global-install projects keep working; a jaclang-only env + project `jac.toml` starts working.

**1a. Re-run plugin discovery after the venv is on the path.** In `_load_project_config` (`cli.impl.jac:196-208`), after `add_venv_to_path(config)`, call `load_plugins_with_disabling(plugin_manager, get_disabled_plugins())`. Idempotent via the `is_registered` guards (`helpers.jac:604-606`). Provision the venv **without** reinstalling jaclang (reuse the global one) so Milestone 1 cannot create a second jaclang on `sys.path`.

**1b. Kind-aware `jac create` writes `[dependencies]`.** In `create` (`project.impl.jac:106-184`): replace the hard "plugin not installed" gate with a soft notice; add `_ensure_kind_plugin_dep(render_config, kind)` that writes `render_config["dependencies"][plugin] = <spec>` using `kind_required_plugin(kind)`. Version spec reuses the `[jacpack.plugins]` version so create and the template registry can't drift.

Kind→plugin map (already in `KIND_SPECS`): `fullstack`/`client`/`mobile` → `jac-client`; `desktop` → `jac-desktop`; all others → core (no dep).

**Invariant:** from a clean `pip install jaclang` env, `jac create app --kind fullstack && cd app && jac install && jac run` works end-to-end. Global-install projects and the monorepo dev loop still pass.

## Milestone 2 — Re-exec launcher (hermetic north star)

**2a. Pure-Python launcher.** Add `jaclang/_launcher.py` (imports only `os`, `sys`, `tomllib`, `hashlib`, `venv`, `subprocess` — never jaclang or any `.jac` module) and `jaclang/__main__.py` as the re-exec target. Repoint the console script in `jac/pyproject.toml` from `jaclang.jac0core.cli_boot:start_cli` to `jaclang._launcher:main` (required because `cli_boot.jac` is itself a `.jac` module — importing it already triggers the global plugin scan, so the branch point must be pure Python).

`main()` flow:
1. `jac purge` / `--version` / env-independent commands → fall through immediately (works even on a broken venv).
2. `JAC_NO_VENV`/`JAC_NO_SYNC` set, or already inside the venv (`JAC_IN_VENV=1`) → run in-process (escape hatch for monorepo maintainers and CI editable installs).
3. Pure-Python upward walk for `jac.toml`. None found → fall through (zero-ceremony fast path on global jaclang).
4. Freshness check: lock hash over `[dependencies]`/`[dev-dependencies]`/git deps + kind plugin + jaclang version, stored at `.jac/venv/.jac-lock`. Stale → provision (venv.EnvBuilder + `uv pip`); `--frozen` → error instead.
5. `os.execv(<venv>/bin/python, ["-m", "jaclang", *argv[1:]])` with `JAC_IN_VENV=1`.

Factor provisioning mechanics into a pure-Python helper that **both** the launcher and the existing `DependencyInstaller`/`jac install` delegate to.

**2b. Auto-provision UX.** `jac run`/`jac start` auto-provision behind the freshness check with a one-line notice and `--frozen`/`JAC_NO_SYNC` to suppress. `jac install` remains the explicit verbose path.

**Invariant:** the global env contains only jaclang; a project pinning a specific jaclang runs on it; `jac purge` and no-project scripts stay fast; `JAC_NO_VENV=1` reproduces today's in-process behavior.

## Milestone 3 — Delete the `jaseci` meta-package and redundant global-install paths

Land only after M1–M2. The repo's own dev loop never used the meta-package (test workflows use `pip install -e jac` + `jac install -e <plugin>`), which makes this low-risk.

- **Meta-package:** delete `jaseci-package/` entirely (no runtime code imports `jaseci`).
- **Release tooling:** remove the `"jaseci"` entry from `PACKAGES` and `INTERNAL_DEPS` in `scripts/release_utils.jac` (`DEPENDENTS` is computed, auto-drops). Re-run `release.jac`/`check_manifest_sync.jac`/`validate_release.jac` dry-run.
- **Installer:** `scripts/install.sh` — drop the `--with byllm --with jac-client …` block, make `uv tool install jaclang` the whole story, make `--standalone` jaclang-only (or rename the full asset away from `jaseci`), drop `uv tool uninstall jaseci`. Coordinate asset names with `build-standalone.yml`.
- **CI/workflows:** delete `.github/workflows/release-jaseci.yml`; remove `inputs.jaseci`/`IN_JASECI`/`publish-tier4` from `publish-release.yml` (keep the `@jaseci/runtime` npm package — that's the jaclang runtime, not the meta-package); rewrite `test-installer.yml`'s "versions differ" assertion; repoint `release-github.yml`'s version source off `jaseci-package/jac.toml`; resolve/strip `build-jaseci` in `build-standalone.yml`.
- **Docs:** rewrite `docs/docs/quick-guide/install.md` and `README.md` from "pip install jaseci / install plugins globally" to "pip install jaclang; plugins are declared per-project / by kind." Sweep remaining `pip install jaseci` / `--with` references across `docs/docs/**`.
- **Disable logic (do NOT delete):** `JAC_DISABLED_PLUGINS`, `get_disabled_plugins`, `[plugins].disabled`, and `jac plugins enable/disable` stay as an escape hatch; only narrow the docs.

## Verification

- **M1, clean env:** in a fresh venv with only `pip install -e jac` (no plugins): `jac create app --kind fullstack` → assert `jac.toml` has `jac-client` under `[dependencies]`; `jac install` → assert `.jac/venv` gets `jac-client`; `jac run` → assert the client plugin's hooks fire. Core kind → `[dependencies]` unchanged vs today.
- **Regression:** `python -m pytest tests/ -x -n 20` in `jac/`, plus plugin repos' test workflows, with `JAC_NO_VENV=1` for the editable dev loop once M2 lands.
- **M2:** in a jaclang-only env, a project pinning a different jaclang runs on the pinned version; `jac purge` and a no-`jac.toml` script run with no provisioning; `JAC_NO_VENV=1` reproduces in-process behavior; second `jac run` on a fresh venv does no pip (lock hit).
- **M3:** `git grep jaseci` returns only intentional/historical hits; release dry-run emits no tier-4 row; `install.sh` on a clean box yields jaclang-only and `jac create … && jac run` self-provisions.

marsninja:
## Summary

Moves Jac from "plugins installed globally alongside jaclang" to **per-project provisioning**. A developer now runs `pip install jaclang` only; plugins (`byllm`, `jac-client`, `jac-scale`, ...) are declared per-project in `jac.toml` and provisioned into the project's own `.jac/venv` — nothing but jaclang lives in the environment the `jac` CLI runs from.

Closes #6824.

## How it works

**M1 — project-venv plugins load (additive, no behavior change for existing setups)**
- After the project `.jac/venv` is added to `sys.path`, plugin discovery is re-run so venv-provisioned plugins register their hooks. `import jaclang` already discovered global-env plugins before the venv path existed, so without this their modules were importable but their hooks never registered. Idempotent via the existing `is_registered` guard.
- Kind-aware `jac create`: the kind's plugin is written into `[dependencies]`, and core now ships **offline stub templates** for the plugin-backed kinds (`fullstack`/`client`/`mobile`/`desktop`) named identically to the plugin templates so the plugin's richer template overrides the stub when present. The previous hard "plugin not installed" error is now a soft notice.

**M2 — hermetic re-exec launcher**
- New top-level `jac/_jac_launcher.py` becomes the `jac` console-script entry (it is intentionally *not* under the `jaclang` package, so it can decide before `jaclang/__init__` runs its global plugin load). When a project declares deps / a plugin-backed kind (or a `.jac/venv` already exists), it provisions the venv (jaclang + declared deps via `uv pip`/pip) and re-execs `python -m jaclang` inside it, giving a single hermetic jaclang per project. Freshness is tracked by `.jac/venv/.jac-lock`.
- Escape hatches: `JAC_NO_VENV=1` (run in-process on the current env — used by the editable monorepo dev loop / CI), `JAC_NO_SYNC=1` / `--frozen` (use the existing venv as-is, don't auto-provision).
- Fast paths stay in-process: no `jac.toml`, env-independent commands (`purge`), and projects that declare nothing to provision (so the repo's own root `jac.toml` is unaffected).

**M3 — cleanup**
- Delete the `jaseci` PyPI meta-package and every reference (`release_utils`, `parse_release`, `check_manifest_sync`, `install.sh`, CI workflows, docs).
- `install.sh` installs jaclang only; the standalone binary is jaclang-only.
- Rewrite `install.md` / `README` and sweep the tutorial docs to the jaclang-only story.

## Validation

Exercised end-to-end against an editable jaclang (no test-suite run):
- A plugin present **only** in `.jac/venv` (absent from the global env) loads through the real CLI.
- `jac create` across all kinds injects the correct `[dependencies]`; both the plugin's rich template and the offline stub path work.
- Re-exec runs under the venv's interpreter; `JAC_NO_VENV` stays global; venv-direct invocation doesn't loop; lock fresh → re-exec, stale → provision → repair.
- Full from-scratch: a jaclang-only env + a project declaring a PyPI dep → `jac run` auto-provisions a venv and runs inside it with the dep importable.
- Repo root stays in-process (no venv created); release scripts produce a correct matrix with no `jaseci`.

## Notes for reviewers / contributors

The editable monorepo dev loop and CI runs should set `JAC_NO_VENV=1` so `jac` in a project directory with declared deps doesn't spin up a `.jac/venv` instead of using the editable install.


marsninja:
## Roadmap analysis: is this the right shape, and what's left to actually hit the goal?

Did a deep read of the branch (launcher, plugin loader, project/config model). The mechanism here is **complete and works** — the launcher provisions `.jac/venv`, re-execs into it, plugins resolve there. But "drop the jaseci meta-package" is the cosmetic half of the goal. The two things that actually make per-project plugins *simple* are untouched, and in one place this PR makes complexity worse. Flagging that as roadmap, not as a blocker.

(Skipping reproducibility/lockfile here on purpose — that's a separate concern; this is about cleanup + achieving the goal.)

### The two structural problems this PR doesn't solve

**1. Plugin loading is still eager — and it's the root cause of all the contortion.**

`import jaclang` discovers and registers plugins from the current env at module-import time (`load_plugins_with_disabling` in `helpers.jac`). Every awkward thing in this PR is a workaround for that one fact:

- the launcher *must* be a separate top-level module that scrupulously avoids importing jaclang (its own docstring says so) — because importing jaclang would load the wrong env's plugins before the venv decision is made;
- so it has to `os.execve` a whole second Python to get a clean import;
- and there's *still* a `_reload_project_plugins()` band-aid in `cli.impl.jac` that invalidates caches and re-scans entry points for the other activation path (sys.path injection without re-exec).

If plugins loaded **lazily** (resolved on first hook dispatch, not at import), none of this is needed: no re-exec for plugin reasons, no two-phase discovery, no "must not import jaclang" tightrope. **This is the highest-leverage cleanup and it's the keystone — most of the rest gets easier or disappears after it.**

**2. The launcher duplicates jaclang's own project logic in a second language.**

This is where complexity went *up*. `_jac_launcher.py` re-implements in pure Python what already exists in Jac:

| Logic | jaclang (Jac) | re-implemented in launcher |
|---|---|---|
| find project root | `config.find_project_root` | `_find_project_root` (L48) |
| kind → plugin map | `KIND_SPECS` in `kinds.jac` | `_KIND_PLUGIN` dict (L36) |
| pip spec formatting | `DependencyInstaller._make_pip_spec` | `_make_pip_spec` (L88) |
| collect deps incl. git | `DependencyInstaller` | `_collect_specs` (L97) |
| uv detection / venv paths | dependencies impl | L195, `_venv_python` (L82) |

The `_KIND_PLUGIN` dict literally carries the comment *"Kept in sync with KIND_SPECS in jaclang/project/kinds.jac"* — a hand-maintained second source of truth is a guaranteed drift bug. We traded a meta-package for a shadow re-implementation of the project model.

### Dev-experience rough edges left behind

- **Double process start on every in-project command.** `jac --help`, `jac run`, everything pays a jac.toml walk + hash + lock-read, then `execve`s a second Python. Noticeable cold-start cost, on *every* invocation.
- **Implicit provisioning is surprising.** First `jac run` in a project can silently trigger a venv build + network install. And on failure it **silently falls back to the global env** (L295), which can mask a broken project as confusing "module not found" errors downstream.
- **Provisioning is magic, not a verb.** `jac install`/`add` exist, but auto-provisioning is wired implicitly into the launcher — the "when does my env change?" mental model is fuzzy.

### Proposed roadmap

Keep this PR as **Phase 0 (packaging + mechanism)** — it's fine. Then:

- **P1 — Lazy plugin loading (keystone).** Resolve plugins on first hook call instead of at `import jaclang`. Lets us delete the re-exec-for-plugins path and `_reload_project_plugins`, and collapses the launcher's reason to exist.
- **P2 — One source of truth for the project model.** Extract a tiny pure-Python `jaclang._bootstrap` that *both* the launcher and CLI import for root-finding / spec-collection / kind-mapping, so `_KIND_PLUGIN`/`_collect_specs`/`_make_pip_spec` stop being a Python clone of the Jac logic.
- **P3 — Make provisioning a legible verb.** Decide auto-provision policy (prompt-on-first-run, or require explicit `jac install`/`jac sync`) and replace the silent global fallback with a clear "this project isn't provisioned" error.
- **P4 — Decouple kind↔plugin.** Let plugins *declare which kinds they provide* (reverse the arrow) so the kind→plugin map isn't hardcoded in two places — needed anyway if third-party plugins ever add kinds.

Net: the packaging change is the easy 20%. P1 (lazy loading) is the part that actually buys the architectural simplicity the goal is really after.

PR #6845, opened 2026-06-21
fix(na-compiler): resolve transitive .na.jac import type loss on incremental builds (#6793)

### LINKED ISSUE #6793: fix(na-compiler): cross-module type resolver misses transitive imports (E5090/E2018)
## Summary

The NA cross-module type resolver (`_register_imported_struct_types` in `na_ir_gen_pass.impl/core.impl.jac`) can miss transitive `.na.jac` imports, causing build failures with `E5090` / `E2018` errors on types that are reached only via deep import chains.

**Status:** Open — workaround in place on [#6355](https://github.com/jaseci-labs/jaseci/pull/6355); compiler fix not yet landed.

## Root Cause

The BFS that builds `imported_mods` seeds sub-imports by calling `iter_context_stmts(imported_mod)` on each discovered module. When a module is pulled in via on-demand compilation (`prog.compile(..., symtab_ir_only=True)`), its AST body can be stripped, so `iter_context_stmts()` returns no `Import` nodes. The BFS adds the module to `imported_mods` but never enqueues its sub-imports, silently cutting off the transitive walk at that node.

**Example chain that breaks:**
```
tui.na.jac → tui_core → input → overlay → select_list (SelectItem)
```
If `overlay` is on-demand compiled with a stripped body, `select_list` is never discovered. Phase 1a skips registering `SelectItem`, and Phase 1b dies with `E2018: Undefined name 'SelectItem'`.

## Symptom

```
error[E5090] Native pathway does not yet support type 'list[<Unknown>]'
error[E2018] Undefined name 'SelectItem' in native type annotation position
```

The error count and which type failed shifts when the entry module's import list changes, making it look nondeterministic (it is config-deterministic).

## Workaround (active on #6355)

Both `tui.na.jac` and `host.na.jac` carry a **"do not prune these imports or the build regresses"** block that imports the full flat module set at the entry level, priming the resolver into a proven-good order. This is brittle (any new leaf type can break it) and duplicated across two files that must stay in sync.

See comments in those files referencing this issue.

## Proposed Fix (not yet on branch)

Add a **fixpoint scan** in `_register_imported_struct_types` that runs after the initial BFS. Re-check every module in `imported_mods` against the already-populated hub until no new modules surface. Hub lookup only — no on-demand compilation, no recursion. Idempotent and safe to run always.

Once landed, remove the priming import blocks in `tui.na.jac` and `host.na.jac`.

**Target:** [#6355](https://github.com/jaseci-labs/jaseci/pull/6355) (or a follow-up compiler-only PR if #6355 is split).

## Related

- PR [#6355](https://github.com/jaseci-labs/jaseci/pull/6355) (feat: TUI for jac ai) — [compiler fixes comment](https://github.com/jaseci-labs/jaseci/pull/6355#issuecomment-4747419134) documents this alongside other `nacompile --shared` fixes
- Companion tech debt: scalar-glob store codegen bug (bare `glob x: str` writes don't persist across `:pub` calls; tracked in comments in `diff.na.jac` / `ipc.na.jac`)

chess10kp:
## Summary

Fixes #6793.

The NA cross-module type resolver dropped transitively-imported struct types on **JIR-cached incremental builds**, failing with `E5090` (`list[<Unknown>]`) and `E2018` (undefined struct in a native annotation position) on types reached only via deep import chains.

## Root cause

JIR does not serialize `Expr.type`, so a JIR-loaded native tree arrives **untyped** and its cross-module link state is degraded — imported struct types resolve to `<Unknown>` during the dep-sweep re-inference for modules reached only transitively.

The bug only surfaces **across process boundaries**: the first `nacompile` writes the JIR cache; a second, separate `nacompile` reads it and fails. A single in-process build is masked because the in-memory hub retains types from the first compile — which is why fresh builds (and single-process test harnesses) never reproduced it.

## Fix

Skip the JIR fast-path (both read and write) for `.na.jac` modules in `TypeEvaluator._import_module_from_path`, so their dependencies are always fresh-compiled. The type evaluator's recursive import resolution then rebuilds full cross-module links, matching fresh-build behaviour. `.na.jac` modules are always symtab-only (no bytecode), so they gained nothing from the JIR cache to begin with.

## Test

Adds a regression test in `test_native_gen_pass.jac` that runs **two separate `nacompile` processes** over a 6-module transitive chain:

```
entry -> core -> handler -> panel -> selector   (selector defines SelectItem)
         state imported as a shared diamond at every chain level; state -> selector
```

The first build writes the JIR cache, the second reads it. Before the fix the second build failed with `E5090`; now both are clean. The diamond topology — `state` imported at every chain level without its symbols being referenced — is the minimal shape that reproduces the bad dep-sweep order, mirroring the real TUI module graph from #6355 where the bug was first hit.

## Verification

- Without the fix: incremental build fails with `error[E5090]: Native pathway does not yet support type 'list[<Unknown>]'`.
- With the fix: both fresh and incremental builds are clean (new regression test passes).

Note: #6793 sketched an alternative approach (a fixpoint scan in `_register_imported_struct_types`). After reproducing the failure, the actual manifestation turned out to be incremental/JIR-cache–specific, so this PR targets that path directly with a smaller change.


chess10kp:
Addressed the review feedback in 82dcf9813:

- **Failing `Contribution Checks`**: added the missing release-note fragment `docs/docs/community/release_notes/unreleased/jaclang/6845.bugfix.md`.
- **P2 — JIR cache cleanup no-op**: kept the loop but rewrote the comment to explain it is defensive cleanup of *pre-fix* caches that an older build may have left on a developer machine (a stale cache would otherwise re-trigger the bug). Post-fix `.na.jac` never writes one, so it is normally a no-op.
- **P2 — loose error filter**: tightened both filters to `re.search(r"\berror\[E\d+\]", line, re.IGNORECASE)` so they no longer match informational references to error codes (and tolerate a capitalised variant).

Regression test still passes locally (`1 passed`).

chess10kp:
Closing: the core fix (skip the JIR fast-path for `.na.jac` modules so transitively-imported native struct types don't degrade to `<Unknown>` on incremental builds) already landed upstream via #6828 — the change in `imported.impl.jac` is functionally identical. No code delta remains against `main`.

PR #6850, opened 2026-06-21
ci(jac-scale): auto-discover backend tests instead of listing by file

MusabMahmoodh:
## Summary

The `test-scale` job invoked backend tests by listing each file explicitly in the workflow. As a result, a test only ran in CI if its path was also added to the workflow by hand. Over time this meant a number of `test_*.jac` files under `jac-scale/jac_scale/tests` were present in the tree but were never executed by CI, so they provided no signal on pull requests.

This change switches the job to directory-level test discovery, in line with how the other packages in this repository run their suites.

## Changes

- Replace the per-file `pytest` invocations with a single discovery run over `jac-scale/jac_scale/tests`, so every test file is picked up automatically and new files are covered without a workflow edit.
- Install all test extras (`aws`, `data`, `scheduler`) and their helpers (`moto`, `testcontainers`) up front, since discovery runs the whole suite in one pass.
- Ignore the Kubernetes-cluster tests (`test_deploy_k8s`, `test_k8s_utils`, `test_pod_env`, `test_image_build`), which require a live cluster and already run in the dedicated `test-scale-k8s` job.

## Status

This is opened as a draft on purpose. Enabling discovery means the previously unrun test files now execute for the first time in CI, and some are expected to fail initially (external service dependencies, or tests that have drifted). The intent is to use this PR's own CI run as the source of truth, then address the failures individually before marking it ready.

## Follow-up: test coverage visibility

To prevent this class of gap from recurring, we should consider adding a test-coverage tool to the pipeline so that unexercised code is visible rather than silent.

One caveat worth flagging early: standard Python coverage tooling (`coverage.py` / `pytest-cov`) may not map cleanly onto Jac. Execution tracing appears to attribute correctly to `.jac` source files, but the reporting layer re-parses source as Python to compute line counts, which is unlikely to work for Jac syntax. We may therefore need a lighter-weight coverage approach suited to Jac, and any coverage gate should be introduced as non-gating (informational) first until its behaviour on `.jac` sources is validated. This is intentionally left out of the current change to keep it focused.


MusabMahmoodh:
### PR summary - current state

CI is green: `test-scale` now runs as readable subsystem groups
(`server` / `data` / `microservices` / `deploy` / `misc`) and the previously
unrun suite passes, with a small explicit quarantine.

#### What this PR does
1. **Auto-discovery** of `jac-scale/jac_scale/tests` instead of a hand-maintained
   file list, so new test files run automatically (the original gap: only ~20 of
   ~78 files ran in CI).
2. **Per-file process isolation** - each file runs in its own pytest process,
   restoring the isolation the suite relies on (this alone cleared ~110
   cross-contamination failures).
3. **Subsystem groups** across runners (with a `misc` catch-all so a new file is
   never silently skipped) - replaces opaque shard indices; failures are readable
   per subsystem.
4. **`--extras test`** install so every service/client dependency is present
   (redis, pymongo, prometheus-client, kubernetes, testcontainers, moto).

#### Fixes included (surfaced by auto-discovery; pre-existing, not caused here)
- **Real product bug:** `PrometheusMetricsCollector` used the global prometheus
  registry -> duplicate-timeseries collision; now owns an isolated registry.
- **Test fixes:** 4 perf tests' stale `register_user` helper (nested
  identities/credential + login); `ResourceStatus` `.value`; `prometheus_scrape_port`
  expectation; static `KubernetesConfig.from_dict(cls, ...)`; lazy-import patch
  path for `KubernetesTarget`; `hasattr(init)` (Jac maps `init` to the
  constructor); and a shared Jac quirk where a test-local `obj`/`node` breaks
  dataclass processing -> hoisted to module scope (gateway, sv_auth_forward,
  direct_db).

#### Quarantined (6) - tracked in #6853
Explicitly skipped in the workflow (logged as `QUARANTINED`, not silent), to be
fixed under the audit issue: `connection_overhead`, `redis_batch_overhead`,
`redis_memory_leak`, `topology_slice_pushdown`, `file_upload`, `user_identity`.
One of these (`connection_overhead`) encodes a **real product perf finding**: the
system root is read from MongoDB on every request instead of being cached.

#### Follow-ups (in #6853)
Fix the 6 quarantined tests; Jac-core quirks (`init`/constructor collision,
test-local archetype `__module__`); and split unit vs integration tiers.

Issue #6852, opened 2026-06-21
Single self-contained jac binary as the only jaclang distribution (Zig launcher + dlopen embed; jac-native plugins; native-migration substrate)


marsninja:
> **This issue was substantially revised.** It supersedes the original
> *static-link + `jac-pylib` source-wheel* plan. Two things changed after
> prototyping and design: (1) the embed pivoted from **static-linking** libpython
> to **`dlopen`ing** a bundled shared libpython at runtime, and (2) the
> distribution model moved from "binary + a `jaclang` PyPI wheel for plugins" to a
> **complete clean break: the single binary is the *only* jaclang artifact**, with
> a fully jac-native plugin workflow. Prototype (Phase 1): #6855.

## Goal

Ship jaclang as **one self-contained `jac` binary** and make it the *sole*
distribution — no system Python, uv, or pip at install or runtime, and **no
`jaclang` package on PyPI at all**. The binary is also the foundation for
gradually moving the compiler itself onto the native (LLVM) pathway.

Two tracks ride on the same artifact:
1. **Distribution clean break** — the binary is the only jaclang; plugins are
   created, tested, bundled, and installed entirely through `jac`; PyPI is used
   only to *host* jac-built plugin wheels.
2. **Native-migration substrate** — the binary is built to host a
   shrinking-Python / growing-native runtime (`.jac` → `.na.jac`), toward an
   eventual self-hosted compiler.

## Locked decisions

| # | Decision | Choice |
|---|---|---|
| 1 | Loader language | **Zig** |
| 2 | Embed model | **`dlopen` shared libpython at runtime** (not static link) |
| 3 | Distribution | **Single binary is the ONLY jaclang artifact** — no `jaclang` on PyPI |
| 4 | Plugins | **jac-native workflow** (`jac create`/`add`/`bundle`/`test`); loaded into the binary via the project env |
| 5 | PyPI's role | **plugin host only** — `jac bundle` builds wheels, PyPI distributes them |
| 6 | `jac-pylib` | **dropped** (no source-wheel facade) |
| 7 | Targets | linux-x86_64 + macos-aarch64 |
| 8 | Project layout | `jac/` is a Zig project (`jac/build.zig`, `jac/launcher/`) |
| 9 | Zig's role | launcher **+** the runtime/LLVM (`libllvmlite`) FFI substrate native code links against |

## Approach: dlopen, not static-link

Rather than reconstructing CPython's static link line from `PYTHON.json` (100+
[...]

PR #6855, opened 2026-06-21
feat(jac): prototype single-binary jac via Zig launcher + dlopen embed

marsninja:
## Summary

Prototype of jaclang shipping as a single self-contained binary, built as a Zig
project at `jac/`. One `zig build` produces a native `jac` launcher that
**dlopens a bundled shared libpython at runtime** and runs the jaclang core
in-process -- no system Python, uv, or pip needed at install or runtime.

Reference: #6852 (the dlopen variant of Phase 1-2).

## Approach

Rather than statically reconstructing CPython's link line from `PYTHON.json`
(100+ objects, bundled archives, OS frameworks), the launcher links only libc
and `dlopen`s libpython from its self-extracting payload -- the same runtime
dynamic-loading model jac-native already uses for LLVM/native code (llvmlite +
ctypes). The build stays trivial and cross-platform (no per-OS framework
enumeration; `dlopen` is uniform across `.so`/`.dylib`); pbs is only a payload
input, never linked.

Binary shape: `[ launcher stub (libc only) ][ runtime.tar.zst ][ trailer ]`,
trailer = `magic | payload_len(u64 LE) | sha256_hex`.

## What's here

- `launcher/runtime.zig` -- payload materialization (trailer parse, cache
  resolution, zstd+tar extract into `~/.cache/jac/rt/<hash>`, stale-version GC).
  Pure Zig, unit-tested.
- `launcher/launcher.zig` -- process entry: materialize, `dlopen` libpython,
  `dlsym` ~6 `Py_*` entry points, run the boot dance. No `@cImport`, no Python
  headers, no pbs at build time.
- `launcher/pack.zig` -- concat stub + payload + trailer into the final `jac`.
- `launcher/mkpayload.sh` -- assemble the runtime payload: pip-install jaclang
  from source, precompile the JIR bundle, stage CPython + site (keeping
  `lib-dynload`), tar + zstd.
- `build.zig` / `build.zig.zon` -- `zig build` (jac), `zig build stub`,
  `zig build test`.

## Validation (macos-aarch64)

- Launcher stub ~2.3 MB, links only `libSystem` (no libpython at link time).
- `jac --version` and `jac run` (obj + methods + comprehensions) work from a
  clean `HOME`.
- Warm start ~0.3s; the shipped `_precompiled` JIR bundle (334 modules) yields
  **0 live compilations** on a cold run.
- `zig build test` -- 4/4 unit tests (trailer parse, cache resolution, an
  end-to-end zstd+tar materialize round-trip).

## Notes / follow-ups

- Tradeoff vs a static embed: libpython is a shared lib inside the
  self-extracting payload (materialized to cache on first run), not baked into
  the image. Still one distributable file, still no system Python.
- Toolchain pinned to Zig 0.16.0.
- Linux: stage the shared lib as `libpython3.12.so` (pbs may ship
  `libpython3.12.so.1.0`).
- Cold start is dominated by payload extraction + first-time JIR cache
  laundering, not compilation; sealing the runtime (#6852 Phase 4) is the
  further win.


marsninja:
## Linux CI: binary test gate is red on the xdist worker crash

Full diagnosis + repro + fix options committed at [`jac/launcher/XDIST_LINUX_NOTES.md`](jac/launcher/XDIST_LINUX_NOTES.md) (pull the branch on the Linux box). Summary:

**What's failing:** only the new blocking gate `jac single-binary self-test` → step *"Full repo suite via the binary"*. pytest exit code **5 ("no tests ran")** because every xdist worker crashes at startup:
```
ModuleNotFoundError: No module named '_decimal'     # -> _contextvars
xdist: maximum crashed workers reached: 16
```
Not a test failure -- the worker subprocesses can't import the C-extension stdlib modules from `lib-dynload`.

**Why macOS was green:** on macos-aarch64 pbs `_decimal`/`_contextvars` are **builtins**; on linux-x86_64 pgo-full pbs they're **shared `.so` in lib-dynload**, which the worker isn't finding. So `-n auto` passing locally was a false positive for this path.

**The one diagnostic that decides the fix** (run on linux-x86_64 after `zig build` + materialize):
```bash
"$BIN" -c "import sys; print([p for p in sys.path if 'dynload' in p]); import _decimal, _contextvars; print('WORKER IMPORT OK')"
```
- fails -> launcher/worker-mode env bug (fix in `launcher.zig`: drive worker via explicit `PyConfig` with `lib-dynload` in `module_search_paths`, instead of relying on `PYTHONPATH` under `Py_BytesMain`).
- prints OK -> execnet/xdist isn't propagating `PYTHONPATH` to workers (fix at the pytest layer).

**Immediate unblock** if you just want green to proceed: set `JAC_TEST_JOBS=auto` -> `0` (serial) in the gate step. Slower, no workers, no crash.

After green: drop the redundant `test-solid-jsdom`/`test-desktop-native` jobs (keep `test-client` for playwright e2e), migrate the remaining `pip install -e ./jac` workflows, then delete `jac/pyproject.toml`.

PR #6869, opened 2026-06-22
[byllm] retry transient mid-stream drops on the same model

MalithaPrabhashana:
## Problem

Streaming `by llm()` calls on an Anthropic model intermittently hang and then fail the whole turn. The streamed response stalls or drops mid-flight (a silent read-timeout, a dropped connection such as `incomplete chunked read`, or a brief 5xx), and the single-model path (`SessionAwareModel(Model)`, not `ModelPool`) handled neither case:

- no explicit `timeout` on the litellm call, so a stalled stream waited out litellm's ~600s default before failing, and
- no retry on transient transport errors (`num_retries` only applied to `ModelPool`).

So a single transient blip became a ~10 minute hang followed by a hard turn failure. (Surfaced downstream as jaseci-labs/jacBuilder#546.)

## Fix

- **Inter-chunk read timeout** (`make_model_params`): streaming calls get an `httpx.Timeout(read=...)`. It resets on every chunk, so a healthy long generation is never cut, while a silent stall fails in `read_timeout`s instead of ~600s. Non-streaming is left on the default (one read for the whole body, where a read timeout would cap total generation time).
- **Same-model retry** across all five dispatch paths (`dispatch_streaming`, `dispatch_streaming_with_tools`, `adispatch_streaming`, `dispatch_no_streaming`, `adispatch_no_streaming`). Only transient transport errors retry; auth / rate-limit / not-found and everything else propagate unchanged. No silent fallback to a different model.
- **Reset on retry**: a streaming retry resets the partial accumulators and bumps `MTRuntime._stream_reset_count`; the streaming producers emit a `stream_reset` event so consumers discard tokens already shown and the regenerated output is not doubled.
- **Config**: new `[plugins.byllm.streaming]` section (`read_timeout`, `connect_timeout`, `num_retries`, `retry_backoff`) with sane defaults; per-call override via `stream_num_retries`.

Prompt caching already in place makes the retry cheap (the cached prefix is reused).

Verified with a new regression test (`tests/test_byllm.jac` + `tests/fixtures/stream_retry.jac`) covering retry-then-recover, `stream_reset` emission, output de-duplication, and retry-budget exhaustion; the existing byllm suite passes and `jac format --check` is clean.


udithishanka (approved):
Changes LGTM. better to verify with a PR Preview deployment with jachammer.

MalithaPrabhashana:
Yeah verified with jacHammer deployment and previous issue didn't noticed

PR #6875, opened 2026-06-22
test(jac-scale): restore connection-overhead test as efficiency guards (#6862)

### LINKED ISSUE #6862: test_connection_overhead: verify MongoDB connection reuse + system-root caching per request
## Summary
`test_connection_overhead` was quarantined when jac-scale CI switched to auto-discovery (PR #6850). It contains two independent checks that now need verification.

## What this test verifies
That an authenticated request does **not** trigger redundant MongoDB work:
1. The per-request availability probe does not open a fresh MongoDB connection each time.
2. The system root node is not re-fetched from MongoDB on every request.

- **Availability probe (case 1)** looks already fixed: `MongoBackend` now caches the client in `_process_cache`, so 10 requests created **0** new connections. The test still asserts the *old* buggy behaviour (`>= 8` new connections), so it fails. -> likely just needs **inverting** into an efficiency guard.
- **System root (case 2) looks like a real, still-present perf issue**: 10 read-only requests issued ~4 MongoDB queries each because the system root is read from Mongo on every request instead of being cached. This is the assertion failing, and it points at a genuine latency/throughput cost.

## What needs to happen
- [ ] Re-run against a live MongoDB and confirm the connection count (case 1) is ~0 -> invert that assertion to assert efficiency.
- [ ] Decide on the system root (case 2): cache it per process (product fix) **or**, if the per-request read is intentional, adjust the test to the expected number.
- [ ] Move `test_connection_overhead.jac` out of `tests/quarantine/` back into `tests/data/` once green.

## Refs
- File: `jac-scale/jac_scale/tests/quarantine/test_connection_overhead.jac`
- Code: `impl/memory_hierarchy.mongo.impl.jac` (is_available / `_process_cache`), `impl/context.impl.jac` (system root load)

akindu-k:
## What was wrong

`jac-scale` serves every request statelessly: a fresh context and memory are built per request on top of three tiers (L1 in-process, L2 Redis, L3 MongoDB).

There used to be two performance bugs:

1. Every request opened a brand-new MongoDB connection just to check "is the database reachable?", separate from the real connection pool.
2. Every request re-read the immutable system root node from MongoDB, even though it never changes.

A test, `test_connection_overhead.jac`, existed to catch both. But both bugs were already fixed in the product some time ago (the client is now shared per process and its availability is cached; the system root is cached per process and seeded into L1). The test was never updated, so it still asserted that the *old, buggy* behaviour was present. When CI switched to auto-discovering tests (#6850), the test started failing and was parked in `tests/quarantine/` with this issue to decide its fate.

So the failure was not a product bug. The test was simply checking for a problem that no longer exists, and one of its two checks was also measuring the wrong thing (total query count, which legitimately includes the user's own root and data reads, instead of the system root specifically).

## How I fixed it

I confirmed against a live MongoDB (Docker) what actually happens today:

- 10 requests open **0** new MongoDB connections.
- The system root is fetched from MongoDB **0** times after warm-up (the 4 queries per request are the user lookup, alias preload, quarantine check, and the user's own root node, none of which are the system root).

Then I rewrote the test so it guards the *correct* behaviour instead of asserting the old bug, and moved it back into the live suite:

1. **Connection reuse check** now asserts that N authenticated requests open at most one new MongoDB connection, so a future regression that reintroduces a per-request throwaway client or uncached probe is caught.
2. **System root check** now turns on MongoDB server-side profiling and asserts that zero `find(_id = SUPER_ROOT)` queries happen after warm-up, directly proving the system root is served from cache and never re-read. It also asserts the profiler actually captured the per-request user-root reads, so the check can never pass vacuously.
3. Moved the file from `tests/quarantine/` back to `tests/data/`.

## Verification

All checks were run against a **real MongoDB container** (Docker, `mongo:latest` spun up via `testcontainers` `MongoDbContainer`), with a real `jac start` server subprocess handling actual HTTP requests, not mocks.

**Before (the quarantined test on current code):** both checks failed, because they demanded the old bug be present.

```
FAIL: AVAILABILITY PROBE: 10 requests created 0 new MongoDB connections (expected >= 8)
FAIL: COLD SYSTEM ROOT: 10 read-only requests issued 40 MongoDB queries total (4.0/request)
```

**Ground truth I measured** by enabling MongoDB server-side profiling and inspecting every query over 10 requests:

```
 10  query jac_db.users               (user lookup / JWT validation)
 10  query jac_db._anchors_aliases    (alias preload)
 10  query jac_db._anchors_quarantine (quarantine check)
 10  query jac_db._anchors  _id=de89...  (the user's OWN root node)

system root (_id=00000000-...-0) find count over 10 requests: 0
new MongoDB connections over 10 requests: 0
```

This proved the system root is never re-read (the 4 queries/request are all other, legitimate reads) and no per-request connection is opened.

**After (the rewritten test):** both checks pass.

```
Ran 2 tests in ~18s
OK
```

**Non-vacuity check:** to prove the system-root guard actually catches a regression, I temporarily disabled the cache seeding in `context.impl.jac`. The test failed exactly as intended, then passed again once reverted:

```
FAIL: SYSTEM ROOT RE-FETCH REGRESSION: 10 MongoDB finds for
_id=00000000-0000-0000-0000-000000000000 across 10 post-warm-up requests (expected 0)
```

## Notes for reviewers

This PR contains **no product code change**. The performance fixes already shipped earlier; this only restores the regression guard so those fixes stay protected.

image

Closes #6862


kugesan1105 (commented):
**Overall: this is the right approach and looks good to merge.** Converting the quarantined "proof" test (which asserted the old bugs still existed) into guards for the correct behavior is exactly the pattern the testing-trophy migration (#7628) established, and both guards protect real product behavior: the shared `MongoClient` + cached `is_available()` in `memory_hierarchy.mongo.impl.jac`, and the `_process_cache['system_root']` seeding in `context.impl.jac`. It's also not redundant: `test_memory_hierarchy.jac` only checks `_process_cache['system_root']` white-box in-process, and the connection-reuse path had no other coverage at all.

The anti-vacuity design (liveness guard on served requests, profiler-activity guard before the zero-finds assert, plus the demonstrated failure when cache seeding is disabled) is genuinely good test hygiene.

One note for reviewers: the deletion half of this diff is now a no-op — #7628 already removed `tests/quarantine/test_connection_overhead.jac` from main after this branch was cut, so the effective change is purely the new guard file. Might be worth a line in the PR description to avoid confusion.

I've left three minor, non-blocking inline comments (env-var restore, a clarifying comment on per-test containers, and one hardening idea for the profiler filter).

jac/jaclang/scale/tests/data/test_connection_overhead.jac
```diff
@@ -0,0 +1,227 @@
...
+    # would instead create one connection per request.
+    MAX_NEW_CONNECTIONS = 1;
+
+    data_dir = FIXTURES_DIR / ".jac" / "data";
+    if data_dir.exists() {
+        import shutil;
+        shutil.rmtree(data_dir);
+    }
+
+    mongo_container = MongoDbContainer("mongo:latest");
```
kugesan1105, line 88: Non-blocking: sibling tests (e.g. `test_memory_hierarchy.jac`) share one Mongo container across tests via `_get_mongo()` + `atexit`, while this file boots a fresh container per test. Here the hermetic instance is actually justified — connection counting via `totalCreated` and db-wide profiling would be noisy on a shared container — but a one-line comment saying so would prevent a future "optimization" that consolidates the containers and quietly breaks these guards.

jac/jaclang/scale/tests/data/test_connection_overhead.jac
```diff
@@ -0,0 +1,227 @@
...
+    admin_client = None;
+    server = None;
+
+    try {
+        mongo_container.start();
+        mongo_uri = mongo_container.get_connection_url();
+        os.environ["MONGODB_URI"] = mongo_uri;
+        os.environ.pop("REDIS_URL", None);
```
kugesan1105: Non-blocking: `REDIS_URL` is popped here and never restored, and the `finally` block pops `MONGODB_URI` unconditionally rather than restoring any prior value. If these tests run in the same process as others that had those vars set, that's a small cross-test pollution risk. `test_memory_hierarchy.jac` uses an `env_backup` dict restored in `finally` — same pattern would fit here (applies to both tests in this file).

jac/jaclang/scale/tests/data/test_connection_overhead.jac
```diff
@@ -0,0 +1,227 @@
...
+        # Sanity: the profiler must be capturing finds (the user root is read
+        # every request), else a 0 system-root count would be vacuously true.
+        assert total_anchor_finds >= REQUEST_COUNT , (
+            f"PROFILER INACTIVE: only {total_anchor_finds} _anchors finds captured "
+            f"for {REQUEST_COUNT} requests; expected >= {REQUEST_COUNT}. "
+            f"The system-root assertion below would be meaningless without a live "
+            f"profiler."
+        );
+
+        assert super_root_finds == 0 , (
```
kugesan1105, line 248: Non-blocking hardening idea: the non-vacuity check above proves the profiler is capturing `_anchors` finds, but not that the `command.filter._id == <string uuid>` filter matches the stored `_id` encoding. If anchor `_id` storage ever changed representation (e.g. string → BSON UUID), `super_root_finds` would be 0 vacuously even with the regression present. You could close that gap by also counting the user-root reads through the same `id_filter` path and asserting `>= REQUEST_COUNT`, so the filter mechanism itself is proven live.

jac/jaclang/scale/tests/data/test_connection_overhead.jac
akindu-k: Added a one line comment in both tests noting the container is per test on purpose, so a later cleanup does not consolidate them and quietly break the counting.

jac/jaclang/scale/tests/data/test_connection_overhead.jac
akindu-k: Switched both tests to the env_backup pattern from test_memory_hierarchy.jac. MONGODB_URI and REDIS_URL are saved and removed up front, then restored in finally, so nothing leaks into tests that run after.

jac/jaclang/scale/tests/data/test_connection_overhead.jac
akindu-k: Good catch. The test now reads the user root id from jac_db.users and counts its reads through the same id filter, asserting it is at least REQUEST_COUNT before the zero for the system root is trusted. So if the _id encoding ever changes, that check fails first instead of the system root check passing for the wrong reason.

While doing that I also moved the matching out of the server side query into a small helper that reads command.filter._id and normalises it to strings, handling both an equality filter and an $in filter. The old version would have counted zero if a read ever moved to the batch $in path, which was a second way the guard could go quiet.

PR #6877, opened 2026-06-22
test(jac-scale): invert obsolete redis batch test into an MGET efficiency guard (#6863)

### LINKED ISSUE #6863: test_redis_batch_overhead: invert obsolete bug-evidence test now that batch_get uses MGET
## Summary
`test_redis_batch_overhead` was quarantined in PR #6850. It is an obsolete "prove-the-bug" test: the optimisation it was written to demand has since shipped.

## What this test verifies
That reading N anchors from the L2 Redis cache uses a single batched `MGET`/pipeline rather than N individual `GET` round-trips.

The test asserts the *old* inefficient behaviour (Redis command count grows with N, `delta >= ANCHOR_COUNT`). But `batch_get()` already uses a single `mget` (`memory_hierarchy.redis.impl.jac`), so the measured command count is a small constant (~3) and the assertion fails. The test's own message even says: *"If delta < ANCHOR_COUNT the fix is already applied."*

## What needs to happen
- [ ] Re-run against a live Redis and confirm `batch_get` issues a constant number of commands regardless of N.
- [ ] **Invert** the assertions into an efficiency guard (command count stays ~constant / does not scale with N), so it protects the MGET optimisation from regressing.
- [ ] Move the file out of `tests/quarantine/` back into `tests/data/` once green.

## Refs
- File: `jac-scale/jac_scale/tests/quarantine/test_redis_batch_overhead.jac`
- Code: `impl/memory_hierarchy.redis.impl.jac` (`batch_get` -> `mget`)

akindu-k:
## What was wrong

There was a test, `test_redis_batch_overhead`, whose only job was to **prove a bug existed**. The bug: when the runtime loaded N graph nodes from the Redis L2 cache, it fetched them one at a time, firing N separate `GET` commands. That is N network round-trips for what should be one fetch.

To prove that, the test seeded N items, read them back, looked at Redis's command counter, and asserted the count grew by **at least N** ("yes, one command per node, the bug is here"). That assertion only passes while the code is slow. The test even said so in its own failure message: *"If delta < ANCHOR_COUNT the fix is already applied."*

The fix shipped a while ago: `batch_get()` now does a single `MGET` for the whole batch. So the command count stopped growing with N, the old assertion started failing, and PR #6850 had to move the test into `tests/quarantine/` to stop it breaking CI. In short, the test was frozen pointing the wrong way: it guarded the bug instead of guarding the fix.

## How it was fixed

The test now asserts the **opposite** property, which is the thing we actually want to protect forever: reading N anchors from L2 costs a constant number of Redis commands, no matter how big N is.

Three changes, not a one-line operator flip:

1. **Precise measurement instead of a noisy total.** The old test read `total_commands_processed`, which counts *every* command Redis handled (auth probes, root loads, heartbeats), so it needed fudge factors like "delta > 10". The new test reads `INFO commandstats` and isolates the exact `mget` and `get` call counts. The `INFO` call itself is neither a GET nor an MGET, so it never pollutes the numbers.

2. **Assert the invariant, not a magic number.** It reads a small population (5) and a large population (25) and computes the per-anchor marginal cost. With MGET that is about 0; with one GET per node it is about 1. The check `marginal < 0.5` is the clean inversion of the old `ratio >= 2.0`. Because it compares two sizes, all the constant per-request overhead cancels out, so there is no brittle absolute ceiling that rots over time. A second check confirms MGET is actually used, so a regression cannot slip through by accident.

3. **Removed the duplication and re-homed the file.** The original was two nearly identical ~150-line tests. This is one test with the measurement logic pulled into a small helper. The file moves from `tests/quarantine/` back to `tests/data/` alongside the other live integration tests.

## Verification

Run against real `redis:latest` and `mongo:latest` containers via testcontainers:

| Implementation | Result |
| --- | --- |
| Real `batch_get` (single MGET) | **PASS** (`Ran 1 test in 14.1s, OK`) |
| Temporarily reverted to one GET per anchor | **FAIL** (`mget=0`, caught by the MGET guard) |

This proves the guard is real: it passes on the current optimized code and fails the moment someone reverts `batch_get` to N individual GETs. The production code was only patched temporarily to prove the failure case and was restored unchanged.

Closes #6863


akindu-k:
Screenshot 2026-06-22 142318

kugesan1105:
use `server_support`, don't inline it

`get_free_port`, `extract`, `register_user`, `call_walker`, and `start_server `— all of which exist in server_support.jac

Issue #6871, opened 2026-06-22
`auth_redirect` config in file-based routing has never worked


SahanUday:
## Summary

When running an app that uses file-based routing (`pages/` directory), the following warning always appears in the terminal on every build/run:

```
⚠ Failed to load routing config, using default auth_redirect='/login': cannot import name 'PluginConfig' from 'jaclang.runtimelib.client.config_loader'
```

<img width="897" height="100" alt="Image" src="https://github.com/user-attachments/assets/486237a4-6546-45a1-aea5-69f872c30bb2" />

The app still starts and works correctly, but the `auth_redirect` configuration value has never been readable since file-based routing was introduced in PR #4422. There are three stacked bugs in `_create_pages_entry_content()` inside `compiler.impl.jac`.

---

## Affected file

`jac/jaclang/runtimelib/client/impl/compiler.impl.jac` — method `_create_pages_entry_content`, lines ~133–145

```jac
auth_redirect = "/login";
try {
    import from jaclang.runtimelib.client.config_loader { PluginConfig }   # Bug 1
    config = PluginConfig.load(self.project_dir);                           # Bug 2
    if config and config?.routing and config.routing {
        auth_redirect = config.routing.get('auth_redirect', '/login');      # Bug 3
    }
} except Exception as e {
    console.warning(
        f"Failed to load routing config, using default auth_redirect='/login': {e}"
    );
}
```

---

## Bug 1 — Wrong class name: `PluginConfig` does not exist

The import asks for `PluginConfig` but the class in `jaclang.runtimelib.client.config_loader` has always been named `JacClientConfig`. There is no `PluginConfig` at that path. This causes an `ImportError` every time and is the direct cause of the warning the user sees.

**Verified**: I checked the state of `config_loader.jac` at the exact commit of PR #4422 (the commit that introduced this code, `da57cd2f`) and at the refactoring commit PR #6390. The class has only ever been `JacClientConfig`.

---

## Bug 2 — Wrong call pattern: `load()` is an instance method, not a classmethod

```jac
config = PluginConfig.load(self.project_dir);
```

Even if Bug 1 were fixed, this would still fail. `JacClientConfig.load()` is an instance method — it takes `self` as its only argument. The correct usage is:

```jac
cfg = JacClientConfig(project_dir=self.project_dir);
config = cfg.load();
```

[...]

PR #6878, opened 2026-06-22
fix(routing): repair auth_redirect config, broken since file-based routing shipped

SahanUday:
## Summary

`auth_redirect` in `jac.toml` has never worked since file-based routing was introduced in PR #4422. Four stacked bugs prevented the configured value from ever reaching the generated `_entry.js`. Users always got redirected to `/login` regardless of their config.

## Bugs fixed

**Bug 1 - Wrong import in `compiler.impl.jac`**
`PluginConfig` does not exist; the class has always been `JacClientConfig`. Caused an `ImportError` on every build, caught silently by the `except` block.

**Bug 2 - Wrong call pattern in `compiler.impl.jac`**
`PluginConfig.load(self.project_dir)` called `load()` as a classmethod, passing a `Path` as `self`. `load()` is an instance method. Fixed to `JacClientConfig(project_dir=...).get_section('routing')`.

**Bug 3 - `routing` key dropped in `config_loader.impl.jac`**
`JacClientConfig.load()` builds a `user_config` dict that was missing the `routing` key entirely, so `[plugins.client.routing]` values from `jac.toml` were silently discarded. Fixed by adding `routing` to both `get_default_config()` and `user_config`.

**Bug 4 - Wrong `AuthGuard` call convention in `react_entry.jac`**
`build_pages_react_entry_script()` generated `AuthGuard("/dashboard")`, passing a plain string as `props`. Since PR #3736, `AuthGuard` compiles with the props-destructure ABI: `function AuthGuard(props) { const {redirect} = props; }`. `props.redirect` on a string is `undefined`, falling back to `"/login"`. Fixed to `React.createElement(AuthGuard, { redirect: "/hora" })`. This is the same fix PR #6494 applied to `.cl.jac` call sites, now applied to the string-generation path in `react_entry.jac`.

## Files changed

- `jac/jaclang/runtimelib/client/impl/compiler.impl.jac` - Bugs 1, 2
- `jac/jaclang/runtimelib/client/impl/config_loader.impl.jac` - Bug 3
- `jac/jaclang/runtimelib/react_entry.jac` - Bug 4

## How to verify

Add to `jac.toml`:
```toml
[plugins.client.routing]
auth_redirect = "/dashboard"
```

Before this fix: unauthenticated users always land on `/login`.
After this fix: unauthenticated users land on `/dashboard`.

### DIFF
--- a/docs/docs/community/release_notes/unreleased/jaclang/6878.bugfix.md
+++ b/docs/docs/community/release_notes/unreleased/jaclang/6878.bugfix.md
@@ -1,0 +1,1 @@
+- **Fix: `auth_redirect` in `jac.toml` now works in file-based routing**: Setting `[plugins.client.routing] auth_redirect = "/dashboard"` now correctly redirects unauthenticated users to the configured path instead of always falling back to `/login`.
--- a/jac/jaclang/runtimelib/client/impl/compiler.impl.jac
+++ b/jac/jaclang/runtimelib/client/impl/compiler.impl.jac
@@ -140,10 +140,11 @@ impl ViteCompiler._create_pages_entry_content(
     }
     auth_redirect = "/login";
     try {
-        import from jaclang.runtimelib.client.config_loader { PluginConfig }
-        config = PluginConfig.load(self.project_dir);
-        if config and config?.routing and config.routing {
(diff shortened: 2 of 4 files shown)


SahanUday:
#6871

PR #6883, opened 2026-06-22
fix(jac-scale): ensure the referenced ServiceAccount exists in microservice deploy

udithishanka:
## Problem

The microservice deploy generates Deployment pod specs that set `serviceAccountName` from `[plugins.scale.kubernetes] service_account_name` (`microservice/manifest_builder.jac:960-962`), but jac-scale **never creates that ServiceAccount**.

Before #6800 this was latent — `deploy()` applied the manifests and returned without waiting, so callers could provision the SA out-of-band afterwards. **#6800 added a blocking fleet-readiness wait** (`_wait_for_fleet_rollout`, up to 1200s, `target.jac`). Now, on a fresh namespace, the ReplicaSet controller rejects every pod:

```
pods "<app>-..." is forbidden: serviceaccount "<app>-sa" not found
```

→ the fleet never becomes ready → the wait fails → `jac start --scale` exits 1, **before** any out-of-band SA creation can run. Result: every fresh-namespace microservice deploy that relies on a named SA fails.

## Fix

Read-or-create a bare ServiceAccount before applying the Deployments, in `microservice/target.jac` `apply_manifests` (right after `ensure_namespace_exists`). This mirrors the **monolith** path (`kubernetes_target.jac:465-471`), which already ensures its own ServiceAccounts.

- Only acts when `service_account_name` is set (empty → pods use the namespace default SA, unchanged).
- `create-if-missing` (read → on 404 create); never overwrites an existing SA, so caller-managed RBAC/IRSA bindings are preserved.

## Why this is the right layer

jac-scale generates the pod spec that references the SA, so jac-scale should ensure the SA exists — consistent with how it already manages the namespace, user Secret, source ConfigMap, PVCs, and (in the monolith path) ServiceAccounts. Callers shouldn't have to pre-create an SA to satisfy a reference jac-scale itself emitted.

## Validation

- `jac format` clean (no diff beyond the addition).
- `jac check`: no new `E0` parse errors (only pre-existing `E1xxx` type-checker noise on the external `kubernetes` client types).
- Reproduced on a live EKS namespace: with the SA absent, deploy fails exactly as above; creating the SA unblocks pod creation and the fleet rolls out normally.


udithishanka:
Addressed the review:

- **automountServiceAccountToken: False** added to the SA body, matching the monolith path's secure default (`kubernetes_target.jac:463/917`). Pods that need a token (e.g. builder-sv for sandbox spawning) opt in at the pod-spec level, which overrides the SA setting.
- **409 guard** on `create_namespaced_service_account` so a concurrent deploy creating the SA between our read and create no longer crashes the deploy.
- **Tests added**: non-404 read error propagates, 409-on-create is swallowed, and the `automountServiceAccountToken: False` default is asserted. 45 tests pass; the `creates when missing` test still fails if the create is removed (verified).

udithishanka:
Refactor per review feedback — reuse instead of duplicate:

The monolith `KubernetesTarget._create_service_account` already does read-or-create **with `automountServiceAccountToken: False`**, and the microservice target inherits it. Replaced the new `ensure_k8s_service_account` helper with a call to `self._create_service_account(...)`:

- **Smaller diff** — the helper (decl + impl in `kubernetes_utils`) is removed; only a call site remains.
- The earlier **automountServiceAccountToken: False** note is now handled for free (it's already in the monolith helper).
- Tests moved onto the shared `_create_service_account` (which was previously **untested**), so this now covers the monolith path too. They call it unbound with a minimal `self` to stay pure unit tests (no cluster). 42 tests pass; the missing-SA test fails if the create is removed (verified).
- Dropped the explicit 409 guard from the previous round to stay consistent with the established read-or-create helpers (the monolith one doesn't guard 409 either).

marsninja:

PR #6885, opened 2026-06-22
fix(scale): bound L2 Redis memory on every deployment (#6864)

### LINKED ISSUE #6864: test_redis_memory_leak: L2 Redis keys have no TTL by default - verify / set expiry policy
## Summary
`test_redis_memory_leak` was quarantined in PR #6850. Unlike the other perf tests, this one appears to catch a **real, still-present issue**.

## What this test verifies
That anchor keys written to the L2 Redis cache have a TTL, so Redis memory does not grow unbounded as the graph grows.

The test asserts every key has a TTL (`ttl != -1`). It fails because `redis_default_ttl` defaults to `0`, which means keys are written with **no expiry (TTL = -1)** - they never expire. As the graph grows, Redis memory grows without bound. This is a genuine reliability concern, not a stale test.

## What needs to happen
- [ ] Verify against a live Redis that default-config anchor keys really have TTL = -1.
- [ ] Decide the product behaviour: set a sensible `redis_default_ttl` (or document/configure LRU eviction) **or**, if unbounded retention is intentional, change the test to assert the documented policy.
- [ ] Move the file out of `tests/quarantine/` back into `tests/data/` once resolved.

## Refs
- File: `jac-scale/jac_scale/tests/quarantine/test_redis_memory_leak.jac`
- Code: `impl/memory_hierarchy.redis.impl.jac` (`_get_db_config` / `redis_default_ttl`)

akindu-k:
Closes #6864

## What was the problem

Jac-Scale stores graph data in three tiers: an in-process cache (L1), Redis (L2), and MongoDB (L3, the authoritative store). Every anchor that is touched gets written into Redis so the next read is fast.

The worry in #6864 was that these Redis keys could pile up forever and eventually run Redis out of memory. Looking at the code, the real situation was a bit more specific than the issue first described:

- Anchor keys already expire by default. `redis_default_ttl` was changed from `0` to `3600` seconds a while back, so on current `main` cached anchors are written with a one-hour expiry, not "never". The issue's headline ("keys have no TTL") described the old behaviour.
- But a TTL is not actually a hard memory bound. It bounds memory by write-rate, not by a ceiling. The real safety net for a cache that sits over an authoritative store is a server-side `maxmemory` limit plus an eviction policy (an evicted key is just a cache miss, since MongoDB still has it).
- That eviction policy was only configured for the Redis that Jac-Scale provisions itself in Kubernetes. Any Redis it merely connects to (a managed Redis, a docker-compose Redis, a standalone instance) got no limit and no policy, so it could still grow until it ran out of memory.
- Setting `redis_default_ttl = 0` was still a supported foot-gun: it made the cache write keys that never expire, and on a Redis with no eviction that is a genuine unbounded leak.

In short, the cache had no guarantee that its own memory was bounded; it relied on a default TTL plus an eviction policy that only existed in one deployment shape.

## How it was fixed

The fix makes "memory cannot grow without bound" a property of the cache itself, regardless of TTL settings or who provisioned Redis.

1. **Self-enforced memory bound on connect.** The first time the cache reaches Redis, it reads the server's `maxmemory` and eviction policy. If the server is unbounded, it applies the configured `redis_max_memory`, eviction policy, and `maxmemory-samples` itself. If the server refuses (a managed instance that already sets these), it logs one clear warning instead of silently assuming the operator handled it. This closes the gap for non-Kubernetes deployments. The check reruns at most once every five minutes, so a Redis restart that drops runtime config is re-bounded without probing on every request.

2. **No unbounded-write state.** Writes now go through a single TTL resolver. A `redis_default_ttl` of `0` is honored only when the server is confirmed to evict all keys; otherwise it falls back to the default TTL. The cache can no longer write a key that neither expires nor can be evicted. The duplicated, dead "default 0" fallbacks were removed so there is one source of truth.

3. **One source of truth for eviction config.** The eviction policy allowlist and the samples clamp used to live only in the Kubernetes provisioning path, so the new self-configuration path validated nothing: an invalid policy would throw, get swallowed, and silently leave Redis unbounded, and `maxmemory-samples` was ignored entirely. Both checks were extracted into shared `normalize_redis_eviction_policy` and `normalize_redis_eviction_samples` helpers that both paths now call, and the self-configuration path applies `maxmemory-samples` too. Identical config now produces identical Redis tuning whether the Redis was provisioned by Jac-Scale or merely connected to.

4. **The availability check can no longer raise.** The whole memory-bound check is wrapped in one guard that degrades to "treat as unbounded" on any Redis error. It runs on the hot path of `is_available()`, so a transient blip during the post-write config readback must not escape as an exception; it now fails safe (writes fall back to a bounded TTL) and stays throttled so a failing server is not re-probed per request.

5. **New config flag.** `redis_enforce_eviction` (default `true`) lets operators opt out of the self-configuration step.

6. **Test rewritten to assert the guarantee, not the mechanism.** The stale quarantined leak test (which asserted "every key must have a TTL") was replaced with one that asserts bounded memory: anchor keys carry a TTL under default config, and against a bare unconfigured Redis the cache self-enforces a `maxmemory` limit with an eviction policy. The file was moved out of `tests/quarantine/` back into `tests/data/`.

## Verification

- `jac check` was run on every modified source file. No new errors were introduced; the error and warning counts match `main` for these files, which already carry pre-existing type-checker warnings (the eviction-config refactor actually removed one warning).
- The new test file parses and type-checks clean.
- The integration test spins up real Redis and MongoDB containers plus a live server; it should run as part of the normal test suite.


akindu-k:
Closing this as obsolete: the L2 Redis tier this PR hardens no longer exists on `main`.

The Postgres clean-break tail (#8000, merged as #8002) retired the Mongo/Redis/Firestore/SQLAlchemy scaffolding. Every file carrying the actual fix here has since been deleted:

- `jac/jaclang/scale/memory/memory_hierarchy.jac` — the +117-line self-enforced memory bound
- `jac/jaclang/scale/memory/impl/memory_hierarchy.redis.impl.jac`
- `jac/jaclang/scale/deploy/database/kubernetes_redis.jac`
- `jac/jaclang/scale/tests/quarantine/test_redis_memory_leak.jac` — the stale test this PR replaced

`jac/jaclang/scale/config/config_loader.jac` survives, but no `redis_*` keys remain in it, so `redis_default_ttl`, `redis_max_memory`, and the proposed `redis_enforce_eviction` flag have nothing to configure. `deploy/database/` now holds only `kubernetes_postgres.jac`, and the remaining `redis` mentions under `scale/` are negative assertions that the shim is absent.

All six numbered items in the description apply to code that is gone, and the branch is conflicting. Same disposition as #7696, the sibling Redis-eviction fix from #7615, which was closed unmerged for this reason.

The underlying report, #6864, is also obsolete and should be closed as not planned.

PR #6908, opened 2026-06-23
fix(jac-scale): reconcile KEDA TriggerAuthentications by ownership label (part 2 - final - of issue #6535)

### LINKED ISSUE #6535: feat(jac-scale): per-service TriggerAuthentication isolation for KEDA autoscaler (#6534 follow-up)

### LINKED ISSUE #6909: feat(jac-scale): label TriggerAuthentications with owning ScaledObject to enable orphan cleanup in destroy()
# feat(jac-scale): label TriggerAuthentications with owning ScaledObject to enable orphan cleanup in destroy()

## Summary

`KEDAAutoscaler.destroy()` discovers which TriggerAuthentications to delete by reading the live ScaledObject from the cluster. If `apply()` created one or more TriggerAuthentications but crashed before writing the ScaledObject, `destroy()` finds no ScaledObject, skips TA cleanup entirely, and leaves the TAs as permanent orphans in the namespace. For a monolith using `destroy()`, there is no automated recovery — the operator must manually run `kubectl delete triggerauthentication` to clean up.

### Why TriggerAuthentications are created before the ScaledObject

KEDA validates `authenticationRef` at ScaledObject admission time. If the referenced TriggerAuthentication does not already exist when the ScaledObject is created, KEDA rejects the ScaledObject with a 422. This is why `apply()` always creates TriggerAuthentications first, then creates the ScaledObject.

### How the current destroy() works

`destroy()` (Part 2 of #6535) reads `spec.triggers[*].authenticationRef.name` from the live ScaledObject to find TA names, deletes each TA, then deletes the ScaledObject. This works correctly for the normal case where both the ScaledObject and its TAs exist on the cluster.

### The gap

If anything between the last TA write and the ScaledObject create fails — network error, cluster admission rejection, process kill, timeout — the TAs are left on the cluster with no ScaledObject to reference them. When the operator later calls `destroy()`, the ScaledObject GET returns 404, the TA discovery step is skipped, and the TAs remain as orphans indefinitely. They hold stale references to secrets that may be rotated or deleted independently, and accumulate silently across retried deploys.

## Root Cause

`destroy()` has no way to discover TriggerAuthentications when the ScaledObject is absent, because TA names are hash-derived from `(app_name, trigger.name/type, trigger_index)` and `destroy()` only receives `autoscaler_name` and `namespace` — not the trigger list needed to recompute them.

### Step 1 — Label TriggerAuthentications with their owning ScaledObject at creation time

In `_apply_trigger_auth`, add a `scaledobject` label alongside the existing `managed=jac-scale` label:

```jac
"labels": {
    "managed": "jac-scale",
    "scaledobject": autoscaler_name
}
```

This label makes every TA permanently discoverable by the name of the ScaledObject that owns it, independent of whether the ScaledObject itself exists on the cluster.

This requires passing `autoscaler_name` into `_apply_trigger_auth`. The value is already available in `apply()`:

```jac
autoscaler_name = spec.autoscaler_name or f"{spec.scale_target_name}-scaledobject";
```

The microservices equivalent `destroy_collection()` already uses label-based discovery as its primary cleanup mechanism — this approach is inspired by that pattern, adapted for single-service teardown in `destroy()`.

### Step 2 — Add a label-based fallback path in destroy()

When the ScaledObject GET returns 404, instead of skipping cleanup entirely, list all TriggerAuthentications in the namespace matching `managed=jac-scale,scaledobject={autoscaler_name}` and delete each one (404-safe):

```
1. GET ScaledObject by autoscaler_name + namespace
   a. If found:
      - Extract authenticationRef.name from each trigger
      - DELETE each referenced TriggerAuthentication (404-safe)
      - DELETE the ScaledObject (404-safe)
   b. If 404/422:
      - LIST TriggerAuthentications by label selector:
        "managed=jac-scale,scaledobject={autoscaler_name}"
      - DELETE each result (404-safe)
      - Skip ScaledObject delete — it is already gone
```

This closes the partial-apply gap without changing the normal destroy flow and without requiring the ScaledObject to exist.

## Files to Change

| File | Change |
|------|--------|
| `jac_scale/abstractions/autoscalers/keda_autoscaler.jac` | Update `_apply_trigger_auth` declaration to accept `autoscaler_name: str` |
| `jac_scale/abstractions/autoscalers/impl/keda_autoscaler.impl.jac` | Add `scaledobject` label in `_apply_trigger_auth`; update `apply()` to pass `autoscaler_name`; add label-based fallback in `destroy()` |
| `jac_scale/tests/deploy/test_keda_autoscaler.jac` | Add tests for the label-based fallback path |

## Tests to Add

**Test 1**: `destroy() falls back to label-based TA cleanup when ScaledObject is not found`
- Mock GET ScaledObject to raise `ApiException(status=404)`
- Mock list TriggerAuthentications to return two TA items
- Call `destroy()`
- Assert list was called with correct label selector
- Assert both TAs were deleted
- Assert ScaledObject delete was never called

**Test 2**: `destroy() label fallback is a no-op when no orphaned TAs exist`
- Mock GET ScaledObject to raise `ApiException(status=404)`
- Mock list TriggerAuthentications to return empty items
- Call `destroy()`
- Assert delete was never called

**Test 3**: `apply() stamps scaledobject label on created TriggerAuthentication`
- Run `apply()` with one authenticated trigger
- Assert the TriggerAuthentication body passed to `create_namespaced_custom_object` contains `labels["scaledobject"] == autoscaler_name`

## Acceptance Criteria

- Every TriggerAuthentication created by `apply()` carries a `scaledobject` label matching its owning ScaledObject name
- `destroy()` with a missing ScaledObject lists TAs by `managed=jac-scale,scaledobject={autoscaler_name}` and deletes all matches
- `destroy()` with a present ScaledObject continues to use the existing authenticationRef-based path — no regression
- All three new tests pass alongside the existing destroy test suite

## Related

- Part 1: `fix/keda-trigger-auth-per-service-naming` — scoped TriggerAuthentication naming (#6801)
- Part 2: `fix/keda-destroy-trigger-auth-cleanup` — normal destroy gap (#6908)
- This issue closes the remaining edge case left by Part 2

Developer-Linus:
## Problem

`KEDAAutoscaler` orphaned `TriggerAuthentication` resources in two distinct ways:

1. **Redeploy with a changed trigger set.** TA names are derived from the trigger identity (`app + trigger-key`, hashed). When a redeploy renames, removes, or reorders triggers, `apply()` creates new TAs under new names and re-points the ScaledObject at them, but never deletes the TAs from the previous config. `apply()` had no prune step, so every such redeploy left a stale TA behind, unreferenced by the ScaledObject.

2. **destroy().** `destroy()` deleted only the ScaledObject, leaving every TA its `apply()` created behind.

The earlier approach tried to fix (2) by reading `authenticationRef.name` off the live ScaledObject. That could not fix (1) at all (the stale TA is no longer referenced by the ScaledObject, so it is never found), and it left a hole in (2): if the ScaledObject was already gone, the GET returned 404 and `destroy()` returned early, stranding any TAs permanently.

## Root cause

There was no ownership-based reconciliation. TAs are children whose lifecycle should be bound to the autoscaler, but the code bound them only by name derivation plus an imperative "read the ScaledObject, delete what it currently points at". Nothing ever reconciled the actual set of managed TAs in the namespace against the desired set. Compounding this, TAs were stamped only with `managed=jac-scale` and no owner scoping, so they could not even be listed per-autoscaler.

## Fix

Reconcile TAs by an ownership label instead of by parsing the ScaledObject.

- `_build_trigger_auth_manifest()` stamps every managed TA with `jac-scale/owner=<ScaledObject name>` (plus `app` and `managed`). The owner value is the ScaledObject name, hashed if it exceeds the 63-char label limit, via a shared `_owner_label_value()` helper so the apply and destroy selectors always agree. Because the labels live in the manifest builder, `--show-yaml` renders exactly what `apply()` writes.
- `apply()` collects the desired TA names, then `_prune_orphan_trigger_auths()` lists this owner's managed TAs by label and deletes any not in the freshly-applied set. This closes orphan pathway (1).
- `destroy()` lists TAs by `jac-scale/owner=<name>,managed=jac-scale` and deletes them (404/422-safe), then deletes the ScaledObject. It no longer reads the ScaledObject, so cleanup works even when the ScaledObject was already deleted out of band. This closes orphan pathway (2) with no early-return hole.
- `destroy()` (base `Autoscaler`, `HPAAutoscaler`, `KEDAAutoscaler`) now takes `app_name` instead of a raw resource name and resolves the concrete name internally via `resource_name_for()`, matching how the owner label selector is computed. A single `_scaled_object_name(spec, app_name)` helper is the one definition of the ScaledObject name, used by `apply()` and `_build_manifests()`; `destroy()`'s formula is exactly its fallback branch, so the two cannot disagree.

```
apply(spec):
  build manifests (validation runs inside the builder)
  for each manifest in _APPLY_ORDER: create/patch (TAs labelled owner=<SO name>, then the ScaledObject)
  prune: list TA by owner label -> delete any name not in desired set

destroy(app_name, namespace):
  scaled_object_name = resource_name_for(app_name)
  list TA by owner label=scaled_object_name -> delete each (404-safe)
  delete ScaledObject named scaled_object_name (404-safe)
```

## Migration for TAs that predate the owner label

There is deliberately no legacy name-pattern sweep (an earlier revision had one; review showed its truncated-prefix match could cross-delete another app's TAs in a shared namespace, so it was removed rather than narrowed). Instead:

- A pre-label TA whose trigger still exists is adopted implicitly: `apply()` patches it by name and the merge patch adds the owner label. From then on it is fully reconciled.
- A pre-label TA whose trigger was already removed cannot be named from `spec.triggers`, so nothing touches it; the namespace-wide `destroy_collection()` sweep (`managed=jac-scale`) remains the catch-all for those.

## Tests

`test_keda_autoscaler.jac`, all mocked, 42 tests including:

- TriggerAuthentication manifest carries app and owner labels for reconciliation
- apply prunes a TriggerAuthentication the redeployed trigger set no longer references, and keeps one still referenced
- destroy resolves the ScaledObject name from app_name and deletes it
- destroy sweeps TriggerAuthentications owned by this ScaledObject before deleting it (works with the ScaledObject already gone, the case the earlier patch could not handle)
- destroy tolerates an already-deleted TA (404) and an absent auth CRD (404 on list)
- duplicate trigger identities are rejected by the manifest builder, so `--show-yaml` refuses what `apply()` refuses

`test_show_yaml_autoscalers.jac` covers the rendered TA manifests carrying the same labels `apply()` writes. `test_factories.jac` green.

## Live cluster validation

Verified against a real cluster with real KEDA CRDs, driving this branch's `KEDAAutoscaler.apply()`/`destroy()` directly:

| Step | Action | Result |
| --- | --- | --- |
| apply | trigger `cache` + auth | TA created with `jac-scale/owner`, `app`, `managed` labels |
| apply prune | rename trigger `cache` to `queue` | old TA pruned, only the new TA remains |
| destroy | destroy with ScaledObject present | 0 TAs, 0 ScaledObjects |
| destroy (SO gone) | delete ScaledObject out of band, then destroy | TA still swept by label |
| migration | pre-label TA, trigger still live | adopted by apply's patch, swept by destroy |

Both orphan pathways are provably closed on a real cluster.

## Notes

Rebased on the merged #8424: trigger validation lives in `_build_manifests`, so the render and apply paths accept and refuse identical specs, and the manifest list is the single desired-state description that both paths share. This PR supersedes the earlier ScaledObject-parsing approach and resolves the partial-apply / already-gone-ScaledObject limitation tracked in #6909 directly: cleanup is now independent of the ScaledObject.

Closes #6535.
Closes #6909.


kugesan1105 (commented):
Validated against main. The approach is right: label-based reconciliation supersedes both the ScaledObject-parsing path and #6909, and it additionally covers out-of-band `kubectl delete scaledobject`, which neither earlier approach did. Prune ordering is correct too, it runs after the ScaledObject patch so a TA is never deleted while still referenced.

Two things worth addressing:

**1. Owner derivation is duplicated, and the fallback branch disagrees with `destroy()`.**

`apply()` computes `owner = spec.autoscaler_name or f"{spec.scale_target_name}-scaledobject"`, duplicating the expression `_build_manifests` already uses for `scaled_obj_name`. But the only `destroy()` caller passes `resource_name_for(app_name)` = `{app_name}-scaledobject`. Those two agree only while `spec.autoscaler_name` is explicitly set.

It is set today in the sole apply driver (`target.jac:867`), so this is latent rather than live. The trap is that the microservices spec has `scale_target_name = f"{k8s_name}-deployment"` and `app_name = k8s_name`, so if `autoscaler_name` were ever dropped the owner label becomes `{k8s_name}-deployment-scaledobject` while `destroy()` sweeps `{k8s_name}-scaledobject`, and every TA leaks silently with no error. Worth noting that the new test "TriggerAuthentication is stamped with app and owner labels" asserts exactly that divergent shape (`svc-deployment-scaledobject`), which no real `destroy()` call could ever match.

Suggest extracting one `_scaled_object_name(spec)` helper used by both `apply()` and `_build_manifests`, so the owner value has a single definition.

**2. No migration path for TriggerAuthentications created before this PR.**

TAs already on live clusters carry only `managed=jac-scale`. After upgrade, neither `_prune_orphan_trigger_auths` nor `destroy()` will see them, since both select on `jac-scale/owner`. Pre-existing orphans stay orphaned indefinitely. `destroy_collection` still sweeps them namespace-wide so the microservices teardown path recovers, but the `destroy()` path does not. A release note line calling this out would be enough.

Developer-Linus:
Thanks for the review. Both are fixed in b31dacadc.

1. Extracted `_scaled_object_name(spec, app_name)` and made it the single definition, used by both `apply()` and `_build_manifests()`. `destroy()`'s formula (`resource_name_for(app_name)`) is exactly that helper's fallback branch, so the two can no longer disagree. Also updated the test you flagged: it now asserts `svc-scaledobject`, the shape `destroy()` actually reaches, instead of the divergent `svc-deployment-scaledobject`.

2. Added `_legacy_managed_trigger_auths()` to find TAs that carry only `managed=jac-scale` and no owner label, matched by their existing per-app-and-hash name shape. `apply()` now adopts them in place via `_adopt_legacy_trigger_auths()` (relabels with the owner), and `destroy()` sweeps and deletes them directly. Covered by new tests for both paths, plus a prefix-collision guard, and noted in the release note.

MusabMahmoodh (changes requested):
Sorry this sat so long without a review from me - 59 days on a scale PR is my miss, not yours. The core of it is right: ownership labels plus a prune is the correct shape, and doing the cleanup in `apply()` rather than parsing the live ScaledObject is what makes it work when the object was already deleted out of band.

Three things, the first is a stop.

**1. The legacy sweep can delete another app's TriggerAuthentications.** `_legacy_managed_trigger_auths` lists namespace-wide on `managed=jac-scale` and then matches on `re.escape(app_name[:20].rstrip('-') + '-') + r"[0-9a-f]{8}-ta$"`. That prefix is truncated at 20 characters, and the 8 hex characters are matched as a wildcard, so any two apps in one namespace that agree on their first 20 characters match each other's auths. Concrete: `jac-shop-production-api` and `jac-shop-production-web` both truncate to `jac-shop-production-`, so deploying the api app adopts the web app's legacy auths under the api owner label, and the api app's next `apply()` prunes them because they are not in its `desired_auth_names`. The web ScaledObject then references a TriggerAuthentication that no longer exists. `destroy()` has the same reach and deletes them outright. Shared-namespace multi-app deploys are the normal case for jac-scale, so this is reachable.

The fix that keeps the migration honest is to confirm the name rather than pattern-match it: for the app being deployed you can compute the exact legacy names from `spec.triggers` via `_trigger_auth_name`, and adopt or delete only names in that set. Anything you cannot name, leave alone.

**2. This conflicts with #8424, which I approved an hour ago, and the two want merging in the other order.** #8424 deletes `_apply_trigger_auth` entirely: the auth body moves into a pure `_build_trigger_auth_manifest`, and `apply()` writes it through a generic `_apply_custom_object`. Your PR instead adds an `owner` parameter and a return value to `_apply_trigger_auth`. Same function, incompatible directions, and both are yours.

Rebasing this on top of #8424 is also the better result, not just the cheaper one: put the `app` and `jac-scale/owner` labels inside `_build_trigger_auth_manifest`, and then `--show-yaml` renders the auth objects carrying the same labels `apply()` stamps. Without that, #8424 makes the preview show TriggerAuthentications and this PR makes the applied ones differ from the preview by two labels, which is the defect #8409 was about in the first place.

**3. Small one: why adopt legacy auths instead of deleting them here?** `_adopt_legacy_trigger_auths` runs after the prune, so an unlabeled auth that no current trigger wants gets relabelled on this deploy and deleted on the next one. Anything still unlabeled at that point is already known to be orphaned (a live trigger's auth was patched by `_apply_trigger_auth` and so has the label). Was the two-deploy path deliberate?

Nothing else to add on the tests - 245 lines and they cover the prune, the adopt and the destroy sweep. Happy to re-review as soon as it is rebased.

Developer-Linus:
Thanks for the review. Rebased onto #8424 and addressed all three points in 3d53d7a08.

**1. The legacy sweep could delete another app's TriggerAuthentications.** Fixed by removing the legacy adopt and sweep path entirely, not by tightening the regex. With #8424's split, `_build_trigger_auth_manifest` now stamps the `app` and `jac-scale/owner` labels on every TriggerAuthentication it builds. That means a TA still referenced by a current trigger gets its ownership label the moment `apply()` patches it, no separate discovery step needed. A TA whose trigger was already removed can't be named from `spec.triggers` at all, so per your own rule, anything you cannot name, leave alone, nothing touches it. It's left for a full `destroy_collection()` sweep instead. This closes the cross-app collision risk at the source rather than narrowing the match.

**2. Conflicts with #8424, wants merging in the other order.** Rebased onto #8424's tip. The `app` and `jac-scale/owner` labels now live in `_build_trigger_auth_manifest`, so `--show-yaml` and `apply()` render and write identical objects, which was the specific gap you flagged.

**3. Why adopt instead of delete, was the two-deploy path deliberate.** Moot now. There's no separate adopt step left (see point 1), so the two-deploy latency you flagged can't happen anymore.

Also changed to match: `destroy()` (base `Autoscaler`, `HPAAutoscaler`, and `KEDAAutoscaler`) now takes `app_name` instead of a raw resource name, and resolves the concrete name internally via `resource_name_for()`, consistent with how KEDA's `destroy()` needs `app_name` to compute the owner label selector.

Tests: updated the `destroy()` tests for the new signature and the owner-label sweep, added coverage for label stamping on the built manifest, pruning an orphaned TA after a redeploy drops its trigger, and keeping one still referenced. Full `test_keda_autoscaler.jac` (40 tests), `test_show_yaml_autoscalers.jac`, and `test_factories.jac` pass, and formatting is clean.

kugesan1105 (commented):
Re-validated the current head end to end, on a live cluster with real KEDA CRDs. Summary in plain terms.

## What this PR actually fixes, simply

The old code created TriggerAuthentications but could not reliably clean them up, because it discovered them by asking the live ScaledObject "which auths do you use?". An orphaned auth is exactly the one the ScaledObject no longer mentions, and if the ScaledObject itself is gone there is nothing to ask. So leftovers were unfindable by construction.

This PR makes ownership explicit instead: every TriggerAuthentication is stamped with a `jac-scale/owner=<ScaledObject name>` label at build time, `apply()` prunes anything carrying the label that the current trigger set no longer wants, and `destroy()` sweeps by the label without reading the ScaledObject at all. This is the standard Kubernetes reconciliation pattern, and the "nicer" built-in alternative (ownerReferences with GC) is genuinely unavailable here: KEDA admission requires the auths to exist before the ScaledObject, so there is no owner UID to reference at creation time. Labels are the correct tool, not a workaround, and the fix lands at the root cause rather than patching symptoms.

## Live validation (microk8s + KEDA, driving this branch's code directly)

- apply: TA created with `app`, `jac-scale/owner`, `managed` labels
- rename trigger `cache` to `queue`: old TA pruned, only the new one remains
- `kubectl delete scaledobject` out of band, then `destroy()`: TA still swept by label, nothing left
- pre-label TA whose trigger still exists: adopted via the merge patch on the next `apply()`, then swept by `destroy()`

`test_keda_autoscaler.jac` 42/42, `test_show_yaml_autoscalers.jac` and `test_factories.jac` green on the current head.

## Worth follow-up issues (not blockers)

@Developer-Linus these three would make good follow-up issues if you want to pick them up:

- keda to hpa engine switch: `HPAAutoscaler.apply` deletes the competing ScaledObject but not its owner-labeled TriggerAuthentications, so that orphan pathway remains. Trivial to close now that the label exists.
- Pre-upgrade TAs whose trigger was already removed stay invisible to prune and destroy (deliberate; only `destroy_collection` catches them). Worth a line in the docs or release note.
- SDK users who set a custom `spec.autoscaler_name` diverge from `destroy(app_name)`'s owner derivation. All internal call sites agree, but the contract deserves a docstring note.

## Score

**8.5 / 10.** Correct root-cause architecture, real cluster validation, thorough tests, and it closes #6535 gap 3 and #6909 in one coherent design.

PR #6929, opened 2026-06-24
perf(planner): cache filter introspection by lambda __code__

Gorgeous-Patrick:
## Summary

Profile of LittleX5 `load_feed` (10k items) shows `_filter_type_name` + `_filter_has_predicates` consuming ~43% of `osp_spawn` total CPU — about 2.6s out of 6.1s per walker run, across ~20,400 calls each. Both helpers are deterministic per filter source-location, so the work should cache trivially. But Jac emits filter lambdas inline at the path-construction call site, so each walker spawn constructs **fresh function objects**. Default `lru_cache` hashes by identity, so an identity-keyed cache never hits.

This PR caches by `filt.__code__` instead. Lambdas defined at the same source location share their code object across every function-instance Python creates, so the cache hits unconditionally.

- **`_filter_has_predicates`**: body needs only the bytecode, so `_filter_has_predicates_by_code` takes a code object directly. `dis.get_instructions(code)` works against it.
- **`_filter_type_name`**: body needs `inspect.getclosurevars(filt)` (closure cells, not just bytecode). Wrapped in a `_CodeKey` carrier whose `__hash__` and `__eq__` collapse to `__code__`; the lambda is still reachable through the wrapper for the closure walk.

Both helpers (and `_CodeKey`) moved from `planner.impl.jac` into `planner.jac` because module-level `@lru_cache` on a `def` in the `.impl.jac` annex isn't honored. Decorators only bind reliably when the body lives in the interface file (same pattern as `pascal_to_snake` in `helpers.jac`).

## Verification

Local test with **fresh-per-call lambdas** constructed via a helper function, simulating what walker abilities actually emit:

```python
def fresh():
    return (
        lambda i: isinstance(i, TypeA),
        lambda n: isinstance(n, TypeA) and n.val > 5,
    )
for _ in range(100):
    a, b = fresh()
    _filter_type_name(a); _filter_type_name(b)
    _filter_has_predicates(a); _filter_has_predicates(b)
```

Result:
```
type_name_by_code:  CacheInfo(hits=198, misses=2, maxsize=None, currsize=2)
predicates_by_code: CacheInfo(hits=198, misses=2, maxsize=None, currsize=2)
```

200 calls × 2 unique sources → 198 cache hits, 2 misses. With the old identity-keyed approach, every call missed.

## Test plan

- [x] Local verification of cache behavior (see above)
- [x] Behavior correctness: pure `isinstance` → no predicate; predicate lambda → predicate detected; `_filter_type_name` extracts captured Archetype class from closure
- [ ] CI runs the existing planner/test_topology_e2e suite — they exercise both helpers transitively
- [ ] LittleX5 benchmark rerun (to confirm the projected −48% on `osp_spawn`)

## Expected benchmark impact

Based on LittleX5 `load_feed` (limit=10000) profile data:

| Metric | Before | After (projected) |
|---|---|---|
| `getclosurevars` calls | 10,205 | ~10 |
| `dis.get_instructions` calls | 20,410 | ~10 |
| Filter analysis cumulative | ~2,650ms | ~20ms |
| `osp_spawn` total | ~5,560ms | ~2,900ms |


Gorgeous-Patrick:
## LittleX5 cache effect (no prefetcher, no TTG — main-relevant config)

Main has no prefetcher; the closest matching benchmark config is `prefetch_limit=0`, which disables TTG + prefetch entirely (`ttg_ms=0`, `prefetch_ms=0`). All walker time goes through the planner.

### Wall time, `prefetch_limit=0` (3 trials each, averaged)

|  | e2e_ms | walker_ms |
|---|---|---|
| Without cache (main) | 9440 | 8203 |
| With cache (this PR, all 3 commits) | **8082** | **6736** |
| Δ | **−1358ms (−14.4%)** | **−1467ms (−17.9%)** |

A ~1.4s walker-path win on every `load_feed` spawn, with no other variables changed. (The reviewer-prompted closure-type-in-key + drop-filt fixes add ~80ms vs the original key-only cache; well within trial noise and worth the correctness/retention guarantees.)

### cProfile breakdown of the saved work

| Function | Without cache | With cache |
|---|---|---|
| `_filter_has_predicates` cumulative | ~850ms | **7.6ms** |
| `inspect.getclosurevars` calls | 10,205 | **6** |
| `dis.get_instructions` calls | 20,410 | **12** |

The wrapper still records ~20,408 calls to `_filter_has_predicates` (one per filter introspection request from `_build_plan_from_path`), but they collapse to 6 actual closure walks and 12 bytecode disassemblies — one per unique source-location lambda in the `load_feed` walker AST. The work that used to happen ~10,000 times per walker run now happens ~10 times total.

### Raw rows for reproducibility

```
# Without cache (main)
load_feed,0,1,9475.972,0.000,0.000,0.000,8237.675
load_feed,0,2,9467.535,0.000,0.000,0.000,8227.485
load_feed,0,3,9376.004,0.000,0.000,0.000,8144.988

# With cache (this PR, all commits)
load_feed,0,1,8663.592,0.000,0.000,0.000,7141.505
load_feed,0,2,7786.215,0.000,0.000,0.000,6526.493
load_feed,0,3,7797.297,0.000,0.000,0.000,6540.576
```

Gorgeous-Patrick:
Good catch on the silent-miscache risk. Pushed `3cdd8f178` to address it.

`_CodeKey` now also captures the `type`-valued contents of `filt.__closure__` and folds them into both `__hash__` and `__eq__`. Verified locally with the exact shared-helper pattern you described:

```python
def mk(arch): return lambda x: isinstance(x, arch)
f_a = mk(TypeA)  # same __code__ as f_b
f_b = mk(TypeB)
assert f_a.__code__ is f_b.__code__         # True
assert _filter_type_name(f_a) == 'TypeA'    # was incorrectly 'TypeA' for f_b too
assert _filter_type_name(f_b) == 'TypeB'    # now correct
```

Closure inspection cost is a constant-time touch of `filt.__closure__` plus `isinstance(c.cell_contents, type)` per cell — orders of magnitude cheaper than the `inspect.getclosurevars` walk we're caching, so the per-call overhead delta is microseconds. Normal Jac codegen (single-source-location lambdas) still hits the cache fully: 99 hits / 1 miss across 100 calls of `lambda x: isinstance(x, TypeA)`.

`_filter_has_predicates_by_code` is unchanged — its result is fully determined by bytecode, so the code-only key is safe.

On unbounded retention: the cache holds one filt per unique `(code, closure_types)` tuple, bounded by the number of unique filter-source-locations × Archetype-classes the walker exercises. For LittleX5's `load_feed` that's ~10 entries. If reviewers prefer a bounded `maxsize`, easy to swap `None → 1024` or similar — say the word.

Gorgeous-Patrick:
Fixed in `027b3db2d`.

Inside `_filter_type_name_by_code`, the lambda is dropped as soon as the closurevars walk finishes:

```jac
@lru_cache(maxsize=None)
def _filter_type_name_by_code(key: _CodeKey) -> (str | None) {
    result: (str | None) = None;
    try {
        closure = inspect.getclosurevars(key.filt);
        ... // walk for Archetype subclass
    } except Exception { ; }
    key.filt = None;   // release the strong reference
    return result;
}
```

This is safe because `_CodeKey.__hash__` and `__eq__` are defined on `(code, closure_types)` only — `filt` doesn't participate, so mutating it to `None` doesn't break cache lookup or invalidate the existing entry. The mutated key remains stored in the cache; future hits compute `hash(new_key)` and run `__eq__` against it, neither of which touches `filt`.

Verified locally:
- Same-source repeats still cache-hit: `CacheInfo(hits=1, misses=1)` after two calls with same-source lambdas from a helper function.
- The shared-helper miscache fix from the previous commit still holds: `mk(TypeA)` → `'TypeA'`, `mk(TypeB)` → `'TypeB'` with same `__code__`.

Per-entry retention is now bounded to `(code object, tuple of captured type objects)` — neither of which holds the closure cells. A long-running process that exercises N distinct filter source-sites with M captured-class combinations now retains O(N×M) `_CodeKey` instances rather than O(N×M) live lambdas + their closures.

(`_filter_has_predicates_by_code` is unaffected — it caches on `code` directly, no `_CodeKey` involved.)

Gorgeous-Patrick:
Please re-evaluate @greptile-apps

jac/jaclang/runtimelib/planner.jac
```diff
@@ -29,17 +29,173 @@ fallbacks or perf cliffs in CI.
...
+    def init(self: _CodeKey, filt: any) {
+        self.filt = filt;
+        self.code = filt.__code__;
+        cells = filt.__closure__ if filt.__closure__ is not None else ();
+        self.closure_types = tuple(
+            c.cell_contents
+            for c in cells
+            if isinstance(c.cell_contents, `type)
+        );
+    }
```
marsninja: Generall you should follow the separateion of decl and impl convention,

also i think this cna be an obj and you can use postinit if needed,

the model should be abel to knock this out, if anything tricky let me know.

marsninja (changes requested):

jac/jaclang/runtimelib/planner.jac
Gorgeous-Patrick: Done in `bdf6f78ef`.

1. **decl/impl split**: `_CodeKey.postinit`, `_CodeKey.__hash__`, `_CodeKey.__eq__`, `_filter_type_name`, and `_filter_has_predicates` all moved to `planner.impl.jac` with declarations in `planner.jac`.

2. **`class` → `obj` + `postinit`**: `_CodeKey` is now `obj _CodeKey { has filt: any, code: any by postinit, closure_types: tuple by postinit; ... }`. The constructor is `_CodeKey(filt=f)`; postinit fills in the derived fields.

One thing I left as an exception, called out in a comment in the file: `_filter_type_name_by_code` and `_filter_has_predicates_by_code` (the actual `@lru_cache`-decorated helpers) stay inline. We hit this earlier in the PR — the decorator silently doesn't bind when the body is in an impl annex (`cache_info` is missing from the function object). Same behavior whether the decorator is on the interface `def` or on the `impl`. The only working pattern is decorator + body together in the same `.jac` file. Let me know if you have a different approach for getting decorators to bind through the impl split — happy to follow up.

marsninja (approved):

PR #6937, opened 2026-06-24
jac-scale re-architecture 

MusabMahmoodh:
## jac-scale re-architecture (phase 1): a modular ports-and-adapters foundation

This is the first phase of reshaping jac-scale from one large bundle into a clean,
plug-and-unplug platform. It lays the architectural foundation as small, focused,
fully-tested building blocks. **It changes no existing behavior** (everything here
is additive). One deliberate behavior change ships here: microservice routes
are served at a bare `/<service>` instead of `/api/<service>` (see the route-scheme
note below), so callers of `/api/<svc>/...` must drop the prefix.

### Why

The goal is one unified flow where the same app runs as a local monolith or a
distributed k8s fleet with no code changes, where every capability (gateway, auth,
storage, ...) can be plugged or unplugged, and where the default path needs zero
external dependencies (no Docker, no config). To get there we first need the
seams: stable ports with interchangeable adapters, chosen in one place from two
dials.

### The two dials

- **environment**: is `--scale` set? -> local or kubernetes
- **topology**: how many services? (from the `sv import` graph) -> monolith or microservice

A single `RealizationPlanner` reads those and decides which adapter fills each
port. Nothing else branches on topology or environment.

### What is in this PR

| Area | What landed |
|---|---|
| State | `StateBackend` port + in-process default (one seam for rate-limit, scheduler locks, token store; Redis adapter selected by `--scale` later) |
| Code injection | `Injector` port + `PvcInjector` default: pack source, content-address it, ship it on a shared PVC, unpack in an init container. No Docker, no registry, no 1 MiB ConfigMap limit |
| Transport | `ServiceTransport` port + in-process collapse: a cross-service call becomes a direct call in a monolith, a network hop when distributed. The boundary is always declared, so expanding changes only the wire |
| Middleware | `Middleware` port + registry: add a cross-cutting concern in one method and one `register()` call; composes into one onion pipeline; same registry feeds both gateways |
| Gateway | `Gateway` port + a pure request-dispatch classifier (builtin -> admin -> service proxy -> passthrough -> static) |
| Composition root | `RealizationPlanner` that encodes the four-cell assembly table (local/k8s x monolith/microservice) as code |
| Leader-lock (deferred) | `run_once` building block for cross-replica scheduler dedup. NOT yet wired into the scheduler tick - a multi-replica `--scale` scheduler still double-fires; tracked as deferred |
| Route policy | One source for the service route prefix (shared by backend routing, the frontend serviceRoutes map, and OpenAPI), with a reserved-name guard |
| Route scheme | Dropped the `/api/` prefix: services are served at a bare `/<service>`, and the gateway dispatch matches only member subpaths so a same-named SPA route still falls through to static. Breaking for callers of `/api/<svc>/...`. Also removed the `jac setup microservice` ceremony (routes auto-derived from `sv import`) |

### On MinIO

An earlier draft used MinIO for the bundle store. MinIO's server is AGPL-3.0 and
its repo is now unmaintained, so the default is a plain Kubernetes PVC instead:
no extra server, no third-party server license, fewer moving parts. An
S3-compatible store can return later behind the same `Injector` port using a
permissive, maintained option if we ever want it.

### What is intentionally not here (the next, cluster-gated phase)

The deploy-path integration genuinely needs a real cluster to verify, so it is
not in this PR:

- wiring the planner to live adapters (`EmbeddedGateway` / `ProcessGateway`)
- unifying the monolith deploy onto the no-image path (monolith = N=1)
- deleting the old code-shipping paths (ConfigMap and the Docker image pipeline)

These rewrite a working deploy path, so landing them blind would risk breaking
real deploys. They are fully specified and will land as one MicroK8s-verified
pass.

### Safety

- Additive ports. The one behavior change is the `/api/` route-prefix drop
  (breaking for callers of `/api/<svc>/...`); no other existing deploy or serve
  behavior changes.
- The package imports and the existing suites stay green (discovery, plan,
  gateway, drain, rate-limit, ...).

### Testing

Tests focus on complex behavior and integration rather than trivial units. New
coverage includes a runtime integration test that runs the middleware chain, the
dispatch classifier, and the in-process transport end to end, plus the planner's
full four-cell table, the leader-lock `run_once` unit (not wired into the
scheduler), onion ordering and short-circuit,
content-addressed packing with secret exclusion, and PVC pod wiring. All green.


jac-scale/jac_scale/runtime/discovery/route_policy.jac
```diff
@@ -0,0 +1,73 @@
...
+def guard_route_name(module_name: str) -> None {
+    if is_reserved(module_name) {
+        slug = service_slug(module_name);
+        raise ValueError(
+            f"service '{module_name}' derives the route '/{slug}', which collides "
+            + f"with a builtin gateway segment. Rename the service module so its "
+            + f"route does not shadow a builtin path."
+        );
+    }
+}
```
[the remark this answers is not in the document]
MusabMahmoodh: Fixed in 10ad955e2: discover_sv_services now calls guard_route_name per service, so a slug that shadows a builtin segment fails at discovery time.

jac-scale/jac_scale/injector/bundle.jac
```diff
@@ -0,0 +1,186 @@
...
+                continue;
+            }
+        } elif stripped == "plugins.scale" or stripped.startswith("plugins.scale.") {
+            continue;
+        }
+        if not skipping {
+            out.append(line);
+        }
+    }
+    return "\n".join(out) + "\n";
```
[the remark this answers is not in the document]
MusabMahmoodh: Fixed in 10ad955e2: the section name is now taken with split("]")[0], so an inline-comment header (`[plugins.scale] # creds`) no longer slips the section past the credential check. Regression test added (inline comment + top-level dotted key + normal section).

jac-scale/jac_scale/injector/bundle.jac
```diff
@@ -0,0 +1,186 @@
...
+}
```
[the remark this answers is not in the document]
MusabMahmoodh: Fixed in 10ad955e2: the dotted-key branch now fires only at top level (under another section, plugins.scale.x is that section's subkey, not a scale credential), and the unreachable == "plugins.scale" case is removed.

MusabMahmoodh:
All Greptile review items addressed:

- **P0 (security)** `bundle.jac` - inline-comment TOML headers (`[plugins.scale] # creds`) no longer bypass credential stripping; section name is taken via `split("]")[0]`. Regression test added (inline comment + top-level dotted key + normal section all stripped).
- **P2** `route_policy`/`discovery` - `discover_sv_services` now calls `guard_route_name`, so a slug that shadows a builtin segment fails at discovery time.
- **P2** `bundle.jac` - the dotted-key branch now fires only at top level and the unreachable `== "plugins.scale"` case is removed.

Also swept the whole branch for engineering conventions: zero em-dashes (U+2014) in any changed file, no AI attribution in any commit. Ready for re-review.

jac-scale/jac_scale/runtime/discovery/route_policy.jac
```diff
@@ -0,0 +1,71 @@
...
+    if name.endswith("_app") {
+        name = name[:-4];
+    }
+    return name;
+}
+
+
+"""The gateway route prefix for a service module: `/<slug>`, no `/api/`."""
+def derive_route_prefix(module_name: str) -> str {
+    return f"/{service_slug(module_name)}";
```
udithishanka, line 32: this drops the /api prefix right now: derive_route_prefix returns /orders not /api/orders, and the discovery test was updated to assert the bare slug. but the PR body's cleanup row says the change is behavior-neutral and routes stay /api/<name>, and the deferred list says dropping /api is held for the gateway phase. the release note already states the opposite (served at a bare /<service>, no /api/ prefix). so as written this is a breaking change to the public route scheme that the body denies, anything calling /api/<svc>/function/... stops resolving. either keep the /api prefix here until the matching gateway change lands, or update the body to declare the break instead of calling it behavior-neutral. side note, there's still a stale comment in manifest_builder.jac (around L713) saying the gateway owns /api/{svc}/* dispatch, which no longer matches.

udithishanka (commented):
Four notes. The first two are really about the description not matching the code (the /api drop, and the leader-lock not being wired), the other two are smaller. greptile's P0 and P2s already look addressed in 10ad955e2 so I'm not re-flagging those.

jac-scale/jac_scale/runtime/lifecycle/leader_lock.jac
```diff
@@ -0,0 +1,40 @@
...
+
+
+"""Run `task` only if this replica wins the lock `lock_key`.
+
+Returns True when the task ran here, False when a peer already holds the lock
+(this fire is handled elsewhere). On a task exception the lock is left held so the
+failed fire is not retried by a peer within the same tick window; the caller's
+retry policy, not duplicate execution, owns recovery.
+"""
+def run_once(backend: StateBackend, lock_key: str, ttl_seconds: int, task: any) -> bool {
```
udithishanka: run_once is only called from its own unit test. the scheduler path doesn't use it, scheduler.impl.jac and serve.scheduler.impl.jac only got import-path changes in this PR, so nothing actually gates a fire on this lock. that means the once-per-replica double fire the description lists as a landed parity fix is still happening. the test passing against the in-process backend makes it read as covered when the integration isn't there. either wire this into the scheduler tick, or move it to the deferred list the way the gateway and transport are honestly flagged.

jac-scale/jac_scale/state/redis_state_backend.jac
```diff
@@ -0,0 +1,83 @@
...
+        return str(raw);
+    }
+
+    def get(key: str) -> (str | None) {
+        return self._decode(self.client.get(key));
+    }
+
+    def set(key: str, value: str, ttl_seconds: (int | None) = None) -> None {
+        if ttl_seconds is not None {
+            self.client.set(key, value, ex=ttl_seconds);
```
udithishanka: two parity gaps with InProcessStateBackend, worth a note since the whole point of the port is interchangeable adapters. only the in-process backend has tests, this redis adapter has none. and they disagree on ttl_seconds=0: in-process treats it as immediate expiry (the state-backend test relies on that), but redis set with ex=0 here, and acquire_lock's px=0, raise an invalid-expire-time error instead. a parity test running the same assertions over both adapters would surface this and pin the ttl=0 semantics.

jac-scale/jac_scale/injector/pvc_injector.jac
```diff
@@ -0,0 +1,172 @@
...
+    def _seed(app_name: str, namespace: str, raw: bytes) {
+        logger.info(
+            f"pvc-injector: publishing {len(raw)} bytes for app {app_name} " + f"at {bundle_pvc_path(
+                self._object_key
+            )}"
+        );
+    }
+
+    def mutate_pod_spec(pod_spec: dict[str, any], base_image: str) -> dict[str, any] {
+        app_name = "app";
```
udithishanka: when _jac_meta is absent this silently falls back to app_name 'app', so the claim name resolves to app-bundles while prepare() and extra_manifests() used the real app name, and the pod ends up mounting a PVC that was never created. the deploy path isn't wired yet so it won't bite today, but failing loudly when the metadata is missing would be safer than defaulting to a name that can't mount.

marsninja:
@greptileai recheck

jac-scale/jac_scale/runtime/discovery/route_policy.jac
MusabMahmoodh: You're right, the body was stale. Dropping `/api` is intentional here (the gateway member-subpath dispatch landed with it, and the release note already documents the bare `/<service>` scheme), so this isn't deferred anymore. Reconciled the PR body: the cleanup row is now a 'Route scheme' row flagging it as a breaking change, removed it from the deferred/MinIO list, and corrected the 'behavior-neutral' framing to call out the prefix drop.

jac-scale/jac_scale/runtime/lifecycle/leader_lock.jac
MusabMahmoodh: Agreed, `run_once` isn't wired into the scheduler tick, so the cross-replica dedup isn't actually landed (a correct gate also needs a stable per-fire token every replica derives identically, plus a StateBackend threaded into the scheduler). Took the honest-defer path: added a STATUS note to the leader_lock docstring (building block, NOT wired, multi-replica --scale still double-fires) and changed the PR body 'Parity fix' row to 'Leader-lock (deferred)'. e5285e480

jac-scale/jac_scale/state/redis_state_backend.jac
MusabMahmoodh: Good catch on the ttl=0 divergence. set() now drops the key for ttl<=0 (immediate expiry, matching InProcessStateBackend) and acquire_lock clamps px to >=1ms so it no longer hits redis's invalid-expire error. Added a parity test in test_state_backend.jac that runs the ttl=0 assertions over both adapters with a fake redis that rejects ex/px<=0, pinning the semantics (and giving the redis adapter its first test). e5285e480

jac-scale/jac_scale/injector/pvc_injector.jac
MusabMahmoodh: Done. mutate_pod_spec now raises when _jac_meta/app_name is missing instead of defaulting to 'app' (which would mount a claim name that was never created). The manifest builder always sets _jac_meta before delegating, so the real path and the test are unaffected. e5285e480

udithishanka:
Tested this branch end to end by deploying a jacBuilder preview against it (jac-scale built from this branch on both the deploy runner and the runtime image). Install + manifest generation work, but the app pods never schedule:

```
0/15 nodes are available: persistentvolumeclaim "<app>-bundles" not found
```

**Root cause:** `pvc_injector.mutate_pod_spec` unconditionally mounts the `jac-bundles` volume (claim `<app_name>-bundles`) onto every pod, but nothing creates that PVC in this branch. jac-scale provisions the `data` and `mongo` PVCs, but not `<app>-bundles`, so all app pods sit `Pending` and the fleet-wait times out at 1200s.

This looks like a known deferral per the `_seed` docstring:

> a short-lived loader pod mounts the PVC and the deploy host kubectl-cp's the bundle in. Kept out of this commit so the pure manifest and bundle logic stays unit-testable without a cluster; covered by the k8s e2e.

So the pod-spec mutation landed but the PVC creation/seeding is deferred, which makes any real (non-e2e) deploy unschedulable. Two ways to keep deploys working before the seeding lands:
- create the `<app>-bundles` PVC in `prepare()` / `extra_manifests()` even as an empty claim, or
- gate the volume injection in `mutate_pod_spec` so it only mounts the bundle volume when bundles are actually published.

Everything else (install, deploy logic, manifest gen) worked against this branch. Happy to re-run the preview once the PVC provisioning is in.

udithishanka:
Thanks for the quick turnaround. Re-tested against `ac52af2`: the `<app>-bundles` PVC is created now, so that part is sorted. It surfaces one more layer though, the deploy still can't schedule:

```
PVC <app>-bundles   Pending   storageClass=gp3   accessModes=[ReadWriteMany]
Warning  ProvisioningFailed (x9)  ebs.csi.aws.com   # EBS/gp3 is RWO-only, can't satisfy RWX
```

The bundle PVC is `ReadWriteMany` but lands on the cluster's **default** storage class, which on EKS is `gp3` (EBS, RWO-only). EBS CSI returns `ProvisioningFailed`, the PVC stays `Pending`, the deploy waits on it ("once the PVC from extra_manifests() exists"), and times out before any app pod is created.

`build_bundle_pvc`'s own docstring already calls this out:
> storage_class empty means the cluster default; callers set it to an RWX class (e.g. efs-sc) for multi-node.

...but `manifest_builder.jac` instantiates `PvcInjector()` with no `storage_class`, so it falls through to that default. Plumbing an RWX class into the injector (the same way the data PVC gets `efs-sc` from the shared-volume config) should close it. There's no consumer-side fix, since the injector is created with no args so an app's jac.toml storage_class can't reach the bundle PVC.

Everything up to the PVC binding works against this branch.

udithishanka:
Tested this PR end to end: deployed a fullstack microservices app (gateway + 3 services + client) with `jac start --scale --experimental` onto EKS, with jac-scale/byllm installed from this branch and `[plugins.scale.kubernetes].jaseci_repo_url`/`jaseci_branch` pointed back at this fork so the in-pod build uses the rearchitecture code. Hit three issues. The orchestration and the in-pod source build both work; the fleet does not converge. Details and workarounds below.

**1. Bundle PVC sets no storageClassName, so the RWX volume can't bind on EBS-default clusters**

The accessMode fix landed (the `<app>-bundles` PVC is now ReadWriteMany), but it sets no storageClassName, so it falls to the cluster default. On EKS (and any cluster whose default is EBS/gp3) EBS cannot provision ReadWriteMany, so the PVC stays Pending and the whole fleet hangs.

There is no knob for it: `PvcInjector` is constructed with no args at all three call sites (`kubernetes_target.jac:1099`, `manifest_builder.jac:80` and `:938`), and `storage_class` defaults to empty.

Workaround (set the injector default to the cluster's RWX class):
```
# jac_scale/injector/pvc_injector.jac
storage_class: str = "efs-sc",
```
Bound immediately: `<app>-bundles  Bound  1Gi  RWX  efs-sc`.

Suggestion: surface the bundle storage class through `[plugins.scale.kubernetes]` (e.g. `bundle_storage_class`), empty by default but settable to a RWX class, so it stays cloud-agnostic rather than hardcoding efs-sc.

**2. The binary's PYTHONHOME/PYTHONPATH leak into the EKS exec credential, causing a 401**

Deploying to EKS fails at database provisioning:
```
ERROR - exec: process returned 1. Fatal Python error: Failed to import encodings module
ModuleNotFoundError: No module named 'encodings'
[ERROR] Deployment failed: database provisioning failed: (401) Reason: Unauthorized
```
The jac binary exports PYTHONHOME/PYTHONPATH for its bundled CPython, and those leak into subprocesses. jac-scale's kubernetes client runs the kubeconfig exec credential (`aws eks get-token`); `aws` is itself a Python program, inherits the binary's PYTHONHOME/PYTHONPATH, cannot find its own stdlib, and dies. No token is produced, so the API server returns 401. `kubectl` from a normal shell works because it is not running under the jac binary, which masks this.

This will hit any exec-credential kubeconfig (EKS `aws`, GKE `gcloud`, AKS `kubelogin`), since they are all host Python tools sensitive to the inherited env.

Workaround (PATH-shimmed `aws` that strips the two vars):
```
#!/bin/bash
exec env -u PYTHONHOME -u PYTHONPATH /opt/homebrew/bin/aws "$@"
```
Suggestion: strip PYTHONHOME/PYTHONPATH from the subprocess env whenever jac-scale shells out (the k8s exec-credential path and the kubectl seed).

**3. The experimental in-pod build can't fit the startup-probe budget, and the build dir is ephemeral, so the fleet crash-loops**

The app container's startup probe is `/healthz/live`, period 10s, failureThreshold 90, i.e. a 900s (15 min) budget for the app to start serving. In experimental mode the app container first runs the entire build (apt build-essential, fetch-llvm slice, fetch CPython, fetch typeshed, `zig build` jac, jac install) before it serves anything, and that takes longer than 900s. At ~15 min the kubelet kills it:
```
lastState.terminated: exitCode 137, reason Error
Warning Unhealthy  Startup probe failed: ... connect: connection refused   (x537)
Normal  Killing    Container <svc> failed startup probe, will be restarted  (x6)
```
Worse, the build dir `/app` is an emptyDir, so each restart rebuilds from scratch. Net result: 6-9 restarts per pod, 90+ min, never Ready.

Suggestion: run the source build in an init container (so it is not racing the app container's startup probe), and/or build onto a persistent volume so a restart resumes instead of rebuilding from zero. As-is, on any cluster/arch where the build exceeds ~15 min the experimental fleet cannot converge.

**Outcome:** with workarounds 1 and 2 in place the deploy provisions cleanly (namespace, mongodb + redis, bundle PVC seeded, 4 deployments + HPAs + PDBs), and the experimental build runs correctly against this fork (clone + LLVM slice + CPython + typeshed + zig build). Issue 3 is what stops the fleet from coming up. Host: macOS arm64 driving the deploy with the `dev` jac binary; target EKS (gp3 default, efs-sc available).

udithishanka:
Re-tested on the latest branch (`d32df368`) with the same fullstack app, deploying to EKS from a macOS arm64 host. The earlier fixes landed well:

- The bundle PVC now takes a `bundle_storage_class` knob and plumbs it into the PVC (`build_bundle_pvc`), and the bundle-loader pod got a hardened securityContext. That resolves the PVC-can't-bind issue and the PodSecurity warning from the earlier run.
- `--experimental` no longer does the per-pod in-pod build; it host-builds the jac binary once and ships it via `BinaryInjector`. Right direction for the OOM/slow-build problem.

One new blocker on the host-build path: it does not cross-compile, so it only works when the host OS/arch matches the pods.

`build_release_binary` (`jac_scale/injector/binary.jac:29`) runs `zig build --prefix .ship-out` with no `-Dtarget`, so it builds for the host platform. `pack_toolchain` then tars that binary and ships it to the pods. From a macOS arm64 host deploying to linux/amd64 pods, that ships a Mach-O/arm64 binary into linux-x86_64 containers, which cannot exec.

The dry-run stops at:
```
[ERROR] --experimental could not ship a host-built jac binary: [Errno 2] No such file or directory: 'zig'.
Put zig 0.16.0 on PATH and run 'cd jac && zig build --prefix .ship-out' ...
```
and even with zig installed, the produced binary would be host-native (Mach-O/arm64 here), not the linux-x86_64 ELF the pods need.

Suggestion: pass `-Dtarget` for the pod platform (resolve from the node arch, default `x86_64-linux`) so the host build cross-compiles to what the pods run; `build.zig` already supports `-Dtarget`. Failing that, detect host-platform != pod-platform up front and fail with a clear message (or fall back to the default release-binary path) instead of building and shipping an incompatible binary. As-is, `--experimental` is usable only from a linux/amd64 host that matches the pods.

I held the deploy rather than run default mode, so this is from the dry-run plus a read of the injector path on `d32df368`.

udithishanka:
A microservice-mode footgun from deploying `this_is_jac`, which embeds **littleX as a sub-app** (same backend, same graph).

littleX's backend is `littlex/social_graph.jac` (the `setup_profile` / `Profile` / `Tweet` walkers). The parent app pulls it in with `import from .littlex.frontend { app as LittleX }` and the embedded UI calls its walkers via `sv import from .social_graph { setup_profile, ... }`. So `social_graph` is a server-walker service the embedded sub-app depends on.

The parent's jac.toml declares explicit routes for its **own** three services:
```
[plugins.scale.microservices.routes]
guestbook = "/api/guestbook"
source_viewer = "/api/source_viewer"
analytics = "/api/analytics"
```
Deploy that, and the embedded littleX is broken: signing up fires `POST /walker/setup_profile` and the gateway returns `404 {"code":"NOT_FOUND","message":"Not found on any service: /walker/setup_profile"}`. `social_graph` was never deployed.

Root cause: auto-discovery is all-or-nothing. In `plugin.jac`, `discover_sv_services` only runs when no routes are declared:
```
if not ms_config.get("routes", {}) {
    ms_config["routes"] = discover_sv_services(filename);
}
```
So the moment you declare a route for even one service, discovery is fully disabled, and every other `sv import` service (including every embedded sub-app's) must be hand-listed too, or it is silently dropped. There is no deploy-time warning; the only signal is a runtime 404 buried inside the embedded app.

That is the odd part: embedding a sub-app just works in the monolith (one server serves all walkers), but the microservice split silently requires the parent app to know and re-declare all of the sub-app's internal services. Someone embedding littleX has no reason to expect they must add `social_graph` to their routes, and nothing tells them.

Workaround (confirmed: dry-run now plans 5 services, `social_graph` carries `/walker/setup_profile`):
```
social_graph = "/api/social_graph"
```

Suggested fix: don't make routes all-or-nothing. Either run `discover_sv_services` always and merge the discovered services with the explicit routes (discovered-but-unrouted services get a default prefix), so embedded sub-apps deploy with no extra config; or at minimum warn at deploy time when an `sv import` service is not covered by a route, and make the 404 say "service `social_graph` is referenced via `sv import` but missing from `[microservices.routes]`" instead of a generic not-found.

udithishanka:
Follow-up: even after hand-adding the route, the sub-app service still can't deploy, because the deploy resolves a service to `<service_name>.jac` at the project root.

littleX's file is `littlex/social_graph.jac` (a subdir), but the service name derived from `sv import from .social_graph` is just `social_graph`. So the pod runs `jac start social_graph.jac` in `/app` and crash-loops:
```
✖ Error: File 'social_graph.jac' not found.
✖ Error:   Current directory: /app
```

So embedding a sub-app whose services live in a subdir is doubly blocked: discovery skips them (previous comment), and even a hand-added route cannot locate the file, because the service-to-file mapping assumes `<name>.jac` at the project root and ignores the sub-package path. A path-qualified route key (`littlex/social_graph`) might sidestep it, but I have not confirmed that the gateway's walker routing still resolves with a slashed service name.

Suggested fix: resolve a service's source file from the `sv import` module path (so `sv import from .social_graph` inside `littlex/` maps to `littlex/social_graph.jac`), rather than `<service_name>.jac` at the root. Combined with the discovery change in the previous comment, that would let an embedded sub-app deploy with no manual route surgery.

udithishanka:
FYI from the same e2e walkthrough of `this_is_jac` (this one is jaclang native-floor / showcase-app, not jac-scale, but it breaks the deployed demo and would break a microservice e2e that ships this app).

The raylib WASM demo fails to instantiate:
```
LinkError: WebAssembly.instantiate(): Import #45 "env" "strtod": function import requires a callable
```
The native-jac to WASM module imports four libc functions from `env` that the app's `raylib_shim.cl.jac` does not stub: `strchr`, `strlen`, `strtod`, `strtol`. (Inspected the built `.wasm`: 48 `env` imports, the shim provided 44.) These are emitted by the native floor (the Phase 1 work now in merged-main / the `dev` binary). The showcase shim predates them, so the WASM will not link.

Not a jac-scale bug, flagging because (a) the deployed app's demo is broken on current jaclang, and (b) if the microservice e2e you are adding deploys `this_is_jac`, the client build ships a WASM the shim cannot satisfy.

I added the four stubs to the shim locally and the demo links again. Broader point: as the native floor grows its libc surface, the hand-written WASM shims (`raylib_shim` etc.) have to track it, or apps silently break at instantiate.

Issue #6940, opened 2026-06-24
Native stdlib for a release-fetcher/installer: HTTP/zip/json/datetime tier (FFI-floor + pure-Jac surface)


marsninja:
## Summary

A concrete native-pathway standard-library roadmap driven by a real **release-fetcher / installer** workload (download release archives over HTTPS with resume, inflate zips, parse a JSON manifest, write a directory tree, stamp UTC timestamps). This is the umbrella issue tracking that work; it **extends** the general native stdlib roadmap (#6404) into the networking/compression tier that #6404 deliberately marked a 🔴 non-goal.

### Required surface (the workload)

| Cluster | APIs | Real requirement |
|---|---|---|
| HTTP/HTTPS | `urllib.request.urlopen`, `Request` (+ Range / Accept-Ranges) | an HTTP/1.1 client + TLS |
| TLS/transport (implied) | `ssl`, `socket` | a TLS stack (release URLs are `https://`) |
| Archive + compression | `zipfile` (ZipFile, namelist, read, ZipInfo, writestr), `zlib` (DEFLATE) | a DEFLATE engine + ZIP container parsing |
| JSON | `json.loads` (manifest), `json.dumps` | recursive `any` dispatch + float repr |
| os / os.path extras | `os.environ`/`getenv`, `os.makedirs`, `os.path.{dirname,normpath,realpath}`, `os.sep` | mostly pure-string + a couple of libc calls |
| Time | `datetime.now`, `timezone.utc` | calendar math on the existing `time` intercept |

## Architecture: FFI-floor + pure-Jac API surface

The headline decision: **never reimplement TLS or DEFLATE — FFI (Mechanism F) to the system library that already does it, exactly as CPython does** (CPython's `ssl` *is* OpenSSL; its `zlib` *is* libz). The security/perf-critical floor is a thin FFI binding; the Python API surface the workload calls (`urllib`/`http`/`zipfile`/`json`) is pure-Jac (Mechanism B) layered on top. Because the workload codes against the stable *upper* API, the floor underneath is swappable without touching the program — which is why "unblock now" and "proper long-term" are the same roadmap.

The native FFI ABI went production-grade recently (System V + AAPCS, struct-by-value, `sret`), so Mechanism F is now a first-class option.

### Mechanism preference order (multi-target reality)

Native now targets ELF / Mach-O / PE / WASM. That inverts #6404's "A-first" default:

[...]

PR #6941, opened 2026-06-24
feat(native): bundled na_stdlib resolution keystone + os.path (#6940)

marsninja:
## What

Phase 0 of the native standard-library roadmap (#6940): the **bundled
`na_stdlib` resolution keystone** — the delivery vehicle for a Python-congruent
standard library on the native (`na`) pathway — plus the first module riding it.

`import from <mod> { fn }` now resolves a bundled pure-Jac `.na.jac` module
shipped under `jaclang/runtimelib/na_stdlib/` and links it through the
**existing** cross-module machinery (binding population → extern forward-decl →
`link_in`), with **zero per-module backend work**. The same source binds
CPython's real module on the sv pathway and the bundled module on na.

## How

The change is two matching edits to the single shared resolver and its
link-side counterpart, so both `BoundaryAnalysisPass`/`NaIRGenPass` and
`NativeCompilePass` agree:

- **`codeinfo.resolve_native_module`** — new `native_stdlib_root()` and a
  **nearest-wins** fallback to it after the importing project's own tree
  misses (so a user module of the same name always shadows a bundled one).
  The bundled `source_module` string uses the `os.sep` relpath form
  (`os/path.na`) so the linker can re-root it.
- **`na_compile_pass`** — the AOT/JIT link-set discovery previously resolved
  each dep only relative to the importer's tree, so a bundled dep got a
  `W5023` and was never compiled+linked → `undefined symbol
  __jac_glob_init_<mod>` at runtime. Added the matching bundled-root fallback.

## First bundled module: `os.path` (POSIX)

`normpath`, `dirname`, `basename`, `split`, `splitext`, `isabs` — pure-Jac,
**byte-for-byte congruent with CPython `posixpath`** including the tricky
`normpath` (collapsed slashes, `..` resolution, `//` vs `///`) and `splitext`
(leading-dot basenames) edge cases. This fills the documented v1 gap
(`normpath`/`split`/`splitext` were never in the `os.path` A-dispatch).

## Tests

`prim_ospath.jac` — tri-backend equivalence fixture (`require=["na"]`),
registered in `test_prim_equivalence.jac`. Verified on **both** execution
models: `jac nacompile` (AOT standalone binary) and the JIT harness, matching
CPython across 22 path edge cases. Existing cross-module native tests and the
`os` stdlib suite still pass (the shared-resolver blast radius).

## Follow-ups (tracked in #6940)

- Float shortest-round-trip repr (Ryū) — gates `json.dumps` float parity.
- Bare-import `module.attr` congruence (`import os; os.path.normpath()`).
- The `any`-consumption native gaps a pure-Jac `json` surfaces (segfault on
  some `any`→typed-container patterns; `len()` on an untyped-list element).

Closes part of #6940.


marsninja:
Thanks @greptile-apps - addressed all three in 99d29a3:

1. **Repeated `native_stdlib_root()` probe** (P2): agreed. Hoisted it above the work-queue loop so the bundled root + its `os.path.isdir()` are computed once per compile instead of per local-miss dep.
2. **POSIX-only on non-POSIX targets** (P2): agreed, and it's a real latent congruence break (CPython uses `ntpath` on Windows). Building `ntpath`+dispatch is out of scope for this keystone demonstrator, so I've documented the POSIX-only scope explicitly in the module docstring and tracked the Windows `ntpath` follow-up under #6940 rather than ship it silently.
3. **`split` not directly exercised** (P2): agreed. The fixture now calls `split` directly and covers the all-slash-head (`dirname("/")`) and empty-head (`split("file.txt")`) branches that `dirname`/`basename` only reached indirectly.

PR #6949, opened 2026-06-25
feat(native): na_stdlib urllib.request over libcurl floor + variadic FFI (#6940 Phase 3)

marsninja:
## Summary

Implements **Phase 3 (transport + HTTP — the long pole)** of the native-stdlib roadmap in #6940. Ships a bundled `na_stdlib/urllib/request.na.jac` whose `urlopen(url)` performs a real HTTP/1.1 GET (DNS, TLS, redirects) over **libcurl** reached through the native FFI (**Mechanism F**) — the transport floor is not reimplemented, exactly as CPython's `ssl`/`http.client` lean on the system stack. libcurl abstracts the per-OS TLS backend (SecureTransport / OpenSSL / Schannel), so one binding covers all native hosts.

The easy handle streams the response body and headers to two temp `FILE*`s via libcurl's default `fwrite` callback (`CURLOPT_WRITEDATA` / `CURLOPT_HEADERDATA`); the module reads them back with the native `open().read()` builtin and parses status/headers into an `HTTPResponse` record. This keeps the FFI surface to a handful of stable C entry points and needs no Jac↔C write callback.

## Enabler: native FFI variadic declarations

`curl_easy_setopt(CURL*, option, ...)` is variadic in C. On AArch64/AAPCS a variadic argument **must** pass on the stack — a fixed (non-variadic) LLVM declaration passes it in a register, so the URL/options never reach libcurl (`CURLE_URL_MALFORMAT`). A trailing `*rest` param on a clib `def` now marks the symbol variadic:

- `_declare_clib_ability` emits a `var_arg` LLVM declaration over only the fixed params; the variadic tail flows through the existing verbatim call-site path.
- `collect_foreign_fns` excludes the `*rest`/`**kw` marker from the fixed ABI signature.
- Non-variadic clib decls are unaffected (`var_arg=False`).

```
def curl_easy_setopt(handle: int, option: int, *rest: int) -> int;
```

## Test

`test_urllib_equivalence.jac` stands up a loopback `http.server` (no live network, per the issue's policy) and pins the bundled native `urlopen` against CPython's real `urllib.request` on **HTTP status** and **received body length**, for a direct GET and for a **301 redirect both follow**. The native functions return plain `int`s driven through the JIT, and the sv reference is pinned with concrete asserts so a regression in either backend is a hard failure. Verified failing without the bundled module and passing with it.

## Partial coverage (per #6940 policy)

v1 = GET + redirects + status/body. Documented next slices: `HTTPResponse` accessor methods (`read()`/`getheader()`) and cross-module `bytes`/header field marshalling (only scalar `int` fields cross a native module boundary today), a `Request` object (custom method / headers / `Range`), POST `data`, `timeout`, and raising `HTTPError` on 4xx/5xx. Native-host-only (clean E5090 on wasm).

Extends #6404; follows #6941 (the bundled `na_stdlib` keystone + `os.path`).


marsninja:
Addressed the review in fcfe286:

- **P1 — `fopen` NULL not checked**: agreed, real bug. `urlopen` now checks both `fopen` results; on NULL it cleans up the easy handle + temp files and raises, instead of letting libcurl's default writer fall back to stdout and `fclose(NULL)` hit UB.
- **P2 — `"variadic"` stored but unused on the ABI-plan path**: agreed (latent). `is_variadic` is now computed once from `sig.varargs` and threaded into the `ir.FunctionType(..., var_arg=...)` on *both* the ABI-struct plan path and the legacy scalar path, so a future variadic C function carrying a by-value struct param can't silently emit a non-variadic declaration. Dropped the now-redundant `"variadic"` model key (the value is read from the signature directly).
- **P2 — `url` is request URL, not effective**: agreed it was misleading. Documented it on the field, in the `urlopen` body, and in the module's "next slices" — recovering the effective URL needs `curl_easy_getinfo(CURLINFO_EFFECTIVE_URL, &char*)` (a variadic out-param read), which is the documented follow-up alongside the `geturl()` accessor.
- **P2 — CRLF assumption**: agreed, added a comment documenting that the header file is read as text and `rstrip("\r")`-normalized, and what that trims.

Also fixed the failing `jac-check` (ran `jac format --lintfix`).

Issue #6956, opened 2026-06-25
perf(jac-scale): server test suite is slow (~7m47s) from per-boot rehash and subprocess recompile tax


akindu-k:
## Summary

The `jac-scale` server test group takes about **7m47s**, and three files account for ~84% of it. The slowness is not the assertions, it is fixed per-test overhead: every test boots a fresh server, and each boot pays costs that are identical across tests and therefore pure waste.

This is an umbrella issue to track bringing that suite down. The first fix is already up as #6955.

## Cost breakdown (server group, ~7m47s)

| File | Time | % | Tests |
| --- | --- | --- | --- |
| test_serve.jac | 191.3s | 41.0% | 31 |
| test_admin.jac | 101.7s | 21.8% | 29 |
| test_webhook.jac | 98.9s | 21.2% | 23 |
| ...remaining files | ~76s | ~16% | — |

For contrast, test_sso.jac runs 51 tests in 4.4s because it exercises components directly without booting a full server. That is the target shape.

### 1. Per-boot password rehash (done — #6955)

Profiling the in-process boot path showed ~1.5s of every ~1.8s boot was bcrypt: each boot hashes two fixed bootstrap credentials (admin + internal guest user) at the production work factor. The in-process suite boots a fresh server per test, so it paid this on every test.

Fix: make the bcrypt work factor configurable (`password_hash_rounds` auth setting / `JAC_SCALE_BCRYPT_ROUNDS` env override, secure default unchanged); tests opt into the minimum factor. Full per-test isolation preserved.

CI result on test_admin.jac (same base commit, with vs without the fix): **64.66s -> 43.40s**, 29/29 passing.

- [x] Configurable bcrypt work factor (#6955)

### 2. Subprocess tests cold-recompile on every spawn

test_serve and test_webhook are already ~half in-process (`make_client`), so #6955 helps those. The remaining subprocess tests are slow because the test harness deletes the entire `.jac/` directory (the on-disk compile cache) before every server spawn, forcing each `jac start` to cold-compile the fixture through full type inference (the "~60s budget / ~25s window" noted in `_start_server`). With ~31 subprocess spawns across the two files, this recompilation is the bulk of their runtime.

- [ ] Scope the pre-spawn cleanup to data/DB artifacts (and client `dist` if needed) instead of wiping the compiled-module cache, so 2nd..Nth spawns are warm. Pre-warm the fixture once so the first spawn is warm too.
[...]

PR #6961, opened 2026-06-25
perf(scale): reuse one MongoDB container across webhook tests (#6956)

akindu-k:
Part of #6956.

## What this does

The webhook test file has 10 "with mongo" tests. Each one started its own `mongo:latest` testcontainer, ran a single assertion, then stopped the container, so the file paid 10 full container start/stop cycles.

This shares a single MongoDB container per worker process: it is started lazily on the first "with mongo" test, reused by the rest, and stopped at process exit. Each test drops the `jac_db` database before running so isolation is identical to a fresh container.

## Why

Container lifecycle, not the assertions, was the cost in those tests. Spinning one up per test is pure repeated overhead since they all need the same thing.

## Result

Measured in isolation (container lifecycle only, the part this PR changes):

- Per test: a fresh container per test is ~1.9s; a shared container plus a `jac_db` drop is ~0.7s.
- About 2.8x faster on that portion. The saving scales with container start cost, which is typically higher on CI than on a local Docker daemon.

All 23 webhook tests still pass. No test cases were removed and isolation is unchanged (each test still starts from an empty `jac_db`).

## Scope

This addresses the MongoDB container cost only. The webhook subprocess `jac start` cold-compiles are a separate cost tracked under #6956.

## Note

Rebased onto the `jac-scale` → core `jaclang` fold (#7038). The test now lives at `jac/jaclang/scale/tests/server/test_webhook.jac`; the change is unaffected by the move.


akindu-k:
image

PR #6982, opened 2026-06-26
fix(py2jac): emit correct operator token in MatchKVPair for valid Jac output

### LINKED ISSUE #6645: py2jac produces invalid Jac for match statements with nested class patterns (case Container(inner=Inner(...)))
## Bug

`jac py2jac` produces invalid Jac source that then fails to re-parse during the format step. Reproduces on latest `main` (`ee186618d`) with a cleared cache.

## Reproduction

```python
class Inner:
    x: int
    y: int

class Container:
    inner: Inner

match Container(inner=Inner(x=1, y=2)):
    case Container(inner=Inner(x=1, y=0)):
        print("1")
    case Container(inner=Inner(x=0, y=2)):
        print("2")
    case _:
        print("none")
```

```bash
rm -rf ~/.jac
jac py2jac repro.py
```

## Expected

Valid Jac output (with the nested class patterns converted to Jac match-arch patterns).

## Actual

All errors report `line 1` with large column numbers, indicating the unparse step produced malformed text that the re-parse step can't handle:

```
✖ Error: repro.py, line 1, col 181: Missing ',

chess10kp:
Fixes #6645

## Problem
py2jac was producing invalid Jac code for match statements involving mapping patterns and class patterns. The unparse pass always emitted a hardcoded colon separator (`:`) for `MatchKVPair`, even when the pattern was a class keyword pattern which requires `=`.

## Root Cause
`MatchKVPair`'s `kid` field only contained the key and value nodes, missing the operator token (`:` for mappings, `=` for class patterns). The unparse pass then couldn't distinguish between the two cases.

## Fix
1. **pyast_load_pass.impl.jac**: When building `MatchKVPair` nodes, include the appropriate operator token in `kid`:
   - Colon token for mapping patterns (`proc_match_mapping`)
   - Equals token for class keyword patterns (`proc_match_class`)
2. **unparse_pass.impl.jac**: Updated `exit_match_k_v_pair` to use `nd.kid[1]` (the operator token) instead of a hardcoded colon.

## Testing
- Verified that py2jac output for match statements now produces valid Jac syntax in both mapping and class pattern contexts.


chess10kp:
**Re: Greptile P1 (loop bound mismatch):** Already fixed in this PR — the refactoring builds `valid_kwd_patterns` and `names` together in a first pass over `range(len(nd.kwd_attrs))`, then a second loop uses `range(len(valid_kwd_patterns))`. The pre-existing `range(len(kwd_patterns))` mismatch is gone.

**Re: Greptile P2 (no automated test):** Added in this PR — `jac/tests/language/test_cli.jac` now includes a test "py to jac match mapping and class patterns" with fixture `jac/tests/language/fixtures/py2jac_match.py` covering both mapping patterns (`:`) and class keyword patterns (`=`). Also fixed the assertion strings to match Jac formatter output (spaces around operators: `"key" : 1`, `x = 0`).

PR #6984, opened 2026-06-26
fix(compiler): support subscripted/generic types in lambda param annotations

### LINKED ISSUE #6773: Lambda parameter annotation rejects subscripted/generic types (`lambda d: list[int] : ...`)

Developer-Linus:
## Problem

Lambda parameter type annotations that used subscripted or generic types (e.g. `list[int]`, `dict[str, int]`, `int | str`) caused a parse failure. The previous heuristic checked only a fixed two-token lookahead and could not recognise a colon followed by a bracket-containing type expression.

The same heuristic introduced regressions in the opposite direction: lambda bodies containing arithmetic (`x * 2`), function calls (`len(s) > 2`), subscripts (`d["active"]`), or unary expressions (`-x`) were sometimes misidentified as type annotations, corrupting the transpiled output.

Additional lambda forms that were also broken:
- **Return-hint form** — `lambda e: any -> None { body }`: the scan never found a second `:`, so the annotation was silently dropped and `any -> None { ... }` was handed to the expression parser, causing a parse error.
- **Brace-body form** — `lambda x: int { body }`: the scan crossed the opening `{`, exited the body's `}` back to depth 0, then hit `;` with no annotation found.
- **Sibling lambdas** — `[lambda x: x * 2, lambda y: y + 3]`: the scan crossed the `,` and found the second lambda's body colon, misidentifying the first lambda's colon as an annotation separator.
- **Dict-value lambdas** — `{"k": lambda x: x * 2, "k2": val}`: after the lambda's body expression set `in_expr`, the scan continued past the dict comma and found the dict key separator `:`, misidentifying it as an annotation separator. Bare-name dict-value lambdas (`{"k": lambda x: x}`) were also affected via the same comma-crossing path.

## Solution

Replaced the speculative two-token lookahead with a **bracket-depth scan** (`_lambda_colon_is_annotation`). When the parser encounters a `COLON` inside a lambda parameter list it now:

1. Saves the current token position.
2. Advances past the colon and scans forward, tracking bracket depth across `()`, `[]`, and `{}`.
3. Returns `True` (annotation separator) when any of the following appear at depth 0:
   - A second `COLON` — standard `lambda x: type : body` form.
   - `RETURN_HINT` (`->`) — `lambda x: type -> ReturnType { body }` form.
   - `LBRACE` (`{`) after at least one type token and no expression operator — `lambda x: type { body }` form.
4. Returns `False` (body colon) when a closing bracket, `SEMI`, or `KW_LAMBDA` is hit first.
5. Restores the saved position unconditionally before returning.

An `in_expr` flag is set on arithmetic and comparison operators (`+`, `-`, `*`, `==`, `<`, etc.) to prevent a dict/set literal body (`lambda x: {k: v}`) from being mistaken for the brace-body annotation form. A `saw_type_token` flag ensures `{` and `:` are only treated as annotation markers when at least one type token was seen first. The `saw_type_token` flag is reset on each depth-0 comma, so the token segment after the comma must independently look like a type — this prevents the dict key separator after `{"k": lambda x: expr,` from firing as an annotation colon. The COLON terminator additionally requires `not in_expr`. A `KW_LAMBDA` terminator at depth 0 stops the scan from crossing into sibling lambda expressions.

## Changes

| File | Description |
|------|-------------|
| `jac/jaclang/jac0core/parser/parser.jac` | Added `_lambda_colon_is_annotation` method declaration |
| `jac/jaclang/jac0core/parser/impl/parser.impl.jac` | Implemented bracket-depth scan; replaced old speculative-parse block |
| `jac/tests/language/fixtures/lambda_arg_annotation.jac` | Added fixtures for subscripted/generic/union, return-hint, brace-body, sibling lambda, and expression-body regression cases |
| `jac/tests/language/test_cli.jac` | Extended transpile test; added runtime test `lambda arg annotation runtime` |
| `docs/.../unreleased/jaclang/6984.bugfix.md` | Release note fragment |

## Test coverage

**Fixtures** (`lambda_arg_annotation.jac`):
- `list[int]`, `dict[str, int]`, `int | str` — subscripted/generic/union annotations parse and execute correctly (labeled output: `g:10`, `h:42`, `k:7`).
- `lambda x: any -> None { body }` — return-hint form parses correctly (`inc:10`).
- Regression cases: `lambda x: x * 2`, `lambda s: len(s) > 2`, `lambda d: d["active"]`, `lambda x: -x` — expression bodies intact in transpiled output and at runtime.
- Sibling lambdas: `[lambda x: x * 2, lambda y: y + 3]` — each element executes independently (`8`, `7`).

**Tests**:
- `lambda arg annotation` — asserts transpiled output preserves all body forms intact.
- `lambda arg annotation runtime` — asserts correct output values end-to-end, including `r6:12` and `r7:10` (dict-value lambdas).

## Related

- Closes #6773
- See also #6782 (original release note PR)
- Release note: #6984


kugesan1105:
https://github.com/jaseci-labs/jac/pull/7376

PR #6991, opened 2026-06-27
ci(installer): verify install.sh against real releases + harden it

marsninja:
## Why

The user-facing install path — `curl -fsSL .../scripts/install.sh | bash` — broke in production while CI stayed green:

```
[jac] Could not determine latest version from GitHub Releases.
[jac] Please specify a version with --version.
```

Root cause was release-side (a binary release landed on a non-canonical `jaclang-<ver>` tag, fixed separately by #6972 + a fresh `v<version>` release), but the **test gaps that let it ship silently** are what this PR closes.

## Why CI didn't catch it

- **`test-installer.yml` only ran on PRs touching `install.sh`**, and its asset "gate" turned a missing/mis-tagged latest release into a **silent green skip** (`::notice::`). The breakage was caused by a release change, not an `install.sh` edit, so the workflow never ran — and even if it had, it would have skipped.
- **`release-jaclang.yml` built and uploaded the binary but never ran the installer**, so tag/installer drift only surfaced in users' terminals.

## Changes

- **`release-jaclang.yml`**: after each platform uploads its asset, run `scripts/install.sh --version <ver>` against the just-published release (isolated `$HOME`), run a program, and uninstall — failing the release if `install.sh` can't fetch/run the binary it just shipped. Runs on the same native hardware the asset targets.
- **`test-installer.yml`**: add a **daily schedule** (installer breakage is usually caused by release changes, not `install.sh` edits, so a path-filtered PR trigger alone never catches it) and convert the skip-gate into a **hard failure** when the latest release lacks this platform's binary.
- **`install.sh`**: fix a `set -u` *"unbound variable"* crash in the EXIT trap — the trap referenced a function-local `tmpdir` that is out of scope when it fires at exit. (Latent because the gate had never let the real install run in CI.)

## Note on ordering

`install.sh` is intentionally kept strict to canonical `v<version>` tags (matching the corrected pipeline in #6972), so these checks only go green once a properly `v`-tagged release is the latest. The companion `v0.30.2` release re-establishes that.

### DIFF
--- a/.github/workflows/release-jaclang.yml
+++ b/.github/workflows/release-jaclang.yml
@@ -116,6 +116,30 @@ jobs:
             "dist/${{ steps.pkg.outputs.asset }}.sha256" \
             --clobber
 
+      # End-to-end check of the user-facing install path. Now that THIS
+      # platform's asset is attached to the release, `scripts/install.sh` must be
+      # able to resolve, download, checksum-verify, and run it. This is the gate
+      # that catches release-tag/installer scheme drift (the kind that silently
+      # breaks `curl ... | bash`) at release time, on the same native hardware
+      # the asset targets, instead of in users' terminals. Gated like the upload
+      # above -- only meaningful once the asset is actually on a release. Runs in
+      # an isolated $HOME so the runner's PATH/profile is left untouched.
+      - name: Verify install.sh can fetch + run this release
+        if: github.event_name == 'release' || inputs.tag != ''
+        env:
+          JAC_VER: ${{ steps.ver.outputs.version }}
+        run: |
+          set -euxo pipefail
+          HOME="$(mktemp -d)"; export HOME
+          bash scripts/install.sh --version "$JAC_VER"
+          BIN="$HOME/.local/bin/jac"
+          test -x "$BIN"
+          "$BIN" --version
+          printf 'with entry { print("installed ok"); }\n' > "$HOME/h.jac"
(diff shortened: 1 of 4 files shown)


marsninja:
Added a second commit: **`build(gitignore): ignore root-level zig build artifacts`**.

Surfaced while cutting `v0.30.2` via the **Create Release PR** workflow — it failed at *Create branch and commit*:

\`\`\`
remote: error: File .zig-cache/o/.../jac is 104.64 MB; this exceeds GitHub's file size limit of 100.00 MB
remote: error: File .zig-cache/o/.../libjacllvm.so is 123.34 MB; this exceeds GitHub's file size limit of 100.00 MB
 ! [remote rejected]     release/jaclang-0.30.2 (pre-receive hook declined)
\`\`\`

`create-release-pr.yml` runs `jac run scripts/release.jac` from the **repo root**; the native backend emits a `.zig-cache/` (with the >100 MB `jac` / `libjacllvm.so` outputs) into that cwd, and the release step's `git add -A` swept them into the commit. `jac/.gitignore` only ignores the build under `jac/`. This breaks releases on any `setup-jac` cache **miss** (0.30.1 was likely cut on a cache hit, which skips `zig build`, so it never surfaced). Root-level ignore patterns fix it.

> Note: this fix must be on `main` before re-dispatching **Create Release PR** for 0.30.2 (the workflow checks out the triggering ref).

Issue #6978, opened 2026-06-26
Native stdlib & self-contained `nacompile` - final plan


marsninja:
# Native stdlib & self-contained `nacompile` - final plan

Status: proposed · Supersedes the phasing in **#6404** and **#6976** · Owner: TBD

This is the consolidated roadmap for the native (`na`) backend standard library
**and** for making `jac nacompile` binaries genuinely self-contained. It merges
two issues that have been tracked separately but are really one decision:

- **#6404** - Python-congruent stdlib on native (sv↔na parity).
- **#6976** - Self-contained native binaries; move FFI floors off host shared
  libs; drop libcurl.

It is written to be the single source of truth. Where it conflicts with the
issue text, this document wins - both issues predate the single-binary
evolution and a large amount of landed work, and their phasing is stale.

---

## 1. Thesis

**The native backend's FFI floor should *be* the set of C libraries the bundled
CPython is built on.** The single-binary distribution already fetches a
python-build-standalone (pbs) tarball and ships it; that tarball carries, as
plain relocatable static archives, the exact C libraries CPython's own C
extensions link against. We link `na` programs against *those* archives, from
our own bundled copies - never against whatever happens to be on the host.

Decision rule for every stdlib module:

| If CPython gets the capability from… | …then `na` uses… | Mechanism |
|---|---|---|
| a reusable C library (libssl, libsqlite3, libmpdec, liblzma…) | the **same pbs-bundled archive**, statically linked | **F** (bundled-C floor) |
| pure Python | pure Jac (`.na.jac`) | **B** |
| bespoke CPython-C-API C (`_json`, `_datetime`, `_struct`, `_random`) | a from-scratch reimplementation (intrinsic or Jac) | **A** / **B** |
| libc/libm/syscalls (universal) | compiler-lowered intrinsics | **A** |

This is what "leverage Python's native libraries fully" means precisely: reuse
the C **libraries**, not the interpreter and not the PyObject-C-API extension
**modules**.

### Verified foundation

The archives are already on disk after a normal build (confirmed on
`macos-aarch64`):

```
.pbs-build/<osarch>/python/build/lib/
    libcrypto.a  libssl.a  libsqlite3.a  libmpdec.a
    liblzma.a    libbz2.a  libexpat.a    libffi.a   libzstd.a
```

Members are ordinary relocatable objects (treeshakeable), **not** LLVM bitcode -
[...]

PR #7002, opened 2026-06-28
feat(native): bundled secrets over a libcrypto CSPRNG floor (#6978 Phase 1)

marsninja:
Part of #6978 (Phase 1). Second F-module on the bundled `libcrypto.a` floor that the hashlib/hmac slice (#6997) stood up.

## What this adds

`import secrets` now resolves on the native (`nacompile`/JIT) pathway to a bundled `na_stdlib` Mechanism-F module: a pure-Jac surface (`secrets.na.jac`) over a thin `RAND_bytes` FFI floor (`_csprng_native.na.jac`) on the same bundled `libcrypto.a` hashlib/hmac use. OpenSSL's `RAND_bytes` is seeded from the OS CSPRNG — the same source CPython's `secrets` draws from via `os.urandom`.

- **Coverage:** `token_bytes`, `token_hex`, `token_urlsafe` (hand-rolled base64-url, `=`-stripped, exactly as CPython renders), and `randbelow` (uniform via rejection sampling over the smallest covering number of CSPRNG bytes, matching CPython's `SystemRandom._randbelow`). A `RAND_bytes` failure raises `ValueError` rather than handing back the zero buffer.
- An `nacompile`d binary statically links the bundled `libcrypto.a` and runs self-contained.

## Validation

New tri-backend equivalence test `prim_secrets` (`test_prim_equivalence.jac::secrets bundled`, `require=["na"]`). Since the output is random, it pins the two backends congruent on the deterministic **properties** rather than exact bytes: token lengths (incl. the empty case), `randbelow(1) == 0`, the in-range bound, and that two draws differ. Verified on both the JIT harness and a self-contained `nacompile` binary.

## Note on `uuid`

`uuid.uuid4` was prototyped on the same CSPRNG floor but **deferred**: the native pathway has no `__str__`/`str(obj)` support yet (`str(obj)` returns garbage), so the canonical `str(uuid4())` form can't be rendered congruently. It's a clean follow-up once native `__str__` lands (which also benefits general object stringification).


marsninja:
`uuid` is deferred from this PR on a concrete native gap — `str(obj)` ignores `__str__` so the canonical `str(uuid4())` form can't render. Tracked in #7003; `uuid` lands once that's fixed.

Issue #7003, opened 2026-06-28
Native: `str(obj)` ignores `__str__` (returns garbage) — blocks `uuid` + object stringification


marsninja:
## Summary

On the native (`na`) pathway, `str(obj)` on a user object does **not** dispatch to the object's `__str__` method — it returns garbage (an unrendered pointer/value) instead of the string. There is no `__str__` lowering, so the canonical Python stringification protocol is unavailable on native.

Discovered while building the bundled native `uuid` module in #7002 (Phase 1 of #6978); it is the blocker that forced `uuid` to be deferred.

## Repro

```jac
obj P {
    has v: str;
    def __str__ -> str { return self.v; }
}

with entry {
    p = P(v="hello");
    print(str(p));   # expected: hello
}
```

`jac nacompile`d (and run): prints garbage (e.g. `�@`), not `hello`. (Same on the JIT path.)

- **Expected:** `str(p)` calls `P.__str__` → `hello`.
- **Actual:** `__str__` is ignored; `str()` of a struct pointer yields garbage.

## Impact

- **`uuid`** (#6978 Phase 1): `uuid.uuid4()`'s primary use is `str(uuid4())` → the `8-4-4-4-12` form. Without `__str__` that can't be rendered congruently, so `uuid` is deferred in #7002.
- **General:** `str(obj)`, `print(obj)`, and `f"{obj}"` for any user object all need `__str__` dispatch — broadly useful well beyond the stdlib (logging, debugging, CLIs).
- **Compounding factor:** a bundled `na_stdlib` module whose name matches a CPython stdlib module type-resolves to **CPython's typeshed stub**, so its public surface must match CPython's API exactly. A workaround method like `str_form()` is rejected by the type checker (`E1030: Type "UUID" has no attribute "str_form"`) because CPython's `uuid.UUID` has no such method — so `__str__` is the *only* congruent way to expose the string form. (This is also why `hashlib`'s `Hash.hexdigest()` worked: the names already matched CPython.)

## Related native gap (lesser; worked around)

`bytes` are immutable on native — subscript assignment is rejected:

```jac
with entry { b = b"\x00\x00"; b[0] = 65; }   # error[E5090]: Native pathway does not yet support subscript assignment 'b'
```

[...]

PR #7006, opened 2026-06-28
feat(native): bundled uuid (v4) on the CSPRNG floor, via native __str__ (#6978 Phase 1)

marsninja:
Part of #6978 (Phase 1). Third F-module on the bundled `libcrypto.a` floor, unblocked by native `str()`/`__str__` dispatch (#7003, landed in #7005).

## What this adds

`import uuid` now resolves on the native (`nacompile`/JIT) pathway to a bundled `na_stdlib` Mechanism-F module (`uuid.na.jac`) over the same `_csprng_native` `RAND_bytes` floor `secrets` uses.

- `uuid4()` draws 16 CSPRNG bytes and renders the 32-char `hex`, applying the v4 **version** (byte 6 high nibble -> `4`) and **RFC-4122 variant** (byte 8 -> `10xx`) bits during the render. Native `bytes` are immutable (the secondary note in #7003), so the bits are applied at render time rather than rewritten into the buffer.
- `UUID.hex`, `str(UUID)` (the canonical `8-4-4-4-12` form, via `__str__`), and `UUID.version` match CPython's API. `.bytes`/`.int`/`.urn` (which need a constructed `bytes`) are follow-ups.
- An `nacompile`d binary statically links `libcrypto.a` and runs self-contained.

## Validation

New tri-backend equivalence test `prim_uuid` (`test_prim_equivalence.jac::uuid bundled`, `require=["na"]`). The output being random, it pins the backends congruent on the deterministic v4 **structure**: `hex`/`str()` lengths, the hyphen layout (positions 8/13/18/23), the version nibble (`4`), the RFC-4122 variant nibble (`8`/`9`/`a`/`b`), `.version == 4`, and that two draws differ. Verified on the JIT harness and a self-contained `nacompile` binary. Exercises native `str(obj)`/`__str__` dispatch end to end.

Closes the `uuid` follow-up noted in #7002.


marsninja:
Thanks @greptile-apps — went through both:

1. **`% 16` in the hex render** — agreed, switched `uuid4`'s render to `v >> 4` / `v & 15` (the `% 16` was a no-op on a byte, and the shift/mask is the unambiguous nibble extraction). Left the identical pattern in the already-merged `_csprng_native.to_hex` / `_hashlib_native.to_hex` untouched — out of scope for this PR and a harmless no-op there.

2. **`uuid_py` / `uuid_na` duplication** — keeping as-is, by design. The two functions run through *different backends*: `uuid_py` executes on the sv backend (CPython's `uuid`) while `uuid_na` is inside the `na {}` block (the bundled native module), and the tri-backend harness compares their outputs. They can't share one helper because the whole point is to run the same logic through each backend independently — that's why every sibling `prim_*` fixture has the same shape.

(Also: deslopped the file per the repo-wide strip-comments/strip-docstrings rules from #7001, and merged latest main.)

Issue #7013, opened 2026-06-28
Native: two foundational gaps blocking #6978 Phase 1 — cross-module exception unwinding aborts; nacompile doesn't link bundled-module IR on Linux


marsninja:
Two foundational gaps in the native (`na`) backend surfaced while building the **#6978 Phase 1** network tier (`socket`/`ssl`/`http`/`urllib`). Both are pre-existing and general (they affect every bundled `na_stdlib` module, not just the new ones), and both block doing the rest of Phase 1 — especially `ssl` — to a production bar. The `socket` module is complete and passes its JIT tri-backend equivalence test, but I'm holding it/the rest until these are fixed, because `ssl`'s fail-closed verification is built directly on #1.

---

## Finding 1 — A native exception cannot cross a `.na.jac` module boundary (aborts)

A `raise` in one `.na.jac` module that is caught (or even left uncaught) by a **different** module SIGABRTs (exit 134) instead of unwinding to the handler. A `try`/`except` **within the same module** works fine. It is not exception-type specific — `OSError`, `ValueError`, and `Exception` all abort across the boundary.

### Minimal repro

`_raiser.na.jac`:
```jac
def:pub do_raise(x: int) -> int {
    if x < 0 { raise OSError("neg"); }
    return x;
}
```

`main.jac`:
```jac
na {
    import from _raiser { do_raise }
    def run -> int {
        try { return do_raise(-3); } except OSError as e { return -999; }
    }
}
with entry { print("caught (expect -999):", run()); }
```

```
$ jac run main.jac
Aborted (core dumped)        # exit 134; "caught" never prints
```

- Same code with the `try`/`except` **inside** `_raiser.na.jac` (a `safe_call` wrapper that catches `do_raise`) returns `-999` correctly.
- An **uncaught** cross-module raise (no `try`/`except`, propagating to `with entry`) also aborts rather than printing a clean traceback.

### Impact
Every bundled `na_stdlib` module that raises is affected latently — `hashlib`/`secrets`/`_csprng_native` all `raise ValueError` on their error paths; they just never fire in the happy-path tests. It directly blocks **`ssl` fail-closed verification** (#6978 Phase 1): a cert/hostname mismatch must surface as a catchable `ssl.SSLCertVerificationError`, but today it can only abort the process or be faked with CPython-divergent sentinel returns (which the issue explicitly rules out).

### Expected
[...]

PR #7024, opened 2026-06-28
feat(native): bundled socket (TCP client) over a libc BSD-sockets floor (#6978 Phase 1)

marsninja:
## Summary

First slice of the **#6978 Phase 1** network tier: a bundled native `socket` module (TCP client). `import socket` now resolves on the native (`nacompile`/JIT) pathway to a bundled `na_stdlib` module — a pure-Jac client surface (`socket.na.jac`) over a thin libc FFI floor (`_socket_native.na.jac`).

Unlike the `hashlib`/`secrets` floors that statically link a bundled `libcrypto.a`, the socket primitives are plain BSD sockets in libc (`socket`/`connect`/`send`/`recv`/`getaddrinfo`), so a socket program links nothing beyond libc/libm and is self-contained with **no bundled archive — and no libcurl dependency.** That last point is the foundation for moving `http`/`urllib` off the libcurl floor later in Phase 1.

DNS resolution walks `getaddrinfo`'s `struct addrinfo` list with **no new compiler intrinsic**: a libc `memcpy` whose source pointer is carried as a plain Jac `int` (the established `FILE*`-as-`int` FFI convention from the libcurl floor) copies the sockaddr out, and `int.from_bytes` reads the fields. The whole transport tier stays a swappable pure-Jac surface with zero backend work.

## API surface

A blocking TCP **client**:

- `socket(family, type, proto)` object — `connect`/`send`/`sendall`/`recv`/`close`/`fileno`
- `create_connection((host, port))` — family-spanning (IPv4/IPv6) helper, the entry point the `ssl` wrapper and `http`/`urllib` build on
- `AF_*` / `SOCK_*` constants

Setup failures and transport errors raise `OSError` (mirroring CPython, whose `socket.error` *is* `OSError`); the raise unwinds cleanly to a caller's `except`, including across a module boundary — relying on the native cross-module exception fix (#7021). The libc floor stays the clean sentinel-returning layer; the surface is where the `OSError`s originate (bar the one `recv` hard-error case the floor raises directly).

## Tests

`test_socket_equivalence` stands up a loopback HTTP server and pins the native client congruent with CPython's `socket` on:
- the HTTP status, and
- the exact total response byte count of a hand-driven `create_connection`/`sendall`/`recv`-drain GET,

and asserts that connecting to a closed port raises a **catchable `OSError`** (not an abort). Green locally on a `-Ddev` Linux build.

## Scope / follow-ups

- Server-side (`bind`/`listen`/`accept`), UDP (`sendto`/`recvfrom`), `settimeout`, and `makefile` are documented follow-ups — the client surface is what the Phase-1 `ssl` + `http`/`urllib` tier needs.
- Linux/glibc `struct addrinfo` offsets make this native-host-Linux-first; macOS/Windows ride the cross-platform follow-up (#6978 Phase 4).
- A Linux AOT self-contained-binary test is deferred until `nacompile` links bundled-module IR on Linux (the remaining open half of #7013); DT_NEEDED is already correct (a socket binary needs only `libc`/`libm`).

Part of #6978 Phase 1. Depends on #7021 (merged).


marsninja:
Superseded by #7052 (merged), which was stacked on this branch and carried the bundled `socket` module into main -- including the later review fix (the `resolve_all` -> `(addresses, rc)` / `gai_str` change). main already has those files with that fix, so merging this PR would revert it and re-add already-merged code. Closing.

Follow-up: the socket release-note fragment (`7024.feature.md`) was dropped from #7052 to keep that PR to its own fragment, so it didn't land -- adding it separately so the socket feature is documented.

Issue #7037, opened 2026-06-29
Native: bundled na_stdlib calls resolve against CPython typeshed signature, not the bundled .na.jac signature — blocks ssl (#6978 Phase 1)


marsninja:
On the native (`na`) pathway, a call to a function/method imported from a bundled `na_stdlib` stdlib module (e.g. `ssl.create_default_context`, `SSLContext.wrap_socket`) is type-checked, keyword-bound, AND default-filled against **CPython's typeshed stub** for that module — not against the bundled `.na.jac` module's own signature (the one that is actually linked and run).

This is **latent** for modules whose API matches CPython trivially and positionally (`hashlib.sha256(data)`, `socket.create_connection(addr)`), which is why hashlib/secrets/socket are unaffected. It **breaks `ssl`** (#6978 Phase 1), whose CPython signature diverges:

- `ssl.create_default_context(purpose=Purpose.SERVER_AUTH, *, cafile=None, capath=None, cadata=None)` — `purpose` first; `cafile` keyword-only; default is the `Purpose.SERVER_AUTH` enum.
- `SSLContext.wrap_socket(sock, server_side=False, do_handshake_on_connect=True, suppress_ragged_eofs=True, server_hostname=None, session=None)` — `server_hostname` is the 5th parameter.

The bundled `na_stdlib/ssl.na.jac` instead defines `create_default_context(cafile: str = "")` and `wrap_socket(sock, server_hostname: str)`.

### Symptoms (all from na using the typeshed signature)

1. **Non-lowerable typeshed defaults.** Omitting `purpose` makes na fill it from typeshed's default `Purpose.SERVER_AUTH` → `error[E5090]: Native pathway does not yet support attribute access 'SERVER_AUTH'` (and `Ellipsis` for other typeshed defaults). These don't lower on na.
2. **Keyword args bind to the typeshed slot.** `ctx.wrap_socket(sock, server_hostname="localhost")` binds `server_hostname` per typeshed's layout (slot 5), mismatching the bundled `(sock, server_hostname)` → segfault (`0x1`). `create_default_context(cafile="x")` lands `cafile` at the wrong slot → NULL → `strlen` segfault (`0x0`).
[...]

PR #7052, opened 2026-06-29
feat(native): ssl (verifying TLS client) + urllib over socket+ssl; remove libcurl floor (#6978 Phase 1)

marsninja:
## Summary

Completes the **#6978 Phase 1** network tier: a verifying-TLS-client `ssl` over OpenSSL, `urllib.urlopen` reimplemented over `socket`+`ssl`, and **the libcurl floor removed**. Builds on the `socket` slice (#7024).

> **Stacked on #7024** (bundled `socket`). The first commit here is that socket module; it drops out on rebase once #7024 merges. Review the **ssl** and **urllib** commits.

### `ssl` — verifying TLS client over `libssl`/`libcrypto`

`import ssl` resolves on the native pathway to a bundled `na_stdlib` module: a pure-Jac surface (`ssl.na.jac`) over an OpenSSL FFI floor (`_ssl_native.na.jac`, reusing the bundled `libcrypto` the `hashlib`/`secrets` floors already link).

**Verification posture is CPython's, never hand-rolled** — `SSL_CTX_set_verify(SSL_VERIFY_PEER)` + a loaded CA bundle (system / `SSL_CERT_FILE`, or an explicit `cafile`) + SNI + `SSL_set1_host` for the in-handshake hostname check. So `SSL_connect` **fails closed** on an untrusted chain OR a hostname mismatch, and the surface raises a catchable `SSLError`.

Surface: `SSLError`, `SSLContext`, `create_default_context`, `SSLContext.wrap_socket(sock, server_hostname=...) -> SSLSocket` (`send`/`sendall`/`recv`/`close`).

### `urllib.urlopen` over `socket`+`ssl`; libcurl removed

`na_stdlib/urllib/request.na.jac` is now pure Jac over the bundled `socket`/`ssl` modules — it parses the URL, opens the connection (TLS-wrapped + cert/hostname-verified for `https`), sends an HTTP/1.0 GET, drains the response, splits headers/body, and follows redirects. The `urlopen(url) -> HTTPResponse` surface is unchanged.

A `nacompile`d program using `urlopen` now links **no foreign C beyond libc/libssl/libcrypto — no libcurl** (the last non-pbs C dependency, and the only user-facing variadic-FFI caller, are gone).

## Tests (all green locally)

- `test_ssl_equivalence` — loopback HTTPS server with a test-generated self-signed cert; asserts in CI that a trusted cert + correct hostname verifies and GETs 200 (congruent with CPython), and that **an untrusted chain and a wrong hostname each fail closed with `SSLError`**.
- `test_urllib_equivalence` — keeps pinning `urlopen` congruent with CPython's `urllib.request` against a loopback server (direct GET + 301 redirect), now over the socket+ssl transport.
- Also smoke-tested against real HTTPS (`example.com` → 200, verified).

## Security review

Fail-closed throughout: `SSL_VERIFY_PEER` is set before the handshake; the CA store must load or `wrap_socket` raises (no insecure fallback); the hostname is pinned; a handshake/verify failure raises `SSLError` and returns no `SSLSocket`. The negative tests prove untrusted-cert and hostname-mismatch rejection. `urllib`'s `https` path uses the verifying default context with `server_hostname`.

## Notes / follow-ups

- A single `SSLError` type is used (not CPython's `SSLCertVerificationError` subclass) pending native base-class `except` matching.
- Linux/glibc-first (struct offsets); macOS/Windows TLS-stack choice is #6978 Phase 4.
- Relies on the native fixes #7021 (cross-module exceptions), #7028 (keyword args), and #7037 (bundled-stdlib signature resolution) that make the bundled `socket`/`ssl` APIs callable congruently.

Part of #6978 Phase 1.


marsninja:
Thanks @greptile — addressed all four in `3dc1c805b` (and dropped the stray `7024.feature.md`, which belongs to the stacked socket PR #7024):

**P1 — TCP socket leaked on TLS handshake failure** ✅
Fixed at the source: `SSLContext.wrap_socket` now closes `sock` on *every* failure path (CTX alloc, CA load, `SSL_new`, handshake), so a rejected cert/hostname no longer leaks the fd. This covers the `urllib` path you flagged (it goes through `wrap_socket`).

**P1/security — TLS read errors silently returned as clean EOF** ✅
`ssl_read` now branches on `SSL_get_error`: `SSL_ERROR_ZERO_RETURN` → EOF; `SSL_ERROR_SSL` with reason `UNEXPECTED_EOF_WHILE_READING` (a server closing without `close_notify`) → EOF for compat; any other `SSL_ERROR_SSL` (e.g. bad record MAC / truncation alert) now **raises**, which `SSLSocket.recv` surfaces as `SSLError` rather than a silent `status=0` truncation. Verified real HTTPS (`example.com` → 200) still works and the loopback fail-closed test stays green.

**P2 — `SSL_get_error` / `_SSL_ERROR_ZERO_RETURN` unused** ✅
Now used by the `ssl_read` change above.

**P2 — `gai_message` re-runs `getaddrinfo`** ✅
Took your suggested approach: `resolve_all` now returns `(list[bytes], rc)`, and the surface uses `gai_str(rc)` for the error message — no second DNS round-trip, and the actual first-call error code is reported. `gai_message` is removed.

All three equivalence tests (socket/ssl/urllib) remain green.

PR #7094, opened 2026-07-01
ci: simplify release publishing to the jaclang-only, idempotent path

kugesan1105:
## What

Simplify the release-publishing pipeline to match reality: with byLLM, scale, and mcp folded into the `jac` binary, there are **no PyPI packages left to publish**. A release is now exactly two things: tag `v<version>` and cut the GitHub Release (which fires `release-jaclang.yml` to build and attach the native binaries).

## Changes

- **`publish-release.yml`** (663 → 158 lines): drop the dead PyPI tier1-4 / wait-for-pypi / npm / per-package build machinery. New flow: `resolve` (version from `jac.toml`) → `approve` (pypi gate) → `tag-and-release` → `summary`. Tag and release creation are **idempotent** (`git tag -f` + create-or-edit), so a re-run after a partial failure converges instead of erroring.
- **`release-jaclang.yml`**: wait for this platform's asset to be visible in the GitHub API before the `install.sh` verify step, avoiding a release-propagation race.
- Remove the now-unused `scripts/parse_release.jac` and `scripts/wait_for_pypi.jac`.

## Why

The last release surfaced two failure modes this fixes:
- `push-tags` crashed recreating an existing tag (non-idempotent) → now `tag -f`.
- The release was cut but never published because a later step failed → now create-or-edit converges, and the manual `workflow_dispatch` recovery path re-runs the same idempotent flow from `jac.toml`.
- The macOS `install.sh` verify raced asset propagation → now waits for the asset first.

`release-jaclang.yml` stays a separate workflow on purpose: it builds on native runners per platform, and its `release: [published]` + `workflow_dispatch` triggers double as the rolling-dev build path and a one-click binary re-attach.

## Release flow (unchanged for the operator)

1. Create Release PR (bumps `jac.toml`).
2. Merge it → publish-release fires → approve the `pypi` gate → tag + release cut.
3. Automatic: `release: published` → release-jaclang builds + attaches binaries.


kugesan1105:
Thanks @greptile-apps. Triaged the four findings; two are stale against the current diff, one is real and now fixed.

- **Published Event Is Suppressed** / **Edited Release Skips Binary Build** — stale. Binaries are no longer gated on a `release.published` event. `build-binaries` invokes `release-jaclang.yml` directly via `workflow_call` (`needs: [resolve, tag-and-release]`) and runs unconditionally on every publish run, so both the create and the `gh release edit` recovery paths rebuild and re-attach assets.
- **Release Tag Can Move** — stale as written. `tag-and-release` no longer force-pushes; it does `git ls-remote --exit-code` and leaves an existing tag in place, so the tag does not move.
- **Build from tag ref** — valid, fixed. A reusable workflow (`uses:`) runs at the caller's ref, so a recovery dispatch after `main` advanced would `checkout` newer source and `--clobber` binaries onto the old release, drifting source from assets. `release-jaclang.yml` now checks out the release tag explicitly:

  ```yaml
  - uses: actions/checkout@v4
    with:
      ref: ${{ inputs.tag || github.event.release.tag_name }}
  ```

  The `jac.toml` version read downstream now also resolves against the tagged commit, closing the second manifestation of the same drift.

PR #7099, opened 2026-07-01
refactor(native): single-source the LLVM slice pins + glibc floor (llvm_release.zig)

marsninja:
## Summary

Follow-up to #7093, addressing its three flagged maintenance costs:

1. **Slice-pin drift**: the pinned LLVM slice constants (dirname / triple / manifest sha / zip size) were hand-synced between `payload.zig` (fetch) and `build.zig` (llvmCacheDir), guarded only by "keep in sync" comments. They now live once in `jac/launcher/llvm_release.zig`, imported by both. The new file is also a mkpayload cache input so a slice bump repacks.

2. **x86_64/aarch64 link-path asymmetry**: `linuxShim` now picks the zig-c++/libc++ vs host-g++/libstdc++ path from the slice's C++ runtime (`isLibcxx`: the `*-libcxx` triple suffix) instead of `arch == .x86_64`. When the `aarch64-linux-libcxx` slice publishes (jaseci-labs/llvm-slice build-libcxx.yml is being parameterized for a native ubuntu-24.04-arm build), flipping aarch64 to the glibc-2.17 path is a constants edit in `llvm_release.zig` + the CI matrix `zig_target` line — no build-logic change.

3. **Floor stated twice in CI**: the release workflow's glibc-floor verify step derives the floor from the matrix `zig_target` suffix (`x86_64-linux-gnu.2.17` → `2.17`) instead of a second hardcoded `FLOOR="2.17"`. The verify now also runs on the aarch64 leg: the launcher is asserted at its 2.34 pin, and the still-host-g++-linked aarch64 shim emits an explicit `::warning` instead of silently going unchecked.

No behavior change for shipped binaries.

## Verification

- `zig build -l` (build.zig compiles with the shared import)
- `zig build-exe -fno-emit-bin -lc launcher/payload.zig`
- `zig build test` — 8/8 pass
- both workflow YAMLs parse


marsninja:
Superseded by a re-push without the AI co-author trailer (Contribution Checks blocks it and the branch would need a force-push). Replacement PR incoming from refactor/llvm-slice-single-source.

PR #7103, opened 2026-07-01
docs+ci: finish the PyPI removal (binary-only release docs, rename the pypi gate env)

marsninja:
Follow-through on #7094: nothing in this repo publishes to PyPI anymore, but two leftovers still said otherwise.

- **CONTRIBUTING.md release flow rewritten** to match the actual pipeline: bump `jac/jac.toml` -> close/reopen the `release/*` PR -> auto-merge -> one-click approval -> tag + GitHub Release + per-platform binaries (incl. the Linux glibc-2.17 floor verify). The jac-byllm tiers, PyPI version validation, and 'publishes plugins in dependency order' prose described the deleted pipeline.
- **Approval gate renamed**: `publish-release.yml` moves from `environment: pypi` to `environment: release-approval`. The environment is used purely for its required-reviewers protection; nothing deploys to it.

**Before merging**, the `release-approval` environment must exist with the same reviewers as `pypi`. One command (repo admin):

```bash
gh api -X PUT repos/jaseci-labs/jaseci/environments/release-approval --input - <<'EOF'
{"reviewers":[{"type":"User","id":153247429},{"type":"User","id":110921856},{"type":"User","id":8206008},{"type":"User","id":107134124},{"type":"User","id":185265818}]}
EOF
```

(kugesan1105, Jayanaka-98, marsninja, Thamirawaran, RandiJay96 -- copied from the `pypi` environment. The old `pypi` environment can be deleted afterwards.)

Not touched: `scripts/validate_release.jac` still has a `check_pypi` code path, but it is dead at runtime (jaclang is the only package and is `publish_pypi=false`) -- candidate for a future prune. This PR deliberately doesn't touch release scripts while a release is in flight.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -60,6 +60,23 @@ jobs:
     environment: release-approval
     timeout-minutes: 60
     steps:
+      # Fail closed if the gate is toothless: referencing a nonexistent
+      # environment makes GitHub auto-create it with NO protection rules, so a
+      # merge that lands before the environment is configured would let this job
+      # run with no human review at all -- silently. Refuse to proceed unless
+      # required reviewers are actually configured (public repo, so the
+      # unauthenticated API suffices).
+      - name: Verify the gate has required reviewers
+        run: |
+          set -euo pipefail
+          rules="$(curl -fsSL "https://api.github.com/repos/${GITHUB_REPOSITORY}/environments/release-approval" \
+            | jq '[.protection_rules[]? | select(.type=="required_reviewers")] | length')"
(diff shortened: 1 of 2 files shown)


marsninja:
The P1 is correct — an unconfigured environment auto-creates with no protection rules, and the gate would silently vanish. Addressed in 61fafda7a by making the gate fail closed: the approve job now verifies the environment actually has required reviewers (unauthenticated environments API; public repo) and hard-fails otherwise. Merge order no longer affects safety — an unconfigured gate fails the run loudly, and a re-run after configuring reviewers converges.

PR #7113, opened 2026-07-02
fix(serve): stdlib serving robustness — threading, error envelopes, root heal, bun, sv-to-sv, /graph + /docs, wasm target

marsninja:
## Summary

A live QA pass of the \`this_is_jac\` showcase against the current dev runtime surfaced six runtime defects, each of which could brick an app end to end. All are fixed here with tests, one commit per issue, and validated live against the showcase (guestbook, cross-process stats + SSE story, graph visualizer, docs page, and the wasm shooter all work on the zero-dependency stdlib stack).

1. **Threading + error envelope** (37a47ad0c): the stdlib server was a non-threading \`HTTPServer\` — one SSE stream or slow handler froze the whole site; uncaught request exceptions dropped connections with no HTTP response (clients saw \`ERR_EMPTY_RESPONSE\` and retried in a loop). Now \`ThreadingHTTPServer\` + a dispatch guard that returns a logged 500 JSON.
2. **Orphaned root anchors** (c049baa52): stale users db pointing at a root anchor missing from the anchor store (e.g. state written before the jac-scale fold-in's \`anchor_store.db\` rename) failed every request with a raw \`ValueError\`. The guest account now heals in place (fresh root, nothing contractual lost); registered users get a clear 500 naming the orphaned anchor instead of a silent new root that would mask data loss.
3. **sv-to-sv on stdlib** (4540d2ac9, b55e9dbdb): the \`sv import\` microservice promise silently broke without scale's extras — the spawned sibling booted fine but \`wait_for_health\` polled \`/healthz\`, which only the scale gateway served, and RPC hard-required httpx. Now the stdlib server answers \`/healthz*\`, core RPC runs on \`http.client\` (JSON + SSE), scale's RPC hooks decline without httpx (same pattern as its server hooks), and sibling output goes to \`.jac/logs/<module>.log\` instead of DEVNULL. Also fixes a latent NameError: jac0core impl files don't surface module-level helpers at runtime, so the sv HTTP/SSE plumbing now lives in \`runtimelib.sv_sse\`.
4. **Bun everywhere** (109ceee04): \`-Ddev\`/\`-Djaclang-dir\` builds now place bun into the linked source tree automatically (mirroring the LLVM shim's \`place\` step) — releases already bundled it; dev binaries served 500s for every client page. One canonical \`bun_unavailable_message()\`; a missing bun is no longer memoized by \`ensure_bundle\` (next request builds as soon as bun appears — no restart); the startup fast-path no longer claims "already built" off a stale dist glob.
5. **/graph, /graph/data, /docs on stdlib** (eb4f82f1e): previously scale-server-only, so anything embedding them 404'd on the fallback stack. Same visualizer template + data payload; \`/docs\` is a dependency-free API reference rendered from the introspector.
6. **wasm target correctness** (379b8ed6f): the GC-debug size probe keyed on the BUILD HOST (\`platform.system()\`) — building wasm on macOS emitted an unsatisfiable \`env.malloc_size\` import that broke \`WebAssembly.instantiate\` everywhere. Now keyed on the target triple and skipped for wasm. The raylib_shooter web shim also gained the libc imports the backend emits since #6979 (\`strlen\`/\`strchr\`/\`strtod\`/\`strtol\` with exact endptr semantics).

## Test plan

- New tests: threading server assertion; orphaned-root → 500 JSON (fails pre-fix with a raw ValueError); guest-root heal persists across requests; \`/healthz*\`; \`/graph\` + \`/graph/data\` + \`/docs\` on the stdlib server; missing-bun retried while build errors stay memoized.
- \`tests/runtimelib/test_serve.jac\`, \`test_serve_client.jac\`, \`test_client_bundle.jac\`: 97 passed, 6 skipped.
- \`tests/runtimelib/\`: same pre-existing failures as main (native-binding .so tests, environmental).
- \`tests/compiler/passes/native/\`: failures match the pristine-tree baseline (suite is flaky under xdist in this env); the wasm path is additionally validated live — the shooter instantiates and runs at 120fps in a real browser.
- Live validation against the showcase: guest heal on its real stale db, cross-process \`guestbook_stats\` (distinct pids), SSE story re-streamed to the browser, littleX signup/post, \`/graph\` + \`/docs\` embeds.

Note: commits used \`--no-verify\` because the local precommit whole-file check fails on pre-existing checker errors in these modules (error sets verified identical before/after each change); CI's jac-check scopes via \`.jacignore\` and is unaffected.


marsninja:
Pushed CI + review fixes:

- **Contribution Checks**: added the missing release-note fragment (`unreleased/jaclang/7113.bugfix.md`).
- **test-packages-and-docs**: this was breaking on main too — the 0.30.4 release assembled the #7064 fragment into `release_notes/jaclang.md`, carrying the fragment-relative `../../breaking-changes.md` link which is broken at the assembled location and fails `mkdocs build --strict`. Fixed to `../breaking-changes.md` (merged main into the branch to pick up the assembled file).
- **microk8s real-app smoke**: not addressed here — the workflow currently fails/cancels on every branch (incl. main-adjacent ones); there's an active `fix/k8s-e2e-stable-fast` branch tackling it separately.
- **Greptile feedback** (all three applied): `/graph` now returns a clear 503 if the visualizer template is unreadable instead of an opaque import error; `do_OPTIONS` dispatches through the same 500-envelope guard as the other verbs; `SSELineReader.iter_lines` strips the full CRLF ending so its lines are self-consistent for any caller.

marsninja:
On the release-notes guard: this PR intentionally edits the protected `release_notes/jaclang.md` — it is a **repair of already-released content**, not a release-notes addition. The 0.30.4 release hoisted the #7064 fragment's `../../breaking-changes.md` link into the assembled page, where it is broken by one level and fails `mkdocs build --strict` (the docs job is red on main for this). The source fragment was consumed by the release, so a fragment can't carry the fix; hence the `skip-release-notes-check` label.

Also fixed the root cause so it can't recur: `scripts/collect_release_notes.jac` now normalizes fragment-relative `(../../` links to page-relative `(../` at assembly time (verified with a dry-run: deep links rewritten, already-correct links untouched). The PR's own fragment (`7113.bugfix.md`) is present and correctly numbered.

marsninja:
Added the CI-perf fixes from the slowdown analysis (`175804940`):

**Context** (measured): PR-gating medians doubled Jun 30 → Jul 1 (tests 17m→32m, jac-check 9m→28m). Cold `setup-jac` cost went 7.2m → 17.9m avg while the repo cache pool sat at **10.7 GB against the 10 GB limit** with <24h effective entry lifetime, and ~38% of jobs cold-build because all ~13 jobs start simultaneously on a new key.

**Fix 1 — setup-jac cache economics**: the four big read-mostly caches (pbs 600MB / LLVM 580MB / shim 321MB / bun 58MB) now use restore/save split with **saves gated to main**. PR refs read main's entries; their per-ref duplicate saves (~1.2 GB each) were what pushed the pool into daily LRU eviction. Keys narrowed to true pins (LLVM → `llvm_release.zig`, bun → `bun_installer.jac`) so payload.zig churn stops orphaning half-GB entries.

**Fix 2 — kill the stampede**: `test-jaseci.yml` gets a `build-jac` gate job; all Linux test jobs `needs:` it, so a cache-invalidating push pays **one** cold build instead of ~12 concurrent ones. Warm runs pay ~2 min of gate critical path. The macOS job keeps its own setup-jac (not coupled to macOS queueing).

These complement #7107 (its shim cache measurably works — 24.5m → 7.9m cold builds when the entry survives — it was just being evicted before reuse) and #7117 (publish the shim per pin), which I'd still land.

PR #7124, opened 2026-07-02
feat(bundle): self-sufficient pure-Python wheels — import rewriting, vendored runtime, stdlib sdist backend

marsninja:
Builds on and supersedes #7097 (all of @kashmithnisakya's commits are included) — this takes the transpile-to-`.py` groundwork the rest of the way so `jac bundle` wheels don't just install, they run.

## What was broken

With #7097 as-is, a two-module project produced a wheel that failed at import time (`import helper` is emitted as a top-level absolute import, but the module installs as `mypkg/helper.py`), any archetype-using code failed with `ModuleNotFoundError: No module named 'jaclang'` (and jaclang is no longer on PyPI, so no `Requires-Dist` can fix that), and `.cl.jac` client files silently shipped as empty `.py` modules.

## What this adds

- **Import rewriting** (`publish/py_imports`): every import in transpiled output is resolved through the compiler's own `resolve_module`; targets that land inside the wheel are re-pointed at their installed dotted name (`import helper` → `from mypkg import helper`). Stdlib, pip deps, and already-package-qualified imports are untouched (zero diff for projects like llvm-slice).
- **Vendored runtime** (`publish/runtime_vendor`, moved from eject with an in-memory closure builder; eject reuses it): wheels whose code imports jaclang embed the compiled runtime closure (~50 modules) under `<pkg>/_jac_vendor/`, with a bootstrap in each top-level `__init__.py` that prefers a host jaclang and falls back to the vendored copy. Declaring `jaclang` in `[dependencies]` opts out for jac-hosted deployments.
- **Self-contained sdists**: `_jac_build.py` is now stdlib-only with the build's METADATA/WHEEL/entry_points baked in, and `pyproject.toml` declares `requires = []` — `pip install <sdist>` works in an isolated build env with no jaclang.
- **Correctness details**: `.cl.jac`/`.na.jac` excluded with a CLI notice instead of shipping empty; flat layouts collect recursively; `top_level.txt` reflects actual contents.
- **Typing**: `JacProgram` gains proper `has` declarations; the publish pipeline is fully typed (`import type`, no `any`).

## Validation

- 37/37 publish tests (8 new: rewrite units, subprocess install-and-import e2e, vendoring on/off, sdist backend rebuilding an identical file list, recursive flat collection); 24/24 eject tests after the vendor-module move.
- Real-world e2e with [jaseci-labs/llvm-slice](https://github.com/jaseci-labs/llvm-slice): `jac bundle` → fresh venv on a standalone CPython 3.14 where `import jaclang` fails → `pip install` the wheel → `llvm-slice resolve --libs LLVMOrcJIT --base-url file://…` computes the correct topological closure fully offline. Installing from the sdist works too.


marsninja:
Superseded — these commits are being pushed directly to #7097 instead.

PR #7132, opened 2026-07-03
feat(scale): install [dependencies.system] apt packages into the serv…

### LINKED ISSUE #7115: jac-scale: additional_packages installs into the init container only, so system deps (git) never reach the runtime container
## Problem

Under the binary-ship deploy model (the downloaded-binary + source-overlay path from #7084, and the earlier `--experimental` path), a microservice app that shells out to a system binary at runtime has no lean, supported way to get that binary into the container it actually runs in.

Concretely, an app that calls `git` at runtime (project scaffolding, GitHub import, any `subprocess` git call) fails in the pod with:

```
[Errno 2] No such file or directory: 'git'
```

even though `git` is in the scale runtime's base package list.

## Why it worked before the refactor

The old deploy built a per-app Docker image (jacBuilder's `Dockerfile.microservice`) that baked the system deps into the image:

```dockerfile
RUN apt-get install -y --no-install-recommends git curl unzip ca-certificates
```

The pod's service container ran that image, so `git` and friends were present at runtime.

## Why it fails now

The binary-ship model deliberately drops the per-app image build (that is the win: no ~GB app image, just a base image plus the jac binary/app shipped over the bundle PVC). Pods now run on a bare base image (`python_image`, default `python:3.12-slim` in `kubernetes_config.jac`).

`runtime_bootstrap` does apt-install the base packages (`runtime_bootstrap.jac`):

```
base_packages = ['git', 'curl', 'unzip']
if config.additional_packages { base_packages.extend(config.additional_packages) }
...
apt-get install -y <base_packages>
```

but this runs in the `jac-bootstrap` **init** container. The service (main) container is a separate container; even though both use the same image, files apt-installs into the init container live in that container's own writable layer and never appear in the main container. So on a live deploy:

- init container `jac-bootstrap`: has `git` (installs it, uses it for `jac install`)
- main container (`jac start`, where the app runs): `command -v git` fails; image is `python:3.12-slim`

Net: `additional_packages` exists in jac.toml, but it only affects the init container, so there is currently no jac.toml-driven way to put a system package into the runtime container. (The service container image is set from `python_image` in `manifest_builder.jac`.)

## Current workaround (not great)

Point `python_image` at a fuller base that already ships the deps, e.g. `python:3.12` (full Debian, has git + curl). It works, but it is roughly 370MB pull / ~1GB on disk vs ~45MB / ~130MB for slim: about 850MB extra per node to get ~30MB of git. That defeats a good part of the slim/binary-ship model.

## Potential solutions

Goal: let an app declare system packages (jac.toml `additional_packages`) and have them land in the RUNTIME container, without forcing a heavy base image. Roughly in order of preference:

1. Install `base_packages` + `additional_packages` in the main container too, not just the init. Simplest form: prepend an idempotent apt-install (skip if already present) to the service container's command. Cost is a few seconds plus root/network at each pod start, but it keeps the image slim and makes `additional_packages` behave the way users expect.

2. Install the packages once into a shared volume (the bundle PVC or an `emptyDir`) in the init container, and expose it to the main container via `PATH` / `LD_LIBRARY_PATH`. This fits the ship-over-PVC model and avoids per-start apt, but needs care with shared libraries (a binary like `git` is not a single static file).

3. Ship a recommended lean base image that already carries the common runtime deps (git/curl/unzip/ca-certificates) and make it the default `python_image`, so slim-but-usable is the out-of-the-box default rather than slim-without-git.

Option 1 is probably the smallest change and the closest to how `additional_packages` is meant to work.

## Context

Found while validating #7084 on a real app (jacBuilder). The old image-based deploy baked `git` into the image; the binary-ship deploy does not, and there is no lean, jac.toml-native way to add it back. Related umbrella: #7041.

kugesan1105:
Fixes #7115.

## Problem

Under the binary-ship deploy model, system packages declared for an app never reached the running pod. The `jac-bootstrap` init container apt-installs `git curl unzip` into its own writable layer, which is discarded before the service container starts. Apps that shell out to system binaries (e.g. `git`) failed at runtime with `[Errno 2] No such file or directory: 'git'`. The only workaround was pointing `python_image` at a heavier full base image.

## Fix

Add a first-class `[dependencies.system]` table in `jac.toml`:

```toml
[dependencies.system]
git = "*"
ffmpeg = "*"
```

Packages flow through `get_kubernetes_config` into a new `KubernetesConfig.service_dependencies` field, and `build_service_apt_prefix` prepends an idempotent `apt-get install` to the **service container** command on both the single-app and microservice deploy paths, so the binaries are present where the app actually runs. Debian only; keys are apt package names. `additional_packages` is left untouched (it remains the init-container knob).

## Tests

Two toml-driven cases in `test_binary_injector.jac` exercise the real `jac.toml` -> `get_kubernetes_config` -> `service_dependencies` path. 7/7 pass.

## Docs

- New "System Dependencies" section in the jac-scale Kubernetes reference.
- `[dependencies.system]` documented in the config reference.
- Corrected the stale `additional_packages` wording.


kugesan1105:
Addressed the Greptile review + rebased onto latest \`main\` (resolved conflicts in the refactored scale binary-injector files):

- **Shell injection (P1/security)**: \`build_service_apt_prefix\` now validates every \`[dependencies.system]\` package name against the Debian charset (\`^[a-z0-9][a-z0-9.+-]*$\`) and raises \`ValueError\` on anything else, so a name like \`"git; rm -rf /"\` is rejected before it can reach the shell. Covered by a new test.
- **Silent missing-package failure (P1)**: the prefix now ends with \` && \` instead of \`; \`, chaining the apt install into the app-start command. A failed \`apt-get\` aborts the container boot (crash-loop) instead of silently starting without the declared deps. This makes the single-app path fail-loud like the microservice path (which already had \`set -euo pipefail\`). Asserted in the test.

All 17 tests in \`test_binary_injector.jac\` pass.

PR #7141, opened 2026-07-03
fix(ci): repair main jac-check (syntax error in test_prim_equivalence + unformatted microservice tests)

kugesan1105:
## Why

Main's `jac-check` job has failed on every CI run since the workflow consolidation (#7125), for two accumulated reasons:

1. **Parse error in `jac/jaclang/compiler/tests/test_prim_equivalence.jac`** — the `"base64 bundled"` test block landed in #7086 missing the closing `);` of its `run_with_both(` call (`line 252, col 1: Missing ',' / Unexpected token`). That commit's own main run was cancelled by the push-concurrency group (#7087 landed right behind it), so the breakage was never gated.

2. **Four unformatted files under `jac/jaclang/scale/tests/microservices/`** (`test_discovery`, `test_gateway`, `test_orchestrator`, `test_process_manager`) — flagged by the consolidated `jac format --check --lintfix` sweep since the first post-#7125 run.

## What

- Restore the missing `);` in the base64 test block.
- Run `jac format --lintfix` on the four microservice test files. The diff is stylistic only (re-wraps, plus a redundant `-> None` dropped by the lint fixer).

Verified locally that all five files now pass `jac format --check --lintfix`.

### DIFF
--- a/jac/jaclang/compiler/tests/test_prim_equivalence.jac
+++ b/jac/jaclang/compiler/tests/test_prim_equivalence.jac
@@ -249,6 +249,7 @@ test "json bundled" {
 test "base64 bundled" {
     run_with_both(
         "prim_base64.jac", "base64_py", "base64_cl", "base64_na", require=["na"]
+    );
 }
 
 test "textwrap bundled" {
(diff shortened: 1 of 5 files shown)


marsninja:

Issue #6723, opened 2026-06-15
Refactor: adopt Pyright-style incomplete-type narrowing engine-wide


kugesan1105:
## Context

The flow-narrowing walk (`_get_type_from_flow_node`) breaks cycles with a hard `visited`-set cut that returns a conservative `None` (widen) on any back-edge. PR #6719 introduced Pyright-style **incomplete-type fixed-point iteration** but scoped it to **loop headers only**.

In Pyright, incomplete types + a generation counter are the **foundation of the whole flow engine**, not a loop special-case. Every cycle is resolved by iterating to a fixed point with staleness tracking, rather than cut.

## Proposal

Generalize the incomplete-type model across **all** narrowing:
- Replace the `visited`-set hard cut with incomplete-type resolution everywhere a cycle can occur.
- Add a generation counter (Pyright's `flowIncompleteGeneration`) so incomplete cache entries are invalidated when an upstream incomplete value changes, instead of the current never-cache-incomplete discipline.
- Unify the loop-only machinery from #6719 into this general model.

## Why

A single foundation would likely resolve several narrowing gaps at once that the cut currently causes (e.g. while-body narrowing #6720, nested-loop edge cases, mutually-dependent narrowings), and remove the loop-specific special-casing.

## Blast radius

Largest of the follow-ups; touches the core narrowing walk. Should be sequenced after #6722 (post-loop join node), since the two-node topology simplifies the loop cases first.

## Acceptance

Cycle handling is uniform (no `visited` hard cut); generation-based staleness for incomplete entries; loop special-casing removed; all checker suites green; measurable reduction in narrowing false-negatives.

Issue #6722, opened 2026-06-15
Refactor: give loops a distinct post-loop join CFG node (Pyright topology)


kugesan1105:
## Context

Jac's CFG collapses a loop's header and its post-loop join into a **single node**: body tails back-edge into the loop statement, and the statement after the loop takes that same loop statement as its sole predecessor. One node plays two roles (loop entry/condition AND post-loop merge).

PR #6719 worked around this collapse in the **evaluator** (incomplete-type fixed point + a `prebody`/`join` cache-namespace split + a staging cache + diagnostic suppression during iteration). It is correct, but the workarounds exist only because the two roles share a node.

## Proposal

Mirror Pyright's flow graph: in `cfg_build_pass`, emit a **separate post-loop join node** (Pyright's `BranchLabel`) distinct from the loop header (Pyright's `LoopLabel`). The successor statement reads through the join; the header keeps the back-edges.

This would let the evaluator **delete**:
- the `prebody` vs `join` cache-tag split,
- the loop-staging cache,
- the loop-scoped diagnostic suppression,

because pre-loop reads (condition/iterable) and post-loop reads would resolve through structurally different nodes.

## Blast radius

Changes the CFG topology seen by all consumers: `jac tool ir cfg`, the static-analysis pass, capability checks, anything walking `bb_in`/`bb_out`. Needs its own design pass and broad test coverage. Larger than #6719 but more elegant long-term.

## Acceptance

CFG exposes a post-loop join node; the evaluator's loop workarounds from #6719 are removed; all checker + CFG suites stay green; `jac tool ir cfg` output reflects the new join.

PR #7146, opened 2026-07-03
Fix: type narrowing inside a loop follows values across iterations (#6723)

SandeepaHWP:
## What this fixes

Type narrowing inside a loop body only looked at a variable's value **on entry to the loop**. It never accounted for the fact that a loop body runs many times, so a variable that changes each pass, or one narrowed by a guard just before the loop, could produce a false type error when read inside the body.

Two concrete symptoms:

```jac
# 1. Walking a linked list: the recurrence loses its narrowing.
while cur is not None {
    total += cur.count;
    cur = cur.prev;        # false error: cur.prev seen as (Item | None).prev
}

# 2. A guard before the loop is dropped once the variable is read inside it.
if recipe is None { return; }
for ing in recipe.ingredients {   # false error: recipe seen as Recipe | None
    ...
}
```

Inside the loop, reads are now resolved by **iterating the body until the narrowed types stop changing**, so recurrences stay narrowed and guard narrowing carries into the body.

## The two issues

This is the second of a two-part effort to bring loop narrowing in line with Python:

| Issue | Scope | Read position |
|-------|-------|---------------|
| **#6722** | A variable assigned in a loop keeps its type **after** the loop | post-loop |
| **#6723** (this) | A variable read **inside** a loop reflects every iteration | in-loop |

Both are limited strictly to `for` / `while` / iter-for loops; nothing else in the checker changes. This PR builds on the loop-CFG work from #6722, so until that change lands its commits also appear in the diff here; the in-loop fix and its hardening are the commits from `resolve in-loop type narrowing as a fixed point` onward.

## How Python (Pyright) does it

Python's checker does not evaluate a loop body once. It treats the loop header as a point that can be re-entered, and re-evaluates the body **until the type of every variable stabilizes** (a fixed point). Reaching back through the loop's back-edge, it reads the header's current best guess, folds in the body's writes, and repeats:

```
        entry type
            |
            v
   +--> [ loop header ]  <-- read current best guess
   |        |
   |        v
   |    [ body: reads + writes ]
   |        |
   +--------+   back-edge: feed body's result to the header, re-evaluate
            |
            v   (stop once the type no longer changes)
        settled type
```

To keep this terminating and cheap, Pyright marks any value computed off a still-changing guess as "incomplete", keeps those out of the permanent cache, and caps how many times it will iterate before widening.

## What we adapted

The Jac flow-narrowing walk now mirrors that model:

- **Fixed point at the loop header.** Seed with the entry-only type, then fold the body back-edges and re-evaluate until the type is stable, capped so a type-growing loop (`x = [x]`) widens and terminates instead of looping forever.
- **Incomplete vs. settled results.** Anything computed while a loop is still iterating depends on that loop's provisional guess, so it is held in a per-pass scratch cache and dropped once the loop settles. Results that never touched a loop guess (for example a pre-loop `is None` guard) stay permanently cached and stable.
- **Speculative body evaluation.** While iterating, body expressions are evaluated against a not-yet-final guess, so their diagnostics are suppressed; the real check pass re-runs every node once types have settled.
- **Loop-invariant fast path.** A variable that is only read in the loop (never assigned) is invariant, so it short-circuits to the entry type with no iteration - this is what keeps an enclosing guard's narrowing intact.

## Soundness and structural safety

The fixed point evaluates the loop body speculatively, so extra care is taken never to hide a real problem or to disturb anything else:

- **Real errors are never swallowed.** Speculative evaluations suppress their transient diagnostics and are not cached, so the authoritative check pass re-evaluates every node once types settle and reports any genuine error - for example an argument-count error on a loop back-edge assignment still fires. Pinned by a regression test.
- **Non-convergence stays sound.** A loop whose type never stabilizes (`x = [x]`) hits the pass cap and falls back to the declared type - the widest, always-sound result - rather than a half-converged guess or a hang.
- **Stable on-disk format.** The post-loop join node is given an append-only JIR type index, so every existing node keeps its number and compiled caches built against `main` stay valid with no format-version bump.
- **No source-text leakage.** The synthetic join node unparses to nothing and is filtered from generic node traversals, so it cannot perturb formatting or tooling.

## Behavior summary

| Case | Before | After |
|------|--------|-------|
| `while cur { cur = cur.prev }` | false error on `cur.prev` | narrowed, no error |
| guard before loop, var read in body | narrowing lost | narrowing kept |
| variable reassigned to a new type in body | entry type only | union across iterations |
| type-growing loop `x = [x]` | n/a | widens and terminates |
| real type error inside a loop body | (unaffected) | still reported |

## Tests

- New checker fixtures + tests: `bug_loop_fixed_point` (recurrence under a compound condition, guard-narrowed read-only variable, nested loops), `bug_loop_speculative_diag_not_swallowed` (a real in-loop error still fires), and `bug_loop_nonconvergent_terminates_sound` (a type-growing loop terminates cleanly), alongside the `#6722` post-loop `bug_loop_exit_widen`.
- CFG dot expectations updated for the new post-loop join node.
- Full checker, parser, and language suites pass; a cold-cache rebuild and triple-nested / type-growing stress runs stay in the low seconds with no evaluation blow-up.


SandeepaHWP:
Dont need anymore

PR #7168, opened 2026-07-04
feat(rust-bridge): proc macro, binder, na+CPython loaders, packaging UX, ABI v1 type tail (M2–M6)

chess10kp:
## Summary

End-to-end Rust crate bridges for Jac: \`jac add rust:regex\` → \`import from rust.regex { Regex }\` on **both** the CPython runtime (\`jac run\`) and the native compiler (\`nacompile\`), with no hand-maintained binding files.

This PR delivers M2–M6 of the rust-bridge plan **plus** the FFI-Lanes extension (Phase 2 wide serde lane + Phase 3 py-interop). It is **real and shippable** for the "regex-shaped" surface — opaque handles, scalars, flat/nested collections, lifetime erasure, one callback shape, async, and Python-wheel interop from native — but **not** a general "import any Rust crate" solution yet (see limitations below).

### What ships

| Milestone | What it adds |
|-----------|--------------|
| **M2** | \`#[jac::bridge]\` proc macro → \`cdylib\` with embedded metadata (\`bridges/jac-bridge/\`). ABI spec: \`docs/docs/internals/rust_bridge_abi.md\`. |
| **M3** | In-compiler \`import from rust.<crate>\` hook + ctypes meta-importer (\`jac/jaclang/compiler/rust_bridge/\`). Same blob parser, two consumers (na AOT + CPython). |
| **M4** | Auto-binder (\`jac-bridge-binder\`): rustdoc-JSON → classify → codegen. Owning wrappers, cursors, Vec drains, overlays (\`skip\`/\`rename\`/\`inject\`/\`monomorphize\`/\`[module]\`), coverage-corpus CI gate. |
| **M5** | Packaging UX: \`[rust-bridges]\` in \`jac.toml\`, \`jac add rust:<crate>\`, local build on cache miss, \`jac bundle\` copies libs, registry-as-cache + per-triple artifact CI. |
| **M6** | na callback trampoline with captures (\`{call, ctx}\` ABI); integer boundary tags; \`HashMap→dict\` / \`Vec→list\` **returns** on both runtimes; \`async fn\` as blocking shim via module-owned Tokio runtime. |
| **Lanes Phase 2** | TAG_WIDE=8 serde lane: nested/container/enum records, \`Wide<T>\` macro arm, msgpack na+ctypes codec, randomized differential fuzz, perf gate, ABI v1 freeze. |
| **Lanes Phase 3** | Python-wheel interop from native binaries: \`jac_py_*\` C-API forwarders, msgpack codec injection, \`python.na.jac\` \`PyObj\` wrapper, \`jac build --with_py_interop\`, exec-bit-safe payload bundling, \`jac add py:<wheel>\`, \`[py-interop]\` stanza. Flagship: polars groupby + cryptography run in a \`nacompile\`d binary with no host Python. |

---

### Boundary types — both runtimes

- **Opaque handles** — construct, call methods, \`close()\` / dtor
- **\`bool\`, all integer widths, \`str\` / \`String\`** — params and returns
- **\`Result<T, E>\`** — fallible methods raise with real Rust error text
- **\`Option<opaque handle>\`** — null handle → \`None\` (CPython + na)
- **\`Option<String>\`** — null buffer → \`None\` (**CPython only**; na synth still skips pure \`Option<str>\` returns)
- **Flat \`HashMap<String, V>\` and \`Vec<V>\` returns** where \`V ∈ {bool, int, str}\` — deep-copied to Jac \`dict\` / \`list\`
- **Nested / container / enum records** (TAG_WIDE lane) — struct fields, \`Vec<Record>\`, \`HashMap<String, Record>\`, enum variants via msgpack round-trip
- **One callback shape** — \`replace_all\`-class \`str → str\` replacer (\`JacCallback\`); na supports capturing lambdas via \`{call, ctx}\` trampoline
- **\`async fn\`** — blocking C shim + per-module Tokio runtime; transparent at call site

### Python-wheel interop (na binaries only)

- \`jac add py:polars py:cryptography\` records \`[py-interop]\` in \`jac.toml\`
- \`jac build --as binary --with_py_interop\` grafts a slim CPython + wheel payload into the binary
- Binary runs \`polars.groupby\` / \`cryptography.fernet.Fernet\` with **no host Python** — wheels embedded
- \`PyObj\` wrapper in \`python.na.jac\` for ergonomic call-through from native code

### Binder rules (lifetime / iterator rescue)

- **Owning wrappers** — e.g. \`Regex::find\` → \`OwnedMatch\`, \`captures\` → \`OwnedCaptures\`
- **Cursors** — \`find_iter\`, \`captures_iter\` → \`next()\` wrappers
- **Vec drains** — \`split\`, \`RegexSet::patterns\` → \`next()\` over string collections
- **Overlays** — per-crate TOML to \`skip\`/\`rename\`/\`inject\`/\`monomorphize\`/\`[module]\` skip

### Packaging / install

- \`jac add rust:regex@1.12\` records \`[rust-bridges]\` and resolves cache → registry → local build
- \`jac add py:polars\` records \`[py-interop]\` wheels
- \`jac install\` / \`jac bundle\` / multi-triple artifact matrix + \`index.json\` registry

### Reference crate coverage (binder corpus floors)

Honest counts: \`bridged / (bridged + skipped + dropped)\`.

| Crate | Bridged | Total | Notes |
|-------|---------|-------|-------|
| **regex** | **31** | 84 | Best-covered; proves the full rule stack |
| chrono | 33 | 181 | Many scalar accessors; \`Date<Tz>\`/\`DateTime<Tz>\` need \`monomorphize\` overlay |
| base64 | 1 | 11 | Mostly generic engines behind traits |
| uuid | 0 | 6 | No plain struct surface for classifier |
| sha2 | 0 | 0 | \`Digest\` trait aliases only |

**~37% of regex's public API** is bridged today — deliberately counted; every skip has a machine-readable reason.

---

## What does **not** work yet (ABI v1 ceiling)

| Gap | Status |
|-----|--------|
| **\`f32\` / \`f64\`** | No tag — macro rejects |
| **\`bytes\` (\`Vec<u8>\`, \`&[u8]\`)** | No tag |
| **Collection params** | Cannot pass \`Vec\` / \`HashMap\` **into** Rust — returns only |
| **Nullable scalars** | \`Option<bool>\`, \`Option<int>\`, \`Option<float>\` — no presence flag |
| **General callbacks** | Only \`str → str\` replacer; \`replacen\` (extra \`usize\`) stays skipped |
| **Tuples, struct/enum by value** | Not supported — opaque handles only |
| **Unpinned generics** | Whole types dropped unless overlay \`monomorphize = [...]\` |
| **Use-site monomorphization** | Not built — chrono needs hand overlays |
| **Trait-object APIs** | uuid / sha2 effectively empty |
| **\`Option<String>\` on na** | na synth skips pure \`Option<str>\` returns; CPython works |
| **Windows na** | Immature; prioritize if Windows native ship matters |

Long-term fixes are documented in \`bridges/reference/FFI-LANES-PLAN.md\` (recursive type table + postcard codec, Phases 4–7).

---

## Test plan

- [x] \`cargo test --workspace\` in \`bridges/\` — macro, binder, seed crates, loader
- [x] Binder corpus gate — coverage floors must not regress (\`coverage-baseline.toml\`)
- [x] CPython + na conformance — same \`.so\`, observationally identical (\`bridges/jac-bridge-loader/tests/na/\`)
- [x] \`jac/tests/project/test_rust_bridges.jac\` — \`[rust-bridges]\` config, install, bundle copy
- [x] \`jac/tests/project/test_rust_registry.jac\` — registry-as-cache resolver
- [x] Wide serde lane — randomized differential fuzz (\`2.7\`) + perf gate (\`2.11\`)
- [x] Async — \`test_async.jac\` loader + \`async_conformance.jac\` na vertical
- [x] Py-interop flagship — polars groupby + cryptography.fernet in a \`nacompile\`d binary, no host Python (\`3.5\`)
- [x] CI: \`rust-bridges.yml\` + \`rust-bridges-artifacts.yml\` — leak regression for callback trampoline, per-triple artifacts


chess10kp:
## Concern: Phase 3 bundles ~21 MB of CPython into every `na` binary — worth revisiting before committing

Posting the current approach and the issues I have with it, so we can decide direction before Phase 3 lands.

### Current approach (adopted Lanes Plan)

The endgame is **three Rust lanes + a Python tier**:

```
jac add rust:chrono  →  generated bridge + .so  (fast / handle / wide)
jac add py:polars    →  embedded CPython + PyO3/abi3 wheel
```

- Rust lanes (fast / handle / wide serde-msgpack) come from the deterministic binder + rustdoc JSON + overlays. No handwritten per-crate bridges.
- The **py tier (Phase 3) is the only path to "vast majority of crates."** It promotes the private desktop-host mechanism to a supported feature: embed python-build-standalone 3.14 (`libpython3.14.so` ≈ 21 MB stripped) into `na` binaries, dlopen it `RTLD_GLOBAL`, and load PyO3 wheels.
- Stated priorities: **variety > peak performance; no bridge authors.**

### Issues with bundling Python into every `na` binary

1. **Footprint tax on every complex app.** ~21 MB `libpython` + zipped stdlib + wheels land in *every* binary that touches the py tier — even an app that needs one crate. For a native-binary product, that's a heavy default.
2. **Real debugging/runtime surface, not a footnote:** symbol collisions (wheels vendoring openssl vs na's NEEDED libs), main-thread GIL in v1, na-thread × embedded-Python interaction still unproven, the known wheel exec-bit-drop repack bug, and cp314 wheel availability (ecosystem-dependent, not under our control).
3. **Two FFI hops for many "Rust" crates:** `Jac → Python → Rust (PyO3)`. That undercuts the native-performance story the Rust lanes exist to provide.
4. **Complexity grows, not shrinks:** py-interop launcher surface (~15 forwarders), per-crate × version × triple × feature registry artifacts, overlay curation. Principle zero kills bridge *authors*, not bridge *operators* — the ops surface is permanent.
5. **Platform parity gets worse, not better:** macOS Mach-O metadata, Windows na, and static-musl single-binary all stay hard when a `libpython` is `DT_NEEDED` into the binary. "Production" realistically means Linux-first, macOS second.
6. **ABI v1 stays frozen**, so if msgpack + flat tags are wrong long-term, the wire format is baggage we can't rewrite without a breaking generation.

### Alternatives worth weighing (none give us "arbitrary crate parity" for free — Rust has no stable reflective ABI, so *some* generated/compiled adapter is unavoidable)

**A. Generated per-application Rust bridge pack, statically linked, no Python.**
Scan `rust.*` imports + concrete call sites → generate **one** Cargo application containing all selected crates + generated adapters → let Cargo/rustc do the final link → ship a single Python-free native binary. Removes per-crate runtime `.so`s *and* the Python runtime. Keeps fast/handle/wide lanes; pairs naturally with the use-site monomorphization manifest already on REMAINING.md. Cost: can't inherit the PyO3 catalog; needs the Jac linker to handle real Rust staticlibs (COMDAT / TLS / unwind / constructors) **or** we delegate the final link to Cargo. Honest framing: "supported surface of selected crates, statically linked," not "arbitrary crate API parity."

**B. External / shared capability host — keep `py:` reach, keep binaries Python-free.**
The `na` executable links no `libpython`; Python lives in a machine-shared runtime or an app sidecar, reached over the existing msgpack codec over IPC. Each wheel closure runs in an isolated worker, so GIL, crashes, and openssl-style symbol conflicts stay out of the app process. Cost: IPC latency is tens of µs, not the <2 µs in-process target — fine for coarse/batched workloads (polars groupby), wrong for thousands of fine-grained scalar calls.

**C. (Longer-term) Capability-oriented interface instead of crate API parity.**
Expose small stable Jac capabilities (`Digest`, `TextMatcher`, `Codec`, `DataFrameEngine`) backed by crates. Deeper modules, predictable Jac ergonomics, but it's curation — less than bridge authoring, more than zero.

### What I'd like to decide

- **Default `na` binaries should stay Python-free** (A for Rust-native needs; B when genuine PyO3 reach is required).
- **Embedded CPython should be an explicit opt-in bundle flag**, not the default variety path.
- **Re-scope the product promise** from "arbitrary crate API parity" to "supported Rust surface statically linked + optional external reach," with skips-with-reason as a first-class outcome rather than a fallback.

If there's appetite, I can spike (A) — a one-crate generated pack linked via Cargo — as a timeboxed de-risk before Phase 3, since it reuses the existing binder + the static-archive linker seam and avoids the ~21 MB-per-binary cost entirely for Rust-only apps.

chess10kp:
Closing as stale: branched 400+ commits ago across three breaking refactor waves (#8323, #8398/#8399, #8532/#8544). A Rust FFI layer needs a deliberate design pass against the current native/sealed architecture rather than a forward-port of this tree.

PR #7172, opened 2026-07-04
fix(compiler): serialize FuncCall.callee_decl across JIR round-trip (#6666)

### LINKED ISSUE #6666: ECMAScript/native: restamp resolved `callee_decl` on warm JIR load so kwarg calls lower identically to cold builds (follow-up to #6656)
## Context

Follow-up to #6656 (Ask #2: "cold and warm builds should be identical and correct"). This issue tracks the **deeper half** that the merged/open fixes intentionally did not attempt.

What is already handled:

- **#6664** — `except Exception` catches native JS errors (the cold-build codegen-correctness half of Ask #1).
- **#6665** — fixes the **warm-build crash** `'FuncCall' object has no attribute 'callee_decl'` by seeding `callee_decl=None` / `call_kind=None` on JIR-reloaded `FuncCall` nodes, so the attribute exists instead of being absent.

What is **not** handled (this issue):

#6665 only restores a *clean unresolved-call* default (`None`). It does **not** re-derive the **resolved** `callee_decl` on a warm JIR load.

## Problem

`callee_decl` (and `call_kind`) are stamped by `TypeCheckPass.exit_func_call` via `ability_of_symbol(resolve_expr_symbol(nd.target))`. By design they are **never serialized to JIR** (same as `Expr.type`); the field comment in `unitree.jac` states they are *"restamped on load"* by inference.

But on a **warm build** the cached module is reused without re-running the inference schedule, so the resolved declaration is never restamped — the call node keeps the `None` default from #6665. The native and client (JS) backends read the resolved signature (ordered parameters, default-arg exprs) off `callee_decl` to lower **keyword arguments** to ordered positional args, and `esast_gen_pass` has **no re-resolution fallback**. So:

- **Cold build** → `callee_decl` resolved → kwarg call lowers to correct positional order.
- **Warm build** → `callee_decl is None` → kwarg lowering has nothing to resolve against, so a call with keyword args can lower differently (or incorrectly) vs the cold build.

Calls **without** kwargs are unaffected (that is the common case #6665 already makes safe).

## Repro sketch

A `cl def` callee with parameters that are passed by keyword (especially out of declaration order), e.g.:

```jac
cl def make(a: int, b: int) -> int { return a - b; }
cl def use() -> int { return make(b=1, a=10); }   # expect 9
```

Build cold vs warm (warm = JIR cache populated, source unchanged but forced through a real recompile) and diff the emitted JS for `use` — the positional ordering of the lowered `make(...)` call should be identical and correct in both.

## Possible directions

1. Re-run the inference schedule (the `callee_decl` restamp) whenever a JIR-loaded module is about to be codegen'd, so warm == cold. (Matches the field's documented "restamped on load" contract — the gap is that the restamp doesn't fire on the cached-reuse path.)
2. Or serialize enough resolution info to reconstruct the resolved signature on load (heavier; departs from the "never serialized" design).
3. Or give `esast_gen_pass` a re-resolution fallback when `callee_decl is None` but the target resolves to a known ability.

## Acceptance

Cold and warm builds emit **identical** JS/native for calls that use keyword arguments (and for default-arg filling), with no dependence on `.jac` cache state.

chess10kp:
## Summary

Closes #6666. Follow-up to #6665.

`FuncCall.callee_decl` is a checker-stamped, never-serialized fact (reset to `None` on JIR reload per #6665). The client (`cl`/JS) backend reads it (via `esast_gen_pass`, which has no re-resolution fallback) to lower keyword arguments to ordered positional args — especially for method calls, whose resolved target symbol is itself an inference-time fact JIR does not carry.

The compiler's dependency re-inference sweep — the mechanism that restamps `callee_decl` for JIR-loaded modules so warm builds match cold — previously covered **only native** deps:

```jac
if _is_na(dep_path) {
    dep_mod._inference_started = True;
    ...
}
```

So a **warm-loaded client dependency** never got restamped: its `callee_decl` stayed `None`, and a kwarg method call lowered as a packed object (`x.m({b: 1, a: 10})`) instead of positionally (`x.m(10, 1)`), diverging from the cold build.

### Fix

Extend the sweep to also re-infer client modules (the issue's Direction #1, matching the field's documented "restamped on load" contract):

```jac
if _is_na(dep_path) or _is_cl(dep_path) {
    dep_mod._inference_started = True;
    ...
}
```

### Verification

Cold and warm builds now emit identical JS for kwarg (and default-arg) calls for client code. New regression test `warm client dep restamps callee_decl for kwarg method calls` in `jac/tests/language/test_cli.jac`:
- cold-compiles a `cl` lib with `x.m(b=1, a=10)` (expect `9`),
- round-trips it through JIR (resetting `callee_decl` to `None`),
- parks the untyped copy in the hub and compiles a client entry (triggering the dep-sweep),
- asserts `callee_decl` is restamped and the emitted JS contains `x.m(10, 1)` and not the packed `x.m({` form.

This test **fails on `main`** (`AssertionError: warm client dep callee_decl not restamped by dep-sweep (#6666)`) and **passes with this PR**.

Affected suites all green: `test_js_generation` (82), `test_esast_gen_pass` (40), the sibling `#6665` test, and the new `#6666` regression.

---

<details>
<summary><b>Original issue #6666</b></summary>


chess10kp:

PR #7183, opened 2026-07-05
fix(checker): widen literal types inferred for mutable declarations

### LINKED ISSUE #7159: Type checker: Literal types inferred from initializers do not widen for mutable variables/containers
## Root cause

When a mutable variable, attribute, or container element type is inferred from a literal initializer, the checker keeps the `Literal[...]` type instead of widening to the base type (`str`, `int`, `bool`). Any subsequent store of a different value then fails. Standard checkers (mypy, pyright) widen literals in these positions and reserve `Literal` for explicit annotations, `Final`, and expression-position inference.

## Repro (should pass, currently fails)

```jac
def make_row -> None {
    row = ["."] * 3;
    row[0] = "B";
}
```

```
$ jac check repro.jac --nowarn
✖ Error: error[E1001]: Cannot assign Literal["B"] to Literal["."]
```

## Proper fix

Widen literal types to their base type when inferring the **declared** type of:
- mutable local/global variables (`x = ""` → `str`, not `Literal[""]`)
- object attributes inferred from initializers
- container element types in list/dict/set literal inference (`["."] * 3` → `list[str]`)

Keep `Literal` for explicitly annotated types, `Final` bindings, and the type of the expression itself (so equality/identity narrowing still works).

## Impact

164 instances in a full-`jac check` sweep (audit that produced #7157). Real sites: `jac/jaclang/compiler/passes/native/na_ir_gen_pass.impl/builtins.impl.jac:100` (`global_fmt.linkage = "private"` fails because `linkage` was inferred as `Literal[""]` from its `= ""` default), `jac/jaclang/byllm/examples/core_examples/level_genarator.jac` (game grid inferred as `list[list[Literal["."]]]`, no tile can ever be placed).

marsninja:
## Summary

Fixes #7159.

When a mutable variable or container element type is inferred from a literal initializer, the type checker kept the `Literal[...]` type instead of widening to the base type (`str`/`int`/`bool`). Any subsequent store of a different value then failed with `E1001`, e.g.:

```jac
def make_row -> None {
    row = ["."] * 3;
    row[0] = "B";   // was: error[E1001]: Cannot assign Literal["B"] to Literal["."]
}
```

Standard checkers (mypy, pyright) widen literals in these mutable-declaration positions and reserve `Literal` for explicit annotations, `Final`, and expression-position inference.

## Approach

Reuses the existing `_widen_literal_for_declaration` primitive rather than adding a parallel mechanism. The primitive already widened top-level literals and unions for the declared type of local/global variables (and the reassignment path); it did not descend into container element types. It now recurses into container `type_args` (list/dict/set/tuple), so `["."] * 3` declares `list[str]`, a nested grid declares `list[list[str]]`, and dict/set literals widen their element types. The recursion returns the original object unchanged (and skips the `specialize_generics` allocation) when no element actually widens.

Scalar widening for class-body attribute defaults (e.g. `linkage = ""`) already flowed through the existing declaration path, so this is a focused, single-function change. Widening only affects the **declared** type; the initializer's own expression type keeps its `Literal`, and `assign_type` already allows literal to base-class assignment, so equality/identity narrowing is unaffected and no spurious errors are introduced.

## Tests

- New fixture `checker_bug_literal_widening.jac` and regression test `bug_literal_widening_for_mutable_declarations` in `test_checker_bugs.jac`, covering list literals, `[x] * n`, nested grids, dict values, and a narrowing-preservation guard (4 errors before the fix, 0 after).
- Full checker suite green: `test_checker_bugs`, `test_checker_pass`, `test_checker_gaps`, `test_checker_shortcomings`, `test_checker_production`, `test_new_diagnostics`, `test_typevar` (no regressions).


marsninja:
Thanks for the review. Addressed both points in d43aaca:

**Attribute widening paths lack test coverage (P2):** Good catch, and it led me to tighten the scope. On investigation the `AtomTrailer` (`self.attr = ...`) and `HasVar` (`has attr = v`) widening branches were no-ops: class-body attribute defaults like `linkage = ""` already widen through the existing `Assignment` declared-type path, an unannotated `has x = v` is a syntax error in Jac (so that branch is unreachable), and the `self.attr = ...` instance case is short-circuited upstream by the cached `node_.type`. Rather than add tests for unreachable code, I reverted those two branches. The load-bearing change is the container-element recursion in `_widen_literal_for_declaration`, which the new fixture exercises directly (lists, `[x] * n`, nested grids, dict values).

**Unnecessary `specialize_generics` call (P2):** Added a short-circuit. The recursion returns the original object unchanged when no element widens, and `specialize_generics` is now only called when at least one arg actually changed, so already non-literal containers (e.g. `list[str]`) avoid the extra allocation.

PR #7184, opened 2026-07-05
fix(typecheck): dispatch subscript assignment through __setitem__ (#7160)

### LINKED ISSUE #7160: Type checker: bytearray slice assignment rejected — missing __setitem__ slice overload in builtin stubs
## Root cause

The builtin stubs model `bytearray.__setitem__` (and `MutableSequence.__setitem__` generally) with a single signature requiring the RHS to be the container's own type. At runtime, slice assignment accepts any buffer / iterable of ints: `bytearray[slice] = bytes` is valid Python.

## Repro (should pass, currently fails)

```jac
def make_header -> None {
    ident = bytearray(16);
    ident[0:4] = b"\x7fELF";
}
```

```
$ jac check repro.jac --nowarn
✖ Error: error[E1001]: Cannot assign bytes to bytearray
```

## Proper fix

Give `__setitem__` proper overloads in the builtin stubs, following typeshed:
- `bytearray.__setitem__(i: int, v: int)` and `bytearray.__setitem__(s: slice, v: Iterable[int] | Buffer)`
- `list[T].__setitem__(i: int, v: T)` and `list[T].__setitem__(s: slice, v: Iterable[T])`

Audit sibling builtins (`memoryview`, `array`) for the same missing slice-overload shape, and make sure the subscript-assignment checker actually dispatches on the index type (int vs slice) rather than assuming element assignment.

## Impact

142 instances in a full-`jac check` sweep (audit that produced #7157), essentially all in the native linkers which build binary images via slice assignment: `jac/jaclang/compiler/passes/native/impl/elf_linker.impl.jac` (e.g. line 1098, `ident[0:4] = ELFMAG`) and `macho_linker.impl.jac`. Fixing this plus the vendored-llvm stub issue would unlock the whole native-pass family from `.jacignore`.

marsninja:
## Summary

Fixes #7160. Range-slice assignment (`container[a:b] = value`) was type-checked by comparing the RHS against the `__getitem__` **read** type of the target (for a range slice, the container type) rather than resolving `__setitem__`. This rejected valid slice writes whose RHS is legitimately broader than the container, most visibly `bytearray[a:b] = bytes`:

```jac
def make_header -> None {
    ident = bytearray(16);
    ident[0:4] = b"\x7fELF";   // error[E1001]: Cannot assign bytes to bytearray
}
```

This pattern drives the native ELF/Mach-O linkers (`elf_linker.impl.jac`, `macho_linker.impl.jac`), and was one of the blockers keeping the native-pass family in `.jacignore`.

## Root cause

For a slice target, the checker derived the LHS type via `get_type_of_expression(target)` (a `__getitem__` read, which for a range slice is the container type) and then required the RHS to be assignable to it. The correct model is a `__setitem__(index, value)` call, where slice overloads accept any buffer / iterable of the element type. The vendored typeshed stubs already carry the right `__setitem__` overloads for `bytearray`, `list`, `memoryview`, and `array`; only the dispatch was missing.

## Change

- A slice expression (`a[0:4]`) now types as the builtin `slice`.
- When an assignment target is a **range slice**, the checker probes `__setitem__` through the existing overload-resolution machinery (`get_type_of_magic_method_call`) and only **suppresses** a would-be `E1001` when the write is genuinely valid.

The probe can only remove a false positive (never add a diagnostic). It is scoped to range slices: plain element-index writes (`items[0] = x`) already type-check correctly via the existing path and keep their current behavior, so the change stays focused on the reported slice bug.

## Tests

`jac/jaclang/compiler/tests/test_subscript_assignment.jac`:
- valid `bytearray` slice write and `list` slice write from an iterable now type-check;
- over-suppression guards: non-buffer / non-iterable RHS still error;
- augmented slice writes (`b[0:2] += ...`) documented as passing via the aug path.

## Review notes

Addressed the Greptile review: the `IndexSlice` case now returns `UnknownType` explicitly for a non-range node with no resolvable start; the duplicated plain-index guard was removed by scoping the probe to range slices; and an augmented-subscript test was added.


marsninja:
Thanks for the review. Addressed the feedback:

- **IndexSlice non-range fall-through** (P2): the `IndexSlice` case now branches on `is_range` first and returns `UnknownType` explicitly for a non-range node with no resolvable start, so a structurally-odd plain-index node can no longer advertise type `slice`.
- **Duplicated plain-index guard** (P2): the `__setitem__` probe is now scoped to range slices, so it passes the `IndexSlice` node directly and the duplicated `not is_range and slices and slices[0].start is not None` guard is gone. Element-index writes already type-check via the existing path, so limiting the probe to slices keeps the fix focused on the reported bug (and avoids re-widening interactions on narrowed element writes).
- **Augmented subscript assignment untested** (P2): added a test asserting `b[0:2] += ...` / `xs[0:2] += ...` pass via the augmented-operator path (which the probe intentionally does not touch).

PR #7191, opened 2026-07-06
fix(error-reporting): enhance client-side error reporting to include browser overlay

### LINKED ISSUE #7190: Both Dev and Prod Modes Silently Swallow Some Runtime JS Errors
## Summary

Certain categories of JavaScript errors cause both **development mode** (`jac start -d`) and **production mode** (`jac start`) to show a **blank white page** with **no error output in the terminal**, **no error overlay in the browser UI**, and **no error boundary message**. The only place the error appears is buried inside the browser's DevTools console — which developers must manually open to discover.

The error categories affected differ between modes:

| Error type | `jac start -d` (dev) | `jac start` (prod) |
|---|---|---|
| Named export mismatch (`SyntaxError`) | ✗ Silent | ✓ Reported (Rollup catches at build time) |
| Class constructor without `new` (`TypeError`) | ✗ Silent | ✗ Silent |
| Top-level module evaluation errors | ✗ Silent | ✗ Silent |

Production mode correctly catches *static* errors (named export mismatches) at build time via Rollup. However, *runtime* errors — those that only appear when JavaScript actually executes in the browser — are silently swallowed in **both** modes.

### 1. Named Export Mismatch — `SyntaxError` (dev mode only; prod mode catches this at build time)

Occurs when a file imports a name that does not exist in the imported module's exports.

**Common triggers in Jac:**
- Using `glob` (private by default) when `glob:pub` is needed for a cross-file import
- Importing a named export from an npm package that doesn't expose it (e.g. CommonJS packages pre-bundled by Vite/esbuild that don't statically expose all exports)

**In dev mode** (`jac start -d`): error is silent — blank page, nothing in terminal.

**In prod mode** (`jac start`): Rollup catches this at build time and prints a formatted diagnostic to the terminal. ✓

**Example browser console output (dev mode, only visible in DevTools):**
```
Uncaught SyntaxError: The requested module '/compiled/lib/mockData.js'
does not provide an export named 'PRODUCTS' (at HomePage.cl.jac:7:2)
```
```
Uncaught SyntaxError: The requested module
'/.vite/deps/@hugeicons_core-free-icons.js?v=08534a5f'
does not provide an export named 'ShieldCheckmarkIcon' (at HomePage.cl.jac:6:2)
```

### 2. Class Constructor Called Without `new` — `TypeError` (both dev and prod modes)

Occurs when an `obj` instance is constructed in top-level module code in a way that the compiled JavaScript calls the class constructor as a plain function instead of using `new`.

This is a **runtime error** — Rollup's static analysis cannot detect it at build time, so it passes through `vite build` silently. The error only manifests when the browser executes the bundle.

**Example browser console output (only visible in DevTools — same in both dev and prod):**
```
Uncaught TypeError: Class constructor En cannot be invoked without 'new'
  at mockData.js:5:11
```
```
Uncaught TypeError: Class constructor Product cannot be invoked without 'new'
  at ProductCard.cl.jac:4
```

## Evidence  from JacHammer

Image

## Expected Behavior

Errors like these should be reported in **at least one** of the three developer-facing channels that dev mode already uses for other errors:

1. **Terminal output** — same as `jac start` (prod mode) does at build time
2. **Browser error overlay** — the red overlay that already appears for Jac compilation errors
3. **Error boundary** — the React error boundary that catches runtime render errors

At minimum, the terminal should print something equivalent to what `jac start` (prod) shows:
```
⚠ Failed to build client bundle

┌ JAC_CLIENT_999: Build Failed ─────────────────────────────────────┐
│  Client bundle build failed.                                       │
└────────────────────────────────────────────────────────────────────┘

--- Raw Error ---
compiled/pages/HomePage.js (7:1): "PRODUCTS" is not exported by
"compiled/lib/mockData.js", imported by "compiled/pages/HomePage.js".
--- End Raw Error ---
```

## Impact

- **Human developers** see a blank white page with no actionable feedback in both dev and prod modes for runtime errors. They must remember to open DevTools and check the Console tab.
- **AI coding agents** (e.g., `jac-code`) that monitor terminal output and browser overlays to diagnose errors have no visibility into this class of failure in either mode. These are among the most common errors when building new Jac client code.
- **Error boundaries and overlays** — the existing mechanisms (React error boundary, Jac error overlay, `__build_status` endpoint) all remain clean/OK even during these failures, giving a false signal that the app is healthy.
- **Prod mode false confidence** — because `vite build` exits with code 0 and the server starts successfully, there is no terminal-level warning that the app will fail in the browser. A developer who skips opening DevTools will not know anything is wrong.

SahanUday:
## Problem

When a Jac app has certain JavaScript errors, the browser shows a **blank white page**  with no output in the terminal, no overlay, and no error boundary message. The only place the error appears is buried in the browser's DevTools console — invisible to both developers and AI coding agents monitoring the terminal.

Two error types were affected:

- **`SyntaxError`** — importing a name that isn't exported (e.g. `glob` instead of `glob:pub`,  or a named import from a CJS npm package)
- **`TypeError`** — class constructor called without `new` in top-level module data  (a known Jac compiler output pattern)

Close #7190 

## Root Cause

It's a **script ordering problem**. The error handler (the code responsible for sending errors to the terminal) was being installed *after* the app module had already loaded and errored out.

**Dev mode** — three scripts ran in this order:
```
1. /@vite/client ← Vite's own script
2. _entry.js ← app loads here → ERROR FIRES HERE 💥
3. error handler ← installed here → TOO LATE ❌
```

**Prod mode** — inside the single `client.js` bundle:
```
line 431k:  function installErrorHandler() { ... }   ← defined
line 443k:  Product({...}), Product({...}), ...       ← ERROR FIRES HERE 💥
line 468k:  installErrorHandler()                     ← TOO LATE ❌
```

### Fix 1 — Dev: Load the app module inside the handler, not before it
`vite_bundler.impl.jac` → `_ensure_dev_index_html()`

Removed the standalone `<script src="/compiled/_entry.js">` tag. Instead, the entry  module is now loaded via `import("/compiled/_entry.js").catch(report)` **inside** the  same script block that installs the handlers. The handler is always ready before the  app starts loading.

Both `SyntaxError` (module linking) and `TypeError` (top-level evaluation) become  rejected promises that `.catch()` intercepts, forwarding the error through the existing  `jac:client-error` WebSocket channel to the terminal.

### Fix 2 — Dev: Show Vite's built-in red error overlay on module errors
`vite_bundler.impl.jac` → `jacErrorReporter()` plugin

After logging to the terminal, the plugin now also calls:
```javascript
server.ws.send({ type: "error", err: { message, stack, id, plugin } });
```

### Fix 3 — Prod: Add an early error handler before the bundle loads 
`serving.impl.jac → JacClient.render_page()`

Added a small regular `<script>` (non-module) before the `<script type="module">` bundle tag. Regular scripts execute synchronously; module scripts are deferred. This guarantees the handler is active before the bundle evaluates.

The handler POSTs to `/cl/__error__` (the same endpoint the bundle uses) and includes a guard: if the bundle loads successfully and installs its own handler (`window.__jacErrorHandlersInstalled`), the early handler steps aside — no duplicate reports.

## Evidence after the fixes

** Terminal output: **
Screenshot 2026-07-05 000531

** Browser overlay: **
Screenshot 2026-07-05 000600


MalithaPrabhashana:
@SahanUday can we add few test cases to this?

MalithaPrabhashana (approved):
Looks good!

PR #7193, opened 2026-07-06
fix(scale): ship the app's modules precompiled in the source bundle (#7179)

### LINKED ISSUE #7179: scale: #7144 jir seeds only ship in the [dev] overlay; stable-channel deploys still cold-compile the app in every pod
Follow-up to #7138 after benchmarking #7144 on a real EKS deploy. #7144's boot optimizations only partially reach production-shaped deploys: the .jir seeds ride the `[dev]` overlay tarball, so a stable-channel deploy (every jacBuilder user app, and generally any deploy whose project has no `[dev]`) ships no seeds, and every pod still cold-compiles the entire app before binding the server.

## Measured (EKS, 5-service sv-import app, intent-less jac.toml, cold namespace, post-#7140/#7144 operator)

| Phase | pre-#7144 baseline | this run | notes |
|---|---:|---:|---|
| host prep (deploy fired -> app pods created) | ~8 min | ~15.8 min | polluted by two concurrent operator-fleet rollouts; the deferred items (binary re-download + repack, force client rebuild, full re-seed) still dominate either way |
| pod boot (pods created -> all Ready) | ~8 min | **~4-6 min** (4/5 in 4m14s, last gateway 6m) | improvement is real but comes only from the apt/bun removal; compile + pip remain |
| crash-loops | 1 restart/pod + pymongo crash-loop without workaround | **0 crash-loops** | #7140's baked database intent verified end to end on an intent-less app |
| total to URL 200 | ~17 min | ~22 min (churn-polluted) | |

## Why the boot is still minutes

The pods have no seeds to adopt. Verified on the live pods and in source:

- Init logs contain no adoption lines; `/app/_precompiled` does not exist in the shipped bundle.
- The seed machinery lives in the overlay packer (`injector/binary.jac`, `JIR_SEED_SUBDIR = "jir-seed"`, the re-homing script at ~:260). Seeds are the jaclang tree's modules, produced when a `[dev]` overlay is packed. The e2e that measured "adopted 244, compile -> ~0" runs a dev-overlay deploy of a small fixture on kind.
- A stable-channel pod already had jaclang internals precompiled inside the binary payload, so overlay seeds add nothing there, and the **user app's own modules** are exactly what still compiles cold, once per pod, before uvicorn binds.

So OPT-0 solved #7138's item 12 (dev-overlay jaclang recompile) but not item 6 (app-module precompile), which is the one user-facing deploys need.

## Ask

Extend seeding to the source bundle: during `pack_source`, run the existing precompiler (`utils/precompile_bytecode.jac`) over the project's modules with the already-downloaded pod toolchain binary, and pack `_precompiled/cpython-XY/...` (or reuse the jir-seed manifest mechanism) into the source bundle. The pod-side loader already walks up from the module dir and validates content-keyed .jir (`jac0core/impl/compiler.impl.jac:409-447`), so no pod-side changes should be needed. That removes the remaining per-pod compile for all N pods on every channel.

Remaining boot cost after that is the per-pod pip closure (#7138 item 8, shared wheel cache on the already-mounted RWX PVC) and the mongo TCP-vs-auth-ready race (#7138 item 9; we still saw one restart on a pod that booted into the window).

Also worth noting for expectation-setting: the 134s e2e headline is kind + local registry + tiny fixture + dev overlay. On EKS with EFS seeding and a real fullstack app, prep alone still runs many minutes because the deferred host-side items (binary cache, seed skip-if-present, client-build fingerprint) have not landed yet.

MusabMahmoodh:
## What

Fixes #7179: `#7144`'s `.jir` seeds ride the `[dev]` overlay tarball only, so **stable-channel deploys (every production/jacBuilder app) shipped no seeds** - each pod cold-compiled the entire app before binding the server (~minutes per pod on EKS, repeated on every pod churn).

## How

`pack_source` now:
1. stages a copy of the project (same exclusion rules as packing),
2. precompiles it with the **already-downloaded pod binary** via the existing `utils/precompile_bytecode.jac` (`.jir` is version-locked, so the driver's own jac cannot produce pod-valid artifacts),
3. packs `_precompiled/cpython-XY/*.jir` at the bundle root.

The core loader already walks up from each module and **content-validates** `_precompiled/` entries ([compiler.impl.jac](../blob/main/jac/jaclang/jac0core/impl/compiler.impl.jac)), so:
- **no pod-side changes** - pods load instead of compiling, on **stable and dev alike**
- stale/mismatched seeds are ignored safely (content-keyed)

Flat app projects are staged with a temporary `__init__.jac` package marker for the precompiler's discovery; the marker's own `.jir` is not packed.

**Best-effort end to end**: a missing/unrunnable binary or failed precompile logs a warning and ships a source-only bundle - pods compile exactly as before.

## Wiring

`BinaryInjector.prepare` downloads the pod binary **before** `super.prepare()` and exposes it; `PvcInjector.prepare` threads it into `pack_source`. Plain PVC (image-based) deploys are unchanged.

## Test

- `test_app_precompile.jac`: real-binary test proves `_precompiled/cpython-XY/svc.jir` lands in the bundle; graceful-skip tests for no-binary and unrunnable-binary paths.
- Injector suites pass (28/28).
- The k8s e2e's per-pod compile counts (from #7144) verify pod-side loading; the honest production acceptance is re-running the #7179 EKS benchmark, where pod boot should drop by the app-compile time on all N pods.

## Notes for reviewers

This deliberately reuses the `_precompiled/` sealed-bundle convention rather than #7144's jir-seed manifest: it is content-keyed (no path re-homing needed) and already supported by the loader. Remaining #7179/#7138 items (per-pod pip closure, mongo auth-ready race, host prep caching) are out of scope here.


MusabMahmoodh:
## EKS validation (real-cluster benchmark, #7179's scenario)

Ran this PR's driver against **`this_is_jac`** (4 services + gateway, intent-less jac.toml) on the **`jaseci-cluster` EKS** (us-east-2), **stable channel**, `efs-sc` RWX bundle storage, fresh namespace.

### The fix works on production-shaped deploys

- Host: **`Bundled 76 precompiled app-module artifacts`** - `pack_source` precompiled the app with the downloaded stable pod binary and shipped `_precompiled/cpython-314/` in the source bundle (234 KB).
- Pods: `_precompiled/` present at `/app`, and **zero `Compiling` lines** in any app container - modules load from the shipped `.jir`.
- Deploy returned `success=True` in 770s end-to-end; all 5 workloads Ready.

### Per-pod boot breakdown (guestbook, representative)

| Step | Duration |
|---|---:|
| wait-for-mongodb / redis | 1s |
| untar 170 MB toolchain from EFS | 8s |
| jac-bootstrap (dependency install) | **3m57s** |
| **main start -> Ready (app loads precompiled)** | **18s** |

The compile that #7179 measured in minutes-per-pod is gone; app start is now seconds. The dominant remaining boot cost is exactly #7179's predicted next item: the **per-pod dependency closure** - pip packages plus **504 npm packages installed in every pod**, including service pods that never serve a client. That npm-closure-on-services observation is a cheap follow-up win (skip npm deps for non-gateway pods, or ship `node_modules`/client dist in the bundle).

### Caveats
- The bench host's client build was skipped (`jac` not on the subprocess PATH), so the gateway built the client on boot - matching the #7179 baseline conditions.
- Absolute end-to-end times on this run aren't directly comparable to #7179's (different app, one warm mongo/redis reuse); the load-bearing numbers are the **0-compile pods** and the **18s app start**.

PR #7196, opened 2026-07-06
fix(scale): resolve the driver's own binary via JAC_EXECUTABLE for host builds

MusabMahmoodh:
## What

The deploy driver's host-side client and admin-console builds invoked a bare `jac` and relied on it being on PATH. When it isn't (operator boxes, CI images, jacBuilder pods), the builds **silently fell back to building on the gateway pod at boot** - adding a minute-plus of boot time and a `bun` memory spike, with only a log line to notice.

Seen live on the EKS benchmark for #7179: the gateway rebuilt the client on boot purely because the driver binary lived at `~/bin/jac` off-PATH.

## How

The launcher already exports its own absolute path as `JAC_EXECUTABLE`. Both self-invocation sites now use `jac_executable()`: `JAC_EXECUTABLE` first, `shutil.which("jac")` fallback, bare `jac` last.

## Achieved

Host builds run wherever the driver runs - the gateway pod stops paying the on-boot client build (~1 min + the OOM-risk bun spike) whenever the host could have done it.

## Test

Covered by the existing host-build paths in the k8s e2e; the change is a two-site argv substitution with a total fallback chain, so no new unit test.


kugesan1105:
Jac's path isn't fixed. `install.sh` defaults to `~/.local/bin`, but that's a default — it varies by installer, by `$HOME`, by container user (root → `/root/.local/bin`), 

when `jac` is running, it resolved its own path via `/proc/self/exe` at boot and published it as `JAC_EXECUTABLE`


but @Thamirawaran we might need to check why installed paths are differnt

PR #7198, opened 2026-07-06
perf(scale): skip re-seeding content-addressed bundles; deterministic archives

MusabMahmoodh:
## What

Two entangled fixes for the #7179 finding that **a warm redeploy is no faster than a cold deploy**:

1. **Seed skip-if-present**: the PVC seeder re-uploaded every artifact (~170 MB toolchain + app bundle) on every deploy. Object keys are `sha256(bytes)`, so if the file already exists on the bundle PVC it *is* this payload - the seeder now `test -f`s before uploading and skips hits.

2. **Deterministic archives** - the bug that would have made (1) useless: gzip stamps wall-clock time into its header, and tar members carried disk mtimes/uid/gid, so identical content got a fresh digest on every pack. Observed live on the EKS benchmark: two byte-identical deploys produced two different object keys. All bundle archives now write a zeroed gzip timestamp and normalized members.

## Achieved

Warm redeploys stop re-shipping ~170 MB to EFS; combined with the sibling binary-cache PR, the measured warm redeploy dropped from ~24 min (baseline in #7179) to ~5 min, with more to come once this determinism fix makes the skip actually fire.

## Test

One unit test: `pack_binary_bytes` is byte-identical across calls, and a source bundle's content-address survives an mtime change. The skip path itself is exercised by the k8s e2e (second deploy to the same namespace).


MusabMahmoodh:
Closing: this PR's full content (deterministic archives via open_det_tar/det_tar_member, the seed skip-if-present probe, and the determinism test) was merged to main as part of the #7193 integration - verified on main at 071be0503. The mechanism is also field-verified: the warm bench observed 'toolchain skipped' with the 170MB re-upload gone.

Issue #7138, opened 2026-07-03
scale: cold k8s deploy pays uncached prep + N-times first-boot compile; ~17 min for a 5-service app


udithishanka:
Companion to jaseci-labs/jacBuilder#618 (measured breakdown of a real deploy there). This half covers what jac-scale itself can fix in the #7084 binary-ship model's cold-deploy path.

Measured: a 5-service `sv import` app deployed cold to EKS takes **~17 min** end to end. The same model applies manifests for a 6-service warm fleet in 2m43s (jacBuilder deploy-dev run 28661896401), and the pre-#7084 path (pods pip-installing wheels, work overlapped) handled comparable apps in ~3 min. The cold-path costs:

## 1. Prep is serial and precedes everything

`deploy()` provisions databases early, then does binary fetch, host client build, bundle pack, and PVC seed **serially** before any app manifest is applied. On an operator-style caller (a pod driving KubernetesTarget in-process with fractional CPU) that prep alone is ~8 min; the DB pods sit idle the whole time. The prep does not depend on the DBs -- it could run concurrently with provisioning, and the binary fetch could be cached across deploys instead of re-downloaded.

## 2. Every pod compiles the whole app on first boot

Each of the N service pods runs the full bootstrap: apt for system deps, binary install, `jac install`, then a **first-boot compile of the entire app, independently in every pod** (~8 min for 5 pods to all reach ready, and it repeats on every pod churn until the rt cache PVC warms). The bundle already ships the source; it could ship the **precompiled JIR** too, produced once on the host during prep -- #7118's content-validated precompile cache is exactly this shape and could be reused, so pods validate instead of recompile.

## 3. Bundle is re-seeded in full every deploy

The seeder ships binary + code + client dist to the bundle PVC wholesale each deploy. Content-addressing the bundle (seed only changed entries; the ~200MB binary almost never changes) would cut both seed time and EFS load.

Happy to provide the full timing logs from the measured deploy.

udithishanka:
[...]

PR #7204, opened 2026-07-06
perf(scale): share pip/bun caches across pods via the RWX bundle PVC

MusabMahmoodh:
## What

After the compile fixes, the dominant remaining pod-boot cost (#7138 item 8) is dependency resolution from the network: every pod pip-installs its full closure, and the gateway additionally bun-installs the client packages - **~4 min per pod measured on the EKS benchmark**, re-paid on every boot, restart, and scale-up.

## How

The bundle PVC (RWX, already mounted read-only for unpacking) is now also mounted read-write in the `jac-bootstrap` init container, and `PIP_CACHE_DIR` / `BUN_INSTALL_CACHE_DIR` point into `<pvc>/.cache/`. The first pod warms the cache; every sibling, restart, and redeploy hits it. Wrapped in an `if mkdir` guard: an unwritable PVC silently degrades to today's behavior.

No new infrastructure - it reuses the volume every scale deploy already has.

## Achieved

Downloads collapse to cache reads for all pods after the first (and across redeploys, since the PVC outlives pods). pip still resolves versions, so pinned-dependency correctness is unchanged - only the network fetch disappears.

## Test

One unit test: the bootstrap exports both cache vars behind the mkdir guard, before `jac install`, and mounts the bundle volume.


jac/jaclang/scale/tests/microservices/test_shared_dep_cache.jac
```diff
@@ -0,0 +1,26 @@
...
+    # bun's cache is deliberately NOT shared: thousands of tiny files are
+    # pathological on EFS/NFS; it stays on the pod-local disk.
+    assert "BUN_INSTALL_CACHE_DIR" not in script;
+    # mkdir-guarded (unwritable PVC degrades to plain installs), size-bounded.
+    assert script.index("if mkdir") < script.index("PIP_CACHE_DIR");
+    assert "du -sm" in script;
+    assert script.index("PIP_CACHE_DIR") < script.index("jac install");
+    mounts = [m["name"] for m in ctr["volumeMounts"]];
+    assert BUNDLE_VOLUME_NAME in mounts;
+}
```
kugesan1105, line 26: @Thamirawaran  these kind of sting test needs to be improved with real world e2e test.
instaed of manual testing

kugesan1105:
@Thamirawaran these kind of sting test needs to be improved with real world e2e test.
instaed of manual testing

Issue #6905, opened 2026-06-23
Opt-in ownership & borrow checking for Jac: stratified, lifetime-free ownership over the managed (RC/GC) floor — design + phased plan


marsninja:
## Summary

Add **opt-in, ownership-checked memory management** to Jac, layered *on top of* the existing managed (RC/GC) floor — so a developer can selectively get Rust-style static guarantees (no use-after-move, no aliased mutation, race-free message passing) and native-backend performance (elided refcount traffic, arena allocation) **without** giving up Jac's productivity defaults and **without** lifetime annotations.

Ownership is expressed **at use sites** — on bindings (`own`), in reference expressions (`&`/`&mut`), and in parameter/return types — like Rust, but with one irreducible difference: **Jac's default stays managed**, so the `own` annotation *is* the opt-in.

Framing: ownership annotations are the **static evidence that lets specific values step off the managed floor, value by value** — with GC/RC remaining underneath for everything not proven, and for the entire Object-Spatial (graph) world, which is never borrow-checked.

Full design doc accompanies this work at `docs/proposals/opt-in-ownership-borrow-checking.md` (not yet on `main`). This issue is the self-contained tracking RFC.

## Why now / motivation

- **Native correctness gap.** The native backend uses automatic reference counting (`jaclang/compiler/passes/native/na_ir_gen_pass.impl/refcount.impl.jac`) with **no cycle collector — cyclic native objects leak.**
- **Native performance.** Refcount retain/release traffic is overhead that a static ownership proof can elide; hot loops want arena allocation.
- **Static bug-catching.** Today there is no static protection against use-after-move / aliased mutation for value data.
- **Concurrency/distribution safety.** `flow`/`thread_run` and the distributed graph have no static data-race protection for the payloads they pass.

## The key enabler (why this is tractable)

Jac already has the two hard prerequisites:

1. **A static ownership lattice** — `jaclang/compiler/ownership.jac` already classifies every native expression `OWNED`/`BORROWED`/`STATIC`/`UNKNOWN` to drive RC. Parameters already default to *borrowed*; a rebound parameter is already *promoted to owner* (a move rule in all but name).
2. **A flow-sensitive analysis engine** — a CFG (`passes/main/cfg_build_pass.jac`) plus backward-walk type narrowing with assignment barriers (`type_evaluator.jac`). This is structurally the same machinery a move/borrow checker needs (per-binding state instead of per-binding type).

[...]

PR #7208, opened 2026-07-06
feat(ownership): Phase 0/1 — RC cycle collector + diagnostics-only borrow checker (#6905)

chess10kp:
## Summary

Implements Phases 0 and 1 of the opt-in ownership & borrow checking system described in #6905.

**Phase 0 — Native RC cycle collector**
- Synchronous Bacon–Rajan cycle collector for the native ARC runtime (`cycle_collector.impl.jac`, `trace_fn.impl.jac`)
- Per-type `__rc_trace_<T>` pointer-field enumeration emitted from existing field-layout metadata
- Threshold-triggered trial deletion (mark-gray → scan → collect-white)
- `JAC_GC_CYCLES` env gate + wasm `realloc` adapter
- Test: `cycle_collect_storm.na.jac` builds a native cycle and asserts reclamation

**Phase 1 — Diagnostics-only ownership checker**
- New keywords: `own`, `&`, `&mut` at use sites; `OwnershipKind` SubTag on annotations and `UnaryExpr`
- `OwnershipCheckPass` wired into `get_type_check_sched()`; engages only on annotated bindings
- Errors: `E1301` use-after-move, `E1302` conflicting `&mut`, `E1303` escaping/ambiguous borrow, `E1304` scope-outlives-owner, `E1306` borrow-escaped-through-return
- Borrow-typed and `own` parameters; passthrough-return exception
- Loop-depth move rule; membrane seal-in semantics
- All backends; **no codegen change**

Includes a preview of RC elision (`Symbol.is_proven_unique`) which is replaced by the full `RcElisionProofPass` in Phase 2.

## Test plan
- [x] `pytest jac/tests/compiler/passes/main/test_ownership_check_pass.jac` — all ownership fixtures pass
- [x] `pytest jac/tests/compiler/passes/native/` — cycle collector test passes
- [ ] Un-annotated code produces zero new diagnostics (regression)

Closes part of #6905. Followed by: Phase 2 RC-elision PR, Phase 3 regions/sendability/markers PR.


chess10kp:
Closing as superseded. The Phase 0/1 diagnostics-only borrow checker + RC cycle collector from this branch (originally #6905) already landed on `main` via #7149 and evolved substantially through follow-ups (#7398 packed 32-byte RC header, #7418/#7435 single-source ownership facts, #7453/#7457 nogc enforcement, #7485 `val`→`imm`, #7491/#7494 first-class regions, #7547). Merging `main` into this branch collapses its diff to release-notes only, so there is nothing left to land here.

Note for follow-up: this branch carried 4 late GC fixes (masked color store, per-edge Bacon–Rajan `scan_black`, deferred-free) against the original color-*byte* RC design. `main` has since replaced that with a packed 64-bit color *word*, so those patches don't port directly — if any of them fixed a real bug not covered by the rewrite, it should be reopened as a fresh targeted PR.

PR #7218, opened 2026-07-06
Remove dead code across the compiler, runtimes, and plugins (1/3)

ayushmk7:
# Remove dead code across the compiler, runtimes, and plugins (1/3)

**+111 / −3,390 across 158 files (net −3,279), 16 commits.** First of a three-PR stack (`prune/dead-code-1-source` → `-2-dedup` → `-3-unreferenced`); this one is pure source deletions — every removed symbol was verified unreachable before cutting.

## What's removed (by commit)

| Commit | Area | What it deletes |
|---|---|---|
| `refactor(langserve/lsp)` | langserve | The incremental sem-token machinery (`update_sem_tokens` + six `handle_*` methods + their exclusive utils + async `debounce`): only reachable from its own unit test — the live server rebuilds semantic tokens via full recompile on every `didChange`, and only `SEMANTIC_TOKENS_FULL` is registered |
| `refactor(native)` | compiler/native | The LLVM type read-back path: `TypeRef`/`TypeKind` + iterators + ffi regs, the eight circular `Type.from_llvm` constructors, their `binding` producers, the never-included `transforms.jac`, and unreferenced IRBuilder surface (`insert_element`, `branch_indirect`, `addrspacecast`, `fptoui`, `IndirectBranch`, `InsertElement`, `BlockAddress`, …). The na backend only emits IR — nothing reads types back |
| `refactor(runtimelib)` | runtimelib | Planner slimmed to its live `execute_path` fast path (the resolver rewrite superseded `execute_plan`/`plan_chain`/`plan_chain_ordered`); fourteen client-rewrite orphans (`copy_typescript_files`, `check_runtime_version`, vite plugin-option formatters, …); write-only `JacTestClient._current_user` |
| `refactor(jac0core)` ×2 | jac0core | The `filter_visitable_by` reflection cluster (its byllm caller was removed upstream in #6767 — the helper chain outlived it), write-only JirReader trailing-section fields, never-called `without_plugins`, dead mtp type-tag helpers, never-emitted diagnostic codes (E0011/E0033/E5070/E5071/W5072 + docs rows), and the orphaned raw-f-string token family (`RF_TEXT_*`, `F_FORMAT_TEXT`, `NEWLINE` + seven legacy `TOKEN_MAP` keys) across all three token registries |
[...]

PR #7219, opened 2026-07-06
Consolidate duplicated implementations behind single-source helpers (2/3)

ayushmk7:
# Consolidate duplicated implementations behind single-source helpers (2/3)

> **Stacked on #7218** — the diff below includes it until that merges; review the 13 commits from this branch's first `refactor(parser)` commit onward.

**This PR's own segment: +1,231 / −2,383 across 63 files (net −1,152), 13 commits.** Byte-identical (or provably equivalent) copies collapse into one owner; public surfaces and behavior are unchanged — several equivalences are proven at the artifact level (emitted LLVM IR, `@jac/runtime` exports).

## The consolidations

| Commit | Copies → owner |
|---|---|
| parser literal makers | 6 identical `make_*` constructors → `_make_literal` (binary/bool precedence levels deliberately left expanded: `grammar_extract_pass` reads those bodies) |
| `JacIntrospection.list_walkers/nodes/edges` | 3 identical `inspect.getmembers` filters → `_list_members_of(module, archetype_cls)` |
| compiler pass handlers | doc_ir switch/match, esast tuple/list, normalize if/elif/while + global/nonlocal, pyast_load set/list/generator comprehensions → five small private helpers, bodies byte-identical |
| native set emitters | `emit_issubset/issuperset/isdisjoint` (~49 lines each) → `_emit_set_relation`; **emitted `.ll` verified byte-identical** for all three before/after; the `_emit_` prefix keeps it out of the `emit_*` dispatch table |
| `Memory.find/find_one` | five byte-equivalent copies (Sqlite, Volatile, Redis, Mongo, ScaleTiered) → one canonical base impl routing through `self.get`. One documented broadening: plain `TieredMemory.find` now consults L2/L3 via `get`, consistent with `get` itself — the only in-repo caller requires exactly that |
| spawn OCC retries | both spawn paths → the existing `_run_function_with_occ`, extended with `deserialize_args` (spawn deserializes before the request context is pushed — `$ref` resolution needs the outer context) and `include_traceback`; retry counts, 409 payloads, backoff identical |
| scale observability URLs + test helpers | the tempo/prometheus/loki env→k8s→localhost resolution ladder → `service_urls.jac`; six-way-duplicated `register_user`/`call_walker`/`start_server`/`extract` → the existing `server_support` module (md5-verified identical copies only; divergent variants stay local) |
| byllm error/compaction/stream helpers | three identical provider-exception cascades → `_raise_mapped_provider_error` (clause order, wrap types, log text preserved); verbatim compaction blocks shared by the react loop and its streaming twin → `_maybe_compact`/`_emergency_compact_retry`; stream-retry decision → `_should_retry_stream` with sync/async sleeps left at call sites |
| cli/project | mcp snippet tempfile lifecycle ×8 → `_snippet_task`; duplicate poll-server-log loops → `_poll_for_server_url`; `JacConfig`'s twice-defined section map → `_section_map` |
| SSO providers | google/apple/github were byte-identical modulo the fastapi_sso class → `FastAPISSOBase`; class names, module paths, and positional ctor order preserved (tests patch FQ paths) |
| client runtime shims | `ErrorFallback`, `errorOverlay`, `__jacReactErrorHandler`, and the zod schema-unwrap block were byte-identical between the React and Solid shims and contain no framework globals → moved once into `client_runtime_core`, re-exported through the existing `glob:pub` pattern so the `@jac/runtime` surface is unchanged |
| admin ops-console | eight identical page headers + three unavailable-banners → `PageHeader`/`OpsUnavailableBanner` (right-slot for deviant pages; LogsPage's structurally different header stays inline); eight inline error fallbacks → the existing `getErrorMessage`. Roughly line-neutral — eleven copies became one |

## Verification

| Gate | Result |
|---|---|
| Compiler suite | 1,982 passed / 1 known baseline fail |
| Scale sso+admin (incl. the real vite build of the admin UI) | **80/80** |
| Scale server/k8s/inprocess | failures = the known port/dev-mode env set only |
| byllm suite | known set only (pre-existing mtir cache bug disclosed in #7218, + one flaky parallel test) |
| Solid jsdom runtime | 5/5 |
| Solid backend export/parity suite | 28/28 — one parity assertion updated: `ErrorFallback` is now a re-exported binding rather than a shim-local function definition |
| React integration (`JAC_CLIENT_INTEGRATION=1`) | failure set **byte-identical** to the pre-change baseline |
| Emitted LLVM IR for the set emitters | byte-identical `.ll` before/after |


ayushmk7:
Closing to restructure: splitting into smaller PRs (<100 files each) for reviewability. Replacements incoming.

PR #7232, opened 2026-07-06
Remove dead runtimelib/client code and orphaned CLI/project accessors

ayushmk7:
# Remove dead runtimelib/client code and orphaned CLI/project accessors

**+43 / −693 across 54 files (net −650), 7 commits.** Pure source deletions plus two restores the verification pass proved alive through non-obvious references. One of six sibling PRs splitting a larger dead-code effort into reviewable units.

## What's removed

| Commit | What it deletes |
|---|---|
| `refactor(runtimelib)` | The planner slimmed to its live `execute_path` fast path — the resolver rewrite superseded `execute_plan`/`plan_chain`/`plan_chain_ordered` and their seven private helpers (zero callers; `execute_path` keeps its API and tests). Fourteen client-rewrite orphans: `copy_typescript_files`, `check_runtime_version`, `sync_runtime_version`, `process_vite_imports`, `should_process_import`, the vite plugin-option formatters, `cleanup_temp_dir`, `run_install_cmd`, `check_command_available`, `Memory.query_by_type`, and more. The write-only `JacTestClient._current_user` field and dead imports across the module |
| `refactor(cli/project)` | The unreachable `get_command_groups` help accessor and its `_groups` global (only `get_help_formatter` is used); `DependencyRegistry`'s speculative CRUD surface (`get_by_name`/`get_all_flags`/`has_type`/`clear` — only `get_by_flag`/`get_all`/`register` are consumed); test-only `TemplateRegistry.has_template` |
| `refactor(misc)` | The never-raised `UnknownToolError` (its raiser died with an earlier byllm re-architecture), `sealed.is_sealed`, two `ext_registry` constants, an unreachable duplicate `return`, and two parameters every caller passed but no body read (`_parse_mappings(sources_count)`, `_parse_export_list(type_only)`) |
| `docs(byllm)` | Drops `UnknownToolError` from the byllm reference, the internals abstraction list, and the shipped `jac-by-llm` agent skill — their import examples would now `ImportError` |
| `test(runtimelib)` | Two test callers updated to `_parse_mappings`' slimmed signature |
| `fix(client)` ×2 | **Restores** `desktop_sdk_path` — the launcher bakes `desktop = …:desktop_sdk_path` into the fused binary's `entry_points.txt` (`payload.zig`) and modresolver loads `@jac/desktop` through it behind a blanket `except`, so removal would have silently dropped the desktop SDK from fused builds. **Restores** `TargetFactory.create_by_type` — a test asserts the method's body text in the file source. Both are references no symbol search sees; they're why every deletion here was also swept against `.zig`/`.lua`/config strings and test source-text |

## Verification

| Gate | Result |
|---|---|
| Compiler suite | pass modulo the known baseline + the host-env native class (reproduces on pristine `main`) |
| Runtime + project + cli suites | 1,174 passed; every failure in the known bun/server/env classes (each non-obvious name solo-rerun: e.g. `test_plugin_loader_toml_error` 2/2, source-map suites 28/28, `test_mobile_target` 7/7) |
| Desktop entry-point smoke (`desktop_sdk_path` resolves the SDK file) | True |
| Format sweep | clean |

Known-flaky disclosures (reproduce identically on pristine `main`): the bun/server client test classes, `test_dependencies` venv-PATH pair. Separately, this effort **discovered** (not caused) an upstream bug: JIR-cached in-package modules silently omit `SEC_MTIR` because the cache writer snapshots the internal program's always-empty `mtir_map` (`jac0core/impl/compiler.impl.jac:222`), so `test_mtir_integration` fails on any warm cache — reproducible on pristine `main` with a two-run probe. Happy to file it with the full analysis.


ayushmk7:
The `test-scale (server)` failure is a runner death, not a test failure: the "Run server tests" step ended with a null conclusion (all subsequent steps null, job log contains zero FAILED lines) — the runner was killed mid-step. Locally this branch's scale server suite runs green (sso 51/51, admin vite-build 29/29). A re-run should clear it.

ayushmk7:
Sharper root cause for both `test-scale` failures: the runner's zig caches missed ("Cache not found for input keys: jac-shim-…" at job start), so the test step spent its whole budget compiling the LLVM shim from libcxx sources — the log's final lines are C++ compilation, 14+ minutes in, when the step hit its timeout. Zero tests executed, zero FAILED lines. Locally this branch's `scale/tests/data` + `server` complete in 92s with 279 passed (failures only in the known port/mongo env set). A re-run with a warm cache should clear both jobs.

ayushmk7:
`test-compiler`'s 4 `test_native_error_gate` failures are not from this PR — reproduced identically on pristine `origin/main` (see #7241, which appears to have merged with this regression). Everything else on this branch is green.

PR #7249, opened 2026-07-07
fix(type-system): correct genuine type errors in the type checker's own source

SandeepaHWP:
## What this does

Fixes the genuinely-real type errors found in the compiler's own type checker while running `jac check` — wrong or missing annotations, and optional values used where a concrete type was required. Most reports from the wider sweep are checker false positives (tracked in #6809) and are left unchanged.

### Type checker (`compiler/type_system`)

**Annotations that were `any` or bare**
- `get_type_of_member` and `_get_method_with_overloads` took `member`/`method_member` as `any`, leaking `Any` into a `Symbol` parameter; both are now `ClassMember`.
- The `found_member` local in `_get_type_of_expression_core` was `any`; it is now `ClassMember | None`.
- `_find_overload_by_arg_count` took `overload_symbols: list`; it is now `list[Symbol]`.

**Parameters widened to what callers actually pass**
- `match_args_to_params` and `get_type_of_magic_method_call` only accepted `list[Expr]` / `list[Expr | KWPair]`, but callers pass `expr.params`, which is `Sequence[Expr | KWPair] | None`. The parameter type is widened accordingly and `match_args_to_params` iterates `arg_nodes or []`.

**Optionals guarded**
- `validate_call_args` could return the optional `prefetch.int_class`; it now falls back to `UnknownType`.
- The `typing.cast` special-case called `len(expr.params)` without a None check; it now guards `expr.params` first.

## Notes

- No behaviour change beyond the None handling, which previously would have crashed.
- Supersedes #6810.
- Rebased onto current `main`. Several fixes originally in this PR were fixed independently on `main` in the meantime and are no longer part of the diff:
  - `validate_arg_types` was declared `-> None` but returns a bool — already corrected on `main` (which also added a `solved_type_vars` parameter).
  - `_collect_boundary_type` and `boundary_analysis_pass`.
  - The two `jac0core/passes` codegen fixes (the `_py_name` None-guard and the exception-swallowing `pass;`). `pyast_gen_pass` has since been replaced by `jcir_gen_pass`, whose `_py_name` call sites already guard for `None` and whose `_get_ambient_typing_names` no longer has a `try`/`except` at all.
- Remaining `jac check` errors on these files are false positives; see #6809.

## Test plan
- [x] Existing test suite (unchanged behaviour beyond None-guards)
- [x] `jac fmt --check` clean on all touched files


jac/jaclang/compiler/type_system/type_evaluator.impl/parameter_type_check.impl.jac
```diff
@@ -61,7 +61,8 @@ impl TypeEvaluator.validate_call_args(expr: uni.FuncCall) -> TypeBase {
     if isinstance(caller_type, types.ClassType)
     and caller_type.is_builtin("auto")
     and (not expr.params or len(expr.params) == 0) {
-        return self.prefetch.int_class if self.prefetch else types.UnknownType();
+        return (self.prefetch.int_class if self.prefetch else None)
+        or types.UnknownType();
```
SandeepaHWP: Re #6810 — @kugesan1105 asked if this is a false positive. It's not: `int_class` is genuinely `ClassType | None` (can be `None` if the builtins stub lookup fails), and the old guard checked `self.prefetch` (always truthy) instead of `int_class` itself. Tested a minimal repro of both the broken and fixed versions — the broken one fails with the same E1002, the fixed `or` version passes because the checker strips `None` from non-final `or` operands.

SandeepaHWP:
Re #6810 — @kugesan1105 flagged `boundary_analysis_pass.impl.jac` line 738 (`arch_kind is not None` guard) as a possible false positive. Agreed, it was — the checker has no narrowing rule for `x in (tuple of literals)` at all, so no guard on that pattern will ever satisfy it; the `is not None` addition only worked by piggybacking on a different narrowing rule. Reverted it here since it's not a real fix. Filed the checker gap separately: #7256.

PR #7262, opened 2026-07-07
fix(checker): narrow on `in` and `not in` against literals

### LINKED ISSUE #7256: Type checker: 'x in (literal, literal)' does not narrow away None from str | None
## Summary

The type checker's narrowing engine does not understand membership tests against a tuple of literals (`x in ("a", "b")`), so it fails to narrow `x: str | None` down to `str` in the true branch of a conditional, even though `None` can never satisfy that membership test at runtime. This produces a spurious `E1001: Cannot assign str | NoneType to str` error on code that is actually safe.

## Minimal reproduction

```jac
def pick_kind(ak: str | None) -> str {
    result: str = ak if ak in ("node", "walker") else "obj";
    return result;
}

with entry {
    print(pick_kind("node"));
    print(pick_kind(None));
}
```

Running `jac check` on this file gives:

```
✖ Error: error[E1001]: Cannot assign str | NoneType to str
  --> repro.jac:2:5
    2 |     result: str = ak if ak in ("node", "walker") else "obj";
      |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

At runtime this code is completely safe: `None in ("node", "walker")` evaluates to `False`, so `ak` can never be `None` when the true branch is taken. The checker is wrong to flag it.

## Context

This surfaced while working on #7249 (rebase of #6810, "Fix real type errors in the compiler's type-checker and codegen passes"). A workaround for this exact pattern in `BoundaryAnalysisPass._collect_boundary_type` (`jac/jaclang/jac0core/passes/impl/boundary_analysis_pass.impl.jac`) was flagged in review as unnecessary once this checker limitation is fixed — see review discussion on #6810.

Image

SandeepaHWP:
Fixes #7256.

## Problem

Checking a variable against a tuple or list of literals with `in` or `not in` did not narrow its type.

```jac
def pick_kind(ak: str | None) -> str {
    return ak if ak in ("node", "walker") else "obj";  # error: ak may be None
}
```

`None` cannot be in `("node", "walker")`, so this error is wrong.

## Change

- Two new narrowing kinds are added, `MEMBERSHIP_IN` and `MEMBERSHIP_NOT_IN`.
- They apply when the right side is a tuple or list and every element is a literal. The variable is narrowed to that literal type, or to the union of the literal types.
- Each element is checked with `_extract_literal_operand_type` from #8432. Variables, f-strings and `typing.cast(...)` are not literals. If any element is not a literal, no narrowing happens.
- Only the safe branch is narrowed, following the rule from #8957:
  - `x in (...)` narrows `x` only inside the `if`.
  - `x not in (...)` narrows `x` only inside the `else`.
  - The other branch keeps the original type.

## Known limitation

`not in` does not reduce a union of literals yet. For example, with `k: Literal["a", "b", "c"]`, `k` keeps its full type inside `if k not in ("a", "b")`. This is safe, and the tests record it as an expected error. Fixing it needs literal-aware type exclusion, which will be a separate PR.

## Tests

`checker_membership_narrowing.jac`, run by the `membership_narrowing` test.

- **Pass with no errors (3 cases):**
  - `in` with a tuple narrows to exactly `Literal["node", "walker"]`
  - `not in` with a list narrows after an early `return`
  - `in` with a single-element tuple narrows to `Literal["node"]`
- **Report the expected error (4 cases):**
  - a variable element, so nothing is narrowed
  - after `if x in (...)`, the variable keeps its full type
  - inside `if x not in (...)`, the variable keeps its full type
  - the known limitation above

Each expected error is checked by its full message, which includes the type.

The `literal_eq_narrowing` and `literal_ne_does_not_over_narrow` tests from #8432 also pass. `jac check` on the type evaluator and `jac fmt --check --lintfix` on the changed files pass.


SandeepaHWP:
Addressed both greptile comments: assertions now check against the joined error text instead of `errors_had[0]`/`errors_had[1]` by position (order-independent), and added `test_membership_not_in_list` plus positive coverage for int/float/bool literal elements in a membership tuple.

kugesan1105 (changes requested):
Thanks for this, the classifier diagnosis is right and I confirmed both reported repros pass with the diff applied locally. But I can't merge it as is: the narrowing engine's negative direction is unsound, and this PR extends that unsoundness to strings and to the new membership paths instead of containing it.

The deeper problem is that `exclude_type_from_union` is literal-blind. It keys union members by `shared.canonical_key()`, and `clone_with_literal` keeps the same `shared`, so excluding `Literal["node"] | Literal["walker"]` from `str | None` deletes the entire `str` member and leaves `None`. (This already bites `int`/`float`/`bool` on main, e.g. `if v != 5` on `int | None` narrows the true branch to `None`, so it's not something this PR introduced, but the PR spreads it to three new places.)

With this PR applied, all three functions below type-check clean; at runtime each `None`-typed variable holds the string `"other"`, and native lowering demotes all three functions (E5092, "binding a void-typed value"):

```jac
def ne_unsound(v: str | None) -> object {
    if v != "node" {
        n: None = v;   # accepted; at runtime n == "other"
        return n;
    }
    return "eq";
}

def notin_unsound(ak: str | None) -> object {
    if ak not in ("node", "walker") {
        n: None = ak;  # accepted; at runtime n == "other"
        return n;
    }
    return "in";
}

def in_false_branch(ak: str | None) -> object {
    if ak in ("node", "walker") {
        return "matched";
    }
    n: None = ak;      # accepted; at runtime n == "other"
    return n;
}
```

Two more specific issues:

1. F-string hole: `isinstance(right_expr, (..., MultiString))` also matches f-strings, whose evaluated type is bare `str`, not a literal. So `v != f"no{x}e"` and `ak not in (f"{x}", "walker")` narrow `str | None` to `None` as well (validated live, same wrong runtime value). The classifier should decide by the evaluated type (`is_literal_type()`), not by node shape.

2. `test_error_membership_false_branch_unnarrowed` passes for the wrong reason: with this PR, `ak` on the false branch is over-narrowed to `None`, and returning `None` against `-> str` still errors, so the assertion holds while the comment's claim ("ak is still str | None") is false.

What I'd like to see before this lands:

- Literal-aware exclusion in `exclude_type_from_union`: drop a union member only when it is the same literal; excluding `Literal["x"]` from a bare `str` member must be a no-op. This is the actual root cause and fixing it also repairs the existing `int`/`float`/`bool` hole.
- The classifier recognizes literals by evaluated type, so f-strings and references never qualify.
- If you'd rather keep the exclusion fix out of scope, then the inverted directions (`!=` true branch, `not in` true branch, `in` false branch) must produce no narrowing at all until it lands; only the positive directions are sound today.
- Tests that assert the narrowed type or the specific expected error on the negative direction over union targets, not just an error count.

The membership feature itself is a good addition and the positive-direction narrowing is sound, so this is close; it just can't ride on the current exclusion machinery in the negative direction.

SandeepaHWP:
Thanks for the detailed review. I rebuilt the PR around it:

- **Inverted directions:** I took the "no narrowing" option. Following #8957, `in` narrows only its true branch and `not in` only its `else`. The fixture checks both negative directions against the exact message, `Cannot return str | NoneType, expected str`. That proves `ak` stays `str | None` there, so your `n: None = ak` repros are rejected.
- **Literal recognition:** each element goes through `_extract_literal_operand_type` from #8432. It accepts literal syntax only, so f-strings, variables, references and `typing.cast(...)` never qualify.
- **Scope:** string `==` / `!=` moved to #8432, and this PR is now membership only, stacked on it.
- **Literal-aware exclusion** is left for a follow-up. The one case it would fix, `not in` over a literal union, is recorded in the fixture as an expected error.

SandeepaHWP:
Example:


```jac
# Inside the `if`, kind is "node" or "walker", so it can't be None.
def label_in(kind: str | None) -> str {
    if kind in ("node", "walker") {
        return kind;
    }
    return "other";
}

# `not in` narrows the other way: past the early return, kind is "node" or "walker".
def label_not_in(kind: str | None) -> str {
    if kind not in ("node", "walker") {
        return "other";
    }
    return kind;
}

# Still an error on purpose: outside the tuple, kind may be any str or None.
def label_rest(kind: str | None) -> str {
    if kind in ("node", "walker") {
        return "known";
    }
    return kind;
}
```

kugesan1105 (approved):
@SandeepaHWP  good fix; check the limitations

--------------------------

Image

PR #7263, opened 2026-07-07
feat(byllm): terminal-tool designation and strict structured-output type adherence

Jayanaka-98:
# feat(byllm): terminal-tool designation, omittable tool params, and ReAct completion fixes

## Summary

Lets a caller designate their own tool as the one that ends a ReAct loop and have byLLM enforce it, fixes several silent-failure paths in the structured-output and tool-calling machinery, and records parameter defaults in the MTIR so tool schemas stop forcing optional arguments.

Motivated by an AppWorld benchmark report (context in BYLLM_TERMINAL_TOOL_ISSUE.md): reasoning models routinely ended a tool session without calling the required terminal tool, so byLLM's loop exited "successfully" while the task was never completed and scores silently zeroed.

## Features

**Terminal-tool designation and enforcement.** `by llm(finish_tool=fn)` marks a caller's own function as the tool that ends the ReAct loop and yields the call's output, instead of relying on byLLM's auto-injected `finish_tool`:

```jac
def agent_step(task: str) -> str by llm(
    tools=[show_api_docs, execute_code],
    finish_tool=finish,
    max_react_iterations=50
);
```

Terminal-ness moved from the hardcoded `finish_tool` name to an `is_terminal` flag on the `Tool` instance (call-scoped, no mutation of the function). Detection funnels through a new `MTRuntime.get_terminal_tool()`. When a terminal is designated, byLLM does not inject its generic `finish_tool`, and it enforces the designated one: if the model tries to end without calling it, byLLM re-prompts once with only that tool available. Two non-blocking warnings flag footguns (a tool literally named `finish_tool`, and an undesignated `finish`-named tool).

**Omittable tool parameters (defaults recorded in MTIR).** Tool-schema generation marked every parameter `required`, so the model was pushed to supply optional args like `finish(answer: str = "")` and padded them with filler. The compiler now records whether a parameter has a default (`ParamInfo.has_default`, a syntactic property available in static analysis), and schema generation omits defaulted params from `required`. For plain Python tool functions (no MTIR), it falls back to `inspect.signature`.

## Bug fixes

**Structured output could silently degrade to a raw string.** When the ReAct loop exhausted its iterations, `_force_final_answer` returned `str(output)` for any return type, so a call declared to return an object could hand back a bare string with no error. It now validates for structured return types and raises into the correction-retry loop.

**finish_tool validation errors bypassed the correction-retry loop.** `FinishToolError` was a sibling of `OutputConversionError`, so a finish-tool output that failed validation escaped as a hard error. It is now caught in the typed-retry loops and carries `raw_output` so the correction prompt can echo the rejected value.

**Str-return agents could exit without calling a required terminal tool.** `_attempt_recovery` was gated off whenever the return type was `str`, so a designated terminal tool was never enforced for the common `-> str` agent. The gate now keys on whether a terminal tool is designated, giving those agents the completion guarantee while still allowing a plain text answer when no terminal is designated.

**MockLLM tool calls did not carry runtime tool flags.** `MockToolCall` wrapped a fresh, disconnected `Tool`, while real dispatch resolves calls through `mt_run.get_tool()` in `_parse_tool_calls`. The mock now re-points each tool call to the registered tool, so it mirrors real flow (and terminal detection works under test).

## Design notes

- Extra fields in a structured response are dropped (pydantic's default): the required data is present, so the typed object is built and re-prompting is avoided. The raw response with the extras stays visible in the LiteLLM call log. Missing or wrong-typed required fields still raise and drive the retry loop. (An earlier strict "reject undeclared fields" approach was considered and dropped as the wrong trade for the common case.)

## Tests

- Terminal tool: instance flagging, terminal ends the loop with its output, designated terminal forces recovery on a natural exit, plain str agent answers in text without a forced turn.
- Defaults in MTIR: the compiler-pass test compiles a real fixture where `analyze_person(person, detail_level: int = 3)` and asserts `has_default` is captured (True for `detail_level`, False for `person`); a schema test confirms a defaulted param is omitted from `required`; the Python-mode `inspect` fallback is covered too.
- Structured output: undeclared fields are dropped (top-level and nested), leaf coercion stays lenient, a missing required field raises.
- finish_tool validation error carries `raw_output`.


marsninja:

PR #7273, opened 2026-07-08
fix(config): handle [plugins.*] sections on config load + ensure [client] on create

chess10kp:
## Problem

When a jac.toml has `[plugins.client]` instead of `[client]` — either from a custom jacpack that omits the section, or from a save that rewrites the structure — `_parse_toml_data` only scans top-level keys (`"byllm" in data`, `"client" in data`, …). A `[plugins.client]` section lands in `data["plugins"]["client"]` and is **silently dropped**. `config.plugins["client"]` stays empty → `is_client` is `False` → `@tanstack/react-form`, `zod`, and other client npm deps are never auto-injected → Vite build fails with "Rollup failed to resolve import @tanstack/react-form".

## Fix

Three targeted changes:

1. **`_parse_toml_data`** — after the existing top-level feature-key loop, also iterate `data["plugins"]` so `[plugins.client]` sections are recognised.

2. **`merge_from_toml_file` + `apply_profile`** — same fix for config merges (profile overlays, `jac.local.toml`).

3. **`_stamp_kind`** — belt-and-suspenders: when creating a project with a client kind (web-app, web-static, desktop, mobile), always stamp an empty `[client]` into the template config so the auto-injection fires even when a custom jacpack omits the section.

## Verification

- All 15 `create_kinds` tests pass (web-app, web-static, desktop, mobile, 11 language-level kinds)
- Config parsing tests pass


chess10kp (commented):
Addressed Greptile feedback: `_parse_toml_data` now uses `_deep_merge` when both top-level `[client]` and `[plugins.client]` coexist, matching the behavior of `merge_from_toml_file` and `apply_profile`.

marsninja:
we dont want to lock in permanent support for this, perhaps we can have a more helpful error if anything, ultimately jac.tomls should be updated.

chess10kp:

Issue #7250, opened 2026-07-07
Jac Scale: auto-detect target node arch for jac start --scale (cross-arch deploy silently crashloops with Exec format error)


udithishanka:
## Summary

`jac start --scale` selects the pod's `jac` binary architecture from the **operator's host arch**, not the **cluster's node arch**. Deploying from an Apple Silicon Mac (arm64) to amd64 nodes ships `jac-*-linux-aarch64` to x86_64 pods, so the `jac-bootstrap` init container dies with `Exec format error` (exit 126) and every pod sits in `Init:CrashLoopBackOff` indefinitely — while the deploy reports success ("Fleet ready: 5/5 deployments serving"). The deploy already has an authenticated Kubernetes client, so it should auto-detect the node architecture (with `JAC_NODE_ARCH` kept as an explicit override) and/or fail fast on mismatch, rather than shipping a binary the nodes can't execute.

## Environment

| | |
|---|---|
| jac | 0.30.8 |
| Operator host | macOS arm64 (Apple Silicon) |
| Target | EKS, nodes `amd64`/linux |
| Command | `jac start --scale` (microservice app, 5 deployments) |

## Reproduction (executed)

From an arm64 Mac, deploy a scale app to an amd64 cluster with no `JAC_NODE_ARCH` set:

```bash
jac start --scale
```

The rollout applies everything and eventually prints `Fleet ready`, but pods never become Ready:

```
$ kubectl -n this-is-jac get pods
gateway-deployment-...        0/1  Init:CrashLoopBackOff  8
analytics-deployment-...      0/1  Init:CrashLoopBackOff  8
guestbook-deployment-...      0/1  Init:CrashLoopBackOff  8
social-graph-deployment-...   0/1  Init:CrashLoopBackOff  8
source-viewer-deployment-...  0/1  Init:CrashLoopBackOff  8

$ kubectl -n this-is-jac get pod <gateway> \
    -o jsonpath='{.status.initContainerStatuses[?(@.name=="jac-bootstrap")].lastState.terminated.exitCode}'
126

$ kubectl -n this-is-jac logs <gateway> -c jac-bootstrap
bash: line 1: /app/.jachome/.local/bin/jac: cannot execute binary file: Exec format error
```

Root cause is a host-vs-node arch mismatch:

```
$ kubectl get nodes -o custom-columns=ARCH:.status.nodeInfo.architecture
amd64            # all nodes

$ file jac/zig-out/bin/jac
[...]

PR #7280, opened 2026-07-08
fix(scale): make the bootstrap guard execute the shipped jac binary

MusabMahmoodh:
## Summary

Follow-up to #7250. The `jac-bootstrap` init container guards the shipped binary with `command -v jac`, which only checks the file exists on PATH with the execute bit set. A binary built for the wrong architecture passes that check, so the pod died later on the first real `jac` invocation with a bare `bash: cannot execute binary file: Exec format error` (exit 126) and the FATAL message with the `JAC_NODE_ARCH` hint never fired; it was effectively dead code for the exact failure it names.

The guard now runs `jac --version`, so an unrunnable binary (wrong arch, corrupt unpack) fails at the guard and the init-container log shows the actionable message instead of the raw exec error.

## Tests

`jac test jac/jaclang/scale/tests/microservices/test_binary_injector.jac` (20 passed), including a new case asserting the guard executes the binary (`jac --version`), no longer uses `command -v`, and carries the `JAC_NODE_ARCH` guidance.

### DIFF
--- a/docs/docs/community/release_notes/unreleased/jaclang/7280.bugfix.md
+++ b/docs/docs/community/release_notes/unreleased/jaclang/7280.bugfix.md
@@ -1,0 +1,1 @@
+- **Fix: pod bootstrap guard detects unrunnable binaries**: the init-container guard now executes the shipped `jac` binary instead of only checking it exists on PATH, so a wrong-arch or corrupt binary fails at the guard with the `JAC_NODE_ARCH` guidance instead of a bare `Exec format error` later in the boot.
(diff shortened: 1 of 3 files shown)


MusabMahmoodh:
Closing: the bootstrap has since been rewritten around this diff (checksum-gated reuse of the base image binary), and the arch-mismatch failure this guarded against is better caught at the driver by #7278 before anything ships. If an execute-not-just-locate guard is still wanted in the new bootstrap, it deserves a fresh small diff there.

PR #7282, opened 2026-07-08
fix(scale): judge fleet readiness by the current generation, incl. init containers

MusabMahmoodh:
## Summary

Follow-up to #7250, which reported `jac start --scale` printing `Fleet ready: 5/5 deployments serving` while every pod sat in `Init:CrashLoopBackOff`. Two blind spots in `_wait_for_fleet_rollout` made that possible:

- **Stale-generation readiness**: a deployment counted as ready when `ready_replicas >= 1`, which pods of the previous ReplicaSet satisfy during a rolling update. On a redeploy the wait could report the whole fleet serving while every new pod crash-looped. Readiness now follows the `kubectl rollout status` rules: generation observed, all replicas updated, old pods gone, updated pods available.
- **Init containers invisible to the crash fast-fail**: the detector only read `container_statuses`, so an `Init:CrashLoopBackOff` (exactly what a wrong-arch `jac-bootstrap` produces) never tripped it and a fresh deploy burned the full 20-minute timeout before failing with a generic message. Crash detection now covers `init_container_statuses` too.

The readiness and crash checks are extracted as `deployment_rolled_out()` / `crashloop_reason()` so they are testable in isolation.

## Tests

New `jac test jac/jaclang/scale/tests/microservices/test_fleet_rollout.jac` (9 passed):
- rolled-out / stale-generation / draining-old-pods / missing-status cases for `deployment_rolled_out`
- init-container and app-container crash loops for `crashloop_reason` (and the below-threshold case)
- `_wait_for_fleet_rollout` integration with a mocked cluster: returns once the new generation serves; raises `crash-looping` on an init-container loop


MusabMahmoodh:
Closing as superseded: main now scopes the fleet-wait crash check to the current ReplicaSet per deployment (#7076) and surfaces init-container crashloops in fleet-wait (#7295), which together cover what this PR set out to fix. The branch also conflicts with those merges.

Issue #7244, opened 2026-07-07
Jac Scale: deploy manifest apply is non-transactional - mid-apply failure leaves Deployments+Services running but no Ingress and a failed status


udithishanka:
## Summary

A jac-scale production deploy applies its Kubernetes resources non-transactionally: it creates the Deployments and Services first, then applies autoscalers (HPA), then the Ingress. When a step in the middle fails, the apply aborts but the already-created resources are left running with no rollback. The observed result is a namespace full of healthy, ready pods that have no Ingress and therefore no public URL, while the deploy is correctly reported as `failed`. From inside the cluster everything looks alive; from outside the app is unreachable. A single failure in the optional HPA/autoscaler step should not take the whole deploy down, and the Ingress (the thing that actually makes the app reachable) should not be gated behind it.

## Environment

| | |
|---|---|
| jac binary | 0.30.8 (rolling dev channel) |
| Platform | Linux x86_64 |
| Runner | jacBuilder deploy_ops (jac-scale KubernetesTarget), PR-578 preview on `jaseci-cluster` (EKS us-east-2) |
| App | multi-service Jac app: gateway + analytics + guestbook + social-graph + source-viewer + mongodb + redis |

## Reproduction (executed, not hypothetical)

Deployed a multi-service app to production through the Deploy wizard on the live PR-578 preview. The manifest apply failed at the HPA step (see the sibling issue for the file-read cause). The deploy reported failed, but the workload had already been created. Observed live:

```
$ kubectl -n jb-952b87a0-imported-app get pods
NAME                                       READY   STATUS    AGE
analytics-deployment-...                   1/1     Running   5m44s
gateway-deployment-...                     1/1     Running   5m44s
guestbook-deployment-...                   1/1     Running   5m44s
social-graph-deployment-...                1/1     Running   5m44s
source-viewer-deployment-...               1/1     Running   5m44s
imported-app-mongodb-0                     1/1     Running   24m
imported-app-redis-...                     1/1     Running   26m

$ kubectl -n jb-952b87a0-imported-app get ingress
No resources found in jb-952b87a0-imported-app namespace.

$ kubectl -n jb-952b87a0-imported-app get hpa
No resources found in jb-952b87a0-imported-app namespace.

# gateway is genuinely serving inside the cluster
$ kubectl -n jb-952b87a0-imported-app logs <gateway-pod> | tail
INFO:     "GET /healthz/ready HTTP/1.1" 200 OK
[...]

Issue #7243, opened 2026-07-07
Jac Scale: prebuilt-binary runtime cache non-deterministically missing deploy resource files (redis.conf.template, hpa_autoscaler.jac) -> [Errno 2] deploy failures


udithishanka:
## Summary

Production deploys through jac-scale fail non-deterministically with `[Errno 2] No such file or directory` on deploy support files that live under `jaclang/scale/deploy/`. Two consecutive deploys of the same app failed on two different files: `autoscale/hpa_autoscaler.jac` (a lazily imported module) and `database/redis.conf.template` (a data file read by filesystem path). Both files exist in source and are present in the pod's runtime cache after the fact, so this is not a source problem: the prebuilt-binary runtime cache (`~/.jachome/.cache/jac/rt/<hash>/site/...`) does not reliably contain these file-path-read resources and lazily-imported modules at the moment the deploy reads them. The result is a hard deploy failure whose specific file varies run to run.

## Environment

| | |
|---|---|
| jac binary | 0.30.8 (rolling dev channel, prebuilt binary) |
| Platform | Linux x86_64 |
| Runner | jacBuilder deploy_ops (jac-scale KubernetesTarget), PR-578 preview on `jaseci-cluster` (EKS us-east-2) |
| App | multi-service (microservice) Jac app: gateway + 4 services + mongodb + redis |

Reproduced twice, live, on two independent deploy attempts of the same imported project.

## Reproduction (executed, not hypothetical)

On the live PR-578 preview (`https://jac-builder-pr-578.jaseci.org`), imported a multi-service Jac app via the Deploy wizard and clicked Deploy to production. Ran it twice.

**Attempt 1** — verbatim `deployment_message` from the backend project record:

```
manifest apply failed: Could not read file /app/.jachome/.cache/jac/rt/ff303679af15630e-91159af80697e1c8/site/jaclang/scale/deploy/autoscale/hpa_autoscaler.jac: [Errno 2] No such file or directory: '/app/.jachome/.cache/jac/rt/ff303679af15630e-91159af80697e1c8/site/jaclang/scale/deploy/autoscale/hpa_autoscaler.jac'. Report this issue: https://github.com/jaseci-labs/jaseci/issues
```

**Attempt 2** (retry, minutes later) — failed earlier, on a different file:

```
[...]

PR #7285, opened 2026-07-08
fix(scale): import the default HPA autoscaler eagerly so deploys don't race the runtime-cache extraction (#7243)

udithishanka:
## Summary

Fixes the autoscaler case of #7243. Production deploys through jac-scale fail non-deterministically with `[Errno 2] No such file or directory` on `jaclang/scale/deploy/autoscale/hpa_autoscaler.jac`, aborting the manifest apply (and, per #7244, leaving a half-created workload with no Ingress).

## Root cause

`AutoscalerFactory.create()` imports the HPA autoscaler **lazily, inside the method** (runs at deploy time):

```jac
if engine is None or engine == "hpa" {
    import from jaclang.scale.deploy.autoscale.hpa_autoscaler { HPAAutoscaler }  # lazy
    return HPAAutoscaler(logger=logger);
}
```

On the prebuilt `jac` binary, the runtime cache (`~/.cache/jac/rt/<hash>/site/...`) extracts lazily-imported modules on first access. A deploy-time import races that extraction: the manifest apply reads the path before the unpack lands, so it fails on a file that exists in source and is present moments later. Which file varies run to run; `hpa_autoscaler.jac` is the one on the HPA path because it is the only deploy module still imported lazily (the rest are imported at module load and unpack early).

## Fix

Import the default HPA engine at **module load** instead of lazily in `create()`, so the binary extracts it before any deploy runs:

```jac
import from jaclang.scale.deploy.autoscale.hpa_autoscaler { HPAAutoscaler }
...
if engine is None or engine == "hpa" {
    return HPAAutoscaler(logger=logger);
}
```

- **No circular import**: `hpa_autoscaler` imports only `autoscaler` / `_optdeps.kubernetes` / logger / k8s utils, never `factory`, and the built-in engines are not self-registered.
- **KEDA stays lazy** (optional engine, rarely used, may pull optional deps) — only the always-used default is made eager.
- `factory` is already imported at module load by `kubernetes_target` / the microservice target, so HPA now unpacks when the deploy target loads, not mid-apply.

## Scope / follow-ups

This closes the observed autoscaler failure. The sibling data-file case in #7243 (`database/redis.conf.template`, read via `Path(__file__).parent`) is the same class and warrants the same eager treatment. The complete fix is upstream in the binary's runtime-cache extraction (make lazy unpack race-free / eager for deploy resources); #7244 is the complementary hardening (an optional autoscaler failure should not abort the deploy or gate the Ingress).

## Testing

Verified by source inspection against `main` (no circular import; import path unchanged). I could not build the prebuilt binary locally to reproduce the race end to end; relying on CI plus a jac-scale deploy to confirm the autoscaler path.

Related: #7243, #7244.

### DIFF
--- a/docs/docs/community/release_notes/unreleased/jac-scale/7243.bugfix.md
+++ b/docs/docs/community/release_notes/unreleased/jac-scale/7243.bugfix.md
@@ -1,0 +1,1 @@
+Fixed Jac Scale deploys failing with `[Errno 2] No such file ... hpa_autoscaler.jac`: the default HPA autoscaler is now imported eagerly so the prebuilt binary extracts it at module load instead of racing the lazy extraction during manifest apply. Fixes #7243 (autoscaler case).
--- a/docs/docs/community/release_notes/unreleased/jac-scale/7243.bugfix.md
+++ b/docs/docs/community/release_notes/unreleased/jac-scale/7243.bugfix.md
@@ -1,1 +1,0 @@
-Fixed Jac Scale deploys failing with `[Errno 2] No such file ... hpa_autoscaler.jac`: the default HPA autoscaler is now imported eagerly so the prebuilt binary extracts it at module load instead of racing the lazy extraction during manifest apply. Fixes #7243 (autoscaler case).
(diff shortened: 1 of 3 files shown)


udithishanka:
this is just a symptom fix. not the root fix. closing now.

PR #7301, opened 2026-07-08
fix(release): refine release process and add verification for install.sh

Thamirawaran:
Cut the GitHub Release as a draft, attach every native binary, then publish as latest in one API call so "Latest" never appears mid-build with assets missing. The publish job polls until all three platform binaries are present (assets API can lag uploads) before flipping the draft to published+latest, and a post-publish matrix verifies the real curl | install.sh path per platform.


Thamirawaran:
Superseded by #7317 — rebased onto latest main as a single commit. Closing this one.

PR #7306, opened 2026-07-08
feat: scale K8s deploy ships a sealed .jab app artifact (#7288)

### LINKED ISSUE #7288: Jac Scale: migrate the K8s deploy artifact from ad-hoc .tar.gz to the sealed .jab
## Summary

The scale deploy path still ships the pod toolchain as an ad-hoc `bundles/{app}/{digest}.tar.gz` and unpacks it pod-side with `tar xzf`, even though the sealed **`.jab`** artifact now exists in `runtimelib/`. The injector never imports it. This issue tracks migrating the K8s deploy path to seed and consume a single `.jab` instead of the bespoke tar, so the deploy artifact matches the rest of the build system (one sealed, self-describing, precompiled bundle).

It also corrects two architecture-diagram/mental-model mismatches that surfaced while reviewing this path.

## Current state (verified against `main`)

**The `.jab` format exists but the scale injector doesn't use it.**

- `jac/jaclang/runtimelib/jab.jac` is present.
- `grep` for `.jab` under `jac/jaclang/scale/injector/` — no references.
- `pack_source()` (`bundle.jac:270`) packs a plain tar; `bundle_object_key()` (`bundle.jac:327`) returns `bundles/{app_name}/{digest}.tar.gz`.
- Pod-side unpack is `tar xzf` (`pvc_injector.jac:67`).

**What is already correct (no change needed):** the ~170 MB `jac` binary already flows **driver → content-addressed PVC → pod**, not downloaded in-cluster:

- `BinaryInjector.prepare()` downloads + checksum-verifies the binary on the driver (`binary_injector.jac:28`).
- It's appended to `_seed_artifacts()` and seeded once via the content-addressed PVC path (`binary_injector.jac:49`, `pvc_injector.jac:204`).
- The init container installs it from the PVC: `install -m 0755 "$TC/bin/jac"` (`runtime_bootstrap.jac:70`).

## Proposed Architecture

Image

## Deploy path today

```
ci / local machine (driver)
  ├─ pack bundle  (pack_source → .tar.gz)   ← should become .jab
  ├─ download + checksum jac binary          ← already correct
  └─ object key = sha256(bytes)
        │  kubectl cp via short-lived bundle-loader pod
        ▼
     [ PVC ]  ReadWriteMany · content-addressed
        │      (test -f <sha256> → skip if already seeded)
        ▼
  Worker Node · app pod (Deployment template)
     ├─ init: jac-pvc-bootstrap   →  tar xzf bundle from PVC   ← should become .jab materialize
     ├─ init: jac-bootstrap       →  install bin/jac from PVC · jac install
     └─ app container             →  runs jac + bundle   (Internet = app egress only)
```

Note the init containers live in the **app pod on the Worker Node** (appended to the Deployment template's `initContainers` — `kubernetes_target.jac:1073`/`1088`), **not** in the control-plane/Master. Payload bytes ride the PVC data plane; the API server carries control traffic only — nothing large is stored in ETCD.

## Scope of change

**1. Emit `.jab` from the injector (the real work)**
- Route `PvcInjector.prepare()` / `bundle_object_key()` to produce the sealed `.jab` instead of `.tar.gz`.
- Keep the content-addressed (`sha256`) object key and the driver→PVC seed transport unchanged.

**2. Consume `.jab` pod-side**
- Replace the `tar xzf` unpack in `build_pvc_init_container()` (`pvc_injector.jac:67`) with `.jab` materialization.

**3. Keep the binary path as-is** — it already does the right thing.

## Files

| Concern | File |
| --- | --- |
| Packing / object key | `jac/jaclang/scale/injector/bundle.jac:270`, `:327` |
| Seed + pod-side unpack | `jac/jaclang/scale/injector/pvc_injector.jac:67`, `:204` |
| Binary seed (already OK) | `jac/jaclang/scale/injector/binary_injector.jac:28`, `:49` |
| Init container wiring | `jac/jaclang/scale/deploy/target/kubernetes/kubernetes_target.jac:1073`, `:1088` |
| `.jab` format | `jac/jaclang/runtimelib/jab.jac` |

## Out of scope / follow-ups

- Diagram corrections (init container is on the Worker Node, not Master; the Internet arrow is app egress only, not a toolchain fetch) — these are doc/mental-model fixes, tracked here for context.
- Graduating the PVC seed transport from `kubectl cp` to an OCI/object-store pull — a separate, larger change.

## Acceptance

- Scale deploys seed and consume a `.jab`; no `.tar.gz` toolchain bundle remains in the deploy path.
- `sha256` content-addressed dedupe + warm-redeploy skip still holds.
- The jac binary is still driver-downloaded and PVC-seeded (unchanged).
- Scale K8s e2e still green.

kugesan1105:
## Summary

Migrates the Jac Scale Kubernetes deploy artifact from an ad-hoc `.tar.gz` bundle to a sealed **`.jab`** image (issue #7288). Both single-app and microservice deploys now ship a manifest-verified, precompiled image; the seal is mandatory.

## The key insight

A `.jab` **is** a gzip tar containing `_precompiled/MANIFEST.json` at the tar root. `pack_source` already `--seal`ed the app and placed `_precompiled/` at the root, so the old `.tar.gz` was structurally almost a `.jab` already. That makes this migration small and keeps the pod run-model unchanged:

- The init container's `tar xzf app.jab -C /app` produces the **same** `/app` tree as before (source + `_precompiled/` + `jac.toml`) - a `.jab` untars like any gzip tar.
- When a pod runs its entrypoint, the compiler **auto-loads the sibling `_precompiled/` sealed image** (`_get_precompiled_jir_bytecode` walks up to find `_precompiled/`), so both `jac start <file>` (single-app / service) and `jac scale gateway` get the precompiled modules with **no cold-compile step** and no pod-command change.

## What changed

- **`bundle.jac`**: `pack_source` → **`pack_jab`**. Seals the app with the pod binary into `_precompiled/MANIFEST.json`; the seal is now **mandatory** - an incomplete seal raises instead of silently shipping a source-only cold-compile bundle. `bundle_object_key` → `bundles/{app}/{digest}.jab`.
- **`pvc_injector.jac`**: calls `pack_jab`; the staging temp suffix is neutralized (the app bundle is `.jab`, the toolchain bundle stays `.tar.gz`). `tar xzf` is unchanged - it accepts a `.jab`.
- **Container commands**: unchanged. Single-app `jac start`, microservice `jac scale gateway` / `jac start <svc>` all run against the extracted `/app` tree as before.
- **Tests**: updated to the new contract - `pack_jab` requires the pod binary; toml-sanitization / db-baking tests assert `sanitized_jac_toml` directly (format-independent, no binary needed); seal-dependent assertions guard on the checkout binary (`zig-out/bin/jac`) and run in CI. App bundle key asserts `.jab`; toolchain bundle stays `.tar.gz`.
- **Docs**: new "App Artifact (`.jab`)" section in the K8s reference + a breaking-change release note.

## Breaking change

Deploys that previously tolerated an incomplete seal (shipping source that cold-compiled on the pod's first boot) now **fail loudly** if the app cannot be sealed. Fix or exclude the offending module and redeploy.

## Microservice mode

`.jab` is single-`entry`, but that only constrains the `jac run app.jab` shortcut - **not** what a pod executes. Because the init container extracts the whole `.jab` to `/app` and the compiler auto-loads the `_precompiled/` image, each microservice pod still selects its member (`$JAC_SV_NAME` → `jac start <svc>.jac`) or runs `jac scale gateway`, now backed by the sealed precompiled modules. Kubernetes still owns the service fan-out (one Deployment per route). No new multi-entry run contract was needed.

## Testing

- 26 scale unit tests pass locally (`test_app_precompile`, `test_pvc_injector`, `test_seed_determinism`, `test_binary_injector`, `test_service_pod_npm`).
- Seal-dependent tests skip locally (no `zig-out/bin/jac`) and run in CI where `build-jac` provides the binary.
- The real-app K8s e2e (which ships the PR-built binary via the `local` channel) exercises the full seal → `.jab` → extract → run path end-to-end.

Closes #7288.


kugesan1105:
Thanks — both addressed in 25c43c6c32:

1. **Staged-dir leak (real bug).** `pack_jab` cleaned up `staged.parent` inside the `with tar:` block, so a `tar.add` failing after the seal would leak a full project copy in /tmp. Moved the cleanup to an outer `finally` that wraps the whole tar-building block, so it runs on every exit path.

2. **Dropped tests.** Correct — the `test_pvc_injector` rewrite inadvertently dropped three tests for unchanged code. Restored the two `build_bundle_pvc` error-path tests verbatim (RWX access-mode, no-storage-class guard), and re-added the `mutate_pod_spec` wiring test adapted to inject seeded state directly (since `prepare()` now requires a pod binary to seal). All 9 pvc tests pass.

Note: the `.bundle` temp suffix Greptile flagged as cosmetic was already removed in an earlier commit (0e1e891b5d) — the scratch file now has no suffix, since the real PVC name comes from the content-addressed object_key.

PR #7307, opened 2026-07-08
fix(scale): make provisioned Redis pod resources configurable via [scale.database]

### LINKED ISSUE #7293: Jac Scale: provisioned Redis pod resources hardcoded (256Mi) and OOMKilled in prod; expose as [scale.database] keys
## Summary

The Redis instance that `jac-scale` provisions for every deployed app has its container CPU/memory **requests and limits hardcoded** in `jac/jaclang/scale/deploy/database/kubernetes_redis.jac` (limit `cpu 500m / memory 256Mi`). There is no `[scale.database]` key to raise them. The only memory knob exposed to `jac.toml` is `redis_max_memory` (Redis' internal `maxmemory`, default `200mb`), which controls key eviction, not the container cgroup limit.

In production this 256Mi cap is too small for Redis' real RSS (dataset + BGSAVE fork copy-on-write + fragmentation + client/replication buffers), so the pod is repeatedly **OOMKilled**. Each kill takes Redis down for the restart window, during which clients get `Error 111 connecting to <app>-redis-service:6379. Connection refused`. That is the user-facing failure tracked in jacBuilder#634 (most-reproduced dogfood failure of the July 2026 round). Same undersized-Redis root as jaseci#551, different data path.

## Environment

- Repo/branch: `jaseci-labs/jaseci` main.
- File: `jac/jaclang/scale/deploy/database/kubernetes_redis.jac` (redis container spec at line 301; hardcoded resources at lines 349-350).
- Observed live on: EKS `production-cluster` (us-east-2), namespace `jachammer` (a jac-builder/jac-scale app), Redis image `redis:7.2`, provisioned by jac-scale deploy.
- Also affects: every jac-scale-deployed app on that cluster gets the same 256Mi Redis (verified: 11 `*-redis` Deployments in the cluster, all from the same template).

Observed against the live jachammer Redis on production-cluster on 2026-07-08.

1. Container limit as deployed by jac-scale:

```
$ kubectl -n jachammer get deploy jachammer-redis \
    -o jsonpath='{.spec.template.spec.containers[0].resources}'
{"limits":{"cpu":"500m","memory":"256Mi"},"requests":{"cpu":"100m","memory":"128Mi"}}
```

2. Restart count and last termination reason:

```
$ kubectl -n jachammer get pod jachammer-redis-6df5456d6f-v425t
NAME                               READY   STATUS    RESTARTS       AGE
jachammer-redis-6df5456d6f-v425t   1/1     Running   92 (45m ago)   31d

$ kubectl -n jachammer get pod jachammer-redis-6df5456d6f-v425t \
    -o jsonpath='{.status.containerStatuses[0].lastState}'
{"terminated":{"exitCode":137,"reason":"OOMKilled",
 "finishedAt":"2026-07-08T11:18:28Z","startedAt":"2026-07-08T11:11:11Z"}}
```

92 restarts in 31 days; most recent `OOMKilled` (exit 137) ~45 min before inspection.

3. Redis' own view (rules out logical eviction as the cause):

```
$ redis-cli config get maxmemory        -> 209715200   (200 MB)
$ redis-cli config get maxmemory-policy -> allkeys-lru
$ redis-cli info memory | grep -E 'used_memory_human|used_memory_peak_human'
used_memory_human:122.59M
used_memory_peak_human:123.32M
```

Logical dataset is ~122M against a 200M `maxmemory`, so keys are **not** being evicted. The container RSS is what exceeds the 256Mi cgroup limit (fork COW during BGSAVE roughly doubles pages, plus fragmentation and buffers), so the kernel OOMKills the process.

4. Source confirming the values are not configurable (main):

```
jac/jaclang/scale/deploy/database/kubernetes_redis.jac
 62:  redis_max_memory = db_config.get('redis_max_memory', '200mb');   # only exposed knob
301:  'name': 'redis',                                                 # the redis container
349:  'requests': {'cpu': '100m', 'memory': '128Mi'},                  # hardcoded
350:  'limits':   {'cpu': '500m', 'memory': '256Mi'},                  # hardcoded
```

No `db_config.get(...)` lookup exists for the container `requests`/`limits`; grepping `[scale.database]` config keys (`config/impl/config_loader.impl.jac`) shows `redis_max_memory`, `redis_eviction_policy`, `redis_eviction_samples`, `redis_default_ttl`, `redis_max_connections`, `redis_enable_keyspace_notifications`, `redis_l1_invalidation_*`, but nothing for pod CPU/memory.

## Expected

Operators can size the provisioned Redis pod from `jac.toml` `[scale.database]`, e.g.:

```toml
[scale.database]
type = "redis"
redis_memory_limit   = "1Gi"
redis_memory_request = "512Mi"
redis_cpu_limit      = "1000m"
redis_cpu_request    = "250m"
```

so the container has enough headroom above `redis_max_memory` for fork/COW and Redis is not OOMKilled. Defaults should equal today's hardcoded values so existing deploys are unchanged.

## Actual

Container is pinned at `cpu 500m / memory 256Mi` for every app. When Redis RSS crosses 256Mi (readily, given a 200M dataset plus a BGSAVE fork), the pod is OOMKilled and clients see `Connection refused` until it restarts. Root-cause anchor: `jac/jaclang/scale/deploy/database/kubernetes_redis.jac:349-350`.

## Impact

- Every jac-scale-deployed app runs the same 256Mi Redis; any app whose working set + fork overhead exceeds 256Mi hits recurring OOMKills.
- User-facing: chat/build turns fail with `Error 111 ... Connection refused` and require a hard page refresh to recover (jacBuilder#634, hit by 2+ testers, high priority). Also plausibly implicated in jaseci#551 (Redis OOMKill wiping log streams).
- 92 restarts in 31 days on a single production Redis quantifies how often the window is open.

## Workaround (verified by inspection, not applied)

A live `kubectl set resources deployment/<app>-redis -c redis --limits=memory=1Gi,cpu=1000m --requests=memory=512Mi,cpu=250m` raises the cap, but a subsequent jac-scale redeploy re-templates the Deployment back to 256Mi, so it is not durable. (Not applied here.) The client-side mitigation for the symptom is an auto-retry of the chat-turn start on transient `Connection refused` (jacBuilder#634), but that does not address the root OOM.

## Suggested fix

In `jac/jaclang/scale/deploy/database/kubernetes_redis.jac`, read the redis container `requests`/`limits` from the database config (mirroring how `redis_max_memory` is already read at line 62), defaulting to the current `128Mi/256Mi` + `100m/500m` so behavior is unchanged unless overridden. Add the corresponding keys to the `[scale.database]` schema/defaults in `config/impl/config_loader.impl.jac` and `config/plugin_config.jac`.

Optionally validate/warn when `redis_max_memory` is set within a small margin of `redis_memory_limit` (fork COW needs headroom), so operators don't set `maxmemory >= limit` and reintroduce the OOM.

Regression test: extend the k8s deploy factory tests (`jac-scale/.../tests/deploy/`) to assert that setting `redis_memory_limit`/`redis_cpu_limit` in the database config propagates into the generated redis Deployment container `resources`, and that the defaults match the historical values when unset.

Related: jacBuilder#634 (client symptom), jaseci#551 (Redis OOMKill wiping log streams).

kashmithnisakya:
Fixes #7293.

## Problem

`jac-scale` provisions a Redis pod whose container CPU/memory requests and limits were **hardcoded** in [kubernetes_redis.jac](jac/jaclang/scale/deploy/database/kubernetes_redis.jac), with a memory limit of `256Mi`. The only memory-related knob exposed to `jac.toml`, `redis_max_memory` (default `200mb`), sets Redis' internal `maxmemory` eviction, **not** the container cgroup limit.

So once RSS crossed `256Mi` (readily: a ~122M working set, plus BGSAVE fork copy-on-write roughly doubling pages, plus fragmentation and buffers) the kernel OOMKilled the pod, and clients saw `Error 111 ... Connection refused` during each restart window. On `production-cluster` this was 92 restarts in 31 days on a single Redis. There was no `[scale.database]` key to raise the limit, and a live `kubectl set resources` was undone by the next redeploy re-templating the Deployment.

## Change

Expose four keys under `[scale.database]` and read them when building the Redis Deployment, mirroring how `redis_max_memory` is already read:

```toml
[scale.database]
type = "redis"
redis_cpu_request    = "100m"
redis_cpu_limit      = "500m"
redis_memory_request = "128Mi"
redis_memory_limit   = "2Gi"
```

### Exactly one default changes

The OOMKill is a **limit** problem. The kernel consults the cgroup limit, not the request, when it kills the process, and an unused limit reserves nothing. A request, by contrast, is a hard scheduler reservation held whether or not the pod uses it. CPU was never implicated at all.

So only `redis_memory_limit` moves. Everything else keeps its historical value, and no extra cluster capacity is reserved per app:

| key | before (hardcoded) | default now | changed? |
|---|---|---|---|
| `redis_cpu_request` | 100m | 100m | no |
| `redis_cpu_limit` | 500m | 500m | no |
| `redis_memory_request` | 128Mi | 128Mi | no |
| `redis_memory_limit` | **256Mi** | **1Gi** | **yes, this is the fix** |

`1Gi` clears the default `redis_max_memory` of `200mb` with room for the fork, so an unconfigured deploy no longer OOMKills. Operators can tune any of the four per app.

### Deploy-time headroom warning

Setting `redis_memory_limit` at or near `redis_max_memory` silently reproduces the original OOM, so `deploy()` now logs a warning when the limit is below roughly 2x maxmemory (the fork copy-on-write rule of thumb). A `_parse_size_to_bytes` helper handles both Redis units (`200mb`) and Kubernetes quantities (`512Mi`, `1Gi`).

The four values land in the Deployment verbatim, so the schema descriptions spell out that they must be valid Kubernetes quantities (`1Gi`, not `1gb`, which the API server rejects at apply time).

## Files

- [kubernetes_redis.jac](jac/jaclang/scale/deploy/database/kubernetes_redis.jac): read the four keys in `deploy()` and use them in the container `resources` block instead of literals; add `_parse_size_to_bytes` and the headroom warning.
- [config_loader.impl.jac](jac/jaclang/scale/config/impl/config_loader.impl.jac): add the keys to the `database` defaults and to `get_database_config()`.
- [plugin_config.jac](jac/jaclang/scale/config/plugin_config.jac): document the keys and their expected units in the `[scale.database]` schema.
- [test_deploy_k8s.jac](jac/jaclang/scale/tests/test_deploy_k8s.jac): four new tests.

## Verification

Four new tests, all passing:

- `redis pod resources default to configured values` asserts the defaults land in the generated Deployment.
- `redis pod resources propagate from [scale.database] config` asserts overrides reach the container `resources`.
- `redis warns when memory limit lacks headroom over maxmemory` asserts the old `256Mi` limit trips the warning against a `200mb` maxmemory, so the guard would have caught the original bug.
- `default redis memory limit does not emit a headroom warning` is the control, asserting `1Gi` stays silent.

Related: jacBuilder#634 (client-side `Connection refused` symptom), jaseci#551 (Redis OOMKill wiping log streams).


jac/jaclang/scale/deploy/database/kubernetes_redis.jac
```diff
@@ -198,6 +198,10 @@ obj KubernetesRedisProvider(DatabaseProvider) {
         redis_enable_keyspace_notifications = db_config.get(
             'redis_enable_keyspace_notifications', False
         );
+        redis_cpu_request = str(db_config.get('redis_cpu_request', '250m'));
+        redis_cpu_limit = str(db_config.get('redis_cpu_limit', '1000m'));
+        redis_memory_request = str(db_config.get('redis_memory_request', '512Mi'));
+        redis_memory_limit = str(db_config.get('redis_memory_limit', '1Gi'));
```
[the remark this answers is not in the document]
kashmithnisakya: good catch, dropped the fallbacks and read the four keys via subscript since get_database_config() already guarantees them. removes the third source of truth.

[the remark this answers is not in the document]
kashmithnisakya: added a deploy-time warning when redis_memory_limit is under ~2x redis_max_memory (needs headroom for the BGSAVE fork), plus tests covering the warn and no-warn cases.

jac/jaclang/scale/config/impl/config_loader.impl.jac
```diff
@@ -30,7 +30,11 @@ impl JacScaleConfig.get_default_config(self: JacScaleConfig) -> dict[str, any] {
             'redis_max_connections': 20,
             'redis_enable_keyspace_notifications': False,
             'redis_l1_invalidation_enabled': True,
-            'redis_l1_invalidation_channel': 'jac:anchor:invalidate'
+            'redis_l1_invalidation_channel': 'jac:anchor:invalidate',
+            'redis_cpu_request': '250m',
+            'redis_cpu_limit': '1000m',
+            'redis_memory_request': '512Mi',
```
udithishanka: the memory request jumping 128Mi to 512Mi is the part i'd question. the OOM was a limit problem, RSS spiking during the BGSAVE fork, and steady-state on the live jachammer pod was only ~122M. raising the limit to 1Gi is what actually fixes it. the request though is a hard scheduler reservation, so 512Mi per redis times every provisioned app (11 on production-cluster right now) is a chunk of capacity the scheduler can't hand to anything else. any reason not to keep the request modest, say 128-256Mi, and let just the limit carry the fork headroom? wondering if the request needs to move at all.

jac/jaclang/scale/config/plugin_config.jac
```diff
@@ -91,6 +91,26 @@ class JacScalePluginConfig {
...
+                        },
+                        "redis_memory_request": {
+                            "type": "string",
+                            "default": "512Mi",
+                            "description": "K8s memory request for the jac-scale provisioned Redis pod."
+                        },
+                        "redis_memory_limit": {
+                            "type": "string",
+                            "default": "1Gi",
+                            "description": "K8s memory limit (cgroup) for the jac-scale provisioned Redis pod. Must exceed redis_max_memory with headroom for BGSAVE fork copy-on-write, fragmentation, and buffers, or the pod is OOMKilled."
```
udithishanka: worth calling out here that this value lands in the k8s manifest verbatim, so it has to be a valid k8s quantity like 1Gi or 512Mi. _parse_size_to_bytes will happily accept a lowercase 1gb for the headroom check, but the api server rejects that at apply time, so someone could set it, get no warning, and then hit a confusing manifest failure. a quick 'e.g. 1Gi' in the description would steer people to the right units.

udithishanka (commented):
two notes, both non-blocking. nice that this reads the keys the same way redis_max_memory is already read.

jac/jaclang/scale/config/impl/config_loader.impl.jac
kashmithnisakya: agreed, reverted. request stays 128Mi and cpu stays 100m/500m, only the limit moves to 1Gi since thats the value the kernel checks on OOMKill. nothing extra reserved per pod now.

jac/jaclang/scale/config/plugin_config.jac
kashmithnisakya: good point, added the unit hint. descriptions now say it goes into the manifest verbatim and must be a valid k8s quantity like 1Gi, not 1gb.

udithishanka (approved):

PR #7316, opened 2026-07-09
fix(jaclang): resolve TypeVar scope through detached impl blocks

### LINKED ISSUE #7315: Type checker: class type parameter `[T]` not in scope inside separate `impl` blocks (false E1065)
# Bug: Class type parameter `[T]` is not in scope inside separate `impl` blocks (false E1065)

## Summary

When a generic archetype is declared with a class type parameter (`obj Box[T]`) and its methods are implemented in a **separate `impl` block**, the type checker reports `error[E1065]: Type variable "T" has no meaning in this context` for every use of `T` inside that `impl`. The type variable `T` clearly belongs to the target class `Box[T]`, so it should be in scope in `impl Box.<method>`.

The error is a **static-checker false positive**: the program compiles and runs correctly (`jac run`), because generic type parameters are erased at runtime. Only `jac check` fails.

This is already tracked in-tree as **"Root Cause 10"** — see the fixture `jac/tests/compiler/passes/main/fixtures/checker/checker_typevar_impl.jac`, whose docstring says: *"TypeVar T defined at module level should be valid in impl blocks. Currently the checker reports E1065."*

## Environment

- **jaclang version:** 0.30.9 (dev mode, compiler source in `jac/`)
- **Diagnostic:** `E1065` — "Type variable \"{name}\" has no meaning in this context" (`docs/docs/reference/diagnostics.md:222`)

## Reproduction

`ll.jac`:

```jac
obj Box[T] {
    has item: T;
    def flatten(items: list[T | None]) -> list[T];
}

impl Box.flatten(items: list[T | None]) -> list[T] {
    out: list[T] = [];
    for it in items {
        if it is not None {
            out.append(it);
        }
    }
    return out;
}
```

### `jac check ll.jac` → FAILS

```
✖ Error: error[E1065]: Type variable "T" has no meaning in this context
  --> ll.jac:6:30
    6 | impl Box.flatten(items: list[T | None]) -> list[T] {
      |                              ^
✖ Error: error[E1065]: Type variable "T" has no meaning in this context
  --> ll.jac:7:15
    7 |     out: list[T] = [];
      |               ^
ll.jac - 2 errors
```

### `jac run` → PASSES (runtime is correct)

Adding a call and running:

```jac
with entry {
    b = Box[int](item=1);
    print(b.flatten([1, None, 2]));   # prints: [1, 2]
}
```

```
$ jac run ll.jac
[1, 2]
```

Runtime works whether or not the method is ever called — the failure is purely in `jac check`.

## Expected behavior

`jac check` should accept `T` inside `impl Box.flatten(...)` (both in the signature **and** in the body), because the `impl` targets `Box[T]`, whose class type parameter list declares `T`. No E1065 should be emitted.

## Root cause

The emit sites are in `TypeEvaluator.validate_typevar_scope`:
- `jac/jaclang/compiler/type_system/type_evaluator.impl/evaluator_util_methods.impl.jac:1448`
- `jac/jaclang/compiler/type_system/type_evaluator.impl/evaluator_util_methods.impl.jac:1460`
- `jac/jaclang/compiler/type_system/type_evaluator.impl/evaluator_util_methods.impl.jac:1464`

To decide whether `T` is legal, `validate_typevar_scope` walks the **lexical parent chain** of the usage to find an enclosing class / function (`evaluator_util_methods.impl.jac:1428-1438`):

```jac
enclosing_class: uni.Archetype | None = None;
enclosing_func: uni.Ability | None = None;
current: uni.UniNode | None = expr.parent;

while current {
    if isinstance(current, uni.Archetype) and enclosing_class is None { enclosing_class = current; }
    if isinstance(current, uni.Ability)   and enclosing_func  is None { enclosing_func  = current; }
    current = current.parent;
}
```

A standalone `impl Box.flatten` block is lexically a **module-level sibling** of `obj Box[T]`, not nested inside it. So walking up from the `impl` never reaches `Box`; `enclosing_class` stays `None`, and the class's `[T]` parameter list is never consulted. The checker only inspects the impl ability's *own* type params (which don't contain `T`) and then falls through to `emit_diag(E1065, ...)`.

**The resolution is purely lexical and never links the `impl` back to its target archetype (`Box`).**

## The existing module-level `TypeVar` workaround is only half-implemented

`validate_typevar_scope` has a special case (`evaluator_util_methods.impl.jac:1409-1422`) for a module-level `glob T = TypeVar("T")`: it returns "valid" **only when the usage is inside a `FuncSignature`**. This means the workaround covers the signature but not the body.

Test with `glob T = TypeVar("T")`:

```jac
import from typing { TypeVar }
glob T = TypeVar("T");

obj Box[T] {
    has item: T;
    def flatten(items: list[T | None]) -> list[T];
}

impl Box.flatten(items: list[T | None]) -> list[T] {   # signature: OK
    out: list[T] = [];                                  # body: STILL E1065
    for it in items {
        if it is not None { out.append(it); }
    }
    return out;
}
```

Result — the signature use of `T` is accepted, but the body use still errors:

```
✖ Error: error[E1065]: Type variable "T" has no meaning in this context
  --> tv_glob.jac:10:15
   10 |     out: list[T] = [];
```

## Proposed fix

In `TypeEvaluator.validate_typevar_scope`:

1. **Link `impl` blocks to their target archetype.** When the usage is inside an `impl` ability, resolve the impl's target archetype (`Box`) and include *its* class type parameters when checking whether `T` is in scope — instead of relying solely on the lexical parent chain (`evaluator_util_methods.impl.jac:1428-1438` / `_get_enclosing_class_type_params`).
2. **Extend the module-level `TypeVar` allowance** (`evaluator_util_methods.impl.jac:1409-1422`) from signature-only to the entire ability body, so body uses of a valid module-level `TypeVar` are also accepted.

```

### PR DESCRIPTION
**Pull request by SandeepaHWP**
## Problem

Closes #7315.

A generic archetype whose methods are implemented in a **separate `impl` block** reports a false `error[E1065]: Type variable "T" has no meaning in this context` for every use of `T` in that block:

```jac
obj Box[T] {
    has item: T;
    def flatten(items: list[T | None]) -> list[T];
}

impl Box.flatten(items: list[T | None]) -> list[T] {   # E1065
    out: list[T] = [];                                  # E1065
    for it in items {
        if it is not None { out.append(it); }
    }
    return out;
}
```

`jac run` executes this correctly (generic parameters are erased at runtime); only `jac check` fails. The type variable plainly belongs to `Box[T]`, so it should be in scope in `impl Box.flatten`.

## Root cause

`TypeEvaluator.validate_typevar_scope` resolved type-variable scope **purely lexically**, walking `expr.parent` to find an enclosing `Archetype` or `Ability`.

A detached `impl Box.flatten` is an `ImplDef` that is lexically a **sibling** of `obj Box[T]`, so the parent chain is:

```
FuncSignature -> ImplDef -> Module      (signature)
Assignment    -> ImplDef -> Module      (body)
```

Neither an `Archetype` nor an `Ability` ever appears. Both `enclosing_class` and `enclosing_func` stayed `None`, and the function fell through to the final `emit_diag(E1065, ...)`. The checker never linked the `impl` back to the declaration it implements.

## Fix

- Track the enclosing `ImplDef` and resolve `decl_link` to the declaration it implements, honoring the type parameters of both the ability and its owning archetype (`method_owner`). This puts `T` in scope in the impl's signature **and** body.
- Also consult `method_owner` in the `enclosing_func` branch, so a class type parameter is accepted at use sites outside the method signature (e.g. `y: list[T] = [];` in a method body).

## Not a blanket suppression

Genuinely unbound type variables still report E1065. Verified:

```jac
glob T = TypeVar("T");
glob x: list[T] = [];        # E1065 (module level)

obj Plain { has bad: list[T]; }   # E1065 (non-generic archetype)
```

An earlier iteration also relaxed the module-level-`TypeVar` allowance from the signature into the body. That was **reverted**: `typeVar3.jac` correctly requires a module-level TypeVar used in a function body to be bound by the signature or the enclosing class, and the relaxation silently dropped 2 expected errors. The `ImplDef` resolution alone covers the impl-body case, since `_get_function_type_params` picks the TypeVar up from the declaration's signature.

## Tests

Adds `tests/compiler/passes/main/fixtures/checker/typevar/typevar_impl_scope.jac` and a `typevar_impl_scope` test asserting exactly one E1065 (the unbound negative guard). Confirmed the test **fails without the fix** (5 errors instead of 1).

- `test_typevar.jac`: 10 passed
- `test_checker_pass.jac`: 237 passed
- `jac check ll.jac` passes; `jac run` prints `[1, 2]`


SandeepaHWP:
> **Generic Method Params Are Missed** — When a detached impl targets a generic ability whose type parameter is declared on the ability but only used in the impl body, this branch only checks TypeVars collected from the declaration signature.

Good catch, and confirmed. Fixed in 03bdb4ae4.

The mechanism is exactly as described: `_get_function_type_params` collected type variables only by walking the signature (parameter type tags + return type) and never read `func.type_params`. So `def make[U] -> int;` with `U` used solely in the body fell through to E1065.

One correction on scope: this was **not** introduced by the `ImplDef` branch. The same helper backs the pre-existing `enclosing_func` path, so an *inline* method had the identical bug. Reproduced before the fix:

```jac
obj Box {
    def make[U] -> int {        # inline: E1065 on `U`
        held: list[U] = [];
        return len(held);
    }
}
```

Since the new branch mirrored existing semantics rather than diverging from them, I fixed it at the root instead of patching the impl path: `_get_function_type_params` now also collects the ability's declared type parameters. That makes the inline and detached-impl paths agree, and it fixes both cases in one place.

Regression coverage added to `typevar_impl_scope.jac` for both shapes (`Gen.inline_only[U]` and `Gen.impl_only[U]`, each using `U` only in the body). The fixture still asserts exactly one E1065, from the unbound negative guard, so this is not a suppression:

```jac
obj Plain { has bad: list[_T]; }   # still E1065
```

`test_typevar` 10 passed, `test_checker_pass` 238 passed.

SandeepaHWP:
Closing this: the bug it fixes has since been fixed on `main` independently, and more generally than this PR does.

`validate_typevar_scope` no longer resolves type-variable scope lexically. It now delegates to `_type_var_bound_in_scope()`, which handles `ImplDef` by following `decl_link` back to the implemented `Ability` and on to its owning `Archetype` -- the same resolution this PR adds, plus the impl's own signature type params. The `_get_enclosing_class_type_params` / `_get_function_type_params` pair this PR extended was replaced by `_get_signature_type_params`.

Verified by running this PR's own fixture (`typevar_impl_scope.jac`) against clean `main`: it reports **exactly 1 error**, the intentional unbound-`_T` guard on the non-generic `Plain` archetype. That is precisely what this PR's test asserts after its fix, so every false `E1065` it was filed for is already gone. Every hunk of the production diff also landed inside functions `main` has since rewritten, so there was nothing left to merge.

The one thing here that `main` does not have is the regression coverage -- the `typevar_impl_scope.jac` fixture and its `typevar_impl_scope` test, which cover the class type parameter in an impl signature and body, a signature-bound module-level `TypeVar`, an ability-declared `[U]` both inline and through a detached impl, and the unbound negative guard. Happy to open a test-only PR for that if it is wanted; it passes on current `main` as written (with `TypeCheckPass(ir_in=..., prog=...)` updated to `run_pass(TypeCheckPass, mod, program)`, since the bare constructor no longer runs the pass).

Note that #7315 is still open and can be closed as fixed.

PR #7317, opened 2026-07-09
fix(release): draft-then-publish so Latest never shows without assets

Thamirawaran:
## What

Cut the GitHub Release as a draft, attach every native binary, then publish it
as latest in one API call, so the "Latest" release is never visible while its
assets are still being built.

## Why

Previously the release was created and marked "Latest" before the binaries were
built. Anyone running `install.sh` during the build window (or browsing the
Releases page) could see a "Latest" release with missing or partial assets, and
`install.sh` would fail to resolve/download the binary.

## How

Only touches `.github/workflows/release.yml`:

- `tag-and-release` now creates the release as a **draft** (no `--latest`). A
  draft is hidden from `/releases/latest` and its assets are not anonymously
  downloadable, so `install.sh` cannot resolve it until it is complete.
- A new `publish` job polls until all three platform binaries are attached to
  the draft (distinguishing a transient release-API error from missing assets),
  then flips it to published + latest in a single `gh release edit` call. It
  reads `isDraft` first and skips the flip if the release is already published,
  so a re-run (or an `action=publish` dispatch from an older commit) can never
  drag the `latest` pointer backward onto an older version.

Completeness is guaranteed by `needs: build-binaries`: if any platform leg
fails, `publish` is skipped and the draft is never revealed, so `/releases/latest`
keeps pointing at the previous good release.

install.sh's real `curl | bash` path stays verified out-of-band by nightly.yml
`installer-live` and ci.yml `installer-test`, not as a blocking release gate.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,8 +2,9 @@ name: Release
 
 # The release lifecycle in one workflow:
 #   - action=create-pr (dispatch): bump versions and open the release PR.
-#   - merge of a release/* PR (or action=publish): tag, cut the GitHub Release,
-#     and build + attach the native binaries (build-binaries.yml).
+#   - merge of a release/* PR (or action=publish): tag, cut a DRAFT Release,
+#     attach every native binary (build-binaries.yml), then publish as latest --
+#     so "Latest" never appears mid-build with assets missing.
 #   - action=create-github-release (dispatch): recovery -- cut a bare Release
 #     for an existing version; binaries attach via the release.published event.
 # jaclang ships as the native `jac` binary, not PyPI.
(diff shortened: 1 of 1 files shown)


Thamirawaran:
Addressed the Greptile review:

- **P1 (verify-install can fail an already-live release):** removed the post-publish `verify-install` job. install.sh's real `curl | bash` path is verified out-of-band by nightly.yml `installer-live` and ci.yml `installer-test`, so this drops a flaky release gate (unauthenticated GitHub API, macOS anon rate limit) without losing coverage. This also matches the reasoning behind removing the in-leg install verify from build-binaries.yml.

- **P2 (asset poll hid API failures):** the poll now folds gh's stderr into the captured output and branches on its exit status, so a transient release-API error is logged and retried as an API error (and reported as such on timeout), instead of being misreported as missing assets.

PR #7319, opened 2026-07-09
fix(checker): trust bare-generic-default any in strict assignability checks (#6761)

### LINKED ISSUE #6761: Type checker: inline key=lambda rejected (E1054) for sorted/.sort over list[dict]; named key fn works
## Problem

Sorting a `list[dict]` by a field with an **inline `key=lambda`** fails type-checking with `E1054: No matching overload found`, even though a **named key function** with the same signature is accepted. This affects both `.jac` (server) and `.cl.jac` (client). The error message gives no hint about the real cause.

## Repro (`jac check`)

```jac
with entry {
    rows: list[dict] = [{"n": 2}, {"n": 1}];
    a = sorted(rows, key=lambda d: dict : d["n"]);   # error[E1054]: No matching overload
}
```
Same failure with the method form `rows.sort(key=lambda d: dict : d["n"]);` and with an explicit cast in the lambda body (`(d["n"] as int)`).

## What works (the asymmetry)

```jac
def keyfn(d: dict) -> int { return d["n"] as int; }
a = sorted(rows, key=keyfn);                  # OK - named key function
b = sorted([3, 1, 2], key=lambda x: int : x); # OK - inline lambda over list[int]
c = sorted(rows);                             # OK - no key
```

So inline lambdas are accepted as `key` over `list[int]`, and named functions are accepted as `key` over `list[dict]` - only the combination "inline lambda key + `list[dict]`" fails.

## Expected

`sorted(rows, key=lambda d: dict : d["n"])` should type-check (the runtime supports `sorted(iterable, key, reverse)`), or at least produce a clearer diagnostic than "no matching overload".

## Impact

Sorting records by a field is a very common operation, and the natural inline-lambda form is rejected with an opaque error. In practice an AI/code-gen agent (and humans) burn several iterations guessing at the lambda's type annotation before giving up.

## Workaround

Extract a named key function with an explicit comparable return type, then pass it as `key`.

Developer-Linus:
## Summary

Closes #6761. 

`sorted()` and `.sort()` reject an inline `key=lambda d: dict : d["n"]` over `list[dict]` with an opaque `E1054: No matching overload found`, even though a named key function with the identical signature works fine. The root cause lives in the type checker's handling of bare (unparameterized) generics, not in `sorted()`'s overload matching itself, so the fix is general rather than scoped to `sorted`/`.sort`.

## Problem

```jac
rows: list[dict] = [{"n": 2}, {"n": 1}];
a = sorted(rows, key=lambda d: dict : d["n"]);   // error[E1054]
rows.sort(key=lambda d: dict : d["n"]);          // same error
```

A named key function with the same signature works, and the same lambda pattern works over `list[int]`. Only the combination of an inline lambda key and a bare `dict` fails.

## Root cause

`build_type_var_solution` already defaults a bare generic's unresolved type parameters (`dict`'s `_KT`/`_VT`) to `any`, which is correct behavior. The actual defect is downstream: `assign_type`'s strict mode rule, which applies to every native Jac module by default, only trusts an `any` value against a non type variable destination if it is marked `is_foreign`, a flag reserved for values originating from untyped Python or JS interop. The `any` produced for an unresolved bare generic parameter was never marked that way, so strict mode treated it like a user hand wrote `Any`, and correctly (but wrongly, in this case) rejected it against `sorted`'s `Callable[[_T], SupportsRichComparison]` overload.

## Fix

- `build_type_var_solution` (`type_utils.impl.jac`) now marks the substituted `any` as `is_foreign=True`. This single line resolves the bug and applies to every bare generic, not just `dict` or `sorted`.
- `parameter_type_check.impl.jac` adds a second diagnostic path for `E1054`, `_unresolved_lambda_body`, for the case where a lambda argument's return type genuinely stays `Unknown` and is not just a bare generic default. It is guarded against two false positive cases found during review: an unrelated argument count or keyword mismatch, and an unresolvable parameter annotation. Both previously would have been misattributed to the lambda body.
- Extracted `_lambda_args_of` to remove duplicated boilerplate between the two lambda diagnostic helpers.

## Files changed

| File | Change | Lines | Purpose |
|---|---|---|---|
| `jac/jaclang/compiler/type_system/impl/type_utils.impl.jac` | Modified | +1 / -1 | Primary fix, marks bare-generic default `any` as `is_foreign` |
| `jac/jaclang/compiler/type_system/type_evaluator.impl/parameter_type_check.impl.jac` | Modified | +59 / -4 | Adds `_unresolved_lambda_body`, guards against false positives, adds `_lambda_args_of` |
| `jac/jaclang/compiler/type_system/type_evaluator.jac` | Modified | +2 / -0 | Interface declarations for the two new helpers |
| `jac/tests/compiler/passes/main/test_checker_pass.jac` | Modified | +119 / -0 | New and updated test cases |
| `jac/tests/compiler/passes/main/fixtures/checker/checker_bare_generic_lambda_key.jac` | Added | 21 | Primary repro, `sorted`/`.sort` with bare `dict` lambda key |
| `jac/tests/compiler/passes/main/fixtures/checker/checker_bare_generic_lambda_key_negative.jac` | Added | 16 | Negative control, whole `dict` as key must still fail |
| `jac/tests/compiler/passes/main/fixtures/checker/checker_bare_generic_any_direct_strict_call.jac` | Added | 23 | Same bug reached through a plain direct call, not `sorted` |
| `jac/tests/compiler/passes/main/fixtures/checker/checker_bare_generic_any_attr_chain.jac` | Added | 24 | Attribute access on bare-generic default `any` propagates correctly |
| `jac/tests/compiler/passes/main/fixtures/checker/checker_bare_generic_lambda_key_cast.jac` | Added | 20 | Explicit cast variant named in the issue repro |
| `jac/tests/compiler/passes/main/fixtures/checker/checker_lambda_return_unresolved.jac` | Added | 22 | Genuinely unresolved lambda body, unrelated to bare generics |

## Tests

| Test | Fixture | Confirms |
|---|---|---|
| `bare_generic_lambda_key_type_checks` | `checker_bare_generic_lambda_key.jac` | Primary repro type-checks |
| `bare_generic_lambda_key_whole_dict_still_errors` | `checker_bare_generic_lambda_key_negative.jac` | Fix does not mask a genuine `SupportsRichComparison` mismatch |
| `bare_generic_any_direct_strict_call_type_checks` | `checker_bare_generic_any_direct_strict_call.jac` | Fix is general, not `sorted` specific |
| `bare_generic_any_attr_chain_propagates_as_any` | `checker_bare_generic_any_attr_chain.jac` | Attribute chains on the trusted `any` resolve end to end |
| `bare_generic_lambda_key_cast_still_type_checks` | `checker_bare_generic_lambda_key_cast.jac` | Explicit cast variant from the issue keeps working |
| `lambda_return_unresolved_gets_help_text` | `checker_lambda_return_unresolved.jac` | Genuinely unresolved lambda body gets a specific `E1054` help text |

## Design notes

- `is_foreign` is reused rather than adding a new flag on `AnyType`. It has exactly two read sites in the type checker, attribute chain propagation and the strict mode assignability check. Both were audited and both behave correctly under the new meaning. This trades a small amount of semantic precision, the flag now means "trust this any" rather than only "from untyped interop", for not introducing a new field.
- `sorted(rows, key=lambda d: dict : (d["n"] as int))`, also named in the issue, already type-checked before this fix and is covered here as a plain regression test.
- `.cl.jac` client modules go through the same `TypeEvaluator` pipeline as `.jac`, so no separate fix was needed for the client side case mentioned in the issue.


MalithaPrabhashana:
Some overlap here worth a maintainer call before this moves.

**On current main (`3f7fd4f46`) the #6761 repro type-checks:**

```jac
rows: list[dict] = [{"n": 2}, {"n": 1}];
a = sorted(rows, key=lambda (d: dict) { d["n"] });   // no error
```

Negative controls (`key={ d }`, `key={ Foo() }`) still error correctly, so it is not a blanket loosening.

Two things shifted since this was opened on Jul 9:

- The `assign_type` strict-mode block this patch targets was removed by #7418 (Jul 14). `is_foreign` is no longer read anywhere in `assign_type` on main, so the `type_utils.impl.jac` one-liner has no consumer left on that path.
- Lambda syntax is now `lambda (d: dict) { ... }`. The six new fixtures use the old `lambda d: dict : ...` form and fail to parse with E0022.

Separately, #7459 (@SandeepaHWP) also references #6761 and covers the E1054 help-text half, including the same unannotated-lambda message updated to current syntax. That overlaps `parameter_type_check.impl.jac` and `test_checker_pass.jac` here.

@kugesan1105 @RandiJay96 could you two check how #7319 and #7459 should be reconciled, and what if anything from here is still worth rebasing?

@Developer-Linus the root-cause writeup in this description is solid and was correct against main as of Jul 9 - this is five weeks of churn under an unreviewed PR, not a problem with the analysis.

---

**Related**

- Issue this PR addresses: #6761
- Overlapping PR (E1054 help text): #7459
- Refactor that removed the targeted `assign_type` path: #7418

marsninja:

Issue #7291, opened 2026-07-08
Jac Scale: jac start --scale deploy time is dominated by in-pod pip install + gateway admin build, not app compile (persists even with #7254 fixed)


udithishanka:
## Summary

A cold `jac start --scale` of a 5 service app (this_is_jac) from a Mac takes ~10 min. It is tempting to blame the skipped host side app precompile (#7254), but measuring the actual pod startup shows the app compile is NOT the dominant cost. Even if precompile/`.jab` worked perfectly, the deploy would stay roughly this slow, because the fleet rollout is dominated by two things sealing does not touch:

1. in-pod `pip install` of the serve + app dependencies: **163 to 227s per pod**
2. the gateway building its admin client from source on boot: **~148s**

This issue documents the measured breakdown so the real levers (warm pod pool, prebuilt admin `_dist`) are visible, and so #7254 is prioritized as the seconds-per-pod win that it is rather than the deploy-time fix.

## Environment

| | |
|---|---|
| jac | 0.30.9 (rolling `dev` / main HEAD) |
| Operator host | macOS arm64 |
| Target | EKS (nodes x86_64), microservice app with 5 deployments + gateway |
| Command | `jac start --scale main.jac` |

## Measured breakdown (executed)

Full deploy timed with per-line elapsed timestamps (PTY wrapper around `jac start --scale`); pod internals from `kubectl get pod <p> -o json` init/main container and condition timestamps.

Operator side (`jac start --scale`, total **617.5s**):

| phase | window | duration |
|---|---|---|
| namespace + Redis (SA/ConfigMap/Deployment) | 0 to 10.6s | ~11s |
| client bundle build (bun/vite + `main.wasm`) | 10.6 to 24.9s | ~14s |
| seal / precompile attempt, **skipped** (#7254) | 24.9 to 58.5s | ~34s (wasted) |
| download rolling `dev` pod binary | 58.5 to 65.7s | ~7s |
| PVC bundle seed (1.36 MB source + **140.6 MB toolchain** to EFS RWX) | 70 to 169s | ~99s |
| apply 25 manifests (svc/deploy/HPA/PDB x6) | 169 to 196s | ~27s |
| **fleet rollout (pods install jac + boot)** | 196 to 617s | **~421s** |

Per pod (init/main container timings):

```
service pod (analytics):
  jac-pvc-bootstrap                                                    3s
  jac-bootstrap  (pip install fastapi, uvicorn, aiohttp, pydantic,
[...]

PR #7322, opened 2026-07-09
feat(scale): ship the admin console as a release asset, pulled at deploy like the binary

MusabMahmoodh:
## What this solves

Every stable-channel scale gateway spends ~148s (plus a bun dependency pull) rebuilding the admin console from source on every pod boot, restart, and HPA scale-up - the release binary ships without a prebuilt `_dist`.

## How

The console becomes a first-class release asset, downloaded at deploy time exactly like the jac binary:

**Release pipeline** - a new `admin-dist` job in `build-binaries.yml` runs after the binary legs, builds the console once (it is platform-independent html/js/css) with the linux-x86_64 binary the same run just produced (same-commit compiler, no skew), and attaches `jac-<version>-admin-dist.tar.gz` + `.sha256` to the release (`jac-dev-admin-dist.tar.gz` on the rolling dev prerelease).

**Deploy driver** - `stage_admin_dist` resolves the asset per channel with the same rules as the binary (fixed URL on dev to avoid the anonymous API quota, API resolution for versioned stable assets), verifies the sha256, caches it beside the binary cache, and extracts it into the app bundle at `.jac/admin` (temp-dir + rename, so a killed deploy can't leave a partial dir that passes the gateway's `index.html` sentinel). The shared download/verify/cache logic is factored into `_fetch_verified_asset`, now used by both the binary and the console. Only the driver touches the network - pods receive the console inside the bundle, so NetworkPolicies and pod egress are unaffected.

**Gateway** - unchanged: it already serves `.jac/admin` when `index.html` is present.

Best-effort degradation: releases published before this asset existed, a driver without network, or the local dev channel fall back to today's on-boot build - the pre-existing local-development behavior.

## Tests

Integration tests drive the real flow against a local HTTP server: download, checksum verification, extraction, and staging into the bundle; a checksum-mismatched asset is refused and nothing is staged; a fresh deploy stages from the verified cache after the asset disappears from the server; the local dev channel never touches the download path.

## Relation to #7302

Alternative implementation of the same goal (#7291 time audit: gateway main start -> Ready 163s -> 23s). #7302 bakes the console into the binary payload; this PR keeps binaries untouched and publishes the console separately. Merge one, close the other.


jac/jaclang/scale/injector/binary.jac
```diff
@@ -193,6 +211,109 @@ def pack_binary_bytes(binary_bytes: bytes) -> bytes {
...
+    # Same release-asset resolution as the binary: fixed names on the rolling
+    # dev prerelease (no API, its anonymous quota is exhausted on CI), and
+    # the API for version-named stable assets.
+    repo = "jaseci-labs/jaseci";
+    suffix = "-admin-dist.tar.gz";
+    if channel == "dev" {
+        url = f"https://github.com/{repo}/releases/download/dev/jac-dev{suffix}";
+        return (url, url + ".sha256");
+    }
+    api = f"https://api.github.com/repos/{repo}/releases/latest";
```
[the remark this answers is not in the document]
MusabMahmoodh: Good catch - applied in c217f7c54: the local channel returns no URLs and stages only the package's own `_dist`, with a test pinning that no network is touched.

jac/jaclang/scale/injector/binary.jac
```diff
@@ -193,6 +211,109 @@ def pack_binary_bytes(binary_bytes: bytes) -> bytes {
...
+        return _fetch_verified_asset(
+            dist_url,
+            checksum_url,
+            _admin_dist_cache_path(channel),
+            f"{channel} admin console"
+        );
+    } except Exception as e {
+        logger.warning(f"Could not download the admin console asset ({e}).");
+        return None;
+    }
```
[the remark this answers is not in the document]
MusabMahmoodh: Intentional: the console is an optional asset, so a broken/unreachable asset must not fail the deploy - the gateway's on-boot build is the correctness fallback and the warning carries the underlying error. Failing hard here would turn a cosmetic release problem into a deploy outage.

jac/jaclang/scale/injector/binary.jac
```diff
@@ -232,6 +250,103 @@ def pack_binary_bytes(binary_bytes: bytes) -> bytes {
...
+    repo = "jaseci-labs/jaseci";
+    suffix = "-admin-dist.tar.gz";
+    if channel == "dev" {
+        url = f"https://github.com/{repo}/releases/download/dev/jac-dev{suffix}";
+        return (url, url + ".sha256");
+    }
+    if channel == "local" {
+        # The binary comes from the developer's machine; a store-bought
+        # console could skew against it, so only the package _dist is used.
+        return ("", "");
```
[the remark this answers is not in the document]
MusabMahmoodh: Intentional as of 64b6c252b: staging is download-only by design, and nothing ships a package `_dist` anymore (the wheel-era build script and both bundled-dist consumers were removed in that commit), so there is no local package console to stage. For the local channel the designed behavior is the pre-existing gateway boot-time build from the `ui/` sources - a local dev flow where the ~2min build is acceptable and version skew against a developer-supplied binary is not.

jac/jaclang/scale/injector/binary.jac
```diff
@@ -133,84 +134,101 @@ def download_release_binary(channel: str, arch: str) -> bytes {
...
-            raise RuntimeError(
-                f"No `jac-*{suffix}` binary published for this arch on the "
-                f"{channel} release; set JAC_NODE_ARCH or check "
-                f"https://github.com/{repo}/releases."
-            );
-        }
-        raise;
-    }
+    payload = _http_get_bytes(url);
     if want {
```
kugesan1105, line 191: One thing that stood out: if `checksum_url` comes back empty here, `want` stays `""` and we just accept + stage the payload without any integrity check. For the binary path that can never happen (it hard-fails earlier without the asset), but `_admin_dist_urls` resolves `dist_url` and `checksum_url` independently from the release's asset list — so if a stable release ever ends up with the dist but a missing/misnamed `.sha256`, we'd pull the console over the network and extract it to disk unverified.

The pipeline always uploads both, so this is really guarding against a broken release rather than a live bug. But since this asset gets untarred straight onto the pod, could we refuse to stage when `what` is a downloaded (non-local) asset and no checksum resolved? Feels cheap and worth it for something that touches the filesystem.

jac/jaclang/scale/runtime/gateway/microservice_gateway.impl.jac
```diff
@@ -165,53 +165,20 @@ impl MicroserviceGateway.serve_spa_fallback -> Response | None {
...
-            } else {
-                shutil.copy2(item, dest);
-            }
-        }
+        built = self.build_admin_client();
     } except (ImportError, ModuleNotFoundError) {
         logger.debug("Admin module not available, skipping admin UI bootstrap");
+        return;
+    }
+    if not built {
```
kugesan1105: Small thing: `built` only gets assigned inside the `try`. The `except` catches `ImportError`/`ModuleNotFoundError` and returns, but if `build_admin_client()` throws anything else, we fall through to `if not built` with `built` never bound. The old code sidestepped this by returning inside each branch.

Build failures are normally caught internally and returned as `False` so this is unlikely to bite in practice — but a `built = False` before the `try` would make it airtight.

jac/jaclang/scale/injector/binary.jac
```diff
@@ -232,6 +250,97 @@ def pack_binary_bytes(binary_bytes: bytes) -> bytes {
...
+    suffix = "-admin-dist.tar.gz";
+    if channel == "dev" {
+        url = f"https://github.com/{repo}/releases/download/dev/jac-dev{suffix}";
+        return (url, url + ".sha256");
+    }
+    if channel == "local" {
+        # A downloaded console could skew against a developer-supplied binary.
+        return ("", "");
+    }
+    api = f"https://api.github.com/repos/{repo}/releases/latest";
```
kugesan1105, line 264: Not a blocker, just noting: this hits the unauthenticated `releases/latest` API, same as the binary resolver already does — so on stable we now make two anonymous API calls per deploy (binary + console) instead of one, sharing the same quota the dev channel deliberately dodges with a fixed URL. Fine as-is since it mirrors existing behavior, but worth keeping in mind if the anonymous quota ever gets tight on busy CI.

kugesan1105:
Validated the console build time on a GitHub-hosted `ubuntu-latest` runner (the same runner the `admin-dist` job uses) — ran the build+package 3× in a throwaway workflow, each a cold build:

| Run | Build + package | Tarball |
|-----|-----------------|---------|
| 1 (first) | 83s | 752K |
| 2 | 44s | 752K |
| 3 | 44s | 752K |

So the `admin-dist` job adds roughly **~45–85s** for the build+package (first run warm-up, ~44s steady-state), plus artifact download/upload overhead — comfortably under the ~148s on-pod build the PR replaces. The bun dependency pull the runner already had warm accounts for most of that gap. Artifact is tiny (752K), so attaching + pulling it at deploy time is cheap, which backs the approach here.

Measurement details / raw logs: #7347 (throwaway, already closed — nothing published).

jac/jaclang/scale/injector/binary.jac
kugesan1105: Addressed in 9559b1f — `download_admin_dist` now refuses to fetch/stage when a non-local channel resolves a dist URL but no checksum, so a release missing its `.sha256` falls back to the on-boot build instead of extracting an unverified asset. Added a test that asserts it stages nothing and never touches the network. Resolving.

jac/jaclang/scale/runtime/gateway/microservice_gateway.impl.jac
kugesan1105: Fixed in 9559b1f — dropped the `built` flag entirely and restructured to a plain try/except: quietly skip on a missing admin module, log any other build failure. No unbound-variable path left. Resolving.

jac/jaclang/scale/injector/binary.jac
kugesan1105: Resolving as acceptable-as-is: it mirrors the binary resolver's existing behavior and isn't a blocker. Leaving a note here so it's easy to revisit if the anonymous API quota ever gets tight on busy CI (e.g. caching the `releases/latest` response so binary + console share one call).

PR #7347, opened 2026-07-10
[DO NOT MERGE] temp: time the admin-console build in CI

kugesan1105:
**Do not merge / do not review.** Throwaway PR to measure how long the admin-console build+package (the `admin-dist` job from #7322) takes on a GitHub-hosted `ubuntu-latest` runner.

It downloads the existing `jac-dev-linux-x86_64` release binary, then builds + tars the console **3 times** in a loop and prints per-run timing. It uploads / releases / publishes **nothing**.

Will close and delete the branch once I have the numbers.

### DIFF
--- a/.github/workflows/time-admin-dist.yml
+++ b/.github/workflows/time-admin-dist.yml
@@ -1,11 +1,17 @@
-# TEMPORARY throwaway workflow — measures how long the admin-console
+# TEMPORARY / DO NOT MERGE — measures how long the admin-console
 # build+package (the `admin-dist` job in build-binaries.yml) takes.
 # Runs the real build steps 3x in a loop and prints per-run timing.
-# Does NOT upload, release, or publish anything. Delete after use.
+# Does NOT upload, release, or publish any artifact. Delete after use.
 name: time-admin-dist (temp)
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


kugesan1105:
## Admin-console build timing — 3 runs on `ubuntu-latest`

Measured how long the admin-console build+package (the `admin-dist` job from #7322) actually takes. Each loop clears `.jac` first so every run is a cold build.

| Run | Build + package | Tarball |
|-----|-----------------|---------|
| 1 (first) | **83s** | 752K |
| 2 | **44s** | 752K |
| 3 | **44s** | 752K |

Total job wall-clock ~3 min (incl. checkout + downloading the `jac-dev-linux-x86_64` binary).

**Takeaways**
- First build ~83s, steady-state ~44s. Run 1 is slower purely from toolchain warm-up.
- Comfortably under the ~148s cited for on-pod builds — that figure includes the bun dependency pull, which the runner already had warm here. The artifact-build portion the release job adds is really ~45–85s.
- Artifact is tiny (752K), so attaching it as a release asset and pulling it at deploy time is cheap — supports the #7322 approach.

So expect roughly **1–1.5 min** of added CI time for the console build+package in the real `admin-dist` job.

Closing — this was a throwaway measurement PR, nothing published.

PR #7352, opened 2026-07-10
feat(lang): add `forever {}` loop keyword

marsninja:
## Summary

Adds a `forever { ... }` statement to Jac — first-class sugar for an unconditional infinite loop, semantically equivalent to `while True { }`:

```jac
forever {
    do_work();
    if done { break; }
}
```

Following the codebase's existing "one node per surface form" convention (`InForStmt` / `IterForStmt` / `WhileStmt` are distinct siblings), `forever` is modeled as its own `ForeverStmt` AST node rather than by overloading `WhileStmt`. It carries **no condition** and **no `else` clause** (an infinite loop's `else` is unreachable), which keeps every node's contract clean and lets the kid-list-driven formatter round-trip `forever` for free.

## What's included

- **Lexer / parser**: `forever` keyword (`KW_FOREVER`) + `parse_forever_stmt`; reserved as a keyword (requires backtick-escaping as an identifier), and usable inside JSX/view bodies.
- **Codegen — all three backends**:
  - Python AST → `while True:`
  - Native / LLVM → unconditional branch into the loop body, with `break`/`continue` via the loop stack
  - JavaScript → `while (true)`
- **Passes**: symbol-table scope, CFG loop wiring, type-evaluator loop-header narrowing + infinite-loop divergence analysis, static-analysis unreachable-code descent, `break`/`continue` validation (E0058), formatter (DocIR) / legacy unparse / normalize, and langserve document outline.
- **Generated artifacts** regenerated with their tools: `jir_registry.jac` (`jac gen-jir-registry`) and `jac.spec` (`jac tool grammar`), adding `forever_stmt ::= "forever" "{" code_block_stmts "}"`.

## Design constraints (enforced by the parser)

- `forever true { }` → error (no condition)
- `forever { } else { }` → error (no `else`)
- `def foo(forever: int)` → error (reserved keyword)

## Tests

- Execution: `break`, `continue`, and nested `forever` (Python path)
- Native LLVM: `break` and `continue` return-value checks
- JS: emits `while (true)`
- Formatter: round-trip idempotency + asserts `forever` is preserved (not rewritten to `while True`)
- Grammar: `forever_stmt` structure + spec snapshot
- Parser negatives: condition / `else` / reserved-keyword-as-param all rejected

Targeted tests for all of the above pass; the full grammar-extract (24), formatter (30), and parser-validation (192) suites pass unchanged.


marsninja:
This one's for Xayden <3

PR #7359, opened 2026-07-10
fix(shadcn): use valid Jac module names across the component registry

MalithaPrabhashana:
## Problem

The bundled shadcn registry shipped primitives with **hyphenated file names** (`dropdown-menu.cl.jac`), and two of them imported a hyphenated peer internally:

```
combobox.cl.jac   import from ".input-group" { ... }
command.cl.jac    import from ".input-group" { InputGroup, InputGroupAddon }
```

The installer renames files `-` -> `_` on copy but writes their content verbatim (it never rewrites imports). So after `jac add --shadcn combobox`, the peer landed as `input_group.cl.jac` while `combobox` still imported `.input-group`. The compiler emits `./input-group.js`, which does not exist:

```
Could not resolve "./input-group.js"
```

So `combobox` and `command` did **not build out of the box**, with no agent involved. `jac check` passes; only the bundler catches it. Separately, neither declared `input-group` as a peer, so it was never even installed.

## Fix

Make the registry correct at the source, so it uses valid Jac module names throughout (a hyphen is the minus operator and cannot be a Jac module name):

- Rename the 13 hyphenated component files to underscores.
- Update the matching `file` fields in `registry.json`.
- Fix the two internal `.input-group` imports to `.input_group`.
- Declare `input-group` as a peer of `combobox` and `command`.

Component keys and the `jac install --shadcn <name>` command keep their kebab-case names (they match the shadcn ecosystem); only file names and internal imports change. The installer's existing filename normalization stays as a defensive fallback.

This complements the docs fix in #7358 (which corrected the agent-facing guidance): that fixed what agents write, this fixes the primitives themselves.

## Verification

End to end, against the fixed registry: `jac add --shadcn combobox` now pulls `input-group` automatically and the app **builds clean** (the resolve error is gone).

Guard tests in `test_shadcn.jac`:
- **registry uses valid Jac module names throughout** - scans every registry source: no file name and no dot-relative import may contain a hyphen.
- **component with a peer resolves to a matching underscored import** - installs `combobox`, asserts `input_group.cl.jac` is pulled and the import matches.

Both are mutation-checked: reintroducing the hyphen fails exactly these two and nothing else. Full file: 32 pass.


MalithaPrabhashana:
Folded into #7358 — both fix the same shadcn hyphenated-import issue (docs + registry), so they read better as one PR. The registry rename, peer fix, and guard tests are all in #7358 now.

PR #7361, opened 2026-07-10
fix(checker): resolve enum metaclass data-attributes (__members__, _member_names_, …)

### LINKED ISSUE #7360: Type checker: false `error[E1030]` on enum metaclass attributes (`__members__`, `_member_names_`, …)
## Summary

`jac check` reports a false `error[E1030]: Type "<Enum>" has no attribute "__members__"` for
valid, running code that accesses standard enum attributes such as `__members__`,
`_member_names_`, `_member_map_`, and `_value2member_map_`. The program executes correctly
(`jac run` produces the expected output); only the type checker rejects it.

These attributes are real. They live on the enum **metaclass** (`EnumType`/`EnumMeta`), not in
the enum class's own MRO, and the checker does not search the metaclass when resolving an
attribute on a class object.

## Reproduction

```jac
import from enum { StrEnum }

class Tokens(StrEnum) {
    A = "a";
    B = "b";
}

def names() -> list[str] {
    out: list[str] = [];
    for k in Tokens.__members__ {
        out.append(k);
    }
    return out;
}

with entry { print(names()); }
```

### Actual behavior

```
$ jac run  repro.jac
['A', 'B']

$ jac check repro.jac
✖ Error: error[E1030]: Type "Tokens" has no attribute "__members__"
✖ Error: error[E1053]: Cannot assign <Unknown> to parameter 'object' of type str
```

The second error (`E1053`) is cascading fallout: once `__members__` resolves to `<Unknown>`,
the loop variable `k` becomes `<Unknown>`, so `out.append(k)` fails too.

### Expected behavior

`jac check` should pass. `Tokens.__members__` should resolve to `dict[str, Tokens]`
(runtime type is `mappingproxy[str, Tokens]`), so `for k in Tokens.__members__` yields `str`.

For reference, pyright accepts the equivalent Python and reports:

```
Type of "Tokens.__members__" is "MappingProxyType[str, Tokens]"
Type of "k" is "str"
0 errors
```

## Scope

The bug is not specific to `StrEnum` or to `__members__`. It affects:

- **All enum forms**: class-based `StrEnum` / `Enum` / `IntEnum`, and Jac-native `enum Foo { ... }`.
- **All metaclass data-attributes**: `__members__`, `_member_names_`, `_member_map_`,
  `_value2member_map_`.

Member-level access (e.g. `Tokens.A.__members__`) is correctly invalid — the metaclass is not in
an instance's MRO, and it raises `AttributeError` at runtime — so an error there is expected.

Attributes that already work (each hand-special-cased elsewhere) and are **not** affected:
`Tokens.A`, `.name`, `.value`, `len(Tokens)`, `Tokens["A"]`, `"a" in Tokens`, `list(Tokens)`,
iteration `for m in Tokens`.

## Related

This is a continuation of the enum-metaclass gaps already tracked and fixed incrementally in
`jac/tests/compiler/passes/main/test_checker_gaps.jac`:

- **Gap #45** — enum iteration false "not iterable" (EnumMeta `__iter__` not modeled) — fixed.
- **Gap #52** — enum iteration variable resolved to `Unknown` — fixed.

Those addressed `__iter__` and the iteration variable's type; the metaclass **data attributes**
(`__members__`, `_member_names_`, …) are the next unaddressed slice of the same gap.

## Environment

- jaclang 0.31.2
- Reproducible via `jac check` in dev mode against the in-repo compiler source.

SandeepaHWP:
## Summary

`jac check` reported a false `error[E1030]: has no attribute` when accessing the enum metaclass data-attributes `__members__`, `_member_names_`, `_member_map_` and `_value2member_map_`, even though the code runs correctly. These attributes live on the enum metaclass (`EnumType`), not in the enum's own MRO, so `_lookup_class_member` (which only walks the class MRO) misses them and the checker emits `E1030`.

Minimal repro that failed `jac check` but printed `['A', 'B']` under `jac run`:

```jac
import from enum { StrEnum }
class Tokens(StrEnum) { A = "a"; B = "b"; }
def names() -> list[str] {
    out: list[str] = [];
    for k in Tokens.__members__ { out.append(k); }
    return out;
}
with entry { print(names()); }
```

## Fix

Synthesize the metaclass data-attribute types in the class-object attribute branch, right after the `_lookup_class_member` miss and before the `E1030` emit, mirroring the existing enum iteration and `.name` / `.value` special cases:

- `__members__` / `_member_map_` -> `dict[str, <Enum>]`
- `_member_names_` -> `list[str]`
- `_value2member_map_` -> `dict[<value type>, <Enum>]`

The `.value` branch is refactored to reuse the extracted `_resolve_enum_value_type` helper (removes a duplicated block). No general metaclass modeling is introduced here; the general fix (a metaclass-MRO fallback on class-object attribute misses) is tracked in #8450.

Member-level access (e.g. `Tokens.A.__members__`) remains correctly rejected, matching the runtime `AttributeError`.

## Scope

Covers all enum forms (`StrEnum` / `Enum` / `IntEnum` and native `enum`) and all four metaclass data-attributes. This is the next slice of the enum-metaclass work: enum iteration and the iteration variable's type were addressed in #5329 and #5790 (tracked as the `gap 45` / `gap 52` tests in `test_checker_gaps.jac`); this PR handles the metaclass data-attributes.

### Scoped to enums for now: general metaclass lookup tracked in #8450

This PR deliberately does **not** add general metaclass modeling; it is the next slice of the enum-metaclass work from #5329 and #5790. Jac's grammar has no `metaclass=` syntax and user-defined metaclasses are an intentional non-feature of the language (`docs/docs/quick-guide/what-makes-jac-different.md:83`, `docs/docs/reference/language/functions-objects.md:496`), so nothing here expands the language surface.

However, the underlying checker gap is broader than enums, and an earlier revision of this description overstated the scope by claiming `EnumType` is the only metaclass the checker can reach. That is not accurate:

- Attributes of plain `type` (the default metaclass) are invisible on every class object: `Foo.__name__` emits a false `E1030` today.
- Stub-imported classes carry their declared metaclass into the checker through the documented `class` interop escape hatch: `class Base(ABC)` then `Base.register(Square)` runs fine but emits a false `E1030`, because `register` lives on `ABCMeta`.
- `Tokens["A"]` is still misread as generic specialization instead of `EnumMeta.__getitem__`, and the false error demotes the module out of native lowering.

The general fix (fall back to the metaclass MRO on a class-object attribute miss, reading the declarations typeshed already provides) is tracked in #8450. When it lands, the enum-specific synthesis added here should be deleted in its favor; #8450 records that cleanup. Modeling how the runtime resolves class attributes does not add metaclass support to the language, so this remains consistent with the design constraints above.

## Tests

- New `enum_metaclass_attr_type_checking` in `test_checker_pass.jac` with fixture `checker_enum_metaclass_attrs.jac`: valid `__members__` / `_member_names_` / `_member_map_` usage type-checks; one deliberate misuse (`bad: int = Tokens.__members__`) still errors, proving the synthesized type is real (`dict[str, Tokens]`), not `Any`.
- Full `test_checker_pass` (238) and `test_checker_gaps` (21) suites pass.

Fixes #7360


kugesan1105 (commented):
Validated this locally on the branch: the repro passes, the synthesized types are correct (typed assignments against `dict[str, Tokens]` etc. work, and the fixture's deliberate misuse still errors), the new test passes, and the `.value` refactor into `_resolve_enum_value_type` doesn't regress `enum_type_checking`, `class_based_enum_type_checking`, `enum_member_assignable_to_base_type`, or gaps 45/52. As a scoped fix this is solid.

Two things on the "Non-goal" section though, because I don't think it should be accepted as settled:

**The "EnumType is the only metaclass that ever reaches the type checker" claim is not correct.** `ABCMeta` reaches it through the same `class` interop escape hatch the PR cites:

```jac
import from abc { ABC, abstractmethod }

class Base(ABC) {
    @abstractmethod
    def area(self: Base) -> float;
}

class Square {
    def area(self: Square) -> float { return 1.0; }
}

with entry {
    Base.register(Square);  # error[E1030]: Type "Base" has no attribute "register"
}
```

This runs fine (`issubclass(Square, Base)` is `True` afterwards) but still emits E1030 on this branch, since `register` lives on `ABCMeta`.

**The underlying gap is not even enum-specific.** `_lookup_class_member` walks only the class's own MRO, so attributes of plain `type` (the default metaclass) are invisible on every class object:

```jac
obj Foo {}
with entry {
    nm: str = Foo.__name__;  # error[E1030]: Type "Foo" has no attribute "__name__"
}
```

Still false-flagged on this branch, all verified working under `jac run`:

- `Tokens["A"]`: misread as generic subscription, errors with `Cannot assign Tokens[Literal["A"]] to Tokens`, because `EnumMeta.__getitem__` is never found. This one also demotes the module out of native lowering.
- `Tokens.__name__` / `Tokens.__qualname__`: E1030.
- `Base.register(...)` as above.

Note that our vendored typeshed already models all of this: `enum.pyi` declares `__members__`, `_member_names_`, `_member_map_` and `_value2member_map_` on `EnumMeta(type)`, with `class Enum(metaclass=EnumMeta)`. So this PR hand-duplicates information the checker has on disk but never consults, and it's the third enum-metaclass special case after #5329 and #5790, with `__getitem__` next in line.

The root-cause shape would be: on a class-object attribute miss, fall back to the metaclass MRO (default `type`, `EnumMeta` for enums, the declared metaclass for stub-imported classes), binding the `self: type[_EnumMemberT]` style generics typeshed already provides. That one mechanism would subsume this synthesis and the earlier special cases.

I'm fine landing this as an incremental fix, but let's track the metaclass-MRO fallback as the real fix rather than treating general metaclass lookup as out of scope for the checker.

kugesan1105:
@SandeepaHWP  check the #8450 . followup

Issue #7397, opened 2026-07-12
RFC: ownership endgame — #7149 alignment asks + Phases 4–6 (nogc enforcement, headerless owned codegen, pay-for-use collector)


marsninja:
## Ultimate goal

The claim we want to be able to make — truthfully, and mechanically verifiably from the produced artifact:

> **In ownership-checked native Jac, memory management compiles to what Rust would emit: allocations and frees at statically determined points, no reference counting, no collector — and the *absence* of that machinery is checkable in the binary.**

#6905 / #7149 build the hard 80% of this (the use-site `own`/`&`/`&mut` surface, the CFG-based `OwnershipCheckPass`, a liveness-based `RcElisionProofPass`, and a leak-free managed floor via the Bacon–Rajan recycler). But as designed they can never *license the claim*, for a structural reason: **elision is opportunistic**. Any value the proof misses silently stays managed, so even a fully-annotated program still carries, and may execute, the RC/GC machinery:

- a **48-byte header on every heap object** (magic/alloc-id, trace-fn pointer, color word, type tag, refcount, dtor pointer — grown from 32 bytes by #7149),
- an **rc=1 stamp + trace-pointer store at every instantiation**,
- an **atomic acq_rel RMW on every non-elided release**, plus (post-#7149) **three global loads and the purple-buffer branch on the release hot path**, a buffered-bit check per free, and a mid-collection flag load per destructor,
- **collector + trace code linked into every binary**, regardless of whether `JAC_GC_CYCLES` is ever set.

That is a Swift/Nim-class ARC story — genuinely good, and the right *default* for Jac — but it is not a near-Rust story. This issue proposes the subtractive endgame on top of #7149: three phases that turn ownership from an optimization hint into an enforceable, verifiable guarantee. It formally revisits locked decision #6 of #6905 ("no header-free per-object allocation — revisit only on demonstrated demand"): the positioning goal above **is** the demonstrated demand.

The resulting ladder, each rung opt-in from the previous:

| Rung | What you write | Guarantee | How it's verified |
|---|---|---|---|
| Default (today + #7149 Phase 0) | nothing | managed, leak-free (ARC + recycler) | recycler regression tests |
| Annotated (#7149 Phases 1–2) | `own` / `&` / `&mut` where you care | checked semantics + elided RC where proven | elision-coverage stat (A3 below) |
[...]

PR #7398, opened 2026-07-12
feat(native): ownership alignment Part A (#7397) — 32-byte RC header, dtor-keyed trace lookup, emit-time --gc modes, own/mut keywords, rc-stats

marsninja:
Implements **Part A of #7397** — the pre-merge alignment asks for the ownership work — now rebased onto `main` as a single commit on top of the merged #7149 (`ac863a891`). The three asks trim the per-object cost the cycle collector introduced and lock the qualifier grammar before more lands on top of it.

## A1 — recycler costs gated at emit time, not just run time

#7149 gates cycle *collection* at runtime (`JAC_GC_CYCLES`, default off), but its costs were emitted unconditionally: a 48-byte header on every object (trace-fn pointer + color word), a per-object trace-pointer store at instantiation, and purple-buffer branches on every release. This removes all per-object cost while keeping the recycler fully functional:

- **RC header is back to 32 bytes.** Collector color/buffered state packs into spare high bits of the alloc-id header word — layout `[63:51 zero][50:48 buffered|color][47:32 magic 0xAC1D][31:0 alloc id]` — via the existing `_emit_rc_color_load/store` helpers (callers unchanged). The magic check masks to bits [47:32]. Fresh allocations are black/unbuffered for free, since the magic stamp already zeroes the high bits.
- **No per-object trace pointer.** Trace functions are per-*type*, so the per-object slot was redundant: collection-time dispatch resolves the trace function through `__rc_trace_lookup`, a compile-time-generated map keyed on the destructor pointer already stamped in the header (type-unique for both structs and containers). At `-O2` LLVM inlines the lookup chain and devirtualizes the trace calls.
- **`--gc` emit-time knob** (`CompileOptions.na_gc_mode`, `jac nacompile --gc <mode>`):
  - `cycles` (default) — full ARC + recycler, exactly #7149's behavior;
  - `rc` — plain ARC: no collector functions, no trace functions, no roots buffer, no purple-buffer branch in `__rc_release_simple`, no `JAC_GC_CYCLES` entry probe (IR-verified: zero collector symbols);
  - `none` — additionally emits no retain/release call sites at all (the emit-time analogue of `JAC_NO_GC`; memory is never reclaimed — IR-verified: zero RC call sites).

## A2 — qualifier spelling pinned in the grammar

`own` and `mut` are now reserved keywords (`KW_OWN`/`KW_MUT`); `ownership_prefix` no longer accepts arbitrary NAMEs (previously any identifier parsed as an ownership prefix in annotation position, and promoting contextual names to keywords later would have been a breaking change). `val` deliberately stays contextual — reserving it would break far more existing code, and it only ever appears in prefix position. `jac.spec` updated:

```
ownership_prefix ::= "own" | "val" | "&" "mut"?
ref ::= "&" ("(" await_expr | "mut"? await_expr) | await_expr
```

The handful of in-tree identifiers named `own`/`mut` (6 files) are renamed.

## A3 — elision coverage stat

`NaIRGenPass` counts retains/releases emitted vs RC ops elided (move-lowering hits + owned-move transfers into locals/globals). `JAC_RC_STATS=1` prints a per-module line:

```
rc-stats [gcprobe.na.jac] gc=cycles retains=1 releases=10 elided=3 coverage=21.4%
```

This is the bridge metric between opportunistic elision (#7149 Phase 2) and the enforcement mode proposed in #7397 Phase 4, and gives the Phase 2 benchmarks a denominator.

## Validation

Locally (before the rebase onto merged main): 128 ownership + parser tests pass; the full native-suite failure set was identical to the #7149 branch baseline in this environment (link-dependent native tests need a vendored musl runtime that isn't present here, so they fail on the base branch too); IR inspection of a cycle-heavy probe in all three `--gc` modes confirmed the symbol guarantees above. Relying on CI for the authoritative run on this base.

Refs #7397, #6905. Follows up #7149.


marsninja:
Pushed fixes for CI:

- **`jac fmt` (jac-check job)**: the A3 rc-stats call used `print()` (banned by the `[no-print]` lint rule) → switched to `sys.stderr.write`; ran `jac fmt --lintfix` on the changed files.
- **`test-compiler` — 3 regressions from the `own`/`mut` keyword promotion (A2)**, now fixed: a constructed-source string in `test_checker_ts_interface.jac` used `own` as a local name (renamed to `extra_val`), and `jac.spec`'s `ownership_prefix` rule now matches the grammar extractor (`"own" | NAME | "&" "mut"?` — the `val` branch extracts as `NAME`).
- **`test-compiler` — 4 pre-existing failures on `main`** (run 29199755490 is red on the same `test_rc_elision_proof.jac` cases): the `rc_elision_proof_cases.jac` fixture used pre-brace `lambda : a` syntax, which stopped parsing after the lambda-braces change landed. Fixed the fixture (`lambda { a; }`) here since it sits in this PR's native RC-elision area and otherwise blocks the job from going green; happy to split it out if preferred.

marsninja:
Correction to my previous comment on the `test_rc_elision_proof.jac` failures — I initially patched that fixture's lambda syntax, but backed it out after diagnosing the real cause; **this PR no longer touches it.**

The 4 `test_rc_elision_proof.jac` failures are **pre-existing on `main`** (run 29199755490 is red on the identical cases) and are two independent breakages unrelated to this PR:
1. The fixture `rc_elision_proof_cases.jac` uses pre-brace `lambda : a` syntax, which stopped parsing after the lambda-braces change (#6773/#7376) — so the test can't even compile the fixture.
2. Even with that fixed locally, `RcElisionProofPass` produces **no marks at all** for this compile (I dumped the full `na_move_lowerable` dict — every entry is `False`). The pass is native-schedule-only, and `move_marks()` compiles a plain `.jac` with `type_check=False`; after the native-backend gating (#7368) that path no longer runs the pass. So the positive assertions (`b1`/`b9`) can never hold here regardless of the lambda fix.

Both are #7149/#7368 interactions that deserve their own fix, so I've left the fixture untouched to keep this PR scoped to Part A. My own `own`/`mut`-keyword regressions (the `test_checker_ts_interface.jac` and grammar-extract failures) are fixed. Net effect: this PR's `test-compiler` failure set is identical to `main`'s.

PR #7408, opened 2026-07-13
fix(type): enhance call classification and symbol resolution in evaluator

### LINKED ISSUE #7394: Cross-module `obj` construction breaks in generated client JS
## Description

When an `obj` type is defined in one `.cl.jac` file and imported into another with a plain `import from`, then constructed from that other file, the generated JavaScript is wrong: it compiles to a bare function call instead of a proper `new` expression. Since the type always compiles to a real ES `class`, calling it without `new` is a guaranteed runtime crash.

This is `obj`'s normal, documented usage — it's the one archetype meant to be shared and constructed across client files. It's also **silent at every build step**: `jac check` passes with zero warnings, `jac build` succeeds. Nothing indicates a problem until a user's browser actually runs the code.

A closely related finding: whether you can even detect this bug depends on which command builds/serves the project. `jac build`'s static output can look completely correct while the identical source is broken the moment it's actually served by `jac start`. This isn't a separate bug — it's the same root cause, visible through a different lens.

---

## Steps to Reproduce

1. Scaffold a `web-app` project (`jac create app --kind web-app`).
2. `lib/types.cl.jac`:
   ```jac
   obj:pub Widget {
       has id: str;
       has name: str;
   }
   ```
3. `frontend.cl.jac` (the app's stateful shell — imported by `main.jac`, not itself the compile entry point):
   ```jac
   import from .lib.types { Widget }
   ```
4. `frontend.impl.jac` — construct it from a real event handler:
   ```jac
   impl app.addWidget -> None {
       n = str(len(widgets) + 1);
       widgets = widgets + [Widget(id=n, name="Widget " + n)];
   }
   ```
5. `jac check` passes with zero warnings. `jac start`, open the app, trigger the handler.

**Expected:** a new `Widget` instance is created and added to the list.
**Actual:** `TypeError: Class constructor Widget cannot be invoked without 'new'`, thrown in the browser console.

Image

---

## The three cases

| # | Shape | Broken? |
|---|---|---|
| 1 | `type.jac → frontend.cl.jac → main.jac` (imported into a **non-entry** file) | **Yes** — the real bug |
| 2 | `type.jac → main.jac` (imported directly into the **entry** file) | No — entry file always gets full resolution |
| 3 | Defined and constructed in the **same file** | No — no import binding involved at all |

Case 1 is the common shape in real apps — almost every component file is imported by `main.jac`, not passed to the compiler directly.

---

## The related finding — `jac build` and `jac start` disagree about whether this is even visible

1. On the same project, run `jac build` on a **completely fresh** checkout (no prior `.jac/` cache):
   ```bash
   jac build
   grep -o '.\{30\}Local Widget.\{30\}' .jac/client/dist/client.*.js
   ```
   Result: `new Xx({id:"1",name:"Local Widget"})` — **correct**, `new` present.
2. On an equally fresh copy, run `jac start` directly (no prior `jac build`):
   ```bash
   jac start
   curl -s "http://localhost:8000/static/<served bundle path>" | grep -o '.\{40\}Local Widget.\{20\}'
   ```
   Result: `Fm({id:"1",name:"Local Widget"})` — **broken**, `new` missing.

Both runs use **identical source**; the only difference is which command built the project. `--dev` mode (`jac start --dev`) is broken the same way as plain `jac start`.

---

## Expected vs. Actual Behavior

| | Expected | Actual |
|---|---|---|
| `Widget(id="1", name="test")`, imported via plain `import from` into a non-entry `.cl.jac` file | `new Widget({id: "1", name: "test"})` | `Widget({id: "1", name: "test"})` — throws `TypeError` at runtime |
| Running `jac build` then `jac start` on the same, unmodified source | Identical generated JavaScript from both | `jac build`'s static output can show correct `new Widget(...)`, while `jac start` serves `Widget(...)` (no `new`) for the identical source |
| A file that's the app's conceptual "entry" (e.g. the top-level page component) | Cross-file archetype imports resolve correctly there, same as the literal compile entry point | Only the file passed directly as `main.jac`'s target gets correct resolution; every other file is treated as reached transitively and gets the wrong symbol |

---

## Root Cause

When an archetype is imported into a file that isn't the literal compile entry point (`main.jac`), the imported name resolves to a generic placeholder symbol (`SymbolType.VAR`) instead of its real archetype symbol (`SymbolType.OBJECT_ARCH`). Confirmed directly with a compiler instrumentation script that reads the symbol-resolution output for a real multi-file project.

`classify_call()` (`jaclang/compiler/symbol_utils.jac:50`) decides `new X(...)` vs. plain `X(...)` by checking this symbol's type. With the placeholder, it wrongly picks the plain-call path, and `esast_gen_pass.impl.jac:1890` emits a `CallExpression` instead of a `NewExpression`.

There's already a utility built to resolve exactly this kind of placeholder to its real type — `resolve_imported_symbols()` (`type_evaluator.impl/evaluator_util_methods.impl.jac:698`) — but `classify_call()` never calls it. Wiring it in naively wouldn't help either: it fixes a *different* `Symbol` object than the one the construction call site actually reads (the fix creates a replacement object instead of mutating the shared one in place), confirmed with object-identity tracing.

**`jac build` vs. `jac start`:** whether the placeholder actually gets hit depends on which command builds the project. `jac build` incidentally pre-resolves every file's symbols correctly, as a side effect of an unrelated manifest-building step — so it never shows the bug. `jac start` (production and `--dev`) has no equivalent step, so it does. Confirmed with direct in-process instrumentation of both commands' real code paths, not just reading the source. Practical effect: a project can pass `jac build` cleanly and still be broken the moment it's served by `jac start` — the only reliable check is a live `jac start` server, not `jac build`'s static output.

SahanUday:
## Problem

An `obj` imported with a plain `import from` and constructed in a file that isn't the literal compile entry point compiles to a plain call instead of `new`. Since the archetype always lowers to a real ES `class`, this is a guaranteed runtime crash:

```
TypeError: Class constructor Widget cannot be invoked without 'new'
```

Silent through `jac check` and `jac build` — only visible once a real browser runs the code.

close #7394 

## Root Cause

When an archetype is imported into a file reached only *transitively* (imported by another file, never the file passed directly to the compiler), its symbol resolves to a generic placeholder (`SymbolType.VAR`) instead of its real archetype kind (`SymbolType.OBJECT_ARCH`, etc.). `classify_call()` reads that placeholder to decide `new X(...)` vs. plain `X(...)`, and picks wrong.

A dereferencing utility for exactly this (`TypeEvaluator.resolve_imported_symbols()`) already existed but was never called from `classify_call()`. Its own guard was also broken — it checked "does a symbol exist" instead of "is it correct," so even a direct call to it would have returned the same wrong placeholder.

### The three cases

| # | Shape | Broken? |
|---|---|---|
| 1 | `type.jac → frontend.cl.jac → main.jac` (imported into a **non-entry** file) | **Yes** — the real bug |
| 2 | `type.jac → main.jac` (imported directly into the **entry** file) | No — entry file always gets full resolution |
| 3 | Defined and constructed in the **same file** | No — no import binding involved at all |

Case 1 is the common shape in real apps — almost every component file is imported by `main.jac`, not passed to the compiler directly.

## Fix

Four small edits, no new APIs:

- **`type_evaluator.impl/evaluator_util_methods.impl.jac`** — fixed `resolve_imported_symbols()`'s guard to check the symbol is actually resolved (`sym_type != SymbolType.VAR`), not just present.
- **`symbol_utils.jac`** — `classify_call()` takes an optional `evaluator` and dereferences the symbol through it before deciding.
- **`type_checker_pass.impl.jac`** — both call sites pass `self.evaluator` (already available).
- **`esast_gen_pass.impl.jac`** — the fallback `classify_call` call site passes `self.prog.get_type_evaluator()`; `_sv_import_item_is_type()` gets the same fix (see note below).

**Note on `_sv_import_item_is_type()`:** this second call site isn't optional. `classify_call()`'s fix applies to *every* archetype symbol, including `sv import`ed ones — so it starts emitting `new` for sv-imported bare archetypes regardless. Without the matching fix here, that `new` lands on an RPC stub (`async function`), producing `TypeError: X is not a constructor` — confirmed by temporarily reverting just this one call site. Both edits ship together.

## Tests

- **New regression test** — `test_esast_gen_pass.jac::"obj construction across a transitively-imported file emits new"`. Builds Case 1's exact 3-file shape inline (`TemporaryDirectory`, no new fixture files), compiles the entry file, then invokes `EsastGenPass` directly on the transitively-discovered middle module straight from `prog.mod.hub` — mirroring exactly what the client bundler's `_ensure_js_generated` does in production, not just a single `compile()` call. Verified: fails with the original bug (`return Widget({...})`, no `new`), passes with the fix.
- **`test_backend_purity.jac`** — an existing architectural check that flags backends re-running central type analysis. My fix's two `self.prog.get_type_evaluator()` calls in `esast_gen_pass.impl.jac` correctly tripped it. Added a `SANCTIONED` entry (count `2`, same shape as the existing `symbol_utils` entry right above it): both calls consume the *one* central evaluator to dereference an import placeholder for modules that never ran `TypeCheckPass` — not a duplicate analyzer.

## Verification

- [x] Live `jac start`: `Widget` constructs correctly, zero page/console errors
- [x] Live `jac start --dev`: same result
- [x] Live `jac build`: served bundle matches
- [x] Targeted suites: `test_typed_interop`, `test_client_codegen`, `test_esast_gen_pass`, `test_js_generation` — all pass
- [x] Full `tests/compiler` suite (2150 items): 2072 passed, 62 skipped, 16 failed — all 16 pre-existing environment gaps (musl toolchain, subprocess `PYTHONPATH`), confirmed unrelated
- [x] New regression test confirmed to fail without the fix, pass with it

### DIFF
--- a/docs/docs/community/release_notes/unreleased/jaclang/7408.bugfix.md
+++ b/docs/docs/community/release_notes/unreleased/jaclang/7408.bugfix.md
@@ -1,0 +1,1 @@
+- **Fix: `obj` construction loses `new` when imported into a non-entry(transitive) client file**: constructing an archetype imported into a `.cl.jac` file that isn't the compile entry point no longer drops `new`, which previously crashed at runtime despite passing `jac check` and `jac build`.
--- a/jac/jaclang/compiler/passes/ecmascript/impl/esast_gen_pass.impl.jac
+++ b/jac/jaclang/compiler/passes/ecmascript/impl/esast_gen_pass.impl.jac
@@ -1781,7 +1781,7 @@ impl EsastGenPass.exit_func_call(nd: uni.FuncCall) -> None {
     _call_kind = nd.call_kind;
     if _call_kind is None {
         import from jaclang.compiler.symbol_utils { classify_call }
-        _call_kind = classify_call(nd).value;
+        _call_kind = classify_call(nd, self.prog.get_type_evaluator()).value;
     }
     if _call_kind == "builtin" and isinstance(nd.target, uni.Name) {
         import from jaclang.compiler.passes.ecmascript.es_unparse { JSCodeGenerator }
@@ -4814,6 +4814,9 @@ impl EsastGenPass._sv_import_item_is_type(
         SymbolType.ENUM_ARCH
     };
     sym = item.name.sym;
+    if sym {
+        sym = self.prog.get_type_evaluator().resolve_imported_symbols(sym);
+    }
     if sym and (sym.sym_type in type_kinds) {
         return True;
     }
--- a/jac/jaclang/compiler/passes/main/impl/type_checker_pass.impl.jac
+++ b/jac/jaclang/compiler/passes/main/impl/type_checker_pass.impl.jac
@@ -150,13 +150,13 @@ impl TypeCheckPass.exit_func_call(self: TypeCheckPass, nd: uni.FuncCall) -> None
     }
 
     if self._lambda_depth > 0 {
-        nd.call_kind = classify_call(nd).value;
+        nd.call_kind = classify_call(nd, self.evaluator).value;
         nd.callee_decl = ability_of_symbol(resolve_expr_symbol(nd.target));
         return;
     }
     self.evaluator.get_type_of_expression(nd);
 
-    nd.call_kind = classify_call(nd).value;
+    nd.call_kind = classify_call(nd, self.evaluator).value;
 
     nd.callee_decl = ability_of_symbol(resolve_expr_symbol(nd.target));
 
--- a/jac/jaclang/compiler/symbol_utils.jac
+++ b/jac/jaclang/compiler/symbol_utils.jac
@@ -47,7 +47,7 @@ def _is_user_symbol(sym: uni.Symbol) -> bool {
     return False;
 }
 
-def classify_call(nd: uni.FuncCall) -> CallKind {
+def classify_call(nd: uni.FuncCall, evaluator: any = None) -> CallKind {
     import from jaclang.compiler.type_registry { JAC_TYPE_REGISTRY }
 
     if isinstance(nd.target, uni.AtomTrailer) and nd.target.is_attr {
@@ -55,6 +55,9 @@ def classify_call(nd: uni.FuncCall) -> CallKind {
     }
 
     sym = resolve_expr_symbol(nd.target);
+    if sym is not None and evaluator is not None {
+        sym = evaluator.resolve_imported_symbols(sym);
+    }
 
     if sym is not None
     and sym.decl is not None
--- a/jac/jaclang/compiler/type_system/type_evaluator.impl/evaluator_util_methods.impl.jac
+++ b/jac/jaclang/compiler/type_system/type_evaluator.impl/evaluator_util_methods.impl.jac
@@ -826,7 +826,11 @@ impl TypeEvaluator.resolve_imported_symbols(sym: uni.Symbol) -> uni.Symbol {
         mod_item = sym.decl.find_parent_of_type(uni.ModuleItem);
         assert (mod_item is not None);
         self.get_type_of_module(mod_item.from_mod_path);
-        if isinstance(mod_item.name, uni.Name) and mod_item.name.sym {
+        if (
+            isinstance(mod_item.name, uni.Name)
+            and mod_item.name.sym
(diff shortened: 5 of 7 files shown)


marsninja:

PR #7409, opened 2026-07-13
fix(scale): invalidate topology-index roots cross-pod on edge-list-delta (#7334)

### LINKED ISSUE #7334: ScaleTieredMemory: topology index (GTI) not invalidated cross-pod on edge-list-delta writes
## Summary

In a multi-pod deployment backed by `ScaleTieredMemory` (Mongo L3 + Redis L2), adding or removing an edge on a **root** updates the root's `topology_index_data` in Mongo (L3) but does **not** refresh it in Redis (L2) and does **not** publish a cross-pod invalidation. Other pods therefore keep serving a **stale topology index**, so GTI-narrowed traversals (`[root ->[?:Type]]`) return stale results on any pod that didn't perform the write.

## Root cause

`ScaleTieredMemory.commit` (`jac/jaclang/scale/memory/impl/memory_hierarchy.main.impl.jac`) deliberately excludes `EDGE_LIST_DELTA` intents from both `l2.put` and `publish_invalidation`:

```jac
self_broadcasting = {
    intent.anchor.id
    for intent in self.changes.intents.values()
    if intent.op == WriteOp.EDGE_LIST_DELTA
};
apply_report = self.l3.apply(self.changes);   // Mongo (L3) IS updated
if self.l2 {
    for intent in ... {
        if intent.is_delete() {
            self.l2.invalidate(...); publish_invalidation(...);
        } elif intent.anchor.id not in self_broadcasting {   // <-- EDGE_LIST_DELTA excluded
            self.l2.put(intent.anchor); publish_invalidation(...);
        }
        // else (edge-list-delta): neither L2 update nor cross-pod invalidation
    }
}
```

Adding/removing an edge produces an `EDGE_LIST_DELTA` intent for the endpoint node (`jac/jaclang/runtimelib/changeset.jac`). The exclusion is intentional and reasonable in general — it avoids invalidation storms when many edges are concurrently added to a hot node, and a node's *edge list* can be reconstructed from the edge anchors (which **are** invalidated).

**The topology index is the exception:** it's an opaque blob stored on the root (`NodeAnchor.topology_index_data`), it only ever changes via `EDGE_LIST_DELTA`, and it **cannot** be reconstructed from the edge anchors. So it is exactly the thing that falls through this crack.

## Reproduction

Two `ScaleTieredMemory` instances ("pods") sharing one Mongo + Redis:

1. Builder connects `root` to 100 `TypeA` nodes, commits.
2. Pod A queries `[root ->[?:TypeA]]` → 100 (warm).
3. Pod B adds one more `TypeA` edge to `root`, commits.
4. Inspect storage directly:
   - **Mongo** root: `edges = 101`, decoded index len `101` ✓
   - **Redis** root: `edges = 100`, decoded index len `100` ✗
5. Pod A re-queries (served from the stale Redis L2) → **still 100**.

No `publish_invalidation` is emitted for the root, so even pods running the L1-invalidation listener never mark it stale / evict it.

## Impact

- **Correctness bug** in multi-pod deployments: GTI-narrowed traversals return stale results on pods that did not perform the edge write, until the root is otherwise evicted/reloaded.
- The whole root anchor is affected (its edge list too), but the topology index is the most severe case because it is not reconstructable from edges.

## Proposed fix (scoped)

In the `commit` L2 loop, when an `EDGE_LIST_DELTA` intent's anchor **carries a topology index** (a `NodeAnchor` with `topology_index_data is not None` — i.e. a root), still emit `l2.invalidate` + `publish_invalidation` for it (mirroring the delete branch), so other pods evict their L1 and reload the fresh root/index from L3. Keep the storm-avoidance exclusion for ordinary nodes without an index.

```jac
} elif intent.anchor.id not in self_broadcasting {
    self.l2.put(intent.anchor); publish_invalidation(...);
} elif _carries_topology_index(intent.anchor) {
    // edge-list-delta on a root whose index blob can't be reconstructed from
    // its edges -> must invalidate cross-pod so other pods reload the fresh index.
    self.l2.invalidate(intent.anchor.id); publish_invalidation(...);
}
```
where `_carries_topology_index(a) = isinstance(a, NodeAnchor) and a.topology_index_data is not None`.

**Trade-off:** ~1 extra invalidation per edge add/remove on a root. Roots are relatively few (per-user / per-session), so this is a good correctness/cost trade and preserves storm-avoidance for hot content nodes.

## Related (perf, optional)

Separately, `NodeAnchor.get_topology_index()` re-decodes `topology_index_data` on **every** call (no cache), which is O(edges) per traversal (~11 ms for a 4k-edge index measured locally; grows with graph size). A decoded-index cache invalidated on the same write paths (`set_topology_index`, deserialize, and the in-place `topology_index_data` refresh in `SqliteMemory`/merge) removes that cost with no behavioral change. Happy to include this alongside the fix.

## Environment

- `upstream/main` @ `333fe145e`
- Backend: MongoDB 7 + Redis 7 via `ScaleTieredMemory` (Mongo L3, Redis L2)

akindu-k:
## Summary

Fixes #7334.

In multi-pod `ScaleTieredMemory` deployments (MongoDB L3 + Redis L2), adding or removing an edge on a **root** updates the root's topology index in MongoDB but never refreshes or invalidates it in Redis, and publishes no cross-pod invalidation. Other pods therefore keep serving a **stale topology index**, so type-filtered traversals from the root (`[root ->[?:Type]]`) return stale results on any pod that did not perform the write.

The commit path deliberately excludes `EDGE_LIST_DELTA` intents from both L2 updates and cross-pod invalidation to avoid invalidation storms on hot nodes. The topology index is the exception: it is an opaque blob carried on the root anchor and **cannot** be reconstructed from the edge anchors, so it falls through this gap.

This change adds one scoped exception in the commit L2 loop: a root that carries a topology index now emits `l2.invalidate` + `publish_invalidation` on an edge-list-delta, reusing the existing Redis pub/sub invalidation channel, so other pods evict their copy and reload the fresh index from L3. Storm-avoidance is preserved for ordinary content nodes (roughly one extra invalidation per edge write on a root; roots are scarce).

## How this was verified

- **End-to-end against real MongoDB + Redis** (testcontainers): a topology-index-carrying root is warmed into the Redis L2 cache, an edge-list-delta is committed on it, and the stale L2 blob is confirmed evicted so a subsequent read reloads the fresh index from MongoDB.
- **Negative control**: with the new branch disabled the same scenario leaves the stale L2 blob in place (reproducing the bug); with the fix in place it is evicted.
- `jac check` on the scale memory module reports the same diagnostics count as `main` (no new errors introduced).

image
image
image


akindu-k:
Closing this PR as it is being implemented by https://github.com/jaseci-labs/jac/pull/7335

PR #7421, opened 2026-07-13
feat(jac-scale): implement KEDA HTTP Add-on activation resources

### LINKED ISSUE #7403: feat(jac-scale): add first-class KEDA HTTP Add-on activation resources
## Summary

Extend the existing `jac-scale` KEDA autoscaler support with first-class resources for HTTP request activation and request buffering through the KEDA HTTP Add-on.

The existing KEDA engine from #6046 / #6220 creates `ScaledObject` and `TriggerAuthentication`, but true HTTP scale-to-zero additionally requires an `InterceptorRoute` and an `external-push` trigger wired to the HTTP Add-on scaler.

This is needed by applications such as JacHammer IDE previews, where a stable URL should wake a zero-replica workload on an authorized request.

## Scale-target prerequisite

KEDA cannot scale a standalone Pod. The target must expose Kubernetes' `/scale` subresource:

- `Deployment` is the default and recommended target for stateless HTTP workloads.
- `StatefulSet` or a compatible custom resource may be used when its lifecycle requires it.
- A Service must select the target workload's pods and remain present while replicas are zero.

This feature manages HTTP activation resources around an existing scalable workload and Service. It does not convert standalone Pods into Deployments.

## Proposed scope

Add a public HTTP activation abstraction that can be called programmatically at runtime, without requiring a complete `jac start --scale` deployment.

The model must preserve the KEDA `InterceptorRoute.rules[]` structure. Fields inside one rule use AND semantics; separate rules use OR semantics.

A representative typed shape is:

```jac
obj HTTPPathMatch {
    has value: str;
}

obj HTTPHeaderMatch {
    has name: str,
        value: (str | None) = None;
}

obj HTTPRoutingRule {
    has hosts: list[str] = [],
        paths: list[HTTPPathMatch] = [],
        headers: list[HTTPHeaderMatch] = [];
}

obj HTTPServiceTarget {
    has service: str,
        port: (int | None) = None,
        port_name: (str | None) = None;
}

obj HTTPConcurrencyMetric {
    has target_value: int;
}

obj HTTPRequestRateMetric {
    has target_value: int,
        window: str = "1m",
        granularity: str = "1s";
}

obj HTTPStaticResponse {
    has status_code: int = 503,
        body: (str | None) = None,
        body_from_config_map: (dict[str, str] | None) = None,
        headers: dict[str, str] = {};
}

obj HTTPColdStartSpec {
    has placeholder: (HTTPStaticResponse | None) = None,
        fallback_service: (HTTPServiceTarget | None) = None;
}

obj HTTPTimeoutSpec {
    has readiness: (str | None) = None,
        request: (str | None) = None,
        response_header: (str | None) = None;
}

obj HTTPActivationSpec {
    has name: str,
        namespace: str,
        scale_target_name: str,
        scale_target_api_version: str = "apps/v1",
        scale_target_kind: str = "Deployment",
        target: HTTPServiceTarget,
        rules: list[HTTPRoutingRule] = [],
        min_replicas: int = 0,
        max_replicas: int = 1,
        polling_interval: int = 30,
        cooldown_period: int = 300,
        concurrency: (HTTPConcurrencyMetric | None) = None,
        request_rate: (HTTPRequestRateMetric | None) = None,
        cold_start: (HTTPColdStartSpec | None) = None,
        timeouts: (HTTPTimeoutSpec | None) = None;
}
```

The exact names may change during implementation, but callers should not need to hand-build unvalidated KEDA dictionaries for routing, metrics, cold-start behavior, or timeouts.

The concrete KEDA implementation should reconcile:

1. `http.keda.sh/v1beta1 InterceptorRoute`
2. `keda.sh/v1alpha1 ScaledObject` with an `external-push` trigger

The `InterceptorRoute` must be created before the `ScaledObject`, because the HTTP external scaler resolves metric metadata from the route when KEDA reconciles the trigger.

## Required behavior

- Idempotent create-or-patch and 404-safe destroy.
- Deterministic, collision-safe names and ownership labels linking both resources to the scale target.
- Preserve routing-rule grouping and host/path/header matching semantics.
- Validate that exactly one of Service `port` or `port_name` is set.
- Require at least one of concurrency or request-rate scaling.
- Support request-rate window and granularity.
- Support `minReplicaCount: 0`, configurable maximum replicas, polling interval, and cooldown.
- Strongly typed cold-start placeholder/fallback and readiness/request/response-header timeouts.
- Configurable HTTP Add-on scaler address and namespace.
- Validate that the referenced scale target and Service exist before apply.
- Cleanup both resources without affecting unrelated KEDA resources.
- Support both normal deployment-time callers and dynamic control-plane callers such as JacHammer.
- Authentication and application-specific wake authorization remain upstream of the interceptor and are out of scope.

## Acceptance criteria

- [ ] A public runtime-callable abstraction represents an HTTP-activated scalable target.
- [ ] Standalone Pods are rejected with a clear error explaining the `/scale` requirement.
- [ ] Apply validates the target workload and Service, then creates or patches the `InterceptorRoute` before the `ScaledObject`.
- [ ] Generated routing rules preserve AND-within-rule and OR-between-rules semantics.
- [ ] The generated `ScaledObject` uses `external-push` with `scalerAddress` and `interceptorRoute` metadata.
- [ ] Service port number/name validation and scaling-metric validation match the `InterceptorRoute` API.
- [ ] Destroy removes both managed resources safely and supports partial-apply cleanup.
- [ ] Unit tests cover typed model validation, route grouping, apply ordering, patching, cleanup, and naming isolation.
- [ ] An integration test begins with a zero-replica Deployment, sends an HTTP request through the interceptor, waits for readiness, and observes scale-down after cooldown.
- [ ] Documentation includes an end-to-end HTTP scale-to-zero example and the required traffic topology.
- [ ] Release notes describe the new capability.

## Out of scope

- Creating the target Deployment or Service.
- Installing KEDA core or the HTTP Add-on.
- Application authentication/share policy.
- Application source restoration or dependency caching.
- Performance claims for large numbers of routes; tracked separately.

## References

- Parent KEDA abstraction: #6046
- KEDA engine implementation: #6220 / #6534
- KEDA HTTP Add-on architecture: https://keda.sh/http-add-on/0.15/concepts/architecture/
- InterceptorRoute API: https://keda.sh/http-add-on/0.15/reference/interceptorroute/
- Autoscale an app: https://keda.sh/http-add-on/0.15/user-guide/autoscale-an-app/
- JacHammer preview use case: https://github.com/jaseci-labs/jacBuilder/issues/603

Developer-Linus:
Closes #7403

## Overview

Adds first-class KEDA HTTP Add-on activation resources on top of the existing KEDA engine (#6046 / #6220 / #6534). Where the current `KEDAAutoscaler` scales on CPU/memory/queue-depth metrics that can be polled even at zero replicas, HTTP traffic can't be polled at zero — there's nothing listening on the Service to observe. The HTTP Add-on solves this with an interceptor that holds an incoming request, signals KEDA to scale up, waits for readiness, then forwards the request. This PR adds a typed spec model and the `KEDAAutoscaler` methods needed to reconcile the two Kubernetes resources that make that work: an `InterceptorRoute` (routing/cold-start/timeouts) and a `ScaledObject` using an `external-push` trigger wired to the interceptor.

This is a callable abstraction, not a config-wired deploy path yet — it can be invoked programmatically (e.g. by a control-plane caller like JacHammer waking an IDE preview) without going through a full `jac start --scale` deployment. Wiring it into `jac.toml`/the standard deploy flow is left for a follow-up (see below).

### New file: `jac/jaclang/scale/deploy/autoscale/http_activation.jac`

A typed, validation-free data model (mirrors `AutoscalerSpec`/`Trigger` in `autoscaler.jac` — plain `obj`s, no self-validation) so callers configure HTTP activation declaratively instead of hand-building KEDA YAML dicts:

- `HTTPPathMatch`, `HTTPHeaderMatch`, `HTTPRoutingRule` — routing match criteria. Fields within one rule (`hosts`/`paths`/`headers`) are AND-ed; separate rules in a list are OR-ed.
- `HTTPServiceTarget` — the Service fronting the scale target; exactly one of `port`/`port_name` is valid.
- `HTTPConcurrencyMetric`, `HTTPRequestRateMetric` — the two supported HTTP scaling metrics (request-rate defaults to a `1m` window / `1s` granularity).
- `HTTPStaticResponse`, `HTTPColdStartSpec` — what the interceptor serves while the target wakes from zero (a static placeholder, a fallback service, or both).
- `HTTPTimeoutSpec` — readiness/request/response-header timeouts applied by the interceptor.
- `HTTPActivationSpec` — the full spec tying the above together, plus `scale_target_name`/`scale_target_kind`/`scale_target_api_version`/`scale_target_plural` (optional, enables existence validation for custom-resource targets), replica bounds, and polling/cooldown.

### `KEDAAutoscaler` extended (`keda_autoscaler.jac` / `.impl.jac`)

New fields:
- `_apps_api`, `_core_api` — injectable Kubernetes API clients (mirrors the existing `_custom_api`/`_v2_api` mock-injection pattern) used to validate the scale target and Service exist before writing anything.
- `http_scaler_address` — configurable HTTP Add-on scaler address, defaulting to `keda-add-ons-http-external-scaler.keda:9090`.
- `_http_preflight_cache` (static) — a second, independent preflight cache from the existing `_preflight_cache`, since the HTTP Add-on CRDs are a separately-installed component from core KEDA.

New methods:
- `_build_routing_rules(rules)` — pure function; expands each `HTTPRoutingRule` into one match-block dict per configured path (or a single block with no `path` key when none is set), carrying only the non-empty `hosts`/`path`/`headers` keys, preserving AND-within-rule / OR-between-rules grouping.
- `_build_interceptor_route(spec)` — builds the full `InterceptorRoute` manifest (`http.keda.sh/v1beta1`): `target` (service name, port xor portName), `scalingMetric` (`concurrency` or `requestRate`), `rules`, and conditional `coldStart` (`placeholder.response` / `fallback.service`) / `timeouts` blocks (only emitted when set, matching how `_build_manifests` already omits `idleReplicaCount`/`initialCooldownPeriod` when unset).
- `_build_http_scaled_object(spec)` — builds the `ScaledObject` manifest (`keda.sh/v1alpha1`) with a single `external-push` trigger carrying `scalerAddress`/`interceptorRoute`, plus `targetValue` (and `window`/`granularity` for request-rate) depending on which metric is set.
- `interceptor_route_name_for(target)` / `http_scaled_object_name_for(target)` — deterministic, collision-safe naming (`{target}-http-route` / `{target}-http-scaledobject`).
- `_validate_http_activation_spec(spec)` — rejects standalone `Pod` targets with a message explaining the `/scale` subresource requirement; enforces the `port`/`port_name` XOR; requires at least one of `concurrency`/`request_rate`.
- `apply_http_activation(spec)` — orchestrates the above: validate → HTTP Add-on preflight check (own cache, separate from core KEDA's) → confirm the target and Service exist via `_apps_api`/`_core_api` (`ValueError` on 404): `Deployment`/`StatefulSet` targets are checked directly, a custom-resource target is checked via `CustomObjectsApi` when `scale_target_plural` is set (group/version parsed from `scale_target_api_version`), and any other `scale_target_kind` without a plural raises `ValueError` before writing anything, since it can't be verified → create-or-patch `InterceptorRoute` → create-or-patch `ScaledObject`, in that order, since the external scaler resolves its metadata from the route. Returns `False` (no cluster writes) when the HTTP Add-on CRDs aren't installed, mirroring the existing `preflight()`/`apply()` degrade-gracefully pattern.
- `destroy_http_activation(target, namespace)` — deletes both resources, 404-safe independently, so a partial-apply state (one resource created, one not) can still be cleaned up.

### `destroy_collection` extended

Generalized the existing plural-sweep loop from a flat list of `keda.sh` pluralities to `(group, version, plural)` tuples, since `interceptorroutes` lives under a different API group (`http.keda.sh/v1beta1`) than `scaledobjects`/`triggerauthentications` (`keda.sh/v1alpha1`). Behavior for the two existing resource kinds is unchanged; a namespace-wide `destroy_collection` call now also sweeps orphaned `InterceptorRoute`s by the same `managed: jac-scale` label selector.

## CRD-Schema Fix (Found via Live-Cluster Testing)

The unit tests above are all mocked, so nothing caught that `_build_interceptor_route`'s original output did not match the real `InterceptorRoute` CRD. Applying it against a cluster running KEDA 2.20.1 and HTTP Add-on 0.15.0 was rejected outright:

- `scaleTargetRef` is not a field on this CRD at all; it was pure dead weight.
- The Service reference belongs under `target`, not `service`.
- `scalingMetric` (`concurrency` or `requestRate`) is a required field on the route itself; the old code only ever wrote it into the `ScaledObject`'s trigger metadata, so the route was rejected with `spec.scalingMetric: Required value`.
- `coldStart.placeholder` and `coldStart.fallbackService` were nested one level too shallow (`coldStart.placeholder.response`, `coldStart.fallback.service`).

Fixed `_build_interceptor_route` to match the real schema and updated the five affected unit tests, plus added a new one covering `scalingMetric.requestRate` (previously untested even by the mocks). Re-verified end to end against the same live cluster: `apply_http_activation` now reconciles cleanly, a request through the interceptor wakes the zero-replica Deployment, and it scales back to zero after cooldown.

## Real-Cluster Integration Test

Mocked unit tests can prove the manifests are shaped correctly; they cannot prove KEDA actually reconciles them or that the interceptor actually wakes a zero-replica workload. Added a real-cluster integration test that starts from a zero-replica Deployment, sends an HTTP request through the interceptor, waits for readiness, and observes scale-down after cooldown, matching the issue's acceptance criterion directly. It also calls `apply_http_activation` twice in a row to exercise the patch branch of create-or-patch against a real API server, not just a mock.

- `jac/jaclang/scale/tests/fixtures/keda_http_activation_e2e/fixture.yaml` — the zero-replica Deployment + Service the test starts from. This feature does not create either, so the fixture stands in for a workload deployed some other way.
- `jac/jaclang/scale/scripts/keda_http_activation_verify.jac` — a thin, environment-variable-parameterized driver around `apply_http_activation`/`destroy_http_activation`.
- `jac/jaclang/scale/scripts/keda_http_activation_real_e2e.sh` — orchestrates the fixture, the driver (create, then patch), the interceptor request, and readiness/cooldown polling, with a diagnostics dump on failure matching the existing `k8s_microservice_real_e2e.sh` convention.

Run manually against a real cluster (KEDA + HTTP Add-on installed); not yet wired into CI, since that trigger/job placement decision belongs to a maintainer. See the script's own preflight check and header comment for how to run it.

## Tests

30 unit tests in `test_keda_http_activation.jac` (29 from the original implementation, plus one added alongside the CRD-schema fix above covering `scalingMetric.requestRate`), plus one existing `test_keda_autoscaler.jac` assertion updated for `destroy_collection`'s new third resource kind:

- Typed model validation: port/port_name XOR, concurrency-or-request_rate requirement, request-rate defaults.
- Routing-rule grouping: single-rule AND grouping, multi-rule OR grouping, multi-path expansion within one rule, omission of empty keys.
- `InterceptorRoute` manifest building: full-body shape, port vs portName, conditional cold-start placeholder/fallback-service, conditional timeouts.
- `ScaledObject` manifest building: full external-push body, concurrency vs request-rate trigger metadata.
- Standalone-Pod rejection, StatefulSet acceptance, custom-target validation via `scale_target_plural` (both found and missing), and rejection of unrecognized `scale_target_kind` values left without a plural.
- Naming determinism/distinctness.
- `apply_http_activation` ordering (InterceptorRoute before ScaledObject, both on create and patch paths), target/Service existence validation, and HTTP Add-on preflight skip.
- `destroy_http_activation` full removal and partial-apply (one resource already gone) cleanup.
- `destroy_collection` sweeping `InterceptorRoute`s alongside the existing two resource kinds.

All 60 tests pass (`jac test jac/jaclang/scale/tests/deploy/test_keda_http_activation.jac jac/jaclang/scale/tests/deploy/test_keda_autoscaler.jac`). See "Real-Cluster Integration Test" above for the non-mocked coverage these unit tests can't provide on their own.

## Files Changed

| Status | File | Purpose |
|---|---|---|
| Added | `jac/jaclang/scale/deploy/autoscale/http_activation.jac` | Typed `HTTPActivationSpec` model and its component objects |
| Added | `jac/jaclang/scale/tests/deploy/test_keda_http_activation.jac` | 30 unit tests covering the full HTTP activation contract |
| Modified | `jac/jaclang/scale/deploy/autoscale/keda_autoscaler.jac` | New fields (`_apps_api`, `_core_api`, `http_scaler_address`, `_http_preflight_cache`) and HTTP activation method declarations |
| Modified | `jac/jaclang/scale/deploy/autoscale/keda_autoscaler.impl.jac` | Full implementation: routing-rule/manifest builders, validation, `apply_http_activation`, `destroy_http_activation`, `destroy_collection` generalized to multiple API groups. Also includes the CRD-schema fix to `_build_interceptor_route` found via live-cluster testing. |
| Modified | `jac/jaclang/scale/tests/deploy/test_keda_autoscaler.jac` | Updated `destroy_collection` test to account for the new `interceptorroutes` sweep |
| Added | `jac/jaclang/scale/tests/fixtures/keda_http_activation_e2e/fixture.yaml` | Zero-replica Deployment + Service fixture for the real-cluster integration test |
| Added | `jac/jaclang/scale/scripts/keda_http_activation_verify.jac` | Env-var-parameterized `apply`/`destroy` driver used by the integration test |
| Added | `jac/jaclang/scale/scripts/keda_http_activation_real_e2e.sh` | Orchestrates the real-cluster integration test end to end |
| Modified | `docs/docs/reference/plugins/jac-scale-kubernetes.md` | New "HTTP Add-on Activation" subsection: prerequisites, example, traffic topology |

## Behavior Preserved

- Core KEDA CPU/memory/queue-depth scaling (`apply`, `destroy`, `preflight`, `_preflight_cache`) is untouched — HTTP activation is fully additive via new methods and a separate preflight cache.
- `destroy_collection`'s behavior for `scaledobjects`/`triggerauthentications` is unchanged; only a third resource kind was added to its sweep.

## Out of Scope (per the issue) / Follow-up

- Wiring `apply_http_activation`/`destroy_http_activation` into `jac.toml` config and the standard `jac start --scale` deploy path — this PR ships the callable abstraction; config/deploy-path wiring is a natural next PR.
- Issue: #7475


jac/jaclang/scale/deploy/autoscale/keda_autoscaler.impl.jac
```diff
@@ -315,3 +319,391 @@ impl KEDAAutoscaler.destroy_collection{
...
+    result: list[dict[str, any]] = [];
+    for rule in rules {
+        header_matches: list[dict[str, any]] = [];
+        for header in rule.headers {
+            header_entry: dict[str, any] = {"name": header.name};
+            if header.value is not None {
+                header_entry["value"] = header.value;
+            }
+            header_matches.append(header_entry);
+        }
```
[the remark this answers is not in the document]
Developer-Linus: This isn't a mismatch — spec.rules[].headers in the InterceptorRoute CRD is a list of {name, value} objects, not a string map. From the official reference:

rules:
  - hosts:
      - api.example.com
    headers:
      - name: Authorization
        value: "Bearer token123"
      - name: Content-Type   # value omitted = matches any value for this header

That's exactly what _build_routing_rules builds: a list of {"name": ..., "value": ...} dicts, with value omitted when header.value is None (which the CRD docs confirm means "match any value for this header name"). No change needed here.

Developer-Linus:
> ### Greptile Summary
> This PR adds programmatic KEDA HTTP Add-on activation for jac-scale. The main changes are:
> 
> * Typed HTTP activation spec objects for routes, targets, metrics, cold-start behavior, and timeouts.
> * `KEDAAutoscaler` helpers to validate targets and reconcile `InterceptorRoute` plus HTTP `ScaledObject` resources.
> * Cleanup support for HTTP activation resources and namespace-wide collection sweeps.
> * Unit tests, a real-cluster verification script, fixture docs, release notes, and user documentation.
> 
> ### Confidence Score: 4/5
> This is close, but the header route shape should be fixed before merging.
> 
> * Header-constrained activation rules can be written in a shape the HTTP Add-on does not accept.
> * Routes that depend on request headers can fail to apply or fail to match traffic.
> * The latest custom-target validation and multi-path routing changes otherwise look aligned with the intended behavior.
> 
> jac/jaclang/scale/deploy/autoscale/keda_autoscaler.impl.jac
> 
> ### Important Files Changed
> Filename	Overview
> jac/jaclang/scale/deploy/autoscale/keda_autoscaler.impl.jac	Adds HTTP activation validation, manifest builders, apply/destroy flows, and collection cleanup; header route matching still uses the wrong resource shape.
> jac/jaclang/scale/deploy/autoscale/http_activation.jac	Adds the typed HTTP activation configuration model used by the autoscaler builders.
> jac/jaclang/scale/deploy/autoscale/keda_autoscaler.jac	Adds HTTP activation fields, imports, caches, and method declarations.
> Reviews (8): Last reviewed commit: ["test(jac-scale): add readiness checks an..."](https://github.com/jaseci-labs/jac/commit/a19e88310bdb99cd1de1b2b8f7041f5e12fd5cc0) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=43871873)

The header-route shape flagged here (and in the confidence score above) is not a mismatch. spec.rules[].headers on the InterceptorRoute CRD is documented as a list of {name, value} objects, not a string map — see the official reference:

rules:
  - hosts:
      - api.example.com
    headers:
      - name: Authorization
        value: "Bearer token123"
      - name: Content-Type   # value omitted = matches any value

_build_routing_rules (keda_autoscaler.impl.jac:517-527) already emits exactly this: a list of {"name": ..., "value": ...} entries, omitting value when header.value is None, which the CRD docs confirm means "match any value for that header name." Changing this to a map would actually break header-constrained routes against the real API server, not fix them.

No change needed on this point. Confirming the rest of the review (custom-target validation, multi-path routing) looks good as noted.

docs/docs/reference/plugins/jac-scale-kubernetes.md
```diff
@@ -454,6 +454,83 @@ metadata = { queueName = "orders", mode = "QueueLength", value = "50", protocol
...
+        max_replicas=1,
+        cooldown_period=300,
+        concurrency=HTTPConcurrencyMetric(target_value=1)
+    );
+
+    keda.apply_http_activation(spec);
+
+    # Later, once the preview is torn down:
+    keda.destroy_http_activation("preview-abc123", "jachammer-previews");
+}
```
Thamirawaran: in jac code, why we need to import?

kugesan1105:
1. remove the `internal imports`  from docs
2. move the test scripts into teh `test/`
3. e2e can be added in teh followup PR : `jac.toml`

Thamirawaran (changes requested):
consider @kugesan1105 comments too

docs/docs/reference/plugins/jac-scale-kubernetes.md
Developer-Linus: Addressed. I have removed from the documentation. I will update once jac.toml configuration lands.

Developer-Linus:
> consider @kugesan1105 comments too

I have addressed them.

Thamirawaran (approved):
lets do the follow up pr with proper toml config and also replace that yaml file too

Issue #5958, opened 2026-05-13
[Performance]: jaclang first-use compile peaks past 8GB and crash-loops on small (<=16GB) nodes


MusabMahmoodh:
# jaclang first-use compile peaks past 8GB and crash-loops on small (<=16GB) nodes

## TL;DR

jaclang's first-use compile peak memory **scales linearly with the plugin import surface and is never released**. When pluggy discovers a registered plugin, jaclang compiles its `plugin.jac` + every transitively imported `.jac` module in the same Python process as jaclang's own 38-module first-use compile. Nothing gets freed.

Today the combined peak overruns a 16 GB node by ~1-2 GB. The CI test passes only because pod crash-loop restarts eventually slot in by luck. This is not microservice-specific: any future feature PR of similar size triggers the same regression, on any plugin package, and each one chips deeper into the available headroom.

## Where it crashes

Kubernetes pod startup chain runs `jac install` then `jac start app.jac`. Pluggy discovers `jac-scale:scale`, imports `jac_scale.plugin`, which makes jaclang compile `plugin.jac` + its transitive `.jac` imports in the same Python process as jaclang's own first-use compile.

Pod previous-instance log (truncated):

```
Successfully installed jaclang-0.14.1
Setting up Jac for first use (compiling and caching compiler)...
Precompiling compiler for cpython-312...
  Found 38 .jac files
<process killed, exit_code=137>
```

K8s reports `last_terminated=reason=Error exit_code=137` (SIGKILL by kernel OOM-killer).

## Why it worked before the microservice PR and not after

This is **not** specific to microservices. The microservice PR (#5846) happened to be the first feature large enough to push the combined compile peak past the 16 GB ceiling. Any feature PR of similar size would have triggered the same regression. The root cause is that **jaclang's compile memory scales linearly with the plugin import surface and never releases**.

K-track did not touch jaclang core. It grew `jac-scale/` by ~5800 lines (microservice deploy targets, gateway internals, image distributors, manifest builders). At runtime in monolith mode (`target=kubernetes`) none of that microservice code is even loaded. But at *compile-discovery* time, pluggy imports `jac_scale.plugin`, which forces jaclang to compile `plugin.jac` + every `.jac` module it transitively imports, **in the same Python heap** as jaclang's own 38-module first-use compile.

| State | jac-scale plugin compile surface | Combined peak vs 16 GB runner headroom |
|---|---|---|
[...]

PR #7422, opened 2026-07-13
perf(core): cut bootstrap RSS −319 MB (−16%) — guard-check materialization + py_ast eviction

chess10kp:
## Summary

Cuts cold-bootstrap peak RSS a further **−319 MB (−15.9%)** on top of #7393 (2007 → 1688 MB for a 205-module `JAC_REBUILD=1` bootstrap), plus a **−19% bootstrap speedup** (102s → 83s). Two independent, complementary fixes targeting the two distinct causes of post-#7393 retained memory: **spurious allocation** and **over-retention**.

Partially addresses #5958. Stacks on #7393.

## Background: two problems, not one

After #7393 made `CodeGenTarget` manifests lazy, ~2 GB was still retained after a cold bootstrap. Investigation showed two distinct root causes mashed together:

1. **Spurious allocation** — the lazy `UniNode.gen` getter allocates a `CodeGenTarget` on first *read*. Codegen guard checks (`if nd.gen.py_ast:`) therefore materialize an empty target on every node they touch, even when no output is ever produced. Instrumentation showed **~38% of all `CodeGenTarget` allocations are these empty read-only artifacts** (351k of 953k in the full build), and **97.4% of allocations come from the `enter_node` guard checks** themselves.

2. **Over-retention** — after `PyBytecodeGenPass` marshals `py_ast` into `py_bytecode`, the in-memory Python AST forest on each node's `_gen.py_ast` is dead weight. Worse, every generated `ast3` node carries a `jac_link` list back to source `UniNode`s, pinning the whole graph via reference cycles that are never collected.

## Fix 1: stop guard-check materialization (commit 1)

Switch the three hot guard sites to test the backing `_gen` field directly (non-allocating):

- `PyastGenPass.enter_node` — `if nd.gen.py_ast:` → `if (nd._gen is not None and nd._gen.py_ast):`
- `EsastGenPass.enter_node` — same pattern for `es_ast`
- `UniNode.unparse` / `Token.unparse` — same pattern for `jac`

Semantically identical (truthy under exactly the same conditions), but skips the allocating getter when no target exists yet. Measured on a single heavy module: **0 empty targets** after the fix (was 13,901 / 37.7%), total allocations −38%, identical generated output.

## Fix 2: `PyAstEvictPass` (commit 2)

New pass appended to the end of `get_py_code_gen`, after `PyBytecodeGenPass`. For the **internal compiler program's** modules, clears `_gen.py_ast` after bytecode is produced — freeing the ast3 forest and breaking the `jac_link` retention cycles.

**Gating:**
- Default: evict only when `self.prog is compiler.internal_program` (the compiler's own modules — their `py_ast` is never consumed by user tooling; `jac2py`/`lang_tools` recompile fresh). User-program modules keep their `py_ast` so cross-module `decl_link` reads during later compiles stay intact.
- `JAC_PYAST_EVICT_ALL=1` — broaden to every program (testing).
- `JAC_NO_PYAST_EVICT=1` — disable entirely (matches the `JAC_NO_UNIT_EVICT` convention).

Uses `flatten()` rather than `get_all_sub_nodes`, which relies on a sub-node cache that is not reliably populated at this point in the schedule.

**Safety verified:** the only non-producer readers of `py_ast` after bytecode gen are `DeleteStmt.py_ast_targets` (read *during* `pyast_gen_pass`, before bytecode — not post-codegen) and `jac2py`/`lang_tools` (which **recompile fresh** and never read a cached module's `py_ast`).

## Measurements

**205-module cold bootstrap** (`JAC_REBUILD=1`, peak RSS via `/proc/<pid>/status` VmHWM):

| configuration | peak RSS | time |
|---|---|---|
| No fixes (pure `upstream/main`) | 2007 MB | 102s |
| Both fixes | **1688 MB** | **83s** |
| **Delta** | **−319 MB (−15.9%)** | **−19s (−19%)** |

Decomposition (run-to-run variance ~±10 MB):

| fix | contribution |
|---|---|
| Guard fix alone (fewer spurious `CodeGenTarget` allocations) | **−211 MB (~−10.5%)** |
| `PyAstEvictPass` on top (free the `py_ast` forest + break `jac_link` cycles) | **−108 MB (~−5.4%)** |

The guard fix is the bigger RSS lever because each spurious `CodeGenTarget` drags ~6 empty containers (lists/dicts) — the container bloat is heavy in bytes. The evict pass frees more *node count* (ast3 nodes **−44%**: 821k → 456k on a controlled compile) but lighter per-node objects.

## Caveats / follow-ups

- **`PyAstEvictPass` reaches ~302k of ~596k `py_ast`-bearing targets.** `flatten()` walks the compiled module trees; the remaining ~294k sit on a secondary path / cross-referenced nodes. Completing this would roughly double the evict pass's contribution. Tracked as a follow-up.
- **`CodeLocInfo` (~1.38M objects) is untouched** and is now the single largest retained population. It's eager in `UniNode.postinit` and never freed. Separate investigation (lazy `loc`).
- A ~19% bootstrap **speedup** came for free (fewer allocations → less GC pressure).

## How to reproduce

```
# baseline (no fix): temporarily revert both commits
# with fix:
JAC_REBUILD=1 jac --version   # measure peak RSS via /proc or VmHWM

# toggles
JAC_NO_PYAST_EVICT=1   # disable the evict pass (isolate the guard fix)
JAC_PYAST_EVICT_ALL=1  # force eviction for all programs (testing)
```

## Notes

- `pre-commit` clean (`jac-format` + `ruff`).
- Cold self-compile (177–205 modules) green with the pass wired.
- No behavior change to generated code (`py_ast` count identical pre/post guard fix; `py_ast` is dead after `py_bytecode` is produced).


marsninja:
Dont think we need a whole pass, i believe it might be better to use "after_pass" or on module exit.

PR #7427, opened 2026-07-13
feat(client): devtools overlay (network + LLM panels)

Ashuwat:
- **feat(client): add dev-only devtools overlay (isolated shell + Environment panel)**
- **feat: dialog box fix, with environment and network intercepts/debug statements**

## **Description**
it's a dialog box for debugging apps specifically in the jac-client section. this is meant for both client side and server side applications as you can see the network requests made and whenever the byllm() function is invoked. you can see the metadata of the request and the byllm() call.

any improvements would be advised. the dialog box is v1, and can change over time but this is a significant improvement from before.


chess10kp:
@Ashuwat does this work for any frontend or just React?

marsninja:

PR #7437, opened 2026-07-13
fix(scale): add bundle_host_path so local clusters deploy without an RWX bundle PVC

### LINKED ISSUE #7433: bug(scale): RWX bundle PVC is mandatory — documented `host_path` / `kind load` local-cluster support does not exist in the source
## Summary

A `[scale.microservices]` deploy **requires** a `ReadWriteMany` StorageClass for the bundle PVC, with no escape hatch. Local clusters (kind / k3d / minikube) have no RWX provisioner out of the box, so `jac start --scale` cannot target them without hand-provisioning storage.

Separately: the docs advertise a `host_path` volume mode and native `kind load docker-image` image distribution *specifically for this case*, but **neither exists in the source tree**.

Environment: `jac 0.32.0 (Linux x86_64)`, jaseci @ `ace73e7a5`.

## Repro — RWX is mandatory both ways

**With `bundle_storage_class` set to a class the local cluster can't satisfy:**

```toml
[scale.kubernetes]
```

kind's default provisioner refuses:

```
failed to provision volume with StorageClass "efs-sc":
  NodePath only supports ReadWriteOnce and ReadWriteOncePod (1.22+) access modes
```

**With `bundle_storage_class` omitted**, jac-scale fails fast:

```
[ERROR] Error: --dry-run failed: Bundle PVC 'jaseci-bundles' requests accessModes
[ReadWriteMany] but no bundle_storage_class is set, so it would use the cluster
default StorageClass. Most cloud defaults (e.g. EKS gp3 / ebs.csi.aws.com) are
ReadWriteOnce and cannot provision a ReadWriteMany volume, leaving the bundle-loader
pod Pending. Set 'bundle_storage_class' under [scale.kubernetes] to a RWX-capable
StorageClass (e.g. 'efs-sc').
```

So RWX is required no matter what, and there is no supported way to deploy to a single-node local cluster.

## Workaround

Static hostPath PVs, pre-bound to the exact claims jac-scale generates. Legitimate on a single node, where RWX degenerates to a shared hostPath:

```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata: {name: efs-sc}
provisioner: kubernetes.io/no-provisioner
volumeBindingMode: WaitForFirstConsumer
---
apiVersion: v1
kind: PersistentVolume
metadata: {name: bundles-pv}
spec:
  capacity: {storage: 1Gi}
  accessModes: ["ReadWriteMany"]
  storageClassName: efs-sc
  hostPath: {path: /mnt/jac/bundles, type: DirectoryOrCreate}
  claimRef: {namespace: <ns>, name: <app>-bundles}
---
# ...same for the shared-data PVC
```

(plus a `chmod 777` on the hostPath — see the separate bundle-loader `fsGroup` issue.)

With this, the full microservice topology deploys to kind and runs healthy.

## The documented local-cluster features don't exist

`docs/docs/reference/plugins/jac-scale-kubernetes.md:921` documents exactly the mode that would solve this:

> | `host_path` | yes (hostPath mode) | Local-cluster-only alternative; binds the volume to a directory on the host node. Use only on k3d / kind / Minikube; ... |

and `:875`:

> **Local clusters** (Minikube, Docker Desktop, k3d, kind): if `image_registry` is unset, the built image is loaded directly into the cluster's runtime (`minikube image load`, `k3d image import`, `kind load docker-image`).

The jac-scale release notes describe the implementation:

> New `_cluster_detect.jac` classifies the active kubeconfig context (minikube / k3d / kind / remote); `_image_build.jac` resolves the right Dockerfile and dispatches build/distribute per cluster type (minikube docker-env, `k3d image import`, `kind load docker-image`, or `docker push` for remote).

But in the tree at `ace73e7a5`:

```console
$ ls jac/jaclang/scale/ | grep -iE "cluster_detect|image_build"
$ grep -rn "kind load" jac/jaclang/scale/
$ grep -rni "host_path" jac/jaclang/scale/
```

All three return nothing. `_cluster_detect.jac` and `_image_build.jac` do not exist, `kind load` appears nowhere in the source, and the microservice target hardcodes `no_image: bool = True` (`deploy/target/kubernetes/microservice/target.jac:45`).

So either the docs are describing a feature that was removed/never landed, or it lives somewhere I couldn't find. Either way the documented local-cluster story is not reachable.

## Expected

Any one of:

1. Ship the documented `host_path` volume mode so local clusters work without hand-written PVs.
2. Auto-relax `ReadWriteMany` → `ReadWriteOnce` when the target is a detected single-node/local cluster (RWX is meaningless there anyway).
3. If neither is planned, correct the docs — they currently promise kind/k3d/minikube support that isn't in the code.

## Note — the registry-free path *does* work

Credit where due: leaving `image_registry` unset selects the "copy local source into the cluster via PVC + install jac at pod startup" path, and that works on kind with **no registry at all**. It's the single reason this was viable. It just collides with the mandatory-RWX requirement above.

marsninja:
## Summary

A `[scale]` deploy unconditionally required a `ReadWriteMany` bundle PVC plus an explicit `bundle_storage_class`, so single-node local clusters (kind / k3d / minikube), which ship no RWX provisioner out of the box, could not be targeted at all without hand-provisioning static hostPath PVs pre-bound to the generated claims. The docs meanwhile referenced a hostPath-style local-cluster escape hatch for the bundle store that did not exist in the source.

This ships the capability for real: a new `bundle_host_path` key under `[scale.kubernetes]` stages the source bundles in a directory on the host node instead of a PVC. On a single node every pod sees the same filesystem, so `ReadWriteMany` degenerates to a shared hostPath and no RWX StorageClass is required.

```toml
[scale.kubernetes]
bundle_host_path = "/mnt/jac/bundles"
```

## Changes

- `PvcInjector` gains `host_path`: the bundle volume becomes a hostPath mount on both the app pods and the bundle-loader pod, `extra_manifests()` emits no PVC, and the loader pod runs as root in this mode (kubelet never applies `fsGroup` to hostPath volumes and creates missing directories root-owned, so a non-root loader could not write the staged bundles).
- The loader pod body is extracted into `_loader_pod_manifest()` so both modes are unit-testable without a cluster.
- `select_injector()` threads the key through both the single-app and microservice deploy paths, and fails fast with an actionable `ValueError` when `bundle_storage_class` and `bundle_host_path` are both set (they are mutually exclusive).
- The single-app path skips `ensure_pvc_exists` in hostPath mode; `jac scale destroy` already tolerates the absent PVC.
- The missing-storage-class fail-fast error now points local-cluster users at `bundle_host_path` instead of dead-ending them.
- Docs (`jac-scale-kubernetes.md`): both bundle storage keys are now documented under Persistent Storage with a dedicated local-clusters section, the Source Distribution steps name the two bundle-store modes, and the troubleshooting entry that referenced a nonexistent bundle `host_path` option now describes the real key.

## Testing

New tests in `jac/jaclang/scale/tests/microservices/test_pvc_injector.jac` (all fail before this change, pass after; none contact a cluster):

- hostPath mode emits no bundle PVC and mounts the host directory in pod specs
- the loader pod mounts the host directory (and drops non-root only in that mode); PVC mode keeps the non-root loader and the PVC claim
- `bundle_storage_class` + `bundle_host_path` together raise a `ValueError`
- the missing-storage-class error mentions `bundle_host_path`
- `bundle_host_path` round-trips through `KubernetesConfig`

Targeted suites run clean: `test_pvc_injector` (16 passed), `test_binary_injector` (31 passed), `test_deploy_k8s` (24 passed, 1 pre-existing environmental failure identical on main), plus the manifest-builder/factory/dry-run suites (`test_service_pod_npm`, `test_shared_dep_cache`, `test_fleet_rollout`, `test_tempo_collection`, `test_database_provisioner`, `test_ingress_tls_preservation`, `test_factories`, `test_dry_run_purity`) all passing.

Fixes #7433


MusabMahmoodh (commented):
Same sweep as #7441. Read it start to end: the config plumbing, the mutual-exclusivity guard in `select_injector`, `PvcInjector`'s split between `_bundle_volume_source` (PVC vs hostPath) and `_loader_pod_manifest` (root vs non-root loader), and the manifest-builder wiring that skips PVC creation when host_path is set. `build_loader_pod` being deleted outright and folded into the instance method is real consolidation, not addition - one branch point for the loader manifest instead of two parallel pod-builders.

Confirmed against `main` this morning: `bundle_host_path` still does not exist anywhere in the tree, so nothing has superseded this in 43 days. It will need a rebase though - `kubernetes_config.jac` is now under `scale/deploy/target/kubernetes/`, and `select_injector` now lives in `scale/injector/binary_injector.jac` and `scale/deploy/target/kubernetes/manifest_builder.jac`.

**The one finding.** Nothing in the code stops `bundle_host_path` from being set on a multi-node cluster. The docs warn against it, and the failure mode is the quiet kind - a pod scheduled onto a different node than the loader simply never sees the seeded bundle - and this is also the only mode where the loader runs as root. Both are fine as documented local-only behavior, but neither `select_injector` nor `KubernetesConfig.from_dict` consults `cluster_type()` before allowing it, so a local `jac.toml` block copied into a real deploy fails partially rather than fast.

Is a `cluster_type() != "local"` fail-fast worth adding here, or is that deliberately left to the operator? Asking rather than asserting because #8405 has `cluster_type()` swallowing every exception and answering `"local"` - which would make that guard fail open, so it may not be worth wiring until that one is fixed.

Second, smaller: the new coverage constructs `PvcInjector` directly and asserts on the generated manifest dict, which is solid for the branching itself. Given the premise is "these clusters could not deploy at all before", was there a kind or k3d run behind the hostPath walkthrough in the docs? Not asking for it to block anything - I have a rig with local clusters and can run it against this branch if a trace would be useful.

PR #7441, opened 2026-07-13
fix(jac-scale): close the deploy.k8s capability over python-dotenv and sync it before first use

### LINKED ISSUE #7430: bug(scale): `jac start --scale` fails with ModuleNotFoundError (dotenv, bcrypt) — `deploy.k8s` capability dep set is incomplete
## Summary

On a machine where the `deploy.k8s` capability packages aren't already present, `jac start <app> --scale` fails with `ModuleNotFoundError`. jac auto-installs `docker` and `kubernetes` for the capability, but the deploy path also hard-imports **`dotenv`** and **`bcrypt`**, which are not in the capability's dependency set.

Environment: `jac 0.32.0 (Linux x86_64)`, deploying a `[scale.microservices]` app to Kubernetes.

## Repro

From a jac-scale project (any project that reaches the k8s realizer), remove the two packages from the jac global site to simulate a clean machine:

```console
$ jac -m pip uninstall -y python-dotenv bcrypt
  Successfully uninstalled python-dotenv-1.2.2

$ jac start main.jac --scale --dry-run
✖ Error: Error executing 'start': No module named 'dotenv'
ModuleNotFoundError: No module named 'dotenv'
✖ Error: No module named 'dotenv'
```

Traceback points at the realizer:

```
File ".../jaclang/scale/plugin.jac", line 125, in scale_start
    import from jaclang.scale.runtime.realizer.kubernetes_realizer { KubernetesRealizer }
File ".../jaclang/scale/runtime/realizer/kubernetes_realizer.jac", line 3, in <module>
    import from dotenv { load_dotenv }
ModuleNotFoundError: No module named 'dotenv'
```

With `dotenv` restored, the next failure is `bcrypt`:

```console
$ jac start main.jac --scale --dry-run
Deploy tooling missing; installing docker>=7.1.0,<8.0.0, kubernetes>=34.1.0,<35.0.0
  (capability 'deploy.k8s') into the jac global site.
[ERROR] Error: Deploy failed: No module named 'bcrypt'
```

Note the sequencing: jac **itself** announces it is installing the deploy tooling, and then immediately crashes on a module that tooling needs but never installed.

## Why declaring them in `jac.toml` does not help

The obvious workaround — declaring the deps in the project's `[dependencies]` — does **not** fix it. The scale plugin runs from the jac **runtime global site**, not the project venv, so a project-level dependency is invisible to it.

The affected project already had all of these declared:

```toml
[dependencies]
kubernetes = ">=34.1.0,<35.0.0"
docker = ">=7.1.0,<8.0.0"
python-dotenv = ">=1.0.0"
bcrypt = ">=4.0.0,<5.0.0"
```

…with a comment reading *"the binary cannot see a system pip install"* — and the deploy still failed with `ModuleNotFoundError`.

## Workaround

```console
$ jac install python-dotenv bcrypt --global
```

before every `jac start --scale`. (`--global` is required; `PYTHONPATH` does not work — jac runs isolated and ignores it.)

## Expected

Add `bcrypt` and `python-dotenv` to the `deploy.k8s` capability pins, alongside `docker` and `kubernetes`, so that the auto-install produces a working deploy environment.

## Secondary observation (lower confidence)

While bringing this up I also saw the jac global site get **reset at least once** — packages installed via `jac install --global` (and the `docker`/`kubernetes` that jac auto-installed) disappeared, and `~/.cache/jac/rt/<hash>/` was re-created. This produced confusing, seemingly random failures across otherwise identical invocations.

I could **not** reduce this to a reliable repro — three consecutive `--scale` runs on a minimal project did not reproduce it, so I'm reporting it only as a note rather than a separate issue. My best guess is that the first `Preparing the 'dev' jac runtime for pods` / rolling-dev-binary download re-extracts the runtime cache and wipes the global site, and it is stable afterwards because the binary is then cached. Flagging in case it's a known hazard.

The more serious aspect of that episode: on one such degraded run the deploy did **not** error — it silently fell back to **default scale config** (auto-derived routes, and `Bundle PVC ... no bundle_storage_class is set` even though `[scale.kubernetes] bundle_storage_class` *was* set). Silently deploying a different topology than the one configured is dangerous; a half-initialized runtime should fail loudly rather than degrade to defaults.

marsninja:
Fixes #7430

## What was still broken after #7425

#7425 closed the `deploy.k8s` capability over `bcrypt`, but the *first* failure in the report was `dotenv`, and that gap survived:

```
File ".../jaclang/scale/plugin.jac", line 125, in scale_start
    import from jaclang.scale.runtime.realizer.kubernetes_realizer { KubernetesRealizer }
File ".../jaclang/scale/runtime/realizer/kubernetes_realizer.jac", line 3, in <module>
    import from dotenv { load_dotenv }
ModuleNotFoundError: No module named 'dotenv'
```

Two compounding problems:

1. **The pin set was incomplete.** The deploy path imports `dotenv` in the realizer (loads the project `.env`) and in `kubernetes_utils` (`dotenv_values` bakes `.env` into the env Secret), but `python-dotenv` was declared only in the `serve` and `env` capabilities, never in `deploy.k8s`. It usually *looked* fine because serve-closure machines already had it.
2. **The sync ran too late.** Even with the pin, the auto-install lives in `DeploymentTargetFactory.create()`, and both dotenv imports execute *before* the factory runs (the realizer's at module import time, i.e. before `realize()` is even called). So a clean operator machine crashed before jac's own install had a chance to happen. This is exactly the sequencing the report calls out: jac announces it is installing the deploy tooling, then dies on a module that tooling needs but never installed.

As the report also notes, project `[dependencies]` cannot paper over this: the deploy path resolves from the jac global site, not the project venv.

## Fix

- **Pin** `python-dotenv >=1.2.1,<2.0.0` in `deploy.k8s` alongside `kubernetes`, `docker`, `bcrypt` (`jac/jaclang/project/capabilities.jac`).
- **Sync before first use.** `ensure_deploy_tooling()` is now public on the target factory, and every deploy entry point (`KubernetesRealizer.realize`, `deploy_destroy`, `deploy_status`) calls it before its first third-party import. The factory keeps its own idempotent call for direct users.
- **Point-of-use imports.** The realizer's and `kubernetes_utils`' module-scope dotenv imports move to their point of use (the same treatment #7425 gave bcrypt in `ingress.jac`), so importing the deploy path never requires the tooling. `load_env_variables` now returns early when no `.env` exists, so a `--dry-run` on a machine that has never installed the tooling still renders manifests.
- **Loud, actionable failure.** Where a dry run does need dotenv (an actual `.env` to read) on an unsynced machine, `require_optional("dotenv", "deploy.k8s")` raises the capability-oriented install guidance instead of a bare `ModuleNotFoundError`. Dry-run purity from #7425 is preserved: a dry run still installs nothing.

## Tests

Every test failed before the fix and passes after (`jac/jaclang/scale/tests/deploy/test_dry_run_purity.jac`):

| Test | Failed before with |
|---|---|
| `deploy.k8s declares every dist the deploy path imports` | `deploy.k8s must close over 'python-dotenv': the deploy path imports 'dotenv' (runtime/realizer/kubernetes_realizer.jac:3)` |
| `importing the kubernetes deploy path does not require dotenv` | `the realizer must import dotenv at its point of use, not at module scope` |
| `a real deploy syncs deploy tooling before reading .env` | `module ... kubernetes_realizer does not have the attribute 'ensure_deploy_tooling'` |

The closure test is no longer a hardcoded membership list: it scans every `.jac` source reachable from the deploy path (`deploy/`, `injector/`, `runtime/cli/`, the kubernetes realizer) for third-party import roots and requires each to be a declared `deploy.k8s` pin, a documented transitive of one (`yaml`/`requests`/`urllib3` ship with the kubernetes client), or guarded behind `require_optional` for its owning capability (`pymongo`/`redis`). New imports that would recreate this bug now fail the suite by construction.

`jac/tests/project/test_capabilities.jac` updated where it hardcoded the exact `deploy.k8s` dist list.

## Verification

- `jac test jac/jaclang/scale/tests/deploy/test_dry_run_purity.jac`: 9 passed (3 failed before the fix, 6 pre-existing kept green, including both #7425 dry-run purity guards and `a real deploy still syncs missing deploy tooling`).
- `jac test jac/tests/project/test_capabilities.jac`: 38 passed.
- `jac test jac/jaclang/scale/tests/deploy/test_factories.jac`: 27 passed.
- `jac test jac/jaclang/scale/tests/test_k8s_utils.jac`: 39 passed.
- `jac test jac/jaclang/scale/tests/misc/test_optional_deps.jac`: 26 passed; `jac test jac/jaclang/scale/tests/misc/test_hooks.jac`: 1 passed (with the serve closure installed, as CI does).
- `jac fmt --check --lintfix` clean on every changed file.
- Smoke: `jac start main.jac --scale --dry-run` from a scratch project proceeds past every dependency import and reaches manifest planning (it stops at pod-binary/cluster prep on this laptop, which needs a cluster).


MusabMahmoodh (commented):
Picked this up in the scale review sweep. Re-verified the premise against `main` this morning rather than against the diff, because the tree has moved a long way under this PR in 43 days.

**The gap this closes is still open.** `jac/jaclang/project/capabilities.jac:44-51` today:

```jac
"deploy.k8s": {
    "context": "deploy",
    "deps": {
        "kubernetes": ">=34.1.0,<35.0.0",
        "docker": ">=7.1.0,<8.0.0",
        "requests": ">=2.32.0,<3.0.0"
    }
},
```

No `python-dotenv`. It is pinned under `serve` and `env`, both `context: "runtime"`, so a machine that syncs only the deploy context still has no dotenv. That is exactly the crash this PR describes, and nothing since July has closed it.

**What has changed is where everything lives.** This will not rebase cleanly - the four files it edits have all moved:

- `kubernetes_utils.impl.jac` / `.jac` -> `jac/jaclang/scale/deploy/target/kubernetes/utils/`
- `kubernetes_realizer.jac` -> `jac/jaclang/scale/runtime/realizer/`
- `CAPABILITY_SPECS` -> `jac/jaclang/project/capabilities.jac` (it is no longer in the plugin file this PR edits)

Also worth knowing before you rebase: #8334 (merged Aug 19) already did the point-of-use dotenv move in `kubernetes_utils`, with a capability-guarded reader that names the missing capability instead of dying at the import line. So that third of this diff is redundant now - #8334 made the failure legible, it did not close the dependency gap.

The two parts still worth landing, and I think they are the valuable two:

1. the actual `python-dotenv` pin in the `deploy.k8s` spec, and
2. `_deploy_path_third_party_imports` - the closure scan that walks every source under deploy/injector/runtime/cli and asserts each third-party import root is pinned, transitive, or guarded by another capability. That is a class gate rather than an instance fix; it would have caught the bcrypt and dotenv gaps as one thing instead of two separate fires.

One thing I checked and it holds: the ordering fix (`ensure_deploy_tooling` before the first third-party import in realize/destroy/status) matches the stated crash mechanism, and the "syncs before reading env" test asserts the sync precedes `load_dotenv` by call order, which is the right way to pin ordering.

Want me to sketch the post-rebase diff against current main so you can see how much of it survives, or are you taking it from here?

Issue #7255, opened 2026-07-07
[Feature]: One artifact, one gate, a clean CLI — `.jab` rebased on the sealed image + intent-organized verbs (supersedes #6280/#6335)


marsninja:
## Summary

Two moves, one plan:

1. **Make `.jab` real by generalizing the sealed image that landed in #7142** — not by building the parallel whole-program JIR container specified in #6280/#6335. Every valid Jac project builds, through one pipeline and behind one fail-closed whole-program type-check gate, into one self-describing sealed artifact that runs with **zero live compilation**. Wheel / npm tarball / static binary / client shells / ejected source are *projections* of that artifact, not sibling build paths.
2. **A clean break on the CLI, reorganized by intent.** ~48 top-level verbs (43 core + plugin injections) become ~22. Verbs name intents; formats become flags; plugins are confined to one namespaced noun each; removed verbs tombstone for exactly one release, then vanish.

The one-sentence model: **`jac build` type-checks the whole program and seals it into one `.jab`; `jac run` / `jac start` consume it with zero compilation; everything else you might ship is `jac build --as …`.**

**Supersedes #6280 and #6335** (and closes their stale PRs #6312/#6345). Their thesis — one universal, type-aware, kind-carrying artifact — is kept wholesale. Their physical format (a new whole-program JIR container: `FLAG_PROGRAM`, inner-framed `SEC_MODULE_TABLE`, `SEC_CLIENT`, Tier-0 hub preload) is dropped: none of it landed, and the sealed image shipped in the meantime delivers the same guarantees with a simpler mechanism that jaclang itself now boots from. The `kind`/capability ideas prototyped in #6345 are folded into Manifest v3 below.

## Ground truth (verified against the tree, 2026-07-07)

**What landed — the substrate to generalize (#6852 Phase 4 → #7135 → #7142):**

- The **sealed image**: `_precompiled/MANIFEST.json` (format 2) maps module fullnames → per-module `.jir` + sha256; resolution is name-keyed ahead of any filesystem probing; every JIR is hash-verified at `register_image` (`jac0core/sealed.py:302`); **fail-closed** — a bad image fails the *build*, never degrades to live compilation. Manifest fields today: `format`, `package`, `python_tag`, `jir_format_version`, `jaclang_version`, `modules` (`utils/precompile_bytecode.jac:310`). **No `kind` field.**
[...]

PR #7447, opened 2026-07-14
perf(cli): lazy CLI startup — manifest routing, dispatch, and lazy init

chess10kp:
## Summary

Refactors the Jac CLI to keep a single Jac-native entry path while lazily loading only what each invocation needs. Phases 0–3 land as incremental commits:

| Phase | Commit | Change |
|-------|--------|--------|
| 0/1 | ed1ffc4c3 | Manifest-backed info paths (`--version`/`--help`/bare/tombstone/unknown) before `registry.finalize()` |
| 2 | bc5a1208b | Lazy selected-command dispatch — known commands skip `register_feature_commands()`/`finalize()` |
| 3 | 15e1821fc | Lazy `jaclang.__init__` (PEP 562 exports) + manifest routing in `cli_boot` before `cli.cli` import |

Earlier commits on this branch defer feature-plugin imports and fast-path `--version` in `cli_boot`.

## Performance (warm dev tree)

| Path | Before | After |
|------|--------|-------|
| `import jaclang` | ~359ms | ~35ms |
| `jac --version` | ~945ms | ~360ms |
| `jac --help` | ~1157ms | ~380ms |

Phase 2 alone was performance-neutral on warm cache because `import jaclang` still paid for runtime+compiler; Phase 3 removes that floor.

## Test plan

- [x] `pytest jac/tests/cli/test_startup_isolation.py` — import isolation for `--version`, `--help`, `purge --help`
- [x] `jac --version` / `jac --help` byte-identical to legacy for all visible commands (warm cache)
- [x] `guide`, `check` execute end-to-end through lazy dispatch
- [x] `from jaclang import JacRuntime, JacRuntimeInterface, compiler` backward compat
- [x] Manifest regenerated (includes `build --fat` from #7465)
- [x] Release note updated in `7447.feature.md`


marsninja:
This could be fine for these commands, but would want to really deep dive why things are slow for a more root cause fix.

marsninja:
Really like where this landed — the two-layer shape (lazy full definitions + a cheap always-loaded projection for routing/help) is the *correct* architecture for the startup contract, and the budget/isolation tests locking it in are great. I took a deep pass over the whole metadata/config surface with an eye toward where this design should go next, and I want to write it down here because I think we're one step away from a genuine single-source-of-truth CLI — and that step also deletes a lot of code.

## Where command metadata lives after this PR

Five places:

1. **Definitions** — `@registry.command(...)` + `Arg.create(...)` in `cli/commands/*.jac` and the feature modules (scale/byllm/shadcn/client). Full truth for args/help/examples; lazily imported per verb. 193 `Arg.create` sites total.
2. **The route table** — 32 hand-authored `CommandMeta` rows in `cli/impl/manifest.impl.jac` (name, help, group, tier, hidden, handler_module, handler_name, needs).
3. **Help** — rendered from the manifest by `manifest_help` (zero command imports — nice).
4. **MCP bridge** — `mcp_manifest`: `list_commands` reads the manifest, but `get_command` lazy-imports the handler module just to read the arg schema back out of the registry.
5. **Sync** — `test_manifest_parity.jac`, which imports every command module and asserts both directions agree field-by-field.

The observation that reframes everything: **the parity test already does all the work of a generator — it just checks instead of writes.** The route table is a projection of the decorators, maintained by hand, with a test as the only thing keeping it honest.

## The config side has the same disease, smaller and worse

The only declarative flag↔`jac.toml` link is the `bridges` table in `_apply_config_defaults` (`dispatch.impl.jac`): four hand-paired tuples (`port`↔`[serve].port`, `cache`, `main`, `verbose`), each restating the flag's built-in default as a sentinel — so `8000` lives in both `Arg.create("port", default=8000, ...)` and the bridge tuple, with no test guarding the pair. Everything else is ad-hoc `get_config()` inside handlers (fine, honestly).

## Proposal: promote the projection from hand-authored to generated

We already have this exact idiom twice in-tree: `jac gen-jir-registry` and `jac gen-uni-dispatch` — committed generated artifacts with freshness enforcement. The manifest is the third instance of the pattern, just missing its generator.

1. **`jac gen-cli-manifest`** — does what the parity test does (import all command modules, walk the registry) and *emits* `manifest.impl.jac`. The parity test collapses to "regenerate == committed". Adding a command becomes: write the decorator, run the generator. One edit site. CI's existing autofix push could even regen it.
2. **`needs=` moves into the decorator.** Today `needs` (templates/providers/client) exists *only* in manifest rows, which means the definition site is currently incomplete — the decorator can't fully describe its own command. Moving it makes the decorator the complete truth.
3. **`config_key=` (optionally `env=`) on `Arg`** — e.g. `Arg.create("port", typ=int, default=8000, config_key="serve.port")`. The bridges table becomes a generic loop over spec args; the duplicated sentinel disappears; precedence (explicit flag > toml > default) is implemented once; help can show provenance ("overridable via `[serve] port`"); and config-default support scales from 4 args to all 193 for free.
4. **Bake full arg schemas into the generated manifest.** I checked: every `Arg` default in the tree is a literal, so this is mechanically safe. This unlocks the deep cut below.

## What a clean break gets to delete

**Tier 0 — dead already, could ride on this PR:**
- `cli/help.jac` + `cli/impl/help.impl.jac` (219 lines) — the old `HelpFormatter` was fully superseded by `manifest_help` but left in place; its only remaining consumer is its own test (`test_help_color.jac`).
- `_pending_extensions` in `registry.impl.jac` — written once, read nowhere.

**Tier 1 — removed by the generator:**
- Hand-maintenance of the 334-line route table (file stays as a committed generated artifact).
- The guts of `test_manifest_parity.jac` (~50 of 62 lines) → a freshness diff.
- `validate_manifest` as runtime code (~40 lines) — it's only called by the parity test today; duplicate/budget/tombstone/tier checks become generator-time build failures.
- `register_feature_commands` / `feature_command_modules` / `_CORE_REGISTRATION_MODULES` (~40 lines) — the eager-import machinery exists solely for tests; the generator *is* that job.
- `test_verb_surface.jac` shrinks — the #7255 budget/tombstone policy stays, enforced at generation time.

**Tier 2 — removed by `config_key`:**
- The `bridges` table (~20 lines, but the worst code-per-line in the system).

**Tier 3 — the deep cut, once schemas are in the manifest:**
- `_load_command_spec` dies; one loader path (`load_handler`), not two.
- `CommandRegistry`'s *runtime* machinery (singleton, commands dict, get/get_all, registration side effects — roughly half of `registry.impl.jac`'s 271 lines) stops existing at runtime; the decorator becomes inert metadata only the generator reads. `Arg`/`ArgKind`/`CommandSpec` survive as the declaration vocabulary; `Arg`→argparse translation moves to dispatch and reads manifest rows.
- MCP `get_command`'s deep path (~50 lines) — serves schemas from static data; the whole MCP command surface goes zero-import.
- Structural win: `jac <cmd> --help` stops importing the command module at all — the budget contract tightens from "≤4 imports" to **0**, and chunks of `test_command_isolation.jac`'s probe harness lose their reason to exist.

**Ledger:** ~500+ lines of authored logic deleted, 334 lines of hand-maintained data converted to generated, against ~150 lines of new generator (most of whose logic already exists inside the parity test). What stays hand-authored, correctly: `TOMBSTONED_VERBS`, `VERB_BUDGET`, group/tier ordering — those are decisions, not derivations.

## Sequencing

None of this needs to block this PR — the two-layer architecture it introduces is exactly right, and the tiers above are a natural follow-up sequence (Tier 0 is a trivial cleanup that could land here or immediately after; Tiers 1–2 are the clean break proper; Tier 3 once the generator has proven itself). Happy to discuss / split issues for the tiers if this direction sounds right.

PR #7467, opened 2026-07-15
feat(cli): fat jab vendors the Python dependency closure into the .jab bundle (#7465)

### LINKED ISSUE #7465: Fat jab: vendor the Python dependency closure into the .jab bundle
## Motivation

A Spring Boot executable jar nests every dependency jar under `BOOT-INF/lib/` inside the archive, so the app is self-contained at the dependency level: copy one file to any machine with a JVM and it runs, no `mvn install`, no network.

A `.jab` today is close to that but not quite. It already bundles the precompiled app, the prebuilt client dist, and native binaries, all hash-sealed. But Python dependencies are only *declared* in the sealed manifest as `capabilities`; they get pip-installed on the target at run/deploy time. So running a jab still assumes the target can reach PyPI (or has the deps preinstalled), and in jac-scale every pod pays the capability install cost at startup.

Proposal: a **fat jab** that vendors the Python dependency closure into the bundle, the same way Spring Boot vendors jars.

## Current state

- `jac build` emits `dist/<app>.jab`: a tarball with `_precompiled/` (sealed manifest, compiled modules), client dist, native binaries.
- Manifest carries `kind`, `entry`, `capabilities`, `package` (`jaclang/runtimelib/jab.jac`).
- `jac run` / `jac start` materialize the jab into a content-addressed cache dir and execute; capability deps are resolved outside the bundle.
- The wheel projection (`jac build --as wheel`) already knows how to vendor the jac runtime itself, so we have precedent for vendoring, just not for third-party deps in the jab path.

## Proposal

- `jac build --fat` (or a `[build]` toml flag) resolves the capability closure at build time and packs the wheels into the archive, e.g. `_vendor/wheels/`.
- At materialize time, if `_vendor/wheels/` is present, install from it with `--no-index --find-links` into the app's cache-scoped site dir (or make it importable directly), and skip the normal capability install entirely.
- The vendored wheels are inside the tarball, so they are covered by the existing sha256 content addressing for free: bump a dep, get a new digest, new cache dir. No stale-dep aliasing.
- jac-scale: when the runtime bootstrap sees a fat jab, pods skip the capability install step. Faster cold start and no PyPI dependency inside the cluster (same spirit as the gateway skipping the npm closure when the bundle ships a prebuilt client dist).

## Open questions

1. **Platform-specific wheels.** Pure-python wheels are trivially portable; native wheels (numpy etc.) are tagged per platform/arch/python. Options: (a) vendor for a target platform triple chosen at build time (matches how we already treat `.jir` as version-locked to the building binary), (b) vendor multiple platforms behind a `--platforms` flag, (c) fat-vendor only pure wheels and fall back to install for native ones. I lean (a) with (c) as the fallback.
2. **Size.** A fat jab with a heavy ML stack could be hundreds of MB. Probably fine (it is opt-in and Spring Boot fat jars have the same property), but we should print the vendored size in the build summary.
3. **Default or opt-in?** Spring Boot made fat the default and nobody looked back. Starting opt-in seems safer given the size question, could flip later.
4. **Interaction with `--as binary`.** The binary projection appends the jab to the launcher; a fat jab would make that artifact truly offline-capable end to end. Worth verifying nothing assumes the jab payload stays small.
5. **Resolution environment.** Wheels should be resolved with the same python version the sealed runtime targets, not whatever happens to be on the build host.

## Risks

- Cross-platform builds (build on mac, deploy to linux pods) need pip's `--platform`/`--only-binary` download mode rather than resolving from the local env, or the vendored wheels will be wrong.
- License/redistribution: bundling wheels is redistribution; same situation as every fat-jar ecosystem, but worth a note in docs.
- Cache dir growth: vendored site dirs multiply per digest; may need the app cache GC to account for them.

marsninja:
## Fat jab: vendor the Python dependency closure into the `.jab` bundle

Closes #7465.

Today a `.jab` bundles the precompiled app, prebuilt client dist, and native binaries, all hash-sealed, but its Python dependencies are only *declared*; they get pip-installed on the target at run/deploy time. So running a jab still assumes the target can reach PyPI, and in jac-scale every pod pays the capability install cost at startup.

This adds a **fat jab** that vendors the Python dependency closure into the bundle, the same way a Spring Boot fat jar nests every dependency jar.

### What it does

- **`jac build --fat`** (jab and binary projections) resolves the app's runtime Python closure and packs the wheels into the archive under `_vendor/wheels/`, with a `_vendor/VENDOR.json` sidecar recording what was vendored. The build summary prints the vendored wheel count and size.
- **At materialize time**, if `_vendor/wheels/` is present, the wheels install offline into a cache-scoped `_vendor/site/` and that dir goes on `sys.path`, so the app imports its deps with no PyPI access. The install runs once per content-addressed image (guarded by a ready-marker).

### Design notes

- `_vendor/` is a **sibling of `_precompiled/`**, so the sealed manifest's hash-checked module tree is left untouched. The wheels ride inside the tarball, so the existing sha256 content addressing covers them for free: bump a dep, get a new digest, get a new cache dir, with no stale-dep aliasing.
- The build side **reuses `resolve_plan`**, the same resolver `jac install` uses, so the vendored set is exactly the runtime closure: declared deps plus the capability deps derived from `jac.toml` intents. Git deps are skipped (a wheel download cannot represent them) and still install normally.
- Both halves use `--only-binary=:all:` / `--no-index` so they **fail closed** rather than silently reach the network or ship an sdist that cannot install offline. Cross-platform resolution is a pip concern (honors `PIP_INDEX_URL` / `PIP_FIND_LINKS` / `PIP_NO_INDEX`; `--platform` / `--python-version` when a target triple is chosen). The default resolves for the build host, matching how `.jir` is version-locked to the building binary (open question 1, option a).
- New vendoring primitives live in a small, dependency-light `jaclang/runtimelib/jab_vendor.jac` so they are importable on the materialize path alongside `jab.jac`.

### Scope / follow-ups

This lands the core build → materialize → offline-import mechanism for local `jac run <app>.jab` and `--as binary`. The jac-scale bootstrap skip (pods reading a fat jab and skipping the capability install step) is a natural follow-up on top of the `VENDOR.json` marker.

### Tests

`jac/tests/publish/test_fat_jab.jac` (network-free, builds against a local wheelhouse):

1. `jac build --fat` vendors the dependency closure into the jab (`_vendor/wheels/*.whl` + `_vendor/VENDOR.json`).
2. A fat jab runs offline (`PIP_NO_INDEX=1`, wheelhouse removed) and imports its vendored dependency.


marsninja:
Thanks for the review. Addressed all four in the latest commit:

- **P1 silent failure (jab.jac):** `activate_vendored` now fails closed. A failed offline wheel install raises `RuntimeError` (surfaced by the materialize path, e.g. `Cannot load app.jab: ...`) instead of returning `None` and letting the app start only to `ModuleNotFoundError` on first import.
- **P2 race on the site-ready marker (jab_vendor.jac):** `_install_offline` now no-ops when the site dir already exists and tolerates a lost `rename` race (another process won), so concurrent materializations against a shared cache -- the jac-scale multi-pod case -- no longer end up without vendored deps on `sys.path`.
- **P2 missing `wheels_dir` (jab_vendor.jac):** `write_vendor_manifest` guards the glob with an `is_dir()` check, so a zero-dep fat jab no longer relies on `Path.glob` swallowing a missing directory.
- **P2 wording (build.impl.jac):** the message now reads "`--fat` is not supported for the '<projection>' projection" to match the hard error.

Also fixed the failing `jac-check` (the repo's deslop lint wanted docstrings/comments stripped from the new non-test `.jac` files) and documented `--fat` in the `jac build` CLI reference.

PR #7478, opened 2026-07-15
fix(skills): clarify usage of sv import and type construction in Skills

SahanUday:
## Summary

Fixes gaps identified in `issues/skill-issue.md` — an audit of the AI-facing skill docs in `jac/jaclang/cli/skills/` — where a real compiler failure mode (misusing `sv import`ed archetypes / the `new()` builtin) wasn't documented anywhere, leaving agents to discover it the hard way.

All changes are additive caveats/cross-references next to existing correct content — no existing guidance was changed or removed.

## Changes

- **`jac-cl-js-interop.md`** — warns that `new()` must never be used for user-defined `obj`/`node`/`edge` types. It silently produces an instance with every field `null` instead of throwing, because `new()` splats args positionally while Jac constructors expect a single `props` object.
- **`jac-fullstack-patterns.md`** — adds the core rule: a type pulled in via `sv import` is for *type-annotating* data only, never for constructing an instance directly. Doing so compiles into a broken RPC call to a nonexistent endpoint (looks like an auth `401`/`404`, not a construction error). Also cross-references the existing `ModuleNotFoundError` troubleshooting entry in `jac-sv-microservices.md`, which wasn't linked from anywhere before.
- **`jac-cl-components.md`** — one-line pointer to the same "never construct" rule, next to the existing `sv`-typed-state guidance.
- **`jac-sv-microservices.md`** — clarifies that the "construct = remote execution" pattern is unique to `walker:pub`; plain `obj`/`node`/`edge` types are never constructed via their `sv import`ed name, in either direction.
- **`jac-cl-organization.md`** — note at the point where the hook example swaps a locally-declared type for an `sv import`ed one, flagging that the swap changes what's safe to do with that name.

## Why

These docs are what coding agents (e.g. JacCoder) read to learn Jac conventions. The gap was narrow but costly: the failure modes involved (silent data corruption, misleading `401`/`404`) have no compile-time signal, so an agent following the docs as written had no way to know it had crossed an unsupported line until runtime.

Full investigation and rationale: `issues/skill-issue.md`.


marsninja:

PR #7489, opened 2026-07-15
fix(scale): RedisEventStream self-heals after a stream key is evicted (#7487)

### LINKED ISSUE #7487: jac-scale: RedisEventStream stops delivering events after a stream is evicted (NOGROUP)
## Summary

`jaclang.scale.events.streams.redis.RedisEventStream` stops delivering events on a stream forever once that stream's Redis key is evicted. When a stream key is dropped (e.g. under `allkeys-lru` memory pressure) its consumer groups are destroyed with it; a later `XADD` recreates the stream without the group, but the broker still has the group memoized in `_ensured_groups`, so it skips group creation and every `XREADGROUP` returns `NOGROUP`. `consume()` then returns `[]` on every call and (because `NOGROUP` fails instantly, so the `block=` timeout does not apply) the surrounding read loops spin with no backoff, flooding logs and pinning a CPU core. The root cause is a poisoned, never-invalidated group cache in `_ensure_group`.

## Environment

- `jac 0.32.0 (Darwin arm64)`, jaseci-labs/jac `origin/main` @ `afd43a9bc`.
- Redis 7 (repro uses a local `redis://localhost:6379`).
- Reproduces on any deployment whose Redis can evict stream keys (memory pressure or an explicit `maxmemory` + `allkeys-lru` policy).

`nogroup_repro.jac`, run against `origin/main`:

```jac
import from jaclang.scale.events.broker { Event }
import from jaclang.scale.events.streams.redis { RedisEventStream }
import from jaclang.scale.persistence.db { get_redis_client }

with entry {
    url = "redis://localhost:6379/0";
    cfg = {"url": url, "consumer_group": "t_evict",
           "retry": {"max_attempts": 3, "backoff_seconds": [0.01, 0.01, 0.01], "dead_letter_suffix": ".dlq"}};
    b = RedisEventStream(config=cfg);
    b.start();
    client = get_redis_client(url);
    client.delete("evict.topic");

    # 1) normal: publish + consume creates the consumer group and caches it
    b.publish("evict.topic", Event(event_type="e.one", data={"n": 1}));
    got1 = b.consume("evict.topic", group="t_evict", max_messages=1, timeout_seconds=2.0, start_from="earliest");
    print(f"[step1] got1={len(got1)}  (expect 1)");

    # 2) Redis evicts the stream key -> destroys the stream AND its consumer group
    client.delete("evict.topic");

    # 3) a later publish re-creates the stream via XADD, but NOT the group
    b.publish("evict.topic", Event(event_type="e.two", data={"n": 2}));

    # 4) consume: the group is still cached as 'ensured', so XREADGROUP -> NOGROUP
    got2 = b.consume("evict.topic", group="t_evict", max_messages=1, timeout_seconds=2.0, start_from="earliest");
    print(f"[step4] got2={len(got2)}  (expect 1; bug = 0)");
    b.stop(drain=False);
}
```

Result:

```
ERROR - XREADGROUP failed for evict.topic/t_evict: NOGROUP No such key 'evict.topic' or consumer group 't_evict' in XREADGROUP with GROUP option
[step1] got1=1   (expect 1)
[step4] got2=0   (expect 1; bug = 0)
```

Step 1 delivers normally. After the stream key is evicted (step 2) and re-created by a publisher (step 3), `consume()` returns empty forever (step 4). In a running service the read loop re-invokes `consume()` immediately, so this becomes a tight `NOGROUP` error loop rather than a single empty return.

## Expected

After a stream key is evicted and re-created, `consume()` should detect the missing group, recreate it (`MKSTREAM`), and resume delivering events. A destroyed group is a recoverable condition, not a permanent dead stream.

## Actual

`consume()` returns `[]` on every call and logs `XREADGROUP failed ... NOGROUP` on every call. Root cause, code-anchored (`jac/jaclang/scale/events/streams/redis.jac` @ `afd43a9bc`):

- `_ensure_group` (line ~109) memoizes `f"{topic}::{group}"` into `self._ensured_groups` **unconditionally** at the end of the function (line ~123), including when `xgroup_create` raised something other than `BUSYGROUP`. Once cached, `_ensure_group` short-circuits on the stale entry and never attempts to recreate the group.
- `consume` (line ~163) calls `XREADGROUP` and, on any exception, logs at ERROR and returns `[]`. It does not special-case `NOGROUP`, does not invalidate the cache, and does not back off. Because `NOGROUP` fails instantly, the `block=` timeout provides no pacing, so callers that re-invoke on empty results hot-loop.

## Impact

This is the streaming layer jacBuilder's AI-chat uses. The chat event drain path is:

`browser WS ai_events` -> `ideServer.drain_events` -> `jaccoder_client._drain_run_events` -> `broker.consume(jc.run.{run_id}, group=conn:{cid})`, where `broker` is exactly this `RedisEventStream` (`services/jaccoder_client.jac:17,24,25,706`).

So when a chat run's stream (`jc.run.<id>` / `bridge.run.<id>`) is evicted mid-turn, `consume()` returns empty and the browser's poll receives no further events: **the live chat log stops updating mid-turn while the agent is still working**, with no recovery until the process restarts. Separately, on prod this produced ~600 ERROR log lines/sec per active run (77k+ in a 15 min window) and pinned CPU on the coder service. The failure is invisible to the higher-level relay/resume logic because `consume()` swallows the error and returns `[]` rather than raising.

## Workaround

Not fully verified in a live deployment. Mechanism-implied only: since the trigger is stream-key eviction, avoiding eviction (raising Redis `maxmemory`, or not using an `allkeys-lru` policy on the broker's Redis) prevents the group from being destroyed and so avoids the bug. This does not fix the underlying non-recovery.

## Suggested fix

In `RedisEventStream` (`jac/jaclang/scale/events/streams/redis.jac`):

1. `_ensure_group`: memoize **only** on a clean `xgroup_create` or a `BUSYGROUP` error. Never cache a `(topic, group)` whose create genuinely failed, so the next call retries the create. Add a `force` flag to bypass the cache for an explicit recreation.
2. `consume`: on a `NOGROUP` error, discard the stale `_ensured_groups` entry, recreate the group with `MKSTREAM`, and retry once. On any unrecoverable error, pace the return (e.g. a short sleep) so callers (`consumer_loop`, and jac-coder / jacBuilder `stream_events` / `drain_events`) cannot hot-loop.

Regression test: publish + consume (creates the group), `DEL` the stream key to simulate eviction, re-publish, then assert `consume` self-heals and delivers the next event. The executed repro above is the basis for it.

## Related

- Introduced with `RedisEventStream` in #5836.
- Amplified by the Redis `allkeys-lru` eviction in jaseci-labs/jacBuilder#551; surfaces as the mid-turn chat-log drop in jaseci-labs/jacBuilder#687.

akindu-k:
## The problem

`RedisEventStream` would stop delivering events on a stream **forever** once that stream's Redis key was evicted, and there was no recovery short of restarting the process. Fixes #7487.

### Why it happened

Redis Streams store a topic as a key (e.g. `jc.run.<id>`), and a **consumer group** (Redis' per-reader bookmark) lives *inside* that key. Two facts combine badly:

1. If the stream key is deleted, its consumer groups are destroyed with it.
2. Redis can evict keys on its own under memory pressure (`allkeys-lru`) — nobody has to call `delete`.

`RedisEventStream` kept an in-memory cache, `_ensured_groups`, as a shortcut meaning "I've already created this group, no need to ask Redis again." **That cache was never invalidated.** So the failure unfolded as:

1. A consumer reads a stream → the group is created and cached. Works fine.
2. Redis evicts the stream key → the group is destroyed, but the cache still says "exists".
3. A publisher writes again → `XADD` re-creates the stream key, but **not** the group.
4. A consumer calls `consume()` → the cache short-circuits group creation, so `XREADGROUP` hits a group that no longer exists and Redis returns `NOGROUP`.

From there the stream was dead: `consume()` caught the error and returned `[]`, the read loop immediately called `consume()` again, and because `NOGROUP` fails *instantly* the blocking timeout never applied — so it became a tight spin loop.

### Impact

This is the streaming layer jacBuilder's AI-chat drains through. When a chat run's stream was evicted mid-turn, `consume()` returned empty and the browser stopped receiving events: **the live chat log froze mid-answer while the agent was still working**, with no recovery until restart. On prod this also produced ~600 ERROR lines/sec per active run and pinned a CPU core.

## The fix

Two ideas: **stop trusting the cache blindly**, and **self-heal on `NOGROUP`** instead of dying.

- **`_ensure_group` only caches a group when Redis confirms it exists.** A clean create or a `BUSYGROUP` ("already exists") response is cached; any other failure now discards the cache entry and returns, so the next call retries instead of trusting a stale "exists". Added a `force` flag to bypass the cache for an explicit recreate.

- **New `_read_group` helper wraps the read with recovery.** On a `NOGROUP` error it forgets the stale cache entry, recreates the group (`force=True`, `MKSTREAM`), and retries the read once — so an evicted-and-recreated stream resumes delivering automatically. Any other error is re-raised. The raw Redis call is factored into `_xreadgroup` so both the first attempt and the retry share one path.

- **`consume` paces on unrecoverable errors.** If a read still fails after recovery, it backs off with the existing shutdown-aware `wait_or_stop` (0.5s) before returning empty, so callers can no longer hot-loop and flood the logs. The normal "empty stream / timed out" path is unchanged and adds no latency.

### Follow-up (from review): honor the handler's `start_from` on recovery

Background handlers registered via `register_handler(..., start_from="earliest")` run their read loop in the shared `consumer_loop`, which called `consume()` **without** `start_from` — so it always defaulted to `"latest"`. Harmless in steady state (the group already exists), but on the new recovery path `start_from` decides where the recreated group begins. An `"earliest"` handler would have its group silently recreated as `"latest"` after an eviction, dropping messages already sitting in the re-created stream. `start_from` is now threaded through `consumer_loop` (and both `register_handler` callers) so recovery recreates the group at the position the handler actually asked for. `local.jac` is behavior-unchanged (it can't be evicted) and is kept consistent.

## Verification

Ran the issue's exact repro against Redis 7, comparing `main` vs this branch:

| | after evict + republish |
|---|---|
| `main` | `consume()` returns empty (`got2=0`), logs `NOGROUP` every call |
| this branch | recovers and delivers the event (`got2=1`), one "recreating" warning |

Tests:
- `test_redis_stream.jac` — 10 passed, incl. a new end-to-end self-heal test (publish + consume, `DEL` the key to simulate eviction, re-publish, assert `consume` delivers the next event).
- `test_events.jac` — 15 passed, incl. two deterministic tests that `consumer_loop` forwards `start_from` (and defaults to `latest`).
- `test_local_stream.jac` — 8 passed (unchanged), confirming the shared-loop change is safe for the in-memory broker.


udithishanka (approved):
Approve. Correct fix for the root cause and verified end to end.

Verified: ran the issue's exact repro against this branch (Redis 7). After evict + republish, `consume()` recovers and delivers (`got2=1`) with a single "recreating" warning, vs `got2=0` + a `NOGROUP` flood on `main`.

Logic reviewed:
- `_ensure_group` caches only on a clean create or `BUSYGROUP`; discards the entry on any other failure, so the poisoned-cache condition can no longer occur. `force` bypass for recreate is right.
- `_read_group` recreates on `NOGROUP` (`force=True`, `MKSTREAM`) and retries once; non-`NOGROUP` errors re-raise.
- `consume` paces with `wait_or_stop(0.5s)` on an unrecoverable read, so the `NOGROUP` hot-loop is gone; happy path unchanged.
- Threading `start_from` through `consumer_loop` is a good extra catch: an `earliest` handler would otherwise have its group silently recreated as `latest` on recovery and drop messages.

Nits (non-blocking):
- Recovery restores forward delivery only; events already evicted are unrecoverable (inherent).
- `"NOGROUP" in str(e)` string-match is fragile but matches redis-py convention (no dedicated exception class).

Nice work.

akindu-k:
image

akindu-k:
image

akindu-k:
Learning ref: https://chatgpt.com/share/6a57d36a-65a0-83ee-a56d-5999d1b31c75

PR #7504, opened 2026-07-16
fix(ci): skip whole-program gate for the client-only admin-dist build

Thamirawaran:
The admin-dist job runs `jac build main.jac`, whose whole-program type-check gate runs the native checker over the console's `.cl.jac` client code -- before the client bundle step installs the npm deps, and with native (not JS) type semantics. So `react`, `document`, and JS array/string methods surface as errors and fail the dev-binary release.

The admin console is a pure client app; its client bundle is type-checked by the client toolchain during the same build. Skip the gate for this step with `--no_typecheck`.


kugesan1105:
It's a temp escape to do release

PR #7522, opened 2026-07-16
scale: fat bundle support - pods materialize vendored wheels offline

MusabMahmoodh:
Implements the jac-scale half of #7465 (fat jab; the CLI side landed in #7467).

## Problem

`jac build --fat` vendors the wheel closure into the bundle, but jac-scale
does not go through `jac build`: `pack_jab` seals a staged tree and tars it
directly, and pods unpack that tar with a plain `tar xzf` (never
`materialize_jab`, so `activate_vendored` never runs). Every pod still
pip-installs the full closure from PyPI at bootstrap.

## Change (`[scale] fat_bundle = false`, default true)

- **pack_jab** (`injector/bundle.jac`): when enabled, stages
  `_vendor/wheels/*.whl` + `_vendor/VENDOR.json` into the sealed tree
  (sibling of `_precompiled/`, same layout as `jac build --fat`) and tars it,
  so the wheels ride the existing content digest for free (the deterministic
  tar + sha256 mechanism is unchanged).
- **vendor_wheels.jac** (new, `scale/utils/`): runs under the staged
  `.pod-jac` binary - the SAME binary that seals the app and that pods run -
  against the SAME sanitized jac.toml the pod unpacks (the sanitizer can bake
  a provisioned-database intent that widens the capability closure), so the
  vendored closure cannot drift from what the pod's `jac install` resolves.
  Non-pypi deps are warned and left to runtime install; a missing wheel fails
  the deploy loudly.
- **jac-bootstrap** (`runtime_bootstrap.jac`): when
  `/app/_vendor/wheels/*.whl` exists, exports `PIP_NO_INDEX=1
  PIP_FIND_LINKS=/app/_vendor/wheels` and runs the unchanged `jac install`.
  Same venv, same closure, zero PyPI traffic; an incomplete closure fails the
  boot instead of silently reaching for the network. Thin bundles keep the
  exact old path.

## Testing

- 4 new tests in `scale/tests/microservices/test_fat_bundle.jac`: config
  parsing; thin bundle ships no `_vendor`; fat bundle stages wheels with
  `VENDOR.json python_tag == MANIFEST.json python_tag` (same-binary
  invariant); digest deterministic across identical rebuilds and moves when a
  pinned wheel changes.
- Neighboring suite (app_precompile, pvc_injector, binary_injector,
  service_pod_npm, shared_dep_cache): 52 passed. The full-scale-suite
  failures (gcs/firestore/gateway-metrics/server) reproduce identically on a
  pristine main worktree - pre-existing environment failures, not from this
  branch.
- EKS validation (jaseci-cluster, this_is_jac, dev channel, fresh namespaces,
  baseline vs fat): all 5 pods log "vendored wheels found; python deps
  install offline", installs succeed under `PIP_NO_INDEX=1` (any index access
  would hard-fail), gateway jac-bootstrap 126s -> 15s, app healthy end to end
  (register/login/walkers exercised through a port-forward). Bundle grew
  2.3MB -> 22MB for this app's 43-wheel closure; jac-pvc-bootstrap untar
  2-4s -> 8-22s on EFS. A heavy-closure run (jacBuilder, ~1GB of PyPI writes
  per pod today) is in progress; results will be posted on this PR.

Side finding from the baseline runs (will file separately): the shared pip
cache on the bundle PVC has been a no-op all along - pip disables caching
because the fsGroup-owned cache dir is not owned by the root bootstrap user -
so every pod re-downloads from PyPI today. Fat bundles sidestep that entirely
for python deps.


MusabMahmoodh:
Update per review direction: fat bundles are now ON BY DEFAULT (6973fee57). Semantics: unset -> vendor the closure, and if vendoring fails (e.g. an sdist-only dep has no wheel) fall back to the pre-fat thin bundle with a warning so existing deploys never break; explicit `[scale] fat_bundle = true` -> vendoring failure fails the deploy; `fat_bundle = false` -> opt out. Tests updated for the new default (5/5 pass, plus app_precompile/pvc_injector 14/14 under default-on).

Thamirawaran:
merging this now, the thin-bundle fallback means nothing breaks and 126s to 15s on gateway boot is too good to sit on.

@MusabMahmoodh jacBuilder numbers still pending though. it's default-on, so a heavy closure would just get quietly slower instead of failing. can you run it soon? if it looks bad we add a size threshold in the follow-up.

also for that follow-up: description still says `fat_bundle = false, default true` (stale), and `vendor_wheels.jac` lost its docstring to strip-docstrings.

PR #7528, opened 2026-07-16
feat(lang): clean-break language cleanup — lexical rebinding, operator and keyword removals (#7514)

### LINKED ISSUE #7514: Language clean break: glob/global merge, can/def enforcement, operator and keyword cleanup
An audit of the grammar, lexer, and the self-hosted `.jac` sources surfaced a set of language inconsistencies worth fixing in one coordinated break. This issue tracks the agreed scope.

## Breaking language changes

- [ ] **Remove the `global`/`nonlocal` directives; adopt standard lexical rebinding** — `x = 5;` in a function assigns to the nearest existing binding (local scope, then enclosing function scopes, then module `glob`); it creates a new local only when no binding exists. Creating a local that shadows an outer binding requires a typed declaration (`x: int = 5;`), and that declaration must precede any use of the name in the scope. Loop, `except ... as`, and `with ... as` targets always bind fresh locals. The module-level `glob` declarator is unchanged and remains the only globals keyword. Python codegen synthesizes `global`/`nonlocal` statements in emitted code from symbol resolution; the directives, `KW_GLOBAL_REF`, and `KW_NONLOCAL` disappear from the language. This also fixes the inherited Python gotcha where `x += 1` against a global raises `UnboundLocalError`. (Replaces the earlier plan of merging `glob` into a context-dependent `global`.)
- [ ] **Enforce the `can`/`def` split grammatically** — `can` requires an event signature (`with entry` / `with exit`); `def` is the only spelling for function-style declarations. Today both dispatch to the same `parse_ability` and `can foo()` parses with only a warning (E0034). Promote E0034 to a hard error.
- [ ] **Drop `&&` and `||`** — `and` / `or` become the only spellings. The symbol forms currently lex directly to `KW_AND` / `KW_OR`, so the AST cannot even distinguish them; they are pure aliases.
- [ ] **Rename the abstract keyword `abs` → `abst`** — frees `abs` to be the builtin function it already is in practice (the current keyword shadows `abs()` and relies on context-sensitive handling in expression position).
- [ ] **Finish deprecated syntax migrations** — remove the parenthesized filter form `(?:...)` in favor of `[?:...]` (currently W0061), and remove `root()` in favor of bare `root` (currently W0062). Both warnings become errors or are retired along with the old forms.
- [ ] **`has x: T by postinit;` → `has x: T postinit;`** — drop the `by` from deferred-init fields.
- [ ] **C-style for-loop header `to`/`by` → `while`/`with`**:

  ```jac
  # before
  for i = 0 to i < 10 by i += 1 { }
  # after
  for i = 0 while i < 10 with i += 1 { }
  ```

  Neither `while` nor `with` has an expression-position meaning (`as` is the cast operator and `and`/`or` appear inside conditions, which ruled those out), so both clauses are full expressions with no ambiguity; `while` reuses its loop-condition meaning. Together with the `postinit` change this takes `by` from five grammatical roles down to three (ability body binder, impl body binder, infix operator), and `to` stops being a keyword entirely (the for-header was its only parse site).

## Removals (zero breakage)

- [ ] **Delete dead tokens `:g:`, `:nl:`, `:typ:`** (`GLOBAL_OP`, `NONLOCAL_OP`, `TYPE_OP_ALIAS`) — emitted by the lexer, consumed by zero parse rules, absent from the parser token-name map, and written nowhere in the repo.

## Fixes (non-breaking)

- [ ] **Repair the backtick keyword-escape mechanism**, which is currently inconsistent in three ways:
  - a backtick-escaped `has` field passes `jac check` but raises `SyntaxError` at runtime inside the dataclass machinery;
  - the Python import loader auto-escapes keywords with an ad-hoc exclusion list (`['float','int','str','bool','self']` in `pyast_load_pass.impl.jac`) rather than one consistent rule;
  - the W3006 over-escape lint should agree with whatever the unified rule is.

## Explicitly out of scope (for now)

- The broader reserved-word overhaul (soft/contextual keywords, un-keywording type names like `list`/`type`/`dict`).
- Removing `switch` (vs `match`) and `forever`.
- Pipe-operator family consolidation (`|>`/`<|` vs `:>`/`<:`).

## Implementation notes

- Blast radius for each syntax change includes the full bootstrap chain: seed compiler → Layer 1 bootstrap `.jac` files → `jac0core` parser/lexer → all `.jac` sources in the repo (compiler, examples, tests, fixtures), plus the formatter and grammar-extract pass.
- Precedent for the migration mechanics: the lambda-braces clean break (#7376) and the `jac add`→`jac install` tombstoning (#7336).

marsninja:
Implements the full clean-break plan from #7514. **All changes are breaking**; every in-repo usage is migrated in this PR.

## Scoping overhaul: `global`/`nonlocal` removed

Jac now uses standard lexical rebinding instead of Python's directive-based scoping:

- A bare assignment (including `+=` and `:=`) inside a function binds to the nearest enclosing binding — an enclosing function's local, or a module-level `glob`-declared variable. A new local is created only when no such binding exists.
- Only `glob` variables are implicitly rebindable at module scope. Assigning a name that matches a module-level import/function/class still creates a local (Python behavior preserved where it matters).
- Shadowing an outer binding requires a typed declaration (`x: int = 5;`) placed before any use of the name in that scope; violations emit the new **E0064**.
- Loop targets, `except ... as`, and `with ... as` targets always bind fresh locals.
- Python codegen synthesizes `global`/`nonlocal` statements from symbol resolution (per-scope `implicit_rebinds` recorded during symbol-table construction). The na and es backends were already resolution-driven and needed no changes.
- Fixes the inherited Python gotcha where `x += 1` against a module global raised `UnboundLocalError`.

`glob` remains the only globals keyword. `KW_GLOBAL_REF`, `KW_NONLOCAL`, the `GlobalStmt`/`NonLocalStmt` unitree nodes, and all pass plumbing are deleted; the Python loader maps imported `global`/`nonlocal` statements to no-ops.

## Other language changes

| Change | Old | New |
|---|---|---|
| Boolean operators | `&&` / `\|\|` (silent aliases) | `and` / `or` only |
| Abstract marker | `def area -> float abs;` | `def area -> float abst;` (`abs` is only the builtin) |
| Filter comprehension | `(?:Type, cond)` | `[?:Type, cond]` — old form is hard error **E0048** (was W0061) |
| Graph anchor | `root()` | bare `root` — old form is hard error **E0049** (was W0062) |
| Deferred field init | `has x: T by postinit;` | `has x: T postinit;` |
| C-style for | `for i = 0 to i < 10 by i += 1 {}` | `for i = 0 while i < 10 with i += 1 {}` (keyword separators with no expression-position collisions; `to` is no longer a keyword, and the condition is now a full expression) |
| Dead tokens | `:g:`, `:nl:`, `:typ:` lexed but never parsed | deleted |

## Backtick escape fixes

- New **E0067**: a Python reserved word (`class`, `lambda`, ...) cannot be a `has`-field or parameter name even with backtick escaping — previously this passed `jac check` and crashed at runtime inside the dataclass machinery.
- The Python import loader and lint now derive keyword-escaping from a single `RESERVED_IDENT_NAMES` constant (previously one module-level list plus two divergent local recomputes with an ad-hoc exclusion list).

## Migration & fixes found en route

- 144 directive sites deleted across 66 files; every `global` target verified against a real `glob` declaration; shadow-collision sweep added 19 typed declarations where an old local would have silently become a glob rebind; 4 jac0core (seed-compiled) sites rewritten to the holder pattern; seed compiler support for removed syntax stripped.
- Pre-existing bugs fixed: unparse emitted `glob` declarations as `global` (round-trip corruption); impl-body nested rebinds skipped directive synthesis (ImplDef `decl_link` deep-lookup fallback); one fixture carried a historically corrupted `glob al_counter` (bad replace of `global_counter`); two latent write-to-global bugs (jsx-guard suggestion cache, webview handle) now behave as intended under the new semantics.
- Docs: 7 skill/MCP guides updated to the new syntax and diagnostics.

## Testing

- Cold bootstrap from empty cache: seed → jac0core → full compiler → full CLI surface (103 modules) compiles clean.
- End-to-end semantics smoke: glob rebinding, `+=`, typed-decl shadowing, closure (`nonlocal`) synthesis, walrus rebinding, loop-target freshness — all verified in emitted Python and at runtime.
- Targeted suites: `test_parser` 73/73, `test_rd_parser_validation` 194/194, `test_unparse_validate` OK, `test_ts_decl_parser` OK (exercises the 19 migrated `nonlocal` sites); E0048/E0049/E0064/E0067 assertions verified.
- Known env caveats (pre-existing, unrelated): per-file `jac check` has an OwnershipCheckPass scheduling failure that fires on pristine files; subprocess-based tests (`test_console`, `test_cli`) shell out to the prebuilt `jac` binary and need a binary rebuild to exercise this change.

Closes #7514.


marsninja:
CI fixes pushed in 76cb4bb12 (formatter compliance, regenerated jir_registry, release-note fragment rename, em-dash cleanup). One cross-repo dependency: the fullstack smoke job clones jaseci-labs/this_is_jac at main, which still uses the removed `global` directives — companion PR jaseci-labs/this_is_jac#6 migrates it (verified 0 errors/0 warnings with this branch's compiler). That job stays red here until #6 merges.

PR #7533, opened 2026-07-17
fix(check): unbreak main - baseline the impl files the full-repo jac check now walks individually

MusabMahmoodh:
Main's CI has failed on every push since ~22:17 UTC yesterday (runs 29539022951, 29539067424, 29541700035): the jac-check job reports 81 error sites across `type_utils.impl.jac`, `template_registry.impl.jac`, and `lang_tools.impl.jac`.

What happened: #7532 cleared the full-repo check baseline, and its own CI was green - but against a merge ref that predated #7512/#7531 landing the same hour. On main, the checker walks `.impl.jac` files as their own units; `.jacignore` still lists the three BASE modules (`type_utils.jac` etc.) by exact name, which does not cover the impl files, so their pre-existing errors now fail every run.

This adds the three `.impl.jac` names to `.jacignore`, next to their base entries, until the 81 sites are actually fixed (they are real errors worth a follow-up, mostly enum/int returns and Optional narrowing in the language tooling).

Not touched here:
- `test-runtime` on main also failed, but that was setup-bun failing to reach GitHub's API (transient; unrelated).
- `Release jac dev binary` fails in `prepare` with HTTP 422 updating the `dev` prerelease - repo release state, needs a maintainer look if the next main push does not clear it. Until one of these passes, no new `:dev` binaries/images publish.

### DIFF
--- a/.jacignore
+++ b/.jacignore
@@ -34,6 +34,7 @@ jac_formatter_pass.jac
 jfast_api.jac
 kubernetes_utils.impl.jac
 kubernetes_utils.jac
+lang_tools.impl.jac
 lang_tools.jac
 llm.jac
 macho_linker.impl.jac
@@ -75,6 +76,7 @@ server.jac
 sym_tab_build_pass.jac
 template_loader.impl.jac
 template_loader.jac
+template_registry.impl.jac
 template_registry.jac
 testing.impl.jac
 testing.jac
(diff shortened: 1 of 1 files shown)


MusabMahmoodh:
For visibility: main still fails after #7534 - run 29545666026 (commit 40c0e2d) fails jac-check (same 81 sites in the three .impl.jac files this PR baselined) and test-pypi-build (the packaged template's main.jac hits E1055 overload-resolution errors on __mul__/__add__/__mod__, which looks like #7512's type-system changes racing the template code). Open PRs are inheriting both through their merge refs. Happy to reopen this for the jac-check half if useful; the pypi-build half needs a template or type-system fix.

Issue #7486, opened 2026-07-15
main red: jac check reports ~34 false-positive type errors — diagnostic-timing regression from #7418 (and #7392)


kugesan1105:
## Summary

`main` CI has been red since **2026-07-14** because the `jac-check` job fails: `jac check .` over the whole repo reports **~33–34 files failing** with type errors. These are **false positives** — the files themselves are unchanged and were type-check-clean immediately before the regression.

This is a **diagnostic-timing regression in the type checker** introduced by #7418, with a second contributing regression from #7392.

## Evidence it's a regression, not real errors

Same command (`jac check . --ignore … --nowarn`), same files, before vs after:

| Commit | Result |
|---|---|
| `629e3014d` (last green, 2026-07-14 01:02, full-repo `push` mode) | `504 passed in 75.06s` |
| current `main` | `471 passed, 33 failed in 158.68s` |

The failing files did not change between these commits — only the compiler did.

## Why PR CI didn't catch it

`.github/workflows/ci.yml` runs `jac-check` in two modes:
- **`pull_request`** → *scoped*: only checks the `.jac` files the PR changed.
- **`push` to `main`** → *full-repo*: `jac check .`.

The culprit PRs changed the **checker itself** but touched none of the 34 failing `.jac` files, so their scoped PR runs were green — and the breakage only surfaced on the full-repo `push` run after merge. **A checker/compiler change has a repo-wide blast radius that scoped PR checking cannot see.**

## Attribution (bisected)

| Root cause | Files | Notes |
|---|---|---|
| **#7418** (`a1644756c`, single-source analyses) | ~27 | Dominant cause. Pre-#7418 all 27 pass; at #7418 they fail. |
| **#7392** (`d6f22ebef`, release per-file compile closure) | 6 | `JAC_NO_UNIT_EVICT=1` fixes these 6; it wipes compile state mid-run and re-exposes the #7418 bug. Also made `jac check` *slower* (75s→158s). |

Confirmed by: git-worktree bisect, swapping the pre-#7418 `type_checker_pass.impl.jac` into the tree (false positives vanish), and the `JAC_NO_UNIT_EVICT` toggle.

## Minimal reproduction

```jac
def add(a: int, b: int) {
    return a + b;
}
```

```
$ jac check add.jac
✖ Error: error[E1003]: Return type annotation required when function returns a value
```

`a + b` is trivially inferable, so no annotation is required. Passes on the pre-#7418 compiler; fails on current `main`.

[...]

PR #7538, opened 2026-07-17
fix(typecheck): make type_utils + version-condition helpers type-clean (greens main jac-check)

kugesan1105:
## Summary

`type_utils.impl.jac` was the last compiler file failing the full-repo `jac check .` on `main` (the `jac-check` CI job). This makes it type-clean, so `jac check .` passes **483 passed / 0 failed**.

## Changes

- **`expr_primitive_name`**: annotate `prog` as `JacProgram` (via a type-only import to avoid the runtime circular) instead of `object`, so `prog.get_type_evaluator()` resolves; guard `expr_node` to `uni.Expr` before `get_type_of_expression`.
- **version-condition helpers** (`helpers.jac`): the params used a forward-ref `'uni.Expr'` (because `uni` is imported only inside the bodies), which the checker read as `Literal["uni.Expr"]` and rejected at the call site. Add a module-level **type-only** import and type them as `UniNode` (they unwrap `AtomUnit.value`, which is broader than `Expr`).
- **`doc_of_sym`**: `lit_value` is `str | bytes`; return it only when it is `str`.
- **`get_completion_items`**: guard `None` scope / module `symbol_table` before `.names_in_scope`.
- **`completion_kind_from_sym`**: return the enum member's int `.value` (declared `-> int`).
- **type-arg specialization**: guard `None` `type_params` / `type_args` before `enumerate` / `len`.

## Validation

`jac check .` (same invocation as CI) → **483 passed, 0 failed**. `jac fmt --check` clean on the touched files.

Context: part of the `jac-check` regression tracked in #7486.


marsninja:

PR #7545, opened 2026-07-17
fix(jac): route jac lint through the report-only pass and make it preserve the mutation pass's warnings

chess10kp:
## Summary

`get_lint_sched()` (the report path behind `jac lint` and `JacProgram.jac_file_linter`) returned `[JacAutoLintPass]` with `report_only=False`, so it ran the full AST-rewriting auto-lint pass and then discarded the rewritten module. Only warnings/errors are read, so the mutation work was pure waste. This switches the report path to `[JacLintCheckPass]` (`report_only=True`), matching what `jac check` already uses.

But `report_only` mode was itself under-reporting: nearly every mutating handler early-returned *before* emitting its diagnostic, so the report path silently dropped all autofix-class warnings. The switch alone would have made `jac lint` under-report on any file that actually contains those issues (the tracked tree only looked "identical" because it is already deslopped and triggers none of them). This PR fixes `report_only` so the report path preserves the mutation pass's warnings.

## Changes

- **`jac0core/compiler.jac`** — `get_lint_sched()` returns `JacLintCheckPass` instead of `JacAutoLintPass`.
- **`passes/tool/impl/jac_auto_lint_pass.impl.jac`** — split each mutating handler so the diagnostic is emitted in *both* modes and only the AST rewrite is gated on `not report_only`. This covers empty parens, combine-`has`/`glob`, `staticmethod`->`static`, `__init__`->`init`, keyword-escape, redundant `from __future__ import annotations`, import semicolons, ternary/getattr/hasattr rewrites, `useState`->`has`, map->comprehension, and property conversions. Also runs `combine_consecutive_glob` detection in report mode (its rewrite was already guarded but the call was unreachable) and normalizes the `__init__` name so follow-on messages match the renamed tree.
- **`tests/compiler/passes/tool/test_jac_lint_report.jac`** — regression tests: the scheduler is locked to `JacLintCheckPass`, the report path matches the full mutation pass exactly across the `auto_lint` fixtures, and a strict *subset* invariant asserts the report path never emits a warning the full pass would not (no spurious/wrong warnings on any fixture).

## Behavior notes

- **Faster:** ~6x for the lint pass, ~19% for `jac lint` end-to-end.
- **`jac check` also benefits:** it shares this report path and was under-reporting the same warnings; it now surfaces them too. CI runs `jac check --nowarn` and `jac fmt --lintfix` (mutation path, unchanged), so this does not gate the build.
- **One intentional divergence:** *follow-on* autofixes the mutation pass can only see after rewriting the tree — combining `glob`s it just extracted from a `with entry`, simplifying a ternary once its `hasattr` condition is rewritten, combining `has` decls synthesized from `useState`. The report path emits the *primary* warning; the follow-on surfaces on the next lint after the primary fix is applied (normal iterative linting). These four fixtures are exempt from exact parity but still held to the subset invariant.

## Checklist
- [x] Code change is minimal and idiomatic
- [x] Regression test added
- [x] Release note added


marsninja:
The fail on compiler doesnt look like a flake.

chess10kp:
Closing as superseded: this fix landed on main via #7496 (ebe266823, squash merge) and was carried through the passes/tool → tools/ rename in #8682. Verified the scheduler routing, split report-only handlers, and both key tests exist on upstream/main today. Two extra tests from this branch were never ported — if they're still wanted, a fresh test-only PR against current main would be the path.

PR #7551, opened 2026-07-18
fix(launcher): preserve executable bit through payload pack/extract

chess10kp:
## Summary

Bundled .sh scripts (e.g. build_libwebview.sh) lost their exec bit through the launcher's payload pipeline, surfacing as EACCES / PermissionError when jac start spawned them on the desktop target.

## Root Cause

Two independent, compounding defects — either alone causes the symptom:

**Packer** (payload.zig tarZstDir): every regular file archived via writeFileBytes(path, bytes, .{}). The empty options {} → mode: 0, and std/tar/Writer.zig's writeHeader only calls setMode when mode != 0 — the header keeps the default 0o664 with no exec bit. The on-disk mode was never consulted.

**Extractor** (runtime.zig extractPayload): even if the tar carried an exec bit, .mode_mode = .ignore makes std/tar.zig:1164 return .default_file regardless of the archived mode — stripping the exec bit on materialization.

Both were confirmed against the real production functions.

## Fix

- Packer (payload.zig): new sourceFileMode() helper stats each file's on-disk mode and passes it to writeFileBytes(.{ .mode = ... }). Cross-platform-safe via @hasDecl(Perm, toMode) guard.
- Extractor (runtime.zig): .mode_mode = .ignore → .executable_bit_only, mirroring the already-correct fetch-pbs extract path.
- Regression test: extended the existing round-trip test (which only asserted bytes) to stage a 0o755 script and assert IXUSR survives the full pack→extract pipe with .executable_bit_only.

Both fixes required together. The test proves the combined fix end to end.

## Verification

zig build test --summary all:
  runtime-tests 11 pass (11 total)
  payload-tests 14 pass (14 total)
  25/25 passed, 5/5 steps succeeded

## Implementation Notes

- The fetch-pbs extract path already had a correct .executable_bit_only with a comment pointing to the EACCES footgun (line 255-258) — the runtime extract simply missed that lesson.
- The existing test assertions on bytes pass unchanged; the new exec-bit assertion would fail if either fix is reverted.


chess10kp:
Verified and cleaned up. `git diff --stat` shows `payload.zig` is back to original — no instrumentation residue.

marsninja:
Feels like this should be tested, wouldn't want to do a very heavy one where there is a whole build on ci to check it, but perhaps can piggy back on an existing test to just add a check for the exec bit.

PR #7553, opened 2026-07-18
feat(py2jac): roundtrip class keyword args (metaclass=)

chess10kp:
## What

Preserve class-definition keyword arguments (e.g. `metaclass=Meta`) across the Python → Jac → Python pipeline. Previously the class keyword was dropped, so a `class Foo(metaclass=Meta)` lifted from Python lost its metaclass and CPython no longer ran it.

## Changes

- Add a `keywords: Sequence[KWPair] | None` field to the `Archetype` unitree node.
- Parse call args in archetype base lists, splitting `KWPair` keywords from positional base classes (parser).
- Carry keywords through pyast load (Python AST → Jac) and pyast gen (Jac → Python AST).
- Emit keywords in the normalize / unparse tool passes.
- Add roundtrip fixtures and tests:
  - `metaclass_py_to_jac_roundtrip` — a `class Foo(metaclass=Meta)` survives Python AST → Jac IR → Python AST and CPython runs the metaclass natively.
  - `py2jac metaclass` — `class Foo(Base, metaclass=Meta)` survives py2jac alongside its base class.

## Testing

- New tests in `test_pyast_gen_pass.jac` and `test_py_infra.jac` pass.


marsninja:
This is a clean pr. My only hesitation is on reasoning what we do on the native pathway for metaclass and do we just treat it as out of the congruence set like we do monkeypatching? I'm also wondering what was the use case that motivated this pr out of curiositiy.

PR #7559, opened 2026-07-18
refactor(lint): dedup identical null-ok propagation in exit_if_else_expr/exit_bool_expr

chess10kp:
## What

`exit_if_else_expr` and `exit_bool_expr` contained a byte-for-byte identical block: collect atom trailers, gather the keys of the null-ok ones, then mark every matching non-null-ok sibling trailer as null-ok. This extracts that block into a single `_propagate_null_ok_siblings(root)` helper.

## Why

The two handlers differed only in their guard preamble:
- `exit_if_else_expr` first runs the ternary-to-`or` conversion (`W3008`) and then the `W3007` suppression check;
- `exit_bool_expr` only has the `W3007` guard.

Everything after the guards was duplicated (~24 lines each). Sharing it removes the duplication and keeps the two entry points as thin guard + delegate.

## Behavior

No functional change. The guard order in each handler is preserved exactly; only the shared tail is factored out.


jac/tests/compiler/backends/native/fixtures/clib_logical_import.jac
```diff
@@ -7,11 +7,10 @@ extensionless surface instead of an explicit string path.
 """
 
 import from raylib {
-    def InitWindow(width: i32, height: i32, title: str) -> None;
-    def CloseWindow -> None;
+    def InitWindow(width: i32, height: i32, title: str) -> None;  def CloseWindow -> None;
 }
 
-def open_then_close(w: int, h: int, title: str) -> None {
+def open_then_close(w: int, h: int, title: str) {
```

christianwilkins (changes requested):
Reviewed `1804b3d4d86bbc7334983ae12aea336e798b504e`, including the complete diff, surrounding code, existing conversation/reviews, and checks for this exact head. The extracted helper retains the two-pass key collection/propagation logic and both callers’ existing guards, including W3008 conversion before W3007 suppression. Its declaration matches the implementation. Local validation was static inspection and a passing `git diff --check`; I did not independently rerun the lint suite.

P2: the unrelated `clib_logical_import.jac` edit removes `-> None` from `open_then_close`. The existing logical-import test requires a `define void` wrapper and fails on this head with “wrapper not defined” (native/tools lane: 1,516 passed, 1 failed, 4 skipped). Please restore the annotation; changing the fixture’s return contract is outside this refactor.

Contribution Checks also fails because `.refactor.md` is a retired release-note category. For this internal refactor, remove the fragment and have a maintainer apply `skip-release-notes-check` if needed, as the repository’s check instructs. GitHub reports no merge conflicts. Changes requested: restore the fixture signature, correct the release-note handling, and obtain passing checks on the resulting head.

chess10kp:

PR #7561, opened 2026-07-18
perf(lint): replace O(RxW) stale-has-read scan with an earliest-write map

chess10kp:
## What

`_check_stale_has_read` (the `W3041` reactive-`has` staleness check) flagged a read when any write to the same name occurred at an earlier source position. It did this with a nested loop over every (read, write) pair -- O(R x W).

Since the check only needs "does an earlier write to this name exist", it is equivalent to precomputing the earliest write position per name and comparing once per read. This adds an `earliest_write: dict[str, (int, int)]` map (one pass over writes) and reduces the read loop to a single dict lookup -- O(R + W).

## Behavior

Identical results. "There exists a write before the read" is exactly "the minimum write position for that name is before the read". Emit order (reads in list order), the per-read dedup via `flagged`, and the `node_override`/`name` arguments are all unchanged.


christianwilkins (commented):
Reviewed `26166d9198f9556f7d97424e2e3ed994012edd97`, including the complete two-file diff, read/write collection, suppression/context gates, and existing discussion. The earliest-write comparison preserves the original strict source-position ordering, read emission order, and duplicate suppression. I found no correctness issue in that optimization. Local validation was static inspection and a passing `git diff --check`; I did not independently rerun the lint suite.

P3: please move the fragment to `release_notes/unreleased/jaclang/7561.refactor.md`. At this head, `scripts/collect_release_notes.jac` reads that directory; the added `docs/docs/community/release_notes/unreleased/...` path is outside its collection root, so the note will be omitted from release assembly.

Current-head CI is not green: `jac-check` reports 915 files passed and 1 failed, with E1001 diagnostics for `sock.recv(8192)` assigned to bytes in `runtime/na_stdlib/urllib/request.jac` at lines 77 and 80. I have not established a clean-base reproduction, so I am not labeling that failure pre-existing. GitHub reports no conflicts. Not yet ready: correct the fragment location and resolve or establish the baseline for the checker failure, then obtain passing checks on the resulting head.

christianwilkins:
Correction to the release-note recommendation in my review: I checked the collector path but missed the category policy in `scripts/check-release-notes.sh`. It explicitly retires `refactor` fragments. For this internal optimization, please remove the obsolete fragment rather than moving it; the PR already has `skip-release-notes-check`, which is the intended path. The earlier code assessment and current-head checker-failure report still stand.

chess10kp:
CI red note: `test_dangling_ref_heal` failed once with `SQLSTATE 08006 connection lost mid-transaction` (1960/1961 runtime tests passed) — a parallel-DB-load flake unrelated to this branch (it edits only `jac_auto_lint_pass.impl.jac`). The earlier smoke-journey flake did not reproduce on retrigger. Lint-pass tests are green.

PR #7564, opened 2026-07-18
refactor(lint): dedup combine_consecutive_glob/has kid-rebuild via _merge_run_into_base

chess10kp:
## What

`combine_consecutive_glob` and `combine_consecutive_has` shared a large, near-identical body: scan a statement list for a run of consecutive combinable nodes, then rewrite the base node's `kid` list to fold every run member's items into one comma-separated declaration terminated by a single `;`.

This extracts the kid-rebuild half into one `_merge_run_into_base(base, combined_items, items_attr, strip_semi)` helper, parameterized by the only two things that actually differed:

- `items_attr`: `"assignments"` for globals, `"vars"` for `has` (read/written via `getattr`/`setattr`, a pattern already used elsewhere in this pass);
- `strip_semi`: globals strip a trailing `;` off each merged assignment's kids; `has` vars do not.

The two callers keep their own scan loop and their own gating, which genuinely differ and are deliberately left in place:

- globals require every merged declaration to be single-assignment (`all_single`) and emit `W3003` inside the combine step;
- `has` emits `W3002` at its callers (`_combine_has_in_node` / `enter_impl_def`), not here.

## Behavior

No functional change. The extracted helper reproduces each original rewrite exactly (item order, comma insertion, per-item semi stripping for globals only, `items_attr` assignment before `set_kids`). Full `test_jac_auto_lint_pass` suite passes (68/68), including `TestCombineConsecutiveHas` and the glob-combination assertions.


jac/tests/compiler/backends/native/fixtures/clib_logical_import.jac
```diff
@@ -7,11 +7,10 @@ extensionless surface instead of an explicit string path.
 """
 
 import from raylib {
-    def InitWindow(width: i32, height: i32, title: str) -> None;
-    def CloseWindow -> None;
+    def InitWindow(width: i32, height: i32, title: str) -> None;  def CloseWindow -> None;
 }
 
-def open_then_close(w: int, h: int, title: str) -> None {
+def open_then_close(w: int, h: int, title: str) {
```
christianwilkins, line 13: [P2] Restore the fixture’s explicit void return. This unrelated deletion changes the generated wrapper contract; the existing native logical-import test asserts `define void`, and current-head CI fails with “wrapper not defined”. Keep `-> None` here.

christianwilkins (changes requested):
Reviewed `cd998a0629d40a47950f9f99fa593bec769d7018`: full diff, both combine callers and their gates, role/kid update ordering, existing conversation, and current-head CI. The helper retains prefix kids, item order, comma insertion, glob-only semicolon stripping, role assignment, and final set_kids ordering. The current role argument matches AssignmentsRole/VarsRole at the two call sites. Local validation was static inspection and a passing `git diff --check`; I did not rerun the lint suite.

P2: the unrelated fixture edit removes `-> None` from `open_then_close`, changing its wrapper contract. Current-head native/tools CI fails the existing logical-import assertion with “wrapper not defined” (1,516 passed, 1 failed, 4 skipped). Restore that annotation.

Contribution Checks rejects the `.refactor.md` fragment because that category is retired. For this internal refactor, remove the fragment and have a maintainer apply `skip-release-notes-check` if needed, as the check directs. The separate jac-check failure is in lsp/server/engine.jac (915 passed, 1 failed); I have not established it as introduced by the refactor. GitHub reports no merge conflicts. No code changes were made. Changes requested: restore the fixture, correct release-note handling, and obtain passing checks on the resulting head.

chess10kp:

PR #7546, opened 2026-07-17
fix(jac2js): prepend @jac/runtime imports on all JS emission paths

chess10kp:
Fixes #7540

## Summary

Codegen now injects required `@jac/runtime` imports (`__jacJsx`, `__jacSpawn`) through the established ESTree import mechanism in `EsastGenPass`, so `mod.gen.js` is emission-ready without consumer-side regex finalizers.

- **Compiler-owned imports**: `ReactBackend.view_runtime_symbols()` returns the JSX factory symbol so ordinary React/Preact JSX elements get imports at codegen time. Remote spawn injects `__jacSpawn` the same way.
- **Backend-neutral fragments**: Compiler-created fragments lower through `ViewElement`/`FragmentTag` IR so Solid keeps native `<>...</>` fragments instead of importing `__jacJsx`.
- **Removed regex finalizer**: Deleted `js_runtime_globals.jac` and removed `finalize_js_for_emission` / `prepend_active_runtime_globals` from CLI, MCP, bundle builder, HMR, eject, test runner, and publisher paths.
- **npm publishing**: `_wire_externals` remaps `@jac/runtime` to the configured runtime package (`@jaseci/runtime`) instead of prepending a duplicate import.
- **Runtime tarball**: Building `@jac/runtime` packages no longer prepends consumer runtime imports onto the runtime shim itself.

## Root cause

JSX lowering emits calls to `__jacJsx`, but several emission paths returned raw `mod.gen.js` without the ESM import from `@jac/runtime`. The initial fix prepended imports with a regex-based finalizer at every consumer, which duplicated responsibility, missed some compiler paths, and could double-import on npm publish.

## Design

```
Jac source
    ↓
EsastGenPass (backend lowers runtime ops; _inject_runtime_import owns ESM)
    ↓
emission-ready mod.gen.js
    ↓
Consumers (CLI / MCP / Vite / HMR / publisher) consume output without finalizing
```

## Test plan

- [x] `jac test tests/runtimelib/client/test_js_runtime_globals_emission.jac` — focused behavior tests:
  - Raw React/Preact JSX imports `__jacJsx` from `@jac/runtime`
  - Raw Solid JSX does not reference/import `__jacJsx`
  - Solid compiler-created fragments stay backend-native (`<>...</>`)
  - Remote spawn imports `__jacSpawn`
  - Runtime source does not self-import `@jac/runtime`
  - Published npm JS remaps the runtime import exactly once
- [ ] CI full suite
[...]

PR #7567, opened 2026-07-18
fix(cli): jac2js CLI and MCP bridge must prepend @jac/runtime globals

### LINKED ISSUE #7541: jac tool jac2js omits @jac/runtime globals (__jacJsx ReferenceError in ESM)
## Summary

`jac tool jac2js` emits raw `ir.gen.js` without calling `prepend_active_runtime_globals()`, so JSX client modules reference `__jacJsx` with no import. Loading the output as ESM throws `ReferenceError: __jacJsx is not defined`.

Confirmed still reproducing on jac **0.32.1**.

## Root cause

In `jac/jaclang/cli/commands/impl/transform.impl.jac` (lines 66–71), the `jac2js` implementation prints codegen output directly:

```jac
js_output = ir.gen.js or '';
...
console.print(js_output, markup=False, highlight=False);   // raw print, no wrapper
```

No call to `prepend_active_runtime_globals()` anywhere in this path.

The injection helper exists and works at `jac/jaclang/runtimelib/client_surface.jac:82` (`prepend_active_runtime_globals` → `with_runtime_globals`).

## Other consumers (all correct)

| Caller | File |
|---|---|
| `jac_to_js.impl.jac` | `return prepend_active_runtime_globals(js_code)` |
| `compiler.impl.jac` | `core_js = prepend_active_runtime_globals(core_js)` |
| `eject.impl.jac` | `prepend_active_runtime_globals(js_code)` |
| `cl_test_runner.impl.jac` | `module_js = prepend_active_runtime_globals(module_js)` |
| `hmr.impl.jac` | `js_code = prepend_active_runtime_globals(js_code)` |
| `jac_client_compiler.impl.jac` | `add_runtime_imports(module_js)` |

The standalone `jac tool jac2js` CLI path is the **only** transform consumer that skips it.

## Also affected

- MCP `CompilerBridge.jac_to_js` (`jac/jaclang/cli/mcp/impl/compiler_bridge.impl.jac:427`) returns raw `ir.gen.js` without prepend.

## Repro

Compile any `.cl.jac` with JSX via `jac tool jac2js`:

```bash
jac tool jac2js app.cl.jac
```

Output contains `__jacJsx(...)` but **no** `import { __jacJsx } from "@jac/runtime";`.

Full build (`jac start`) works because `compiler.impl.jac` injects globals — only standalone emission paths are broken.

## Suggested fix (short term)

In `transform.impl.jac`:

```jac
import from jaclang.runtimelib.client_surface { prepend_active_runtime_globals }
...
console.print(prepend_active_runtime_globals(js_output), markup=False, highlight=False);
```

Same for MCP `compiler_bridge.impl.jac`.

## Suggested fix (long term / root cause)

**Centralize finalization at codegen** so `ir.gen.js` is always emission-ready:

```jac
// esast_gen_pass.impl.jac, when assigning nd.gen.js:
nd.gen.js = prepend_active_runtime_globals(es_to_js(nd.gen.es_ast));
```

`with_runtime_globals` is already idempotent (skips when imports present or symbols defined inline), so existing consumer-level calls become harmless redundancy rather than required correctness.

This closes the entire class of bugs: any new emission surface that reads `mod.gen.js` / `ir.gen.js` automatically gets required `@jac/runtime` globals.

## Bonus finding

The old top-level `jac jac2js` command was removed in #7255 — only `jac tool jac2js` remains. Dead-code fallback in `scripts/jac-client-compile.mjs` (`catch → execSync('jac jac2js …')`) is moot.

## Environment

- jac 0.32.1
- Repro: any client `.jac` / `.cl.jac` using JSX

chess10kp:
Closes #7541

## Problem

`jac tool jac2js` printed raw `ir.gen.js` without calling `prepend_active_runtime_globals()`, so JSX client modules referenced `__jacJsx` (and other `@jac/runtime` globals) with no matching import. Loading the emitted output as ESM threw:

```
ReferenceError: __jacJsx is not defined
```

The MCP `CompilerBridge.jac_to_js` handler returned raw `ir.gen.js` with the same gap.

Every other emission consumer (`jac_to_js`, `compiler`, `eject`, `cl_test_runner`, `hmr`, `jac_client_compiler`) already injects the runtime globals — these two standalone paths were the only ones that skipped it.

## Fix

Wrap the emitted JS in `prepend_active_runtime_globals()` in both paths:

- `jac/jaclang/cli/commands/impl/transform.impl.jac` (`jac2js`)
- `jac/jaclang/cli/mcp/impl/compiler_bridge.impl.jac` (`CompilerBridge.jac_to_js`)

`prepend_active_runtime_globals()` is idempotent (it skips when the imports are already present or the symbols are defined inline), so wrapping the output is safe.

## Verification

Before:

```js
function App() {
  return __jacJsx("div", {"className": "hello"}, [__jacJsx("span", {}, ["Hi"])]);
}
```

Loaded as ESM under Node -> `ReferenceError: __jacJsx is not defined`.

After:

```js
import {__jacJsx, __jacSpawn} from "@jac/runtime";
function App() {
  return __jacJsx("div", {"className": "hello"}, [__jacJsx("span", {}, ["Hi"])]);
}
```

Loaded as ESM under Node (v26) with a `@jac/runtime` shim -> runs cleanly, no ReferenceError.

## Tests

Added regression test `jac2js cli prepends @jac/runtime globals for jsx` in `tests/compiler/passes/ecmascript/test_js_generation.jac`, which runs `jac tool jac2js` on the JSX fixture via subprocess and asserts the emitted output contains both the `__jacJsx(...)` calls and the `import { ... __jacJsx ... } from "@jac/runtime";` line.

Full `test_js_generation.jac` module: **88 passed**.


chess10kp:
Closing in favor of #7546, which moves runtime-import injection into a single choke point (`EsastGenPass._inject_runtime_import`) covering every JS emission path. #7567's consumer-side prepending in `transform.impl.jac`/`compiler_bridge.impl.jac` wraps exactly the paths #7546 removes — the two approaches are mutually exclusive and #7546 is the long-term-correct architecture.

PR #7573, opened 2026-07-18
fix(typecheck): resolve JSX tags by scope, not name case (W1050)

### LINKED ISSUE #6825: Lowercase JSX component names emit false-positive W1050 (component-vs-intrinsic decided by name case, not scope)
## Summary

A JSX tag whose name resolves to a real component (locally defined or imported) is misclassified as an unknown HTML intrinsic when its first letter is **lowercase**, emitting a false-positive `W1050: Unknown intrinsic JSX element '<...>'`.

The JSX pipeline decides "component vs. HTML intrinsic" by the **first letter of the tag name** (uppercase = component, lowercase = intrinsic HTML), rather than by whether the name resolves to a symbol in scope. A lowercase identifier that *does* resolve to a real component function is wrongly treated as an HTML host tag, looked up against the DOM intrinsic registry, and warned as unknown when it isn't there.

This bakes React's capitalization convention into the compiler as the only way to spell a component, and produces spurious diagnostics for perfectly valid code.

## Reproduce

`jac.toml`:

```toml
[project]
name = "repro"
kind = "web"
```

`app.jac`:

```jac
def myWidget(label: str) -> JsxElement {
    return <div>{label}</div>;
}

def App -> JsxElement {
    return <myWidget label="hi"/>;
}
```

```
$ jac check app.jac
⚠ warning[W1050]: Unknown intrinsic JSX element '<myWidget>'
  --> app.jac:6:15
    6 |     return <myWidget label="hi"/>;
      |             ^^^^^^^^
```

The same false positive appears when the lowercase component is **imported** rather than defined locally:

```jac
import from widgets { myWidget }

def App -> JsxElement {
    return <myWidget label="hi"/>;
}
```

## Expected

No warning. `<myWidget/>` resolves to the `myWidget` definition (local or imported) and is a valid component reference.

## Actual

`W1050: Unknown intrinsic JSX element '<myWidget>'`.

## Why this matters

- The case heuristic is the only thing standing between a valid component and a false diagnostic. Renaming `myWidget` → `MyWidget` makes the warning disappear, which confirms the warning is driven purely by case, not by whether the reference is valid.
- The same case-based classification is the basis for `E1105` in `kind = "universal"` projects (raw HTML host tags forbidden). There, the heuristic both **over-rejects** (a legitimate `<myWidget/>` is flagged) and is the only thing preventing **under-rejection** of an out-of-scope uppercase tag. Case is a poor proxy for "is this a real reference" in both directions.
- Correctly fixing this means resolving JSX tag names through the symbol table (is the name in scope?) instead of branching on the first character. That same resolution should feed both the intrinsic-vs-component decision and any project-kind-specific enforcement, so the classification happens once, in one place.

## Related

- #6262 — type-checker false positives on client (`.cl.jac`) code; overlaps in that the client JSX path is the most affected.
- #6417 — roadmap to clarify compiler analysis boundaries; relevant because "is this tag a component?" is a semantic fact that should be owned by a shared frontend pass, not derived independently by the type evaluator and the universal-project guard.

chess10kp:
## Summary

A lowercase JSX tag that resolves to a real in-scope component (locally defined or imported) was misclassified as an unknown HTML intrinsic and emitted a false-positive `W1050: Unknown intrinsic JSX element '<...>'` in web/universal projects. The component-vs-intrinsic decision branched on the **first letter's case** (uppercase = component, lowercase = intrinsic) rather than on whether the name resolves to a symbol in scope.

This baked React's capitalization convention into the compiler as the only valid way to spell a component. A prior fix already addressed this for mobUI projects; this extends the same scope-based classification to the default (web) path.

## Fix

In the type evaluator's JSX handling:
- `_resolve_intrinsic_jsx_tag` now short-circuits when the tag name resolves to a `FunctionType` (a real component): it resolves the symbol and skips `W1050`. Genuinely unknown host tags still emit `W1050`.
- `_resolve_jsx_component_type` no longer gates resolution on the first letter's case; it classifies purely by symbol resolution.
- `_validate_jsx_props` routes lowercase, in-scope components through component prop validation instead of intrinsic validation.

## Tests

Added `web_project_lowercase_component_no_w1050` plus a fixture asserting:
- an in-scope lowercase component (`<myWidget/>`) emits no `W1050`, and
- a genuinely unknown lowercase host tag (`<blepblorp/>`) still emits exactly one `W1050`.

All 18 JSX-related checker tests pass.

Closes #6825


jac/jaclang/compiler/types/type_evaluator.impl/jsx_type_check.impl.jac
```diff
@@ -69,6 +65,14 @@ impl TypeEvaluator._get_intrinsic_props_type(tag_name: str) -> ClassType | None
 }
 
 impl TypeEvaluator._resolve_intrinsic_jsx_tag(part: Name, tag_name: str) -> None {
+    if part.sym is not None {
+        sym_type = self.get_type_of_symbol(part.sym);
+        if isinstance(sym_type, FunctionType) {
```
christianwilkins, line 70: [P2] Update JSX emission together with this new component classification. Compiling the added fixture on this head produces `__jacJsx("myWidget", {"label": "hi"}, [])` even though `myWidget` is a bound function: EsJsxProcessor.element still chooses HostTag by lowercase spelling and element_name still emits a literal. This return removes W1050 without making the use invoke the component. Resolve the tag consistently in codegen and add generated-output or runtime coverage alongside the warning assertion.

jac/tests/compiler/backends/native/fixtures/clib_logical_import.jac
```diff
@@ -7,11 +7,10 @@ extensionless surface instead of an explicit string path.
 """
 
 import from raylib {
-    def InitWindow(width: i32, height: i32, title: str) -> None;
-    def CloseWindow -> None;
+    def InitWindow(width: i32, height: i32, title: str) -> None;  def CloseWindow -> None;
 }
 
-def open_then_close(w: int, h: int, title: str) -> None {
+def open_then_close(w: int, h: int, title: str) {
```
christianwilkins, line 13: [P2] Keep the unrelated fixture’s explicit void return. Removing `-> None` changes its wrapper contract; the existing native logical-import test requires `define void`, and this head’s CI fails with “wrapper not defined”. Restore the annotation.

christianwilkins (changes requested):
Reviewed `a544a02ebf9363360947d10a072dbe02c9f10a7d`: full diff, JSX type resolution and prop checks, JavaScript JSX processing, existing conversation/threads, and current-head CI. I compiled the new web fixture with the production JacProgram and EsastGenPass on this head. Compilation reports no errors and only the unknown `blepblorp` receives W1050, but the generated App body is `return __jacJsx("myWidget", {"label": "hi"}, []);`. The component is emitted separately as `MyWidget` with `const myWidget = MyWidget`.

P2: accepting a lowercase in-scope function in the checker is incomplete while the emitter still produces a host-tag string. EsJsxProcessor.element and element_name still classify single lowercase names by case, so the PR suppresses the warning without making this use invoke the component. Update code generation to use the resolved component reference and add an emitted-JS or runtime regression; retain the unknown intrinsic control.

P2: the unrelated removal of `-> None` from `open_then_close` breaks the existing logical-import fixture. Current-head native/tools CI reports “wrapper not defined” (1,516 passed, 1 failed, 4 skipped). Restore the annotation. Contribution Checks also rejects the release note's plain paragraph; format it as the required bullet. The separate jac-check failure is in lsp/server/engine.jac (915 passed, 1 failed), and I have not attributed it to this patch.

Local `git diff --check` passed. The production compilation/emission probe above completed; I did not run a browser or full suites. GitHub reports no conflicts. No code changes were made. Changes requested: align the emitter with the checker, restore the fixture signature, correct the release-note format, and obtain passing checks on the resulting head.

Issue #7418, opened 2026-07-13
refactor(compiler): single-source analyses; native backend consumes core facts instead of shadow analyses


marsninja:
# Summary

Re-architect the compiler so that every semantic analysis has exactly one implementation in core, backends declare which analyses they require and consume the stamped facts, and the required analyses (type inference, CFG, ownership dataflow) are always on for the native pathway.

This supersedes the "diagnostics-only contract" in `docs/docs/reference/language/ownership-checker-spec.md` and deletes the native backend's shadow analyses. Two audit sweeps produced the inventory below: Part A covers the core-analysis shadows (types, ownership, liveness, CFG, def-use), Part B covers eight further duplicated or misplaced fact families (layout, capability checks, interop types, codespace classification, call dispatch, method-override binding, enums, OSP invocation).

The end state is a well-organized compiler: analyses compute facts once, diagnostics and codegen are both thin consumers of those facts, and pass scheduling is dependency-driven instead of hand-maintained conditional lists.

# Motivation

Today the ownership checker is a lint by design: `OwnershipCheckPass` writes `Symbol.ownership` and its dataflow sets, but no backend is allowed to read them, because the checker only runs when `type_check=True` and `nacompile`/JIT compile with `type_check=False`. Codegen depending on a maybe-ran analysis would mean a lint toggle changes program behavior, so the contract forbids it.

The consequence is systematic duplication: everything the native backend needs, it re-derives locally, usually as a coarser approximation. That costs us:

- **Correctness surface.** Two implementations of the same fact drift. The backend's approximations (scope-exit decref of everything, syntactic ownership guesses, default-to-`i64` type fallbacks) each carry their own bug classes. In one case the shadow copy is now *smarter* than the single source (see B5: `.pyi` overload disambiguation exists only in the backend).
- **Performance left on the table.** Coarse liveness means redundant retain/release pairs; no escape analysis means every managed value is heap-allocated; no acyclicity facts means every ref-bearing object pays cycle-collector registration and tracing.
- **Untrustworthy annotations.** On the native path, `own`/`val`/`linear`/borrow annotations are unverified claims (a use-after-move compiles), so the backend cannot exploit them.

Making the required analyses mandatory on the native path dissolves the reason the diagnostics-only contract exists: codegen is deterministic again, just deterministic-with-facts.

# Current state (pipeline reality)

The native path is already halfway there, which makes this refactor tractable:

[...]

PR #7576, opened 2026-07-18
fix(native): diagnose unresolved annotation types instead of silent i64

### LINKED ISSUE #7139: Native: _resolve_jac_type silently int-ifies unresolved annotation types (bypasses the _emit_unlowerable_type backstop) — broken IR, no diagnostic
## Summary

`NaIRGenPass._resolve_jac_type` (`jac/jaclang/compiler/passes/native/na_ir_gen_pass.impl/types.impl.jac`) has **three `return ir.IntType(64)` paths that silently default an unresolved type to `i64` with no diagnostic**. They bypass the pass's own `_emit_unlowerable_type` backstop, which is the helper every *other* unresolvable-type path routes through to emit `E2018`/`E5090`.

This matters because **every caller of `_resolve_jac_type` is a type-annotation / signature position** — function params/returns, global var decls, object fields, set/var decls, lambda types. So a wrong or unresolved annotation silently becomes `i64` storage/ABI in the emitted LLVM IR instead of failing (or at least warning). It is also inconsistent with the stated direction in #6542's release note:

> A missing semantic annotation at codegen is now an `E9002` internal error instead of a silent `i64` fallback.

…which these three paths still are.

## The three silent paths

```jac
impl NaIRGenPass._resolve_jac_type(type_expr: (uni.UniNode | None)) -> ir.Type {
    if type_expr is None {
        return ir.IntType(64);          # (1) line 537 — silent
    }
    ...
    if _ptb is not None and not isinstance(type_expr, uni.Expr) {
        lowered = self._lower_type(_ptb);
        if lowered is not None { return lowered; }
        return self._emit_unlowerable_type(type_expr, _ptb);   # GOOD — emits E2018/E5090
    }
    if isinstance(type_expr, uni.Expr) {
        if type_expr.type is None {
            try { self.prog.get_type_evaluator().get_type_of_expression(type_expr); }
            except Exception as _type_err { ...DEBUG log only... }
        }
        if type_expr.type is None {
            _prim = self._resolve_primitive_name(type_expr);
            if _prim is not None { return _prim; }
            _sname = self._get_name(type_expr);
            if _sname is not None and _sname in self.struct_types { return ...; }
            return ir.IntType(64);      # (2) line 577 — silent (the main one)
        }
        lowered = self._lower_type(type_expr.type);
        if lowered is not None { return lowered; }
        return self._emit_unlowerable_type(type_expr, type_expr.type);  # GOOD
    }
    return ir.IntType(64);              # (3) line 585 — silent
}
```

- **(2) line 577** is the worst: on-demand type eval failed *and* primitive-name lookup failed *and* struct-name lookup failed → `i64`, no `emit(...)`, no `E9002`. This is the one surfaced by the #6355 review.
- **(1) line 537** (`type_expr is None`) and **(3) line 585** (final fallthrough) are the same shape.
- Contrast the two `return self._emit_unlowerable_type(...)` sites — those DO emit a diagnostic (E2018 for `UnknownType`, else E5090) and *then* return `i64`.

`_emit_unlowerable_type` even logs the intent:

```
[NaIRGenPass] _resolve_jac_type: unresolved type %r (type=%s) — defaulting to i64;
consider adding an explicit type annotation
```

…but the three silent paths never reach it, so neither the diagnostic nor that log fires.

## Blast radius — all callers are annotation positions

```
func.impl.jac:32,36,371,379   — function param types + return type
globals.impl.jac:22           — global var declared type
objects.impl.jac:352,397      — object field declared types
objects.impl.jac:452,456,787  — object method param/return + local type
core.impl.jac:737             — var declared type
expr.impl.jac:129,138         — closure/lambda param + return types
stmt.impl.jac:616,767,1342    — set type, redeclared type, var declared type
types.impl.jac:600            — _try_resolve_field_type fallback (object fields!)
```

A wrong annotation in any of these silently becomes `i64`. Concretely, an object field whose type tag on-demand-evaluates to `None` reaches line 577 via `_try_resolve_field_type` → `_resolve_jac_type`, so the field is laid out as `i64` and **#6715's `_emit_unlowerable_type`-based guardrail never fires** (it lives in the helper these paths skip).

## Pre-existing

`git blame` confirms all three silent `return ir.IntType(64)` lines predate the current cycle:

- line 537 / 577 → `f856cc7fd4` (2026-06-10)
- line 585 → `9574e3349c` (2026-06-16)

PR #6355 only added the `DEBUG`-level log in the `except` at lines 556–564 (so on-demand failures are at least *diagnosable*), which is a helpful step but does not change the silent fallback — "fallback behavior is unchanged," per its release note. The silent `i64` itself long predates it.

## Suggested fix

Route the three silent paths through a diagnostic instead of bare `i64`. Two options:

1. **Minimal / consistency:** make (1)/(2)/(3) call a helper shaped like `_emit_unlowerable_type` (which already exists and emits E2018/E5090 + the debug log). (2) has no `type_base` to pass, so it would need an `E9002` ("missing semantic annotation") or an `E2018` variant keyed on the node label — consistent with the #6542 direction quoted above.
2. **Stricter / aligned with #6542:** treat a `None` `.type` at these annotation positions as an `E9002` ICE (the stated "missing semantic annotation → E9002 instead of silent i64" contract), gated behind a flag if needed to avoid regressing programs that currently lean on the i64 default.

Either way, the fix should come with a test that an unresolvable annotation in a native signature/field position produces a diagnostic (not silent `i64`). Note the front-end (`jac check`) catches *some* of these via `E2018` (see #6739), so the repro needs an annotation whose `.type` is `None` at codegen after on-demand eval — the case the front-end doesn't currently preempt.

## Related

- This is the hole **below** the backstop that #6715 hardened and #6739 relies on: #6739's premise is *"Codegen does catch it as E2018 today"* via `_emit_unlowerable_type` — but these three paths never reach that helper.
- #6676 — clib import decls silently defaulting to signed i64 (same class of silent-i64 hazard).
- #6386 — umbrella "Native backend re-derives type/layout/symbol info" (P1) work this slots under.
- Surfaced during review of PR #6355 (the `jaclang/native` debug-log addition in this same function).

chess10kp:
## Summary

`NaIRGenPass._resolve_jac_type` had a bare `return ir.IntType(64)`
fallthrough that silently defaulted a genuinely-unresolved annotation to
`i64` with no diagnostic. Every caller of this helper is a
type-annotation / signature position (function params/returns, object
fields, global/var/set decls, lambda types), so an unresolved type there
became `i64` storage/ABI in the emitted IR — bypassing the pass's own
`_emit_unlowerable_type` backstop (`E5092`), the same helper every other
unresolvable-type path routes through.

Two of the three silent paths the issue originally flagged were already
closed by the single-source-analyses refactor (#7418), which routes the
`uni.Expr`-with-`None`-type case through `_emit_unlowerable_type`. This
PR closes the remaining hole: the non-`None`, non-`Expr` fallthrough
(e.g. an annotation node whose semantic `.type` was never stamped).

## Change

- Route the fallthrough through `_emit_unlowerable_type(type_expr, _ptb)`
  so an unresolvable annotation fails loudly (`E5092`) and withholds the
  native artifact instead of emitting broken IR.
- The `type_expr is None` path is **intentionally** left as `i64`: an
  absent annotation (e.g. an ability with no `-> T` return type) is a
  legitimate default, not an unresolved type. Verified this fires on
  ordinary valid native functions, so diagnosing it would be a false
  positive.

## Testing

Added two cases to `test_native_error_gate.jac`:
- an unresolvable type in a native signature position (`x: type`)
  produces `E5092` and publishes no llvm ir;
- an ability with no return annotation still compiles clean (absent !=
  unresolved).

All 7 tests in the native error-gate suite pass.

Closes #7139


SupulHeshan:
Filed a follow-up: #8931.

This PR fixes the one spot inside `_resolve_jac_type` that still silently defaulted to `i64`. Turns out the same shortcut ("no type written down? just assume `i64`") also shows up in a few places that call `_resolve_jac_type` — they check for a type *before* calling it, and hardcode `i64` if none was written.

I checked all of those. Every one of them guards a named function or method parameter, and Jac already refuses to compile those without a type — so in practice those shortcuts can never fire. Except one: lambda parameters. Lambdas are the one place Jac lets you skip the type, since it can usually figure it out from context (e.g. what the lambda gets passed to).

That's exactly where the bug was live: if a lambda's parameter type *was* correctly figured out (say, inferred as a string), the code generator ignored that and used `i64` anyway — silently, with no error. #8931 fixes it the same way this PR does: ask for the real type first, and only fall back to `i64` when there truly isn't one.

Different file than this PR (`expr.impl.jac` vs `types.impl.jac`), just flagging it as related since both trace back to #7139.

jac/tests/compiler/backends/native/test_native_error_gate.jac
```diff
@@ -139,6 +181,43 @@ test "check-path compile surfaces native codegen errors (check/run parity)" {
...
+    prog = JacProgram();
+    mod = prog.compile(file_path=src, options=native_force_opts());
+    assert not has_error_code(prog, "E5092") , (
+        f"a `type` param annotation resolves structurally and must not be "
+        f"diagnosed, got: {fmt_errors(prog)}"
+    );
+}
+
+test "absent return annotation is a legitimate i64 default, not a diagnostic" {
+    src = write_module("gate_no_ret_annot.na.jac", NO_RET_ANNOT_SOURCE);
```
christianwilkins, line 208: [P2] Use the supported filename and force the native options explicitly. The `.na.jac` marker is rejected on this head, so CI raises ValueError at the following compile call before checking the missing-return behavior. Write `gate_no_ret_annot.jac` and pass `options=native_force_opts()`, like the adjacent native cases, so this regression actually exercises code generation.

christianwilkins (changes requested):
Reviewed `0e9bc302ea216d2a4fab5def6cb476d7c58941f1`, including the complete diff, resolver/backstop context, new cases, existing conversation, and current-head CI. The changed final fallthrough now uses the existing unlowerable-type diagnostic, while the early None case and structural/semantic lowering paths remain intact. The separate lambda-inference follow-up #8931 is already documented in the conversation. Local validation was static inspection and a passing `git diff --check`; I did not rerun the native suite.

P2: the new absent-return regression writes `gate_no_ret_annot.na.jac`. This head rejects the retired `.na.jac` marker before the test reaches any assertion. Current-head workflow 33571341635 reports exactly that ValueError at the new compile call (native/tools: 1,519 passed, 1 error, 4 skipped). Use a `.jac` filename and explicit `native_force_opts()`, as the adjacent native cases do, then verify the no-annotation assertions actually execute.

The separate jac-check lane fails in lsp/server/engine.jac and dist/publish/npm_sources.jac (914 passed, 2 failed); I have not established those as regressions from this one-line resolver change. GitHub reports merge conflicts. No code changes were made. Changes requested: repair the regression test, resolve conflicts, and obtain passing checks on the resulting head.

RandiJay96:
closing this. The complete solution will be added https://github.com/jaseci-labs/jac/pull/8931

PR #7579, opened 2026-07-18
fix(scale): deterministic guest root id via UUIDv5 to eliminate dangling-pointer repair (#6698)

### LINKED ISSUE #6698: fix(jac-scale): deterministic guest root id via UUIDv5 to eliminate dangling-pointer repair
## Problem

`ensure_root_id` mints a random UUID for the guest root the first time a project is accessed. If that anchor is later lost (DB migration, pod restart with partial data, row pruned), the stored id becomes a dangling pointer. The current fix (#6498) repairs this by re-invoking the resolver and writing a new random id — but:

- **In-process**: the `threading.Lock` in `get_shared_root` ensures only one thread re-resolves. ✓
- **Cross-pod**: two pods can each detect the stale id and independently write different replacement ids, producing divergent guest roots with separate graphs. The process-local lock does not coordinate across pods.

## Proposed fix

Derive the guest root id deterministically:

```
guest_root_id = UUIDv5(namespace=project_uuid, name="__guest__")
```

Because the id is a pure function of the project, any pod that "repairs" it will always write the same id. The dangling-pointer class disappears entirely — no repair loop needed, just a single idempotent upsert on cold boot.

## Migration

Existing deployments have rows keyed by a random UUID. A one-time migration (or a dual-lookup fallback that checks the old random id before writing the deterministic one) will be needed.

## Scope

- `jac-scale`: `ensure_root_id` / `IdentityStorage`
- Possibly `jaclang` core `get_shared_root` (the self-heal retry can be simplified or removed once ids are deterministic)

## Related

- #6498 (the band-aid repair this would replace)
- `_shared_root_lock` comment in `runtime.jac` references this issue

akindu-k:
Fixes #6698.

## The problem

Every anonymous visitor to a Jac Scale app shares one "guest" workspace: a single root node that holds the public/anonymous graph. To find that workspace on each request, the server looks up an ID that points at it.

Until now, that ID was a **random** value generated the first time the guest workspace was created, then saved in the user database. The catch with a random value is that it can never be recomputed. If the database record holding it was ever lost (a migration, a data prune, a partial restore) or the workspace node it pointed at disappeared, nothing in the system could work out what the ID was supposed to be.

That produced two failures:

- **Broken pointer:** the ID survived but the workspace node was gone, so `get_shared_root` couldn't resolve it and every anonymous request errored out.
- **Split-brain guest workspaces:** when the server tried to auto-repair by generating a *new* random ID, different server instances (pods) each generated their *own* different ID at roughly the same time. Visitors then landed on several different, disconnected guest workspaces depending on which pod happened to answer them.

The earlier attempt to patch this (PR #6498) healed the broken pointer at runtime, but because each pod invented a different replacement value, it made the split-brain worse rather than better. The real issue was never the missing node; it was that the ID was random and therefore unrecoverable.

## How we fix it

Instead of a random ID, we now **calculate** the guest workspace ID from stable inputs using UUIDv5 (a UUID that is a pure function of its inputs):

```
guest_root_id = uuid5( uuid5(fixed_namespace, project_name), "__guest__" )
```

Because the ID is calculated rather than random:

- **Every server instance arrives at the exact same ID**, every time, with no coordination between pods. Split-brain becomes impossible: any "repair" is just every pod writing the same value.
- **The ID can never be "lost."** If the database record disappears, the server simply recomputes the same value and points straight back at the existing workspace.
- **It is scoped per project** (via the project name from `jac.toml`), so different apps sharing one backend get their own separate, stable guest workspaces instead of colliding on one.

We also made the lookup self-healing and safe for existing deployments:

- If the workspace node is missing, `get_shared_root` now recreates it at the resolved ID and saves it durably, instead of throwing an error. Since every pod computes the same ID, this is an idempotent get-or-create with no locks and no races.
- Existing deployments that already have a random guest ID keep using it for as long as it still resolves, so no anonymous data is stranded and no migration is forced. They only converge to the calculated ID if their old one is already broken (in which case there was nothing left to lose).

## What changed

- New helper that computes the deterministic, per-project guest root ID.
- The guest user is now created with that deterministic ID.
- The shared-root resolver keeps a still-valid existing ID, otherwise falls back to (and records) the deterministic one.
- `get_shared_root` now get-or-creates the shared root instead of raising when the node is missing.
- Removes the need for the random-replacement repair path from #6498.

## Testing

- All changed modules compile cleanly.
- Verified the ID is identical across repeated computation for the same project and differs across projects.
- Existing shared-root behaviour is covered by `test_shared_root` (runs in CI).


MusabMahmoodh (commented):
Apologies for the 32-day wait. Rather than leave line comments on a branch
this old, I checked the whole thing against current `main` first. Three
things moved under you while it sat, and they change what this PR costs to
land. Laying them out so the decision is yours with the facts in hand.

## 1. Half the problem you named is now handled, by code written after this PR

Your body leads with split-brain: pods each inventing a different replacement
id. `IdentityStorage.ensure_user_by_id` on `main` today
(`identity_storage.impl.jac:99-135`) reconciles that race - when `_store_user`
fails on a concurrent create it re-reads and returns the winner's `root_id`
rather than keeping its own:

```
} except Exception {
    winner = self.get_user_by_id(user_id);
    if winner {
        return winner.get('root_id');
    }
    raise;
}
```

That is not a computable id and it is not what you proposed, but it does close
the specific mechanism the PR body leads with: concurrent pods now converge on
one row. It landed with the Postgres work, well after you opened this. Whatever
gets re-filed should be argued against this code, not against the July code.

## 2. The other half is still live, so the issue should stay open

The id is still random and still unrecoverable. On `main`:

- `identity_storage.impl.jac:73`, `:111` and `:148` all mint
  `root_id = user_root.__jac__.id.hex` from a fresh `Root()` - a random node id.
- `get_root_id` (`identity_storage.impl.jac:283-285`) reads it straight back off
  the user row.
- `serve.core.impl.jac:279-285` still resolves the shared root through
  `_um.get_root_id(Con.GUEST.value)`.
- `reset_root` (`:137-151`) "repairs" by minting **another** random id.

So the failure you actually cared about - lose the row, lose the workspace,
with nothing in the system able to recompute the id - is exactly as true today
as when you filed it. `#6698` should stay open.

## 3. The layer the diff patches was deleted

The branch is **373 commits behind** `main` and conflicts in
`user_manager.impl.jac`. Under it, since your merge-base:

| file | what happened |
|---|---|
| `identity_storage.sqlite.impl.jac` | deleted, -377 |
| `identity_storage.mongo.impl.jac` | deleted, -178 |
| `identity_storage.pg.impl.jac` | new, +360 |
| `identity_storage.impl.jac` | +76 |
| `user_manager.impl.jac` | +102, and the conflict |
| `serve.core.impl.jac` | 289 lines changed |

Eight commits touched `scale/identity/` in that window (`#7945`/`#7947`
Postgres-native persistence, then `#8000`/`#8002` retiring the Mongo/Redis/
SQLAlchemy scaffolding). Resolving this is not a rebase, it is writing the
change again against a storage layer that did not exist in July.

## Your call, and both paths look reasonable to me

**Refresh in place:** resolve the conflict and rewrite the storage-side hunks
against `identity_storage.pg.impl.jac`. Keeps the history and the review
context on one thread.

**Or close this and re-file against `#6698`:** one deterministic
`guest_root_id` derived per project, read at `get_root_id` time, argued
against `ensure_user_by_id` as it now stands. That version should come out
considerably smaller than the 103 lines here, because the two storage backends
you had to thread it through no longer exist.

I lean slightly toward the second only because point 1 means the PR body
argues for something the code now partly does, so the case has to be rewritten
either way - but that is a preference, not a request, and you know this code
better than I do.

Whichever you pick: the UUIDv5 derivation itself I still think is right, and
the reasoning in your body is the clearest statement of this bug anyone has
written. If the PR does close, please lift that text into `#6698` verbatim so
it is not lost. And if you would rather hand `#6698` off than carry it, say so
and I will take it or find it an owner. **Rig offer:** the multi-pod convergence leg is the part
that is genuinely hard to prove locally. I have an EKS/kind rig on jac-ec2 that
can run a real 2-replica deploy, drop the guest row, and show whether both
replicas land on the same workspace. That evidence is yours if you want it.

marsninja:

Issue #10, opened 2021-10-30
Need to create pypi project and prep jaseci core package accordingly


marsninja:
We need to create the public jaseci pypi project so anyone can `pip install jaseci`

PR #7603, opened 2026-07-20
fix(runtime): dedup Serializer unregistered-class warnings to one per class per process (#7602)

### LINKED ISSUE #7602: runtime: Serializer logs an unbounded WARNING per anchor for unregistered classes
## Summary

`Serializer._get_class` logs an unconditional `WARNING` every time it meets an unregistered class, with no dedup and no rate limit, on the per-anchor deserialization hot path. Any process that reads graph anchors whose archetype classes it does not define emits one warning per anchor read, forever, for the same handful of class names. In production (jachammer, jac-builder) this reaches roughly 85 lines/sec/service and buries real errors. Encountering an unregistered class is a supported, expected outcome of `_get_class` (there is a test asserting it returns `None`), so `WARNING`-per-occurrence is the wrong level and the wrong cardinality.

## Environment

| | |
|---|---|
| Binary | `jac 0.34.0 (Darwin arm64)`, released asset `jac-0.34.0-macos-aarch64` from `v0.34.0` |
| Source | `origin/main` @ `b7c06cded`, code path unchanged |
| Repro dir | fresh `/tmp` dir, outside any jaseci checkout |
| Also observed | prod EKS, namespace `jachammer`, services `builder-sv` and `jac-coder-sv`, 2026-07-14 |

Two programs. The first writes the on-disk anchor shape for a node class it owns. The second reads that same anchor without defining the class, which is exactly what a service does when it loads anchors written by another program sharing the store.

`dump.jac`:

```jac
import json;
import from jaclang.runtimelib.serializer { Serializer }

node DonationNode { has amount: int; }

with entry {
    n = DonationNode(amount=42);
    blob = Serializer.serialize(n.__jac__, include_type=True);
    with open("anchor.json", "w") as f {
        json.dump(blob, f);
    }
    print(json.dumps(blob));
}
```

`load.jac` (note: `DonationNode` is deliberately not defined here):

```jac
import json;
import logging;
import from jaclang.runtimelib.serializer { Serializer }

with entry {
    # a jac-scale service logs at WARNING; `jac run` installs no handler by default
    logging.basicConfig(level=logging.WARNING);
    with open("anchor.json") as f {
        blob = json.load(f);
    }
    for _ in range(200) {
        Serializer.deserialize(blob);
    }
    print("deserialized the SAME foreign anchor 200 times");
}
```

Run:

```
$ ./jac-0.34.0-macos-aarch64 --version
jac 0.34.0  (Darwin arm64)

$ ./jac-0.34.0-macos-aarch64 run dump.jac
{"__type__": "NodeAnchor", "__module__": "jaclang.jac0core.archetype", "id": "7a45bcf9-...",
 "root": null, "persistent": false, "access": {"all": -1, "roots": {"anchors": {}}},
 "archetype": {"__type__": "DonationNode", "__module__": "__main__", "amount": 42},
 "edges": [], "version": 0}

$ ./jac-0.34.0-macos-aarch64 run load.jac > load.log 2>&1 ; echo "exit=$?"
exit=0
```

Result:

```
$ grep -c "Refused to deserialize unregistered class: __main__.DonationNode" load.log
200
$ grep -c "archetype reconstruction failed" load.log
200
$ grep -c WARNING load.log
400
```

Verbatim, repeated 200 times each:

```
WARNING - Refused to deserialize unregistered class: __main__.DonationNode
WARNING - Serializer: archetype reconstruction failed for __main__.DonationNode; dropping anchor (caller should quarantine).
```

200 reads of one anchor, one class name, produce 400 WARNING lines. There is no bound on this.

## Expected

Unresolvable classes are a normal condition, not an error condition, so the diagnostic should be constant-size in the number of anchors read: logged once per distinct class per process (or at `debug`). A service reading N foreign anchors should not produce O(N) warnings for a set of distinct classes whose size is small and fixed.

## Actual

Both warnings are unconditional, in the innermost per-anchor path.

- `jac/jaclang/runtimelib/impl/serializer.impl.jac:523` in `Serializer._get_class`, reached from:
  - `_deserialize_archetype` (`:653`), once per anchor loaded from the store
  - `deserialize` (`:273`), once per nested typed value
- `jac/jaclang/runtimelib/impl/serializer.impl.jac:562`, the follow-on `archetype reconstruction failed` warning, once per anchor

`_get_class` returning `None` for an unknown name is deliberate behavior with a test covering it (`jac/tests/runtimelib/test_layer3_coercion_alias.jac:59`, `"_get_class returns None for truly unregistered class names"`), which is why `WARNING` is the wrong level here.

Field measurement, `builder-sv`, 5 minutes of one production namespace, counting only the first of the two lines:

| class | lines / 5 min |
|---|---|
| `__main__.DonationNode` | 4740 |
| `__main__.VolunteerNode` | 3950 |
| `__main__.ShelterNode` | 3950 |
| `__main__.BusinessNode` | 3950 |
| `services.ideas.Idea` | 3160 |
| `__main__.HubNode` | 2765 |
| `__main__.Offers` | 790 |

Seven distinct class names, 23305 log lines. The useful information content is seven lines.

## Impact

Anyone running a jac-scale service against a store shared with other programs, which is the normal multi-app deployment shape. Two costs:

1. Logging overhead on the anchor-read hot path, under exactly the load where it hurts most. Each line is formatted and emitted even when the handler drops it, because the f-string is evaluated at the call site.
2. Observability loss. At tens of thousands of identical lines per minute, genuine errors are not findable. This was first noticed during a live load test, when it mattered.

## Workaround (verified)

None at the jac level. A logging filter on `jaclang.runtimelib.serializer` in the host application suppresses it, but that also hides the first, useful occurrence, and it requires every downstream service to know to do this.

## Suggested fix

In `Serializer._get_class`, keep a module-level `set[str]` of already-reported keys and log at most once per key per process, or drop the line to `debug` outright. Same treatment for the paired warning at `:562`, keyed on the same class name. Both are on the same per-anchor path, so fixing only one leaves half the volume.

Regression test: deserialize the same unregistered-class anchor N times under `caplog` and assert exactly one `WARNING` record for that class, not N. That is a direct extension of the existing `_get_class` coverage in `jac/tests/runtimelib/test_layer3_coercion_alias.jac`.

## Related

Split out from jaseci-labs/jacBuilder#689, which tracks the separate question of why those services traverse foreign anchors in the first place. That half is not a jaclang bug. This half is, and it is worth fixing independently, because a service reading anchors it does not own is legitimate in a shared store and should not be punished with unbounded logging.

akindu-k:
## What was wrong

When `Serializer` deserializes a stored anchor and cannot fully reconstruct its archetype, it logged a `WARNING` **for every anchor read**, with no dedup and no rate limit, on the innermost per-anchor path. In a shared store where multiple apps read each other's anchors (the normal multi-app shape), a service reading N anchors produced O(N) warnings for a small, fixed set of problems. In production this reached ~85 lines/sec/service; in one 5-minute window seven distinct problems produced 23,305 log lines whose useful content was seven lines. Two costs: genuine errors get buried, and each f-string is formatted on every read even when the handler drops it.

There are **two distinct flood situations**, and both are fixed here:

- **Situation A — the class is unknown in this process.** The registry lookup fails and the code logged, per anchor:
  - `Refused to deserialize unregistered class: <class>` (`_get_class`)
  - `Serializer: archetype reconstruction failed for <class>; dropping anchor` (`_deserialize_anchor`)
- **Situation B — the class IS registered but its stored data has drifted.** The lookup *succeeds*, then `_deserialize_archetype` logged **per field, per anchor** (coercion fallback, stored-None, required-missing). This is worse than A because it scales with fields, not just anchors. The first version of this PR did not cover it; this was raised in review and is now fixed.

Fixes #7602.

## How it's fixed

An unknown class or a drifted field is a normal, expected condition, so the diagnostic should be **constant-size in the number of anchors read** — logged once per distinct problem per process, not once per anchor.

**A single dedup primitive, `Serializer._should_warn_once(scope, key)`.** It is backed by a process-wide `set[str]` and stores `f"{scope}:{key}"`:

- `scope` names the *failure mode* (a fixed label); `key` names the *thing it happened to* (the class name, or `module.type.field`).
- Every warning is wrapped as `if Serializer._should_warn_once(scope, key) { logger.warning(...) }`, so the f-string is not even built on repeat occurrences — this removes the per-call formatting cost, not just the emit cost.

**All seven per-anchor warnings on the load path are now gated:**

| scope | fires when | dedup key |
|---|---|---|
| `unregistered` | class not registered here | `module.type` |
| `alias-missing` | rename-alias points to an unregistered target | `module.type` |
| `archetype-failed` | archetype reconstruction returned `None`; anchor dropped | `module.type` |
| `coercion-fallback` | a field value can't be coerced to its declared type | `module.type.field` |
| `stored-none-no-default` | required field stored `None`, no default to recover | `module.type.field` |
| `stored-none-defaulted` | required field stored `None`, recovered from default | `module.type.field` |
| `required-missing` | required field absent from stored data | `module.type.field` |

**Keys are namespaced per failure mode (the `scope`).** Distinct failure modes about the same class no longer collapse into one another. In particular the `unregistered` and `archetype-failed` lines for the same class each emit once (they are *not* collapsed to a single line), and a class that first fails as unregistered and later fails per-field still gets a line for each mode.

**Level stays at WARNING.** jac-scale services log at WARNING, so the first, useful occurrence stays visible in production. The bug was cardinality, not level.

**Thread-safe (double-checked locking).** The steady-state path (already warned) is a lock-free `key in seen` check; the lock is taken only on the first occurrence of a key, where a second check under the lock guarantees exactly one thread records-and-logs it under concurrency.

**Bounded memory, with DEBUG past the cap.** The set is bounded (`_MAX_WARNED_CLASS_KEYS = 1024`, far beyond the realistic count) to prevent unbounded growth on a pathological store; it emits a one-time saturation notice and then downgrades further new keys to `DEBUG`, guarded by `isEnabledFor` so it costs nothing when DEBUG is off, in a lock-free branch.

**Test seam.** `Serializer._reset_warn_cache()` clears the set and the saturation flag so tests are order-independent instead of relying on unique names to dodge the process-global set (same pattern `prometheus_client` uses for its once-only warnings).

## Verification

Reproduced the issue from the report (`dump.jac` / `load.jac`): write an anchor for a class one program owns, read it 200 times in a program that does not define the class.

| Situation A (200 reads of one unknown-class anchor) | Before | After |
|---|---|---|
| `Refused to deserialize unregistered class` | 200 | **1** |
| `archetype reconstruction failed` | 200 | **1** |
| **Total WARNING lines** | **400** | **2** (one per failure mode, namespaced) |

Situation B (100 reads of one drifted-but-registered archetype) drops the per-field `coercion fallback` lines from 100 to **1** in the same way.

### Regression tests (in `test_layer3_coercion_alias.jac`)

- **Situation A** — deserialize the same unknown-class anchor 50 times; assert the `unregistered` line and the `archetype-failed` line each appear **exactly once** (namespaced, so each mode fires once).
- **Situation B** — deserialize the same drifted registered archetype 50 times; assert the per-field `coercion fallback` warning appears **exactly once**.
- Both reset the dedup cache before and after via `_reset_warn_cache()` so they are order-independent.

```
test_layer3_coercion_alias.jac::unregistered-class warnings are emitted once per class, not once per anchor PASSED
test_layer3_coercion_alias.jac::per-field coercion-fallback warning is emitted once per field, not once per anchor PASSED
```

## Backward compatibility

- Every log **message string is unchanged** — greps and alerts that match those messages still work; only the repeated-line volume drops.
- No public API, config, or behavior change for correct programs. All new fields/methods are private to the serializer.

## Note (not addressed here)

One more per-anchor warning, `archetype field is not a dict; dropping anchor` in `_deserialize_anchor`, is left ungated. It fires only for structurally malformed anchors (the archetype field is not a dict at all), a rare corruption case rather than the drift/unknown-class flood this PR targets, and was outside the reviewed set. It can be gated the same way if we want the constant-size-logging invariant to be absolute.


jac/jaclang/runtimelib/impl/serializer.impl.jac
```diff
@@ -514,16 +514,51 @@ impl Serializer._get_class(module_name: str, class_name: str) -> (type | None) {
...
         return None;
     }
-    logger.warning(f"Refused to deserialize unregistered class: {key}");
+    if Serializer._should_warn_once(key) {
+        logger.warning(f"Refused to deserialize unregistered class: {key}");
+    }
     return None;
 }
 
+impl Serializer._should_warn_once(key: str) -> bool {
```
udithishanka: the dedup primitive itself is fine, but it only gates 3 of the 7 unbounded warnings on this path. _deserialize_archetype still has four more at L745, L752, L760 and L790, and those are worse than the two you fixed because they fire per field per anchor rather than per anchor.

checked against the released 0.34.0 binary, 100 deserializes of one drifted archetype whose class is registered:

```
$ grep -c "coercion fallback" drift.log
100
WARNING - Serializer: coercion fallback on __main__.Drifted.count: stored=str, target=<class 'int'>; stored value kept as-is.
```

your dedup can't help there because _get_class succeeds, so nothing reaches _should_warn_once at all. so the constant-size-in-anchors-read claim in the description only holds for the unregistered-class case, and a store with schema drift floods exactly like it did before. same gate works here, just key it on module.type.field so distinct fields still each get their one line.

jac/jaclang/runtimelib/impl/serializer.impl.jac
```diff
@@ -558,11 +593,13 @@ impl Serializer._deserialize_anchor(data: dict[str, object]) -> (Anchor | None)
             }
             archetype = Serializer._deserialize_archetype(arch_data, anchor);
             if archetype is None {
-                logger.warning(
-                    "Serializer: archetype reconstruction failed for "
-                    f"{arch_data.get('__module__')}.{arch_data.get('__type__')}; "
-                    "dropping anchor (caller should quarantine)."
-                );
+                arch_key = f"{arch_data.get('__module__')}.{arch_data.get('__type__')}";
+                if Serializer._should_warn_once(arch_key) {
```
udithishanka: arch_key here is byte-identical to the key _get_class builds, same module and type strings, and _get_class always runs first and claims the slot. that's why this line goes to 0 in your table rather than 1.

collapsing the two for the unregistered case is reasonable. the problem is the key isn't namespaced per call site, so the collapse also crosses unrelated failure modes. a class that first fails as unregistered, then gets registered once its module is imported, then starts failing per-field, never emits this line even once. prefixing at each site, unregistered: vs archetype-failed:, keeps it constant-size and costs a string concat.

not urgent since jac db quarantine list gives operators the structured channel for dropped anchors, so this is log fidelity rather than data-loss visibility.

jac/tests/runtimelib/test_layer3_coercion_alias.jac
```diff
@@ -62,6 +64,41 @@ test "_get_class returns None for truly unregistered class names" {
...
+    try {
+        for _ in range(50) {
+            assert Serializer.deserialize(blob) is None;
+        }
+    } finally {
+        logger.removeHandler(capture);
+        logger.setLevel(prev_level);
+    }
+    count = buf.getvalue().count(key);
+    assert count == 1 , f"expected exactly one warning for {key}, got {count}";
```
udithishanka: this locks in the collapse from the other comment. if you namespace the dedup keys it becomes 2, which is the count you'd actually want to be asserting.

also the Ghost_uuid4 name is dodging the process-global set that never clears. that's consistent with _registry which never clears either so I won't push hard, but any future test asserting a first-occurrence warning is going to be order-dependent. prometheus_client handles it by having its test clear _warned_once, probably worth the same hook here.

udithishanka (changes requested):
3 issues, the main one being that this only covers part of the flooding.

udithishanka (approved):

Thamirawaran:
merging since PM needs it urgent. 

but not a complete fix. the same per anchor logging pattern is still there in a few info/error paths in serializer.impl.jac, including the drift case this pr claims to cover. we also lost the occurrence count, so we can't tell 1 from 1M anymore.

@akindu-k can u do a followup pr for these

akindu-k:
> merging since PM needs it urgent.
> 
> but not a complete fix. the same per anchor logging pattern is still there in a few info/error paths in serializer.impl.jac, including the drift case this pr claims to cover. we also lost the occurrence count, so we can't tell 1 from 1M anymore.
> 
> @akindu-k can u do a followup pr for these

sure I'll do them

akindu-k:
Follow-ups are up, split along the two gaps raised here:

- **#7629** — completes the coverage. Gates the remaining ungated per-anchor info/warning/error paths (alias-resolution success path, the `repair`/drift info lines, attic-preserved, non-dict archetype, and the three outer `Failed to deserialize` error paths), each namespaced with its own scope.
- **#7630** — restores the occurrence count. Replaces the boolean once-gate with a counter that re-emits each line on a power-of-10 ladder (`#1, #10, #100, ...`) with the running count in the message, so 1 vs 1M is distinguishable again while volume stays O(log N). Stacked on #7629.

@Thamirawaran @udithishanka

PR #7608, opened 2026-07-20
fix(compiler): backticked `root entry trigger resolves to the Root class

### LINKED ISSUE #7607: Backticked `root entry trigger resolves to the root() function instead of the Root class
### Summary

A backticked `` `root `` entry trigger compiles to the jaclib `root()` **function** instead of the `Root` **class**, so spawning the walker fails at runtime.

### Reproduce

```jac
walker Ping {
    can go with `root entry {
        report "pong";
    }
}

with entry {
    root spawn Ping();
}
```

### Actual

```
TypeError: issubclass() arg 2 must be a class, a tuple of classes, or a union
```

The trigger lowers to a reference to the `root()` accessor function. The dispatch path then calls `issubclass(node_type, trigger)` against a function object.

### Expected

`` `root `` names the `Root` archetype, so the trigger should resolve to the `Root` class and the walker should run, reporting `pong`.

### Cause

`PyastGenPass.exit_ability` does not normalize the backticked `root` name, and `exit_event_signature` normalizes only the already-capitalized `Root`:

```jac
if (isinstance(n, ast3.Name) and (n.id == 'Root')) {
    self.jaclib_imports.add('Root');
}
```

Neither path rewrites the lowercase `root` produced by the backtick escape, so codegen emits the function reference.

### Scope

Affects the shared Python codegen path, so `jac start`, `jac run`, and `jac build` are all impacted equally. Not specific to any one command.

### Fix

Normalize `root` to `Root` in both `exit_ability` and `exit_event_signature`, and register the `Root` jaclib import. Covered by a regression test that actually spawns the walker rather than only checking codegen text.

kashmithnisakya:
Fixes #7607.

## Summary

A backticked `` `root `` entry trigger compiled to the jaclib `root()` **function** instead of the `Root` **class**, so spawning the walker died with `TypeError: issubclass() arg 2 must be a class`.

`PyastGenPass.exit_ability` never normalized the backticked name, and `exit_event_signature` matched only the already-capitalized `Root`:

```jac
if (isinstance(n, ast3.Name) and (n.id == 'Root')) {
    self.jaclib_imports.add('Root');
}
```

Neither path rewrote the lowercase `root` the backtick escape produces, so codegen emitted a reference to the accessor function and dispatch then ran `issubclass()` against it.

Both paths now normalize `root` to `Root` and register the jaclib import.

## Scope

This is in the shared Python codegen path, so `jac run`, `jac start`, and `jac build` were all affected equally.

## Verification

`jac test jac/tests/language/test_bugs.jac`: 2/2.

The regression test spawns the walker and asserts on its report rather than grepping generated source, so it fails on the original `TypeError` rather than on a codegen text change.

## Note

Split out of #7517, where this was found while validating `jac build --as source`. It is unrelated to source conversion and stands on its own.

### DIFF
--- a/docs/docs/community/release_notes/unreleased/jaclang/7608.bugfix.md
+++ b/docs/docs/community/release_notes/unreleased/jaclang/7608.bugfix.md
@@ -1,0 +1,1 @@
+- **Fix**: a backticked `` `root `` entry trigger compiled to the jaclib `root()` function instead of the `Root` class, so spawning the walker failed with `TypeError: issubclass() arg 2 must be a class`. Both `exit_ability` and `exit_event_signature` now normalize the backticked `root` to `Root` and register the jaclib import.
--- a/jac/jaclang/jac0core/passes/impl/pyast_gen_pass.impl.jac
+++ b/jac/jaclang/jac0core/passes/impl/pyast_gen_pass.impl.jac
@@ -1564,6 +1564,12 @@ impl PyastGenPass.exit_ability(nd: uni.Ability) -> None {
                 trigger_arg = tinfo;
             }
         }
+        for n in ast3.walk(trigger_arg) {
+            if (isinstance(n, ast3.Name) and (n.id in ('Root', 'root'))) {
+                n.id = 'Root';
+                self.jaclib_imports.add('Root');
+            }
+        }
 
         trigger_thunk = self.sync(
             ast3.Lambda(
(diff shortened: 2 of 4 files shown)


Thamirawaran:
thanks for the writeup, but i think this one's a non-issue.

`root as a walker trigger was deliberately removed in #4496 (0.10.0) along with the backtick type operator. backtick is now only the keyword escape, so `root is the root value, not the Root class. using it as a trigger is an intentional hard error (E1116), finalized for the backticked form in #7512.

normalizing root to Root in codegen makes jac check reject the file while jac run happily executes it. it also makes the escaped spelling more permissive than the plain one: `with root entry` still errors with "use `with Root entry`", but `with `root entry` would now silently work. the new test calls JacRuntime.jac_import directly, so it bypasses the checker and never sees the conflict.

the real gap is that the runtime guard in archetype.impl.jac:358 only catches Archetype instances, so the jaclib root function slips past it into a raw issubclass TypeError instead of the clear "use Root" message. that's worth a narrow issue on its own.

migration is just `with Root entry`.

closing this one.

PR #7612, opened 2026-07-21
fix(scale): converge concurrent find-or-create on system_root (#7605)

### LINKED ISSUE #7605: jac-scale: concurrent find-or-create on system_root duplicates singletons and raises WriteConflict on superroot anchor
## Summary

Concurrent requests that find-or-create a singleton node anchored to `system_root` (the all-zeros superroot, `00000000-0000-0000-0000-000000000000`) race each other: every request that loses the race still creates its own copy of the "singleton" and attaches another edge to the shared root, and the optimistic version check on that anchor then fails with `WriteConflict`. Requests surface the conflict as a 500, and the graph is left with duplicate singletons that no later request cleans up. In production this has accumulated 82 duplicates of one singleton and 68 of another on a single superroot.

The root of it is that `system_root` is a process-global anchor shared by every request context (`jaclang/scale/memory/impl/context.impl.jac:19-29`), and the commit that writes it has no conflict handling at the layer where it runs.

## Environment

- `jac 0.34.0  (Darwin arm64)`, released binary `jac-0.34.0-macos-aarch64` from the v0.34.0 release, run from a fresh directory outside any jaclang checkout.
- MongoDB 7 (docker, single-node replica set) as L3, Redis as L2. Both are required to see the conflict: with Mongo alone the duplication still happens but the CAS does not fire (see Actual).
- Also observed in production on jachammer (EKS, 8 `builder-sv` replicas) running jac-scale from main, which is what prompted this. Downstream report: jaseci-labs/jacBuilder#688.

`app.jac` - find-or-create a singleton `Hub` on the system root:

```jac
import from jaclang { JacRuntimeInterface as Jac }

node Hub {
    has label: str;
}

"""Find-or-create a singleton node anchored to the shared system root."""
walker:pub touch_hub {
    can go with entry {
        sys = Jac.get_context().system_root.archetype;
        hubs = [n for n in [sys-->] if isinstance(n, Hub)];
        if not hubs {
            h = Hub(label="community");
            sys ++> h;
            hubs = [h];
        }
        report {"hubs": len(hubs)};
    }
}

with entry {
    # Let request-context walkers write the shared system root.
    sys_root = Jac.get_context().system_root.archetype;
    grant(sys_root, level=WritePerm);
}
```

`jac.toml`:

```toml
[project]
name = "superroot-repro"
version = "0.1.0"
entry-point = "app.jac"

[dependencies]
pymongo = ">=4.6"
redis = ">=5.0"
```

`hammer.py` - fire N concurrent requests, round-robined across the replicas:

```python
import json, sys, urllib.request, urllib.error
from concurrent.futures import ThreadPoolExecutor

PORTS = [8000, 8001, 8002]
N = int(sys.argv[1]) if len(sys.argv) > 1 else 40


def call(i):
    req = urllib.request.Request(
        f"http://localhost:{PORTS[i % len(PORTS)]}/walker/touch_hub",
        data=b"{}", headers={"Content-Type": "application/json"}, method="POST")
    try:
        with urllib.request.urlopen(req, timeout=60) as r:
            return r.status, r.read().decode()
    except urllib.error.HTTPError as e:
        return e.code, e.read().decode()


with ThreadPoolExecutor(max_workers=N) as ex:
    res = list(ex.map(call, range(N)))

codes = {}
for s, _ in res:
    codes[s] = codes.get(s, 0) + 1
print("status counts:", codes)
for s, b in res:
    if s != 200:
        print("FAILED:", b[:200])
```

Run it:

```bash
# 1. backing stores
docker run -d --name superroot-mongo -p 27019:27017 mongo:7 --replSet rs0
docker exec superroot-mongo mongosh --quiet --eval 'rs.initiate()'
# redis on localhost:6379, db 11 used below

# 2. deps
jac install

# 3. three replicas against the same Mongo + Redis, cold database
for p in 8000 8001 8002; do
  MONGODB_URI="mongodb://localhost:27019/?directConnection=true" \
  REDIS_URL="redis://localhost:6379/11" \
  jac start app.jac --port $p &
  sleep 14
done

# 4. 40 concurrent find-or-create calls
python3 hammer.py 40
```

Result (exit code 0 from the driver; the failures are in the responses):

```
status counts: {200: 36, 500: 4}
FAILED: {"error": "WriteConflict: WriteConflict: anchor 00000000-0000-0000-0000-000000000000 changed concurrently (expected v0, found v1)"}
FAILED: {"error": "WriteConflict: WriteConflict: anchor 00000000-0000-0000-0000-000000000000 changed concurrently (expected v0, found v1)"}
FAILED: {"error": "WriteConflict: WriteConflict: anchor 00000000-0000-0000-0000-000000000000 changed concurrently (expected v0, found v1)"}
FAILED: {"error": "WriteConflict: WriteConflict: anchor 00000000-0000-0000-0000-000000000000 changed concurrently (expected v0, found v1)"}
```

Graph state after that single burst (paths shortened):

```
superroot version=2 edges=2 Hub anchors=3
```

Three `Hub` anchors exist where the walker's logic allows exactly one, and two of them are attached to the superroot.

Server-side traceback for the conflict:

```
  File "<jac>/jaclang/runtimelib/impl/server.impl.jac", line 94, in reset_root
  File "<jac>/jaclang/jac0core/runtime.jac", line 204, in commit
  File "<jac>/jaclang/scale/memory/impl/memory_hierarchy.main.impl.jac", line 139, in commit
  File "<jac>/jaclang/scale/memory/impl/memory_hierarchy.mongo.impl.jac", line 613, in apply
  File "<jac>/jaclang/scale/memory/impl/memory_hierarchy.mongo.impl.jac", line 649, in _apply_one
  File "<jac>/jaclang/scale/memory/impl/memory_hierarchy.mongo.impl.jac", line 772, in _put_node_atomic
jaclang.runtimelib.exceptions.WriteConflict: WriteConflict: anchor 00000000-0000-0000-0000-000000000000 changed concurrently (expected v0, found v1)
```

Control runs, same setup:

| run | requests | outcome |
|---|---|---|
| Mongo + Redis, cold DB | 40 concurrent | 4x 500 `WriteConflict` on `000...0`, 3 `Hub` anchors |
| Mongo only, cold DB | 40 concurrent | no `WriteConflict`, but **7** `Hub` anchors created, only 1 edge on the superroot |
| Mongo + Redis, hub already exists | 400 concurrent | all 200, no new hubs |

So the duplication is the primary defect and happens with or without the L2 cache; the `WriteConflict` is what the CAS reports when Redis is in play. Once the singleton exists the race is over, which is why in production this shows up in bursts after restarts rather than continuously.

## Expected

Two concurrent find-or-create passes over the same shared root should converge on one node. At minimum, a losing writer should be able to re-read and retry rather than have its request fail, and the graph should not end up with N copies of a singleton whose creation code is guarded by an existence check.

## Actual

Every concurrent request that finds no hub creates one and attaches it to the superroot. The commit is version-gated on that anchor, so all but one lose, and the loser's request fails outright. The duplicates that did land stay in the graph forever.

Root-cause pointers on `main` (27dbb50):

- `jac/jaclang/scale/memory/impl/context.impl.jac:19-29` - the superroot anchor is cached in a process-global `_process_cache` and the same object is injected into every request context's `__mem__`, so one request's mutation is visible to all in-flight contexts in that process.
- `jac/jaclang/scale/memory/impl/memory_hierarchy.mongo.impl.jac:728-776` - `_put_node_atomic` CAS-filters on `data.version` and raises `WriteConflict` when the document moved.
- `jac/jaclang/scale/memory/impl/memory_hierarchy.main.impl.jac:135-138` - `commit()` calls `self.changes.clear()` **before** re-raising the conflict, so the losing changeset cannot be replayed by a caller that catches it.
- `jac/jaclang/scale/jserver/impl/jfast_api.impl.jac:1096-1101` - the request-context close commits inside a `finally` and swallows any exception into `logger.warning("Failed to clear request context: ...")`. The response has already been sent, so a conflict there silently drops the write. This is the path seen in production.
- The OCC retry that does exist (`jac/jaclang/runtimelib/impl/server.impl.jac:394-420`, 5 attempts) wraps the walker call only, which has already returned by the time the close-path commit runs.

## Impact

`system_root` is shared by every user of a deployment, so this is a system-wide contention point rather than a per-user one, and it worsens with replica count and concurrency.

Measured on a production deployment (8 replicas), the superroot anchor is at `version=571` with 151 edges:

| node type attached to superroot | count | expected |
|---|---|---|
| CommunityHub | 82 | 1 |
| StripeEventIndex | 68 | 1 |
| McpRegistry | 1 | 1 |

Each duplicate is one lost race. Downstream code that traverses from the root and takes the first match now picks a different copy depending on ordering, so state written through one copy is invisible through another. During a load event the same deployment logged 294 `WriteConflict` warnings on `000...0` in 15 minutes, all from the context-close path where they are swallowed and the write is lost.

## Workaround (verified)

None that is safe to apply from user code. Once the singleton exists the race stops, so serializing the first creation (a single bootstrap before traffic is admitted) avoids it, but that is not verified here as a general fix and does nothing about the swallowed close-path conflict. I did verify the negative case: with the hub already present, 400 concurrent requests produce no conflicts and no new hubs.

## Suggested fix

1. Make creation of a node anchored to the shared root atomic rather than check-then-write. A conditional insert or a unique-index-backed upsert in the Mongo backend would let exactly one writer win and let the others re-read the winner, instead of every loser materializing a duplicate.
2. Give the request-context close path bounded optimistic retry (re-read, reapply, commit) instead of `logger.warning`, and stop clearing the changeset before the conflict is raised (`memory_hierarchy.main.impl.jac:135-138`) so a retry has something to replay.
3. Consider not sharing one mutable superroot anchor object across concurrent contexts in a process (`context.impl.jac:19-29`); a per-context copy would keep one request's mutation out of another's changeset.

Regression test: spawn K concurrent requests through a find-or-create on `system_root` against a cold store and assert both that exactly one node was created and that no request failed with `WriteConflict`. The repro above is that test with K=40 and 3 processes; a single-process variant with threads should also show the duplication.

akindu-k:
Fixes #7605

## What was going wrong

When several requests hit the server at the same time and each one tried to "find or create" a singleton node hanging off the shared `system_root` (the all-zeros super root), they raced each other. Every request that lost the race still created its own copy of the "singleton" and attached it to the super root, so the graph ended up with many duplicates of a node that was supposed to exist exactly once. On top of that, the version check on the super root would reject the losers and the request came back as a `500 WriteConflict`. In production this had piled up 82 copies of one singleton and 68 of another on a single super root.

The real culprit was that **the whole process shared one single super root object across every in-flight request.** Each request context was handed the exact same `system_root` anchor (it was cached live in `_process_cache`). Because they all pointed at the same object, one request's edits were instantly visible to the others, which scrambled the per-request bookkeeping used to detect changes and to do the optimistic version check. That corruption is what let multiple duplicates get linked to the root and what defeated the retry logic that was supposed to handle exactly this situation.

image

## How we fixed it

The core idea: stop sharing one mutable super root between concurrent requests, and let the existing optimistic-concurrency retry actually do its job.

1. **Each request now gets its own copy of the super root.** We still keep a cached snapshot (the "template") so a new request doesn't have to hit the database every time, but that snapshot is detached, and every request works on its own independent copy. Now one request's in-progress edits can never leak into another's.

2. **After a super root write, the cached snapshot is refreshed from the committed store state, not from the writer's own copy.** Once a request persists a change to the super root, the template is rebuilt by re-reading the merged row straight from the store (the source of truth) under a lock. Doing the read and the publish together under one lock keeps refreshes in commit order, so the cached snapshot can never regress to an older or partial view: a later refresh always reflects later committed state. Because it comes from the store, the snapshot also carries whatever is durably committed on the super root, including access grants set on it (for example a `grant(system_root, ...)` in an app's startup entry) and its current set of edges, so the next request sees consistent permissions and topology and does not re-create the singleton. This refresh keys off what actually persisted, so even when the same commit later hits a conflict on a different write and has to retry, a super root edge that did land in the store still makes it into the cached snapshot rather than being skipped.

3. **A request that loses the race re-reads the super root and retries.** When the version check rejects a writer, the retry re-reads the super root fresh from the store. The walker runs again, sees the winner's node, and simply returns it instead of creating a duplicate. So losers converge on the one real node instead of failing with a 500.

4. **Rolling back a failed attempt no longer deletes the shared super root.** The rollback path used to delete any locally-modified node from the store. For a node created inside the failed attempt that is correct, but for a pre-existing shared node like the super root it would have wiped out another writer's committed data. Rollback now only removes nodes that were brand new in the aborted attempt.

Together these make concurrent find-or-create on the super root converge on exactly one node, with no failed requests and no duplicates left behind.

image

image

## Verification

Reproduced the bug with a walker that find-or-creates a `Hub` singleton on `system_root`, hit with 40 concurrent requests:

- **Before the fix:** 40 requests produced **8** hubs linked to the super root (should be 1).
- **After the fix:** 40 requests produced exactly **1** hub, with zero failed requests, and the super root keeps the access grant set at startup.

Added a regression test (`test_superroot_singleton_race.jac`) that runs this exact scenario against both the SQLite backend (no containers) and the Mongo + Redis backend, asserting that exactly one node is linked and that no request fails with a conflict.

## Follow-ups (not blocking, tracked separately to keep this PR focused on correctness)

- **Copy-on-write isolation.** Today each request deep-copies the super root at context init, even requests that never touch it. For a super root with large fan-out this is per-request overhead; isolating lazily (only when a request actually mutates the super root) would remove it.
- **`arefresh_system_root` placement.** The retry re-read helper lives on the base `ExecutionContext` in `runtimelib` while being specific to the scale super root path. It works (the super root UUID is a runtimelib constant), but a later cleanup could push the scale-specific bits down into the scale layer.
- **OCC retry re-executes the walker.** On a write conflict the whole walker runs again rather than replaying a change set. This is correct and idempotent for find-or-create, but any non-idempotent side effect in a walker body would re-run on retry. Noted here as existing behavior this fix relies on.


RandiJay96:
https://github.com/jaseci-labs/jacBuilder/pull/874

PR #7620, opened 2026-07-21
feat(jac-scale): add KEDA HTTP Add-on capability discovery and preflight diagnostics

### LINKED ISSUE #7404: feat(jac-scale): add KEDA HTTP Add-on capability discovery and preflight diagnostics
## Summary

Add explicit capability discovery and actionable preflight diagnostics for the KEDA HTTP Add-on.

The current `KEDAAutoscaler.preflight()` verifies only that the core `scaledobjects.keda.sh` API exists. That is sufficient for queue/metric scalers, but an HTTP-activated workload also depends on:

- `http.keda.sh/v1beta1 InterceptorRoute`
- the HTTP external scaler service
- the interceptor proxy service
- compatible KEDA core and HTTP Add-on versions
- traffic being routed through the interceptor

Without these checks, deployment can create a valid core `ScaledObject` that never activates from HTTP traffic.

## Proposed scope

Introduce structured capability discovery, either as an extension of autoscaler preflight or as a reusable Kubernetes add-on capability service.

Example result:

```jac
obj KEDACapabilities {
    has core_available: bool,
        http_addon_available: bool,
        interceptor_route_api_version: (str | None),
        external_scaler_address: (str | None),
        interceptor_service_address: (str | None),
        warnings: list[str] = [],
        errors: list[str] = [];
}
```

The capability check should be cached per cluster/context, with an explicit invalidation or refresh path.

## Required checks

- Discover the `keda.sh/v1alpha1 ScaledObject` API.
- Discover the `http.keda.sh/v1beta1 InterceptorRoute` API.
- Verify that the configured HTTP external scaler service resolves or exists.
- Verify that the configured interceptor proxy service exists.
- Distinguish missing KEDA core from missing HTTP Add-on.
- Detect RBAC-denied discovery separately from a genuinely absent API.
- Validate that the selected HTTP API version is supported by jac-scale.
- Return structured results for programmatic callers and concise actionable messages for CLI users.
- Provide the official Helm installation/upgrade command when the HTTP Add-on is absent.
- Do not auto-install cluster operators by default.

## Acceptance criteria

- [ ] Core-only clusters report KEDA core as available and HTTP activation as unavailable.
- [ ] Clusters with both components report HTTP activation as ready.
- [ ] Missing APIs, missing services, unsupported versions, and RBAC failures produce distinct diagnostics.
- [ ] Results are cached per cluster identity and can be refreshed.
- [ ] HTTP activation deployment refuses to proceed when required capabilities are unavailable instead of silently producing a nonfunctional workload.
- [ ] Tests cover core-only, full-install, missing-service, unsupported-version, and 403/RBAC paths.
- [ ] Documentation shows the supported Helm install/upgrade flow.

## Out of scope

- Automatically installing or upgrading KEDA.
- Application ingress creation.
- Application authentication and authorization policy.
- General-purpose health checking of every KEDA scaler type.

## References

- Existing KEDA preflight: #6046 / #6220 / #6534
- HTTP activation resources: #7403
- KEDA HTTP Add-on installation: https://keda.sh/http-add-on/0.15/operations/installation/
- JacHammer preview use case: https://github.com/jaseci-labs/jacBuilder/issues/603

Developer-Linus:
## Summary

Closes #7404.

`KEDAAutoscaler.preflight()` only ever verified that the core `keda.sh/v1alpha1 ScaledObject` API existed on a cluster. That check is not sufficient for HTTP activated workloads, which also depend on the KEDA HTTP Add-on's `InterceptorRoute` API, two Kubernetes Services (the external scaler and the interceptor proxy), and a compatible Add-on version. Without these checks, `apply_http_activation` could create a valid `ScaledObject` that never actually activates from HTTP traffic, with no clear diagnostic explaining why.

This PR introduces structured capability discovery for the KEDA HTTP Add-on and wires it into `apply_http_activation`, so a broken or partially installed cluster fails with an actionable error instead of silently deploying a nonfunctional workload.

## Changes

| File | Changes |
|---|---|
| `jac/jaclang/scale/deploy/autoscale/keda_capabilities.jac` (new file) | Adds the `KEDACapabilities` data object, matching the shape specified in the issue: `core_available`, `http_addon_available`, `interceptor_route_api_version`, `external_scaler_address`, `interceptor_service_address`, `warnings`, and `errors`. Also adds `SUPPORTED_HTTP_INTERCEPTOR_API_VERSIONS`. |
| `jac/jaclang/scale/deploy/autoscale/keda_autoscaler.jac` | Imports `KEDACapabilities`.<br>Adds a new `interceptor_service_address` field, defaulted the same way `http_scaler_address` already is.<br>Replaces the old bool only `_http_preflight_cache` with `_capabilities_cache`, which holds a full `KEDACapabilities` object per cluster.<br>Declares `discover_capabilities`, `invalidate_capabilities`, `_check_service_address`, `format_capabilities`, and `_capabilities_cache_key`. |
| `jac/jaclang/scale/deploy/autoscale/keda_autoscaler.impl.jac` | `discover_capabilities` checks the core `ScaledObject` API and the HTTP Add-on's `InterceptorRoute` API independently, so a missing core install is distinguished from a missing Add-on. A 403 on either check is reported as a distinct permission error rather than being folded into "not installed."<br>If the `v1beta1 InterceptorRoute` API is absent (not RBAC denied), it additionally probes the legacy `v1alpha1 HTTPScaledObject` API. If that is present, the result is a distinct "upgrade" warning with the exact `helm upgrade` command, rather than being reported as "Add-on not installed." A 403 on this legacy probe is reported as its own permission error rather than being folded into "not installed."<br>`_check_service_address` verifies the external scaler and interceptor proxy Services actually resolve, and reports which one is missing.<br>Results are cached per cluster identity via a new `_capabilities_cache_key` helper, which combines `_get_cluster_key()` with this instance's configured `http_scaler_address` and `interceptor_service_address`. This keeps two instances that target the same cluster but are configured with different service addresses from reusing an incompatible cached result. `discover_capabilities(refresh=True)` and `invalidate_capabilities()` remain the explicit refresh paths.<br>`format_capabilities` produces a single concise, human readable line for CLI output.<br>`preflight()` now also catches a 403 and returns a distinct permission error instead of letting the `ApiException` propagate unhandled, which is what happened previously.<br>`apply_http_activation` now calls `discover_capabilities()` instead of its own inline, duplicate check. If discovery reports any errors (missing service, RBAC denied, or similar), it raises `ValueError` with the details. If the core KEDA API or the HTTP Add-on is simply absent (no errors, just not installed), it keeps the existing behavior of logging a warning and returning `False` rather than raising. A partial install, where the Add-on and its services are healthy but core KEDA itself is missing, now also falls into this gate instead of proceeding to create resources that would fail against a nonexistent API. |
| `jac/jaclang/scale/tests/deploy/test_keda_http_activation.jac` | Adds a new capability discovery test section covering: core only clusters, full installs, missing external scaler service, missing interceptor service, legacy Add-on detection, RBAC denied on the core check, the HTTP Add-on check, and the legacy probe, cache and refresh behavior, cache isolation between instances with different service addresses, explicit invalidation, the raise on broken capabilities and skip on missing core behavior in `apply_http_activation`, `format_capabilities` output, and the `preflight()` 403 fix. Test setup uses small composable mock builders (`_custom_api_for`, `_core_api_for`) rather than one mock per scenario, and one table driven test covers the distinct diagnostic cases to avoid duplicating near identical assertions. |

## Out of scope

As noted in the issue: no auto install or upgrade of KEDA, no ingress creation, no general health checking of every scaler type. This PR also does not verify that a workload's traffic is actually routed through the interceptor at runtime, only that the interceptor's own API and Service objects exist and resolve. That distinction is called out explicitly in the implementation plan for this issue.

## Test plan

- [x] Core only clusters report core available and HTTP unavailable.
- [x] Fully installed clusters report HTTP ready with no errors.
- [x] Missing external scaler service, missing interceptor service, legacy Add-on version, and RBAC denied cases (core check, HTTP Add-on check, and legacy probe) each produce a distinct diagnostic.
- [x] Results are cached per cluster identity, including the configured service addresses, and refresh via `discover_capabilities(refresh=True)` and `invalidate_capabilities()`.
- [x] `apply_http_activation` raises when capabilities are broken, returns `False` gracefully when the Add-on or core KEDA is simply absent, and does not proceed to create resources on a partial install.
- [x] Documentation update showing the supported Helm install and upgrade flow.
- [x] Full deploy test suite (169 tests) passes with no regressions.


kugesan1105:
image
image

PR #7628, opened 2026-07-21
test(scale): testing-trophy migration - honest tests + real infra (no hand-rolled fakes)

MusabMahmoodh:
## Summary

Hardens the jac-scale test suite so tests fail for real reasons. Hand-rolled fakes and mocks of the system's own components are replaced with real infrastructure, a few genuinely-untested modules get coverage, and the real Kubernetes end-to-end suite now runs on scale changes.

## What changed

**Remove tests that could not fail**
- Deleted tests that passed while asserting nothing (a swallowed `pytest.fail`, silent skips on missing infra, `assert True`, a swallowed `except` body) and removed the dead code around them.
- Removed "proof" tests that asserted a bug still exists - they invert the normal contract and break the moment the bug is fixed.
- Wired two tests that existed but no CI job ran (password hashing, firebase env) into the suite.

**Replace fakes with real infrastructure**
- Redis, S3 (LocalStack), GCS (fake-gcs-server), and Firestore (emulator) now run against real containers instead of hand-rolled fakes or in-process mocks; Mongo-backed paths use a real container.
- SSO and firebase auth run against a real `UserManager` and store - only the external identity provider is stubbed at the network boundary.
- WebSocket hardening runs against a real `UserManager`. This exposed that the token-swap defense was never actually exercised (the fake short-circuited before reaching it); it now is.
- Scheduler, manifest-builder, and plan tests drive the real code instead of patching their own internal methods.

**Fill coverage gaps**
- Added real tests for the state-backend factory, the service-URL builder, the k8s ops helpers, SSE framing, and the WebSocket forward proxy.

**End-to-end**
- The Kubernetes e2e job now runs on scale-only changes (it was gated to core changes only).
- Added an e2e journey that verifies an order survives a service-pod restart: state persists in Mongo and the gateway-issued token is accepted by a downstream service.

## Testing

Converted suites pass locally against real containers (Redis, Mongo, LocalStack, fake-gcs-server, the Firestore emulator). The e2e journey was validated on a local kind cluster.

## Notes

Making these tests real surfaced two pre-existing issues: a rate-limit config key mismatch between the loader and the plan renderer (filed separately as #7635), and the WebSocket identity-mismatch defense that had no coverage (now covered).


MusabMahmoodh:
Validation (matching jac binary, local + EC2 with docker):

- Directly-edited tests pass: `test_memory_tracing`, `test_async_drivers` (SkipTest conversions, real Mongo/Redis testcontainers), `test_metrics`, `test_k8s_config_tracing`, both CI-wired root tests (`test_identity_password`, `test_firebase_env`), and the rewritten read-only `deployment target methods` assertion against a real EKS context.
- `test_gateway`: 64 passed, 3 failed. The 3 failures (`/metrics` tests I did not touch) reproduce identically on **untouched upstream/main** (3 failed, 66 passed), root cause `no such column: identity_users.profile` - a pre-existing local binary/source schema mismatch, not this change.
- The EC2 rc=1s on `test_serve`/`test_webhook` were environmental: a full disk (Mongo testcontainer OutOfDiskSpace) and a stale 0.32 binary vs newer source (sse.jac compiler-pass error). Not caused by the edits.

Net: the honesty edits add no new failures.

MusabMahmoodh:
@greptile review

The one blocker in the previous summary (Confidence 4/5) - "an accepted WebSocket that closes before sending a frame still satisfies the rejection checks" - is fixed in 5e8a6e475. The WS-rejection tests now assert the connection did **not** succeed AND that the failure carries a 4xx handshake status (`InvalidStatus.response.status_code`), so an accepted-then-closed connection or a timeout no longer counts as a rejection. Please re-review the current HEAD.

PR #7643, opened 2026-07-21
feat(client): derive route membership from exports, not filename skip-suffix (#7640)

### LINKED ISSUE #7640: Route scanner uses `.cl.` skip-suffix, blocking fully-markerless co-location in pages/
## Context

Follow-up from #7639 (shadcn + skill-docs markerless migration).

The markerless migration lets client placement be inferred, so `.cl.jac` is no longer needed to mark client code. But the file-based route scanner still relies on the `.cl.` filename infix as its "this file is not a route" signal, so co-located components in a `pages/` directory cannot go fully markerless.

## The gap

`jaclang/runtimelib/client/route_scanner.jac:20`:

```
SKIP_SUFFIXES = ['.cl.', '.impl.', '.test.'];
```

`jaclang/runtimelib/client/impl/route_scanner.impl.jac:57` skips any entry in a `pages/` dir whose name contains one of those substrings; every other plain `.jac` becomes a route.

Consequence: a component co-located beside route pages in `pages/` must be named `*.cl.jac` (or `.impl.` / `.test.`) or the scanner auto-registers it as a broken route. `.cl.` is the only standalone-importable skip token (`.impl.` alone is not importable; `.test.` carries test semantics), so it cannot simply be renamed to a plain `.jac`.

This is the one spot where the markerless migration is incomplete: in #7639, `jac-cl-routing.md` intentionally keeps `.cl.` and reframes it as a route-scanner skip signal (placement is still inferred either way) rather than removing it.

## Options

- Teach the scanner a markerless skip convention, for example:
  - skip files/dirs with a leading underscore (`_budget_ui.jac`, `_components/`), Next-style;
  - or require route files to be named `page.jac` / `index.jac` and treat everything else as non-route;
  - or a `components/` subdir convention under `pages/`.
- Whichever is chosen, update `SKIP_SUFFIXES` and the `jac-cl-routing.md` co-location guidance together.

No behavior change was made in #7639; this issue tracks the design decision.

marsninja:
## Summary

Closes #7640. The file-based route scanner decided whether a `pages/` file was a route from a filename skip-suffix list (`.cl.`/`.impl.`/`.test.`). Markerless codespace inference retired `.cl.` as a placement marker, but the scanner still leaned on it as the "not a route" signal, so a markerless co-located component in `pages/` was auto-registered as a broken route (its generated `import { page as ... }` resolved to nothing).

## Root cause

A route is definitionally a module that exports `page` (a layout `layout`) : that is the exact contract `_generate_routes_manifest` imports (`import { page as ... }` / `import { layout as ... }`). Keying route-ness off the *filename* instead of that *export set* is the mismatch; the skip-suffix list was a placement-marker proxy for a signal the module already carries. The three options in the issue (leading underscore, `page.jac`-only, `components/` subdir) all just swap one filename marker for another, so none of them close the gap.

## Change

- `RouteScanner` takes an injected `exports_of` resolver. A file is a route iff it exports `page`, a layout iff it exports `layout`. `JacClientCompiler` backs the resolver with `compile_module(...).exports`, which is hub-cached (`_compile_to_js` resolves from `program.mod.hub`) so `pages/` files are not compiled twice.
- `.cl.` leaves the skip list. Legacy `.cl.jac` components stay out of the route table because they export a non-`page` symbol; markerless co-located components now need no marker at all.
- `.impl.`/`.test.` remain structural skips (module roles, not standalone route modules) : filename-cheap, never compiled.
- With no resolver injected the scanner falls back to the previous filename behavior, so standalone construction is unchanged.

## Tests

`tests/runtimelib/client/test_route_scanner.jac` (new, 5 cases): the markerless co-located component is not a route, a layout registers as a layout, a legacy `.cl.jac` is skipped via its exports, `.impl.`/`.test.` sidecars are skipped, and dynamic / index / catch-all path derivation still resolves. Red before the change (route membership was filename-driven, so the plain-`.jac` component was registered as `/budget_ui`), green after.

`jac check` on `jac_client_compiler.jac` reports the same pre-existing type-noise as `main` (0 new errors).

Doc: `jac-cl-routing.md` co-location guidance rewritten to the export model.


marsninja:
Addressed the Greptile P1 and the resulting `test-runtime` failure (same root cause).

Route membership was keyed on the client compiler's export list, which is `:pub`-gated. The `test_hmr_client` pages are written as `cl { def page() {...} }` (non-pub), so `page` was absent from the export list and every route was dropped, leaving no `_routes.js`.

Changes since the first push:
- Route membership now reads the compiled module's **symbol table** (`sym_tab.names_in_scope`), which carries `page`/`layout` whether the def is `:pub` or a plain `def` inside `cl {}`. (Diagnostic: `cl { def page }` yields symtab `['page']` but exports `[]`.)
- **Greptile P1** (silent route drop on compile error): the resolver returns `None` when a page cannot be resolved, and the scanner keeps it as a route (fail-safe), so the real compile still surfaces the error and HMR never removes an edited route.
- Renamed `exports_of`/`_page_exports` → `symbols_of`/`_page_symbols` to match the signal.
- Added regression tests: a non-pub `cl {}` page is a route, and an unresolvable page stays a route.

Local: the 5 routing `test_hmr_client` cases that failed in CI now pass; `test_route_scanner` 7/7. (The 10 server-HMR failures seen locally are a pre-existing subprocess/env issue that reproduces identically on `main` and passed in CI.)

PR #7665, opened 2026-07-22
fix(byllm): log handled tool-call failures at WARNING, not ERROR with traceback

### LINKED ISSUE #7657: byllm: Tool.__call__ logs ERROR+traceback for every handled tool-call failure (log flood under retry loops)
## Summary

`Tool.__call__` in byLLM catches a failing tool call, returns the error to the LLM as a recoverable `"Tool error: ..."` observation (correct), but ALSO logs a full `ERROR` traceback via `logger.exception("Tool execution failed")` on every occurrence, with no dedup and no rate limit. A tool call failing and being handed back to the model is an expected, handled control-flow outcome, not an exceptional error. When a model repeatedly emits the same malformed tool call (e.g. a required-arg-less call), this floods the logs with identical `ERROR` tracebacks and buries real errors. This is the same class of defect as jaseci-labs/jac#7602 (wrong log level and cardinality for a supported outcome), on a different code path.

## Environment

| | |
|---|---|
| Binary | `jac 0.34.2 (Darwin arm64)`, released asset `jac-0.34.2-macos-aarch64` from `v0.34.2`; also reproduces on a `main` build (`afd43a9bc`) |
| Source | `jaclang/byllm/types.impl/tool.impl.jac`, `impl Tool.__call__`, unchanged on `main` |
| Also observed | prod EKS `jachammer`, service `jac-coder-sv` on `jac 0.34.1`, 2026-07-22 UAT: 52 identical tracebacks in ~3 min on one pod, recurring across multiple pods and multiple models (gpt-5.2, claude-sonnet-4-6) |

`repro_tool_log.jac`:

```jac
import logging;
import from jaclang.byllm.lib { Tool }

"""Write code to a file."""
def write_code(file_path: str, content: str) -> str {
    return f"wrote {len(content)} bytes to {file_path}";
}

with entry {
    logging.basicConfig(level=logging.INFO);
    t = Tool(func=write_code);
    # Simulate a model emitting write_code with missing args, repeatedly.
    # (In prod a confused model did this 52 times in ~3 min on one pod.)
    for i in range(200) {
        result = t();  # missing file_path, content -> handled, returned to model
        print(f"[caller] iteration {i}: returned -> {result!r}");
    }
}
```

Run against the released binary and count what 200 handled failures emit:

```
$ ./jac-0.34.2-macos-aarch64 run repro_tool_log.jac > out.txt 2>&1

$ grep -c 'returned ->'                    out.txt   # handled tool calls (input)
200
$ grep -c 'Tool execution failed'          out.txt   # ERROR log lines
200
$ grep -c 'Traceback (most recent call last)' out.txt # full tracebacks
200
$ grep 'Tool execution failed' out.txt | sort -u | wc -l   # distinct messages
1
$ wc -l < out.txt                                     # total log lines
1000
```

Each ERROR line looks like:

```
ERROR - Tool execution failed
  File ".../site/jaclang/byllm/types.impl/tool.impl.jac", line 30, in __call__
TypeError: write_code() missing 2 required positional arguments: 'file_path' and 'content'
```

The logging is exactly 1:1 with the number of calls and unbounded: **200 recoverable failures produce 200 `ERROR` tracebacks (1000 log lines), all one identical message (zero dedup / rate limit).** The `returned ->` count confirms every one was handled and returned to the model as a `"Tool error: ..."` observation, i.e. none of these were unrecoverable.

## Expected

A caught tool-call failure that is deliberately returned to the model as a recoverable observation should be logged at `WARNING` (or `DEBUG`), message only, and ideally deduped/rate-limited per `(tool, error)`. It should not emit an `ERROR`-level stack trace per occurrence on the hot path.

## Actual

`impl Tool.__call__` (`jaclang/byllm/types.impl/tool.impl.jac`):

```jac
try {
    return self.func(*args, **kwargs);
} except Exception as e {
    # For the LLM if the tool failed, it'll see the error message
    logger.exception("Tool execution failed");   # line 38: ERROR + full traceback, every time
    return f"Tool error: {e}";                    # line 39: handled, returned to the model
}
```

The comment on the next line shows the failure is intentionally recoverable, yet `logger.exception` logs it at `ERROR` with a traceback unconditionally. In prod this reached 52 identical tracebacks in ~3 min on a single `jac-coder-sv` pod during the 2026-07-22 UAT, burying real errors in the log stream.

## Impact

- Log flood during exactly the moments operators are triaging: a confused model that repeats a malformed tool call turns one recoverable event into dozens of `ERROR` tracebacks per minute per pod, across pods.
- Buries genuine errors and inflates log ingestion/storage.
- Mirrors jaseci-labs/jac#7602, observed together with it during the same UAT (see jaseci-labs/jacBuilder#689).

Note: the user-facing symptom (the agent hot-looping on the same failing tool call) is a separate concern in the agent loop that drives the retries (bounding identical repeated failing tool calls); that lives outside byLLM. This issue is scoped to the logging level/cardinality in byLLM.

## Suggested fix

- In `Tool.__call__`, log the caught-and-returned tool error at `WARNING` (message only, e.g. `logger.warning("Tool %s failed: %s", self.get_name(), e)`), not `logger.exception` at `ERROR`. Optionally dedup/rate-limit per `(tool name, error text)` within a turn.
- Regression test: assert that a tool invoked with invalid args (a) returns the `"Tool error: ..."` string and (b) does not emit an `ERROR`-level record (caplog at `ERROR` is empty; a single `WARNING` is present).

## Related

- jaseci-labs/jac#7602 (same class: wrong log level/cardinality for a supported outcome; runtime deserializer)
- jaseci-labs/jacBuilder#689 (both floods observed together under prod load)

Hirudika4:
Fixes #7657

## Summary

`Tool.__call__` catches a failing tool call and hands the error back to the model as a recoverable `Tool error: ...` observation (correct), but it also logged every occurrence with `logger.exception("Tool execution failed")`. That records at ERROR level with a full traceback, once per call, with no dedup or rate limit. A tool call failing and being returned to the model is an expected, handled control-flow outcome, not an exceptional error. When a model repeats the same malformed tool call (e.g. an argument-less call), this floods the logs with identical ERROR tracebacks and buries real errors. Same defect class as #7602 (wrong log level/cardinality for a supported outcome), on a different code path.

Observed in prod: 52 identical tracebacks in ~3 min on one `jac-coder-sv` pod during the 2026-07-22 UAT, recurring across pods and models (gpt-5.2, claude-sonnet-4-6).

## Root cause

`jaclang/byllm/types.impl/tool.impl.jac`, `impl Tool.__call__`:

```jac
} except Exception as e {
    logger.exception("Tool execution failed");   # ERROR + full traceback, every time
    return f"Tool error: {e}";                    # handled, returned to the model
}
```

The exception is intentionally recoverable (returned to the model on the very next line), yet `logger.exception` logs it at ERROR with a traceback unconditionally.

## Fix

Log the caught-and-returned tool error at WARNING, message only, naming the tool:

```jac
logger.warning("Tool %s failed: %s", self.get_name(), e);
```

The observation returned to the model is byte-for-byte unchanged, so the ReAct loop, recovery paths, and every caller behave identically. Only the log level and format change.

## Behavior (issue repro: 200 handled tool failures)

|                               | before | after |
|-------------------------------|:------:|:-----:|
| handled and returned to model | 200    | 200   |
| ERROR log lines               | 200    | 0     |
| full tracebacks               | 200    | 0     |
| WARNING lines (message only)  | 0      | 200   |

## Test

Adds a regression test in `jaclang/byllm/tests/test_byllm.jac`: a tool invoked with invalid args

- returns the `Tool error: ...` string,
- emits no ERROR-level record and no traceback,
- logs exactly one WARNING naming the tool.


MalithaPrabhashana:
Adding context on the two commits I pushed here, because the root cause turned out to be different from what #7657 (and my own first commit message) says.

## The issue is not a confused model

#7657 attributes the flood to "a model repeatedly emitting the same malformed tool call". Instrumenting byLLM's dispatch path during a real JacCoder build shows the model's calls are fine — byLLM is corrupting them.

When tool calls are issued in parallel, their streamed argument fragments get spliced together during reconstruction. One call's arguments run into another's:

```
', "background": false}{"command": "mkdir -p /home/...
 └─ tail of call A ────┘└─ head of call B ───────────
```

The result is unparseable JSON. `_sanitize_tool_calls` then replaced it with `"{}"` and recorded nothing, and `"{}"` is indistinguishable from a genuine no-argument call, so the tool was invoked bare and the model was told:

```
write_code() missing 2 required positional arguments: 'file_path' and 'content'
```

which is false — it sent both. With nothing to correct it re-sends the same call, which is the loop that produces the identical-traceback flood this PR quietens.

`finish_reason` for a spliced batch is `tool_calls`, never `length`, so the existing truncation guard could not report these.

## Evidence

One JacCoder build (Sonnet 4.6), tapping every tool call:

| | tool calls | corruptions | reached the tool with arguments erased |
|---|---|---|---|
| this branch, before my commits | - | 1 | **1** |
| with the fix | 85 | **9 (10.6%)** | **0** |

Specimens from the fixed run:

```
run_command     230 chars   (spliced: tail of one call + head of the next)
write_code      107 chars   components/Landing.cl.jac
write_code      107 chars   components/Landing.cl.jac   <- same cut point, twice
write_code   10,221 chars   components/Courses.cl.jac
write_code   11,010 chars   components/Quiz.cl.jac
write_code   11,618 chars   components/Quiz.cl.jac
edit_code       406 chars   components/Exam.cl.jac
write_code   16,833 chars   components/Profile.cl.jac
write_code   13,593 chars   components/Lecturer.cl.jac
```

It affects every tool that carries arguments, not just `write_code`, and it scales
with payload size: the 10-17 KB components are hit, small calls are not.

Across that one 28 minute build the splice destroyed 64,125 characters of
correctly generated code (~16k output tokens) and forced nine extra model
round-trips. The build still completed - 40 `.jac` files, server booting and
answering `/healthz` - because every corruption came back as something the model
could act on.

`Landing.cl.jac` took three attempts: corrupted, corrupted, then written (25 KB on disk). Under the current code each of those attempts would have come back as "you omitted required arguments" instead of something the model could act on.

Replaying captured payloads through byLLM directly: **0/3 reported before, 3/3 after.**

## What the commits change

`_sanitize_tool_calls` now returns the ids whose arguments it discarded, and the three dispatch paths forward them to `_parse_tool_calls` as a separate set from `truncated_ids`, so each cause gets advice that fits it:

- genuine `max_tokens` truncation -> "retry with a smaller payload, split the work" (unchanged)
- corrupted in transit -> "the arguments you chose were not the problem and the payload was not too large: send the same call again unchanged"

Routing corruption into the truncation message would tell the agent to fragment its files for no reason, so the two are kept distinct.

A call carrying no argument text at all is not treated as damage, so argument-less tools are unaffected.

Tests cover both directions through `_parse_tool_calls`: a truncated call still
gets the split-the-payload advice, a damaged one gets re-send-unchanged and must
not get the split advice. Verified to fail when the damaged branch is pointed
back at the truncation message, so re-uniting the two sets cannot pass silently.

## Scope

This makes the failure recoverable; it does not stop the splice. Every occurrence still costs a wasted tool call and a model round-trip - nine in the build above. The splice itself lives in streaming delta accumulation (litellm's `stream_chunk_builder` or byLLM's chunk collection); I have not traced which, and it deserves its own issue. The two 107-char cuts at an identical position suggest it is deterministic rather than random stream noise, which should help narrow it.

Two things worth deciding before this merges:

1. #7657's "confused model" framing is wrong and should be corrected, so the splice still gets chased after this lands.
2. The existing release-note bullet repeats that attribution ("A confused model repeating the same malformed tool call") and would ship inaccurate history.

Issue #7606, opened 2026-07-20
jac-scale: embedder-supplied microservices_config bypasses schema defaults, so microservice deploys silently ship no client


udithishanka:
## Summary

`MicroserviceManifestBuilder._ms_cfg()` treats a caller-supplied `microservices_config` as
authoritative and returns it verbatim, skipping the schema defaults that
`get_scale_config().get_microservices_config()` would have filled in. One of those defaults
is `client.entry = "main.jac"`.

So an embedder that passes its own config, as jacBuilder does, loses the client entry and
every other defaulted key. Client detection then falls through to a `**/*.cl.jac` glob,
which misses any app whose client is written in plain `.jac`. `_build_client_bundle()`
returns without building, without logging, and without failing, and the fleet deploys as a
headless API with `/` returning a JSON 404 while the deploy reports success.

This is also why the bug has gone unnoticed: on the `jac start --scale` path the
caller-supplied config is empty, the ambient defaults apply, `client.entry` is always
populated, and **the `.cl.jac` glob is never reached**. The glob only ever executes on the
embedded path.

## Environment

| Item | Value |
|---|---|
| jac | **0.34.0** (verified), also reproduced on 0.32.0; logic identical on `main` |
| Code | `jac/jaclang/scale/deploy/target/kubernetes/microservice/manifest_builder.jac` (line numbers below are `main`) |
| Seen in production | jachammer (`production-cluster`, us-east-2), app `this-is-jac`, pods on `jaseci/jaclang:dev` |
| Embedder | jacBuilder `deploy_manager`, which passes `microservices_config=` parsed from the app's own `jac.toml` |

## Reproduction (executed on 0.34.0)

Two app layouts differing only in how the client is spelled:

```bash
mkdir -p clirepro/app_clj clirepro/app_plain

# A: client written as .cl.jac
printf 'def:pub App() -> any {\n    return <div>hi</div>;\n}\n' > clirepro/app_clj/frontend.cl.jac
printf '[project]\nname="a"\nentry-point="main.jac"\n' > clirepro/app_clj/jac.toml

# B: identical client written as plain .jac, declared via [client.*]
printf 'def:pub App() -> any {\n    return <div>hi</div>;\n}\n' > clirepro/app_plain/frontend.jac
[...]

PR #7668, opened 2026-07-22
fix(scale): gateway pod client-build fallback uses --client web, skipping the whole-project typecheck gate

### LINKED ISSUE #7667: jac-scale: in-pod client build runs the strict whole-project type-check gate, so type-error apps deploy green but serve JSON 404 at /
## Summary

Now that #7613 makes microservice deploys actually build the client (fixing #7606), the in-pod client build hits the next wall: the gateway pod runs plain `jac build "<client_entry>"`, and `jac build` hard-fails on the **whole-project type-check gate** before building anything. Any app with type-check errors (including the pervasive `.cl.jac` Any-typed false-positive class) deploys "successfully" but serves **API-only**: `/` returns the gateway's JSON 404 while `/docs` and every `/function/*` route work. The failure is a one-line `echo` in the pod log; the deploy itself reports success.

The dev-time gate is fine. The problem is running that same strict gate **inside the deploy pod as a packaging step**, with no way for an embedder to relax it: post-fold `AppConfig` has no typecheck knob, and the entrypoint template does not pass `--no-typecheck`.

## Environment

- jac 0.34.2 binary in the gateway pod (`jaseci/jaclang` image, binary-ship), jaseci-cluster (EKS us-east-2), namespace `jb-e0de7c76-imported-app`
- Deployed by jacBuilder `deploy_manager` via `KubernetesMicroserviceTarget` (embedder-supplied `microservices_config`, i.e. the #7606/#7613 path)

## Reproduction (executed, live deploy)

1. Deploy a fullstack microservice app whose project has type-check errors (the app: 5 `sv` services + a `.cl.jac` client with the usual Any-typed noise) through jac-scale's `KubernetesMicroserviceTarget`.
2. Deploy reports success; all 5 service deployments and the gateway go 1/1 Ready.
3. Gateway pod log (verbatim):

```
/app/components/ui/tabs.jac - 0 errors, 17 warnings
/app/components/ui/textarea.jac - 0 errors, 3 warnings
/app/components/ui/tooltip.jac - 0 errors, 22 warnings
...
✖ Error: error[E1001]: Cannot assign <Unknown> to float
✖ Error: error[E1055]: No matching overload found for method "__add__" with the given arguments
✖ Error: error[E1022]: Type <any> is not iterable (no __iter__ method)
...
✖ Error: Build gate failed: the project does not type-check.
[jac-scale] client build failed; / will 404
INFO - Registered route: analytics -> /analytics
INFO - Registered route: guestbook -> /guestbook
...
INFO - Starting gateway-only mode on 0.0.0.0:8000
```

4. Result, verbatim:

```
$ curl -s https://thisisjac-test.jaseci.app/
{"ok":false,"error":{"code":"NOT_FOUND","message":"Not found: /", ...},"meta":{"http_status":404}}
$ curl -s -o /dev/null -w "%{http_code}" https://thisisjac-test.jaseci.app/docs
200
```

(30 client components were type-checked in the pod, i.e. #7613's client detection works; the build then dies at the gate.)

## Root cause

Two pieces:

1. `jac/jaclang/scale/deploy/target/kubernetes/microservice/manifest_builder.jac` `_pod_entrypoint()` (~line 770):

```
f'if [ ! -f .jac/client/dist/index.html ]; then '
f'jac build "{client_entry}" '
f'|| echo "[jac-scale] client build failed; / will 404"; '
f'fi; '
```

   No `--no-typecheck`, and the `||` swallows the failure into an echo, so the deploy still reports success.

2. `jac/jaclang/cli/commands/impl/build.impl.jac` (~line 38): unless `no_typecheck`, `jac build` runs `check(paths=[project_root], ...)` over the **whole project** and returns non-zero on any `error[Exxxx]`.

So the pod's client build is gated on the entire app type-checking cleanly. There is no escape hatch at this layer: post-fold `AppConfig` is `{code_folder, file_name, app_name, testing, no_image, dev}` (no typecheck field), and the entrypoint string is not configurable by the embedder.

## Expected

Deploy-time client bundling is packaging, not validation: the code was already accepted for deploy. The pod build should produce the client bundle regardless of type-check noise (or at minimum the embedder must be able to choose that), and a client-build failure should be loud, not an `echo` on a "successful" deploy.

## Actual

Strict whole-project gate inside the pod -> client silently skipped -> app serves JSON 404 at `/` while the deploy is green.

## Impact

Every jacBuilder user-app deploy whose project carries any type error ships without its frontend. `.cl.jac` apps routinely carry Any-typed E1001/E1055/E1022 diagnostics that do not affect the emitted bundle (jacBuilder's own production files produce 1000+ of them and ship fine), so this is the common case, not the corner case. This is the direct successor to #7606: #7613 got the client build to run; this gate stops it from succeeding.

## Workaround (verified)

None at the config layer. jacBuilder is temporarily patching the gateway Deployment's container args post-deploy to insert `--no-typecheck` into the `jac build` invocation (k8s API patch + gateway restart). Works, but it is an embedder monkey-patch of an internal command string.

## Suggested fix

1. In `_pod_entrypoint()`, pass `--no-typecheck` to the in-pod client build (packaging step; the dev-time `jac build` gate is unchanged). Optionally make it a `client.typecheck` key in the microservices config schema (default lenient in-pod) for embedders that want the strict gate.
2. Make the failure loud: surface a client-build failure into the deploy result / readiness instead of `|| echo`, so a deploy without its client cannot report success silently.
3. Regression test: manifest render includes `--no-typecheck` in the gateway entrypoint (extend `tests/deploy/test_deployment_overlay.jac`, which #7613 already touched), plus an e2e that a type-error-bearing fullstack app still serves `/`.

Refs: #7606 (client silently not built), #7613 (fixed detection; exposed this gate).

udithishanka:
Fixes #7667. Follow-up to #7606 / #7613: now that the client is detected and built, the in-pod fallback build dies on the whole-project type-check gate for any app with type errors, so the app deploys green but serves the JSON NOT_FOUND envelope at `/` (observed live: 5-service fullstack app, all pods 1/1, gateway logs `Build gate failed: the project does not type-check.` then `[jac-scale] client build failed; / will 404`).

## Root cause
`_pod_entrypoint()` rendered `jac build "<entry>" || echo ...`. A bare `jac build` runs `check(paths=[project_root])` first (`cli/commands/impl/build.impl.jac`) and returns non-zero on any `error[Exxxx]`, and the `||` swallows that into an echo on a green deploy. The host-side `_build_client_bundle` already avoids this by building with `--client web`, which short-circuits before the gate; the pod fallback did not.

## Fix
The fallback now renders `jac build "<entry>" --client web`: gate-free by construction, builds only the client bundle (faster pod boot), and identical semantics to the host-side primary path. The dev-time `jac build` gate is unchanged.

## Tests
Two new tests in `scale/tests/test_pod_env.jac` (the manifest-builder suite): the rendered gateway entrypoint must contain `jac build "main.jac" --client web` plus the dist-exists guard, and must not contain the bare gated form; no client entry renders no build. Verified: **1 failed on unfixed main, 7 passed with the fix**.

## Notes
- The `|| echo` still hides a genuine client-build failure from the deploy result; making that loud is left as a follow-up (tracked in #7667).
- Embedder context: jacBuilder builder pods have no bun, so the host-side build falls through to this pod fallback routinely; with this fix the fallback actually works.


udithishanka:
End-to-end verification, run against a real cluster through the programmatic deploy API (the jacBuilder embedding path), using the editable dev loop so the PR code drove the deploy:

**Setup**: the `k8s_e2e` fixture (fullstack shop: client + 3 `sv` services) with a deliberate type error injected; `jac check` confirmed failing (the #7667 precondition). Deployed via `KubernetesMicroserviceTarget.deploy(AppConfig(...))` from a builder pod, scratch namespace `jac-fixverify-7668`.

**Results**:
1. Rendered gateway Deployment args contain the fix verbatim: `jac build "main.jac" --client web`.
2. Host-side path: client bundle built on the type-error app (2.7s Mac, 8.4s builder pod).
3. In-pod fallback path: the exact entrypoint command executed inside the deployed gateway pod on the type-error app: `Client bundle built (1.9s)`, `dist/index.html` produced.
4. `GET /` on the deployed app returns **200 text/html** (the client SPA shell) instead of the JSON `NOT_FOUND` envelope from the incident. Verified via port-forward and publicly through the shared ALB.
5. Unit tests: 1 failed on unfixed main, 7 passed with the fix.

One additional data point for the issue: the host-side `_build_client_bundle` works on jacBuilder builder pods after all (the jac client tooling bootstraps bun on demand), so the pod fallback is the safety net for cold/toolchain-less environments, and with this fix it actually functions.

PR #7670, opened 2026-07-22
fix(compiler): resolve entry-module relative imports at compile time

### LINKED ISSUE #7661: Relative import in entry-point file fails at runtime with "attempted relative import with no known parent package"
## Bug: relative import in the entry-point file fails at runtime with "attempted relative import with no known parent package", even though `jac check` reports no errors

**Jac version:- 0.31.1**

### Summary

A `.jac` file that is the project's **entry point** (`entry-point` in `jac.toml`, or the file passed directly to `jac run`) cannot use a relative import (`import from .services.x { ... }` / `import .services.x;`) to reach a sibling module. `jac run` fails with:

```
✖ Error: Error: attempted relative import with no known parent package
```

`jac check` on the same file passes with 0 errors, so the static checker and the runtime disagree on whether the file is valid — the failure only surfaces when the program is actually executed (or after `jac build`, in the sealed `.jab`).

### Minimal reproduction

Project layout:

```
relimport-demo/
├── jac.toml
├── main.jac
└── services/
    └── greeter.jac
```

`jac.toml`:
```toml
[project]
name = "relimport-demo"
entry-point = "main.jac"
kind = "cli"
```

`services/greeter.jac`:
```jac
def greet() -> str {
    return "hello from greeter";
}
```

`main.jac`:
```jac
import from .services.greeter { greet }

with entry {
    print(greet());
}
```

Steps:
```console
$ jac check main.jac
  Checking main.jac...
main.jac PASSED [100%]
============================== 1 passed in 0.80s ===============================

$ jac run main.jac
✖ Error: Error: attempted relative import with no known parent package
    1 | import from .services.greeter { greet }
      | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    2 |
    3 | with entry {
    4 |     print(greet());
    5 | }
  at <module> /path/to/relimport-demo/main.jac:1
  ... [internal runtime calls]
```

The same failure reproduces after `jac build` + running the sealed artifact:
```console
$ jac build main.jac
...
✔ Built dist/relimport-demo.jab (1,893 bytes)

$ jac run dist/relimport-demo.jab
✖ Error: Error: attempted relative import with no known parent package
    1 | import from .services.greeter { greet }
...
```

**Workaround found:** switching the same import to an absolute path works fine:
```jac
import from services.greeter { greet }   # no leading dot
```
This confirms `services/greeter.jac` itself, its symbols, and the type checker's resolution are all fine — only the *relative* form breaks, and only for the entry file, and only at runtime.

### Actual behavior

- `jac check main.jac` — passes, no diagnostics for the import.
- `jac run main.jac` — crashes with `attempted relative import with no known parent package`.
- `jac build` + running the resulting `.jab` — same crash.
- Root cause appears to be that when `main.jac` is executed as the entry point, Jac loads it into `sys.modules['__main__']` (see `jac/jaclang/jac0core/impl/runtime.impl.jac` around line 848, `module = types.ModuleType('__main__')`) without ever setting `module.__package__` / `module.__spec__.parent`. Python's import machinery requires that context to resolve a leading-dot relative import, so any relative import inside the entry file hits Python's own `ImportError: attempted relative import with no known parent package` — the same error you'd get running a plain Python script directly (`python main.py`) that contains `from .sibling import x`.
- Non-entry `.jac` files (i.e. files that are themselves *imported* by something else, not run directly) are unaffected — only the file that becomes `__main__` at execution time hits this.

### Expected behavior

Relative imports (`import from .services.x { ... }`, `import .services.x;`) should work the same way regardless of whether the file containing them happens to be the project's entry point or a module imported by something else. In particular:

- `jac check` and `jac run` should agree on whether a given import is valid — a file that type-checks cleanly should not fail purely because of how it's invoked.
- Since Jac already knows the project root via `jac.toml` (and computes `caller_dir` / `project_root` in `runtime.impl.jac`), the entry-point module should be given a synthetic package context (e.g. set `__package__` to the project's top-level package name, matching however non-entry-file relative imports already resolve `.services.x` against `caller_dir`/`project_root`) so leading-dot imports resolve correctly whether or not the file is the one being executed directly.
- At minimum, if relative imports in an entry file are intentionally unsupported, `jac check` should raise a diagnostic on the import so the failure is caught at type-check time instead of only at runtime / after `jac build`.

SandeepaHWP:
## Summary

A relative import in a project's entry file fails at runtime with `attempted relative import with no known parent package`, while `jac check` reports the same file as clean.

This resolves the import at compile time instead, which is what the native backend already does for the same source.

## Scope

**Covered:** `jac run`, `jac run --entry` / `jac enter`, and `jac start`.

**Not covered:** executing a sealed `.jab`. That case is unchanged from `main` and is left for a follow-up. Reasoning is in the last section.

## Root cause

A leading-dot import is resolved against the module's `__package__`. A module only gets a `__package__` when it is loaded as a member of a package, which happens when something imports it. The project entry file is not imported, it is loaded standalone, so its `__package__` is empty and Python has nothing to resolve the dot against.

Every other module in a project is reached by being imported, so it does receive a `__package__` and its relative imports already work. The entry file is the single file where the dot breaks.

## Fix

`PyastGenPass` emits Python imports and previously passed the relative level through unchanged, leaving `from .services.greeter import greet` for Python to resolve at runtime.

The fix rewrites a single leading dot to the absolute form for the file that is loaded standalone, and only that file:

```
before   from .services.greeter import greet
after    from services.greeter import greet
```

This is safe precisely because that file has no parent package: `.services.greeter` means "services next to me", the file's own directory is on the search path, so it resolves to the same module as the plain `services.greeter`. The two spellings point at the same file, so emitting the absolute one changes nothing about what is imported.

The compiler cannot see `__package__` (it only exists at runtime), but it can read `jac.toml`, which declares the `entry-point`. That is exactly the file that will be loaded standalone, so `flatten_import_level` flattens the single leading dot only when the module being compiled is the declared entry-point, and returns the level unchanged for every other module. Deeper levels (`..x`) are left alone since they walk up a directory, and the bare `import from . { x }` form is skipped because flattening it would leave no module name at all.

Because the dot is gone before the program runs, it no longer matters which execution path a command takes.

Keying on the declared entry-point, rather than on whether the file's directory looks like a package, is deliberate. A directory can be a package without any `__init__` file (an implicit namespace package, which the `deep/` test fixtures rely on), so a filesystem check misclassifies those and rewrites imports that are genuinely package-relative. Whether a file is loaded standalone is a load-time property, and the declared entry-point is its precise answer.

## Why compile time rather than runtime

An earlier revision of this PR anchored the entry at load time. That worked, but resolved the import to `<entry>.services.greeter`, producing a second module object for one file with its own archetype classes, so `isinstance` failed across the two copies and archetypes registered twice. Resolving at compile time keeps one file as one module and makes the Python path agree with native rather than inventing a third answer:

```
entry Token : <class 'services.greeter.Token'>
other Token : <class 'services.greeter.Token'>
same class? True
```

`find_spec` is deliberately untouched. Giving a single file `submodule_search_locations` there turns it into a package, and `runpy` then refuses to execute it without a `__main__` submodule, which breaks `python -m`. `tests/compiler/test_importer.jac::python dash m jac module` covers that.

## Tests

`test_entry_relative_import.jac` loads an entry the way `jac start` does and asserts the leading-dot import resolves. Its fixture ships a `jac.toml` declaring the entry-point, since the rule keys on that. Verified to fail against the unpatched compiler with `ImportError: attempted relative import with no known parent package`.

## Verification

Each suite was also run against `main` to separate regressions from pre-existing failures. Caches were cleared with `jac purge` plus `rm -rf ~/.cache/jac/apps` between comparisons, since stale caches produce misleading results here.

- `tests/language/test_language.jac`: 96 passed, including the deep-import fixtures.
- `tests/language/test_py_infra.jac`: 32 passed, including the deep-import fixtures.
- `tests/language/test_cli.jac`: 97 passed, matching `main`. No existing test needed changing.
- `tests/compiler/test_importer.jac`: 17 passed, matching `main`.
- `jaclang/compiler/tests/test_batch_decl_identity.jac`: 3 passed. Uses `import from . { box }`, the form the rule skips.
- `jac run`, `jac run --entry` and `jac start` all pass on two projects that differ only in whether the sibling is a `.jac` or a `.sv.jac`.

## Why `.jab` is out of scope

A sealed image is staged as a Python package. The entry inside that image is still launched standalone, so it still has no parent package at runtime and the dot still fails there. Removing the `__init__.jac` the sealer writes is not viable on its own: the seal step then fails with `Could not find a Python package`, so it is load-bearing for packaging. Closing that case needs a decision about sealed-image load semantics rather than a change to this rule, so it is deliberately excluded here and tracked separately.

Fixes #7661


RandiJay96:
this will fix in the Jac-coder side

Issue #7626, opened 2026-07-21
na completeness: residual lowerings, OSP depth, and platform/self-hosting maturity (post-#7593)


marsninja:
Successor to #7593. That issue's inventory is complete: every demotion class either lowers natively or routes server for an honest, labeled reason. This issue goes deep on the three workstreams that finish the na pathway's *language and platform* story: (A) the documented residuals #7593's waves deliberately scoped out as structured errors, (B) OSP depth beyond single-hop adjacency, and (C) platform breadth and the self-hosting arc. (The other two known frontiers -- census-ranked stdlib next-tier (`re`, `subprocess`) and the native persistence ABI -- are deliberately NOT here: the first is engine-building that deserves its own issues, the second needs a design doc before an inventory.)

The framing carries over from #7593: every gap below is currently a *loud* boundary (structured E5090/E5092, a classifier gate, or a platform guard), and finishing an item means walking it up the layers until it disappears. Prioritization stays measurement-driven: `jac check --native-census` over real projects ranks what actually blocks code; the rankings cited below come from the repo-internal corpus probe.

**Status legend:** ✅ landed · 🟨 partially landed / scoped residual · ⬜ not started.

**Update (2026-07-22):** the full inventory below is executed in #7646 -- A1-A6, B1-B4, and C1-C6 are ✅; C7 is 🟨 at its stage-0+1 scope with the census-ranked stage-2 blockers recorded in its row.

## A. Residual lowerings (bounded extensions of landed machinery)

| # | Gap | Detected at | Status | Difficulty | Long-term approach (summary) |
|---|-----|-------------|--------|:----------:|------------------------------|
| A1 | `with entry` assignments are frame-local natively, module-global in Python | E5092 (entry-lambda body) | ✅ landed (#7646) -- entry-block assignments register as native module globals (frontend already stamps them MODULE_GLOBAL), initialized in jac_entry after `__jac_glob_init`; lambdas/defs read them late-bound; the entry-lambda E5092 guardrail flipped to a behavioral pin | **M** | Globalize: lower entry-block top-level assignments to native module globals (same registration path as `glob`, initialized in `jac_entry` after `__jac_glob_init`), so lambdas/nested fns/defs resolve them as globals exactly like CPython. Delete the entry-lambda E5092 when the scan confirms no frame-local reads remain |
[...]

Issue #6942, opened 2026-06-24
Epic: Native self-hosting — a CPython-free `jac nacompile` / `jac run`


marsninja:
## Goal

Make the Jac toolchain reach native programs **without touching CPython** — both `jac run <native program>` and `jac nacompile` should execute with no `libpython` in the loop. The governing constraint:

> **The compiler must run as native code.** Today the *artifact* is already CPython-free (the in-Jac ELF/Mach-O/PE linkers emit a `libc`-only executable — `jaclang/compiler/passes/native/elf_linker.jac` links only `libc.so.6`). What still touches CPython is the **toolchain**: the `jac` driver boots `libpython` (`jac/launcher/launcher.zig` `BOOT_SRC` → `Py_Initialize`), and the compiler pipeline (parse → passes → `na_ir_gen` → `na_compile` → link) is Jac-transpiled-to-Python running on CPython. The endgame is **self-hosting the compiler on the native backend** — a new stage-2 (`compiler-on-native`) above today's stage-1 (`compiler-on-CPython`).

This is the umbrella epic. **#6930** (native parser + unitree) and **#6936** (decouple the `na` LLVM backend from the `llvmlite` Python wheel) are workstreams *inside* this goal; this issue tracks the whole ladder to a CPython-free toolchain.

---

## Two targets, very different difficulty

| Target | Status today | Distance |
|---|---|---|
| **`jac run <native program>` with no CPython** | The AOT artifact is already `libc`-only; the *only* CPython touch is the driver booting `libpython` to dispatch the CLI | **Near-term.** Either run the AOT binary directly (works today), or add a native fast-path to the launcher that execs/JITs without `Py_Initialize`. |
| **`jac nacompile` with no CPython** | The compiler *is* a CPython program | **The real project — requires self-hosting.** |

## The bootstrap ladder (how self-hosting lands)

1. **Stage-0 (exists):** `jac0.py` (Python subset transpiler) + `meta_importer.py` bring up the Jac-written compiler on CPython.
2. **Stage-1 (exists):** the Jac compiler runs on CPython (the `sv` backend). This is what the CGO papers mean by "~98% self-hosted" — *written in Jac*, **not** native.
3. **Stage-2 (this epic):** point the native backend at the **compiler's own source** → a native `jac` compiler. Then:
   - `jac1` = stage-1 (CPython) compiles the compiler source → native binary.
   - `jac2` = `jac1` recompiles the compiler source → native binary.
[...]

PR #7672, opened 2026-07-22
na self-hosting stage 2: clear the six census-ranked compiler-tree blockers (#7626 C7)

marsninja:
## Summary

#7626 closed with its C7 (self-hosting) row at stage 0+1 and a census-ranked list of the compiler-tree blockers standing between the native pathway and the compiler's own source. This PR executes that list: all six blockers are implemented, each one red-first with a pinned cross-backend equivalence fixture or checker regression test. Several of the census labels turned out to name a *symptom* whose root cause was elsewhere; each section below records what the actual defect was, since that is what the next census wave will care about.

Modules that previously demoted and now lower natively: `type_registry.jac`, `llvm/__init__.jac`, plus `wasm_linker.jac` / `macho_linker.jac` now check clean. `ts_decl_parser.jac` clears its recorded blocker (transitive captures) and now surfaces its *next* wall (a ternary shape), which is new census input rather than a regression.

## 1. Checker: Any/Unknown cascades (E1053 wasm_linker, E1055 macho_linker)

The census pointed at overload resolution, but the linkers' errors were the tail of a cascade: values that should have been `Any` degraded to strict `Unknown`, and the strict-Unknown discipline then hard-errored every downstream use. Four seams fixed:

- **Variadic tuple extraction** (`_extract_tuple_element_type`): `tuple[T, ...]` stores its ellipsis as an Unknown type-arg, so element 1+ of a `struct.unpack_from` result came back strict-Unknown instead of `T`. Any index of a homogeneous tuple now yields `T`.
- **Binary ops** (`operations.jac`): the existing `Any` early-out only matched the `AnyType` special form, not the typeshed `Any` class form (`is_explicit_any` covers both), and `Unknown` operands fell through to magic-method resolution that could only fail. `Any op Any` is now `Any`; an Unknown operand propagates a *diagnosed* Unknown silently — the operand site already carries its W1051, a second error on the op is cascade noise.
- **Subscripts** (`type_evaluator`): indexing an `Any`/`Unknown` base returned strict Unknown; it now propagates (`Any` → `Any`, Unknown → diagnosed Unknown).
- **Augmented assignment** (`type_checker_pass`): `x op= y` validated `assign(typeof(y), typeof(x))` — the raw rhs — so valid `bytearray += bytes` failed while the semantically meaningful check (the *operator result* must fit the target) was missing. The aug path now resolves `__iadd__`/fallbacks as before and assign-checks the result: `bytearray += bytes` passes, `x: int; x += 0.5` still errors.

Genuine mismatches on resolved types still report (pinned by control cases). `test_unknown_cascade.jac` covers all shapes; both linkers check clean.

## 2. Native dicts: helper-family collision + `dict.get` use-after-free (type_registry TypeError)

`TypeError: Type of #3 arg mismatch: %"JacTypeEntry"* != i8*` was not an optional-coercion bug. Dict helper families were cached under `"str:ptr"` for *every* pointer-valued dict while the family's LLVM signatures were built from whichever dict was emitted first — a module holding `dict[str, Entry]` next to `dict[str, str | None]` fed `Entry*` into an `i8*` slot and the whole module demoted on the raw llvmlite error.

- `_dict_val_spec` now names pointer value families by pointee identity (`ptr_Entry` vs `ptr`), and `_codegen_dict_val` prefers the checker-declared family, coercing every pair into it, instead of trusting the first value's LLVM type.
- Fixing that made struct-valued dicts *reachable* and exposed a latent use-after-free: `dict.get` returned the stored pointer borrowed while scope-exit released it as an owned local — the release stole the dict's reference and the entry was freed while still held (crash timing depended on heap state). `emit_get` now retains its pointer result (null-safe, no-op for static allocations).

Fixture: `prim_dict_struct_vals.jac` — the exact registry shape (`dict[str, Obj]` beside `dict[str, str | None]`, `.get` with defaults, optional-field reads), stable across repeated runs. `type_registry.jac` lowers natively.

## 3. os.environ lowers natively (llvm/__init__.jac)

Native `os` had `getenv` but `environ` is an attribute, so `os.environ.get(...)` died at the capability gate. The compiler tree only uses `.get` (including nested defaults), three `[]=` writes, and `.pop`:

- `os.environ.get(k[, d])` delegates to the existing `getenv` intrinsic — identical contract (found → owned copy, missing → default/None).
- `os.environ.pop(k[, d])` → getenv-copy + `unsetenv`; `os.environ[k] = v` → `setenv(k, v, 1)`.
- Narrow capability-gate carve-out for the `os.environ` attribute shape.

Fixture: `prim_os_environ.jac`. `llvm/__init__.jac` lowers natively.

## 4. Transitive sibling captures (ts_decl_parser kind_at)

Native closures resolved captures textually, one frame deep. The parser family shape — sibling nested fns sharing `toks`/`pos` where `is_p` reaches them only by *calling* `kind_at`, and `parse_type_full` is called before its definition — either failed loudly or, for the forward-reference case, crashed the IR builder with an IndexError from a short argument list.

- `_predeclare_nested_siblings`: the first time any nested def in a frame lowers, all top-level siblings are prototyped together with a **fixpoint over transitive capture sets** — a sibling forwards the full capture set of every sibling it references, and a var any callee holds by pointer stays a pointer through the whole chain (writes land in the defining frame, not a stale copy). Captures of locals declared between defs prototype from their checker type; cell-vars defer to the old per-def flow.
- The capture-miss diagnostic path now pads the call to keep the IR well-typed: a structured E5092 instead of a compiler crash.
- `boundary_analysis_pass`: nested fns no longer enter the cross-codespace interop registry. A forward sibling call inside an `na` block used to bind by bare name against the *server-side* twin and emit a module-level `__jac_interop_py_funcs__['parse_b'] = parse_b` stub referencing a nested name — a NameError at import for any module containing the pattern.

Fixture: `prim_closures_transitive.jac` (shared mutated `pos` through two levels of sibling calls, plus forward + mutual recursion). All pre-existing closure suites stay green.

## 5. Archetype-level generics and explicit-self methods (primitives.jac)

Two layers here. Generic class templates (`class IntEmitter[V, C]`) are now excluded from direct lowering — they are uninstantiable templates; subclasses that fix the type args (`NativeIntEmitter(IntEmitter[(ir.Value, NativeEmitCtx)])`) lower as ordinary archetypes.

Behind that label sat a broader gap: **class-style methods with an explicit `self` parameter had never lowered at all**. The checker leaves the explicit `self` param Unknown-typed and the native prototype counted it as a second parameter after the receiver, so every such method failed and misaligned call-site argument mapping. Prototype building, body binding, and bound-call arg mapping now treat a leading explicit `self` as the receiver slot it is.

Fixture: `prim_generic_classes.jac` (generic all-abstract base + specialized concrete subclass, dispatched calls). The emitter *reflection* layer (`getattr`/`dir`/`inspect.signature` in `_ensure_emitters`) is intentionally out of scope — it needs a static registration table and can continue riding per-method demotion until then.

## 6. bytearray lowers natively (linker_common + all object-file linkers' blocker)

The growable byte buffer every linker and the JIR marshaller is built on, previously aliased to immutable `jacbytes` so every constructor and mutation failed. It now lowers as a `List.u8` container — `{i64 len, i64 cap, i8* data}` — reusing the standard container new/append/get/set/len/RC machinery, plus a dedicated runtime (`bytearray_na.impl.jac`):

- `__bytearray_reserve` (doubling growth with a zero-cap floor), `__bytearray_extend_raw`, and `__bytearray_setslice` (full Python splice semantics: clamping, tail memmove, length change).
- Constructors: `bytearray()`, `bytearray(n)` (zero-filled), `bytearray(bytes)`, `bytearray(bytearray)`. Zero-arg builtin calls previously never reached the emitter at all — the builtin dispatch required at least one argument, so `bytearray()` silently server-demoted its module; that gate is fixed.
- `.append(int)` / `.extend(bytes|bytearray)`; index reads **zero-extend** (bytes are unsigned — pinned with >127 values); index writes truncate; slice reads produce new buffers; `buf[a:b] = src` splices; `+=` extends in place — sourcing from the raw rhs, since the pre-coerced value bitcasts a `bytes` rhs into the buffer type and misreads its layout (this was a runtime segfault).
- `struct.pack_into(fmt, buf, off, vals...)` with literal formats and a runtime bounds check, sharing the endian-explicit byte-store scheme of the existing `struct.pack`.
- `bytes(bytearray)` freezes via raw-buffer copy; `len()` and `for b in buf` ride the container machinery.

Fixture: `prim_bytearray.jac` — three test groups (core ops, slices/iteration/`+=`, pack_into), byte-for-byte against the CPython reference.

## Validation

Targeted suites only, fresh caches:

- `test_prim_equivalence.jac`: **106/107** — the one failure is the pre-existing `string escapes` test whose required `cl` leg needs a Node.js binary this environment lacks; the same environmental cause accounts for all OSP/shared-types `cl`-leg failures. Every native leg passes, including the nine new tests added by this PR.
- `test_unknown_cascade.jac` (new), `test_global_narrowing.jac`, `test_untyped_third_party.jac`, `test_subscript_assignment.jac`: 20/20.
- All previously-passing closure, dict, os, and generics equivalence tests stay green.

## Notes for reviewers

- Every fix follows the #7626 discipline: failing test first, then the lowering/checker change; loud boundaries stay loud where the semantics require it (control cases pin that genuine type errors and unsupported shapes still report).
- One diagnosability gap surfaced while debugging and is worth its own issue: when an `na` block fails to lower inside a server module, the demote driver truncates the native errors, so the fallback is completely silent outside `--native-census` runs.
- `ts_decl_parser.jac` now demotes on an `IfElseExpr` shape further down its body — the next-layer finding this PR's fixes exposed, not a regression.

Closes out the stage-2 blocker list recorded in #7626 (C7); the standing arc continues in #6942.


marsninja:
Addressed the review findings in 983b38711:

- **sym.decl guard** (boundary_analysis): added the None guard before nested-fn classification.
- **fwd_mutual_cl stub**: added; the fixture now follows the same convention as its siblings.
- **pack_into offset**: the offset expression is now checked for lowering failure and reports a structured E5092 instead of crashing the coercion.
- **setslice aliasing**: real bug — the source can alias the buffer (`b[a:c] = b`), and both the growth realloc and the tail shift would corrupt it mid-splice. setslice now snapshots its source; extend_raw re-derives an in-buffer source from its offset after growth (`b += b` / `b.extend(b)` are safe, with no copy overhead in the common non-aliased case).

The same commit also reconciles the checker changes with the strict-Unknown production pins (ops on Any stay strict at typed seams; only Any-class *pairs* — the `tuple[Any, ...]` unpack case — take the silent early-out), types wasm_linker's bare containers at the source, and restores the `bytearray([ints])` constructor the earlier aliasing to `emit_bytes` had provided.

Issue #5628, opened 2026-04-20
App-author requirements for multi-pod jac-scale deployments (5 coordination needs)


udithishanka:
# App-author requirements for multi-pod jac-scale deployments

> Reframing note (2026-04-22): earlier revisions of this issue prescribed a specific API shape (`@restspec(lock=..., stream=...)`, `Jac.kv`, etc.). The jac-scale team flagged reasonable concern that those look tailored to one consumer (jacBuilder). This rewrite inverts that. I show **the actual code we run today, the concrete failure modes in multi-pod, and the user-visible symptoms** — and leave the implementation shape to the jac-scale team. Prior API proposals are at the bottom as reference only.

## TL;DR

jac-scale already makes **graph persistence** and **L2 graph-read caching** fully transparent. For a single-pod app, that's enough. For a **multi-pod app**, five additional coordination problems show up that graph persistence cannot solve.

Below I walk through each one with the **real code from jacBuilder today**, explain exactly why it breaks at N > 1 pods, and describe the user-visible failure. These same failure patterns will appear in any Jaseci app that shares state via EFS, a database, or WebSockets — they're not jacBuilder-specific.

---

- [ ] **Need 1 — Mutual exclusion across pods for shared-resource writes** @ahzan-dev 

### Current code: `services/community_manager.jac`

```jac
import from threading {Lock}

glob _community_lock = Lock();

def submit_community_template(user_id, name, description, tags, jacpack_path) -> dict {
    ...
    _community_lock.acquire();
    try {
        comm_id = "comm_" + str(uuid4()).replace("-", "")[:16];
        dest_path = os.path.join(COMMUNITY_ROOT, comm_id + ".jacpack");
        shutil.copy2(jacpack_path, dest_path);
        items = _read_manifest_unlocked();      # reads manifest.json from EFS
        items.append(new_item);
        _write_manifest_unlocked(items);        # writes manifest.json to EFS
    } finally {
        _community_lock.release();
    }
}
```

### What's wrong with this

`threading.Lock` is a **per-process** primitive. Each jacBuilder pod instantiates its own `_community_lock` at module load — they don't share. On a 3-pod deployment behind a load balancer, two users submitting templates simultaneously route to different pods:

1. Pod A acquires its local `_community_lock`, reads `manifest.json` (100 items), appends, writes (101 items)
[...]

PR #7691, opened 2026-07-23
fix(runtimelib): converge concurrent find-or-create on the system root

### LINKED ISSUE #7605: jac-scale: concurrent find-or-create on system_root duplicates singletons and raises WriteConflict on superroot anchor

Thamirawaran:
Fixes #7605.

## What was happening

Concurrent requests that find-or-create a singleton node on the shared system root duplicated the singleton and returned WriteConflict errors as 500s. Two defects compounded:

1. **One mutable super-root object shared by every request.** `JScaleExecutionContext.init` injected the anchor object from a process-global cache into every request context's L1, so all in-flight requests aliased the same `NodeAnchor` — one request's uncommitted `++>` edges leaked into every other request's traversal and commit before anything reached the store.
2. **OCC replay never re-resolved the system root.** The retry loop in `_run_function_with_occ` re-resolved the user root between attempts ([server.impl.jac:419](https://github.com/jaseci-labs/jac/blob/main/jac/jaclang/runtimelib/impl/server.impl.jac#L419)) but not the system root, so every replay traversed the same stale anchor, re-created the "missing" singleton, and re-conflicted on the same stale CAS version until attempts were exhausted.

## The fix (deliberately minimal)

- **Drop the process-cache injection** ([scale/memory/impl/context.impl.jac](https://github.com/jaseci-labs/jac/blob/main/jac/jaclang/scale/memory/impl/context.impl.jac)): the super root is now resolved per context through the memory hierarchy like any other anchor. No snapshots, locks, or template publishing — the shared-mutable-object special case is simply removed.
- **`ExecutionContext.arefresh_system_root()`** (new, [runtimelib/impl/context.impl.jac](https://github.com/jaseci-labs/jac/blob/main/jac/jaclang/runtimelib/impl/context.impl.jac)): evicts the super root from L1 *and* the persistence layer's read cache (`l3.delete` is cache-evict for backends that keep one — without this, SQLite's internal `__mem__` serves the loser's own dirty object right back), re-fetches, and rebinds `system_root`/`user_root`/`entry_node`.
- **Call it on OCC replay** ([runtimelib/impl/server.impl.jac](https://github.com/jaseci-labs/jac/blob/main/jac/jaclang/runtimelib/impl/server.impl.jac)): losers now replay against the winner's committed state, find the singleton, and skip creation.

`TieredMemory.abort` is untouched: its `l3.delete` is already the correct cache-eviction step in-tree (SQLite pops its read cache; scale Mongo delete is a no-op).

## Regression test

`test_superroot_singleton_race.jac` drives 40 concurrent unauthenticated find-or-create requests against a real `jac start` server — Mongo+Redis via testcontainers plus a SQLite variant — and asserts every request succeeds, a fresh observer sees exactly 1 singleton, and a late find-or-create creates nothing. The fixture (`superroot_singleton_app.jac`) uses only public builtins (`jobj`/`grant`/`commit`).

| suite | main | this branch |
|---|---|---|
| new race test (SQLite) | FAIL — 9 duplicate singletons | pass |
| new race test (Mongo+Redis) | — | pass |
| test_occ_replay (7) | pass | pass |
| test_persistence_race (2) | pass | pass |

## For reviewers comparing with #7612

#7612 fixes the same issue but keeps a process-level anchor cache alive via per-request deep copies, a version-gated template publisher, a threading lock, and a commit hook. Since the copy already pays serialize/deserialize per request, the cache only saves a backend read — this PR removes the cache instead, so the super root follows the exact same load/CAS/replay path as every other anchor. Also note #7612's `brand_new`-only `l3.delete` in `abort()` removes the SQLite read-cache eviction that replay convergence depends on for non-root anchors.

## Follow-ups (out of scope)

- No public user-land name for the system root exists (`root` and `root.shared` both resolve to the guest root under serve) — apps must hardcode the well-known UUID or import runtime internals. A `root.system` builtin mirroring `root.shared` would close that gap.
- The request-close commit path still demotes conflicts to warnings outside the OCC loop.
- Import-time (`with entry`) bootstrap across multiple pods remains outside OCC; a storage-atomic upsert (#5628) is the backstop there.


Thamirawaran:
mistakenly created pull request

Issue #7680, opened 2026-07-23
jac-client 0.34.3: #7654 route-as-type silently drops old page-name-convention routes; app builds green, serves blank


udithishanka:
## Summary

#7654 (jac 0.34.3) changed route detection from the `page`/`layout` **name convention** to a **type** (`JsxPage`/`JsxLayout`), but a `pages/` module that still uses the old convention (`def:pub page() -> JsxElement`) is now **silently** treated as not-a-route. `jac build` exits 0, reports "Compiled N/N modules", and emits **no error or warning** — yet no `_routes.js` is generated and the client entry is built without a router (`root.render(<App/>)`). The app builds green and deploys green, then serves a **blank page** (React mounts into an empty tree; base-route `app` returns `props.children`, which is empty). There is no runtime error, no console output, and nothing in CI/deploy logs to indicate the app is broken.

Root cause: `RouteScanner` now requires a public export returning the ambient `JsxPage` type; the previous marker (a symbol named `page`) is dropped with no back-compat and no diagnostic.

## Environment

- Regressed in: `jac 0.34.3` (macOS arm64 and Linux x86_64, release binaries).
- Works in: `jac 0.34.2`.
- Cause: #7654 "feat(client): make a route a type - nominal JsxPage/JsxLayout replace the page/layout name convention" (commit `abbebc7c6`, merged 2026-07-22 11:18Z). v0.34.2 was published 06:03Z (before the merge); v0.34.3 at 23:52Z (after). The version boundary straddles #7654.
- Observed in production: docs.jachammer.ai (a jac-scale web-app deployed on 0.34.3) serves HTTP 200 with a blank page; the served client bundle contains zero route paths and a bare `root.render(<App/>)` entry.

## Reproduction (executed, not hypothetical)

Minimal file-based-routing web-app, identical for both versions:

```bash
jac create --kind web-app routeapp && cd routeapp

# base-route wrapper (renders the matched route)
cat > main.jac <<'EOF'
cl { def:pub app(props: any) -> any { return props.children; } }
EOF

# a page using the pre-0.34.3 name convention
mkdir -p pages/intro
cat > pages/intro/welcome.jac <<'EOF'
cl { def:pub page() -> any { return <div>welcome page</div>; } }
EOF

echo Y | jac build main.jac
ls .jac/client/compiled/_routes.js
grep -c 'react-router' .jac/client/compiled/_entry.js
```

Results, same app, same command:

[...]

PR #7698, opened 2026-07-24
feat(jac-scale): pin the pod runtime to a published release via [scale-runtime] version / JAC_SCALE_VERSION

### LINKED ISSUE #7682: jac-scale: no way to deploy on a specific/earlier stable release; STABLE channel hardcoded to releases/latest (no rollback escape hatch)
## Summary

jac-scale has no supported way to deploy against a specific or *earlier* published stable release. The pod-runtime channel resolver (`dev_config.jac`) offers only three modes — `STABLE` (hardcoded to `releases/latest`), `DEV` (rolling `dev` prerelease, i.e. newer), and `LOCAL` (a binary file the operator supplies) — with no "pin to release tag `vX.Y.Z`" option. When `releases/latest` ships a regression, every `jac start --scale` deploy is forced onto the broken version, and the only escape is the `LOCAL` channel (manually download the good release binary and point `JAC_SCALE_BINARY_PATH` at it). There is no `releases/tags/<tag>` path and no version/tag env.

This just bit us in production: jac 0.34.3 shipped #7680 (route-as-type silently drops old-convention routes), which made a deployed jac-scale web-app serve a blank page. There was no way to say "deploy this app on the last known-good stable, 0.34.2" — pinning the deploy runner's binary does nothing, because `jac start --scale` downloads its own pod runtime from `releases/latest` regardless.

## Environment

- jac 0.34.3 (Linux x86_64) pod runtime; deploy driven by `jac start --scale`.
- `jac/jaclang/scale/config/dev_config.jac` (channel resolver), `jac/jaclang/scale/injector/binary.jac` (binary download).

## Reproduction (executed)

**1. The channel resolver has no version-pin path** — `dev_config.jac`:

```jac
def resolve_dev_channel(project_dir: str) -> Channel {
    if os.environ.get(BINARY_PATH_ENV, "") {   # JAC_SCALE_BINARY_PATH -> LOCAL
        return Channel.LOCAL;
    }
    config = get_config_for_path(Path(project_dir));
    if config is None or config.get_section("dev") is None {
        return Channel.STABLE;                  # default
    }
    return Channel.DEV;                          # [dev] stanza -> rolling prerelease
}
```

**2. STABLE is hardcoded to `releases/latest`** — `binary.jac`:

```jac
# Stable asset names carry the release version, so resolve them through the API.
api = f"https://api.github.com/repos/{repo}/releases/latest";
```

There is no `releases/tags/<tag>` branch and no `JAC_SCALE_VERSION` / `JAC_SCALE_RELEASE_TAG` env anywhere in `jac/jaclang/scale/` (grep confirms none).

**3. Real deploy showing the gap.** We pinned the deploy *runner's* binary to v0.34.2 and tried to deploy. The runner ran 0.34.2, but jac-scale still pulled latest for the pods, so the pod came up on 0.34.3:

```
Install step: Pinned: v0.34.2
Install step: jac 0.34.2  (Linux x86_64)
Deploy step:  Ready: downloading the latest published (stable) jac release.
=> pod runtime: jac 0.34.3  (Linux x86_64)
```

So even with the runner explicitly on 0.34.2, there was no way to get 0.34.2 onto the pods short of the `LOCAL` file workaround.

## Expected

An app should be able to deploy on a specific published stable release — at minimum to roll back off a regressed `latest` — via a documented config/env, e.g. `JAC_SCALE_RELEASE_TAG=v0.34.2` or a `[scale] runtime_version = "0.34.2"` field, resolving `releases/tags/<tag>` instead of `releases/latest`.

## Actual

The only ways to avoid `latest` are: `[dev]` (a *newer*, rolling prerelease — not a rollback) or `LOCAL` (operator manually fetches the release binary and sets `JAC_SCALE_BINARY_PATH`). There is no first-class "deploy on release vX.Y.Z."

## Impact

`latest`-only means every jac-scale deploy auto-adopts each new stable the moment it publishes, with no pinning and no rollback escape hatch. A single regressed release (like 0.34.3 / #7680) silently breaks all jac-scale deploys, and operators have no supported knob to hold a known-good version while the fix lands. This is a significant supply-chain / release-safety gap for anything deployed via `jac start --scale`.

## Workaround (verified)

Use the `LOCAL` channel: on the deploy runner, download the desired release binary (e.g. `jac-0.34.2-linux-x86_64`) and export `JAC_SCALE_BINARY_PATH=/path/to/jac` before `jac start --scale`. `resolve_dev_channel` then returns `LOCAL` and ships that exact binary to the pods. Verified this is the only mechanism that pins the pod runtime to an older release; it works but requires the operator to manage the binary out-of-band.

## Suggested fix

Add a release-tag pin to the `STABLE` path: an env (`JAC_SCALE_RELEASE_TAG`) and/or a `[scale]` config field that, when set, resolves `https://api.github.com/repos/{repo}/releases/tags/<tag>` instead of `releases/latest`, with a clear error if the tag or per-arch asset is missing. This gives a first-class, auditable "deploy on release vX.Y.Z" without the operator hand-managing binaries, and a real rollback path when `latest` regresses. A regression test: `resolve_dev_channel` / the binary resolver honors the tag env and hits the tags endpoint.

kashmithnisakya:
## Summary

Fixes #7682: jac-scale had no supported way to deploy pods on a specific or earlier published stable release. The STABLE channel was hardcoded to `releases/latest` in two places, so when a regressed release published (0.34.3 / #7680), every default deploy adopted it immediately and the only rollback was hand-managing a binary via `JAC_SCALE_BINARY_PATH`.

This adds a first-class version pin:

```toml
[scale-runtime]
version = "0.34.2"
```

overridable per-deploy with `JAC_SCALE_VERSION=0.34.2` (a leading `v` is accepted and stripped).

## Design

- **One pin, every consumer.** A pin resolves the runtime binary and the admin console from `releases/tags/v<version>` (both former `releases/latest` hardcodes go through one new `_release_meta()` helper), moves the pod base image to `jaseci/jaclang:<version>`, caches assets in version-scoped slots, and stamps pods with a `JAC_SCALE_RUNTIME_VERSION` container env so `kubectl` can answer "which jac is this pod running". Deploy output names the pinned release instead of "stable".
- **No fifth `Channel` member.** A pin is `Channel.STABLE` plus a non-empty version string, resolved once (`resolve_pinned_version`), so pinned-ness cannot desync from the version. `peek_pin()` is the never-raising lookup used on paths that must stay total (the LOCAL escape hatch, explicit `python_image`).
- **Precedence** (highest first): `JAC_SCALE_BINARY_PATH` (pin ignored with a warning) > `JAC_SCALE_VERSION` > `[scale-runtime] version` > `[experimental]` > `[dev]`. A jac.toml committing both `[scale-runtime]` and `[experimental]` is a hard error; the env pin overrides `[experimental]` with a loud warning instead, so an exported variable can never brick an unrelated project's deploy. The pin outranks `[dev]` because `[dev]` is the host dev stanza, present incidentally in contributor checkouts.
- **Why `[scale-runtime]` and not `[scale] version`**: the capability resolver treats a non-empty `[scale]` section as deploy intent and adds `pymongo` to the closure (`capabilities.jac`), and `examples/mini_todo` ships an empty `[scale]` table, so any key under `[scale]` would silently change a plain `jac serve`'s dependency set. A new top-level table is inert by construction (`get_section` is called with exactly three names repo-wide).
- **Supported window**: the current minor and one back, anchored to the deploy runner's own version (auto-advances, nothing to rot). Older pins are rejected at resolve time with an error that names the window and the `JAC_SCALE_BINARY_PATH` escape, because the seal step runs the pinned binary against the current tree's precompile script. Unknown runner version skips the check.
- **Base image moves in lockstep** so the existing sha256 gate in the pod bootstrap *matches* (the image binary is built from the same release asset) and the pod symlinks instead of re-extracting. `prefer_image_binary` is deliberately not widened: the sealed `.jir` format version is fail-closed at pod boot, and the artifact is sealed by the injected binary, so coherence comes from making the checksum pass, never from bypassing it. Hub-probe behavior for a pin: definitive 404 falls back to `python:3.12-slim` with a warning naming the 0.32.0 image floor; any other probe failure keeps the versioned tag.
- **Pinned failures are loud.** A pinned admin console with a missing checksum or missing index.html aborts the deploy instead of warn-and-continue; a release that genuinely publishes no console asset stays non-fatal (pre-0.31 releases). Unpinned behavior is unchanged.
- **Byte-identity when unpinned**: with no pin set, every resolved URL, cache filename and console string is identical to before (the one deliberate exception: the transient download tmp-file name switched from `with_suffix` to `with_name`, since `with_suffix` would mangle dotted version cache names like `jac-0.34.2-linux-x86_64` and collide the two arches on one tmp path). New tests lock the unpinned `releases/latest` URL and `jac-stable-linux-x86_64` cache name in.

## Tests

24 new tests across `test_binary_injector.jac` and `test_binary_cache.jac`: the full precedence ladder (including both-pins-set and both-conflict cases), validation edges (malformed, numeric-toml, leading-v), window edges (below/at/above, unknown runner), the tags-endpoint URL for binary and admin dist, version-scoped vs unpinned cache filenames, the pinned admin-dist loud/quiet forks, base-image 404/probe-failure/digest paths with URL assertions, the dry-run end-to-end pin threading, and the env stamp. The two admin-dist test fakes gained the new signature plus a call-count assertion so a silent TypeError can never hide again.

`jac check` clean on every touched checkable file; microservices and deploy test legs fully green locally.

## Out of scope (follow-ups)

- Deleting the inert pinning knobs (`jaseci_repo_url`/`jaseci_branch`/`jaseci_commit`/`install_jaseci`, `[project] jac-version`, `image_tag`) - breaking change, docs now mark them inert and point at the real pin.
- `GITHUB_TOKEN` forwarding in `_http_get_bytes` (shared with the Docker Hub probe; an ungated header would leak the credential cross-host).
- The missing `experimental -> dev` rewrite in `_admin_dist_urls` (pre-existing, changes what a live channel ships).


kashmithnisakya:
superseded by #7713, which landed the pin via [project] jac-version with the same lockstep console/image resolution. closing

PR #7701, opened 2026-07-24
fix(byllm): stop forcing a default temperature on every LLM call

### LINKED ISSUE #7687: byllm: forced temperature breaks GPT-5/o-series (OpenRouter) and Claude thinking mode
## Summary

byLLM always puts an explicit `temperature` into every `litellm.completion` call (`BaseLLM.make_model_params`, `jac/jaclang/byllm/llm.impl/basellm.impl.jac:960`, default `0.7` from `:863`). Newer OpenAI models only accept the default `temperature=1`: the GPT-5 family (including gpt-5-codex) and the o-series reasoning models (o1/o3/o4). byLLM sets `litellm.drop_params = True` (`model.impl.jac:31`), which strips the unsupported value on the DIRECT openai route, but litellm does NOT apply that per-model rule when the same model is reached through an aggregator provider such as OpenRouter (`openrouter/openai/gpt-5.x`). On that route litellm forwards `temperature=0.2` to the upstream, which returns:

```
litellm.BadRequestError: OpenAIException - Unsupported value: 'temperature' does not support 0.2 with this model. Only the default (1) value is supported.
```

So any GPT-5 / o-series turn routed through OpenRouter fails outright. This is exactly the BYOK path in jacBuilder (jaseci-labs/jacBuilder#839), where users bring their own OpenRouter key and pick `openrouter/openai/gpt-5.x`.

This is really a provider-generic bug: byLLM forces a call param the caller never set, and it breaks on any model that constrains that param. GPT-5 / o-series is the loud case today. The newest Anthropic models are the next one: plain Claude calls accept `temperature=0.7`, but with extended thinking enabled Anthropic requires `temperature=1`, and litellm forwards byLLM's forced value regardless (see Part 3 below). Fixing this generically is what keeps byLLM working as providers keep shipping models with new constraints.

## Environment

- jac-byllm, jaseci-labs/jac `main` @ `23a7f88e` (byLLM source now folded into `jac/jaclang/byllm/`).
- litellm `1.93.0` (local reproduction below).
- Production sighting: jacBuilder PR-839 preview (jac dev channel), model `openrouter/openai/gpt-5.x` invoked with a user BYOK OpenRouter key. The verbatim `OpenAIException` above was read from the `jac-coder-sv` pod logs there.

Two parts: a deterministic litellm-level repro that shows byLLM's `temperature` survives to the wire on the aggregator route (no API key needed), and the verbatim production error.

Part 1, litellm param resolution (mirrors what byLLM sends: `temperature=0.2`, `litellm.drop_params=True`):

```python
import litellm
for model, prov in [("gpt-5","openai"), ("o1","openai"),
                    ("openrouter/openai/gpt-5","openrouter"),
                    ("openrouter/openai/gpt-5.1","openrouter")]:
    op = litellm.utils.get_optional_params(
        model=model, custom_llm_provider=prov, temperature=0.2, drop_params=True)
    print(model, prov, "->", op.get("temperature", "<dropped>"))
```

Output (exit 0):

```
litellm==1.93.0

byLLM sends: temperature=0.2 (call_params), litellm.drop_params=True (byLLM default)
------------------------------------------------------------------------
gpt-5                      via openai     -> SAFE (dropped)
o1                         via openai     -> SAFE (dropped)
openrouter/openai/gpt-5    via openrouter -> FORWARDS temperature=0.2 -> upstream 400
openrouter/openai/gpt-5.1  via openrouter -> FORWARDS temperature=0.2 -> upstream 400
```

For completeness, with `drop_params=False` litellm rejects it client-side even on the direct route, which is the same root cause surfaced earlier in the chain:

```
model=gpt-5  drop_params=False -> RAISES UnsupportedParamsError: gpt-5 models (including gpt-5-codex) don't support temperature=0.2. Only temperature=1 is supported
model=o1     drop_params=False -> RAISES UnsupportedParamsError: O-series models don't support temperature=0.2. Only temperature=1 is supported
```

Part 2, the same failure end to end in production (jacBuilder PR-839, GPT-5 via OpenRouter BYOK):

```
Error: litellm.BadRequestError: OpenAIException - Unsupported value: 'temperature' does not support 0.2 with this model. Only the default (1) value is supported.
```

Part 3, the newest Anthropic models (same forced-param design, different constraint). Verified with the same litellm param resolution, `temperature=0.7`, `drop_params=True`:

```
[A] plain temperature=0.7 -- direct anthropic vs openrouter
  claude-sonnet-4-5-20250929                 [anthropic ] temperature=0.7
  openrouter/anthropic/claude-sonnet-4-5-20250929 [openrouter] temperature=0.7
  claude-opus-4-20250514                     [anthropic ] temperature=0.7
  (all Claude 3.5 / 3.7 / 4 / 4.5 ids: temperature=0.7 accepted, no drop, no raise)

[B] extended thinking + temperature=0.7 (Anthropic requires temperature=1 with thinking)
  claude-3-7-sonnet-20250219       drop_params=True  -> temperature=0.7 thinking=True
  claude-sonnet-4-20250514         drop_params=True  -> temperature=0.7 thinking=True
```

So plain Claude usage is fine today (Anthropic accepts `temperature` in 0..1, and 0.7 is valid), which is why this has not surfaced for Claude the way it has for GPT-5. But [B] shows the latent break: when extended thinking / reasoning is enabled, Anthropic's API rejects any `temperature` other than 1, and litellm forwards byLLM's forced 0.7 anyway (it is not dropped even with `drop_params=True`). Any byLLM path that turns on thinking for a Claude model while `make_model_params` still injects the default temperature will 400 the same way. The generic fix below covers this without a per-provider special case.

## Expected

A GPT-5 / o-series model, and any current or future Anthropic model, completes regardless of how it is routed (direct provider or through an aggregator like OpenRouter) and regardless of mode (including extended thinking). byLLM should not force a `temperature` (or any call param) the model rejects when the caller never asked for one.

## Actual

`make_model_params` unconditionally adds `temperature` (`basellm.impl.jac:960`, default `0.7` from `:863`). `litellm.drop_params = True` (`model.impl.jac:31`) only saves the direct-provider route, because litellm's value-constraint knowledge for GPT-5/o-series is keyed to `custom_llm_provider="openai"` and does not fire for `custom_llm_provider="openrouter"` (or any other proxy/aggregator provider). So the param is forwarded and the upstream 400s.

## Impact

All GPT-5 and o1/o3/o4 usage through OpenRouter, or any proxy/aggregator provider, is broken. In jacBuilder this is the BYOK flow (jaseci-labs/jacBuilder#839): a user adds an OpenRouter key, picks a GPT-5 model, and every turn fails with the error above. The same class of failure hits any newly shipped model that constrains `temperature`, since byLLM always sends one. For the newest Anthropic models the break is latent rather than immediate: normal Claude turns work, but enabling extended thinking on a Claude model surfaces the identical forced-param 400 (temperature must be 1 when thinking is on). As providers keep shipping models with new param constraints, the forced-param design keeps breaking them one by one.

## Workaround (verified)

Verified at the litellm param layer (Part 1 above): the direct openai route drops the unsupported value, so `gpt-5` / `o1` reached as `custom_llm_provider="openai"` work, while the `openrouter/openai/...` route forwards it and fails. So a user whose key is a direct OpenAI key can select the direct model id (`gpt-5`) rather than `openrouter/openai/gpt-5`. For OpenRouter keys there is no user-side workaround, the model id must go through the openrouter route.

## Suggested fix

Only include `temperature` (and `max_tokens`) in the params dict when the caller explicitly set them, and omit them otherwise so the provider's own default applies. That fixes every route including aggregators, which `drop_params` cannot cover. Concretely, in `BaseLLM.make_model_params` (`basellm.impl.jac:960`) stop folding in `default_temp`; add the key only when `temperature` is present in the instance/config call_params that the user actually supplied. Optionally, keep sending an explicit value only for models known to accept arbitrary temperatures.

Regression test: assert `make_model_params` for a GPT-5 / o-series model, and for its `openrouter/openai/...` variant, does not carry a non-default `temperature` when the caller did not set one. Add an Anthropic case too: a Claude model with extended thinking enabled and no caller-set temperature must not carry `temperature` in the params (Anthropic requires 1 when thinking is on).

Related: jaseci-labs/jac#5863 (pass-through arbitrary litellm call params) covers the inverse need (letting callers set params); this issue is about not forcing one they never set.

SupulHeshan:
## Summary

`make_model_params()` always injected `temperature=0.7` (and a `max_tokens` default) into every `litellm.completion()` call, even when the caller never asked for one. `litellm.drop_params=True` only strips unsupported values on the direct-provider route (`custom_llm_provider="openai"`); it does **not** fire for aggregator routes such as OpenRouter, so `openrouter/openai/gpt-5.x` (and o-series models) forwarded the forced value straight to the upstream, which rejects it:

```
litellm.BadRequestError: OpenAIException - Unsupported value: 'temperature' does not support 0.7 with this model. Only the default (1) value is supported.
```

The same forced-param design also latently breaks Claude with extended thinking enabled, which requires `temperature=1`.

Fixes #7687.

## Before

```mermaid
flowchart TD
    A[Caller invokes LLM] --> B["make_model_params()"]
    B --> C["Always sets temperature=0.7\n(forced, caller never asked)"]
    C --> D{custom_llm_provider}
    D -->|"openai (direct)"| E["drop_params strips the\nunsupported value"]
    E --> F["✅ call succeeds"]
    D -->|"openrouter (aggregator)"| G["drop_params does NOT apply here\ntemperature forwarded as-is"]
    G --> H["❌ 400 BadRequestError\n'temperature does not support 0.7'"]
```

## After

```mermaid
flowchart TD
    A[Caller invokes LLM] --> B["make_model_params()"]
    B --> C{"Did the caller set temperature?\n(call-site kwargs or jac.toml)"}
    C -->|yes| D[Include temperature in params]
    C -->|no| E[Omit temperature entirely]
    D --> F["litellm.completion(...)"]
    E --> F
    F --> G["✅ works on every route\n(direct, OpenRouter, ...)\nand every model\n(GPT-5/o-series, Claude+thinking)"]
```

## Changes

- `temperature` / `max_tokens` are now only included in the litellm call params when the caller actually set them — either at the call site (`llm(temperature=...)`) or via `jac.toml`'s `[byllm.call_params]` — otherwise the key is omitted entirely so the provider's own default applies. This mirrors how `tool_choice` is already handled.
- `get_call_params_config()` / `get_default_config()` no longer bake in a hardcoded `temperature: 0.7` default, so config-level detection of "did the user actually set this" is possible.
- Updated the `call_params.temperature` config-schema description to reflect the new behavior.
- Updated the `llm_params` fixture/test to set `temperature`/`max_tokens` explicitly (verifying pass-through), and added regression tests covering: no forced temperature/max_tokens by default, explicit call-site values passing through, and jac.toml-configured values applying only when configured.

### DIFF
--- a/jac/jaclang/byllm/impl/config_loader.impl.jac
+++ b/jac/jaclang/byllm/impl/config_loader.impl.jac
@@ -13,7 +13,7 @@ impl JacByllmConfig.get_default_config(self: JacByllmConfig) -> dict[str, any] {
             'proxy': False,
             'verbose': False
         },
-        'call_params': {'temperature': 0.7, 'max_tokens': 0, 'max_output_retries': 3},
+        'call_params': {'max_tokens': 0, 'max_output_retries': 3},
         'litellm': {'local_cost_map': True, 'drop_params': True, 'debug': False},
         'fallback': {'strategy': 'fallback', 'num_retries': 1, 'timeout': 60.0},
         'prompt_caching': {'enabled': True},
@@ -60,12 +60,14 @@ impl JacByllmConfig.get_model_config(self: JacByllmConfig) -> dict[str, any] {
 impl JacByllmConfig.get_call_params_config(self: JacByllmConfig) -> dict[str, any] {
     config = self.load();
     call_params = config.get('call_params', {});
-    temperature = call_params.get('temperature', 0.7);
     max_tokens = call_params.get('max_tokens', 0);
     result: dict[str, any] = {
-        'temperature': temperature,
         'max_output_retries': int(call_params.get('max_output_retries', 3))
     };
+
+    if 'temperature' in call_params {
+        result['temperature'] = call_params['temperature'];
+    }
 
     if max_tokens > 0 {
         result['max_tokens'] = max_tokens;
--- a/jac/jaclang/byllm/llm.impl/basellm.impl.jac
+++ b/jac/jaclang/byllm/llm.impl/basellm.impl.jac
@@ -860,7 +860,7 @@ impl BaseLLM._invoke_react_loop(mt_run: MTRuntime) -> object {
 }
 
 impl BaseLLM.make_model_params(mt_run: MTRuntime) -> dict {
-    default_temp = _call_params_config.get("temperature", 0.7);
+    default_temp = _call_params_config.get("temperature");
     default_max_tokens = _call_params_config.get("max_tokens");
 
     agent_metadata = {};
@@ -957,11 +957,21 @@ impl BaseLLM.make_model_params(mt_run: MTRuntime) -> dict {
         "messages": messages,
         "tools": tools,
         "response_format": mt_run.get_output_schema(),
-        "temperature": self.call_params.get("temperature", default_temp),
-        "max_tokens": self.call_params.get("max_tokens", default_max_tokens),
         "metadata": agent_metadata,
 
     };
+
+    if "temperature" in self.call_params {
+        params["temperature"] = self.call_params["temperature"];
+    } elif default_temp is not None {
+        params["temperature"] = default_temp;
+    }
+
+    if "max_tokens" in self.call_params {
+        params["max_tokens"] = self.call_params["max_tokens"];
+    } elif default_max_tokens is not None {
+        params["max_tokens"] = default_max_tokens;
+    }
 
     if "tool_choice" in mt_run.call_params {
         params["tool_choice"] = mt_run.call_params["tool_choice"];
--- a/jac/jaclang/byllm/plugin_config.jac
+++ b/jac/jaclang/byllm/plugin_config.jac
@@ -50,8 +50,8 @@ class JacByllmPluginConfig {
                     "nested": {
                         "temperature": {
                             "type": "float",
-                            "default": 0.7,
-                            "description": "Model creativity/randomness (0.0-2.0, lower is more deterministic)"
+                            "default": None,
+                            "description": "Model creativity/randomness (0.0-2.0, lower is more deterministic). Unset by default; some models reject non-default values."
                         },
                         "max_tokens": {
                             "type": "int",
--- a/jac/jaclang/byllm/tests/fixtures/llm_params.jac
+++ b/jac/jaclang/byllm/tests/fixtures/llm_params.jac
@@ -17,7 +17,9 @@ def get_wind_speed_report(city: str) -> str by MockLLM(
         "verbose": True
     },
 )(
-    tools=[get_live_wind_speed]
+    tools=[get_live_wind_speed],
+    temperature=0.5,
+    max_tokens=256
 );
 
 glob result = get_wind_speed_report("Puttalam");
--- a/jac/jaclang/byllm/tests/test_byllm.jac
+++ b/jac/jaclang/byllm/tests/test_byllm.jac
@@ -98,6 +98,8 @@ test "params format" {
     for key in required_keys {
         assert key in extracted_dict , f"Missing key: {key}";
     }
+    assert extracted_dict["temperature"] == 0.5 , "Explicit call-site temperature should be passed through";
+    assert extracted_dict["max_tokens"] == 256 , "Explicit call-site max_tokens should be passed through";
 
     add_message = extracted_dict["messages"];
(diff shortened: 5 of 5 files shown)


RandiJay96:
https://github.com/jaseci-labs/jac/pull/7699 fixed with this PR

PR #7702, opened 2026-07-24
fix(jac-scale): deploy status probe falsely fails on TLS-enabled ingresses

udithishanka:
## Problem

`check_deployment_status` (the final gate of a K8s deploy) probes `http://<nlb-hostname>/<path>` and requires a 200. Once `enable_tls` has patched the app ingress with `ssl-redirect` / `force-ssl-redirect`, that probe receives a 308 upgrade to https **on the same host** - and following it can never pass certificate verification, because the certificate is issued for the app's custom domain, not the load balancer hostname. `requests` raises `SSLError` on every attempt, all 48 retries burn (12 minutes at 15s intervals), and the deploy is reported as:

```
Deployment failed for application '<app>': Deployment failed: Application failed to deploy.
```

...even though the rollout is Available and the app is serving fine.

## Reproduction (executed, deterministic)

On a jacBuilder-deployed app with a custom domain (`builder.protomcp.io`, legacy per-app NLB path):

1. First deploy (no TLS yet): probe gets HTTP 200, deploy succeeds in ~6 min (DeployJob timestamps: 351s).
2. `enable_tls` patches the ingress (cert issued, Ready in 35s, app serves 200 over https).
3. Every subsequent redeploy: rollout goes Available in ~2 min, then the probe burns exactly 12 minutes and false-fails. Observed twice back-to-back on independent runs.

Probe behavior confirmed from outside the cluster:

```
$ curl -s -o /dev/null -w '%{http_code}' http://k8s-jba79f62-...elb.us-east-2.amazonaws.com/
308
$ curl -sL http://k8s-jba79f62-...elb.us-east-2.amazonaws.com/ ; echo $?
60   # SSL certificate verification failure following the redirect
```

## Fix

When the probe request dies on `SSLError`, retry the same request with `verify=False`. The probe is a reachability check of the app through the ingress, not a trust decision, and the NLB hostname is never the name the certificate is issued for. Every other path is unchanged: non-TLS apps still require a genuine 200, restart detection and timeout diagnostics are untouched.

## Release notes

- Fixed permanent deploys falsely reporting "Application failed to deploy" (after a 12-minute stall) when redeploying an app whose ingress has HTTPS enabled; the deploy status probe now tolerates the ssl-redirect certificate mismatch on the load balancer hostname.


udithishanka:
Added a regression test (`test_k8s_utils.jac`): `requests.get` is stubbed to raise `SSLError` whenever verification is on and return 200 only with `verify=False`, asserting the probe succeeds via exactly one verified attempt plus one unverified retry. Confirmed non-tautological: it fails against the unfixed probe with the exact production symptom (`certificate verify failed: hostname mismatch` after burning retries) and passes with the fix. Full suite: 40/40.

PR #7703, opened 2026-07-24
feat(jac-scale): authenticate GitHub release lookups with GITHUB_TOKEN, host-gated to api.github.com

kashmithnisakya:
## Summary

Follow-up 2 from the #7682 cleanup: jac-scale's release lookups hit anonymous `api.github.com`, which rate-limits at 60 requests/hour per IP (403/429). CI runners and shared-NAT offices hit this in practice, and a rate-limited lookup during a rollback is the worst possible timing. This forwards `GITHUB_TOKEN` (fallback `GH_TOKEN`) as `Authorization: Bearer <token>`, mirroring `scripts/install.sh`'s `api_curl` exactly.

## Why it was deferred from #7698, and how the gate works

`_http_get_bytes` is shared between the GitHub API lookups, the `github.com` asset downloads (which redirect to `objects.githubusercontent.com`), and the Docker Hub probe. An unconditionally attached header would leak the credential cross-host, and urllib copies `req.headers` onto redirected requests.

The gate closes all of it:

- The header is attached only when the URL's scheme is `https` and its parsed hostname is exactly `api.github.com`. The Docker Hub probe and `github.com` downloads never get a header, so nothing can ride their redirects.
- The header is attached via `Request.add_unredirected_header`, the same mechanism urllib's own auth handlers use: `unredirected_hdrs` are excluded from redirect header-copying, so even a hypothetical cross-host redirect from `api.github.com` itself cannot carry the token. The no-leak guarantee is structural, not dependent on GitHub's redirect topology.
- `urlsplit().hostname` parsing (not substring matching) fails closed on lookalike hosts: `api.github.com.evil.example`, `evil.example/api.github.com`, and the userinfo trick `api.github.com@evil.example` all get no header, covered by tests.

With no token env set, the request pipeline is byte-identical to before (`urlopen` wraps bare URLs in a `Request` internally).

## Tests

Four new tests with a recording `urlopen` fake: token sent to `api.github.com` and nowhere else (asset download and Docker Hub URLs asserted header-free), no header when no env is set, `GITHUB_TOKEN` > `GH_TOKEN` precedence, and the lookalike/userinfo/plain-`http` fail-closed matrix.

## Merge interaction with #7698

`binary.jac` auto-merges cleanly (the edits end above #7698's first hunk). The docs paragraph and the test-file tail are the two textual conflict points; whoever merges second re-places them. Post-merge, #7698's pinned `releases/tags/v<version>` lookups flow through the authed path automatically, and its pinned rate-limit error message is the natural place to mention `GITHUB_TOKEN` (deliberately not touched here since `_release_meta` lives on that branch).


kashmithnisakya:
closing with the #7682 batch. note #7713 does not add api auth, so anonymous release lookups still hit the 60/hr rate limit; this one is standalone and rebases cleanly if wanted later

PR #7704, opened 2026-07-24
fix(jac-scale): experimental deploys stage the dev admin console, matching their dev binary

kashmithnisakya:
## Summary

Follow-up 3 from the #7682 cleanup. `download_release_binary` rewrites `experimental -> dev` before downloading (the experimental channel ships the `dev` binary; pods run the PR's image), but the admin-console path never did: the literal string `"experimental"` fell through `_admin_dist_urls`'s dev/local branches into the release-API branch. Every `[experimental]` deploy therefore paired a **dev binary** with a **latest-stable admin console**, cached under `jac-experimental-admin-dist.tar.gz`.

The fix mirrors the binary path's rewrite at the top of `download_admin_dist`, so one rewrite covers the URL, the cache slot, and the log label together: experimental deploys now fetch `releases/download/dev/jac-dev-admin-dist.tar.gz` and share the `jac-dev-admin-dist.tar.gz` cache slot (content-identical, same dedup the binary path uses).

## Behavior change

`[experimental]` deploys switch from the latest-stable console to the dev console, which matches the binary they run. The stale `jac-experimental-admin-dist.tar.gz` cache entry is simply no longer read; nothing migrates or deletes it. Stable, dev, and local channels are byte-identical to before.

## Test

New regression test drives the real `download_admin_dist("experimental")` through a checksum-serving HTTP fake and asserts the dev URL is fetched, the API is never consulted, and the cache lands in the dev slot, not an experimental one. Fails on the pre-fix code at the API assertion.

## Merge interaction

#7698 touches the same function (adds a `version` parameter); whichever merges second resolves a small conflict in `download_admin_dist`'s opening lines. Semantics compose: the rewrite stays version-free because the resolver never produces a pinned experimental deploy.

### DIFF
--- a/docs/docs/community/release_notes/unreleased/jaclang/7704.bugfix.md
+++ b/docs/docs/community/release_notes/unreleased/jaclang/7704.bugfix.md
@@ -1,0 +1,1 @@
+- **jac-scale: experimental deploys now stage the dev admin console**: the `[experimental]` channel ships the `dev` jac binary, but its admin console was silently resolved from the latest stable release, pairing mismatched versions inside the app bundle. The console now follows the binary to the `dev` release and shares its download cache slot.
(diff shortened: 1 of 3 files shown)


kashmithnisakya:
closing with the #7682 batch. the experimental-channel console mismatch (dev binary + stable console) looks untouched by #7713, so this 4-line fix may be worth reviving standalone

Issue #7707, opened 2026-07-24
jac-scale: redeploy resets the ingress host binding, silently breaking HTTPS on TLS-enabled apps


udithishanka:
## Summary

Redeploying an app whose TLS was enabled via `enable_tls` leaves the app serving nginx's self-signed **"Kubernetes Ingress Controller Fake Certificate"** instead of its valid Let's Encrypt cert. Browsers show `NET::ERR_CERT_AUTHORITY_INVALID` until `enable_tls` is run again. The certificate itself is untouched and Ready the whole time.

## Root cause

`IngressDeployer.enable_tls` patches every ingress rule's `host` to the app domain (plus `spec.tls` + ssl-redirect annotations). ingress-nginx binds a certificate only to a server whose hostname matches, so that host patch is what makes the cert serve.

A subsequent deploy re-renders the ingress in its pre-TLS shape with an **empty rule host** (catch-all). Observed state after redeploy on a live app:

```
host:                                          # <- reset by redeploy
tls: [{"hosts":["builder.protomcp.io"],"secretName":"sdsdfsdfsdf-tls"}]
ssl-redirect: true
```

`spec.tls` and the annotations survive, but with no rule host there is no nginx server for the domain, so the cert has nothing to bind to and the default fake cert serves on 443. Because everything else looks enabled (cert Ready, tls status enabled), nothing signals the breakage - HTTPS is just silently broken after every redeploy.

## Executed reproduction (jacBuilder-deployed app, legacy per-app NLB path)

1. Deploy with a custom domain, run `enable_tls` - cert issues, `https://builder.protomcp.io` serves 200 with the LE cert.
2. Redeploy the same app (rollout succeeds; probe passes post-#7702).
3. `openssl s_client -servername builder.protomcp.io` now returns `subject=O=Acme Co, CN=Kubernetes Ingress Controller Fake Certificate`; browser shows `ERR_CERT_AUTHORITY_INVALID`. The Certificate resource is still `READY True`.
4. Re-running `KubernetesTarget.enable_tls` (no email needed - read back from the `jac-scale/cert-email` ingress annotation) restores the host binding; the LE cert serves again within seconds.

## Suggested fix

The deploy path should not clobber ingress fields `enable_tls` owns: when rendering the ingress for an app whose config/annotations carry a domain (`jac-scale/domain`), keep the rule host set to that domain (and preserve the TLS wiring), so a redeploy leaves HTTPS exactly as `enable_tls` configured it.

[...]

PR #7716, opened 2026-07-24
fix(jac-scale): annotation fallback for the ingress host binding (follow-up to #7708)

kugesan1105:
**Stacked on #7708** — the first five commits are that PR unchanged; only the last commit (`a8549cf69`) is new. Review just that commit; merge after #7708 lands (this then rebases to the one commit).

## Gap in #7708

The dual-rule render keys on `config.domain`. But jac-scale supports the domain living **only** in the `jac-scale/domain` ingress annotation — `enable_tls` itself reads it back to re-run with no arguments (#7707, repro step 4). A redeploy driven without the domain in config (it may only ever have been supplied at `enable_tls` time) still re-rendered a single catch-all rule and silently wiped the host binding — an unfixed instance of #7707.

## Fix

The per-app render resolves its domain exactly the way `enable_tls` already does — explicit `config.domain` first, `jac-scale/domain` annotation as fallback. The fallback read of the live ingress happens **only** when config carries no domain (zero extra API calls for jacBuilder-style deploys, which always pass the domain; shared-ingress paths never read). The resolved domain also feeds the `jac-scale/domain` annotation write and the redis-insight rules.

## Validation

- **Unit**: 6 new tests in `test_ingress_deployer.jac` (annotation fallback for main + redis rules, config-wins-without-reading, no-domain-anywhere stays catch-all, 404/500 read handling). The 3 behavior tests fail against plain #7708 and pass here; all 173 tests in `tests/deploy/` pass.
- **Live E2E (microk8s)**: deploy → enable_tls → redeploy → **domainless redeploy**:
  ```
  STEP3 after-redeploy      host=myapp.e2e.local  tls intact   (base #7708 behavior)
  STEP4 domainless-redeploy host=myapp.e2e.local  tls intact   (this PR - FALLBACK-OK;
                                                   wiped on plain #7708)
  ```
- Two pre-existing no-domain tests taught to 404 the new live-state read (bare MagicMock reads would otherwise coerce to a garbage domain).


MusabMahmoodh (changes requested):
checked it against current `main` before
writing, and the short version is: **the gap you describe is still there, the
fix is still the right shape, and about a third of the diff no longer has a
subject.**

## The gap is unfixed on main

Nothing has closed it in the 27 days. On `upstream/main` today:

- `_deploy_ingress_resource` still keys everything on `self.k8s_config.domain`
  alone: the host rule at `ingress.jac:414-418`, and the annotation write at
  `ingress.jac:459-460`.
- `enable_tls` still resolves the other way, at `ingress.jac:725`:
  `domain = ann.get('jac-scale/domain', '') or self.k8s_config.domain`.

So the asymmetry your PR body describes is intact: the TLS path reads the
annotation, the deploy path writes it and never reads it, and a domainless
redeploy still re-renders a catch-all rule over the host binding. Your
microk8s STEP4 result is still reproducible.

`#7707` was closed as completed by `#7708`, so **this PR is currently the only
record that the follow-up gap exists.** That is the main reason I would rather
see it rebased than dropped.

## The base landed, so the stacking note is stale

`#7708` merged at `2026-07-24T17:48:39Z`, thirteen minutes before this PR
opened. The header instruction ("review just the last commit, merge after
#7708 lands") no longer applies - it is in `main`.

## What a rebase costs

The branch is **318 commits behind** `main` and `git merge-tree` conflicts in
`ingress.jac`. Splitting your diff by whether its target survived:

**Survives, apply as written:**
- `_render_domain` - the new helper is self-contained, reads the live ingress
  only when config carries no domain, and handles 404 the way it should.
- The `main_rules` change and the `jac-scale/domain` annotation write.

**Does not survive - the code is gone:**
- The `redis_rules` block. Redis-insight ingress was deleted from `main` by the
  Postgres clean-break (`768124c78a`, `#8000`/`#8002`), together with the
  `redis_dashboard` and `mongo_dashboard` parameters of
  `_deploy_ingress_resource`. Those hunks have nothing to attach to, and the
  tests covering "annotation fallback for redis rules" go with them.

Same file also moved under you in ways that will show up in the rebase but are
mechanical: `Any` to `any` on the signatures, and `_htpasswd_line` /
`require_optional` removed.

## What I am asking for

Rebase onto current `main` and drop the redis third. What is left is the
helper, the two call sites, and the three tests that still have a subject -
roughly thirty lines. That is a much easier PR to land than the one that has
been sitting here, and it is the part that was always the fix.

If you would rather not carry it after this long, say so and I will file the
gap as a standalone issue crediting your `#7707` and this PR, so it stays
tracked either way. I would rather it stayed yours.

**Rig offer, and it is a real one:** the domainless-redeploy leg is the
awkward one to re-prove after a rebase. I have a kind rig on jac-ec2 that can
run deploy -> enable_tls -> redeploy -> domainless-redeploy and hand you back
the STEP3/STEP4 output for the PR body. Ask and I will run it.

marsninja:

PR #7719, opened 2026-07-24
fix(byllm): retry temperature rejection and content-filter false-positives instead of hard-failing

SupulHeshan:
## Summary

Two related failure modes made byLLM turns die silently or loudly on newer/stricter models, both surfaced by `claude-fable-5` failing every build in jaseci-labs/jacBuilder#896:

- **`temperature` rejection** — models not yet in litellm's `drop_params` registry 400 outright when `temperature` is explicitly set (e.g. Anthropic's `` `temperature` is deprecated for this model ``). `Model.model_call_no_stream`/`model_call_with_stream` (and their async variants) now retry once with `temperature` stripped instead of surfacing the provider's 400.
- **Content-filter false positives on byLLM's own prompt** — the un-dedented system persona and Python-`repr()`-style argument rendering (`question = ('5 + 5?', '')`) were enough to trip some models' content filters. The provider still returns `200 OK` with `finish_reason="content_filter"` and empty content, so byLLM read that as a normal, empty completion and silently returned `''` — no error, nothing in the logs to explain why.

  ```
  # before
  question = ('5 + 5?', '')
  ...
  LLM call completed with response:
  ANSWER REPR: ''

  # after
  question = "5 + 5?"
  ...
  LLM call completed with response: 10
  ANSWER REPR: '10'
  ```

  Fixed by dedenting the system prompt and rendering arguments with `json.dumps` instead of `repr()`, so formatting artifacts stop tripping the filter in the first place. Belt-and-suspenders: a `content_filter` response is now retried automatically (it can be probabilistic), and a still-blocked turn raises a new `ContentFilteredError` instead of returning silently empty.


marsninja:

PR #7721, opened 2026-07-24
refactor(assets): unify static asset serving into one resolver

marsninja:
Static asset serving lived in four places with four different extension lists, and the two servers disagreed on nearly every axis.

## Why

The dev server used its extension set as a **routing discriminator**, not just a servability check. `[client.assets] custom_extensions` replaced the defaults rather than extending them, via `set(...) or {defaults}` (`X or Y` returns `Y` only when `X` is empty, so the "default when unset" reading was accidental). Miss the gate and the request fell past the asset branch into the SPA fallback, which rendered the root app as HTML with status 404. Fonts came back as a 404 whose body was a full web page.

Scale could not hit that, because its routing came from a registered catch-all. It had its own hardwired set plus `.pdf/.txt/.xml` and a blanket `Cache-Control: max-age=31536000`. Beyond the extension lists, the two also diverged on:

| | dev | scale |
|---|---|---|
| root search order | `.jac/client/dist` first | `.jac/client/dist` **last** |
| extra roots | none | `public/`, `src/`, bare project root |
| cache header | `max-age=3600` binary, `no-cache` CSS | `max-age=31536000` root, none under `/static/` |
| 404 body | JSON | `text/plain` |

So a stale `dist/` shadowed fresh build output in production but not in dev, `src/` and any root-level `.json` were downloadable in production only, and neither canonicalized or confined the joined path (no `resolve()`, `is_relative_to`, or `commonpath` anywhere). A fourth list, `AssetProcessor.ASSET_EXTENSIONS`, omitted `.js`/`.json` while both servers served them, so those assets never reached `dist/`.

The root cause is that the two servers share no response type (one writes to `handler.wfile`, the other returns a `fastapi.Response`), so the shared logic had nowhere to live and got copy-pasted. The tell is the CSS special case: a 6-branch `if/elif` ladder duplicated four times, walking exactly the candidates the generic loop 10 lines below already walks.

## What this does

Adds `AssetResolver` in `runtimelib/client/assets.jac` as the single source of truth. `resolve(url, query)` returns an `AssetResolution` value tagged `SERVED` / `NOT_AN_ASSET` / `NOT_FOUND` / `FORBIDDEN`. Each server only translates that value into its own response type.

- **Routing stops sniffing extensions.** A suffixed path that exists under a declared root is served whatever its extension; a miss is a hard `text/plain` 404, never an SPA render. The font failure mode is structurally impossible now, not just fixed.
- **One media table**, consulted instead of the OS mime database, so a URL yields the same `Content-Type` on every platform. `ASSET_COPY_EXTENSIONS` derives from it, closing the `.js`/`.json` gap.
- **Paths are percent-decoded, then confined** with `resolve()` + `is_relative_to`. Dotfiles and bundler sources (`.scss`, `.ts`, `.jac`, ...) are rejected. `src/` and the bare project root are no longer served.
- **Cache follows provenance:** `immutable` for content-hashed names and `?hash=` URLs, `no-cache` plus a strong ETag otherwise, `no-store` for unhashed files on the local server. Adds `Last-Modified`, 304s, `Accept-Ranges` and single-range 206 responses (Safari needs the latter for `<video>`/`<audio>`).
- **`[client.assets]` is schema'd:** ordered `roots`, `extensions.add`/`.remove`, `immutable_max_age`. There is no spelling of the new key that reproduces the old failure.

## Compatibility

`custom_extensions` still reads, is now **additive**, and warns with the new key name. That alone fixes affected projects with no config edit. The basename fallback (`dist/<file>` for `/a/b/<file>`) is preserved on the historical roots as a per-root `flatten` flag, defaulted on there and off for new roots, so existing apps keep resolving.

Two intentional behavior changes: a missed asset now returns `text/plain` instead of JSON on the dev server (a browser fetching an image should not get JSON), and unhashed assets are revalidated rather than pinned for a year in production.

## Removed

`serve_root_asset`, `serve_static_file`, `send_static_file`, both hardcoded extension sets, all four copies of the CSS ladder, and the 127-line dev asset branch (now 9 lines). The `web-static` preview server keeps its own handler but reads the shared media table instead of hand-rolling a `.wasm` case.

## Testing

`jac/tests/runtimelib/client/test_asset_resolver.jac` is table-driven: one set of URLs replayed through the resolver and both adapters, asserting they agree on status, content type, cache directive, ETag and body. This is the part that keeps them from re-diverging. 13 tests, all passing.

Type errors dropped against baseline (`runtimelib/server.jac` 121 to 86; scale unchanged at 453). Targeted suites match baseline exactly: `test_serve` 7F/5P, `test_serve_client` 30F/56P/8S, `scale/tests/server` 3F/253P, `test_client_config` 8/8 across four runs. Pre-existing failures are unrelated (node toolchain, native linking).

One thing to watch: a full-directory `tests/runtimelib/client/` run showed two `test_client_config` failures that do not reproduce in isolation (8/8 four times) or in a three-file run. Likely vite-build contention under `-n 18`, but unconfirmed.

### DIFF
--- a/jac/jaclang/runtimelib/client/assets.jac
+++ b/jac/jaclang/runtimelib/client/assets.jac
@@ -1,0 +1,291 @@
+import re;
+import from enum { StrEnum }
+import from pathlib { Path }
+
+glob ASSET_MEDIA_TYPES: dict[str, str] = {
+         '.avif': 'image/avif',
+         '.css': 'text/css; charset=utf-8',
+         '.eot': 'application/vnd.ms-fontobject',
+         '.gif': 'image/gif',
+         '.htm': 'text/html; charset=utf-8',
+         '.html': 'text/html; charset=utf-8',
+         '.ico': 'image/x-icon',
+         '.jpeg': 'image/jpeg',
+         '.jpg': 'image/jpeg',
+         '.js': 'text/javascript; charset=utf-8',
+         '.json': 'application/json; charset=utf-8',
+         '.map': 'application/json; charset=utf-8',
+         '.mjs': 'text/javascript; charset=utf-8',
+         '.mp3': 'audio/mpeg',
+         '.mp4': 'video/mp4',
+         '.otf': 'font/otf',
+         '.pdf': 'application/pdf',
+         '.png': 'image/png',
+         '.svg': 'image/svg+xml',
+         '.ttf': 'font/ttf',
+         '.txt': 'text/plain; charset=utf-8',
+         '.wasm': 'application/wasm',
+         '.wav': 'audio/wav',
+         '.webm': 'video/webm',
+         '.webp': 'image/webp',
+         '.woff': 'font/woff',
+         '.woff2': 'font/woff2',
+         '.xml': 'application/xml; charset=utf-8'
+     };
+
+glob SOURCE_ONLY_EXTENSIONS: set[str] = {
+         '.jac',
+         '.jsx',
+         '.less',
+         '.py',
+         '.pyc',
+         '.sass',
+         '.scss',
+         '.styl',
+         '.ts',
+         '.tsx'
+     };
+
+glob ASSET_COPY_EXTENSIONS: set[str] = set(ASSET_MEDIA_TYPES.keys()) | {
+         '.less',
+         '.sass',
+         '.scss',
+         '.styl'
+     };
+
+glob RESERVED_URL_PREFIXES: list[str] = [
+         'walker/',
+         'walkers',
+         'function/',
+         'functions',
+         'user/',
+         'protected',
+         'healthz',
+         'graph',
+         'docs',
+         '__build_status',
+         '__jac'
+     ];
+
+glob _HASHED_TOKEN_RE: str = r'[._-]([A-Za-z0-9_]{8,})(?:\.[A-Za-z0-9]+)+$',
+     _CACHE_BUST_KEYS: list[str] = ['hash', 'v', 'version'],
+     _HEX_DIGITS: str = '0123456789abcdefABCDEF';
+
+enum AssetOutcome(StrEnum) {
+    SERVED = 'served',
+    NOT_AN_ASSET = 'not_an_asset',
+    NOT_FOUND = 'not_found',
+    FORBIDDEN = 'forbidden'
+}
+
+enum AssetRootMode(StrEnum) {
+    BUILD_OUTPUT = 'build-output',
+    PUBLIC = 'public',
+    SOURCE_ASSET = 'source-asset'
+}
+
+obj AssetRoot {
+    has path: Path,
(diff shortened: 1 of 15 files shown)


jac/jaclang/runtimelib/client/assets.jac
```diff
@@ -0,0 +1,291 @@
...
+     };
+
+glob ASSET_COPY_EXTENSIONS: set[str] = set(ASSET_MEDIA_TYPES.keys()) | {
+         '.less',
+         '.sass',
+         '.scss',
+         '.styl'
+     };
+
+glob RESERVED_URL_PREFIXES: list[str] = [
```
MusabMahmoodh, line 56: This is a second reserved-path list. `scale/runtime/routing.jac:23` already carries one, and it is *derived* rather than written out:

```jac
_RESERVED_SEGMENTS: set[str] = (
    set(GATEWAY_API_SEGMENTS)
    | set(GATEWAY_WELLKNOWN_SEGMENTS)
    | {GATEWAY_CLIENT_SEGMENT, "api", "admin", "static"}
)
```

The two already disagree - `admin`, `api` and `static` are reserved there and absent here - so a path the gateway treats as its own is a candidate asset to this resolver. Hand-synced duplicate lists, and this one starts out of sync on day one.

Could this import the derived set instead, or is there a reason the resolver's notion of reserved has to be narrower than routing's?

MusabMahmoodh (commented):
Picking this up because it lands in `scale/server/impl/serve.static.impl.jac` and `scale/server/serve.jac`, and it has sat unreviewed since July 24.

**The thing worth raising first: #8175 is rewriting the same two files right now.** Developer-Linus opened it Aug 14 and it is active - it adds `runtimelib/serving/static_assets.jac` as *its* shared module and rewrites `serve.static.impl.jac` and `serve.jac` the way this PR does. So there are two independent implementations of the same consolidation open at once, and they cannot both land. I would rather ask than guess: is this one still the intended landing spot, with #8175 rebased onto it, or has it been superseded and should be closed so the reviewers there stop splitting? I have no stake in which - I just do not think anyone else is positioned to notice the collision, and 26 days is a long time for it to sit.

Same note for my own RFC #8124: the `AssetRoot` / `AssetRootMode(BUILD_OUTPUT | PUBLIC | SOURCE_ASSET)` table with declared precedence here is the shape that RFC argues for, and this PR got there in July, before I wrote it. Saying so explicitly so the sequencing is not misread.

The confinement is right, and it is the part both old paths were missing: `confined_path` resolves both sides and uses `is_relative_to`, and the per-segment gate rejects empty, dot-leading, backslash and NUL segments before any join happens.

CI has been red for 26 days - Contribution Checks, test-scale (server), test-scale (data), test-pypi-build. Worth knowing whether that is rot or real, and it needs a rebase regardless now that #8315 has rewritten 74 files under `scale/`.

PR #7731, opened 2026-07-25
fix(native): diagnose a stale vendored wasm libc instead of blaming the app

marsninja:
## The bug

Any project whose native code reaches `realloc` fails its `na`->wasm build:

```
⚠ Skipped na->wasm for 'main.jac': wasm import '__jac64_realloc' is neither part
  of the jac_host1 host ABI nor declared by the app; the native floor references
  a symbol missing from the vendored wasm_rt libc (compiler drift) -- add it to
  wasm_rt or to wasm_host_abi.jac
```

**Both suggestions are wrong.**

`__jac64_realloc` is already in wasm_rt -- [`jac_abi64.c:31`](jac/jaclang/compiler/passes/native/wasm_rt/jac_abi64.c#L31), added alongside the native cycle collector in #7149.

And `WASM_HOST_ABI` is syscalls only: `write`, `fopen`, `stat`, `clock_gettime`, `exit`. It contains no allocator **by design** -- the allocator operates on the module's own linear memory, so a host-side `realloc` cannot return a valid wasm pointer. Following the advice produces a module that links cleanly and then corrupts memory, which is strictly worse than the current loud failure.

## The actual root cause

A build artifact with no recorded link to its sources.

- `.pbs-build/wasm32/libc` is a **gitignored local cache**.
- `vendor-wasm-libc` ([build.zig:280](jac/build.zig#L280)) is a standalone step nothing depends on, so editing `wasm_rt/*.c` never triggers a rebuild.
- `_wasm_libc_dir()` selects that cache on `os.path.isdir(dev) and os.listdir(dev)` alone -- no hash, no mtime, no version -- **in preference to the complete shipped floor**.

So a checkout whose floor was built before #7149 keeps serving a 12-of-13 adapter set. Measured on my machine:

| | dev `.pbs-build` | shipped floor |
|---|---|---|
| strlen, malloc, calloc, malloc_usable_size, memcpy, memmove, memset, memcmp, memmem, strncmp, strtol, snprintf | yes | yes |
| **realloc** | **MISS** | yes |

Artifact built Jul 11 16:17; `jac_abi64.c` gained the adapter Jul 12 11:10.

**Released binaries are unaffected** -- the shipped floor carries all 13. This only bites contributors with a dev checkout, silently, and it looks like a compiler bug.

## The fix

Three guards, weakest to strongest:

**1. `_stale_dev_libc()` warns** when the dev cache predates `wasm_rt/`. Deliberately a warning, not an error: mtime is a heuristic, and a `git checkout` that rewrites `wasm_rt/` without changing it semantically must not break every wasm build. An earlier revision of this patch made it fatal and turned two green tests red -- that is how the warning/error line got drawn.

**2. `_assert_abi64_adapters()` hard-fails** before the rename, naming exactly what is missing:

```
vendored wasm libc at <dir> is stale: the native floor rewrites realloc to
__jac64_realloc, which the linked bitcode does not define (see _ABI64_ADAPTED
and wasm_rt/jac_abi64.c). The adapters exist in source; the compiled bitcode
predates them -- run `zig build vendor-wasm-libc` to rebuild it.
```

Scoped to the adapters *this module actually renames*, not all of `_ABI64_ADAPTED`, so a module that never calls `realloc` still builds against an otherwise-good floor.

**3. The linker** recognises `__jac64_*` as a floor adapter and says so, instead of offering the `wasm_host_abi.jac` suggestion that can never apply to an allocator.

## Tests

Two added to `test_wasm_build.jac`:

- **adapter-gap diagnosis** -- copies the floor, removes `jac_abi64.bc` (the shape of a pre-adapter build), asserts the error names the floor and **never mentions `wasm_host_abi`**.
- **source-level contract** -- every `_ABI64_ADAPTED` name has a `__jac64_` counterpart in `jac_abi64.c`. Fails the moment the two drift again, with no bitcode required.

## Verification

`jac -m pytest jac/tests/compiler/passes/native/test_wasm_build.jac` -- **2 failed, 5 passed**, the same two failures by name as on `main` (this machine's floor is genuinely stale, which is what surfaced the bug). No previously-green test regressed; both new tests pass.

End to end on a real project with a `realloc`-using native module:

- stale floor -> warning + the precise adapter error above
- complete floor -> `Native wasm module: /static/main.wasm (39213 bytes)`, no warnings

`jac check` error counts on both edited files are unchanged from `main` (58 and 7 -- pre-existing `object`-typed llvm handle noise).


marsninja:
Pushed Tier 0, 1 and 4 into this PR. The first commit made the failure legible; these make it not happen.

## Tier 0 -- the actual root cause

`build.zig` kept the `vendor-wasm-libc` run inside the `link_dir == null` guard that also does the *bundling*:

```zig
if (link_dir == null) {          // link_dir != null  ==  -Ddev
    const vendor_wasm = ...;
    mk.step.dependOn(&vendor_wasm.step);
    mk.addArg(b.fmt("--wasm-libc={s}", .{wasm_libc}));
}
```

So a distributable `zig build` compiled the floor, but **`zig build -Ddev` skipped it entirely** -- and `-Ddev` is the one configuration whose runtime reads `.pbs-build/wasm32/libc` directly. The only mode that consumed the floor was the only mode that never produced it, so it kept serving whatever the last non-dev build happened to leave behind. That is the whole incident.

The step is now hoisted out; only `--wasm-libc=` stays conditional, since a linked binary has no payload to carry the floor in.

## Tier 1 -- a real build-graph node

`addTreeInputs` now declares the `wasm_rt` sources, so Zig can see the step is out of date. This file already documents the exact hazard for the payload step ("a bare dir arg silently never invalidates ... `addFileInput` content-hashes each file") -- the floor step was the one place that lesson had not been applied. `has_side_effects` stays set deliberately: the output lives outside the cache, so a deleted `.pbs-build` must repopulate.

With the producer fixed, **`_stale_dev_libc()` and its mtime helpers are deleted** (-52 lines). That heuristic only existed because nothing rebuilt the floor in dev. A build that keeps it fresh, plus the exact adapter assertion, covers the same ground without guessing from timestamps -- and it removes the false-positive risk I flagged when I first added it.

## Tier 4 -- the coverage hole

`test_wasm_build.jac` **ran nowhere in CI** and `vendor-wasm-libc` was never invoked there. That is why an adapter gap could only be found by a user. It now runs in `test-native-macos`, where `setup-jac` has already built the binary and therefore the floor, so the job doubles as a check that the shipped floor and the source tree agree.

Plus a test that scans the **built bitcode** for every `_ABI64_ADAPTED` adapter. The source-level test in the first commit passes on a stale floor; this one reads the artifact, so it catches a floor compiled from an older tree no matter which symbols a fixture happens to reach.

## Verification

- `zig build -Ddev` now compiles `wasm_rt` -- recompiled a touched `jac_abi64.c`, no-ops on a clean tree, exits 0.
- The rebuilt floor carries all 13 adapters (`realloc` included).
- **Suite went from 2 failed / 5 passed to 8 passed.** Those two failures were the stale floor, not the tests -- fixing the producer turned them green, which is the clearest evidence the diagnosis was right.
- `jac check` on `wasm_build.jac` unchanged at 58 pre-existing errors.
- End to end on a real `realloc`-using project: `Native wasm module: /static/main.wasm (79925 bytes)`, no env override, no warnings.

Tiers 2 (content-addressed floor path) and 3 (generate `jac_abi64.c` from `_ABI64_ADAPTED` instead of hand-syncing three copies) are the "never again" versions and are not in this PR.

marsninja:
Third commit: the source manifest. This closes the residual gap -- drift that keeps the symbol set and changes behaviour.

## One deviation from the plan, and it matters

I did **not** build a whole-tree digest protocol. That design needs two implementations to agree on ordering, separators and framing forever, and any disagreement escalates to "every wasm build fails everywhere" -- worse than the bug it fixes.

Instead the floor carries `SOURCES.sha256`, sha256sum's format, one line per input. Per-file sha256 has no protocol to agree on, so there is nothing to keep in sync and no agreement test to write. **That is what makes fatal severity safe**, and fatal is what covers behavioural drift.

## It went in the place that was already deciding this, badly

`buildWasmLibc` compared **mtimes** to pick what to recompile. It now hashes and writes the manifest. So the same file is the producer's incremental-rebuild key *and* what `wasm_build.jac` reads back -- one artifact serving both sides, not a second staleness system bolted onto the first. Net effect on the producer, all verified:

| change | before | after |
|---|---|---|
| `touch` (mtime bumped, bytes identical) | rebuilt | **no-op** |
| one `.c` edited | rebuilt 1 | rebuilt 1 |
| `prelude.h` edited | rebuilt 107 | rebuilt 107 |
| `README.md` edited | rebuilt 107 | **no-op** (only `.c`/`.h` are inputs now) |

The first row is the false-positive class that forced the earlier warning/error compromise. It is gone at the producer too, not just papered over at the consumer.

## The gap it closes

```
⚠ Skipped na->wasm for 'main.jac': vendored wasm libc at .../wasm32/libc was built
  from different wasm_rt sources than this tree: jac_alloc.c. Linking it would fold
  the old floor into the module and ship behaviour that no longer matches the
  source -- run `zig build vendor-wasm-libc` to rebuild it.
```

That is a behavioural edit to `jac_alloc.c` with an unchanged symbol set -- previously silent past all three guards, now refused by name.

## Scope I deliberately left open

The manifest is authoritative about which files count. Re-deriving that rule consumer-side would be exactly the duplicate-implementation problem this avoids, so a source *added* since the floor was built is not caught here -- it stays the adapter and import checks' problem, and only matters if the new file defines something.

A floor with no manifest is **unverifiable, not wrong**: older artifacts and hand-built floors via `JAC_NATIVE_WASM_LIBC_DIR` keep linking, warning only for a dev floor where it is actionable.

## Verification

- **10 passed.** Two new tests: a floor whose recorded bytes no longer match the tree (refused, names the file), and a floor with no manifest (still links).
- Producer/consumer agree on a fresh floor -- clean build, no warnings, `main.wasm` 79925 bytes.
- `zig build -Ddev` green; `jac check` on `wasm_build.jac` unchanged at 58 pre-existing errors.

PR #7745, opened 2026-07-26
fix(byllm): remove indentation causing claude-fable-5 to silently filter responses

SupulHeshan:
## Summary

`SYSTEM_PERSONA` and `INSTRUCTION_TOOL` in `mtir.impl.jac` were defined as triple-quoted string literals written with source-level indentation for readability. Since neither string was dedented, that indentation — leading spaces on every line, plus a trailing blank/whitespace line — became part of the literal prompt text sent to the model on every `by llm()` call.

## Why this matters

Some LLM providers run safety/content filters that treat oddly indented or padded "instruction" blocks as a red flag, since that kind of formatting is also a common technique for smuggling/obfuscating hidden instructions in prompt-injection attempts. With the stray indentation in place, a plain, benign call like:

```jac
def get_answer(question: str) -> str by llm();
```

could come back with an empty response and a `content_filter` finish reason from the provider — silently, with no exception raised — even though nothing in the actual request was unsafe. This was reproduced consistently while testing against the `fable-5` model: the exact same request succeeded every time once the system prompt's stray whitespace was removed, and failed every time it was present, independent of tool usage, `metadata`, or other request fields.

## Before / After (observed behavior)

Given this trivial call:

```jac
def get_answer(question: str) -> str by llm();

glob answer = get_answer("5 + 5?");

with entry {
    print(f"ANSWER REPR: {answer!r}");
}
```

**Before this fix** — silent failure, no exception, just an empty answer:
```
ANSWER REPR: ''
```

**After this fix** — the same call, same question, returns an actual answer:
```
ANSWER REPR: '10'
```

Nothing about the question or the call changed between the two runs — only the stray whitespace in `SYSTEM_PERSONA` was removed.

## Change

Rebuilt both globals using implicit string concatenation instead of triple-quoted blocks, so the literal string value is exactly the intended clean text with no incidental leading whitespace or trailing blank line:

```jac
glob SYSTEM_PERSONA = (
    "This is a task you must complete by returning only the output.\n"
    "Do not include explanations, code, or extra text—only the result."
),
     INSTRUCTION_TOOL = (
    "Use the tools provided to reach the goal. Call one tool at a time with "
    "proper args—no explanations, no narration. Think step by step, invoking tools "
    "as needed. When done, always call finish_tool(output) to return the final "
    "output. Only use tools."
);
```

And the tool-enabled concatenation site now inserts the separator explicitly rather than relying on the literals' incidental whitespace:

```jac
content=system_content + ("\n\n" + INSTRUCTION_TOOL if tools else ""),
```

No behavioral change intended beyond removing the stray whitespace and keeping the tool-prompt spacing correct — the wording itself is unchanged.


MalithaPrabhashana:
Nice find. I hit this same bug independently and can confirm your fix works: `''` before, `'10'` after, consistently on `claude-fable-5`.

While digging in I collected some evidence on *why* it happens. Sharing it because it affects how much this fix covers.

**Short version: the empty response is not a bug in our parsing. The model is refusing the request, and byllm turns that refusal into `""` silently.** Anthropic documents this behavior here: https://platform.claude.com/docs/en/build-with-claude/refusals-and-fallback

That page is worth a read. Key line from it:

> Claude Fable 5 and Claude Opus 5 include safety classifiers that can decline a request. When that happens, you receive a normal response, not an error, with `stop_reason: "refusal"`.

So we get HTTP 200, no exception, empty content. Exactly the symptom you saw.

## What the provider actually sends back

It is not a content filter in the usual sense. Anthropic returns a normal HTTP 200 with:

```json
{
  "content": [],
  "stop_reason": "refusal",
  "stop_details": {
    "category": "reasoning_extraction",
    "explanation": "This request was blocked as it seems to violate Anthropic's Terms of Service restrictions on reverse engineering or duplicating model outputs..."
  }
}
```

Two things matter here:

- The category is `reasoning_extraction`, a safeguard that applies to **reasoning models**. `fable-5` returns thinking blocks; the others do not, which is why only fable is affected.
- **litellm drops `stop_details` entirely.** byllm only ever sees `finish_reason="content_filter"` and `content=None`. That is why it surfaces as a silent `''`.

Same prompt, 10 runs per model:

| model | refusals |
|---|---|
| claude-fable-5 | 10/10 |
| claude-sonnet-5 | 0/10 |
| claude-opus-4-8 | 0/10 |
| claude-haiku-4-5 | 0/10 |

## How to see the refusal message yourself

byllm cannot show you this today, because litellm strips it before byllm sees it. But the raw body is reachable through a litellm callback. Drop this in before making a `by llm()` call:

```python
import litellm

def show_raw(kwargs, response, start_time, end_time):
    raw = kwargs.get("original_response")
    if raw and "stop_details" in str(raw):
        print("REFUSED BY MODEL ->", raw)

litellm.success_callback = [show_raw]
```

Then run any failing call (a `by llm()` function with no docstring and no `sem`, on `fable-5`, with the old indented persona). You will see the refusal printed with its full explanation, instead of an empty string.

That is the whole story in one line of output: the model refused, and byllm reported it as `''`.

## One correction on the "why"

The PR describes this as filters treating odd indentation as a prompt-injection red flag. The data does not support that. Varying **only** the number of leading spaces, everything else identical (each reproduced 10/10):

| leading spaces | refusals |
|---|---|
| 0 | 0/10 |
| 1 | **10/10** |
| 2 | **0/10** |
| 3 | 10/10 |
| 4 (byllm's) | 10/10 |

One space refuses, two spaces is fine, three refuses again. If the filter were reacting to "suspicious looking formatting" this would be monotonic. It is not. The classifier boundary is just very thin, and byllm's prompt happens to sit right on it.

We also found the **necessary** trigger is the user message, not the system prompt. With no docstring or `sem` string, byllm sends:

```
get_answer

question = ('5 + 5 ?', '')
```

That is a code fragment with no actual request in it. Add a docstring or `sem` and **nothing refuses, at any indentation**.

## What this means for the fix

Your change is correct and worth merging. byllm should not ship accidental source indentation regardless.

The thing to be aware of: it lands on a safe point rather than removing the cause. 0 spaces works, 2 also works, 1 and 3 do not. So any future edit to that string can flip it back silently, with no error.

## Suggested follow-ups

**1. Stop swallowing refusals.** In `basellm.impl.jac`:

```jac
output_content: str = message.get("content") or "";
```

This turns a provider refusal into `""` with no error. Raising something like `LLMRefusalError` when `finish_reason == "content_filter"` and there is no content would have made this a two second diagnosis instead of a hunt. It is model and provider agnostic, and it is the piece that protects you when the boundary shifts.

The `explanation` text is recoverable, by the way: register a `litellm.success_callback` and read `kwargs["original_response"]`, which holds the raw body before litellm strips `stop_details`.

**2. The prompt shape is the real root.** `question = ('5 + 5 ?', '')` is a Python tuple repr of `(value, semstr)`, including a stray `''` when no sem string exists. Making that read as a request instead of code removes the trigger entirely, and then the system prompt whitespace stops mattering.

## One more note

Anthropic's own suggested remedy in that message is "configure a fallback model." I tested it with `fallbacks: "default"` and the beta header: `reasoning_extraction` has no recommended fallback, so the refusal still stands. That route will not help here.

MalithaPrabhashana:
Here is a small experiment you can run yourself that shows this clearly.

I **reverted your fix completely**, so `SYSTEM_PERSONA` still has all its original indentation. Then I ran the same file twice, changing **one line only**.

### Run A: no `sem` string

```jac
import from jaclang.byllm.lib { Model }
import from dotenv { load_dotenv }

glob llm = Model(model_name="anthropic/claude-fable-5");

def get_answer(question: str) -> str by llm();

with entry {
    load_dotenv();
    answer = get_answer("5 + 5 ?");
    print(f"ANSWER REPR: {answer!r}");
}
```

```
ANSWER REPR: ''
```

### Run B: identical, plus one `sem` line

```jac
def get_answer(question: str) -> str by llm();
sem get_answer = "Given a question, return the answer in a single line. If the question is not answerable, return 'I don't know.'";
```

```
ANSWER REPR: '10'
```

Same indented system prompt in both runs. Fix not applied in either. The only difference is that one `sem` line.

## What this tells us

The indentation is **not** what the model is reacting to. What actually triggers the refusal is the *user* message byllm builds.

With no docstring and no `sem`, byllm has no description to work with, so it falls back to the bare function name and sends this as the user message:

```
get_answer

question = ('5 + 5 ?', '')
```

That is a code fragment. There is no question in it, nothing is being asked. Add a `sem` (or a docstring) and the user message starts with a real sentence instead, and the model answers normally, indentation and all.

## Why your fix still works

Both things move the request off the classifier's boundary, just from different sides:

- **Your fix** cleans up the system prompt.
- **A `sem` / docstring** makes the user message read like a request.

Either one is enough on its own. That is why you saw a clean before/after, and it is a real improvement worth merging.

The reason I am flagging it: dedenting works, but it is not removing the cause. In my testing, 0 spaces works, 2 spaces also works, but 1 and 3 spaces both fail. So there is no rule making 0 safe, it just happens to land on the safe side. If someone later rewords `SYSTEM_PERSONA`, or a user sets their own `system_prompt`, it can silently break again.

## Direction

Two things that would make this properly robust:

1. **Fix the user message shape.** A `by llm()` function with no docstring or `sem` should still send something that reads as a request, not a bare function name plus a Python tuple repr. That removes the trigger entirely, and then the system prompt formatting stops mattering.

2. **Never return `""` on a refusal.** Right now `message.get("content") or ""` in `basellm.impl.jac` hides the problem. If it raised instead, this whole thing would have been obvious immediately rather than looking like a broken model.

SupulHeshan:
Thanks for digging into this @MalithaPrabhashana - agreed on the main point:  the empty string is a refusal we're hiding, not a parsing bug. I checked all three claims against the code and they hold. Note there are two more copies of the content or "" swallow, in the async and streaming-with-tools paths.

One thing to add: `_structured_output_ok` returns `True` immediately for `str`, which is why only `-> str` shows `''`. A typed return instead burns all three typed retries against a refusal that can never succeed, then fails with a misleading "non-conforming value" error. Same cause, and arguably the worse symptom.

That makes your 1 comment more than diagnostics. We already ship `ModelPool(strategy="fallback")`, which fires on exceptions from the router - today a refusal is a successful 200, so the pool sails past it. Raising is what makes our own fallback able to rescue a refusal at all.

Small correction: Anthropic's `fallbacks` is a Claude-API-only Messages API parameter, and we go through `litellm`, so we likely never pass it through regardless of category. Your conclusion stands, just for a simpler reason.

When we build the raise, I'd match a set (`content_filter`, `refusal`, plus empty-content-and-no-tool-calls) rather than `content_filter` alone — that's litellm's current mapping and it could change.

Plan: keep this PR as-is and note in the description that it moves the prompt off the boundary rather than removing the cause. Then two follow-ups — `LLMRefusalError` raised from all three dispatch sites, and the user-message shape fix (that one changes every prompt, so it needs its own PR).

What do you think about this?

marsninja:

PR #7762, opened 2026-07-28
fix(client): always inject core client runtime deps into the effective npm dep set

### LINKED ISSUE #7323: jac-client: OPTIONAL_RUNTIME_DEPS imported by client_runtime.js but not installed -> client build fails
## Summary

A fullstack app whose client bundle imports the jac-client runtime's `OPTIONAL_RUNTIME_DEPS` (`@tanstack/react-form`, `zod`) but does not declare them in `jac.toml [dependencies.npm]` **fails the client build** and aborts `jac build` / the scale deploy. The generated `.jac/client/compiled/client_runtime.js` imports `@tanstack/react-form`, but `jac install` only installs those optional deps when the app itself lists them, so Rollup cannot resolve the import.

## Environment

| | |
|---|---|
| jac | 0.31.0 (stable release binary), linux x86_64 |
| App | a fullstack app (`has-client` + `has-native`) whose `jac.toml [dependencies.npm]` declares `zod` but not `@tanstack/react-form` |

## Reproduction (executed)

App `jac.toml` lists `zod` under `[dependencies.npm]` but not `@tanstack/react-form`.

```
$ jac --version            # Version: 0.31.0
$ jac build main.jac
  ✔ Gate passed
  Sealing ...
  Building client bundle (web)...
  ⏳ Installing dependencies...
  ✔ Dependencies installed (6.9s)
  ⏳ Building client bundle...
✖ Error: Client build failed: Vite build failed after 1.0s
✗ Build failed in 171ms
error during build:
[vite]: Rollup failed to resolve import "@tanstack/react-form" from ".jac/client/compiled/client_runtime.js".
```

Add the one missing dep and nothing else:

```toml
[dependencies.npm]
"@tanstack/react-form" = "^1.33.0"
```

```
$ jac build main.jac
  ✔ Client bundle built (18.2s)
Done! Compiled 10/10 modules (0 reused, 0 failed)
  ✔ Built dist/<app>.jab (1,572,849 bytes)
```

So the single undeclared optional dep is the whole failure.

## Expected

If the client runtime emits an import for `@tanstack/react-form` / `zod`, `jac install` should install it, so the client build resolves it without the app author having to hand-list a runtime-internal dependency.

## Actual

`@tanstack/react-form` + `zod` are declared as **optional** and are not installed unless the app opts in, yet the generated client runtime imports them:

- `jac/jaclang/runtimelib/client/npm_provider.jac:18-20` - `OPTIONAL_RUNTIME_DEPS = {"@tanstack/react-form": "^1.33.0", "zod": "^4.3.6"}`
- `jac/jaclang/runtimelib/client/impl/client_deps.impl.jac` - `_default_client_deps()` also lists `@tanstack/react-form`
- `.jac/client/compiled/client_runtime.js` imports `@tanstack/react-form` (the Rollup error above)

Note `npm_provider.jac:135` already merges `{**CORE_RUNTIME_DEPS, **OPTIONAL_RUNTIME_DEPS}` into `core_runtime` in one path, so the "optional" classification is inconsistent with the runtime unconditionally importing them.

## Impact

Every fullstack app that uses the runtime surface which pulls in `@tanstack/react-form` (forms) or `zod` (validation) but does not manually declare it fails to build and therefore to deploy. This is what broke jacBuilder / jachammer user-app deploys: a user app imported without that dep aborts the deploy. The failure is easy to misread because the scale deploy truncates the error and can surface the seal step instead of the client build.

## Workaround (verified)

- Add `"@tanstack/react-form" = "^1.33.0"` to the app's `[dependencies.npm]` (verified above: build then succeeds and emits the `.jab`).
- jacBuilder now auto-injects the runtime's `OPTIONAL_RUNTIME_DEPS` into every staged user app at deploy time (jaseci-labs/jacBuilder#578) so direct users of the IDE do not hit it. Direct `jac build` / `jac start --scale` users still do.

## Suggested fix

Install `OPTIONAL_RUNTIME_DEPS` whenever the client runtime emits their import (tie the npm-install set to what `client_runtime.js` actually imports), or promote them to `CORE_RUNTIME_DEPS`, or only emit the import when the dep is present. A build-time check that every import in `client_runtime.js` has a resolvable npm dep would have caught this.

kashmithnisakya:
## Summary

Fixes #7323.

The generated client runtime unconditionally imports `@tanstack/react-form` and `zod` (`client_runtime.cl.jac:15-16`), but injection of the core/optional runtime deps into the generated `.jac/client/configs/package.json` was gated on `'client' in jac_config.plugins`, which is populated only from a literal `[client]` table in `jac.toml`. The client build itself activates from the compiled manifest's `has_client` flag, independently of the toml, so a fullstack app with client code but no `[client]` table built a bundle whose imports were never installed and failed with:

```
[vite]: Rollup failed to resolve import "@tanstack/react-form" from ".jac/client/compiled/client_runtime.js"
```

## Change

Move the two `_inject_missing` calls in `JacClientConfig.load()` out of the `if 'client' in jac_config.plugins` block so the core runtime/dev dep closure is always part of the effective in-memory dependency set. The legacy meta-package migration (the only path that writes `jac.toml`) stays inside the gate, so the #6164 invariant is preserved: reading client config never mutates `jac.toml`. That invariant is still asserted byte-for-byte in the updated test.

The gate dates from #6164, when `load()` could write deps back to the manifest; since the #7237 providers refactor the injection is a pure in-memory merge, so gating it no longer protects anything and only starves the bundler of deps the emitted runtime requires.

## Tests

- Rewrote test 1 of `test_config_dep_injection.jac`: a no-`[client]` project now asserts `react`, `@tanstack/react-form`, `zod`, and `vite` are present in the effective package config while `jac.toml` stays byte-identical (this doubles as the #7323 regression test).
- `jac test jac/tests/runtimelib/client/test_config_dep_injection.jac`: 4 passed.

Screen Shot 2026-07-28 at 7 48 25 PM


kugesan1105:
Validated against a live repro (client code, `zod` declared, no `[client]` table): fails on main with the exact #7323 error, builds with this PR merged onto main — the generated package.json carries `@tanstack/react-form`. The ungating rationale checks out: post-#7237 the injection is a pure in-memory merge, so the #6164 gate only starved the bundler.

Noting for the record: the `[client]`-table gate still exists in `NpmProvider.resolve()` (affects `jac install` / `jac eject`), which #7784 addresses — so merge order matters: this first, then #7769/#7784.

PR #7769, opened 2026-07-28
feat(client): fingerprint npm installs and preflight client runtime imports

kashmithnisakya:
## Summary

Hardening follow-up to #7762 for #7323, closing the two failure modes that remain after core deps are always injected:

1. **Stale `node_modules` is never healed.** `install_packages` early-returned whenever `node_modules/.bin/vite` existed, without ever comparing against the regenerated `configs/package.json`. Any dependency-set change (new injected dep, version bump, edited jac.toml) after the first install was silently ignored, resurfacing the Rollup unresolved-import failure.
2. **Missing deps fail with a cryptic Rollup trace.** When the compiled client runtime imports a package that is not installed, the failure surfaces as a Rollup resolve stack trace that deploy tooling truncates, which is what originally misdirected the jacBuilder debugging in #7323.

## Changes

- `install_packages` computes a sha256 fingerprint of `configs/package.json` + `configs/.npmrc`, stamps it into `node_modules/.jac-package-hash` after a successful `bun install`, and short-circuits only when the stamp matches. A stampless or mismatched `node_modules` reinstalls (bun is incremental, so an in-sync tree is a fast no-op). The existing missing-`.bin/vite` clean-stale wipe is unchanged, and `bun` is now only required when an install actually runs.
- New `ViteBundler._preflight_runtime_imports`, called in `build()` right after installation: scans the compiler-emitted `compiled/client_runtime.js` / `client_runtime_core.js` for bare import specifiers and raises `ClientBundleError` naming the missing package(s) with the remedy (`jac install` / `jac add --npm <pkg>`). Only compiler-emitted files are scanned; user-authored .js copied into `compiled/` is deliberately out of scope to avoid false positives blocking builds.
- Bundler installs deliberately do NOT honor `JAC_CLIENT_SKIP_NPM_INSTALL`. An earlier revision of this branch made `install_packages` skip when the var is set, and CI caught the mistake: the var's established meaning is 'skip npm during provisioning' (`jac install`, `jac create --skip`), and scaffolded projects rely on the first `jac build` installing. (`test_create_kinds.jac` also leaks the var process-wide into sibling tests, which is how the littlex/wasm e2e builds failed.) The preflight's no-`node_modules` error still names the var when it is set, so scale pods that skipped provisioning get an actionable message instead of a truncated trace.

## Tests

New `jac/tests/runtimelib/client/test_install_fingerprint.jac`:
- missing stamp installs and writes the stamp; matching stamp short-circuits; changed package.json reinstalls
- preflight names the missing package with the `jac add --npm` remedy, does not flag installed packages, and passes once the package is present
- missing `node_modules` raises the actionable error, including the env-var hint when set

Verified locally with `JAC_CLIENT_SKIP_NPM_INSTALL=1` exported to reproduce the CI worker state: all tests pass. `test_cli.jac` 27/28 locally with the one failure (`start dev with client does initial compilation`) reproducing identically on clean `main`, pre-existing local-environment flake, unrelated to this change.


MusabMahmoodh (commented):
The preflight is a good idea and I like that you wrote down why the earlier revision honouring `JAC_CLIENT_SKIP_NPM_INSTALL` was wrong. That note is what stops someone re-adding it in six months.

The fingerprint half is already on main though. `client_deps_stale()` landed on 2026-08-06 in #7927, same file, stamping `node_modules/.jac-deps-hash` with a sha256 of `configs/package.json` and gating the early return exactly this way. This branch is based on pre-#7927 main and hasn't moved since 2026-07-30, so on rebase we end up with two stamp files and two staleness rules answering the same question.

Two things your version does better and I'd keep: it hashes `configs/.npmrc` as well, and it calls `create_package_json()` before fingerprinting rather than treating a missing manifest as not-stale. Both fold into the existing helper in a few lines, and then `_package_fingerprint`, `_read_stamp` and `.jac-package-hash` can go. What's left is the preflight, which has no equivalent on main.

On the tests: they do fail on main, so they're not theatre, but nothing here ever runs a real install or looks at a real `node_modules`, and the fingerprint one patches `subprocess.run` process-wide. The defect in #7323 is a `jac build` that dies in Rollup, and `test_littlex_e2e.jac` in the same directory already boots a real project. One leg there, a fixture whose generated package.json is missing a package the compiled runtime imports, asserting the build fails naming the package instead of dumping a Rollup trace, is the test that proves the error a user actually sees.

Smaller one: `_preflight_runtime_imports` returns silently when neither `client_runtime.js` nor `client_runtime_core.js` is in `compiled/`. By the time `build()` calls it the compiler has run, so isn't the empty case "the compiler emitted nothing under the name we expect", which is the case where skipping quietly is worst? Is there a real config where neither file exists?

I did check whether the vite aliases could make the preflight fail a build that works today (`react` under the preact backend, `react-native` to `react-native-web`) and couldn't get one to fire, since `config_loader` injects the core deps regardless. So no concern there.

jac/jaclang/runtimelib/client/impl/vite_bundler.impl.jac
```diff
@@ -13,6 +13,31 @@ def _require_bun -> str {
 }
 
 
+def _package_fingerprint(package_json: Path, npmrc: Path) -> str {
```
MusabMahmoodh, line 16: This is already on main. `client_deps_stale()` landed 2026-08-06 in #7927, same file (currently ~:175), stamping `node_modules/.jac-deps-hash` with a sha256 of `configs/package.json` and gating the early return exactly this way. This branch is based on pre-#7927 main and hasn't moved since 2026-07-30, so on rebase we get two stamp files and two staleness rules for one question.

Two things yours does better and I'd keep: hashing `configs/.npmrc`, and calling `create_package_json()` before fingerprinting instead of treating a missing manifest as not-stale. Both fold into the existing helper in a few lines, then `_package_fingerprint`, `_read_stamp` and `.jac-package-hash` can go and what's left is the preflight, which has no equivalent on main.

jac/jaclang/runtimelib/client/impl/vite_bundler.impl.jac
```diff
@@ -258,6 +284,71 @@ impl ViteBundler.install_packages(
...
+    runtime_files = [
+        compiled_dir / 'client_runtime.js',
+        compiled_dir / 'client_runtime_core.js'
+    ];
+    present = [
+        f
+        for f in runtime_files
+        if f.exists()
+    ];
+    if not present {
```
MusabMahmoodh, line 302: This makes the whole check a no-op when neither runtime file is in `compiled/`. By the time `build()` calls it the compiler has run, so the empty case reads to me as "the compiler emitted nothing under the name we expect" - the case where skipping quietly is worst. Is there a real configuration where neither file exists, or should this raise?

jac/jaclang/runtimelib/client/impl/vite_bundler.impl.jac
```diff
@@ -239,6 +262,9 @@ impl ViteBundler.install_packages(
             raise ClientBundleError("Failed to install dependencies") from None;
         }
         console.print(f"\n  ✔ Dependencies installed ({elapsed:.1f}s)");
+        if node_modules.exists() {
+            stamp_file.write_text(fingerprint, encoding='utf-8');
```
MusabMahmoodh, line 266: If `bun install` returns 0 and `node_modules` isn't there, we skip the stamp silently and the preflight two lines later raises the no-`node_modules` error, so the user gets "run 'jac install'" for an install that just claimed success. Small, but the honest message is that bun produced no `node_modules`.

jac/tests/runtimelib/client/test_install_fingerprint.jac
```diff
@@ -0,0 +1,140 @@
...
+
+test "install packages fingerprint gates reinstalls" {
+    import from jaclang.runtimelib.client.vite_bundler { ViteBundler }
+
+    with tempfile.TemporaryDirectory() as temp_dir {
+        build_dir = _make_build_dir(temp_dir, {"react": "^18.2.0"});
+        node_modules = _fake_node_modules(build_dir);
+        bundler = ViteBundler(Path(temp_dir));
+        ok_result = type("Result", (), {"returncode": 0});
+        with unittest.mock.patch(
```
MusabMahmoodh, line 42: These do fail on main (pre-fix the first call short-circuits and `call_count` stays 0), so they're not theatre - but patching `subprocess.run` process-wide means nothing here ever runs a real install or looks at a real `node_modules`.

The defect in #7323 is a `jac build` dying in Rollup, and `test_littlex_e2e.jac` in this directory already boots a real project. One leg there - fixture whose generated package.json is missing a package the compiled runtime imports, asserting the build fails naming it rather than dumping a Rollup trace - is what proves the error a user actually sees.

MusabMahmoodh (commented):
Preflight is a good idea, and writing down why the earlier revision honouring `JAC_CLIENT_SKIP_NPM_INSTALL` was wrong is the note that stops someone re-adding it in six months. Main thing is the rebase, inline below.

marsninja:

PR #7771, opened 2026-07-28
test: restore JAC_CLIENT_SKIP_NPM_INSTALL after scaffolding in test_create_kinds

kashmithnisakya:
## Summary

`_scaffold` in `test_create_kinds.jac` sets `os.environ["JAC_CLIENT_SKIP_NPM_INSTALL"] = "1"` and never restores it. Under pytest-xdist, one worker process runs many test files in sequence, so once this file runs, every later test in that worker (and every subprocess it spawns with an environment derived from `os.environ`, e.g. the littlex e2e builds) inherits the variable.

Today the leak is invisible because only `jac install`'s npm closure honors the variable, but it makes test outcomes depend on worker scheduling and silently constrains any future code that reads it: an early revision of #7769 made the bundler honor the variable and three CI jobs failed from this leak rather than from the change under test.

## Change

Save the prior value and restore it in a `try/finally` (also covering the `os.chdir` restore), the same save-and-restore pattern `jac create --skip` itself uses in `project.impl.jac`.

## Tests

`jac test jac/tests/runtimelib/client/test_create_kinds.jac`: 3 passed; scaffolds still skip npm install during `project.create` and the variable no longer outlives the helper.

### DIFF
--- a/jac/tests/runtimelib/client/test_create_kinds.jac
+++ b/jac/tests/runtimelib/client/test_create_kinds.jac
@@ -18,13 +18,22 @@ glob _tmp_dirs: list[str] = [],
      _orig_cwd: str = os.getcwd();
 
 def _scaffold(kind: str) -> Path {
+    prev_skip = os.environ.get("JAC_CLIENT_SKIP_NPM_INSTALL");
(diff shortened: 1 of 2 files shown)


marsninja:

PR #7784, opened 2026-07-30
feat(client): framework-keyed form deps and compiled-dir client signal for resolve

kashmithnisakya:
## Summary

Phase 3 of the #7323 remediation, extending the dependency fix to the paths #7762 and #7769 left out. **Stacked on #7762** (its commits are included in this branch); review the last two commits, and merge #7762 first.

Three problems addressed:

1. **Solid apps fail exactly like #7323.** `solid_runtime.cl.jac` unconditionally imports `@tanstack/solid-form`, which was in neither dep constant, so even unconditional injection never provided it.
2. **`jac install` and `jac eject` still used the `[client]`-table gate.** `NpmProvider.resolve()` emitted no self-heal deps for no-`[client]` projects, so ejected frontends were missing the runtime closure and installs could not heal such projects.
3. **Version pins were triplicated** across `npm_provider.jac`, `_default_client_deps()`, and the prompt defaults, guaranteeing eventual drift.

## Changes

- `OPTIONAL_RUNTIME_DEPS` is replaced by a framework-keyed `FRAMEWORK_FORM_DEPS` map (react/preact get `@tanstack/react-form` + `zod`, solid gets `@tanstack/solid-form` + `zod`) with `form_deps_for_framework()` / `resolve_client_framework_name()` helpers. Injection in `JacClientConfig.load()`, the legacy meta-package migration, and `NpmProvider.resolve()` all pick the entry matching the project's `[client] framework`.
- `NpmProvider.resolve()` widens `is_client` to `'client' in config.plugins or <client_dir>/compiled exists`. The compiled dir is the compiler-derived signal that the project has client code, so `jac install` and `jac eject` now emit the full closure for no-`[client]` projects that have built, while genuinely clientless backend projects that merely declare an npm dep still resolve to the manifest slice only (guardrail asserted in tests).
- `_default_client_deps()` now sources its version literals from the `npm_provider` constants, ending the triplication. (Note: this aligns the react prompt defaults with the canonical pins, e.g. `@vitejs/plugin-react` `^4.2.1` -> `4.7.0`, which is the point of a single source.)

## Housekeeping

- The branch deletes `release_notes/unreleased/jaclang/7762.bugfix.md` from its own diff only because the Contribution Checks job forbids a PR touching a fragment that is not its own number and the stacked #7762 commits carry that file; the deletion commit is dropped during the post-#7762 rebase, so #7762's fragment is unaffected. (No fork permissions for the `skip-release-notes-check` label.)
- Rebased onto current `main`; `test_cli.jac` picked up the updated `jac fmt` line-wrapping on the shared Phase 1 commits.

## Tests

- New: a `[client] framework = "solid"` project gets `@tanstack/solid-form` injected (not `@tanstack/react-form`), with jac.toml byte-identical, fails at HEAD today.
- Extended: after creating `.jac/client/compiled`, a no-`[client]` project's resolve plan carries `self-heal:client-core` entries including the form dep; without the compiled dir it still resolves to `[]`.
- `jac test jac/tests/runtimelib/client/test_config_dep_injection.jac`: 5 passed.


marsninja:

PR #7789, opened 2026-07-30
docs(jac-scale): add scheduler and background jobs guide

kashmithnisakya:
## Summary

Adds a user-facing guide for the scheduler at `reference/plugins/jac-scale-scheduler.md`, linked in the nav between Data & Storage and Kubernetes & Operations. Until now the scheduler had no docs page; only a one-line mention in the Scale overview.

The guide covers:

- Quick start: `[scale.scheduler]` intent in `jac.toml`, `jac install`, `jac start`, and the startup log lines to expect
- The `@schedule` decorator on walkers and functions, STATIC vs DYNAMIC, and the three trigger types (interval, cron, date)
- Cron format reference, including the day-of-week numbering (0 = Monday) since it differs from classic Unix cron
- The full `/jobs` REST API with curl examples: auth token flow, create/list/get/update/delete, pagination and filters
- Ownership and execution identity (users manage their own jobs, dynamic jobs run on the creator's root, static tasks run as system)
- Persistence (in-memory vs Mongo job store), run history fields, `[scale.scheduler]` config keys, and an error code reference

## Verification

Every example was run against a live server built from this repo (demo project with static interval/cron tasks and DYNAMIC targets). All response JSON in the doc is captured from those runs, including the 400/401 error shapes. The static interval task was confirmed firing (12 runs in ~60s at interval=5).


MusabMahmoodh (commented):
Structure and the API walkthrough are good, and I checked the mechanical claims against the code: the five endpoints, all five error codes, and the UTC framing all match (`BackgroundScheduler(timezone=utc)` at `scheduler/impl/scheduler.impl.jac:374`). Calling out that day-of-week is 0 = Monday unlike classic cron is the most useful line on the page, and it's right: `apscheduler_kwargs` passes the field straight through to APScheduler.

The persistence chapter is the problem. `MONGODB_URI` doesn't appear anywhere in the repo. The store is picked in `JacScaleScheduler.setup` (`scheduler/impl/scheduler.impl.jac:341-356`): `open_docs_store()` succeeds and you get `PgJobStore`, writing to the `jac_docs` table with `collection` as a column; it fails and you get `MemoryJobStore`. What decides it is `[scale.database]` url / `JAC_DB_URL`. That makes four places wrong the same way: both bullets under Persistence, the `collection` row ("MongoDB collection"), and the 503 row ("The MongoDB job store"). Worth rewriting rather than swapping the word, since the knob the reader has to set changes too.

Second one: `enabled` isn't a runtime switch. `_setup_scheduler()` is called unconditionally on server boot (`serve.core.impl.jac:287`) and registers `/jobs` and the static tasks regardless; `enabled` is an install-time capability flag (`capabilities.jac:96`) so `jac install` pulls apscheduler, and its default is `None`, not `false`. As written someone sets `enabled = false`, expects the scheduler off, and still gets static tasks firing.

Third, and this is the one that bites in production: with no reachable database `_acquire_fire_lease` returns True unconditionally unless the store is `PgJobStore` (`scheduler.impl.jac:724-728`), so every replica fires the same job on every tick. The code's own warning says it out loud. One sentence next to the in-memory bullet, no database means duplicate runs per replica and not just non-durable jobs, is the difference between a dev caveat and a production one.

Two smaller things. The config table is missing `system_user_password` (default `__no_login__`, and static tasks run as that user) and `user_exists_ttl`, both read in `get_scheduler_config()`. And the error table promises 400 for an invalid trigger spec, but `TriggerSpec.validate` only checks that `date` is non-empty, so `"date": "tomorrow"` is accepted at POST and fails later at schedule time. Which did you hit when you ran the examples?

jac/jaclang/cli/docs/reference/plugins/jac-scale-scheduler.md
```diff
@@ -0,0 +1,351 @@
...
+| `401` | `UNAUTHORIZED` | Missing or invalid `Authorization` header |
+| `404` | `NOT_FOUND` | Unknown walker/function name, or job id that does not exist (or belongs to another user) |
+| `503` | `STORAGE_UNAVAILABLE` | The MongoDB job store is temporarily unreachable |
+| `500` | `SCHEDULER_ERROR` | Unexpected internal failure |
+
+## Persistence and Run History
+
+Where jobs live depends on your database configuration:
+
+- **No `MONGODB_URI`**: jobs are held in memory. Everything works, but dynamic jobs disappear when the server restarts. Fine for local development.
```
MusabMahmoodh: `MONGODB_URI` doesn't appear anywhere in the repo. The store is picked in `JacScaleScheduler.setup` (`scale/scheduler/impl/scheduler.impl.jac:341-356`): `open_docs_store()` succeeds and you get `PgJobStore` writing to the `jac_docs` table with `collection` as a column; it fails and you get `MemoryJobStore`. What decides it is `[scale.database]` url / `JAC_DB_URL`.

Worth rewriting rather than swapping the word, since the knob the reader has to set changes too.

jac/jaclang/cli/docs/reference/plugins/jac-scale-scheduler.md
```diff
@@ -0,0 +1,351 @@
...
+- **`MONGODB_URI` set** (or provisioned automatically by `--scale` on Kubernetes): jobs are persisted to the `scheduled_jobs` collection, survive restarts, and are re-registered on boot.
+
+Each execution updates the job record with run bookkeeping, visible via `GET /jobs/{job_id}`:
```
MusabMahmoodh, line 325: The bigger production trap isn't durability, it's duplicate runs: `_acquire_fire_lease` returns True unconditionally unless the store is `PgJobStore` (`scheduler.impl.jac:724-728`), so with no database every replica fires the same job on every tick. The code's own warning says exactly that. One sentence here turns this from a dev caveat into the thing someone needs before they run it on more than one pod.

jac/jaclang/cli/docs/reference/plugins/jac-scale-scheduler.md
```diff
@@ -0,0 +1,351 @@
...
+enabled = true
+collection = "scheduled_jobs"
+thread_pool_size = 10
+misfire_grace_time = 60
+shutdown_timeout = 10
+```
+
+| Key | Default | Meaning |
+|-----|---------|---------|
+| `enabled` | `false` | Turns the scheduler subsystem on and makes `jac install` resolve its dependencies (`apscheduler`) |
```
MusabMahmoodh: Two things off. The default is `None`, not `false` (`plugin_config.jac:1062`), and nothing reads it at runtime: `_setup_scheduler()` is called unconditionally on boot (`serve.core.impl.jac:287`) and registers `/jobs` and the static tasks regardless. It's an install-time capability flag (`capabilities.jac:96`) so `jac install` pulls apscheduler.

As written, someone sets `enabled = false`, expects the scheduler off, and still gets static tasks firing.

jac/jaclang/cli/docs/reference/plugins/jac-scale-scheduler.md
```diff
@@ -0,0 +1,351 @@
...
+| `collection` | `"scheduled_jobs"` | MongoDB collection used for the job store |
+| `thread_pool_size` | `10` | Worker threads available for concurrently firing jobs |
+| `misfire_grace_time` | `60` | Seconds a late job may still fire after its scheduled time (for example after a restart) before the run is skipped |
+| `shutdown_timeout` | `10` | Seconds to wait for in-flight jobs when the server stops |
```
MusabMahmoodh, line 353: `get_scheduler_config()` (`config_loader.impl.jac:458-470`) also reads `system_user_password` (default `__no_login__`, and static tasks run as that user) and `user_exists_ttl`. The password one in particular is the kind of key someone should know exists before they find out the hard way.

jac/jaclang/cli/docs/reference/plugins/jac-scale-scheduler.md
```diff
@@ -0,0 +1,351 @@
...
+    }
+}
+```
+
+| Argument | Type | Meaning |
+|----------|------|---------|
+| `trigger` | `ScheduleTrigger.STATIC` or `ScheduleTrigger.DYNAMIC` | `STATIC` starts running as soon as the server boots. `DYNAMIC` marks the target as schedulable through the `/jobs` REST API. |
+| `interval` | `float` | Seconds between runs. |
+| `cron` | `str` | 5-field cron expression, evaluated in UTC. |
+| `date` | `str` | One-shot fire time, `"YYYY-MM-DD HH:MM:SS"` in UTC. |
```
MusabMahmoodh: The error table promises 400 for an invalid trigger spec, but `TriggerSpec.validate` only checks that `date` is non-empty (`scheduler.impl.jac:35-39`) - no format parsing, unlike the cron branch which round-trips through `from_crontab`. So `"date": "tomorrow"` is accepted at POST and fails later at fire time. Which did you get when you ran the examples? If it's the latter, validate is probably the place to fix it and the doc is already right.

MusabMahmoodh (commented):
Structure and the API walkthrough are good, and the mechanical claims check out against the code: the five endpoints, all five error codes, and the UTC framing (`BackgroundScheduler(timezone=utc)`, `scheduler.impl.jac:374`). The 0 = Monday note is right too, and it's the most useful line on the page. The persistence chapter is the problem, inline below.

jac/jaclang/cli/docs/reference/plugins/jac-scale-scheduler.md
kashmithnisakya: rewrote the persistence chapter around the `[scale.database]` url / `JAC_DB_URL` knob and fixed the collection and 503 rows, no mongo mentions left

jac/jaclang/cli/docs/reference/plugins/jac-scale-scheduler.md
kashmithnisakya: added the duplicate-runs caveat to the in-memory bullet, plus a per-replica note where static schedules are introduced since those fire on every replica regardless

jac/jaclang/cli/docs/reference/plugins/jac-scale-scheduler.md
kashmithnisakya: reworded it as an install-time capability flag, default `None`, with an explicit note that it is not a runtime switch and static schedules fire regardless

jac/jaclang/cli/docs/reference/plugins/jac-scale-scheduler.md
kashmithnisakya: added `system_user_password` and `user_exists_ttl` rows to the config table

jac/jaclang/cli/docs/reference/plugins/jac-scale-scheduler.md
kashmithnisakya: the examples all used valid dates so i never hit it; keeping this PR docs-only, so the error table now documents the current 500-at-schedule-time behavior and the validate round-trip fix can land in a follow-up code PR

PR #7791, opened 2026-07-30
style: reformat .jac files under the new operator-chain wrapping rule

kugesan1105:
## Why

#7781 taught `jac fmt` to wrap over-width operator chains (`BinaryExpr`, `CompareExpr`, `MatchOr`) at width, but landed without reformatting the repo's own `.jac` files under the new rule. The full-repo sweep in `jac-check` ("Fail if formatting was not clean") has failed on every push to main since, flagging 254 files — see e.g. run 30521419682.

## What

Purely mechanical: the output of the exact command CI's full-repo mode runs,

```sh
git ls-files -z '*.jac' \
  | grep -zv -E '(/fixtures/|^scripts/|/passes/native/llvm/|_err\.jac$|_syntax_err\.jac$)' \
  | xargs -0 jac fmt --lintfix
```

run with the in-repo dev-mode compiler (so the new formatter did the reformatting). No hand edits.

## Verification

- `jac fmt --check --lintfix` over the same file set now exits 0.
- Re-running the fixer changes nothing further (idempotent — no autofix ping-pong risk).
- Remaining lint output is warnings only (docstring placement, mutable defaults, etc.) which do not fail the check; left untouched to keep this diff mechanical.

## Follow-ups: everything this uncovered

Getting this PR green surfaced a chain of pre-existing bugs, each now tracked:

- **#7792 — fmt layout quality.** The new operator-chain wrapping puts continuations at the statement column with no indent and no parens (`assert x\n== y;`, `if a\nand b {`, `x = a\n| b;`, ...). Five minimally-reproduced cases from this reformat's own diff.
- **#7793 — jac0 bootstrap transpiler corrupts multiline f-strings.** This PR's first CI run failed `test_jaseci.jac::traversing save`. Bisecting 254 files down to one hunk showed the reformat's collapse of a multiline f-string in `printgraph` *changed the compiled string constant*: `jac0.py`'s `_read_string` whitespace collapse (`" ".join(inner.split())`, `jac0.py:206-212`) eats significant whitespace from literal segments of any f-string whose `{...}` field spans lines — and jac0core is the only code compiled by jac0, which is why no standalone repro exists. The test baseline encoded the mangled output (source says two spaces; jac0 emitted one; the test asserts one). The squashed commit carries the one-character mitigation: drop the accidental doubled space in `printgraph` so the source, the correct compile, and the baseline all agree. A 20-line oracle repro (jac0 vs Python on the same expression) is on the issue and doubles as the regression test for the real fix.
- **#7795 — the CI loophole that let this happen.** `jac-check` scopes fmt/check to the PR's changed files, so #7781 changed formatter behavior, its own files passed the scoped check, it merged, and the first full-repo sweep on main flagged 254 files nobody had touched. That PR makes the fmt sweep always full-repo (autofix included, so a formatter-behavior PR carries its repo-wide reformat atomically) and escalates `jac check` to full-repo when `jac/jaclang/compiler/` or `jac/jaclang/jac0core/` is touched. Merge order: this PR first, then #7795.
- **Contribution Checks** failed only on the release-notes fragment; style-only sweep, so the `skip-release-notes-check` label is applied rather than a fragment.


kugesan1105:
CI triage on the first run:

- **Contribution Checks** failed only on the release-notes fragment check — added the `skip-release-notes-check` label (style-only sweep, no user-facing change).
- **test-runtime** failed on `test_jaseci.jac::traversing save`. Bisected it to the reformat of `jac/jaclang/jac0core/impl/runtime.impl.jac`: collapsing the multiline f-string in `printgraph` changed the compiled string constant (one space vs two before `[label=`). That is a compiler bug — layout-dependent lexing of multiline f-string replacement fields in the core impl path — filed with full evidence as #7793. The doubled space in the source was always accidental, so 207dc95 removes it; the dot output now matches the test baseline under correct lexing. Full `test_jaseci.jac` suite (18 tests) passes locally with the dev-source compiler.

PR #7794, opened 2026-07-30
chore(jacignore): drop 82 entries that no longer suppress anything

SandeepaHWP:
The ignore list consumed by CI's `jac check` step has accumulated entries that are now inert. This removes 81 of them, taking the list from **570 to 488**. Nothing CI checks today changes, and no source file is touched.

## What is being removed

**14 entries name files that no longer exist.** Renamed, deleted, or moved since the entry was added: `plugin.impl.jac` became `cli.impl.jac`, `dependencies.jac` and `dep_registry.jac` are gone, the four `cef_*.na.jac` and two `na_stdlib/*.na.jac` paths no longer resolve, `compiler/ownership.jac` and `scale/scripts/build_admin_ui.jac` are absent. Verified by expanding every pattern against `git ls-files`: these match zero files.

**68 entries name files that already pass `jac check`.** Their exemptions were earned by problems that have since been fixed, so the entries were quietly holding clean modules out of the type checker. Every one of the 61 modules behind these entries was re-checked before removal: 61 ok, 0 failed.

## Verification

- Each removed pattern was expanded against `git ls-files` using the same `fnmatch` semantics as `path_matches_exclude` in `jac/jaclang/jac0core/helpers.jac`, so the matching here is what the check step actually does.
- The 61 affected modules were run through `jac check --nowarn` in chunks, with declaring modules passed directly and `.impl.jac` annexes left to be pulled in transitively (passing an impl directly double-counts its diagnostics against the module that declares it).
- Plugin deps were installed locally first, mirroring the `jac install` block in `.github/workflows/ci.yml`, so import resolution matched CI.

## Why so many had gone stale

An entry without a `/` is matched against *every path component*, so a bare filename exempts every file of that name repo-wide. That makes entries cheap to add and hard to notice once they stop being needed. `frontend.jac`, for instance, covers nine files across examples, a project template, and the CLI.

For the same reason, entries whose bare name also matches a **still-failing** module were deliberately left in place, even where the specific module named in the original comment now passes. Seven entries fall in that group (`api.jac`, `frontend.jac`, `registry.jac`, `registry.impl.jac`, `server.jac`, `server.impl.jac`, `transport.jac`); they can only go once their siblings are clean. Converting surviving bare names to full paths as they are worked through would make the remaining list state honestly what is exempt.


kugesan1105:
good to go; resolve the conflict and update this branch with lastest main

Issue #7793, opened 2026-07-30
jac0 bootstrap transpiler destroys significant whitespace in multiline f-strings (jac0.py _read_string collapse) -- source layout changes jac0core runtime behavior


kugesan1105:
## Root cause (exact lines)

`jaclang/jac0.py` — the single-file bootstrap transpiler that compiles the `jac0core/` tree — destroys significant whitespace in any **non-triple-quoted string that spans lines** (i.e. every f-string with a multiline replacement field). In `_read_string`, `jac0.py:206-212`:

```python
        # Collapse newlines in non-triple-quoted strings (py2jac line wrapping)
        if not triple and "\n" in value:
            # Find the opening quote position in the value
            qi = len(value) - len(value.lstrip("fFrRbBuU")) + 1
            inner = value[qi:-1]  # strip prefix+quote and closing quote
            inner = " ".join(inner.split())  # collapse whitespace
            value = value[:qi] + inner + value[-1]
```

`str.split()` drops leading/trailing whitespace and collapses every internal whitespace run — across the **entire token**, literal segments included. Collapsing whitespace is only safe *inside* the `{...}` replacement field (where it is expression whitespace); applying it to the literal parts changes the string constant the program builds.

Ten-line demonstration on the exact token from `jac0core/impl/runtime.impl.jac` (`printgraph`'s dot edge label):

```python
value = 'f\' [label="{edge_label\n    if "GenericEdge" not in edge_label\n    else ""}"];\\n\''
qi = len(value) - len(value.lstrip("fFrRbBuU")) + 1
inner = value[qi:-1]
inner = " ".join(inner.split())
print(value[:qi] + inner + value[-1])
# f'[label="{edge_label if "GenericEdge" not in edge_label else ""}"];\n'
#   ^ the literal's leading space is gone
```

## Why it makes source layout change program behavior

The same token stream compiles to **different string constants** depending on where the line breaks fall:

| second f-string layout | compiled edge line |
|---|---|
| multiline replacement field (as on main) | `0 -> 1 [label=""];` — one space |
| identical tokens on one line | `0 -> 1  [label=""];` — two spaces |

Two spaces is the faithful compile (the two concatenated literals in the source each contribute one: `"...index(target)} "` + `f' [label...'`); the multiline form loses the leading space of `' [label="'` to the collapse above.

[...]

PR #7795, opened 2026-07-30
ci: close the scoped-check loophole for toolchain PRs

kugesan1105:
## The loophole

`jac-check` scoped both `jac fmt` and `jac check` to the PR's changed `.jac` files. That optimization silently assumes the formatter/checker itself is fixed — but a PR can change them. That's exactly how main went red this week:

1. #7781 changed `jac fmt` behavior (operator-chain wrapping). Its own changed files were formatted under the new rules, so the **scoped** check passed and the PR merged.
2. The first push to main ran **full-repo** mode and flagged 254 files the PR never touched. Every main push stayed red until the mop-up reformat (#7791).

The failure was structurally guaranteed before the merge but only observable after it. `jac check` has the same blind spot for compiler/jac0core PRs.

## The fix

- **`jac fmt`: never scoped.** Formatting is a property of *(file, formatter)*, so checking only the PR's files proves nothing about the merge result. The full sweep costs ~90s on the warm binary; the only skip is PRs touching no `.jac` files and no toolchain paths. This is the same contract Black uses: a formatter-behavior change carries the repo-wide reformat in the same PR.
- **Autofix pushes the repo-wide fix.** The autofix step previously fixed only `changed-jac.txt`; against a full-sweep failure it would fix nothing relevant and leave the check permanently red. It now runs the same sweep the check ran, so the behavior change and the reformat land atomically in the offending PR.
- **`jac check`: scoped stays, with an escalation.** Full-repo `jac check` is the expensive step, so ordinary PRs keep the scoped fast path — but a diff touching `jac/jaclang/compiler/` or `jac/jaclang/jac0core/` (the bootstrap) escalates to full-repo, since those can introduce type errors in files the PR never touches.

## Ordering note

This PR only touches `ci.yml`, so it can merge independently — but **#7791 should merge first**: until the repo is reformatted, the full fmt sweep fails on every `.jac`-touching PR (and same-repo PRs would receive a 254-file autofix commit).

## What this would not have caught

The runtime-behavior change from the reformat (#7793) needs the test suites, not the fmt sweep — with this change those suites run against the repo-wide reformat *inside the formatter PR*, which is where that failure belongs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -274,8 +274,11 @@ jobs:
           submodules: true
           fetch-depth: 0
 
-      # On PRs, scope format/check to the .jac files changed vs the merge-base;
-      # otherwise run the whole repo.
+      # On PRs, `jac check` is scoped to the .jac files changed vs the
+      # merge-base -- UNLESS the PR touches the toolchain (compiler or the
+      # jac0core bootstrap), which can change checker behavior for files the
+      # PR never touches; those PRs escalate to full-repo mode. The fmt check
+      # below is never scoped (see its comment).
       - name: Determine changed .jac files
         id: changes
         run: |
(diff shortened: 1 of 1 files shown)


marsninja:

PR #7797, opened 2026-07-30
fix(ecmascript,native): correct declaration errors in the es backends and emitter tables

SandeepaHWP:
Genuine type errors found by running `jac check` against files currently exempted in `.jacignore`. Nothing here is annotated around a checker limitation: where a diagnostic turned out to be a checker gap rather than a defect, the file stays on the ignore list instead.

## `self` was unannotated on every emitter impl signature

`ESIntEmitter` and friends are declared with `class`, where `self` is an explicit parameter, and Jac requires every parameter to carry a type. The impl signatures omitted it. All 561 `E0052` in the repository came from these two files: 332 in the ecmascript emitter table, 229 in the native one.

This is not cosmetic. `E0052` blocks the build: a method whose `self` is unannotated does not run at all. A reduced case, where the only difference is the annotation and only the second executes:

```jac
class C { def bar(self: C, x: int) -> int; }

impl C.bar(self, x: int) -> int { return x * 2; }      # E0052, refuses to run
impl C.bar(self: C, x: int) -> int { return x * 2; }   # clean, runs
```

Annotating also hands the checker a receiver type it did not have before, which is why the native table fell to 947 rather than the arithmetic 884. Those 63 extra diagnostics were being suppressed: with `self` untyped, every `self.<attr>` read resolved to Unknown and was accepted silently. They are pre-existing problems in that file, now simply visible.

## `FragmentTag` was the only view type demanding a source node

`view_ir` has seventeen types carrying a `jac_node`. Sixteen declare it `(uni.UniNode | None)`; `FragmentTag` alone declared it required. That made `FragmentTag(jac_node=jn)` an error wherever no node exists, which is the ordinary case for a synthesized fragment. Aligned with the other sixteen.

## The child-list parameters had no element type

`_jsx_child_nodes` and `_view_children_as_expr` both took `children: list`. With no element type every child resolved to Unknown, including the `isinstance` narrowing the whole dispatch depends on. The six-member child union was already written out inline five times in `view_ir`, so it is now a named `ViewChild` alias and both signatures use it.

## `_jsx_tag_name` rejected the value it is actually passed

It declared `expr: es.Expression`, but its only in-tree caller passes `expr.object` off a `MemberExpression`, typed `Expression | None`. The body already returns `None` for anything it does not recognise, so the parameter is widened rather than the call site guarded.

## `_jsx_attr` widened a literal-only local to the whole expression union

`_jsx_attr` builds its result into a local declared to hold three things:

```jac
value: (es.Literal | es.JsxExpressionContainer | None) = None;
```

The final `else` assigned `attr.value` straight into it. But `StaticAttr.value` is declared `es.Expression`, a twenty-member union, so the assignment offered far more than the local accepts:

```
error[E1001]: Cannot assign Identifier | ThisExpression | ArrayExpression | ... | Literal
to Literal | JsxExpressionContainer | NoneType
```

The catch-all is now split in two. An explicit `elif isinstance(attr.value, es.Literal)` narrows the value before the direct assignment, and the remaining `else` wraps anything else in a `JsxExpressionContainer`, which the local does accept.

Behaviour is unchanged. Both places that construct a `StaticAttr` pass an `es.Literal` (`normal_attribute` in `jsx_processor.impl.jac`, at the no-value and string-value branches); every other path there returns `EventAttr`, `RefAttr` or `DynamicAttr`. The `elif` therefore carries all real traffic and the new `else` is unreachable today. It is a declaration made honest rather than a new code path, and the fallback it picks is the one the unsafe-string branch directly above already uses.

## `JSCodeGenerator` never declared its indent fields

`init` assigns `self.indent_str` and `self.indent_level`, but the `has` block declared only `line_map`, `_current_line` and `_line_to_src`. Every read of either was an attribute the type does not have, including the `indent` accessor, whose entire body is `self.indent_str * self.indent_level`.

## `BlockStatement.body` referenced an unimported `List`

It was annotated `List[Statement]`, but `List` is never imported in that module. It appears exactly once in the file, at that line. The name resolved to nothing, so `body` degraded to `<Unknown>` and every consumer of a block's statements lost its type with it. Corrected to the builtin `list`, which is what the rest of the codebase uses (4,825 occurrences against 12).

Worth noting how it survived: an unresolved name in an annotation raises nothing of its own. It becomes `<Unknown>` and resurfaces later as an unrelated-looking assignment error, so nothing ever pointed at the annotation.

## Ignore-list changes

`primitives_es` and the solid backend both reach zero, so their four entries are removed.

Two files stay on the list deliberately:

- **`primitives_native`**: 947 diagnostics remain, originating in the vendored llvm tree under `passes/native/llvm/`, which is kept diff-comparable with upstream and is not touched here.
- **`es_unparse`**: 3 remain, all checker limitations rather than defects: two are missing truthiness narrowing, where `if v { ... }` still sees an optional local as `str | None`, and one is `str.join` over an iterable of `str` being typed `LiteralString`, pinning the local so a later plain `str` assignment is rejected.

Those are tracked separately. Annotating around them would hide the gaps rather than close them.

## Checks

`view_ir` is already covered by CI and stays clean. Since `FragmentTag` is shared, its consumers were measured before and after the change. `react`, `esast_gen_pass` and `jsx_processor` are unmoved at 4, 145 and 30 diagnostics respectively. Generated JavaScript is unchanged throughout; these were mismatches between what the code does and what it declares.


jac/jaclang/compiler/passes/ecmascript/view_ir.jac
```diff
@@ -12,7 +12,7 @@ obj ComponentTag {
 }
 
 obj FragmentTag {
-    has jac_node: uni.UniNode;
+    has jac_node: (uni.UniNode | None);
```
kugesan1105, line 15: when this can be `None` type

SandeepaHWP: This fragment is synthesized by the compiler rather than written by the user, so no corresponding Jac source node exists. lower_async_boundary receives only an AsyncBoundary (reactive_intent.jac:26), which carries lowered ES expressions and no node reference, so jn = None (solid.impl.jac:479) is passed to the FragmentTag at line 497. Both lines already exist on main; this PR only corrects the annotation to match the existing behaviour.

The None value is safely handled downstream: sync_loc falls back to the current node via if not jac_node { jac_node = self.cur_node; } (esast_gen_pass.impl.jac:6276), and this particular path uses _noloc, which ignores the argument entirely (solid.impl.jac:89).

All sibling tags constructed from the same jn in that block already declared (uni.UniNode | None), leaving FragmentTag as the only outlier. Reverting the annotation reproduces error[E1053] at line 497.

kugesan1105:
Merging this, the code changes all check out (no correctness issues found, generated JS confirmed unchanged since both `StaticAttr` producers only ever pass `es.Literal`). Leaving follow-ups here so they don't get lost:

**1. Release note over-claims what was un-ignored.** Only `solid.jac`, `solid.impl.jac`, `primitives_es.jac`, and `primitives_es.impl.jac` were removed from `.jacignore`. The unparser and the native emitter table are still exempt: the bare `es_unparse.jac` / `es_unparse.impl.jac` entries (lines 102-103) match by basename per `path_matches_exclude`, and `primitives_native.impl.jac` (line 174) plus bare `primitives_native.jac` (line 444) survive too. `estree.jac` was never in `.jacignore` at all. The note should be trimmed to what actually got enforced, or the follow-ups below close the gap.

**2. es_unparse is exactly one error away from being enforceable.** Running `jac check` on the pair with the ignore entries dropped fails on a single pre-existing E1001 at `impl/es_unparse.impl.jac:322`: `params` gets inferred as `LiteralString` from its initial assignment and then rejects `str`. Annotating that one local as `str` and deleting the two bare entries would put the new `indent_str` / `indent_level` declarations under the check gate. Right now nothing stops them regressing.

**3. The `self: NativeXEmitter` sweep in `primitives_native.impl.jac` is unenforced.** With its entries dropped, the pair fails with ~1600 errors, dominated by llvmlite Unknown-attribute noise from the still-ignored `passes/native/llvm/*` bindings, so it can't be un-ignored as-is. Fine to keep the annotations, but worth knowing no CI job validates them.

**4. The new non-Literal `else` in `_jsx_attr` is unreachable and duplicates the unsafe-string `elif` wrap verbatim.** Both `StaticAttr` construction sites (`jsx_processor.impl.jac:329,340`) always pass `es.Literal`. The deeper fix is narrowing `StaticAttr.value` from `es.Expression` to `es.Literal` in `view_ir.jac:25`: the impossible branch goes away and any future producer that routes an expression into `StaticAttr` gets caught at check time instead of being silently reshaped.

**5. The `ViewChild` glob could replace the four inline copies of the same union** in `view_ir.jac` itself (`IfChild.consequent`, `IfChild.alternate`, `EachChild.body`, `ViewElement.children`). Forward references are fine, `estree.jac` already does this with `Statement`. Five spellings of one union is how a seventh child kind gets missed in one spot later.

Minor: the `has indent_str: str = '  '` default duplicates `init`'s `indent` parameter default one line above, and `init` overwrites every field, so the has-default is dead. A bare declaration would keep one copy of the truth.

SandeepaHWP:
Thanks for writing these up. Picked them up as a follow-up in #8474:

- **2**: `params` annotated `str`, both `es_unparse` entries off `.jacignore`, the pair passes clean, so the `indent_str` / `indent_level` declarations are gated now.
- **4**: took the deeper fix rather than just deleting the branch. `StaticAttr.value` is narrowed to `es.Literal`, and the unreachable `else` plus the three `isinstance` guards fall out of that.
- **5**: the four inline copies now use `ViewChild`.
- **minor**: the dead `indent_str` has-default is gone.

**3** stays as-is, the llvm bindings still block un-ignoring the native table.

**1** I left alone on purpose, rewriting a merged PR's fragment from an unrelated PR felt worse than the over-claim. Item 2 makes that pair genuinely enforced anyway, so what is left overstated is the native table and `estree.jac`. Happy to trim it if you would rather it be exact.

SandeepaHWP:
Update on these, since two of them resolved themselves.

**2 (es_unparse enforceable)** is done, but not by me. #8399 rewrote type checking, and `str.join` no longer pins a local to `LiteralString`, so the E1001 at `es_unparse.impl.jac:322` is simply gone. The annotation I had in #8474 was fixing an error that no longer exists. #8399 removed 53 entries from `.jacignore` on that basis, the `es_unparse` pair among them, so the pair is enforced now and the `indent_str` / `indent_level` declarations from this PR are under the gate. I verified with `jac check` on current main: both files pass clean with the line unannotated.

**1 (release note over-claims)** is therefore mostly closed too. The unparser really is enforced now. What is left overstated in that fragment is the native emitter table and `estree.jac`.

**3 (native table unenforced)** unchanged. `primitives_native.jac` and its impl are still on the list.

**4 and 5** are still open and still in #8474, which I have rescoped to just those plus the `indent_str` minor. Title and description updated so it no longer claims to be about un-ignoring files.

So #8474 is now three declaration cleanups, no type-error fixes, and its `.jacignore` diff is empty.

PR #7803, opened 2026-07-30
perf(native): reuse parsed kernel modules and archive parses across links

kugesan1105:
## Why

`test-compiler` on main has grown from ~15 min to ~29 min since the na roadmap waves landed (Jul 20-22): the suite now collects 2,560 tests and ~1,000 of them go through the native pipeline, where every build repeats identical work.

## What

Two process-level caches, both behavior-preserving:

**Kernel modules** — the osp/fmt kernel IR was already cached, but only as *text*: every native build re-ran `parse_assembly` + `verify()` on the same string, and `link_in` then destroyed the parsed module. Now the verified `ModuleRef` is cached once per process and linked via `link_in(preserve=True)`, which clones internally. The linked result is identical — IR-level link followed by whole-module optimization is unchanged.

**Archive selection** — `select_ar_members` re-parsed the same musl/floor archives (symbol index + member table, pure-Jac byte parsing) on every native link. Both are now memoized content-keyed on the archive bytes, and each member's undef list is cached across selections. Selection traversal per root set is unchanged.

## Validation

- `tests/compiler/passes/native/test_exec_link.jac`: 9/9 pass
- exec_link + cycle_collector + na_xmod_gc subset: passes except one `test_na_xmod_gc` host-path case that fails identically **without** this change (pre-existing, environment-related)
- This PR's CI run on the Blacksmith fleet is the real measurement against the ~29 min baseline


marsninja:

PR #7807, opened 2026-07-30
Store topology index on system root

Gorgeous-Patrick:
## **Description**

Moves the Global Topology Index (GTI) off per-user roots and onto the single **system root**, so every graph mutation is recorded in one authoritative index. This removes the cross-root bookkeeping that the index previously carried and simplifies both the on-disk encoding and the query path.

### Motivation

Previously each user root owned its own topology index, and edges/nodes that crossed roots had to be indexed into *both* roots. That required tracking an `owner_root` per node, a `SELF_ROOT` sentinel, per-hop "bail on foreign owner" logic, and duplicate indexing work. Storing the index once on the system root makes it a true *global* index: resolution no longer needs to reason about which root owns a node, and cross-user graphs resolve correctly from a single source of truth.

### What changed

**Index storage & maintenance (`topo_utils`)**
- Added `_get_index_anchor()`, which always resolves the system root (`Constants.SUPER_ROOT_UUID`) and lazily promotes/persists `ctx.system_root` when needed.
- `on_edge_created` / `on_edge_destroyed` / `on_node_saved` / `on_node_destroyed` now index into the single system-root anchor instead of resolving and double-writing source and target roots.

**Index data structure (`topology_index`)**
- Dropped the per-node `owner_root` column and the `SELF_ROOT` sentinel; `node_table` entries are now `(node_id, type_id)`.
- Removed cross-root machinery: `get_cross_root_ids`, the `bail_on_intermediate_foreign` path in `resolve_chain` (now returns a plain `set[UUID]`), and the v1→v2 migration helper.
- Bumped `TOPO_VERSION` to 3 with a slimmer node encoding, while `decode` still reads legacy v1/v2 blobs (skipping the old owner-root bytes) for backward compatibility.

**Resolver (`resolver`)**
- Added `_get_system_root_anchor()` and resolve queries against the system-root index.
- Simplified the fast-path hop check to "all origin ids present in the index" now that there is no per-node owner to inspect.

**Persistence (core + scale)**
- `memory.impl` records create/update of the system-root anchor's `topology_index_data` even without normal write access, so GTI metadata is durably persisted.
- Scale/Mongo backends persist and invalidate system-root GTI state across pods (`context`, `memory_hierarchy.mongo`, `l1_invalidation`).

### Tests

- Updated/added coverage for user-root and cross-user topology graphs: `test_topology_index`, `test_topology_integration`, `test_resolver_per_hop`, `test_cross_user_topo_index` (renamed fixture), `test_topology_e2e`.
- Scale tests updated for cross-pod invalidation and system-root persistence.

Verified locally:
```
PYTHONPATH=jac python -m jaclang test jac/tests/runtimelib/test_topology_index.jac
PYTHONPATH=jac python -m jaclang test jac/tests/runtimelib/test_topology_integration.jac
PYTHONPATH=jac python -m jaclang test jac/tests/runtimelib/test_resolver_per_hop.jac
PYTHONPATH=jac python -m jaclang test jac/tests/runtimelib/test_occ_read_capture.jac
```

> Note: the full `test_topology_e2e` run against the repo default DB failed due to stale system-root state in `./.jac/data/main.db` left over from repeated local runs, not a regression in this change.

### DIFF
--- a/jac/jaclang/runtimelib/impl/memory.impl.jac
+++ b/jac/jaclang/runtimelib/impl/memory.impl.jac
@@ -24,6 +24,22 @@ import from jaclang.runtimelib.typecache { get_field_types }
 glob logger = logging.getLogger(__name__),
      SQLITE_DB_FORMAT_VERSION: int = 1;
 
+def _is_system_root_topology_anchor(anchor: Anchor) -> bool {
+    if not (
+        isinstance(anchor, NodeAnchor)
+        and isinstance(anchor.archetype, Root)
+        and anchor.topology_index_data is not None
+    ) {
+        return False;
+    }
+    try {
+        import from jaclang.jac0core.constant { Constants as Con }
+        return anchor.id == UUID(Con.SUPER_ROOT_UUID);
+    } except Exception {
+        return False;
+    }
+}
+
 impl Memory.aget(id: UUID) -> (Anchor | None) {
     return await asyncio.to_thread(self.get, id);
 }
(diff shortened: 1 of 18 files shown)


marsninja:

Issue #7790, opened 2026-07-30
Topology index fast path: `[-->[?:T]][?f==x]` silently returns all nodes; O(N^2) edge writes; field pushdown unreachable


MalithaPrabhashana:
## Where this came from

jacBuilder#895 reports rolling OOMKills on jachammer prod. Its measured root cause is that every session lookup runs `[root()-->][?:Session][?id==session_id]`, which "cannot push the `id` predicate down, so the runtime materializes EVERY Session node ... then filters in Python" (1048 sessions, ~435MB per context).

We set out to verify that claim against jaclang itself. It is correct. While verifying it we found two further problems in the same subsystem, one of which is a **silent correctness bug** that is more severe than the performance issue we started with.

All measurements below are on jaseci `main` at `64b62314b`, local SQLite backend, Python 3.14, WSL2. The topology index is ON by default whenever a `jac.toml` is present, which is the case for every real project.

## Summary

Three independent problems, all in the topology-index path:

| # | Kind | Symptom |
|---|---|---|
| 1 | **Correctness** | `[root-->[?:T]][?field==x]` silently returns every node, unfiltered |
| 2 | **Performance** | Attaching N edges to one root is O(N^2). 8000 nodes takes ~7 minutes |
| 3 | **Missing capability** | Field predicates never reach storage, even on Mongo where support exists |

They are separable and can be split into separate issues if preferred. Problem 1 is the one we would prioritise.

---

# Problem 1: `[root-->[?:T]][?field==x]` silently returns unfiltered results

## Repro

`jac.toml` (any minimal project file; the index is on by default):

```toml
[project]
name = "bug"
version = "0.1.0"
```

`bug.jac`:

```jac
node A {
    has val: int = 0;
}

node B {
    has tag: str = "";
}

with entry {
    root ++> A(val=1);
    root ++> A(val=2);
    root ++> A(val=3);
    root ++> B(tag="x");

    print(f"nested  = {[root-->[?:A, val == 2]]}");
    print(f"chained = {[root-->][?:A][?val == 2]}");
    print(f"mixed   = {[root-->[?:A]][?val == 2]}");
}
```

`rm -rf .jac && jac run bug.jac`

## Expected

All three forms ask for "nodes of type `A` where `val == 2`", so all three should print `[A(val=2)]`.

## Actual

```
nested  = [A(val=2)]
chained = [A(val=2)]
mixed   = [A(val=1), A(val=2), A(val=3), B(tag='x')]
```

[...]

PR #7813, opened 2026-07-31
fix(runtime): keep a trailing filter separate from the hop's type filter

akindu-k:
Fixes the correctness problem in #7790 (Problem 1). The O(N^2) edge writes (Problem 2) and the field pushdown gap (Problem 3) are left for separate PRs.

## What was broken

There are three ways to write "give me the `A` nodes off root where `val == 2`". Two of them worked. One returned every node in the graph.

```jac
[root-->[?:A, val == 2]]     # type and condition inside the arrow
[root-->][?:A][?val == 2]    # type and condition both chained on after
[root-->[?:A]][?val == 2]    # type inside the arrow, condition chained after
```

```
nested  = [A(val=2)]
chained = [A(val=2)]
mixed   = [A(val=1), A(val=2), A(val=3), B(tag='x')]   <- every node
```

The third form dropped the condition *and* the type filter, so it returned everything connected to the root, including nodes of a completely unrelated type. There was no error and no warning. Setting `topology_index = false` made it correct again, so the problem was specific to the index fast path, which is on by default in any project that has a `jac.toml`.

This is dangerous in exactly the code that reported the original issue. jac-code looks up sessions with `[root()-->][?:Session][?id == session_id]`, which is the second form and is correct today. Moving the type filter inside the arrow is the kind of tidy-up anyone might do without thinking, and it would quietly turn a single-session lookup into "return every session".

## Why it happened

When the runtime plans a graph query, it has to decide whether it can answer from the topology index or whether it has to walk the graph. It made that decision by *inspecting the filter function the compiler generated*: disassembling its bytecode to guess "does this test a field?" and inspecting its closure to guess "which node type does this want?".

That works while the filter is a single flat lambda. The third form is different: it arrives as two separate lambdas, one for the type and one for the condition, and `refs()` glued them together by wrapping them in a third lambda.

That wrapper is opaque to both guesses. Its own bytecode contains no comparison, just "call these two functions". Its closure contains two functions rather than a node type. So the runtime concluded there was nothing to filter, took the index fast path with an empty filter, and never called the lambda at all.

image

## Why only one of the three forms broke

Worth recording, because it explains how this survived so long: the three spellings do not lower to the same shape. `jac tool jac2py` over them gives three different call shapes, and the output is byte-identical on `main` and on this branch.

```python
# [root-->[?:A, val == 2]]
refs(OPath(root()).edge_out(nd=lambda i: isinstance(i, A) and i.val == 2))

# [root-->][?:A][?val == 2]
filter_on(items=refs(OPath(root()).edge_out(), post_filter=lambda i: isinstance(i, A)), func=lambda i: i.val == 2)

# [root-->[?:A]][?val == 2]
refs(OPath(root()).edge_out(nd=lambda i: isinstance(i, A)), post_filter=lambda i: i.val == 2)
```

`refs()` accepts exactly one trailing filter, as `post_filter`. The first chained filter claims that slot; every one after it becomes a `filter_on(...)` wrapper around the result. `filter_on` is a plain list comprehension that runs after `refs()` has returned, over the materialised list, entirely outside the query engine.

Line the three forms up by what actually lands inside a single `refs()` call:

| Form | `nd` | `post_filter` | Safety net outside `refs()` | On `main` |
|---|---|---|---|---|
| `[-->[?:A, val==2]]` | one flat lambda | — | none needed | correct |
| `[-->][?:A][?val==2]` | — | type | `filter_on` applies `val==2` | correct |
| `[-->[?:A]][?val==2]` | type | condition | none | **broken** |

The first form's `nd` is a single flat lambda the compiler wrote itself, so its bytecode contains the comparison and the inspection reads it correctly. The second never collides: `nd` is `None`, so the old code stored `post_filter` straight into it with no wrapper, and the condition sat safely outside in `filter_on`, where the query engine could not lose it.

The third is the only shape that puts two filters into one `refs()` call, which is exactly the branch that built the wrapper:

```jac
if last.nd is None {
    last.nd = post_filter;                                           // second form -> safe
} else {
    orig_nd = last.nd;
    last.nd = lambda (i: any) { orig_nd(i) and post_filter(i); };    // third form -> the bug
}
```

That `else` was reachable only by the third form, and the third form is also the only one with no `filter_on` downstream to re-apply what the wrapper hid. So this was never "trailing filters are broken" — it was one lowering shape, and the other two spellings are correct for two entirely different reasons.

Codegen being identical on both branches also means this is a pure runtime fix: no compiler pass changes, and `.jir` artifacts cached from `main` stay valid.

## How it is fixed

The trailing filter is no longer glued onto the hop's filter. It keeps its own field on the destination, so both halves stay readable on their own and neither has to be recovered from a wrapper.

Both halves are then inspected when deciding strategy and when recovering the type to narrow by. That distinction matters, because a trailing filter is not always a condition:

- `[-->][?:T]` arrives as a trailing filter that carries only a *type*. It still reaches the index and still narrows by that type.
- `[-->[?:T]][?f == x]` carries a real condition, so the hop moves to the walk path and the condition is applied there.

I also made `_filter_has_predicates` treat any filter that closes over other filters as condition-bearing. The property the fast path depends on is that this function never under-reports. With that in place, a future change that wraps or composes a filter degrades to the slower path instead of silently dropping filters, which is how this class of bug came back a second time.

That guard is a mitigation rather than a cure: the planner still cannot read a composed filter, it just knows to stop trusting itself. The real fix is to stop reverse-engineering intent from compiled lambdas at all — the compiler already holds the type and the conditions as separate fields on `FilterCompr` and flattens them away. Filed as #7831, which would make `_wraps_callables` and the rest of the bytecode inspection unnecessary.

image

## Result

All three spellings now agree, and the query plans for the fused and in-arrow forms are identical:

```
[query] hop[0]:index:ordered edge=* node=A dir=2 in=1 out=50 slc=0:50   # [root-->][?:A][0:50]
[query] hop[0]:index:ordered edge=* node=A dir=2 in=1 out=200           # [root-->[?:A]]
[query] hop[0]:walk:set      edge=* node=A dir=2 in=1 out=1             # [root-->[?:A]][?val==2]
[query] hop[0]:walk:set      edge=* node=A dir=2 in=1 out=1             # [root-->[?:A, val==2]]
[query] hop[0]:index:ordered edge=* node=* dir=2 in=1 out=250           # [root-->]
```

Type-only queries, unfiltered queries, and sliced queries all keep the index fast path they had before. The previously broken form now costs what the equivalent in-arrow form has always cost. Getting index narrowing back for the type half of a condition query is Problem 3's job, not this PR's.

## Tests

- `test_topology_e2e.jac`: end to end, asserts the three spellings agree and that the type-only query still returns everything it should.
- `test_resolver_per_hop.jac`: pins the structure directly. A composed lambda is treated conservatively, a trailing condition keeps its hop off the fast path, a trailing filter carrying only a type stays on it, the type stays recoverable in both cases, and the resolver actually applies the condition to its result.


akindu-k:
Closing this in favour of #7856.

#7856 is the queries-as-data clean break from #7855, and it deletes the subsystem this PR patches instead of hardening it. Everything I touched here is gone there: `query_utils` (both files), `resolver.impl.jac`, the `post_filter` fusion in `runtime.impl.jac`, and the `Destination`/`Filter` classes in `archetype`. `_wraps_callables` and `_filter_has_predicates` have nothing left to guard once the planner reads a typed IR instead of picking apart lambda bytecode.

It fixes #7790 Problem 1 as well. The repro prints `[A(val=2)]` for all three spellings, fresh and against a persisted DB. The three-spelling check I added to `test_topology_e2e.jac` is covered by `test_query_differential.jac` over a wider spelling matrix, cold and warm, so nothing is lost by dropping this.

This is basically what I asked for in #7831, just taken further. The right fix is to stop guessing intent from compiled lambdas at all, not to make the guessing fail safe.

Issue #1072, opened 2023-04-20
 jsctl jac build main.jac will not work at all.


WrappedUsername:
**Describe the bug**

I ran this command while following the instructions from [Jaseci docs](https://docs.jaseci.org/docs/docs/getting_started/3_setup_jaseci_serv.html): jsctl jac build main.jac, this command will not work at all, and returns this error:

```bash
2023-04-20 08:01:11,519 - ERROR - interface_error: Internal Exception: [Errno 2] No such file or directory: 'main.jac'
{
  "response": "Internal Exception: [Errno 2] No such file or directory: 'main.jac'",
  "success": false,
  "errors": [
    "Internal Exception: [Errno 2] No such file or directory: 'main.jac'"
  ],
  "stack_trace": [
    "Traceback (most recent call last):\n",
    "  File \"/usr/local/lib/python3.10/dist-packages/jaseci/extens/api/interface.py\", line 192, in general_interface_to_api\n    ret = getattr(_caller, api_name)(**param_map)\n",
    "  File \"/usr/local/lib/python3.10/dist-packages/jaseci/extens/api/jac_api.py\", line 29, in jac_build\n    with open(file, \"r\") as file:\n",
    "FileNotFoundError: [Errno 2] No such file or directory: 'main.jac'\n"
  ]
}
```

**To Reproduce**

- I followed each step in the [docs](https://docs.jaseci.org/docs/docs/getting_started/3_setup_jaseci_serv.html):
![image](https://user-images.githubusercontent.com/104662990/233389069-422f9c40-dde5-4956-a67b-94195ad8bd24.png)

**Expected behavior**

Well I expected it to compile the Jac program to JIR bit. Honestly I only have around a year and a half of experience as a self-taught dev, so I am not sure what else to expect, because if you don't know what to expect, because of a lack of experience and knowledge, how can you know about any other expectations that might be possible?

**Screenshots**

- Started dev server on first terminal:
![image](https://user-images.githubusercontent.com/104662990/233394199-12e95eea-2428-4510-bd56-4548f66b0e09.png)

- Development server started and running on first terminal:
![image](https://user-images.githubusercontent.com/104662990/233388235-74710119-cd86-4e64-b86a-c1b2e6b078eb.png)

- Logged into jaseci shell with new second terminal
![image](https://user-images.githubusercontent.com/104662990/233401907-c5655f1e-fdfe-4345-8412-067fb9f6741c.png)

[...]

PR #7816, opened 2026-07-31
feat(cli): seal-producing builds honor the [project] jac-version pin

### LINKED ISSUE #7864: scale: [project] jac-version pin governs the pod runtime but not manifest generation
## Summary

`[project] jac-version` resolves the pod binary, admin console and base image, but **not the code that generates the Kubernetes manifests**. Manifest generation always runs on whatever jaclang the *deploying* process has installed. So an app that pins a release specifically to pick up a deploy-path fix still gets the older, buggy manifest, with no warning.

This is not hypothetical. It is how a real user lost a day of data on `stand-up.jachammer.app`: the app pinned `==0.34.9` (which contains the #7828 fix for the dropped `MONGODB_URI`), the deploy ran from a jacBuilder pod on 0.34.8, and the Deployment was written by 0.34.8's broken code path. The app silently fell back to a local SQLite store on an `emptyDir` and lost everything on the next pod reschedule.

The underlying `MONGODB_URI` drop is already fixed (introduced by #6937, fixed by #7828, shipped in v0.34.9). This issue is about the pin gap that let a *pinned-to-fixed* app still receive the broken manifest.

## Environment

| | |
|---|---|
| Binaries under test | `jac 0.34.8 (Darwin arm64)`, `jac 0.34.9 (Darwin arm64)` (released assets `jac-0.34.<n>-macos-aarch64`) |
| Also observed | production-cluster (EKS 1.35, us-east-2), jacBuilder pods on `jac 0.34.8 (Linux x86_64)` |
| Affected releases (manifest bug) | v0.32.0, v0.34.8, v0.35.0, v0.35.1 |
| First fixed release | v0.34.9 (published 2026-08-03T10:50:32Z) |

### A. Hermetic: the two versions do not share a monolith deploy path

Fresh scratch dir, released binaries, nothing from a jaseci checkout.

```bash
mkdir -p repro/app && cd repro/app
cat > jac.toml <<'EOF'
[project]
name = "dbpin"
entry-point = "main.jac"
kind = "web-app"
jac-version = "==0.34.9"

[scale.database]
backend = "mongo"

[dependencies]
python-dotenv = "*"
kubernetes = "*"

[scale.kubernetes]
EOF
echo 'with entry { print("dbpin"); }' > main.jac
jac install
jac start main.jac --scale --dry-run
```

Run with **0.34.8** (exit 1):

```
Installing Jaseci packages from PyPI...
⚠ --dry-run only supported with kubernetes-microservice target (got 'kubernetes').
```

Run the identical project with **0.34.9** (exit 0):

```
No [scale.microservices.routes] declared; deploying the app as a single service behind a gateway.
Preparing deployment...
ℹ Generated 2 pod-spec stubs (1 gateway + 1 services) | Context: {'image': 'jaseci/jaclang:0.34.9', 'namespace': 'default'}

Totals
  2 deployments, 2 services, 2 Autoscalers, 2 PDBs, 1 PVC
```

Two things this shows:

1. 0.34.8 has a **separate, non-unified monolith target** (the one carrying the bug); 0.34.9 routes the same project through the unified path per #7828.
2. 0.34.9 correctly resolves the image from the pin (`jaseci/jaclang:0.34.9`), confirming the pin *is* honored for the image while the generating code is whatever binary you invoked.

### B. Hermetic dry-run cannot exercise this bug at all

With `--show-yaml` on 0.34.9, the generated app Deployment env is:

```yaml
env:
- name: JAC_SV_NAME
  value: dbpin
- name: K8S_APP_NAME
  value: dbpin
- name: K8S_NAMESPACE
  value: default
- name: JAC_SV_FILE
  value: main.jac
```

`grep -c MONGODB_URI` over the full YAML returns `0`, on the fixed version, because `DatabaseProvisioner` performs cluster I/O and is skipped under the hermetic `--dry-run` introduced in #7438. **The only code path that can produce `MONGODB_URI` is the one no offline test executes.** That is why this shipped.

### C. Live: a pinned-to-fixed app deployed by an older builder

On production-cluster, namespace `jb-299afd1f-standup`, app pinned `==0.34.9`:

```
$ kubectl exec -n jb-299afd1f-standup <pod> -c jac-bootstrap -- grep -E "jac-version|backend" /app/jac.toml
jac-version = "==0.34.9"
backend = "mongo"
```

The Deployment was created fresh (not a stale patch):

```
creationTimestamp: 2026-08-03T14:33:08Z
generation: 1
revision: 1        (single ReplicaSet)
```

Yet its container env is:

```
['K8S_APP_NAME', 'K8S_NAMESPACE', 'HOME', 'PATH']
```

The deploying pods are all on the older release:

```
builder-sv-...-82fb6 -> jac 0.34.8  (Linux x86_64)
builder-sv-...-h4ctc -> jac 0.34.8
builder-sv-...-kr8xl -> jac 0.34.8
builder-sv-...-rr2p2 -> jac 0.34.8
builder-sv-...-x6hdd -> jac 0.34.8
```

Proof the fixed code never executed: `database_provisioner.jac` in 0.34.9 logs `Database env injected (N vars), init containers (M)`, a string that exists only in the fixed implementation.

```
$ for p in $(kubectl get pods -n jachammer -l app=builder-sv -o name); do
    kubectl logs -n jachammer $p --since=120m | grep -c 'Database env injected'; done
0
0
0
0
0
0
0
```

Result on the app's own database:

```
DB: admin   0.10 MB
DB: config  0.06 MB
DB: local   0.07 MB
```

No application database was ever created. The app log contains zero occurrences of `mongo`; it never attempted a connection. It persisted instead to `/app/.jac/data/users.db` and `anchor_store.db`, and:

```
$ findmnt -no SOURCE,TARGET,FSTYPE /app
/dev/nvme0n1p1[/var/lib/kubelet/pods/.../volumes/kubernetes.io~empty-dir/jac-app] /app xfs
```

An `emptyDir` on node-local NVMe. Every pod reschedule destroys it.

## Expected

Per #7713, the pin "resolves to one release tag, and the pod binary, admin console, and base image all come from that same tag, **so the deployed pod is coherent**." A user reading that reasonably concludes that pinning a release gets them that release's deploy behavior. Pinning to a version specifically to obtain a deploy-path fix should obtain it, or the deploy should refuse loudly.

## Actual

The pin governs runtime artifacts only. `DatabaseProvisioner` -> `ManifestBuilder` -> `apply_manifests` all execute in the deploying process against its own installed jaclang, and there is no version check between the two. A pinned-newer app receives an older manifest with no warning at any layer.

Relevant code in v0.34.9, `jac/jaclang/scale/deploy/target/kubernetes/target.jac`:

```
L75   DatabaseProvisioner(...).provision(...)        cluster I/O
L689  ManifestBuilder(...).generate_manifests(...)   pure computation, returns a dict
L106  self.apply_manifests(bundle)                   cluster I/O
```

`manifest_builder.jac` contains **zero** Kubernetes API calls (`grep -c` for `client.|apps_v1|core_v1|read_namespaced|create_namespaced|patch_namespaced` returns 0). It takes `AppConfig` + `DeployContext` + `KubernetesConfig` and returns manifest dicts, and #7438 already threads `dry_run` through it. The middle phase is already pure and already returns a serializable value.

## Impact

Every jacBuilder-deployed single app inherits the deploying pod's jac version regardless of what it pins. At the time of writing, **22 apps on production-cluster** have a MongoDB StatefulSet provisioned but no `MONGODB_URI` on the app container, so each silently writes to an `emptyDir` and loses all user data on any pod reschedule:

`ai-recruitment, event-planner, blackstar, subway (x2 namespaces), goodmiles, standup, att5, nearnest, build-a-car-racing, build-a-doctor-appoi, let-s-build-a-librar, jaclang-website, marsninja-site, billbacked, goodbyeexcel, jachacks-sf-2026-v2, jachacks26, blastradius, jachammer-dashboard, create-an-agent-like, hey-i-want-to`

With Karpenter consolidation active these reschedule routinely (6 node events in 80 minutes observed), so the loss happens unattended. The failure is silent in the worst way: the app serves HTTP 200 throughout, so uptime checks stay green while data evaporates.

The general form outlives this particular bug. Any future manifest-shape fix (probes, resources, volumes, ingress) has the same property: pinning to the release containing it does not get you it.

## Workaround (verified)

Upgrade the **deploying** environment, not the app. On our cluster that means rebuilding the jacBuilder image on >= 0.34.9 and redeploying each app. Changing the app's `jac-version` pin was verified to have no effect (section C above: pinned `==0.34.9`, still no `MONGODB_URI`).

Per-app stopgap: patch the Deployment to add `MONGODB_URI` from the existing `<app>-mongodb-secret` `connection-string` key. This only protects data written after the patch.

## Suggested fix

The three phases are already cleanly separated and the middle one is already pure, so the smallest correct change is to run `generate_manifests` under the pinned binary and hand the resulting bundle back to the deployer:

```
deployer (any version)   provision DBs / namespace / SA        [cluster I/O]
        |  serialize DeployContext + AppConfig as JSON
        v
pinned binary            emit the manifest bundle as JSON      [pure]
        |
        v
deployer                 apply_manifests(bundle)               [cluster I/O]
```

The mechanism already exists. #7811 established exactly this pattern for the seal step in `jac/jaclang/scale/injector/bundle.jac`, writing the pinned binary to disk and bootstrapping into its own module:

```
L207  jac_path.write_bytes(pod_binary);
L211  "import from jaclang.utils.precompile_bytecode { main }\n"
L221  result = subprocess.run(...)
```

The open design question, which I do not want to paper over: `DeployContext` in and the manifest bundle out become a versioned compatibility contract. An older deployer cannot apply a manifest kind a newer version invents. That bounds how far the skew can stretch, and the bound should be explicit and checked rather than discovered in production.

If full delegation is judged too large, the much cheaper alternative is a **loud abort**: when the pinned version differs from the deploying jaclang, refuse the deploy with a clear error, exactly as #7713 already does for a pin that cannot be honored. Silently shipping a manifest from a different version than the user pinned is the part that caused harm.

### Regression test that would have caught it

The gap is that no test exercises the DB-env path, because `--dry-run` skips cluster I/O (section B). Either:

1. A test that asserts, for a project with `[scale.database] backend = "mongo"`, that the generated Deployment carries a `MONGODB_URI` entry sourced from the mongo secret. This needs `DatabaseProvisioner` to be faked rather than skipped, so the manifest builder is exercised with a populated `DeployContext`.
2. A kind-based E2E that writes a record, deletes the app pod, waits for reschedule, and reads the record back. An HTTP 200 assertion is insufficient: the broken app served 200 the entire time it had no database.

Related: #6937 (introduced the drop), #7828 (fixed it), #7713 (added the pin), #7811 (precedent for running the pinned binary's own code), #7438 (hermetic dry-run).

The missing test coverage that let this reach users is filed separately as jaseci-labs/jacBuilder#1060.

MusabMahmoodh:
## Summary

`[project] jac-version` (stamped by `jac create`, honored by `jac start --scale` since #7713) selects the pod runtime at deploy time, and the deploy seals the app .jab with that same pinned binary. The host-built seal path had no such guard: `jac build --as jab` / `sealed` / `binary` sealed with whatever jac happened to be running, silently ignoring the pin - and a sealed bundle is locked to the toolchain that builds it (`python_tag` / `jir_format_version` are fail-closed at load).

This closes that gap with the simple rule: **to produce a sealed artifact, the running jac must satisfy the pin.**

- `jac build --as jab|sealed|binary` (and `jac bundle`) now errors when the running `jac_version()` does not satisfy `[project] jac-version`, naming the pin and the running version.
- Spec matching reuses the deploy resolver's matcher (`version_satisfies`, a public wrapper over the existing `_satisfies` in `jaclang.scale.injector.binary`), so build and deploy agree on what a spec means.
- Source distributions (`wheel`, `npm`) carry no seal and are not gated.
- Docs: `jac-version` reference section documents the build-time gate.

## Tests

- `jac test jac/tests/publish/test_bundle_cli.jac`: 6 passed (3 new - unsatisfied pin refuses the jab and produces nothing; satisfied pin builds; wheel is not gated), run against a `-Ddev` linked binary so the suite exercises this branch's source.
- `jac test jac/jaclang/scale/tests/test_version_pin.jac`: 17 passed (deploy-side pin resolution untouched).


MusabMahmoodh:
The rule this lands ("to produce a sealed artifact, the running jac must satisfy the pin") is the right one, and exporting `version_satisfies` is the piece that makes it reusable. A few things it could grow into, from hitting the same class of problem on the jacBuilder side.

**1. The guard refuses, but everything needed to resolve is already here.**

Right now an unsatisfied pin is a dead end: the user is told the running version does not match and is left to go find the right toolchain themselves. `binary.jac` already has the whole resolution path:

- `_resolve_target_version(repo, spec)` picks the release that satisfies a spec
- `download_release_binary(channel, arch, version)` fetches it
- `_fetch_verified_asset` checksum-verifies and caches it

So the natural next step is offering the fix rather than only naming the problem: fetch the pinned toolchain and either re-exec with it or tell the user the exact command. Behind a flag (`--auto-toolchain`) or an interactive prompt, so CI stays deterministic and non-interactive. For someone testing locally against a pinned app this is the difference between "build refused" and "build worked", and it is a small wrapper over functions that already exist.

**2. `_binary_cache_path` is not version keyed.**

```jac
def _binary_cache_path(channel: str, arch: str) -> Path {
    return _scale_cache_root() / f"jac-{channel}-linux-{arch}";
}
```

Two different pinned versions on the same channel and arch resolve to the same cache file. That is fine while a machine only ever holds one binary per channel, but it blocks anything that wants several pinned toolchains present at once, and it can silently serve the wrong bytes if the resolved version changes between runs. Putting the version in the path is a prerequisite for any multi-version work on top of this.

**3. The error could carry the resolved version, not just the spec.**

The message names the pin and the running version. Since `_resolve_target_version` can already turn a spec like `>=0.34.0,<0.35.0` into a concrete release, including that ("pin resolves to 0.34.9") saves the reader a lookup and makes the message directly actionable.

**4. Where this is heading for us.**

We are looking at supporting several toolchain versions side by side, so a deployed app can run on the version its own `jac.toml` pins instead of being forced onto the builder's version. The shape is versioned aliases (`jac<version>`) staged next to each other, with the pod starting the one matching the app's pin. `version_satisfies` plus the resolver in this PR is exactly the primitive that needs, which is why the cache keying in (2) matters beyond this change. Proposal is written up in jaseci-labs/jacBuilder#1072 if it is useful context for where the guard gets used.

Worth saying that the fail-closed direction is right regardless. Sealing with the wrong toolchain does not fail at build time, it fails much later as a pod that crash-loops with no obvious cause, so catching it at the build is a real improvement on its own.

MusabMahmoodh:
Thanks - three of the four are in as of \`161684bb2\` (branch also rebased onto current main).

**1. Auto-fetch: shipped, but default-on rather than behind a flag.**

This landed in \`b3079c4c3\`, before the comment - worth a re-read since the shape differs from what you sketched. An unsatisfied pin now resolves the spec, downloads the matching release binary for the *host* platform (same \`jac-<ver>-<os>-<arch>\` asset naming \`scripts/install.sh\` uses, checksum-verified) into \`~/.cache/jac/toolchains/\`, and re-execs the build under it. \`JAC_TOOLCHAIN_REEXEC\` guards against bootstrap loops, and \`JAC_NO_DEV_SOURCE\` makes the fetched binary seal with its own bundled compiler.

I went default-on with an env opt-out (\`JAC_NO_TOOLCHAIN_FETCH=1\`) instead of opt-in, for two reasons. It matches deploy, which already downloads the pinned binary with no flag - if the same pin silently works for \`jac start --scale\` but needs \`--auto-toolchain\` for \`jac build\`, that gap is its own trap. And on CI determinism I think it cuts the other way: with default-on, the artifact is sealed by the version the pin names no matter which jac the CI image happens to ship, which is *more* deterministic than either the old silent-wrong-seal behavior or a hard failure. The download itself is deterministic (concrete release, checksum-verified, cached). Happy to add a real \`--no-toolchain-fetch\` flag alongside the env var if you want it discoverable in \`--help\`; I skipped the interactive prompt deliberately so behavior is identical in a terminal and in CI.

**2. Version-keyed cache: done.**

\`\`\`jac
def _binary_cache_path(channel: str, arch: str, version: str = "") -> Path {
    return _scale_cache_root() / f"jac-{version or channel}-linux-{arch}";
}
\`\`\`

Pinned deploys key by resolved version; unpinned ones keep the channel key. Covered by a test asserting two versions on one channel+arch land on distinct paths and that the unpinned path is unchanged. The build-side toolchain cache was already version-keyed - verified two pinned toolchains coexisting (\`jac-0.34.7-linux-x86_64\`, \`jac-0.34.9-linux-x86_64\`).

**3. Resolved version in the error: done, with one narrowing.**

A refused build now reports `The pin resolves to jac 0.34.9.` for a range. I left exact pins without the clause: `_exact_of` returns the version without checking it was ever published, so for `==0.0.1` the message would read "resolves to jac 0.0.1" about a release that does not exist. The spec already names the version in that case, so the clause only adds value for ranges.

**4. Multi-version side-by-side.**

`resolve_pin_target(spec)` is now exported next to `version_satisfies` - spec to concrete release, `""` when none satisfies. Between those two and the now version-keyed cache, the primitives your versioned-alias plan needs should be here; the staging/`jac<version>` alias layer is the part that belongs on your side. I read jacBuilder#1072 - the one thing I would flag is that the pod picks its runtime from the app's pin, which means the pin has to be resolvable at pod start or baked at deploy; baking it (the deploy already resolves it once and memoizes) avoids every pod re-hitting the releases API.

Verification: 20 passing in `test_version_pin.jac` (3 new), 7 in `test_bundle_cli.jac`, plus a real e2e - a project pinned `>=0.34,<0.35` built by a running jac 0.35.1 fetched 0.34.9 and produced a .jab whose sealed manifest reports `jaclang_version: 0.34.9`.

ahzan-dev:
Following up from the jacHammer side, since this is the primitive #1072 wants to build on.

The auto-fetch + re-exec under the pin, the version-keyed `_binary_cache_path`, and the exported `version_satisfies` / `resolve_pin_target` are exactly what unblocks running a deployed app on its own pinned generation instead of the builder's. 

One thing I want to confirm before we wire the jacHammer alias layer: all of this lands on the `jac build --as jab|sealed|binary` path. jacHammer doesn't deploy through `jac build` — it seals through the jac-scale injector (`bundle.jac` / `resolve_deploy_version`), which runs the precompile script under `pod_binary` in a subprocess. **Does that injector/deploy seal path also fetch-and-re-exec under the app's resolved pinned toolchain, or is the re-exec behavior CLI-only?** If it's CLI-only, the "seal with the resolved toolchain" step is ours to add on the deploy path in #1072 — just want to know which side owns it before we start.

And do you have a rough merge / target-release for this? We can't begin #1072 until we're on a jac release that carries `version_satisfies` / `resolve_pin_target` + the version-keyed cache.

MusabMahmoodh:
Not CLI-only, and you don't need to add it.

The injector path never seals in-process. `BinaryInjector.prepare` resolves the pin, downloads that binary, and passes the bytes to `pack_jab`, which writes them to `.pod-jac` and runs the precompile as a subprocess of that binary. `pack_jab` refuses to run at all without them. So the thing doing the seal is always the exact binary the pod will run, on any channel, and that has been true since #7713. The re-exec here only exists because `jac build` seals in-process, which is the one place the two could drift.

That makes me want to double-check the crash in #1075 though. If that seal went through the injector, a 0.34.9 bundle landing on a 0.34.5 pod shouldn't be reachable. It reads more like the jab was built by jacHammer's own jac and then deployed, which is exactly what this PR fixes on the build side. Worth pinning down which path you are actually on before the alias layer gets built on top of the stopgap, since it decides whether you need the stopgap at all. If it is the injector path and you can repro, send me the deploy and I will look.

On timing: no date I can promise, it is open and unreviewed. Whenever it merges it rides the next release.

PR #7817, opened 2026-07-31
ci: capture per-test timings in the compiler and equivalence suites

kugesan1105:
## What

Adds per-test timing capture to the two binary suite steps in `test-compiler`, via `PYTEST_ADDOPTS` (no harness or compiler code changes):

- `--durations=0 --durations-min=0.001` — pytest prints every test's wall time (setup/call/teardown) at the end of the run, visible directly in the CI log.
- `--junitxml` — the same timing data uploaded as a structured artifact (`compiler-suite-timings`, `equivalence-suite-timings`) for offline sorting/aggregation.

## Why

The compiler suite step grew from ~12.4m (Jul 14) to ~22.8m (Jul 30) and we need ground-truth per-test numbers from the real Blacksmith runners to see where the time goes. Prior profiling was done on fork runners (different hardware), so absolute numbers were not trustworthy.

## Notes

- Measurement only — no behavior change to the suites.
- The suites run with `JAC_TEST_JOBS=auto` (pytest-xdist, 4 workers on the 4vcpu runner), so recorded times include cross-worker contention. They answer "where does the wall time go", not "what does this test cost solo".
- `if-no-files-found: warn` keeps the upload steps from failing runs where the suite dies before pytest writes the XML.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -191,7 +191,13 @@ jobs:
           set -uxo pipefail
           WORK="$(mktemp -d)"
           rc=0
-          HOME="$WORK" JAC_TEST_JOBS=auto jac test tests/compiler | tee /tmp/compiler.log || rc=$?
+          # Per-test timings: --durations prints every test's time in the log;
+          # junitxml uploads the same data as a structured artifact. Under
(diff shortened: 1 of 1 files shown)


marsninja:

PR #7819, opened 2026-07-31
ci: docker-retag workflow to repoint jaclang:latest at a published version

kugesan1105:
## Summary

Backport releases steal the `jaseci/jaclang:latest` Docker tag: the release pipeline's docker job always pushes `<version>` + `latest`, so publishing a patch for an older line (today: 0.34.8 after 0.35.0 was already out) leaves `:latest` pointing at the older version.

This adds a small dispatch-only recovery workflow that repoints `:latest` at any already-published version's image via `docker buildx imagetools create` — a multi-arch manifest alias, no rebuild, no asset churn, using the existing `DOCKERHUB_*` org secrets.

Usage: dispatch `Docker retag latest` with `version: 0.35.0`.

A follow-up PR adds guards so the release pipeline skips `--latest`/`:latest` on its own when releasing a version older than the current latest.

### DIFF
--- a/.github/workflows/docker-retag.yml
+++ b/.github/workflows/docker-retag.yml
@@ -1,0 +1,42 @@
+name: Docker retag latest
+
+# Repoint jaseci/jaclang:latest at an already-published version's image by
+# creating a manifest alias -- no rebuild, no asset churn. Recovery tool for
+# backport releases: publishing a patch for an older line (e.g. 0.34.8 after
+# 0.35.0) must not leave :latest on the older version.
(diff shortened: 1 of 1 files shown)


kugesan1105:
Closing unmerged: covered by dispatching the existing build-binaries workflow with tag=v0.35.0 (it checks out the tag on the release channel and re-pushes jaseci/jaclang:0.35.0 + :latest). The permanent fix for backport releases moving the latest pointers is #7820.

Issue #7724, opened 2026-07-25
Per-project jac runtime pinning: version-addressable release binaries, honor [project] jac-version, per-deploy pod binary selection


marsninja:
## Motivation

jacBuilder should let every user project be associated with **any jac single-binary release** published on this repo's Releases page, including releases **newer than the jac runtime jacBuilder itself is running on**. A user opens a project, picks `0.35.x` from a dropdown, and their previews, checks, and deploys all run under that exact release, while the platform (builder_sv / jac_coder_sv pods) keeps running its own pinned runtime.

Most of the machinery for this already exists; what is missing is the ability to *name a version*. This issue tracks the jac-side enablers, records the full scoping (including the downstream jacBuilder work for context), and documents the local kind development environment used to develop and test against.

## What already exists (foundations to build on)

1. **The deploy path already executes a foreign jac binary to compile apps.** `jac/jaclang/scale/injector/bundle.jac` (`_precompile_app`, ~L189-232) stages the app, writes the downloaded pod binary to `.pod-jac`, and runs `[.pod-jac, "run", precompile_bytecode.jac, --seal]` as a subprocess, because `.jir` is version-locked and the driver's own runtime cannot produce pod-valid artifacts. An older driver compiling a newer app is already the designed case.
2. **Verified download + content cache exist.** `_fetch_verified_asset` in `jac/jaclang/scale/injector/binary.jac` (~L159-205) downloads release assets with sha256 sidecar verification and caches them.
3. **Multiple binaries coexist safely on one machine.** The launcher's runtime cache keys extractions by payload hash + binary path (`jac/launcher/runtime.zig` ~L247-268), and GC is path-scoped (~L617-644), so N versions at N paths never interfere.
4. **`[project] jac-version` already exists in the schema.** Declared in `jac/jaclang/project/config.jac` (~L32), parsed in `jac/jaclang/project/impl/config.impl.jac` (~L1054), and documented in `docs/docs/reference/config/index.md` as a functional compiler pin.
5. **Release assets are uniform and addressable.** Every tagged release ships `jac-<ver>-{linux-x86_64,linux-aarch64,macos-aarch64}` plus `.sha256` sidecars and a `jac-<ver>-admin-dist.tar.gz`, built by `.github/workflows/build-binaries.yml`.

## Gaps (jac side)

[...]

PR #7824, opened 2026-07-31
scale: decouple the seal toolchain from the shipped pod binary; self-version the seal scripts (fixes #7810)

### LINKED ISSUE #7810: scale: cross-version deploy seal runs the pinned pod binary against the BUILDER's precompile_bytecode.jac, aborting on any version break
## Summary

A jac-scale deploy that pins `[project] jac-version` to a version different from the builder's own jac cannot produce a `.jab` and aborts with "App seal did not complete." The deploy correctly downloads the pinned pod binary, but the seal step (`_precompile_app` in `jaclang/scale/injector/bundle.jac`) runs that pinned binary against the **builder's own** copy of `precompile_bytecode.jac`. Since jac does clean breaks across versions, the builder's seal script calls APIs that no longer exist in the pinned runtime (reproduced: `ext_registry.is_native_module`, removed in 0.35.0), so the pinned binary cannot run the older script and never emits a manifest. Only `pinned == builder` works. This defeats the whole point of the pin (#7713: "deploy on / roll back to a specific release") for any cross-version case, and it means an older platform can never ship a newer pinned app.

## Environment

- Repo: jaseci-labs/jac. Code path present in v0.35.0 (current) and 0.34.x.
- Builder jac 0.34.7 (Linux x86_64); app pins `jac-version = "0.35.0"`.
- Reproduced live from a jachammer prod builder pod (0.34.7). The buggy line is identical in v0.35.0: `jac/jaclang/scale/injector/bundle.jac`.

`_precompile_app` runs `jac_path run <builder's precompile_bytecode.jac> --seal`, where `jac_path` is the downloaded pinned binary and the script comes from the builder via `__file__`. Reproduced that exact invocation.

### A. The failure: pinned 0.35.0 binary + builder's 0.34.7 seal script

```bash
# stage the app, download the pinned (0.35.0) binary, run the BUILDER's 0.34.7 script
/tmp/pod-jac run <0.34.7>/jaclang/utils/precompile_bytecode.jac /tmp/stage --seal
```
```text
Precompiling app for cpython-314...
  Package directory: /tmp/stage/app
✖ Error: module 'jaclang.jac0core.ext_registry' has no attribute 'is_native_module'
  118 |         and not ext_registry.is_native_module(p)
      |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  at sealable_sources() .../jaclang/utils/precompile_bytecode.jac:118
  at main() .../jaclang/utils/precompile_bytecode.jac:499
```
No `_precompiled/MANIFEST.json` is written, so `_precompile_app` raises "App seal did not complete." The 0.34.7 script uses `ext_registry.is_native_module`, which the 0.35.0 runtime no longer exposes.

### B. The fix, proven: pinned 0.35.0 binary + its OWN 0.35.0 seal script

Same binary, same app, but run against the 0.35.0 `precompile_bytecode.jac` (the version-matched script the pinned binary itself ships):
```bash
/tmp/pod-jac run <0.35.0>/precompile_bytecode.jac /tmp/stage --seal
```
```text
  Found 79 .jac files
  ... [1/79] ... [79/79]
  Freezing jac0core bootstrap layer...
  Manifest modules: 79 (0 bootstrap)
  App: kind=web-app caps=['has-server', 'has-entry', 'has-client']
  Manifest: /tmp/stage/app/_precompiled/MANIFEST.json
# rc=0, is_native_module errors: 0, MANIFEST.json produced
```
Seal succeeds and emits a valid manifest. The only variable changed is the seal script version. (App is the jaclang.org site, a real web-app; both runs used the same staged sources.)

## Expected

Pinning `jac-version` to a release different from the builder should produce a valid `.jab` for that release. That is the stated purpose of the pin (#7713) and the "older platform ships a newer app" capability. The seal must be performed with a script that matches the pinned binary, not the builder.

## Actual

`jaclang/scale/injector/bundle.jac`, `_precompile_app` (v0.35.0):

```jac
jac_path = stage_root / ".pod-jac";
jac_path.write_bytes(pod_binary);            # bundle.jac:207  (the PINNED binary)
script = Path(str(__file__)).parent.parent.parent
    / "utils"
    / "precompile_bytecode.jac";             # bundle.jac:209-211  (the BUILDER's script)
result = subprocess.run(
    [str(jac_path), "run", str(script), str(stage_root), "--seal"],  # bundle.jac:214
    ...
);
if result.returncode != 0
   or not (stage / "_precompiled" / "MANIFEST.json").is_file() {
    ... raise "App seal did not complete ..."  # bundle.jac:225
}
```

The pinned binary is fed the builder's `__file__` script. When the builder and pin differ across a version break, the script references APIs the pinned runtime removed, so the run fails and no manifest appears. `_vendor_stage` (bundle.jac:261-272) has the same pattern: it runs the builder's `utils/vendor_wheels.jac` via the pinned binary, so it will break the same way on a vendor-script API change.

Note: the operator-facing error ("Could not prepare the '<channel>' jac runtime ... Check network access to github.com/jaseci-labs/jaseci releases") is misleading. Network was verified fine; the real cause is the seal-script/binary version mismatch.

## Impact

Cross-version deploy is broken. A platform on jac 0.34.7 (e.g. a hosted jacBuilder/jachammer) cannot deploy any user app that pins a newer jac, even though the pin machinery, binary download, and admin-console staging all succeed. It only works when the app pins exactly the builder's version, which makes the pin a no-op rather than the "deploy on a specific release" knob #7713 advertises. Because the coupling is by `__file__`, every future clean break re-breaks the pinned-newer-than-builder case, silently, behind a network-access red herring.

## Workaround (verified)

Pin to the builder's own version (or remove the pin). Verified live: with `jac-version` matching the builder (0.34.7), the seal runs the 0.34.7 script with the 0.34.7 binary and completes. To ship a newer release, the platform image itself must be upgraded to that release so builder == pin.

## Suggested fix

Decouple the seal script from the builder: run the pinned binary against **its own** seal logic. Cleanest is a first-class jac CLI entrypoint so the binary uses its internal, version-matched `precompile_bytecode` rather than an external script path. Add e.g. `jac __seal <dir>` (and `__seal-finalize`, mirroring the existing `--seal` flags) that dispatches to the binary's own `precompile_bytecode.main`, then in `_precompile_app` replace:

```jac
[str(jac_path), "run", str(script), str(stage_root), "--seal"]
```
with
```jac
[str(jac_path), "__seal", str(stage_root)]
```
so no builder-side script crosses the version boundary. Apply the same to `_vendor_stage` (a `jac __vendor-wheels <stage>` subcommand). This makes any builder able to seal any pinned release, which is the actual cross-version deploy capability.

Regression test (network-free): seal a trivial app where the "pinned binary" ships a `precompile_bytecode`/`ext_registry` whose API differs from the builder's script, and assert a valid `_precompiled/MANIFEST.json` is produced. A test that only exercises `pinned == builder` (as today's do) cannot catch this; the drift is exactly the pinned-differs-from-builder path.

Related: #7713 (the `jac-version` pin feature this breaks), #7682.

marsninja:
## Summary

Clean break of the deploy toolchain plumbing so a driver on any platform can deploy any pinned jac version. Fixes #7810 and lands the platform-generality half of #7724 (P0 version-addressable exec/ship resolution, P2 per-deploy pinning hook).

### ResolvedToolchain: exec vs ship
`injector/binary.jac` gains one value object resolved per deploy that separates the binary that PERFORMS the seal/vendor (`exec_path()`, driver platform, served from the verified cache and exec'd in place) from the binary PODS BOOT (`ship_bytes(arch)`, `linux-<node-arch>`). On an all-linux deploy both resolve to the same cached file, so the common path is behavior-identical. On a macOS driver or a cross-arch builder pod (x86_64 builder deploying to arm64 nodes) they differ and the deploy works instead of dying on `[Errno 8] Exec format error`. This retires the 'drive the deploy from a linux container' constraint documented in #7724's kind runbook.

### Self-versioned seal and vendor scripts (#7810)
`_precompile_app` / `_vendor_stage` no longer donate the driver's `precompile_bytecode.jac` / `vendor_wheels.jac` to the pinned binary. A generated 3-line shim (`import from jaclang.utils.precompile_bytecode { main } with entry { main(); }`) makes the pinned binary import the sealer from its OWN payload, so the script version always matches the runtime pods boot. The cross-version contract shrinks from 'every internal API the script touches' to 'module path + argv', which holds for 0.34/0.35. This is exactly the fix experiment B in #7810 validated. The `.pod-jac` write/chmod/unlink temp dance is deleted from both sites.

### Version-keyed binary cache
`scale-binaries` entries are now keyed `jac-<version>-<platform>`; a stable deploy with no pin resolves the latest tag ONCE per deploy and threads that version to both exec and ship downloads, so 'stable' no longer silently mutates under a long-running builder and exec/ship can never skew across a mid-deploy release. Admin-dist caching is version-keyed the same way when pinned.

### Cross-platform fat bundles
`vendor_wheels.jac` accepts `--platform=<os-arch>` and resolves manylinux wheels for the pod platform when the driver host is not the ship platform (`pip download --platform manylinux_2_28/_2_17/2014 --only-binary=:all:`). `_vendor_stage` verifies the VENDOR.json platform stamp, so an older pinned vendorer that silently ignores the flag falls back to a thin bundle instead of shipping wrong-platform wheels.

### Per-deploy channel override
`select_injector(..., channel_override=...)` lets embedders (jacBuilder's threaded deploy manager) pin the channel per deploy instead of racing on process-global `JAC_SCALE_BINARY_PATH`.

### Honest errors + launcher hardening
- ENOEXEC on the seal subprocess now reports 'the driver host cannot execute the jac binary' instead of the misleading check-github-network hint.
- `launcher.zig` reports the actual bring-up error name (`dieBringup`) instead of the catch-all 'payload not materialized?'. This is how the next item was found at all.
- Found while testing: concurrent same-binary launches can transiently fail the launcher's `cacheRoot` probe with `NoWritableCacheDir` (load-dependent, ~50% repro rate with 18 pytest workers sharing one exec path on macOS; never narrated before the error-name change). The seal/vendor runner now retries twice with backoff. The underlying `dirWritable` race deserves its own launcher issue; I will file it separately.

## Testing
- `test_app_precompile` (real end-to-end shim seal), `test_fat_bundle` (6 consecutive green parallel runs, previously ~50% flaky via the launcher race), `test_seed_determinism`, `test_pvc_injector`, `test_binary_cache` all pass.
- `test_binary_injector`: 38 passed plus the pre-existing 7 local failures from the missing `kubernetes` package (identical set on main).
- New tests: local-toolchain exec/ship identity, stable resolve-once pinning, experimental-to-dev mapping, version-keyed cache paths, host_platform naming, shim never leaks into the packed .jab.

## Follow-ups
- CI invariant job: seal one fixture on macos-aarch64 and linux-x86_64, assert identical content_address (guards the bytecode-is-platform-independent assumption at fixed CPython minor).
- Launcher issue for the `NoWritableCacheDir` concurrency race in `dirWritable`.
- jacBuilder-side adoption: replace its five binary-resolution ladders with `resolve_toolchain`/`ResolvedToolchain` and pass `channel_override` per deploy.

### DIFF
--- a/jac/jaclang/scale/injector/bundle.jac
+++ b/jac/jaclang/scale/injector/bundle.jac
@@ -1,7 +1,9 @@
 import io;
 import os;
+import errno;
 import gzip;
 import hashlib;
+import json;
 import logging;
 import shutil;
 import subprocess;
@@ -9,13 +11,13 @@ import tarfile;
 import tempfile;
 import tomllib;
 import from pathlib { Path }
+import from jaclang.scale.injector.binary { ResolvedToolchain, host_platform }
 
 glob logger = logging.getLogger(__name__);
 
 glob APP_VOLUME_NAME: str = "jac-app",
      APP_MOUNT_PATH: str = "/app",
-     JAC_HOME: str = "/app/.jachome",
-     TOOLCHAIN_SUBDIR: str = ".jactoolchain";
+     JAC_HOME: str = "/app/.jachome";
 
 glob _EXCLUDE_DIRS: set[str] = {
          ".jac",
@@ -186,7 +188,68 @@ def open_det_tar(buf: io.BytesIO) -> tuple[tarfile.TarFile, any] {
 }
 
 
-def _precompile_app(base: Path, pod_binary: bytes) -> Path {
+def _write_run_shim(dir_: Path, module: str) -> Path {
+    shim = dir_ / ("__" + module.split(".")[-1] + "_shim.jac");
+    shim.write_text(
+        "import from " + module + " { main }\n\nwith entry { main(); }\n",
+        encoding="utf-8"
+    );
+    return shim;
+}
+
+
+def _run_toolchain_script(
+    toolchain: ResolvedToolchain,
+    workdir: Path,
+    module: str,
+    args: list[str],
+    timeout: int,
+    what: str
+) -> any {
+    import time;
+    exec_bin = toolchain.exec_path();
+    shim = _write_run_shim(workdir, module);
+    try {
+        for attempt in range(3) {
+            try {
+                result = subprocess.run(
+                    [str(exec_bin), "run", str(shim)] + args,
+                    capture_output=True,
+                    text=True,
+                    timeout=timeout
+                );
+            } except OSError as e {
+                if e?.errno == errno.ENOEXEC {
+                    raise RuntimeError(
+                        f"The driver host ({host_platform()}) cannot execute "
+                        f"the jac binary at {exec_bin} for the {what}. Point "
+                        f"JAC_SCALE_BINARY_PATH at a binary built for this "
+                        f"host; the pod binary is downloaded separately for "
+                        f"the cluster's arch."
+                    );
+                }
+                raise;
+            }
+
+            transient = result.returncode != 0
+                and "runtime bring-up failed" in (result.stderr or "");
+            if not transient or attempt == 2 {
+                return result;
+            }
+            logger.warning(
+                f"jac launcher bring-up transiently failed during the {what} "
+                f"(attempt {attempt + 1}); retrying: "
+                f"{(result.stderr or '').strip()[-200:]}"
+            );
+            time.sleep(0.5 * (attempt + 1));
+        }
+    } finally {
+        shim.unlink(missing_ok=True);
+    }
+}
+
+
+def _precompile_app(base: Path, toolchain: ResolvedToolchain) -> Path {
     stage_root = Path(tempfile.mkdtemp(prefix="jac-app-precompile-"));
     stage = stage_root / "app";
     try {
@@ -203,20 +266,15 @@ def _precompile_app(base: Path, pod_binary: bytes) -> Path {
         if not marker.exists() {
             marker.write_text("");
         }
-        jac_path = stage_root / ".pod-jac";
-        jac_path.write_bytes(pod_binary);
-        jac_path.chmod(0o755);
-        script = Path(str(__file__)).parent.parent.parent
-            / "utils"
-            / "precompile_bytecode.jac";
 
-        result = subprocess.run(
-            [str(jac_path), "run", str(script), str(stage_root), "--seal"],
-            capture_output=True,
-            text=True,
-            timeout=600
+        result = _run_toolchain_script(
+            toolchain,
+            stage_root,
+            "jaclang.utils.precompile_bytecode",
+            [str(stage_root), "--seal"],
+            600,
+            "app seal"
         );
-        jac_path.unlink();
         if (
             result.returncode != 0
             or not (stage / "_precompiled" / "MANIFEST.json").is_file()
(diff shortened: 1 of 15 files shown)


jac/jaclang/scale/injector/bundle.jac
```diff
@@ -186,7 +188,68 @@ def open_det_tar(buf: io.BytesIO) -> tuple[tarfile.TarFile, any] {
...
+    toolchain: ResolvedToolchain,
+    workdir: Path,
+    module: str,
+    args: list[str],
+    timeout: int,
+    what: str
+) -> any {
+    import time;
+    exec_bin = toolchain.exec_path();
+    shim = _write_run_shim(workdir, module);
```
MusabMahmoodh, line 211: `workdir` here is `stage_root`, and `stage_root` is exactly what gets handed to the sealer as its argument (`[str(stage_root), "--seal"]`). So while the seal runs, the directory it is seal­ing contains this shim alongside `app/`.

The `finally { shim.unlink(missing_ok=True) }` handles the source file, and it runs before the pack, so `__precompile_bytecode_shim.jac` genuinely cannot reach the tar - the new assertion in `test_app_precompile` proves that much.

What I do not think anything covers: whether the sealer *compiled* the shim on its way past, and left an artifact for it inside `_precompiled/`. `_precompiled/**` is packed (the test at the bottom filters on `n.startswith("_precompiled/")`), and the guard is `assert not any(["_shim.jac" in n for n in names])`, which a compiled artifact would not match on name.

Does the seal skip loose `.jac` files at the root it is given, or would this put a shim entry in `_precompiled/MANIFEST.json`? If it can, that is the bundle-membership class from #7957 - something in the sealed closure that no declared entry reaches - and the cheap fix is writing the shim to a sibling tempdir rather than into the tree being sealed.

MusabMahmoodh (commented):
Reviewing this because it is the oldest open PR in `scale/` and it lands in the injector, which is the surface my bundle-membership RFC (#8077) and #7957 are about. It has never had a review, and it should have had one in July.

The exec/ship split is the right shape. `ResolvedToolchain` separating `exec_path()` (driver platform, performs the seal) from `ship_bytes(arch)` (what pods boot) names the host-vs-target conflation exactly, and it turns an `[Errno 8] Exec format error` into a deploy that works. That is the part I would keep whatever happens to the rest.

**The cross-version contract became implicit.** The body is right that "module path + argv" is a much smaller surface than "every internal API the script touches", and that it holds for 0.34/0.35. But nothing pins it: the pinned binary is asked to `import from jaclang.utils.precompile_bytecode { main }`, and if some future generation moves or renames that module, the failure is an import error inside a `jac run` invoked by the sealer, on a version the driver did not build. Is there a reason not to preflight it - probe the pinned binary once and fail on the driver with "this jac version cannot self-seal" rather than at seal time? That would make the contract checkable instead of assumed.

**It cannot be judged at this head.** CI is red on 3 checks and has been for 20 days, and #8315 landed yesterday across `injector/binary.jac`, `injector/bundle.jac`, `runtimelib/jab_vendor.jac` and `utils/vendor_wheels.jac` - the four files this PR changes most. It needs a rebase before a regression claim on it means anything, mine included.

PR #7826, opened 2026-08-01
feat(desktop): in-process dispatch streaming, shared invocation surface, HTTP sidecar, shared-root fix

chess10kp:
Consolidates the desktop-probe WIP across three concerns, plus a shared desktop HTTP surface.

## Streaming
Generator walkers/functions now drain correctly on the in-process dispatch path (native push channel + local CEF OAuth broker), not only over HTTP SSE. Request context is snapshotted before the invocation coroutine returns and committed once at drain time, fixing the "request context torn down before the generator drains" failure. The async drain loop is consolidated into a transport-neutral helper in `runtimelib/streaming.jac`, shared by the HTTP, broker, and native transports. The base server's `/function` and `/walker` endpoints frame generator results as `text/event-stream` via `sse_frame_stream` (restored after the upstream merge dropped it).

## InvocationDispatcher
Extracts a transport-neutral dispatcher (`runtimelib/invocation.jac` + `impl/invocation.impl.jac`) owning auth, guest-selection, node-targeting, and error normalization. Desktop boot calls `dispatcher.invoke()` directly, keeping a minimal app lean (no FastAPI/Starlette/Pydantic loaded).

**BREAKING:** when a transport pre-validates identity (`identity_prevalidated`, e.g. JWT middleware) and the supplied token resolves to no user, the request now returns 401 on public walkers instead of silently downgrading to guest. Anonymous access with no token is unchanged.

## Shared desktop surface
A shared HTTP surface builder (`runtimelib/http_surface.jac`) and an opt-in HTTP sidecar (`runtimelib/client/targets/desktop/native/http_sidecar.jac`, configured under `[desktop.http]` with `multipart`/`webhooks`/`websockets`/`openapi` capabilities) so the same `InvocationDispatcher` serves browser/native traffic without forking execution semantics. Transport outcomes carry a `DispatchMode` enum (str-based, so `DispatchMode.SSE == "sse"` holds for the broker), `InvocationResult.to_body()` is the single canonical `{ok, data | error}` envelope shared by the webview/bridge/sidecar renderers, and scalar coercion lives once in `invocation`. Design doc: `docs/desktop-portal-architecture.md`; usage guide: `jaclang/cli/skills/jac-desktop-app.md`. Covered by `test_http_sidecar`, `test_http_surface_bridge`, `test_cross_transport_parity`, `test_desktop_routing`, and `test_desktop_oauth_broker`.

## Shared-root fix
`JacRuntime.get_shared_root` no longer goes permanently stale after a guest-root heal. When the cached root id fails to resolve (data store reset, or the execution context swapped out from under a booted runtime), it now invalidates the cache, re-queries the registered resolver once, and retries, so `root.shared` follows the healed root. This also unblocks graph-writing streaming requests that run after a context reset (e.g. across pytest test blocks).

## Also fixed along the way
- `client_runtime_core.impl.jac`: invalid Jac source `typeof raw == "string"` -> `isinstance(raw, str)` (the file isn't compiled by `jac test`, so the syntax error was latent).
- Test helper `_dispatch_request` re-boots when `_runtime is None`.
- CI regressions from the upstream merge: stale symbol names in tests (`forward_outcome`, `AccessLevel.READ`), guide example parse failure, and SSE framing test semantics aligned with upstream's eager-`TypeError` `sse_frame_stream`.


MusabMahmoodh (commented):
This has been open 23 days with no human review and it should not have been - that is on me, it is a scale-touching PR and it was never in any of my sweeps. Read it today.

The part I want on the record, because I think it is right and I think it is mislabelled. The breaking-change note frames the new 401 as a transport-adapter concern ("when a transport adapter pre-validates identity, e.g. JWT middleware"). On the scale path it is unconditional: `scale/server/impl/serve.endpoints.impl.jac` lifts the raw `Authorization: Bearer` value into `token` and passes `identity_prevalidated=True` as a literal at both new dispatch sites. The gate in `runtimelib/impl/invocation.impl.jac` then reads

```
if request.identity_prevalidated {
    username = request.resolved_user;
    if request.token and username is None { ... 401 }
}
```

so on every jac-scale served walker, a request carrying a Bearer token that upstream did not resolve to a username now 401s, public walkers included. The line it replaces was `(username or Con.GUEST.value)`.

I think the change is correct - silently downgrading a rejected credential to guest at an auth decision point is the exact defect class #7900 theme 1 is about, and this deletes it. But the blast radius is the whole scale serving surface, not desktop: a browser holding an expired session that hits a public endpoint is served as guest today and gets a 401 after this lands. That belongs in the release note in those words, because scale operators are the people it will surprise.

Two coordination things before this can go anywhere:

**#8537 overlaps it.** Your own PR from Aug 22 touches `jac0core/impl/runtime.impl.jac` and `runtimelib/client/targets/desktop/native/inprocess_dispatch.jac`, and rewrites the desktop dispatch surface this one restructures. They cannot both land as written - which supersedes which?

**The test file needs a rebase before its claims mean anything.** `scale/tests/server/test_serve.jac` is touched here, and #8287 replaced pytest with the native jac test runner across that suite on Aug 20.

chess10kp:
@MusabMahmoodh — responses to your three points:

**The breaking note** is reworded (32f5ca2df): it now leads with the unconditional scale-server behavior — every `jac-scale`-served walker and function, public endpoints included, 401s a present-but-rejected Bearer — and addresses scale operators directly. You were right that the transport-adapter framing under-described the blast radius.

**The #8537 overlap**: this PR supersedes; #8537 rebases on top. It is currently red on 8 lanes and already conflicts with plain `main`, before any of this branch's changes. I trial-merged the two branches and the actual collision is two hunks — one in `inprocess_dispatch.jac` (which #8537 rewrites anyway, so it resolves in its favor) and one in `store.impl.jac` from its stale main merge. Everything else auto-merges. This branch also keeps the surface #8537 consumes: the bridge-resolved `forward()`/`forward_outcome()` and the widened `set_context`.

**The test file**: the branch had already absorbed #8287's native-runner migration before this landed — `scale/tests/server/test_serve.jac` runs under the jac runner with the bad-token assertions flipped to the new 401. No rebase needed on that account.

The branch is on current `main` (merge of #8648 head) and CI is fully green — ready for review.

MusabMahmoodh:
Taking your three points in order:

**Breaking note** - agreed, that reads right now. Leading with the unconditional scale-server behavior and naming public endpoints explicitly is the version an operator can act on.

**The #8537 overlap** - your call, and I think it is the correct one. You did the trial merge and I did not, so the two-hunk collision finding is yours. `inprocess_dispatch.jac` resolving in #8537's favour makes sense given it rewrites the file anyway. This PR going first, #8537 rebasing on top: settled from my side.

**The test file** - accepted. I checked and `scale/tests/server/test_serve.jac` does carry #8287's native-runner migration with the bad-token assertions on the new 401, so there was nothing for you to do there.

**Now the part that has changed since you wrote that.** "CI is fully green, ready for review" was true when you posted it. About ten hours later #8682 merged (Aug 25 12:11Z, "the program state splits into lifetime organs") and it dissolved `jaclang/runtimelib/` and `jaclang/jac0core/` as packages - 2,573 renamed files. Your last push is from Aug 25 01:44Z, so the green you are looking at was measured against a tree that no longer exists, and GitHub currently reports this PR's mergeability as `UNKNOWN`.

The good news is that every single path you touch is a clean rename in that commit. Nothing was deleted, nothing was redesigned. I mapped the whole set so you do not have to.

**Import rewrites (8):**

| old | new |
|---|---|
| `jaclang.jac0core.constant` | `jaclang.compiler.frontend.constant` |
| `jaclang.jac0core.constructs` | `jaclang.runtime.constructs` |
| `jaclang.runtimelib.builtin` | `jaclang.runtime.builtin` |
| `jaclang.runtimelib.transport` | `jaclang.server.transport` |
| `jaclang.runtimelib.server` | `jaclang.server.server` |
| `jaclang.runtimelib.serving.datatypes` | `jaclang.server.serving.datatypes` |
| `jaclang.runtimelib.serving.sse` | `jaclang.server.serving.sse` |
| `jaclang.runtimelib.client.*` | `jaclang.client.*` |

`jaclang.jac0core.ext_registry` is the one that does **not** move - `jac0core/` survives as a residual Python bootstrap package, so `ext_registry`, `sealed` and `cache_paths` stay exactly where they are. Worth knowing so you do not rewrite that one by reflex.

**File relocations (23 you modify):** `runtimelib/client/**` -> `client/**` (all of `desktop_config`, `desktop_plugin_config`, `targets/desktop/_host_bootstrap`, both desktop target impls, `native/cef/cef_dispatch`, `native/host_boot`, `native/inprocess_dispatch`, `native/oauth_broker`, `client_runtime`, `client_runtime_core`); `runtimelib/server.jac` and `impl/server.impl.jac` -> `server/`; `jac0core/runtime.jac` and its impl -> `runtime/`. Tests move too: `jac/tests/runtimelib/client/**` -> `jac/tests/client/**`, which covers `test_cef_binding`, `test_desktop_native_target`, `test_desktop_oauth_broker`, `test_inprocess_dispatch`, `test_integration` and the `inprocess_app` fixture. `jac/tests/runtimelib/` itself still exists, so `test_serve_client`, `test_serve_streaming` and `test_shared_root` stay put.

**The one real decision is where your five new source files go.** These are yours, so no rename to follow:

- `runtimelib/client/targets/desktop/native/http_sidecar.jac` - unambiguous, it goes to `client/targets/desktop/native/`.
- `runtimelib/http_surface.jac`, `runtimelib/invocation.jac`, `runtimelib/impl/invocation.impl.jac`, `runtimelib/streaming.jac` - these are the genuinely open ones, because they are the shared surface both desktop and the server consume, and shared is exactly what #8682 removed as a category.

The pattern that commit used is **split by consumer**: `client/`, `server/`, `runtime/`, `data/`, `compiler/`. So the question for each is which one owns it. My read, worth what you paid for it: `streaming.jac` and `http_surface.jac` look like `server/` (they sit next to `serving/sse` and `serving/datatypes`, which both landed there), and `invocation` is the harder one because the whole point of your PR is that it is the *shared* invocation surface - `runtime/` is probably the honest home for it, since that is where the consumer-neutral things went. If you disagree I would rather hear why than have you take mine.

I have not done the full review pass yet, deliberately - reviewing 52 files that are about to be relocated wholesale would waste both our time. Push the rebase and I will take it properly, and I will not sit on it for two days this time.

One thing you may want to raise upstream rather than absorb: you are not alone here. I scanned the open queue this morning and 14 open scale and deploy PRs name modules #8682 retired, including two of Jason's (#7721, #8210) and all five of the #7621-#7625 scheduler PRs. Yours is the largest by a distance, which is an argument that the rebase burden is worth flagging rather than quietly eating.

MusabMahmoodh:
Unrelated to my earlier review, but it will bite this branch: #8736 (merged Aug 26) now rejects `refactor` release-note fragments in Contribution Checks, so `release_notes/unreleased/jaclang/7826.refactor.md` fails that lane.

The intended path per `release_notes/unreleased/README.md` is to delete the fragment and get `skip-release-notes-check` added by a maintainer. Deleting it on its own is not sufficient - `check-release-notes.sh` then reports the folder under `MISSING_NOTES` instead - so both halves are needed. That said, given this PR carries a user-visible dispatch change, a `bugfix` or `breaking` fragment may genuinely be the more accurate answer here rather than the label.

chess10kp:
Closing as superseded: the in-process dispatcher (/), desktop SSE streaming (cef/oauth/serve endpoints), and the ServingApp rewrite have all landed upstream through other work. The invocation-envelope model from this branch (InvocationRequest/InvocationResult, coerce_scalar, stream_ack credits) was never integrated; if it's still wanted it should be re-designed against the current ServingApp dispatcher on a fresh branch.

PR #7828, opened 2026-08-01
refactor(scale)!: unify the Kubernetes deploy path - monolith is the N=1 case

MusabMahmoodh:
`jac start --scale` has shipped two parallel deploy engines since the re-architecture: the route-table fleet target, and a ~900-line legacy monolith path that any app without `[scale.microservices.routes]` silently falls into. The legacy path redeploys by **deleting the Deployment, sleeping 5s, and recreating it** - guaranteed downtime on every redeploy - and never got rolling updates, PDBs, `--dry-run`, or the crash-loop-aware rollout wait. It is also the one deploy path with zero e2e coverage in CI.

## Monolith = the N=1 case of the one pipeline

The manifest builder is route-table driven for every N. With no declared routes it derives a solo route from the entry module (`main.jac` -> `main` at `/main`, reserved-name guard applied) and deploys one service behind the gateway - through the same builder, the same idempotent patch-or-create apply, and the same fleet rollout wait as an N-service app. The realizer loses the mode branch entirely; `--dry-run` now works for every app.

Every Deployment - monoliths included - rolls with `RollingUpdate maxSurge=1 / maxUnavailable=0` plus a PodDisruptionBudget. The delete-then-recreate path is deleted, not bypassed.

## One target, honest names

- `KubernetesMicroserviceTarget` -> **`KubernetesTarget`**; the shared base (databases, TLS, destroy, status, scale) is `KubernetesTargetBase`. `deploy/target/kubernetes/microservice/` collapses into `deploy/target/kubernetes/`; `Microservice*` class prefixes drop (`ManifestBuilder`, `DatabaseProvisioner`).
- `kubernetes-microservice` remains a working factory alias.
- The SDK builds one target per `DeploySpec`: `_ms_target`/`_plain_target`/`_namespace_has_fleet` collapse into `_make_target`, and `preview()` works for every spec instead of raising on plain-kubernetes.
- The vestigial `no_image` flag and the provisioner `enabled` gate (both frozen `True`/passthrough since the Docker pipeline was removed) are gone.
- The `shared_ingress`-without-domain fail-fast is ported from the legacy path so it survives the deletion.

## Behavior changes for former monolith deploys

- Served behind the gateway: builtin `/walker/*`, `/function/*`, ... fan out exactly as before, so existing clients keep working unprefixed.
- No auto-installed per-app nginx ingress controller and no auto NetworkPolicies; ingress is the declarative `[scale.microservices.ingress]` / `shared_ingress` config, same as fleets.
- Cost: one gateway pod per app. Buys monoliths zero-downtime rollout, PDB, HPA, dry-run, and rollout self-heal for free.

## Tests

Three additions in the trophy style - real builder, real config, no mocks: the no-routes solo shape (service + gateway, derived route), the reserved-name guard on solo entries, and a manifest render asserting every Deployment carries the surge strategy. Legacy-only tests (the deleted `_build_*`/`_deploy_network_policy` helpers and the NodePort-validation path) are removed with the code they tested. Deploy suites 39/39 green; the microservices suite's failure set is byte-identical to main's baseline.

Net: **+184 / -1330**.

Live validation on a real cluster (littleX as the no-routes monolith, the jac-shop e2e fixture as the 3-service fleet, plus rolling-redeploy verification on both) is running now; results will be posted here.


MusabMahmoodh:
## Cluster validation (EKS, real branch driver)

Deployed both shapes against a live EKS cluster with the branch as the deploy driver.

**N=1, no declared routes** (`jac/examples/todo_app` - walkers + JSX in one file):

```
todo-app-deployment   1/1   RollingUpdate  maxSurge=1  maxUnavailable=0
gateway-deployment    1/1   RollingUpdate  maxSurge=1  maxUnavailable=0
```

Serving, through the gateway:

| request | result |
|---|---|
| `POST /todo-app/walker/read_todos` | 200 - `Proxy:` -> `todo-app-service...:8000/walker/read_todos` |
| `POST /walker/read_todos` (unprefixed) | 200 - `Passthrough:` -> same service |
| `GET /healthz/live`, `/openapi.json` | 200 |

The second row is the compatibility case that mattered most: an existing monolith's client calls `/walker/x` with no prefix, and the gateway's builtin fan-out keeps that working unchanged.

**N=3, declared routes** (`tests/fixtures/k8s_e2e`):

```
products-app-deployment  1/1   RollingUpdate  1 / 0
orders-app-deployment    1/1   RollingUpdate  1 / 0
cart-app-deployment      1/1   RollingUpdate  1 / 0
gateway-deployment       1/1   RollingUpdate  1 / 0
```

`GET /cart/function/view_cart` -> 200, `/products/function/list_products` -> 200, `/cart/function/add_to_cart` -> 422 (route registered, argument validation). Provisioned Redis/Mongo deployments still show the k8s default `25%/25%` - they come from the database providers, untouched here - while every deployment the unified builder emits carries surge-first rolling.

## Two fixes the validation forced

- **Solo service naming.** `_solo_route` derived the service name from the entry module, so every routeless app deployed as `main-deployment` / `main-service` - the app's identity vanished and two apps in a namespace would collide. The name was coupled to the file because the pod entrypoint resolved its module via `find . -name "$JAC_SV_NAME.jac"`. Broken by recording the entry file as the service's `file` config, shipping it as `JAC_SV_FILE`, and having the entrypoint prefer it. This also honors the per-service `file` key that `[scale.microservices.services.NAME]` already accepted but pods ignored.
- **Single-file client detection.** `_has_client_source` only probed for `pages/` or `components/` directories, so an app whose JSX lives in the entry module got no client bundle and its gateway served 404 at `/`. Monoliths never hit this before because they did not go through the gateway. Falls back to the declared `[client]` section.

## Follow-up

#7829 tracks collapsing the gateway pod at N=1 (`gateway_embedded`), which the planner already models but the Kubernetes realizer never consults. Deliberately out of scope here: the current always-gateway behavior is correct, and the collapse has a compatibility surface (static serving, builtin fan-out, unified OpenAPI, rate limiting) that deserves its own PR and e2e leg.

MusabMahmoodh:
Verified the single-file client fix (0fb5c3a61) on EKS end to end.

`jac/examples/todo_app` has no `pages/` or `components/` directory - its JSX lives in the entry module - which is the shape that previously fell through `_has_client_source` and left the gateway with nothing to serve.

| | before | after |
|---|---|---|
| `_has_client_source()` | `False` | `True` |
| `_client_entry()` | `""` | `main.jac` |
| deploy log | *(silent)* | `Building the client bundle on the host (main.jac).` |
| gateway `GET /` | **404** | **200** |

Deployed as `todo-app-deployment` + `gateway-deployment`, both `1/1`, `Fleet ready: 2/2`. The EKS namespaces used for this validation have been torn down, including their retained `efs-sc` PVs.

PR #7838, opened 2026-08-01
fix(checker): type 'default as X' npm imports from the package d.ts

### LINKED ISSUE #7836: checker: `{ default as X }` npm import items always typed Unknown (spurious W1051 at every use) — keyword Token short-circuits get_type_of_module_item before .d.ts default-export lookup
## Summary

Any client-side npm import of the form `import from "pkg" { default as X }` produces a symbol typed **Unknown**, which then emits a spurious `W1051: Expression type could not be resolved (Unknown)` at **every use site** of `X` (never at the import line itself, which makes it confusing to trace).

Named items from the same package are typed correctly from the package's `.d.ts` — the entire TS-declaration pipeline (npm resolution → `types` field → `parse_dts` → `_map_ts_decl`) fully supports default exports and produces a correct type for `default`. It is simply never reached, because `get_type_of_module_item` bails out early for any item whose name is not a `uni.Name` node — and `default` is parsed as a keyword `Token`.

## Reproduction

Fixture layout (no `npm install` needed — the fake package is just checked-in files):

```
repro_default_import/
├── jac.toml
├── repro.jac
└── .jac/client/node_modules/fakepkg/
    ├── package.json
    ├── index.js
    └── index.d.ts
```

`jac.toml`:

```toml
[project]
name = "repro_default_import"
entry-point = "repro.jac"
kind = "web-app"
```

`.jac/client/node_modules/fakepkg/package.json`:

```json
{
  "name": "fakepkg",
  "version": "1.0.0",
  "main": "index.js",
  "types": "index.d.ts"
}
```

`.jac/client/node_modules/fakepkg/index.js`:

```js
export function helper(x) { return x; }
const thing = { render: (s) => s };
export default thing;
```

`.jac/client/node_modules/fakepkg/index.d.ts`:

```ts
export declare function helper(x: string): string;
export interface Thing {
    render(s: string): string;
}
declare const thing: Thing;
export default thing;
```

`repro.jac`:

```jac
import from "fakepkg" { default as thing, helper }

def:pub Repro -> JsxElement {
    a = thing;
    b = helper("ok");
    return <div>{str(a)}{b}</div>;
}
```

Run:

```
$ jac check repro.jac
⚠ warning[W1051]: Expression type could not be resolved (Unknown)
  --> repro.jac:4:9        # a = thing;   <- default import: Unknown
⚠ warning[W1051]: Expression type could not be resolved (Unknown)
  --> repro.jac:6:22       # str(a)       <- fallout from `a` being Unknown
```

Note the asymmetry: `helper` (named item) gets **no** warning — it is fully typed from `index.d.ts`. Proof that the named path is live-typed rather than silently `any`:

```jac
import from "fakepkg" { helper }

def:pub Repro2 -> JsxElement {
    b = helper(42);
    return <div>{b}</div>;
}
```

```
✖ Error: error[E1053]: Cannot assign int to parameter 'x' of type str
  --> repro_typed.jac:4:16
```

**Expected:** `thing` is typed from the `.d.ts` `export default` declaration (a synthesized `Thing` class instance — exactly what `_map_ts_decl` produces for the `default` decl that `parse_dts` already returns for this file). No W1051.

**Actual:** `thing` is `UnknownType` (undiagnosed), so W1051 fires at every use site; nothing is reported on the import line.

## Root cause

1. **Parser** — `jaclang/jac0core/parser/impl/parser.impl.jac` (~L5080): in `{ default as X }`, `default` is `TokenKind.KW_DEFAULT`, so the item name is built with `make_uni_token(...)` — a raw `Token` — whereas ordinary items go through `make_name(...)` and become `Name` nodes. (`ModuleItem.name` is declared `(Name | Token)` to allow this.)

2. **Type evaluator** — `jaclang/compiler/type_system/type_evaluator.impl/imported.impl.jac` (~L6, top of `get_type_of_module_item`):

   ```jac
   if isinstance(node_.name, uni.Name) { } else {
       return types.UnknownType();
   }
   ```

   Every `default as X` item hits this and returns a fresh **undiagnosed** `UnknownType()` before the external-module branch — the one that calls `_get_ts_declared_item_type(from_loc, item_name)` and falls back to `AnyType(is_foreign=True)` — ever runs.

3. Because the Unknown is undiagnosed, no diagnostic is attached to the import statement. The alias symbol carries Unknown, and W1051 is emitted at the first *expression* use instead (`get_type_of_expression`, `type_evaluator.impl.jac` ~L3170).

Everything downstream already supports default exports: verified that for real-world `mermaid@11`, `resolve_client_npm_declarations` finds `dist/mermaid.d.ts` via the package `types` field, and `parse_dts` returns `default → TsDecl(name='mermaid', kind=CONST, ty=REF Mermaid)` with the `Mermaid` interface (15 members) available for class synthesis. The lookup key `"default"` is exactly what `_get_ts_declared_item_type` would be handed if the early return didn't fire first.

## Real-world impact

Hit in a web-app project on both of its `default as` imports:

```jac
import from "mermaid" { default as mermaid }                        # W1051 at every use
import from "embla-carousel-react" { default as useEmblaCarousel }  # W1051 at every use
```

while `import from react { useEffect }` in the same files types perfectly (full `FunctionType` from react's `.d.ts`). Default exports are the *primary* export convention for a large share of npm packages, so client code importing them gets untyped symbols plus warning noise that trains users to ignore W1051.

## Suggested fix

In `get_type_of_module_item`, instead of bailing for non-`Name` items, recognize a `Token` item whose value is `default` and route it through the existing external-module branch (`_get_ts_declared_item_type(from_loc, "default")`, falling back to `AnyType(is_foreign=True)` + the untyped-external note).

One caveat for the implementer: the surrounding code caches results via `node_.name.type = ...`, and `Token` has **no `.type` attribute** (accessing it raises `AttributeError: 'Token' object has no attribute 'type'`). So either skip caching for token-named items, cache on `node_.alias`/the `ModuleItem` itself, or have the parser wrap `default` in a `Name`-like node (which would also fix any other consumer that assumes `ModuleItem.name` has `Name` capabilities).

The fixture above is designed to drop into `tests/compiler/passes/main/fixtures/checker/` — asserting (a) no W1051 for either item, and (b) `helper(42)` still errors with E1053 (guards against "fixing" this by typing everything `any`).

## Environment

- jac 0.35.1 (Darwin arm64), also reproduced against current jaclang source
- Project kind: `web-app` (module in client codespace; `in_client_context()` is True)

marsninja:
Fixes #7836.

## Problem

Any client-side npm import of the form `import from "pkg" { default as X }` produced a symbol typed **Unknown**, emitting a spurious `W1051` at every use site of `X` while named items from the same package typed correctly from the package's `.d.ts`. The entire TS-declaration pipeline (npm resolution, `types` field, `parse_dts`, `_map_ts_decl`) already supports default exports; it was never reached because `get_type_of_module_item` bailed out early for any item whose name is not a `uni.Name` node, and `default` parses as a keyword `Token`.

## Fix

The Token representation for `default`/`*` items is intentional and load-bearing (ES codegen emits `ImportDefaultSpecifier`/namespace specifiers from it, and the symtab pass special-cases it), so the fix lands in the type evaluator rather than the parser:

- **`uni.ModuleItem.binding_name`** (new getter): the `Name` that carries the item's symbol and type -- the item name when it is a `Name`, or the alias for the keyword form (`default as X`). `None` when the item binds no Name (star imports, unaliased `default`).
- **`get_type_of_module_item`** now routes through `binding_name` plus the token's value as the lookup key instead of assuming `node_.name` is a `Name`. The existing external-module branch then resolves `"default"` from the package `.d.ts` exactly like named items, including the synthesized-interface path and the foreign-any `W1102` fallback (and `E1120` under the strict knob) when no declarations exist. No logic was duplicated; every branch of the function now works uniformly for default items.
- **`_set_module_item_symbol_and_type`** is guarded so a symtab hit on a Token-named item cannot touch `Name`-only attributes.

Star imports keep their current behavior (bare Unknown); typing `* as ns` namespace imports properly needs a module-namespace type and is left as follow-up.

## Behavior after the fix

```jac
import from "clsx" { default as cx }   # cx: (...inputs: ClassValue[]) => string
```

- Declared default exports get their real declared type: misuse (e.g. returning `cx("a")` as `int`) is a genuine `E1002`, and member access on a synthesized default-export interface enforces parameter types (`E1053`).
- Packages without declarations fall back to gradual foreign any with `W1102` at the import line instead of an undiagnosed Unknown with `W1051` noise at every use site.

## Tests

Three new tests in `test_checker_npm_imports.jac` (reusing the existing `checker_npm_harness.jac` fixtures; `clsx`'s d.mts already ships `export default clsx`):

1. `default as` import types cleanly: no errors, no `W1051`, no `W1102`.
2. Misuse of the declared default export is a genuine `E1002` (guards against fixing this by typing everything `any`).
3. Undeclared default import stays gradual with `W1102` and no `W1051`.

Targeted runs: `test_checker_npm_imports.jac` (9 passed), `test_checker_ts_interface.jac` (26 passed), `test_checker_ts_modifiers.jac` (2 passed), `test_ts_decl_parser.jac` (23 passed), `test_import_universe_resolution.jac` (5 passed), `test_import_pass.jac` (7 passed). Also verified live against the issue's repro fixture: both `W1051`s gone, named-import `E1053` asymmetry guard intact.


marsninja:
@greptile-apps review

PR #7841, opened 2026-08-01
fix(langserve): goto-definition for string-form imports (npm items, string paths, self-jump guard)

### LINKED ISSUE #7839: LSP goto-definition broken for string-form imports: npm package items self-jump, module-path strings return nothing
## Summary

Goto-definition (LSP `textDocument/definition`) is broken for string-form imports. QA'd by driving `jac lsp` directly with definition requests at every relevant cursor position:

| Cursor on | Actual | Expected |
|---|---|---|
| Item imported from an npm package (`ArrowRight01Icon` from `"@hugeicons/core-free-icons"`) | **Self-jump** — returns the position of the very token queried; in VSCode the cursor silently hops to the start of the same identifier | Jump to `declare const ArrowRight01Icon: IconSvgObject;` in the package's `.d.ts` (or at minimum open the package entry file) |
| Anywhere on a string module path (`"@hugeicons/core-free-icons"`, `"@jac/runtime"`) | **null** — nothing happens, for *all* string-form paths, even ones that resolve to a real `.jac` file | Open the resolved module at the top of the file, same as dotted paths do |
| Dotted module path (`.helper`) — control | ✅ opens `helper.jac` at 0:0 | — |
| Item imported from a `.jac`-backed string module (`Link` from `"@jac/runtime"`) — control | ✅ jumps to `client_runtime.jac:118` (`Link = ReactRouterLink,`) | — |

Found while working on the jaclang.org site (`docs/Markdown.jac`, hugeicons imports). A full sweep there (6 icon names × start/middle of token, 4 positions inside the path string) is consistent: every npm item self-jumps, every string-path position returns null.

## Environment

- jac 0.35.1 (Darwin arm64) — release binary, no dev mode
- Also reproduces against current `main` (7b02170) compiler source via `[dev] jaclang_source`

## Minimal repro project

```
gotodef_repro/
├── jac.toml
├── helper.jac
└── main.jac
```

**jac.toml**
```toml
[project]
name = "gotodef_repro"
entry-point = "main.jac"
kind = "web-app"

[dependencies.npm]
"@hugeicons/core-free-icons" = "*"
```

**helper.jac**
```jac
def:pub helper_fn -> str {
    return "hi";
}
```

**main.jac**
```jac
import from .helper { helper_fn }
import from "@jac/runtime" { Link }
import from "@hugeicons/core-free-icons" { ArrowRight01Icon }

with entry {
    print(helper_fn());
    print(ArrowRight01Icon);
    print(Link);
}
```

Make sure client deps are installed so `.jac/client/node_modules/@hugeicons/core-free-icons/` exists (run a build/serve once, or copy the package in — the resolver only needs the directory present).

### Bug 1 — item imported from an npm package: silent self-jump

1. Open `main.jac` in VSCode with the Jac extension.
2. F12 (Go to Definition) on `ArrowRight01Icon` (line 3).
3. **Actual:** cursor hops to the start of the same identifier — looks like a no-op, no "No definition found" message either.
4. **Expected:** jump to `.jac/client/node_modules/@hugeicons/core-free-icons/dist/types/index.d.ts:705`:
   ```ts
   declare const ArrowRight01Icon: IconSvgObject;
   ```
   The type evaluator already resolves exactly this file via `resolve_client_npm_declarations` for type info, so the target is known to the toolchain.

Script output (harness below):
```
BUG 1:   item          ArrowRight01Icon    L3:C46 -> .../main.jac:3:43   <-- SELF-JUMP (points at the very token queried)
```

### Bug 2 — string module path: no definition at all

1. F12 anywhere inside `"@hugeicons/core-free-icons"` on line 3 (tested at the quote, first char, middle, near end).
2. **Actual:** null / "No definition found".
3. **Expected:** open the package (its types/entry file) at 0:0 — the dotted-path control `.helper` does exactly this.
4. Note this is **not npm-specific**: F12 inside `"@jac/runtime"` (line 2) also returns null, even though that spec resolves to a real file (`jaclang/runtimelib/client_runtime.jac`) and its imported *item* `Link` jumps there fine.

```
BUG 2:   string path  "@jac/runtime"                  L2:C15 -> null (no definition)
BUG 2:   string path  "@hugeicons/core-free-icons"    L3:C15 -> null (no definition)
CONTROL: item          Link                           L2:C30 -> .../jaclang/runtimelib/client_runtime.jac:118:5
CONTROL: module path  .helper                         L1:C15 -> .../gotodef_repro/helper.jac:1:0
```

### Bug 3 — generic fallback returns the queried token itself

`get_definition`'s last-resort branch returns `sym.defn[0]` even when that node **is** the token under the cursor. Any symbol whose only "definition" is itself produces a silent self-jump instead of a clean "no definition" response. This is the mechanism behind Bug 1's symptom, but deserves its own guard so future resolution gaps degrade to null rather than a fake result.

## Root cause

All in [`jac/jaclang/langserve/impl/engine.impl.jac` (`JacLangServer.get_definition`)](https://github.com/jaseci-labs/jac/blob/7b02170f63407b8d95e5bcac942132c319866e27/jac/jaclang/langserve/impl/engine.impl.jac#L294):

- **Bug 1:** the [`uni.ModuleItem` branch (L354)](https://github.com/jaseci-labs/jac/blob/7b02170f63407b8d95e5bcac942132c319866e27/jac/jaclang/langserve/impl/engine.impl.jac#L354) first checks `sym.defn[0]` for a cross-file definition — but for npm items the symbol's only defn is the import token itself, so that fails. It then tries `import_node.from_loc.resolve_relative_path()`, which calls [`resolve_module` (`modresolver.jac` L253)](https://github.com/jaseci-labs/jac/blob/7b02170f63407b8d95e5bcac942132c319866e27/jac/jaclang/jac0core/modresolver.jac#L253) — and `resolve_module` **never consults the npm resolvers**; it naively joins the spec onto the importing file's directory:
  ```
  resolve_module("@hugeicons/core-free-icons", <file>) -> <file_dir>/@hugeicons/core-free-icons   (lang 'py', doesn't exist)
  ```
  `Path(...).exists()` fails, control falls to the [generic `sym.defn[0]` fallback (L412)](https://github.com/jaseci-labs/jac/blob/7b02170f63407b8d95e5bcac942132c319866e27/jac/jaclang/langserve/impl/engine.impl.jac#L412), which returns the queried token → self-jump.
- **Bug 2:** the [`uni.ModulePath` branch (L333)](https://github.com/jaseci-labs/jac/blob/7b02170f63407b8d95e5bcac942132c319866e27/jac/jaclang/langserve/impl/engine.impl.jac#L333) only handles dotted paths via `parent.path` segment matching; string-form paths match no branch and fall through to null.
- The pieces needed already exist: [`resolve_client_npm_declarations` (`modresolver.jac` L589)](https://github.com/jaseci-labs/jac/blob/7b02170f63407b8d95e5bcac942132c319866e27/jac/jaclang/jac0core/modresolver.jac#L589) is what the type evaluator uses (`type_evaluator.impl/ts_declarations.impl.jac`) to find the package's `.d.ts` — the LSP just never calls it.

## Suggested fixes

1. **`ModuleItem` branch:** when the resolved path doesn't exist and the spec is a client npm package, fall back to `resolve_client_npm_declarations` / `resolve_client_npm_package`; ideally locate the named export (`declare const <name>` / `export { <name> }`) in the `.d.ts` for an exact line, else return the entry file at 0:0.
2. **`ModulePath` branch:** handle string-form paths — `resolve_relative_path()` for jac-resolvable strings (covers `"@jac/runtime"`), npm resolvers for package specs.
3. **Fallback guard:** if `sym.defn[0]` is the node under the cursor (or same location), return `None` so editors correctly report "No definition found".

## LSP repro harness

Drop this next to `main.jac` and run `python3 lsp_gotodef_check.py` from the project root (uses `jac lsp` over stdio, no editor needed):

<details>
<summary><code>lsp_gotodef_check.py</code></summary>

```python
#!/usr/bin/env python3
"""Repro: drive `jac lsp` over stdio and print textDocument/definition results."""
import json
import os
import queue
import subprocess
import threading
import time

ROOT = os.path.dirname(os.path.abspath(__file__))
FILE = os.path.join(ROOT, "main.jac")
URI = "file://" + FILE
TEXT = open(FILE, encoding="utf-8").read()
LINES = TEXT.splitlines()

# (label, token to place the cursor on, 0-indexed line, extra col offset)
TARGETS = [
    ("CONTROL: module path  .helper", "helper", 0, 2),
    ("CONTROL: item          helper_fn", "helper_fn", 0, 2),
    ("BUG 2:   string path  \"@jac/runtime\"", "@jac/runtime", 1, 2),
    ("CONTROL: item          Link", "Link", 1, 1),
    ("BUG 2:   string path  \"@hugeicons/core-free-icons\"", "@hugeicons/core-free-icons", 2, 2),
    ("BUG 1:   item          ArrowRight01Icon", "ArrowRight01Icon", 2, 3),
]

proc = subprocess.Popen(
    ["jac", "lsp"], stdin=subprocess.PIPE, stdout=subprocess.PIPE,
    stderr=subprocess.DEVNULL, cwd=ROOT,
)
msgs = queue.Queue()

def reader():
    while True:
        n = None
        while True:
            line = proc.stdout.readline()
            if not line:
                return
            s = line.decode("utf-8", "replace").strip()
            if s.lower().startswith("content-length:"):
                n = int(s.split(":")[1])
            elif s == "" and n is not None:
                break
        body = proc.stdout.read(n)
        if not body:
            return
        try:
            msgs.put(json.loads(body))
        except ValueError:
            pass

threading.Thread(target=reader, daemon=True).start()
rid = [0]

def send(p):
    d = json.dumps(p).encode()
    proc.stdin.write(b"Content-Length: %d\r\n\r\n" % len(d) + d)
    proc.stdin.flush()

def request(method, params, timeout=60):
    rid[0] += 1
    send({"jsonrpc": "2.0", "id": rid[0], "method": method, "params": params})
    end = time.time() + timeout
    while time.time() < end:
        try:
            m = msgs.get(timeout=1)
        except queue.Empty:
            continue
        if m.get("id") == rid[0]:
            return m
    return None

request("initialize", {"processId": os.getpid(), "rootUri": "file://" + ROOT,
                       "capabilities": {}})
send({"jsonrpc": "2.0", "method": "initialized", "params": {}})
send({"jsonrpc": "2.0", "method": "textDocument/didOpen", "params": {
    "textDocument": {"uri": URI, "languageId": "jac", "version": 1, "text": TEXT}}})
time.sleep(20)  # let the server compile/type-check

for label, token, line, off in TARGETS:
    col = LINES[line].index(token) + off
    r = request("textDocument/definition", {
        "textDocument": {"uri": URI},
        "position": {"line": line, "character": col}})
    res = (r or {}).get("result")
    if not res:
        out = "null (no definition)"
    else:
        loc = res[0] if isinstance(res, list) else res
        p = loc["uri"].replace("file://", "")
        s = loc["range"]["start"]
        out = "%s:%d:%d" % (p, s["line"] + 1, s["character"])
        if p == FILE and s["line"] == line:
            out += "   <-- SELF-JUMP (points at the very token queried)"
    print("%-55s L%d:C%d -> %s" % (label, line + 1, col, out))

proc.kill()
```

</details>

Full output on jac 0.35.1:

```
CONTROL: module path  .helper                           L1:C15 -> .../gotodef_repro/helper.jac:1:0
CONTROL: item          helper_fn                        L1:C24 -> .../gotodef_repro/helper.jac:1:8
BUG 2:   string path  "@jac/runtime"                    L2:C15 -> null (no definition)
CONTROL: item          Link                             L2:C30 -> .../jaclang/runtimelib/client_runtime.jac:118:5
BUG 2:   string path  "@hugeicons/core-free-icons"      L3:C15 -> null (no definition)
BUG 1:   item          ArrowRight01Icon                 L3:C46 -> .../gotodef_repro/main.jac:3:43   <-- SELF-JUMP (points at the very token queried)
```

marsninja:
Fixes #7839.

## Problem

Goto-definition was broken for string-form imports, QA'd by driving `jac lsp` directly (full matrix in the issue):

1. **npm items self-jump**: F12 on an item imported from a client npm package (`ArrowRight01Icon` from `"@hugeicons/core-free-icons"`) returned the position of the very token queried -- a silent no-op in VSCode. `resolve_module` never consults the npm resolvers, so the resolution attempt fell through to the generic `sym.defn[0]` fallback, which returned the import token itself.
2. **String module paths return nothing**: F12 anywhere inside a string path (`"@jac/runtime"`, `"@hugeicons/core-free-icons"`) returned null for *all* string-form paths, even ones that resolve to a real `.jac` file. Root cause is one level below the branch logic: string path tokens carry no semantic token, so `get_token_at_position` returns None and `get_definition` bailed before reaching any branch.
3. **Self-jump fallback**: the last-resort `sym.defn[0]` branch happily returned the queried token whenever a symbol's only "definition" is itself, so every future resolution gap degrades to a fake result instead of a clean "no definition".

## Fix

All in the langserve engine, reusing the existing npm resolution stack (`is_client_npm_import` / `resolve_client_npm_declarations` / `resolve_client_npm_package`) that the type evaluator already uses for `.d.ts` typing -- the LSP just never called it:

- **`ModuleItem` branch**: when the joined module path does not exist, `_npm_item_location` resolves the package declarations and lands on the item's exact declaration line inside the `.d.ts` via a new `utils.find_dts_decl_range` scanner (declaration statements win over export lists; `default` maps to `export default`). Falls back to the declarations file at 0:0, then the runtime entry file.
- **String paths**: `get_definition` now routes the no-semantic-token case to `_string_import_definition`, which finds the raw `String` token via the existing `get_node_at_position`, then resolves through `_module_spec_target`: `resolve_relative_path()` first (covers `"@jac/runtime"` and other jac-resolvable specs), npm declarations/entry second.
- **Fallback guard**: if `sym.defn[0]` *is* the token under the cursor (identity or same location), return None so editors report "No definition found".

`_module_spec_target` mirrors the type evaluator's spec derivation (`string_path_value or dot_path_str`), so bare-name client npm imports (e.g. `import from react`) get the same treatment for free.

## Validation

Reproduced the issue's repro project (synthetic `@hugeicons/core-free-icons` under `.jac/client/node_modules/`) against current main, then re-ran post-fix:

```
before:
  string path "@jac/runtime"               -> null
  string path "@hugeicons/core-free-icons" -> null
  item ArrowRight01Icon                    -> main.jac:2:43 (self-jump)
after:
  string path "@jac/runtime"               -> jaclang/runtimelib/client_runtime.jac:0:0
  string path "@hugeicons/core-free-icons" -> .../core-free-icons/dist/types/index.d.ts:0:0
  item ArrowRight01Icon                    -> .../dist/types/index.d.ts:5:14  (declare const ArrowRight01Icon)
```

Controls (dotted `.helper` path, `helper_fn`, `Link` from `"@jac/runtime"`) unchanged.

## Tests

New `test_go_to_definition_string_imports` in `jac/tests/langserve/test_server.jac`, reusing the shared `checker_npm_harness` npm-project fixtures. Covers all three bugs: exact `.d.ts` declaration ranges for `clsx` (exports-map `.d.mts`) and `twMerge` (`declare const` + export list), string-path resolution for npm packages and `"@jac/runtime"`, and None (not a self-jump) for an item from a missing package. Written first against main to confirm it fails with the issue's exact symptoms.

Targeted runs: `jac/tests/langserve/` 39 passed + 1 pre-existing failure (`test_cache_invalidation`, fails identically on clean main), `test_checker_npm_imports.jac` 9 passed. `jac fmt` applied; `jac check` on `engine.jac`/`utils.jac` stays exactly at the main baselines (no new diagnostics).


marsninja:
CI status: the jac-check failure was the lintfix gate stripping explanatory comments in engine.impl.jac -- fixed in d8e49db04f (comments removed via jac fmt --lintfix).

The test-runtime failure is not from this PR: the guide-examples gate is red on main since #7840 (jac-cl-components block #4 is a bare ability fragment failing E2018), so every PR inherits it through the merge preview. Fix is up separately in #7844; once that merges, a re-run here should go green.

PR #7845, opened 2026-08-02
probe: nightshift gh seam check (closing immediately)

ayushmk7:
One-off probe of the harness gh write seam. Closing within a minute.


ayushmk7:
Probe complete: gh write seam verified. Closing.

PR #7856, opened 2026-08-02
feat(runtime)!: queries as data — GraphQuery IR, planner, storage-owned graph indexes (clean break, #7855)

### LINKED ISSUE #7790: Topology index fast path: `[-->[?:T]][?f==x]` silently returns all nodes; O(N^2) edge writes; field pushdown unreachable

### LINKED ISSUE #7855: Clean break: queries as data — GraphQuery IR, planner floors, and storage-owned graph indexes (retires the topology-index blob)
## Summary

#7790 documented three problems in the topology-index path: a silent wrong-answer bug in `[-->[?:T]][?f==x]`, O(N²) edge writes, and field pushdown that never reaches storage. They are three symptoms of one architectural absence: **queries never exist as data.** The compiler lowers the graph-query surface syntax into Python lambdas; the runtime then reverse-engineers intent out of bytecode (`dis` opcode scans, `getclosurevars`) and fails silently in the unsafe direction whenever a filter is wrapped or composed; and the storage layer never hears about intent at all, so predicates arrive as opaque callables and the topology index has to live as an application-level blob bolted onto anchor persistence.

This issue is the clean break: **queries compile to a closed, typed IR; a planner executes it against storage-owned indexes; predicates never touch objects; and the old machinery is deleted, not wrapped.** It supersedes the topology-index blob entirely, and it supersedes #7807 (moving the blob to the system root centralizes a structure this plan deletes; the access-bypass and blob-shipping machinery there should not land).

## Design tenets

These are decided, not open for per-PR relitigation:

1. **No fallbacks.** There is no legacy path, no compat shim, no feature flag toggling old vs. new behavior, and no "try the new way, fall back to the old way" anywhere. Each capability has exactly one implementation; where the old code disagrees with the new model, the old code is deleted in the same PR.
2. **Closed query language.** `[?...]` admits only the typed, structural predicate forms the spec enumerates. Anything else is a **compile error** with a fix-it (move it to a list comprehension, where object-space cost is visible). There is no residual-lambda floor and no callable-filter API in the query path.
3. **Predicates never see objects.** Every predicate evaluates either inside a storage engine (when translation is exactness-proven) or in Python against ~60-byte index rows. Object deserialization happens exactly once per query, for winners.
4. **Indexes are derived state.** Maintained inside the storage write path, rebuildable from committed edges, never persisted as user-data fields, never shipped as encoded values between tiers.
5. **Everything is counted.** Deserialization counters, per-tier resolution counts, and `EXPLAIN` are part of the runtime, not debug scaffolding. Invariants are enforced by tests, not review vigilance.

## The root cause, stated once

Every database separates three things Jac currently conflates:

| Layer | What a DB does | What Jac does today |
|---|---|---|
| Query representation | AST → logical plan (data) | Python closures (opaque) |
| Execution strategy | derived from the plan | guessed from bytecode of the closures |
| Derived state (indexes) | maintained inside the storage engine's write path | serialized blob on a root anchor, decode-modify-encode per edge |

All three #7790 problems are leakage across those missing boundaries:

- **P1 (correctness):** `refs()` fuses a trailing `[?pred]` into the hop filter by wrapping two lambdas in a third (`jac/jaclang/jac0core/impl/runtime.impl.jac`). The wrapper's bytecode contains no recognizable opcodes and its closure holds no archetype type, so `_filter_has_predicates` / `_filter_type_name` (`jac/jaclang/runtimelib/query_utils.jac`) report "no predicates, no type", the index fast path runs with an empty filter triple, and the lambda is never called. Every node comes back, silently.
- **P2 (writes):** `NodeAnchor.get_topology_index` / `set_topology_index` are unconditional full decode / full encode (`jac/jaclang/jac0core/impl/archetype.impl.jac`), and the `on_edge_created` hook runs them per edge (`jac/jaclang/runtimelib/impl/topo_utils.impl.jac`). Edge N pays for the N−1 before it: 8,000 nodes ≈ 7 minutes.
- **P3 (reads):** `QueryPlan.field_predicates` exists, `MongoBackend` declares and implements `field_pushdown`, but the single `QueryPlan(...)` construction site passes the predicate as a Python callable and always builds `chain=[]`. The intended plan/capability architecture was half-built and then bypassed.

## Spec decisions

1. **Filter purity contract.** `[?...]` predicates are pure and may be evaluated zero or more times, in any order, against any candidate set (SQL `WHERE` semantics). Comparator right-hand sides evaluate exactly once, at query construction.
2. **Closed, typed filter sublanguage.** The spec enumerates the admitted predicate forms: `field op value` where `field` is a declared `has` field of the hop's node/edge type, `value` is a scalar expression (`str`, `int`, `float`, `bool`, `UUID`, ISO datetime string, `None`), and `op` ∈ {`==`, `!=`, `<`, `<=`, `>`, `>=`, `in`}. Each (type, op) pair has exact semantics defined once in the spec; a backend may execute a predicate only if its translation is proven equivalent, otherwise the predicate evaluates in row space (below) — never against objects. Anything outside the sublanguage is a **compile error** (new E-code, replacing today's silent `W1051`) with a fix-it pointing to a list comprehension.
3. **Correctness invariant (normative):** every spelling denoting the same query — `[-->[?:A, p]]`, `[-->[?:A]][?p]`, `[-->][?:A][?p]` — produces identical results under every execution strategy, index projections warm or cold. Enforced by a differential harness across the spelling matrix. This bug class has occurred twice ("same fast path, different trigger"); only an equivalence test catches the third.
4. **Cost invariant (universal):** anchors deserialized per query == |result|, for **every** query — the closed sublanguage leaves no exceptions. Enforced by a per-query deserialization counter asserted in tests and surfaced in `EXPLAIN`.
5. **The OSP memory model — consistency is named, not implied.** **default** — read-your-writes for the writing request, per-object repeatable read within a request, cross-pod eventual consistency with a bounded staleness window, no cross-hop snapshot guarantee; **`fresh`** — a per-query hint that executes the read against the system of record at cold-path cost; **`snapshot`** — reserved and defined now, unimplemented, so programs are written against a contract that only ever strengthens. Governing rule: a cache may never advertise more than the transaction layer beneath it.

### GraphQuery IR

Edge-ref trailer chains — including trailing `[?...]` filters, slices, and the chained spellings that today become `filter_on` wrappers outside the query — compile to one structured value:

```
GraphQuery { origin, hops: list[HopIR], edges_only, slc }
HopIR      { direction, edge_type: type|None, node_type: type|None,
             edge_preds / preds: list[(field, op, value)] }
```

Types are class references, predicates are data, comparators are evaluated once. **There are no callables in the IR** — it is fully serializable, and the native runtime and future wire-level pushdown consume the same value. Consecutive trailing `[?...]` on an edge-ref fuse structurally into the last hop, which rescues the `[root-->][?:Session][?id==sid]` spelling (9+ live call sites) with zero application rewrites. `refs(GraphQuery)` is the only runtime entry point; the callable-accepting `ObjectSpatialFilter` surface is deleted. `filter_on` survives solely as a plain list utility outside the query path.

### Planner: two execution tiers, one overlay, one verification

The planner consumes the IR with zero introspection and resolves each hop in **id space**:

- **Engine execution** (cold): the backend executes type + exactness-proven predicates + slice inside its own indexes; only matching rows/ids cross the wire. Predicates whose backend translation is not exactness-proven are returned as id+column rows and evaluated by the planner **in row space** — Python semantics, native exactness, still no objects.
- **Row execution** (warm): the process-scoped projection resolves hops against locally resident index rows (adjacency + predicate columns), microseconds per hop.
- **Overlay composition:** the request's uncommitted-write overlay composes over every result — read-your-writes is unconditional.
- **Winner verification:** materialization re-checks predicates against the fetched winner objects. This is part of the consistency contract (it is what makes stale-row admits self-healing), not a fallback — it costs |result| comparisons on already-fetched objects.

There is **no walk tier**. Volatile memory maintains live adjacency like every other backend, so even in-memory graphs resolve via a projection; `edges_to_nodes` survives only for walker `visit` semantics, which are not queries. There is no residual tier; the compile-time sublanguage boundary makes one unnecessary. `EXPLAIN` (grown from `JAC_QUERY_TRACE`) prints the logical IR, the tier that resolved each hop, and projected vs. actual counts.

## Storage-owned indexes (index = derived state)

- **SQLite:** a `graph_edges(src, dst, edge_type, undirected)` table with indexes on `(src, edge_type)` / `(dst, edge_type)`, maintained transactionally inside `apply(ChangeSet)` — the ChangeSet already carries edge creates/deletes and edge-list deltas. Typed hops join against the existing `arch_type` column; exactness-proven predicates use `json_extract(data, '$.archetype.<field>')`. Per-edge maintenance is O(1) forever; P2's quadratic is structurally gone, not cached away.
- **Mongo:** compound indexes plus chain execution in `execute_plan` as iterative `id_in` batches per hop (one strategy, deliberately — no `$graphLookup` variant to maintain). The existing `data.archetype.<field>` predicate translation finally becomes reachable, gated per (type, op) on proven equivalence.
- **Volatile:** live in-memory adjacency + column dicts (today's `TopologyIndex` structure minus encode/decode), maintained on mutation, never serialized.
- **`jac tool reindex`:** rebuild-from-committed-edges as a first-class operation. A derived structure's corruption story is a rebuild, never a correctness event — and nothing about the index is user data, so no access-check bypasses exist anywhere.

## Process-scoped index projection (final phase)

Three explicit lifetimes: **request** (`__mem__` objects + uncommitted-write overlay), **process** (the index projection), **cluster** (Redis anchor cache + per-anchor invalidation bus; Mongo truth).

The projection holds, per completeness-marked node: full typed adjacency **plus the predicate columns** (~50–60 B/node — two orders of magnitude smaller than objects, which is why rows can live at process scope while objects stay request-scoped). Warm-on-scan: engine responses piggyback index rows for the frontier they scanned, size-capped; a capped-out frontier page-faults later with one narrowed ids-only sub-plan and converges (cold → one shrinking fault → fully local). Buffer-pool semantics — cache index pages with per-node completeness, never query results, so invalidation stays per-node and query-cache invalidation problems never exist.

Redis keeps exactly its two current roles — anchor cache (winners only, by construction: losers never exist as anchors) and per-anchor invalidation bus — plus one designed-for future slot: `RedisBackend.capabilities()` may later declare `hop_pushdown` with native adjacency structures if fleet p99s justify it. **Never** store an encoded index structure in Redis as a value — that recreates the blob one tier up.

## Consistency model

The projection and caches implement the **named** contract of spec decision 5 — not incidental cache behavior.

| Level | Guarantee | Cost |
|---|---|---|
| default | RYW for the writer · per-object repeatable read within a request · cross-pod eventual with a bounded window · no cross-hop snapshot | warm path (µs–ms) |
| `fresh` (per-query hint) | reads the system of record via engine execution | cold path (ms) |
| `snapshot` (reserved) | per-request consistent snapshot — future MVCC upgrade | n/a today |

**Staleness mechanics.** Same-pod is never stale: the writing pod updates its own projection synchronously in the write path, and the request overlay covers uncommitted writes. Cross-pod staleness is confined to the commit→eviction window (~bus latency):

- *False positives self-heal.* Winner verification re-checks predicates against fetched objects, so a stale-row admit is caught at materialization (a deleted node degenerates to a dangling id that materialization drops). At worst the query serves an internally consistent, slightly-old committed snapshot.
- *False negatives are the honest cost.* A just-committed remote change can be missed for the window. This is the platform's existing consistency class — Redis L2 anchor invalidation has the identical window today, and the blob GTI had it with whole-index blast radius — narrowed here to per-node granularity.
- *What staleness can never do:* skip a predicate (stale ≠ unfiltered — the P1 bug class stays unreachable), expose uncommitted remote data (the projection holds only committed rows), or corrupt anything (derived state; the worst outcome of doubt is eviction plus one round trip).

**Required backstops for lost invalidations** (pub/sub is fire-and-forget; these land with the projection PR, non-negotiable): on any bus disconnect/reconnect, drop **all** completeness marks; a periodic bus heartbeat/epoch so a wedged-but-connected subscription triggers the same flush (or migrate the bus to Redis Streams for replayable delivery); a TTL on completeness marks bounding worst-case staleness even for a silently broken pod. Write ordering is commit-then-publish, and messages are id-only evict-and-refault — they carry no state, so they cannot be applied out of order.

**Write atomicity.** The commit unit is the request's ChangeSet, applied atomically per backend (Mongo multi-document transaction, SQLite transaction). Read-consistency guarantees are meaningless if a crash can apply half an edge.

**Upgrade path to `snapshot`.** Because queries are data and execution tiers are negotiated, snapshot isolation arrives later as a plan attribute (read timestamp) plus versioned index rows and Mongo snapshot sessions; tiers that cannot honor the requested level are skipped for that query. No re-architecture — deferred until traversal-anomaly evidence justifies MVCC cost (see non-goals).

Ops note: session-affine routing stretches read-your-writes into read-your-recent-writes for the human-visible case without changing the model.

## What we remove

- The `post_filter` lambda fusion in `refs()` and the entire lambda-filter plumbing: `ObjectSpatialFilter`, callable `nd`/`edge` on `ObjectSpatialDestination`, the path-builder query surface. `refs(GraphQuery)` is the only entry point.
- **All of `query_utils`** (interface + impl): `dis` opcode scanning, `getclosurevars`, `_CodeKey`, the code-object LRU caches. The archaeology is not hardened; it ceases to exist.
- **The walk execution path for queries**: `_hop_via_walk`, `_hop_via_walk_ordered`, and the resolver's per-hop fold. `edges_to_nodes` remains only for walker `visit`.
- `TopologyIndex.encode`/`decode`, `TOPO_VERSION` + migrations, `NodeAnchor.topology_index_data` and its accessors, the `__getstate__` special case, and the serializer's b64 handling in both directions. Stale `topology_index_data` keys in stored anchors are ignored on read and dropped on next write — a key filter, not a decode path.
- **All of `topo_utils`**: the four mutation hooks and their call sites in `runtime.impl.jac`.
- The `[run] topology_index` config knob and `JAC_TOPOLOGY_INDEX` env var — access paths are not user-visible configuration once they cannot change results.
- The cross-pod GTI blob invalidation machinery (#7335 codepaths) and Mongo GTI blob persistence.
- **#7807**, closed rather than rebased: its motivation (a global view for cross-user resolution) is served by backend-owned indexes, which are naturally global; its persistence machinery (access-check bypass in `_collect_into`, `put_partial` blob shipping, process-cache eviction dance) is scaffolding for the model this removes.

## Worked example and cost model

```jac
sprint_pool = [root
    -->[?:Workspace, name == "eng", active == True]
    ->:Contains:archived == False:->[?:Project, status == "active", priority >= min_pri]
    -->[?:Task, state == "open", estimate <= 8, due <= sprint_cutoff(days=14)]
][0:25];
```

Population: 40 workspaces → 1; 400 projects → 12 (edge + node predicates); 2,400 reachable tasks → ~180 → slice 25. Every clause above is inside the closed sublanguage (`sprint_cutoff(...)` is a comparator, evaluated once at query construction). A correlated check like "has a verified owner assigned" is written as a list comprehension over `sprint_pool` — deliberately outside the query, where its object-space cost is visible.

| | Today | Clean break (cold) | Clean break (warm pod, 2nd request) |
|---|---|---|---|
| Mongo round trips | many (per-frontier batch gets) | 1 (engine execution) | 0 (row execution; Redis materializes) |
| Anchors deserialized | **2,840** | **25** | **25** |
| Redis after the query | ~2,815 losers promoted | 25 winners | 25 winners |
| Wall clock (order) | 100s of ms | ~5–12 ms | ~1 ms |

## Landing plan

The break is complete and shippable after PR 3; PR 4 adds the warm path.

- [ ] **PR 0 — spec + checker:** filter purity contract; the closed, typed filter sublanguage (admitted types/ops with exact per-backend semantics; new E-code + fix-it replacing `W1051`); the OSP memory model (default + `fresh`; `snapshot` reserved); release notes for the breaking semantics. The differential spelling-matrix harness lands here and gates everything after.
- [ ] **PR 1 — IR + codegen + planner:** `GraphQuery` emitted by `exit_filter_compr` / `exit_atom_trailer`; planner with engine/row execution and overlay composition; **deletes** the `refs()` fusion, all of `query_utils`, the walk path, and the callable-filter API. From this commit the #7790 P1 bug class is unrepresentable and all spellings are equivalent. `EXPLAIN` v1.
- [ ] **PR 2 — storage-owned indexes:** SQLite `graph_edges` + ChangeSet-transactional maintenance; atomic ChangeSet application per backend; volatile adjacency; `jac tool reindex`; auto-reindex on first open of an existing DB (`db_format_version` bump; rebuild reads committed edges, never the old blob); **deletes** the blob (`topology_index_data`, encode/decode, `topo_utils`, config knob, GTI cross-pod machinery). Deserialization counter + universal cost-invariant tests.
- [ ] **PR 3 — engine pushdown:** Mongo chain execution (iterative `id_in` batches) + field predicates gated on exactness proofs; SQLite `json_extract` predicates; row-space evaluation for unproven (type, op) pairs; winner verification at materialization.
- [ ] **PR 4 — process-scoped index projection:** completeness-marked adjacency + predicate columns; warm-on-scan with size cap; page-fault sub-plans; bus-driven per-node eviction; the `fresh` hint; the lost-invalidation backstops (reconnect flush, heartbeat/epoch, completeness TTL); two-pod stale-read tests.

Each PR is independently green under the harness; none requires the next to be correct.

## Acceptance criteria

1. The #7790 P1 repro (`nested` / `chained` / `mixed` spellings) returns identical, correct results warm and cold — as a permanent differential test, not a one-off.
2. Attaching 8,000 nodes to one root is linear: flat per-batch time, and per-edge index maintenance does no full-structure re-encode (asserted by counter, not wall clock).
3. `[root-->[?:Session, id == sid]]` on Mongo executes with `field_predicates` populated and deserializes exactly |result| anchors.
4. `[root-->[?:A, val == 2]]` is at least as fast as `[root-->[?:A]]` on the same graph — a more selective query never reads more.
5. A predicate outside the closed sublanguage fails at compile time with the new E-code and a fix-it; nothing outside the sublanguage reaches the runtime.
6. `jac tool explain` (or `JAC_QUERY_TRACE=plan`) prints logical IR, per-hop resolution tier, and projected vs. actual counts for any query.
7. Grep-clean: no `dis.get_instructions`, no `getclosurevars`, no `topology_index_data`, no `TOPO_VERSION`, no `_hop_via_walk`, no `ObjectSpatialFilter` in the tree.
8. A fresh checkout on an existing `.jac/data` DB auto-reindexes on first open with no data loss and no user action.
9. Two-pod consistency test: a write committed on pod B is (a) reflected immediately by a `fresh`-hinted query on pod A, and (b) reflected by a default query on pod A once invalidation is delivered; with the bus severed, pod A converges within the completeness TTL and never returns an unfiltered or uncommitted result at any point.

## Breaking changes

- **Purity contract:** side-effecting `[?...]` filters change behavior (comparators evaluate once; predicates may run zero-or-more times). Loud release note.
- **Closed sublanguage:** filter forms outside the spec — arbitrary expressions, method calls, cross-field compares, callable filters via the runtime API — become compile errors with fix-its. The migration for each is mechanical (a list comprehension), and the error message says so.
- **Query API:** the callable-accepting `refs`/path-builder surface is removed; `refs(GraphQuery)` only.
- **DB format:** one-time auto-reindex on first open; `.jir` caches invalidate via the normal compiler-version bump.
- **Access control in id space:** the index projection and `graph_edges` carry the access-relevant columns (owner root, permission level) so permission filtering runs in row space — materialize-to-check is not reintroduced. (#7807's removal of ownership from the index moves opposite to this and is another reason to supersede it.)

## Non-goals (deliberately deferred)

- **Snapshot isolation per request** (MVCC in the projection, plan-carried read timestamps, Mongo snapshot sessions) — the upgrade path is designed and the `snapshot` level is reserved in the spec; build it when traversal-anomaly reports justify MVCC cost, not before.
- Redis as a query-executing tier (`hop_pushdown` via native adjacency structures) and RediSearch-style field indexes — the capability slot exists; build only when fleet p99s justify it. (Migrating the invalidation bus to Redis Streams is a mechanical hardening, not a model change, and can happen any time.)
- Semi-join planning for correlated sub-queries; cross-field compares (`spent > estimate`); aggregations; ordering-by-field; field projection — each admitted later only with an answer to "does it run in id space?" and a sublanguage spec amendment.
- Ids-only engine responses when L2 is likely warm (covering-index-then-key-lookup knob) — measure first.

## References

- #7790 — the measured analysis this plan roots in (P1/P2/P3, repros, numbers).
- #7807 — superseded by this plan (see "What we remove").
- jacBuilder#895 — the production OOM report (session lookups materializing every session); fixed at the runtime layer by PR 3 with zero app rewrites.
- Prior art in-tree: `QueryPlan.chain` / `needs()` / `capabilities()` / `execute_plan` — the dormant skeleton this plan completes; the native runtime's per-node adjacency store — the same direction, already taken on the native path.

marsninja:
Implements #7855 — the queries-as-data clean break — phases 0–3 complete on one branch. No fallbacks, no legacy paths, no toggles: **+~1,900 / −4,900 net**, with the deletion sweep alone removing 4,488 lines.

## What landed

- [x] **IR** — `GraphQuery` / `QHop` / `QPred` in `jac0core.graph_query`: types as class references, predicates as validated `(field, op, value)` data, comparators evaluated once, zero callables. `QPred.matches` is the reference semantics every engine must agree with. Identity filters ride `QHop.only`; trailing filters compose via `GraphQuery.narrow()`.
- [x] **Storage-owned index (SQLite)** — `graph_edges` + `graph_types` (type ancestry for subtype-correct hops) maintained transactionally inside `apply(ChangeSet)`; O(1) per edge, forever. Opening a pre-index DB auto-reindexes from committed anchors (never any legacy encoding) and stamps `db_format_version 2`. `hop_rows` returns type-narrowed `(edge_id, dst_id)` pairs; `filter_ids` evaluates predicates with `json_type`-guarded conditions that can neither over- nor under-approximate `QPred.matches`.
- [x] **Planner** — hops resolve in id space: engine rows + request-overlay composition (fresh uncommitted edges in, detached edges out) with predicates routed by residency (live `matches` for resident anchors → read-your-writes on dirty fields in both directions; engine `filter_ids` otherwise). Objects deserialize exactly once, for winners, where the last hop re-verifies (the stale-admit self-heal). `JAC_QUERY_TRACE` prints per-hop tier counts.
- [x] **Codegen swap** — `exit_edge_ref_trailer` emits `refs(GraphQuery(...))`; chained trailing filters fuse structurally through `_is_edge_ref_target` (the jacBuilder#895 spelling becomes one query); anything outside the closed sublanguage is **E5094** at compile time with a fix-it. `visit`, edge-only queries, identity filters, and undirected traversal verified through the new pipeline.
- [x] **Mongo** — speaks the graph-index contract natively: `hop_rows` over edge docs (source/target already in `data`), `graph_types` ancestry collection maintained on write, `filter_ids` via type-bracketed BSON comparisons (same exactness class as the SQLite guard). `QueryPlan`/`capabilities`/`execute_plan` deleted. *Not integration-tested in this environment (no Mongo) — needs a scale-CI pass.*
- [x] **Deletions** — `query_utils` (all bytecode archaeology), resolver, old planner, the object-walk query path, `ObjectSpatialPath`/`Destination`/`Filter`, the `post_filter` fusion, `TopologyIndex` + `TOPO_VERSION` + `topology_index_data` + serializer/`__getstate__` special cases, all of `topo_utils`, the `[run] topology_index` knob, the scale GTI L2 special case.
- [x] **Differential harness** — spelling matrix (`nested`/`chained`/`mixed` + slices, ordering, edge preds, two-hop chains) asserted equivalent cold and warm through real syntax; cost invariant asserted through real syntax (`L3 fetches == |result|` for pushable queries; selective never reads more than broad).
- [ ] **Process-scoped index projection** (issue phase PR 4 — the warm path: completeness marks, pushable columns, `fresh` hint, lost-invalidation backstops, two-pod tests). Per #7855 the break is complete and shippable without it; it needs a Redis/multi-pod harness this branch's environment lacks. Follow-up PR.

## Verification (targeted, per owner instruction — no broad suite runs)

- 36 tests across 9 runtime suites green (`graph_query`, `graph_edges`, `query_planner`, `query_differential`, `osp_kernel`, `changeset`, `occ_read_capture`, `occ_atomicity`, `occ_sqlite`).
- The #7790 P1 repro prints `[A(val=2)]` for all three spellings, both fresh (overlay) and against a persisted DB (engine), including a mixed engine+overlay run.
- Walker smoke: `visit [-->]`, typed+predicated visit narrowing, `[edge a -->]`, `[a ->:T:w > 8:->]`, identity `[a -->(b)]`, undirected `[b <-->]` all correct.
- E5094 fires with a precise reason (`operator 'Is' is not admitted`) and blocks codegen.
- Grep-clean: no `getclosurevars`, no query-path `dis`, no `topology_index_data`, no `TOPO_VERSION`, no `post_filter`, no `ObjectSpatialPath` outside the ES/native backends' own AST lowering.

## Notes for review

- Pre-existing, unrelated: walker `spawn` with a backtick-`root` trigger fails in scratch projects (`issubclass() arg 2` error) on clean main too — reproduced on an upstream/main worktree; explicit `Root` triggers work.
- The deleted walk helpers applied `check_read_access` mid-traversal while the old index fast path did not (inconsistent on main); the planner uniformly matches the fast-path behavior. Access-in-id-space (owner/permission columns in the projection) is called out in #7855's breaking-changes section as the follow-up.
- `[?...]` remains lambda-based for plain **list** filtering via `filter_on` — that's outside the query language by design.

Closes #7855. Supersedes #7807. Fixes #7790.


marsninja:
## A/B benchmarks: this branch vs upstream/main, real scale stack in k8s

Harness: `_planning/qbench` (branch) / worktree of upstream/main — identical scripts, `ScaleTieredMemory` with **Redis L2 + Mongo L3 running in a kind cluster**, full db+cache wipe between runs, per-request fresh `JScaleExecutionContext` (request-scoped L1, warm L2/L3 — production shape). `redis_fetches` = anchors materialized per query (L2 fetch-counter delta).

### Correctness, live in a scale deployment

| query (600 Sessions + org tree under root) | main | branch |
|---|---|---|
| `[root-->][?:Session][?sid==X]` (chained) | 1 ✓ | 1 ✓ |
| `[root-->[?:Session, sid==X]]` (nested) | 1 ✓ | 1 ✓ |
| `[root-->[?:Session]][?sid==X]` (mixed) | **640 — every root child, filter silently dropped (#7790 P1)** | **1 ✓** |

### Read path (cold p50 / p95 ms, per fresh request; anchors materialized per query)

| query | main cold | branch cold | main anchors | branch anchors |
|---|---|---|---|---|
| session lookup, chained | 23.1 / 40.9 | 23.1 / 24.9 | **600** | **1** |
| session lookup, nested | 380.3 / 395.5 | 25.8 / 27.3 | **1,280** | **1** |
| live-session scan (400 hits) | 391.7 / 412.8 | 32.2 / 35.7 | 1,280 | 400 |
| org 3-hop pred chain (8 hits) | 384.6 / 401.6 | 56.2 / 63.6 | **1,332** | **8** |
| page slice [0:25] | 9.7 / 10.1 | 16.1 / 17.9 | 25 | 25 |

Any predicate on main forces the full walk: ~1,280–1,332 anchors deserialized per request to return 1–8 results (the jacBuilder#895 OOM shape). The branch holds `materialized == |result|` on every query — the cost invariant, live on Mongo. Main's chained lookup is latency-competitive only because Redis pipelines the 600-anchor flood in one RTT; it still moves and deserializes 600 objects per request.

### Write path (attach N items to one root, batches of 250)

| | main | branch |
|---|---|---|
| batch @ N=500 | 682 ms | 288 ms |
| batch @ N=2000 | **2,490 ms (growing every batch)** | **288 ms (flat)** |
| total, 2000 nodes | 11.8 s | 2.2 s |
| seed 4,240-edge graph | 51.1 s | 4.8 s |

Main's per-batch cost grows linearly with graph size (the #7790 P2 quadratic, blob re-encode per edge); the branch is O(1)/edge and flat.

### Honest notes

- **Main wins warm same-context repeats** (8–12 ms vs 20–56 ms): after main's first query floods L1, repeats are list-filters over resident objects, while the branch re-resolves through the engine each time. This is exactly the gap the phase-4 process-scoped projection closes (#7855) — and the branch's warm number is what main's *cold* number pretends to be.
- This run was also the Mongo path's first live execution; it caught and fixed two real defects (per-stub heal probes → batched `existing_ids`, 567→24 ms on typed hops; benchmark-db hygiene). Zero quarantines clean-room.

PR #7868, opened 2026-08-03
[client] support custom <script> tags in the generated index.html head

### LINKED ISSUE #7865: feat(client): support custom <script> tags in the generated index.html head
## Problem

There is no way to add a `<script>` tag to the `<head>` of the generated `index.html` via `jac.toml`.

The head is fully synthesized in `jac/jaclang/runtimelib/client/targets/impl/web_target.impl.jac` (`_generate_index_html`), and the only configurable surface is `[client.app_meta_data]`, whose `HeaderBuilder` (`jac/jaclang/runtimelib/client/impl/serving.impl.jac`) emits only `<title>`, `<meta>` (incl. `og_*` / `twitter_*` / `meta_*`), and `<link>` (`link_*`) tags. All values are passed through `html.escape`, so a script tag cannot be expressed, and there is no custom index.html template override.

Common use cases that need this: analytics snippets, external SDKs (Stripe, maps), font/script preloads.

## Current workarounds

- Inject at runtime from client code with `document.createElement("script")` — loads only after the bundle boots.
- Post-process `dist/index.html` after `jac build` — fragile.

## Proposal

Add a scripts key under `[client.app_meta_data]`, e.g.:

```toml
[client.app_meta_data]
scripts = [
  { src = "https://example.com/thing.js", defer = true },
]
```

Emitted by `HeaderBuilder` alongside the link tags, with attribute escaping. Docs update in `reference/plugins/jac-client.md` (App Metadata section).

MalithaPrabhashana:
Closes #7865

## Problem

There is no way to put a `<script>` tag in the `<head>` of the generated `index.html` from `jac.toml`. The head is fully synthesized by `HeaderBuilder`, which only emits `<title>`, `<meta>`, and `<link>`, and every value goes through `html.escape`, so a script tag cannot be expressed. That leaves analytics snippets and third-party SDKs (Stripe, maps) to either be injected at runtime after the bundle has already booted, or post-processed into `dist/index.html` after `jac build`.

## Fix

Add a `scripts` key under `[client.app_meta_data]`:

```toml
[client.app_meta_data]
scripts = [
  { src = "https://js.stripe.com/v3", defer = true },
]
```

The array-of-tables form works too, and a bare string is shorthand for `src`:

```toml
[[client.app_meta_data.scripts]]
src = "https://cdn.example.com/maps.js"
async = true
crossorigin = "anonymous"

[[client.app_meta_data.scripts]]
content = """
window.dataLayer = window.dataLayer || [];
function gtag(){dataLayer.push(arguments);}
"""
```

Each entry takes either `src` or an inline `content` body. Every other key becomes an attribute on the tag, validated by shape rather than by an attribute whitelist, so `async`, `defer`, `type`, `crossorigin`, `integrity`, `nonce`, `data-*`, and any attribute the platform adds later all work without touching jaclang:

- Underscores in keys fold to hyphens (`data_site_id` renders `data-site-id`).
- A boolean `true` renders a bare flag (`defer = true` renders `defer`); a boolean `false` omits the attribute. Only a boolean omits, so `tabindex = 0` and `data_empty = ""` render as real values.
- Any other value renders as `name="value"`, HTML-escaped.
- Keys that do not fold to a well-formed lowercase attribute name are dropped, so a stray space or quote in a key cannot break out of the tag and forge extra attributes. A literal `</script>` inside an inline body is escaped so it cannot close the tag early.

Event handlers are ordinary attributes here, which the CDN fallback pattern depends on:

```toml
[[client.app_meta_data.scripts]]
src = "https://cdn.example.com/lib.js"
onerror = "loadLocalFallback()"
```

`jac.toml` can already execute arbitrary JavaScript through an inline `content` entry, so there is no trust boundary to defend by refusing `onerror` on a sibling entry, and refusing it would block a standard use of `<script>`.

The rules encode frozen invariants (the HTML attribute-name grammar and raw-text element parsing) and derive flag-vs-valued rendering from the TOML value type, rather than enumerating the HTML vocabulary.

### Files

- `runtimelib/client/serving.jac` + `impl/serving.impl.jac`: new `build_scripts` / `_script_tag` on `HeaderBuilder`, called from `build_head()`. This is the shared chokepoint, so both `jac build` (`_generate_index_html`) and `jac start` (`JacClient.render_page`) pick it up with no change at either call site. PWA, static, and mobile targets post-process the same file, so they inherit it.
- `runtimelib/client/impl/vite_bundler.impl.jac`: `_ensure_dev_index_html` writes its own head and never goes through `HeaderBuilder`, so the scripts are injected there too. Without this they would silently vanish under `jac start --dev`, which reads as a bug when an app depends on an SDK locally.
- `cli/docs/reference/plugins/jac-client.md`: new Head Scripts section under App Metadata.

## Verification

- 6 dense script tests in `test_header_builder.jac` (10 total in the file) (written red-green, then consolidated), each pinning one concern: an exact-string render of a fully loaded entry (escaping, bare flags, folding, `0` and `""` surviving, `false` omitting, unlisted-but-well-formed attributes like `blocking`, insertion order, scripts-last placement), the three input shapes, malformed-key rejection, event handler support, inline-body breakout protection in any case, and silent degradation of malformed config; plus 1 dev-index test in `test_web_target_dev.jac`.
- Validated end to end on a scaffolded `jac create --kind web-static` app declaring scripts across the whole attribute surface, including deliberately hostile keys. The built `index.html` was executed in a real HTML parser and every attribute the parser actually constructed was enumerated, rather than asserting on generated strings: the inline script runs before the bundle, `defer`/`async` scripts execute, escaped values round-trip, a configured `onerror` fires on a missing `src`, `tabindex="0"` and `data-empty=""` survive, malformed keys build no attribute, and an inline `</script>` payload creates no element. The same page was verified served by `jac start` and by `jac start --dev` (vite prepends its HMR scripts and leaves the injected tags intact). That enumeration approach found two bugs the unit tests missed; see the comments below.


MalithaPrabhashana:
> **Superseded in part.** The `on*` rejection and the falsy-value omission described below were removed in 96d5d2109; see the correction comment further down. The attribute-name grammar findings and the `data_` bug remain accurate.

## Verification evidence

Beyond the unit tests, this was validated end to end on a project scaffolded from scratch with `jac create --kind web-static`, pointed at this branch via `[dev] jaclang_source`.

### Adversarial testing found a real bug

The interesting result: a hostile config pushed through a **real HTML parser** (jsdom), enumerating every attribute the parser actually built rather than asserting on strings the implementation produced. That caught a leak the unit tests missed.

A key of `"data_" = "x"` folded to `data-`, which passed the original name grammar `[a-z][a-z0-9-]*` (a `d`, then `ata-` are all permitted characters) and emitted a junk attribute:

```html
<script src="/guard.js" id="guard" data-="x"></script>
```

Fixed in 7153c9717 by tightening to `[a-z][a-z0-9]*(-[a-z0-9]+)*`, so every hyphen must be followed by at least one alphanumeric. Rejects `data-` and `data--x`; keeps `data-site-id`, `blocking`, `tabindex`. A regression test covers the bare `data_`, doubled `data__x`, and leading-digit `9data_lead` cases.

Worth noting the unit tests could not have found this on their own: they assert on strings the implementation generates, so they only cover cases the author already imagined. Enumerating the parser's actual output is what surfaced it.

### Hostile keys, verified dropped in a browser parser

| Key in `jac.toml` | Result | Rule |
|---|---|---|
| `onload`, `onerror`, `onmouseover` | dropped | `on*` class |
| `"data_x onmouseover"` | dropped | space fails grammar |
| `'data_y" onclick="...'` | dropped | quote fails grammar |
| `"data_z><span data_q"` | dropped | fails grammar |
| `data_UPPER` | dropped | uppercase fails grammar |
| `"data_"` | dropped | trailing hyphen (the fix above) |
| `"9data_lead"` | dropped | must start with a letter |
| `nomodule = false` | omitted | falsy |

The parser sweep additionally asserts that no `on*` attribute exists anywhere in the document, and that no ``/`<span>`/`<iframe>` element materialized in `<head>` from the inline `</script>` payload. Both clean.

`/guard.js` is deliberately a 404 in the demo, which is the exact trigger a surviving `onerror` would need. No handler fired.

### Well-formed attributes pass through with no whitelist

`blocking="render"`, `tabindex="3"`, `data-site-id="abc &amp; co"` all render correctly. None of these were in the original whitelist, which is the point of the grammar based approach: new platform attributes work without a jaclang release.

### Coverage

| | Verified |
|---|---|
| `jac build` output | yes, parsed and executed |
| `jac start` | yes, tags present in served HTML |
| `jac start --dev` (HMR) | yes, vite prepends its own scripts and leaves the injected tags intact |
| `scripts = [{...}]` inline array | yes, built |
| `[[client.app_meta_data.scripts]]` tables | yes, built |
| `scripts = ["https://..."]` bare string | yes, built |

Dev mode note: vite injects `/@react-refresh` and `/@vite/client` into the head, so the script count there is higher than configured. Expected, not a defect.

### Ordering guarantee

Emission follows `jac.toml` order exactly, and scripts land after the title/meta/link tags so crawler-relevant markup stays at the top of the head.

An external CDN library declared in the head was confirmed callable during head parsing (`typeof confetti === "function"` at 270ms), before the app bundle executed. That is the case the issue describes and that a runtime `document.createElement` injection cannot serve, since the app has not booted yet.

### Known gap, not addressed here

Dropped keys are silent. A typo such as `defre = true` is well formed, so it renders as an inert unknown attribute rather than being dropped or reported; a malformed key like `data_UPPER` disappears with no warning. Adding a build time warning is possible but needs care, since `HeaderBuilder` also runs per request inside `render_page`, so a naive warning would spam the server log. Happy to follow up in a separate PR if reviewers want it.

Separately, and unrelated to this change: web-static serves no loose `.js` files (`publicDir: false` plus the static server's routing), so `src = "/local.js"` will 404 in that app kind. CDN URLs, `data:` URIs, and inline `content` all work. Worth its own issue.

MalithaPrabhashana:
## Correction to the evidence comment above

Two restrictions in my earlier comment were wrong and have been removed in 96d5d2109. Flagging explicitly because the comment above presents one of them as a security property, which it is not.

### 1. Event handler attributes are now supported

The earlier version dropped any `on*` key and I documented that as a safety guarantee. That justification does not hold: `jac.toml` is developer authored and can already execute arbitrary JavaScript through an inline `content` entry, so refusing `onerror` on a sibling entry protects nobody. It is a lock on the window next to an open door.

Worse, it blocked the single most common reason to put an attribute on an external script tag:

```toml
[[client.app_meta_data.scripts]]
src = "https://cdn.example.com/lib.js"
onerror = "loadLocalFallback()"
```

`onload` and `onerror` are spec defined event handler content attributes on `<script>`, and the CDN fallback pattern depends on them. They now render like any other attribute, with the value HTML escaped.

Verified end to end: with a deliberately missing `src`, the configured `onerror` fired in a real parser and set its attribute, confirming the fallback path actually works rather than merely rendering.

### 2. Falsy values were silently dropped

The omission rule was written as "skip falsy", intending "a `false` boolean omits the flag". It also swallowed every other falsy value:

| Config | Before | Now |
|---|---|---|
| `tabindex = 0` | dropped | `tabindex="0"` |
| `data_count = 0` | dropped | `data-count="0"` |
| `data_empty = ""` | dropped | `data-empty=""` |

`tabindex="0"` is meaningful HTML (focusable in natural tab order), and an empty attribute value is valid. Only a boolean `false` omits an attribute now.

### What still holds from the original comment

The attribute name grammar is unchanged and remains the real defense. A key that cannot fold to a well formed lowercase attribute name is still dropped, so it cannot break out of the tag and forge extra attributes:

| Key | Result |
|---|---|
| `"data_x onmouseover"` | dropped, space |
| `'data_y" onclick="..."'` | dropped, quote |
| `"data_z>"` | dropped |
| `data_UPPER` | dropped, uppercase |
| `"data_"` | dropped, trailing hyphen |
| `"9data_lead"` | dropped, must start with a letter |

Re-verified after the change by enumerating every attribute a real HTML parser built: the tag carries only `src`, `id`, `onerror`, `tabindex="0"`, `data-empty=""`, and no malformed name or injected element appears anywhere. The `</script>` breakout protection is also unchanged and still verified.

Both build output and `jac start --dev` produce a byte identical tag. One dev only note for anyone testing this: vite runs with `appType: 'spa'`, so a missing `src` returns the SPA fallback with HTTP 200 rather than a 404, and an `onerror` handler therefore will not fire under `--dev`. That is vite dev server behavior, not a difference in what jac emits.

### Why this took an extra round

The original restriction was reflexive pattern matching: "config value reaching an HTML attribute name" resembles XSS, and I applied that instinct without checking whether the trust boundary it assumes exists here. It then accumulated a test, a docs paragraph, and a PR comment asserting it, none of which examined the premise. Tests that encode an assumption do not validate it, they entrench it. Worth stating plainly since the earlier comment reads with more confidence than it earned.

PR #7875, opened 2026-08-04
fix(scale): stop deploys inheriting the deploying process's microservices config

### LINKED ISSUE #7874: Jac Scale: 0.34.9 single-service deploys inherit the host process's shared_volumes, unresolved placeholder becomes a PVC name and fails apply with 422
## Summary

Since 0.34.9 unified the Kubernetes deploy targets, a single-service app deployed from a host process that itself has a jac-scale config inherits that host's `[scale.microservices].shared_volumes`. When the host's volume name is an env placeholder the host process cannot resolve, the literal `${...}` becomes a PVC name and the deploy dies at apply with a raw 422 from the API server.

Root cause is two long-standing latent defects that 0.34.9 made reachable for the first time: `_ms_cfg()` silently falls back to the process-global `get_scale_config()` when handed an empty microservices config, and `k8s_safe_name()` does not enforce RFC 1123, so the unresolved placeholder is never rejected before apply.

This is a regression in reachability, not in either defect. Both behave identically in 0.34.8.

## Environment

- jac 0.34.9 (pod runtime and deploy driver), pinned via `[project] jac-version = "==0.34.9"`
- Deploy target: EKS 1.35, `production-cluster`, us-east-2
- Host process: jacBuilder builder-sv, which deploys user apps in-process
- Last known good: jac 0.34.8, same cluster, same app

Observed in production on 2026-08-04T04:26:36Z deploying a single-service user app (`standup`, namespace `jb-299afd1f-standup`). The user app's `jac.toml` declares `[scale.sso]` and `[scale.monitoring]` and has no `[scale.microservices]` section at all. Verbatim failure returned to the deploy panel:

```
manifest apply failed: (422)
Reason: Unprocessable Entity
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure",
"message":"PersistentVolumeClaim \"${shared-data-pvc}\" is invalid: metadata.name: Invalid
value: \"${shared-data-pvc}\": a lowercase RFC 1123 subdomain must consist of lower case
alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character
(e.g. 'example.com', regex used for validation is
'[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')","reason":"Invalid",
"details":{"name":"${shared-data-pvc}","kind":"PersistentVolumeClaim","causes":[{"reason":
"FieldValueInvalid","field":"metadata.name"}]},"code":422}
```

The PVC name is not from the user app. It is the host's value after `k8s_safe_name`. Executed:

```python
GATEWAY_NAME = "__gateway__"
def k8s_safe_name(svc_name):     # manifest_builder.jac:16-21, verbatim
    if svc_name == GATEWAY_NAME: return "gateway"
    return svc_name.lower().replace("_", "-")

k8s_safe_name("${SHARED_DATA_PVC}")
```

```
  input : ${SHARED_DATA_PVC}
  output: ${shared-data-pvc}
  matches the 422 name '${shared-data-pvc}': True
```

`${SHARED_DATA_PVC}` is the host's own `shared_volumes` entry, still literal in the running pod because the variable is only defined at CI deploy time and is absent from the pod environment. Executed inside the host pod:

```
$ kubectl exec -n jachammer builder-sv-deployment-... -c builder-sv -- \
    sh -c 'echo "SHARED_DATA_PVC=[${SHARED_DATA_PVC:-<UNSET>}]"'
SHARED_DATA_PVC=[<UNSET>]

$ ... -- sh -c 'grep -n SHARED_DATA_PVC /app/jac.toml'
302:shared_volumes = [{ name = "${SHARED_DATA_PVC}", mount_path = "/app/.jachome/.jac-ide",
    services = ["builder_sv", "jac_coder_sv"], size = "10Gi",
    access_mode = "ReadWriteMany", storage_class = "efs-sc" }]
```

Minimal reproduction from a clean checkout, no jacBuilder required:

1. Build any app whose `jac.toml` has `[scale.microservices]` with a `shared_volumes` entry whose `name` is `"${SOME_VAR}"`, and do not export `SOME_VAR`.
2. From that app's own process (so `get_scale_config()` is populated), deploy a second, different app whose `jac.toml` has no `[scale.microservices]` section, passing an empty microservices config.
3. The second app's manifest bundle contains a PVC named `${some-var}` and apply fails with the 422 above.

## Expected

Either of these would have prevented it:

- An app that declares no `[scale.microservices]` should not inherit another app's `shared_volumes`. An empty config should mean "no microservices config", not "fall back to whatever is loaded in this process".
- `k8s_safe_name()` should produce an RFC 1123 compliant name or fail loudly. A name containing `$`, `{` or `}` should never reach the API server.

`plan.jac` already enumerates `pvcs`, so an invalid name is catchable before a single manifest is applied.

## Actual

`_ms_cfg()` at `jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac:87-100`:

```jac
if self.microservices_config {
    return normalize_microservices_config(self.microservices_config);
}
try {
    return get_scale_config().get_microservices_config();
} except Exception {
    return {};
}
```

An empty or absent `[scale.microservices]` is falsy, so this returns the process-global config, which inside a host that is itself a jac-scale app is that host's config. `_build_shared_pvc_manifests()` (same file, around line 426) then builds a PVC per `shared_volumes` entry using `pvc_name = k8s_safe_name(name)`, and `k8s_safe_name` only lowercases and swaps underscores. Executed against a range of inputs:

| input | k8s_safe_name output | valid RFC 1123 |
|---|---|---|
| `${SHARED_DATA_PVC}` | `${shared-data-pvc}` | no, 422 |
| `-lead` | `-lead` | no, 422 |
| `vol!name` | `vol!name` | no, 422 |
| 260 chars | unchanged | no, 422 |
| `shared_data_pvc` | `shared-data-pvc` | yes |

Separately, `normalize_microservices_config` (`config/impl/config_loader.impl.jac:544-560`) falls back to the `JAC_SV_ROUTES` env var whenever routes are empty, which is the same class of leak through a different channel. In the host pod that variable is populated with the host's own routes, so simply adding an empty `[scale.microservices]` to the guest app swaps the PVC leak for a routes leak.

## Why this is a 0.34.9 regression

Neither defect changed. `_ms_cfg` and `k8s_safe_name` are byte-identical in 0.34.8. What changed is which deploys reach them.

In 0.34.8 the shared_volumes code lived in the microservice-only builder:

```
$ gh api ".../kubernetes/manifest_builder.jac?ref=v0.34.8"
gh: Not Found (HTTP 404)

$ gh api ".../kubernetes?ref=v0.34.8" --jq '.[].name'
cluster_provider.jac
ingress.jac
kubernetes_config.jac
kubernetes_target.jac
microservice          <- manifest_builder.jac lived in here
monitoring.jac
runtime_bootstrap.jac
utils
```

`shared_volumes` occurrences at v0.34.8: 6 in `microservice/manifest_builder.jac`, 0 in `kubernetes_target.jac`. Single-service apps went through `kubernetes_target.jac` and never touched shared_volumes or the global fallback. 0.34.9 collapsed both paths into one `kubernetes/manifest_builder.jac`, so single-service apps now run the microservice-only code for the first time.

## Impact

Every single-service app deployed in-process by a jac-scale host that declares `shared_volumes`. On jacBuilder that is all user apps, since none of them declare `[scale.microservices]`.

The failure mode depends on whether the host's variable resolves:

- Unresolved, as here: every user-app deploy fails at apply with an opaque 422. Already-running deploys are unaffected, so this presents as "redeploy is broken" rather than an outage.
- Resolved: much worse and near-silent. The apply succeeds, so the host's ENTIRE microservices fleet is created in the guest namespace (verified: builder_sv + jac_coder_sv at 5 replicas each with HPAs to 45, a gateway, Services, PDBs and the PVC). The guest's own app keeps running and serving, so nothing looks broken from the outside while ten crashlooping pods sit in the user's namespace consuming schedulable capacity.

## Workaround (verified)

Setting the host's variable in the host pod environment so the placeholder resolves:

```bash
kubectl patch deploy -n <ns> <host-deploy> --type=strategic \
  -p '{"spec":{"template":{"spec":{"containers":[{"name":"<c>","env":[{"name":"SHARED_DATA_PVC","value":"jachammer-data-pvc"}]}]}}}}'
```

DO NOT USE THIS. Tried on 2026-08-04 and it is far worse than the 422. Setting the
variable makes the PVC name valid, which lets the apply SUCCEED, and the rest of the
host's microservices config then deploys into the guest namespace: builder_sv and
jac_coder_sv Deployments (5 replicas each, HPAs min 5 max 45), a gateway Deployment,
their Services and PDBs, plus the PVC. The guest pods crashloop with
`couldn't read builder_sv.jac: No such file or directory`, because they are running
the host's service manifests against the guest's bundle.

The 422 is load-bearing. It fails the apply before any Deployment is created, so the
invalid PVC name is the only thing standing between this bug and the host's entire
fleet landing in every guest namespace. Reverted, and the leaked objects were deleted
by hand.

There is no safe user-side workaround either. Adding an empty [scale.microservices]
to the guest stops the PVC leak and triggers the JAC_SV_ROUTES leak instead, injecting
the host's routes into the guest deploy.

Adding an empty `[scale.microservices]` to the guest app does NOT work. It stops the PVC leak and triggers the `JAC_SV_ROUTES` leak instead, injecting the host's routes into the guest deploy.

## Suggested fix

Make the fallback explicit rather than implicit. `_ms_cfg()` should distinguish "caller passed nothing, use the global" from "caller passed an empty config, honor it", for example by defaulting `microservices_config` to `None` and only falling back when it is `None`. The same applies to the `JAC_SV_ROUTES` fallback in `normalize_microservices_config`, which should not fire when the caller supplied a config object.

Independently, `k8s_safe_name()` should enforce RFC 1123: strip characters outside `[a-z0-9.-]`, trim leading and trailing non-alphanumerics, cap at 253, and raise on an empty result. Its current name promises a guarantee it does not provide, and PVC names from `injector.extra_manifests()` (`manifest_builder.jac:200-202`) are not passed through it at all.

Regression tests that would have caught this:

- Build a manifest bundle for an app with no `[scale.microservices]` while a different config is loaded process-globally, and assert the bundle contains no PVCs from the global config.
- Assert `k8s_safe_name` output matches `^[a-z0-9]([-a-z0-9]*[a-z0-9])?$` and is 253 chars or fewer, over a table including `${VAR}`, a leading hyphen, and an over-long name.

MusabMahmoodh:
Fixes #7874.

The report is accurate and I reproduced both defects verbatim on current main. The `_ms_cfg` fallback and `k8s_safe_name` are byte-identical to what the issue quotes, and the executed `k8s_safe_name("${SHARED_DATA_PVC}") -> "${shared-data-pvc}"` matches the 422 name exactly.

## Root cause

The deeper cause is narrower than "two latent defects": **a per-deploy operation resolves an ambiguous input by reaching for process-global ambient state.**

`ManifestBuilder`/`KubernetesTarget` cannot distinguish "caller supplied nothing" from "caller supplied an empty config", because both are falsy. It resolves that ambiguity by reading `get_scale_config()`. For a CLI deploy the process-global config *is* the app being deployed, so the fallback is invisible and harmless. For an in-process deploy the process-global config belongs to the **host**, and the guest silently inherits it.

That makes this specifically a violation of the SDK's documented contract. #7809 introduced `ScaleClient` as a deploy that "takes the whole configuration in memory ... never reads `jac.toml`" - and `_make_target` passes `spec.to_microservices_config()`, which is `{}` for an app declaring no microservices. Falsy. So the one path built for embedding is exactly the path that fell through to the host's `jac.toml`. The unification in #7828 then routed single-service apps through this builder for the first time, which is why it surfaced now.

## The fix

**Make the input a real sentinel.** `microservices_config: dict | None = None`. `None` means "not supplied, fall back to the loaded config" (the CLI path, unchanged). Anything supplied is honored exactly, empty or not. Applied to both `ManifestBuilder` and `KubernetesTarget`, which carried independent copies of the same falsy check.

**Gate the environment fallback the same way.** `normalize_microservices_config` filled empty routes from `JAC_SV_ROUTES`, which is how the *gateway pod* learns its routes - legitimate there, wrong for a deploy driver reading the host's environment. It now takes `env_routes_fallback: bool = True`, left on for the runtime callers (`get_microservices_config`, used throughout the gateway, rpc, admin and context modules) and off when the builder normalizes a caller-supplied config. Without this, adding an empty `[scale.microservices]` to the guest app just swapped the PVC leak for a routes leak, as the issue notes.

**Make `k8s_safe_name` deliver the guarantee its name makes.** It now strips characters outside `[a-z0-9.-]`, trims to a leading and trailing alphanumeric, caps at 253, and raises when nothing usable remains.

One deliberate deviation from the suggested fix: an unresolved `${...}` placeholder **raises** rather than being sanitized. Cleaning `${SHARED_DATA_PVC}` into `shared-data-pvc` would turn a loud 422 into a silently provisioned volume named after a variable the operator never resolved - the issue's own "resolved" branch, which it rates as *worse and silent*. The error names the variable and says what to do:

```
'${SHARED_DATA_PVC}' still contains an unresolved ${...} placeholder, so the
environment variable it names was not set where this deploy ran. Export it, or
replace it with a literal name.
```

Also hoisted the `get_scale_config` / `normalize_microservices_config` imports in `manifest_builder` from function-local to module level. The function-local import is why this seam had no test: it is not patchable, and the first version of the regression test failed with `does not have the attribute 'get_scale_config'`.

## One claim I could not confirm

The issue notes that PVC names from `injector.extra_manifests()` are "not passed through it at all". On current main the injector's `app_name` *is* sanitized at every call site (`manifest_builder.jac:133`, `:217`, `:1015`, `target.jac:812`), so `bundle_pvc_name` receives an already-safe name. Sanitizing inside `bundle_pvc_name` would need `injector` to import from `manifest_builder`, which imports `pvc_injector` - a cycle. Left as is.

## Tests

`tests/deploy/test_ms_config_isolation.jac`, 8 tests. Six fail on the unfixed tree; all pass after:

- a guest with `{}` inherits neither the host's `shared_volumes` nor its routes
- the guest's bundle contains no PVC built from the host's config (the exact production failure)
- an **omitted** config still falls back to the loaded one, so the CLI path is pinned against over-correction
- a supplied config takes no routes from `JAC_SV_ROUTES`
- `k8s_safe_name` output matches `^[a-z0-9]([-a-z0-9]*[a-z0-9])?$` and is <= 253 over a table covering a leading hyphen, a trailing hyphen, `vol!name`, 300 chars and doubled underscores
- an unresolved placeholder raises, naming the variable
- a name with nothing usable raises
- the gateway keeps its reserved short name

Targeted runs: `tests/deploy/` + `tests/test_pod_env.jac` -> 110 passed, and `tests/misc/test_config_loader.jac` + `tests/microservices/test_plan.jac` + the new file -> 38 passed. The single `test_keda_http_activation` failure is present on unmodified main. `jac check` on the touched files is unchanged from baseline (`target.jac` 64, `config_loader.impl.jac` 4); the new test file's two errors are the untyped `unittest.mock.patch.dict`, matching the existing `test_dry_run_purity.jac`.

## Deployment note

For anyone on the workaround in the issue: the env var patched into the host pod can be removed after upgrading, and any spurious per-namespace PVC it provisioned can be deleted. With `efs-sc` on `Retain` those leave Released PVs behind, so they need deleting explicitly.


MusabMahmoodh:
Reworked the regression tests to drop the mocks.

The first version MagicMocked `get_scale_config`, which is one of the system's own collaborators rather than an external boundary, so it asserted against our *assumption* of the loader instead of the loader itself. They now seed a real `jac.toml` and load it through the real config loader, following the existing pattern in `tests/microservices/test_plan.jac`, and drive `JAC_SV_ROUTES` through the real environment with save/restore. No test doubles anywhere in the file.

That also made the tests meaningfully stronger: they now exercise real TOML parsing of the inline `shared_volumes = [{ name = "${SHARED_DATA_PVC}", ... }]` table, the real `normalize_microservices_config`, and the real process-global instance that `reset_scale_config()` clears. One added test pins the setup itself (`the host's config really loads`) so the leak assertions cannot pass vacuously if the fixture ever stops loading.

Re-verified against the unfixed builder by reverting only the two source files: **6 of 9 fail**, with the guest inheriting the host's volume verbatim -

```
AssertionError: guest inherited the host's shared_volumes:
[{'name': '${SHARED_DATA_PVC}', 'mount_path': '/app/.jachome',
  'services': ['builder_sv'], 'size': '10Gi',
  'access_mode': 'ReadWriteMany', 'storage_class': 'efs-sc'}]
```

Targeted re-run: `tests/deploy/` + `tests/misc/test_config_loader.jac` + `tests/microservices/test_plan.jac` -> 131 passed, with only the pre-existing `test_keda_http_activation` failure that also fails on unmodified main.

Issue #6732, opened 2026-06-16
Bare expression in a statement slot silently renders nothing


MalithaPrabhashana:
## Description

Inside a JSX statement slot, a JSX element is auto-pushed into the slot's child
accumulator, but a bare expression statement (e.g. a string) is not - it is evaluated
and discarded. So conditional content written as a bare expression renders **nothing**.
The only diagnostic is a misleading `W0060` that reports the string as a *docstring*, so
nothing tells the author the content was dropped.

## Reproduce

Same slot shape, three ways to write the body - only A is broken:

```jac
cl def:pub app() -> JsxElement {
    has show: bool = True;
    return <div>
        <div data-test="bare">              <-- A
            {if show {
                "BARE_STRING_A";
            }}
        </div>

        <div data-test="element">           <-- B
            {if show {
                <span>ELEMENT_STRING_B</span>
            }}
        </div>

        <div data-test="ternary">           <-- C
            {"TERNARY_STRING_C" if show else ""}
        </div>
    </div>;
}
```

<img width="471" height="566" alt="Image" src="https://github.com/user-attachments/assets/902b493f-1160-47f9-9bee-9dad2788ba56" />

## Expected

A renders `BARE_STRING_A` (like B and C), or the compiler clearly warns that a bare
expression in a statement slot produces no child.

## Actual

**Compile (`jac jac2js`)** - A's string is never pushed; B is, C renders inline:

```js
// A:  __jac_view_kids stays []  ->  renders nothing
if (show) { "BARE_STRING_A"; }
// B:
if (show) { __jac_view_kids_1.push(__jacJsx("span", {}, ["ELEMENT_STRING_B"])); }
// C:
(show ? "TERNARY_STRING_C" : "")
```

**Only diagnostic** is on A, and it's misleading - the bare string is parsed as a docstring:

```
W0060: Docstrings in Jac go before the declaration, not inside the body.
```

**Browser (rendered DOM, Chromium via Playwright):**

| box | body form | rendered text |
|---|---|---|
| A | `"BARE_STRING_A";` (bare expr) | `""` &nbsp;← **blank** |
| B | `<span>ELEMENT_STRING_B</span>` | `ELEMENT_STRING_B` |
| C | `{"TERNARY_STRING_C" if show else ""}` | `TERNARY_STRING_C` |

A renders empty with no runtime error. (Screenshot: `issue2_proof.png`.)

## Why it matters

[...]

PR #7876, opened 2026-08-04
fix(client): keep JSX statement-slot child positions stable (#7859)

### LINKED ISSUE #7859: JSX statement slots (`if`/`for` as a JSX child) emit an extra unkeyed array nesting level, causing a spurious "unique key prop" warning
## Summary

When an `if {..} else {..}` block or a `for {..}` loop is used as a JSX **statement slot** (i.e. as one of possibly several children inside a parent JSX element, rather than as a top-level return value), the generated React code wraps the slot's output in an extra, unnecessary array-inside-array. React interprets this as an unkeyed list and logs:

```
Warning: Each child in a list should have a unique "key" prop.
```

This fires even when there is nothing list-like about the actual content — including on a single, non-repeating conditional branch with zero iteration involved.

**Scope: `jac start --dev` only.** Verified this does not appear under a production `jac start` (no `--dev`): the served production bundle is React's production build, which strips dev-only `console.error` diagnostics (including this key-validation check) out of the shipped code entirely — not just silenced, absent. Confirmed by building the same component both ways and diffing bundle content: the dev bundle contains the warning string and fires it; the prod bundle contains no trace of the warning string or any dev-mode markers (`development`, `__DEV__`, `NODE_ENV`), and the console is clean. So this is a development-experience issue (affects local dev, hot-reload workflows, and anything that inspects a dev-mode console, e.g. automated dev-mode testing), not something that reaches production users.

## Environment

- `jac` version: reproduced on `0.34.8` (confirmed identical codegen present through `main`, i.e. this is not fixed by any later 0.34.x/0.35.x release)
- Target: `react` client backend (default web target)
- OS: Linux x86_64, but this is a pure codegen issue — not platform-specific

## Reproduction

Minimal project, any fresh `jac create` scaffold works. Component:

```jac
node Todo {
    has text: str;
}

def:pub TodoList() -> JsxElement {
    has todos: list[Todo] = [];

    return <div className="p-4">
        {if len(todos) == 0 {
            <p className="text-sm">No todos yet.</p>
        } else {
            <ul>
                {for todo in todos {
                    <li key={todo.text}>
                        <span>{todo.text}</span>
                    </li>
                }}
            </ul>
        }}
    </div>;
}
```

Steps:
1. `jac start --dev main.jac` (with `main.jac` rendering `<TodoList />`)
2. Open the app in a browser with the console open
3. Observe the console — the warning fires immediately on first render, with `todos` still empty and no list ever rendered

## Actual behavior

Console (dev mode):

```
Warning: Each child in a list should have a unique "key" prop.
Check the render method of `TodoList`.
    at p
    at TodoList
```

Image

Note the warning names the `<p>` element from the **empty-state branch** — there is no iteration, no repeated element, and no missing key on anything a developer actually wrote. `key={todo.text}` on the `<li>` (the only element in the whole component with a real, repeating identity) is set correctly and is not what's being complained about.

Compiled output (`components/TodoList.js`) shows why:

```js
function TodoList() {
  const [todos, setTodos] = useState([]);
  return __jacJsx("div", {"className": "p-4"}, [(() => {
    let __jac_view_kids = [];
    if (todos.length === 0) {
      __jac_view_kids.push(__jacJsx("p", {"className": "text-sm"}, ["No todos yet."]));
    } else {
      __jac_view_kids.push(__jacJsx("ul", {}, [(() => {
        let __jac_view_kids_1 = [];
        for (const todo of todos) {
          __jac_view_kids_1.push(__jacJsx("li", {"key": todo.text}, [__jacJsx("span", {}, [todo.text])]));
        }
        return __jacJsx(null, {}, [__jac_view_kids_1]);   // <-- array nested inside array
      })()]));
    }
    return __jacJsx(null, {}, [__jac_view_kids]);          // <-- array nested inside array
  })()]);
}
```

Both statement slots (`if/else` and `for`) return `__jacJsx(null, {}, [accumulator])` : a Fragment whose children array has exactly one item, and that one item is *itself an array* (`__jac_view_kids`), rather than the accumulator's contents being passed/spread directly as the Fragment's children.

`__jacJsx`'s runtime spreads its `children` argument as individual positional arguments to `React.createElement` (the same calling convention as plain JSX siblings, which never requires keys). That works correctly everywhere *except* here: because the accumulator array is nested one level too deep, it gets pushed through as a single opaque "child" whose value is itself an array — the one shape where `React.createElement`'s resulting `props.children` is a raw array, which is exactly what triggers React's key-validation warning.

## Expected behavior

No console warning. The Fragment's children should be the accumulator array's *contents*, not the accumulator array itself as one nested item — e.g. `__jacJsx(null, {}, __jac_view_kids)` (pass the array directly, no extra wrapping) or `__jacJsx(null, {}, [...__jac_view_kids])` (spread) instead of `__jacJsx(null, {}, [__jac_view_kids])`.

## Root cause

`jaclang/compiler/passes/ecmascript/impl/esast_gen_pass.impl.jac`, function `_view_fragment_return` (builds the `return` for every JSX statement-slot IIFE):

```jac
impl EsastGenPass._view_fragment_return(
    acc_name: str, anchor: uni.UniNode
) -> es.ReturnStatement {
    frag = self._jsx_fragment(
        [self.sync_loc(es.Identifier(name=acc_name), jac_node=anchor)], anchor
    );
    return self.sync_loc(es.ReturnStatement(argument=frag), jac_node=anchor);
}
```

`_jsx_fragment(children, anchor)` builds `__jacJsx(null, {}, ArrayExpression(children))`. Here `children` is `[Identifier(acc_name)]` — a one-element array literal whose single element is the accumulator identifier — instead of the accumulator's contents being used as the Fragment's children directly.

This affects the `react` backend specifically because, for `IfChild`/`EachChild` view nodes, `ReactBackend.lower_view` (`compiler/passes/ecmascript/backends/impl/react.impl.jac`, around the `isinstance(child, (IfChild, EachChild))` branch) doesn't use its own key-aware element lowering for these nodes — it falls back to `child.js_fallback`, which is exactly this generically-compiled (non-JSX-aware) `if`/`for`-as-expression output. A properly key-aware `EachChild.key` field does exist in the view IR (extracted via `_split_key` in `jsx_processor.impl.jac`) but is unused by the React backend; it appears to be consumed only by the `solid` backend, which needs real per-item keys for its fine-grained reactivity model.

## Impact

- **Severity: low / cosmetic in practice, but broad in reach.** Every generated app that nests an `if`/`else` or `for` block as a JSX child (i.e. the officially recommended "statement slot" iteration idiom, as opposed to `.map()`) hits this on every render of that slot.
- The warning does not indicate an actual reconciliation problem: it always fires on a wrapper array containing exactly one item (either one resolved `if/else` branch, or one fully-formed list whose own elements — if any — already carry correct, unique keys of their own). There is no code shape under this bug where React could actually mismatch or lose state on add/remove/reorder.
- Still meaningfully bad in practice: React logs this class of dev diagnostic via `console.error()`, so it's indistinguishable at the console-severity level from a real crash. Anything that gates on "no console errors" (automated testing,CI checks, agentic build/validate loops) will treat a perfectly working app as broken.

## Where a fix would go

`jaclang/compiler/passes/ecmascript/impl/esast_gen_pass.impl.jac`, `_view_fragment_return` (and/or `_jsx_fragment`'s caller there) — pass the accumulator array as the Fragment's children directly (or spread it) instead of wrapping it in an outer one-element array. This is backend-specific to `react` (`compiler/passes/ecmascript/backends/impl/react.impl.jac`'s `lower_view`), so a fix should be verified against the `solid` backend's own `IfChild`/`EachChild` handling to confirm it isn't relying on the same fallback path in a way that would regress.

SandeepaHWP:
Fixes #7859.

## What was wrong

A JSX statement slot collected its children with `push`, so a child's position depended on how many earlier statements happened to emit. A false branch contributed nothing and every later sibling slid down an index. React matches keyless children by position, so it paired those siblings with the wrong fibers and remounted them, discarding their state. The slot fragment also nested the accumulator inside a second array, handing React a single child whose value is an array, which is the one shape React key-validates, producing a spurious "unique key prop" warning against content that never repeats.

The issue as filed treats the array nesting as the defect and the warning as cosmetic. Both are understatements, and @MalithaPrabhashana's analysis in the thread is what established that. Un-nesting the array alone silences the warning while leaving the remount, so the warning was closer to a true positive than the report credited.

## Root cause

Slot bodies had a single lowering and it was the dynamic-list one. `if` and `for` compiled to structurally identical code even though their arity differs completely: a conditional emits at most one child at a position known at compile time, while a loop emits a count known only at run time. The front end already draws this line, since W2021 fires for keyless `for` and `while` slots and `IfStmt` is deliberately excluded, but codegen did not.

## What changed

The count is known at compile time wherever it matters, so the accumulator no longer grows. Each statement claims an index as it is lowered, the accumulator is emitted at its final length with `false` in every slot, and statements assign rather than append.

### Before

```js
let __jac_view_kids = [];
if (show) {
  __jac_view_kids.push(__jacJsx("p", {}, ["header"]));
}
for (const t of tags) {
  __jac_view_kids.push(__jacJsx("li", {"key": t}, [t]));
}
return __jacJsx(null, {}, [__jac_view_kids]);
```

### After

```js
let __jac_view_kids = [false, false];
if (show) {
  __jac_view_kids[0] = __jacJsx("p", {}, ["header"]);
}
__jac_view_kids[1] = (() => {
  let __jac_view_kids_1 = [];
  for (const t of tags) {
    __jac_view_kids_1.push(__jacJsx("li", {"key": t}, [t]));
  }
  return __jac_view_kids_1;
})();
return __jacJsx(null, {}, __jac_view_kids);
```

`false` renders as nothing while holding its position, and unlike `null` it survives the runtime's empty-child filter in `client_runtime.impl.jac`, so no runtime change is needed and `props.children` keeps the shape every existing consumer expects.

An `if` chain reserves one range shared by every arm, sized by its widest arm, so `if`/`else` reuse the element across a toggle the way hand-written `{cond ? <a/> : <b/>}` does:

```js
let __jac_view_kids = [false, false, false];
if (a) {
  __jac_view_kids[0] = X;
  __jac_view_kids[1] = Y;
} else {
  __jac_view_kids[0] = Z;
}
__jac_view_kids[2] = __jacJsx("hr", {}, []);
```

Loops keep their own appending accumulator returned as a raw array, since their length is only known at run time and the author's per-item keys are what identify those children. That is the one place React should still require keys, and it still does.

## Statements stay statements

Nothing is lifted into an expression. No closure sits between a `skip`, `break`, or `return` and the loop it was written in, and no `if` chain can disagree with itself about its own shape.

This matters because the first version of this PR did lift conditionals into ternaries, and that produced two silent miscompiles: `skip` returned from the lifted closure instead of the enclosing loop, and an `elif` arm whose own subtree was clean lifted away from a parent that could not, stranding its expression as a discarded `else` clause that ASI accepts. Both needed guards, and the invariant those guards enforced, that every arm of a chain agrees on statement-or-expression, was not expressible anywhere in the compiler.

Under this lowering both cases are correct with no special handling at all, and the guards are gone with the approach that needed them. The history is kept as three commits, a revert then the replacement then the tests, so the two shapes can be compared directly.

## Verification

Nine tests in `test_slot_child_positions.jac` covering the fragment shape, a single conditional, sibling conditionals, `if`/`else` sharing a position, an `if` reserving a range sized by its widest arm, loop nesting, a loop not shifting its siblings, `skip` reaching its loop, and an `elif` chain keeping every arm on the same footing.

Running the compiled shapes against react-dom 18.3.1 under jsdom, with a component holding `useState` and a mount counter, mounting with `a=true, b=true`, bumping B's state, then re-rendering with `a=false`:

| lowering | key warning | B mounts | node reused | B's state after |
| --- | --- | --- | --- | --- |
| before | yes | 2 | no | 1 -> 0 |
| after | no | 1 | yes | 1 -> 1 |

Also confirmed in a browser on a `jac start --dev` scaffold: the page renders and the console is clean.

## Tests run

`tests/compiler/passes/ecmascript/`, 217 passed, including the Solid and Preact backend suites since the change sits in the shared pass. Re-run after rebasing onto the latest `main` merge. Broader suites were not run.

## Assertions touched outside the new tests

Three existing assertions matched the old `push` shape and are worth a reviewer's eyes, since editing someone else's regression test to make a change pass is exactly how a regression hides.

- `#6732` accepts either write form now, `[n] = ` or `.push(`, because a counted slot assigns while a loop body appends. Same guarantee, that a bare expression statement reaches the accumulator rather than being discarded, with slightly less precision about which form. The trailing `.push(label)` assertion is untouched, so the loop half stays exactly as strict.
- `sibling slots` matches any initializer rather than the literal `[];`. Its bare-reassignment guard, which exists for a real `ReferenceError` crash, got broader in the process and now also catches a bare `acc = [false]`.
- `test_esast_gen_pass` asserts the reserved write `[0] = ` instead of `.push(`, which pins the index rather than only that some write happened.

## Related

#7877 covers the architectural half, that `ReactBackend.lower_view` routes `IfChild`/`EachChild` through `js_fallback` rather than lowering the view IR, which is why this fix lives in the shared pass rather than in the backend. This PR does not depend on it.


MalithaPrabhashana:
Verified the fix first: the spurious key warning from #7859 fires on `main` and is gone on this branch. Sibling conditionals keep their state now, and so does the sibling after a growing loop, which wasn't in the report. All good.

One shape looks off though and is worth a check: a `for` directly inside a `for` in the same slot.

```jac
def:pub Groups(rows: list[list[str]], extra: bool) -> JsxElement {
    return
        <div>
            {for r in rows {
                for c in r {
                    <span key={c}>{c}</span>
                }
            }<input id="footer" />}
        </div>;
}
```

The inner loop opens its own accumulator, so the children come out grouped instead of flat:

```js
render 1:  [ [a1, a2], [b1, b2] ]
render 2:  [ [b1, b2], [a1, a2] ]
```

Those inner arrays carry no key, so React pairs them by position and only compares the `key={c}` values inside each pair. Reordering `rows` rebuilds every item even though every key is correct.

Mounted under react-dom 18.3.1, `rows=[["a1","a2"],["b1","b2"]]` then re-rendered reversed:

| | keys survive reorder | sibling survives growth |
| --- | --- | --- |
| main | 4/4 | fails |
| this branch | 0/4 | passes |
| with the guard below | 4/4 | passes |

Also checked in a browser through `jac start --dev`: put a text box in each item, type, reorder, and every box comes back empty. No console error and no key warning, so nothing surfaces it.

The guard already exists in this PR, in `_view_chain_open` -- it asks both whether there is an enclosing accumulator and whether that accumulator counts positions. `_view_open_loop_acc` only asks the first:

```diff
 impl EsastGenPass._view_open_loop_acc(nd: uni.UniNode) -> None {
-    if self._cur_view_acc() is None {
+    acc = self._cur_view_acc();
+    if (acc is None) or (acc in self._view_dynamic_accs) {
         return;
     }
     self._view_dynamic_accs.add(self._push_view_acc());
     self._view_slot_loops.add(id(nd));
 }
```

Inside another loop there are no reserved indices to protect, so the wrapping isn't needed there. With it applied the emitted diff is confined to nested-loop components, the outer loop still takes exactly one reserved index so sibling stability is unaffected, and the numbers go to the last row above.

How much this matters: I scanned 714 client view modules across the examples, tests, and a pile of generated apps and found zero occurrences, because the usual grouped-list shape wraps each group in a keyed element, which gives every loop its own slot and compiles correctly. So nothing is broken today. It's reachable where a wrapper is illegal though, such as rows grouped inside a `<tbody>` or items in one CSS grid, and there `<React.Fragment key=...>` isn't expressible in Jac, so there's no way to write around it.

Not blocking from my side, but the repair is one line and it holds both properties at once, which neither `main` nor this branch does.

Separately: the release note says conditionals "lower to a single conditional expression", which describes the ternary version this PR reverts rather than what it ships.

SandeepaHWP:
Thanks for catching this, reproduces exactly as you describe and the guard is applied in 528268e41 with a `nested_loops` fixture and test. Release note updated too in ea6190d58.

kugesan1105:
image

----------------------

image

PR #7949, opened 2026-08-07
fix(scale): gate fleet readiness on the applied revision, add an opt-in full-rollout wait (#7768)

### LINKED ISSUE #7768: jac-scale: fleet-ready gate counts >=1 ready replica per service; add an opt-in full-rollout wait
## Summary

`jac start --scale` gates its success on `_wait_for_fleet_rollout`, which is genuinely useful (crash-loop and image-pull detection, self-healing, a clear `"Fleet ready: N/N deployments serving"` signal) — but its readiness bar is **≥1 ready replica per deployment**, which is weaker than `kubectl rollout status` (all updated replicas of the new ReplicaSet available). For multi-replica fleets this reports success while most of the fleet is still rolling, so CI pipelines that care about full capacity have to follow up with their own `kubectl rollout status` loops.

## Current behavior

`_wait_for_fleet_rollout` (`jac/jaclang/scale/deploy/target/kubernetes/microservice/target.jac`) counts a deployment ready when:

```
d.status.ready_replicas >= 1
```

With `replicas: 30` and a rolling update in flight, "Fleet ready" fires as soon as one new pod per service is serving. That's a reasonable fast-feedback default (the gateway can route, the app is up), but it is not "the rollout completed":

- old-ReplicaSet pods may still be draining/replacing for many minutes;
- a rollout that later stalls (quota, scheduling, a node-pool issue) is reported as success;
- pipelines that need full capacity before cutting traffic must re-implement the wait with `kubectl rollout status --timeout=... ` per deployment.

## Proposed fix

Keep the current fast gate as the default and add an opt-in strict mode, e.g.:

```
jac start main.jac --scale --wait=full     # or [scale.kubernetes] rollout_wait = "full"
```

Semantics of `full`, per deployment: `status.updatedReplicas == spec.replicas` AND `status.availableReplicas >= spec.replicas` AND `observedGeneration >= metadata.generation` (the same conditions `kubectl rollout status` checks), bounded by the existing timeout and reusing the existing crash-loop/image-pull failure detection.

The existing `progressDeadlineSeconds` support (settable via `deployment_overlay`) composes with this: a stalled rollout would surface as a failure instead of an indefinite wait.

## Related

- #7767 — the process currently never exits after the wait completes, so the fleet-ready result cannot be consumed as an exit code by CI at all; that issue is a prerequisite for this one to pay off in automation.

## Acceptance criteria

- Default behavior unchanged (fast ≥1-ready gate).
- `--wait=full` (or config equivalent) only reports "Fleet ready" when every deployment's new ReplicaSet is fully available, and fails (non-zero, with the pod-level diagnosis the waiter already collects) on stall/timeout.

akindu-k:
Closes #7768.

## The problem

`_wait_for_fleet_rollout` counted a Deployment ready at `status.readyReplicas >= 1`. That counter spans **every** ReplicaSet the Deployment owns, and nothing checked `observedGeneration`.

The Deployment name is stable across deploys and each deploy patches it, so on any re-deploy the **old** revision's pods already satisfied the gate. `jac start --scale` printed "Fleet ready" on the first poll — before the new ReplicaSet had scheduled a pod, sometimes before the controller had even observed the applied spec. It was reporting success against the pods it was about to delete.

Only the readiness half was affected; the failure-detection half (`_resolve_current_rs_hashes`) was already revision-aware.

### Visual walkthrough

<details>
<summary>Where the problem occurs, and how ReplicaSets make it possible</summary>

phase where the problem occurs
understanding replica sets
issue location
issue location detail
issue location detail
fix explanation

[Full explanation](https://drive.google.com/file/d/1553orvKVrt-a4RubmvjTZG_RI_rni_Tq/view?usp=sharing)

</details>

## The fix

Readiness moved into a new `rollout.jac` and is now measured against the applied revision:

- `observedGeneration` must have caught up with `metadata.generation`;
- an available replica counts only when it provably belongs to the new ReplicaSet: `availableReplicas - (replicas - updatedReplicas) >= 1`. Conservative by construction — it can undercount while old pods drain, never overcount.

`rollout_wait` selects how far to wait:

| Mode | Returns when |
|------|--------------|
| `ready` (default) | every service has ≥1 new-revision replica available |
| `full` | all replicas updated and available, no old replicas left — the `kubectl rollout status` bar |

Set it under `[scale.kubernetes]` in `jac.toml`, or per run with `jac start --scale --wait full` (the flag wins). `rollout_timeout` (default 1200s) bounds both.

## ⚠️ The default is stricter for everyone, not just `--wait full`

On a redeploy `jac start --scale` used to return on the first poll. It now blocks until a new-revision replica is *available* — scheduling, image pull, app boot, readiness probe, `minReadySeconds`. **Seconds before, minutes now.**

That time was always being spent; the command just wasn't waiting for it. But anything shelling out under its own timeout (jacBuilder, jachammer) sized that timeout against the old behaviour. #7768 asked for the default to stay put — a default that reports success against the previous revision isn't fast, it's wrong, so this changes it deliberately. Ships as its own `7949.breaking.md`.

## Attributing a stall to the deploy that caused it

A rollout past its `progressDeadlineSeconds` now fails with the stall reason and pod-level diagnosis instead of spinning to the timeout.

The subtlety: `ProgressDeadlineExceeded` is **sticky**. `DeploymentTimedOut` short-circuits on its own prior verdict, and right after a patch the controller re-asserts the *previous* deploy's timeout against the new rollout (`observedGeneration` behind → not complete; `updatedReplicas` dropped → not progressing). Failing on sight would reject the deploy that fixes a timed-out rollout, using the timeout it is fixing.

So a verdict is keyed to the condition the controller stamped, not to elapsed polls. The waiter records each Deployment's `stall_mark()` — `reason@lastUpdateTime` — on first sight, and `stall(inherited)` stays silent while the mark matches. A stall re-stamped for *this* rollout fails immediately; a predecessor's never fails it, however long the first pod takes.

This replaces an earlier three-poll confirmation window, which was wrong in both directions: too long to sit on a real stall, too short when PVC binding delays the first pod.

## One definition of "complete", not three

`_rollout_status` (ops deploy view) and `KubernetesTargetBase.get_status` each had weaker copies of this logic — no `observedGeneration`, no old-replica check. `get_status` is what `ScaleClient.resource_status()` exposes, so a caller polling after a deploy got `RUNNING` mid-rollout off old-revision pods.

Both now call the shared reader and ask `satisfies(ROLLOUT_WAIT_FULL)`. `read_rollout_status` accepts both client objects and `to_dict()` payloads, so there is one reader and one definition.

Neither is a *waiter*, though — both are stateless single reads with no first-poll mark to pass to `stall(inherited)`, so each can surface the predecessor's sticky condition. Suppressing it would hide real stalls; reporting it bare implies certainty they don't have. So both report it and say what it's worth: `stall_caveat()` is the one definition of the hedge, carrying the controller's own `lastUpdateTime` — a stamp older than the deploy you just triggered is the previous one's.

## Where the config is validated

On the deploy path in `_rollout_gate()`, not in the `KubernetesConfig` constructor. `from_dict` is shared and runs in-pod, where callers swallow exceptions — validating there meant an unusable `rollout_timeout` inside the pod would quietly cost those readers the whole config. Raising on the deploy path fails the one operation the knob governs and leaves every other reader intact.

Numeric strings stay accepted: `jac.toml` env interpolation always yields a string, so `rollout_timeout = "${ROLLOUT_TIMEOUT:-1200}"` is supported. Strict on the *value*, not the carrier — `""`, `"soon"`, `30.5`, `true`, `0` and negatives are rejected.

## Tests

21 new unit tests, plus a redeploy leg in the k8s e2e that runs once per wait mode.

- **`test_fleet_rollout.jac` (12)** — the redeploy case, an unobserved generation, each `full` holding condition, the default gate clearing at exactly one available new replica, the progress-deadline failure, a stall re-stamped for this rollout vs. one inherited from the previous deploy, a stall that clears, a Deployment the API server refuses to serve being named in the timeout, and a read that fails once then recovers not being blamed for it.
- **`test_deploy_k8s.jac` (7)** — the config surface, rejected values for both knobs (with the in-pod read asserted to survive them), the env-interpolated string contract, and `resource_status()` agreeing with the gate.
- **`test_ops_deploy.jac` (2)** — the dashboard agreeing with the gate on stale and draining states.

Unit tests run against mocked API objects, so they pin the decision table but can't prove a live Deployment populates the counters as the arithmetic assumes. The e2e previously deployed once into a fresh namespace, so nothing walked the path being fixed — the defect only exists on a **second** deploy. It now bumps the overlay marker and redeploys last, once per mode: the `ready` leg asserts `deploy()` returns only once every Deployment is observed with `updated_available() >= 1`; the `full` leg asserts `is_complete()` on an independent read. Each leg asserts the revision advanced, so neither passes vacuously the day the bump stops producing a new ReplicaSet.

## Judgment calls

- `full` reads `spec.replicas` live each poll, so an autoscaler resizing mid-rollout moves the target — matching `kubectl rollout status` rather than pinning the applied value.
- `resource_status()` reports `PENDING` when `updated > 0` with nothing available yet, so a started rollout is never indistinguishable from a failed one.
- The ops view still calls a baseline-free condition `stuck`: for an advisory page a false `stuck` costs a glance, a missed one costs the thing the page exists for.


jac/jaclang/scale/deploy/target/kubernetes/rollout.jac
```diff
@@ -0,0 +1,134 @@
...
+        (seconds, is_whole) = (0, False);
+    }
+    if isinstance(value, bool) or seconds < 1 or not is_whole {
+        raise ValueError(
+            f"invalid rollout timeout '{value}': expected a positive whole "
+            "number of seconds"
+        );
+    }
+    return seconds;
+}
```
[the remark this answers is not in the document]
akindu-k: Numeric strings have to stay accepted here. jac.toml values support env interpolation, and the interpolated result is always delivered as a string, so a valid timeout can legitimately reach this validator as `"1200"`:

```toml
[scale.kubernetes]
rollout_timeout = "${ROLLOUT_TIMEOUT:-1200}"
```

The suggested `isinstance(value, int)` check would reject that supported pattern (and whole TOML floats like `120.0`). Validation here is intentionally strict on the value rather than the carrier type: `"soon"`, `"30.5"`, `30.5`, booleans, `0` and negatives are all still rejected. Pinned this contract in a3ae633 with the test "an env interpolated rollout_timeout string stays accepted" so it cannot regress silently.

MusabMahmoodh (commented):
Read this against #7768 and walked the waiter and its callers. The diagnosis is better than the issue's - the old gate counted `ready_replicas` across every ReplicaSet, so on a redeploy it passed on the pods the deploy was about to replace, and `available - (total - updated)` is a sound conservative floor for "belongs to the new RS". Deleting the second copy of this logic out of `deploy_health.impl.jac`, dropping the unused `summary()`, and pulling `_fleet_pod_details` out of the timeout path in the same PR that adds the shared reader is the right instinct - the diff is smaller than the problem in places, which is rare.

Approving with comments; the one I'd like an answer on before merge is the stall confirmation, inline.

Worth being explicit about in the release note: this changes the default gate for everyone, not just `--wait full` users. On a redeploy `jac start --scale` now blocks until a new-revision replica is available where it used to return on the first poll - minutes instead of seconds on this stack, once image pull and app boot are in the path. That's the fix working, and I think it's the right call even though #7768 asked for the default to stay put, but anything that shells out to `jac start --scale` under its own timeout (jacBuilder, jachammer) will feel it before anyone reads the note.

The mocked API objects in `test_fleet_rollout.jac` are pre-existing and I'm not asking you to rewrite the file - the ask is the e2e redeploy, inline. `_k8s_config_for_toml` builds a real temp project through the real loader, and the config tests are the strongest part of the suite.

jac/jaclang/scale/deploy/target/kubernetes/rollout.jac
```diff
@@ -0,0 +1,134 @@
+glob ROLLOUT_WAIT_READY: str = "ready",
+     ROLLOUT_WAIT_FULL: str = "full",
+     ROLLOUT_WAIT_MODES: tuple = ("ready", "full"),
+     PROGRESS_DEADLINE_REASON: str = "ProgressDeadlineExceeded",
+     STALL_CONFIRM_POLLS: int = 3;
```
MusabMahmoodh: The write-up on why this can't fail on sight is the best part of the PR, but three polls is 30s and what it's waiting on is the new ReplicaSet creating its first pod - on this stack that can sit behind PVC binding or scheduling for longer than that. When it does, the sticky `ProgressDeadlineExceeded` from the *previous* deploy survives all three polls and we abort the deploy that was fixing it, which is the exact case this guard exists for.

Would keying off the condition rather than the poll count close it - only trust a stall whose `last_update_time` is newer than the moment we patched, or require `updated > 0` before a condition counts as this rollout's? Both are already in hand at that point.

jac/jaclang/scale/tests/microservices/test_fleet_rollout.jac
```diff
@@ -383,3 +410,163 @@ test "a terminating pod is invisible to the scanner (no false re-abort)" {
...
+        _rolled_out(3)
+    ];
+
+    err = _run(target, _bundle("gateway-deployment", "gateway"), apps, core, 60);
+
+    assert err is None , (
+        "the controller re-asserts a stale timeout until the new ReplicaSet "
+        f"makes progress; one poll of it is not a stalled rollout: {err}"
+    );
+}
```
MusabMahmoodh, line 600: You flagged this in the description, so taking it as a question rather than a finding: the defect only exists on a *second* deploy over an existing Deployment, and `k8s_microservice_real_e2e.sh` deploys once into a fresh namespace and then asserts `RS_COUNT` for the first rollout - so nothing outside these fixtures ever walks the path being fixed. A second `target.deploy()` with a bumped env var, asserting the call returns only once the new RS has an available pod, is the one test that proves `updated_replicas` / `available_replicas` behave the way the arithmetic assumes. Worth adding there in this PR rather than waiting for the k8s job to notice?

jac/jaclang/scale/admin/impl/deploy_health.impl.jac
```diff
@@ -24,31 +28,22 @@ def _s(v: any) -> str {
...
+    if rollout.stall() {
+        return {
+            "status": "stuck",
+            "reason": rollout.stall_reason,
+            "message": rollout.stall_message
+        };
     }
-
-    if updated == desired and available == desired {
+    if rollout.satisfies(ROLLOUT_WAIT_FULL) {
```
MusabMahmoodh, line 39: This consolidation is the right move, but "one definition of complete, not two" isn't quite true yet - `kubernetes_target.jac:578` still decides RUNNING on `ready_replicas == replicas`, which counts pods from every ReplicaSet for the same reason the gate used to. That's the surface `ScaleClient.get_status()` (`sdk/client.jac:103`) exposes, so a caller polling status after a deploy still gets RUNNING mid-rollout. Now that `read_rollout_status` exists, can that one route through it too - and does `ResourceStatusInfo.is_ready()` want the same treatment?

jac/jaclang/scale/deploy/target/kubernetes/kubernetes_config.jac
```diff
@@ -177,6 +186,12 @@ obj KubernetesConfig(BaseConfig) {
             bundle_access_mode=config.get('bundle_access_mode', 'ReadWriteMany'),
             resource_deletion_wait=config.get('resource_deletion_wait', 5),
             aws_nlb_wait=config.get('aws_nlb_wait', 60),
+            rollout_wait=normalize_rollout_wait(
+                config.get('rollout_wait', ROLLOUT_WAIT_READY)
+            ),
+            rollout_timeout=normalize_rollout_timeout(
+                config.get('rollout_timeout', 1200)
+            ),
```
MusabMahmoodh: `from_dict` is shared - `get_kubernetes_config()` calls it, and that runs in-pod as well (`admin/impl/logs.impl.jac:29`, `admin/service_urls.jac:14`). Both swallow exceptions, so a `rollout_timeout` that interpolates to something unusable inside the pod (env var set on the deploy host, absent in the pod) doesn't crash anything - it quietly makes the admin logs view resolve an empty namespace and show nothing. Would these two knobs be better validated where the deploy reads them than in the constructor every caller shares?

MusabMahmoodh (approved):
Approving. The stall attribution is the design comment 1 was hoping for: a verdict only counts once its stamp differs from the one inherited at deploy start, so this rollout's own ProgressDeadlineExceeded fails on sight instead of after three polls, while a predecessor's never fails it, however long the first pod takes to schedule. The redeploy leg in the e2e is the one test that proves the gate against a real cluster - revision bump asserted, updated_available >= 1 on return. And get_status routing through read_rollout_status closes the mid-rollout RUNNING gap that jachammer polls.

jac/jaclang/scale/deploy/target/kubernetes/kubernetes_target.jac
```diff
@@ -574,26 +578,29 @@ obj KubernetesTargetBase(DeploymentTarget) {
...
+            ready_replicas = rollout.updated_available();
 
-            if ready_replicas == replicas and replicas > 0 {
+            if rollout.satisfies(ROLLOUT_WAIT_FULL) and replicas > 0 {
                 status = ResourceStatus.RUNNING;
-            } elif ready_replicas > 0 {
+            } elif ready_replicas > 0 or rollout.updated > 0 {
                 status = ResourceStatus.PENDING;
             } else {
                 status = ResourceStatus.FAILED;
```
MusabMahmoodh, line 597: On a redeploy where the new ReplicaSet can't create pods yet (quota, admission webhook), `updated` is 0 while the old replicas are all still available, so this reports FAILED with no message until the progress deadline stamps - main reported RUNNING there, and the truth is PENDING. Should something-still-serving hold it at PENDING?

```
} elif ready_replicas > 0 or rollout.updated > 0 or rollout.available > 0 {
    status = ResourceStatus.PENDING;
}
```

jac/jaclang/scale/deploy/target/kubernetes/kubernetes_target.jac
```diff
@@ -574,26 +578,29 @@ obj KubernetesTargetBase(DeploymentTarget) {
...
             }
 
             return ResourceStatusInfo(
-                status=status, replicas=replicas, ready_replicas=ready_replicas
+                status=status,
+                replicas=replicas,
+                ready_replicas=ready_replicas,
+                message=rollout.stall() or None
```
MusabMahmoodh: The deploy gate goes to some trouble to ignore a stall inherited from the previous deploy, but `get_status` has no baseline so `rollout.stall()` surfaces the predecessor's sticky ProgressDeadlineExceeded right after a fresh redeploy - a status poller can see PENDING plus the old deploy's stall message on the deploy that fixes it. Intended, or worth noting in the message that the verdict may predate the current rollout?

jac/jaclang/scale/runtime/realizer/kubernetes_realizer.jac
```diff
@@ -69,6 +71,9 @@ obj KubernetesRealizer {
         } else {
             target_config = scale_config.get_section(target);
         }
+        if rollout_wait {
+            target_config['rollout_wait'] = normalize_rollout_wait(rollout_wait);
```
MusabMahmoodh, line 74: This override seam is the one piece of new plumbing no test walks - the config tests stop at `_rollout_gate` and the wait tests call `_wait_for_fleet_rollout` directly. A small test that `realize` with `rollout_wait="full"` lands `"full"` in the created target's `k8s_config` would pin the flag-beats-toml contract the help text promises.

jac/jaclang/cli/docs/reference/cli/index.md
```diff
@@ -273,6 +273,7 @@ jac start [-h] [-p PORT] [-m] [--no-main] [-f] [--no-faux] [-d] [--no-dev] [-a A
 | `--platform` | Mobile start/dev platform selector for `--client mobile` (`auto`, `android`, `ios`) | `auto` |
 | `--scale` | Deploy to Kubernetes (built-in scale subsystem) | `False` |
 | `--target` | Deployment target (with `--scale`) | `kubernetes` |
+| `--wait` | How far to wait for the rollout: `ready` (a new-revision replica per service) or `full` (every replica available) | `ready` |
```
MusabMahmoodh: The table lists the `--wait` default as `ready`, but the flag's actual default is empty, meaning defer to `[scale.kubernetes] rollout_wait` - someone with `rollout_wait = "full"` in jac.toml would read this as the flag silently resetting to `ready`. Maybe state the default as "from jac.toml (`ready`)".

jac/jaclang/scale/deploy/target/kubernetes/kubernetes_target.jac
akindu-k: Valid, and it's a regression this PR introduced rather than something `main` had - worth saying plainly. `main` reported RUNNING there off the old replicas; routing through the shared reader turned that into FAILED with no message, on an app that is up and serving. Both are wrong and FAILED is the worse of the two, since a poller can reasonably act on it.

Fixed in c24c9ae5e, though not literally as suggested - `ready_replicas` here is `rollout.updated_available()`, which is `max(available - outdated, 0)` and so always `<= rollout.available`. The `ready_replicas > 0` clause can never be true when `available > 0` is false, so it is fully subsumed and the third clause alone covers it:

```jac
} elif rollout.available > 0 or rollout.updated > 0 {
```

What falls out reads the way the words should:

| Condition | Status |
|---|---|
| new revision fully available | `RUNNING` |
| anything serving, or any new-revision replica | `PENDING` |
| nothing serving and nothing rolling out | `FAILED` |

Pinned by `test "status stays PENDING while old replicas serve a redeploy that placed no pod"`, which asserts both directions - `updated=0, available=3` is PENDING, and `updated=0, available=0, total=0` is still FAILED. Without the second case "return PENDING more often" would pass vacuously.

jac/jaclang/scale/deploy/target/kubernetes/kubernetes_target.jac
akindu-k: Not intended, and worth the note - `get_status` is a single stateless read, so unlike the gate it has no "before" to compare against and structurally cannot tell this rollout's stall from the predecessor's. Suppressing it would hide real ones; reporting it bare claims a certainty the call doesn't have.

So it now reports it and says what it's worth. `stall_note()` in c24c9ae5e:

```jac
def stall_note -> str {
    stall = self.stall();
    if not stall {
        return "";
    }
    stamped = f"recorded {self.stall_marked_at}, " if self.stall_marked_at else "";
    return f"{stall} ({stamped}may predate the current rollout)";
}
```

and `get_status` passes `message=rollout.stall_note() or None`.

The timestamp is the part that does the work rather than the caveat text: it's the controller's own `lastUpdateTime`, so a caller holding the moment it triggered the deploy has the one fact needed to attribute the verdict itself - a stamp older than the patch is the previous deploy's. It composes over `stall()` rather than reimplementing it, so the "silent until `observedGeneration` catches up" rule is inherited and there's still one stall definition.

`test "status reports the stall the deploy gate would fail on"` now asserts both the stamp and "may predate" reach the caller.

For the record on the sibling surface: `_rollout_status` in the ops view has the same missing baseline and was deliberately left alone - it returns `reason` and `message` as separate fields so the machine-readable verdict is already isolated, the ambiguity there lives in the `"stuck"` status itself which a message caveat wouldn't fix, and `DeployPage` renders only `status`. One line reusing `stall_note()` if you'd rather have it consistent.

jac/jaclang/scale/runtime/realizer/kubernetes_realizer.jac
akindu-k: Agreed - the config tests prove the read and the wait tests prove the waiter, and nothing proved the join. A break anywhere along `execution.impl.jac` -> realizer -> `DeploymentTargetFactory.create` would have shipped silently: `--wait full` accepted, nothing printed, `ready` used.

`test "--wait overrides jac.toml on the target that runs the deploy"` in c24c9ae5e. It runs a real `KubernetesRealizer(...).realize(..., dry_run=True, rollout_wait=...)` with `print_plan` patched, and pulls the target out of `print_plan.call_args.args[0]` - so it asserts on the deployment target the deploy would actually have run with, not on the dict key the realizer sets. The end of the seam rather than the middle, so any break along it fails the test.

Both directions, since the second is really the contract:

- toml `ready` + `--wait full` -> `"full"` (the flag wins)
- toml `"full"` + no flag -> `"full"` (an absent flag does not reset to the default)

jac/jaclang/cli/docs/reference/cli/index.md
akindu-k: Valid, and the misreading has an edge to it: someone who set `rollout_wait = "full"` deliberately would read that table and conclude they now have to pass `--wait full` on every invocation to keep the behaviour they already had.

Reworded in c24c9ae5e to state the rule rather than a value:

| `--wait` | ... | `[scale.kubernetes] rollout_wait` in `jac.toml` (`ready`) |

That names the key that actually supplies the default and keeps `ready` visible as the effective value when nothing is configured, so the column still answers "what if I do nothing?" in one glance. The `--help` text already said "Overrides [scale.kubernetes] rollout_wait", so the two surfaces agree now.

The precedence this sentence describes is also the `_realized_target("full", "")` half of the new realizer test from your comment above, so the docs and the test now assert the same rule.

MusabMahmoodh (approved):
This is a much better shape than the version I read before, and the things I raised last time all came back answered rather than papered over.

The stall handling in particular. Keying the verdict to `reason@lastUpdateTime` instead of a poll count is the right move, and it holds for the reason you'd want: k8s `SetDeploymentCondition` returns early when status and reason are both unchanged, so a re-asserted identical timeout keeps its `lastUpdateTime` and a predecessor's condition can never fail the deploy that's fixing it. No fixed window to be wrong in either direction. The e2e redeploy leg, `get_status` going through `read_rollout_status`, and moving validation out of the shared `from_dict` into `_rollout_gate()` all landed too.

The one that didn't move is the release note. The description covers the slowdown thoroughly, but the fragment is what ships and it reads as a pure correctness fix. Two caller-visible changes belong in it: `jac start --scale` on a redeploy now blocks for the real rollout, and `ScaleClient.get_status()` (`sdk/client.jac:103`, what jachammer polls) stops reporting RUNNING mid-rollout. Anything shelling out under its own timeout feels both before it reads anything else. #8175 in this same window ships a separate `.breaking.md` alongside its feature fragment for exactly this. Is there a jacBuilder/jachammer issue tracking the timeout bump, or should one go up with this?

Two smaller ones.

`deploy_health.impl.jac:31` calls `rollout.stall()` with no inherited mark, while the waiter is careful to pass one and `stall_note()` hedges its text with "may predate the current rollout". So the ops deploy view can report `"status": "stuck"` off the previous deploy's condition during exactly the window a new rollout is placing its first pod. Deliberate, on the grounds that the dashboard is advisory and a false stuck is cheaper than a missed one? If so it's worth a line saying that, since the reason `stall()` takes an `inherited` argument at all is right above it.

`target.jac:479`, the `except Exception { _ = 0; }` in the poll loop means an unreachable API server, an RBAC refusal and a genuinely un-ready fleet all end as `timed out after 1200s (0/N deployments serving)` twenty minutes later. Pre-existing, but this PR rewrites the loop and now owns that message. Holding the last exception and appending it next to `_fleet_pod_details` turns a twenty-minute mystery into one line.

And `full` mode is still only proven against `SimpleNamespace` fixtures. The redeploy leg exercises the default gate, which is the actual defect, so that was the right leg to add first, but `full` is what the issue asked for and its arithmetic is the load-bearing part. The Deployment is already up at that point, so a second `deploy()` with `rollout_wait="full"` is a few lines onto the block you just wrote. Worth it, or does the runtime cost push it past what the kind job can carry?

The `after <= before` revision check at the end of that leg is the detail that makes it a real test, by the way. Without it the whole block passes vacuously the day the overlay bump stops producing a new ReplicaSet, which is how this class of e2e usually rots.

release_notes/unreleased/jaclang/7949.bugfix.md
```diff
@@ -0,0 +1 @@
+- **Fix: `jac start --scale` reported "Fleet ready" while the old pods were still serving**: on a re-deploy the readiness gate counted replicas from the previous revision, so the command returned success before the new pods had started, and a rollout that stalled past its progress deadline waited out the full timeout with no explanation. Readiness is now measured against the revision this deploy applied, a stalled rollout fails with the reason and per-pod diagnosis as soon as the verdict belongs to this deploy rather than a previous one, deployment status and the ops deploy view report that same verdict, and `jac start --scale --wait full` (or `rollout_wait = "full"` under `[scale.kubernetes]`, bounded by the new `rollout_timeout`) holds until every replica of every service is available for pipelines that need full capacity before cutting traffic.
```
MusabMahmoodh, line 1: This is the one that didn't move. The description covers the slowdown thoroughly, but the fragment is what ships and it reads as a pure correctness fix. Two caller-visible changes belong in it: `jac start --scale` on a redeploy now blocks for the real rollout, and `ScaleClient.get_status()` (`sdk/client.jac:103`, what jachammer polls) stops reporting RUNNING mid-rollout. Anything shelling out under its own timeout feels both before it reads anything else.

#8175 in this same window ships a separate `.breaking.md` alongside its feature fragment for exactly this. Is there a jacBuilder/jachammer issue tracking the timeout bump, or should one go up with this?

jac/jaclang/scale/admin/impl/deploy_health.impl.jac
```diff
@@ -24,31 +28,22 @@ def _s(v: any) -> str {
...
-        if c.get("type") == "Progressing"
-            and str(c.get("reason", "")) == "ProgressDeadlineExceeded" {
-            return {
-                "status": "stuck",
-                "reason": "ProgressDeadlineExceeded",
-                "message": _s(c.get("message"))
-            };
-        }
+    rollout = read_rollout_status(dep);
+    if rollout.stall() {
```
MusabMahmoodh, line 32: This calls `stall()` with no inherited mark, while the waiter is careful to pass one and `stall_note()` hedges its text with "may predate the current rollout". So the ops deploy view can report `"status": "stuck"` off the previous deploy's condition during exactly the window a new rollout is placing its first pod.

Deliberate, on the grounds the dashboard is advisory and a false stuck is cheaper than a missed one? If so it's worth a line saying that, since the reason `stall()` takes an `inherited` argument at all is right above it.

jac/jaclang/scale/deploy/target/kubernetes/target.jac
```diff
@@ -422,20 +468,29 @@ obj KubernetesTarget(KubernetesTargetBase) {
...
                             ready += 1;
                         }
+                        if name not in inherited_stalls {
+                            inherited_stalls[name] = rollout.stall_mark();
+                        }
+                        stall = rollout.stall(inherited_stalls[name]);
+                        if stall {
+                            stalls[name] = stall;
+                        }
                     } except Exception {
```
MusabMahmoodh: An unreachable API server, an RBAC refusal and a genuinely un-ready fleet all end up as `timed out after 1200s (0/N deployments serving)` twenty minutes later. Pre-existing, but this PR rewrites the loop and now owns that message - holding the last exception and appending it next to `_fleet_pod_details` turns a twenty-minute mystery into one line.

jac/jaclang/scale/scripts/k8s_microservice_real_e2e.sh
```diff
@@ -688,6 +688,74 @@ else
         "200|404|405|000" "${FIRST_SVC}-deployment" "" "5"
 fi
 
+# The readiness gate only has a revision to get wrong on a SECOND deploy over
+# an existing Deployment, so it runs last: bump the overlay marker, redeploy,
+# and assert the call returned with a new-revision replica already serving.
+echo "=== redeploy: the wait gates on the applied revision ==="
```
MusabMahmoodh, line 694: The `after <= before` revision check at the end here is what makes this a real test - without it the block passes vacuously the day the overlay bump stops producing a new ReplicaSet, which is how this class of e2e usually rots.

One gap: this exercises the default gate, which is the actual defect and the right leg to add first, but `full` is what the issue asked for and its arithmetic is the load-bearing part, and it's still only proven against `SimpleNamespace` fixtures. The Deployment is already up at this point, so a second `deploy()` with `rollout_wait="full"` is a few lines onto the block you just wrote. Worth it, or does the runtime cost push it past what the kind job can carry?

MusabMahmoodh (commented):
Much better shape than the version I read before, and everything I raised last time came back answered rather than papered over. Keying the stall verdict to `reason@lastUpdateTime` instead of a poll count is the right move, and it holds for the reason you'd want: k8s `SetDeploymentCondition` returns early when status and reason are both unchanged, so a re-asserted identical timeout keeps its `lastUpdateTime` and a predecessor's condition can never fail the deploy that's fixing it. Four follow-ups inline, none blocking.

release_notes/unreleased/jaclang/7949.bugfix.md
akindu-k: Agreed, and the fragment is the right place to lose that argument. Split in 4c91fcda6, following #8175's shape: the bugfix fragment stays the correctness story, and a new `7949.breaking.md` carries the two things a caller feels.

It leads with the redeploy wait and names the cost in the units that matter ("seconds before, minutes now"), says plainly that anything running `jac start --scale` under its own timeout needs that timeout raised, and names `rollout_timeout` as the bound. Then the status half, since it is the one that can be misread silently: `ScaleClient.resource_status()` reports `PENDING` while a rollout is in flight and `RUNNING` only once the new revision is fully available, so a poller that read `RUNNING` as "the new code is live" was being answered off the previous revision.

On the tracking issue: there is nothing open on the builder side for it today. The nearest are a deploy timeout issue about in-process app seal and the deploy-reliability follow-up train, and neither covers a bump for this. One should go up alongside this PR rather than after someone hits it, so I will open it there with the fragment's wording and link it back here.

jac/jaclang/scale/admin/impl/deploy_health.impl.jac
akindu-k: Deliberate, and you are right that it should say so rather than leave the reader to infer it from the argument sitting one line above. The dashboard is advisory and a false `stuck` costs a glance, where a missed one costs the thing the page exists for, so the verdict stays. Fixed in d421d55d5 by making the call carry what it is worth instead of only being defensible.

Three parts:

- `stall_note()` was doing two jobs, so the caveat is now its own `stall_caveat()` on `RolloutStatus`. Both baseline-free readers compose it, so the hedge has one definition rather than a copy per surface.
- `_rollout_status` keeps `reason` isolated as the machine-readable verdict and appends the caveat plus the controller's `lastUpdateTime` to `message`. That timestamp is the point: a stamp older than the deploy you just triggered is the predecessor's, so the reader gets the evidence rather than only the hedge.
- The comment you asked for is there, and it names the reason the gate can do better: a dashboard read has no previous poll to inherit a mark from.

One thing that fell out while writing it, which is why this went further than a comment. `DeployPage` rendered only `rollout.status` and never `message`, so all of the above would have been invisible on the one surface that shows the word. The chip now carries `message` as its `title`, so the `stuck` badge explains itself on hover.

`test "a baseline-free stall verdict carries when the controller stamped it"` pins both halves: the verdict is still `stuck`, `reason` is still the bare `ProgressDeadlineExceeded`, and the message carries the stamp.

jac/jaclang/scale/deploy/target/kubernetes/target.jac
akindu-k: Taken, and taking ownership of it rather than leaving it as inherited. Fixed in bdceb5dc2, exactly as you framed it: the poll loop holds the last read failure and the timeout message reports it next to `_fleet_pod_details`.

```
timed out after 1200s (0/3 deployments serving). Last deployment read failed: gateway-deployment: (403) Reason: Forbidden: deployments.apps is forbidden. [pod ...]
```

The read stays inside the loop rather than failing on the first exception, since a transient API blip mid-rollout should not abort a healthy deploy, and an undercounted `ready` cannot satisfy the gate anyway. So the loop still runs out, it just stops arriving with nothing to say.

`test "a deployment the API server refuses to serve is named in the timeout"` covers it, and asserts the whole point rather than the string: an unreachable API server, an RBAC refusal and a genuinely un-ready fleet no longer read as the same bare timeout.

jac/jaclang/scale/scripts/k8s_microservice_real_e2e.sh
akindu-k: Worth it, and the cost lands well short of what would make it a judgment call. Added in b3dbf058d.

The Deployment is already up at that point and the fixture runs one replica per service, so `full` over `ready` is the old pod's drain, not a second rollout. The block now runs `redeploy(mode)` twice, once per mode, each bumping the overlay marker and asserting the revision advanced, so neither leg can pass vacuously the day the bump stops producing a new ReplicaSet. The marker bump moved out of the `sed` and into the block as `bump_marker()`, so both legs share one definition and the second leg cannot silently reuse the first's revision.

The `full` leg asserts `is_complete()` on an independent read: every replica updated, every one available, no old replicas left. That is the arithmetic, checked against a live Deployment rather than a `SimpleNamespace`, which was the gap.

Both legs print their own elapsed time, so if the kind job does start feeling it the number is in the log to argue from rather than a guess.

One exposure worth naming since it is mine and not in your comment: the namespace has managed HPAs by then, so an autoscaler resizing a service between `full` returning and the re-read would fail the assertion. The window is milliseconds against a 15s reconcile on an idle namespace, and the failure prints the replica counts, so it fails legibly rather than mysteriously. I kept the check strict rather than loosening it to absorb a race that would itself be worth knowing about.

MusabMahmoodh (commented):
Re-reviewed the four new commits. My approval stands.

All four of my earlier points are resolved, two beyond what I asked for: `stall_caveat()` extracted to `rollout.jac:128-131` so both baseline-free readers compose one definition, rather than the code comment I actually suggested, and the k8s e2e gained a real `full` leg with per-leg revision assertions. You are also right that it is `ScaleClient.resource_status()` and not `get_status()`; my mistake.

**The red `jac-check` is my fault, and it is two lines.** Run `32109524937`, job `95625756670`: `jac fmt --check --lintfix` reports

```
⚠ jac/jaclang/scale/admin/impl/deploy_health.impl.jac, line 32, col 5: Comment can be removed [strip-comments]
⚠ jac/jaclang/scale/admin/impl/deploy_health.impl.jac, line 33, col 5: Comment can be removed [strip-comments]
✖ Error:   jac fmt --lintfix jac/jaclang/scale/admin/impl/deploy_health.impl.jac
```

That is the only blocking diagnostic in the run, and it is the explanatory comment I asked for in my 04:17Z note on that exact file. Because this is a fork PR the CI autofix push does not fire, so it needs a local `jac fmt --lintfix jac/jaclang/scale/admin/impl/deploy_health.impl.jac`. The reasoning survives the comment being stripped anyway, since `stall_caveat()` names itself and your new test's assertion message spells it out. `Everything Passed` is red only because it aggregates this check; the other 20 jobs pass, including both k8s microservice e2e runs and all five `test-scale` shards.

Two more.

**`deploy/target/kubernetes/target.jac:488`** - `last_read_error` is never cleared, so it holds the last failure *ever* rather than the last poll's. A transient 500 on the first poll of a fleet that then reads clean for twenty minutes still ends with "Last deployment read failed: ..." on the timeout, which points the reader at the API server when the real answer was an un-ready fleet. Setting `last_read_error = "";` next to `ready = 0;` at `:471` fixes it. Your new test makes every read fail forever, so it passes either way; a leg where the read fails once and then recovers is the one that pins it.

**`deploy/target/kubernetes/rollout.jac:96`** - merge-order flag rather than a change here. #8304 adds `deployment_rollout_complete()` in `target.jac`, the same file whose `_wait_for_fleet_rollout` you rewrite, and it is a second definition of this predicate that already disagrees: its last clause is `available >= updated` where yours is `available >= desired`, and it defaults a `spec.replicas` of None to 1 where you route `desired <= 0` through `satisfies()`. Whichever lands second gets the conflict. I have raised it there too; can #8304 be pointed at `read_rollout_status(...).satisfies(ROLLOUT_WAIT_FULL)` before it does?

MusabMahmoodh (approved):
Re-read at `e9093720`. Approval stands, and both of my last two points are closed. Half a day late on this one, sorry.

`last_read_error = ""` now sits next to `ready = 0` inside the poll loop (`target.jac:484`), so the timeout names the poll that actually failed rather than the worst thing that ever happened during the wait. And you added the leg I said would pin it: "a read that fails once and then recovers is not blamed for the timeout" injects one 500, then 64 clean reads of a fleet that never becomes ready, and asserts `"Last deployment read failed"` is absent from the message. That fails on the old code and passes on this one, which is the whole point. The sibling test that asserts a 403 *is* named is the right pair for it - an RBAC refusal, an unreachable API server and a slow fleet reading as three different messages is exactly the diagnostic you were after. `jac-check` is green now too.

**The #8304 collision is still live and neither side has moved.** I checked its head just now: `target.jac:130` still defines `deployment_rollout_complete(dep)` and `:557` still calls it, so once both land the same file carries two definitions of the rollout predicate that disagree on two points - its final clause is `available >= updated` where `read_rollout_status().satisfies()` uses `available >= desired`, and it defaults a `spec.replicas` of `None` to 1 where you route `desired <= 0` through `satisfies()`. I have approved both PRs, so I am not blocking either, but whichever merges second inherits a real reconciliation rather than a textual conflict.

@Thamirawaran, restating the ask from your thread since this is the concrete cost: pointing `deployment_rollout_complete` at `read_rollout_status(d).satisfies(ROLLOUT_WAIT_FULL)` before #8304 lands makes the second merge a no-op instead of a decision someone has to make under time pressure. If you would rather not touch it, say so and I will carry the reconciliation myself in a follow-up once both are in.

Issue #7900, opened 2026-08-05
Jac Scale runtime core: 15 validated defects across the plugin surface, gateway, RPC, middleware, config and dry-run CLI (tracking)


MusabMahmoodh:
Tracking issue for a module-by-module review of the jac-scale **runtime core**: the plugin hook surface, routing, planner, realizers, local orchestrator, gateway, transport, RPC, request context, lifecycle, middleware, dry-run CLI and configuration.

Every item below was validated by reading the source as it stands today, not inferred from naming or docs. Each gets its own issue carrying the mechanism, the exact location, the code evidence and a suggested fix.

## Scope reviewed

`plugin.jac`, `runtime/` (routing, planner, realizer, orchestrator, gateway, transport, rpc, context, lifecycle, middleware, cli), `config/`, `_optdeps/`. Roughly 6,376 lines across 58 files.

### Correctness and security

- [x] #7901 Jac Scale: gateway builtin passthrough replays the buffered request body to every service until one accepts, so a non-idempotent walker can execute more than once @kashmithnisakya 
- [x] #7902 Jac Scale: sv-to-sv RPC retries walker and function POSTs on read timeout, re-executing walkers that already ran @kashmithnisakya 
- [x] #7903 Jac Scale: rate limiter trusts X-Forwarded-For unconditionally, so any client bypasses per-IP limits with a single header
- [ ] #7904 Jac Scale: StateBackendRateLimiter performs its read-modify-write even when it never acquired the lock, failing open under contention

### Performance and robustness

- [x] #7905 Jac Scale: a failure resolving microservice routes is swallowed with no log and silently starts the app as a monolith
- [x] #7906 Jac Scale: get_scale_config() ignores project_dir after first construction and its loaded dict is mutated in place by consumers
- [ ] #7907 Jac Scale: the gateway opens a new aiohttp ClientSession per forwarded request, so every proxied call pays full connection setup
- [ ] #7908 Jac Scale: gateway proxy and passthrough buffer the entire request body in memory with no size cap
- [ ] #7909 Jac Scale: circuit breaker records one failure per exhausted retry loop and admits unlimited concurrent half-open probes @kashmithnisakya 
- [x] #7910 Jac Scale: install_structured_logging replaces all root log handlers and forces INFO, wiping host application logging config
- [x] #7911 Jac Scale: aggregate_openapi uses blocking urllib inside an async function, stalling the event loop up to 5s per service

### Structural gaps

[...]

Issue #8000, opened 2026-08-09
Postgres clean-break tail: retire the Mongo/Redis/Firestore/SQLAlchemy scaffolding #7947 left behind


marsninja:
## Summary

#7947 replaced the l1/l2/l3 + ChangeSet stack with a Postgres-native architecture and deleted ~10k lines. The demolition was correct; what it left behind is the scaffolding that used to hold the old stack up. This issue is the second half of that clean break: retire every remaining Mongo/Redis/Firestore/SQLAlchemy affordance so `runtimelib/store.jac` + `runtimelib/session.jac` is the only persistence story in the tree.

Read against `main` @ `b73f8ebd6` (0.36.0). Five of the findings are live defects, not cosmetics: an ejected backend that cannot import, a documented config value that raises on boot, a 503 path that is unreachable, a missing index, and the admin ops dashboard silently hiding the Postgres health signal.

Scope: ~2,000 lines of source/test deletion, ~800 lines of docs rewrite, and four packages (`sqlalchemy`, `pymongo`, `redis`, `testcontainers[mongodb,redis]`) off the CI install list in six jobs.

---

### 1.1 `jac eject` emits a backend that cannot import

`jac/jaclang/cli/commands/eject_targets/backend.jac:561` writes into the generated `main.py`:

```python
from jaclang.runtimelib.storage.sqlalchemy_memory import SqlAlchemyMemory
```

and line 124 seeds the vendor closure with the same module:

```jac
sql_seeds = ["jaclang.runtimelib.storage.sqlalchemy_memory"] if sql else [];
```

#7947 deleted `runtimelib/storage/sqlalchemy_memory.jac` and its impl. The vendorer does not notice: `_module_to_file` returns `None` for a missing module (`jac/jaclang/publish/runtime_vendor.jac:99-112`) and the closure walk just skips it (`runtime_vendor.jac:257-260`). So eject prints

```
✔ SQLAlchemy persistence (vendored runtime backend)
✔ vendored jaclang runtime (N modules)
```

and exits 0, having produced a tree that dies with `ModuleNotFoundError` on first run.

Triggered by `[eject.db] driver = "sqlite" | "sql" | "sqlalchemy"` (`backend.jac:103`).

Two fixes, both wanted: point the emitted backend at the Postgres store, and make `build_runtime_closure` fail loudly on an `extra_seeds` entry that does not resolve. A silent `continue` on an explicitly requested seed is the wrong default.

[...]

PR #8014, opened 2026-08-10
fix(scale): validate [scale.*] jac.toml keys against the config schema

### LINKED ISSUE #7914: Jac Scale: unknown keys under [scale.*] are silently ignored because the declared config schema is never used to validate
**Area:** runtime core / config
**Severity:** gap (diagnosability)

### What happens

Configuration is loaded from `jac.toml` with no validation against the declared schema. `JacScalePluginConfig.get_config_schema()` exists and documents the `[scale.*]` surface, but nothing consults it when loading, so unknown or misspelled keys and sections are silently ignored.

### Where

`jac/jaclang/scale/config/impl/config_loader.impl.jac` (load path, no validation present)
`jac/jaclang/scale/config/plugin_config.jac:10` (`get_config_schema`, no validating consumer)

### Why it matters

This is the exact class of failure as the historical `[plugins.scale.*]` regression, where a section rename meant pod dependency configuration was silently dropped and nothing reported it. A user who writes `[scale.microservice]` instead of `[scale.microservices]`, or `cpu_limits` instead of `cpu_limit`, gets defaults with no warning anywhere in the CLI, the dry run or the logs.

### Additional drift

`get_default_config` and `get_config_schema` define overlapping but different key sets, maintained by hand. The schema omits the `websocket`, `admin`, `scheduler`, `auth`, `emailer` and `telemetry` sections that the defaults define, and the two disagree on `microservices.enabled` semantics (schema documents a `False` default; the normaliser treats absent as `None`, meaning auto-enable when routes exist).

### Suggested fix

Validate the loaded config against the schema at load time and emit a warning (or an error under `--strict`) listing unknown keys with their nearest valid neighbour. Generate the defaults from the schema, or the schema from the defaults, so there is one source of truth rather than two hand-synced ones.

Part of #7900

kashmithnisakya:
Fixes #7914 (part of #7900).

## What

Configuration under `[scale.*]` was loaded from jac.toml with no validation. `PluginConfigBase.load` deep-merges the raw table over the defaults, every getter cherry-picks the keys it knows with `.get(key, default)`, and unknown or misspelled keys and sections are silently ignored. `JacScalePluginConfig.get_config_schema()` documented the surface but had zero callers anywhere in the repo, and it had drifted from the runtime: the `websocket`, `webhook`, `admin`, `scheduler`, `auth`, `emailer` and `telemetry` sections were missing, only 16 of the ~60 accepted `[scale.kubernetes]` keys were declared, and it disagreed with the runtime on `microservices.enabled` (tri-state auto-detect, not `False`) and `sso.host`.

## How

- **Schema completed** (`scale/config/plugin_config.jac`): every section and key the runtime reads is now declared, including the keys the deploy pipeline itself bakes into shipped bundles (`database.backend`), the install-time capability surfaces (`[scale.storage]`, `[scale.deploy]`, `[scale.tracing]`, `scheduler.enabled`), the documented microservices boot knobs and `client.static_mounts`, per-service overrides as a machine-readable `nested_each` spec, and top-level `fat_bundle` / `seal_timeout`. The two wrong defaults were corrected. Deliberately free-form tables (`secrets`, `env`, `labels`, `deployment_overlay`, `hpa.behavior`, `routes`) stay open.
- **Validator** (`scale/config/validation.jac` + impl): walks the raw `[scale]` table (pre-merge, so defaults never mask a typo) against the schema and reports unknown keys with a nearest-neighbour suggestion via `difflib.get_close_matches`, e.g. `jac.toml [scale.kubernetes] has unknown key 'cpu_limits'; did you mean 'cpu_limit'? Unknown keys are ignored, so defaults apply instead.` A section named by `[scale.deploy] target` is treated as owned by that registered deploy-target plugin. Non-dict values and malformed shapes are tolerated; the validator can warn, never crash a boot.
- **Wired at the choke point** (`get_scale_config()`): runs once per process at singleton creation and logs each finding as a warning. `JAC_SCALE_CONFIG_STRICT=1` escalates to a `ValueError`. An unparseable jac.toml defers to the pre-existing error path.
- **Dry run** (`scale/runtime/cli/diagnostics.jac`): `PlanValidator` gained a `_check_config_keys` pass emitting warn-severity diagnostics under Global findings in `jac start app.jac --scale --dry-run`.
- **Normalizer pass-through** (`normalize_microservices_config`): the documented `boot_health_timeout` / `boot_max_wait` / `health_monitor_interval` and `client.static_mounts` keys were being stripped before the orchestrator could read them, so those documented features were dead from jac.toml; they now pass through.

## Tests

17 new tests in `scale/tests/misc/test_config_validation.jac`: unknown section/nested/per-service detection with suggestions, open-dict tolerance, capability-intent and documented-key acceptance, plugin-target section ownership, strict-mode raise/clean paths, normalizer pass-through, an end-to-end on-disk jac.toml loaded through the factory, and two drift locks that keep the schema load-bearing (every default key must be declared in the schema, and every scalar schema default must equal the runtime default). Existing suites (config-loader 13, events 15, dispatch 6, service-urls 5) still pass.

## Deeper root cause: the legacy `[plugins.*]` table

Validating inside `[scale]` is not enough to prevent the historical regression this issue cites: a project still carrying `[plugins.scale.websocket]` gets zero warning even in strict mode, because `JacConfig` drops the unknown top-level `[plugins]` table before the scale validator ever sees an entry (`config.plugins["scale"]` stays empty, so there is nothing to validate). Reproduced with a failing test first: loading such a jac.toml produced no output and the runtime used defaults.

`JacConfig` now warns at both parse sites (primary load and toml-file merge) when a non-empty `[plugins]` table is present, naming the sections found and pointing at the flattened location (`[plugins.scale.database]` -> `[scale.database]`). The check is scoped to the `plugins` key only, so arbitrary user tables are untouched and there are no new false positives. Two tests in `jac/tests/project/test_config.jac` pin the warning and its absence on clean configs.

Targeted runs: `test_config.jac` 136 passed, `test_config_validation.jac` 17, `test_config_loader.jac` 13, `test_pvc_injector.jac` 11 (the suite whose fixtures still parse legacy tables). `scripts/check-release-notes.sh` passes.


jac/jaclang/scale/config/plugin_config.jac
```diff
@@ -356,6 +779,38 @@ class JacScalePluginConfig {
...
+                            "default": {},
+                            "description": "Module-name -> entry-file map for services whose file differs from <name>.jac; normally maintained by the deploy pipeline rather than by hand"
+                        },
+                        "shared_volumes": {
+                            "type": "list",
+                            "default": [],
+                            "description": "Volume declarations mounted into every service pod"
                         }
                     }
                 },
```
MusabMahmoodh, line 801: `[scale.events] backend` and `broker` (the block just below) are read by `INTENT_RULES` in `jac/jaclang/project/capabilities.jac:94-98` to pull in `data.redis`, but neither is declared here - so a config that installs fine today warns "unknown key 'backend'", and hard-fails under `JAC_SCALE_CONFIG_STRICT=1`.

And this is the part that worries me: `INTENT_RULES` is already a machine-readable list of the `[scale.*]` sections and keys, so re-typing them by hand into the schema feels like a workaround for the drift the issue is about - it drifted on the first write. Can the capability keys be pulled from `INTENT_RULES` instead? Failing that:

```jac
"backend": {
    "type": "string",
    "default": None,
    "description": "Event broker backend ('redis'); read at install time for capability detection"
},
"broker": {
    "type": "string",
    "default": None,
    "description": "Alias for backend; read at install time for capability detection"
},
```

jac/jaclang/scale/config/impl/config_loader.impl.jac
```diff
@@ -391,6 +391,16 @@ impl get_scale_config(project_dir: Path | None = None) -> JacScaleConfig {
         return _scale_config_instance;
     }
     new_instance = JacScaleConfig(project_dir);
+    raw_plugin_config: dict[str, any] | None = None;
+    try {
+        raw_plugin_config = new_instance.get_jac_config().get_plugin_config('scale');
+    } except Exception {
```
MusabMahmoodh: This swallows the exact failure the validator is here to report - if `get_jac_config()` throws we skip validation and nothing says so anywhere, which is the same silent shape as the bug we're fixing. If the throw is real, `.load()` raises it a moment later anyway so the catch buys nothing; if it can't throw, the try is dead. I'd drop it, or log before falling through.

jac/jaclang/project/impl/config.impl.jac
```diff
@@ -407,6 +408,22 @@ impl JacConfig.get_plugin_config(plugin_name: str) -> dict[str, any] {
...
+    plugins_table = data.get("plugins");
+    if not isinstance(plugins_table, dict) or not plugins_table {
+        return;
+    }
+    legacy: dict[str, any] = dict(plugins_table);
+    sections = ", ".join(sorted(str(k) for k in legacy.keys()));
+    import from jaclang.cli.console { console }
+    console.warning(
+        f"jac.toml has a [plugins] table ({sections}); the [plugins.*] "
+        "namespace was removed and these settings are ignored. Move "
```
MusabMahmoodh, line 421: The message says these settings are ignored, but `jac/jaclang/jac0core/service_cut.jac:29` still falls back to `plugins.scale.microservices.routes` - so a legacy project gets its compile-time service cut applied while the runtime routes come up empty, which is exactly the half-applied config this warning is meant to prevent. Warning about it rather than deleting that fallback feels like a workaround; it's the last reader of the old namespace, can it go in this PR?

MusabMahmoodh (commented):
Read this against #7914 and walked the surrounding code; the schema completion and the choke-point wiring at `get_scale_config()` are the right shape, and the two drift-lock tests are the thing that makes the schema load-bearing instead of decorative. Ran the new suite on a test box, 17 pass.

One that needs fixing before merge: `[scale.events] backend` / `broker` are read by `INTENT_RULES` for capability install but aren't declared in the schema, so the validator warns on a config that works today and hard-fails it under `JAC_SCALE_CONFIG_STRICT=1`. Details inline.

The broader thing I keep coming back to: the issue asked for one source of truth, and hand-completing the schema gives us a fourth hand-synced list of `[scale.*]` keys (`get_default_config`, `get_config_schema`, `KubernetesConfig`'s fields, `INTENT_RULES`). The drift locks catch two of those four, and the `events` miss is the other two drifting on the first write. Not asking you to rebuild it here, but worth deciding whether the schema should be derived rather than typed - and `byllm`, `mcp`, `client` and `desktop` all expose `get_config_schema()` with the identical silently-ignored-keys gap, so a validator that lives next to the convention would cover five plugins instead of one.

Nothing here breaks a running deployment by default - the non-strict path only logs. Two to be aware of: anyone running `JAC_SCALE_CONFIG_STRICT=1` in CI with `[scale.events] backend` set gets a hard failure, and the normalizer pass-through means `boot_health_timeout` / `boot_max_wait` / `health_monitor_interval` now actually reach the orchestrator, so a project that already set those in jac.toml gets different local boot timing on upgrade. Both fine, just worth a line in the release note.

jac/jaclang/scale/config/impl/validation.impl.jac
```diff
@@ -0,0 +1,93 @@
...
+    options = dict(known_scale_options());
+    deploy_cfg = raw.get("deploy");
+    if isinstance(deploy_cfg, dict) {
+        target = deploy_cfg.get("target");
+        if isinstance(target, str) and target and target not in options {
+            options[target] = {
+                "type": "dict",
+                "description": "Config section owned by the registered deploy target plugin"
+            };
+        }
```
MusabMahmoodh: Reading ownership off the `target` string cuts both ways: keeping a second target's config around (`[scale.fly]` while `target = "kubernetes"`) now warns and strict-fails, and a typo'd target quietly silences its own section. `DeploymentTargetFactory._registry` (`deploy/target/factory.jac:41`) already knows the registered target names - any reason not to ask it, and warn when `target` isn't one of them?

jac/jaclang/scale/tests/misc/test_config_validation.jac
```diff
@@ -0,0 +1,228 @@
...
+                "static_mounts": [{"url_prefix": "/assets", "local_path": "public"}]
+            }
+        },
+        env_routes_fallback=False
+    );
+    assert normalized.get("boot_max_wait") == 120 , normalized;
+    assert "boot_health_timeout" not in normalized , normalized;
+    mounts = normalized["client"]["static_mounts"];
+    assert mounts == [{"url_prefix": "/assets", "local_path": "public"}] , mounts;
+}
```
MusabMahmoodh, line 330: Nothing here checks our own shipped configs validate clean, which is what would have caught `events.backend`. Same for `PlanValidator._check_config_keys` - no test, and its body is wrapped in a bare `except Exception { return; }`, so if it throws the dry run just reports nothing. The sweep is cheap and it's the test that keeps the schema honest as the runtime grows:

```jac
test "shipped jac.toml files validate clean" {
    import tomllib;
    root = Path(__file__).parents[4];
    for p in sorted((root / "jac").rglob("jac*.toml")) {
        data = tomllib.loads(p.read_text());
        assert validate_scale_config(data.get("scale", {})) == [] , str(p);
    }
}
```

jac/jaclang/scale/config/plugin_config.jac
kashmithnisakya: the events INTENT_RULES rule was retired by #8000 (no events entry and no data.redis capability left in capabilities.jac), so these keys are read by nothing now and the warning is correct - left them undeclared.

jac/jaclang/scale/config/impl/config_loader.impl.jac
kashmithnisakya: dropped: validation moved into PluginConfigBase.load where the raw table is already in hand, so there is nothing left to guard.

jac/jaclang/project/impl/config.impl.jac
kashmithnisakya: deleted, and there were two readers not one - service_cut.jac:29 and plugin.jac:344 (jac scale split) - so both are gone and the warning now only fires for config nothing reads.

jac/jaclang/scale/config/impl/validation.impl.jac
kashmithnisakya: switched to DeploymentTargetFactory.known_targets() (builtins + registry) so sections come from the real target list, and an unrecognised [scale.deploy] target is now reported instead of silencing its own section.

jac/jaclang/scale/tests/misc/test_config_validation.jac
kashmithnisakya: added the sweep (15 shipped tomls, clean) plus tests for _check_config_keys, whose bare except now emits an honest diagnostic; note your snippet needed parents[5] and a rename off 'root', which is a reserved builtin in jac.

MusabMahmoodh (approved):
All five addressed, and the big one landed at the right altitude: validation now lives with the get_config_schema() convention on PluginConfigBase, opt-in per plugin, so byllm/mcp/client/desktop can adopt it by overriding one method. Verified the events call against main: INTENT_RULES has no events entry since #8000, so leaving backend/broker undeclared is correct. Deleting the two [plugins.*] readers instead of warning past them, and sourcing target sections from DeploymentTargetFactory.known_targets() with a misspelled target reported, both kill the class rather than the instance. The sweep test asserting checked > 0 is the detail that keeps it honest. Approving.

Issue #7909, opened 2026-08-05
Jac Scale: circuit breaker records one failure per exhausted retry loop and admits unlimited concurrent half-open probes


MusabMahmoodh:
**Area:** runtime core / rpc
**Severity:** high (robustness)

### What happens

Two independent problems in the circuit breaker:

1. **Coarse accounting.** `breaker_record_failure` is called only after the RPC retry loop has exhausted all attempts, so three wire failures count as one breaker failure. With a trip threshold of 5, the breaker opens after roughly 15 failed round trips.
2. **Unbounded half-open probes.** `breaker_should_skip` flips `OPEN` to `HALF_OPEN` and returns `False`, then releases the lock. Every concurrent caller during that window also falls through the final `return False`, so an unlimited number of probes hit a service that was just declared unhealthy.

### Where

`jac/jaclang/scale/runtime/rpc/breaker.jac:19-23`
`jac/jaclang/scale/runtime/rpc/rpc.jac:110` and `:195`

### Evidence

```jac
if st.get("state") == "OPEN" {
    elapsed = time.time() - st.get("opened_at", 0.0);
    if elapsed < _BREAKER_COOLDOWN_SECONDS {
        return True;
    }

    st["state"] = "HALF_OPEN";
    return False;
}
return False;
```

```jac
    }
}

breaker_record_failure(module_name);

raise RuntimeError(
    f"sv-to-sv walker spawn '{label}' failed after "
        + f"{max_attempts} attempts: {last_err}"
);
```

### Why it matters

The breaker is intended to shed load from a failing dependency. Opening only after ~15 round trips, then admitting a full concurrent burst on the first probe, means it provides much less protection than the constants suggest.

### Suggested fix

Record a failure per attempt rather than per exhausted retry loop. Add a `PROBING` state (or a probe token) set under the same lock so only one caller performs the half-open trial while others continue to be skipped.

Breaker state is also per process, so each replica learns independently, and no metrics are emitted for breaker transitions.

Part of #7900

marsninja:
Closing: no activity in 10+ days. Reopen if still relevant.

MusabMahmoodh:
Flagging rather than reopening, since I do not have triage here.

#8064 is open against this and names it in the title ("#7909 - Bug 2"). @akindu-k is assigned and it was last pushed on Aug 28, so the work is in flight rather than stale. Worth reopening so the PR still has something to close.

PR #8029, opened 2026-08-11
fix(scale): record a circuit breaker failure per RPC round trip (#7909 - Bug 1)

akindu-k:
Issue #7909 (bug 1). Bug 2, the unbounded concurrent HALF_OPEN probes, is left for a follow-up PR.

## Problem

`breaker_record_failure` was called only *after* the sv-to-sv retry loop had exhausted all three attempts, so three failed wire round trips counted as a single breaker failure. With `_BREAKER_TRIP_THRESHOLD = 5`, the breaker did not open until roughly 15 failed round trips, each costing up to the 10s RPC timeout plus backoff. The constant said "shed load after 5 failures"; the code delivered about a third of that protection.

image

---
image

## Root cause

The surface symptom is one misplaced call, but it sat outside the loop in *two* places: `service_call` and `walker_call` each hand-rolled an identical retry-with-backoff loop and bolted the breaker onto the outside of it. The breaker's unit of health is a wire round trip, while the recording site was a whole retry loop, and the duplication is what let those two drift apart. Patching the call site twice would leave the same class of bug open for the next RPC path added.

## Fix

Collapse both loops into a single `_rpc_with_retries` helper that owns the attempt lifecycle:

- Records one breaker failure per failed round trip, so the trip threshold now counts what its name says.
- Stops retrying as soon as the breaker opens mid-call. Without this, per-attempt recording would introduce a new problem: each further attempt re-enters the `fails >= threshold` branch and refreshes `opened_at`, pushing the cooldown window forward. It also gives the half-open probe the right semantics, spending exactly one round trip instead of three before re-opening.
- Reports the attempts actually made in the failure message, which stays byte-identical to the old text when the loop runs to completion.

Retry count and initial backoff move to module-level constants instead of being redeclared in each function. Success recording keeps its existing position relative to envelope decoding, so an application-level error envelope still does not trip the breaker.

## Tests

Three tests were added, all failing before the change:

- The breaker opens after exactly `_BREAKER_TRIP_THRESHOLD` failed round trips (was 15).
- A breaker that trips mid-call abandons its remaining retries, so a half-open probe spends one round trip (was 3).
- The walker path records one failure per round trip, covering the second call site through the shared helper.

`jac test jac/jaclang/scale/tests/server/test_sv_auth_forward.jac` passes 29/29, and `test_sv_streaming.jac` and `microservices/test_runtime_integration.jac` were run as the other users of this path.

## Note on scope

The issue also mentions that breaker state is per process and that no metrics are emitted for transitions. Both are real, but neither is part of this defect; they are left alone here.


jac/jaclang/scale/runtime/rpc/rpc.jac
```diff
@@ -5,9 +5,44 @@ import from .breaker {
...
+
+    backoff: float = _RPC_INITIAL_BACKOFF_SECONDS;
+    last_err: Exception | None = None;
+    attempts: int = 0;
+
+    for attempt in range(_RPC_MAX_ATTEMPTS) {
+        attempts = attempt + 1;
+        try {
+            payload = send();
+            breaker_record_success(module_name);
```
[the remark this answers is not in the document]
akindu-k: I think this one is a false positive. The success recording was already in that position before this PR, so nothing moved.

Here is `service_call` on `main` today, unchanged by this PR:

```jac
(kind, payload) = post_stream_aware_sync(...);
breaker_record_success(module_name);
if kind == 'sse' {
    return payload;
}
return _unwrap_sv_envelope(payload, module_name, func_name);
```

`walker_call` was the same shape: `resp.json()`, then `breaker_record_success`, then `_hydrate_walker_envelope`. The refactor kept that ordering on purpose so the breaker semantics would not shift while I was fixing the failure accounting.

I also checked it rather than just reading the diff. I wrote a temporary test that records two transport failures, then makes a call whose transport succeeds but returns an `{"ok": false, ...}` envelope, and then reads the breaker state. Running it against this branch and against the current `main` version of `rpc.jac` gives the same result both times:

```
branch: OBSERVED fails=0 state=CLOSED
main:   OBSERVED fails=0 state=CLOSED
```

So the behavior you are describing is real, but it predates this PR and is not affected by it.

On whether it should change anyway, I do not think it should. This is a transport breaker rather than an application health breaker. It only records failures for `ClientConnectionError` and `OSError`, and there is already a test called "application error envelope does NOT trip the breaker" that pins that contract. A completed round trip that comes back with a well formed error envelope is evidence that the dependency is reachable and answering, so clearing the transport failure count is the correct reading of "the transport worked".

Making the suggested change would also leave the contract in a strange half state, where application errors start affecting the breaker in one direction by blocking the reset, but still do not affect it in the other direction by never incrementing. And in the flapping case, half the round trips are succeeding at the transport layer, so opening the breaker would mean refusing to call a service that is demonstrably up.

There is a fair question buried in here though, which is whether some responses should count as negative health signals, for example a 503 or an error code that means the database pool is exhausted. A service that is up but failing everything under load will never trip this breaker today. That seems worth doing, but it changes the contract the existing test asserts and it needs a decision about which responses qualify, so I would rather handle it as its own issue than fold it into a PR about failure counting.

jac/jaclang/scale/runtime/rpc/rpc.jac
```diff
@@ -5,9 +5,44 @@ import from .breaker {
     _BREAKER_COOLDOWN_SECONDS
 }
 
+glob _RPC_MAX_ATTEMPTS: int = 3,
```
MusabMahmoodh, line 8: The counting fix is right, but it changes what `_BREAKER_TRIP_THRESHOLD = 5` *means* without changing the number: a failure used to be a whole retry loop, now it is a round trip, so with 3 attempts the breaker opens after two failed calls instead of about five. During a rolling restart of a downstream service - or at fleet boot, where a service calls a peer that is not up yet - that is 30s of fail-fast after two calls, per process.

Was leaving the threshold at 5 deliberate? If that number was calibrated against observed behaviour rather than against the docstring, this quietly triples the sensitivity, and re-tuning belongs in the same PR that moved the goalposts.

jac/jaclang/scale/runtime/rpc/rpc.jac
```diff
@@ -5,9 +5,44 @@ import from .breaker {
...
+            }
+            if attempt < _RPC_MAX_ATTEMPTS - 1 {
+                time.sleep(backoff);
+                backoff = backoff * 2.0;
+            }
+        }
+    }
+
+    raise RuntimeError(
+        f"sv-to-sv {label} failed after {attempts} attempts: {last_err}"
```
MusabMahmoodh: When the breaker cuts the loop short this reads "failed after 1 attempts", which says neither that the breaker stopped it nor why the count is not 3 - and the first thing anyone does with that message is go looking for the other two attempts.

```suggestion
        f"sv-to-sv {label} failed after {attempts} attempts"
            + (" (circuit breaker opened)" if breaker_should_skip(module_name) else "")
            + f": {last_err}"
```

MusabMahmoodh (commented):
Read this against #7909 and walked the two call sites it merges. The diagnosis goes past the symptom: the breaker was bolted outside two hand-rolled retry loops, and collapsing them into `_rpc_with_retries` makes the fix come out at net -72 lines. Breaking out of the loop once the breaker opens is the part I would have missed - without it, per-attempt recording would refresh `opened_at` on every remaining attempt and stretch the cooldown, and the half-open probe would spend three round trips instead of one.

Checked the three claims that were cheap to get wrong, all hold: `func_label` really is `module_name.func_name`, so the failure message is byte-identical when the loop runs to completion; `breaker_record_success` still fires before envelope decoding on both paths, so an application error envelope still cannot trip the breaker; and `rpc.jac` is the breaker's only consumer, so there is no second accounting site left on the old semantics.

The thing I would like an answer on before merge is the threshold, inline. Worth a sentence in the release note too: with `_RPC_MAX_ATTEMPTS = 3` and the threshold left at 5, two failed sv-to-sv calls now open the breaker for 30s where it used to take about five, per process - during a rolling restart of a downstream service, or at fleet boot when a service calls a peer that is not up yet, callers start shedding load a lot sooner than they did.

jac/jaclang/scale/tests/server/test_sv_auth_forward.jac
```diff
@@ -361,6 +387,96 @@ test "HALF_OPEN probe failure re-opens the breaker with a fresh cooldown" {
...
+        sv_client.register(mod, "http://demo.invalid");
+        try {
+            JacScalePlugin.sv_walker_call(mod, "Ping", {}, None);
+        } except RuntimeError {
+            ;
+        }
+
+        st = _breaker_peek(mod);
+        assert st is not None;
+        assert round_trips["n"] > 1;
```
MusabMahmoodh: This is the assertion carrying the walker half of the fix, and `> 1` passes on two round trips as happily as three - it would not catch a regression that dropped an attempt. The number is known here.

```suggestion
        assert round_trips["n"] == _RPC_MAX_ATTEMPTS;
```

jac/jaclang/scale/tests/server/test_sv_auth_forward.jac
```diff
@@ -361,6 +387,96 @@ test "HALF_OPEN probe failure re-opens the breaker with a fresh cooldown" {
...
+
+test "breaker opens after exactly the trip threshold of failed round trips" {
+    mod = "breaker_trip_roundtrips_test";
+    _breaker_reset(mod);
+    round_trips: dict = {"n": 0};
+    original_post = _install_unreachable_transport(round_trips);
+
+    try {
+        sv_client.register(mod, "http://demo.invalid");
+        while not breaker_should_skip(mod) {
```
MusabMahmoodh: If a regression stops the breaker opening, this spins forever and the job dies on the runner timeout with no useful output, instead of failing on the assert below. Bounding it keeps the failure readable:

```suggestion
        guard = 0;
        while not breaker_should_skip(mod) and guard < 50 {
            guard += 1;
```

jac/jaclang/scale/tests/server/test_sv_auth_forward.jac
```diff
@@ -361,6 +387,96 @@ test "HALF_OPEN probe failure re-opens the breaker with a fresh cooldown" {
...
+        st = _breaker_peek(mod);
+        assert st is not None and st["state"] == "OPEN";
+    } finally {
+        sv_http.post_stream_aware_sync = original_post;
+        sv_client.unregister(mod);
+        _breaker_reset(mod);
+    }
+}
+
 test "application error envelope does NOT trip the breaker" {
```
MusabMahmoodh, line 487: Stepping back from these three tests: the breaker has no coverage outside this file - `breaker` appears in exactly one test file in the repo, and nothing exercises it against a real cluster.

`k8s_microservice_real_e2e.sh` does cross the breaker-protected hop (`orders_app` sv-imports from `cart_app`, so `create_order` is a real `service_call`), but only on the happy path, so the only breaker function that ever runs there is `breaker_record_success`. The failure it does inject - `kubectl delete pod` on orders-app at line 300 - is the gateway-to-orders hop, which is a proxy forward with no breaker, and it waits for `kubectl rollout status` and then curls with `--retry 6 --retry-all-errors`, so nothing ever observes a failed round trip. `test_runtime_integration.jac` uses `InProcessTransport` and never enters `service_call` at all.

The failing case is ~12 lines in the existing e2e and it is the only thing that proves the accounting against a live cluster:

```bash
kubectl scale -n "${NAMESPACE}" deploy/cart-app-deployment --replicas=0
kubectl rollout status -n "${NAMESPACE}" deploy/cart-app-deployment --timeout=60s
# two create_order calls = 6 round trips, enough to open the breaker
curl -sS -X POST "${GW_URL}/orders/function/create_order" -H "${AUTH}" -d '{}' >/dev/null
curl -sS -X POST "${GW_URL}/orders/function/create_order" -H "${AUTH}" -d '{}' >/dev/null
START=$(date +%s)
OUT=$(curl -sS -X POST "${GW_URL}/orders/function/create_order" -H "${AUTH}" -d '{}')
[ $(( $(date +%s) - START )) -lt 5 ] || { echo "FAIL: breaker did not shed load" >&2; exit 1; }
printf '%s' "${OUT}" | grep -q "open circuit breaker" || { echo "FAIL: no breaker message" >&2; exit 1; }
kubectl scale -n "${NAMESPACE}" deploy/cart-app-deployment --replicas=1
```

jac/jaclang/scale/runtime/rpc/rpc.jac
akindu-k: Yes, deliberate, and you are right that it deserves to be stated rather than left implicit.

I went back through the history on this. The constant came in with 121fc6757 (#5483), and the docstring shipped alongside it said:

> Per-provider circuit breaker **wraps the retry loop**. After `_BREAKER_TRIP_THRESHOLD` consecutive **transport-failure cycles** the breaker opens...

So the original placement outside the loop was not an accident. Counting cycles was the design, and 5 meant 5 cycles, which is roughly 15 round trips. #7909 reframed that same design as the bug. That makes this a deliberate policy change rather than a code-versus-intent correction, which is exactly the distinction you are pointing at. That docstring no longer exists in `breaker.jac`, the `strip-docstrings` deslop rule removed it, so the only surviving description of the unit was in the issue.

On why I would keep 5 rather than retune to 15, the thing that makes it safer than it first looks is that the threshold counts *consecutive* failures. `breaker_record_success` resets the whole record to `{"state": "CLOSED", "fails": 0, "opened_at": 0.0}`, so a single successful round trip wipes the counter.

That changes the rolling restart case. The breaker keys on the service, not the pod, and during a rolling restart most replicas are still serving, so most round trips succeed and keep resetting the counter to zero. Reaching 5 consecutive failures requires essentially every round trip to fail, which means the dependency is fully unavailable rather than partially degraded. That is the case where shedding load fast is the behaviour we want.

Fleet boot is the same shape. If the peer is not up yet then every round trip fails, and opening after 5 beats grinding through 15 with a 10s timeout on each. The real cost in that scenario is not the trip, it is the fixed 30s cooldown, which can keep shunning a peer that came up 5 seconds later. That is a genuine weakness, but it is the cooldown constant and the absence of any fast recovery path, and it is unchanged by this PR.

Retuning to 15 would keep the accounting fix while giving up the protection the issue was filed to get, and it would leave a magic 15 that silently depends on `_RPC_MAX_ATTEMPTS` staying at 3.

I have added the sentence to the release note so the sensitivity change is user visible:

> This makes callers shed load from an unreachable dependency sooner: five consecutive failed round trips now open the breaker for the cooldown period, and any successful round trip resets the count to zero.

If you would rather this be tunable per dependency instead of a fixed default, `_sv_rpc_timeout` already reads per-service values out of the scale config, so a `breaker_trip_threshold` could ride the same path. Happy to add it here or file it separately, whichever you prefer.

jac/jaclang/scale/runtime/rpc/rpc.jac
akindu-k: Agreed on the problem. "failed after 1 attempts" makes the reader go hunting for two attempts that never happened, and the message should say why the count is short.

I went with a local flag rather than the re-query, because `breaker_should_skip` is not a pure read. It takes the lock and, on line 20 of `breaker.jac`, flips `OPEN` to `HALF_OPEN`:

```jac
if st.get("state") == "OPEN" {
    elapsed = time.time() - st.get("opened_at", 0.0);
    if elapsed < _BREAKER_COOLDOWN_SECONDS {
        return True;
    }

    st["state"] = "HALF_OPEN";
    return False;
}
```

So calling it to build the string can consume the probe slot as a side effect of formatting an error. It can also disagree with the reason we actually stopped: if the cooldown elapses between the `break` and the `raise`, the re-query returns False and the message drops the annotation in exactly the case where the breaker did open.

We already know why we left the loop at the point we leave it, so I recorded it there:

```jac
if breaker_should_skip(module_name) {
    tripped = True;
    break;
}
...
stopped_by = " (stopped by open circuit breaker)" if tripped else "";
raise RuntimeError(
    f"sv-to-sv {label} failed after {attempts} attempts{stopped_by}: {last_err}"
);
```

Same information, no second lock acquisition, no state mutation, and it cannot go stale. When the loop runs to completion `stopped_by` is empty, so the message stays byte-identical to the old text and the existing "3 attempts" assertion still holds.

I also pinned the new wording in the mid-call test rather than adding another test:

```jac
assert "1 attempts (stopped by open circuit breaker)" in raised_msg;
```

jac/jaclang/scale/tests/server/test_sv_auth_forward.jac
akindu-k: Taken, thanks. `> 1` would have passed on two round trips as happily as three, so it was not actually guarding the thing the test exists to guard.

Used the constant rather than a literal so it tracks if `_RPC_MAX_ATTEMPTS` ever moves:

```jac
import from jaclang.scale.runtime.rpc.rpc { _RPC_MAX_ATTEMPTS }
...
assert round_trips["n"] == _RPC_MAX_ATTEMPTS;
```

jac/jaclang/scale/tests/server/test_sv_auth_forward.jac
akindu-k: Good catch, a hang with no output is the worst way for this to fail.

I bounded it with `_BREAKER_TRIP_THRESHOLD` instead of a literal 50, because that is the real ceiling: even in the degenerate case of one failure recorded per call, the breaker cannot need more than `_BREAKER_TRIP_THRESHOLD` calls to open. So the guard can never cut a correct run short, and it doubles as documentation of why the bound is what it is.

```jac
calls = 0;
while not breaker_should_skip(mod) and calls < _BREAKER_TRIP_THRESHOLD {
    calls += 1;
```

The old buggy behaviour needed exactly 5 calls, so it still reaches the assert below and fails with the informative "opened after 15 round trips, expected 5" rather than timing out.

jac/jaclang/scale/tests/server/test_sv_auth_forward.jac
akindu-k: Agreed on the gap, and your reading of the existing e2e matches what I found. `create_order` does call `view_cart()` on `cart_app` through `sv import`, so that hop is a real `service_call`, and the `kubectl delete pod` at line 300 is the gateway-to-orders proxy hop with no breaker on it. The `--retry 6 --retry-all-errors` afterwards would mask a failed round trip even if there were one. So today the only breaker function that ever executes against a cluster is `breaker_record_success`.

Added the check. I changed two things from your sketch:

**It drives calls until one trips instead of assuming two.** Breaker state is per process, so it lives in whichever orders-app pod served the request. These services carry managed HPAs, so a second replica would split the failures across two independent breakers and the fixed two-call version would flake. Looping to a bound is robust to replica count and does not re-encode the threshold arithmetic, so it survives the threshold constant changing.

**It restores cart_app before asserting.** The script runs under `set -euo pipefail`, so an `exit 1` between the scale-down and the scale-up would leave `cart_app` at zero replicas for every later section. Scaling back first means a failure here cannot cascade.

```bash
kubectl scale -n "${NAMESPACE}" deploy/cart-app-deployment --replicas=0 >/dev/null
kubectl wait --for=delete pod -n "${NAMESPACE}" -l app=cart-app --timeout=120s >/dev/null 2>&1 || true

while [ "${BREAKER_CALLS}" -lt 10 ]; do
    ...
    if printf '%s' "${CALL_BODY}" | grep -q "open circuit breaker"; then
        BREAKER_BODY="${CALL_BODY}"
        break
    fi
done

# restore, then assert
kubectl scale -n "${NAMESPACE}" deploy/cart-app-deployment --replicas=1 >/dev/null
```

It asserts both halves: that the breaker message appears at all, and that the call carrying it returned in under 5s, which is the part that proves it actually shed load rather than just reporting after doing the work.

One caveat worth stating plainly: I cannot run this locally, so unlike the unit tests it has not been executed. `bash -n` passes and I checked that the message survives to the response body (`transport_to_response` puts it in `error.message`, and the unhandled path in `app.jac` puts it in `error`), but the first real run will be in CI.

MusabMahmoodh (dismissed):
Approving. The local tripped flag is the right call over my re-query suggestion - breaker_should_skip flipping OPEN to HALF_OPEN makes calling it twice wrong, good catch. The e2e step driving calls until the breaker opens instead of assuming a fixed count survives a second orders replica, and restoring cart-app before asserting means a failed check cannot leave the fixture scaled to zero. Release note now states the operational change plainly.

PR #8064, opened 2026-08-12
fix(scale): admit a single half-open probe through the circuit breaker (#7909 - Bug 2)

akindu-k:
Issue #7909 (bug 2). Bug 1, the coarse per-retry-loop failure accounting, is #8029, now merged.

explanation: https://claude.ai/code/artifact/6e5047f4-86e1-4a58-b6e5-c03a3cbf6116

## The breaker, and the state this PR is about

```mermaid
stateDiagram-v2
    direction LR
    [*] --> CLOSED
    CLOSED --> OPEN: 5 consecutive failed round trips
    OPEN --> HALF_OPEN: cooldown elapsed and one caller took the probe
    HALF_OPEN --> CLOSED: the probe holder succeeded
    HALF_OPEN --> OPEN: the probe holder failed, or handed the lease back
```

`HALF_OPEN` is supposed to admit exactly one trial call. It admitted all of them.

## Problem

`breaker_should_skip` flipped `OPEN` to `HALF_OPEN` and returned `False`, then released the lock. Every caller that arrived afterwards saw `HALF_OPEN`, fell through the final `return False`, and was also admitted. The half-open trial was supposed to be one call; instead the full concurrent burst hit a dependency that had just been declared unhealthy, which is the load the breaker exists to shed.

A test that opens the breaker, lets the cooldown elapse, and then fires nine concurrent callers at a transport that blocks measures **17 round trips** on the old semantics, against the 1 the design intends.

```mermaid
flowchart TD
    subgraph before["BEFORE - 9 callers arrive the moment the cooldown elapses"]
        direction LR
        A1["caller 1"] --> G1{{"breaker_should_skip"}}
        A2["callers 2-9"] --> G1
        G1 -- "flips OPEN to HALF_OPEN, returns False" --> S1[("service B, just declared unhealthy")]
        G1 -- "sees HALF_OPEN, falls through the final return False" --> S1
    end
    subgraph after["AFTER - the same 9 callers"]
        direction LR
        B1["caller 1"] --> G2{{"breaker_admit"}}
        B2["callers 2-9"] --> G2
        G2 -- "leases probe 7, admits" --> S2[("service B")]
        G2 -- "shed, never touches the network" --> X2(["RuntimeError, fails fast"])
    end
    before ~~~ after
    classDef bad fill:#f7e2e0,stroke:#a8433a,color:#3d1512
    classDef good fill:#dff0e8,stroke:#2c7a5c,color:#123528
    classDef probe fill:#f8ecd6,stroke:#a9700a,color:#3b2705
    class S1 bad
    class X2 good
    class S2,G2 probe
```

**17 round trips against the 1 the design intends.**

## Root cause

Not the missing counter. `breaker_should_skip` is named as a query and used as one, but it performed a state transition, and the transition it performed had no notion of who owned the trial. That conflation is what makes the bug unfixable in place:

- There is nowhere to put "one probe is already in flight", because the function that would record it is also the function every caller polls.
- #8029's retry helper calls `breaker_should_skip` mid-loop to ask "did the breaker just open on me?". Under the old semantics that question mutates state, and any probe-token scheme bolted onto the same function would have that inner poll silently acquiring or stealing probes.

So the fix is to separate the question from the decision. Three functions, each answering one question for one audience, and only one of them mutates:

- **`breaker_should_skip(provider)`** is a pure predicate: is this provider shedding load right now? That is the question a new arrival asks at the door. It never transitions, so it is safe to call twice.
- **`breaker_admit(provider, probe_ttl)`** is the single mutating gate. It admits one caller and, when the breaker is recovering, leases the half-open probe to exactly that caller. Everyone else is shed.
- **`breaker_still_admitted(provider, probe)`** is a pure predicate for a caller already inside the retry loop: is the breaker still letting *me* through? The breaker is shedding nobody, or you are the one it is letting through.

Admission then falls out as the mirror of shedding: a non-`CLOSED` caller that is not being shed *is* the probe holder, by construction, under one lock acquisition.

## The failure mode a naive token would introduce

A probe lease that is never returned is strictly worse than the bug being fixed: the breaker sits in `HALF_OPEN` forever and sheds 100% of traffic to a dependency that may be perfectly healthy. Three things close that off:

1. **The lease is released in `finally`.** #8029 collapsed both call sites into one retry helper, which is what makes a single guaranteed release path possible. An exception the retry loop does not classify as a transport failure (a `ValueError` out of `resp.json()` on a malformed 200, say) unwinds through the release instead of leaking the lease. The release restamps the cooldown, which is deliberate and asserted: releasing to the old `opened_at` would hand the next caller an immediate probe, so a peer that answers but breaks every probe would be probed as fast as callers arrive. Restamping holds it to one probe per cooldown.
2. **A probe outstanding past its lease deadline is reclaimed** by the next caller. This is the backstop for a wedged probe holder that never reaches its `finally`. The deadline comes from the caller, `rpc_timeout + _BREAKER_PROBE_GRACE_SECONDS` (30s), not from a fixed constant: `rpc_timeout` is per service and the schema docs set it to `120.0` for `llm_app`, so a fixed 60s window would expire under a live probe on such a service and admit a second one alongside it. One round trip is the whole budget to size against, because a half-open probe never retries: its first failure reopens the breaker, and the `breaker_still_admitted` on the next line breaks the loop before any backoff.
3. **Verdicts carry the lease they belong to.** In `HALF_OPEN` only the lease holder's failure re-opens the breaker. Without this, a straggler admitted before the trip, still retrying when the cooldown elapsed, could re-open the breaker under an unrelated probe and cancel a trial it was never part of.

### The lease deadline, drawn to scale

```text
                       0s          60s         120s        150s
  the call may run     |===========================|                rpc_timeout = 120.0 (llm_app)
  a fixed 60s window   |============|                               <-- expires under a live probe,
                                                                        a second probe is admitted
  rpc_timeout + 30s    |==================================|         always outlives the call
```

### What the backstop actually does

```mermaid
sequenceDiagram
    participant A as caller A
    participant BR as breaker
    participant B as caller B
    A->>BR: breaker_admit
    BR-->>A: admitted, probe 7, lease until rpc_timeout + 30s
    Note over BR: HALF_OPEN, sheds every other arrival
    B->>BR: breaker_admit
    BR-->>B: not admitted, probe 0
    Note over A: wedged, never reaches its finally
    Note over BR: lease deadline passes
    B->>BR: breaker_admit
    BR-->>B: admitted, probe 8
    A->>BR: breaker_record_failure with probe 7
    Note over BR: verdict is not current, live probe is 8 - dropped
```

## Where each gate sits

Each gate sits where its question is asked:

- `breaker_should_skip` stays at the top of `service_call` and `walker_call`, ahead of URL resolution and header building. A shed call still costs nothing, and an unregistered service with an open breaker still reports the breaker instead of "not registered". Being a pure predicate, it takes no lease and mutates nothing, so having it there cannot interfere with a probe.
- `breaker_admit` lives in the shared retry helper, because that is the only scope that can release the lease on every exit path.
- `breaker_still_admitted` sits between retries, next to `breaker_record_failure`, because that is the only point where a caller's own standing can have changed since it was admitted.

```mermaid
flowchart TD
    call["service_call / walker_call"] --> gate1{"breaker_should_skip<br/>pure predicate, at the door"}
    gate1 -- "shedding" --> shed["raise _breaker_shed_error<br/>one message, both raise sites"]
    gate1 -- "not shedding" --> resolve["resolve URL, build headers,<br/>size rpc_timeout"]
    resolve --> gate2{"breaker_admit<br/>the only mutator, in _rpc_with_retries"}
    gate2 -- "not admitted" --> shed
    gate2 -- "admitted, carrying probe" --> attempt["send - one round trip"]
    attempt -- "transport failure" --> rec2["breaker_record_failure with probe"]
    rec2 --> gate3{"breaker_still_admitted<br/>pure predicate, between retries"}
    gate3 -- "still mine" --> backoff["backoff, next attempt"]
    backoff --> attempt
    gate3 -- "no longer mine" --> stop["break - stopped by open circuit breaker"]
    attempt -- "success" --> rec1["breaker_record_success with probe"]
    rec1 --> fin["finally: breaker_release_probe"]
    stop --> fin
    classDef mut fill:#f8ecd6,stroke:#a9700a,color:#3b2705
    classDef pure fill:#e7ecef,stroke:#5a666b,color:#1a2226
    class gate2,fin mut
    class gate1,gate3 pure
```

One mutator, two pure predicates, and a single release path that every exit unwinds through.

Both raise sites build the message through `_breaker_shed_error`, so it is composed from the existing `label` in one place and stays identical on both paths.

## A caller admitted before the trip

The retry loop has to re-check the breaker between attempts, because the breaker can trip while a call is on the wire. It used to ask `breaker_should_skip`, which is the door's question, and that is the wrong question for a caller already in flight.

The two diverge in exactly two states, both of which are moments when the breaker is *not* shedding precisely because it is about to hand one probe to the next arrival:

- `OPEN` with the cooldown elapsed and no lease taken yet.
- `HALF_OPEN` with a lease past its deadline that nobody has reclaimed.

In both, a caller admitted while the breaker was still `CLOSED` holds no lease, so its verdicts are already dropped by `_breaker_verdict_is_current`, but nothing stopped it retrying. Its next attempt goes to the wire beside the recovery probe: load on a dependency that has just been declared unhealthy, with no ability to open, close, or count against the breaker. On a service configured `rpc_timeout = 120.0` a single attempt can span the whole 30s cooldown, so this needs no contrived timing to reach.

`breaker_still_admitted` is what the loop asks now, and it breaks with `shed = True` when the answer is no. Two rows change; the rest is pinned by tests that already passed:

| state at the retry check | before | after |
| --- | --- | --- |
| `CLOSED` | retry | retry |
| `OPEN`, still cooling down | break | break |
| **`OPEN`, cooldown elapsed, no lease** | **retry** | **break** |
| `HALF_OPEN`, live probe, not the holder | break | break |
| **`HALF_OPEN`, expired lease, not the holder** | **retry** | **break** |
| `HALF_OPEN`, the holder | unreachable, its own failure reopens first | same |
| `CLOSED`, retired probe holder | retry | retry |

The obvious alternative, having `breaker_record_failure` report whether the verdict counted, is not equivalent. It diverges in one state: a retired probe holder in a breaker that has since closed. Its verdict is not current, since `_breaker_verdict_is_current` requires `probe == _BREAKER_NO_PROBE` when `CLOSED`, but the breaker is shedding nobody there, so breaking would abandon a caller against a healthy dependency and blame a closed breaker for it.

## The shed error

`_breaker_shed_error` used to quote `(cooldown 30s)`. Once the probe lease is sized from the caller, that number is in force on only one of the two paths the message covers: a `HALF_OPEN` shed on a service at `rpc_timeout = 120.0` runs up to 150s. A constant that is not the window actually being served is worse than no number, since it sends whoever is debugging somewhere else with confidence. The message now names the shed without quoting a window, and `_BREAKER_COOLDOWN_SECONDS` is no longer imported into `rpc.jac`. The substring the tests assert on is unchanged.

## A success reported while the breaker is OPEN

It is dropped, where previously it closed the breaker outright. That caller was admitted while the breaker was still `CLOSED`, so its result describes the dependency before the trip. Letting it close the breaker means that under partial failure one overlapping success cancels every trip and the breaker can never stay open. Recovery is still bounded by the cooldown, and the first caller after it probes immediately; the old fast path only fired when a straggler happened to be in flight. Called out in the release note.

```mermaid
sequenceDiagram
    participant X as caller X
    participant BR as breaker
    participant Y as callers Y1-Y5
    X->>BR: admitted while CLOSED, probe 0
    Note over X: on the wire, describing the dependency before the trip
    Y->>BR: 5 failed round trips
    Note over BR: OPEN, cooldown running
    X->>BR: breaker_record_success with probe 0
    Note over BR: state is OPEN, verdict is not current - dropped
```

## Tests

Added, plus the existing half-open tests updated to take a lease the way a real caller does:

**One probe at a time**
- Only one caller is admitted while a probe is in flight, and the second is told to skip.
- Nine concurrent callers against a blocked transport produce exactly one round trip; the other eight are shed. Fails at 17 round trips without the fix.

**The lease always comes back**
- A probe that dies without a verdict leaves the breaker `OPEN` on a fresh cooldown and admits the next probe, rather than wedging it shut or handing out a free one.
- A probe that never reports back is reclaimed at its deadline, and its late verdict cannot clobber the live probe.
- A lease is reclaimed at the deadline its caller set, and the replacement sheds the caller behind it.
- A real call configured with `rpc_timeout = 120.0` runs under that timeout and holds a lease that outlives it. Fails against a fixed 60s window.

**Only current verdicts count**
- A stale probe's success neither closes the breaker around a live probe nor erases current failures, and its failure does not count toward the trip threshold.
- A success from before the breaker opened does not cancel the cooldown.

**Shedding happens at the door**
- An unregistered service with an open breaker reports the breaker, not "not registered", on both `service_call` and `walker_call`.

**A caller stops once the breaker stops admitting it**
- A caller admitted before the trip spends one round trip instead of retrying into the recovery window, and says why it stopped short. Reads 3 round trips without the fix.
- Admission survives only in `CLOSED` or while holding the live probe, walked across every state in the table above, including an expired lease and a retired holder in a closed breaker.

The core invariants were mutation-checked: restoring the old half-open fallthrough fails the single-probe tests, and neutering the lease release fails the reclaim test. The `rpc_timeout = 120` test asserts a lease deadline past 120s, which the old fixed 60s window cannot produce. Restoring `breaker_should_skip` in the retry loop fails the straggler test at 3 round trips.

Suite is `jac test jac/jaclang/scale/tests/server/test_sv_auth_forward.jac`, with `test_sv_streaming.jac` and `microservices/test_runtime_integration.jac` as the other users of this path. CI has not run them on this revision yet: `build-jac` is currently failing on `main` in the toolchain step (`NATIVE SEAL: error[E1053]` out of `native_materialize.jac`), before any test executes.

## Note on scope

The issue also notes that breaker state is per process and that no metrics are emitted for transitions. Both are real and neither is part of this defect; they are left alone, as in #8029, and both are now filed so they do not close with #7909: #8420 for the per-process state, where the one-probe-per-cooldown guarantee this PR establishes becomes one per replica, and #8421 for the missing transition metrics.

The breaker tests are now the majority of `test_sv_auth_forward.jac`, at 22 of 42 tests. Moving them into their own file is #8311, kept out of this PR because a 660-line pure move would bury the fix in the diff.


jac/jaclang/scale/runtime/rpc/breaker.jac
```diff
@@ -1,52 +1,137 @@
 import from threading { Lock as _BreakerLock }
 glob _BREAKER_TRIP_THRESHOLD: int = 5,
      _BREAKER_COOLDOWN_SECONDS: float = 30.0,
+     _BREAKER_PROBE_TIMEOUT_SECONDS: float = 60.0,
```
MusabMahmoodh: `_BREAKER_PROBE_TIMEOUT_SECONDS = 60.0` assumes a probe finishes within the default 10s `rpc_timeout`, but `rpc_timeout` is per-service config and the schema docs' own example sets `rpc_timeout = 120.0` for `llm_app` (plugin_config.jac:250). For such a service a live probe loses its lease at 60s, a second real probe is admitted alongside it, and the first probe's verdict is then discarded as stale - the exact double-probing this PR removes. Could the lease TTL be derived from the caller instead, e.g.:

```
(admitted, probe) = breaker_admit(module_name, ttl=rpc_timeout + _BREAKER_PROBE_GRACE_SECONDS);
```

with the deadline stored on the lease, falling back to 60s when no timeout is known?

jac/jaclang/scale/runtime/rpc/breaker.jac
```diff
@@ -1,52 +1,137 @@
...
-def breaker_record_success(provider: str) {
+def breaker_record_success(provider: str, probe: int = _BREAKER_NO_PROBE) {
     with _breaker_lock {
-        _breaker_state[provider] = {"state": "CLOSED", "fails": 0, "opened_at": 0.0};
+        st = _breaker_state.get(provider);
+        if st is None {
+            _breaker_state[provider] = _breaker_new_state();
+            return;
+        }
+        if _breaker_verdict_is_current(st, probe) {
```
MusabMahmoodh, line 120: Before this PR `breaker_record_success` closed the breaker unconditionally; now a success reported while OPEN is dropped by `_breaker_verdict_is_current`, so an in-flight call that proves the service healthy one second after the trip no longer short-circuits the 30s cooldown - everyone gets shed until a probe runs. Ignoring stale verdicts around a live probe is clearly right, but while OPEN there is no probe to protect, and this is a recovery-latency regression for running deployments. Was dropping the OPEN-state success a deliberate trade (the result predates the trip so it may be stale evidence), or just fallout from the generation check? Worth a sentence in the release note either way.

jac/jaclang/scale/runtime/rpc/rpc.jac
```diff
@@ -12,30 +14,42 @@ def _rpc_with_retries(module_name: str, label: str, send: any) -> any {
     import time;
     import from jaclang.runtimelib.serving.client { ClientConnectionError }
 
+    (admitted, probe) = breaker_admit(module_name);
```
MusabMahmoodh: Moving admission into `_rpc_with_retries` puts it after URL resolution and header building, so a call shed by an open breaker now needs a resolvable URL first - which is why the fail-fast test had to gain `sv_client.register(mod, ...)`. Probably fine since a tripped breaker implies the service was resolvable recently, but a service `unregister`ed during teardown while its breaker is open will now surface "not registered" instead of the breaker error. Just confirming that contract change is intended.

jac/jaclang/scale/runtime/rpc/breaker.jac
akindu-k: Confirmed. 60s is wrong for any service configured above it, and the `llm_app` example at 120s is exactly that case.

Done the way you suggested: the lease now stores a deadline instead of being compared against a constant, and `breaker_admit` takes the TTL from the caller. `_rpc_with_retries` passes `rpc_timeout + _BREAKER_PROBE_GRACE_SECONDS` (30s), with 60s kept as the fallback for callers that have no timeout in hand. One round trip is the right budget to size against: a half-open probe never retries, since its first failure reopens the breaker and breaks the loop.

Two tests: a lease past its deadline is reclaimed and the next caller is shed, and a real call driven at `rpc_timeout=120` asserts the lease outlives it. The second one fails on the old constant.

jac/jaclang/scale/runtime/rpc/breaker.jac
akindu-k: Deliberate, and pinned by a test ("a success from before the breaker opened does not cancel the cooldown").

The straggler was admitted while the breaker was still CLOSED, so its result describes the dependency before the trip. Letting it close the breaker means that under partial failure, one overlapping success cancels the trip and the breaker can never stay open, which is the flapping the trip exists to stop. Recovery stays bounded by the cooldown and the first caller after it probes immediately, whereas the old fast path only fired when a straggler happened to be in flight, so it was never something a deployment could rely on.

Release note now says it.

jac/jaclang/scale/runtime/rpc/rpc.jac
akindu-k: Not intended, that was fallout from moving the gate.

Restored: `service_call` and `walker_call` call `_breaker_shed_if_open` before resolution and header building. It uses the pure predicate, so it takes no lease, and `_rpc_with_retries` keeps `breaker_admit` as the only mutating gate. Both raise sites build the message through `_breaker_shed_error`, so it stays identical on either path and there is still one copy of it.

The fail-fast test drops the `register` it had to grow, and a new test asserts an unregistered module with an open breaker reports the breaker rather than "not registered", on both `service_call` and `walker_call`.

jac/jaclang/scale/runtime/rpc/rpc.jac
```diff
@@ -1,41 +1,70 @@
...
+
+def _rpc_with_retries(
+    module_name: str, label: str, send: any, rpc_timeout: float
+) -> any {
     import time;
     import from jaclang.runtimelib.serving.client { ClientConnectionError }
 
+    (admitted, probe) = breaker_admit(
+        module_name, rpc_timeout + _BREAKER_PROBE_GRACE_SECONDS
+    );
```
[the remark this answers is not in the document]
akindu-k: A lease holder never gets a second attempt, so the retry budget never multiplies into the lease.

After the first transport failure, `breaker_record_failure(module_name, probe)` sees a current verdict (the caller is the live probe holder) in `HALF_OPEN` and reopens the breaker. The next line, `breaker_should_skip`, then reads `OPEN` with `opened_at` at roughly now, so it sheds and the loop breaks with `tripped = True`, before `time.sleep(backoff)` is ever reached. The lease therefore only has to cover one `send()`, which is bounded by `rpc_timeout`, and the grace sits on top of that.

This is already pinned by "a breaker that trips mid-call abandons the remaining retries", which drives a real half-open probe at an unreachable transport and asserts `round_trips == 1` and `"1 attempts (stopped by open circuit breaker)"` in the error. Under the behavior described here that assert would read 3.

The loop does keep retrying when the caller holds no lease, that is `probe == _BREAKER_NO_PROBE` because the breaker was CLOSED at admit time. There is no single-probe invariant in play there, and those verdicts are gated by `_breaker_verdict_is_current` anyway.

MusabMahmoodh (commented):
Re-reviewed the three follow-ups. All three landed and all three check out.

1. **Fixed 60s TTL vs per-service `rpc_timeout`** - `breaker_admit(provider, probe_ttl)` now stores `probe_expires_at = now + ttl` and `_rpc_with_retries` passes `rpc_timeout + _BREAKER_PROBE_GRACE_SECONDS`. `_breaker_is_shedding` reads the deadline rather than a constant, so nothing else has to know the TTL. Two tests pin it, including the 120s case.
2. **Straggler success dropped while OPEN** - deliberate, pinned by "a success from before the breaker opened does not cancel the cooldown", and now stated in the fragment. The reasoning holds: a CLOSED-era result is evidence about the pre-trip dependency, and letting it close the breaker makes the trip uncancellable-by-luck.
3. **Admission behind URL resolution** - restored. `_breaker_shed_if_open` (pure predicate, takes no lease) is back at the top of both entry points, the message is built once in `_breaker_shed_error`, the fail-fast test dropped the `register` it had grown, and a new test asserts an unregistered module with an open breaker reports the breaker on both paths.

I also re-walked the lease state machine on this head: success closes before the `finally` so the release no-ops; a HALF_OPEN failure opens so the release no-ops; a stale verdict from a retired probe is dropped in every state. And the "lease expires during retries" concern is wrong for the reason you gave - the probe holder's first failure reopens the breaker and `breaker_should_skip` breaks the loop before any backoff, so one lease only ever covers one `send()`.

Separating the query from the command is a real fix rather than a patch - the probe-token scheme could not have been bolted onto the old `breaker_should_skip` without the inner poll stealing probes, and this PR removes that possibility rather than working around it. Nothing dead left behind that I can find, other than the note in comment 2. Can this break something already running: no - the admit path is check-and-claim under the same single lock, the shed message text is preserved on both paths, monolith and test-client calls still bypass the breaker entirely, and the one intended behaviour change is now in the release note.

Two small comments, neither blocking.

jac/jaclang/scale/runtime/rpc/breaker.jac
```diff
@@ -1,52 +1,140 @@
...
+
+def breaker_release_probe(provider: str, probe: int) {
+    import time;
+    if probe == _BREAKER_NO_PROBE {
+        return;
+    }
+    with _breaker_lock {
+        st = _breaker_state.get(provider);
+        if st is not None and _breaker_holds_probe(st, probe) {
+            _breaker_open(st, time.time());
```
MusabMahmoodh, line 108: The release stamps a *fresh* `opened_at`, so a non-transport exception out of `send()` reopens the breaker for a full cooldown for every caller of the module even though the peer answered. That is reachable: `walker_call`'s `send` ends in `resp.json()`, so a 200 with a malformed body raises `ValueError`, skips both `except` arms, and lands here. Repeat it and the module stays shed at one call per 30s while the service is otherwise healthy - and pre-PR that case didn't shed at all.

Fail-closed is the right default and re-stamping does rate-limit the retry, so this may well be the trade you want; I'd just like it to be a decision rather than fallout. If it's deliberate, the abandoned-probe test is the place to say so in one line.

(Separately: `_BREAKER_DEFAULT_PROBE_TTL_SECONDS` is now only reached from tests, since the sole production caller always passes a TTL. Fine as a default for future callers - just flagging it's no longer exercised in production.)

jac/jaclang/scale/tests/server/test_sv_auth_forward.jac
```diff
@@ -484,6 +499,301 @@ test "a breaker that trips mid-call abandons the remaining retries" {
...
+        in_flight["at"] = time.time();
+        raise ClientConnectionError(detail="connection refused");
+    }
+    sv_http.post_stream_aware_sync = fake;
+
+    original_resolver = getattr(plugin_mod, "_sv_rpc_timeout");
+    def fake_resolver(module_name: str) -> float {
+        return slow_timeout;
+    }
+    setattr(plugin_mod, "_sv_rpc_timeout", fake_resolver);
```
MusabMahmoodh: Patching the transport is the established seam in this file, but replacing the *config resolver* is a new kind of stub, and `rpc_timeout` has a real one: `[scale.microservices.services.<name>] rpc_timeout = 120.0` is what `test_gateway.jac` drives it with. Setting the config keeps the assertion - "the lease outlives the configured timeout" - measuring the production path rather than a function this test installed.

Also worth knowing: #8075 stubs the same `_sv_rpc_timeout` in a different shape (direct assignment vs `setattr`). If the config isn't usable in-process, one shared helper in this file beats a third copy landing next month.

jac/jaclang/scale/runtime/rpc/breaker.jac
akindu-k: Reachable as described, and it is the trade I want, so it should not have been implicit. `walker_call`'s `send` ends in `resp.json()`, a 200 with a malformed body raises `ValueError`, and the release stamps a fresh `opened_at`.

Fail-closed is the choice: the alternative, releasing to the old `opened_at`, hands the next caller an immediate probe, so a peer that answers but breaks every probe gets probed as fast as callers arrive, which is the burst this branch exists to stop. Restamping keeps it at one probe per cooldown. Worth noting the blast radius is narrower than the whole module: this only fires while the caller holds a lease, meaning the breaker was already open from real transport failures. A `ValueError` on a healthy CLOSED breaker takes the `probe == _BREAKER_NO_PROBE` early return and sheds nothing.

The abandoned-probe test now asserts `opened_at >= abandoned_at` with that reasoning in the message, so restoring the old stamp fails it.

On the default TTL: dropped it and made `probe_ttl` required. The only production caller always passes one, and a silent 60s fallback is exactly the fixed window this branch removed, so a future caller should have to state its own deadline. Tests pass an explicit `_PROBE_TTL`.

jac/jaclang/scale/tests/server/test_sv_auth_forward.jac
akindu-k: Switched to the config. The file already had the seam I should have used: the K8s test stubs `get_scale_config` with an object exposing `get_microservices_config`, which is what `_sv_rpc_timeout` actually reads.

The lease test now installs `{"services": {mod: {"rpc_timeout": 120.0}}}` and additionally asserts the transport was called with `timeout=120.0`, so `_sv_rpc_timeout` and the lease plumbing are both on the production path instead of one of them being stubbed out.

On the third copy: the stub object plus an `_install_ms_config` / `_restore_ms_config` pair are now shared at the top of the file, and the K8s test uses them, so `_K8sStubMsCfg` and its inline `setattr` are gone. Anything landing later has one seam to reach for.

jac/jaclang/scale/runtime/rpc/rpc.jac
```diff
@@ -1,41 +1,70 @@
...
+                last_err = err_ref;
+                breaker_record_failure(module_name, probe);
+                if breaker_should_skip(module_name) {
+                    tripped = True;
+                    break;
+                }
+                if attempt < _RPC_MAX_ATTEMPTS - 1 {
+                    time.sleep(backoff);
+                    backoff = backoff * 2.0;
+                }
```
MusabMahmoodh: A caller admitted while CLOSED that is still mid-retry when the cooldown elapses can send its next attempt straight into the recovery window: this inner `breaker_should_skip` returns False in the OPEN-past-cooldown gap, so attempt 2/3 goes to the wire with no lease, racing the probe - and its verdict is dropped either way, since it holds no probe and the state is non-CLOSED. Concrete case: `llm_app` at `rpc_timeout = 120` has attempt 1 hang across the whole 30s cooldown, then retries alongside the probe. Worth breaking out once the caller's verdicts can no longer count - e.g. `breaker_record_failure` returning whether the failure was current - or was staying admitted-until-shed deliberate?

jac/jaclang/scale/runtime/rpc/rpc.jac
```diff
@@ -1,41 +1,70 @@
...
 }
 
 glob _RPC_MAX_ATTEMPTS: int = 3,
      _RPC_INITIAL_BACKOFF_SECONDS: float = 0.1;
 
-def _rpc_with_retries(module_name: str, label: str, send: any) -> any {
+def _breaker_shed_error(label: str) -> RuntimeError {
+    return RuntimeError(
+        f"sv-to-sv {label} blocked by open circuit breaker "
+            + f"(cooldown {int(_BREAKER_COOLDOWN_SECONDS)}s)"
```
MusabMahmoodh: This message now covers two different sheds, and for the HALF_OPEN one the "cooldown 30s" claim is wrong - a probe window on a `rpc_timeout = 120` service sheds for up to 150s, and whoever debugs it will go looking for a 30s cooldown that isn't the cause. This keeps the "open circuit breaker" substring the tests assert on (and the `_BREAKER_COOLDOWN_SECONDS` import can go with it):

```suggestion
        f"sv-to-sv {label} blocked by open circuit breaker shedding load"
```

jac/jaclang/scale/tests/server/test_sv_auth_forward.jac
```diff
@@ -484,6 +522,309 @@ test "a breaker that trips mid-call abandons the remaining retries" {
     }
 }
 
+test "only one caller is admitted while a HALF_OPEN probe is in flight" {
```
MusabMahmoodh, line 625: After this PR the majority of this file is breaker tests. Fine to land here, but a follow-up that moves them to a test_sv_breaker.jac would spare the next person grepping for auth tests. Not blocking.

MusabMahmoodh (approved):
The mechanism is right: pure predicate split from a single mutating admit gate, lease released in `finally`, caller-sized reclaim deadline, verdict-currency checks. I traced the success / failure / unclassified-exception / stale-probe / concurrent-admit paths and found no hole, and the deletions (`_K8sStubMsCfg`, the duplicated shed messages) came with it. Three minor comments below, none blocking.

One caveat: CI hasn't executed the suite on this revision (build-jac is red on main in the toolchain step), so worth a re-run before merge once main is green.

jac/jaclang/scale/runtime/rpc/rpc.jac
akindu-k: Confirmed, and it was not deliberate. Traced it: a caller admitted at `CLOSED` holds `probe == _BREAKER_NO_PROBE`, its `send()` raises after the trip, `breaker_record_failure` drops the verdict on the `OPEN` branch of `_breaker_verdict_is_current`, and then `_breaker_is_shedding` reads `(now - opened_at) < _BREAKER_COOLDOWN_SECONDS` as false and lets the loop run on. Attempt 2 goes to the wire with no lease, exactly as you describe.

There is a second door into the same state, which the same fix closes: `HALF_OPEN` with a lease past its deadline that nobody has reclaimed yet. `_breaker_is_shedding` checks `now < probe_expires_at` there, which is also false, so a straggler retries into a live probe window too.

The root cause is that the inner check was asking the door's question. `breaker_should_skip` answers "is this provider shedding load right now?", which is right for a new arrival with no history. A caller already in the loop needs "am I still allowed on the wire?". They agree everywhere except the two states above, both of which are moments when the breaker is not shedding precisely because it is about to hand one probe to the next arrival, and the straggler cannot be that arrival: taking a lease from inside the loop is the probe-stealing this branch exists to prevent.

I did not key the break on verdict currency, though. "Verdict counted" and "still admitted" diverge in one state: a retired probe holder in a breaker that has since closed. Its verdict is not current, since `_breaker_verdict_is_current` requires `probe == _BREAKER_NO_PROBE` when `CLOSED`, but the breaker is shedding nobody at that point, so breaking there would abandon a caller against a healthy dependency and report "stopped by open circuit breaker" while the breaker is closed.

So the loop asks a third pure predicate instead, mirroring `breaker_admit`:

```jac
def breaker_still_admitted(provider: str, probe: int) -> bool {
    with _breaker_lock {
        st = _breaker_state.get(provider);
        if st is None or st.get("state") == "CLOSED" {
            return True;
        }
        return _breaker_holds_probe(st, probe);
    }
}
```

The breaker is shedding nobody, or you are the one it is letting through. Same lock, and it reuses the `_breaker_holds_probe` that the verdict check uses, so admission and verdict currency cannot drift apart in `HALF_OPEN`. `tripped` became `shed`, since the flag now means the breaker stopped admitting this caller rather than that it tripped during the call.

Two rows of the behaviour table move; the rest is pinned by tests that already passed:

| state at the retry check | before | after |
| --- | --- | --- |
| `CLOSED` | retry | retry |
| `OPEN`, still cooling down | break | break |
| **`OPEN`, cooldown elapsed, no lease** | **retry** | **break** |
| `HALF_OPEN`, live probe, not the holder | break | break |
| **`HALF_OPEN`, expired lease, not the holder** | **retry** | **break** |
| `HALF_OPEN`, the holder | unreachable, its own failure reopens first | same |
| `CLOSED`, retired probe holder | retry | retry |

Tests: "a caller admitted before the trip stops instead of retrying into recovery" drives your case through `sv_service_call`, with a transport that trips the breaker and elapses the cooldown on its first round trip, then asserts one round trip and `1 attempts (stopped by open circuit breaker)`. It reads 3 round trips without the fix. "admission survives only in CLOSED or while holding the live probe" walks the predicate through every row above.

jac/jaclang/scale/runtime/rpc/rpc.jac
akindu-k: Confirmed, and the constant was wrong from the moment this branch added the second shed. `_breaker_is_shedding` returns true in `HALF_OPEN` while `now < probe_expires_at`, and `probe_expires_at` is now `now + rpc_timeout + _BREAKER_PROBE_GRACE_SECONDS`, so the message quoted a number that is only in force on one of the two paths it covers. A constant that is not the window actually being served is worse than no number at all, since it sends the reader somewhere else with confidence.

Took your text. `_BREAKER_COOLDOWN_SECONDS` goes out of rpc.jac with it, it was imported only to build that string. Both raise sites still compose through `_breaker_shed_error`, so the two paths stay identical and the substring the tests assert on is unchanged.

I did consider reporting the real remaining window instead, which is the message I would want while debugging. It needs `breaker_should_skip` to return a duration rather than a bool, and re-opening the pure predicate's signature for the sake of a log string is not a trade I want on this branch. That belongs with the breaker metrics the issue lists as out of scope.

jac/jaclang/scale/tests/server/test_sv_auth_forward.jac
akindu-k: Numbers agree: after this branch, 22 of 42 tests and roughly 660 of 1156 lines in the file are breaker tests.

Keeping it out of this PR. It is a pure move, so it is not risky, but a 660-line diff would bury a 30-line fix, retire the approval, and restart the bot review from scratch. Folding a mechanical refactor into a bug fix also makes the fix harder to revert on its own if it ever needs to be.

Filed as #8311. It also records the one part that is not mechanical: `_install_unreachable_transport`, `_install_unreachable_walker_transport`, `_StubMsCfg`, `_install_ms_config` and `_restore_ms_config` are used by both halves of the file, so they need a shared home rather than a second copy, which is the same point you made about the `_sv_rpc_timeout` stub last round.

MusabMahmoodh (approved):
Re-review at head. I approved on 08-17; the substantive commit after that is the straggler fix, so this pass is about that.

`breaker_still_admitted` checks out, and your reason for not taking the "`breaker_record_failure` reports whether the verdict counted" variant I suggested is right: a retired probe holder in a since-CLOSED breaker has `_breaker_verdict_is_current` false but is being shed by nobody, so that version would abandon a caller against a healthy dependency and blame a closed breaker for it. The predicate form gets that row right. I walked the other rows of your table against the code as well: OPEN with no lease and HALF_OPEN with an expired lease both break, the live holder cannot reach the check because its own failure reopens the breaker first, and a reclaimed holder's stale verdict is dropped in every state. Dropping the cooldown number from the shed message took `_BREAKER_COOLDOWN_SECONDS` out of `rpc.jac` with it.

Re-affirming the approval. Two notes, neither about the code.

**1. The suite still has not run in CI on this branch.**

`build-jac` is red again, but not on your code this time: it dies fetching a zig package (`zig-pkg/neovim-.../build.zig.zon:13:20: error: bad HTTP response code: '404 Not Found'`), and every test job is skipped behind it, so the 22 breaker tests have never executed in CI on any revision of this branch. The other scale PRs on the same base are green after merging main on 08-18, so a re-run or a fresh merge from main should turn the matrix on. Worth doing before merge, since this is a concurrency change and a careful read is not a test run.

**2. The two scope notes die with the issue.**

#7909 closes with this PR, and its last paragraph goes with it: breaker state is per process, so each replica learns independently, and no metrics are emitted for transitions. I cannot find a follow-up issue for either, and #8311 only covers the test move. Both are real for a fleet: with N replicas the "one probe per cooldown" guarantee is really N per cooldown, and today the only evidence a breaker is shedding is the RuntimeError text in a caller's traceback. Worth filing them before this merges so they do not vanish with the issue.

PR #8075, opened 2026-08-12
fix(scale): retry sv-to-sv RPC only when the request provably never reached the peer (#7902)

### LINKED ISSUE #7902: Jac Scale: sv-to-sv RPC retries walker and function POSTs on read timeout, re-executing walkers that already ran
**Area:** runtime core / rpc
**Severity:** critical (correctness)

### What happens

`walker_call` and `service_call` both wrap their POST in a 3-attempt retry loop that catches `httpx.TimeoutException` alongside `httpx.RequestError`. A read timeout means the request was sent and may have been fully processed by the peer. Retrying re-executes it.

### Where

`jac/jaclang/scale/runtime/rpc/rpc.jac:169-194` (walker_call)
`jac/jaclang/scale/runtime/rpc/rpc.jac:78-110` (service_call)

### Evidence

```jac
max_attempts: int = 3;
backoff: float = 0.1;
...
for attempt in range(max_attempts) {
    try {
        client = httpx.Client(timeout=rpc_timeout);
        try {
            resp = client.post(
                f"{url}/walker/{walker_name}", json=args, headers=headers
            );
            data = resp.json();
        } finally {
            client.close();
        }
        breaker_record_success(module_name);
        return _hydrate_walker_envelope(data, module_name, walker_name, stub_cls);
    } except (httpx.RequestError, httpx.TimeoutException) as e {
        last_err = e;
        if attempt < max_attempts - 1 {
            time.sleep(backoff);
            backoff = backoff * 2.0;
        }
    }
}
```

### Why it matters

A slow walker that exceeds `rpc_timeout` gets spawned up to three times. Every non-idempotent sv-to-sv call is affected, and the failure mode is silent duplicate work rather than an error.

### Suggested fix

Retry only on connect-phase failures (`httpx.ConnectError`, `httpx.ConnectTimeout`), where it is certain the request never reached the peer. Read timeouts should surface immediately. If retry on read timeout is wanted, it needs an idempotency key or an explicit per-walker opt-in via `restspec`.

The retry policy is also hardcoded (`max_attempts = 3`, `backoff = 0.1`) while the timeout beside it is configurable.

Part of #7900

kashmithnisakya:
Closes #7902 (part of https://github.com/jaseci-labs/jac/issues/7900)

## What happens today

`walker_call` and `service_call` wrap their POST in a 3-attempt retry loop. A peer that accepts the request, executes the walker, and then fails to answer within `rpc_timeout` gets the same bytes replayed up to two more times. Reproduced against a real socket: three identical `POST /walker/...` deliveries in ~2s, three executions, one recorded breaker failure, and a terminal error that ends in `: ` because `str(TimeoutError())` is empty.

## Root cause

The issue as filed blames the retry loop's `except` tuple, but narrowing that tuple changes nothing. The failure is classified a layer lower and the classification is destroyed before rpc.jac can see it:

1. The whole request leaves in one fire-and-forget `writer.write()` (client.jac). From that instant the peer may hold and execute the request.
2. The response-head reads sit in `asyncio.wait_for` with nothing catching the timeout. On Python 3.11+, `asyncio.TimeoutError` is builtin `TimeoutError`, which subclasses `OSError`.
3. `arequest`'s catch-all (`except (ConnectionError, OSError, IncompleteReadError)`) spans write, drain, and read-head, and re-raises everything as a plain `ClientConnectionError(detail=str(e))`, i.e. the exact type the retry arm treats as "safe to retry", with an empty message.

The audit found the filed head-timeout is one of three replay paths: the non-streaming body read was completely unwrapped (a mid-body reset escaped as a raw `OSError` and was retried), and streaming HTTP 500s were surfaced as generic connection errors and retried too.

## The fix

Retry-safety becomes a property of the exception type, encoded where the information exists:

- `client.jac` gains three empty-body subclasses of the existing `ClientConnectionError`: `ClientRequestNotSent` (provably never left the process), `ClientRequestSent` (peer may have executed), and `ClientReadTimeout(ClientRequestSent)`. None of them subclass `OSError`, since inheriting retry semantics from `TimeoutError -> OSError` is the mechanism that caused this bug.
- Every raise site in the request path is classified by lifecycle phase: connect failures are `ClientRequestNotSent`; everything after the first write (head timeout, bad status line, mid-body loss) is `ClientRequestSent`. Read timeouts get a synthesized, non-empty message.
- `post_stream_aware_sync` carries the exception object across its pump-thread handshake instead of a string, so the type survives into the sv function-call half; a wedged pump (`queue.Empty`) and streaming HTTP >= 400 are typed as `ClientRequestSent` as well.
- Side effect worth calling out: `_pump` now puts the error ahead of the sentinel, so a stream that dies mid-flight raises out of the consumer's iterator instead of ending cleanly. On `main` that case truncated silently despite `docs.md` promising a `RuntimeError`. Covered by the mid-stream RST test.
- rpc.jac splits the single catch-all arm in `_rpc_with_retries`: `except ClientRequestNotSent` keeps the existing retry-with-backoff (rolling-deploy resilience is unchanged), and a following `except (ClientConnectionError, OSError)` records the error and breaks without replaying. `OSError` stays in the second arm as a fail-closed backstop.

The invariant: only a failure the client can prove happened before the first `writer.write()` is ever replayed. A future raise site that forgets to classify gets the ambiguous parent type, which is not retried, so contributor error degrades to "no retry", never to "replay a walker".

Because the new types subclass `ClientConnectionError`, all existing catch sites (gateway `raw_forward`/`stream_forward`, ws forwarding, admin Loki/Tempo helpers) keep working with zero edits, verified case by case. This was the deciding argument against a sibling exception type, which would have turned an admin-UI read timeout into an uncaught 500.

## Tests

- Three new socket-level tests in `test_sv_auth_forward.jac` assert the literal invariant in the issue title: a deaf peer and a 500-returning peer each receive exactly one POST. All three fail on `main` (the deaf-peer tests observe 3 accepts).
- New `jac/tests/runtimelib/test_serving_client_errors.jac` is the first direct test file for the serving client: taxonomy invariants (including `not issubclass(ClientRequestSent, OSError)`) and end-to-end classification against real loopback sockets, including type preservation across the streaming thread boundary.
- The two existing retry fixtures simulated "connection refused" while raising the ambiguous parent type; they now raise `ClientRequestNotSent`, which is what connection refused is. Their assertions (2 attempts, then 3 attempts + "3 attempts" in the message) are unchanged.
- Non-regression runs: `test_sv_auth_forward.jac` 29 passed, `test_stream_ws.jac` 14 passed, `test_sso.jac` 51 passed, `test_serving_client_errors.jac` 7 passed. `test_gateway.jac` has 3 pre-existing metrics failures that reproduce identically on `main`.

## Notes for review

- **Overlap with #7909 (#8029, #8064):** #8029 has since merged and the arm split is folded into its `_rpc_with_retries` helper, so `attempts` is the real per-round-trip count and the fixtures assert `"after 1 attempts"`. #8064 is still open and rewrites the same helper: it wraps the loop in `try/finally` for the probe lease and adds an `rpc_timeout` parameter. Whichever lands second must preserve the other's structure; in particular the new fail-fast arm has to keep falling through the `finally` that releases the probe.
- **Breaker timing side effect:** an unsafe failure now records one breaker failure and stops, instead of three across an exhausted retry loop, so the counter tracks calls rather than attempts. Measured against a peer that is alive but always exceeds `rpc_timeout` (trip threshold 5):

  ```
  before:  call 1 -> fails=3 CLOSED,  call 2 -> fails=5 OPEN      (opens on call 2)
  after:   call 1 -> fails=1 ...      call 5 -> fails=5 OPEN      (opens on call 5)
  ```

  Five delivered POSTs before the circuit opens either way, but they are now five distinct calls instead of two calls' worth of replays. The direction is less trigger-happy per call, not more: a persistently slow service takes five calls rather than two to shed load. Compounds with #8064's half-open rule.
- **Slow-to-first-token endpoints:** an SSE/LLM endpoint that needs longer than `rpc_timeout` to produce its first byte previously sometimes succeeded on attempt 2 or 3, at the cost of 2-3 executions. It now fails fast; `services.<name>.rpc_timeout` is the intended knob and the docs section updated in this PR already documents it.
- The retry-policy statement in `runtime/docs.md` already described this behaviour as the intent for streaming RPC ("connect-time failures still retry"); this PR generalizes that statement to all sv-to-sv RPC and removes three stale references to the removed httpx transport.
- Pre-existing and deliberately untouched, now each tracked so the class has a home once this closes #7902:
  - #8429: `aiter_raw` has no read timeouts, so a peer that answers the head and then stalls hangs the caller forever. `rpc_timeout` bounds connect and the response head only, which the docs section in this PR now states outright.
  - #8430: the breaker counts HTTP status errors, records a success on an error envelope that arrives with a 2xx, and never hears about a stream that dies mid-flight.
  - #8431: the retry-on-timeout opt-in (restspec) and configurable `rpc_max_attempts` / `rpc_backoff` that #7902 also asked for, so idempotent read-style calls keep a path back to timeout resilience.

  Still unfiled and left alone: `resp.json()` decode errors escape both arms uncaught (skipping the breaker), the head timeout is per-readline rather than a wall-clock deadline, and `aws_connect`'s write/upgrade path is unclassified (idempotent GET, no caller retries it).


jac/jaclang/scale/runtime/rpc/rpc.jac
```diff
@@ -80,13 +81,17 @@ def service_call(module_name: str, func_name: str, args: dict) -> any {
...
             err_ref: any = e;
             last_err = err_ref;
             if attempt < max_attempts - 1 {
                 time.sleep(backoff);
                 backoff = backoff * 2.0;
             }
+        } except (ClientConnectionError, OSError) as e {
+            sent_ref: any = e;
+            last_err = sent_ref;
+            break;
```
MusabMahmoodh: After this break the terminal error at L100 still says `failed after 3 attempts`, but exactly one POST was delivered - the same deliveries-vs-attempts ambiguity this PR is removing. `attempt` survives the loop and the exhausted NotSent path still yields "3 attempts", so the existing fixture assertion keeps passing:

```
f"sv-to-sv RPC '{module_name}.{func_name}' failed after "
    + f"{attempt + 1} attempts: {last_err}"
```

Same in walker_call at L188.

jac/jaclang/scale/runtime/rpc/rpc.jac
```diff
@@ -80,13 +81,17 @@ def service_call(module_name: str, func_name: str, args: dict) -> any {
...
             return _unwrap_sv_envelope(payload, module_name, func_name);
-        } except (ClientConnectionError, OSError) as e {
+        } except ClientRequestNotSent as e {
             err_ref: any = e;
             last_err = err_ref;
             if attempt < max_attempts - 1 {
                 time.sleep(backoff);
                 backoff = backoff * 2.0;
             }
+        } except (ClientConnectionError, OSError) as e {
```
MusabMahmoodh: A streaming HTTP 4xx lands in this arm now and records a breaker failure with no retry - one caller with an expired token can 401 the circuit open for every well-authed caller of the module. Not a regression (main recorded it too, after 3 wasted replays), but since the classification now lives here: should a status error count against the breaker at all, given the peer demonstrably responded? Fine as a #7900 follow-up. Related: the rewrap in `post_stream_aware_sync` flattens the status into the detail string; `ClientConnectionError` already has a `status_code` field, and carrying it through would let this arm tell a 4xx from transport loss.

jac/jaclang/runtimelib/serving/client.jac
```diff
@@ -403,6 +437,9 @@ def post_stream_aware_sync(
                     events.put(item);
                 }
             }
+        } except Exception as e {
+            events.put(('error', repr(e)));
```
MusabMahmoodh, line 445: This does more than carry the type across the thread: on main, a mid-stream failure hit the `finally` sentinel before `_runner` put the error into `events`, so the consumer's generator returned cleanly and the stream truncated silently - despite docs.md promising a RuntimeError out of the iterator. The new ordering (error, then sentinel) actually delivers that promise. Worth a test so it doesn't regress: a peer that sends an SSE head plus one event and then RSTs, asserting the consumer's `for` loop raises instead of ending cleanly.

jac/jaclang/runtimelib/serving/client.jac
```diff
@@ -236,13 +252,22 @@ async def arequest(
...
+        raise;
     } except (ConnectionError, OSError, asyncio.IncompleteReadError) as e {
         try {
             writer.close();
         } except Exception {
             ;
         }
-        raise ClientConnectionError(detail=str(e));
+        raise ClientRequestSent(
+            detail=f"connection lost after request was sent: {e!r}"
```
MusabMahmoodh: Typing a write/drain failure as sent is the right fail-closed call, but at drain time the bytes may never have left the buffer, so "after request was sent" overclaims. "connection lost during or after send" keeps the incident text honest without weakening the classification.

MusabMahmoodh (commented):
The invariant and the altitude both check out - classification lives in the client where the phase information exists, policy stays in rpc.jac, and an unclassified future raise site degrades to no-retry rather than a replay. I verified the catch-site audit case by case (gateway http_forward, ws_forward, admin logs/traces, sso), plus the jac0core fallback, which has no retry loop, so no sibling defect is left behind. Four comments; only the attempts-count message is worth changing before merge.

jac/jaclang/scale/runtime/rpc/rpc.jac
kashmithnisakya: fixed, folded into main's _rpc_with_retries so the error reports the real attempt count

jac/jaclang/scale/runtime/rpc/rpc.jac
kashmithnisakya: status_code now carried through as a typed ClientRequestSent; breaker policy left for #7900

jac/jaclang/runtimelib/serving/client.jac
kashmithnisakya: added the mid-stream RST test, the consumer's for loop now must raise

jac/jaclang/runtimelib/serving/client.jac
kashmithnisakya: reworded to "connection lost during or after send"

jac/jaclang/scale/tests/server/test_sv_auth_forward.jac
```diff
@@ -748,3 +748,149 @@ test "ensure_sv_service early-returns when peer already registered" {
...
+    rpc_mod._sv_rpc_timeout = _fast_rpc_timeout;
+    try {
+        sv_client.register(mod, f"http://127.0.0.1:{port}");
+        raised_msg = "";
+        try {
+            JacScalePlugin.sv_walker_call(mod, "Ping", {}, None);
+        } except RuntimeError as e {
+            raised_msg = str(e);
+        }
+        assert counts["n"] == 1;
```
MusabMahmoodh: `rpc_mod._sv_rpc_timeout = _fast_rpc_timeout` shortens both connect and read, so a loaded runner that takes longer than the shortened window to connect on loopback classifies as `ClientRequestNotSent`, retries, and fails this count - which is what the "widen deaf-peer timeouts" commit was chasing. Widening lowers the probability rather than removing it. The `"no response head within"` assertion does distinguish the two outcomes, so putting it first makes a flake report its real cause:

```jac
        assert "no response head within" in raised_msg , raised_msg;
        assert counts["n"] == 1;
```

jac/jaclang/scale/tests/server/test_sv_auth_forward.jac
```diff
@@ -748,3 +748,149 @@ test "ensure_sv_service early-returns when peer already registered" {
...
+}
+
+
+test "walker_call sends exactly one POST to a peer that never replies" {
+    counts: dict = {"n": 0};
+    (srv, held, port) = _start_counting_server(counts, None);
+    mod = "deaf_walker_mod";
+    _breaker_reset(mod);
+    original_timeout = rpc_mod._sv_rpc_timeout;
+    rpc_mod._sv_rpc_timeout = _fast_rpc_timeout;
```
MusabMahmoodh: Small one: #8064 stubs this same `_sv_rpc_timeout` in a different shape (`setattr` there, direct assignment here), and it's repeated three times in this file. If the config seam (`[scale.microservices.services.<name>] rpc_timeout`) isn't usable in-process, one shared `_with_rpc_timeout(mod, seconds)` helper beats a fourth copy landing next month, and gives the next person one place to change when the resolver moves.

MusabMahmoodh (commented):
Re-reviewed after your four fixes. All four landed.

1. **`failed after 3 attempts` when one POST was delivered** - resolved by the merge with main; `attempts` is the real per-round-trip count from #8029, and the three socket tests assert `"after 1 attempts"`.
2. **Streaming 4xx recording a breaker failure** - the status is now carried as a typed `ClientRequestSent(detail=..., status_code=...)` (the field already exists on `ClientConnectionError`, client.jac:20), so the breaker-policy question stays answerable later without another classification pass. Deferring the policy to #7900 is the right call.
3. **Error-before-sentinel ordering** - test added; `_pump`'s new `except Exception { events.put(('error', repr(e))); raise; }` puts the error ahead of the sentinel and `_runner` now only adds the sentinel, so the consumer's `for` loop raises instead of truncating, which is what docs.md already promised.
4. **"after request was sent" overclaiming at drain time** - reworded.

One improvement you didn't mention: moving `ssl_lib.create_default_context()` inside the try in `_open_stream` means a missing CA bundle now classifies as `ClientRequestNotSent` and retries, where before it escaped raw.

This is a proper fix rather than a patch - retry-safety becomes a property of the exception type, encoded at the raise site where the phase is known, and an unclassified future raise site degrades to no-retry rather than to a replay. The catch-site audit still holds: every new type subclasses `ClientConnectionError`, so gateway forwards, ws forwarding and the admin Loki/Tempo helpers are unchanged. No dead code left behind - the `OSError` in the second arm is a deliberate backstop, not a leftover. Can this break something already running: no; the only behaviour change is fewer retries, in exactly the cases where a retry was re-executing a walker, and connect-phase failures keep the old retry-with-backoff.

The three socket-level tests assert the literal invariant from the issue title - one POST on the wire to a peer that never answers, measured at the socket - which is the right level for this bug.

**Merge-order note:** this and #8064 both rewrite `_rpc_with_retries` and will conflict. #8064 wraps the loop in `try/finally` for the probe lease and adds an `rpc_timeout` parameter; this one splits the `except`. Whichever lands second needs the other's structure preserved - the fail-fast arm must still fall through the `finally` that releases the probe.

Approving. Two test notes below.

jac/jaclang/scale/tests/server/test_sv_auth_forward.jac
kashmithnisakya: done, and took it to the root in 930d73579: the counting server now tells round trips apart from delivered POSTs, the message assert comes first and every assert carries raised_msg, so a misclassified connect neither trips the count nor hides its cause

jac/jaclang/scale/tests/server/test_sv_auth_forward.jac
kashmithnisakya: done in 930d73579, one `_install_rpc_timeout(mod, seconds)` / `_restore_rpc_timeout` seam and all four stubs in the file go through it (8064 goes through `_install_ms_config` rather than the resolver, so post-merge that stays one shape per layer)

kashmithnisakya:
ack on the merge order with #8064: trial-merged c687a588c locally and only `_rpc_with_retries` conflicts (the test file auto-merges); the split except drops straight inside 8064's try/finally so the fail-fast break still releases the probe. happy to rebase whichever of us lands second

MusabMahmoodh (approved):
Re-reviewed at `5f3c2a20e`. Approving.

I went back to the question I raised on #7900, which was whether the retry axis is "evidence about the provider" or "evidence about *this request*", since a 503 is a response but is not evidence the request was rejected before it executed. Walking it on this head, the predicate is right.

The only retried arm is `rpc.jac:29`, `except ClientRequestNotSent`. No status can reach it: HTTP status errors come back as `ClientRequestSent(status_code=...)` (`client.jac:417-420`), read timeouts as `ClientReadTimeout` (`client.jac:175`, `:189`), which subclasses `ClientRequestSent` (`client.jac:32`), and mid-body loss at `client.jac:268`/`:292`. Those all land on the `ClientConnectionError, OSError` arm at `rpc.jac:41`, which records one breaker failure and breaks without retrying. I checked the inverse too: every `ClientRequestNotSent` raise (`client.jac:165`, `:167`, `:221`) happens before the `writer.write` at `client.jac:252`. So the retry set is a strict subset of "no evidence this ran", which is the conservative side of the line.

Both of my 08-15 points are resolved, and the first is better than what I asked for: `930d73579` counts accepted connections separately from delivered POSTs and cross-checks rpc's attempt count against observed connections, which pins the "never sent" claim to something the test can actually see.

Two non-blocking follow-ups.

**`rpc.jac:44`** - `gh issue list --search breaker` comes back empty, so the deferral we agreed on in #7900 has nowhere to live once this merges. Could you file it before merge and link it from this line? The scope is settled between us: split `breaker_record_failure` out of this arm and decide it on `status_code` (absent or 5xx counts, 4xx and 429 do not), and fix the mirror at `breaker.jac:27-31` where a status error returning normally resets `fails` to 0.

**`rpc.jac:118`** - pre-existing, and I think it belongs in that same follow-up rather than here: on the SSE return, `_rpc_with_retries` has already called `breaker_record_success` (`rpc.jac:27`) by the time the head arrives, and the generator is consumed afterwards, so a stream that dies mid-flight raises at `client.jac:494` without the breaker ever hearing about it. A provider whose streams always truncate looks permanently healthy. Worth one line in the issue so the policy covers both call shapes.

jac/tests/runtimelib/test_serving_client_errors.jac
```diff
@@ -0,0 +1,253 @@
...
+def _sse_one_event_then_reset(conn: any) {
+    import time;
+    conn.recv(65536);
+    conn.sendall(
+        b"HTTP/1.1 200 OK\r\n"
+            + b"Content-Type: text/event-stream\r\n"
+            + b"Connection: close\r\n\r\n"
+            + b"data: {\"n\": 1}\n\n"
+    );
+    time.sleep(0.5);
```
MusabMahmoodh: The 0.5s sleep before the RST is an ordering assumption: if the pump thread hasn't pulled the event bytes off the socket yet, the LINGER-0 RST discards them from the receive buffer and `got == [{"n": 1}]` fails on a loaded CI runner. Consuming the first event before letting the server reset makes it deterministic:

```jac
go = threading.Event();
# handler: conn.sendall(<event>); go.wait(5.0); then linger + close
(kind, gen) = post_stream_aware_sync(...);
it = iter(gen);
got.append(next(it));  # proves the client parsed event 1
go.set();              # now the server may reset
```

jac/jaclang/runtimelib/serving/client.jac
```diff
@@ -256,8 +279,17 @@ async def arequest(
...
+        async for block in response.aiter_raw() {
+            collected.append(block);
+        }
+    } except (ConnectionError, OSError, asyncio.IncompleteReadError) as e {
+        try {
+            writer.close();
+        } except Exception {
+            ;
+        }
+        raise ClientRequestSent(detail=f"connection lost mid-body: {e!r}");
```
MusabMahmoodh, line 292: Now that a stalled head fails fast instead of being replayed, the one remaining hang path is the body: `aiter_raw`'s reads have no `wait_for`, so a peer that returns the head and then stalls mid-body hangs `walker_call` forever (`rpc_timeout` covers connect + head only, as the updated docs say). Pre-existing and rightly out of scope here - worth filing as its own issue so the class has a home?

jac/jaclang/scale/runtime/rpc/rpc.jac
```diff
@@ -35,6 +38,11 @@ def _rpc_with_retries(module_name: str, label: str, send: any) -> any {
...
```
MusabMahmoodh, line 41: #7902 also asked for an idempotency-key / per-walker restspec opt-in for retry-on-timeout, and flagged the hardcoded max_attempts/backoff; this PR (correctly) ships only the safety half. Is there a follow-up tracking the opt-in + configurable policy? Without it, idempotent read-style sv functions have no path back to timeout resilience during rolling deploys.

MusabMahmoodh (commented):
The fail-closed shape of the taxonomy is the right invariant: an unclassified future raise site degrades to "no retry", never to "replay a walker". Keeping the new types under `ClientConnectionError` so every existing catch site works unmodified was the right call over a sibling type - I checked the gateway forwards, ws_forward, the admin Loki/Tempo helpers, and the jac0core fallback path, all fine as-is.

jac/tests/runtimelib/test_serving_client_errors.jac
kashmithnisakya: done in 8e47ee23a, the handler now waits on an event the test sets after next(it) has returned event 1, so the reset can't race the pump thread

jac/jaclang/runtimelib/serving/client.jac
kashmithnisakya: filed as #8429, and in 8e47ee23a the docs paragraph that still promised a per-read timeout between events now says the body is unbounded and points there

jac/jaclang/scale/runtime/rpc/rpc.jac
kashmithnisakya: filed as #8431 (restspec opt-in plus rpc_max_attempts / rpc_backoff next to rpc_timeout), so the second half of 7902 keeps a home after this closes it

jac/jaclang/scale/runtime/rpc/rpc.jac
```diff
@@ -35,6 +38,11 @@ def _rpc_with_retries(module_name: str, label: str, send: any) -> any {
...
+            sent_ref: any = e;
+            last_err = sent_ref;
+            breaker_record_failure(module_name);
```
kashmithnisakya, line 44: breaker policy for status errors is now #8430: split breaker_record_failure out of this arm and decide on status_code, fix the success-on-500-envelope mirror in walker_call / breaker.jac, and record the SSE outcome after the head rather than at it (covers the note on the sse return too)

MusabMahmoodh (approved):
Re-read the new commit. Approval stands.

Replacing `time.sleep(0.5)` with `_sse_reset_gate.wait(5.0)`, set only after the consumer has parsed event 1, turns the one genuinely racy assertion in this PR into a deterministic one. That test was asserting "the stream yielded exactly `[{"n": 1}]` before the RST", which is a claim about ordering that a sleep can only make probabilistically, and on a loaded CI runner the losing side of that race is a confusing failure in a file about error classification. Setting the gate in the `finally` too, so the handler thread never sits out its full 5s on an early failure, is the part people usually forget.

One note, not an ask: `_sse_reset_gate` is a module-level global, so it is shared state between tests in this file. It is safe as written because the test clears it on entry and sets it in `finally`, and these run serially - but if this file ever runs under parallel execution, or a second test reuses `_sse_one_event_then_reset`, that global is what will bite. Passing the event into the handler through a closure would make each test own its own gate and cost about two lines. Fine to leave.

Docs change reads correctly: `rpc_timeout` stopping at the response head, not covering the body, is the distinction that makes the SSE case explicable.

PR #8080, opened 2026-08-12
feat(checker): require type arguments on dict, list and tuple

### LINKED ISSUE #8082: Track: feat(checker): require type arguments on dict, list and tuple (#8080)
Tracking issue for #8080.

## Summary

Bare `dict`, `list` and `tuple` are now a hard error (`E1036`) rather than a warning (previously `W1036`). The bare-generic fallback in `build_type_var_solution` now resolves to `Unknown` instead of `any`, so values read out of an unparameterized container are flagged rather than silently accepted.

```jac
d: dict = src();
return d["a"]["b"];        # E1036 on the annotation, Unknown through the reads

d: dict[str, any] = src();
return d["a"]["b"];        # fine, and says what it means
```

Scoped to `dict`, `list`, and `tuple` via `BARE_GENERIC_REQUIRED`; `set`, `frozenset`, `Queue`, and other unparameterized generic classes are untouched. `_check_bare_generic` now also descends into type arguments, so nested cases like `list[dict]` are caught.

The PR migrates ~395 annotation sites across ~140 files to satisfy the new check.

## Known follow-ups noted in the PR

- `.jacignore`d files (e.g. `na_ir_gen_pass.jac`) still contain bare generics of their own; they're invisible to `jac check` and left for follow-up work.
- `jac fmt` refuses files that contain errors, so the checker fixtures that must contain bare generics by design can't be formatted (CI's format step already excludes `/fixtures/` and is `continue-on-error`, but worth tracking).

See PR for full details: https://github.com/jaseci-labs/jac/pull/8080

SandeepaHWP:
Bare `dict`, `list` and `tuple` are now a hard error rather than a warning.

## What changed

A container type written without its element types used to draw `W1036` and, for type-checking purposes, behave as if `any` had been written, so `dict` and `dict[str, any]` read differently but checked the same, and the checker handed `any` to the native backend where a real type would have produced better code.

The two spellings are not identical everywhere, though. Under `--memory nogc` the ownership checker tracks the elements of a written `list[any]` but not those of a bare `list`, so `l: list; return l[1];` compiles today and `l: list[any]; return l[1];` is `E1317` (move out of an element). That is pre-existing and `nogc`-only, `rc` and `managed` are unaffected, but it means naming the element types can surface an ownership error at sites that read an owned element out of a container. Move it out with `pop()` or borrow it with `&l[...]`. Three native fixtures needed exactly that change here.

Writing `dict`, `list` or `tuple` bare in an annotation is now `E1036`:

```jac
d: dict = src();           # E1036: say what the dict holds
d: dict[str, int] = src(); # fine
d: dict[str, any] = src(); # fine, and says the values are mixed
```

The check also looks inside type arguments, so `list[dict]` and `dict[str, list]` are reported. Before, only the outer container was checked, and the inner bare one went unreported.

## What did not change

Type inference is unchanged. The error is about annotations the user writes; the checker still treats a container it builds on its own as holding `any` when it has nothing to infer from, exactly as before:

```jac
def f(x: object) {
    if isinstance(x, list) {   # narrows to a bare list, no annotation involved
        for e in x { ... }     # e is any, as on main
    }
}
```

An earlier revision of this branch made those internally created containers resolve to `Unknown`. That punished code with no annotation to fix (`isinstance` narrowing, `list(...)` calls with nothing to infer from) and forced typed-local workarounds across the migration. It is reverted, and so are the workarounds it required; every such line is back to what `main` has.

## Scope

The rule covers `dict`, `list` and `tuple`, listed in one shared `BARE_GENERIC_REQUIRED`. `set`, `frozenset` and generic classes used without type arguments (such as `tempfile.TemporaryDirectory()`, or a user-defined `class Box[T]`) keep the existing `W1036` advisory rather than the error, so nothing loses coverage.

## The migration

Every bare container annotation in the repo is typed, about 1,060 files. The rule applied:

- one knowable element type gets that type: `list[float]` for flattened 4x4 matrices, `dict[int, int]` for keycode state, `dict[str, SemTokManager]`, `tuple[bool, set[tuple[str, str]]]`
- genuinely mixed values (JSON payloads, LLM request params, config tables) get `any`, as in `dict[str, any]`, rather than a long union

Two situations needed more than a substitution:

- **Cross-file cascades.** Files that were never edited went red because a bare container in a different file, including a `.jacignore`d one, leaked into them.
- **Inference quirks.** `[text] if text else []` inferred `list[LiteralString]` and rejected a plain `str`; an explicit `if` fixed it.

## The native build

`.jacignore` does not protect the native build of the compiler. It compiles the compiler's own sources into the binary, so `E1036` there is fatal even in files `jac check` skips, and it stops at the first broken module. That surfaced about 217 sites across about 25 files in `jac0core`.

One of those needed judgement rather than an annotation:

- `unitree.jac`'s `py_ast_targets` and the `py_ast` field behind it are `list[any]`: neither module imports `ast3`, and every consumer already casts to `list[ast3.expr]`, so `any` keeps the exact prior behaviour without adding an import.

## Docs

Nine skill guides taught the old style in `jac`-fenced examples, which `test_guide.jac` requires to pass `jac check`. Fifteen sites, several nested (`list[dict]`) or inside a union (`dict | None`), forms the old warning never reached.

## Tests

`test_checker_pass.jac`:

- `bare_generic_hard_error_at_every_site`: return type, param type, local annotation, and nested inside a parameterised generic
- `bare_generic_iteration_is_a_hard_error`: each bare annotation is an `E1036`, and reads through it still behave as `any`, so only the annotation is reported
- `bare_generic_in_signatures_emits_e1036`: 6 `E1036` sites, plus the 2 `W1036` advisories `set` and `frozenset` still draw
- `explicit_any_propagates_through_operations`: `any` keeps working through chained subscripts, attribute access and iteration
- `written_empty_tuple_arguments_are_not_bare`: `tuple[()]` is the only spelling of the empty tuple, so it is not reported, nested (`list[tuple[()]]`) or in a union (`tuple[()] | None`) as well as at the top level
- `typing.Optional[dict]` is reported: the brackets belong to `Optional`, so they do not excuse the `dict`

`test_unpack_any_element.jac` uses the explicit `tuple[any, any]` / `list[any]` spellings, since the bare forms are now an error; they reach the same element-extraction path the test guards.

`prod b1`-`b5` in `test_checker_production.jac` still pass.

## Gaps closed along the way

`list[any]` is the spelling a bare `list` migrates to, so two native gaps it exposed are fixed on this branch: `repr()` now lowers for `any`-element containers, and the boxed elements of an `any`-valued container are reclaimed under `--gc none`.

## Known limits

CI's `jac check` skips `.jacignore`d files and test directories, so those areas rely on the native build and the test runs. A scan of every `.jacignore`d file finds no bare container annotations left.

`jac fmt` is not idempotent in a single pass on this change: wrapping one signature can push the next over the line limit, so it needs running until it converges.

## Open question for reviewers

Decision 3 in #8323 names `list`, `dict` and `set`. This PR errors on `dict`, `list` and `tuple`, and leaves `set` and `frozenset` on the `W1036` advisory rather than dropping them. Worth settling whether `set` should be promoted to the error too, and whether `tuple` stays.


kugesan1105 (dismissed):
Thanks for this, the migration work is genuinely careful. I checked all 1,063 files mechanically (normalise each changed line, strip container type args, then diff) and **824 of them are provably pure annotation edits** with nothing else hidden in them. The nested detection (`list[dict]`, `dict[str, list]`) is real new coverage, and the two native fixes are correct: `repr` lowering and the refcount drop path both matched only lowercase `"any"` and missed the static key's `"Any"`, so neither was ever reached.

Five things before this can go in.

**1. CI is red on the rule this PR adds.** Two bare `dict` sites came in from main via #8565 and were not migrated: `cli/commands/impl/analysis.impl.jac:1382` and `tests/cli/test_fmt_exit_codes.jac:161`. Both `jac-check` and `test-runtime` fail on these.

**2. W1036 is deleted rather than promoted.** The new `is_builtin(BARE_GENERIC_REQUIRED)` gate means every other generic now reports nothing at all. Verified against main on one probe file: `set`, `frozenset`, `deque` and a user-defined `class Box[T]` all draw W1036 on main and nothing here. The description frames set/frozenset as "out of scope", which reads as "keeps its warning", but the diagnostic no longer exists. Worth deciding explicitly whether that advisory should survive for non-container generics, since it also covers user generics.

**3. `_writes_type_args` tests the spelling, not the type, and that breaks both ways.**

```jac
def a2(x: dict | None) -> None {}      # E1036, correct
def a3(x: Optional[dict]) -> None {}   # silent, same type

glob G2: list[tuple[()]] = [];         # E1036, but this is valid
```

`Optional[dict]` slips through because the annotation node carries an `IndexSlice`, so the union branch inherits the escape hatch. `list[tuple[()]]` is wrongly reported because the type-arg recursion hardcodes `written=False`, so nested positions can never reach the hatch. Both fall out of the same place, so gating on the type being reported rather than on the whole annotation node should fix them together. `checker_empty_tuple_arg.jac` covers top-level `tuple[()]` and a nested bare tuple, but not a nested `tuple[()]`.

**4. Bare `list` and `list[any]` are not interchangeable, which the description assumes.** On main's own compiler, two files differing by one token, under `--memory nogc`:

```jac
l: list = [1, s + "!", 2.5];       return l[1];   # builds, runs, exit 0
l: list[any] = [1, s + "!", 2.5];  return l[1];   # E1317, cannot move out of the element
```

Same for `dict`, and `rc`/`managed` are unaffected. It is pre-existing, not introduced here, but it means the migration converts legal programs into compile errors at those sites. That is what forced the fixture rewrites (`l[1]` to `l.pop(1)`, `len(m[1])` to `len(m)`, and `hetero_nogc`'s expected output from `["1200", "1200", "done"]` to `["600", "400", "done"]`). Could the description say so, since anyone migrating a `nogc` build will hit it?

**5. The release note says `jac run` still executes.** It does not, `jac run` exits 1 without running. The `test-runtime` failure above is the same thing. The `suppress = ["E1036"]` guidance does work.

Smaller things: `bare_generic_iteration` dropped its total-error-count assertion, so extra errors would now pass unnoticed; the tuple case was not carried over into `checker_explicit_any_propagation.jac`; `entry_meta` and `caps` in `execution.impl.jac` became blanket `any` rather than `dict[str, any]` / `list[str]`; and the "repr() has no native lowering" comments in `hetero_matrix.jac` are now contradicted by the repr fix in this same branch. The `jir_registry.jac` section of the description covers files deleted upstream by #8190/#8194 and can be dropped.

Issue #7907, opened 2026-08-05
Jac Scale: the gateway opens a new aiohttp ClientSession per forwarded request, so every proxied call pays full connection setup


MusabMahmoodh:
**Area:** runtime core / transport
**Severity:** high (performance)

### What happens

Both forwarding helpers construct a brand new `aiohttp.ClientSession` per call. There is no shared session, connector or connection pool anywhere in the gateway hot path.

### Where

`jac/jaclang/scale/runtime/transport/http_forward.jac:13` (`raw_forward`)
`jac/jaclang/scale/runtime/transport/http_forward.jac:97` (`stream_forward`)

### Evidence

```jac
async def stream_forward(
    ...
) -> ForwardResult | None {
    session = aiohttp.ClientSession(timeout=aiohttp.ClientTimeout(total=timeout));
```

### Why it matters

Every proxied request pays full TCP setup (and TLS handshake where applicable) plus session construction and teardown. Since the gateway is the single ingress for the whole fleet, this is the first throughput ceiling the system hits, and it also inflates tail latency under load.

### Suggested fix

Create one `ClientSession` with a `TCPConnector` (per-host and total connection limits) on gateway startup, reuse it for all forwards, and close it on shutdown. `stream_forward`'s existing ownership split already separates response cleanup from session lifetime, so the change is contained.

Related: the single `ClientTimeout(total=...)` conflates connect and read timeouts, and `_HOP_BY_HOP` omits several RFC-mandated hop-by-hop headers (`te`, `trailer`, `proxy-authorization`, `proxy-authenticate`, `upgrade`).

Part of #7900

MusabMahmoodh:
Revalidated against main (89391f3d2). aiohttp is gone from the forward path (the #7969 serving rewrite), but the cost this issue is about is unchanged in the replacement: serving/client.jac opens a fresh asyncio.open_connection per call, builds a new TLS context per HTTPS call, and hard-codes Connection: close, so connection reuse is impossible and every proxied request still pays full TCP+TLS setup with no pool anywhere. The single timeout still conflates connect and read phases, and the _HOP_BY_HOP list in http_forward.jac still omits te/trailer/proxy-authorization/proxy-authenticate/upgrade. The mechanism named in the title changed; the defect did not.

marsninja:
Closing: no activity in 10+ days. Reopen if still relevant.

PR #8081, opened 2026-08-12
fix(scale): resolve the owning service before forwarding builtin passthrough bodies (#7901)

### LINKED ISSUE #7901: Jac Scale: gateway builtin passthrough replays the buffered request body to every service until one accepts, so a non-idempotent walker can execute more than once
**Area:** runtime core / gateway
**Severity:** critical (correctness)

### What happens

`handle_builtin_passthrough` buffers the entire request body once, then walks every registered service in registry order and forwards that same body to each in turn. Any service replying 404 or 405 causes a `continue` to the next service.

The loop is not restricted by HTTP method, so a `POST /walker/<name>` is re-sent to service after service until one returns something other than 404/405.

### Where

`jac/jaclang/scale/runtime/gateway/microservice_gateway.impl.jac:362-410`

### Evidence

```jac
body: bytes = await request.body();
query: str = str(request.url.query);

for (svc_name, svc_entry) in self.registry.entries.items() {
    ...
    fr = await stream_forward(
        request.method, target_url, fwd_headers, body, timeout=int(timeout)
    );
    ...
    if fr.status in (404, 405) {
        await fr.cleanup();
        continue;
    }
    return fr.to_response();
}
```

### Why it matters

- A walker that performs side effects and *then* 404s (or whose service 404s after partially handling the request) is executed against more than one service.
- Latency on a miss is O(number of services), sequential.
- The terminal case is a 404 returned only after every service has been contacted.

This is invisible in development and surfaces under production traffic patterns.

### Suggested fix

Resolve the owning service before forwarding, rather than discovering it by trial. The registry already knows which services declare which walkers via the OpenAPI aggregation path. Failing that, restrict the fan-out to safe methods (GET/HEAD/OPTIONS) and return 404 immediately for unsafe methods that do not match a known owner.

Part of #7900

kashmithnisakya:
Closes #7901 (Part of https://github.com/jaseci-labs/jac/issues/7900)

## What happens today

`handle_builtin_passthrough` buffers the request body once, then walks every HEALTHY/STARTING service in registration order and forwards the same bytes to each, treating any 404/405 as "try the next one". There is no HTTP-method guard and no ownership lookup, so a `POST /walker/<name>` is delivered to service after service, with terminal 404 latency of O(services), sequential.

## Root cause

This is an ordering bug, not a missing feature. The gateway already learns exact per-walker ownership: every scale service registers concrete `POST /walker/<name>` / `/function/<name>` / `/webhook/<name>` routes in its openapi, and `ServiceRouteTable` ingests them. But the only consumer of that table (`handle_app_route`) runs *after* the passthrough fan-out in the fallback chain, and passthrough terminates every builtin request, so builtin traffic can neither consult the table nor trigger the lazy refresh that would populate it. A cold gateway fans out forever while holding the information that would let it route.

Validating the issue also sharpened it in both directions:

- **Overstated for scale walkers**: an executed walker/function/webhook callback can only produce 200/401/500 or SSE, never 404/405, and each route is registered only by its owning service, so a 404 in the loop is router-emitted before any handler ran. In a pure scale deployment the walker executes at most once; what replays is byte delivery and latency.
- **Understated for the identity fallthrough**: every service registers the full auth surface, and several handlers execute before returning a semantic 404: `POST /user/verify-identity` attempts the write and emits audit events, then 404s; `PATCH /user/me` runs `update_profile` before its 404. The dispatcher treats the gateway app's 404 as "not handled" and drops the request into the fan-out, so the next service executes the same credential-bearing body again. This is the real multi-execution vector, and it was not in the issue.

## The fix: builtin families, learned per service

Builtin segments are classified as data in `routing.jac`, next to the tables #8006 centralized, and the three families partition `GATEWAY_API_SEGMENTS` (pinned by a test):

- **Replicated** (`GATEWAY_REPLICATED_SEGMENTS`: `/user`, `/sso`, `/api-key`, `/graph`): the same handler runs on every service against the shared identity store, so ownership is meaningless. Delivered exactly once (`deliver_once` -> `forward_to`), never learned. This closes the identity multi-execution vector, and stops the per-crawl clash warnings for the replicated auth surface.
- **Service-owned** (`GATEWAY_SERVICE_OWNED_SEGMENTS`, derived as the remainder: `/walker`, `/function`, `/webhook`, `/ws`): each concrete route is registered by exactly one service, so the gateway resolves the learned owner first, for every HTTP method (`resolve_learned_owner`, shared with `handle_app_route`, both forwarding through one `forward_root_path`), and forwards once. `learn()` only ingests concrete members of these families and app-declared routes (`is_learnable_prefix`), so a templated `/walker/{name}` or `/cl/{page}` cannot capture a family.
- **Discovery** (`GATEWAY_DISCOVERY_SEGMENTS`: `/jobs`, `/cl`): the 404 is a verified pre-side-effect routing signal (`POST /jobs` 404s before scheduling anything, `/cl` pages are GET-only), so the try-each-service probe stays exactly as it was.

The refresh throttle is no longer a single global timestamp. `ServiceRouteTable` records, per service, when it last received a *definitive* spec answer (any HTTP status below 500) at a given url and health status. A miss re-asks only services that were never asked, whose url or status changed, or whose answer is older than `route_refresh_seconds`; the crawl is parallel (`asyncio.gather`, applied in registry order so first-learner-wins stays deterministic) and coalesced across concurrent misses through one shared in-flight task. An unanswered fetch (connection refused, timeout, 5xx) is not knowledge, so a service that was booting on the last crawl is asked again on the very next miss. A newly registered service, a local service flipping STARTING to HEALTHY, or a pod that comes up after the gateway are therefore learned immediately, not after a window.

When no learned service declares a service-owned path, the gateway probes **only the services whose routes are unknown** (never asked, unreachable, or publishing no spec): a learned service's spec is authoritative, so its 404 would be a wasted delivery. Once every eligible service has published its routes an unknown path 404s without the body reaching anybody. A fleet with `docs_enabled = false` (the documented production setting, which removes `/openapi.json`) publishes nothing, so it keeps main's discovery for walkers rather than losing routing.

## Behaviour changes to note

- An owner-routed or replicated request to a dead backend returns `forward_to`'s 502 rather than the probe's 503. The pinned bare-`/jobs` 503 tests still pass unchanged (jobs is a discovery family).
- Safe methods on service-owned families are owner-routed too; the probe is only the fallback for a miss, and then only against services with unknown routes.
- Duplicate walker names across services stop being registration-order roulette: first-learner-wins in registry order, with the existing clash warning.
- A walker added to an already-learned service (redeploy, dev HMR) can 404 for up to `route_refresh_seconds` (default 30s) after its openapi changes, because that service's previous answer is authoritative until it ages out. That is the only residual window; the fan-out that used to cover it is the bug.
- A service that starts answering its spec with a non-200 has its routes unlearned and rejoins the unknown set (with a warning) until it publishes again.

## Tests

Real loopback HTTP backends (draining the request body to `Content-Length` before replying), two services, counting hits per method+path:

- Owner-routed walker POST reaches only the owner; cold gateway learns ownership from openapi before the first forward; a GET walker with a learned owner skips the probe.
- A path no learned service declares is 404'd with zero deliveries; an unsafe miss probes only the services whose routes are unknown; a spec-less fleet keeps walker discovery.
- Identity fallthrough POST delivered exactly once; replicated auth families delivered once and never learned.
- A service registered after the first crawl is learned on the next miss (the learned sibling is not re-asked); a service unreachable at the first crawl is learned once it answers; a walker added to an already-learned service waits for the refresh window (pins the residual contract).
- Stale learned owner is forgotten and re-resolved; `learn()` ignores every non-owned builtin family and bare owned roots from templated paths.
- `POST /jobs` and `GET /cl/<page>` cross-service discovery preserved.
- `test_routing_guard`: the families partition the API segments; only concrete service-owned members are learnable.
- Pinned suites stay green: bare-`/jobs` 503 trio, `test_app_routes` (8), `test_routing_guard` (14, the `is_gateway_owned` assert is gone with the function), `test_dispatch` (6), `test_embedded_gateway` (5), `test_openapi_agg` (3), `test_microservices_registry` (14), `test_stream_ws` (14). The 3 metrics tests in `test_gateway.jac` fail identically on `main` in this environment (pre-existing, prometheus-related).

## Notes for review

- #7907/#7908 target `http_forward.jac` and the body-buffering lines; this PR touches neither, so it is sequence-agnostic with them.
- Follow-ups deliberately not here: send-aware discovery on top of the #8075 exception taxonomy (continue the probe only past provably-never-sent forwards, honest 502 for sent-phase failures), target-aware `POST /jobs` routing, a real `ROUTE_PASSTHROUGH` dispatcher branch, and the `/ws` openapi ownership gap.


jac/jaclang/scale/runtime/gateway/microservice_gateway.impl.jac
```diff
@@ -373,6 +376,52 @@ impl MicroserviceGateway.handle_builtin_passthrough(
...
+            target_url: str = f"{self.service_base_url(owner)}{path}";
+            if query {
+                target_url = f"{target_url}?{query}";
+            }
+            logger.info(f"Passthrough owner: {request.method} {path} -> {target_url}");
+            return await self.forward_to(request, owner, target_url);
+        }
+    }
+
+    return await self.fanout_discovery(request, path, False);
```
MusabMahmoodh: This is where the 30s window bites: with no learned owner we deliver to the first registered service and hand back its 404. Two different cases are sharing one fallback - `/user` (never learnable, so single-shot *is* the delivery mechanism) and `/walker|/function|/webhook` (learnable, just not learned yet). Can we split them, and let a new registration invalidate the throttle? Smallest version is on the registry side:

```jac
# service_registry.impl.jac, in register()
self._routes_refreshed_at = None;   # a new service must be learnable immediately
```

Otherwise the issue's own suggestion for the learnable prefixes - 404 without delivering to a non-owner at all - is better than delivering to the wrong one and 404ing anyway.

jac/jaclang/scale/runtime/gateway/microservice_gateway.impl.jac
```diff
@@ -373,6 +376,52 @@ impl MicroserviceGateway.handle_builtin_passthrough(
...
+}
+
+impl MicroserviceGateway.resolve_learned_owner(path: str) -> ServiceEntry | None {
+    svc_name: str = self.route_table.owner(path);
+    if not svc_name {
+        if not await self.refresh_app_routes() {
```
MusabMahmoodh, line 416: This puts `refresh_app_routes` -> `learn_service_routes` on the builtin request path for the first time, and that loop fetches `/openapi.json` from every service sequentially at 5s each (openapi_agg.jac:25), inline in the request. On a cold gateway with one wedged service the first `POST /walker/...` now blocks for the whole chain where main forwarded immediately. Was that considered? A miss that kicks the refresh off in the background and falls through this once would keep it off the hot path.

jac/jaclang/scale/runtime/gateway/microservice_gateway.impl.jac
```diff
@@ -373,6 +376,52 @@ impl MicroserviceGateway.handle_builtin_passthrough(
...
+    first_seg: str = stripped.split("/")[0] if stripped else "";
+    if first_seg in _DISCOVERY_SEGMENTS or request.method in _SAFE_METHODS {
+        return await self.fanout_discovery(request, path, True);
+    }
+
+    if not is_gateway_owned(path) {
+        owner = await self.resolve_learned_owner(path);
+        if owner is not None {
+            query: str = str(request.url.query);
+            target_url: str = f"{self.service_base_url(owner)}{path}";
```
MusabMahmoodh: This tail (query, target_url, log, `forward_to`) is a verbatim copy of `handle_app_route`'s, and the two have already drifted: this one uses `self.service_base_url(owner)`, which prefers `JacRuntime.get_sv_registry()`, while `fanout_discovery` uses `svc_entry.url` directly - so the same service gets addressed at two different hosts depending on which branch handled the request. Since `resolve_learned_owner` was just extracted, worth extracting the forward with it.

MusabMahmoodh (commented):
The ownership-first direction is right, and `resolve_learned_owner` being shared with `handle_app_route` is the reuse I'd have asked for. One thing I'd like settled before this lands, plus four smaller notes.

Can this break something already running: yes, bounded. An unsafe builtin request with no learned owner is now delivered to whichever service is first in registry order and that service's 404 goes back to the client. The refresh is throttled at 30s (`_DEFAULT_ROUTE_REFRESH_SECONDS`, context/util.jac:28), so a service that registers just inside a throttle window has its walkers 404'd by a sibling for up to 30s, where main's fan-out found them. It's a 404, not a double execution - the 404 is router-emitted - but it lands on the gateway hot path during every rolling deploy and scale-up.

Everything else checks out: safe methods, `/jobs` and `/cl` keep the exact old fan-out, `/graph/data` stays ahead of the owner logic, and the `handle_app_route` refactor is behaviour-identical to the code it replaced.

jac/jaclang/scale/runtime/gateway/route_table.impl.jac
```diff
@@ -8,6 +9,9 @@ impl ServiceRouteTable.learn(service_name: str, templates: list[str]) -> list[st
         if not prefix or is_gateway_owned(prefix) {
             continue;
         }
+        if prefix.strip("/") in GATEWAY_API_SEGMENTS {
```
MusabMahmoodh: The guard covers `GATEWAY_API_SEGMENTS` but not `GATEWAY_CLIENT_SEGMENT`, so a core server publishing `/cl/{page}` still captures `/cl` wholesale. Inert today only because `/cl` short-circuits into `fanout_discovery` before the owner lookup, which is what makes it the kind of latent capture the next routing change re-arms:

```jac
        if prefix.strip("/") in set(GATEWAY_API_SEGMENTS) | {GATEWAY_CLIENT_SEGMENT} {
```

Separately, the PR body says `/sso` and `/api-key` are never learnable - `is_gateway_owned` is only `/user` + `/admin` + well-knowns (dispatch.jac:63-72), so those two do get learned and do take the owner branch.

jac/jaclang/scale/tests/microservices/test_gateway.jac
```diff
@@ -1509,3 +1509,313 @@ test "bare /jobs (create + list) is forwarded as a built-in route, not 404'd" {
...
+
+    def serve {
+        while True {
+            try {
+                (conn, _addr) = srv.accept();
+            } except OSError {
+                return;
+            }
+            try {
+                raw = conn.recv(65536);
```
MusabMahmoodh: Real sockets and per-delivery counting is the right call here, and the `hits_a == 0` assertions genuinely fail on main. Two things: the handler replies and closes after a single `recv` without draining the body, so a body that doesn't fit the first segment can hand the forwarder an RST and turn the asserted 404/200 into a 503 - reading to `Content-Length` first makes these deterministic under CI load. And nothing covers the one state that got worse: owner unknown *because the refresh is throttled*. A test that calls `refresh_app_routes()` once, then registers svc_b, then POSTs one of svc_b's walkers, states whatever contract you settle on above.

jac/jaclang/scale/runtime/gateway/microservice_gateway.impl.jac
```diff
@@ -373,6 +376,52 @@ impl MicroserviceGateway.handle_builtin_passthrough(
...
+        self.route_table.forget(svc_name);
+        return None;
+    }
+    return svc;
+}
+
+impl MicroserviceGateway.fanout_discovery(
+    request: Request, path: str, allow_probe: bool
+) -> Response {
     fwd_headers = self.build_forward_headers(request);
```
MusabMahmoodh, line 462: Worth stepping back on this function. "Deliver exactly once" is implemented as "the fan-out loop with `continue` switched off", so one loop now serves two contracts through a boolean. In `allow_probe=False` mode the `for` can only ever run one iteration, which makes the `continue`, the loop, and the terminal `not_found(path, message="Not found on any service")` all dead or wrong in that mode - the message says every service was tried when exactly one was.

A single-shot delivery that picks its target and calls `forward_to` (the same thing the owner branch already does) would say what it means, delete the flag, and let `fanout_discovery` go back to being only the discovery probe. Happy for that to be a follow-up if you'd rather keep this diff tight, but the flag reads as a workaround for reusing the loop rather than as the intended design.

jac/jaclang/scale/runtime/routing.jac
```diff
@@ -41,7 +41,8 @@ def _derive_builtin_prefixes -> tuple {
 glob GATEWAY_BUILTIN_PREFIXES: tuple = _derive_builtin_prefixes(),
      GATEWAY_BARE_API_PATHS: set[str] = {
          f"/{seg}" for seg in GATEWAY_BARE_API_SEGMENTS
-     };
+     },
+     GATEWAY_DISCOVERY_SEGMENTS: tuple = ("jobs", "cl");
```
MusabMahmoodh: `"cl"` here is `GATEWAY_CLIENT_SEGMENT` and `"jobs"` is already in both `GATEWAY_API_SEGMENTS` and `GATEWAY_BARE_API_SEGMENTS`, so this is a fourth hand-maintained segment list sitting next to the three #8006 centralized - the exact second-source-of-truth shape that centralization was for. At minimum reference the existing constant so one of the two can't drift:

```jac
     GATEWAY_DISCOVERY_SEGMENTS: tuple = ("jobs", GATEWAY_CLIENT_SEGMENT);
```

And a line saying *why* those two are probe-safe (the `/jobs` create-404 fires before any scheduling side effect, `/cl` pages are GET-only) would keep the next person from adding a third entry that isn't.

jac/jaclang/scale/runtime/gateway/microservice_gateway.impl.jac
```diff
@@ -373,6 +376,52 @@ impl MicroserviceGateway.handle_builtin_passthrough(
         }
     }
 
+    import from jaclang.scale.runtime.gateway.dispatch { is_gateway_owned }
+
+    stripped: str = path.strip("/");
+    first_seg: str = stripped.split("/")[0] if stripped else "";
+    if first_seg in _DISCOVERY_SEGMENTS or request.method in _SAFE_METHODS {
+        return await self.fanout_discovery(request, path, True);
```
MusabMahmoodh: This bypasses ownership for every safe method, so GET/HEAD/OPTIONS builtin traffic still fans out across all services and still pays the O(services) sequential cost - which is the second of the three symptoms the issue lists, and it survives this PR untouched even though the table that would fix it is now reachable from here.

Was that deliberate scoping (fix the correctness half, leave latency for later)? If so it's worth saying in the PR body and the fragment, because the release note currently reads as though builtin routing is now owner-first, and for the majority of gateway traffic by volume it isn't. Otherwise: a safe method with a known owner could take the owner branch too, and only fall back to the probe on a miss.

MusabMahmoodh (commented):
Second pass, on the "is this the real fix or a patch, and what does it leave behind" axis rather than on correctness. Three things, all cheap to address.

jac/jaclang/scale/runtime/gateway/microservice_gateway.impl.jac
kashmithnisakya: split: /user and the other replicated families are deliver-once and never learned; owned families are owner-first and a miss probes only services whose routes are still unknown. the throttle is per service now (a new registration, a url/status change, or an unanswered fetch is due on the very next miss), so the register() hook isn't needed (_routes_refreshed_at lived on the gateway anyway). 0c604aa87

jac/jaclang/scale/runtime/gateway/microservice_gateway.impl.jac
kashmithnisakya: the crawl is parallel and shared across concurrent misses, and only asks services that are unknown, changed, or older than route_refresh_seconds, so a fully learned fleet never touches the network on a miss. kept it inline instead of background because a cold miss delivering to the first service is the exact bug. 0c604aa87

jac/jaclang/scale/runtime/gateway/microservice_gateway.impl.jac
kashmithnisakya: extracted forward_root_path, used by the owner branch, deliver_once and handle_app_route; probe_services and the graph aggregation address through service_base_url too now. 0c604aa87

jac/jaclang/scale/runtime/gateway/route_table.impl.jac
kashmithnisakya: learn() goes through is_learnable_prefix in routing.jac now: only concrete members of service-owned families or app routes, so /cl, /sso, /api-key, /graph, /jobs are never learned. PR body corrected, those are replicated families and are delivered once. 0c604aa87

jac/jaclang/scale/tests/microservices/test_gateway.jac
kashmithnisakya: backend drains to content-length before replying. added the throttled-window contract (a walker added to an already-learned service waits for the window) plus late registration and unreachable-then-up being learned on the next miss. 0c604aa87

jac/jaclang/scale/runtime/gateway/microservice_gateway.impl.jac
kashmithnisakya: gone: probe_services is only the probe (with an explicit candidate list) and deliver_once picks the first routable service through forward_to. 0c604aa87

jac/jaclang/scale/runtime/routing.jac
kashmithnisakya: now ("jobs", GATEWAY_CLIENT_SEGMENT), with the replicated and service-owned families beside it and a test pinning that the three partition GATEWAY_API_SEGMENTS. the why lives in the fragment and PR body since strip-comments would eat an inline note. 0c604aa87

jac/jaclang/scale/runtime/gateway/microservice_gateway.impl.jac
kashmithnisakya: safe methods take the owner branch now; the probe is only the miss fallback, and only against services with unknown routes. 0c604aa87

MusabMahmoodh (commented):
Re-reviewed at `a505f9ed0`. All eight of my 08-15 points are addressed, so this moves off needs-changes.

Worth saying which way you went: I suggested a `register()` hook to close the 30s dark window, and you did per-service definitive-answer tracking instead (`route_table.jac:38`, `spec_due` at `route_table.impl.jac:49-62`). That is the better design, because it distinguishes "not asked yet" from "asked and got nothing" rather than just shrinking the window. Same for `is_learnable_prefix` (`routing.jac:62-69`) turning my deny-list guard into an allow-list with an enumerating test.

Four follow-ups, the first is the one I would want before merge.

**`route_table.jac:63`** - the spec fetch runs inline on the miss path at the 5s default, and an unanswered fetch is never `mark_spec_seen` (`route_table.impl.jac:70-73`), so a service that is HEALTHY but wedged stays due forever. Every 404-bound builtin request then becomes a 5-second held connection on the gateway. A `timeout=1.0` on this call bounds it without changing the learning behaviour.

**`dispatch.jac:63`** - `route_table.impl.jac:2` was `is_gateway_owned`'s only production caller and this PR replaces it with `is_learnable_prefix`, so it now has no callers outside `test_routing_guard.jac:121-123`. Two functions answering "what does the gateway own", one reachable only from its own test, is the same second-source-of-truth shape the family tables just removed. Worth deleting it and those three asserts here, while it is still obvious why they went.

**`microservice_gateway.impl.jac:442`** - `eligible[0]` is registry order, so the whole fleet's identity traffic lands on whichever service registered first, permanently. Was picking `eligible[hash(path) % len(eligible)]`, or rotating, considered, or is pinning deliberate for cache or session reasons I am not seeing?

**`test_gateway.jac`, "identity fallthrough"** - `assert total <= 1` also passes when the credential-bearing body reached nobody, which is the failure this test exists to catch. `== 1` makes it discriminate.

PR #8119, opened 2026-08-13
fix(runtime): pool and reuse HTTP client connections (#7907)

### LINKED ISSUE #7907: Jac Scale: the gateway opens a new aiohttp ClientSession per forwarded request, so every proxied call pays full connection setup

akindu-k:
Fixes #7907.

Explanation: https://claude.ai/code/artifact/c9906fe0-8b85-4440-a3aa-f03b8aa01a25

## What was wrong

The issue was filed against the gateway's aiohttp forwarders. Those are gone (the #7969 serving rewrite), but the cost moved rather than disappeared: `jaclang.server.serving.client.arequest` opens a fresh `asyncio.open_connection` per call, hard-codes `Connection: close`, and builds a new `ssl.create_default_context()` on every HTTPS request. Three sequential requests to the same origin opened three TCP connections.

That is one layer below the gateway, and it is the layer worth fixing: `arequest` also backs the SSO token exchange and the admin probe/log/trace fetches, all of which paid the same handshake. A gateway-local session would have left those callers behind.

image

image

## What changed

- **Connection pool.** Idle keep-alive connections are cached per origin in a pool owned by the running event loop, bounded per host (32) and in total (256), with an idle expiry (60s) swept across every origin on release rather than only on the origin next asked for. The pool is stored as an attribute on the loop itself, so the loop, its pool and their sockets form one cycle the collector can take as a unit. A module-level registry keyed by loop would not do: the pooled writers reach the loop through their transports, so even a weak key would be pinned by its own value. Loop ownership also keeps connections from crossing event loops, which asyncio streams cannot do.
- **Maximum connection age.** A pooled connection is retired 120s after it was opened, however busy it is. Its idle deadline is capped at that age and `release` closes it instead of pooling it once the age is reached. Without this, a connection in constant use never expires. Behind a ClusterIP Service, kube-proxy picks the pod once per connection, so after an HPA scale-out the old pods would keep all the traffic and the new ones would get none. `Connection: close` used to rebalance on every request.
- **Keep-alive.** `Connection: close` is no longer forced. A response is returned to the pool only once it is fully drained and its framing says the connection is still usable: no `Connection: close`, HTTP/1.1 (or an explicit `keep-alive`), and a body delimited by `Content-Length` or chunked encoding rather than by EOF. Reuse additionally requires an empty reader, checked both when a connection is parked and when it is taken back out. An upstream that overruns its `Content-Length` is caught outright when the extra bytes have already arrived; if they land later, the next response head is malformed and the request retries under the rule below.
- **Stale-connection retry.** An *idempotent* request that fails on a pooled connection before a usable response head is read is retried once, on a connection opened fresh rather than drawn from the pool again, per RFC 7230 6.3.1. Retrying from the pool is what makes this fragile in practice: a server expiring its keep-alive timer kills the whole idle bucket at once, so the retry would draw a second dead socket and fail. A connection that dies before the head arrives leaves the request's fate unknown, so `POST` and `PATCH` surface the failure rather than being replayed. Connect failures and read timeouts are not retried.
- **Worker recycling closes its last connection.** With `[serve.workers] max_requests` set, the jac server used to answer the budget-hitting request with `Connection: keep-alive` and only then stop the worker, handing the pool a connection that was about to die. A `POST` landing on it came back as a 502. The recycle is now decided before the response is written, so that response goes out with `Connection: close`.
- **Correct framing, now that it matters.** The chunked reader stopped at the terminating zero-size chunk without consuming it or its trailers, and there was no notion of a bodiless response. Both were harmless while the socket was thrown away and corrupting once it is reused, so `HEAD`/204/304 now stop at the head and chunked reads consume the trailer section.
- **Timeouts split.** Connect gets its own budget (`min(10s, timeout)`) instead of letting a slow handshake consume the whole read window.
- **TLS context cached** instead of reloading the system CA bundle per HTTPS request.
- **Request framing belongs to the client.** `arequest` computes `Content-Length` itself, so a caller-supplied `content-length` or `transfer-encoding` is dropped rather than emitted alongside it. On a pooled connection a conflicting length is the classic desync, and the gateway forwarded inbound headers without filtering either one.
- **One hop-by-hop set.** The gateway's `HOP_BY_HOP_HEADERS` and `http_forward`'s response filter were separate lists that had drifted apart. They are now one set in `http_forward.jac`, used for both forwarded requests and relayed responses. It includes the RFC-mandated `te`, `trailer`, `upgrade`, `proxy-authenticate` and `proxy-authorization`, plus `content-length`.
- **Lifecycle.** Only the async serving path gets pooling. The serving loop closes its pool on shutdown. The sync entry points (`request_sync`, `post_stream_aware_sync`) own a private loop per call and close their pool with it, so they still pay a handshake per call. That includes service-to-service calls (`sv_client` and `rpc.jac`).

## Tests

`jac/tests/runtimelib/test_serving_client_pool.jac` is a new file rather than an extension of `test_serving_client_errors.jac`. That file owns the phase-typed exception taxonomy and drives a threading-based socket server. Pooling needs a keep-alive asyncio upstream that counts accepted connections and can kill them selectively, which does not fit the existing harness. Serving-client test count goes from 8 to 19.

Eleven tests, tabled wherever the cases share a shape:

- `connection reuse follows the response framing` tables four request sequences against their expected statuses, bodies and connection counts: reuse across three sequential requests, a bodiless 204, `Connection: close`, and a response overrunning its `Content-Length`.
- `a dead pooled connection is retried only for idempotent methods` tables `GET` (retried on a new connection) against `POST` (surfaced, never replayed).
- `the retry opens a fresh connection instead of taking another dead one` pools two connections, then makes the upstream kill every reused connection.
- `the pool keeps no empty, expired or overaged connections` covers idle expiry sweeping, rejected connections leaving no bucket, a connection past its maximum age being closed rather than pooled, and a younger one's idle deadline being capped at that age.
- `a worker at its request budget closes the connection it answers on` runs a real `ServingServer` with `request_budget=1` and asserts the response carries `Connection: close` and nothing is pooled. This is the only new test with a server, because the budget path in the connection loop had no coverage at all. It fails on the previous `server.jac`.
- `the buffered byte check reads asyncio's real buffer` tables an empty reader, a fed reader and a foreign object. It fails loudly if `StreamReader._buffer` is ever renamed.
- The remaining five cover a drained stream handing its connection back, transparent replacement of a dropped connection, the per-host idle bound, a caller-supplied `Content-Length` never reaching the upstream, and a loop with pooled connections still being collectable.

Before the change the first case saw three connections for three requests.

Local runs: 46 passed across the pool file, `test_serving_client_errors`, `test_serve_disconnect`, `test_worker_retirement`, `test_bodiless_status` and `test_stream_ws`. Across every suite that imports the gateway, `http_forward` or the serving client, 374 passed and 4 failed. Three of the failures are the gateway `/metrics` tests, which fail the same way on `main` in this environment. The fourth is fleet edge analysis in `test_fleet_from_apps`, which this PR does not touch. It is left to CI.

## Not addressed here

- Only idle connections are bounded. A fan-out burst still opens as many sockets as it has in-flight requests; capping those needs a semaphore and a wait queue.
- Service-to-service calls through the sync entry points still pay a handshake per call, as described under Lifecycle.
- Draining a small response body on the gateway's 404/405 passthrough hops, so those connections return to the pool instead of being discarded.
- The gateway still fans out serially in `handle_builtin_passthrough` and `aggregate_graph_data`, one service at a time.

All of these sit under #7900.


jac/jaclang/runtimelib/serving/client.jac
```diff
@@ -57,48 +153,72 @@ obj ClientResponse {
...
         }
+        while remaining > 0 {
+            block: bytes = await self.reader.read(min(65536, remaining));
+            if not block {
+                return;
+            }
+            remaining -= len(block);
+            yield block;
+        }
+        self._drained = True;
```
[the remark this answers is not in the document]
akindu-k: Valid, fixed in 4aff44c9d.

`_iter_sized` marked the response drained at the declared length and `release` had no way to see that the reader still held bytes, so an upstream that overran its `Content-Length` left them at the head of a pooled connection.

Reuse now requires an empty reader. `_is_stale` became `_is_unusable` and folds in a `_has_buffered_bytes` check, which runs both when a connection is parked and when it is taken back out, so bytes that land after the connection is already idle are caught before the next request rides it. Since this client never pipelines, any byte remaining after a fully framed response is a protocol violation, and the connection is closed rather than pooled.

Covered by "a response longer than its content length is not pooled", which asserts the first body is still intact, the follow-up response is clean, and a second connection was opened. It fails on the previous code.

jac/jaclang/runtimelib/serving/client.jac
```diff
@@ -219,56 +372,128 @@ async def arequest(
...
+            OSError,
+            asyncio.IncompleteReadError,
+            ClientConnectionError
+        ) as e {
+            _close_writer(writer);
+            if pooled is None or retried {
+                raise ClientConnectionError(detail=str(e));
+            }
+            retried = True;
+            continue;
```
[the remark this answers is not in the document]
akindu-k: Valid for `POST`/`PATCH`, fixed in 4aff44c9d. The retry was method-agnostic, which contradicts the RFC 7230 6.3.1 rationale the PR cites: that text permits automatic retransmission only when the aborted requests have idempotent methods.

`arequest` now gates the retry on the method. `POST` and `PATCH` raise `ClientConnectionError` instead of being resent, so the one-time OAuth code exchange cannot be consumed by the first attempt and replayed by the second.

`PUT` and `DELETE` stay retryable. They are idempotent by definition in RFC 9110 9.2.2, and they are exactly what the cited retry rule covers. `POST` is the method whose application status is genuinely unknowable, and that is the one now excluded.

Two tests cover it, both failing on the previous code: "an idempotent request is retried when its pooled connection dies" and "a post is not replayed when its pooled connection dies". The upstream probe gained a `drop_next` flag so the connection dies after the request is read, which is the ambiguous case; without it the drop is caught at acquire time and the retry path never runs.

MusabMahmoodh (changes requested):
Sorry for the wait on this one, it sat far too long.

Read the whole diff. Retargeting #7907 from the gateway forwarders to `arequest` was the right call, and the reasoning for it in the body is the part I would keep: the handshake cost moved down a layer rather than disappearing, and a gateway-local session would have left RPC, SSO exchange and the admin fetches paying it. Two of the framing fixes are the good kind of unglamorous: consuming the chunked terminator and its trailers, and having a notion of a bodiless response, are both exactly the bugs that are invisible while the socket is discarded and corrupting the moment it is not. The RFC 7230 6.3.1 line on replay is drawn in the right place.

Findings below, heaviest first.

## 1. A caller's `Content-Length` reaches the upstream alongside the computed one, on a connection that is now reused

`_prepare_payload` passes caller-supplied headers through verbatim, and `_request_blob` always emits its own `Content-Length: {len(body)}`. So any caller that puts `Content-Length` in `headers` produces a request with two of them.

That path is reachable from outside. `MicroserviceGateway.build_forward_headers` (`microservice_gateway.impl.jac:242`) copies inbound request headers, filtering only against `HOP_BY_HOP_HEADERS` (`microservice_gateway.jac:36`), which is `{host, transfer-encoding, connection, keep-alive}`. `content-length` is not in it. The response-side `_HOP_BY_HOP` in `http_forward.jac` does have it, so the asymmetry looks accidental rather than deliberate. `stream_forward` and `raw_forward` both hand those headers straight to `arequest`.

Today the blast radius is bounded by the thing this PR removes: `Connection: close` means the upstream reads one request and the socket dies, so a conflicting length costs at most that one exchange. Once the connection is pooled and reused, conflicting framing on a persistent connection is the classic desync shape. If an upstream honors the forwarded value rather than the computed one, the connection is left misaligned, and the pool's protection does not see it: `_has_buffered_bytes` inspects the client's read buffer, which says nothing about a request-side desync at the peer.

`transfer-encoding` being filtered does close the TE.CL variant. CL.CL is open.

The fix I would rather see here than at the gateway: `arequest` owns request framing, so `_prepare_payload` should drop caller-supplied `content-length` and `transfer-encoding` outright. That covers every caller instead of the one I happened to trace. Adding `content-length` to `HOP_BY_HOP_HEADERS` is still worth doing on its own merits, since a gateway should not forward it.

## 2. `_has_buffered_bytes` fails open on the check that prevents cross-request corruption

```
def _has_buffered_bytes(reader: any) -> bool {
    try {
        return len(getattr(reader, '_buffer', b'')) > 0;
    } except Exception {
        return False;
    }
}
```

`False` means "clean, safe to pool". `_is_unusable`'s own outer handler gets this right and returns `True` when it cannot tell. This one returns the unsafe answer.

It matters because `reader._buffer` is a private asyncio attribute, and it is the sole mechanism behind the guarantee that an upstream overrunning its `Content-Length` cannot leave bytes at the head of the next request's response. If that attribute is ever renamed, `getattr` returns the `b''` default, every connection reads as clean, and the suite stays green while the guarantee is gone. The test for the overrun case would keep passing because it also goes through this function.

`return True` on the failure path, and it would be worth asserting somewhere that the attribute exists at all, so a rename fails loudly instead of silently downgrading to trust.

## 3. `idle_timeout` never closes an idle socket, it only refuses to hand it out

Expiry is evaluated in `acquire(origin)` and nowhere else. A bucket for an origin that is not requested again keeps its writers open until `close_all`, well past the 60s the constant advertises.

Those entries also count against the total budget, because `idle_count()` sums every bucket without regard to expiry. So a process that talks to a spread of short-lived origins can fill `MAX_IDLE_TOTAL` with connections that expired minutes ago, and then `release` starts rejecting live connections for the origin it actually uses, quietly turning pooling back off. A sweep of expired entries in `release`, before the budget check, would fix both halves.

## 4. Smaller

- `release` calls `self._idle.setdefault(origin, [])` before the unusable check, so a rejected connection still leaves an empty bucket keyed by origin. Key growth is unbounded across origins, and those empty buckets are what `idle_count` then walks on every release.
- `_POOLS` is a `WeakKeyDictionary` keyed by the loop. If a loop is collected without `close_connection_pool` having run, the pool goes with it and its writers are never closed. `weakref.finalize(loop, pool.close_all)` at creation would make the sockets follow the loop rather than depending on every exit path calling the closer.
- The sync entry points close the pool per call, so `request_sync` and `post_stream_aware_sync` still pay a handshake per call. The body is accurate about the lifecycle, but a reader can come away thinking the win is global. Worth one clause saying only the async serving path pools.
- The overrun test covers the deterministic case, where the extra bytes are already buffered when the connection is released. If they arrive after release, `acquire` catches them; if they arrive after `acquire`, the next response head is malformed, which raises `bad status line`, and that is caught and retried on a fresh connection for idempotent methods. That is the right outcome, but the body reads as a stronger guarantee than the timing actually gives, so it is worth stating as a retry rather than as prevention.

Items 1 and 2 are what I would want changed before this lands. 3 is a real leak but a slower one. The rest are notes.

MusabMahmoodh (approved):
Clearing my changes-requested. Both blockers are properly closed, not patched around:

- `_FRAMING_HEADERS` strips caller-supplied `content-length` and `transfer-encoding` in `_prepare_payload`, so `arequest` owns request framing for every caller rather than for the one path I traced, and `content-length` is now in the gateway's hop-by-hop set as well. The `sent.count("content-length") == 1` assertion is the right shape for a test of this - it catches the duplicate directly instead of asserting the strip happened.
- `_has_buffered_bytes` returns `True` on failure and reads `reader._buffer` directly rather than through `getattr` with a default, so an asyncio rename raises and fails safe instead of silently downgrading the whole pool to trust.
- Item 3 is done too: `_sweep()` runs in `release` before the budget check, and there is a test for an origin nobody asks for again, which was the half I expected to be left.
- `release` no longer `setdefault`s an empty bucket for a connection it then rejects.

The one note I made that you did not take is `weakref.finalize(loop, pool.close_all)` at pool creation, so sockets follow a collected loop. Still a note, not an ask.

**One thing that is new since my last read, and it is between two of your own PRs.** #8449 rewrites `arequest`'s framing in the same function, and it needs the opposite of what this PR does with `content-length`: it captures the caller's value into `declared_length` and re-emits it so a streamed body can be forwarded with a real length instead of being chunk-framed. This PR drops that header outright. Whichever lands second has to reconcile them, and the merged behaviour has to be "drop a caller-supplied length unless a body stream is being passed through, in which case that length is the frame". Worth deciding now which one carries the reconciliation rather than discovering it in a conflict - my read is that this one lands first since #8449 is red and newer, so #8449 picks it up.

jac/jaclang/server/serving/client.jac
```diff
@@ -35,14 +52,140 @@ def _client_json_default(o: any) -> str {
...
+            return;
+        }
+        self._sweep();
+        bucket = self._idle.get(origin) or [];
+        if len(bucket) >= self.max_idle_per_host
+            or self.idle_count() >= self.max_idle_total {
+            _close_writer(writer);
+            return;
+        }
+        bucket.append((reader, writer, time.monotonic() + self.idle_timeout));
```
MusabMahmoodh, line 139: `expires_at` is reset on every release, so a connection that stays in use never ages out. The gateway reaches apps through their ClusterIP Service (`k8s_service_dns_url`), where kube-proxy picks the pod once per connection, so after an HPA scale-out the new pods only get connections opened above what this LIFO stack already holds and the old pods stay hot; `Connection: close` used to rebalance every request for free. Stamping the open time next to `expires_at` and closing instead of pooling once a connection passes a max age (a couple of minutes is plenty) would bound that.

jac/jaclang/server/serving/client.jac
```diff
@@ -233,64 +441,139 @@ async def arequest(
...
+            _close_writer(writer);
+            raise;
+        } except (
+            ConnectionError,
+            OSError,
+            asyncio.IncompleteReadError,
+            ClientConnectionError
+        ) as e {
+            _close_writer(writer);
+            if pooled is not None and not retried and replayable {
```
MusabMahmoodh, line 562: Not replaying the POST is right, but the jac server itself hands this pool that dead connection when `[serve.workers] max_requests` is set: `_connection_loop` writes the budget-hitting response with `Connection: keep-alive` (server.jac:396-399) and only afterwards stops the worker and closes the socket (server.jac:406-412). That connection is the one on top of the LIFO stack, so the next POST to that app can land on it and come back as `ClientRequestSent`, which `forward_to` turns into a 502. Deciding the recycle before `_write_response` so the last response goes out with `Connection: close` fixes it at the source, and server.jac is already in this diff.

release_notes/unreleased/jaclang/8119.bugfix.md
```diff
@@ -0,0 +1 @@
+- **Fix: HTTP connections are reused instead of reopened**: Proxied requests, service-to-service calls and login flows now keep connections alive between requests, so they no longer pay a fresh connection setup every time.
```
MusabMahmoodh, line 1: Service-to-service calls don't get this: `sv_client._http_call`/`_http_spawn` and `rpc.jac` go through `post_stream_aware_sync`/`request_sync`, which run each call on a private loop and close its pool on the way out, so sv-to-sv still pays a handshake per call. Probably worth its own issue under #7900 next to the in-flight cap.
```suggestion
- **Fix: HTTP connections are reused instead of reopened**: Gateway-proxied requests, admin probes and login flows now keep connections alive between requests, so they no longer pay a fresh connection setup every time.
```

MusabMahmoodh (commented):
The delta since my approval is right: the loop registry was a fallback `setattr` never reached, and retrying on a fresh connection instead of the pool is what actually survives a server expiring the whole idle bucket at once, with a test that fails on the old revision. The branch conflicts with main again in `client.jac` after #9068 retyped `aiter_raw`/`aiter_lines` to `AsyncIterator[bytes]`, and `_iter_chunked`/`_iter_sized` should take the same annotation in that merge. The first two comments are about pool lifetime against a real multi-replica fleet and are the ones I'd want settled before this merges.

jac/jaclang/scale/runtime/gateway/microservice_gateway.jac
```diff
@@ -35,6 +35,7 @@ glob logger = logging.getLogger(__name__),
      HOP_BY_HOP_HEADERS: set[str] = {
          "host",
          "transfer-encoding",
+         "content-length",
```
MusabMahmoodh, line 38: This is the same hop-by-hop list as `_HOP_BY_HOP` in http_forward.jac, which this file already imports from, and after this PR the two disagree on `te`, `trailer`, `upgrade` and `proxy-*`: the gateway now drops those from upstream responses but still forwards a client's `TE` and `Proxy-Authorization` upstream. One set in http_forward.jac used by both `build_forward_headers` and `_filtered_headers` keeps them from drifting again.

jac/jaclang/server/serving/client.jac
akindu-k: Valid, fixed in d5d7312d3.

Each pooled connection now carries the time it was opened. `release` closes it instead of pooling it once it is older than `MAX_CONNECTION_AGE` (120s), and its idle deadline is capped at that age so it also leaves the pool on the sweep. A busy connection is therefore reopened at least every two minutes, which gives kube-proxy a chance to route it to a new pod after a scale-out.

Covered in "the pool keeps no empty, expired or overaged connections": a connection past its age is not pooled, and a younger one gets an idle deadline no later than its age limit.

jac/jaclang/server/serving/client.jac
akindu-k: Valid, fixed at the source in d5d7312d3.

`_connection_loop` now counts the request and decides the recycle right after dispatch, before `_write_response`, and passes `keep_alive = False` when the budget is spent. The response that exhausts the budget goes out with `Connection: close`, and the worker is stopped after it is written, as before.

New test "a worker at its request budget closes the connection it answers on" runs a `ServingServer` with `request_budget=1` and asserts the response carries `Connection: close` and nothing is pooled. The budget path had no coverage before, and the test fails against the previous `server.jac`.

release_notes/unreleased/jaclang/8119.bugfix.md
akindu-k: Agreed on both. I took the suggested wording in d5d7312d3, and the PR body now says outright that sv-to-sv calls through `request_sync` / `post_stream_aware_sync` still pay a handshake per call. Filed as #9300 under #7900.

jac/jaclang/scale/runtime/gateway/microservice_gateway.jac
akindu-k: Valid, fixed in d5d7312d3. The set now lives only in `http_forward.jac` as `HOP_BY_HOP_HEADERS`. `microservice_gateway.jac` imports it, so `build_forward_headers` and `_filtered_headers` filter against the same list. A client's `TE`, `Trailer`, `Upgrade` and `Proxy-*` headers are no longer forwarded upstream.

akindu-k:
Merged current `main` in 6535e79d0. The only conflict was in `client.jac` after #9068. `_iter_chunked`, `_iter_sized` and `aiter_lines` now all return `AsyncIterator[bytes]`, matching `aiter_raw`. #8080 also landed, so the remaining bare `dict` annotations in `_prepare_payload` and the pool test file now have type arguments.

All four inline comments are addressed in d5d7312d3, with replies on each thread.

MusabMahmoodh (approved):
The age cap, the `Connection: close` on the response that spends the worker budget, and the single hop-by-hop set all check out, and the budget test fails against the old `server.jac`. The sync sv-to-sv path is rightly left to #9300.

PR #8120, opened 2026-08-13
feat(scale): read JAC_SV_<PEER>_URL in the k8s service registry

### LINKED ISSUE #8108: [Feature]: k8s service registry should honor the JAC_SV_<PEER>_URL env it already injects
## Summary

In Kubernetes mode, peer service URLs are always derived by naming convention and the per-peer env overrides that jac-scale itself injects are never read. Make the k8s service registry honor `JAC_SV_<PEER>_URL` from the environment when present, falling back to the convention-derived URL otherwise.

## Current behavior

- `manifest_builder.jac:712` injects `JAC_SV_{PEER}_URL` env vars into every pod.
- At runtime, `get_sv_registry` (`jaclang/scale/plugin.jac:772-797`) detects `KUBERNETES_SERVICE_HOST` and resolves every peer via `build_k8s_registry` (`deploy/target/kubernetes/runtime_resolver.jac:84-95`), which constructs `http://{svc}-service.{namespace}.svc.cluster.local:{port}` purely from the route name.
- No code anywhere reads `JAC_SV_<PEER>_URL` (only `JAC_SV_ROUTES` is consumed, and only for the route name set, `config_loader.impl.jac:512`). The injected URLs are dead.

## Requested behavior

In `build_k8s_registry` (or its caller), for each route name, use `os.environ.get(f"JAC_SV_{name.upper()}_URL")` when set and non-empty; otherwise keep the derived URL. No other behavior change.

## Why

1. It makes the env jac-scale already injects meaningful instead of dead, so a pod's routing is reproducible from its spec.
2. It gives platforms built on jac-scale a supported way to point a pod set at alternate service names without raw manipulation of jac-scale-owned objects. Concretely, jacBuilder's blue-green deployment work (jaseci-labs/jacBuilder#1232) keeps the previous release as a warm standby stack with `-standby` suffixed Services and needs standby pods to resolve peers to their own color; today the convention-derived URLs make that impossible (analysis: jaseci-labs/jacBuilder#1233).

## Backward compatibility

Pods without overrides resolve byte-identically to today. Pods with the injected defaults also resolve identically, since the injected values equal the derived ones. Only a deliberately rewritten override changes anything.

## Notes

- Sibling ask, different layer: #8089 covers revision listing/rollback. This issue is only about env-route precedence in the runtime resolver.
- Suggested tests: unit test on `build_k8s_registry` (override wins, empty/missing falls back, malformed name untouched), plus a pod-env round-trip in the existing microservices test suite.

Thamirawaran:
## **Description**

Closes #8108.

In Kubernetes mode every peer URL was derived from the route name, so the `JAC_SV_<PEER>_URL` env vars jac-scale injects into every pod were never read. `get_sv_registry` returns `build_k8s_registry`, which emits an entry for every declared route, so the `sv_client` fallback that does read the env was unreachable for any declared peer. The injected values were dead, and a platform could not point a pod set at alternate Services without editing jac-scale-owned objects.

`build_k8s_registry` now takes the env value when it is set and non-empty, and derives otherwise. Deployed pods are unaffected, because the value jac-scale injects is the value it would have derived.

### The key format now has one definition

It was being assembled by hand in three places: the k8s manifest builder, the local process manager, and `sv_client`. It now lives once in `sv_client.peer_url_env_key`, and every writer plus the reader share it.

That matters because the two sides disagree by design. The env key keeps the raw route name, while Kubernetes object names are RFC1123-safe. A peer named `orders_app` is injected as `JAC_SV_ORDERS_APP_URL` beside a Service named `orders-app-service`, so a reader keyed off the safe name would look up `JAC_SV_ORDERS-APP_URL`, miss, and silently fall back to the derived URL. That failure is indistinguishable from the bug being fixed, which is why the tests assert the literal key rather than calling the helper.

### Behavior change

A peer URL that is set but blank now reads as absent. `resolve_url` raises naming the variable to set, and `_ensure_available` no longer caches an empty URL as a registration. Previously both treated a blank value as present, which moved the failure to the first request where an empty URL was much harder to place.

## **Tests**

Added to the existing suites rather than new files. All real objects and real environment variables, no mocks or fakes.

`test_pod_env.jac`, which already owns the pod env contract and the `ManifestBuilder` fixture:
- a peer with no override resolves by naming convention
- an override wins for its own peer and leaves the others derived
- a blank override falls back to the derived URL
- the key the builder injects is the key the registry reads
- the injected default equals the derived URL under matching config

`test_eager_spawn.jac`, beside its existing env-to-registry test:
- the env key keeps the raw module name
- unset, blank and whitespace all read as absent
- an in-process registration wins over the env
- the env is used when nothing is registered
- a blank env value raises instead of resolving to an empty URL
- a blank env value is never registered

Each was verified to fail when the behavior it guards is removed. Reverting the registry change fails the override and key round-trip cases; reverting the `sv_client` change fails the blank-value case.

## **Verification**

On a kind cluster, a real three-service deploy. The running pod carries:

```
JAC_SV_ROUTES          {"orders_app": "/orders", "main": "/__main__"}
JAC_SV_ORDERS_APP_URL  http://orders-app-service.default.svc.cluster.local:8000
```

beside Deployments named `orders-app-deployment` and a Service named `orders-app-service`, which is the underscore and hyphen split confirmed against a real cluster rather than assumed.

Running this branch's resolver on Linux with those exact routes and values resolves an override to the standby Service and leaves unrelated peers derived.

## **Notes**

- Existing deployments are unaffected, and the last test asserts that rather than claiming it.
- `get_sv_registry` still ignores `sv_client._registry` in-cluster, so a programmatic `register()` is still a no-op in a pod. Out of scope here, but it is the same class of gap in the same function.


jac/jaclang/scale/deploy/target/kubernetes/runtime_resolver.jac
```diff
@@ -87,9 +88,9 @@ def build_k8s_registry(
     namespace: str = in_cluster_namespace();
     registry: dict[str, str] = {};
     for svc_name in routes.keys() {
-        registry[svc_name] = k8s_service_dns_url(
-            k8s_safe_name(svc_name), namespace, port
-        );
+        override: str = sv_client.peer_url_from_env(svc_name);
+        registry[svc_name] = override
+            or k8s_service_dns_url(k8s_safe_name(svc_name), namespace, port);
```
MusabMahmoodh: A non-blank override is consumed verbatim, so a value without a scheme (say `orders-app-standby-service`) lands in the registry and only fails at the first RPC with a connection error far from the cause - the exact failure shape this PR removed for blanks. The issue scoped this as "no other behavior change", so a question: would you take a warning (or a hard fail) here when the value does not start with `http://` or `https://`, or is that for a follow-up?

jac/jaclang/scale/deploy/target/kubernetes/runtime_resolver.jac
```diff
@@ -87,9 +88,9 @@ def build_k8s_registry(
     namespace: str = in_cluster_namespace();
     registry: dict[str, str] = {};
     for svc_name in routes.keys() {
-        registry[svc_name] = k8s_service_dns_url(
-            k8s_safe_name(svc_name), namespace, port
-        );
+        override: str = sv_client.peer_url_from_env(svc_name);
```
MusabMahmoodh: A present-but-blank override falls back to the derived URL with no trace. Blank-as-absent is the right semantic (clearing a value restores routing, and the test documents why), but a one-line debug log naming the key when it is set-but-blank would make a templating bug findable instead of invisible:

```jac
if not override and sv_client.peer_url_env_key(svc_name) in os.environ {
    logger.debug(f"blank {sv_client.peer_url_env_key(svc_name)} ignored; using derived URL");
}
```

MusabMahmoodh (commented):
jac/jaclang/scale/deploy/target/kubernetes/runtime_resolver.jac:24 (url_for):
`url_for` a few lines above still derives purely by convention, so it is now the one k8s resolution site that ignores the override. As far as I can tell nothing constructs `K8sRuntimeServiceResolver` yet, so it is dormant, but the day the "not yet wired" admin ops wire it up it silently bypasses the feature. Was leaving it for the admin-ops work intentional, or should it share the fallthrough now:

```jac
def url_for(service_entry: ServiceEntry) -> str {
    return sv_client.peer_url_from_env(service_entry.name)
        or k8s_service_dns_url(
            k8s_safe_name(service_entry.name), self.namespace, self.port
        );
}
```

jac/jaclang/scale/deploy/target/kubernetes/runtime_resolver.jac
Thamirawaran: warning. raising would turn a typo into a pod that won't start, and a bad override can't fall back to derived the way a blank one does.

though it belongs in `peer_url_from_env`, not here. `resolve_url` and `_ensure_available` share it, so guarding this line alone leaves them open.

will add it here unless you'd rather a follow-up.

PR #8121, opened 2026-08-13
fix(scale): honor X-Forwarded-For only from configured trusted proxies in the rate limiter

### LINKED ISSUE #7903: Jac Scale: rate limiter trusts X-Forwarded-For unconditionally, so any client bypasses per-IP limits with a single header
**Area:** runtime core / middleware
**Severity:** critical (security)

### What happens

The rate limiter derives its per-IP key from the `X-Forwarded-For` request header whenever that header is present, with no trusted-proxy check. Any client can set the header to an arbitrary value and get a fresh bucket per request.

### Where

`jac/jaclang/scale/runtime/middleware/rate_limit.jac:122-129`

### Evidence

```jac
def _resolve_client_ip(request: any) -> str {
    xff = request.headers.get("X-Forwarded-For");
    if xff {
        return xff.split(",")[0].strip();
    }
    client = request?.client;
    return getattr(client, "host", "unknown") if client else "unknown";
}
```

### Why it matters

Per-IP rate limiting is fully bypassable with one header, which also defeats it as a brute-force and resource-exhaustion control. When the gateway is directly exposed (NodePort, or a load balancer that does not strip the header), this is trivially exploitable.

### Suggested fix

Add a trusted-proxy configuration: only honor `X-Forwarded-For` when the immediate peer is in the configured trusted set, and take the rightmost untrusted hop rather than the leftmost entry. Default to the peer address when no trusted proxies are configured.

Part of #7900

kashmithnisakya:
Fixes #7903 (Part of https://github.com/jaseci-labs/jac/issues/7900)

## Root cause

Two defects in `_resolve_client_ip` (`runtime/middleware/rate_limit.jac`):

1. **No trust boundary on forwarded headers.** Whenever `X-Forwarded-For` was present its *leftmost* entry became the per-IP bucket key. The leftmost entry is the one the original client writes, so any caller minted a fresh bucket per request with a one-line header, which fully defeats per-IP limiting as a brute-force and resource-exhaustion control.

2. **The fallback path was dead code.** The serving `Request` carries the TCP peer as `client_host`; it has no Starlette-style `.client` attribute. `request?.client` therefore resolved to None and every client that did not send the header keyed to the literal string `"unknown"`: one shared bucket for all direct traffic, so per-IP limiting never actually keyed on a client IP, and any single client could exhaust the shared bucket and 429 everyone else. The suite missed this because its `_FakeReq` defined `.client.host`, an API production does not have, and one test asserted the spoofable behavior as correct.

The gateway's `forward_to` had the same dead `request?.client` read and forwarded `X-Forwarded-For: unknown` to every downstream service.

## Fix

Standard trusted-proxy resolution, the same model as nginx `real_ip` and Starlette's `ProxyHeadersMiddleware`:

- New `[scale.microservices.rate_limit] trusted_proxies` key (IPs or CIDRs), parsed once at middleware build time with `ipaddress.ip_network`; invalid entries log a warning naming the bad value and are skipped. The default is `[]`: the header is never trusted unless the operator declares which proxies front the gateway.
- The resolver takes the peer from `client_host`. If the peer is not a trusted proxy, the header is ignored entirely. If it is, all `X-Forwarded-For` values are collected via `headers.getlist` and walked right to left, skipping trusted hops; the first untrusted entry (normalized for whitespace, `[v6]` brackets and a trailing port) is the client address as attested by the nearest trusted proxy, and everything the client wrote further left is unreachable. Unparseable hops and all-trusted chains fall back to the peer.
- `forward_to` now sources the forwarded address from `request.client_host`, so services receive a real address instead of `unknown`.
- The identity audit path (`identity/audit.jac` re-exports the resolver as `client_ip` for the forgot-password/verification handlers) keeps its one-argument call: `trusted_networks` defaults to no-trust, which resolves to the TCP peer, so audit events now record an unspoofable address there too (previously they logged whatever the client put in the header, or `unknown`).
- The tests are rebuilt on the real `Request`/`Headers` objects so the double cannot drift from production again. The test that asserted the spoof is replaced by its inverse, with new coverage for rightmost-untrusted selection, CIDR matching, all-hops-trusted fallback, port/bracket normalization, absent-header peer keying, and invalid config entries.

## Behavior after the fix

| Deployment | Before | After |
| --- | --- | --- |
| Exposed directly, honest client | all clients share one `unknown` bucket | keyed by real peer IP |
| Exposed directly, forged XFF | unlimited buckets, limiter bypassed | header ignored, keyed by peer IP |
| Behind ingress, `trusted_proxies` set | leftmost XFF, forgeable | rightmost untrusted hop, forgery-proof |
| Behind ingress, `trusted_proxies` unset | leftmost XFF, forgeable | shared but unforgeable bucket (fail-safe) |


jac/jaclang/scale/runtime/gateway/microservice_gateway.impl.jac
```diff
@@ -436,7 +436,7 @@ impl MicroserviceGateway.forward_to(
     }
 
     fwd_headers = self.build_forward_headers(request);
-    client_addr: str = getattr(request?.client, "host", "unknown");
+    client_addr: str = request.client_host or "unknown";
     fwd_headers["X-Forwarded-For"] = client_addr;
```
MusabMahmoodh: this stamps X-Forwarded-For on forward_to only - the other two build_forward_headers callers (handle_builtin_passthrough and aggregate_graph_data) still forward the client's own X-Forwarded-For verbatim, since build_forward_headers copies every non-hop-by-hop header. moving the stamp into build_forward_headers makes all three paths agree, and appending the peer instead of overwriting keeps the ingress-attested chain so a downstream service can still resolve the real client with its own trust config:

```jac
    prior = request.headers.get("X-Forwarded-For");
    peer = request.client_host or "unknown";
    fwd["X-Forwarded-For"] = f"{prior}, {peer}" if prior else peer;
```

(in build_forward_headers, dropping the two lines here)

jac/jaclang/scale/runtime/middleware/rate_limit.jac
```diff
@@ -146,6 +215,9 @@ def build_rate_limit_middleware(cfg: dict[str, any]) -> any {
     per_user_rpm: int = int(cfg.get("per_user_rpm", 0));
     burst: float = float(cfg.get("burst_multiplier", 2.0));
     exempt: set[str] = set(cfg.get("exempt_paths", []));
+    trusted_networks: list = _parse_trusted_proxies(
```
MusabMahmoodh, line 227: an already-deployed gateway behind an ingress upgrades into one shared bucket at per_ip_rpm for every user (the fail-safe row in your table), and the only symptom is 429s. worth a one-time logger.warning when the limiter drops an XFF from an untrusted peer while trusted_proxies is empty, so operators find the new key before their users do.

jac/jaclang/scale/runtime/middleware/rate_limit.jac
```diff
@@ -119,13 +120,81 @@ impl StateBackendRateLimiter.consume(key: str) -> tuple[bool, float] {
...
+}
+
+
+def _resolve_peer(request: any) -> str {
+    peer = request?.client_host or "";
+    if not peer {
+        peer = request?.client?.host or "";
+    }
+    return peer or "unknown";
+}
```
MusabMahmoodh: the PR body proves `.client` is dead on the serving Request, and the tests now run on the real Request - this fallback resurrects it. keeping a can't-happen fallback is the same silent-fallback class this PR is killing:

```suggestion
def _resolve_peer(request: any) -> str {
    return request?.client_host or "unknown";
}
```

jac/jaclang/scale/runtime/middleware/rate_limit.jac
```diff
@@ -119,13 +120,81 @@ impl StateBackendRateLimiter.consume(key: str) -> tuple[bool, float] {
...
+
+
+def _resolve_client_ip(request: any, trusted_networks: list | None = None) -> str {
```
MusabMahmoodh, line 179: with the one-arg call, identity audit now always records the TCP peer, so behind an ingress every forgot-password/verification event logs the ingress IP. intentional for now? the trust config exists after this PR - threading it into client_ip's callers looks like a small follow-up issue rather than leaving audit permanently peer-only.

jac/jaclang/scale/tests/microservices/test_rate_limit.jac
```diff
@@ -51,50 +54,14 @@ def _run(coro: object) -> object {
...
 def _req(
     path: str, headers: dict[str, str] | None = None, client_host: str = "1.2.3.4"
-) -> _FakeReq {
-    return _FakeReq(
-        url=_URL(path=path),
-        headers=_Headers(_d=headers or {}),
-        client=_Client(host=client_host)
-    );
+) -> Request {
+    h = Headers();
```
MusabMahmoodh, line 62: rebuilding the suite on the real Request/Headers is the right fix for the root cause here - the _FakeReq drift is exactly how the dead `.client` read went unnoticed, and inverting the spoof-asserting test closes that door.

jac/jaclang/scale/runtime/gateway/microservice_gateway.impl.jac
kashmithnisakya: moved the stamp into build_forward_headers with the append form, so all three paths agree now (adc5b6553)

jac/jaclang/scale/runtime/middleware/rate_limit.jac
kashmithnisakya: added a one-time warning when an xff gets dropped with no trusted_proxies configured, plus a test (adc5b6553)

jac/jaclang/scale/runtime/middleware/rate_limit.jac
kashmithnisakya: dropped it, the resolver reads client_host directly now so drift fails loud instead of silently (adc5b6553)

jac/jaclang/scale/runtime/middleware/rate_limit.jac
kashmithnisakya: intentional for now, opened #8135 to thread the trust config into the audit callers

jac/jaclang/scale/tests/microservices/test_rate_limit.jac
kashmithnisakya: thanks!

jac/jaclang/scale/runtime/middleware/rate_limit.jac
```diff
@@ -119,13 +122,90 @@ impl StateBackendRateLimiter.consume(key: str) -> tuple[bool, float] {
...
+    peer_ip = _parse_forwarded_ip(peer);
+    if peer_ip is None or not _is_trusted(peer_ip, trusted_networks) {
+        return peer;
+    }
+    getlist = request.headers?.getlist;
+    if callable(getlist) {
+        raw_values = getlist("X-Forwarded-For");
+    } else {
+        value = request.headers.get("X-Forwarded-For");
+        raw_values = [value] if value else [];
```
MusabMahmoodh: The real `Headers` always has `getlist` (the tests now run on it), so this `callable(getlist)` branch is the same can't-happen fallback class this PR just removed for `.client`. Worse, the fallback variant reads `.get`, which returns only the first matching header line, so if it were ever hit with multiple XFF headers it would resolve from the client-written line and be spoofable again. I'd call `getlist` directly and let drift fail loud:

```jac
raw_values = request.headers.getlist("X-Forwarded-For");
```

release_notes/unreleased/jaclang/8121.bugfix.md
```diff
@@ -0,0 +1 @@
+- **Fix: Gateway rate limiter honors X-Forwarded-For only from trusted proxies**: per-IP bucket keys now come from the TCP peer address unless the peer is listed in the new `[scale.microservices.rate_limit] trusted_proxies` key (IPs or CIDRs), in which case the rightmost forwarded hop not itself trusted is used, so clients can no longer bypass rate limits by forging the header; the peer fallback also reads the field the serving Request actually has, replacing the dead `.client` access that keyed all direct traffic to one shared "unknown" bucket.
```
MusabMahmoodh: The fragment describes the fix but not the operator action: anyone running behind an ingress upgrades into one shared bucket until they set `trusted_proxies`. One added sentence like "deployments behind a proxy or ingress must set `trusted_proxies` to keep per-client limiting" turns a support ticket into a config change.

MusabMahmoodh (commented):
jac/jaclang/scale/runtime/middleware/ws_forward.jac:38 (_select_handshake_headers):
The HTTP forward paths now stamp the peer onto the XFF chain in `build_forward_headers`, but this WS handshake path still forwards the client's own `X-Forwarded-For` verbatim, so downstream services get a fully client-forged chain with no gateway-attested hop. Can we do the same append here? The websocket object carries `client_host` too:

```jac
prior = client_headers.get("X-Forwarded-For") or client_headers.get("x-forwarded-for");
peer = getattr(websocket, "client_host", "") or "unknown";
out["X-Forwarded-For"] = f"{prior}, {peer}" if prior else peer;
```

(would need the peer threaded into `_select_handshake_headers` from `ws_proxy`; fine as a small follow-up if you'd rather keep this PR HTTP-only, but then let's file it.)

jac/jaclang/scale/runtime/docs.md:589:
This still documents the old behavior ("Per-IP key falls back from `X-Forwarded-For` (first hop) to `request.client.host`") and the toml example above it has no `trusted_proxies`. Worth updating in this PR since it now describes the exact spoofable behavior the PR removes.

jac/jaclang/scale/runtime/middleware/rate_limit.jac
kashmithnisakya: dropped the fallback, the resolver calls headers.getlist directly now (1354f1c67)

release_notes/unreleased/jaclang/8121.bugfix.md
kashmithnisakya: added the operator sentence to the fragment (1354f1c67)

jac/jaclang/scale/runtime/gateway/microservice_gateway.impl.jac
```diff
@@ -246,11 +246,16 @@ impl MicroserviceGateway.build_forward_headers(request: Request) -> dict[str, st
...
     for (key, value) in request.headers.items() {
-        if key.lower() not in HOP_BY_HOP_HEADERS {
+        if key.lower() not in HOP_BY_HOP_HEADERS and key.lower() != "x-forwarded-for" {
             fwd[key] = value;
         }
     }
 
+    peer: str = request.client_host or "unknown";
+    fwd["X-Forwarded-For"] = ", ".join(
+        request.headers.getlist("X-Forwarded-For") + [peer]
```
MusabMahmoodh, line 260: Appending rather than replacing is right, but it means services receive a chain whose left-hand entries are client-controlled. That's inert today because every service resolves with `trusted_proxies = []` and keys on its TCP peer. It stops being inert the moment #8135 threads trust config into the audit callers: the gateway's pod IP has to be in the *service's* `trusted_proxies` for the chain to be readable at all, and if it isn't, the service must keep ignoring the header entirely. Worth writing that into #8135 now, while the reasoning is fresh, so it doesn't get implemented as "read XFF[0]".

jac/jaclang/scale/tests/microservices/test_rate_limit.jac
```diff
@@ -239,22 +290,102 @@ test "X-Forwarded-For takes precedence over client.host for bucket key" {
...
+            "exempt_paths": []
+        }
+    );
+    async def next_handler(req: object) -> object {
+        import from jaclang.runtimelib.serving.datatypes { JSONResponse }
+        return JSONResponse({"ok": True});
+    }
+    handler = MemoryHandler(capacity=100);
+    _rl_logger.addHandler(handler);
+    _untrusted_xff_warned.clear();
```
MusabMahmoodh, line 389: `_untrusted_xff_warned.clear()` runs before the assertions but not in the `finally`, so the Event stays set for whatever runs next in the process. Nothing else asserts on it today, so this is only a trap for the next warning test someone adds - moving the `clear()` into the `finally` as well costs one line.

MusabMahmoodh (commented):
Re-reviewed after the two rounds of follow-ups. All seven items landed:

- XFF stamped on one path only -> moved into `build_forward_headers` in the append form, with `forward_to`'s local stamp deleted, so passthrough and `aggregate_graph_data` agree with it.
- Silent upgrade into one shared bucket -> `_warn_dropped_xff_once`, with a test.
- `request?.client` fallback -> gone; the resolver reads `client_host` directly.
- `callable(getlist)` fallback -> gone. I checked that `getlist` exists on both `Headers` implementations (datatypes.jac:39 and :104), so there's no path where the dropped fallback was load-bearing.
- Audit permanently peer-only -> tracked as #8135.
- docs.md:589 -> rewritten, including the operator action.
- Fragment -> operator sentence added.
- ws handshake -> `_select_handshake_headers` takes the peer and appends it, assertion tightened to the exact chain.

I re-walked the new resolver: right-to-left with trusted hops skipped, unparseable hop or all-trusted chain falls back to the peer, `strict=False` on the CIDR parse, and `IPv4Address in IPv6Network` returning False rather than raising - every branch fails safe toward the TCP peer. `_parse_forwarded_ip`'s port/bracket handling is right for bare IPv6 too (`::1` has two colons, so the port strip doesn't fire).

This is the standard trusted-proxy model rather than a workaround, and it deleted two dead fallbacks on the way in rather than leaving them - I found nothing dead left behind. Can this break something already running: yes, and it's the intended break, now documented in three places - a gateway behind an ingress with no `trusted_proxies` set collapses to one shared per-IP bucket until the operator sets the key, and the warning makes that findable. I'd call this production ready.

Approving. One forward-looking note and one test nit.

jac/jaclang/scale/runtime/gateway/microservice_gateway.impl.jac
kashmithnisakya: written into #8135's body as a trust-chain constraint section: resolve right-to-left via _resolve_client_ip, never XFF[0], and the gateway address has to be in the service's trusted_proxies or the header stays ignored entirely

jac/jaclang/scale/tests/microservices/test_rate_limit.jac
kashmithnisakya: moved the clear() into the finally as well so the Event is reset on exit (a042ce648)

MusabMahmoodh (approved):
Re-review at head. Since my 08-15 pass the only change is the one I asked for, `_untrusted_xff_warned.clear()` moved into the `finally` so the warning test leaves no process-global state behind, plus two merges from main. Everything else stands: the resolver fails safe toward the TCP peer on every branch, both dead fallbacks (`.client`, the `callable(getlist)` variant) went out rather than being left beside the new code, and the suite runs on the real `Request`/`Headers` so the double cannot drift from production again. CI is green on this head.

Approving. Two forward-looking things, neither blocking.

**1. The deploy knows about the ingress before the operator does.** `rate_limit.jac`, `_warn_dropped_xff_once`.

The one-time warning is the right runtime signal, but it only fires once traffic is already keyed on the wrong address. jac-scale generates the ingress itself, so the deploy can see the whole picture: rate limiting enabled, an ingress configured, `trusted_proxies` empty. A deploy-time warning naming the key reaches the operator while they can still act, instead of after the first 429. Worth a follow-up on the manifest builder?

**2. The key's trip from jac.toml to the middleware is untested.** `test_rate_limit.jac`, "distinct clients behind a trusted proxy get distinct buckets".

Every new test hands `trusted_proxies` straight to `build_rate_limit_middleware`, so a break anywhere in `normalize_microservices_config` -> `get_microservices_config` -> `install_middleware` would leave all of them green while the key is silently ignored in production. That is the same shape as the `[plugins.scale.*]` section that went quiet in #7098. One test that runs a jac.toml through `normalize_microservices_config` and asserts `trusted_proxies` survives into `rate_limit` closes it for good.

MusabMahmoodh (approved):
Re-reviewed at `a151c48b2`. Approving. All eight prior points are resolved, and I appreciate the trust-chain constraint going into #8135's body rather than just being filed.

I also checked this against #8300 and #8333, since all three touch the same config surface: `trusted_proxies` does go through the loader contract properly, emitted at `config_loader.impl.jac:602-604` into the same normalized block as `per_ip_rpm` and consumed by that exact name at `rate_limit.jac:223-225`. And it is independent of #8333: `rate_limit.jac:227` is byte-identical to main.

Three follow-ups, none blocking.

**`rate_limit.jac:257`** - the warning only fires when `trusted_proxies` is empty, so the misconfiguration that worries me more stays silent: an operator who sets the wrong CIDR (ingress moved subnets, or they guessed the pod range) gets the same one-shared-bucket collapse while believing they configured it. Same one-time signal, one branch wider: resolve the client IP first and warn when it comes back equal to the peer address on a request that carried an `X-Forwarded-For`.

**`plan.jac:283`** - this is the line I added in #8300 so a dry run can confirm the limit before it deploys, and `trusted_proxies` is now the key whose absence silently changes what that limit *means*. Worth appending a trust clause to the rendered line so the dry run says which of the two regimes you are getting.

**`tests/`** - `grep -rn trusted_proxies jac/jaclang/scale/tests/` is empty. A break anywhere in `normalize_microservices_config` -> `get_microservices_config` -> `install_middleware` leaves all fourteen tests green while the key is ignored in production. That is exactly the shape of #8333, and of the `[plugins.scale.*]` section that went quiet in #7098. One case in `tests/misc/test_config_validation.jac` closes it.

Issue #8104, opened 2026-08-13
bundle membership WI-1: host client build runs on the filtered stage, not the raw tree


MusabMahmoodh:
---

## Reframed 2026-08-13 per marsninja's review on #8077 (move 1: the stage is the manifest)

This is no longer "run the client build somewhere else"; it is hoisting stage ownership out of `_precompile_app` (scale/injector/bundle.jac:416) into a deploy-scoped object, with everything downstream seeing only the stage:

- Stage phases in order: `jac install` (cwd=stage), client build (cwd=stage), seal. The deploy stops mutating the user's working tree (today: `jac install` side effects plus `.jac/client/dist` written into their repo and allowlisted back into the bundle).
- `pack_jab` collapses to "tar the stage". This deletes the second walk, the second `source_filter_for`, the duplicate nested-toml emit, and the live seal-window drift (the tar currently re-walks the working tree after a seal that can run 1800s + 900s vendoring; anything changed in that window ships unsealed).
- Membership becomes a place you cannot route around, not a predicate call sites must remember to make.

Additional acceptance from the review:

- [ ] The npm cache is carried into the stage (or the stage dir is stable per project); the deploy-time delta vs today is measured and posted in the PR
- [ ] A file touched in the working tree during the seal does not reach the .jab (seal-window drift test)
- [ ] The user's working tree is byte-identical before and after a deploy

marsninja:
Closing: no activity in 10+ days. Reopen if still relevant.

Issue #8077, opened 2026-08-12
(design) bundle membership from first principles: one declared source of truth for what ships to a pod


MusabMahmoodh:
> **Status 2026-08-13:** #8097 (@marsninja) implements the host-side half
> of this design: a declared `BundleSourceFilter` read by the staging
> copy, the .jab pack, and the nested-toml walk, with the `.jacignore`
> loader placed canonically in `jac0core/helpers.jac` beside the matcher
> `jac check` already uses. That settles the class framing, the placement
> question, the `.jacignore` semantics, the baseline floor, and the
> fail-loud-on-parked-entry behavior (boxes ticked below with notes).
> Still open and **assignable**: WI-1 (client build runs on the filtered
> stage) and WI-2 (member list ships in the .jab, pod boots from it, the
> entrypoint `find` dies) - see "Remaining work" below.

> **Update 2026-08-13 (2):** marsninja's design review below is folded in
> wholesale: I0 added above the invariants, the derivation count is eight,
> "the stage is the manifest" replaces the `manifest.json` mechanism, and
> ship vs must-compile is split (a `.jacignore` is no longer required to
> survive a stale broken module). WI-2 re-scoped to S, WI-4/WI-5 added.

## Tracking

- [ ] #8044 - the instance (@MalithaPrabhashana's finding and provenance forensics; closes when #8097 merges)
- [ ] #8104 - WI-1: deploy-scoped stage, client build as a stage phase, pack_jab tars the stage (size M)
- [ ] #8105 - WI-2 re-scoped: service file resolved host-side, JAC_SV_FILE always set, the pod `find` deleted (size S)
- [ ] #8106 - WI-3: retire jacBuilder's park/restore workaround, the class acceptance test (size S, after WI-1/2)
- [ ] #8171 - WI-4: seal the closure of declared entries hard, ship the rest inert (size M, after WI-1)
- [ ] #8172 - WI-5: split _EXCLUDE_DIRS into security floor and logged heuristic (size S)

## Summary

Every stage of a `--scale` deploy decides for itself which files are part
of the app: the staging copy, the .jab pack, the seal walk, the host client
build, and the in-pod boot each derive membership ad hoc from the raw
filesystem, with three different hardcoded exclude vocabularies between
them and no project-level input. `.jacignore` is consulted by none of them
(its only consumer in the repo is a shell glob in CI). This RFC proposes a
single compiler-derived bundle manifest that every consumer reads, with
[...]

Issue #8010, opened 2026-08-10
Jac Scale: 0.34.13 host client build fails (fatal) because the deploy never installs the app's [dependencies] into the host build env


udithishanka:
## Summary

Since 0.34.13, a permanent deploy hard-fails at the host client-bundle build for any app whose client build transitively imports a Python module backed by a declared `[dependencies]` package. The deploy runs `jac build <entry> --client web` on the host (`manifest_builder.jac:166`) **without first `jac install`-ing the app's `[dependencies]`**, so the build's module traversal hits the missing package and dies with `ModuleNotFoundError`. #7936 then makes that a fatal deploy error ("the pod never rebuilds"), where it used to be a tolerated warning. Net effect: an app that deployed fine on 0.34.11 cannot deploy at all on 0.34.13.

This is a regression in reachability + fatality, not a new build defect: the host build never had the app's deps, but on 0.34.11 the failure was swallowed and the pod rebuilt on boot (where the deps ARE present, via the vendored wheels). #7936 removed that fallback without adding the missing `jac install`.

## Environment

- jac **0.34.13** (platform image + pinned deploy binary), `production-cluster` (jachammer), EKS 1.35, us-east-2.
- Minimal repro also run locally with the released `jac-0.34.13-macos-aarch64`.
- Regressing change: **#7936** (`7de5533c`, merged 2026-08-06) "fix(scale): fail the deploy fast on a broken client build; the pod never rebuilds", shipped in 0.34.13 (published 2026-08-07). Its own problem statement quotes the pre-0.34.13 non-fatal message this issue is about.

### 1. Minimal app, the exact command the deploy runs

`jac.toml` declares a Python dep and a client; a server module imports the dep; the client entry imports the server module (the common jac-client shape):

```toml
# jac.toml
[project]
name = "clientdep-repro"
entry-point = "main.jac"
jac-version = "==0.34.13"
[dependencies]
"psycopg[binary]" = ">=0.0.0"
[dependencies.npm]
react = "^18.2.0"
react-dom = "^18.2.0"
```
```jac
// db.jac        (server module, imports a [dependencies] package)
import from psycopg { connect }
def find_user(email: str) -> str { return email; }

// main.jac      (client entry, imports the server module)
import from db { find_user }
with entry { print("server entry"); }

// app.cl.jac    (a client component)
def:pub App() -> any { return <div>hello</div>; }
```

Run the deploy's exact host-build command **without** `jac install` (which is what the deploy does):

[...]

PR #8128, opened 2026-08-13
fix(scale): run the host client build on the bundle's filtered stage, not the raw tree

kashmithnisakya:
Part of #8104 (the WI-1 build-on-stage half; stage ownership, `pack_jab` tarring the stage, and the byte-identical working tree stay open on the issue). Part of #8077 (bundle membership: one declared source of truth), WI-1.

## Problem

#8097 gave the deploy a declared `BundleSourceFilter`, but the host client build still ran on the raw project tree: `_build_client_bundle` invoked `jac install` and `jac build --client web` with `cwd=project_dir`. Files the bundle strips still reached the client build (`.env`, the unsanitized `jac.toml`, `.jacignore`d trees, key material), and the resulting `.jac/client/dist` is force-shipped into the `.jab` by `_is_bundle_required`. The generated Vite config loads `.env` from the project root (`envDir: projectRoot` in `vite_bundler.impl.jac`), so values the bundle strips could be inlined into the served JS, where they can never be audited after the fact.

## Fix

The client build now runs inside the same staged, filtered copy that defines the bundle's membership, so host build inputs equal pod inputs by construction:

- `_build_client_bundle` stages with `_stage_app(base, sanitized_jac_toml(base))` (the #8097 machinery, reused as is), runs `jac install` and the build with `cwd=stage`, gates on `index.html` (#7936, unchanged), copies the dist back to `<project>/.jac/client/dist` where `pack_jab` ships it from, and removes the stage in a `finally`.
- A client entry parked by `.jacignore` (or otherwise outside membership) fails the deploy immediately, naming the file, the matching pattern, and the fix; previously that case silently skipped the build.
- On a failed build, a membership diagnosis maps path/module tokens in the build error to raw-tree files the filter excludes and appends the file, the exclusion reason, and the fix. `BundleSourceFilter` gains `matching_ignore_pattern` and `exclusion_reason` to support this.
- #8013's dependency install stays in the build environment (now the stage), and the missing-install cause chain on a failed build (#8010) is unchanged.
- A `.env` (or `.env.local` / `.env.production` / `.env.production.local`, the files a production Vite build loads) that declares `VITE_`-prefixed keys is called out by file and key before the build, since `envDir` is the stage and those now compile to `undefined` rather than failing the build. The warning points at `[client.vite.define]` for public build-time values; secrets never reach the client. Keys only, never values.
- A `[scale.microservices.client] entry` the user declared that names a missing file fails the deploy (previously a silent skip that shipped a gateway with nothing to serve). An undeclared, guessed default that is absent still skips the build, but says so.
- The client entry is read once, from the raw `[scale.microservices.client]` table, by `_client_entry_setting`, which returns both the resolved entry and whether the user declared it; neither reader swallows exceptions any more, so a `jac.toml` the loader rejects fails the deploy instead of downgrading a declared entry to a guessed one and skipping. The remaining skip (entry unset or `false`, no client source) says so.
- The membership diagnosis only names exclusions a user can act on (`.jacignore` patterns, `.env`, the raw `jac.toml`, key material); build and cache trees (`node_modules/`, `dist/`, `.jac/`) that vite and bun print on most failures are filtered through `BundleSourceFilter.is_build_tree`, so it never advises dropping a tooling-tree exclusion.
- The built dist is copied to `<project>/.jac/client/dist.incoming` and swapped in with `os.replace`, so an interrupt or a full disk mid-copy leaves the previous dist in place rather than none.
- Dry-run behavior is unchanged: no staging, no install, no build.

## Acceptance

- Host client build inputs == pod bundle inputs over the #8097 fixture shape: `test_client_build_stage.jac` records the build's working tree through the `JAC_EXECUTABLE` seam and asserts it equals `bundle_source_files(base)` plus the sanitized `jac.toml` (parked `docsite/` and `evaluation/` absent, `.env` absent, `[scale.secrets]` stripped).
- An asset outside membership fails the build loud on the host with the file and the matching `.jacignore` pattern named, never in the pod: covered for both a parked entry and a parked reference.
- #8010's scenario still fails loud with the actionable message: "a client build that dies on a missing import names the failed install" passes unmodified.
- The motivating case end to end with the real toolchain: a client that reads `import.meta.env.VITE_API_KEY` gets `sk-live-8104` inlined by a raw-tree control build, and the deploy build's shipped dist never contains it (`test_client_build_stage.jac`, "a VITE_ value in .env never reaches the shipped dist").
- The deps test observes the install where the venv now lives: a `JAC_EXECUTABLE` wrapper hands every call to the real jac and, after `install`, records that `toposort` landed in the stage's `.jac/venv`, so the guarantee is the install itself, not "the build did not raise".
- Regression floor: `test_client_build_fail_fast.jac` (9 passed), `test_client_build_deps.jac` (4 passed), `test_bundle_jacignore.jac` (6 passed), `test_client_build_stage.jac` (6 passed, new). The #7936 broken-dist gate is unchanged.

One existing-test edit: the deps test's venv-location assertion (`<project>/.jac/venv`) is replaced by asserting the build succeeds. The venv now lives in the ephemeral stage, and the build only succeeds if `[dependencies]` installed there, so the guarantee it protected still holds.

## Deploy-time delta (measured)

Host client build wall-clock, this machine (M-series Mac), global pip and bun caches warm; `main` builds on the raw tree with `.jac/venv` and `node_modules` reused across deploys, this branch builds in a fresh stage every deploy:

| app | pre-PR raw tree, warm | this PR, fresh stage | delta |
|---|---|---|---|
| 2 py deps + 2 npm | 4.8s | 9.2s | +4.4s |
| 10 py deps + 6 npm | 5.2s | 11.9s | +6.7s |
| 14 py deps (numpy, pandas, fastapi) + 6 npm, `jac install` only | 1.2s | 9.8s | +8.5s |

The delta is the venv creation plus pip install-from-cache; bun install from its global cache is under a second, and one staging walk is 0.2s at 10k tree entries (node_modules present), so staging the tree twice per deploy is not where the time goes. First-ever (cold cache) builds are 16-18s either way. That per-deploy cost is the price of host inputs equalling pod inputs by construction, and it is bounded by the global caches; a stage kept alive across deploys would trade it for a per-project build cache with its own staleness and locking, which is #8104's "stage is the manifest" reframe rather than this fix.

## Behavior change

Client builds that latently depended on raw-tree files start failing loud at deploy time, with the file, the exclusion, and the fix named (per the #8077 risk register). `.env` values no longer reach the client build: server-side values belong in `[scale.secrets]` or pod env, and public build-time client values under `[client.vite.define]`; a `.env` with `VITE_` keys is called out at deploy time so this cannot go silent. A declared client entry that names a missing file now fails the deploy. The stage starts without `.jac/venv` and `node_modules`, so pip and bun re-resolve per deploy (numbers above).


jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac
```diff
@@ -233,76 +321,97 @@ obj ManifestBuilder {
...
-            attempt += 1;
-            try {
-                result = subprocess.run(
-                    [jac_bin, "build", client_entry, "--client", "web"],
-                    cwd=project_dir,
-                    capture_output=True,
-                    text=True,
-                    timeout=900,
-                    env=jac_env
+        stage_root = _stage_app(base, sanitized_jac_toml(base));
```
MusabMahmoodh, line 446: The stage has no `.env` and `envDir` is the stage, so `import.meta.env.VITE_X` compiles to `undefined` rather than failing the build - the one exclusion whose symptom is a blank page at mount instead of a build error. `_membership_diagnosis` can't reach it because it only runs on a non-zero exit. Can we catch it here, where the raw tree is still in hand?

```jac
        env_file = base / ".env";
        if env_file.is_file() and "VITE_" in env_file.read_text(encoding="utf-8") {
            raise ValueError(
                f"'{env_file.name}' declares VITE_-prefixed values and the client "
                f"build no longer reads it, so those become undefined in the served "
                f"JS. Move them to [scale.secrets] or pod env."
            );
        }
```

A `console.warning` naming the keys would do. Silently is the one way this must not go.

jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac
```diff
@@ -203,27 +252,66 @@ obj ManifestBuilder {
...
+        }
+        if ".." in entry_rel.parts {
+            raise ValueError(
+                f"The client entry '{client_entry}' reaches outside the project "
+                f"tree, so the bundle cannot ship it. Point "
+                f"[scale.microservices.client] entry at a path relative to "
+                f"the project root."
+            );
+        }
+        if not (base / entry_rel).is_file() {
```
MusabMahmoodh, line 392: Two lines above the new loud raise for a parked entry, a *missing* entry still returns silently and ships a gateway with nothing to serve. Same defect class this PR is closing, same function, and now inconsistent: `entry = "man.jac"` (typo) is quiet, `entry = "main.jac"` (parked) is loud. Worth making both raise here.

jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac
```diff
@@ -233,76 +321,97 @@ obj ManifestBuilder {
...
-                    capture_output=True,
-                    text=True,
-                    timeout=900,
-                    env=jac_env
+        stage_root = _stage_app(base, sanitized_jac_toml(base));
+        stage = str(stage_root / "app");
+        dist = os.path.join(stage, ".jac", "client", "dist");
+        try {
+            shutil.rmtree(dist, ignore_errors=True);
+            dep_error = self._install_app_deps(stage, jac_bin, jac_env);
```
MusabMahmoodh, line 451: `_install_app_deps` now targets the ephemeral stage, so `.jac/venv` is created and populated from scratch on every deploy instead of being reused - and on top of that the tree is staged twice per deploy, once here and again in `pack_jab`. Have you measured the deploy delta on a real app? If it's material, keeping this stage alive and letting the pack reuse it pays for both at once.

jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac
```diff
@@ -233,76 +321,97 @@ obj ManifestBuilder {
...
-                    cwd=project_dir,
-                    capture_output=True,
-                    text=True,
-                    timeout=900,
-                    env=jac_env
+        stage_root = _stage_app(base, sanitized_jac_toml(base));
+        stage = str(stage_root / "app");
+        dist = os.path.join(stage, ".jac", "client", "dist");
+        try {
+            shutil.rmtree(dist, ignore_errors=True);
```
MusabMahmoodh, line 450: Dead line: `.jac` is in `_EXCLUDE_DIRS` (bundle.jac:28), so a freshly staged tree can never contain `.jac/client/dist`. Worth dropping rather than leaving a reader wondering what state it defends against.

Related, in `_membership_diagnosis`: `cand.rstrip(".,;:")` can only ever strip `.`, since the regex `[\w.\\/-]+` never matches `,`, `;` or `:`. And more generally - that function is a regex heuristic over build output that stats every token, so it will silently miss any error message that names a file in a form it doesn't recognise. Fine as a hint, but the error text should not read as authoritative ("these files it references never ship") when it is best-effort.

jac/jaclang/scale/tests/deploy/test_client_build_stage.jac
```diff
@@ -0,0 +1,256 @@
...
+}
+
+
+def _with_recording_build(
+    project_dir: str, build_exit: int = 0
+) -> tuple[(Exception | None), str] {
+    tmp = tempfile.mkdtemp(prefix="client-stage-rec-");
+    script = _recording_jac(tmp, tmp, build_exit);
+    saved = os.environ.get("JAC_EXECUTABLE");
+    os.environ["JAC_EXECUTABLE"] = script;
```
MusabMahmoodh, line 145: Recording the build's working tree through `JAC_EXECUTABLE` with a real script is the right seam (real subprocess, no mock), and `staged == bundle_source_files(base)` is a much stronger assertion than "the parked file is absent".

The gap is the motivating case: nothing shows that a client which *reads* `VITE_API_KEY` no longer gets `sk-live-8104` inlined into its dist - the suite asserts the mechanism (`.env` not staged), not the outcome. One real (non-recording) build asserting `sk-live-8104` appears nowhere under `.jac/client/dist` would pin the actual security property, and would have surfaced the silent-undefined case above.

Minor: `_with_recording_build` mutates `os.environ["JAC_EXECUTABLE"]` process-wide. It restores in a `finally`, but any parallel test in the same process sees it.

jac/jaclang/scale/tests/deploy/test_client_build_deps.jac
```diff
@@ -72,25 +71,22 @@ def _project(dependencies: str, server_import: str) -> str {
...
         ).is_file();
-    } finally {
         shutil.rmtree(project_dir, ignore_errors=True);
     }
 
-    assert installed , (
-        "the deploy never installed the app's [dependencies] into the host "
-        f"build env, so any client entry reaching a {APP_DEP}-backed module "
-        "cannot build here"
+    assert raised is None , (
```
MusabMahmoodh: This swaps a specific guarantee (the app's dependencies landed in the build env) for a general one (the build didn't raise). The build failing on a missing import is a real proxy, so I'm not blocking - but it now also passes if the build succeeds for some unrelated reason, and the test name still promises the install. Asserting the dep is importable from the stage, or asserting the install step ran at all, keeps the original guarantee where the venv moved.

MusabMahmoodh (commented):
Building on the same staged copy that defines membership is the right fix - host inputs equal pod inputs by construction rather than by two lists agreeing - and reusing `_stage_app`/`sanitized_jac_toml` as-is is exactly right.

Can this break something already running: yes, and it's the one case that does *not* fail loud. `.env` is in `_EXCLUDE_FILES` (bundle.jac:41) and `envDir: projectRoot` (vite_bundler.impl.jac:828) now points at a stage with no `.env`, so every `import.meta.env.VITE_*` compiles to `undefined`. Vite does not error on a missing env var, so the build is green, the deploy is green, and the app breaks in the browser with nothing in the deploy output pointing at the cause. The intended breaks (parked entry, parked reference) do fail loud and are fine.

Is it production ready as a whole: not quite, for two reasons - that silent case, and the deploy-time cost nobody has measured yet (comment 3). The rest is solid. Dead code I found is in comment 4.

jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac
kashmithnisakya: Done in 113b5a6: before staging, the deploy scans the .env files a production Vite build loads (.env, .env.local, .env.production, .env.production.local) and warns per file with the VITE_ key names (never values), pointing at [client.vite.define] for build-time values since [scale.secrets]/pod env cannot reach a host-built client; covered by a recording test and by a real-build test that shows the raw tree inlines sk-live-8104 and the shipped dist never contains it.

jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac
kashmithnisakya: Done in 113b5a6: an entry the user declared in [scale.microservices.client] that names a missing file now raises with the path and the entry = false way out, while an undeclared, guessed default that is absent still skips but prints why, since raising on a heuristic guess would break API-only apps that merely have a components/ dir; both paths are tested.

jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac
kashmithnisakya: Measured and posted in the PR body: +4.4s / +6.7s / +8.5s per deploy for a 2, 10 and 14 Python-dep app (warm global caches), all of it venv creation plus pip install-from-cache; bun is under a second and one staging walk is 0.2s at 10k tree entries, so the second staging is not where the time goes, and keeping a stage alive across deploys would trade this for a per-project build cache with its own staleness/locking, which is #8104's stage-is-the-manifest reframe rather than this fix.

jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac
kashmithnisakya: The rmtree is live: excludes() checks _is_bundle_required before _EXCLUDE_DIRS and .jac/client/dist is bundle-required, so a prior deploy's dist is copied into the stage and this line is what stops a build that emits nothing from passing the index.html gate on last deploy's output (verified with bundle_source_files on such a tree); the rstrip is now just "." and the diagnosis header reads "files named in its output" with the footer pointing at [client.vite.define] rather than [scale.secrets] for client build-time values (113b5a6).

jac/jaclang/scale/tests/deploy/test_client_build_stage.jac
kashmithnisakya: Added in 113b5a6: one real (non-recording) build where main.jac reads import.meta.env.VITE_API_KEY, with a raw-tree control build asserting sk-live-8104 IS inlined and the deploy build asserting it appears in no file under the shipped dist, plus the warning naming the key; on the env mutation, jac test forks one process per worker and runs a file's tests serially, and test_client_build_fail_fast.jac (#8037) already uses the same seam the same way, so it is process-local.

jac/jaclang/scale/tests/deploy/test_client_build_deps.jac
kashmithnisakya: Done in 113b5a6: the test now goes through a JAC_EXECUTABLE wrapper that hands every call to the real jac and, after install exits 0, records the cwd and the toposort entries in that tree's .jac/venv site-packages, so it asserts the install ran, ran in the stage rather than the raw tree, and landed the dep there, with the build result checked after that rather than standing in for it.

MusabMahmoodh (commented):
Re-review at head (113b5a6 plus the merge from main). All six from 08-15 landed: the silent `VITE_*` to `undefined` case now warns per env file with the key names and never the values, a declared-but-missing client entry raises while an undeclared guess still skips and says so, the deploy-time delta is measured and posted, the deps test asserts the install ran in the stage and landed `toposort` there, and the motivating case is now a real build with a control leg. You were right about the `rmtree` and I was wrong: `_is_bundle_required` runs before `_EXCLUDE_DIRS`, so a prior dist does reach the stage and that line is load-bearing.

The control-build-plus-deploy-build pair asserting `sk-live-8104` is inlined on the raw tree and absent from the shipped dist is the strongest test I have read in this stack, precisely because it cannot pass vacuously.

Four things, none blocking.

**1. The membership diagnosis will accuse node_modules.** `manifest_builder.jac`, `_membership_diagnosis`.

`_EXCLUDE_DIRS` holds `node_modules`, `dist` and `build`, and vite/bun print those paths relative to the build cwd on almost every real failure, so on a raw tree that has ever been built locally I think this prints "these files named in its output never ship: node_modules/vite/... Drop the exclusion or remove the reference." That is wrong advice at the exact moment someone is stuck. Can it skip the build and cache trees and keep the surprising exclusions (.jacignore, key material, .env)?

```jac
reason = excluded.exclusion_reason(rel);
if "build or cache tree" in reason {
    continue;
}
```

**2. "Fixes #8104" while the deploy still writes into the user's repo.** End of `_build_client_bundle`, the `shutil.rmtree(target)` plus `copytree`.

#8104 as reframed asks for stage ownership: `pack_jab` collapses to "tar the stage", and the working tree is byte-identical before and after a deploy. This still deletes and rewrites `<project>/.jac/client/dist` on every deploy, which your reply on the cost thread agrees is the reframe's job rather than this PR's. That split is fine, but then this is "Part of #8104", not "Fixes", or the issue closes with its acceptance list unmet.

**3. The new fail-loud path can still go quiet.** `_declared_client_entry`.

`except Exception -> ""` makes a declared entry look undeclared on any config read hiccup, which drops the typo case straight back into the silent skip this PR just removed: it prints "no client entry is declared" and ships a gateway with nothing to serve. `_client_entry` reads `self._ms_cfg()`; can this use the same path and let a config error surface?

**4. The control build has no retry, in a test for code that retries.** `test_client_build_stage.jac`, "a VITE_ value in .env never reaches the shipped dist".

The control `subprocess.run([... build --client web])` asserts `returncode == 0` with no retry, while the code under test retries exactly this on `_is_transient_client_build_failure`. One flaky bun install fails the test for a reason that has nothing to do with the assertion. Worth reusing that same check, or skipping the leg when the control cannot build at all?

MusabMahmoodh (commented):
Re-reviewed at head. Approving with comments. All six prior findings are resolved, two of them by you correcting me: the `rmtree(dist)` is live because `excludes()` checks `_is_bundle_required` first (`bundle.jac:103-106`), and the `os.environ` mutation is process-local given how `jac test` forks. Both stand.

The real-toolchain leg with a raw-tree control build is the strongest test in this stack: asserting the secret **is** inlined on the control before asserting it is absent from the deploy dist means the negative can never quietly go vacuous. That is the pattern I would like to see more of in this suite.

Four new things at this head.

**1. `manifest_builder.jac:87` - the diagnosis will accuse `node_modules`**

`_EXCLUDE_DIRS` holds `node_modules`, `dist` and `build` (`bundle.jac:27-38`), and vite and bun name paths inside those in most real build failures. On any tree that has been built locally I think this prints something like `node_modules/vite/...: 'node_modules/' is a build or cache tree the deploy never ships`, followed by "Drop the exclusion or remove the reference" - which is wrong advice at the exact moment the message is read closely. Filtering the build/cache-tree reasons out and keeping only the surprising exclusions (`.jacignore`, key material, `.env`) would leave the diagnosis saying only things a user can act on.

**2. `manifest_builder.jac:687` - there is a third path and it says nothing**

You made two paths loud and a third stayed silent: `_client_entry` returns `""` on any config-read error (`:703`), and `_build_client_bundle` returns at `:343` without printing, so a `jac.toml` the loader chokes on ships a gateway with nothing to serve and no message at all. The new `_declared_client_entry` has the same shape - `except Exception` to `""` downgrades a declared entry to a guessed one, which puts the typo case straight back into the skip you just removed. Can the new one at least let the error out?

**3. `manifest_builder.jac:505-510` - the copy-back is not atomic**

`rmtree(target)` then `copytree(dist, target)` means an interrupt or a full disk between the two leaves the user's repo with no `.jac/client/dist` at all, where before this the deploy built in place. Staging to a sibling and swapping keeps the tree as it was found:

```jac
staged_dist = target + ".deploy-tmp";
shutil.rmtree(staged_dist, ignore_errors=True);
shutil.copytree(dist, staged_dist);
shutil.rmtree(target, ignore_errors=True);
os.replace(staged_dist, target);
```

Worth a release-note line either way: the deploy now deletes and rewrites `<project>/.jac/client/dist` rather than building into it.

**4. `test_client_build_stage.jac`, the VITE_ leg - the control build has no retry**

The control `subprocess.run([jac_executable(), "build", "main.jac", "--client", "web"])` asserts `returncode == 0` with no retry, while the code under test retries exactly this case on `_is_transient_client_build_failure`. One flaky bun install fails the test for a reason unrelated to what it asserts, and this is the leg that leans on the network hardest. Reusing that same predicate, or skipping the leg when the control cannot build at all, would keep it honest.

## Merge order with #8170

These two do not contradict each other. A `merge-tree` of both heads gives three conflict hunks, all in `bundle.jac`, all in the duplicated `matching_ignore_pattern` / `exclusion_reason` pair (same logic, different message strings) - and I would keep this PR's wording, since it names the fix. #8170's tests survive that: they assert only the `".jacignore"`, `"dist"` and `"entry module"` substrings.

They reinforce each other too. #8170 puts service resolution at `generate_manifests:349`, ahead of the client build at `:165` that this PR makes considerably more expensive, so an unresolvable module fails before the venv, pip and bun cost is paid.

Suggested order is this one, then #8170, and nothing in between - the duplicated methods are the whole conflict and every commit landing between them makes that collapse harder to eyeball.

MusabMahmoodh (approved):
Delta review since Aug 19. All six of my earlier findings stayed resolved and the two new commits are both improvements. Approving.

Worth saying because it reads the other way at a glance: **the `is_build_tree` skip in `_membership_diagnosis` is a strengthening, not a weakening.** Without it, `node_modules` and `dist` are under `_EXCLUDE_DIRS`, so any failing client build would have blamed the one tree that is *supposed* to be absent from the bundle and present in the stage. The diagnosis is only useful if it points at files the user actually parked, and now it does.

I also traced the atomic swap and it holds: `incoming` and `outgoing` are siblings of `target` under `.jac/client`, so `os.replace` stays on one filesystem; `target` is vacated into `outgoing` before the second replace, so the directory-onto-non-empty-directory case cannot fire; and the `rmtree(outgoing, ignore_errors=True)` at entry means a process killed mid-swap self-heals on the next deploy rather than leaving a half-state. That is a real improvement over the rmtree-then-copytree it replaces.

**One thing.** `_membership_diagnosis` stops at three files:

```jac
if len(lines) >= 3 {
    break;
}
```

An app that parked a config directory can easily reference six or eight files from it. The operator fixes the three that were named, redeploys, waits out another staged build, and gets three more. Since the whole point of this diagnosis is to turn a silent wrong-dist into one loud actionable message, the truncation is the one place it stops being actionable in a single pass. Cheapest fix is to keep counting and say so:

```jac
if not lines {
    return "";
}
shown = lines[:3];
if len(lines) > 3 {
    shown.append(f"  ... and {len(lines) - 3} more");
}
```

Not blocking - the first three still get someone unstuck.

Issue #8050, opened 2026-08-11
SERIALIZABLE by default exhausts predicate-lock shared memory: every query fails with 53200, health checks included, fleet cannot recover


MalithaPrabhashana:
### Summary

> Severity: production stability. After a few minutes of ordinary application load, the provisioned Postgres runs out of predicate-lock shared memory and then **every** query fails, including the app's own health checks. The fleet goes unready, restarts, and does not recover, because the restarted pods hit the same error immediately.

`PgStore.begin()` opens transactions as `SERIALIZABLE` by default:

```jac
impl PgStore.begin(serializable: bool = True) -> None {
    ...
    self._run(f"BEGIN ISOLATION LEVEL {level}", {});
```

SERIALIZABLE acquires an SIReadLock per row/page read. An application whose reads traverse many rows per request (a graph app, which is the point of jac) accumulates predicate locks far faster than the default pool allows, and jac-scale provisions Postgres with the stock defaults.

Observed on a fleet after roughly three real agent turns:

```
PgWireError: {'S': 'ERROR', 'C': '53200', 'M': 'out of shared memory',
 'H': 'You might need to increase "max_pred_locks_per_transaction".',
 'F': 'predicate.c', 'L': '2491', 'R': 'CreatePredicateLock'}
```

Measured on the provisioned instance at that moment:

| | |
|---|---|
| `max_pred_locks_per_transaction` | 64 (default) |
| `max_connections` | 100 |
| implied predicate-lock pool | ~6,400 |
| **SIReadLocks actually held** | **17,406** |
| active connections | 20 |

That is roughly 870 predicate locks per connection against a budget of 64.

### Why it is worse than a slow query

The failure lands on `JScaleExecutionContext()` construction, i.e. **every request that builds a context**, so it takes down `/healthz/live` and `/healthz/ready` too:

```
Unhandled error for GET /healthz/ready
  ... native_server.jac line 86, in request_context_middleware
      ctx = JScaleExecutionContext();
  ... session.impl.jac line 321, in batch_get
  ... store.impl.jac line 286, in load_full
  ... pgwire.impl.jac line 441, in run
PgWireError: out of shared memory ... CreatePredicateLock
```

Kubernetes then sees liveness/readiness failures and restarts the pods, which changes nothing, so the whole fleet sits in a restart loop with the database still up and healthy. Observed builder-sv at 9 restarts and jac-coder-sv at 11.

### Versions

[...]

Issue #8049, opened 2026-08-11
Archetype registration depends on import order: hooks do not back-fill, so reads silently return nothing for unregistered classes


MalithaPrabhashana:
### Summary

> Severity: not blocking us (we register by hand), but it is an order-dependent, silent data-invisibility bug: the same code passes or fails depending on module import order, and the failure looks exactly like "the data is not there".

Archetype classes are registered with the serializer through `Archetype._subclass_hooks`. `jaclang.runtimelib.serializer` appends its hook at import time:

```jac
glob _register_hook: object = Archetype._subclass_hooks.append(
         lambda (cls: type) { Serializer.register(cls) ... }
     ), ...
```

and `Archetype.__init_subclass__` runs the hooks for each new subclass:

```jac
for hook in Archetype._subclass_hooks {
    hook(cls);
}
```

`__init_subclass__` only fires for classes created **after** a hook is installed, and nothing back-fills classes created **before**. So whether an application's node/edge classes end up in `Serializer._registry` depends entirely on whether `jaclang.runtimelib.serializer` happened to be imported before the module that defines them.

When it loses that race, every read of those anchors fails:

```
WARNING - Refused to deserialize unregistered class: <module>.<Class>
WARNING - Serializer: archetype reconstruction failed for <module>.<Class>; dropping anchor (caller should quarantine).
```

The caller receives no error, just an empty result, which is indistinguishable from the data not existing. The rows are intact in `anchors` and become readable again in a process that wins the race.

### Versions

- jac 0.36.0 (release binary, Linux x86_64), Postgres 16 via `JAC_DB_URL`

### Reproduce (the missing back-fill, deterministic)

```jac
import from jaclang.runtimelib.serializer { Serializer }
import from mymod { Widget }          # any module defining a node

with entry {
    import from jaclang.jac0core.archetype { Archetype }
    seen: list = [];
    Archetype._subclass_hooks.append(lambda (c: type) -> None { seen.append(c.__name__); });
    print("hook back-filled existing classes:", len(seen));
}
```

Output:

```
hook back-filled existing classes: 0
```

A hook installed at time T sees nothing that existed before T. The serializer's own registration hook is installed exactly this way, so any archetype class defined before `runtimelib.serializer` is first imported is permanently unregistered in that process.

### Observed impact

Hit in jacBuilder's 0.36 migration. A regression suite that seeds a node, commits, then reads it back from a fresh context alternated between passing and failing across CI runs **on effectively identical code**:

[...]

Issue #8047, opened 2026-08-11
PgEventStream.consume() omits start_from declared by EventStreamBroker, silently killing every live event stream on Postgres


MalithaPrabhashana:
### Summary

> Severity: BLOCKING. On 0.36 with Postgres this silently kills every live event stream in an app: publishers keep writing, consumers written against the documented interface receive nothing, and the only symptom is a warning in the consumer's own logs. Works fine in local dev (LocalEventStream), fails only once deployed on Postgres.

`EventStreamBroker` declares `consume(topic, group=None, max_messages=1, timeout_seconds=5.0, start_from="latest")`. `LocalEventStream` implements that signature. **`PgEventStream` omits `start_from` entirely**, so any caller written against the interface raises

```
TypeError: PgEventStream.consume() got an unexpected keyword argument 'start_from'
```

Because a stream consumer typically wraps `consume()` in a try/except (a broker hiccup should not kill the consumer), the TypeError is swallowed as a transient error and the consumer just returns "no events" forever. Nothing surfaces at the publisher side, the events accumulate in `jac_events` correctly, and the application looks alive while its live stream is completely dead.

### Versions

- jac 0.36.0 (release binary, Linux x86_64)
- Postgres 16 via `JAC_DB_URL`

### Reproduce

```bash
docker run -d --name jacpg -e POSTGRES_PASSWORD=jac -e POSTGRES_DB=jacdb -p 5433:5432 postgres:16
export JAC_DB_URL="postgresql://postgres:jac@localhost:5433/jacdb"
jac run broker_repro.jac
```

`broker_repro.jac`:

```jac
import inspect;
import from jaclang.scale.events.streams.postgres { PgEventStream }
import from jaclang.scale.events.streams.local { LocalEventStream }
import from jaclang.scale.events.broker { EventStreamBroker, Event }

with entry {
    print("interface  :", str(inspect.signature(EventStreamBroker.consume)));
    print("InProcess  :", str(inspect.signature(LocalEventStream.consume)));
    print("PgEventStream:", str(inspect.signature(PgEventStream.consume)));

    stream = PgEventStream();
    stream.publish("demo.topic", Event(event_type="x", data={"n": 1}));
    try {
        got = stream.consume("demo.topic", group="g1", max_messages=10,
                             timeout_seconds=0.0, start_from="earliest");
        print("consume(start_from=...) ->", len(got), "events");
    } except Exception as e {
[...]

Issue #8046, opened 2026-08-11
PgStore LISTEN/NOTIFY: listen() sanitizes the channel name but notify() publishes raw, silently losing messages and cross-delivering channels


MalithaPrabhashana:
### Summary

> Severity: NOT currently blocking jacBuilder (a workaround is in place on the migrate/jac-0.36 branch); filed so the workaround can be retired and so others do not lose time on the same silent failure.

`PgStore.listen()` sanitizes the channel name before running `LISTEN` (non-ident chars become `_`), but `PgStore.notify()` publishes via `pg_notify` with the RAW channel name. For any channel containing a non-identifier character (a `:` namespace separator is the natural case), this causes two silent failures at once:

1. **Silent message loss**: `notify("cache:invalidate", ...)` is delivered to nobody, because the only thing anyone LISTENs on is the sanitized `cache_invalidate`.
2. **Cross-channel collision**: a subscriber of `cache:invalidate` receives messages published to the logically distinct channel `cache_invalidate`, because the listen loop matches subscribers by sanitized name.

No error or warning is raised on either side.

### Versions

- jac 0.36.0 (release binary, Linux x86_64)
- Postgres 16 via `JAC_DB_URL`

### Reproduce

```bash
docker run -d --name jacpg -e POSTGRES_PASSWORD=jac -e POSTGRES_DB=jacdb -p 5433:5432 postgres:16
export JAC_DB_URL="postgresql://postgres:jac@localhost:5433/jacdb"
jac run notify_repro.jac
```

`notify_repro.jac`:

```jac
import time;

glob got: list = [];

def on_colon(payload: str) -> None { got.append(("cache:invalidate", payload)); }
def on_plain(payload: str) -> None { got.append(("cache_invalidate", payload)); }

with entry {
    import from jaclang.scale.persistence.pg { shared_store }
    store = shared_store();
    store.listen("cache:invalidate", on_colon);
    store.listen("cache_invalidate", on_plain);
    time.sleep(1.5);
    store.notify("cache:invalidate", "A");
    store.notify("cache_invalidate", "B");
    time.sleep(3.0);
    print("EXPECTED: [('cache:invalidate', 'A'), ('cache_invalidate', 'B')]");
    print("ACTUAL:  ", sorted(got));
}
```

### Observed output

```
EXPECTED: [('cache:invalidate', 'A'), ('cache_invalidate', 'B')]
[...]

PR #8132, opened 2026-08-13
fix(runtime): close the follow-ups from the 0.36 Postgres fix batch (#8117)

### LINKED ISSUE #8117: Follow-ups from the 0.36 Postgres fix batch: import-cache poisoning, an ungated ensure_schema, and two footguns
### Summary

> Severity: none of these block the #8044 / #8045 / #8046 / #8047 / #8049 / #8050 fix batch (#8097, #8098, #8099, #8100, #8101, #8102, #8103). I verified all seven of those against their own repros and they do what they claim. Filing the leftovers separately so the batch can merge without carrying them.

Four items came out of reviewing that batch: two that Greptile raised and that are still present at the current PR heads, and two behavioural gaps I found while verifying. Each is small and self-contained.

### 1. A transient import failure permanently poisons the serializer's module cache

`Serializer._unimportable_modules` is written on the first failure and never expires, so a module that fails to import once is never retried in that process.

`jac/jaclang/runtimelib/impl/serializer.impl.jac` at #8099's head:

```jac
if module_name in Serializer._unimportable_modules {
    return None;
}
...
} except Exception as e {
    if len(Serializer._unimportable_modules) < _MAX_UNIMPORTABLE_MODULES {
        Serializer._unimportable_modules.add(module_name);
    }
    ...
    return None;
}
```

Most import failures are deterministic, so this is usually harmless. The case that worries me is a **circular import resolved under concurrency**: in a threaded server two requests can race module initialisation, one sees a partially initialised module and raises, and that module is then marked unimportable for the life of the process. Every later read of those anchors resolves to `None`, gets dropped and quarantined, and the only recovery is a restart. That is the same silent-data-invisibility shape as #8049.

A retry budget, or clearing the entry when the module later appears in `sys.modules`, would bound it.

### 2. The scale shared store calls `ensure_schema()` outside `_schema_gate`

`jac/jaclang/scale/persistence/pg.jac` at #8101's head:

```jac
if store is not None {
    store.ensure_schema({});
}
```

`PgStore` now guards first-use schema DDL behind `_schema_gate` plus `_schema_ready`, but this call site bypasses both. If a scale shared store initialises the same project database while a request-scoped `PgStore` is doing its first graph operation, both can run schema DDL at once, and one side can hit the lock timeout and surface `SessionConflictError` on an ordinary first request.

This is most likely to appear on a cold multi-replica fleet, where several pods take their first request simultaneously against a fresh database, which is exactly jacBuilder's startup shape.

### 3. `Archetype._subclass_hooks.append` still does not back-fill

#8099 fixes the ordering bug by adding `Archetype.install_subclass_hook()`, which appends **and** back-fills, and the serializer now uses it. Verified working: installing a hook that way back-filled 241 existing classes, where the raw append back-fills 0.

The raw list is still public and still has the old behaviour, so the next caller who does `Archetype._subclass_hooks.append(...)` silently reintroduces the exact bug #8049 describes. A short comment on `_subclass_hooks` pointing at `install_subclass_hook`, or making the list private, would close it for good.

### 4. The scratch-database self-heal does not hold in containers

#8103's note says a cluster start reaps scratch databases whose process is gone, "so a crashed run self-heals". The liveness check is:

```jac
impl process_alive(host: str, pid: int) -> bool {
    if pid <= 0 { return False; }
    if host and host != this_host() { return True; }
    try { os.kill(pid, 0); return True; }
    except ProcessLookupError { return False; }
    except PermissionError { return True; }
    except OSError { return True; }
}
```

Every pod gets a fresh hostname, so a record written by a previous pod always reads as alive and is never reaped. PID reuse on a long-lived host has the same effect. Both fail in the safe direction (leak rather than delete, which is the right bias for something that drops databases), but the self-healing property is weakest exactly where crashed runs are most common. Either soften the wording or add a secondary signal such as a last-heartbeat timestamp so an untouched record eventually becomes reapable regardless of host.

### Already addressed, listed so they are not re-reported

These were raised in review on the same batch and are resolved at the current heads: the UTF-8 byte-length limit on notification channel identifiers (#8098), the test fixture that classified setup failures as infrastructure unavailability (#8100), and concurrent schema initialisation between two `PgStore` instances, which `_schema_gate` now covers (#8101).

### Context

Found while verifying the seven-PR batch for jacBuilder's 0.36 migration, tracked at jaseci-labs/jacBuilder#1266. Items 1 and 2 were surfaced by Greptile and confirmed by reading the code at the current PR heads; items 3 and 4 came out of running the fixes. Happy to test a follow-up PR the same way.

marsninja:
Fixes #8117.

Four items from reviewing the #8044 / #8045 / #8046 / #8047 / #8049 / #8050 batch, one commit each, on top of the merged heads. Verifiable item by item.

### 1. A transient import failure no longer poisons the serializer's module cache

`Serializer._unimportable_modules` was a set written on the first failure and never cleared, so one failed import made a module dead for the life of the process: every later read of an anchor whose class lives there resolved to `None`, was dropped and quarantined, and only a restart recovered. The failure that matters is not deterministic, it is a circular import resolved under concurrency, which is transient by definition.

The set is replaced by a bounded backoff record (module name -> attempts + a monotonic retry deadline). The cache keeps its original job, which was to stop a dead name turning every anchor read into a path scan: a failing module is attempted at most once per window, the window doubles per consecutive failure up to a ceiling, and the entry is dropped the moment the module turns up in `sys.modules`, which is exactly what the losing thread of an import race sees a moment later. `Serializer.is_auto_registered_module` and the rest of the resolution allowlist are untouched.

To verify: a module that raises on first import and imports cleanly afterwards is resolvable again without a restart.

### 2. The scale shared store now goes through `_schema_gate`

`PgStore` guards first-use schema DDL behind `_schema_gate` plus `_schema_ready`, and `jaclang/scale/persistence/pg.jac` called `store.ensure_schema({})` past both, so a scale shared store and a request-scoped `PgStore` could run schema DDL against the same database at once and one side could hit the DDL lock timeout and surface `SessionConflictError` on an ordinary first request.

Rather than a second parallel guard, the gate is now a method: `PgStore.ensure_schema_once(promotions)` owns `_schema_gate` + `_schema_ready`, `_ensure_ready` calls it, and the scale shared store and the `db` CLI call it too. No production module calls `ensure_schema` directly any more, and a test asserts that stays true. The scale store also passes the project's real `[database] indexes` promotions instead of `{}`, so whichever side wins the race applies the same schema.

To verify: eight threads racing a store's first use run schema DDL once.

### 3. `Archetype._subclass_hooks.append` is now impossible, not just discouraged

`install_subclass_hook` appends and back-fills, but the raw list stayed mutable, so the next caller who used `.append` silently reintroduced #8049. A comment would not have survived, since production `.jac` files are stripped of comments and docstrings by the repo's own lint rules. The hook list is an immutable tuple now, so `.append` raises `AttributeError` at the call, and installation and removal go through `install_subclass_hook` / the new `uninstall_subclass_hook`. Reads (`len`, iteration) still work, and replacing the tuple wholesale also means `__init_subclass__` never iterates a list that is being mutated.

To verify: `Archetype._subclass_hooks.append(hook)` raises instead of installing a hook that sees nothing that already exists.

### 4. The scratch-database self-heal now holds in containers

`process_alive(host, pid)` returns `True` whenever `host != this_host()`, and every pod gets a fresh hostname, so a scratch database left by a crashed pod always read as alive and was never reaped. PID reuse on a long-lived host did the same.

The registry record carries a `last_heartbeat_at` column now (added in place on existing registries), refreshed every time a process attaches to the database and then every 60s by one daemon heartbeat per scratch database per process, stopped before the database is dropped. `DbRecord.is_dead_scratch()` stays the single predicate shared by the creation-time reap and `jac db prune`.

Reaping on silence alone was not safe, and Greptile was right to keep pushing on it: a silent heartbeat and an idle database are not independent signals, because one partition, one saturated connection pool or one rotated credential stops the heartbeat *and* drops the store's connection at the same instant for the same reason, and `process_alive` cannot break the tie for a foreign pid. So silence observed once is no longer enough:

- The first pass to find a scratch record silent and unused only **condemns** it, stamping `condemned_at`. A drop needs a later pass, at least an hour after that stamp, that still finds it silent and unused, so one outage window cannot supply both observations. Any heartbeat (or a process re-attaching) clears the stamp, so a partition the owner survives costs it nothing.
- Only a pid confirmed gone from this host justifies `DROP DATABASE ... WITH (FORCE)`. A heartbeat-grounded drop is unforced, so an owner that has reconnected vetoes it at the server instead of being terminated by it.
- The reaper re-reads the record immediately before dropping instead of acting on the snapshot `list_databases` took earlier.
- The heartbeat no longer fails silently at `debug`: it warns after five minutes of failure, again when silence passes the grace, and reports recovery, since silent heartbeat failure is exactly the condition that precedes a wrong reclamation.

The rest of the bias is unchanged: a pid this host can check answers the question outright, an unknown heartbeat age never authorises a drop, and the grace is 30 minutes of silence for a host that cannot be checked, 24 hours when the pid answers and may be recycled. `jac db list` shows a condemned store as `silent scratch` and `jac db prune -y` prints what it marked. The `jac db` reference said a cluster start reaps scratch databases "left by processes that are gone" without saying how that is decided, which is the sentence that hid this gap; it now describes both signals and their bias. #8103's own release-note fragment is left untouched (the release-notes gate refuses edits to another PR's fragment, and its "a crashed run self-heals" claim is true once this lands).

To verify: a scratch record from a hostname that never comes back survives the first `jac db prune -y` (which marks it) and is reclaimed by a second pass once that mark has aged; one that is still heartbeating survives both; and one whose owner has reconnected keeps both its database and its session.

### Review record

Greptile filed four findings. The first two are reviews of the tests-only commit and are answered by the four fix commits after it. The third and fourth are the same hazard at two depths: the third (a live owner that keeps its connection) was answered by requiring an idle database, and the fourth (a live owner that holds no connection while its heartbeat fails) was **not** answered by that and required the redesign above. My earlier comment on this PR wrongly lumped the fourth in with the stale ones; that is corrected in a follow-up comment.

### Tests

`jac test tests/runtimelib/test_archetype_registration_order.jac` (13 passed, 4 new), `tests/runtimelib/test_lazy_pg_boot.jac` (11 passed, 2 new), `tests/runtimelib/test_db_lifecycle.jac` (17 passed, 5 new), plus `tests/runtimelib/test_pg_session.jac` (8), `tests/runtimelib/test_schema_repair.jac` (25) and `jaclang/scale/tests/test_identity_fail_loud.jac` (2) as the adjacent existing suites. Each new test was confirmed failing on the unfixed heads first; the failing-test commit is separate and first, so the whole set can be reproduced by checking it out on its own.

### Not addressed

Nothing from the issue is left. The three items listed as already resolved at the merged heads were not touched.


MalithaPrabhashana (commented):
Verified all four items against their own repros, same way as the earlier batch: a worktree at the PR head, the dev lane (`🛠 jac dev mode — using compiler source at .../wt-8132/jac`), and a real Postgres. Baseline for every comparison is `upstream/main` at `5c947b346`, run identically.

**Verdict: good to go.** One scope note on item 2 that is not a defect in this PR and does not block it.

### The tests fail without the fixes

The tests-first commit (`4df31955e`) makes this easy to confirm, so the suite is not just green, it is actually pinning the behaviour:

| file | at `4df31955e` (tests, no fixes) | at PR head |
| --- | --- | --- |
| `test_lazy_pg_boot.jac` | 2 failed, 9 passed, 1 error | **11 passed** |
| `test_archetype_registration_order.jac` | 7 failed, 6 passed | **13 passed** |
| `test_db_lifecycle.jac` | 2 failed, 12 passed | **18 passed** |

Every failing name maps onto an item, with nothing left over.

### Item 1: bounded backoff, verified through the real API

Driving `Serializer` directly on each branch:

```
### PR #8132
1. allowed before any failure : True
2. allowed right after failure: False ( window = 30.0 s )
3. allowed once 31s have passed: True
4. 2nd failure window = 60.0 s; allowed after only 31s: False
5. allowed after 62s: True

### BASELINE upstream/main
1. recorded as unimportable   : True
2. any retry/backoff API?     : False
3. any expiry on the record?  : no - it is a plain set with no timestamp
```

The window doubles as documented, and the permanent-poisoning path is gone. `sys.modules.get(module_name)` is still checked before the gate, so a module that turns up after losing an import race resolves immediately rather than waiting out its window, which is the case that motivated the item.

### Item 3: closed structurally, better than what I suggested

I asked for a comment or a private name. Making it a tuple is stronger, because the footgun becomes impossible rather than discouraged:

```
### PR #8132
_subclass_hooks type: tuple
RESULT: raw append blocked -> 'tuple' object has no attribute 'append'
install/uninstall present: True True

### BASELINE upstream/main
_subclass_hooks type: list
RESULT: raw append SUCCEEDED - footgun still open
install/uninstall present: True False
```

Worth stating plainly since it is a behaviour change: any existing caller doing `_subclass_hooks.append(...)` now gets an `AttributeError` instead of silently skipping the back-fill. That is the right trade, and it fails loudly on a leading-underscore name. `uninstall_subclass_hook` plus the lock are a good addition.

### Item 4: the destructive path got more conservative, not less

This is the one I read hardest, since it drops databases. The two-phase shape is right: `condemn_stale_scratch` only marks on the first prune, `SCRATCH_CONDEMN_SEPARATION_SECONDS = 3600` keeps a later prune an hour away, and `drop_if_still_dead` re-checks at drop time. `force=True` becoming `force=rec.pid_confirmed_dead()` is a real improvement on its own, since force-dropping terminates live connections. The `ADD COLUMN IF NOT EXISTS` migration is idempotent.

### Item 2: correct for what it fixes, but it closes half of what I filed

The fix is right and the extraction is clean, and passing real `promotions` where the old call site passed `{}` is a genuine bonus I did not ask for: the shared store now agrees with the request-scoped store instead of running a weaker schema.

The scope note is that `_schema_gate` is a process-local `Lock` and `_schema_ready` is a process-local set, so this dedupes **within one process**. The test pins exactly that, "even from many threads". But item 2 as I wrote it said the case that worries me is "a cold multi-replica fleet, where several pods take their first request simultaneously against a fresh database", and that is cross-process, which an in-process lock cannot cover. That part is still reproducible at this PR head:

```
### BASELINE upstream/main  (5c947b346)
  trial 1: A=000 B=200 -> ONE DIED [duplicate key ... "pg_class_relname_nsp_index"]
  trial 2: A=200 B=000 -> ONE DIED [duplicate key ... "pg_class_relname_nsp_index"]
  trial 3: A=200 B=000 -> ONE DIED [duplicate key ... "pg_class_relname_nsp_index"]

### PR #8132  (17fa30440)
  trial 1: A=200 B=000 -> ONE DIED [duplicate key ... "pg_class_relname_nsp_index"]
  trial 2: A=000 B=200 -> ONE DIED [duplicate key ... "pg_class_relname_nsp_index"]
  trial 3: A=000 B=200 -> ONE DIED [duplicate key ... "pg_class_relname_nsp_index"]
```

Two `jac start` processes against one fresh database, three trials each. That is my fault for writing one item that contained two different races, not a gap in this PR. I filed the cross-process half separately as #8166 with its own repro, and it wants a different mechanism anyway, either `pg_advisory_lock` around `ensure_schema` or tolerating `23505` / `42P07`. Nothing here needs to change for that.

### Regressions

Nothing found.

- Related suites at PR head: `test_db_cli` 3 passed, `test_db_naming` 2 passed, `test_osp_store_lifecycle` 2 passed, `test_occ_read_capture` 3 passed, `test_dangling_ref_heal` 2 passed, `test_context_isolation` 3 passed.
- The branch is behind main, so I test-merged `upstream/main` locally and re-ran the three item suites on the merged tree: 11, 13 and 18 passed, unchanged. No silent semantic conflict of the kind #8103 had.
- CI is 29/29 green.

### One methodology note for anyone re-running this

The dev lane swaps the **compiler** for the checkout, but a program only picks up the checkout's `jaclang` **runtime** if it runs from the checkout root. My first pass ran the probes from a subdirectory and quietly measured the embedded 0.36.1 runtime instead, which made the PR and the baseline look identical. Confirmed with:

```
archetype module file: .../home-8132/.cache/jac/rt/.../site/jaclang/jac0core/archetype.jac   <- subdirectory, wrong
archetype module file: .../wt-8132/jac/jaclang/jac0core/archetype.jac                        <- checkout root, right
```

`PYTHONPATH` does not override it. Every result above is from the checkout root.

PR #8134, opened 2026-08-13
fix(client): make AuthGuard render its children on React and react-native

### LINKED ISSUE #8023: AuthGuard ignores its children and renders <Outlet/>, silently blanking protected pages in manual routing
## Summary

`AuthGuard` receives its JSX children and silently discards them. On successful authentication it renders `<Outlet/>` unconditionally, so wrapping a page in it renders a **completely blank page with no compile error and no console error**.

```jac
<Route path="/dashboard"
       element={<AuthGuard redirect="/login"><Dashboard /></AuthGuard>} />
```

Log in, land on `/dashboard`, get `<div id="root"></div>`.

This was originally filed as a docs bug in `jac-cl-auth.md`. After reproducing it, the docs are downstream of a runtime defect. The children **are** delivered to the component and are simply never read, and the Solid backend already implements the children-aware behaviour that React and react-native do not. Retitled and rescoped accordingly.

Verified on `jac 0.36.1`, jaclang @ `300e220fe`.

---

## Reproduction

Minimal app, manual flat routing, written exactly as `jac-cl-routing.md` (manual routing section) and `jac-cl-auth.md:121` prescribe.

```jac
# AppShell.jac
import from "@jac/runtime" { Router, Routes, Route, AuthGuard }

def:pub AppShell() -> JsxElement {
    <Router>
        <Routes>
            <Route path="/login" element={<LoginPage />} />

            {#* A: the documented wrapper form *#}
            <Route path="/dashboard"
                   element={<AuthGuard redirect="/login"><Dashboard /></AuthGuard>} />

            {#* B: inline guard *#}
            <Route path="/dashboard-inline" element={<DashboardInline />} />

            {#* C: AuthGuard as a parent route element *#}
            <Route element={<AuthGuard redirect="/login" />}>
                <Route path="/dashboard-nested" element={<Dashboard />} />
            </Route>
        </Routes>
    </Router>
}
```

Real signup against a live `jac start` server (valid JWT in `localStorage`), driven with Playwright:

```
--- LOGGED IN ---                              --- LOGGED OUT ---
  A  AuthGuard as WRAPPER      BLANK             /login  ok
  B  inline jacIsLoggedIn      renders           /login  ok
  C  AuthGuard as PARENT ROUTE renders           /login  ok

jac check: passes
console messages: 0
```

### Only the success path is broken

All three shapes redirect anonymous users correctly. **A test asking "does the guard block logged-out users?" passes on the broken code.** The assertion that catches this is "does the protected page render when logged IN?". This is almost certainly why it shipped and survived.

---

## Root cause

**1. The children are delivered.** The call site compiles to:

```js
__jacJsx(Route, {"path": "/dashboard",
                 "element": __jacJsx(AuthGuard, {"redirect":"/login"}, [ Dashboard ])
                }, [])
```

`__jacJsx` (`jaclang/runtimelib/impl/client_runtime.impl.jac:39`) spreads that array into `React.createElement(tag, props, ...children)`, so `props.children` is populated.

**2. The component never reads them.** `jaclang/runtimelib/impl/client_runtime.impl.jac:124`:

```jac
impl AuthGuard(redirect: str = "/login") -> JsxElement {
    if jacIsLoggedIn() {
        return <ReactRouterOutlet/>;
    }
    <ReactRouterNavigate to={redirect} replace={True}/>
}
```

Compiled:

```js
function AuthGuard(props) {
  const {redirect: redirect = "/login"} = props;   // destructures redirect only
  if (jacIsLoggedIn()) {
    return __jacJsx(ReactRouterOutlet, {}, []);
  }
  ...
}
```

**3. `<Outlet/>` resolves to nothing in a flat route.** `Outlet` renders the *matched child route*, not JSX children. A self-closing `<Route>` has no child routes, so `Outlet` renders `null`. React treats "no matched child" and "component returned null" as entirely normal, so nothing warns. `jac check` does not flag children passed to a component that declares none.

Net effect: the page is handed to the guard, sits unread in `props`, and the guard renders an empty slot instead.

### Confirmation that only the read is missing

Same call site, same everything, one added parameter:

```jac
def:pub FixedAuthGuard(redirect: str = "/login", children: any = None) -> JsxElement {
    if not jacIsLoggedIn() { return <Navigate to={redirect} replace={True} />; }
    if children != None { return <>{children}</>; }
    <Outlet />
}
```

Renders. Nothing upstream needed changing.

---

## The backends disagree

Same `@jac/runtime` export, three implementations, two different contracts:

| Backend | Declaration | Renders children? |
|---|---|---|
| React (web) | `runtimelib/client_runtime.jac:135` | no |
| react-native | `runtimelib/client/client_runtime_native.jac:114` | no |
| Solid | `runtimelib/impl/solid_runtime.impl.jac:114` | **yes** |

```jac
# Solid, already correct
impl AuthGuard(redirect: str = "/login", children: any = None) -> JsxElement {
    if jacIsLoggedIn() { return children; }
    return Navigate({"to": redirect, "replace": True});
}
```

So the docs are correct for Solid and wrong for the backend every web app uses.

---

## Test coverage gap

`AuthGuard` is referenced by exactly four files under `jac/tests/`:

```
jac/tests/compiler/passes/ecmascript/test_solid_backend.jac
jac/tests/runtimelib/test_solid_jsdom.jac
jac/tests/runtimelib/fixtures/solid_routing/dashboard_page.jac
jac/tests/runtimelib/client/fixtures/react_native_multiscreen/main.jac
```

**There is no React/web `AuthGuard` test.** The backend with the correct implementation is the tested one. `test_solid_jsdom.jac:334` even asserts the right thing for Solid:

```js
check("auth route renders the protected page when logged in", present("dashboard"), ...)
```

React has no equivalent. This gap is why both this bug and the `auth_redirect` regression below shipped and survived, so the fix should not land without it. A concrete requested test matrix is in **Proposed fix, item 2**.

---

## History (not a regression)

Behaviour has been identical since the component was introduced.

| Date | PR | Event |
|---|---|---|
| 2026-02-02 | #4422 File based routing | `AuthGuard` introduced for the `(auth)/` codegen. Shipped with a docstring: *"Auth guard component for file-based routing. Renders child routes if authenticated."* Correct and correctly scoped. |
| 2026-06-12 | #6600 Overhaul agent skill guides | `jac-cl-auth.md` gains *"`AuthGuard` ... wraps children"* and *"In manual routing, wrap the protected subtree the same way"*. Docs and code diverge. |
| 2026-06-28 | #7025 chore(lint): deslop core sources | The docstring stating the constraint is deleted. Last in-source record of the scope is gone. |

Nothing regressed. A correct, narrowly-scoped component was exported as public API, then documented as something it never was, then lost the comment that explained the difference.

---

## Proposed fix

**1. Make `AuthGuard` children-aware, keep `<Outlet/>` as the fallback.**

```jac
# runtimelib/client_runtime.jac:135
def:pub AuthGuard(redirect: str = "/login", children: any = None) -> JsxElement;

# runtimelib/impl/client_runtime.impl.jac:124
impl AuthGuard(redirect: str = "/login", children: any = None) -> JsxElement {
    if not jacIsLoggedIn() {
        return <ReactRouterNavigate to={redirect} replace={True}/>;
    }
    if children != None {
        return <>{children}</>;
    }
    <ReactRouterOutlet/>
}
```

Apply the same to `runtimelib/client/impl/client_runtime_native.impl.jac:767` so all three backends agree.

**Backward compatible.** Every existing call site passes no children and keeps the `<Outlet/>` branch:

- the `(auth)/` codegen calls `AuthGuard(...)` with no children
- `<Route element={<AuthGuard />}>` with nested routes (case C above) passes none
- the documented file-based layout `<AuthGuard redirect="/login"><Outlet /></AuthGuard>` currently works by rendering AuthGuard's own `<Outlet/>`; afterwards it renders the user's `<Outlet/>`, which is the same component

**2. Add React/web `AuthGuard` tests.** There are none today, which is why this and the `auth_redirect` regression below both survived. Requested cases, with what each one would catch:

| # | Scenario | Auth state | Expected | Today |
|---|---|---|---|---|
| 1 | `<AuthGuard redirect="/login"><Dashboard/></AuthGuard>` in a flat `<Route element=...>` | logged in | `Dashboard` renders | **FAILS** (this issue) |
| 2 | same | logged out | redirects to `/login` | passes |
| 3 | `<Route element={<AuthGuard/>}>` with a nested child route | logged in | child route renders | passes |
| 4 | same | logged out | redirects to `/login` | passes |
| 5 | file-based `pages/(auth)/dashboard.jac` | logged in | page renders | passes |
| 6 | same | logged out | redirects | passes |
| 7 | file-based layout `<AuthGuard redirect="/login"><Outlet/></AuthGuard>` | logged in | child route renders | passes (must not regress) |
| 8 | `<AuthGuard/>` with neither children nor child routes | logged in | renders nothing, no crash | passes (pins the `<Outlet/>` fallback) |
| 9 | `[client.routing] auth_redirect = "/signin"`, any guarded route | logged out | lands on `/signin` | **FAILS** (see adjacent finding) |

Cases 1 and 9 are the regression tests; they must fail before the patch and pass after. Cases 3 to 8 are the backward-compatibility guard: they pass today and must keep passing, since they cover every shape currently in use (codegen, parent-route, documented layout).

**A test that only asserts the logged-out redirect passes on the broken code.** Cases 2, 4 and 6 alone would have caught neither bug. The logged-in render assertion is the one that matters.

`jac/tests/runtimelib/test_solid_jsdom.jac` is the closest existing model; its `startUrl` x `loggedIn` fixture already expresses cases 5 and 6 for Solid and can be mirrored for React. Parity for react-native (`jac/tests/runtimelib/client/fixtures/react_native_multiscreen/`) is worth adding in the same pass, since it shares the children-less implementation being changed.

**3. Docs.** Once the runtime accepts children, `jac-cl-auth.md:110` and `:121` become true as written. Worth adding regardless:

- the parent-route form, which works today and is the better shape for guarding a group of pages
- restore the scope note that #7025 deleted

Note on rollout: text of the form "wrapping works" is only true on jaclang versions that carry this fix. Guidance that stays correct on both old and new runtimes is *"`AuthGuard` goes above your routes, not around your page"* plus the inline guard for one-offs.

---

## Follow-up worth its own issue

`jac check` does not reject children passed to a component that declares none, and it has the declaration needed to do so (`def:pub AuthGuard(redirect: str = "/login")`). The equivalent mistake in TypeScript React is a compile error.

This is a class, not an instance. `jac-cl-routing.md` already documents a second member of it: an `app` that ignores `children` **silently drops every route**, with `jac check` passing and no error anywhere. Unrendered `children` is currently the most reliable way to produce a blank screen with zero diagnostics.

---

## Adjacent finding: `auth_redirect` is broken again (separate bug, same file)

Flagging because whoever fixes this will be editing the exact line.

`runtimelib/react_entry.jac:69` generates a positional string call into a props-destructuring component:

```js
// generated _entry.js, from a jac.toml with auth_redirect = "/signin"
function _AuthGuardRoute() { return AuthGuard("/signin"); }
```

```
$ node -e 'const {redirect = "/login"} = "/signin"; console.log(redirect)'
/login
```

The configured value is discarded and every user lands on `/login`.

This is a revert of a shipped fix. #6878 (2026-06-23, *"repair auth_redirect config, broken since file-based routing shipped"*) replaced it with `React.createElement(AuthGuard, { redirect: "..." })`. #6227 (2026-07-02, beta react native support) restored the broken form along with this comment:

```jac
# AuthGuard is a compiled Jac function taking the redirect path as
# a positional string, not a props-consuming React component, so
# it needs a wrapper component rather than createElement props.
```

That comment is incorrect. A fresh build compiles `AuthGuard` as `function AuthGuard(props)`. `solid_entry.jac` kept the correct props form; only React regressed.

Verified by inspecting generated `_entry.js` and the compiled runtime; end-to-end browser confirmation not run, though the three facts above determine the outcome.

Fix is to restore #6878's line and delete the comment. Probably belongs in its own issue so it gets its own release note, but it should not be left in place by a PR touching this code.

---

## Repro app

Available on request: minimal `jac create --kind web-app` project with all four guard shapes wired side by side, a signup form, and a Playwright driver that produces the matrix above.

SupulHeshan:
Fixes #8023.

`AuthGuard` receives its JSX children and silently discards them. On the React and react-native backends it renders `<Outlet/>` unconditionally, and `Outlet` resolves the *matched child route* -- which is nothing in a flat route. So the documented form

```jac
<Route path="/dashboard"
       element={<AuthGuard redirect="/login"><Dashboard /></AuthGuard>} />
```

renders a blank page for a logged-in user, with no compile error and no console error. Solid's implementation was already children-aware; the two React backends were not.

## What's in here

1. **`AuthGuard` reads `children`, keeps `<Outlet/>` as the fallback** -- React
   (`runtimelib/impl/client_runtime.impl.jac`) and react-native
   (`runtimelib/client/impl/client_runtime_native.impl.jac`), matching Solid.
   Backward compatible: every existing call site passes no children and keeps
   the `<Outlet/>` branch (the `(auth)/` codegen, `<Route element={<AuthGuard/>}>`
   with nested routes, and the documented `<AuthGuard><Outlet/></AuthGuard>`
   layout, which afterwards renders the user's `<Outlet/>` -- the same component).

2. **`auth_redirect` reaches the guard again.** `react_entry.jac` emitted
   `function _AuthGuardRoute() { return AuthGuard("/signin"); }` -- a positional
   string into `function AuthGuard(props)`, so the configured redirect
   destructured to the `/login` default and every user landed there. Restored to
   the props form, and dropped the comment that justified the positional call.

3. **React/web `AuthGuard` tests, which did not exist.** The only executable
   coverage was Solid's -- the one backend that was already correct.

4. **A jsdom suite that had never run.** `test_react_state_semantics.jac` skips
   itself unless `JAC_CLIENT_INTEGRATION` is set, and no CI step named it, so it
   has never executed. The workflow step now names every jsdom suite and says so
   in a comment, since a suite missing from that list is a suite that never runs.

## Tests

`tests/runtimelib/test_react_auth_jsdom.jac` compiles the real runtime shim and
mounts it in jsdom (opt-in via `JAC_CLIENT_INTEGRATION`, same as
`test_react_state_semantics.jac`), one process per scenario so each starts at
its own URL and auth state:

| Scenario | Auth | Expected | On `main` |
|---|---|---|---|
| guard wrapping a page as a route `element` | in | page renders | **FAILS** |
| same | out | redirects to `/login` | passes |
| guard as a pathless parent route | in | child route renders | passes |
| same | out | redirects | passes |
| layout wrapping `<Outlet/>` in the guard | in | child renders | passes |
| same | out | redirects | passes |
| guard with neither children nor child routes | in | renders nothing, no crash | passes |
| same | out | redirects | passes |

One row fails on `main`, and it is the whole bug. Note that a logged-*out*
assertion passes on the broken code -- a guard that renders nothing blocks
anonymous users perfectly well -- so the logged-in render is the assertion that
matters, and it is the one that was missing.

`tests/compiler/passes/ecmascript/test_react_backend.jac` pins the entry's
calling convention at the source level (the React entry had no test at all,
which is how the positional call came back). Its `auth_redirect` test also fails
on `main`. This is the only coverage for defect 2: it is a string-level
assertion on the generated entry, not a behavioral one. A jsdom scenario driving
the generated `pages/` entry end to end is left as a follow-up.

`tests/runtimelib/test_client_runtime_parity.jac` asserts that every `def:pub`
shared across the three router runtimes and the two cores declares one
signature, rather than merely existing. Signature checks alone cannot pin
behavior, since the compiler never checks an `impl/` body against its
declaration, so the last test reads the compiled `AuthGuard` body instead.

`tests/runtimelib/react_jsdom_harness.jac` extracts the rig shared by the two
React jsdom suites; `test_react_state_semantics.jac` shrinks by ~110 lines onto
it.

Docs: `jac-cl-auth.md` becomes true as written; the parent-route form and the
scope note deleted in #7025 are restored.

## Known limitation, not fixed here

The parent-route form documented here works on React and react-native. It does
**not** work on Solid: `Route` wraps an `element` route in a props-ignoring
lambda (`solid_runtime.impl.jac`), and Solid's `AuthGuard` has no `<Outlet/>`
fallback, so `<Route element={<AuthGuard/>}>` with nested routes renders
nothing. Solid's `component=` form and the generated `pages/` entry both pass
`props.children` explicitly and are unaffected, which is why no test caught it.
Tracked separately; the docs here describe the React behavior.

More broadly, this fixes one component. Any Jac component that declares a
parameter and not `children` still discards its JSX children silently, with
`jac check` clean, because the JSX prop validator inspects attributes only and
never reads an element's children. That is the general fix and it belongs in the
type checker, not here.


MalithaPrabhashana (approved):
Tested this end to end and the fix works. Flagging up front that this is an **application-behavior validation, not a code-quality sign-off**, so please hold the merge until @kugesan1105 and @Thamirawaran have reviewed the code itself. They own this area upstream and their read should be the one that gates it.

## What I ran

A `jac create --kind web-app` project pointed at this branch via `[dev] jaclang_source`, real signup against a live `jac start` server (valid JWT in `localStorage`), driven in Chromium. Same app, same script, run against `main` and against this branch, covering four shapes beyond the ones in `fixtures/react_auth/manual_routing_app.jac`.

| Shape | `main` | this PR |
|---|---|---|
| `<AuthGuard><Page/></AuthGuard>` as a route element | **blank** | renders |
| Two element children | **blank** | both render |
| Text child | **blank** | renders |
| Guard nested inside a guard | **blank** | renders |
| Layout: content beside `<Outlet/>` | header **silently dropped** | header + page |
| `<Route element={<AuthGuard/>}>` with child routes | renders | renders |
| Inline `jacIsLoggedIn()` guard (control) | renders | renders |
| Hand-written children-aware guard (control) | renders | renders |
| All eight, logged out | redirect to `/login` | redirect to `/login` |

**16/16 pass on this branch, 5 fail on `main`, zero console errors on either.** The three shapes that already worked still work and every logged-out redirect is unchanged, so the backward-compatibility claim holds under test. Four of the five failures are shapes this PR does not itself test, so the fix generalizes past its own fixture.

I also confirmed the new jsdom suites genuinely execute rather than skip: `test-solid-and-desktop` reports **7 passed**, which matches 5+1+1 across the three files exactly. Worth stating because a pytest skip also exits 0, so a green step alone would not have proven it.

One implementation detail I checked and can confirm is safe: `children is not None` lowers to a **loose** `!= null` (`esast_gen_pass.impl.jac:2642` special-cases null comparisons), so React passing `undefined` for absent children still falls through to the `<Outlet/>` branch. That is what keeps the `(auth)/` codegen and the pathless-parent shape working, and both pass in the browser.

## Two things for the author, neither blocking

**1. The scenario table in the PR body overstates the jsdom coverage.** It lists the file-based `(auth)/` page and `auth_redirect = "/signin"` as scenarios, but `test_react_auth_jsdom.jac` runs four shapes across two auth states and its own docstring says so honestly, deferring the entry to `test_react_backend.jac`. The `auth_redirect` fix is pinned at the string level only; no behavioral test proves a logged-out user actually lands on `/signin`. The composition covers it, but the table should match the code.

**2. One visible behavior change is not mentioned.** Content placed *beside* `<Outlet/>` inside the guard was previously discarded and now renders:

```jac
def:pub AuthShell() -> JsxLayout {
    <AuthGuard redirect="/login">
        <NavBar/>      # dropped on main, renders after this PR
        <Outlet/>
    </AuthGuard>
}
```

Measured: `main` gives `['page-a']`, this branch gives `['header', 'page-a']`. Nothing breaks and the new behavior is the correct one, but anyone who hit "my navbar will not show inside AuthGuard" and worked around it by also rendering the nav in a parent layout will get two after upgrading. Worth one sentence in the release note. The backward-compat argument in the description covers `<AuthGuard><Outlet/></AuthGuard>` exactly, which is genuinely unaffected, and stops there.

## On the tests

706 test lines against a 14-line runtime change reads heavy until you look at the structure, and I do not think it is bloat:

- `react_jsdom_harness.jac` is real extraction, not new surface. `test_react_state_semantics.jac` gives up 109 lines of its own compile helper, deps dict, bun install and skip guard to use it. One rig, two suites.
- `test_client_runtime_parity.jac` is the strongest file here. It carries anti-vacuity assertions so a parity test cannot silently compare nothing, and after noting that the compiler never checks an `impl/` body against its declaration, its last test reads the **compiled body** rather than the signature. Deleting `children` from the impl would keep every signature test green and revive #8023; that test is what catches it.
- The CI change is the one I would keep even if everything else were cut. These suites skip unless `JAC_CLIENT_INTEGRATION` is set, so a suite not named in that step never runs anywhere. That is the meta-bug that produced this issue in the first place.

@kugesan1105 @Thamirawaran over to you for the code review. I can confirm the runtime behavior is correct in a real app across every guard shape I could think of, but I would like your eyes on the implementation and the test structure before this goes in.

PR #8170, opened 2026-08-14
fix(scale): resolve every service module host-side against what the bundle ships and boot pods from it, replacing the entrypoint find

### LINKED ISSUE #8105: bundle membership WI-2 (re-scoped): resolve the service file host-side, always set JAC_SV_FILE, delete the pod find
---

## Re-scoped 2026-08-13 per marsninja's review on #8077: size S, no format change

`manifest.json` is dropped from v1: `/app` is an emptyDir re-untarred per pod start from several bundles (toolchain + app), so a pod-side member list buys little and costs a format migration. What remains is the valuable half:

- Resolve the service file host-side against the stage, at deploy time.
- Wire the resolved map from `resolve_service_files` (scale/runtime/routing.jac:143) into `_build_env` (deploy/target/kubernetes/manifest_builder.jac:683), which today only reads a user-declared `services.<name>.file` and never sees the resolved map.
- `JAC_SV_FILE` is always set; missing or ambiguous resolution fails the deploy loud.
- The pod-side `find ... | head -n1` fallback (order-dependent, can resolve the wrong copy) is deleted.

Revised acceptance:

- [ ] Every pod env carries an explicit `JAC_SV_FILE`; no `find` in any entrypoint
- [ ] Ambiguous resolution (two candidate files for one service) fails the deploy naming both paths
- [ ] Kind e2e: fleet boots with resolution done host-side
- [ ] The original mixed-version/manifest.json acceptance items are void (superseded)

kashmithnisakya:
Closes #8105. WI-2 of #8077 (re-scoped 2026-08-13: no format change, no `manifest.json`), building on #8097's `BundleSourceFilter`.

## Problem

The service-file decision was made at the point of use, against a filesystem, at the worst possible time. The host enumerates the full member set at pack time and knows every declared service, but threw both away: membership was applied as a filter predicate and never recorded, and service resolution was deferred to a shell `find` inside the pod (`_pod_entrypoint`, manifest_builder.jac:1076). Consequences at head:

- For any routed service without an explicit `[scale.microservices.services.NAME] file`, `find . -name "$JAC_SV_NAME.jac" | head -n1` picks the first match in readdir order. A same-named `.jac` elsewhere in the shipped tree (e.g. a seal-exempt `tests/<svc>.jac`) can be booted as the service, nondeterministically per node.
- A `.jacignore`-parked service module packed fine and crash-looped its pods in-cluster with a missing-file error; #8097's fail-loud guard covered only the entry module.
- The implicit `main` pod ignored `[project] entry-point` entirely and booted whatever `main.jac` `find` turned up.

## Change

- One resolution rule, `resolve_service_members` in `bundle.jac`, applied to the exact set of sources the bundle ships: explicit `[scale.microservices.services.NAME] file` wins and must ship; else root `<name>.jac`; else the unique same-named module anywhere in the tree; zero or several candidates fail naming the module, the `.jacignore` pattern or baseline rule, or every candidate. The rule reuses `routing.jac`'s `resolve_routes` and the (newly extracted) `needs_main_service` predicate rather than re-deriving them.
- The implicit `main` service (routed fleets with no route covering `/`) boots the `[project] entry-point` module at its exact path, so a same-named root stray never shadows a configured `src/app.jac`, and same-named modules elsewhere never make it ambiguous. `[scale.microservices.services.main] file` overrides it, as `resolve_service_files` already allowed in local mode.
- The deploy resolves against the microservices config it emits pods for (`ManifestBuilder._ms_cfg()`, so SDK-declared routes resolve too), does so first thing in `generate_manifests` (before the client build, binary download, or seal), and bakes the result into every service pod's `JAC_SV_FILE`. Any resolution error propagates: `--dry-run` and the SDK `preview()` fail loud with the same message as a real deploy. `pack_jab` applies the same rule to the project's `jac.toml` as the seal-time invariant.
- The entrypoint `find` is deleted. The pod refuses to boot when `JAC_SV_FILE` is unset or names a file the extracted bundle does not contain, with a named error instead of guessing. In-pod JSON parsing was deliberately avoided: the official jac image ships no system `python3` (the embedded runtime is jac-internal), which the kind e2e's official-image leg caught.
- The #8097 entry-module guard now uses `excludes()` rather than `parks()` only, so an entry module dropped by the baseline floor (e.g. under `dist/`) also fails the pack instead of silently shipping nothing.
- `JAC_SV_FILE` joins `JAC_SV_NAME` and `JAC_SV_*_URL` as protected env keys, so overlay env cannot point a pod at an arbitrary file.

## Mixed-version window

Additive: no new bundle member, no format change; old pods never see new bundles at all since object keys are content-addressed and baked into init-container args at deploy.

## Behavior changes (intended, per the epic's risk register)

- Parked/missing declared service modules fail the deploy naming file and setting, instead of crash-looping the fleet.
- Duplicate `<name>.jac` basenames with no explicit `file` key fail the deploy listing candidates; previously an arbitrary one booted per node.
- A fleet whose implicit `main` module cannot resolve fails the deploy; previously the main pod crash-looped in-cluster.

## Tests

`test_bundle_services.jac` (18): resolution rules, entry-point precedence over same-named strays, `services.main.file` override, SDK-shaped configs and the `/__main__` marker, every fail-loud case, the D3 guard, and a sealed .jab that ships exactly the sources the resolution ran against (temp trees are reaped at exit). `test_pod_env.jac` (+7): the entrypoint is executed with a stub `jac` on PATH (boots exactly the resolved module, refuses a missing file, refuses to guess with no `JAC_SV_FILE`), the pod env carries the resolved file for the routes the deploy emits (root wins over nested), and a routed service with no module fails `generate_manifests`, dry-run included. Regression floor green locally: the full `tests/microservices` and `tests/deploy` dirs plus `test_pod_env`. The kind e2e fixture (`k8s_e2e` jac-shop) resolves cleanly under the new rules.

## Notes

- The issue's "stray `.jac` planted on the PVC" criterion is already structurally satisfied at head (per-pod emptyDir populated from a content-addressed tarball; no boot-time tree compile since #8097). The surviving form of that threat is the shadowing case above, which the deterministic resolution closes.
- No textual collision with #8128 (WI-1), but both add `BundleSourceFilter.matching_ignore_pattern` / `exclusion_reason` with matching semantics; whoever lands second collapses the duplicate on rebase.


MusabMahmoodh (commented):
Deleting the in-pod `find` and deciding the service file host-side is the right direction, and the mixed-version reasoning holds (the old entrypoint's `find` only matches `*.jac`, so an old pod ignores the new member).

Can this break something already running: yes, two ways. A routed app deployed as `jac deploy app.jac` whose `jac.toml` has no `[project] entry-point` now fails the pack demanding `main.jac` (comment 1) - `entry_point` defaults to `"main.jac"` at config.impl.jac:1074, while the realizer resolves that pod's file from `os.path.basename(filename)`. And any project with a top-level `manifest.json` now fails the pack outright (comment 3).

Is it a whole solution: not yet, and the dead-code answer is the same finding. `manifest.json` is written into the .jab but nothing reads it - the pod boots from the `JAC_SV_FILE` baked into its env, and the PR deliberately avoids parsing JSON in-pod. So the epic's declared source of truth ships as a write-only artifact, while the value pods actually boot from is computed a second time, separately, from the raw toml (comment 2). Those two are the same computation done twice; that is what stops "consistent by construction" from being true today.

jac/jaclang/scale/injector/bundle.jac
```diff
@@ -166,6 +202,203 @@ def bundle_source_files(base: Path) -> list[Path] {
...
+    routes = _declared_ms_section(str(base)).get("routes");
+    routes = routes if isinstance(routes, dict) else {};
+    for val in routes.values() {
+        sval = str(val or "").strip();
+        if sval and ("/" + sval.strip("/")) in ("/", "/__main__") {
+            return "";
+        }
+    }
+    rel = entry_module_rel(str(base));
+    return rel.name if rel is not None else "main.jac";
```
MusabMahmoodh: This map already exists. `resolve_service_files` + `ensure_main_service` (routing.jac:88-112) produce service -> file, and `kubernetes_realizer.jac:35` fills `ms_config["service_files"]` with it using `os.path.basename(filename)` - the file the user actually passed to the deploy. Re-deriving it from the raw `jac.toml` makes the two disagree in a real case: `jac deploy app.jac` with no `[project] entry-point` (which defaults to `"main.jac"`) makes the realizer intend `main -> app.jac`, while this returns `"main.jac"` and `_default_service_member` fails the pack telling the user to add a module the deploy never meant to boot.

Threading the already-resolved `service_files` into the pack - validate and locate against `member_set` rather than re-derive - is less code and removes the second source of truth, which is what the "consistent by construction" claim needs to actually hold.

jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac
```diff
@@ -467,6 +468,27 @@ obj ManifestBuilder {
...
+            return self._bundle_services_cache;
+        }
+        resolved: dict[str, str] = {};
+        if self.project_dir {
+            try {
+                import from jaclang.scale.injector.bundle { resolved_bundle_services }
+                resolved = resolved_bundle_services(self.project_dir);
+            } except Exception as exc {
+                if self.logger {
+                    self.logger.warning(
```
MusabMahmoodh: Every error this PR introduces is a deliberate fail-loud (parked, missing, ambiguous), and this handler turns all of them into a log line plus an empty map - so the pod ships with no `JAC_SV_FILE` and the new entrypoint exits 1, i.e. crash-loops the fleet, which is the outcome the PR removes.

It's masked on the real path because `_prepare_injector` -> `pack_jab` runs first (manifest_builder.jac:312) and raises. It is *not* masked on `--dry-run`, which skips `prepare()` entirely (binary_injector.jac:90-99): dry-run prints a warning and emits manifests for pods that would exit 1, where a real deploy refuses. Dry-run is exactly where the operator wants the loud version. Let the ValueError through and keep the catch for genuinely unexpected types.

jac/jaclang/scale/injector/bundle.jac
```diff
@@ -15,6 +15,7 @@ glob logger = logging.getLogger(__name__);
 
 glob APP_VOLUME_NAME: str = "jac-app",
      APP_MOUNT_PATH: str = "/app",
+     BUNDLE_MANIFEST_NAME: str = "manifest.json",
```
MusabMahmoodh: The bundle already namespaces everything it generates - `_precompiled/`, `_vendor/`. `manifest.json` at a project root is a common file (web app manifest among others), so reserving that exact name is a break people will hit. `_manifest.json` (or `_bundle/manifest.json`) makes the collision impossible and deletes `assert_manifest_name_free`, its test, and the release-note caveat along with it.

jac/jaclang/scale/injector/bundle.jac
```diff
@@ -166,6 +202,203 @@ def bundle_source_files(base: Path) -> list[Path] {
...
+        f"Service '{name}' points [scale.microservices.services.{name}] file "
+        f"at '{explicit}', but the bundle does not ship it "
+        f"({excluded.exclusion_reason(Path(rel))}), so every '{name}' pod "
+        "would crash-loop on boot. Ship the file, or point the service at a "
+        "module the bundle ships."
+    );
+}
+
+
+def _default_service_member(
```
MusabMahmoodh, line 450: With no root `svc.jac`, a single `tests/svc.jac` in the shipped tree still becomes the service module - deterministic now instead of readdir order, which is the win, but it's still the stray from the issue's own example. Was leaving that in deliberate (any unique basename is fair game) rather than restricting the implicit search to non-test trees? Worth stating in the docs either way, since it's the rule people hit when they add a fixture.

jac/jaclang/scale/tests/microservices/test_bundle_manifest.jac
```diff
@@ -0,0 +1,247 @@
...
+glob _JAC_BIN: Path = Path(__file__).parents[4] / "zig-out" / "bin" / "jac",
+     _ROUTED_TOML: str = (
+         "[project]\nname = \"mini\"\n\n[scale.microservices.routes]\n"
+         "svc = \"/svc\"\n"
+     ),
+     _PING: str = "def:pub ping -> str { return \"pong\"; }\n";
+
+
+def _app(toml: str, files: dict[str, str], jacignore: (str | None) = None) -> Path {
+    d = Path(tempfile.mkdtemp(prefix="jac-manifest-"));
```
MusabMahmoodh: Driving the real `pack_jab` over real temp trees with no mocks is right, and these do fail on main (no resolution existed, so `caught` stays empty). Three things:

- Every `_app()` leaks its `mkdtemp` tree - 14 per run, none cleaned up. A `finally` with `shutil.rmtree(..., ignore_errors=True)` in the helper is cheap.
- The entrypoint tests assert on the text of the shell script (`"find " not in args`), which is the shape of the fix rather than the outcome. Does the kind e2e leg assert which module a pod actually booted - two same-named modules, declared one wins? If not, that's the assertion the issue deserves.
- Nothing covers a resolution error reaching `_bundle_services`, or the dry-run path that reaches it without `pack_jab` having run.

MusabMahmoodh (commented):
Came back to this one and the head is the same as when I last read it, so my earlier comments still stand. One new thing though, and it changes how I read the rest of the PR.

I went looking for what reads `manifest.json` and I can't find anything. The deploy doesn't: `_bundle_services` calls `resolved_bundle_services()`, which re-walks the project tree and re-runs `resolve_service_members` from scratch (`bundle.jac:240`), so `services` in the manifest and `JAC_SV_FILE` in the pod env are two independent computations over the filesystem at two different moments, not one declaration consumed twice. The pod doesn't either, deliberately, since there's no `python3` in the image. Grepping the tree outside tests only turns up the PWA manifest and `.jac/serve_manifest.json`.

That matters because #8105's re-scope is explicit that manifest.json is dropped from v1 and "what remains is the valuable half". The valuable half is already here and it works: host-side resolution, `JAC_SV_FILE` always set, the `find` gone. Dropping `_bundle_manifest_bytes`, `assert_manifest_name_free` and `BUNDLE_MANIFEST_NAME` takes about 110 lines and the reserved-name break out with them. If the manifest is meant to land with the I1/I2 validator that actually reads it, would it be better as that PR's first commit?

Still open from last time, in short form:

- `manifest_builder.jac:468` `_bundle_services` catches every fail-loud this PR adds and turns it into a warning plus an empty map, so the pod ships with no `JAC_SV_FILE` and exits 1. `pack_jab` masks it on the real path but not on `--dry-run`, which skips `prepare()` entirely, and dry run is where you most want the loud version.
- `bundle.jac:248` re-derives a map `resolve_service_files` + `ensure_main_service` already produce (`routing.jac:90-142`), and the two disagree for `jac deploy app.jac` with no `[project] entry-point`. Local mode and k8s now resolve the same jac.toml differently.
- `bundle.jac:223` `_implicit_main_wanted` returns `rel.name`, so a root `app.jac` wins over a configured `src/app.jac`. `entry_module_rel()` already hands you the full path.
- `test_bundle_manifest.jac:672` returns early when `zig-out/bin/jac` isn't built, so the one test that opens a real sealed .jab is green whether or not it ran. Does the scale job build the binary first?

The entrypoint change itself is the right fix, and the mixed-version reasoning holds up: the old `find` only matches `*.jac`, so an old pod ignores the new member.

jac/jaclang/scale/injector/bundle.jac
```diff
@@ -166,6 +202,203 @@ def bundle_source_files(base: Path) -> list[Path] {
...
+    }
+    raise ValueError(
+        f"Service '{name}' resolves to '{wanted}', but the project has no "
+        f"such module, so every '{name}' pod would crash-loop on boot. Add "
+        f"the module, or set [scale.microservices.services.{name}] file."
+    );
+}
+
+
+def _bundle_manifest_bytes(
```
MusabMahmoodh: I went looking for what reads this and couldn't find anything. The deploy doesn't: `_bundle_services` calls `resolved_bundle_services()`, which re-walks the tree and re-resolves from scratch, so the manifest's `services` and the pod's `JAC_SV_FILE` are two independent computations at two different moments rather than one declaration consumed twice. The pod doesn't either, deliberately, since the image has no `python3`.

#8105's re-scope is explicit that manifest.json is dropped from v1 and "what remains is the valuable half" - and the valuable half is already here and working. Dropping this, `assert_manifest_name_free` and `BUNDLE_MANIFEST_NAME` takes ~110 lines and the reserved-name break out with them. If it's meant to land with the I1/I2 validator that reads it, is it better as that PR's first commit?

jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac
```diff
@@ -467,6 +468,27 @@ obj ManifestBuilder {
         }
     }
 
+    def _bundle_services -> dict[str, str] {
```
MusabMahmoodh, line 710: `except Exception -> warning -> {}` turns every fail-loud this PR adds (parked, missing, ambiguous) into a log line plus a pod that ships with no `JAC_SV_FILE` and exits 1 on boot, which is the crash-loop you're removing.

`pack_jab` masks it on the real path because `_prepare_injector` runs first, but not on `--dry-run`, which skips `prepare()` entirely (`binary_injector.jac:89-99`). Dry run is exactly where you want the loud version, and it's the one path where this handler is the only resolution that runs. Can the ValueError just propagate?

jac/jaclang/scale/injector/bundle.jac
```diff
@@ -166,6 +202,203 @@ def bundle_source_files(base: Path) -> list[Path] {
...
+
+def resolved_bundle_services(project_dir: str) -> dict[str, str] {
+    base = Path(project_dir).resolve();
+    excluded = source_filter_for(base);
+    members = [path.relative_to(base).as_posix() for path in bundle_source_files(base)];
+    return resolve_service_members(base, members, excluded);
+}
+
+
+def resolve_service_members(
```
MusabMahmoodh, line 329: This map already exists: `resolve_service_files` + `ensure_main_service` (`routing.jac:90-142`), and `kubernetes_realizer.jac:34` already fills `ms_config["service_files"]` with it from the filename the user actually passed.

The two disagree in a real case: `jac deploy app.jac` with no `[project] entry-point` makes the realizer intend `main -> app.jac`, while this path fails the pack demanding `main.jac`. And local microservice mode keeps routing.jac's rule (root `<name>.jac`, no nested search) while k8s now accepts `src/svc.jac`, so the same jac.toml resolves differently per mode.

jac/jaclang/scale/injector/bundle.jac
```diff
@@ -166,6 +202,203 @@ def bundle_source_files(base: Path) -> list[Path] {
...
+    for name in routes {
+        svc = services.get(str(name)) or {};
+        svc = svc if isinstance(svc, dict) else {};
+        out[str(name)] = str(svc.get("file") or "");
+    }
+    return out;
+}
+
+
+def _implicit_main_wanted(base: Path, declared: dict[str, str]) -> str {
```
MusabMahmoodh: This returns `rel.name`, but `entry_module_rel()` hands you the full relative path - so a root `app.jac` wins over a configured `src/app.jac` via the `wanted in member_set` shortcut. For a PR whose whole premise is "stop booting by basename", the implicit main is the one service still doing it:

```suggestion
    rel = entry_module_rel(str(base));
    return rel.as_posix() if rel is not None else "main.jac";
```

The "implicit main follows the configured entry point" test passes today only because the fixture has no root `app.jac` next to `src/app.jac`.

jac/jaclang/scale/tests/microservices/test_bundle_services.jac
```diff
@@ -0,0 +1,247 @@
...
+    assert first != _bundle_manifest_bytes(base, grown, services);
+    manifest = json.loads(first.decode("utf-8"));
+    assert manifest["format"] == 1;
+    assert manifest["services"] == {"svc": "svc.jac", "main": "main.jac"};
+    assert manifest["members"] == sorted(members);
+}
+
+
+test "a sealed .jab carries the manifest naming exactly what ships" {
+    if not _JAC_BIN.is_file() {
```
MusabMahmoodh, line 313: This makes the one test that opens a real sealed .jab pass silently wherever `zig-out/bin/jac` isn't built, and a skip that returns green looks identical to a pass in the CI log. Does the scale job build the binary before this runs? If not, it's the assertion the PR most needs and the one least likely to execute.

(Also still open: every `_app()` leaks its `mkdtemp` tree, 15 per run now.)

MusabMahmoodh (commented):
Head hasn't moved since I last read this, so my earlier four still stand and I've re-anchored them below. The new one is on `_bundle_manifest_bytes`, and it reframes the rest of the diff for me.

jac/jaclang/scale/injector/bundle.jac
kashmithnisakya: agreed and dropped: nothing read it and #8105's re-scope had already cut it, so pack_jab now only validates against the shipped set, nothing new lands in the .jab, and the reserved-name guard went with it (e3bcdd6e5).

jac/jaclang/scale/injector/bundle.jac
kashmithnisakya: moot now: manifest.json is gone from the bundle, so nothing is reserved and assert_manifest_name_free, its test and the release-note caveat are deleted (e3bcdd6e5).

jac/jaclang/scale/injector/bundle.jac
kashmithnisakya: fixed: the implicit main now resolves the entry module at its exact path (and honors services.main.file), so a root app.jac can't shadow src/app.jac and a lib/app.jac can't make it ambiguous; both cases are tested (e3bcdd6e5).

jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac
kashmithnisakya: fixed: _bundle_services no longer catches anything and generate_manifests resolves first thing, so a bad route fails --dry-run (and the sdk preview) before any download or build; that handler would also have crashed on StandardLogger, which has warn not warning. test added (e3bcdd6e5).

jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac
kashmithnisakya: same fix as above: the ValueError propagates, dry-run and the sdk preview fail with the deploy's own message, and the resolution now runs before the client build / binary download / seal (e3bcdd6e5).

jac/jaclang/scale/injector/bundle.jac
kashmithnisakya: checked this one: normalize_microservices_config already pins main -> main.jac before the realizer runs (routing.jac reads cfg?.entry_point, which doesn't exist on JacConfig), so basename(filename) is dead in every mode and there's no app.jac intent to disagree with, and that map is entry-point-blind. what i did take from it: the pack now reuses resolve_routes plus a shared needs_main_service instead of re-deriving, and the deploy resolves against the ms config it actually emits pods for, sdk routes included (e3bcdd6e5).

jac/jaclang/scale/injector/bundle.jac
kashmithnisakya: see the reply on the first thread: the app.jac disagreement doesn't occur (main is pinned to main.jac before the realizer runs), the duplication is gone via resolve_routes/needs_main_service, and the local-mode root-only rule is pre-existing and out of scope here.

jac/jaclang/scale/injector/bundle.jac
kashmithnisakya: deliberate: a unique shipped basename is fair game (mirrors the old find, now deterministic and refused when ambiguous, and #8105's two-candidates acceptance needs a search); the order is now documented in runtime/docs.md and a new file row in the k8s reference (e3bcdd6e5).

jac/jaclang/scale/tests/microservices/test_bundle_manifest.jac
kashmithnisakya: done: temp trees live under one root reaped at exit, the entrypoint is executed with a stub jac on PATH (boots exactly JAC_SV_FILE, refuses a missing file, refuses to guess) instead of grepping the script, root-vs-nested is asserted through the pod env, and a routed service with no module now fails generate_manifests, dry-run included (e3bcdd6e5).

jac/jaclang/scale/tests/microservices/test_bundle_services.jac
kashmithnisakya: ci's test-scale job downloads the jac-binary artifact to jac/zig-out/bin (same self-skip convention as test_fat_bundle etc.), so the sealed test does run there; it now asserts the .jab ships exactly the sources the resolution ran against, and the mkdtemp leak is fixed (e3bcdd6e5).

MusabMahmoodh (commented):
Re-review at head (e3bcdd6e5 plus the merge from main). Everything from my 08-15 and 08-18 passes landed: `manifest.json` and `assert_manifest_name_free` dropped along with the reserved-name break, `_bundle_services` no longer swallowing so `--dry-run` and the SDK preview fail loud, the implicit `main` resolving the entry module at its exact path, and the entrypoint tests executing the real script with a stub `jac` on PATH instead of grepping its text. That last one is the right shape for this kind of fix.

Two of the five below matter; the rest are small.

**1. The two maps agree only because a config read is broken.** `jac/jaclang/scale/runtime/routing.jac`, `ensure_main_service`: `entry_file = str(cfg?.entry_point or 'main.jac')`.

You closed the second-source-of-truth thread with "main is pinned to `main.jac` before the realizer runs", and that is true today, but only because this line reads `entry_point` off `JacConfig`, where it actually lives at `config.project.entry_point`. The attribute error is swallowed by the `except Exception` below it, so `ensure_main_service`'s entry-file support has never once worked. `normalize_microservices_config` has the same broken read for the client entry (`get_config().entry_point or 'main.jac'`, inside its own try). Fix either typo and local mode resolves `main -> src/app.jac` while nothing else changes, and the two rules disagree again. I would rather this PR pin the agreement with a test than inherit it from an exception.

**2. The route-less deploy still boots by basename, unvalidated.** `manifest_builder.jac`, `generate_service_pod_specs`: `solo_cfg["file"] = app_config.file_name`.

`file_name` is `os.path.basename(filename)` (`kubernetes_realizer.jac`), and `generate_manifests` calls `_bundle_services()` before `generate_service_pod_specs` invents the solo route, so the cache is already `{}` when the solo pod's env is built and its `JAC_SV_FILE` never passes through `resolve_service_members`. `jac deploy src/app.jac` therefore ships `JAC_SV_FILE=app.jac` for a bundle that has the module at `src/app.jac`, and the new entrypoint exits 1 with "the shipped bundle does not contain": the exact crash-loop this PR removes, on the single-service path. Can the solo route resolve through the same rule, or at least validate that basename against the member set before it is baked in?

**3. The boot error names the one cause it cannot be.** `manifest_builder.jac`, the entrypoint's "no service file was resolved for $JAC_SV_NAME at deploy time; re-deploy with a matching jac version".

Your own mixed-version section argues an old pod never sees a new bundle, since object keys are content-addressed and baked into init-container args, so version skew is precisely what this cannot be. Whoever reads that in pod logs goes to check versions instead of their routes or their overlay env. Name the state instead: `JAC_SV_FILE` is unset for this pod.

**4. Two things still decide `JAC_SV_FILE`.** `manifest_builder.jac:733-735`.

`resolve_service_members` already resolves and validates the explicit `services.<name> file`, so taking `svc_cfg["file"]` first means the raw declared string wins over the one we proved ships. Flipping it leaves one decision point:

```jac
svc_file: str = self._bundle_services().get(svc_name, "");
if not svc_file {
    svc_file = str(svc_cfg.get("file") or "");
}
```

**5. An unparseable jac.toml silently skips the seal-time invariant.** `bundle.jac`, `_declared_ms_section`.

`except Exception -> warning -> {}` means `pack_jab`'s new `resolve_service_members` call resolves nothing and the .jab seals as though no services were declared: the same swallowed-config-failure shape the `_bundle_services` handler had before this round. Is there a case for sealing at all when the toml cannot be read?

MusabMahmoodh (commented):
Re-reviewed at head. Approving with comments. All nine of my earlier threads are settled, and I want to withdraw one of my own first: my solo-route basename claim was wrong. `kubernetes_realizer.jac:61`/`:99` and `sdk/spec.jac:98` both return `(dirname, basename)`, so the bundle base really is the file's directory. Sorry for the noise on that one.

Deleting the entrypoint `find` and replacing it with a named refusal is the right mechanism fix, and running the resolution at the top of `generate_manifests` rather than inside the pack is what makes `--dry-run` genuinely useful here: an operator finds out about a parked service module before anything is downloaded or sealed.

Four things.

**1. `injector/bundle.jac:798` - two configs feed one rule**

`_bundle_services()` resolves against `self._ms_cfg()`, which is the config this deploy emits pods for, while `pack_jab` resolves against `_declared_ms_section`, which re-reads `jac.toml`. On the CLI path those are the same dict and this is invisible. On the SDK path they are not: `DeploySpec.to_microservices_config()` (`sdk/spec.jac:135-153`) builds routes from the spec and never reads `jac.toml`, so a project whose `jac.toml` still declares a route the SDK caller is not deploying fails the pack over a service that gets no pod, and skips validating one that does. Could the injector carry the config the deploy actually used, the way it already carries the seal bytes?

**2. `runtime/routing.jac:113` - you are right, and the read it rests on is broken in two places**

You are right that main is pinned before the realizer runs. The reason it holds, though, is a bug: `cfg?.entry_point` reads `entry_point` off `JacConfig`, which does not have it. It lives on `ProjectConfig` (`project/config.jac:18-31`), reached as `config.project.entry_point` - which is exactly how this same repo reads it correctly at `injector/bundle.jac:172` and `:175`. So the `except Exception` below swallows an AttributeError on every call and `entry_file` is always `'main.jac'`.

There is a second copy at `config/impl/config_loader.impl.jac:579` (`get_config().entry_point or 'main.jac'`), same wrong attribute, same swallowing `except Exception`, and it is on main today rather than introduced here.

Net effect: a project that sets `entry-point = "src/app.jac"` has it silently ignored on both paths. This PR is what turns that from harmless into a divergence, because a pod would now boot `src/app.jac` while local microservice mode still boots `service_files['main']`, which is `main.jac` (`orchestrator.jac:129`). Fixing the read also needs the `os.path.basename` on `:119` to go.

I am not asking you to take that on inside this PR. Can this one pin today's agreement with a test, and I will carry the two-site fix separately?

**3. `deploy/target/kubernetes/manifest_builder.jac:1155` - the boot error names the one cause it cannot be**

After this PR `generate_manifests` raises before any pod spec with an unresolved service is written, and your own body argues an old pod never sees a new bundle. So version skew is precisely what an empty `JAC_SV_FILE` cannot be any more. What is left is a hand-edited manifest or an overlay that dropped the key, and this line sends whoever is reading pod logs at 2am to go check jac versions. Naming the actual state reads better: `JAC_SV_FILE is not set on this pod`.

**4. `injector/bundle.jac:210` - unparseable jac.toml skips the invariant you just added**

`except Exception` to a warning to `{}` means `pack_jab`'s new `resolve_service_members` call resolves nothing and the `.jab` seals as if no services were declared. The file already does this for the fat-bundle and seal-timeout checks (`:691`, `:717`), so it is the house style, but those gate optional behaviour and this one gates a fail-loud. Is there a case for sealing at all when the toml cannot be read?

One more for the release note: `JAC_SV_FILE` set in an overlay `env` is now silently dropped at `manifest_builder.jac:779`. Worth a line.

jac/jaclang/scale/injector/bundle.jac
```diff
@@ -227,6 +271,187 @@ def bundle_source_files(base: Path) -> list[Path] {
...
+        }
+    }
+    return resolved;
+}
+
+
+def _implicit_main_member(
+    base: Path, member_set: set[str], excluded: BundleSourceFilter
+) -> str {
+    rel = entry_module_rel(str(base));
```
kugesan1105: `_implicit_main_member` mixes two coordinate systems, so a nested entry module fails the deploy with a message that is factually wrong.

`entry_module_rel` returns `[project] entry-point` **relative to the directory holding jac.toml** (`get_config_for_path` walks up to find it). It is then compared against `member_set`, whose paths are **relative to `base`**, and `base` is the deploy's `code_folder`, which is `os.path.dirname(filename)` (`kubernetes_realizer.jac:61`). The two only coincide when the deployed entry file sits at the project root.

Validated on this branch. Project root holds `jac.toml` with `entry-point = "src/app.jac"`, modules at `src/app.jac` and `src/orders.jac`, deployed as `jac run src/app.jac --scale` so `base = "src"`:

```
entry_module_rel('src') = src/app.jac
RAISED: The routed fleet also deploys an implicit 'main' service for the project
entry module, but 'src/app.jac' does not exist, so every 'main' pod would
crash-loop on boot.
```

`src/app.jac` does exist, and it ships: it is bundle member `app.jac`. The `.is_file()` probe on line 359 checks `base / wanted`, i.e. `src/src/app.jac`, so the resolver takes the "does not exist" branch and hard-fails a deploy that has nothing wrong with it. This is the case the PR description highlights ("a configured `src/app.jac`").

Suggest resolving the entry point against the project root that `get_config_for_path` actually found, then re-expressing it relative to `base` before the `member_set` lookup, and failing only when that relative path is genuinely absent.

jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac
```diff
@@ -207,6 +208,7 @@ obj ManifestBuilder {
             access_mode=self.k8s_config.bundle_access_mode
         );
         injector: PvcInjector = selection[0];
+        injector._ms_config = self._ms_cfg();
```
kugesan1105: This assignment is dead. `select_injector` calls `injector.prepare(...)` (`binary_injector.jac:111`) before it returns, and `BinaryInjector.prepare` calls `super.prepare` which is `PvcInjector.prepare`, which is where `pack_jab(..., self._ms_config)` runs. By the time this line executes on the returned object, the bundle is already packed.

`prepare()` has exactly one call site, so nothing re-packs later.

The effect is that `pack_jab` always receives `ms_config=None` in production and falls back to `_declared_ms_section(project_dir)`, i.e. the raw `jac.toml`. For routes declared in `jac.toml` that happens to resolve identically, so tests pass. For SDK-declared routes the raw section has no routes, `resolve_service_members` returns `{}`, and the seal-time invariant silently no-ops. That makes the PR description's "`pack_jab` applies the same rule ... as the seal-time invariant" untrue for exactly the config shape `_bundle_services()` was extended to cover.

No user-visible bug today, since `_bundle_services()` runs first in `generate_manifests` and raises on the same conditions. But either the config needs threading in before `prepare()` runs (pass it into `select_injector`), or this line and the `pack_jab` parameter should go.

kugesan1105 (dismissed):
Reviewed the mechanism end to end against #8105, running the resolver and the pod entrypoint rather than reading them. The approach is right, and better than what the issue asked for: resolving against `bundle_source_files` instead of wiring in the config-only `resolve_service_files` map is the correct seam, and it is what makes the `find` safe to delete.

Confirmed working on a fixture with a nested service and a non-root entry point:

```
resolved: {'orders': 'services/orders.jac', 'main': 'src/app.jac'}
--- orders JAC_SV_FILE = 'services/orders.jac'
--- main   JAC_SV_FILE = 'src/app.jac'
```

On `main` both of those were unset, `orders` booted only because of the `find`, and `main` crash-looped. A planted `tests/orders.jac` decoy now fails the deploy naming both candidates, which is the ambiguity criterion from #8105.

I merged `main` into the branch and pushed (0146848e2). #8596 retired `jac start`, so both conflicts were in the entrypoint: kept this branch's fail-loud boot, dropped the `find`, exec `jac run --serve`. `test_pod_env.jac`'s stub-argv assertion follows the rename. All touched suites pass locally (`test_pod_env` 36, bundle/pvc/jacignore 42).

Two things to fix before merge, both small, inline below.

Separately, local mode was left on the old path and is still broken by the same root cause: `resolve_service_files` still clobbers the resolved `main`, and a local fleet boots nonexistent modules while the gateway reports success. Filed as #8686 rather than growing this PR.

jac/jaclang/scale/injector/bundle.jac
kashmithnisakya: fixed: the entry point now resolves against the jac.toml root and is re-expressed relative to the deploy base before the member lookup, so a nested deploy like yours passes, and an entry module outside the deployed tree fails loud; both cases tested (ca74874a8).

jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac
kashmithnisakya: fixed: the config now rides into select_injector and lands on the injector at construction, so pack_jab seals against the ms config the deploy emits pods for, sdk routes included; the dead assignment is gone and the wiring is tested at prepare time (ca74874a8).

MusabMahmoodh (commented):
Re-reviewed at ca74874a. The two things kugesan raised are properly fixed - `_ms_config` riding in through `select_injector`'s signature instead of being assigned onto the injector after `prepare()` had already run is the right shape, and splitting `_entry_module_spec` out so the entry point is resolved against `config.project_root` rather than the deploy cwd fixes the coordinate mixing at the source rather than at one call site.

The error text in `_implicit_main_member` is the good kind - it names the module, the tree it resolved outside of, what will happen to the pods, and the two ways out. That is the standard.

**Which is what makes the sibling path stand out.** `entry_module_rel` now folds two different outcomes into `None`:

```jac
try {
    return (proj_root / entry_path).resolve().relative_to(Path(project_dir).resolve());
} except ValueError {
    return None;
}
```

`None` means "no entry point configured" *and* "the entry point resolves outside the deployed tree". Its caller is `assert_entry_module_ships`, which opens with `if rel is None ... { return; }` - so in exactly the subdirectory-deploy case this PR exists to fix, the guard silently skips instead of firing. Same condition, two functions, opposite behavior: `_implicit_main_member` raises a paragraph, `assert_entry_module_ships` returns quietly and the deploy proceeds toward the crash-loop it was written to prevent.

That is the #7900 theme-1 shape - a decision point that answers "fine" when it means "I could not tell". Cheapest fix is to let the caller see the difference:

```jac
def assert_entry_module_ships(base: Path, excluded: BundleSourceFilter) {
    spec = _entry_module_spec(str(base));
    if spec is None {
        return;
    }
    (proj_root, entry_path) = spec;
    try {
        rel = (proj_root / entry_path).resolve().relative_to(base.resolve());
    } except ValueError {
        raise ValueError(
            f"The project entry module '{entry_path.as_posix()}' resolves "
            f"outside the deployed tree '{base}', so the bundle cannot ship "
            "it. Deploy from the project root, or point [project] entry-point "
            "at a module inside the tree."
        );
    }
    ...
}
```

**Second, smaller.** The same function moves from `excluded.parks(rel)` to `excluded.excludes(rel)`. That widens the failure from "parked by .jacignore" to every exclusion reason, so an entry module living under `dist/` or `build/` now fails a deploy that previously went through and shipped a broken bundle. I think that is the right call and the new message handles it, but it is a behavior widening riding inside a fix for something else - worth a line in the PR body so it is not a surprise for anyone whose entry point happens to sit in a directory `_EXCLUDE_DIRS` covers.

Nothing else from my Aug 19 round is outstanding.

Issue #8173, opened 2026-08-14
RFC: One execution contract, two deployment strategies


MusabMahmoodh:
## The promise being broken

jac-scale's central promise is that topology is a *deployment* decision:
write the app once, choose single-process or microservices in `jac.toml`.
The routes table moves modules between pods; the program means the same thing.

Measured against a real application deployed four ways (kind microservice,
hosted microservice, local single-process, hosted single-process), the promise
does not hold. The same source changes meaning with the topology:

| Semantics | single-process | microservice | Evidence |
|---|---|---|---|
| Endpoint registration | entry module only; other modules' `def:pub` silently serve nothing | every routed module registers | #7695 (open, third recurrence; #6221 closed the def:pub case in May) |
| The `include` escape hatch | publishes every absorbed name (`POST /function/uuid4`) | n/a | app 01 F33 |
| `report` in a `-> Generator` | returns `None` | streams over RPC as SSE | app 01 F34 |
| Typed return across a module boundary | real archetype instance | `dict` with `_jac_type` tag left unhydrated | app 01 F19; does NOT reproduce in-process |
| Route shape | `/function/<fn>` | `/<prefix>/function/<fn>`, other form 404s/503s | the stream_reply 404; F31 (one fault, two status codes) |
| Client route resolution | generated stubs adapt; hand-rolled fetch cannot | same | F14 - forced hand-rolled streaming call broke on topology change |
| Failure statuses | 503 `SERVICE_UNAVAILABLE "transport_error"` | 502 `GATEWAY_TIMEOUT` | F31: same dead upstream, two codes by URL shape |

None of these are exotic. An app that (a) has more than one module, (b) streams,
(c) returns a typed object, or (d) has a client that calls a streaming endpoint
hits one of them the first day. App 01 hit all of them.

## Why this is one design gap, not seven bugs

Each row exists because a behaviour was implemented twice - once in the
in-process dispatch path and once in the RPC/gateway path - with no shared
definition to conform to. There is no document, test suite, or interface that
states what `report`, endpoint registration, route shape, or a cross-module
call *means* independent of topology. So the two implementations drift, each
drift ships silently, and fixing any single row (e.g. #6221) does not prevent
the next recurrence (#7695) because nothing enforces the invariant.

The proposal is to create the missing artifact: a single execution contract
that both deployment strategies implement, with a conformance suite that runs
every contract test against BOTH topologies in CI.

## The contract (proposed, v1)

[...]

PR #8174, opened 2026-08-14
fix(runtime): run streamed generator bodies under the caller's request context

### LINKED ISSUE #8136: A generator served as SSE executes against a different root: streamed handlers read an empty graph and their writes are lost
## Summary

A `def:pub` that returns a generator is served as SSE. The generator body runs
against a root that is not the caller's: it reads an empty graph, and anything
it writes is not visible to subsequent requests. A plain function in the same
module, in the same process, on the same root, at the same moment, sees the real
data.

Nothing is raised and nothing is logged. The client receives a complete,
correctly framed SSE stream. The only way to notice is to read the data back.

## Reproduction

Complete program, 24 lines, single process. No scale, no cluster, no
microservices, no byLLM.

```jac
"""Minimal repro: is `root` the caller's root inside a streamed generator?"""

import from typing { Generator }


node Item {
    has v: int = 0;
}


def:pub seed -> int {
    root ++> Item(v=1);
    return len([root-->][?:Item]);
}


def:pub count_direct -> int {
    return len([root-->][?:Item]);
}


def:pub count_streamed -> Generator[dict, None, None] {
    yield {"count": len([root-->][?:Item])};
    yield {"count": len([root-->][?:Item])};
}
```

```
$ jac start main.jac --port 8771
```

```
$ curl -sX POST localhost:8771/function/seed          -d '{}'
{"ok": true, "data": {"result": 2, "reports": []}, "error": null}

$ curl -sX POST localhost:8771/function/count_direct  -d '{}'
{"ok": true, "data": {"result": 2, "reports": []}, "error": null}

$ curl -sNX POST localhost:8771/function/count_streamed -d '{}'
data: {"count": 0}

data: {"count": 0}

event: end
data: {}

$ curl -sX POST localhost:8771/function/count_direct  -d '{}'
{"ok": true, "data": {"result": 2, "reports": []}, "error": null}
```

`count_direct` returns **2** before and after. `count_streamed`, between those
two calls, returns **0**. Same process, same anonymous root, same graph.

## Expected

The generator body should run under the same root as the request that created
it, so `count_streamed` yields the same count `count_direct` returns.

## Impact

Reads are the visible half; **writes are the damaging half**. Anything a
streaming handler does after its first `yield` - persisting the completed
response, recording usage, writing an audit row, invalidating a cache - operates
on a root nobody will read back.

This was found in a streaming chat application, where it means: the tokens
stream correctly to the browser, the client renders the whole reply, and the
conversation is empty on the next page load. The A/B in that app was the same
turn implementation reached two ways - consumed in-request it persisted a
250-character assistant message and cleared its state flag; handed to the
serving layer it delivered 38 identical chunks and persisted nothing. Its own
error guard logged `no such conversation: <id>` for a conversation the same
request had just created.

`report`ing a generator is the documented way to stream, and stream-then-persist
is the defining shape of an LLM chat application, so this sits directly under
the AI-era use case jac-scale is aimed at.

It also cannot be worked around in application code, because the handler cannot
tell it has lost its root: a read simply returns nothing.

## Environment

- jac **0.36.1** release binary, `jac-0.36.1-linux-x86_64`, sha256
  `3edaa74d5f6124c59be7183a58b823da0c325344c8666406332a57e4a5182c71`
- Reproduced in single-process `jac start`, and independently in a deployed
  microservice fleet on kind (Postgres, 6 pods) with the same symptom
- Isolated `XDG_CACHE_HOME`; `jac --version` prints no dev-mode banner

## Suggested direction

Keep the request's root/auth context alive for the lifetime of the streamed
response, so the generator body runs under the caller it was created for.

Failing that, make it loud: raise when a graph operation runs with no root
context rather than resolving to an empty one. The silent empty root is what
turns this from an error into data loss.

Possibly related: #6015 (byllm streaming generator, "Token was created in a
different context") looks like the same family seen from the byllm side.

kashmithnisakya:
Fixes #8136

## Problem

A `def:pub` that returns a generator is served as SSE, but the generator body executes against a root that is not the caller's: it reads an empty graph, and anything it writes after its first `yield` is silently lost. A plain function in the same module, on the same root, at the same moment, sees the real data. Reproduced on current main (and per the issue, back to at least 0.35.1, so unrelated to the Postgres rewrite).

## Root cause

Generators are lazy; the request lifecycle is not.

1. `_run_function_with_occ` pushes the caller's `ExecutionContext` into the `_request_context` ContextVar, calls the endpoint, and tears everything down in its `finally`. For a generator function the call only creates the generator object; not one line of its body has run.
2. The serving layer iterates the escaped generator later, inside the `StreamingResponse`, after teardown. Python generators do not capture ContextVars; each `next()` runs in the caller's context at iteration time.
3. `Jac.get_context()` finds no request context and silently falls back to `JacRuntime.exec_ctx`, the process boot context, whose `user_root` resolves to the empty superroot. Reads see nothing; writes attach where no request lifecycle will ever commit them. The fallback is a fully functional context, which is why nothing raises and nothing logs.

This affects every path that funnels through `ExecutionManager`: `POST /function/*`, `POST /walker/*` (via `report gen;`), webhooks, the WebSocket endpoints, and cluster workers serving RPC calls.

## Fix

A streamed response is treated as a continuation of the request that created it. `_run_function_with_occ` now routes generator results through `_stream_in_request_context(gen, ctx)`; the handler's own teardown is unchanged and stays balanced within the handler task. A wrapper that is never iterated holds nothing.

For a **sync producer** (the common `def:pub` shape), a dedicated per-stream thread owns the generator for its whole life: it pushes the request context once in-thread, iterates, closes the producer, commits, and exits the session unit, handing items to the response through a queue with a small backpressure window. One thread per active stream is the honest cost of a blocking producer, and it keeps three failure modes out by construction:

- pulls never touch the shared request thread pool, so blocked producers cannot starve or deadlock ordinary requests;
- only the owning thread ever calls into the generator, so consumer cancellation cannot race `gen.close()` against an executing frame;
- the end-of-stream commit and the producer's cleanup run off the event loop.

An **async producer** is iterated on the loop with the context pinned around every `__anext__`, committed after clean exhaustion, and closed with the context pinned; `unit_exit` is guarded so it runs even if close raises.

Two deliberate semantics, both matching how the rest of the runtime behaves:

- The commit lands before the SSE `event: end` frame, so a clean terminator implies the stream's writes are durable. A write conflict at that commit cannot be retried (bytes are already on the wire); it surfaces as the existing framed `event: error`.
- Early termination (client disconnect, producer error) does not roll anything back. The session is process-shared, so an abort here would discard concurrent requests' uncommitted writes; instead the stream's unit simply closes and the session's normal unit model decides when remaining state is flushed.

Also fixed along the way:

- the webhook callback ran `'error' in result` on the raw manager result; for a walker reporting a generator that is a crash (`TypeError` on an async generator, silent drain on a sync one). It now has the same generator guard as the REST callbacks and streams the report.
- close-path hardening so early termination tears down promptly instead of at GC: `_sse_frame_stream_inner` closes its source in a `finally`, the native server closes the body stream when the client connection drops (and for HEAD requests carrying a stream), and `_ws_stream_result` closes its stream on all exits.

## Tests

Three regression tests in `tests/runtimelib/test_serve_streaming.jac` (in-process `JacTestClient`, no sockets):

- streamed reads see the caller's graph, not an empty root
- writes made after a `yield` are committed when the stream completes
- a walker reporting a generator streams under the caller's root

Existing streaming and serve suites pass: `test_serve_streaming.jac`, `test_serve.jac`, `test_serve_disconnect.jac`, `test_serve_bridged_endpoints.jac`, `test_serve_client.jac`, `test_sv_streaming.jac`, `test_stream_ws.jac`, `test_webhook.jac`, and the scale `server/test_serve.jac` (its dev-mode HMR test fails identically on a clean main checkout locally and is unrelated).


jac/jaclang/runtimelib/impl/server.impl.jac
```diff
@@ -449,6 +453,119 @@ async def _run_function_with_occ(
...
+        ctx.mem.unit_enter();
+        try {
+            try {
+                while True {
+                    capacity.acquire();
+                    if stopped[0] {
+                        break;
+                    }
+                    try {
+                        item = next(gen);
```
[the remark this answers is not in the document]
kashmithnisakya: true, but a thread blocked inside user code can't be interrupted from outside in python, same as a blocked sync endpoint pinning its to_thread worker today. the thread is a daemon and exits via the stopped flag the moment the producer resumes, closing the gen and releasing the unit. while it's held, only the at-zero backstop commit is suppressed; every normal request still commits explicitly, so staged writes flush with the next commit either way.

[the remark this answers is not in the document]
kashmithnisakya: took another pass at this: the unit is now held only around the end-of-stream commit, not for the stream's life, so a blocked producer holds no session state at all anymore, just its own daemon thread (which python can't interrupt, same as any blocked sync handler). also added a raw-socket regression test that RST-closes mid-SSE and asserts no unit stays held and the server keeps serving.

jac/jaclang/runtimelib/impl/server.impl.jac
```diff
@@ -449,6 +453,129 @@ async def _run_function_with_occ(
...
+                }
+            }
+            _put(('done', None));
+        } except BaseException as e {
+            _put(('error', e));
+        } finally {
+            Jac.reset_request_context(token);
+        }
+    }
+    worker = threading.Thread(target=_produce, daemon=True);
```
MusabMahmoodh, line 606: One uncapped OS thread per open stream, on a public endpoint. A chat-style app with a few hundred concurrent SSE readers holds a few hundred threads, each alive as long as the client keeps reading. The reasoning for not using the shared request pool is right - a blocked producer must not starve requests - but a *dedicated bounded* executor gives that same isolation with a ceiling. Was a cap considered? Even a config'd `max_concurrent_streams` that 503s past the limit turns an unbounded resource into a declared one.

jac/jaclang/runtimelib/impl/server.impl.jac
```diff
@@ -449,6 +453,129 @@ async def _run_function_with_occ(
...
+        while True {
+            (kind, payload) = await q.get();
+            if kind == 'done' {
+                break;
+            }
+            if kind == 'error' {
+                raise payload;
+            }
+            capacity.release();
+            yield payload;
```
MusabMahmoodh, line 623: The context is pinned around `next(gen)` / `__anext__`, then the item crosses to the loop and is serialized by the SSE frame layer with no context pushed. For dicts (what the tests yield) that's fine, but a generator yielding nodes or archetypes gets serialized outside the request context - the same boot-superroot fallback this PR removes, one layer later. Can serialization ever touch the graph on this path? If it can, the pin belongs around the yield, not just around production.

jac/jaclang/runtimelib/serving/server.jac
```diff
@@ -514,10 +517,25 @@ obj ServingServer {
             await writer.drain();
             return conn_override != 'close';
         } except (ConnectionError, OSError) {
+            if chunked {
+                await self._close_stream(stream);
```
MusabMahmoodh: The close hangs off the `except` arm, so it only runs for `(ConnectionError, OSError)`. `CancelledError` is a `BaseException` and is the usual way a connection task dies on server drain or shutdown, so it skips this entirely: the producer thread then runs the generator to exhaustion for a client that's gone, and still commits its writes at the end. A `finally` covers both and is a no-op on the success path:

```jac
        } except (ConnectionError, OSError) {
            return False;
        } finally {
            if chunked { await self._close_stream(stream); }
        }
```

The disconnect test only exercises the RST path (`SO_LINGER` -> `ConnectionResetError`), so this gap isn't covered either.

jac/jaclang/runtimelib/serving/server.jac
```diff
@@ -514,10 +517,25 @@ obj ServingServer {
...
     }
 
+    async def _close_stream(stream: any) {
+        try {
+            if hasattr(stream, 'aclose') {
+                await stream.aclose();
+            } elif hasattr(stream, 'close') {
+                stream.close();
+            }
+        } except Exception {
```
MusabMahmoodh: Two things here. First, `except Exception { ; }` is a bare swallow in a new helper, and the two sibling closes added in this same PR (sse.jac, websocket.impl.jac) both log - so if `aclose()` raises `RuntimeError: asynchronous generator is already running`, the producer never sees `stopped`, the thread leaks, and it's completely silent. `logger.debug(f"stream close failed: {e}")` is enough.

Second, this PR introduces the same `hasattr(x,'aclose') / elif hasattr(x,'close')` block three times - here, `sse.jac:_sse_frame_stream_inner`'s new `finally`, and `_ws_stream_result`'s new `finally`. That's a third source of truth for "close whatever this stream is" being born in a single PR, and the three already differ in their logging. One shared `close_stream_like(obj)` used by all three would be strictly less code.

jac/tests/runtimelib/test_serve_disconnect.jac
```diff
@@ -13,15 +13,98 @@ subsequent requests as if nothing happened.
...
+                );
+                head = _recv_until(sock, b'"n": 0');
+                assert b'"n": 0' in head , f"no first SSE chunk: {head}";
+                sock.close();
+                deadline = time.monotonic() + 5.0;
+                while session._active_units > baseline_units
+                    and time.monotonic() < deadline {
+                    time.sleep(0.05);
+                }
+                assert session._active_units == baseline_units , "stream unit still held after client disconnect";
```
MusabMahmoodh, line 151: Does this test fail on main? Tracing it: pre-fix there's no producer thread and no stream unit, so `session._active_units` never leaves baseline and `/healthz` still answers - both assertions look like they pass on main. If it's here to pin the new thread's release semantics (which the last commit changed), a line in the docstring saying so would help, and the assertion that would actually bite is that the *producer stopped* - e.g. the generator does not keep advancing after the client is gone.

Two gaps worth naming: a write conflict at the end-of-stream commit surfacing as the framed `event: error` (promised in the fragment, untested), and more than one stream open at once - the only case where the per-stream thread model can misbehave.

MusabMahmoodh (commented):
Root cause and fix both check out. A generator returned from `run_attempt` hasn't executed a line when `_run_function_with_occ`'s `finally` resets the request ContextVar, so iteration later falls through `Jac.get_context()` to `JacRuntime.exec_ctx` and the boot superroot. Binding the generator to `ctx` for the stream's life is the right altitude - it fixes every consumer (`/function`, `/walker` via `report`, webhook, ws, cluster RPC) at the one place they funnel through, rather than patching each caller.

`count_direct` and `count_streamed` in the same fixture, asserting `{"count": 1}` present and `{"count": 0}` absent, pins the bug exactly. That's the assertion the issue deserved.

Can this break something already running: yes, under load, by resource model rather than by logic. Every concurrently-open SSE stream from a sync producer now pins a dedicated OS thread for its whole life, uncapped, where before it was iterated on the loop. Correct for one stream; the question is a few hundred (comment 1).

On production readiness: the correctness half I'd ship. What I'd settle first is the thread ceiling and the cancel path (comment 3), because both only show up under real traffic - a drain or shutdown cancels the connection task, which is `BaseException`, so it misses the close entirely and the producer keeps running for a client that's gone.

jac/jaclang/runtimelib/serving/server.jac
kashmithnisakya: moved the close into a `finally` so cancellation (drain/shutdown) closes the stream too, and added a test that cancels the connection task mid-stream and asserts the producer's generator actually gets closed.

jac/jaclang/runtimelib/serving/server.jac
kashmithnisakya: one `close_stream_like` in runtimelib/serving/streams.jac now serves all three sites (http body, sse source, ws stream) and warns on failure. also turned out `logger` was never bound in the runtimelib server module, so those warnings would have NameError'd; bound it.

jac/jaclang/runtimelib/impl/server.impl.jac
kashmithnisakya: added `[serve] max_concurrent_streams` (default 1024, 0 = unlimited): the manager takes a slot before handing out the wrapper and refuses past the cap with a typed 503 `stream_limit` before any bytes go out; a sync stream's slot is released when its worker thread exits, so the count is the thread ceiling. covered by test_serve_stream_limit.jac (side-by-side streams, refuse past cap, slot comes back).

jac/tests/runtimelib/test_serve_disconnect.jac
kashmithnisakya: you're right, it passed on main. the assertion is now that the producer's generator gets closed (its own finally runs) rather than that progress freezes, since blocking on the backpressure semaphore freezes too; checked it fails with the stop flag ignored. added side-by-side streams + cap tests and a producer-raises -> `event: error` test; a conflict at the end-of-stream commit isn't constructible in-process (WriteConflict isn't raised anywhere in the current runtime, conflicts surface as store serialization retries inside Session.commit), so that path is covered through the shared error-frame delivery.

jac/jaclang/runtimelib/impl/server.impl.jac
kashmithnisakya: checked, nothing on this path touches the graph: sse frames with `json.dumps(item, default=str)` and `str(archetype)` is the dataclass field repr (anchor field is repr=False), ws uses `send_json`; only `Serializer.serialize` reaches `Jac.get_context()` and it isn't used for streamed items. left as is; a pin around the yield wouldn't be safe anyway since the consumer can drive the asyncgen from another task and contextvar tokens can't be reset across contexts.

MusabMahmoodh (commented):
Re-review at head. All five from my 08-15 pass landed: the uncapped thread per stream is now `[serve] max_concurrent_streams` with a typed 503, the close moved into a `finally` with a cancel-the-task test behind it, the three copies of the close block collapsed into one `close_stream_like` (along with the `logger` that was never bound in that module), the disconnect test now asserts the producer's generator actually gets closed rather than something that passed on main, and the serialization question came back with a concrete answer. I accept the last one: `json.dumps(item, default=str)` on the SSE path, `send_json` on the WS path, and `Serializer.serialize` is the only thing that reaches `Jac.get_context()`.

Five things below, none of them blocking the correctness fix.

**1. The cap charges async producers for a thread they never take.** `jac/jaclang/runtimelib/impl/server.impl.jac`, the `_stream_pool.acquire(limit)` site in `_run_function_with_occ`.

You framed the slot as the thread ceiling ("a sync stream's slot is released when its worker thread exits"), but the slot is taken for both branches and `_stream_async_gen` runs on the loop with no thread at all. An all-async app starts getting `503 stream_limit` at 1024 open streams for a resource it never consumes. Move it into the sync branch?

```jac
if isgenerator(result) {
    limit = _stream_limit();
    slot = _stream_pool.acquire(limit);
    ...
}
```

**2. The async branch and the webhook fix have no test.** `_stream_async_gen`.

That is ~35 lines with its own commit-after-exhaustion, its own pin around `__anext__` and its own close path, and I cannot find a test anywhere that produces an async generator: all four new tests and the whole existing streaming suite use sync `def:pub` generators. Same for the webhook guard in `serve.endpoints.impl.jac` (`if (isgenerator(result) or isinstance(result, AsyncGenerator))`), which fixes a real crash while `test_webhook.jac` is untouched. Can both get a leg?

**3. `write_conflict` changes status on the monolith endpoints.** `_execution_result_response`.

This also turns the existing `write_conflict` result into a real 409 on `/function/*` and `/walker/*`, where today they answer 200 with the error dict inside `data`. Right status, and I am glad it is fixed, but it breaks anyone reading `data.error` on a conflict and it is a separate change from the streaming fix. Worth its own `.breaking` fragment rather than a clause in the bugfix note?

**4. The ceiling can only decay, and nothing says so.** `StreamPool.acquire`.

A producer blocked inside user code never reaches its `finally`, so its slot is gone for the life of the process. Agreed upstream that Python gives you no way to interrupt it, but it means the effective ceiling only ever drops, and the first symptom is 503s for everyone. Nothing logs on the refusal path. One `logger.warning` when `acquire` returns None, carrying `active` and the limit, makes that visible before someone has to guess.

**5. A bad `max_concurrent_streams` fails to the wrong side.** `_stream_limit`.

`except Exception { ; }` plus `max(0, limit)` means a typo'd value silently keeps 1024 and a negative one silently means unlimited. It matches `_occ_policy` right above it so I would take either, but this one removes a resource ceiling when it misreads. Worth a warning naming the key?

MusabMahmoodh (commented):
Re-reviewed at head. All five of my 08-15 points landed properly, and I withdraw one of my own: the `write_conflict` 200 to 409 change is doc conformance, not a break, since `persistence.md:82-83` and `config/index.md:232` already promise a typed 409.

Good catch on `logger` never being bound in `runtimelib/server.jac`.

On the new bound: `[serve] max_concurrent_streams` reads through the core loader (`project/config.jac:87`, `project/impl/config.impl.jac:1249`) rather than `get_scale_config()`, which is the right call for a runtimelib-level knob, and hitting it is loud - typed 503 `stream_limit` at `server.impl.jac:412-420` before any bytes, producer closed on the way out, asserted at `test_serve_stream_limit.jac:157-160`. No silent drop, so no theme-1 problem.

On RFC #8173 clause 4: this satisfies it in behaviour for both topologies, since the fix sits in `_run_function_with_occ` (`server.impl.jac:381`), the single funnel for all four entry points. It does not yet satisfy it as an *enforced* invariant - every new test lives in `jac/tests/runtimelib/` against the in-process server, and nothing under `jac/jaclang/scale/tests/` or the k8s e2e pins it. That is the difference between closing row 4 of the RFC table and closing the RFC, and I am happy to treat the second as follow-up.

Five things.

**1. `scale/websocket/impl/websocket.impl.jac:273`** - after this PR every generator result comes back from `_run_function_with_occ` already wrapped by `_stream_sync_gen`, which is an `async def`, so `isgenerator(result)` is False here for a sync `def:pub` producer and this whole arm, plus `_ws_stream_next` above it and `_WS_STREAM_STOP` in `websocket.jac:12`, looks unreachable - the `else` at `:282` handles it. I traced the only caller (`serve.endpoints.impl.jac:761` -> `ws_execute_target` -> `execution_manager.spawn_walker`) and think it can all go. But if I have the reachability wrong, then this is the one streaming arm that still runs the producer with no context pushed, which is the worse answer. Which is it?

**2. `runtimelib/impl/server.impl.jac:480`** - `max(0, limit)` plus `if limit > 0` in `acquire` means `max_concurrent_streams = -1` silently means unlimited. A typo'd key falling back to 1024 is fine; this edge fails open and removes the ceiling the key exists to set. Returning 1024 with a warning on negatives keeps it closed.

**3. `runtimelib/impl/server.impl.jac:409-411`** - you framed the slot as the thread ceiling, but it is taken for both branches, and `_stream_async_gen` runs on the loop holding no thread at all. So an all-async app starts getting `503 stream_limit` at 1024 open streams for a resource it never consumes. Either move the acquire into the sync branch, or widen the justification to "open streams" and drop the thread framing. Right now the code and the stated reason disagree.

**4. `runtimelib/impl/server.impl.jac:617`** - `_stream_async_gen` is about 35 lines with its own per-item pin, end-of-stream commit, close path and `slot.release()`, and I cannot find a test that produces an async generator. Same for the webhook guard at `serve.endpoints.impl.jac:512`: real crash fixed, `test_webhook.jac` untouched. Can both get a leg?

**5. `tests/runtimelib/test_serve_stream_limit.jac:128`** - this bound does not discriminate. `_post_stream` defaults to `count=5` and the producer sleeps 0.05s per item, so three streams take about 0.25s each; fully serialized they finish in ~0.75s and the assertion still passes. Raising the count at `:98` to `_post_stream(server.bound_port, tag, 20)` makes it fail when the streams are not actually concurrent.

MusabMahmoodh (commented):
Read the whole diff again against `6d01389ab`, plus the surrounding code in `server.impl.jac`, `sse.jac` and `streams.jac`. The mechanism is right and I would merge it. Two things I want to name as good before the asks: `close_stream_like` serving HTTP, SSE and WebSocket from one helper is single-source-of-truth in the sense this repo means it, and `weakref.finalize(stream, slot.release_unstarted)` closes a created-but-never-started slot leak that would only ever have surfaced as slow capacity loss in production.

**1. The tests encode the correctness half of #8136, not the isolation half.** This is my issue being under-specified, not your code.

`test_serve_streaming.jac`, `test_serve_stream_limit.jac` and `test_serve_disconnect.jac` all seed under `root`, stream `len([root-->][?:Item])`, and assert the count. One user throughout, always authenticated. That asserts "a streamed body sees the caller's graph". What the fix actually establishes is stronger and more important: a streamed body sees *only* the caller's graph, because pushing the caller's `ctx` is what puts the right principal in front of the ACL seam.

Those are different assertions and only the first is tested, so a future regression here comes back as a silent isolation failure with the suite still green. Two cases on top of what is there would close it:

1. two users, the second streams the first's node by jid, assert nothing comes back
2. no credential at all, same call, assert it is refused rather than served

I have a probe that drives both arms and I am happy to hand it over or open it as a follow-up PR against your branch - your call, it is your PR.

I am also going to correct #8136 itself, since the issue text is what shaped these tests: it says writes after a `yield` are never committed, and what I measure is that they are committed under the fallback root rather than lost. That is my error to fix, not yours, and I will do it on the issue.

**2. Question, not a change request: writes on a client disconnect.** In `_stream_sync_gen` the terminating commit is guarded by `if not stopped[0]`, and `stopped[0]` is set in the consumer's `finally`, which is where a disconnect lands. So a stream that produced twenty frames and had the client walk away commits nothing, while the same twenty frames to a client that stayed commit everything. Is that the intended contract? "A disconnected stream is an aborted request" is defensible and matches the `abort` in the conflict path. But an app author will meet it the hard way, and the release note only says disconnected streams close their producer promptly. If it is intended, one line in the note saying a disconnect discards the stream's writes saves someone a long afternoon. If it is not, it wants a test.

**3. Nit.** `except BaseException as e { _put(('error', e)); }` in the producer turns `SystemExit` and `KeyboardInterrupt` into a frame on the queue. Daemon thread running one generator, so the blast radius is small and I see why you cast wide, but `Exception` plus a re-raise for the two base cases is the usual shape and a Ctrl-C arriving as a stream error is confusing in a dev server. Take it or leave it.

**What I validated:** the diff and its surrounding code, by reading. I have reproduced the underlying defect on my own rig on two binaries with positive controls, but I have not yet built this branch and re-run that probe against it - a toolchain build did not fit in the disk I had last night. I will do that run and report the result here rather than leave it implied. `test-jac-pack-smoke` is currently red on the PR.

MusabMahmoodh (commented):
Re-read this against current `main` rather than against my last pass, because a
lot moved underneath it. Three things, in the order they matter.

## 1. Your diagnosis is still exactly right, seven days and five refactors later

I re-verified the mechanism on `main` at `d97a6d0`, since the file it lives in
has been rewritten repeatedly since you last pushed:

- `_run_function_with_occ` is still there, now at
  `jac/jaclang/server/impl/server.impl.jac:423`
- it still resolves the context via `_abegin_user_request_context` (`:432`) ->
  `_begin_request_context` (`:266`), which pushes the ContextVar at `:271`
- it still tears it down unconditionally in its `finally` -
  `Jac.reset_request_context(req_token)` at `:520`

So "the call only creates the generator object; not one line of its body has
run, and the ContextVar is gone by the time the serving layer iterates it" holds
verbatim on today's tree. That is worth saying out loud because the file churned
five times in five days and it would be reasonable to assume the analysis had
rotted. It has not.

## 2. The blocker: `jaclang/runtimelib/` no longer exists

#8682 (merged Aug 26) dissolved `jaclang/jac0core` and `jaclang/runtimelib` -
2,573 renames. About 298 of this PR's lines land in files that are gone:

| this PR | current `main` |
|---|---|
| `runtimelib/impl/server.impl.jac` (+247) | `jac/jaclang/server/impl/server.impl.jac` |
| `runtimelib/server.jac` (+24) | `jac/jaclang/server/server.jac` |
| `runtimelib/serving/server.jac` (+5) | `jac/jaclang/server/serving/server.jac` |
| `runtimelib/serving/sse.jac` (+3) | `jac/jaclang/server/serving/sse.jac` |
| `runtimelib/serving/streams.jac` (new) | `jac/jaclang/server/serving/streams.jac` |

Two warnings so this does not eat your afternoon twice:

- **The 247-line hunk is the hard one.** `server/impl/server.impl.jac` has taken
  five commits in five days - #8682, #8391, then #8711, #8742 and #8745 from
  Jason in the last two. Three of those touch the request/session lifecycle your
  change hooks into. Rebase that file by hand and re-read the surrounding code;
  a `git rebase` that reports clean here is not evidence.
- Your logger name `"jaclang.runtimelib.serving.streams"` in the new
  `streams.jac` is a string, so nothing will flag it. Rename it with the path.

You are one of fourteen open PRs carrying dead `runtimelib`/`jac0core` paths, so
this is not a comment on your PR specifically - it is the tax the refactor put
on everyone. Yours is the one I care most about clearing because it closes
#8136.

## 3. Design: `_stream_pool` is process-global, and that means two different things on our two deploy targets

```jac
glob _stream_pool: StreamPool = StreamPool();
```

`StreamPool` holds `active: int` behind a `threading.Lock`, as a module-level
singleton. Two things follow.

**a. The configured limit does not mean what it says in microservice mode.** In
single-process serving, one process is the whole server and the limit is the
server's limit. Under `--scale` each service is its own process and the fleet
runs N replicas, so the same config value yields an effective ceiling of
`limit x replicas`, and it moves when someone edits `replicas` for unrelated
reasons. An operator who sets "max 100 concurrent streams" to protect a
downstream gets 100 or 600 depending on topology, with no diagnostic either way.

This is the same shape as RFC #8173 ("one execution contract, two deployment
strategies") - a knob whose meaning silently changes with the deployment
strategy. I am not asking you to solve distributed limiting in this PR; that
would be a much larger change and it is not what #8136 is about. What I would
like is for the PR to *say* which it is: name the setting per-process
(`max_concurrent_streams_per_process`), or document in
`cli/docs/reference/config/index.md` that the bound is per process and
multiplies by replica count. Right now the docs hunk does not distinguish, and
whichever way you go the reader stops being surprised.

**b. Module-global mutable state is the #7900 theme-1 shape**, so it will draw
attention on review whether or not it is wrong here. It is defensible - the pool
is genuinely process-scoped and the `StreamSlot.released` flag makes double
release a no-op - but a one-line comment at the `glob` saying "per process by
design; see <doc>" costs nothing and answers the question before it is asked.

## 4. `close_stream_like` - make it a consolidation rather than an addition

New helper, and I checked: there is no existing generic one, so this is not a
duplicate. But there are ad-hoc versions of exactly this already on `main` -
`jac/jaclang/scale/runtime/transport/http_forward.jac:59` and `:70` both do a
bare `await resp.aclose()` with no guard and no logging.

If you route those two through `close_stream_like` in this PR, the helper stops
being a new thing and becomes the single place streams get closed, which is the
framing that survives review here. Small diff, and it makes the "Deleted" story
non-empty.

## Checked and not flagged

- `close_stream_like` preferring `aclose` over `close` and swallowing to a
  warning is right for a cleanup path - a failed close must not mask the
  original error.
- `StreamSlot` splitting `release` from `release_unstarted` reads like it exists
  because acquire-then-fail-before-start is a real path. Assuming that is what
  `test_serve_disconnect.jac` covers.
- The three new test files are real integration tests against a served app, not
  mocks. That is the right shape for this and it is most of why I am comfortable
  with a change this size in the request lifecycle.

Nothing here is a request to change the approach - the approach is right and it
closes the issue I filed. (2) is the gate. Ping me when it is rebased and I will
turn the re-review around same day; this one has waited long enough on my side
already.

PR #8188, opened 2026-08-14
fix(compiler): clear per-node ES caches when placement invalidates a module

### LINKED ISSUE #8184: Client pages importing a demoted module re-lower from stale node caches and ship runtime hooks without their imports
## Summary

When placement invalidates a module's ES output (demotion of a client-pulled module, pull-stamp changes), only the module-level caches (`module.gen.js`, `module.gen.es_ast`) are cleared. The per-node lowering caches (`nd.gen.es_ast` / `nd.gen.js`) survive. `EsastGenPass.enter_node` prunes any subtree whose `gen.es_ast` is already set, so the re-lower stitches the stale fragments back together and skips the lowering side effects that only fire on a real pass, in particular `_inject_runtime_import`.

The result: a client page importing a since-demoted element regenerates with `useJacState` / `useState` / `useEffect` references in its body but without their `@jac/runtime` import lines. The page throws `ReferenceError` at mount and dies behind the error boundary.

## Repro

Two markerless modules:

```jac
# demote_cache_util.jac
glob _seen: list = [];

def note(tag: str) {
    _seen.append(tag);
}

def:pub ping(payload: str) -> str {
    return payload;
}
```

```jac
# demote_cache_app.jac
import from demote_cache_util { note, ping }

def:pub app -> JsxElement {
    has count: int = 0;

    can with entry {
        note("boot");
        count = count + 1;
    }

    msg = ping("x");
    return <div>{count}{msg}</div>;
}
```

Compile the app with codegen, then demote the pulled helper (as `ClientBundleBuilder` does after an ES lowering failure) and re-lower the app:

```jac
prog = JacProgram();
app = prog.compile("demote_cache_app.jac", no_cgen=False);
util = <hub module for demote_cache_util.jac>;
demote_client_pulled_elements(util, prog, len(prog.errors_had), len(prog.warnings_had));
EsastGenPass(ir_in=app, prog=prog);
```

First lower of the app:

```js
import { note, ping } from "./demote_cache_util.js";
import { useJacState } from "@jac/runtime";
import { useEffect } from "@jac/runtime";
function App() {
  const __jacS_count = useJacState(0);
  ...
```

Regenerated app after demotion:

```js
import { __jacCallFunction } from "@jac/runtime";
async function ping() {
  return await __jacCallFunction("ping", {});
}
function App() {
  const __jacS_count = useJacState(0);   // useJacState is never imported
  useEffect(() => { ... }, []);          // useEffect is never imported
  ...
```

The `App` body is the stale first-run fragment. The import declarations are rebuilt per pass run from `_inject_runtime_import` calls, which never fire because `enter_node` pruned the cached subtree.

## Where

- Invalidation sites that clear only module-level caches: `jac/jaclang/jac0core/placement_solver.jac` (`_invalidate_stale_codegen` in `solve_module_placement`, and `demote_client_pulled_elements` for both the demoted module and each client importer).
- Prune-on-cached: `EsastGenPass.enter_node` in `jac/jaclang/compiler/passes/ecmascript/impl/esast_gen_pass.impl.jac` (`if nd.gen.es_ast { ... self.prune(); }`).

## Fix direction

Every invalidation site must clear the per-node ES caches together with the module-level ones, so the re-lower actually re-runs and re-registers its injected runtime imports.

kashmithnisakya:
Fixes #8184

## Problem

When placement invalidates a module's ES output, only the module-level caches (`module.gen.js`, `module.gen.es_ast`) were cleared. The per-node lowering fragments survived, and `EsastGenPass.enter_node` prunes any subtree whose `es_ast` is already cached. The re-lower therefore stitched the stale first-run fragments back together and skipped the lowering side effects that only fire on a real pass, in particular `_inject_runtime_import`. A client page importing a since-demoted element regenerated with `useJacState`/`useState`/`useEffect` references in its body but without their `@jac/runtime` import lines, and died at mount with a `ReferenceError` behind the error boundary (full repro with before/after JS in the issue).

## Fix

All three invalidation sites in the placement solver (`_invalidate_stale_codegen` in `solve_module_placement`, and `demote_client_pulled_elements` for both the demoted module and each of its client importers) now go through a single helper, `_invalidate_module_es`, which walks the module tree and wipes the node-level ES caches together with the module-level ones. The walk reads the raw `_gen` slot rather than the lazily-allocating `gen` getter, so nodes that never had a cache do not get one allocated just to be cleared.

Folding the module-level clears into the one helper also means a future invalidation site cannot repeat the original mistake of clearing only half the state.

## Test

`jac/tests/compiler/test_codespace_inference.jac` gains a regression test backed by two new codespace fixtures: compile a client page with `has` state that imports a pullable helper, demote the helper, re-lower the page, and assert the regenerated JS still imports the state hook it references. Without the fix the page keeps its stale body fragment, references `useJacState`, and never imports it; the test fails with the emitted JS in the assertion message.


jac/jaclang/jac0core/placement_solver.jac
```diff
@@ -496,8 +496,7 @@ def solve_module_placement(
 
     def _invalidate_stale_codegen {
         if stamp_count[0] and (module.gen.js or module.gen.es_ast is not None) {
```
MusabMahmoodh: The state this PR is about is "node fragments cached, module-level output empty" - and that is exactly what a lowering that died before `exit_module` leaves behind, which is the state `demote_client_pulled_elements` exists to handle. In it, this condition is false and the walk never runs, so the stamp path still re-uses stale fragments.

Was the guard kept deliberately as a cheap early-out, or is it left over from when the clear was two assignments? `_invalidate_module_es` already skips nodes with no cache, so dropping the parenthesised half costs one tree walk and closes the case:

```jac
        if stamp_count[0] {
            _invalidate_module_es(module);
        }
```

jac/tests/compiler/test_codespace_inference.jac
```diff
@@ -639,6 +639,38 @@ test "codespace stamps invalidate stale module codegen" {
     );
 }
 
+test "demotion re-lowers importer pages with their injected runtime imports" {
```
MusabMahmoodh, line 655: Both demotion sites are exercised, but `_invalidate_stale_codegen` - the one site that keeps a module-level guard, and so the one most likely to still be wrong - has no node-level coverage; the existing "codespace stamps invalidate stale module codegen" asserts on `module.gen.js` only. A second case that changes a pull stamp and then re-lowers, asserting the same `useJacState` import, would pin the site this PR did not fully change.

jac/tests/compiler/test_codespace_inference.jac
```diff
@@ -639,6 +639,38 @@ test "codespace stamps invalidate stale module codegen" {
...
+        util, prog, len(prog.errors_had), len(prog.warnings_had)
+    );
+    assert demoted , "pulled helper must demote";
+    assert not (app.gen.js or '') , (
+        "demoting a dep must invalidate the client importer's module JS"
+    );
+    EsastGenPass(ir_in=app, prog=prog);
+    js = app.gen.js or '';
+    assert 'useJacState(' in js , "re-lowered page must still lower its state";
+    hook_imported = any(
```
MusabMahmoodh: `ln.strip().startswith('import ') and 'useJacState' in ln` breaks the day the emitter wraps a long specifier list across lines, and it'll read as this bug coming back rather than as a formatting change. Matching the runtime import source is sturdier and still specific - assert the import section contains both `@jac/runtime` and `useJacState` rather than requiring them on one physical line.

jac/jaclang/jac0core/placement_solver.jac
```diff
@@ -905,8 +917,7 @@ def demote_client_pulled_elements(
                     if (
                         isinstance(resolved, str) and os.path.abspath(resolved) == key
                     ) {
-                        m.gen.js = '';
-                        m.gen.es_ast = None;
+                        _invalidate_module_es(m);
```
MusabMahmoodh: This invalidates modules whose own `body` has a CLIENT `Import` resolving to the demoted path - one hop. If A client-imports B and B client-imports the demoted C, A isn't invalidated. Not this PR's bug (A stays internally consistent, so no `ReferenceError`), but is A's lowering guaranteed placement-correct after C demotes? If it is, worth a line in the loop saying why one hop is enough.

MusabMahmoodh (commented):
Diagnosis verified against the code: `EsastGenPass.enter_node` prunes on `nd.gen.es_ast` (esast_gen_pass.impl.jac:7783) and the module-level pair is only written at the end of `exit_module` (:7237-7238), so clearing just `module.gen.js`/`es_ast` left every node fragment cached and the re-lower stitched them back without re-running `_inject_runtime_import`.

This is a real fix rather than a patch - one helper for all three sites means a fourth site can't repeat the half-clear - and the `kid` walk does cover impl/test annexes, since `JacAnnexPass._merge_annex` merges annex statements into `nd.body` and `add_kids_right` (annex_pass.impl.jac:45-46).

Can this break something already running: no. It only widens an invalidation that already ran, at compile time; a wider clear can cost a re-lower but can't produce stale output.

One thing to settle before merge (comment 1) - there is a leftover half of the old condition that keeps the fix from covering the case it most needs to.

jac/jaclang/jac0core/placement_solver.jac
kashmithnisakya: yep, leftover from when the clear was two assignments. dropped the parenthesised half in bd5d9f6f6 so the walk runs on any stamp change.

jac/tests/compiler/test_codespace_inference.jac
kashmithnisakya: added a stamp-path test in bd5d9f6f6: empties the module output, re-pulls the client import, and asserts the useJacState import survives the re-lower. it fails against the old guard.

jac/tests/compiler/test_codespace_inference.jac
kashmithnisakya: swapped the per-line check for a statement-level helper that matches useJacState and @jac/runtime inside the same import statement, so wrapping can't fake a regression.

jac/jaclang/jac0core/placement_solver.jac
kashmithnisakya: one hop per demotion event composes transitively: B's export surface toward A doesn't change when C demotes, and if B itself later fails and demotes, that event runs its own importer invalidation and catches A. can't put the rationale in the loop, the strip-comments lint bans it, so leaving it here. (the loop is now main's invalidate_importers_es after the merge.)

PR #8246, opened 2026-08-15
fix(byllm): structured-output robustness for finish_tool arguments and json_object-downgrading backends

### LINKED ISSUE #8114: byllm: response_format json_schema request rejected by backends that downgrade to json_object (must contain word 'json')
## Summary

A zero-tool, typed `by llm()` call always builds `response_format` with `type: "json_schema"`, and the prompt text byllm generates for it (task instructions + the schema-requirements hint) never contains the literal word "json" anywhere. This is fine against backends that honor `json_schema` as sent. It breaks against any backend that silently downgrades `json_schema` to the legacy `json_object` mode server-side, because OpenAI's own API (and providers that mirror its contract) rejects `json_object` requests whose messages don't contain the word "json" — a documented constraint byllm's generated prompt can never satisfy.

Observed in production through OpenRouter: model `openrouter/qwen/qwen3.7-flash` is routed to an `Alibaba` backend that performs exactly this downgrade, causing every affected call to fail with `litellm.BadRequestError`.

## Environment

- `jac` 0.34.13 and 0.34.14 (both reproduced)
- `litellm` 1.82.6
- Trigger: any `by llm()` function with **zero tools** and a **typed (non-`str`) return type** — this is `MTRuntime.get_output_schema()`'s only branch that builds a `response_format` at all; the moment ≥1 tool is present, byllm switches entirely to tool-calling (`finish_tool`) and `response_format` is `None`.
- Reproduced live via OpenRouter (`openrouter/qwen/qwen3.7-flash` → `provider_name: "Alibaba"`), and reproduced with **zero network calls** (see below) — the request byllm constructs is identical regardless of target model or provider; the downgrade and rejection happen entirely server-side.

## Steps to Reproduce

No live LLM call needed. This monkeypatches `litellm.completion` to capture byllm's outgoing request and abort before any network I/O, so it reproduces deterministically offline:

```jac
import litellm;
import from jaclang.byllm.lib { Model }

obj TurnReflection {
    has issue: str;
    has root_cause: str;
    has resolution: str;
}

glob captured: list = [];

def _fake_completion(*args: object, **kwargs: object) -> object {
    captured.append(kwargs);
    raise RuntimeError("captured-before-network-call");
}

glob llm = Model(model_name="openrouter/qwen/qwen3.7-flash");

def reflect_on_turn(turn_digest: str) -> TurnReflection by llm(max_tokens=200);

with entry {
    litellm.completion = _fake_completion;
    try {
        reflect_on_turn("Turn summary: wrote frontend.cl.jac, fixed a toggle bug, deleted the file.");
    } except Exception as e {
        print(f"aborted before network I/O: {e}");
    }

    params = captured[0];
    print(params.get("response_format"));
    messages_text = " ".join(str(m.get("content", "")) for m in params.get("messages", []));
    print(messages_text);
    print("contains 'json'?", "json" in messages_text.lower());
}
```

Run with `jac run repro.jac`. No API key, no network access, and no target-model-specific behavior required — any `model_name` produces the same captured request, since the downgrade this triggers happens server-side, not client-side.

## Expected Behavior

Either:
- byllm's generated prompt for a `json_schema` structured-output request should satisfy the legacy `json_object` constraint that some providers enforce even when `json_schema` was requested (e.g. by ensuring the word "json" appears somewhere in the messages), **or**
- byllm should detect and adaptively recover from this specific request-level rejection (it already has a retry-with-feedback loop for a related but distinct failure class — see below), rather than the call failing outright with no recovery path.

## Actual Behavior

The captured request:

```
response_format: {'type': 'json_schema', 'json_schema': {'name': 'TurnReflection', 'schema': {...}, 'strict': True}}

messages text: "This is a task you must complete by returning only the output. Do not
include explanations, code, or extra text—only the result. reflect_on_turn

turn_digest = (...)  Schema requirements:
- issue (string)
- root_cause (string)
- resolution (string)"

contains 'json'? False
```

Against a backend that downgrades `json_schema` → `json_object` (confirmed: OpenRouter routing to Alibaba), this request is rejected outright, before the model generates anything:

```
litellm.BadRequestError: OpenrouterException - {"error":{"message":"Provider returned error","code":400,"metadata":{"raw":"data: {\"error\":{\"code\":\"invalid_parameter_error\",\"param\":null,\"message\":\"'messages' must contain the word 'json' in some form, to use 'response_format' of type 'json_object'.\",\"type\":\"invalid_request_error\"},\"id\":\"chatcmpl-a7d8c5a9-03a7-94b1-9b47-967e9377760e\"}\n\n","provider_name":"Alibaba","is_byok":false}}
```

This is a **request-level provider rejection**, not a model output/parsing failure — it happens before the model generates a single token. byllm's existing typed-output retry loop (`_invoke_typed_retry` in `basellm.impl.jac`) only catches `OutputConversionError` (a response the model produced that failed to parse) and feeds back generic "try again" text. It has no path for this failure category, so the call simply fails.

## Evidence/Logs

**Live occurrence** (jaseci-labs/jacBuilder CI, `jac_coder` agent, `todo_app_sentinel` scenario, `openrouter/qwen/qwen3.7-flash`, one instance per stage, 5/5 stages, always immediately after a turn's last tool call):

```
2026-08-12 10:56:13.037 | ERROR | jaclang.byllm.llm:_raise_mapped_provider_error:160 - LLM call failed with BadRequestError:
litellm.BadRequestError: OpenrouterException - {"error":{"message":"Provider returned error","code":400,"metadata":{"raw":"data: {\"error\":{\"code\":\"invalid_parameter_error\",\"param\":null,\"message\":\"'messages' must contain the word 'json' in some form, to use 'response_format' of type 'json_object'.\",\"type\":\"invalid_request_error\"},\"id\":\"chatcmpl-a7d8c5a9-03a7-94b1-9b47-967e9377760e\"}\n\n","provider_name":"Alibaba","is_byok":false}},"user_id":"user_3FtDNdhdf75tGtzIbO6U0T9nzZK"}
```

Non-fatal only because the caller (`jac_coder`) happens to wrap this specific call in a broad `try/except`, best-effort by its own design — byllm itself has no recovery.

**Offline reproduction output** (no live model, from the script above, `jac` 0.34.13):

```
response_format sent to litellm:
{'type': 'json_schema', 'json_schema': {'name': 'TurnReflection', 'schema': {'title': 'Turnreflection', ...}, 'strict': True}}

full prompt text sent to the model:
    This is a task you must complete by returning only the output.
    Do not include explanations, code, or extra text—only the result.
     [{'type': 'text', 'text': "reflect_on_turn\n\nturn_digest = (...)"}, {'type': 'text', 'text': 'Schema requirements:\n- issue (string)\n- root_cause (string)\n- resolution (string)'}]

prompt contains the literal word 'json' (case-insensitive)? False
```

## Severity

Medium. Narrow, backend-specific blast radius — of 8 OpenRouter models spot-checked with an identical request shape, only the one routed to Alibaba rejected it; models routed to OpenAI, Anthropic, Meta, Mistral, DeepSeek, and Z-AI backends accepted the same `json_schema` request without complaint. But for any caller whose model happens to route to an affected backend, every zero-tool typed `by llm()` call fails outright with no recovery, and the caller has no way to know in advance which backends this affects — it depends on server-side routing/downgrade behavior byllm has no visibility into.

## Possible Cause

byllm already has the right shape of infrastructure for this — a typed-output retry loop (`_invoke_typed_retry`, `basellm.impl.jac`) that re-prompts with feedback and retries up to `max_attempts` — but it's scoped only to `OutputConversionError` (the model's response failed to parse). This failure is a different category: a `BadRequestError` raised before the model responds at all, because the *request itself* was rejected by the provider.

Proposed direction: extend the retry loop to also catch known provider request-rejection signatures (starting with this one — Alibaba's `json_object` word requirement) and adaptively repair the request before retrying, rather than either (a) unconditionally stuffing "json" into every prompt regardless of whether the target backend needs it, or (b) statically preferring tool-calling (`finish_tool`) as a blanket substitute — that path has its own distinct, currently open cross-provider failure mode (see jaseci-labs/jac#5442, Groq rejecting the `finish_tool` call envelope for simple return types). Neither mechanism is unconditionally safe; the fix likely belongs in adaptive, signature-driven request repair rather than a static preference for either mechanism.

## Related 

- jaseci-labs/jac#5442 — `finish_tool` fails on Groq for simple return types; relevant because it rules out "just switch this call to tool-calling" as a complete fix — that mechanism has its own provider-specific failure mode.
- jaseci-labs/jac#5837 — recent, merged fix injecting schema hints into prompts for local-model backends; same code area (`schema.impl.jac` / `attach_schema_hint`), different gap (enum member names, not this constraint).
- Also observed in the same investigation: `finish_tool()` itself raised `missing 1 required positional argument: 'final_output'` on a tool-calling structured-output call (`plan_build`) against the same model — a separate, independent failure in the tool-calling path, tracked separately.

### LINKED ISSUE #8169: byllm: finish_tool() crashes with raw TypeError when the model names its argument with wrong name
## Summary

When a `by llm()` call has ≥1 tool, byllm auto-appends a synthesized `finish_tool(final_output: object) -> object` for the model to call with its final answer. `Tool.parse_arguments()` builds that call's arguments by copying only the keys the model happened to name that exactly match a declared parameter — any other key is silently dropped, no error raised. If a model calls `finish_tool` with an argument named anything other than `final_output` (e.g. `output`, `result`), the parsed argument dict ends up empty. Unlike every other tool, `finish_tool`'s invocation isn't wrapped in a try/except, so the resulting `TypeError` propagates out of byllm as a raw, unhandled crash instead of a recoverable tool error.

## Environment/Context

- `jac` 0.34.13 and 0.34.14 (both reproduced)
- `litellm` 1.82.6
- Trigger: any `by llm()` call with ≥1 real tool (this is what causes byllm to synthesize `finish_tool` at all — see `MTRuntime.get_output_schema()` / `Tool.make_finish_tool`)
- Live occurrence: `openrouter/qwen/qwen3.7-flash` via OpenRouter, jaseci-labs/jacBuilder's `jac_coder` agent, `plan_build` tool (a `by llm()` call with tools)

## Steps to Reproduce

No live LLM call needed. This monkeypatches `litellm.completion` to return a fake-but-structurally-real `litellm.ModelResponse` whose `tool_calls` include a call to `finish_tool` with the argument named `output` instead of the tool's actual declared parameter `final_output`, mirroring the naming slip the live model made:

```jac
import litellm;
import from jaclang.byllm.lib { Model }

obj PlanResult {
    has plan: str;
}

def noop_tool(x: str) -> str {
    """Only here to make byllm take the tool-calling path, which is what
    auto-synthesizes finish_tool.""";
    return x;
}

def _fake_completion(*args: object, **kwargs: object) -> object {
    return litellm.ModelResponse(
        model="fake-model",
        choices=[{
            "index": 0,
            "finish_reason": "tool_calls",
            "message": {
                "role": "assistant",
                "content": None,
                "tool_calls": [{
                    "id": "call_fake_001",
                    "type": "function",
                    "function": {
                        "name": "finish_tool",
                        "arguments": '{"output": {"plan": "add input, add button, render list"}}',
                    },
                }],
            },
        }],
        usage={"prompt_tokens": 10, "completion_tokens": 10, "total_tokens": 20},
    );
}

glob llm = Model(model_name="openrouter/qwen/qwen3.7-flash");

def plan_build(task: str) -> PlanResult by llm(tools=[noop_tool], max_tokens=200);

with entry {
    litellm.completion = _fake_completion;
    try {
        result = plan_build("build a todo app");
        print(f"UNEXPECTED: call succeeded with result: {result}");
    } except Exception as e {
        print(f"call raised: {type(e).__name__}: {e}");
    }
}
```

Run with `jac run repro.jac`.

## Expected Behavior

A tool call whose arguments don't fully match the tool's declared parameters should fail gracefully and recoverably, the same way an unknown tool name or malformed JSON arguments already do in this codebase (`_error_tool_call(...)`, fed back to the model as a normal tool-error message it can react to on the next turn).

## Actual Behavior

The call raises a raw, unhandled Python `TypeError` that escapes byllm entirely:

```
call raised: TypeError: Tool.make_finish_tool.<locals>.finish_tool() missing 1 required positional argument: 'final_output'
```

Note the tool's own advertised JSON schema is correct : `final_output` is properly declared as the (only) required property:

```json
{"name": "finish_tool", "parameters": {"type": "object",
  "properties": {"final_output": {"type": "object", "properties": {"plan": {"type": "string"}}, "required": ["plan"]}},
  "required": ["final_output"], "additionalProperties": false}}
```

So this isn't a bad-schema problem — it's that byllm has no fallback when the model's tool call doesn't exactly match that schema.

## Evidence/Logs

**Live occurrence** (jaseci-labs/jacBuilder CI, `jac_coder` agent, `openrouter/qwen/qwen3.7-flash`, `plan_build` → `finish_tool`):

```
{"type": "llm_tool_result", "tool": "plan_build",
 "result": "Error: planning failed - Tool.make_finish_tool.<locals>.finish_tool() missing 1 required positional argument: 'final_output'",
 "call_id": "call_01aefe27d30041b2bc45a6f1", "iteration": 4}
{"type": "llm_thought", "content": "Skipping plan_build", "iteration": 5}
```

Image

The crash wasn't retried — the caller's next action was to abandon planning entirely for that turn ("Skipping plan_build"), losing the delegation rather than recovering from it.

Image

**Offline reproduction output** (no live model, from the script above, `jac` 0.34.13):

```
call raised: TypeError: Tool.make_finish_tool.<locals>.finish_tool() missing 1 required positional argument: 'final_output'
```

Identical error text to production.

## Severity

Medium. `finish_tool` is the universal mechanism byllm uses to end any tool-calling `by llm()` turn, so this can surface for any caller once a model deviates from the exact parameter name — model-dependent, not backend-dependent (unlike the sibling response_format issue below), so any less strictly-aligned model is at risk. Not fatal by itself (the raised exception is typically caught by the caller's own code, as it was here), but the caller only ever sees a confusing Python `TypeError` string instead of a recoverable tool-error message, and loses the entire delegation rather than getting a chance to retry.

## Possible Cause

Two compounding gaps in the same file family:

1. `Tool.parse_arguments()` (`types.impl/tool.impl.jac`) iterates over the arguments *the model provided* and keeps only the ones whose name matches a declared parameter, it never checks that all *required* parameters ended up present. A single misnamed argument silently empties the result instead of raising.
2. `Tool.__call__()`'s finish-tool branch is the one path not wrapped in try/except — every regular tool's exception is caught and turned into a `"Tool error: {e}"` string the model can react to; `finish_tool`'s isn't.

Fixing either one closes this: have `parse_arguments` validate required-parameter coverage and raise (routing through the existing `_error_tool_call(...)` recoverable-error path already used for unknown tools and malformed JSON), and/or wrap `finish_tool`'s invocation the same way every other tool's is wrapped.

## Related

- jaseci-labs/jac#8114 : the sibling `response_format`/"json word" structured-output gap, found in the same investigation. Different trigger (provider-side request rejection vs. model argument-naming), same underlying theme: byllm's structured-output paths (`response_format` and `finish_tool`) each have their own untested failure mode, and neither is a strictly safer default than the other.
- jaseci-labs/jac#5442 : `finish_tool` rejected outright by Groq for simple return types; same synthesized-tool mechanism, a different (provider-side) failure.

MalithaPrabhashana:
Two structured-output failure modes found in the same investigation, one per mechanism: `finish_tool` (tool-calling) and `response_format` (zero-tool). Fixes #8169. Fixes #8114.

## Problem 1: finish_tool crashes on a misnamed argument (#8169)

When a `by llm()` call has tools, byllm synthesizes `finish_tool(final_output)` to end the turn. If the model names that argument anything else (`output`, `result`, a flattened payload), `Tool.parse_arguments()` silently drops it and the invocation crashes with a raw `TypeError` that escapes byllm instead of the recoverable tool-error path every other tool failure gets:

```
TypeError: Tool.make_finish_tool.<locals>.finish_tool() missing 1 required positional argument: 'final_output'
```

### Fix

- `Tool.parse_arguments` (finish tool): resolve arguments by trial-validating candidates against the declared output type: `final_output` if present, then a sole misnamed value, then the whole dict as a flattened payload. The common misname case now succeeds outright with zero extra turns.
- `Tool.parse_arguments` (regular tools): missing required parameters (and non-dict argument payloads) raise a `ValueError` that routes through the existing `_error_tool_call` recoverable path instead of silently parsing to `{}`. The `return` annotation is no longer treated as a parameter.
- `Tool.__call__`: the finish branch wraps failures in `FinishToolError` (never a "Tool error" string, which would corrupt the typed result) and both react loops catch it, feed the error back, and let the model retry.
- `_error_tool_call`: error-report tools are marked so `is_finish_tool()` no longer classifies a finish-call error report as the finish call itself; previously its coaching text could be returned as the typed final output.
- Tool-protocol prompt now names the `final_output` argument explicitly.

## Problem 2: response_format requests rejected by json_object-downgrading backends (#8114)

A zero-tool typed `by llm()` call sends `response_format: json_schema`, and the generated prompt never contains the word "json". Backends that silently downgrade `json_schema` to legacy `json_object` (observed: OpenRouter routing to Alibaba) then enforce OpenAI's documented `json_object` constraint - the messages must contain the word "json" - and reject every such call before the model generates a token:

```
litellm.BadRequestError: ... "'messages' must contain the word 'json' in some form,
to use 'response_format' of type 'json_object'."
```

This is a request-level rejection, so the typed-output retry loop (which only catches `OutputConversionError`) never sees it.

### Fix

The schema hint injected into every `response_format` request now opens with "Respond with a single JSON object." The word appears exactly when a JSON response is requested and nowhere else (plain `str` and tool-calling requests carry no `response_format`, hence no hint). On a downgrading backend the sentence also becomes the model's only format instruction, since the downgrade strips the schema; an off-shape reply then lands in `OutputConversionError`, which the existing typed-retry loop already handles. A test pins the invariant: any request carrying a `response_format` contains "json" in its messages; requests without one do not.

## Verification

- byllm suites: 238 passing locally (the 3 `test_mtir_integration` failures reproduce identically on clean `main` with these changes stashed - environmental, pre-existing).
- New table-driven unit tests plus end-to-end tests through `dispatch_no_streaming`, `adispatch_no_streaming` (async), `_force_final_answer`, `_attempt_recovery`, and the streaming react loop.
- Live (`claude-haiku-4-5`, `claude-sonnet-4-6`): a litellm proxy rewrites the model's real `finish_tool` call into each #8169 failure shape - pre-fix all three crash with the exact `TypeError` above; post-fix misname/flatten heal with no extra turns and unusable arguments recover in one retry turn. For #8114, typed calls across nested-object, object-with-list, scalar-int, and list-of-str shapes all return typed results with the json word present, and plain-`str` calls confirm no hint leaks.


jac/jaclang/byllm/types.impl/tool.impl.jac
```diff
@@ -65,25 +71,105 @@ impl Tool.make_finish_tool(resp_type: type) -> Tool {
...
+        );
+    }
     return args;
 }
+
+impl Tool._parse_finish_arguments(args_json: dict, params: dict) -> dict {
+    resp_type = params.get("final_output", object);
+
+    candidates: list = [];
+    if "final_output" in args_json {
```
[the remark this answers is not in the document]
MalithaPrabhashana: Valid catch. Reproduced with an output type whose own field is named final_output sent flattened: the dict validated against the type yet was rejected into the retry path. Fixed in 252b3004b by making the whole-dict candidate an unconditional fallback rather than part of the elif chain, with a regression row added to the resolution table test.

jac/jaclang/byllm/impl/schema.impl.jac
```diff
@@ -498,7 +498,9 @@ impl schema_to_hint_text(rf: dict | None) -> str {
     if not lines {
         return "";
     }
-    return "Schema requirements:\n" + "\n".join(lines);
+    return (
+        "Respond with a single JSON object.\nSchema requirements:\n" + "\n".join(lines)
```
kugesan1105: The "json" invariant still has a reachable hole: when the schema produces no hint lines, this returns `""` while `response_format` is still sent. `_hints_from_schema` emits nothing for properties without a top-level `"type"` or `"enum"`, and union fields compile to `anyOf` with no `"type"`. So `obj Result { has value: str | int; }` (also a fieldless obj) still sends a request with no "json" in the messages, the exact #8114 rejection.

Suggest decoupling the sentence from the lines: return "Respond with a single JSON object." even when `lines` is empty (or prepend it in `attach_schema_hint` whenever `rf` is a real response_format).

jac/jaclang/byllm/tests/test_byllm.jac
```diff
@@ -1012,6 +1012,210 @@ test "finish tool error not raised for valid type" {
...
+    }
+    assert recovery is not None , "Recovery should run for a structured resp_type";
+    finish_tc = recovery.tool_calls[0];
+    assert finish_tc.is_finish_call() , "Recovered call should be the finish call";
+    assert finish_tc.get_output().name == "Gus" , "Expected healed typed output";
+}
+
+# Some backends downgrade json_schema to json_object and reject requests whose
+# messages lack the word "json" (issue #8114).
+test "response format requests always carry the json word" {
```
kugesan1105, line 1198: Please add a union-field output type (e.g. `has value: str | int;`) and a fieldless obj to this table. Both currently violate the invariant this test pins, which is why the hole in `schema_to_hint_text` slipped through.

jac/jaclang/byllm/impl/schema.impl.jac
MalithaPrabhashana: Confirmed and fixed in 8756cec1a. Reproduced both shapes through real request construction: `has value: str | int` compiles to `{"anyOf": [...], "title": "Value"}` with no top-level "type", and a fieldless obj to `"properties": {}` - both produce zero hint lines, so the whole hint including the lead sentence was dropped while `response_format` was still sent. Exactly the #8114 rejection condition, thanks for catching it.

Took your first suggestion: the lead sentence now rides on the request mode rather than on the schema contents. `schema_to_hint_text` returns the sentence as soon as `rf` is a real JSON response_format (json_schema or json_object), then appends `Schema requirements` only if lines materialize. That closes all three early-return paths that skipped it (missing schema, non-dict properties, empty lines), not just the two you named. A non-JSON `rf` type still returns "" as before.

jac/jaclang/byllm/tests/test_byllm.jac
MalithaPrabhashana: Added both rows in 8756cec1a: `union field` (`has value: str | int`) and `fieldless obj`, alongside two new module-level fixtures.

Verified they actually guard rather than just pass: with the source fix stashed the table fails with `AssertionError: union field: json-word=False but rf=True`, and passes with it restored. Full byllm directory green otherwise (the 3 `test_mtir_integration` failures reproduce identically on clean main with all changes stashed).

Issue #6115, opened 2026-05-22
jac-scale --scale deploys silently no-op: hyphenated bool Args register a phantom `name-with-hyphen`: True namespace key (PR #6024 regression)


udithishanka:
## Problem

Any boolean CLI `Arg` whose name contains a hyphen (e.g. `dry-run`, `show-yaml`, `enable-tls`) ends up with a **phantom namespace key defaulting to `True`** that `HookContext.get_arg("the-name", False)` returns even when the flag was never passed on the command line. This causes pre-hooks that check `context.get_arg("dry-run", False)` to see `True` for every invocation.

Hit in production on `jaseci-labs/jacBuilder@dev`: every `jac start main.jac --scale --experimental` since `6cff699d2` (PR #6024) silently takes the dry-run branch in `jac-scale/jac_scale/plugin.jac:147`, prints the plan, sets `cancel_execution=True`, and returns before `apply_manifests` ever runs. The deploy reports success because the workflow follows it with a hard-coded `echo "Deployment spec applied"`. The cluster has been frozen on a 43h-old image for ~16h before anyone noticed.

## Root cause

[`jac/jaclang/cli/impl/registry.impl.jac:289-304`](https://github.com/jaseci-labs/jaseci/blob/main/jac/jaclang/cli/impl/registry.impl.jac#L289-L304) adds an auto-generated `--no-X` companion for every bool flag:

```jac
} elif arg.kind == ArgKind.FLAG or arg.typ == bool {
    kwargs["action"] = "store_true";
    kwargs["default"] = arg.default if arg.default is not None else False;
    if arg.short {
        parser.add_argument(f"-{arg.short}", f"--{arg.name}", **kwargs);
    } else {
        parser.add_argument(f"--{arg.name}", **kwargs);
    }
    # Add --no-X variant
    no_kwargs = {
        "action": "store_false",
        "dest": arg.name,                         # <-- preserves hyphens
        "help": f"Disable {arg.name}"
    };
    parser.add_argument(f"--no-{arg.name}", **no_kwargs);
}
```

argparse handles this name pair inconsistently:

- `--dry-run` (`store_true`): no explicit `dest`, argparse auto-derives → namespace attribute `dry_run` (underscore), default `False`.
- `--no-dry-run` (`store_false`): explicit `dest="dry-run"` (hyphen), → namespace attribute `dry-run` (literal hyphen), default `True` (the inverse default of `store_false`).

[...]

PR #8298, opened 2026-08-17
fix(cli): tri-state bool flags so jac install --no-npm is honored

### LINKED ISSUE #7464: jac install --no-npm is documented but not honored for full installs
## Summary

`jac install --help` documents `--no-npm: Disable npm`, but the flag has no effect on a full `jac install`: the provider loop in `cli/commands/impl/project.impl.jac` installs every non-python ecosystem unconditionally, and the `npm` parameter is only consulted for the `--npm <named-package>` add-mode. On a client-kind project, `jac install --no-npm` still runs the full bun closure.

## Why it cannot be fixed by just checking the flag

The install signature has `npm: bool = False`, so a plain `jac install` and `jac install --no-npm` are indistinguishable (both `False`). Making it tri-state (`default=None`) does not survive the CLI framework: `registry.impl.jac` registers bool args with `kwargs["default"] = arg.default if arg.default is not None else False`, explicitly coercing `None` to `False`, so `--no-X` forms can never express "explicitly disabled" vs "unspecified" for any bool arg.

## Repro

```
jac create app --use web-app && cd app
jac install --no-npm     # bun installs the full client closure anyway
```

## Options

1. Teach the arg framework tri-state for bool args (preserve `default=None`; `--X` -> True, `--no-X` -> False, unspecified -> None) and gate the npm provider on `npm is False`. This makes every documented `--no-X` flag honest, but changes defaults framework-wide (consumers using truthiness are unaffected; `is False` comparisons need auditing).
2. Or drop the misleading auto-generated `--no-npm` from the help until it does something.

Context: found while making scale pods skip the npm closure (#7462); the working alternative there is the existing `JAC_CLIENT_SKIP_NPM_INSTALL` env knob, which #7462 extends to gate the npm provider's `install_all`.

kugesan1105:
Closes #7464

## Problem

`jac install --help` documents `--no-npm: Disable npm`, but on a client-kind project `jac install --no-npm` still ran the full bun closure. The flag could not work even in principle: the shared arg parser registered every bool flag as a `store_true`/`store_false` pair and coerced `None` defaults to `False`, so an explicit `--no-X` was indistinguishable from the flag not being given. The same defect was latent in the `config_key` overlay, which detects "user did not pass this flag" by comparing against the default: an explicit `--no-X` on a default-False flag could be silently re-enabled by `jac.toml`.

## Fix

**Parser** (`cli/impl/manifest_parse.impl.jac`): bool flags now use `argparse.BooleanOptionalAction` with the declared default preserved.

| Arg declares | unspecified | `--X` | `--no-X` |
|---|---|---|---|
| `default=False` (77 args) | `False` | `True` | `False` (unchanged) |
| `default=True` (4 args) | `True` | `True` | `False` (unchanged) |
| `default=None` (opt-in tri-state) | `None` | `True` | `False` |

All existing bool args declare explicit defaults and the repo has no `is False` identity checks on parsed args, so nothing changes behavior until an arg opts in with `default=None`. Flags already named `no-*` keep a plain `store_true` and get no negated twin, which also drops the nonsensical `--no-no-save` / `--no-no-deps` options from help and completions.

**Config overlay** (`cli/impl/dispatch.impl.jac`): compares against the true declared default instead of coercing `None` to `False` for bools.

**npm** (`cli/commands/project.jac`, `impl/project.impl.jac`): the `npm` arg opts into tri-state and the full-install provider loop skips the npm provider when `npm is False`, printing `Skipping npm dependencies (--no-npm)`. `--npm` add-mode and plain `jac install` are unchanged. The `JAC_CLIENT_SKIP_NPM_INSTALL` env knob stays as the deeper gate for non-CLI callers (scale pods).

`manifest.impl.jac` regenerated via `jac gen-cli-manifest` (`--verify` passes).

## Help output change

Negations now render as paired entries (`--npm, --no-npm`) on one line instead of separate `--no-X  Disable X` rows, and `--no-no-*` entries are gone.

## Tests

- New `tests/runtimelib/test_cli_arg_tristate_bool.jac`: tri-state parse, unchanged two-state behavior, and the `no-*` no-twin rule.
- New case in `tests/project/test_dependencies.jac`: `install(npm=False)` must not invoke the npm provider (suite 83/83).
- Regression suites green: hyphenated-bool (#6115), manifest parity, config-key bridge, run dispatch, help color, verb surface, command isolation.
- Validated end to end on a real `web-app` project: `jac install --no-npm` installs Python deps and never invokes bun; plain `jac install` still runs the full closure; `jac install --npm left-pad` and `jac remove --npm` unaffected.


MusabMahmoodh (approved):
Approving. This closes #7464 properly rather than at the symptom, which I appreciate: the issue's own point was that you cannot fix it by checking the flag, because `default=False` makes a plain `jac install` and `jac install --no-npm` indistinguishable. Teaching the arg layer tri-state is the fix at the right altitude.

I checked the regression I was most worried about, which is that dropping the `None` to `False` coercion in `_apply_config_defaults` (`cli/impl/dispatch.impl.jac:163`) changes behaviour for every other bool arg. It does not: across `cli/commands/*.jac` there are 85 `Arg.create` blocks with `ArgKind.FLAG` or `typ=bool` and **zero** without an explicit `default=`, and in `cli/impl/manifest.impl.jac` there are 86 bool `ArgMeta` entries with exactly one lacking a default, which is `npm` itself. So the only arg whose resolved default moves is the one that is meant to. Adding the explicit `default=False` to the five `nacompile` flags in the same PR is what makes that true, and it is the part that would have been easy to skip.

`test_cli_arg_tristate_bool.jac` is the right test for this: a real parser, real `add_argument`, three assertions that pin unspecified/`--X`/`--no-X`, plus the two-state case so the change is fenced in both directions. No mocks in the part that matters.

Also worth calling out because it is easy to miss in the diff: `tools.impl.jac:181` stops emitting `--no-<name>` for args already named `no-*`, which is what was producing `--no-no-client` in the generated CLI docs. That is a separate real bug fixed in passing.

Two non-blocking notes.

**`cli/impl/manifest.impl.jac`** - this PR had to make the same edit twice, once in `commands/*.jac` and once in the `ArgMeta` list here, and the two have to agree for the parser and the docs to agree. That is a hand-synced pair, and this change is exactly the shape that drifts them: a future arg declared tri-state in one place and two-state in the other would parse one way and document another, silently. If `_command_routes` is generated from the command modules, ignore this. If it is maintained by hand, a test asserting the two declaration sets have identical `(name, typ, default)` triples would make the drift impossible rather than merely unlikely.

**`tests/project/test_dependencies.jac`, the new `--no-npm` case** - it patches `NpmProvider.install_all` and asserts `assert_not_called()`, which matches the rest of that file, so I am not asking you to change it. But the thing a user actually reports is "bun still ran", and the test that would catch a regression in the provider loop rather than in the call is a real `jac install --no-npm` subprocess in a temp project with an npm dependency, asserting no `node_modules` appears. The parser test already covers the flag semantics; this would cover the wiring.

PR #8303, opened 2026-08-17
fix(scale): roll pods when only configuration changed

### LINKED ISSUE #8115: [Bug]: a config-only deploy never reaches running pods
## Summary

A deploy that changes only configuration never reaches running pods. The Secret object is updated, but the pod template comes out byte identical, so the patch is a no-op, no rollout happens, and the app keeps serving the previous values. The deploy reports success.

Split out of jaseci-labs/jac#8089 (item 4) as agreed on jaseci-labs/jacBuilder#1232. This is orthogonal to rollback and affects users today.

## Environment

- jac 0.36.1 (pod binary), deployed with `jac start --scale main.jac -t kubernetes`
- kind, Kubernetes v1.36.1, single node
- No jacBuilder involved, so this is reproducible on the plain jac-scale path

## Steps to reproduce

1. Minimal monolith app whose walker reports an env var:

```jac
import os;

walker:pub ping {
    can run with Root entry {
        report {"marker": os.environ.get("MARKER", "unset")};
    }
}
```

2. `jac.toml` with the value declared as a scale secret:

```toml
[scale.secrets]
MARKER = "v1"
```

3. `jac start --scale main.jac -t kubernetes`. Confirm the app reports `marker: v1`.
4. Change **only** `MARKER` to `v2`. Touch no source file.
5. Redeploy with the same command.

## Observed

| | before | after |
|---|---|---|
| Secret object | `v1` | `v2`, updated correctly |
| Deployment revision | 2 | 2, unchanged |
| Pod name and UID | `bgtest-deployment-76785c59cb-kd8zt` / `23478b7d...` | identical |
| Container restarts | 0 | 0 |
| Value the app serves | `v1` | `v1`, stale |

The deploy printed `Fleet ready: 2/2 deployments serving`.

## Expected

A configuration change should reach running pods, or the deploy should say that it did not.

## Root cause

Three things combine so that a config change is invisible to the pod template:

1. Secret sections are stripped from the bundle before it is packed, so config never affects the content digest. `injector/bundle.jac:111` `sanitized_jac_toml`, using `_is_secret_path` at `injector/bundle.jac:98-103`, which matches `scale.secrets` and `plugins.scale.secrets`.
2. The Secret is referenced by a constant name. `deploy/target/kubernetes/manifest_builder.jac:645` `_user_secret_name` returns `{app}-secrets`, consumed as `envFrom.secretRef` at `manifest_builder.jac:1128-1129`.
3. There is no config hash or checksum annotation on the pod template. A grep for `restartedAt` and for checksum annotations across `jaclang/scale` returns nothing.

The digest is unchanged and the secret name is constant, so the generated pod template is byte identical to the running one. `patch_namespaced_deployment` (`deploy/target/kubernetes/target.jac:937`) is then a no-op, no new ReplicaSet is created, and `envFrom` is only read at container start, so running pods keep the old values.

Note that Kubernetes does not restart pods when a referenced Secret changes, so nothing else recovers this.

## Scope

The same shape applies to consumers that pass `secrets` in rather than declaring them in `jac.toml`. jacBuilder sets `target.secrets` from the project `.env` and strips `.env` from the staged bundle for the same reason, so it inherits the identical behaviour. Reproduced here on the plain jac-scale path to keep it independent of any consumer.

## Suggested fix

Add a config hash annotation to the pod template, derived from the resolved secret contents, so that a config change produces a template change and therefore a rollout. This also makes config participate in revision history, which matters for the rollback work in jaseci-labs/jac#8089.

## Related

- jaseci-labs/jac#8089 (this was item 4 there)
- jaseci-labs/jacBuilder#1232

Thamirawaran:
Fixes #8115.

## Problem

A deploy that changes only configuration never reaches running pods. The
Secret object updates, but the pod template comes out byte identical, so the
patch is a no-op, no rollout happens, and the app keeps serving the previous
values. The deploy reports success.

Three things combine:

1. Secret sections are stripped from the bundle before it is packed
   (`injector/bundle.jac:111`), so config never affects the content digest.
2. The Secret is referenced by a constant name, `{app}-secrets`
   (`manifest_builder.jac:645`).
3. There is no config marker on the pod template.

Kubernetes does not restart pods when a referenced Secret changes, so nothing
else recovers this.

## Fix

The pod template carries `jac-scale.config-revision`, holding the Secret's own
`resourceVersion`, stamped when the Secret is applied. A config change is then
a template change, so the patch is real and the fleet rolls. The Secret is
already applied before the Deployments, so the new pods read the new values.

The marker is deliberately **not** a digest of the resolved values.
`OPS_READ_RULES` grants `get/list/watch` on deployments and replicasets while
withholding secrets, so a hash of the values would put a brute-forceable
fingerprint of them on objects that boundary deliberately exposes, and
superseded ReplicaSets would retain fingerprints of previous values.
`resourceVersion` discloses nothing and Kubernetes advances it exactly when
the object changes, so an unchanged config still yields an unchanged template.

Stamping happens in the apply path rather than the manifest builder, because
that is where the value exists. That also means an overlay can neither drop
nor forge the marker: the stamp is applied after the `deployment_overlay`
merge, so no re-assertion step is needed.

An app with no secrets has no Secret and so no revision, and gets no
annotation at all. It is unaffected and takes no extra rollout.

## Verification

Reproduced and fixed on a real cluster (kind, and a jacBuilder PR preview
running this build): a walker reporting `os.environ.get("MARKER")` returned
`v1` after a config-only redeploy before the change, and `v2` after it.

Tests are in the existing files rather than new ones:

- `test_pod_env.jac`: a changed config changes the pod template; an unchanged
  one does not; an app with no secrets gets no marker; the marker carries no
  secret material.
- `test_deployment_overlay.jac`: an overlay that sets the marker to a forged
  value does not win.

Each assertion was checked by removing the behaviour it guards and confirming
it fails.

## Review notes

Thanks for the review, all three findings were correct. On the suggested keyed
HMAC: it was considered and dropped, because the key has nowhere to live that
is not itself exposed through workload metadata. `resourceVersion` needs no key.


jac/jaclang/scale/deploy/target/kubernetes/utils/kubernetes_utils.impl.jac
```diff
@@ -790,18 +790,21 @@ impl create_k8s_secret(
...
+    applied: any = None;
     try {
-        core_v1.create_namespaced_secret(namespace=namespace, body=secret_body);
+        applied = core_v1.create_namespaced_secret(
+            namespace=namespace, body=secret_body
+        );
     } except ApiException as e {
         if e.status == 409 {
-            core_v1.replace_namespaced_secret(
+            applied = core_v1.replace_namespaced_secret(
```
MusabMahmoodh, line 799: This branch is the one the bug actually travels - a redeploy 409s and the revision comes from replace - but `_FakeSecrets` only implements create, so it has no coverage. If the `applied =` capture here regresses, tests stay green and every real redeploy dies on the new RuntimeError in target.jac. A second fake that raises 409 on create would close it:

```jac
obj _FakeSecretsExisting(_FakeSecrets) {
    def create_namespaced_secret(namespace: str, body: any) -> any {
        raise ApiException(status=409);
    }
    def replace_namespaced_secret(name: str, namespace: str, body: any) -> any {
        return SimpleNamespace(metadata=SimpleNamespace(resource_version=self.version));
    }
}
```

jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac
```diff
@@ -130,6 +131,24 @@ def merge_manifest_overlay(base: any, overlay: any) -> any {
 }
 
 
+def with_config_revision(manifest: dict[str, any], revision: str) -> dict[str, any] {
```
MusabMahmoodh, line 134: `merge_manifest_overlay` ten lines up already does exactly this walk - non-mutating, recursive map merge - so this can collapse to:

```jac
def with_config_revision(manifest: dict[str, any], revision: str) -> dict[str, any] {
    if not revision {
        return manifest;
    }
    return merge_manifest_overlay(
        manifest,
        {"spec": {"template": {"metadata": {"annotations": {CONFIG_REVISION_ANNOTATION: str(revision)}}}}}
    );
}
```

Same semantics for every case the new tests pin down, including the forged-overlay one.

release_notes/unreleased/jaclang/8303.bugfix.md
```diff
@@ -0,0 +1 @@
+- **Scale: a config-only deploy now reaches running pods**: secret sections are stripped from the bundle before it is packed and the Secret is referenced by a constant name, so a deploy that changed only configuration produced a byte identical pod template. The patch was a no-op, no rollout happened, and the app kept serving the previous values while the deploy reported success. The pod template now carries a `jac-scale.config-revision` annotation holding the Secret's `resourceVersion`, stamped when the Secret is applied, so a config change is a template change and the fleet rolls. The marker is opaque: it carries no secret material, which matters because the ops read role grants access to deployments and replicasets while withholding secrets. Apps with no secrets have no Secret and so get no annotation and no rollout.
```
MusabMahmoodh: Worth one more sentence here: an existing app *with* secrets takes a one-time rolling restart on its first deploy after upgrading, because the annotation appears on a template that never had it. It's surge-rolled and it re-asserts current config, so it's benign - but an operator watching a fleet roll after a "no-op" deploy should be able to find that sentence.

MusabMahmoodh (approved):
The fix is right and in the right place: stamping at apply time after the overlay merge means an overlay can neither drop nor forge the marker, and the resourceVersion-over-digest reasoning holds up. Approving with a few comments.

One thing that isn't inline anywhere: the end-to-end claim (config-only redeploy -> pods roll -> app serves the new value) is verified by hand here but not by automation. `sdk_deploy_real_e2e.sh` already deploys todo_app with secrets and asserts they reach the pod via kubectl exec, so a redeploy leg on that script (bump the secret, deploy again, poll the walker until it reports the new value) is basically the #8115 repro and would pin this end to end. Fine as a follow-up if you'd rather not grow this PR.

PR #8304, opened 2026-08-17
feat(scale): expose revision based rollback and a bundle retention contract

### LINKED ISSUE #8089: [Feature]: expose revision based rollback in jac-scale (history limit, revision API, bundle retention)
## Feature Description

Consumers of jac-scale (jacBuilder, see jaseci-labs/jacBuilder#1232) need a rollback that does not rebuild. Today a rollback means checking out an old commit and redeploying it, which takes minutes and fails outright when the old commit no longer builds.

The useful observation is that a jac-scale deploy already leaves a complete rollback target in the cluster, we just do not expose or protect it:

- the container image is constant across versions (`resolved_image = k8s_config.python_image`), app code ships as a content addressed bundle
- the only per version difference in the pod template is the bundle digest inside the `jac-pvc-bootstrap` init container args
- deploy is `patch_namespaced_deployment`, not delete and recreate, so Kubernetes retains previous ReplicaSets

So a previous version is already fully specified in the cluster and could be restored by reverting the pod template, with no rebuild and no second set of pods. Four things are missing to make that a supported operation.

### 1. `revision_history_limit` in `KubernetesConfig`

Never set today, so the rollback depth silently rides the Kubernetes default of 10. If rollback depth becomes a product promise it needs to be a declared value that a consumer can read back and reason about. Natural home is beside `min_replicas` / `max_replicas` / `idle_replicas`.

### 2. A revision API on the deploy target

`list_revisions(app)` returning revision number, bundle digest and created timestamp, plus `rollback_to(app, revision)`.

Without this a consumer writes raw Kubernetes calls against objects jac-scale owns, which is jaseci-labs/jac#7730 all over again.

### 3. Bundle retention contract

jac-scale does no bundle garbage collection at all today, so consumers implement it themselves and hardcode jac-scale internals to do it (the `jac-pvc-bootstrap` container name, the PVC layout, the digest extraction).

The rule that matters here, never delete a bundle referenced by a retained revision, cannot be enforced from outside, because the consumer cannot see `revision_history_limit`. jac-scale owns the bundle format so it should own this guarantee.

### 4. Config changes do not currently produce a pod template change

This one looks like a live bug rather than a feature gap, and it may be worth splitting out.

`.env` keys ship as a Secret referenced by a constant name (`_user_secret_name` returns `{app}-secrets`), and there is no checksum or config hash annotation on the pod template. A grep for `restartedAt` and checksum annotations across `jaclang/scale` finds nothing.

That means a config only redeploy produces a byte identical pod template, the patch is a no op, no rollout happens, and the new values never reach running pods. Worth confirming on a real cluster before treating it as fact, I have traced this from the source but not reproduced it.

A config hash annotation on the pod template fixes it, and has the side benefit of making config participate in revisions.

## Examples & References

Source references, all on `main` at the time of writing:

| what | where |
|---|---|
| image constant across versions | `deploy/target/kubernetes/manifest_builder.jac:1089` |
| bundle digest into init container args | `injector/pvc_injector.jac:191-203` |
| deploy patches, history preserved | `deploy/target/kubernetes/target.jac:937` |
| secret referenced by constant name | `deploy/target/kubernetes/manifest_builder.jac:645` |
| rollout strategy already surge based | `deploy/target/kubernetes/manifest_builder.jac:750-752` |
| startup budget of 15 minutes | `deploy/target/kubernetes/manifest_builder.jac:819-826` |

Consumer side, for context: jacBuilder implements bundle pruning in `deployment/deploy.impl/retention.impl.jac` and builds its keep set from Deployments only, so bundles referenced by older ReplicaSets are already prunable.

Related: jaseci-labs/jac#7730, jaseci-labs/jacBuilder#1232

Note on scope: this is deliberately not blue green. A template revert waits for the app to start, so it is not instant. `startupProbe` currently budgets 900s, and that number has not been measured, so whether anything warmer is needed on top of this is a separate question.

Thamirawaran:
Fixes #8089, except item 4, which shipped separately as #8303.

A jac-scale deploy already leaves a complete rollback target in the cluster: the container image is constant across versions, app code ships as a content addressed bundle, the only per-version difference in the pod template is the bundle digest in the `jac-pvc-bootstrap` init args, and deploy patches rather than recreates, so Kubernetes retains previous ReplicaSets. What was missing was a way to see it, protect it, and use it.

## 1. `revision_history_limit` in `KubernetesConfig`

Never set, so rollback depth silently rode the Kubernetes default of 10 and no consumer could read it back. A consumer cannot offer a rollback it has no way to know is still retained.

It is now a declared field that round-trips through `to_dict` and `from_config` and lands on every generated Deployment. The default stays 10, the value the API already applied, so no existing deployment loses history.

It is also declared in the `[scale.kubernetes]` schema (`config/plugin_config.jac:198`). Without that, `_walk_schema` flags it as an unknown key on every command and hard-fails under `JAC_SCALE_CONFIG_STRICT`, while `from_config` reads it anyway, so the warning would have been wrong on top of being noisy. The description names the consequence nothing validates: `0` keeps no history and makes a rollback structurally impossible.

## 2. Revision API, scoped to the app and coherent across its fleet

`list_revisions(app)` reports the app's releases. A release is identified by the bundle digest baked into every service's bootstrap args, not by a Kubernetes revision number: a service whose pod template did not change takes no new revision, so a fleet drifts apart and "revision 2" names different releases for different services. Each release therefore carries the per service revision that shipped it.

`rollback_to(app, revision)` resolves a release and reverts every service to whichever revision carried it. It refuses when any service has no retained revision for that release, rather than reverting part of the fleet and reporting success, which would leave the app serving two versions at once. A patch that fails partway puts the already reverted services back.

**The fleet is read from the cluster, not from a route table.** An earlier revision of this PR derived the app's Deployments from its declared routes. That is wrong through `ScaleClient`: `_lifecycle_target` builds a bare `DeploySpec`, `spec.to_microservices_config()` always returns a dict with a `routes` key, so `_ms_cfg` short-circuits and never falls back to jac.toml. A microservice app resolved to `["{app}-deployment", "gateway-deployment"]`, so `rollback()` always refused and `revisions()` reported the gateway alone. `FLEET_SELECTOR` is the selector `destroy` already uses (`managed=jac-scale,jac-scale.role in (microservice,gateway)`), and `_fleet_deployment_names` lists against it. `app_deployment_names`, `_app_deployment_names` and `replica_sets_at_revision` are gone with the assumption.

The decisions stay pure functions covered natively: `releases_from_revisions`, `fleet_targets_for_release`, `fleet_revert_order`, and `strip_pod_template_hash`, which drops the controller stamped label that would otherwise pin a reverted template to the old ReplicaSet.

This is not a claim that two apps may share a namespace. Every app generates the same fixed-name `gateway-deployment`, and `destroy` already deletes by `managed=jac-scale` across the whole namespace, so a shared namespace is not a configuration jac-scale supports today, with or without this change.

## 3. Rollback restores code, not configuration

Worth stating plainly, because "revision based rollback" reads like it covers both. Since #8303, a config-only redeploy stamps `jac-scale.config-revision` on the pod template and so produces a real new revision, but it keeps the **same bundle digest**. A release here is its digest, so that revision is grouped with the release it shares the digest with rather than listed separately: it is not selectable in `list_revisions`, and rolling back to that release restores the newest revision carrying the digest. A reverted pod reads whatever the `[scale.secrets]` Secret currently holds.

Configuration is therefore deliberately out of scope rather than partly supported, and the release note says so.

The undo capture takes the live **Deployment** template, which does not carry `pod-template-hash`, so the compensating restore puts back the current config marker intact rather than a stripped one. Without that, a failed rollback would roll the already-restored services once for nothing and the next deploy would roll them again when it restamps.

## 4. Bundle retention contract

jac-scale owns the bundle format and the pod template that mounts it, but did no garbage collection, so consumers built their own keep-set by hardcoding the bootstrap container name and the mount layout (see jaseci-labs/jacBuilder's `retention.impl.jac`).

`bundle_keys_from_pod_spec` reads the keys back out of the init container the injector itself builds, and the container name now has a single definition both sides share, so the writer and the reader cannot drift. `referenced_bundle_keys(app)` is the contract a pruner can honor: never delete a bundle referenced by a retained revision.

It takes `strict=True` on that path. An unreadable cluster returning an empty keep-set is indistinguishable from "nothing is referenced", and for a pruner that means deleting every bundle it was meant to retain. Failing loudly is the only safe answer for a caller whose next step is a delete.

`ScaleClient.bundle_keys` exposes it beside `revisions` and `rollback`, so a consumer reaches all three without constructing a `KubernetesTarget` itself. That is the one the contract exists for, and it was the only one of the three not on the client.

## Tests

In existing files, no new ones. Each assertion was checked by removing the behaviour it guards and confirming it fails, including a fleet missing a member refusing to revert, a release restoring the newest revision of each service whatever order the API lists ReplicaSets in, a patch failing halfway putting the already reverted services back, and `rollback_to` driven end to end over a microservice fleet with the gateway reverting last.

## Review notes

On the suggested app identifier label: considered and dropped, because a label on the pod template changes the template and would roll every existing fleet once. Reading the fleet from `destroy`'s own selector gives the same scoping with no migration and no new label.

Known and left as is, non-blocking: the release-level int still reads like a `kubectl rollout history` revision number to anyone who knows that command, and they are not the same numbering. Worth a line in the CLI help when one exists; the digest is in the `list_revisions` output for callers that want an unambiguous key.


MusabMahmoodh (changes requested):
Read the whole diff against `bee017452b`. The core reasoning holds up: keying a release on the bundle digest rather than a Kubernetes revision number is the right call for a fleet that drifts apart, `strip_pod_template_hash` is a real trap avoided rather than a defensive flourish, and refusing a partial revert instead of half reverting is the right failure mode for this feature. The pure-function split makes the decisions readable.

Findings below, heaviest first. Late review, apologies.

## 1. `revision_history_limit` is not in the config schema, so setting it warns on every command

`KubernetesConfig.from_config` reads `revision_history_limit`, but the `[scale.kubernetes]` block in `config/plugin_config.jac:143` does not declare it. `_walk_schema` (`jac/jaclang/project/impl/config_schema.impl.jac:20`) flags every undeclared key, so a user who writes

```toml
[scale.kubernetes]
revision_history_limit = 3
```

gets `jac.toml [scale.kubernetes] has unknown key 'revision_history_limit'. Unknown keys are ignored, so defaults apply instead.` printed on every command, and a hard `ValueError` when `JAC_SCALE_CONFIG_STRICT` is set. The warning text is also wrong in this case: `from_config` does read the value, so defaults do not apply.

That lands on the PR's own headline claim, that rollback depth becomes something a consumer can set and read back. The fix is one block next to `max_replicas` at `config/plugin_config.jac:193`. Worth spending the `description` on the fact that `0` disables history entirely and makes `rollback_to` structurally impossible, since nothing validates the field.

The existing guard, "every default config key is declared in the schema" (`tests/misc/test_config_validation.jac:107`), walks the default config dict rather than `KubernetesConfig`'s fields, which is why CI is green on this.

## 2. The rollback API takes an int whose meaning differs from `kubectl rollout history`

The premise is right: a Kubernetes revision number names different releases for different services. But the release-level number this introduces has the same problem one level up. `list_revisions` reports `max(revision)` across the release's services while each service reverts to its own, so a release holding `{orders-app-deployment: 4, gateway-deployment: 2}` is "revision 4", and `rollback_to(app, 2)` does not name it. An operator who reads a number off `kubectl rollout history deploy/gateway-deployment` and passes it either gets a refusal or silently resolves a different release, with nothing in the result saying which release was chosen.

The identity you actually built the feature on is the bundle digest, and it is already in the `list_revisions` output. Accepting the digest, keeping the int as a convenience, would remove the overload.

Related: `fleet_targets_for_release` returns the first release whose max revision matches. If two releases ever share a max, the winner is list order.

## 3. `deployment_rollout_complete` rewrites the deploy-time readiness gate, not just rollback's

`target.jac:565` is the deploy path's fleet wait. Replacing `ready_replicas >= 1` with the full check is defensible and matches what `kubectl rollout status` does, but it changes every deploy's wait semantics from inside a rollback feature, and it is a change an operator feels before they ever call `rollback_to`.

akindu has #7949 open on exactly this gate ("gate fleet readiness on the applied revision, add an opt-in full-rollout wait"). Worth sequencing with them rather than resolving the same question twice in two branches.

Separately, `except Exception { return False }` means a status shape the client did not populate reads as "never complete", so the failure surfaces as a 1200s timeout with no reason attached to it.

## 4. `rollback_to` itself is not tested

The pure helpers are covered well and `_restore_templates` has both its paths. What has no test is the function that ties them together: the resolve-then-patch loop, the ordering of the `undo` capture against the patch, and the synthesized `wait_bundle`.

That last one is where I would most want a test, because it reconstructs the pod selector as `name[:-len("-deployment")]` instead of reading the live Deployment's own `spec.selector`. It is correct today, since `_build_deployment_manifest` names deployments `{k8s_name}-deployment` and selects on `{"app": k8s_name}`, but if that ever drifts the wait quietly degrades to a broad pod scan and nothing fails.

## 5. Smaller

- `list_revisions(app_name)` and `referenced_bundle_keys(app_name)` never read `app_name`; scoping comes from `self.k8s_config.app_name`. A caller passing a different app gets this target's app back with no complaint. Either assert on it or drop the parameter.
- `_revision_entries` calls `bundle_keys_from_pod_spec(dict(pod_spec))` twice per ReplicaSet, once for `bundle_keys` and again inside `app_bundle_digest`.
- `_restore_templates` returns without waiting for the compensating rollout, so a `success=False` result can come back while the fleet is still mid-roll. Fine if intended, worth a clause in the message so an operator does not read "put back" as "settled".

Only item 1 needs to change before this lands. Items 2 and 3 are worth a reply on direction, since this PR sets a contract and both are cheaper to settle now than after consumers bind to it.

jac/jaclang/scale/deploy/target/kubernetes/kubernetes_config.jac
```diff
@@ -46,6 +46,7 @@ obj KubernetesConfig(BaseConfig) {
         liveness_failure_threshold: int = 80,
         min_replicas: int = 1,
         max_replicas: int = 3,
+        revision_history_limit: int = 10,
```
MusabMahmoodh, line 49: Still the one blocker: this field is read by `from_config` but never declared in the `[scale.kubernetes]` schema (`config/plugin_config.jac:143`, the block ends around `max_replicas` at :193). `_walk_schema` flags every undeclared key, so a user who writes

```toml
[scale.kubernetes]
revision_history_limit = 3
```

gets `unknown key 'revision_history_limit'. Unknown keys are ignored, so defaults apply instead.` on every command, plus a hard `ValueError` under `JAC_SCALE_CONFIG_STRICT` - and the warning is wrong on top of it, because `from_config` does read the value.

That lands squarely on the PR's headline claim that rollback depth becomes something a consumer can set and read back. One block next to `max_replicas` fixes it. Worth spending the `description` on `0` disabling history entirely and making `rollback_to` structurally impossible, since nothing validates the field.

(CI is green because `test_config_validation.jac:107` walks the default config dict rather than `KubernetesConfig`'s fields.)

jac/jaclang/scale/deploy/target/kubernetes/target.jac
```diff
@@ -36,8 +36,147 @@ import from jaclang.scale.deploy.autoscale.autoscaler {
...
+        text = str(key);
+        if text.startswith("bundles/") and text.endswith(".jab") {
+            return text.rsplit("/", 1)[-1][:-len(".jab")];
+        }
+    }
+    return "";
+}
+
+
+def releases_from_revisions(entries: list[dict[str, any]]) -> list[dict[str, any]] {
```
MusabMahmoodh, line 80: New since I last read this: #8303 merged at 07:51 and stamps `jac-scale.config-revision` onto the pod template from the Secret's resourceVersion, so config-only redeploys now produce a real template change.

That cuts across the release identity here. A config-only rollout creates a new revision carrying the *same* bundle digest, so this collapses it into the existing release and keeps `max(revision)`. Two visible consequences: a config-only rollout never appears in `list_revisions`, and `rollback_to` on that release restores the newest revision holding the digest, which is the post-config-change one. So config changes are silently unrollbackable through this API rather than explicitly out of scope.

Given the commit that says rollback keeps to code and leaves configuration current, is the intent that the release key stays digest-only and the doc says config is not versioned here? If so this is worth stating in the release note, because "revision based rollback" reads like it covers both.

jac/jaclang/scale/deploy/target/kubernetes/target.jac
```diff
@@ -1173,6 +1312,253 @@ obj KubernetesTarget(KubernetesTargetBase) {
...
+        }
+
+        reverted: list[str] = [];
+        undo: dict[str, any] = {};
+        for name in expected {
+            if name not in targets {
+                continue;
+            }
+            rs = targets[name];
+            template = strip_pod_template_hash(
```
MusabMahmoodh: Second half of the #8303 interaction, and this one looks like a real bug rather than a scoping question. The revert replaces `/spec/template` wholesale, so the old `jac-scale.config-revision` annotation comes back with it, while `{app}-secrets` is a constant name still holding current values. The pods come up reading current config (fine, and matches the stated intent) carrying a marker that claims an older config generation - and the next deploy sees that marker differ from the live Secret's resourceVersion and rolls the whole fleet again for nothing.

This is the same shape as `strip_pod_template_hash` right here, which you already got right for the controller's own stamped label: a field the restored template should not carry back. Stripping the config marker (or re-stamping it from the live Secret) alongside the pod-template-hash is the matching move, and it keeps "rollback is code only" true in the template as well as in effect.

jac/jaclang/scale/deploy/target/kubernetes/target.jac
```diff
@@ -426,7 +565,7 @@ obj KubernetesTarget(KubernetesTargetBase) {
                 for name in names {
                     try {
                         d = apps_v1.read_namespaced_deployment(name=name, namespace=ns);
-                        if (d.status.ready_replicas or 0) >= 1 {
+                        if deployment_rollout_complete(d) {
```
MusabMahmoodh: This is the deploy path's fleet gate, not rollback's, and #7949 rewrites this exact line into `read_rollout_status(d).satisfies(mode)`. Textual conflict aside, the two disagree on what the default should be: #7949 makes the full-rollout bar opt-in behind `--wait full` and defaults to one available new-revision replica, precisely because blocking every deploy on a complete rollout lands on callers that shell out under their own timeout. `deployment_rollout_complete` makes that bar unconditional with no flag.

`RolloutStatus.is_complete()` over there is also this function, near enough - the only real difference is `available >= updated` here versus `available >= desired` there. If #7949 lands first, rollback gets what it wants by asking for `full` explicitly and the deploy path keeps its default. Worth sequencing with akindu rather than settling it twice in two branches.

Separately: `except Exception { return False }` means a status shape the client did not populate reads as "never complete", so it surfaces as a 1200s timeout with no reason attached.

jac/jaclang/scale/deploy/target/kubernetes/target.jac
```diff
@@ -1173,6 +1312,253 @@ obj KubernetesTarget(KubernetesTargetBase) {
...
+                    ]
+                );
+            } except Exception {
+                stranded.append(name);
+            }
+        }
+        return stranded;
+    }
+
+    def rollback_to(app_name: str, revision: int) -> DeploymentResult {
```
MusabMahmoodh, line 1465: The pure helpers are covered well and `_restore_templates` has both paths, but this function - the resolve-then-patch loop, the ordering of the `undo` capture against the patch, the synthesized `wait_bundle` - has no test.

The `wait_bundle` is where I'd most want one, because it rebuilds the pod selector as `name[:-len("-deployment")]` instead of reading the live Deployment's own `spec.selector`. Correct today, since `_build_deployment_manifest` names deployments `{k8s_name}-deployment` and selects on `{"app": k8s_name}`, but if that ever drifts the wait quietly degrades to a broad pod scan and nothing fails.

Also still open from last time: `list_revisions(app_name)` and `referenced_bundle_keys(app_name)` never read `app_name` (scoping comes from `self.k8s_config.app_name`), so passing a different app returns this one's revisions with no complaint - assert on it or drop the parameter. And the release-level int still names different releases than `kubectl rollout history` does for any single service; the digest is already in the output and would remove the overload.

MusabMahmoodh (commented):
Re-read at `72961b93`. Nothing has moved since my last pass, so this is the same set anchored to lines rather than left in one block, sorry about that. Two of them are new though, and both come from things that landed in the last hour rather than from the diff: #8303 merged at 07:51 and now puts a config marker in the pod template, and #7949 (approved this morning) rewrites the same gate line this PR touches.

Still think the core reasoning is right: keying a release on the bundle digest rather than a Kubernetes revision number is the correct call for a fleet that drifts apart, `strip_pod_template_hash` is a real trap avoided rather than a defensive flourish, and refusing a partial revert is the right failure mode.

MusabMahmoodh (changes requested):
Re-reviewed at `6affb7e41c`. Thanks for turning these around so fast - of my five earlier points, 1 and 4 are fully resolved (`revision_history_limit` is declared at `plugin_config.jac:208`, and `require_full=False` keeps the deploy path off the full-rollout gate), and 2 is answered as scoped. The compensating restore in `_restore_templates`, with tests driving both its success and its own failure, is a real failure mode handled properly rather than a defensive flourish.

Still needs changes, mainly for one thing I missed on the first pass.

**1. `sdk/client.jac:184` / `deploy/target/kubernetes/target.jac:1334` - the revision API can't see a microservice app**

`_lifecycle_target` builds a bare `DeploySpec(app_name, namespace, target)`, and `_make_target` passes `microservices_config=spec.to_microservices_config()`. That never returns `None`: `sdk/spec.jac:135-136` does `ms['routes'] = dict(ms.get('routes', {}))`, so it always hands back a dict with a `routes` key. `_ms_cfg` at `target.jac:197` then short-circuits on `is not None` and never reaches the `get_scale_config()` fallback that would read jac.toml.

So through `ScaleClient`, a microservice app resolves to `["{app}-deployment", "gateway-deployment"]`. `rollback()` always refuses on a missing `{app}-deployment`, and `revisions()` reports the gateway's releases only.

`destroy` already asks the cluster instead of the local config, at `target.jac:1654`:

```jac
selector: str = "managed=jac-scale,jac-scale.role in (microservice,gateway)";
```

Listing Deployments with that selector and sorting the gateway last gives the same set with no jac.toml dependency, and lets `app_deployment_names`, `_app_deployment_names` and their two tests go.

**2. `config/plugin_config.jac:198` - these two keys aren't `[scale.kubernetes]` inputs**

`monitoring_enabled` and `service_dependencies` are overwritten unconditionally after loading, from `[scale.monitoring]` and from the system plugin deps (`config/impl/config_loader.impl.jac:218` and `:225`). Declaring them here means `[scale.kubernetes] monitoring_enabled = true` now validates clean and is still discarded, which is the same lie this PR is fixing, pointing the other way.

They only surfaced because the new test counts loader-derived keys. I'd drop both declarations and name them in the test instead:

```jac
    derived: set = {"monitoring_enabled", "service_dependencies"};
    missing = sorted(consumed - declared - derived);
```

**3. `deploy/target/kubernetes/target.jac:133` - dead on arrival**

`replica_sets_at_revision` has no caller outside tests. `rollback_to` resolves its ReplicaSets inline at `:1490-1504` and `_revision_entries` keys off the annotation directly. It and the three assertions on it in `test_superseded_deployment_reap.jac` belong in the commit that replaced it.

**4. `deploy/target/kubernetes/target.jac:1533` - the undo capture is stripped too**

The undo template goes through `restorable_pod_template`, so it comes back without the *current* `jac-scale.config-revision`. That makes the compensating patch a real template change on services that were already put back: a failed rollback rolls them once for nothing, and the next deploy rolls them again when it restamps the marker. The undo capture wants the template as it is:

```jac
                undo[name] = dict(
                    serializer.sanitize_for_serialization(live.spec.template)
                );
```

**5. `sdk/client.jac:114` - `referenced_bundle_keys` is the one the contract exists for**

It's the only one of the three not exposed on `ScaleClient`, so jacBuilder's pruner still has to build a `KubernetesTarget` to reach it:

```jac
    def bundle_keys(
        app_name: str, namespace: str, target: str = 'auto'
    ) -> set[str] {
        return self._lifecycle_target(
            app_name, namespace, target
        ).referenced_bundle_keys(app_name);
    }
```

One earlier point still open: `rollback_to` has no test driving it end to end, and findings 1 and 4 both live inside it. That's the smallest addition that would make me comfortable shipping this.

MusabMahmoodh (approved):
All five are answered. Re-read the three new commits against the head:

1. The fleet is resolved from the cluster now - `FLEET_SELECTOR` is jac-scale's own destroy selector and `_fleet_deployment_names` lists against it, so `revisions`/`rollback` no longer depend on a `routes` table the SDK path never populates. `app_deployment_names`, `replica_sets_at_revision` and `_app_deployment_names` are gone with it, and "the fleet is read from the cluster, not from a route table" pins it. Two helpers deleted while fixing the bug is the right trade and I would not have insisted on it - good call taking it that far.
2. `monitoring_enabled` and `service_dependencies` are out of the `[scale.kubernetes]` schema.
3. Settled, and I had it half wrong: the forward path cleans the ReplicaSet template (which does carry `pod-template-hash`) while the undo captures the live *Deployment* template, which does not - so the current marker is restored intact rather than stripped. "a compensating restore puts back the marker the live template carried" is the test for exactly that.
4. Already resolved last round.
5. The two gaps I named are covered - "a rollback reverts the whole microservice fleet, gateway last" for the resolve-then-patch loop and "a compensating restore ..." for the undo capture. `referenced_bundle_keys(strict=True)` is a good catch of your own: an unreadable cluster returning an empty keep-set would have deleted every bundle it was meant to retain.

CI is green across all groups.

Non-blocking, unchanged from last round and fine to leave: the release-level int still reads like a `kubectl rollout history` revision number to anyone who knows that command, and they are not the same numbering. Worth a line in the CLI help at some point, not in this PR.

Issue #8323, opened 2026-08-18
Execution-first: the production binary runs its sealed prefix now, and every seal extends it -- waves to the zero-bytecode invariant


marsninja:
Execution program for #8288 / #8139, reordered **execution-first**: instead of sealing everything and flipping at the end, the production binary starts EXECUTING its sealed prefix now, and every subsequent seal extends the executed set continuously. Supersedes the wave ordering implied in the #8288 tracker comment; milestones and acceptance are unchanged at the endpoint. Checkboxes here are the living tracker, ticked with evidence per landing.

**Current state (main @ c376cd7a6, 2026-08-19):** sixteen sealed roots bind, verify and register; the parser is the only native execution; THE FLIP MACHINERY IS LIVE IN PRODUCTION (every compile routes through the prefix predicate; all modules decline `prefix-entry-absent` pending the A3 crossing); the link bug is fixed at root and the base chain re-landed (#8341); the A3 lowering gate (BASE_CHAIN_ARGUED) is emptied and retired on the crossing branch; the resolution-determinism law + E5029 blast-radius guard protect the sealed-binary context (#8360). The frame-census trend reads ~87.8M (baseline re-measured pre-crossing); completion is 0 with the rim deleted.

**Honest ceiling of the first flip:** the schedule's contiguous sealed prefix is seven passes (ASTValidation, SymTabBuild, DeclImplMatch, SemanticAnalysis, SemDefMatch, CFGBuild; +JsxIntrinsicGuard under no_cgen) plus the parser. The other nine sealed roots sit beyond the unsealed wall and join as it falls -- incrementally, each seal extending production execution rather than waiting for a final event.

---

## Language decisions (maintainer, 2026-08-18)

1. **Type-checking is IN the zero-bytecode scope**; the evaluator's four CPython-`ast` paths get redesigned away, timing allowed post-invariant.
2. **End state is hard error, no fallback** -- when M4 deletes the bytecode twins, a module the fused path cannot compile is refused outright (#8139's letter). Exotic casualties get fixed forward.
3. **Bare container annotations become a hard error repo-wide** (`list`/`dict`/`set` without element types), E1122-playbook migration included. The untyped-membership residual resolves through this: untyped operands in native-lowered membership are a checker error, no dynamic-dispatch tier.
[...]

PR #8175, opened 2026-08-14
fix(serve): one asset URL law — shared root table, `public/` at the web root, every server consumes it

Developer-Linus:
Closes #8125 (child 1 of RFC #8124).

## Summary

Static-asset resolution had drifted four different ways — the monolith
(`serve_static_file` + `serve_root_asset`), the `--scale` gateway
(`serve_static`/`serve_project_static`/`serve_extra_static`), the embedded
gateway (`_serve_static`), and the admin console each hand-rolled their own
"URL → file → response" logic. The same asset URL returned different bytes (or
a different 404) depending on the run mode, `public/` was not served at the web
root at all (the amplify report, #6876), and the monolith's `/*` handler
exposed arbitrary project files (`/package.json`, `src/…`) at the web root.

This PR extracts **resolution** into one shared, dependency-free core that
every run mode consumes, so a given asset URL returns the same bytes (or the
same honest 404) in vite dev, monolith `jac serve`, local `jac start`, and
`--scale`. It closes the source-exposure hole in the same move.

This is a **seam extraction, not a patch**: the per-host resolution logic is
removed and every host is routed through one function. Each host keeps its own
*dispatch/precedence* (which legitimately differs) but shares one *resolution*
law.

## The one law — `jac/jaclang/server/serving/static_assets.jac` (new)

A pure, dependency-free core (this PR added it under `runtimelib/`; an
unrelated in-flight restructure on `main`, #9048, later moved the
`server`/`runtimelib` boundary and this module landed under `server/serving/`
as a result — the monolith and scale still both consume it):

- `asset_root_table(project_root, client_dist, mounts)` → an ordered list of
  `AssetRoot`s, first hit wins:
  1. configured `static_mounts` (exclusive-ownership semantics preserved)
  2. `/static/*` → client dist, then project `assets/` (the #6872 URL space)
  3. `/*` → client dist
  4. `/*` → **`public/`** (the DX fix — new)
  5. `/*` → `assets/` (compat; the monolith already served this)
- `resolve_asset(table, url_path)` → `(Path, AssetRoot) | None`. Exact-path
  only (no basename flattening), jailed on every root via `Path.resolve()` +
  `is_relative_to()`, no extension allowlist, returns `None` on miss (never
  falls back to SPA/index — that stays the host's job). The web-root (`""`)
  roots deliberately do **not** serve `/static/*` URLs, so that tier belongs
  solely to the static-prefixed roots + mounts and the two URL spaces stay
  disjoint.
[...]

PR #8344, opened 2026-08-18
feat(compiler)!: importing an ambient typing name is a hard error, and the repo drops the imports (#8323)

marsninja:
Executes the collision-rule half of #8323 item A10 (language decision 4, as amended): importing a name the language already provides ambiently is a hard error, and the repo migrates off every such import.

## The audit came first, and it settled two open questions

The escalation was gated on a per-name identity audit, because two earlier probe designs disagreed about whether the spellings diverge. Both turned out to be measuring the harness: `Expr.type` is a lazily-stamped cache that NO_TC compiles never populate, and "both spellings in one file" is unmeasurable by construction — an import shadows the ambient binding module-wide. The trustworthy channel is `get_type_of_symbol` after the checker runs, with cross-program identity compared via `ClassDetailsShared.canonical_key()`, the compiler's own identity law, on line-aligned fixture pairs.

On that channel, **all 19 probed names certify IDENTICAL**: the 17 ambient names (ambient vs `typing` vs `collections.abc` spelling) and `Optional`/`Union` vs their PEP-604 rewrites — same canonical declaration, same assignability verdicts, same diagnostic streams. The mechanism is structural: the ambient merge injects the typeshed `typing.pyi` symbol into builtins, `collections/abc.pyi` re-exports from `typing`, and the stub hub memoizes by path — every spelling funnels to one declaration per program.

The previously reported "Optional layout-error divergence" is real behavior with a wrong attribution: within a fixed pipeline the spellings stamp and lower identically. What actually differed was **placement** — a `typing`-sourced import whose names are not all native-erased makes `native_default_verdict` refuse the module ("unresolved import 'typing'"), silently landing it in the server codespace and skipping the native diagnostic rim entirely. The layout warning never ran, not never applied. That placement flip is pinned by its own test, and it means this migration is placement-restoring for user code: modules previously demoted by nothing but their import line become native-eligible. The release fragment says so explicitly.

## E1125

`W1103` ("'{name}' is ambient and does not need to be imported") escalates to `E1125`, `blocks_codegen=True`, message naming the remedy — the third instance of the W1104→E1122 playbook. Scope is exact:

- The plain import of any of the 17 ambient names, from `typing` **or** `collections.abc`, is refused.
- **Aliased imports stay legal** (`Callable as Cb`): the ambient set cannot provide a chosen local name — this is where E1125's scope deliberately differs from E1123.
- `Optional`/`Union` are not ambient; their imports stay legal and the PEP-604 rewrite remains style, audited identical.
- Two carve-outs keep the W1103 advisory instead of the error, each for a proven reason: `jaclang/jac0core/` (jac0.py resolves names solely through real imports — the remedy does not exist there) and `jaclang/runtimelib/builtin.jac` (its typing imports are an incidental export surface: `JcirGenPass.builtin_name()` emits `from jaclang.runtimelib.builtin import ClassVar/override/abstractmethod`, resolvable only through those imports; migrating it detonates the dev-mode tree).

## The migration

114 files, 152 import items dropped (99 whole statements). Two line-pins moved with their deleted import lines. Validation: the CI-mirror `jac check` invocation over every checkable changed file is clean; the fmt gate is delta-clean under CI's exact flags; ~960 tests green across 20+ named suites including all touched-fixture consumers; the identity audit itself ships as `test_ambient_identity_audit.jac` (21 tests: 19 identity, the layout signal, the placement map).

Closes nothing on its own; ticks the collision-rule box of #8323's item A10.


MusabMahmoodh (commented):
I read the `jac/jaclang/scale/**` half of this (16 files) rather than the whole diff, since that's the surface I can speak to. Scoping my verdict to that.

The scale edits are clean: +2/-18, every changed line an import line, no logic touched. I checked that no `Callable`/`Iterator`/`AsyncIterator` is used at runtime anywhere in scale (all annotation or type-alias positions), that codegen still re-emits `from typing import ...` for the Python binding (`jcir_gen_pass.impl.jac:3928-3932`), and that no scale module re-exported a typing name another module imports from it. `admin/impl/admin_portal.impl.jac:2` was already dead - `Callable` was referenced nowhere in the file - so that one is a straight win.

Two things.

**1. `scale/events/broker.jac:1` - six scale modules change placement eligibility, not just typing**

Line 1 drops to `import from typing { cast }`, and `cast` is the whole of `NATIVE_ERASED_IDENTITY_CALLS` (`jac0core/codeinfo.jac:199`), so `placement.jac:617` skips this import outright. With `datetime` and `uuid` both bundled in `runtimelib/na_stdlib/`, `broker.jac` has no blocker left and becomes native-eligible under `build.default_codespace = "native"`. The same flip lands on `events/subscriber.jac`, `events/streams/_stream_common.jac` and `deploy/autoscale/factory.jac`, and `models/progress.jac` and `observability/metrics.jac` end up with zero imports at all.

Was the scale set run through the census? `test_native_sorted_empty.jac:3-11` in this PR records that the identical flip on `jac/tests/cli/` SIGSEGV'd in `sorted()` on its first native run, and the identity audit is checker-tier on synthetic probes, so it certifies types rather than lowering. `models/progress.jac:20-23` is a function-valued field that gets called and `observability/metrics.jac:52-55` returns a nested closure, both first-time native exposures.

This doesn't affect default `jac serve` or `jac start --scale`, since `effective_default_codespace()` is `"server"` (`codeinfo.jac:236-251`). It's the native-default and binary-toolchain builds I'd want the census for.

**2. `scale/server/impl/serve.static.impl.jac:1` - merge order with #8175**

This hunk deletes line 2 (`import from collections.abc { Callable }`) while #8175 deletes line 1 (`import mimetypes;`) of the same file. Both hunks start at line 1, so whichever lands second conflicts here, and the resolution has to keep *both* deletions - if #8175's side wins wholesale, `Callable` comes back and E1125 blocks codegen on a serving hot-path file.

Easiest order is this one first, then #8175 rebases onto it. I'll flag the same thing on #8175 so it isn't a surprise there. #8170, #8128 and my #8334 have no real overlap with this (8334 touches `kubernetes_utils.jac:9`, four lines from your deletion at `:5`, and git merges it).

Issue #8311, opened 2026-08-17
Jac Scale: move the circuit breaker tests out of test_sv_auth_forward.jac


akindu-k:
**Area:** runtime core / rpc (tests)
**Severity:** low (maintainability)

### What happens

`jac/jaclang/scale/tests/server/test_sv_auth_forward.jac` is now majority circuit-breaker tests. After #8064 it holds 42 tests across 1156 lines, of which 22 tests and roughly 660 lines (from the `jaclang.scale.runtime.rpc.breaker` import block through `sv_service_call does NOT retry on application error envelope`) are breaker tests that have nothing to do with auth forwarding.

The file's name says auth forwarding, so anyone opening it to add or debug an auth header test has to scroll past the breaker state machine, its concurrency tests, and its lease-reclamation tests to find the six tests the filename promises.

### Why it is not urgent

Every test passes and lives in a coherent order today. This is purely about where the next person looks.

### What to do

Move the breaker tests into a sibling `jac/jaclang/scale/tests/server/test_sv_breaker.jac`:

- The 22 breaker tests, plus the `_trip_breaker`, `_take_probe` and `_PROBE_TTL` helpers that only they use.
- `_install_unreachable_transport`, `_install_unreachable_walker_transport`, `_StubMsCfg`, `_install_ms_config` and `_restore_ms_config` are shared with the auth and timeout tests that stay behind, so they need a shared home rather than a second copy. A `_sv_test_support.jac` next to both files is the obvious place.

This should be a pure move: no test body edits, no assertion changes. Anything that has to change to make the move work is a finding worth raising separately.

### Origin

Raised by @MusabMahmoodh as a non-blocking follow-up during review of #8064:

> After this PR the majority of this file is breaker tests. Fine to land here, but a follow-up that moves them to a test_sv_breaker.jac would spare the next person grepping for auth tests. Not blocking.

christianwilkins:
Test split proposed in draft PR #8351.

Current `main` has 29 tests in the 750-line auth-forwarding file; 9 breaker/retry tests remain in the contiguous breaker block described by this issue. The PR moves that block unchanged into `test_sv_breaker.jac` and moves shared mocks/transports into `_sv_test_support.jac`. Both test files pass together (29 tests), and the moved block matches `main` except for its trailing blank line.

christianwilkins:
[...]

PR #8351, opened 2026-08-18
test(scale): split circuit breaker coverage (#8311)

christianwilkins:
## What changed

The 12 circuit breaker and retry tests now live in `test_sv_breaker.jac`. The other 21 tests stay in `test_sv_auth_forward.jac`, including the three current main routing tests. Both files use the request and transport helpers in `_sv_test_support.jac`.

## Validation

Rebased onto `3ad7d91b5`. All 33 test names and complete test blocks are byte identical to that main revision after formatting: 21 auth and routing tests plus 12 breaker and retry tests. Both files passed together (33 passed). Formatting and `git diff --check` pass.

## Follow-up

The transport helper replacement remains a separate, nonblocking follow-up in #8439. The shared helpers still patch the HTTP transport, so that issue remains relevant. GitHub rejected both the REST and CLI reopen attempts; a maintainer needs to reopen it.

Refs #8311.


MusabMahmoodh (commented):
Reviewed the draft since it lands in a file I care about. The split is correct and I verified the conservation claim independently rather than taking it on trust: parsing test blocks out of `test_sv_auth_forward.jac` on main gives 29, and this head gives 20 there plus 9 in `test_sv_breaker.jac`. No test name is dropped, and every moved body is byte-identical to its original. For a move-only PR that is the whole question, so this is good to go from my side once you take it out of draft.

Good call scoping to the block that is actually there now rather than to the 22-tests-in-1156-lines the issue described. The file moved underneath the issue, and saying so in the body is the right way to handle it.

One thing, not a blocker and not really about this PR.

`_sv_test_support.jac:21` and its two siblings work by assigning over module-level functions (`sv_http.post_stream_aware_sync = fake`). I checked the restore discipline before raising this and it is clean: all five call sites capture the original and put it back in a `finally`, so nothing leaks between tests. The concern is direction rather than correctness. Our house rule for scale tests is no mocks and no monkeypatched collaborators, with a real stand-in at the boundary instead, and promoting these helpers from one file into a shared `tests/_sv_test_support.jac` makes them the obvious thing to reach for in the next test that needs a transport.

Since the extraction is happening anyway, this is the cheapest moment to point them at an in-process HTTP server instead of a patched module attribute. The breaker tests would then exercise the real client path, including the connection error they currently synthesize. If that is a bigger job than this PR wants, I would rather it be a follow-up issue than a reason to hold the split, and I am happy to write that one up.

MusabMahmoodh (approved):
Re-verified the conservation claim at this head rather than carrying my Aug 19 answer forward, since the scope grew from 9 moved tests to 12.

It holds exactly. 12 test names leave `test_sv_auth_forward.jac` and the same 12 arrive in `test_sv_breaker.jac` - no name dropped, none invented, and the set difference in both directions is empty. `main` has 32 in the auth file, 12 move, 20 remain, which is what the body says. Diffing the moved bodies line for line, everything that leaves the auth file lands either in `test_sv_breaker.jac` or in `_sv_test_support.jac`, and the only lines in the new file that did not come from the old one are its import block. For a move-only PR that is the whole question, so: approving.

Pulling #8075's three retry/transport tests across with the original nine was the right call. They read as retry tests but they share breaker state, so leaving them behind would have split one subsystem's coverage across two files and produced exactly the confusing failure this PR exists to prevent.

One observation, still not a blocker and still not really about this PR. Lifting `_MockRequest`, `_MockResponse`, `_install_mock`, `_install_unreachable_transport` and their siblings into a named, importable `_sv_test_support.jac` makes them look like a supported seam rather than local scaffolding, and two files now depend on them instead of one. That is a fair consequence of splitting the file and I would not hold this PR for it - but it does raise the stakes on #8439, because the next person writing an sv test will now find a support module and use it rather than reaching for a real in-process server. Worth saying on #8439 that the helper set got more permanent, not less, while that issue was open.

christianwilkins (commented):
Rebased and validated at `cdc953488869dd326da4cfd3f6e69b58ae66b4de`, based on main `3ad7d91b5`.

@MusabMahmoodh The rebase preserves all 33 current-main test names and complete test blocks byte for byte: 21 auth/routing tests remain and the same 12 breaker/retry tests move. The three newer routing tests remain covered. Both files pass together (33 passed); formatting and `git diff --check` pass. GitHub reports no merge conflicts and pre-commit.ci is green for this head.

Current full CI is awaiting maintainer approval: https://github.com/jaseci-labs/jac/actions/runs/34263667983. @akindu-k, please approve this run and review the updated head before merge. The August 26 approval was for the earlier head, so I am not treating it as approval of these rebased commits.

#8439 is still relevant and nonblocking. The shared helpers still patch the transport. GitHub rejected my attempts to reopen the issue after its inactivity closure; a maintainer needs to reopen it. #8311 was also closed for inactivity even though this PR remains open.

Status: locally validated and conflict-free; waiting for full CI approval and maintainer review.

jac/jaclang/scale/tests/server/test_sv_breaker.jac
```diff
@@ -0,0 +1,407 @@
...
+        assert call_count["n"] == 1;
+        assert "deterministic_fail" in raised_msg;
+    } finally {
+        sv_http.post_stream_aware_sync = original_post;
+        sv_client.unregister("app_err_mod");
+    }
+}
+
+"""Count round trips and delivered POSTs apart; an abandoned connect is not a delivery."""
+def _start_counting_server(counts: dict, reply: (bytes | None)) -> tuple {
```
MusabMahmoodh, line 278: This and `_stop_counting_server` now exist twice: the rebase kept main's copies in `test_sv_auth_forward.jac` too (lines 359-409), and nothing in that file calls `_start_counting_server` any more since its three users moved here. Both files use `_stop_counting_server`, so the pair belongs in `_sv_test_support.jac` next to the transport helpers, with the auth file's copies deleted.

MusabMahmoodh (approved):
Re-checked at cdc9534 against main: 33 tests before and after (21 stay, 12 move), every test block byte-identical, and the breaker imports still resolve against `jaclang.scale.runtime.rpc.breaker`. One leftover from the rebase inline.

Issue #8276, opened 2026-08-16
jac check misses undefined bare names that collide with an archetype field, jac run then fails with NameError


MalithaPrabhashana:
## Summary

Inside an archetype method, a bare identifier that happens to match a field name of
that archetype **type-checks as if `self.` were implied**, but at runtime it is not in
scope. `jac check` passes and `jac run` dies with `NameError`.

The checker and the runtime disagree about implicit `self`, and the checker is the
optimistic one, so the failure is invisible until execution.

## Reproduction

```jac
obj Widget {
    has title: str = "hello";
    def size -> int;
}

impl Widget.size {
    return len(title.strip());   # `title` is a field, written without `self.`
}

with entry {
    w = Widget();
    n = w.size();
}
```

```
$ jac check repro.jac
1 passed

$ jac run repro.jac
✖ Error: name 'title' is not defined
```

## What makes it specific

The checker is not simply blind to undefined names. It catches them everywhere **except**
when the name collides with a field of the enclosing archetype:

| case | `jac check` |
|---|---|
| bare name matching **no** field, in a plain `def` | caught, `error[E1032]` |
| bare name matching **no** field, in an `obj` method | caught, `error[E1032]` |
| bare name matching **no** field, in an `impl` block | caught, `error[E1032]` |
| bare name matching **no** field, in `with entry` | caught, `error[E1032]` |
| **bare name matching a field, in an `impl` block** | **passes, then crashes at runtime** |
| **bare name matching a field, in an inline method body** | **passes, then crashes at runtime** |

So it is the field-name collision that flips a correct diagnostic into silence, and it
affects inline method bodies as well as `impl` annexes.

## Versions

Reproduced on:

* `v0.36.0` (released)
* `main` at `8848f9f22`

Both give `check: PASSES` / `run: name 'title' is not defined`.

## Why this is worth fixing

The diagnostic that would have caught it already exists and already fires for the
non-colliding case, so this is a scope-resolution gap rather than a missing feature.

It is also expensive for tool-driven workflows, where `jac check` is treated as ground
truth for "is this file correct". A real JacCoder build hit this while writing a todo
app: the component's own state fields (`newTitle`, `todos`, `adding`, `error`) were
referenced without qualification, and the agent rewrote the same file **five times**
against **byte-identical** error output, because nothing named the actual problem. That
build was on `0.34.14`, where the symptom surfaced as
`Type is Unknown, cannot access attribute "strip"` pointed at the *usage* line rather
[...]

PR #8353, opened 2026-08-18
fix(checker): reject bare archetype fields

christianwilkins:
Refs #8276.

## What changed

Unassigned bare archetype fields used as attribute receivers, such as `title.strip()`, no longer resolve as implicit members. Lexical locals, module globals, imports, builtins, assigned fields, and reactive ability-owned state retain their bindings.

The rebase onto `3ad7d91b5` preserves main's walker-based checker hooks and compile-time and type-only use checks.

## Review follow-up

Added the requested `NestedRebindControl`: an annex method with a same-named local assignment inside an `if` block. The diagnostic assertion now identifies the inline, annex, and node field reads separately. Mutation validation proves the control fails when `_is_runtime_local_name` stops protecting assigned locals; normal checker tests pass.

This PR covers attribute receivers only. Standalone field reads, bare ability calls, and bare subscript receivers remain in #8466. That issue still reproduces on current main; its body now includes the subscript case and current validation. GitHub rejected my reopen attempt, so a maintainer needs to reopen it.

## Validation

- Checker regression suite: 80 passed.
- Mutation test with the runtime-local guard disabled: the new nested annex control fails as expected; production code was restored.
- `jac check` on `project/config.jac` and `scale/websocket/websocket.jac`: 2 passed.
- Full compiler-pass suite: 951 passed, 1 failed, 5 skipped.
- Matching main baseline: 950 passed, the same 1 failed, 5 skipped.
- The shared failure is `bug_cycimp1_reentrant_import_guard_returns_stub`, at the settled-module assertion. It passes in isolation on main and fails in the full suite on both revisions.
- Formatting and `git diff --check` pass.

The initial broader run also hit an outdated LLVM shim. Rebuilding the shim from current source removed the executable-link failures; the full-suite results above are from the corrected environment. The shared main failure and repository CI still need maintainer attention before merge.

### DIFF
--- a/jac/jaclang/compiler/passes/impl/type_checker_pass.impl.jac
+++ b/jac/jaclang/compiler/passes/impl/type_checker_pass.impl.jac
@@ -6,42 +6,16 @@ impl TypeCheckPass.exit_name with Name exit {
     if here.__class__ is not Name {
         return;
     }
+    if self.evaluator._is_non_expression_name(here) {
+        return;
+    }
     is_bare_field = (
         here.sym is not None
         and here.sym.sym_type == SymbolType.HAS_VAR
         and isinstance(here.sym.parent_tab, Archetype)
         and here.expr_ctx == ExprCtx.LOAD
-        and not (
-            isinstance(here.parent, AtomTrailer)
-            and here.parent.is_attr
-            and here.parent.right is here
-        )
     );
     if here.sym is None or is_bare_field {
-        if isinstance(here.parent, AtomTrailer) and here.parent.right is here {
-            return;
-        }
-        if isinstance(here.parent, HasVar)
-            and here.parent.defer
-            and here is not here.parent.name {
-            return;
-        }
-        if isinstance(here.parent, KWPair) and here.parent.key is here {
-            return;
-        }
-        if isinstance(here.parent, SubTag) {
-            return;
-        }
-
-        if isinstance(here.parent, ModulePath) {
-            return;
-        }
-        if isinstance(here.parent, JsxElementName) {
-            return;
-        }
-        if isinstance(here.parent, ModuleCode) and here.parent.name is here {
-            return;
-        }
         self.evaluator.get_type_of_expression(here);
     }
     if here.sym is not None {
--- a/jac/jaclang/compiler/passes/impl/type_checker_pass.impl.jac
+++ b/jac/jaclang/compiler/passes/impl/type_checker_pass.impl.jac
@@ -14,6 +14,7 @@ impl TypeCheckPass.exit_name with Name exit {
         and here.sym.sym_type == SymbolType.HAS_VAR
         and isinstance(here.sym.parent_tab, Archetype)
         and here.expr_ctx == ExprCtx.LOAD
+        and self.evaluator._is_attr_receiver_name(here)
     );
     if here.sym is None or is_bare_field {
         self.evaluator.get_type_of_expression(here);
--- a/jac/jaclang/compiler/passes/impl/type_checker_pass.impl.jac
+++ b/jac/jaclang/compiler/passes/impl/type_checker_pass.impl.jac
@@ -15,6 +15,7 @@ impl TypeCheckPass.exit_name with Name exit {
         and isinstance(here.sym.parent_tab, Archetype)
         and here.expr_ctx == ExprCtx.LOAD
         and self.evaluator._is_attr_receiver_name(here)
+        and not self.evaluator._is_runtime_local_name(here)
     );
     if here.sym is None or is_bare_field {
         self.evaluator.get_type_of_expression(here);
--- a/jac/jaclang/compiler/passes/impl/type_checker_pass.impl.jac
+++ b/jac/jaclang/compiler/passes/impl/type_checker_pass.impl.jac
@@ -6,7 +6,18 @@ impl TypeCheckPass.exit_name with Name exit {
     if here.__class__ is not Name {
         return;
     }
-    if here.sym is None {
+    is_bare_field = (
+        here.sym is not None
+        and here.sym.sym_type == SymbolType.HAS_VAR
+        and isinstance(here.sym.parent_tab, Archetype)
+        and here.expr_ctx == ExprCtx.LOAD
+        and not (
+            isinstance(here.parent, AtomTrailer)
+            and here.parent.is_attr
+            and here.parent.right is here
+        )
+    );
+    if here.sym is None or is_bare_field {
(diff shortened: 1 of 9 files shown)


MalithaPrabhashana (changes requested):
Thanks for the quick turnaround on the assigned-field cases. I re-ran everything against
`80ceba849`, comparing it to its base `d578aa5b7`, and the fix holds up. Note that `main` has
moved 7 commits since that base, so a rebase before merge would be worth it.

## What I verified

- Whole-repo `jac check .` with the CI lane's ignore list: **872 passed, 0 errors**.
- `jac test jac/tests/compiler/passes`: **896 passed, 5 skipped**. Plain `main` gives 895 passed,
  5 skipped, so the only change is your new test.
- `jac check` on all 91 files in `jac/examples`: the diagnostics are identical to `main`. 74 of
  those files emit warnings, so this is a real comparison and not an empty one.
- The `sem` regression is gone, and all 3 `test_sem_def_match_pass` tests pass.
- Reactive `has` in `def:pub` components still works. `mini_todo`'s `if text.strip()` checks
  clean, which was the case I was most worried about.
- The member-access fix still fires everywhere it should: `obj`, `node`, `walker`, `edge`,
  `class`, annex `impl`, f-strings, comprehensions, and walker `can ... with entry`.

Your three false positives reproduce and are fixed. The smallest version I found is an annex
`impl` with the rebind inside a nested block:

```jac
impl W.go -> int {
    if True {
        conn = "local value";
        return len(conn.strip());
    }
    return 0;
}
```

`main` accepts this, `a1c81e66d` rejected it with E1032, and `80ceba849` accepts it again. Both
parts are needed to trigger it. An inline method with a nested block does not reproduce it, and
an annex `impl` with a top-level rebind does not either.

## Before merge

Two small things.

**1. The regression control does not guard the bug.** `RebindControl` uses an inline method
with a top-level assignment, and that shape never broke. Here is the E1032 count in the fixture:

```
a1c81e66d (before the fix): 3
80ceba849 (after the fix):  3
```

The count is the same either way, so `assert len(e1032_errors) == 3` also passes on the broken
commit.

I also removed each guard clause one at a time and re-ran the full pass directory. Four of the
five can be deleted with all 896 tests still green, including `_is_runtime_local_name`. The only
one that is pinned is `_is_attr_receiver_name`, and that is pinned by the older `impl has var
type not shadowed` test, not by anything in this PR.

Could you add this to `checker_bug_bare_field_name.jac`? The assertion stays at 3, since this is
a negative control and should not produce an error.

```jac
obj NestedRebindControl {
    has nested_title: str = "field";

    def size -> int;
}

impl NestedRebindControl.size -> int {
    # A rebind inside a nested block is still a function-scope local,
    # so the field must not be demoted here either.
    if True {
        nested_title = "local";
        return len(nested_title.strip());
    }
    return 0;
}
```

With that block added, `test_checker_bugs.jac` gives `78 passed, 1 failed` on `a1c81e66d` and
`79 passed` on `80ceba849`. That is the difference between catching this regression and missing
it.

**2. `Closes #8276` will close an issue that is only partly fixed.** The body opens with
`Closes #8276`, so merging this will auto-close it. Your own Scope section says this PR covers
the member-access form, and that standalone reads such as `if flag` and bare calls such as
`helper()` are tracked in #8466. Could you change it to `Refs #8276`? Then we can close #8276
once the standalone cases land, instead of leaving them with no open report.

## Not blocking

- `assert len(...) == 3` cannot tell which construct regressed. Every archetype in the fixture
  calls `.strip()`, so any three of them will satisfy both assertions. Checking each construct by
  name would be sturdier.
- One more shape is still silent, and it is not covered by #8466: a bare field under a subscript.
  `return items[0];` only produces warnings and exits 0, then fails at runtime with
  `name 'items' is not defined`. Worth adding to that issue.
- The `SemDef` entry in `_is_non_expression_name` is now redundant. `_is_attr_receiver_name`
  already excludes `sem` targets, and the sem tests still pass if the `SemDef` clause is removed.
  It does no harm, it just is not doing any work now.
- E1032 reads `Type is Unknown, cannot access attribute "strip"` and points at `strip`, so it
  never mentions the missing `self.`. The mechanism is fine, but the wording could be clearer in
  a follow-up, since the goal of #8276 was to make `jac check` trustworthy.

christianwilkins (commented):
Review follow-up completed at `cbef50c79c6a5716ddab12142002d69b8af0f229`, rebased onto main `3ad7d91b5`.

@MalithaPrabhashana Both requested changes are addressed. `69657af0c` adds the nested annex assignment control from your review, and `cbef50c79` identifies each rejected inline, annex, and node field access. The PR now says `Refs #8276`. I also added your bare-subscript case to #8466 and verified that both the standalone field and subscript examples still pass `jac check` but fail at runtime on current main. GitHub rejected my attempt to reopen #8466, so a maintainer needs to reopen it.

The normal checker suite passes all 80 tests. A mutation that makes `_is_runtime_local_name` always return false now fails this regression test on `nested_title.strip()`, proving the new control catches the missing guard. Production code was restored before final validation. Both previously problematic files, `project/config.jac` and `scale/websocket/websocket.jac`, pass `jac check`.

Full compiler-pass validation with the rebuilt current LLVM shim: 951 passed, 1 failed, 5 skipped. The same command on main's compiler source and compiler-pass tests gives 950 passed, the same 1 failed, 5 skipped. The shared failure is `bug_cycimp1_reentrant_import_guard_returns_stub`, at the settled-module assertion; it passes in isolation on main. This remains a baseline suite failure, not a clean full-suite result.

The old August CI failure included native sealing and emitter errors and must not be treated as #8467. Current main has since removed those sealing tests. Formatting and `git diff --check` pass; GitHub reports no merge conflicts and pre-commit.ci is green for this head.

Please re-review the updated head and approve the new full CI run: https://github.com/jaseci-labs/jac/actions/runs/34264651261. Status: requested author fixes are complete; maintainer re-review, full CI approval, and assessment of the shared baseline failure remain before merge.

MusabMahmoodh:
@christianwilkins FYI: #9310 (for #9292) resolves bare names inside method bodies the Python way, which covers the cases #8466 listed (standalone field reads, bare ability calls, subscript receivers) and overlaps this PR's receiver case. It keeps the bindings you found valid: a name assigned in the same function stays a local (so `base.resolve()` in `project/config.jac` and `backplane.subscribe()` in `scale/websocket/websocket.jac` check clean). Happy to rebase on top of this PR if it lands first, or to fold your E1032 cases into the new rule; whichever you and the reviewers prefer.

PR #8370, opened 2026-08-19
fix(native): form na->sv bindings for native-pinned callers (single stamped native verdict)

chess10kp:
# PR 1/4 — fix(native): form na->sv bindings for native-pinned callers

## Summary
- `BoundaryAnalysisPass` computed `pinned_server` only for client-context callers, so a native-pinned module importing from a server-pinned provider never formed an na->sv binding. The pin path now admits `InteropContext.NATIVE`, `native_imports` populates, and E5093 enforces on the new bindings.
- The module-level native verdict is stamped exactly once (`gen.native_codespace`, set by `stamp_boundary_facts`); `_in_native_module()` and every boundary consumer read that single fact. The stamp pass and the analysis pass can no longer disagree under `aot_mode` with a native default codespace (previously the stamp pass marked `pinned_server` while the pass registered a meaningless SERVER->SERVER binding).
- **Breaking change**: with na->sv bindings forming for the first time, the E5093 scalar checks enforce on them — `list`/`dict`/object-typed parameters or returns on native->server signatures that previously compiled unchecked now hard-error. See `release_notes/unreleased/jaclang/8370.bugfix.md`.

This is part 1 of 3 addressing the review in https://github.com/jaseci-labs/jaseci/pull/8370#issuecomment-5426876586. Follow-ups: #9332 (native-build refactor), #9333 (entry-library dispatch lane).

## Test plan
- [ ] `jac test tests/compiler/test_interop_manifest.jac` — na->sv binding formation for native-pinned callers
- [ ] Existing boundary/E5093 checker lanes stay green


chess10kp:
@marsninja Follow-up on your [deep review](https://github.com/jaseci-labs/jaseci/pull/8370#issuecomment-5426876586): all of it is landed, split exactly per your §6. The dispatch-table design from your §5 turned out to be the whole design — #9333 is built on it.

**1. Boundary fix stands alone here (reworked).** This PR is now 6 files: the driver admission of `InteropContext.NATIVE`, the boundary fix, and the interop-manifest tests. The duplicated native-context test is collapsed into one stamped fact — `stamp_boundary_facts` sets `gen.native_codespace` exactly once, and `_in_native_module()` plus every boundary consumer read that single fact, so the `aot_mode` drift from your §3a can't recur. Your §3b is addressed: the E5093 break is called out in `release_notes/unreleased/jaclang/8370.bugfix.md`. The inert mutable-default and style changes are dropped.

**2. #9332 — nacompile refactor on its own.** `_compile_native_module` / `_link_native_artifact` on a `NativeBuildContext`: one compilation, per-artifact linking. `NativeLinkTarget` dropped, the dead `scrub` param dropped, the W3030 suppression gone. Backend-seam warnings moved to `sys.stderr` so the compiler backend no longer imports the CLI layer.

**3. #9333 — the completing step, as you specified.** na->sv callbacks lower to indirect calls through global `__jac_bind_*` dispatch slots (`bind_shared_callbacks`); the sealed entry library carries **no unresolved Python symbols** and loads cleanly under `RTLD_NOW` — the canary problem from your §2 (last row) is gone at the source. Init registration is deferred to the loading host: it sets `__jac_argv`/`__jac_argc`, fills the slots, then calls `__jac_shared_init` explicitly before `jac_entry`. `run_jab_image` loads interop jabs **in-process** through that lane — the dlopen path is now the real path for interop apps, not a corrupted-bundle fallback; non-interop jabs stay exe-first and the docs say so honestly.

Your §4 concrete defects, all resolved on the branches:

| Defect | Resolution |
|---|---|
| Inert mutable-default fix (§4.1) | Dropped from this PR |
| Cross-compile naming regression (§4.2) | Target-aware extension restored (`native_build.impl.jac:39`: `.dll`/`.dylib`/`.so` from the build target) |
| `_native_lib_suffix` twin (§4.3) | Removed; `precompile_bytecode` imports `native_lib_suffix` from `sealed_serve` — one naming convention |
| Bare-`return` gates (§4.4) | All gates in `test_native_entry_lib.jac` use `testskip()` |
| `scrub` / W3030 / `NativeLinkTarget` (§4.5) | All gone (in #9332) |
| `"__jac_argc" not in llvm_ir_str` substring probe | Define-site regex `argv_globals_defined` (`shared_emit.jac:200`), used at `native_build.impl.jac:182` |
| Auto-promote printed twice | Print lives in `_compile_native_module`, which runs once per compile; the sealed pair compiles once and links both artifacts from it |
| Docs say `lib<stem>.so` | Docs now say `<stem>.so` / `.dylib` / `.dll` (`cli-and-native.md:61`) |

On your §2 re-implementation table: the `.interop.json` sidecar no longer describes an unloadable artifact — with the callbacks bound at define time there are no unresolved Python symbols to warn about. The sidecar records the callback contract (name, ctypes-wire signature, source module per callee) and drives the runner's load order (register callbacks → argv → init → entry). Interop seals drop the standalone exe entirely, so nothing ships that no code path can load.

Verification at the current stack head: 20 tests green (14 entry-lib + the interop-manifest lanes), a native smoke binary runs clean, and targeted `jac precommit` passes on all 23 stack `.jac` files. Merge order is #8370 → #9332 → #9333 (cross-fork stacking isn't supported, so all three target `main`; #9333's diff shrinks to its own changes as the earlier two merge).

chess10kp:
Restructured as the bottom of a PR stack: #8370 → #9332 → #9333. Rebased onto current main (was 9 behind; no file overlap, clean rebase), and the empty CI-retrigger commit is gone — force-pushed `88443bb0e` → `317bc5642`, content-identical, new base. This branch should merge first; the other two re-diff automatically as it lands.

Issue #8381, opened 2026-08-19
byLLM: finish_tool exit on a streamed turn shrinks the tools array and breaks the Anthropic prompt cache


SahanUday:
## Summary

When a `by llm()` agent is called with `stream=True` and a `str`/`None` return type, ending a ReAct turn by calling `finish_tool` triggers one extra LLM call (`_stream_final_answer`) that strips the tools array down to `[finish_tool]` and sets `tool_choice="none"`. Anthropic's prompt-cache breakpoint sits on the tools block, which serializes before the message list, so that swap invalidates the cached prefix for the entire conversation — it gets rebilled at the cache-**write** rate instead of the cache-**read** rate. The same conversation can avoid the penalty entirely if the model happens to end the turn with plain text instead of a `finish_tool` call, so the cost is non-deterministic for otherwise-identical turns.

## Where

`jac/jaclang/byllm/llm.impl/basellm.impl.jac`, current `main` (`7932ab7`):

- `_stream_final_answer` (~L1748-1781) — on a successful `finish_tool` call, replaces `mt_run.tools` with `[finish_tool_obj]` and sets `tool_choice="none"` before making another streaming call to narrate the answer.
- `_apply_prompt_caching` (~L234-255) — places the cache breakpoint on `tools[-1]`.
- `_invoke_streaming` (~L1830-2194) — the two legal turn-exits: model calls `finish_tool` (~L2122-2149, triggers `_stream_final_answer`) vs. plain-text stop with no tool call at all (~L2151-2168, where `_attempt_recovery` returns `None` whenever `resp_type in (None, str)` and the text is accepted as-is with no extra call).

Both exits are legal given how the loop is written, and which one a given turn takes is the model's own free choice — same prompt, same tools, same model, can go either way turn to turn.

## Reproduction (no live LLM calls)

A real `Model` instance drives `_invoke_streaming` unmodified; only `model_call_with_stream` is mocked to serve hand-built, litellm-shaped streaming chunks, so `make_model_params`, `_apply_prompt_caching`, `_stream_final_answer`, `_attempt_recovery`, and `litellm.stream_chunk_builder` all run for real.

```
scenario A (model exits via finish_tool):
  call 1: tools=[search_files, write_file, run_command, finish_tool]  (4 tools)
  call 2: tool_choice='none', tools=[finish_tool]                     (1 tool)
  tools payload identical between call 1 and call 2? False

scenario B (model exits via plain text, no tool call):
[...]

PR #8386, opened 2026-08-19
fix(byllm): finish_tool-related calls stop shrinking the tools array (Anthropic cache-prefix fix)

SahanUday:
## Summary

Fixes the prompt-cache invalidation described in #8381: three functions in `basellm.impl.jac` shrank `mt_run.tools` down to `[finish_tool]` before making a follow-up `finish_tool`-related call. Anthropic's cache breakpoint sits on `tools[-1]`, and tools serialize ahead of the message list in the wire request, so that swap invalidated the cached prefix for the *entire* conversation every time one of these calls fired -- rebilling the whole accumulated context at the cache-write rate instead of a cache-read.

## What changed

All three tools-shrinking sites now keep `mt_run.tools` untouched, and use whichever `tool_choice` correctly expresses that call's own intent (cache preservation only requires the `tools` array to be byte-identical to prior calls; `tool_choice` is a separate field and doesn't affect the cached prefix at all):

- **`_stream_final_answer`** (the `finish_tool` exit / `max_react_iterations` cutoff / `on_iteration` `ABORT_WITH_SUMMARY`, all under `stream=True`): `tool_choice="none"` -- the model is only asked to narrate an already-decided answer, so no tool call is legitimate here at all.
- **`_force_final_answer`** (the non-streaming counterpart of the same two forced-cutoff triggers): `tool_choice = {"type": "function", "function": {"name": "finish_tool"}}`. Its own instruction text ("provide only your final answer") never mentions `finish_tool`, and today it sets no `tool_choice` at all -- the model can already legally reply in plain text instead, which the code has always tolerated via a plain-text fallback. Forcing `finish_tool` specifically now gets a validated, typed answer through `get_output()` whenever the model complies, for any `resp_type`; the existing fallback stays in place as defensive code.
- **`_attempt_recovery`** (forces a bare-text reply with a structured `resp_type` back through `finish_tool`): same forced `tool_choice`. Its own message already says *"You must use finish_tool to return your response. Call finish_tool now"* -- forcing `tool_choice` turns that from a hopeful prompt-level instruction into an actual API-level guarantee.

Verified with litellm directly that `{"type": "function", "function": {"name": "finish_tool"}}` translates correctly to Anthropic's native `{"type": "tool", "name": "finish_tool"}`, with the tools array passed through unchanged.

## Verification

**Mocked regression tests** (`jaclang/byllm/tests/test_byllm.jac`), no network, each asserting the tools array sent to the model is unchanged from the prior call:
- `_stream_final_answer keeps the tools array intact (no cache-prefix break)`
- `finish_tool exit sends an identical tools array on the narration call`
- `_force_final_answer keeps the tools array intact and forces finish_tool`
- `_attempt_recovery keeps the tools array intact and forces finish_tool`

All four fail on the pre-fix code and pass after. Full targeted suite: `jac test jaclang/byllm/tests/test_byllm.jac` -- 113 passed, 1 skipped (pre-existing, unrelated) -- including the pre-existing `_force_final_answer`/`_attempt_recovery` tests, confirming no regression in either function's existing behavior.

**Live Anthropic calls** (`claude-sonnet-4-5`):

`_stream_final_answer`, same 3-tool `finish_tool`-exit turn, full `usage.per_call` breakdown, before vs. after:

After (this fix):
```
call 1: prompt_tokens=2803 cache_read_input_tokens=0    cache_creation_input_tokens=2800
call 2: prompt_tokens=3055 cache_read_input_tokens=2800 cache_creation_input_tokens=248
call 3: prompt_tokens=3227 cache_read_input_tokens=3048 cache_creation_input_tokens=176
```

Before (pre-fix):
```
call 1: prompt_tokens=2803 cache_read_input_tokens=2800 cache_creation_input_tokens=0
call 2: prompt_tokens=3069 cache_read_input_tokens=2800 cache_creation_input_tokens=262
call 3: prompt_tokens=3033 cache_read_input_tokens=0    cache_creation_input_tokens=3030
```

`call 3` is the post-`finish_tool` narration call (`steps_done` reports 2 iterations, so it's the untracked 3rd `usage.per_call` entry). Before: cache read collapses to 0 and the call rewrites the whole conversation (3030 fresh tokens). After: it reads 3048 cached tokens and writes only 176 new ones. (`call 1`'s numbers differ between the two runs because the "after" run executed first and its cache was still warm when the "before" run started moments later; `call 3` is unaffected by that.)

`_force_final_answer`, real end-to-end run with `max_react_iterations=1` (forces the cutoff right after the first genuine tool call, so a real prior tool-result message is already in history -- the same shape that crashed against real Anthropic before this codebase's earlier "Anthropic tool-calling compatibility" fix): full tools array sent intact, `finish_tool` forced and called, correctly returns a validated `Person` instance, no crash.

`_attempt_recovery`, called directly against a real Anthropic conversation: full tools array sent intact, `finish_tool` forced and called, correctly returns a validated `Person` instance, no crash.

## Not included

Tier 2 from #8381 (skipping the `_stream_final_answer` narration call entirely and reusing `finish_tool`'s own output, mirroring the pattern already shipped for the plain-text exit) is a separate, bigger change -- it removes token-by-token narration for that exit path, which is documented, user-visible behavior, and needs its own validation before shipping.


MalithaPrabhashana (approved):
Re-checked the new commits - the extension to the other two sites and the compressed tests both hold up.

The part worth verifying was the forced `tool_choice`, since OpenAI-format tool_choice has to survive litellm's translation to Anthropic. Confirmed live on `claude-opus-5`, driving `_force_final_answer` through the `max_react_iterations` cutoff:

```
call 2 (_force_final_answer):
  tools=3 ['search_files', 'run_command', 'finish_tool']
  tool_choice={'type': 'function', 'function': {'name': 'finish_tool'}}
```

Provider accepted it and returned a valid typed answer. Before/after on that same path:

| `_force_final_answer` call | tools sent | cache_read | cache_write |
|---|---|---|---|
| before | `['finish_tool']` | 0 | 24,268 |
| after | all 3, forced choice | 632 | 23,943 |

Also ran all four tests against the pre-fix code: all four fail with the right messages, so they are real guards for all three sites.

Thanks for picking up `_attempt_recovery`.

PR #8391, opened 2026-08-19
feat(cli): bind explicit --port/--api_port strictly instead of relocating

### LINKED ISSUE #8387: [Bug]: jac start silently binds a port different from the explicit -p (0.36 also reports the wrong one)
## Summary

`jac start -p N` treats an explicit port as a preference, not a requirement. If `N` is
occupied, jac walks to the next free port, prints a normal ready banner, and keeps
running with exit status 0. On 0.36.0 it goes further and prints the port you asked
for while listening on a different one.

This is fine on a laptop. It is a silent outage for any managed environment where the
port is a contract (a reverse proxy, a container port mapping, a health probe). We hit
exactly that in production: a preview server relocated from 8000 to 8002, kept reporting
itself healthy, and the proxy in front of it served connection errors for 12 minutes.

A second, related defect made the port busy in the first place: `jac start --dev` does
not terminate its Vite child on SIGTERM. Details in "Related defect" below. Happy to
split these into two issues if preferred.

## Environment

- OS: Ubuntu (WSL2, Linux 6.6.87.2)
- jac `0.36.0` (Linux x86_64) and jac `0.34.15` (Linux x86_64)
- Project: unmodified `jac create <name> --kind web-app`

### Reproduction

```bash
jac create portwalk --kind web-app && cd portwalk

# occupy the port with something that is not jac
python3 -c "import socket,time; s=socket.socket(); \
s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1); \
s.bind(('0.0.0.0', 8000)); s.listen(5); time.sleep(600)" &

jac start main.jac -p 8000 --no-dev > out.log 2>&1 &
sleep 30

grep -E "Gateway|App:|API:" out.log
ss -tlnp | grep -E ":80[0-9][0-9] "
```

### Actual result, jac 0.36.0

```
-- who owns :8000 before jac starts --
LISTEN 0  5    0.0.0.0:8000  0.0.0.0:*  users:(("python3",pid=245259,fd=3))

-- what jac reported --
  Gateway: http://0.0.0.0:8000
  ● endpoints          /endpoints           -> http://127.0.0.1:18021
  ● main               /__main__            -> http://127.0.0.1:18689

-- what jac actually bound --
  LISTEN 0  5    0.0.0.0:8000  0.0.0.0:*  users:(("python3",pid=245259,fd=3))
  LISTEN 0  100  0.0.0.0:8001  0.0.0.0:*  users:(("jac",pid=245272,fd=8))

-- jac process still alive (i.e. it did not fail)? --
  YES, running as pid 245272
```

The string `8001` appears **zero** times in the whole log. Both `--dev` and `--no-dev`
behave identically. With 8000 free, the same command binds 8000, so the 8001 binding is
the fallback, not a fixed second port.

### Actual result, jac 0.34.15

Same relocation, but the banner is at least truthful:

```
-- did it warn about the port? --
  (no warning)
-- what the banner says --
    ➜  Local:     http://localhost:8002/
      App:  http://localhost:8002/
      API:  http://localhost:8001/
```

### Expected result

With an explicitly passed `-p`, jac should fail:

```
Error: port 8000 is already in use
```

Falling back is a reasonable default only when no port was requested.

### Where

Two independent walkers, neither gated on whether the user asked for a specific port,
and no strict-port option exists anywhere in the tree (`grep -r strict_port` returns
nothing):

- `jac/jaclang/scale/server/impl/serve.core.impl.jac:291`
  `self.port = _find_available_port(self.port);` (up to `PORT_RETRY_LIMIT` = 10 ports).
  This one does print `Port N is in use, using port M instead`, but only for the API
  port, and the 0.36 gateway path above reached 8001 without printing it.
- `jac/jaclang/cli/commands/impl/execution.impl.jac:711`
  `free_vite_port = _find_free_port(vite_port, skip_port=actual_port);` (up to 50 ports).
  This one prints nothing at all. It governs the **app** port, the one a browser and a
  proxy actually talk to.

## Related defect: `jac start --dev` orphans its Vite child on SIGTERM

`_start_vite_dev_server` spawns Vite as a subprocess. It is terminated only inside
`except KeyboardInterrupt` (`execution.impl.jac:731`), i.e. SIGINT. There is no SIGTERM
handler on this path, so SIGTERM (a `kill`, or ordinary container shutdown) kills the
jac parent and leaves Vite running and holding the app port.

Reproduced on 0.34.15:

```
===== PHASE 1: normal start on -p 8000 =====
      App:  http://localhost:8000/
      API:  http://localhost:8001/
  LISTEN 0 2048  0.0.0.0:8001  users:(("jac",pid=247444,fd=22))
  LISTEN 0 512         *:8000  users:(("bun",pid=247541,fd=13))
   247444  247441  jac start main.jac -p 8000 --dev
   247541  247444  .../_bun/bun .../vite/bin/vite.js --config configs/vite.dev.config.js --port 8000 --host
  curl :8000 -> 200

===== PHASE 2: pkill -f "jac start" =====
  parent gone
-- what SURVIVED the pkill --
   247541     866  .../_bun/bun .../vite/bin/vite.js --config configs/vite.dev.config.js --port 8000 --host
  LISTEN 0 512  *:8000  users:(("bun",pid=247541,fd=13))
  curl :8000 -> 200
```

Note the child is not reachable by the same pattern: `pkill -f "jac start"` cannot match
`bun .../vite/bin/vite.js`, so the usual "restart the server" incantation leaves the port
held. It reparents (PPID 247444 -> 866) and keeps serving the previous build.

Not reproduced on 0.36.0: the default `--kind web-app` scaffold now takes the process
gateway path, which installs a real SIGTERM drain handler
(`scale/runtime/lifecycle/drain.jac:186`) and shuts its children down cleanly. But the
`jac start --dev` Vite path in `execution.impl.jac` is unchanged on `main`, and
`WebTarget.dev` (`runtimelib/client/targets/impl/web_target.impl.jac:59`) has the same
KeyboardInterrupt-only cleanup, so the gap is still in the source for projects that reach
those paths.

## The two together: the actual failure

Phase 3 of the same script, restarting on the same explicit `-p 8000` while the orphan
still holds it:

```
===== PHASE 3: restart on the SAME -p 8000 =====
-- did it warn about the port? --
  (no warning)
-- what the banner says --
      App:  http://localhost:8002/
      API:  http://localhost:8001/
-- listeners --
  LISTEN 0 2048  0.0.0.0:8001  users:(("jac",pid=247673))   <- new API
  LISTEN 0 512         *:8000  users:(("bun",pid=247541))   <- ORPHAN, previous build
  LISTEN 0 512         *:8002  users:(("bun",pid=247748))   <- new app
-- a proxy pinned to :8000 would get --
  http://localhost:8000/ -> 200
```

Three listeners, two Vite servers fighting over the same generated
`configs/vite.dev.config.js`, and a proxy pinned to 8000 getting HTTP 200 from a stale
build. Every signal says healthy.

## Why it matters

In our managed preview environment the container command is a supervisor loop running
`jac start main.jac -p 8000 --dev`, and a reverse proxy forwards `preview-*.example.com`
to pod IP `:8000`. One SIGTERM produced exactly the sequence above. The supervisor
restarted correctly every time, jac reported ready every time, and the preview was
unreachable for 12 minutes because the app had moved to 8002 and nothing in the system
was able to notice. The agent driving that workspace spent 36 of its 100 tool calls on it.

Strict-port would have converted a silent 12 minute outage into an obvious crash loop in
the first 5 seconds.

## Suggested fix

1. Add `--strict-port`, or make an explicitly passed `-p`/`--port` strict by default,
   raising on `EADDRINUSE` instead of walking. Apply to both `_find_available_port` and
   `_find_free_port`. Keep the walk only when no port was requested.
2. If the walk stays for any path, the ready banner must print the port that was actually
   bound. The 0.36 gateway banner currently prints the requested one.
3. Install a SIGTERM handler on the `jac start --dev` path that runs the same child
   cleanup as the `KeyboardInterrupt` branch, so Vite never outlives its parent.

Repro scripts and the full transcripts are reproducible from the snippets above; happy to
attach them or open a PR for (1) and (3).

MalithaPrabhashana:
Closes #8387. Rewritten after review: the `--strict-port` flag is gone, the two duplicate port finders are one, and strictness now reaches the socket that actually serves.

> **Note on command spelling.** Upstream #8596 ("run absorbs start and dev") removed `jac start` and `jac dev`, and this branch has merged it. Everything below uses the current `jac run --serve` spelling; the two references to `jac start` in the incident description above are left as-is because that was the command at the time of the outage.

## Problem

`jac start -p N` treated the port as a preference. If N was taken, jac walked to the next free port, printed a normal ready banner, and kept running with exit status 0.

That is right for a laptop and wrong for anything that pins the port. A reverse proxy, a container port mapping or a health probe still talks to N while the app answers somewhere else, so the app is healthy and unreachable with every signal green.

It cost us 12 minutes of a dead preview in production: a supervisor restarted `jac start -p 8000 --dev`, 8000 was briefly held, the app came up on 8002, and the proxy in front of it kept talking to 8000.

## What changed

**No new flag.** `port` defaults to a sentinel instead of `8000`, which is the shape `api_port` already used in this same signature. Naming a port is itself the statement that it matters:

```jac
port_pinned = port > 0;
port = port if port_pinned else 8000;
```

| command | behaviour |
|---|---|
| `jac run --serve` (no port named) | port busy, relocates and says so. Unchanged. |
| `jac run --serve -p 8000` | port busy, **fails and names the port** |
| `jac run --serve --api-port 9000` | same rule, independently |
| `[serve] port` in `jac.toml` | same rule, and it reaches `jac dev` too |

Config counts because the rule follows the port being *named*, not the flag being typed. "Named, unless you named it in a file" is harder to explain than the behaviour it replaces, and a project that pins 3000 in config usually has something addressing 3000. `--api_port` has no config key and stays command-line only. The full rule lives in the [local tutorial's Port section](jac/jaclang/cli/docs/tutorials/production/local.md), which is the page someone lands on after hitting the new error.

**Strictness reaches the bind, not just the probe.** `find_available_port` is a probe: it opens a throwaway socket, closes it, and the real socket is bound later by `ServingServer._serve`, which walks `chosen += 1` on EADDRINUSE. Checking only the probe fixes the announcement and not the outcome, and the gap between the two is exactly the restart race this PR exists for. Both server classes now hand the bind a retry budget of 0 when the port was pinned, so the existing loop raises instead of walking.

The second of those classes is not just a base: `get_api_server_class` falls back to `runtimelib`'s `JacAPIServer` when the scale module fails to import, which has happened in production. It accepted `strict_port` and never read it.

**One helper instead of two.** `_find_available_port` (scale/server) and `_find_free_port` (CLI) were the same hand-rolled bind-and-walk retry. They are now one `find_available_port` in `runtimelib/net.jac`, carrying `skip_port` and `strict`. The Vite wrapper keeps its `max_retries=50` and its warn-and-fall-back policy; the scale path keeps its 10.

## Correction: the errno tuple was incomplete, but a string fallback hid it

**This section previously claimed that relocation never worked on macOS. That claim is wrong**, and I am correcting it in place rather than editing it away, because @MusabMahmoodh and @kugesan1105 both signed off on it and it is now in the thread twice.

The two copies did disagree about *which errno counts as "in use"*:

| copy | errnos it knew |
|---|---|
| CLI `_find_free_port` | swallowed **every** `OSError` |
| scale `_find_available_port` | 98 and 10048, but **not 48** |

48 is `EADDRINUSE` on macOS, so the errno arm never matched there. What I missed is that the test has a *second* arm:

```jac
if e.errno in (ERRNO_ADDR_IN_USE_LINUX, ERRNO_ADDR_IN_USE_WINDOWS)
    or "Address already in use" in str(e)
```

On macOS `str(e)` is `[Errno 48] Address already in use`, so the string arm matches and the walker relocates normally. Measured on `main`, on macOS arm64, with the port held by an unrelated listener:

```
$ jac run --serve --port 50759
Port 50759 is in use, using port 50760 instead
```

and the condition itself, evaluated on the same machine:

```
e.errno                             = 48
e.errno in (98, 10048)              -> False
"Address already in use" in str(e)  -> True
combined (what main actually tests) -> True
```

So there is no macOS relocation defect, and there was never one in the field. The claim came from reading the constant list; I should have run it before writing it down, and the two approvals it collected are a good argument for measuring the small claims as well as the large ones.

Adding 48 is still worth doing, for a smaller reason than the one I gave: it makes the check depend on an errno rather than on an English string that a localised `strerror` can change out from under it. That is hardening, not a defect fixed as a side effect. `runtimelib/serving/server.jac` already carried the complete `(98, 48, 10048)`; the merged helper uses all three and keeps both arms. The constants test pins all three, including the 48 the errno arm was missing.

## One refusal, rendered the same way everywhere

Round 3 left the two paths disagreeing about how a refusal *looks*, which I had described as fixed when only the wording was. The gateway path let the `OSError` reach the CLI executor, and an error the executor cannot map is printed as a raw Python traceback -- roughly thirty lines through `local_realizer` to `orchestrator` to `process_gateway` to asyncio internals -- before the line that matters:

```
  File ".../serving/server.jac", line 101, in _serve
    raise OSError(f"Port {chosen} is already in use") from None;
OSError: Port 50710 is already in use
✖ Error: Port 50710 is already in use
```

The plain path never showed this because it catches the error itself.

The fix is a type rather than a message. `runtimelib/net.jac` gains `PortInUseError(OSError)` carrying the port; `find_available_port` and `ServingServer._serve` raise it; `cli/errmap.jac` maps it, which is the mechanism the CLI already uses to turn a known failure into a message and a hint. The plain path re-raises it after its cleanup so both paths render through the same map. Subclassing `OSError` keeps every existing `except OSError` site working, including the Vite wrapper's warn-and-fall-back.

Both paths now print:

```
✖ Error: Port 51150 is already in use
💡 A port you named is a contract, so jac fails instead of moving to another one.
   Free the port (lsof -i :<port>), name a different one, or drop --port to let
   jac pick the next free port.
```

`JAC_DEBUG=1` still prints the full traceback, so nothing is hidden, only moved behind a flag. This also closes the "reads as `Server error: ...` rather than a clean refusal" item I left open for @MusabMahmoodh in round 2.

### Reproducing this locally

The dev loop is the `[dev] jaclang_source` stanza in `jac.toml`, read from the **nearest ancestor**. `jac/jaclang/scale/tests/fixtures/jac.toml` has no such stanza, so inside that directory jac silently runs the binary's bundled jaclang instead of the branch, with no warning. The suite spawns its servers with `cwd` set to exactly that directory.

A `zig build -Ddev` binary reroutes to the build-root source from any directory and is the reliable way to check by hand. `JAC_DEV_SOURCE` is not: it prints a dev-mode banner while the command still runs bundled code, so the banner is not proof. Every transcript below was taken against branch source and confirmed with a marker print.

### The paths a port reaches

```
CASE A   explicit port, held
$ jac run --serve app.jac -p 8765 --no-client
exit=1   ✖ Error: Port 8765 is already in use

CASE B   no -p, default 8000 held
Port 8000 is in use, using port 8001 instead      (unchanged, and the case that keeps everyone safe)

CASE C   --api_port held, --dev
exit=1   ✖ Error: Port 49111 is already in use

CASE D   second server against the port the first holds
exit=1   first server undisturbed, healthz 200
```

**The dev + client path**, which is the one the incident actually hit, verified separately rather than assumed to follow the API path:

```
$ jac run --serve --port 52555 --dev test_api.jac  # 52555 held by another listener
  ✔ Initial client compilation completed
exit=1   ✖ Error: Server error: Port 52555 is already in use

  at _find_free_port()  jaclang/cli/commands/execution.jac:905
  at _announce_ready()  jaclang/cli/commands/execution.jac:720
```

The traceback is the point: the refusal comes from the client-port selection in `_announce_ready`, after the client compiled, so this is the Vite branch and not the API branch producing a similarly worded message.

**A port named in `jac.toml`**, measured rather than inferred, since by the time `start` sees the value the two paths are indistinguishable:

```
port = 54444 in jac.toml, nothing typed    server came up on 54444, healthz 200
54444 held, still nothing typed            exit=1  Port 54444 is already in use
```

**`max_retries=0` does not break the happy path**: strict on a free port still serves, `healthz 200`.

### `jac dev`, red to green

`dev` declared `port: int = 8000` and passed it straight to `start`, so `port_pinned` read True even with no `-p`. That broke case B for the command developers run most:

```
RED    branch, 8000 held, `jac dev`        exit=1  Port 8000 is already in use
       main, same command                  App: http://localhost:8002

GREEN  after the fix, `jac dev`            App: http://localhost:8002, keeps serving
GREEN  `jac dev -p 8000`, 8000 held        exit=1  Port 8000 is already in use
```

### Tests

| level | what it pins |
|---|---|
| unit, `an explicitly requested port is never relocated` | both selections a `-p` reaches, plus the case where the pinned port is also the skipped one |
| unit, `port constants` | all three errnos, including 48 |
| integration, `a second server refuses an already-taken explicit port` | a real second `jac run --serve`: non-zero exit, the port named, first server undisturbed |

The unit tests run in-process, so they always measure this branch and they fail without this patch. The integration test spawns a subprocess into the fixtures directory described above, so CI is where it is meaningful; `test-scale (server)` passes.

## Two tests had to change, and that is the signal

**`setup_dev_mode_server`** pinned **both** `--port` and `--api_port`, reserved them at import time, then spawned a minute later. By then the app port was often gone, and the suite passed only because jac quietly un-pinned it. It now pins only the API port, which is the one all 19 request sites actually use.

**`server startup - port fallback integration`** asserted the old contract directly, that a second server handed a busy explicit port relocates upward. Same scenario, expectation inverted, not a deleted test.

That is the migration cost, stated plainly: code that pins a port it does not care about, and relies on jac to un-pin it, has to stop doing that.

## Docs

The production/local tutorial taught the old behaviour with a transcript showing `--port 3000` relocating to 3001. It now shows the refusal, the `jac.toml` case, and keeps the relocation example under the unset default so both halves of the contract are visible.

Three further defects in that same section, all introduced by this branch and all fixed:

- it referenced `jac start` and `jac dev`, removed in #7255 and absorbed into `jac run` by #8596, which this branch has merged. A leftover from before the rebase.
- it spelled the flag `--api_port`. The parser accepts only `--api-port`; `jac run --serve --api_port N` fails with `unrecognized arguments`.
- it promised "naming the port is what makes it a contract, not where you named it" without the microservice carve-out the release note already admits. The carve-out is now stated where someone will hit it: a `[scale.microservices] gateway_port` in `jac.toml` does not pin, because the loaded config supplies a default for that key; `--port` pins the gateway as usual.

## Not covered

- The refusal in dev mode still surfaces after the API server has started, so it arrives late rather than up front. The wording is now clean on every path (see the refusal section above); moving the app-port check ahead of startup is the remaining half, and I have kept it out to hold this PR to one idea.
- Every `jac` command prints its error twice: `cli/impl/dispatch.impl.jac:247` prints `result.error` after the executor has already printed it. Visible on the port refusal, but not specific to it and not introduced here. Worth its own issue rather than a rider on this one.
- `desktop`, `mobile` and `web-static` client targets are unverified for the port contract.
- `runtimelib/serving/server.jac:30` still declares its own `ERRNO_ADDR_IN_USE` tuple. Different module and a genuinely different use, catching a bind rather than walking candidates, so three copies became two rather than one.
- The second defect in #8387, serving with `--dev` not reaping its Vite child on SIGTERM, is deliberately out of scope. Independent fix, and its only honest test needs a real dev server. The platform-side half is jaseci-labs/jacBuilder#1447.


jac/jaclang/cli/commands/impl/execution.impl.jac
```diff
@@ -625,7 +626,9 @@ impl start(
     }
 
     ServerClass = Jac.get_api_server_class();
-    server = ServerClass(module_name=mod, port=actual_api_port, base_path=base);
+    server = ServerClass(
+        module_name=mod, port=actual_api_port, base_path=base, strict_port=strict_port
```
MusabMahmoodh: This makes the wrong port strict in dev mode. At :620-622, `if dev and not no_client { vite_port = port; actual_api_port = api_port if api_port > 0 else port + 1; }` - so `-p 8000` pins Vite to 8000 and *derives* the API port as 8001. This line then hands `strict_port` to the server that binds 8001.

So `jac start -p 8000 --strict-port` hard-fails when 8001 is busy, with 8000 free and 8000 the only port the user pinned. That is the inverse of #8387: refusing loudly over a port nobody asked for.

I think strict belongs on the `_find_free_port(vite_port, ...)` call at :714 in dev mode, and on `_find_available_port` only when the API port came from an explicit `--api-port` - or from `-p` on the non-dev branch at :624, where `actual_api_port = port` really is what the user named.

jac/jaclang/cli/commands/impl/execution.impl.jac
```diff
@@ -900,7 +905,9 @@ def _find_free_port(start: int, skip_port: int = 0) -> int {
                 }
                 return candidate;
             } except OSError {
-                ;
+                if strict {
+                    raise OSError(f"Port {start} is already in use") from None;
```
MusabMahmoodh: This is under a bare `except OSError`, so the new raise reports "Port N is already in use" for EACCES on a privileged port, a bad bind address, or anything else bind can fail with - and now promotes it to a hard stop instead of a walk to the next candidate.

The sibling already solved this: `_find_available_port` filters on `ERRNO_ADDR_IN_USE_LINUX`/`ERRNO_ADDR_IN_USE_WINDOWS` and re-raises everything else (`serve.core.impl.jac:20-30`). Worth reusing that condition here rather than widening the bare catch.

MusabMahmoodh (changes requested):
Agreed on the premise - a pinned port is a contract and relocating off it is wrong for anything in front of a proxy. Leaving the default alone is the right call too. Two things block it for me, both in the CLI half, plus a reuse point and a test gap.

**The same strict flag is now bolted onto two copies of one loop.** `_find_available_port` (`scale/server/impl/serve.core.impl.jac`) and `_find_free_port` (`cli/commands/impl/execution.impl.jac`) are the same hand-rolled bind-and-walk retry, and this PR adds the same `strict` parameter and the same `f"Port {x} is already in use"` string to both. The inline comment on line 908 is that duplicate's cost made concrete: the two copies already disagree about which errors count. One helper carrying `skip_port` and `strict` would delete the divergence instead of doubling it.

**Tests cover the half the issue title is not about.** The new test exercises `_find_available_port`. Nothing covers `_find_free_port` with `strict=True` - the app/Vite port, which in dev mode is the one `-p` actually pins - and nothing covers the CLI flag reaching `JacAPIServer.strict_port`, which is the part most likely to rot. A real-CLI e2e (`jac start -p <held> --strict-port` exits non-zero and names the port) would cover both, and the cluster-free deploy group already runs that shape.

**One correction to #8387 worth folding in:** on the API path the relocation is not silent - `serve.core.impl.jac:296-300` prints "Port X is in use, using port Y instead". The Vite path is the silent one; `_find_free_port` warns only after 50 failed candidates. Which of the two was the 8000 -> 8002 preview incident on?

And the design question underneath, which I would not block on: if an explicit `-p` is a contract, the default still breaks that contract for everyone who has not heard of the flag. A sentinel default (`-p` unset vs `-p` passed) would make an explicitly passed port strict on its own and leave only the unspecified default relocating - one flag fewer, and it matches your issue title rather than adding a second thing to know about. Keeping the default unchanged is the safer call for Jason though, so your call.

jac/jaclang/cli/commands/impl/execution.impl.jac
```diff
@@ -616,16 +616,25 @@ impl start(
         mach.mem.unit_exit();
     } except Exception { }
 
+    port_pinned = port > 0;
```
MusabMahmoodh, line 784: `jac dev` was not converted to the sentinel, so this line reads True for every `jac dev` invocation.

`execution.jac:133` still declares the dev `port` Arg with `default=8000`, `execution.jac:154` and `execution.impl.jac:492` both declare `port: int = 8000`, and `:516` passes it straight through to `start`. So `port_pinned = 8000 > 0` is True even when the user gave no `-p` at all.

Net effect: **`jac dev` with no port argument now exits 1 when 8000 is busy, instead of relocating.** That is case B from your own table, broken for the command developers run most, and two `jac dev` sessions at once is the ordinary case rather than an edge one.

Three defaults and a help string:
- `execution.jac:133` and `manifest.impl.jac:748`: `default=8000` -> `default=0`
- `execution.jac:154` and `execution.impl.jac:492`: `port: int = 8000` -> `port: int = 0`
- give the dev Arg the same help wording `start` now carries, so the contract reads the same on both commands

jac/jaclang/scale/server/serve.jac
```diff
@@ -108,12 +108,14 @@ glob _jwt_config = get_scale_config().get_jwt_config(),
...
-     logger = logging.getLogger(__name__),
-     PORT_RETRY_LIMIT: int = 10,
-     ERRNO_ADDR_IN_USE_LINUX: int = 98,
-     ERRNO_ADDR_IN_USE_WINDOWS: int = 10048;
+     logger = logging.getLogger(__name__);
 
-def _find_available_port(start_port: int, max_retries: int = PORT_RETRY_LIMIT) -> int;
+import from jaclang.runtimelib.net {
+    ERRNO_ADDR_IN_USE_LINUX,
+    ERRNO_ADDR_IN_USE_WINDOWS,
```
MusabMahmoodh: These two are imported and never used - `find_available_port` and `PORT_RETRY_LIMIT` below them are, but the errno names are not referenced anywhere in `serve.jac` or `serve.core.impl.jac`, and `test_serve.jac:1839-1844` imports them from `jaclang.runtimelib.net` directly rather than through here. Leftovers from the glob block that moved. Safe to drop both lines.

jac/jaclang/scale/tests/server/test_serve.jac
```diff
@@ -1945,7 +1982,7 @@ test "port constants - values are correct" {
     assert ERRNO_ADDR_IN_USE_WINDOWS == 10048 , "ERRNO_ADDR_IN_USE_WINDOWS should be 10048";
```
MusabMahmoodh, line 2040: This pins LINUX and WINDOWS but not `ERRNO_ADDR_IN_USE_MACOS`, which is the constant whose absence caused the macOS relocation bug this PR fixes. Worth adding the third line here - it is the one of the three that has actually gone wrong.

MusabMahmoodh (changes requested):
All five are addressed, and 5 went further than I asked - deleting the flag rather than re-polarising it is the better call and the supervisor argument is the one that makes it obvious.

The errno find is yours, not mine, and it is the better half of the reuse point. My comment said the two copies disagreed about *handling*; you found they disagreed about *which errno*, and that the scale copy never relocated on macOS at all because 48 was missing. That is a real bug fixed as a side effect, and it is worth calling out in the PR body under its own heading rather than leaving it inside the consolidation paragraph, because it is the kind of thing a reader skims past.

I checked the things a merge of two helpers usually breaks and they are all preserved: the Vite wrapper keeps `max_retries=50` and the warn-and-fall-back policy explicitly, and the scale path keeps its 10, which is what it had before.

**One regression, and it is the reason I am not approving yet: `jac dev` was not converted to the sentinel.** Details in the inline comment - it is three defaults and a help string.

Two small ones, neither blocking:

- `net.jac` - when `start_port == skip_port` the candidate advances and can return a different port even under `strict=True`. Only reachable via `-p N --api_port N`, so an edge, but strict is meant to mean never move. A `raise` when the pinned port is the skipped one would close it.
- `runtimelib/serving/server.jac:30` still declares `ERRNO_ADDR_IN_USE: tuple = (98, 48, 10048)` on its own, so three copies became two rather than one. Different module and a genuinely different use (catching a bind on the async server, not walking candidates), so I am not asking for it here - just noting where the last one lives.

jac/jaclang/scale/server/impl/serve.core.impl.jac
```diff
@@ -288,7 +265,7 @@ impl JacAPIServerCore.start(
     self._setup_events();
 
     original_port = self.port;
-    self.port = _find_available_port(self.port);
+    self.port = find_available_port(self.port, strict=self.strict_port);
```
MusabMahmoodh, line 268: This strict check is a probe, but the socket that actually serves is bound at line 281: `run_server` builds a `ServingServer` and calls `run(port_retries=max_retries)` with `max_retries` defaulting to 10, and `ServingServer._serve` walks `chosen += 1` on EADDRINUSE with no strict mode. `on_ready(self.port)` at line 279 has also already announced the probed port by then. So in the restart race from #8387 (port free at probe time, taken again at bind time) the banner and Vite still get N while the app binds N+1. I think strictness has to reach the bind:

```
self.server.run_server(
    port=self.port,
    max_retries=0 if self.strict_port else 10
);
```

With `port_retries=0` the existing loop in `_serve` already raises instead of walking, and the probe becomes a fast-fail nicety rather than the enforcement.

jac/jaclang/cli/commands/execution.jac
```diff
@@ -176,8 +180,12 @@ def enter(
...
-            `default=8000,
-            help="Server port",
+            `default=0,
+            help=(
+                "Server port (default 8000). A port given here must be free: "
+                "it is a contract with whatever sits in front of it, so the "
+                "server fails rather than moving to another port."
+            ),
             short="p",
             config_key="serve.port"
```
MusabMahmoodh, line 100: Does a port set in jac.toml count as pinned on purpose? `_apply_config_defaults` writes an explicit `[serve] port` into the args whenever the CLI value equals the default, which is now 0, so a config-supplied port makes `port_pinned` true exactly like `-p` (for `jac dev` too, since it shares this config_key). I can argue that is right, a port written in jac.toml is also a contract, but it changes behavior for existing projects that set `serve.port`, and the release note only mentions `--port`/`--api_port`. If intended, a line in the note; if not, the pinned check needs to see the pre-config value.

MusabMahmoodh (commented):
Two more with no diff line to hang them on, and one good thing.

`jac/jaclang/runtimelib/impl/server.impl.jac:2438`: the base `JacAPIServer.start` never reads the `strict_port` it now accepts. It builds `ServingServer(app=..., port=self.port)` with default retries, so the port walks on EADDRINUSE regardless. This class is live, not just a base: `get_api_server_class` falls back to it when the scale server module fails to import, which has happened in production before (the lazy-requests fix in the release notes). On that path `jac start -p 8000` accepts the contract and then relocates anyway. Same fix shape as the scale path: thread `strict_port` into the `ServingServer` retries here too.

`jac/jaclang/cli/docs/tutorials/production/local.md:100`: the tutorial still says an in-use `--port 3000` relocates to 3001, transcript included. After this change that section teaches exactly what an explicit `--port` no longer does; it should flip to showing the refusal, since it is the first page someone will land on after hitting the new error.

The rewritten e2e asserting the first server is undisturbed after the second one is refused is exactly the right invariant for this bug.

jac/jaclang/scale/server/serve.jac
MusabMahmoodh: One more in this import: PORT_RETRY_LIMIT is also unused here now. Its only consumer was the deleted _find_available_port, and the test imports it from runtimelib.net directly, so only find_available_port needs importing. I had that backwards above.

MalithaPrabhashana: Dropped both errno names, they had no references left here.

`PORT_RETRY_LIMIT` I kept, and it is worth flagging rather than leaving you to spot it in the diff, because your other comment changed its status. Carrying strictness into the bind means the non-strict path needs an explicit retry budget to hand `run_server`, so `serve.core.impl.jac:282` now reads:

```jac
port=self.port, max_retries=0 if self.strict_port else PORT_RETRY_LIMIT
```

It went from unused back to used between your two comments, so dropping it now would break the build.

jac/jaclang/cli/commands/execution.jac
MalithaPrabhashana: Intended, and now documented rather than left implicit.

I measured it rather than reading it off `_apply_config_defaults`, since the two paths are indistinguishable by the time `start` sees the value. With `port = 54444` in `jac.toml` and nothing typed on the command line:

```
server came up on 54444          healthz = 200      config is honoured
54444 held, still no -p typed    exit=1  "Port 54444 is already in use"
```

So yes, config-supplied ports are strict, on `jac dev` as well since it shares the key.

I went back and forth on this. The argument for carving config out is that two `jac dev` sessions in a project pinning 3000 now fail, which is the same shape as the `jac dev` regression you blocked this on. What decided it the other way is that the carve-out contradicts the rule the PR is built on: the port being named is the contract, and a port written in `jac.toml` is named. A project that pins 3000 in config usually has something addressing 3000, an OAuth callback or a proxy or a hardcoded URL, and moving it to 3001 is the failure this change exists to prevent. Making the rule "named, unless you named it in a file" is harder to explain than the behaviour it replaces.

The note now says so explicitly, including that it governs `jac start` and `jac dev` alike, and that `--api_port` has no config key and stays command-line only.

MusabMahmoodh (approved):
Clearing my changes-requested. The `jac dev` regression I flagged in round 2 is fixed everywhere it lived: `default=8000` is now `default=0` at both Arg declarations, and `dev`'s signature and impl both carry `port: int = 0`, so `port_pinned` is false for a bare `jac dev` and it relocates again instead of exiting 1. The unused `ERRNO_ADDR_IN_USE_LINUX` / `_WINDOWS` imports in `scale/server/serve.jac` are gone - it now imports only `PORT_RETRY_LIMIT` and `find_available_port` - and `test_serve.jac` pins `ERRNO_ADDR_IN_USE_MACOS` alongside the other two, which was the one most worth pinning since its absence is what kept the scale copy from ever relocating on macOS.

Two notes carried forward, neither blocking and neither new to this round:

- `net.jac`: when `start_port == skip_port` the candidate advances and can return a different port even under `strict=True`. Only reachable via `-p N --api_port N`, so it is an edge, but strict is meant to mean never move.
- `runtimelib/serving/server.jac` still declares its own `ERRNO_ADDR_IN_USE = (98, 48, 10048)`. Three copies became two, and this one is a different use (catching a bind on the async server rather than walking candidates), so it is a note rather than an ask.

Good PR. You went further than my comment asked on the flag polarity, and the third defect you turned up on the way - the scale copy missing errno 48 - was worth more than the original report.

kugesan1105 (changes requested):
Validated this end to end rather than by reading: real scaffolded projects, real
servers, a real listener squatting the port. I reproduced the issue on current main
first, then applied this branch and re-ran the identical cases. Both halves of the
patch apply cleanly to `c97ddb0b2`, and the rebase onto "run absorbs start and dev"
looks right.

The diagnosis here is good, and two of its three insights are ones I would not have
found from the issue alone: that strictness has to reach the bind and not just the
probe, and that the two copies disagreed about errno 48. I confirmed both still hold
on main. `scale/server/serve.jac:113-114` still knows only 98 and 10048, so that
walker still never relocates on macOS.

Where it works, it works. `--kind service`, `jac run --port 8765 --no-client` with
8765 held:

```
main:      relocates to 8766, "Server ready", exit 0
this PR:   exit 1,  ✖ Error: Port 8765 is already in use
```

## The gap

It does not fix the case the issue was filed about.

`serve()` dispatches in three stages (`cli/commands/impl/execution.impl.jac:707-719`):

1. `run_client_start(...)`
2. `scale_local_serve(filename, port=port)`, which **returns for any project carrying
   `[scale.microservices.routes]`, so `serve()` returns immediately**
3. everything after, which is where `port_pinned` and `strict_port` live

`jac create --kind web-app` writes `[scale.microservices.routes]` into `jac.toml` by
default. That project therefore takes stage 2: `scale/plugin.jac:57` to the
orchestrator to `ProcessGateway` to `MicroserviceGateway.start`, which calls
`ServingServer(...).run()` at `microservice_gateway.impl.jac:848-853` with the default
`port_retries=10`. None of the files this PR touches are on that path.

Measured, same command, 9100 held by an unrelated listener:

| project | main | this PR |
|---|---|---|
| `--kind service`, `--port 8765 --no-client` | relocates to 8766, exit 0 | exit 1, names the port |
| `--kind web-app`, `--port 9100` | binds 9101 silently, exit 0 | binds 9101 silently, exit 0 |

I think I see how this was missed, and it is an easy one to miss. The dev+client
verification in the description used `jac start test_api.jac --port 52555 --dev`, a
bare `.jac` file. A bare file has no routes table, so it takes stage 3 and the fix does
apply there. The production incident was a `jac create --kind web-app` preview, which
does not. The command you type looks identical in both cases.

## Two more defects on that same path

**The banner still reports the port that was asked for.** With this branch applied and
9100 held:

```
-- actually bound --
  LISTEN 0 100 0.0.0.0:9101  users:(("jac",pid=178139,fd=8))
-- what is printed --
    Gateway: http://0.0.0.0:9100
    Gateway starting on 0.0.0.0:9100
```

`9101` appears nowhere in the output. That is item 2 of the issue, unchanged. The
information already exists: `ServingServer` reads `sockets[0].getsockname()[1]` into
`bound_port` at `runtimelib/serving/server.jac:102-106` and offers it through
`on_ready`. `MicroserviceGateway.start` passes no `on_ready`, and the orchestrator
prints its banner from `gateway_port` before `gateway.serve()` is called.

**An explicit `--port 8000` is discarded, not merely left unpinned.**
`scale/plugin.jac:61-63` gates on `if port and port != 8000`. With
`[scale.microservices] gateway_port = 9300` in `jac.toml`, `jac run --port 8000` binds
9300. The typed flag is thrown away. That is the same "8000 is a magic default rather
than a sentinel" shape this PR set out to remove from `serve()`, living in a file the
PR does not touch.

## Why I am asking for changes rather than merging and filing a follow-up

The release note promises "a named port is now a contract". For the most common project
kind that would ship untrue. Someone reads the note, pins a port in production, and gets
exactly the outage this issue describes, except now with a documented guarantee behind
it. That is worse than the status quo, so I would rather this land complete.

## The ask

Narrow, and it is applying the fix already written here to the second path:

1. Give `MicroserviceGateway.start` a retry budget so a pinned gateway port refuses
   instead of walking, the same way `JacAPIServerCore.start` now does.
2. Pass an `on_ready` through so the gateway announces the port it actually bound, and
   have the orchestrator banner print that rather than the requested value.
3. Replace `scale_local_serve`'s `port != 8000` test with the same sentinel `serve()`
   now uses, so a typed `--port 8000` is honoured and pinned.

Happy to re-run the same cases against the next push.

## Not covered by this review

I could not reproduce the SIGTERM orphan half. The code gap is real and still in the
tree, `except KeyboardInterrupt` only at `execution.impl.jac:892-901` and
`web_target.impl.jac:59,96`, but the current web-app scaffold spawns no Vite process at
all in my runs, only the gateway plus per-service `jac run --serve` children, and the
gateway installs a real signal drain. Confirming that one needs a project that actually
starts Vite, which I did not build. Agreed it is out of scope here either way.

MusabMahmoodh (changes requested):
Re-reviewed at the current head. The code is right and 09ed0ecc is the commit that matters - finding that `serve()` dispatches a routed project to the gateway and returns before the plain path's port handling ever runs, and that `jac create --kind web-app` writes exactly that routes table, means the PR was pinning the wrong path until then. That is the discovery, not the fix, and you found it by re-running against real projects rather than re-reading the diff. Also good: routing `PortInUseError` through `humanize_error` so the gateway path stops printing a raw traceback ahead of the line that matters.

I am marking changes rather than approving, and it is entirely about two of the three new tests in `test_gateway_port_contract.jac`. The first one is solid. The other two cannot fail.

**1. `"an unnamed port still relocates rather than failing"` names a port, and its assertion is unreachable.**

```jac
(code, output) = _serve(["--port", str(free)], timeout=60);
assert code != 0 or "is already in use" not in output;
```

Two problems on top of each other. The test name and comment say "not naming one must keep the old forgiving behaviour", but the body passes `--port`, so the case the comment describes is never exercised. And on the success path a served gateway does not exit - `_serve` hits `TimeoutExpired` and returns `code = None`, so `code != 0` is `True` and the assertion short-circuits before it looks at the output. It also short-circuits on every failure path where the process exits non-zero. The only way this test fails is a clean `exit 0` that also printed "is already in use", which this code cannot produce.

Assert the outcome instead - it bound the port it was given and stayed up:

```jac
(code, output) = _serve(["--port", str(free)], timeout=60);
assert code is None , (
    f"a free named port must be bound and served, not exited: {output[-300:]!r}"
);
assert str(free) in output , (
    f"the banner must name the port it bound: {output[-300:]!r}"
);
```

And if you want the unnamed case the comment describes, that is a fourth test with no `--port` at all.

**2. `"a gateway announces the port it bound, not the one it wanted"` is a negative against a port it was never going to pick.**

`taken` comes from `bind(("0.0.0.0", 0))`, so it is a random ephemeral port. The run passes no `--port`, so the gateway starts from its configured default. Asserting `f"Gateway: http://0.0.0.0:{taken}" not in output` is therefore true almost by construction - the two numbers were never going to coincide, and the holder socket is doing no work.

Item 2 of #8387 is a positive claim: the announced address is reachable. Pin it that way - scrape the port out of the banner and connect to it:

```jac
import re;
(_code, output) = _serve([], timeout=240);
m = re.search(r"Gateway: http://[\d.]+:(\d+)", output);
assert m , f"the gateway announced no address: {output[-400:]!r}";
probe = socket.socket();
try {
    assert probe.connect_ex(("127.0.0.1", int(m.group(1)))) == 0 , (
        f"the announced port {m.group(1)} is not listening, so the banner "
        f"reported a request rather than a bind: {output[-400:]!r}"
    );
} finally {
    probe.close();
}
```

That fails on `main` for the reason the issue was filed and passes here, which is the property the PR needs to hold.

Unrelated but worth knowing before you rebase again: Jason's #8711 also edits `microservice_gateway.impl.jac` inside `MicroserviceGateway.setup` (it changes the `/docs` branch to serve the SPA). It is a different hunk from your port and ready-callback changes, but the same method, and it is likely to land first.

kugesan1105 (approved):

PR #8424, opened 2026-08-20
fix(scale): render KEDA TriggerAuthentication manifests into --show-yaml

### LINKED ISSUE #8409: Jac Scale: the dry-run autoscaler render omits KEDA TriggerAuthentication manifests, so --show-yaml understates what apply() will create
## Mechanism

`render_autoscaler_manifests` (the dry-run/--show-yaml path) renders one
manifest per autoscaled service by calling the engine's `_build_manifests`.
For KEDA, `_build_manifests` returns only the ScaledObject; the
TriggerAuthentication objects for any trigger with `auth.secret_refs` are
built and applied exclusively inside `KEDAAutoscaler.apply` via
`_apply_trigger_auth`, which the render path never touches. So a config with
authenticated triggers (the exact case where a user most wants to preview
what will land in the cluster, since it references Secrets) dry-runs clean
with N ScaledObjects and then applies N ScaledObjects plus M
TriggerAuthentications the preview never showed. Same shape as #7915
(--show-yaml omitting autoscaler manifests the plan counts), one layer down.

Secondary wart in the same function: it discriminates the two engines' return
shapes with `built["body"] if "body" in built else built` - HPA returns a
bare manifest, KEDA returns `{"name", "body"}` - instead of the engines
sharing one contract, which is how the TriggerAuthentications fell through
the seam in the first place.

## Exact location

- Render path: `jac/jaclang/scale/deploy/target/kubernetes/target.jac:871-887`
- KEDA build returns ScaledObject only:
  `jac/jaclang/scale/deploy/autoscale/keda_autoscaler.impl.jac:108-166`
- TriggerAuthentications built only inside apply:
  `keda_autoscaler.impl.jac:168-228` (`_apply_trigger_auth`), called from
  `:254-255` in `apply`
- HPA's divergent return shape: `jac/jaclang/scale/deploy/autoscale/hpa_autoscaler.impl.jac:42-62`

## Code evidence

```jac
// target.jac:879-884
built = autoscaler._build_manifests(
    self._autoscaler_spec(svc_name, cfg, autoscaler, self.k8s_config.namespace)
);
out.append(built["body"] if "body" in built else built);
```

```jac
// keda_autoscaler.impl.jac:254-257 - auth manifests exist only on the apply path
for (i, trigger) in enumerate(spec.triggers) {
    self._apply_trigger_auth(api, trigger, spec.namespace, app_name, i);
}
manifests = self._build_manifests(spec);
```

## Suggested fix

Give `_build_manifests` one contract across engines: return a list of
manifests. HPA returns `[hpa]`; KEDA returns `[scaled_object,
*trigger_authentications]` (extract the auth-body construction out of
`_apply_trigger_auth` so apply and render share the builder). The render path
then flattens without shape-sniffing, and apply iterates the same list. One
seam, both consumers, nothing to fall through.

## How to validate

Configure a KEDA autoscaler with an auth-bearing trigger and run `jac start --scale --show-yaml`: the ScaledObject renders and the TriggerAuthentication does not, then `apply()` creates it. Same class as #8008, which fixed the autoscaler manifests missing from --show-yaml; this is the next instance in the same renderer.

Size: S. Found by a module-by-module read of the jac-scale Kubernetes deploy
surface; the tracking issue collects the rest. I have a kind + EKS rig set up for
this surface and am happy to run a validation pass on a branch if that helps.

Developer-Linus:
Closes #8409.

## Summary

`jac start --scale --show-yaml` (via `render_autoscaler_manifests`) prints
the autoscaler objects a deploy would apply, but for KEDA it only ever
printed the `ScaledObject`. A trigger with `auth.secret_refs` also causes
`apply()` to create a `TriggerAuthentication`, and that half of the
footprint was invisible to the dry-run preview. This PR unifies the
`_build_manifests` contract across both autoscaler engines so the render
path renders the same objects `apply()` creates, with nothing built as a
side effect of a live API call.

## Problem

`_build_manifests` for KEDA only ever constructed the `ScaledObject` body.
The `TriggerAuthentication` for an authenticated trigger was built and
written to the cluster exclusively inside `KEDAAutoscaler.apply` via
`_apply_trigger_auth`, a function that conflated *building* the manifest
with *reconciling* it against the cluster (get → patch-or-create) in one
step. Because the manifest never existed as a standalone, returnable value,
`render_autoscaler_manifests` (the `--show-yaml` path) had no way to see it.

A secondary wart compounded this: `_build_manifests` returned a different
shape per engine (HPA a bare manifest dict, KEDA a `{"name", "body"}`
wrapper), so the render path added defensive shape-sniffing
(`built["body"] if "body" in built else built`) instead of a real contract.
That sniff is what let the KEDA gap go unnoticed - it happily returned
"a manifest" without anyone checking whether it was the *only* manifest
KEDA needed to emit.

Net effect: a KEDA autoscaler config with an auth-bearing trigger dry-ran
clean (N ScaledObjects, 0 TriggerAuthentications) and then `apply()`
silently created N ScaledObjects **and** M TriggerAuthentications the
preview never showed - exactly the class of surprise `--show-yaml` exists
to prevent, and the same shape of gap #8008 already fixed once for
autoscalers generally.

## Fix

`_build_manifests` now returns one `list[dict]` contract for both engines:

- `HPAAutoscaler._build_manifests` returns `[hpa]`.
- `KEDAAutoscaler._build_manifests` returns
  `[scaled_object, *trigger_authentications]`.

To make that possible without duplicating logic between the render and
apply paths:

- The `TriggerAuthentication` body construction is extracted out of the old
  apply-only `_apply_trigger_auth` into a pure builder,
  `_build_trigger_auth_manifest`, which returns the manifest (or `None`
  when a trigger has no auth) and makes no API calls.
- A shared `_apply_custom_object` reconciler (get → patch-or-create)
  replaces the two nearly-identical get/patch/create blocks that
  previously existed separately for `TriggerAuthentication` and
  `ScaledObject` writes in `KEDAAutoscaler.apply`.
- `render_autoscaler_manifests` (`target.jac`) now does
  `out.extend(autoscaler._build_manifests(...))` instead of shape-sniffing
  a single object out of the return value, so it renders every manifest an
  engine produces, not just the first one.

Apply-time ordering is unchanged: `TriggerAuthentication` objects are still
written before the `ScaledObject` that references them via
`authenticationRef`, since KEDA's admission webhook needs the referenced
auth resource to exist first.

## Changed files

| File | Change |
|---|---|
| `jac/jaclang/scale/deploy/autoscale/keda_autoscaler.jac` | Interface updates: `_build_manifests` returns `list[dict[str, any]]`; new `_build_trigger_auth_manifest` and `_apply_custom_object` declarations replace `_apply_trigger_auth`. |
| `jac/jaclang/scale/deploy/autoscale/keda_autoscaler.impl.jac` | `_build_manifests` assembles `[scaled_object, *trigger_auths]`; trigger-auth body construction moved into pure `_build_trigger_auth_manifest`; new `_apply_custom_object` reconciler; `apply()` applies trigger-auth manifests before the ScaledObject via the shared reconciler. |
| `jac/jaclang/scale/deploy/autoscale/hpa_autoscaler.jac` | `_build_manifests` interface updated to `list[dict[str, any]]`. |
| `jac/jaclang/scale/deploy/autoscale/hpa_autoscaler.impl.jac` | `_build_manifests` wraps its single manifest in a list; `apply()` unwraps `[0]`. |
| `jac/jaclang/scale/deploy/target/kubernetes/target.jac` | `render_autoscaler_manifests` flattens the manifest list from each engine instead of shape-sniffing a single object. |
| `jac/jaclang/scale/tests/deploy/test_factories.jac` | HPA `_build_manifests` assertions updated to the list contract. |
| `jac/jaclang/scale/tests/deploy/test_keda_autoscaler.jac` | KEDA `_build_manifests` assertions updated to the list contract (ScaledObject at index 0, no more `name`/`body` wrapper). |
| `jac/jaclang/scale/tests/deploy/test_show_yaml_autoscalers.jac` | New regression test: an authenticated KEDA trigger renders both the `ScaledObject` and its matching `TriggerAuthentication` through `render_autoscaler_manifests`. |
## Testing

Ran the full `deploy` test group locally with `jac test` (one process per
file, matching how CI's `test-scale (deploy)` matrix leg runs it): all 27
files, 293 tests pass, both before-vs-after on the three files this PR
touches and as a full-group run afterward to catch cross-file fallout.

- `test_show_yaml_autoscalers` (4, 1 new) - the regression test for this
  issue. Builds a bundle with one KEDA-engine service carrying an
  authenticated `redis` trigger (`auth.secret_refs`) and asserts
  `render_autoscaler_manifests` returns exactly two manifests: the
  `ScaledObject` at index 0 and a `TriggerAuthentication` at index 1 whose
  `metadata.name` matches the `authenticationRef.name` KEDA embedded in the
  ScaledObject's trigger, with the `secretTargetRef` mapped correctly from
  config. Confirmed this test fails against the pre-fix code (dropping the
  `_build_trigger_auth_manifest` loop reproduces the original bug -
  `len(manifests) == 1`) and passes with the fix. The other two existing
  tests in this file (HPA render, empty-bundle render) are unchanged and
  still pass, guarding against a regression in the ordinary case.
- `test_keda_autoscaler` (34) - full suite for the engine this PR changes
  most: `_build_manifests` shape and field mapping (9 tests, now indexing
  `[0]` into the list), trigger-authentication naming/collision rules,
  `preflight`, and - the part most likely to break silently on a
  build/apply split - `apply()`'s behavior end to end against a mocked
  `CustomObjectsApi`: that a `TriggerAuthentication` is still created
  before its `ScaledObject`, that the name written to the cluster still
  matches the `authenticationRef` the ScaledObject carries, and that a
  KEDA-CRDs-not-installed cluster still short-circuits before either write.
  All pass unmodified in assertion logic (only the `_build_manifests`
  indexing changed), which is the signal that `apply()`'s observable
  behavior toward the cluster is unchanged by the refactor.
- `test_factories` (32) - HPA `_build_manifests` output (exact-dict
  equality against the expected manifest, and the `behavior_overlay`
  merge) re-verified against the new `[hpa]` list contract.

### Live cluster validation

Also validated end to end against a real cluster, not just mocks:

- **Cluster**: local `kind` (v0.32.0) with KEDA core installed via the
  standard Helm chart (`kedacore/keda`, namespace `keda`); `scaledobjects.keda.sh`
  and `triggerauthentications.keda.sh` CRDs confirmed present. A static
  `hostPath` PV under a `no-provisioner` StorageClass (`jac-rwx`) stands in
  for the bundle PVC's `ReadWriteMany` requirement, the same technique
  `jac/jaclang/scale/scripts/e2e_lib.sh`'s `provision_kind_rwx_storage`
  uses for the kind-based CI e2e leg (kind's `local-path` StorageClass is
  RWO-only; a static hostPath PV is safe on a single-node cluster).
- **Project**: a minimal scratch app (`main.jac`, one walker, no client) with
  `[dev] jaclang_source` pointing at this checkout, so both the CLI and the
  deployed pods run the fixed source, not a published release binary.
  `[scale.kubernetes] autoscaler_engine = "keda"` plus one
  `[[scale.microservices.services.<name>.triggers]]` entry with
  `type = "redis"` and `auth.secret_refs` - the exact shape #8409 described.
- **Dry-run** (`jac start --scale --dry-run --show-yaml`, no cluster
  contact): the rendered YAML stream includes both the `ScaledObject` and a
  `TriggerAuthentication` object, with the `ScaledObject`'s
  `authenticationRef.name` matching the `TriggerAuthentication`'s
  `metadata.name` exactly (`kedaval-7dcd81a6-ta`) - this is the bug fixed,
  observed through the real CLI rather than only through the test suite.
- **Live apply** (`jac start --scale`): the same `TriggerAuthentication`
  (`kedaval-7dcd81a6-ta`, `SECRET: redis-secret`) is present on the cluster
  via `kubectl get triggerauthentications`, and the `ScaledObject`'s
  `AUTHENTICATIONS` column names it - dry-run and `apply()` now agree,
  which is the whole point of the fix. `kedaval-scaledobject` reports
  `READY=False`/`ACTIVE=Unknown` only because no real Redis exists at the
  configured address; that's expected and irrelevant to what's being
  validated here (the auth object's existence and wiring, not live scaling).
- **App health**: both `gateway-deployment` and `kedaval-deployment` reached
  `1/1 Running`, the fleet rollout reported `2/2 deployments serving`, and
  `/health` on the gateway reported both services `"status": "healthy"` -
  confirming the fix doesn't regress the ordinary deploy path.

## How to validate

Configure a KEDA autoscaler with an auth-bearing trigger and run
`jac start --scale --show-yaml`: both the `ScaledObject` and its
`TriggerAuthentication` now render, matching what `apply()` creates.

## Post-review updates (included in the merge)

The sections above describe the branch as submitted; the following landed on it before merge:

- **Main merged in** as it advanced (one conflict, the abstract `_build_manifests` signature, resolved keeping the `list[dict]` contract).
- **`_validate_triggers` moved from `apply()` into `_build_manifests`.** With the render path now emitting TriggerAuthentications, an apply-only check let a config whose triggers resolve to colliding identities render clean YAML where both triggers share one TA holding whichever secret kubectl applied last, while `apply()` refused the same config with a `ValueError`. The auto-key variant (an unnamed trigger colliding with one explicitly named `<type>-<index>`) is invisible to KEDA's admission webhook. Validating in the builder gives render and apply one gate; `apply()` now builds manifests before its engine-switch HPA delete, so an invalid spec still fails before any cluster write. Two new tests in `test_keda_autoscaler.jac` (now 36) pin this, including the webhook-invisible collision, and the release-notes fragment covers it. Details in [this review comment](https://github.com/jaseci-labs/jac/pull/8424#issuecomment-5444116316).
- **CLI note**: after #8596 the dry-run spelling is `jac scale deploy main.jac --dry-run --show-yaml`; the `jac start --scale --show-yaml` commands quoted above predate the run/start consolidation.
- The wider render/apply divergences this PR does not attempt (HPA engine fabricating a cpu trigger where apply raises, silent zero on failed preflight, apply-side deletes never shown as removals, TAs orphaned on trigger rename, ops_rbac missing from the plan printer) are tracked in #8750.


MusabMahmoodh (approved):
Fix is the shape I asked for in #8409, and the show-yaml test is the right one: it goes through `render_autoscaler_manifests` and asserts both manifests plus the `authenticationRef` name agreeing, so it fails on main. Apply ordering (auth before the ScaledObject that references it) is preserved. Approving.

Three things, none blocking.

**1. The unified contract still is not declared anywhere.** `Autoscaler` (`autoscale/autoscaler.jac:34`) carries concrete stubs for `apply`, `destroy` and `destroy_collection` that raise `NotImplementedError`, but nothing for `_build_manifests` - and `target.jac:893` calls it on whatever `AutoscalerFactory.create` hands back. That duck-typed seam is how the two shapes diverged in the first place. A stub in the base is what makes "one contract" true for the next engine rather than for these two:

```jac
def _build_manifests(self: Autoscaler, spec: AutoscalerSpec) -> list[dict[str, any]] {
    raise NotImplementedError(
        "_build_manifests() must be implemented by a concrete Autoscaler subclass"
    );
}
```

**2. `apply()` now depends on position rather than kind.** `scaled_object = manifests[0]` with `manifests[1:]` as auth means a future reordering inside `_build_manifests` would apply a TriggerAuthentication through the `scaledobjects` plural and nothing would catch it. The order is load-bearing, so it is worth pinning, and it costs one line in the pure-function test you already have:

```jac
assert [m["kind"] for m in manifests] == ["ScaledObject", "TriggerAuthentication"];
```

**3. Question on `destroy`.** Now that both objects come out of one builder, should `destroy(autoscaler_name, namespace)` remove the TriggerAuthentications too? Today `keda_autoscaler.impl.jac:292` deletes only the scaledobject, and `destroy_collection` is the sole path that sweeps `triggerauthentications` by label, so a single-service teardown leaves the auth object behind pointing at a Secret. If that split is deliberate, ignore me.

PR #8427, opened 2026-08-20
test(scale): lock the monitoring namespace fallback so the dropped declaration cannot return (#8057)

### LINKED ISSUE #8057: Jac Scale: sanitize_prometheus_namespace has no declaration, so any config that reaches it crashes with NameError instead of a sanitized namespace
## Summary

`get_monitoring_config()` falls back to `sanitize_prometheus_namespace(k8s_namespace)` whenever `[scale.monitoring] namespace` is unset-or-falsy on the loaded dict. The function exists only as an `impl` in config_loader.impl.jac with no matching `def` in config_loader.jac, so at runtime the name is unbound: reaching that branch raises `NameError` instead of returning a sanitized namespace. The branch is normally masked because `get_default_config()` merges `monitoring.namespace = 'jaclang_scale'` over jac.toml, so the users who hit it are exactly the ones who set `namespace = ""` explicitly, and they get a crash in place of the sanitizer that was written for them.

## Provenance

Introduced with the jac-scale fold-in (#7038, d4f74f010); the declaration has been missing the whole time. Verified against upstream/main as of 9d2cacecc.

## Reproduction

Two files, fails in under 2 minutes:

```toml
# jac.toml
[project]
name = "repro"
entry-point = "main.jac"

[scale.monitoring]
namespace = ""
```

```jac
# main.jac
import from jaclang.scale.config.config_loader { get_scale_config }
with entry { print(get_scale_config().get_monitoring_config()); }
```

`jac run main.jac` (jac 0.35.1, source at upstream/main 9d2cacecc):

```
Error: name 'sanitize_prometheus_namespace' is not defined
  299 |         k8s_config = config.get('kubernetes', {});
  300 |         k8s_namespace = k8s_config.get('namespace', 'default');
  301 |         namespace = sanitize_prometheus_namespace(k8s_namespace);
      |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  at get_monitoring_config() .../scale/config/impl/config_loader.impl.jac:301
```

## Code evidence

- `config_loader.impl.jac:266`: `impl sanitize_prometheus_namespace(name: str) -> str` with no declaration anywhere in `config_loader.jac` (grep is empty).
- `config_loader.impl.jac:299-304`: the only call site, guarded by `if not namespace`.
- `get_default_config()` fills `monitoring.namespace = 'jaclang_scale'`, which keeps the branch dead on normal loads and is why no test ever executed it.

## Root cause

Two classes stacked:

1. Dead branch behind a default: the defaults merge makes the fallback unreachable except for explicit falsy overrides, so it shipped unexecuted.
2. Orphan impl accepted silently: the compiler compiles an `impl` with no declaration and drops it, converting a build-time mistake into a runtime NameError.

## Suggested fix

One line in `jac/jaclang/scale/config/config_loader.jac`: declare `def sanitize_prometheus_namespace(name: str) -> str;` next to the other module-level defs. Add a regression test that calls `get_monitoring_config()` with `monitoring.namespace` explicitly empty and asserts the sanitized kubernetes namespace comes back; it NameErrors pre-fix through the real path.

## How to validate

`jac test jac/jaclang/scale/tests/misc/test_config_loader.jac` with the new test: it must fail with the NameError before the declaration is added and pass after. The repro above doubles as a manual check. No cluster or deploy needed.

Size: S. No prerequisites.

## Scope note

Deliberately not included, each worth its own look: whether an orphan impl should be a `jac check` error (compiler lane), and whether an empty-string namespace should instead fail loud as invalid config.

akindu-k:
## **Description**

Fixes #8057.

> **What this PR contains, after review and after merging current `main`.**
> The one-line declaration this PR originally added has since landed on `main` on its own, through `86b0c25da` (#8399, "type checking joins the critical path"), which added the same `def sanitize_prometheus_namespace(name: str) -> str;` while making the repo type check. This branch was carrying a second, duplicate copy of that line, so it was removed in `d284df8`.
>
> **The crash is therefore already fixed on `main`, and the diff here is now two files and 27 added lines: the regression coverage that keeps it fixed, and the release note. Nothing is deleted or modified, and no product code is touched.**
>
> ```
>  jac/jaclang/scale/tests/misc/test_config_loader.jac | 26 ++++++++++++++
>  release_notes/unreleased/jaclang/8427.bugfix.md     |  1 +
>  2 files changed, 27 insertions(+)
> ```

### The bug

`get_monitoring_config()` derives the Prometheus namespace from the Kubernetes namespace whenever `[scale.monitoring] namespace` is unset-or-falsy. `sanitize_prometheus_namespace` existed only as an `impl` with no matching `def` in `config_loader.jac`, so the compiler dropped it and the name was unbound at runtime: anyone who set `namespace = ""` explicitly got a `NameError` instead of the sanitizer written for them. `get_default_config()` merges `monitoring.namespace = 'jaclang_scale'` over `jac.toml`, which is why the branch was never executed by a test and the orphan survived unnoticed.

The blast radius was the whole server, not one config read. `get_monitoring_config()` is called in `scale/server/extensions.jac:143` before `collector.is_enabled()` is reached at `:147`, so `monitoring.enabled = false` does not protect startup. The other callers remain in `runtime/gateway/microservice_gateway.impl.jac` and `admin/impl/workloads.impl.jac`.

### Reproduction

```toml
# jac.toml
[project]
name = "repro"
entry-point = "main.jac"

[scale.kubernetes]
namespace = "jac-shop-prod"

[scale.monitoring]
enabled = false
namespace = ""
```

Before:

```
File ".../scale/server/extensions.jac", line 143, in register
    cfg = get_scale_config().get_monitoring_config();
File ".../scale/config/impl/config_loader.impl.jac", line 265, in get_monitoring_config
    namespace = sanitize_prometheus_namespace(k8s_namespace);
NameError: name 'sanitize_prometheus_namespace' is not defined
```

After: `jac run --serve` starts, `/healthz` returns `200`, and the monitoring config carries `namespace = 'jac_shop_prod'`.

### What this PR contains

- **A regression test that drives the real loader.** It writes the issue's two-file repro into a temp dir and reads it back through `get_scale_config()`, rather than assigning `cfg._config` directly, so the default merge that hid the bug is part of what the test exercises.
- **A unit test for the sanitizer's four branches**, including the `"---"` to `jaclang_scale` fallback. Its import of the helper is local to the test body, so a future removal of the declaration fails this one test instead of collapsing every test in the file into a collection error.
- **The release note for #8057**, worded as a failure to start rather than a failed config read.

### Validation

Re-measured on the merge commit `9e9608fcd`, not carried over from the earlier revision. Line numbers below are current as of that commit.

**Pre-fix failure mode.** Deleting the declaration from `config_loader.jac:47` and running the file gives `20 passed, 2 error`, with the two failures landing separately and neither hiding the other:

```
FAILED ...::get_monitoring_config derives the namespace when monitoring.namespace is empty
  .../config_loader.impl.jac:360 in get_monitoring_config
    namespace = sanitize_prometheus_namespace(k8s_namespace);
E   NameError: name 'sanitize_prometheus_namespace' is not defined

FAILED ...::sanitize_prometheus_namespace produces a valid Prometheus namespace
    import from jaclang.scale.config.config_loader { sanitize_prometheus_namespace }
E   ImportError: cannot import name 'sanitize_prometheus_namespace' from '...'

20 passed, 2 error
```

This is the behaviour asked for in review. The first test reports the real `NameError` at the real call site, the second reports its own local `ImportError`, and the other 20 tests still run. Moving that import to module scope is what previously turned the whole file into a single collection error.

**`jac check jac/jaclang/scale/config/config_loader.jac`.** With the declaration present: passes, **0 errors**, warnings only (`W1036`, `W1037`, `W1051`, `W2074`). Deleting the declaration turns that into **2 errors**, `E1054` and `E1032`, both inside the dropped body, so the orphan was compiler-visible the whole time. The `E1099` reported in review does not reproduce on this branch: the module is clean of errors today.

Error locations are cited by code rather than by line, deliberately. The stale `jac check` claim in the previous revision of this description was stale precisely because line numbers moved under it.

**The gate now looks at this module.** #8671 removed both `config_loader` entries from `.jacignore` (58 entries remain), so CI's `jac check` compiles this file and the orphan-caused errors above would fail it. Between that and the tests here, a second removal of the declaration cannot ship silently.

**Test suite.** `jac test jac/jaclang/scale/tests/misc/test_config_loader.jac`: **22 passed**. That is the union of this branch's 2 new tests and the 3 added on `main` while this PR was open; the merge conflict in that file was both sides appending tests, and both sets were kept.

**End to end.** With `[scale.monitoring] enabled = false` and `namespace = ""`, `jac run --serve` starts and `/healthz` returns `200`. Without the declaration it dies in `__post_init__` at the traceback above.

**CI.** Green on `9e9608fcd`: 23 passed, 0 failed, 5 skipped, and the PR is mergeable.

### Review feedback

| Asked | Where it landed |
| --- | --- |
| Drive the monitoring test through the real loader instead of assigning `cfg._config`, since that assignment skips the default merge that hid the bug | `b1b3fc884` |
| Move the sanitizer unit test's import inside the test body so a dropped declaration fails one test rather than collection | `d5b673436`, failure mode re-measured above |
| Reword the release note as a failure to start, not a failed config read | Release note now reads "no longer stops the server from starting", and says `enabled = false` did not protect you |
| Correct the validation claims, particularly `jac check` | Whole Validation section re-measured on `9e9608fcd` and rewritten above |
| `.jacignore` is what let this ship | Agreed, and already resolved on `main` by #8671 |
| Merge current `main` to retrigger the stuck native shard | Done, CI green on `9e9608fcd` |

### Out of scope

Per the issue's scope note, this does not change whether an orphan `impl` should be a `jac check` error, nor whether an empty-string namespace should be rejected as invalid config.

Two further points raised in review were confirmed and remain out of scope for a fix PR. Both were filed as issues, and both have since been closed as not planned, so they are recorded here rather than presented as scheduled work:

- **A bare top-level orphan `impl` is silent.** `E2009` is emitted only when a partial name path resolves (`decl_impl_match_pass.impl.jac:126-131`), so `impl Foo.unknown_method` errors while `impl orphan_func` yields only `W2001`/`W1051`, which CI's `--nowarn` discards. Removing the module from `.jacignore` does not close the class either: `config_loader` failed its check only through collateral `E1054`/`E1032` inside the dropped body, so an orphan whose body happens to type check stays completely silent. Widening that condition is a compiler change. Filed as #8687, closed as not planned.
- **The sanitizer only runs on the Kubernetes fallback**, so an explicitly configured `[scale.monitoring] namespace` is still passed through unsanitized. It does not crash, but `prometheus_client` and the sanitizer normalize differently: `2024app` becomes `_024app_...` through the client and `_2024app` through the sanitizer, so the metric name depends on which path set the namespace. Filed as #8688, closed as not planned.


MusabMahmoodh (approved):
Right fix for #8057 and the sanitizer test covers all four branches, including the `"---"` to `jaclang_scale` fallback. I swept `scale/**/*.impl.jac` for other impls with no matching `def` in their declaration file and this is the only one, so it is an instance and not a class. Approving.

One thing, non-blocking.

The monitoring test sets `cfg._config` directly, which follows the idiom already in this file ("repeated get_kubernetes_config calls return equal results"), but here it skips the exact mechanism that hid the bug. `get_default_config()` merges `monitoring.namespace = 'jaclang_scale'` over jac.toml, and that merge is why the branch shipped unexecuted for the whole life of #7038. Assigning `_config` means the test would still pass if a future defaults change stopped `namespace = ""` from ever reaching the sanitizer, which is the regression most likely to put this back.

The real-load idiom is in the test right below yours. Something like:

```jac
reset_scale_config();
d = tempfile.mkdtemp(prefix="jac-cfg-mon-");
try {
    (Path(d) / "jac.toml").write_text(
        '[project]\nname = "repro"\n\n[scale.kubernetes]\nnamespace = "jac-shop-prod"\n\n[scale.monitoring]\nnamespace = ""\n'
    );
    cfg = get_scale_config(project_dir=Path(d));
    assert cfg.get_monitoring_config()["namespace"] == "jac_shop_prod";
} finally {
    reset_scale_config();
    shutil.rmtree(d, ignore_errors=True);
}
```

That is the two-file repro from the issue, running through the loader instead of around it. Keep the unit test of the sanitizer as it is.

kugesan1105 (commented):
Validated this against `main` (66947e9cd) by applying the patch locally and running the real stack, not just reading the diff. The fix is correct for the reported symptom, but two of the validation claims in the description do not hold, and the deeper causes are untouched.

## What the fix does do

Confirmed working end to end. With the patch applied, `jac start` binds port 8000 under `[scale.monitoring] namespace = ""`, where on `main` it dies in `__post_init__`. `get_monitoring_config()` returns the derived namespace. The one-line declaration is correct and sufficient for the crash, and every assertion in the new sanitizer unit test matches the impl's actual behavior.

## The regression test does not fail the way the description says

> New test in `test_config_loader.jac` fails pre-fix with exactly the reported `NameError` and passes after.

Reverting only the declaration and running the file:

```
File ".../scale/tests/misc/test_config_loader.jac", line 5, in <module>
ImportError: cannot import name 'sanitize_prometheus_namespace' from 'jaclang.scale.config.config_loader'
1 error in 0.70s
```

Pre-fix the file dies at import time. All 20 tests collapse into a single collection error, neither new test ever runs, and the reported `NameError` never surfaces. The cause is the second test: adding `sanitize_prometheus_namespace` to the module-level import block short-circuits the file before the branch-driving test can execute.

The first test on its own is a genuine guard. Removing the direct import and the sanitizer unit test gives exactly what the description claims:

```
FAILED ...::get_monitoring_config derives the namespace when monitoring.namespace is empty
  .../config_loader.impl.jac:320 in get_monitoring_config
E   NameError: name 'sanitize_prometheus_namespace' is not defined
18 passed, 1 error
```

One focused failure at the real call site, everything else still green. Please import the helper locally inside the sanitizer unit test body rather than at module scope. That restores the accurate failure mode and keeps any future removal of the symbol from taking down all 20 tests with a collection error.

## `jac check` is not clean on the changed file

> `jac check` clean on both changed files (the two `W2074` f-string warnings are pre-existing in the impl file and untouched here)

With the patch applied:

```
✖ Error: error[E1099]: Cannot access attribute "resolve" for type "Path | NoneType"
1 failed in 2.65s
```

Pre-existing and unrelated to this fix (`config/impl/config_loader.impl.jac:411`), so not a regression, but the module is not clean and the description reads as though only warnings remain.

Worth adding the flip side, because it is the strongest evidence in favor of this PR: the declaration also clears two real errors that were direct consequences of the orphan, `E1054` at `config_loader.impl.jac:287` and `E1032` at `:289`. The file goes from 3 errors to 1. That is proof the orphan was compiler-visible the whole time.

## The deeper causes are untouched

1. **The `.jacignore` exclusion is what actually let this ship, and it is neither fixed nor mentioned.** `.jacignore:327-328` lists both `jac/jaclang/scale/config/config_loader.jac` and `config/impl/config_loader.impl.jac`, so CI's `jac check` never opens this module, errors included. That is why the two orphan-caused errors above sat in `main` unnoticed. It also means green CI says nothing about this file: the check that would have caught the bug is configured not to look at it. This is the highest-value follow-up and deserves its own issue.

2. **`E2009` already exists and is simply unreachable here.** `jac0core/passes/impl/decl_impl_match_pass.impl.jac:126-131` emits "Implementation could not be matched to a declaration" only `if len(name_of_links) > 0`, that is, only when a partial name path resolved. A bare top-level orphan resolves nothing and falls through to a silent `return`. Verified with a minimal fixture: `impl Foo.unknown_method` gives a hard `E2009`, while `impl orphan_func` gives only `W2001`/`W1051`. Deferring this is defensible for a fix PR, but it is one condition rather than a new diagnostic, and until it changes the class recurs.

3. **The sanitizer is still applied to the wrong input.** It only ever runs on the Kubernetes fallback. A user writing `[scale.monitoring] namespace = "jac--shop"` bypasses it entirely. Notably the new unit test asserts `sanitize_prometheus_namespace("jac--shop") == "jac_shop"`, so the hazard was clearly in view while the primary path stayed unguarded. It does not crash, prometheus_client normalizes at exposition time, but the two disagree: `2024app` emits `_024app_...` through the client versus `_2024app_...` through the sanitizer.

## Release note understates severity

> reading the monitoring config with the namespace left explicitly empty raised `NameError`

`scale/server/impl/serve.core.impl.jac:95` calls `get_monitoring_config()` unconditionally inside the server's `__post_init__`, before any `enabled` gate. Same unconditional call in `scale/runtime/gateway/microservice_gateway.impl.jac:622` and `scale/admin/impl/workloads.impl.jac:68`. The server did not start at all, and `monitoring.enabled = false` did not protect you. Someone skimming the notes should read "the server failed to start", not "reading config raised an error".

## Summary

Happy to approve once the sanitizer unit test's import moves inside the test body and the release note is reworded. Please also correct the two validation claims in the description, particularly the `jac check` one, since it papers over the module's real state.

MusabMahmoodh (approved):
Re-reviewed at the current head. Removing the duplicate declaration in d284df8 was the right call and the PR body is straight about it - I confirmed `def sanitize_prometheus_namespace(name: str) -> str;` is on `main` at `config_loader.jac:41` via #8399 (86b0c25da, Aug 22), so what is left here is the lock, not the fix. That is still worth landing. Approving.

The local import inside the sanitizer test is a genuinely good detail. A dropped declaration failing one test instead of collapsing the file into a collection error is the difference between a five-second diagnosis and a confusing one.

**One thing that needs coordinating, and it is with your own other PR.** #8691 appends to the same file at the same place, and one of its four new tests is this test:

```jac
test "an empty monitoring namespace still derives from the kubernetes namespace" {
    got = _monitoring_namespace(
        '[scale.kubernetes]\nnamespace = "jac-shop-prod"\n\n[scale.monitoring]\nnamespace = ""\n'
    );
    assert got == "jac_shop_prod";
}
```

Same repro, same assertion, same expected value as `"get_monitoring_config derives the namespace when monitoring.namespace is empty"` here. Whichever of the two merges second will conflict on `test_config_loader.jac` and, once resolved, ship the assertion twice under two names. Worth deciding now which PR owns that case - my read is it belongs here, since #8057 is the issue about that exact branch, and #8691 can drop its fourth test on rebase.

Checked and compatible otherwise: #8691 routes the empty-string case through `resolve_prometheus_namespace`, where `not configured` is true for `""`, so it still lands on the Kubernetes fallback and your assertion holds unchanged.

Minor, take it or leave it: the fix this note describes shipped in 0.37.0 (cut this morning) via #8399, which carried no fragment of its own, so this note will appear under 0.38.0 for a crash that was already gone. That is arguably better than the gap staying unrecorded - just flagging it so the wording is a deliberate choice rather than a surprise at release time.

MusabMahmoodh (approved):
Approving. Both tests hold against `3ad35097c9`: the declaration is at `config_loader.jac:80`, `get_monitoring_config` derives through the sanitizer at `config_loader.impl.jac:265`, and the four sanitizer assertions match the impl at `:227-240` including the `'---'` to `jaclang_scale` fallback. The local import inside the test body does what you say it does, and putting both tests in the file that already owns this surface is the right call.

One thing worth fixing in the description before this merges, since the body outlives the diff: `serve.core.impl.jac:73` no longer exists - #8981 deleted the whole `scale/server/serve*` family on Friday. Your blast-radius claim is still true, at a new address: `scale/server/extensions.jac:143` calls `get_monitoring_config()` and only reaches `collector.is_enabled()` at `:147`, so `enabled = false` still does not protect the boot. The other two callers you named (`microservice_gateway.impl.jac`, `admin/impl/workloads.impl.jac`) are unchanged.

PR #8432, opened 2026-08-20
fix(checker): narrow on `==` and `!=` against a string literal

### LINKED ISSUE #7261: Type checker: string-literal equality/inequality narrowing never fires (missing MultiString case)
## Summary

`x == "literal"` / `x != "literal"` narrowing (`LITERAL_EQ`/`LITERAL_NE` in `TypeEvaluator.classify_condition`) silently never applies when the literal is a **string**, even though the exact same pattern works correctly for `int`, `float`, and `bool` literals. This means code like the following incorrectly reports a type error:

```jac
def pick(v: str | None) -> str {
    r: str = "obj";
    if v == "node" {
        r = v;
    }
    return r;
}
```

`jac check` reports:
```
error[E1001]: Cannot assign str | NoneType to str
```
even though inside `if v == "node"`, `v` can only be the string `"node"` — never `None`.

## Confirmed NOT a caching/environment issue

Verified with a fully cleared `~/.cache/jac` (jir + rt) and a clean recompile — same failure. Also verified the equivalent pattern with `int`/`float`/`bool` literals passes cleanly:

```jac
def case_int(v: int | None) -> int {
    r: int = 0;
    if v == 5 { r = v; }
    return r;
}
```
`int`, `float`, and `bool` versions all pass `jac check`; only the `str` version fails.

## Root cause

In `jac/jaclang/compiler/type_system/type_evaluator.impl/type_evaluator.impl.jac`, `classify_condition`'s literal-comparison branch does:

```jac
if left_key is not None
and isinstance(right_expr, (uni.String, uni.Int, uni.Float, uni.Bool)) {
    ...
}
```

Jac's parser wraps **every** string literal -- even a single, plain `"node"` with no interpolation or concatenation -- in a `uni.MultiString` node (confirmed via debug instrumentation: `right_expr_type=MultiString` for a bare string literal operand). `uni.MultiString` is not in that `isinstance` tuple, so the check is always `False` for string literals, `classify_condition` falls through and returns `None`, and no narrowing predicate is ever produced for a string-literal comparison. `int`/`float`/`bool` literals parse as their own atomic node types with no such wrapper, so they're unaffected.

## Fix

Add `uni.MultiString` to the isinstance check, and derive the literal type via the existing `self.get_type_of_expression(right_expr)` call (which already has correct `MultiString` handling elsewhere in this same file, for the case where all constituent strings are literal).

Verified fix (one line):
```jac
if left_key is not None
and isinstance(
    right_expr, (uni.String, uni.Int, uni.Float, uni.Bool, uni.MultiString)
) {
```

With this change, a full narrowing test matrix (isinstance, hasattr, truthiness, `is`/`is not None`, `==`/`!=`, and-chains, walrus -- each in both ternary and if-statement form) passes cleanly, whereas only the string `==`/`!=` cases failed before the fix.

## Context

Found while investigating jaseci-labs/jaseci#7256 (a related but distinct narrowing gap for `in`/`not in` against a tuple/list of literals, surfaced during review of #6810/#7249). This is a separate root cause -- #7256 is about a missing narrowing rule for membership tests; this one is about an existing, correct narrowing rule (`LITERAL_EQ`/`LITERAL_NE`) that never fires for one literal type due to a parser-representation mismatch.

We'll submit the fix for this in #7249 alongside the #7256 fix.

CFG is fine : 
Image

SandeepaHWP:
Fixes #7261.

## Problem

Comparing a variable with a string literal using `==` or `!=` did not narrow its type. Integer, float and boolean literals already worked.

```jac
def pick(v: str | None) -> str {
    r: str = "obj";
    if v == "node" {
        r = v;  # error: Cannot assign str | NoneType to str
    }
    return r;
}
```

Inside `if v == "node"`, `v` can only be `"node"`, so this error is wrong.

## Cause

The checker decided whether the right side was a literal by checking its node class against a fixed list: `String`, `Int`, `Float` and `Bool`. The parser always wraps a string literal in a `MultiString` node, so a string never matched. A literal in parentheses, such as `("node")` or `(5)`, did not match either.

The checker already had a function that recognizes literals, `_literal_type_from_expr`, and it handles `MultiString`. The `==` code did not use it and kept its own incomplete list.

## Change

The `==` code now uses the existing `_literal_type_from_expr` instead of its own list. A small helper, `_extract_literal_operand_type`, adds only two things around it:

- Parentheses are unwrapped, so `("node")` and `(5)` are recognized.
- Floats keep working as before. There is no `Literal[1.5]` type, so a float literal narrows to `float`.

The following are not treated as literals:

- **Variables and references**, even when their declared type is a literal.
- **`typing.cast(Literal["node"], w)`**: its type is a literal, but the value compared at runtime is `w`.
- **f-strings with interpolation.**
- **`None`, `...` and type names**, which have their own handling or are not values.

This PR only changes how the literal is recognized. Which branch gets narrowed is unchanged and follows #8957: `==` narrows only its true branch and `!=` only its `else` branch.

## Known limitation

Because `!=` does not narrow its true branch, a union of literals is not reduced there. For example, `Literal["a"] | Literal["b"]` stays the same inside `if v != "a"`. This is safe, because the type is wider than it could be but never wrong. The tests record it as an expected error. Fixing it needs literal-aware type exclusion, which will be a separate PR.

## Tests

- **`literal_eq_narrowing`:** the reported case, `!=`, the ternary form, string concatenation, a literal in parentheses, and int, float and bool literals all pass with no errors. The only error is the known limitation above.
- **`literal_ne_does_not_over_narrow`:** 10 unsafe cases, including the `typing.cast` case, still report errors.
- **Checks:** `jac check` on the type evaluator and `jac fmt --check --lintfix` on the changed files pass.

## Not included

- **`in` / `not in` narrowing** is handled in #7262, which builds on this PR.
- **Reversed comparisons** such as `"node" == v` still do not narrow, for any literal type. This limitation existed before this PR.


jac/tests/compiler/passes/main/fixtures/checker/checker_literal_eq_narrowing.jac
```diff
@@ -0,0 +1,98 @@
...
+    }
+    return r;
+}
+
+def ne_str(v: str | None) -> str {
+    r: str = "other";
+    if v != "node" {
+        return r;
+    }
+    return v;
```
kugesan1105, line 20: This only exercises the sound direction of `!=`: the false branch (fall-through means `v == "node"`, positive narrowing to `str`). The unsound direction is the true branch, where `LITERAL_NE` excludes `Literal["node"]` from `str | None` and `exclude_type_from_union` deletes the whole `str` member, leaving `None`. A companion case is needed here asserting that this stays an error:

```jac
def ne_str_true_branch(v: str | None) {
    if v != "node" {
        n: None = v;   # must be E1001; currently type-checks with this PR
    }
}
```

As written, the suite passes while `v` is typed `None` over a live `str` (verified at runtime), and native lowering demotes such functions (E5092, "binding a void-typed value").

kugesan1105 (commented):
Validated this locally by applying the diff and running probes against the patched compiler (dev-mode, in-repo source). Summary: the classifier rewrite is the right approach and the reported repro is genuinely fixed, but this does not fix the deeper defect underneath #7261, and it extends that defect to strings. Details below.

## What checks out

- `_extract_literal_operand_type` recognizing operands by the `Literal` base / evaluated `is_literal_type()` is the correct technique, and it is better than the one-line `MultiString` addition proposed in the issue: parens, implicit concatenation, f-string exclusion, and reference exclusion all behave correctly in my testing.
- The issue's repro goes from 1 error to 0, and the int/float/bool baselines hold.

## The deeper problem this arms

Negative literal narrowing is literal-blind. `exclude_type_from_union` keys union members by `shared.canonical_key()`, and `clone_with_literal` reuses the same `shared`, so excluding `Literal["node"]` deletes the entire `str` member. Before this PR strings were immune only because the predicate never fired. With this PR applied, all of the following type-check cleanly:

```jac
def f(v: str | None) {
    if v != "node" {
        n: None = v;   # accepted; at runtime n holds a live str
    }
}

def g(v: str | None) {
    if v == "node" {
    } else {
        n: None = v;   # same hole on the else branch
    }
}
```

I confirmed at runtime that `n` holds the string `"other"` in both. This is not only a silent type lie: native lowering rejects the result ("Native lowering failed for binding a void-typed value to variable 'n'", E5092), demoting the functions to Python-only. Any real guard shaped like `if status != "done" { ...use status... }` with `status: str | None` now types `status` as `None` inside the branch, so `str` operations there can produce new false errors as well.

The int version (`if v != 5` on `int | None` narrowing to `None`) exists on main today and is untouched here, so this PR is not the origin of the bug, it just widens its reach to the most common guard type.

Empirical note: plain non-union targets are fine. `v: str` with `v != "a"` does not over-narrow (the `same_class -> NeverType` arm never triggers for them), so the defect is confined to union targets.

## What I would ask for

Either in this PR or as an immediate companion:

1. Make negative literal narrowing literal-aware: excluding `Literal[X]` may only drop a member that is itself that same literal, and must be a no-op on a bare-class member. That also fixes the pre-existing int/float/bool unsoundness.
2. A fixture asserting the `!=` true branch does NOT narrow `str | None` to `None`, e.g. that `n: None = v;` inside `if v != "node"` is an error. The current `ne_str` fixture only exercises the sound false branch, so the suite cannot see this.

## Confirmed out-of-scope gaps (agree with the PR's list)

- `"node" == v` (literal on the left) still does not narrow, for strings or ints.
- `x in ("a", "b")` membership (#7256).
- Chained comparisons (`a == v == "node"`), excluded by the single-op guard.
- Positive narrowing yields bare `str`, not `Literal["node"]` (`_type_matches_isinstance` compares class identity only); imprecise but sound.

jac/tests/compiler/passes/main/fixtures/checker/checker_literal_eq_narrowing.jac
SandeepaHWP: Fixed. The true branch now keeps `str | None`, and your exact case is asserted as an error in the new `checker_literal_ne_no_over_narrow.jac` fixture, along with the `int`, `float` and `else`-branch equivalents.

SandeepaHWP:
Two updates to this PR:

- **Cast fix (from Greptile's review):** `v == cast(Literal["node"], w)` no longer narrows `v`. A cast only claims a type; the value compared at runtime is `w`. Literals are now recognized by their syntax, not by their type, and a test covers this case.
- **Simpler helper:** `_extract_literal_operand_type` now reuses the existing `_literal_type_from_expr` and only adds parentheses unwrapping and float support. Behaviour is unchanged and the tests pass.

kugesan1105 (approved):
image

The PR fixes string-literal type narrowing in the checker.

Previously, this failed because string literals are parsed as `MultiString`, while the narrowing logic only checked for `String`/`Int`/`Float`/`Bool` nodes:

```jac
def pick(v: str | None) -> str {
    if v == "node" {
        return v;  // incorrectly: str | None -> str
    }
}
```

The PR reuses the existing `_literal_type_from_expr()` recognizer instead of maintaining a separate literal check. This correctly handles strings, implicit concatenation, parenthesized literals, and negative integers.

It also depends on #8957, which prevents unsafe negative literal narrowing. Testing shows the issue repro is fixed, existing literal tests remain clean, the 4 unsoundness cases still fail as expected, and all CI checks are green.

Nice fix @SandeepaHWP, deleting the duplicate literal check is what makes all four spellings work at once, and holding it behind #8957 was the right order.

Two follow-ups worth filing, neither blocking:

**1. `==` replaces the declared type instead of intersecting it.** When no union member matches, `_narrow_type_to` returns `None` and `_apply_narrowing` falls back to the raw `narrow_type`. That is right for `isinstance` (a runtime proof of type) but not for `==`. So `v: int | None` under `if v == "node"` narrows to `Literal["node"]`, and with a custom `__eq__` a `-> str` function can return a non-str with a clean `jac check`. The `int` form already does this on main, so this PR widens it rather than introducing it. Fix: for `LITERAL_EQ`, no related member should mean `Never`, not the literal.

**2. `v == (None)` is routed to `LITERAL_EQ` instead of `NONE_IS`,** because `_literal_type_from_expr` accepts `Null` and the new helper unwraps parens, while the `None` branch above it does not. Its true branch narrows but its else branch does not, so it is inconsistent with bare `None`. One line: unwrap `AtomUnit` in the `Null` branch, or skip `Null` in `_extract_literal_operand_type`.

PR #8449, opened 2026-08-21
fix(scale): the gateway streams an upload through instead of buffering it whole

### LINKED ISSUE #7908: Jac Scale: gateway proxy and passthrough buffer the entire request body in memory with no size cap
**Area:** runtime core / gateway
**Severity:** high (robustness)

### What happens

Both gateway dispatch paths call `await request.body()`, materialising the entire request into gateway memory before forwarding. There is no size cap at the gateway layer.

### Where

`jac/jaclang/scale/runtime/gateway/microservice_gateway.impl.jac:447` (`handle_proxy`)
`jac/jaclang/scale/runtime/gateway/microservice_gateway.impl.jac:378` (`handle_builtin_passthrough`)

### Evidence

```jac
client_addr: str = getattr(request?.client, "host", "unknown");
fwd_headers["X-Forwarded-For"] = client_addr;

body: bytes = await request.body();
```

### Why it matters

A large upload is held whole in the gateway process. Concurrent uploads multiply that. The webhook handler has a `max_body_bytes` cap, but nothing equivalent guards the gateway itself, so the single ingress for the fleet is the least protected component.

### Suggested fix

Stream the request body through to the upstream rather than buffering, and enforce a configurable gateway-level maximum that returns 413 before reading the body. Note that the passthrough replay behaviour (see the fan-out issue in this tracker) currently *requires* buffering, so the two are best fixed together.

Part of #7900

akindu-k:
Fixes #7908.

## What was wrong

The serving core read the entire request body into memory before it routed anything, so `Request.body` was always a complete copy by the time a handler ran. The gateway is a pure forwarder on its proxy and app-route paths, but it had no way to forward anything other than that finished blob, and the outbound client then wrote a second copy into its send buffer.

Measured on this branch's parent, a single 90 MiB `POST` through `MicroserviceGateway` to a registered service:

```
upload            : 90 MiB
gateway peak rss  : 208.2 MiB -> 395.0 MiB   (+186.7 MiB)
```

That is per replica, per in-flight request, on the one process that fronts the whole fleet.

## The fix

The body is handed to the app as an async chunk stream instead of a pre-read blob.

- `Request` carries a `stream` alongside `body`. `load_body()` drains it once; `body_source()` returns whichever of the two is live.
- `ServingServer` validates `Content-Length` at header time exactly as before (400 on a malformed value, 413 over the cap), then attaches a generator that reads the socket on demand and enforces the cap incrementally for chunked bodies.
- `ServingApp._route_and_call` materialises the body before it resolves the route, so every endpoint, multipart parse and form parse sees `request.body` unchanged.
- `MicroserviceGateway.forward_to` passes the stream through to `stream_forward`, and the serving client learned to write an async-iterable body: it owns request framing now, preserving a declared `Content-Length` and falling back to `Transfer-Encoding: chunked` when the length is unknown.

Same reproduction, same 90 MiB upload, on this branch:

```
upload            : 90 MiB
gateway peak rss  : 262.5 MiB -> 262.5 MiB   (+0.0 MiB)
```

The upstream service received all 94371840 bytes in both runs.

image

## Two things deliberately left as they were

**Builtin passthrough still buffers.** `handle_builtin_passthrough` replays one body across services until one stops returning 404, so it calls `load_body()` and keeps the bytes. Streaming it would break the fan-out, which is #7901's subject; a test pins the replay behaviour so it does not regress underneath that work.

**Connections whose body nobody read are closed.** `_should_keep_alive` returns false while a stream is still pending. Bodiless requests never build a stream, so ordinary keep-alive traffic is untouched; a middleware that short-circuits a bodied request (rate limit, drain) now ends the connection rather than leaving unread bytes framed on the socket. `dispatch` also re-raises `ConnectionError`/`IncompleteReadError` so a client that hangs up mid-upload is absorbed by the connection handler as it always was, instead of surfacing as a logged 500.

## Tests

- `jac/tests/runtimelib/test_serve_request_stream.jac` covers the serving core: a middleware observing an unread stream while the endpoint still receives the bytes, chunked reassembly, 413 on an oversized `Content-Length`, 413 mid-stream when a chunked body outgrows the cap, and keep-alive across back-to-back bodied requests.
- `jac/jaclang/scale/tests/microservices/test_gateway_body_stream.jac` covers the gateway: streaming a declared-length body, chunk-framing an undeclared one, and passthrough still replaying past a service that declines.


MusabMahmoodh (commented):
This is the fix #7908 asked for, and it also settles the interaction I flagged in the issue: `handle_builtin_passthrough` keeps buffering because it replays the same body across services, while `forward_to` hands the stream straight through. Splitting `load_body()` from `body_source()` is the right seam for that.

The tests are the best I have read on a scale PR in a while: real `ServingServer` on an ephemeral port, a raw socket writing actual chunked frames, and not a single mock. `an oversized Content-Length is refused before the body is read` and `a chunked body that outgrows the cap is refused mid-stream` are the two cases I would have asked for.

Three things.

**1. An over-limit chunked upload is now partially forwarded before it is refused.** For `Transfer-Encoding: chunked`, `_declared_body_length` returns `CHUNKED_LENGTH`, so the upfront `declared > self.max_body_bytes` check cannot fire and the cap moves inside the generator. On the gateway path that generator is being drained by `_pump_body`, so by the time `HTTPError(413)` is raised the gateway has already written headers and up to `max_body_bytes` to the upstream service and then closes the socket under it with no terminating `0\r\n\r\n`. The client still gets a clean 413 through `ServingApp.dispatch`, so the caller-facing behavior is fine, but before this PR the gateway refused that request without ever opening the upstream connection. A service that streams uploads to disk will now have a partial object for every over-limit request. Is that the trade you want, and can it go in the release note either way? Right now the note only mentions the memory win.

**2. A client that under-delivers a declared Content-Length gets blamed on the service.** With `Content-Length: N` and fewer than N bytes on the wire, `readexactly` raises `IncompleteReadError`, which `arequest` converts to `ClientConnectionError`, which `stream_forward` catches and turns into `None`, which the gateway renders as a 502. So a truncated client upload shows up in fleet metrics as an upstream failure, and the service it names never saw a problem. Pre-fix that read failed on the gateway's own socket before any forwarding. Worth distinguishing the client-side read error from an upstream connection error so the 502 rate stays meaningful.

**3. CI is red and it is your two changed surfaces.** `jac-check` fails on `Run jac check (using .jacignore)` and `test-scale (microservices)` fails on `Run microservices tests`. I could not read the logs from here, so I have no diagnosis, but happy to look again once it is green - the reading above is a static one and does not depend on the run.

PR #8464, opened 2026-08-21
feat(scale): expose autoscaler runtime status and a shared scale-transition observer

### LINKED ISSUE #7405: feat(jac-scale): expose autoscaler runtime status and scale transition observation
## Summary

Extend the autoscaler abstraction with normalized runtime status and scalable transition observation.

Today `Autoscaler` exposes `apply`, `destroy`, `destroy_collection`, `preflight`, and resource naming. Applications that dynamically manage scale-to-zero workloads must bypass the abstraction and inspect `ScaledObject`, HPA, Deployment, and pod status directly to answer basic questions:

- Is the workload inactive, activating, ready, deactivating, or degraded?
- Did the scaler fail to fetch metrics?
- What replica count is desired/current/ready?
- When did this observer see a scale transition?
- When is a cold-started workload ready to receive traffic?

This is needed by control-plane applications such as JacHammer, but should be implemented as a generic Jac Scale capability.

## Scale and durability semantics

The target is a scalable Kubernetes workload, normally a Deployment. A zero-replica target is a healthy inactive state; individual pods are disposable sessions and are not the durable identity.

Kubernetes watches are not durable history. Transition events must carry the time they were observed and the source `resourceVersion`, but callers remain responsible for persisting business timestamps such as `last_suspend_at`. After an observer restart, the current state can be reconstructed, but the exact time of a transition missed during downtime must not be invented.

## Proposed API

The exact shape is open for design, but it should provide:

1. A point-in-time normalized status query.
2. A shared namespace-level observer that multiplexes transitions to many subscribers.

```jac
obj AutoscalerRuntimeStatus {
    has state: str,  # inactive | activating | active | deactivating | degraded | unknown
        desired_replicas: int,
        current_replicas: int,
        ready_replicas: int,
        trigger_active: (bool | None),
        scaler_ready: (bool | None),
        conditions: list[dict[str, any]] = [],
        resource_version: (str | None) = None,
        message: (str | None) = None;
}

obj AutoscalerTransition {
    has autoscaler_name: str,
        scale_target_name: str,
        namespace: str,
        previous_state: (str | None),
        state: str,
        observed_at: str,
        resource_version: str,
        reason: (str | None) = None,
        message: (str | None) = None;
}

class Autoscaler {
    def get_runtime_status(
        autoscaler_name: str,
        scale_target_name: str,
        namespace: str
    ) -> AutoscalerRuntimeStatus;
}

class AutoscalerObserver {
    def watch_namespace(
        namespace: str,
        label_selector: (str | None) = None
    );

    def subscribe(
        autoscaler_name: (str | None) = None,
        scale_target_name: (str | None) = None
    );
}
```

The observer API may expose an iterator, callback, or async stream. The important constraint is architectural: it must not create one Kubernetes watch connection per preview.

## Required behavior

- Normalize KEDA `ScaledObject` conditions, HPA status, and target workload replica status.
- Correctly represent zero replicas as an intentional `inactive` state, not an error.
- Distinguish trigger inactivity from scaler failure.
- Report activation and application readiness as separate states.
- Emit deduplicated transitions for zero-to-one, ready, one-to-zero, and degraded/recovered.
- Use a bounded number of watches per namespace/resource kind and multiplex events in process.
- Support label selectors so callers only observe workloads they own.
- Recover from watch disconnects and Kubernetes `410 Gone` by relisting and resuming from a valid resource version.
- Reconcile relisted state with the observer cache without emitting a storm of duplicate transitions.
- Bound subscriber queues and define backpressure/drop behavior for slow consumers.
- Carry `observed_at` and `resource_version` on transition events.
- Never present `observed_at` as authoritative historical transition time.
- Allow callers to persist business timestamps without requiring per-request activity storage.
- Preserve an engine-neutral normalized model where practical.

## Acceptance criteria

- [ ] `Autoscaler` exposes normalized point-in-time runtime status.
- [ ] KEDA status combines ScaledObject, HPA, and target workload information.
- [ ] A zero-replica healthy workload reports `inactive`, not `missing` or `error`.
- [ ] Scale-up is distinguishable from application readiness.
- [ ] Scaler/metric errors are surfaced with useful condition messages.
- [ ] One observer can monitor many autoscalers in a namespace using a bounded number of Kubernetes watches.
- [ ] No code path creates one watch per scale target or preview.
- [ ] The observer reconnects using resource versions, handles `410 Gone`, and avoids duplicate transition storms after relist.
- [ ] Subscriber backpressure behavior is explicit and tested.
- [ ] Transition events expose observation time and resource version without claiming durable history.
- [ ] Unit tests cover inactive, activating, active, deactivating, degraded, relist, deduplication, and slow-subscriber behavior.
- [ ] A Kubernetes integration test observes a complete 0 -> 1 -> ready -> 0 cycle while multiple targets share the observer.
- [ ] Documentation explains status semantics, watcher cardinality, and persistence responsibility.

## Out of scope

- Persisting transition history or business timestamps.
- Defining application authorization or wake policy.
- Request activity tracking.
- Owning application health endpoints.
- Watching individual pod logs or application-specific events.

## References

- Existing autoscaler abstraction: #6046
- KEDA implementation: #6220 / #6534
- HTTP activation resources: #7403
- JacHammer preview lifecycle: https://github.com/jaseci-labs/jacBuilder/issues/603

Developer-Linus:
Closes #7405

## Summary

`jac-scale` can turn scale-to-zero autoscaling **on** for a workload (`Autoscaler.apply`), but had no way to answer "what state is it in right now?" without a caller reaching past the abstraction and reading raw `ScaledObject` / `HorizontalPodAutoscaler` / `Deployment` objects itself. This PR adds that as a first-class, engine-agnostic capability: a one-shot status query and a shared, registry-backed live observer, for both the plain-HPA and KEDA engines.

The driving consumer is JacHammer, which needs to know when a scale-to-zero preview environment has actually finished waking up before it can serve a stable preview URL.

## Problem

- No method existed to ask "is this workload inactive, activating, active, deactivating, or degraded?" for either autoscaler engine.
- Nothing multiplexed Kubernetes watches: a caller wanting live updates for N targets had no choice but to open N watch connections against the API server.
- Kubernetes watches are not durable history — reconnecting after a drop (`410 Gone` or any other interruption) risked either losing transitions or re-announcing state that never actually changed.

## Solution

**`Autoscaler.get_runtime_status(autoscaler_name, scale_target_name, namespace) -> AutoscalerRuntimeStatus`**
A synchronous, point-in-time query implemented on both `HPAAutoscaler` and `KEDAAutoscaler`, normalizing engine-specific status (HPA conditions; KEDA `ScaledObject`/generated-HPA/target-`Deployment` conditions) into one shared six-state vocabulary: `inactive | activating | active | deactivating | degraded | unknown`.

**`AutoscalerObserverRegistry` / `KubernetesAutoscalerObserver`**
A process-local, namespace-scoped shared observer. `AutoscalerObserverRegistry.get_or_create(namespace, cluster_key)` hands back the same observer instance to every caller for a given `(cluster, namespace)` — this is the structural mechanism that satisfies "no code path may open one watch per preview," not a convention callers have to remember. Each `get_or_create` call is refcounted; pair it with `AutoscalerObserverRegistry.release(namespace, cluster_key)` once done, and the watcher stops only once every caller has released. The observer runs one daemon thread per `(namespace, resource kind)` — `ScaledObject`, `HorizontalPodAutoscaler`, `Deployment` — correlates the three by `scaleTargetRef.name` (each kind's own required, authoritative reference to the target `Deployment`, rather than a shared label) into one normalized status per scale target, and emits an `AutoscalerTransition` to subscribers only when the *normalized state* actually changes (first observation is recorded silently, never announced as a manufactured transition). `AutoscalerSubscription` supports per-subscriber filtering and a bounded queue with a pluggable backpressure policy (`drop_oldest` default, `drop_newest`, `close`).

**Design choices worth calling out:**
- The point-in-time query and the live observer both go through the *same* pure per-engine normalizer function, so they can never disagree about what "active" means.
- The watch/reconnect mechanics deliberately mirror this codebase's existing `EventStreamBroker` / `LocalEventStream` pattern (`Thread` + `threading.Event` stop signal + `stop(drain=True)`/`join()` shutdown) rather than introducing a new concurrency style.
- Any watch interruption (not just `410 Gone`) triggers the same recovery: relist and continue. A relisted object goes through the identical reconcile path as a live event, so an unchanged object naturally produces no transition without a separate raw-diff step.

### New

| File | Lines | Purpose |
|---|---:|---|
| `jac/jaclang/scale/deploy/autoscale/observer.jac` | +130 | Abstract `AutoscalerObserver` interface, `AutoscalerSubscription`, `BackpressurePolicy` |
| `jac/jaclang/scale/deploy/autoscale/kubernetes_observer.jac` | +62 | `KubernetesAutoscalerObserver` declaration |
| `jac/jaclang/scale/deploy/autoscale/kubernetes_observer.impl.jac` | +304 | Watch loop, relist/reconnect, cross-kind join, reconcile |
| `jac/jaclang/scale/deploy/autoscale/observer_registry.jac` | +51 | `AutoscalerObserverRegistry` — shared per-namespace observer |

### Modified

| File | +/- | Purpose |
|---|---:|---|
| `jac/jaclang/scale/deploy/autoscale/autoscaler.jac` | +41 | `AutoscalerState`, `AutoscalerRuntimeStatus`, `AutoscalerTransition`, base `get_runtime_status` stub |
| `jac/jaclang/scale/deploy/autoscale/hpa_autoscaler.jac` / `.impl.jac` | +6 / +70 | `_normalize_hpa_status`, `get_runtime_status` |
| `jac/jaclang/scale/deploy/autoscale/keda_autoscaler.jac` / `.impl.jac` | +11 / +149-5 | `_normalize_keda_status`, `get_runtime_status` |
| `jac/jaclang/scale/_optdeps/kubernetes.jac` | +2-1 | Adds `watch` to the guarded optional-dependency shim |
| `jac/jaclang/scale/deploy/target/kubernetes/utils/kubernetes_utils.jac` / `.impl.jac` | +2 / +9 | Shared `default_cluster_key()` helper |

### Tests (new files)

| File | Tests |
|---|---:|
| `test_autoscaler_runtime_status_contract.jac` | 8 |
| `test_hpa_runtime_status.jac` | 6 |
| `test_hpa_autoscaler.jac` | 5 |
| `test_autoscaler_subscription.jac` | 9 |
| `test_kubernetes_autoscaler_observer.jac` | 22 |
| `test_autoscaler_observer_registry.jac` | 16 |

### Tests (added to existing files)

| File | Tests added |
|---|---:|
| `test_keda_runtime_status.jac` | 17 |
| `test_keda_autoscaler.jac` | 7 |
| `test_optional_deps.jac` | 3 |
| `test_k8s_utils.jac` | 1 |

## Tests

- **94 new tests**, strict TDD throughout (every implementation commit is preceded by a failing-test commit).
- **245 tests passing** across the full targeted suite (autoscale + optional-deps + k8s-utils) with no regressions against the pre-existing suite.
- **Live-cluster validation**: exercised against a real `kind` cluster running KEDA core + the HTTP Add-on, using genuine HTTP traffic through the interceptor to drive a real scale-to-zero → cold-start → active → cooldown → scale-to-zero cycle, observed live via `get_runtime_status` and a real `AutoscalerObserverRegistry` subscription. Reproduced identically on two independently created clusters. This caught two real correctness bugs the unit-test suite had not (see Notes for Reviewers).

## Updates since review

Four of the five issues raised in review are fixed on the branch (the fifth, broader test rigor, is partially addressed and discussed in the comment below). Full details and live-cluster evidence are in [this comment](https://github.com/jaseci-labs/jac/pull/8464#issuecomment-5394446050).

- `label_selector` is no longer a per-caller parameter on `get_or_create`/`watch_namespace`. The observer always watches with a fixed `MANAGED_SELECTOR` constant, since it is one shared watch per `(cluster, namespace)` and cannot have a per-caller selector.
- `get_or_create` now creates, starts, and begins watching as one atomic step under the registry lock, so no caller can observe an instance before it is actually watching.
- `AutoscalerObserverRegistry.release(namespace, cluster_key)` bounds observer lifetime with a refcount, so a long-lived process is no longer stuck accumulating daemon threads with no way to tear one down.
- Live-cluster testing on the fourth point (whether the KEDA-generated HPA carries the `app` label) confirmed it does, but also surfaced a more serious related bug: jac-scale's own microservice/gateway deploy path sets a different `app` label value on the Deployment than on its ScaledObject/HPA, which permanently broke the observer's three-way join for that shape. Fixed by correlating `ScaledObject`/`HPA` by their required `scaleTargetRef.name` instead of by label.

All four fixes have dedicated concurrency or live-cluster regression tests, not mocks. The test counts in the "Tests" section above reflect the branch after these fixes.

## Notes for reviewers

- One known, minor, non-blocking artifact remains: a brief (tens of milliseconds) `activating ↔ deactivating` flicker at the same `resourceVersion` can occur while the three watched kinds (`ScaledObject`/`HPA`/`Deployment`) race to populate a target's cache during a fast transition. It always self-corrects to the right final state and never reports a wrong *final* state


MusabMahmoodh (changes requested):
Sorry this sat 63h - that is my miss, not yours. Read the whole thing; the status normalization is the good part, one shared six-state vocabulary over two engines is the right shape and the join in `_reconcile_target` suppressing unchanged states is exactly right. Four things below, the first is a real bug.

Before any of it: the only red check here is `test-native (passes-native)` -> `Everything Passed`, which is #8583 - it fails on every fork PR right now, mine included. Nothing is failing on this PR's own merits.

**1. `label_selector` is accepted and then silently dropped.** In `observer_registry.jac`, the key is `f"{cluster_key or default_cluster_key()}::{namespace}"` - the selector is not in it. The `if existing is not None { return existing; }` returns before `created.watch_namespace(namespace, label_selector)`, which only runs on the create path. So the second caller into the same namespace with a different selector gets the first caller's filter and is never told. Calling `watch_namespace` again on the hit path would not fix it either: `thread_key` is `f"{namespace}::{kind}"`, so the second call collides with the running thread and is skipped. Either put the selector in the key, or drop the parameter and document the observer as namespace-wide - the current shape is a silent fallback at a decision point.

**2. the instance is published before it is watching.** `_instances[key] = created` is inside the lock; `created.start()` and `created.watch_namespace(...)` are after it is released. A second thread can cache-hit an observer that is registered but has not started its watches yet, subscribe, and silently miss transitions until the first caller gets there. Moving both calls inside the lock, or only publishing after `watch_namespace` returns, closes it.

**3. nothing ever releases an observer.** `get_or_create` has no counterpart but `shutdown_all()`, so three daemon threads per namespace stay alive for process lifetime after the last subscriber is gone. Given the driving consumer creates preview environments continuously, that grows without bound - a refcount or an explicit `release(namespace)` would bound it.

**4. does the KEDA-generated HPA carry the `app` label?** `_target_name_for` correlates on it, with a name fallback only for `deployment`. jac-scale sets `app` on its own HPA (`hpa_autoscaler.impl.jac:48`) and on the ScaledObject (`keda_autoscaler.impl.jac:159`), but the `keda-hpa-*` object is created by the KEDA operator. If it does not propagate `app`, `_handle_event` drops it at the debug log and the `hpa` slot stays `None` forever, so every KEDA target permanently normalizes through `_normalize_keda_status(so, None, dep)`. I could not settle this by reading - do you know from a live cluster?

**5. the tests are all MagicMock.** House rule for scale tests is no mocks, but the specific problem is that a registry, three daemon threads, watch reconnect and a bounded queue are the parts MagicMock cannot prove - and they are 4 of the 5 things above. A kind cluster with a real ScaledObject would cover 1, 2 and 4 in one test. Offer stands: my rig on jac-ec2 can run that, tell me and I will set it up.

MusabMahmoodh (approved):
Sorry for the slow turnaround on this - you answered four days ago and I left it
sitting behind a stale `changes requested`. Clearing that now.

## 1, 2, 3: verified fixed, reading the branch

- **`label_selector`** is gone from `get_or_create` and `watch_namespace`
  entirely, replaced by the `MANAGED_SELECTOR` constant. Removing the parameter
  rather than honouring it in the cache key is the better fix - there is no
  longer a caller-supplied value that *could* be dropped, so the class is closed
  rather than patched.
- **Publication ordering**: `created.start()` and `created.watch_namespace()`
  now both run inside `_lock` before `_instances[key] = created`, so there is no
  window where another thread can take a published instance that is not yet
  watching. Correct.
- **Lifetime**: refcount incremented on the cache-hit path too, `release()`
  decrements, evicts at zero, and - importantly - calls `to_stop.stop()`
  *outside* the lock rather than holding the registry lock across a thread join.
  That is the detail most implementations of this get wrong. `shutdown_all`
  clearing `_refcounts` alongside `_instances` closes the obvious follow-on.

## 4: you found a bigger defect than the one I asked about

This is the part worth calling out. My question was narrow - does the KEDA
generated HPA inherit the app label - and the honest answer would have been
"yes, non-issue, moving on". Instead you took it to a real cluster and found
that `_autoscaler_spec` sets `scale_target_name` to the Deployment *resource*
name (`gateway-deployment`) while the Deployment itself is labelled with the
bare app name (`gateway`), because that label doubles as the pod selector. So
ScaledObject and HPA correlated with each other and never with the Deployment:
two permanently separate `_target_cache` slots per real target, and the observer
could never report `active` for jac-scale's own primary deploy shape.

That is a strictly worse bug than the one in my comment, it is on the main
deploy path rather than an edge case, and correlating by
`spec.scaleTargetRef.name` is the right fix - it is a required field, it is
authoritative, and it makes the watch path agree with `get_runtime_status()`,
which was already doing direct name lookups. Good catch, and the right instinct
to go to a cluster rather than argue it by reading.

## Two residual notes, neither blocking

**a. The error path in `get_or_create` leaks a started observer.** If
`created.start()` succeeds and `created.watch_namespace(namespace)` raises, the
exception unwinds out of the `with` block, `created` is never entered into
`_instances`, and nothing holds a reference to it - but its threads are running.
That is the same "nothing ever releases it" class as finding 3, relocated to the
failure branch. Cheap to close:

```jac
try {
    created.start();
    created.watch_namespace(namespace);
} except Exception {
    created.stop(drain=False);
    raise;
}
```

**b. Eviction and re-creation can briefly overlap.** `release()` pops the
instance under the lock and stops it after, so a concurrent `get_or_create` for
the same key can construct and start a second observer on the same namespace
while the first is still draining. Both are refcounted independently so nothing
corrupts, but subscribers can see duplicate events for the drain window. Worth a
comment on `release()` saying that is intended, or gating on a "stopping" marker
if it is not.

## 5: the rig - yes, and here is the concrete shape

Taking you up on this properly rather than leaving it as an offer.

The protocol I use is written down as **#8654** (`jac/jaclang/scale/TESTING.md`),
open now. Section 1 is the part that matters for what you are trying to capture:
building the binary from the branch under test, and then *proving* the binary
you are about to trust is actually serving your source rather than a baked
release - that trap has cost me whole sessions, and it is the reason a "green
on a real cluster" claim is worth anything. Section 3 is the pre-PR ladder and
3.1 is the cluster discipline. Read those two before we start; if anything in
them is unclear that is a bug in the doc and I would like to know.

What I need from you to get it going:

1. The exact KEDA version and install method you used (you said 2.20.2 - core
   only, or the full bundle?), and the app manifest / `jac.toml` you deployed.
2. Whether you want the captured test to assert the *correlation* invariant (one
   fully populated `_target_cache` slot per real target, which is what you
   verified live) or the end-to-end `active` status. The first is the one that
   would have failed before a0e5c2213 and is the regression test this PR is
   missing; the second is a broader e2e and probably belongs on its own.

One caveat so you are not waiting on me silently: the box is currently out of
disk and the reclaim I need is blocked on a permission I do not have, so I
cannot promise a run today. I will say here the moment it is clear. If you would
rather not wait, everything in #8654 section 1 reproduces on any machine that
can run kind - the doc exists so the rig is not a person.

Approving on the strength of 1-4. (a) is the one I would still like in before
merge, but it does not need to hold the PR.

MusabMahmoodh (commented):
Post-0.37 check, not a new review - my Aug 28 approval stands, and the only
change since is your Aug 30 merge from `upstream/main`.

Confirming for the record: **every check on your head commit is green**,
including the lanes that are currently flaky repo-wide. I checked because
0.37.0 landed on Sunday and several PRs (including four of mine) went red on
`passes-native` and on whole-tree `jac check` failures in files they never
touched. Yours is clean through all of it, so there is nothing for you to
chase.

No further findings from me. This is waiting on a merge, not on review.

jac/jaclang/scale/deploy/autoscale/kubernetes_observer.impl.jac
```diff
@@ -0,0 +1,312 @@
...
+}
+
+impl KubernetesAutoscalerObserver._watch_loop{
+    consecutive_failures = 0;
+    while not self._stop_event.is_set() {
+        try {
+            (resource_version, items, list_fn) = self._list_and_seed(
+                kind, namespace, label_selector
+            );
+            for resource in items {
```
MusabMahmoodh, line 184: The relist only replays objects that still exist, so a ScaledObject, HPA or Deployment deleted while the watch was down keeps its slot and the target reports its old state forever. After seeding, clear `slot[kind]` for any cached `namespace::` target missing from `items` and reconcile those, the usual prune-on-relist step. Greptile raised the same thing and it's still unanswered.

jac/jaclang/scale/deploy/autoscale/kubernetes_observer.impl.jac
```diff
@@ -0,0 +1,312 @@
...
+    return (resource_version, items, list_fn);
+}
+
+impl KubernetesAutoscalerObserver._stream{
+    if self._watch_factory {
+        return self._watch_factory(
+            kind, list_fn, namespace, label_selector, resource_version
+        );
+    }
+    return Watch().stream(
```
MusabMahmoodh, line 167: With no `timeout_seconds` this stream blocks until the next event, and `_stop_event` is only checked between events. In a quiet namespace `release()` hits the 10s join deadline, opens `_stopping`, and the next `get_or_create` starts a new observer while the old threads still hold watches. Passing `timeout_seconds` bounds the stop.

jac/jaclang/scale/tests/deploy/keda_gateway_correlation_verify.jac
```diff
@@ -0,0 +1,172 @@
...
+        if (
+            full_slot.get("scaled_object") is not None
+            and full_slot.get("hpa") is not None
+            and full_slot.get("deployment") is not None
+        ) {
+            break;
+        }
+        time.sleep(1.0);
+    }
+    observer.stop(drain=True);
```
MusabMahmoodh, line 122: After the consolidation nothing calls `get_runtime_status` on either engine, and that's the one-shot API jacHammer is meant to use. This e2e already has a live ScaledObject, HPA and Deployment, so one `KEDAAutoscaler().get_runtime_status(...)` assertion against the observer's state here would cover it on a real API server.

jac/jaclang/cli/docs/reference/plugins/jac-scale-kubernetes.md
```diff
@@ -488,6 +488,42 @@ metadata = { queueName = "orders", mode = "QueueLength", value = "50", protocol
...
+        subscription = observer.subscribe(scale_target_name=scale_target_name);
+        transition = subscription.poll(timeout_seconds=30.0);
+        if transition {
+            print(f"{transition.previous_state} -> {transition.state}");
+        }
+        AutoscalerObserverRegistry.release(namespace);
+    }
+    ```
+
+    For HTTP-activated workloads (`apply_http_activation`, see the KEDA HTTP Add-on note above), `activating` distinguishes "KEDA has requested more replicas" from `active` ("the target Deployment is actually ready to receive traffic") -- watch for `active`, not just a desired-replica bump, before routing a request through.
```
MusabMahmoodh: Since #8779, gateway and sv-to-sv traffic to an HTTP-activated app goes through the interceptor, which holds the request until the target is ready. So waiting for `active` is for UX ("waking up"), not correctness; as written people will build a second gate in front of the interceptor.

jac/jaclang/compiler/tests/fixtures/bundled_stdlib_defaults.jac
```diff
@@ -16,10 +16,10 @@ token_hex of 0 bytes is the empty string instead of 64 hex characters.
 import from zlib { adler32 }
 import from secrets { token_hex }
 
-def adler32_default_seed() -> int {
+def adler32_default_seed -> int {
```
MusabMahmoodh, line 19: This fixture change isn't part of this PR, main still has `adler32_default_seed()`; looks like a merge resolution picked it up. Can you revert it to main's version?

MusabMahmoodh (commented):
Residuals (a) and (b) from my Aug 28 approval are both fixed, and the `_stopping` gate is cleaner than what I suggested. Two watch-loop issues and one coverage gap left, plus a stray compiler fixture that came in through a merge.

jac/jaclang/scale/deploy/autoscale/kubernetes_observer.impl.jac
Developer-Linus: Fixed in 759dacd14. _watch_loop now calls a new _prune_deleted after every relist, which clears the cached slot for that kind on any target missing from the fresh list and reconciles it. Added a step to test_kubernetes_autoscaler_observer_watch.jac that seeds a stale cache entry, calls _prune_deleted directly with an empty items list, and asserts the slot clears and the last state reconciles down to UNKNOWN.

jac/jaclang/scale/deploy/autoscale/kubernetes_observer.impl.jac
Developer-Linus: Fixed in 759dacd14. Added a WATCH_TIMEOUT_SECONDS constant (5.0s) and passed timeout_seconds=WATCH_TIMEOUT_SECONDS to Watch().stream(), so an idle watch now returns control to the loop regularly instead of blocking until the next event. _stop_event is checked right after, well inside the 10s join deadline in stop().

jac/jaclang/scale/tests/deploy/keda_gateway_correlation_verify.jac
Developer-Linus: Fixed in 759dacd14. Added a get_runtime_status assertion right after the correlation check: it computes the observer's own normalized state from the same full_slot (KEDAAutoscaler()._normalize_keda_status) and compares it to KEDAAutoscaler().get_runtime_status(SCALED_OBJECT_NAME, DEPLOYMENT_NAME, NAMESPACE) against the live cluster, failing if they disagree. This exercises the one-shot API against the same real ScaledObject, HPA, and Deployment the correlation check already proved are live.

jac/jaclang/cli/docs/reference/plugins/jac-scale-kubernetes.md
Developer-Linus: Fixed in 759dacd14. Reworded the note: waiting for active is now framed as UX (showing a waking up indicator, or measuring cold-start latency), not a correctness requirement, since gateway and sv-to-sv traffic to an HTTP-activated target already flows through the interceptor, which holds the request until the target is ready.

jac/jaclang/compiler/tests/fixtures/bundled_stdlib_defaults.jac
Developer-Linus: Good catch, but this one turned out not to be a simple revert. I checked out main's exact copy of this file (with the parens), and jac fmt --lintfix immediately stripped them right back off with the same remove-empty-parens warning we get on this branch. Main's fixture just predates that lint rule (it is not alone, several other files under compiler/tests/fixtures still have the old style), so matching main byte for byte fails our own precommit hook, while the current content is already lint-clean and functionally identical to main's version. I left the file as is rather than fight the formatter. Happy to do a separate, unrelated cleanup PR if you would rather main's copy got swept too.

Developer-Linus:
Pushed 759dacd14, merged with main (16 commits, no conflicts) and addressing today's review round:

1. Relist prunes deleted resources: `_watch_loop` now calls a new `_prune_deleted` after every relist, clearing and reconciling any cached target missing from the fresh list. This also closes Greptile's P1 on the same line.
2. Watch shutdown leaves threads active: bounded `Watch().stream()` with a 5s `WATCH_TIMEOUT_SECONDS`, so `_stop_event` gets checked promptly instead of `release()` hitting its join deadline in a quiet namespace. Also closes Greptile's matching P1.
3. Queue capacity lacks validation (Greptile P2, not re-raised today but still open): `AutoscalerSubscription` now rejects a non-positive `max_queue` in `postinit` instead of crashing on `pop(0)` against an empty queue.
4. Coverage gap: added a `get_runtime_status` assertion to `keda_gateway_correlation_verify.jac`, checked against the observer's own normalized state on the same live ScaledObject/HPA/Deployment.
5. Docs: reworded the `active` note to make clear it's for UX, not correctness, since the KEDA HTTP Add-on interceptor already holds the request.
6. Replied inline on the stray fixture comment: matching main's exact bytes there fails our own `jac fmt --lintfix`, since main's copy predates the remove-empty-parens rule. Left as is; happy to sweep it separately if wanted.

Replied inline on each thread with the specifics. All directly relevant tests pass (test_kubernetes_autoscaler_observer_watch.jac, test_autoscaler_observer_registry_lifecycle.jac, keda_gateway_correlation_verify.jac unchanged elsewhere). Full scale/deploy suite has 7 pre-existing failures from the main merge (CLI/entry-point format drift, unrelated to this PR, same ones already visible on other branches re-synced with main today).

MusabMahmoodh (commented):
The prune and the queue guard check out, and the fixture explanation makes sense, so I am leaving that one. The stop bound works, but it turns the watch into a 5s relist loop, and the new e2e assertion compares two different moments.

jac/jaclang/scale/deploy/autoscale/kubernetes_observer.impl.jac
```diff
@@ -0,0 +1,341 @@
...
+        list_fn,
+        namespace=namespace,
+        label_selector=label_selector,
+        resource_version=resource_version,
+        timeout_seconds=WATCH_TIMEOUT_SECONDS
```
MusabMahmoodh, line 172: With `timeout_seconds` set, `Watch.stream` returns after one server-side timeout instead of re-watching, so every 5s this falls out of the `for`, resets `consecutive_failures` and goes back through `_list_and_seed`. That is three LISTs per namespace every ~5s for as long as the observer runs, and it grows with every preview namespace. Keeping the 5s bound but resuming from the last seen `resourceVersion` when the stream ends cleanly, and relisting only on an exception (a 410 already raises), keeps the prompt stop without turning the watch into a poll.

jac/jaclang/scale/tests/deploy/test_kubernetes_autoscaler_observer_watch.jac
```diff
@@ -0,0 +1,331 @@
...
+        obs = KubernetesAutoscalerObserver();
+        key = "default::orders-deployment";
+        obs._target_cache[key] = {
+            "scaled_object": scaled_object(target="orders-deployment"),
+            "hpa": None,
+            "deployment": None
+        };
+        obs._last_state[key] = AutoscalerState.ACTIVE.value;
+
+        obs._prune_deleted("default", "scaled_object", []);
```
MusabMahmoodh: This calls `_prune_deleted` directly, so deleting the new call in `_watch_loop` leaves the step green. `build_observer` already relists about every 50ms because `no_events` ends the stream at once, so start the ScaledObject list with the object in it, switch the list to empty items, and poll until the slot clears. That runs the path the bug actually took.

jac/jaclang/scale/tests/deploy/keda_gateway_correlation_verify.jac
```diff
@@ -0,0 +1,201 @@
...
+        full_slot.get("deployment")
+    );
+    runtime_status = KEDAAutoscaler().get_runtime_status(
+        SCALED_OBJECT_NAME, DEPLOYMENT_NAME, NAMESPACE
+    );
+    print(
+        f"get_runtime_status: state={runtime_status.state} "
+        f"(observer state={observer_status.state})"
+    );
+    if runtime_status.state != observer_status.state {
```
MusabMahmoodh: `full_slot` is the first snapshot where all three objects exist, right after deploy, while the ScaledObject's `Active` condition and the HPA status can still be filling in. `get_runtime_status` reads the cluster later, after `stop()`, so this can fail on a target that is just still settling. Polling `get_runtime_status` until it matches `_last_state[key]` within `OBSERVE_TIMEOUT` with the observer still running, and requiring the state not be `unknown`, keeps two `unknown`s from passing as agreement.

Developer-Linus:
@MusabMahmoodh thanks for the review. All three inline points are addressed and pushed (head `ee7f764d1`, with current `main` merged in at `97147b690`).

**1. The watch no longer relists every 5s** (`98452154d`)

A stream that ends cleanly now resumes from the last seen `resourceVersion`, bookmarks included. Only an error, a 410 among them, goes back through `_list_and_seed`.

While checking this on a kind cluster I found a second problem in the 5s bound itself. The constant was the float `5.0`, and the API server rejected it with `400 strconv.ParseInt: parsing "5.0": invalid syntax`. So every real watch request failed and fell back to a relist. The fake stream in the unit test and the e2e both hid this, because relisting still fills the cache. It is fixed in `0681b8659` by making the value an integer, and the watch test now also checks the value passed to the real `Watch.stream` call.

I ran the same probe against the kind API server for 26 seconds:

| | Lists | Stream opens | Errors |
|---|---|---|---|
| Before these changes | 27 | 27 | 27 (all 400) |
| Now | 3 (one per kind) | 18 | 0 |

**2. The prune test runs through the real watch loop** (`98452154d`)

`step_relist_prunes_deleted_resources` now drives `_watch_loop` using `build_observer`. The ScaledObject starts in the list, the list is then switched to empty, and the step polls until the slot clears and the state reaches `UNKNOWN`. Because a clean stream end no longer relists, the fake stream raises a 410 once to force the relist. The step also checks that clean stream ends do not relist.

I checked that it can fail: removing the prune call from `_watch_loop` fails it, and restoring the old relist-on-every-end loop fails it with "3 lists after 3 streams".

**3. The e2e compares against the observer's live state** (`ee7f764d1`)

`keda_gateway_correlation_verify.jac` now keeps the observer running and polls `get_runtime_status` until it matches `_last_state[key]` within `OBSERVE_TIMEOUT`. An `unknown` state never counts as agreement. On a kind cluster with KEDA, `keda_gateway_correlation_real_e2e.sh` passes, with `get_runtime_status` and the observer both reading `active` and no watch loop errors.

**About the red CI**

The failing check is `build-kit`, and it is not caused by this PR. It fails while building the native compiler kernel (`jac build --native --lib jaclang/compiler/jc_unit.jac`) with `Native dependency analysis failed: .../parser/parser.jac`. All the test jobs were skipped as a result, so this run has not tested the branch yet.

CI builds a PR merged with the current tip of `main`, and `main` moved in the meantime. At `b830db592`, `jac check` on the compiler's own `parser.jac` reports `error[E1053]: Cannot assign list[CodeBlockStmt] to parameter 'iterable' of type Iterable[Token]` at `parser.impl.jac:4183`, on this line:

```jac
awaiting_kid.extend(awaiting_stmts);
```

At `043104a65`, the commit this branch merged, the same file checks clean. Bisecting the commits between them points to `58cf61a44` (#9312, "sum() over floats is a float"): it fails, and its parent `18267fbfa` does not. Other unrelated PRs, for example `fix/9286-sum-comprehension` and `fix/9289-declared-missing-dependency`, fail `build-kit` with the same `parser.jac` error. I only bisected between two adjacent commits, so I have not looked into why the checker infers `Iterable[Token]` there.

I will re-push once `main` is fixed so CI can run the real tests.

Developer-Linus:
Closing in favour of #9432, which replaces this PR's approach to #7405 with a single workload health verdict shared by every status reader. The branch is kept in case any of this work is needed later.

PR #8474, opened 2026-08-21
refactor(ecmascript): tighten jsx attribute and view child declarations

SandeepaHWP:
Three declaration cleanups in the ecmascript backend, from the follow-ups left on #7797 at merge.

Rescoped since it was opened. Two of the original five items landed upstream in #8399: the checker rework there stopped `str.join` pinning a local to `LiteralString`, so the arrow-function emitter needs no annotation, and the `es_unparse` pair came off `.jacignore` on that basis, in a sweep of 53 entries. Both are dropped from here rather than carried as a no-op diff. Nothing below overlaps that work; all three are still present on `main`.

## `StaticAttr.value` is a literal, and now says so

It was declared `es.Expression`, the twenty-member union, but both construction sites build it with `es.Literal`: `normal_attribute` in `jsx_processor.impl.jac` at the no-value branch (`es.Literal(value=True)`) and at the string-value branch (`es.Literal(value=attr_value)`). Every other path there returns `EventAttr`, `RefAttr` or `DynamicAttr`.

Narrowing the declaration deletes the `else` in `_jsx_attr` that the union forced, which was unreachable and duplicated the unsafe-string branch above it verbatim. It also removes the three `isinstance(attr.value, es.Literal)` guards, which the declaration now carries. Same three cases as before: `True` becomes a bare attribute, an unsafe string is wrapped in a `JsxExpressionContainer`, anything else is emitted as the literal.

The point is what happens next time. A producer that routes an expression into `StaticAttr` now fails `jac check` at the construction site, instead of being silently reshaped into a container by a fallback nobody reads.

## One spelling of the child union

The six-member child union is written out five times in `view_ir`. #7797 named it `ViewChild` for the two Solid signatures but left the four copies in the file itself: `IfChild.consequent`, `IfChild.alternate`, `EachChild.body` and `ViewElement.children`. They now use the alias. Forward references are fine here, the same way `estree.jac` uses `Statement` at line 53 and defines it at line 443.

## A dead default on the code generator

`has indent_str: str = '  '` repeats the default `init` already carries one line above, and `init` assigns the field unconditionally, so the has-default is never read. Left as a bare declaration, which keeps one copy of the truth.

## Checks

`jac check` clean on the four changed files against current `main`, with the checker from #8399.

`StaticAttr` is shared, so its consumers were measured either side of the change: `jsx_processor`, the `react` pair and `esast_gen_pass` sit at 175 diagnostics combined, unmoved, with an identical error-code breakdown. All three stay on the ignore list.

`jac test jac/tests/compiler/passes/ecmascript/` passes at 254, and `test_client_codegen` at 14. Generated JavaScript is unchanged.


christianwilkins (commented):
Reviewed head `73c0337494645a10b343c5cc885cdba8691a5a3d`: the full four-file diff, both StaticAttr construction sites, Solid attribute emission, the ViewChild union and JSCodeGenerator initialization. No blocking behavior change found in this static review: both producers supply Literal values; the bare-true, safe-string and escaped-string branches retain their prior output; the alias has the same six members; init sets indent_str unconditionally. The rescope described in the existing author comment is reflected in the current patch.

Validation: `git diff --check` passed. I did not independently rerun the historical 254/14 test counts or the checker comparison cited in the description, so those remain author-reported evidence. Current-head checks show 24 successes and 6 skips, with the native lane and aggregate failing. GitHub no longer supplies that failed job log (97670613547), so I cannot identify or dismiss its cause from the available evidence. No code changes made.

Not ready to merge: GitHub now reports merge conflicts despite the older rebase comment. The author needs to resolve them against current main and run the relevant checker/client codegen suites plus full CI on the resulting head. Maintainer review follows once that head is green.

PR #8479, opened 2026-08-22
fix(byllm): usage/cost reaches every by llm() invocation shape

### LINKED ISSUE #8462: byLLM: token usage is silently dropped for every non-streaming (and several streaming) by llm() call shapes
## Problem Statement / Motivation

byLLM computes real token usage for every LLM call — it's right there in litellm's response — but only *surfaces* it to application code for one specific invocation shape: `stream=True` **and** `logging=True`, via the `usage` `StreamEvent` built in `_invoke_streaming`. Every other shape (plain non-streaming calls, `stream=True` without `logging=True`, no-tools streaming, most of the async surface) computes the same usage data internally and then discards it before it reaches the caller.

This isn't a corner case: it's every non-streaming `by llm()` call, i.e. any structured/typed call that doesn't opt into the ReAct-logging stream shape — a very common pattern (classification, structured extraction, single-shot typed responses). Downstream, this makes it impossible to reconcile token-based accounting against cost-based accounting: cost is captured independently via litellm's own callback hooks (unconditional on streaming), so a `total_cost_usd` figure includes every call, while a `tokens{}`/`num_llm_calls` figure built from byLLM's own `usage` event silently omits every non-streaming call. The two numbers stop agreeing, with no error or warning anywhere.

## Reproducible Minimal Example

No API key required — `mockllm` reproduces this deterministically.

```jac
import from jaclang.byllm.lib { Model, MockToolCall }

glob no_stream_usage: dict[str, int] = {
    "prompt_tokens": 1800, "completion_tokens": 250, "total_tokens": 2050
};
glob stream_usage: dict[str, int] = {
    "prompt_tokens": 900, "completion_tokens": 120, "total_tokens": 1020
};

def finish_tool(final_output: str) -> str {
    return final_output;
}

glob mock_no_stream = Model(model_name="mockllm", config={
    "outputs": [(MockToolCall(tool=finish_tool, args={"final_output": "no-stream reply"}), no_stream_usage)]
});
glob mock_stream = Model(model_name="mockllm", config={
    "outputs": [
        (MockToolCall(tool=finish_tool, args={"final_output": "stream reply"}), stream_usage),
        "stream reply"
    ]
});

"""Shape A: a typed/structured, non-streaming call - e.g. any classifier,
extractor, or single-shot structured response."""
def call_no_stream(prompt: str) -> str by mock_no_stream(max_tokens=200);

"""Shape B: stream=True + logging=True - the one shape byLLM already handles correctly."""
def call_streaming(prompt: str) -> str by mock_stream(tools=[finish_tool], stream=True, logging=True);

with entry {
    print("=== SHAPE A: stream=False ===");
    result_a = call_no_stream("turn digest");
    print("return value:", result_a);
    print("usage_history after call:", mock_no_stream._usage_history);

    print("\n=== SHAPE B: stream=True, logging=True ===");
    for ev in call_streaming("turn digest") {
        if ev.event_type == "usage" { print("usage StreamEvent:", ev.data); }
    }
}
```

Run with: `jac run repro.jac`

## Evidence (captured output, jac 0.34.15)

```
=== SHAPE A: stream=False ===
return value: no-stream reply
usage_history after call: []

=== SHAPE B: stream=True, logging=True ===
usage StreamEvent: {'total': {'prompt_tokens': 900, 'completion_tokens': 120, 'total_tokens': 1020}, 'per_call': [{'prompt_tokens': 900, 'completion_tokens': 120, 'total_tokens': 1020}, {}]}
```

Shape A's call succeeds and returns the correct value — the mock delivered `{"prompt_tokens": 1800, "completion_tokens": 250, "total_tokens": 2050}` for that exact call — but `_usage_history` is empty afterward. That usage was read once, internally, then thrown away. Shape B, hitting the exact same `finish_tool` mechanism, correctly surfaces it.

## Root Cause

Traced through `jaclang/byllm/llm.impl/basellm.impl.jac` (v0.34.15):

`dispatch_no_streaming` (the primitive nearly every call shape routes through) reads real usage off the litellm response and attaches it to `CompletionResult.usage` — the data is never missing at the source:

```jac
usage_data = response.get("usage", {});
return CompletionResult(output=output_value, tool_calls=tool_calls, usage=usage_data);
```

But its only caller for non-streaming calls, `_invoke_react_loop`, uses that usage for one purpose — a local `total_tokens` counter for the compaction threshold check — and drops it on every exit path:

```jac
resp = self.dispatch_no_streaming(mt_run);
usage = resp.usage or {};
total_tokens += int(usage.get("total_tokens", 0) ...);   # used, then gone
just_compacted = self._maybe_compact(mt_run, usage, just_compacted, _cp);
...
return finish_tc.get_output();   # or resp.output — usage never travels with it
```

`self._usage_history` — the one place usage *does* survive to become a `StreamEvent` — is populated by hand inside a handful of streaming-specific functions (`dispatch_streaming`, `_invoke_streaming`) but never inside `dispatch_no_streaming`/`adispatch_no_streaming` themselves. Every shape built on top of the non-streaming primitives inherits the gap.

### Full map of invocation shapes

| Shape | Path | Usage reaches `_usage_history`? | Usage reaches the caller? |
|---|---|---|---|
| `stream=True, logging=True` (tools or not) | `_invoke_streaming` | Yes | **Yes** — `usage` StreamEvent, aggregated |
| `stream=False` (any tools config) | `_invoke_typed_retry` → `_invoke_react_loop` → `dispatch_no_streaming` | No | No |
| `stream=True, logging=False, tools=[...]` | `_invoke_react_loop_streaming` → delegates to the **sync** `_invoke_react_loop` | No | No — despite `stream=True` being set |
| `stream=True, logging=False, tools=[]` | `dispatch_streaming` directly | Yes (per-chunk) | No — nothing ever reads `_usage_history` back; caller gets a raw string generator |
| `stream=True` + `stream_handler=`, no tools, typed | `_invoke_with_handler` branch 1 | Yes | Partial — emits `_usage_history[-1]` as `event_type="llm_timing"`, not `"usage"`; last-chunk snapshot only, requires a custom handler |
| Async, `stream=False` | `_ainvoke_typed_retry` → `adispatch_no_streaming` | No | No. Also has no async ReAct loop at all — async typed calls silently skip tool support too |
| Async, `stream=True, logging=False`, no tools | `adispatch_streaming` directly | Yes (per-chunk) | No |
| Async, anything else | `asyncio.to_thread(self.invoke, ...)` | Inherits the matching sync row | Inherits the matching sync row |

One more adjacent gap: `Model.invoke` (`llm.impl/model.impl.jac`) is the single universal wrapper — every sync call passes through it and unconditionally calls `_emit_agent_telemetry()`. `Model` overrides only `invoke`, not `ainvoke`, so any genuinely-async call bypasses that wrapper (and its telemetry) entirely.

## Proposed Solution

Two independent changes:

**1 — Push usage capture down into the primitives, not the shapes built on top of them.**
Make `dispatch_no_streaming` and `adispatch_no_streaming` append their own `usage_data` to `self._usage_history` right where they build `CompletionResult`, mirroring what `dispatch_streaming`/`adispatch_streaming` already do per chunk:

```jac
usage_data = response.get("usage", {});
self._usage_history.append(usage_data);   # new
return CompletionResult(output=output_value, tool_calls=tool_calls, usage=usage_data);
```

Every caller above these two functions (`_invoke_react_loop`, `_invoke_typed_retry`, `_ainvoke_typed_retry`, `_invoke_react_loop_streaming`) then populates `_usage_history` for free, with no changes to their own control flow. This alone closes 5 of the 6 broken rows in the table above.

**2 — One uniform exposure point, not a per-shape patch.**
`Model.invoke`/`_stream_with_telemetry` already runs after every completed invocation, streaming or not, and already calls `_emit_agent_telemetry()` unconditionally. After (1), `_usage_history` is guaranteed populated by the time that wrapper runs. Factor the aggregation already written at the tail of `_invoke_streaming` into a shared helper, call it once from `Model.invoke`'s success branch, and deliver it through a purpose-built hook:

```jac
def:pub register_usage_callback(callback: Callable) -> None;
```

— symmetric with the existing `register_agent_callback`, firing once per completed invocation with the same `{"total": {...}, "per_call": [...]}` shape the `usage` StreamEvent already uses, so existing `usage`-event consumers need no reshaping — just a second subscription point for non-streaming calls. Also override `Model.ainvoke` the same way `Model.invoke` is wrapped today, closing the async telemetry/usage gap symmetrically.

## Alternatives Considered

- **Set `stream=True` on the affected call sites.** Doesn't actually work: for a typed/no-tools call, `stream=True` without a `stream_handler=` routes through `_invoke_react_loop_streaming`, which — per the table above — just calls the *sync* `_invoke_react_loop` and yields its single return value. Usage is dropped exactly the same way. The one shape where `stream=True` *does* surface something (`stream_handler=` + no tools) exposes a different, non-standard event (`llm_timing`, last-chunk-only) and requires every call site to write a custom handler — inconsistent and not a general fix.
- **Capture usage at the application layer via a litellm `CustomLogger`.** Works today (litellm's callback fires regardless of streaming), and is a reasonable stopgap for a single application. Rejected as the *real* fix because: it's boilerplate every byLLM consumer has to reinvent; it only sees calls that go through litellm — byLLM also supports `http_client`/`proxy` dispatch modes that never touch litellm at all, so those calls would stay uncounted; and it duplicates parsing/aggregation logic byLLM already implements correctly for the streaming case.
- **Read `Model._usage_history` directly from application code after a call.** Rejected: it's a private, underscore-prefixed field, not a stable API, and it's unconditionally reset (`= []`) at the start of the next streaming call on the same instance — unsafe to read post-hoc, especially with any concurrency.

## Impact (without this fix)

For any application built on non-streaming or partially-streaming `by llm()` calls:

- `num_llm_calls`/token totals built from the `usage` event undercount real usage by exactly the number of non-qualifying calls — silently, with no error.
- Cost (captured independently via litellm's own callback, unconditional on streaming) and token counts (captured only via the `usage` event) stop reconciling against each other, since only one of the two paths sees every call.
- Any per-call cost/token analysis built on this data silently excludes affected calls, which skews outlier/anomaly detection toward whatever *is* visible.

This was found via a downstream production system tracking per-call cost and token usage for every LLM call in a multi-stage agent pipeline. Cross-checking `num_llm_calls` and token totals against independently-tracked cost across dozens of real runs showed **zero mismatches** on the streaming calls and a **consistent ~2% cost gap with no matching tokens** attributable to exactly one non-streaming call per run — the pattern in the table above, not a one-off.

## Behavior: Before vs. After

| | Before | After |
|---|---|---|
| Non-streaming typed/structured call | Usage computed, silently discarded | Usage available via `register_usage_callback` |
| `stream=True, logging=False` (with or without tools) | Usage discarded despite `stream=True` | Usage available via `register_usage_callback` |
| Async calls | No telemetry or usage at all (`Model.ainvoke` unwrapped) | Same telemetry + usage treatment as sync |
| Consistency | Only one invocation shape (`stream=True, logging=True`) ever reports usage | Every invocation shape reports usage through one of two consistent channels (`usage` StreamEvent for streaming, `register_usage_callback` for everything) |
| Application-layer workaround needed | Yes — a custom litellm `CustomLogger`, incomplete for non-litellm dispatch modes | No |

## Environment

- `jac 0.34.15` (self-contained binary, Linux x86_64) — reproduced with the exact pinned release a downstream production app tracks
- `mockllm` backend — no external API/model access needed to reproduce
- Also traced against the current `jaclang/byllm` source tree for confirmation the gap is still present past this pinned version

SahanUday:
## Summary

Fixes #8462. byLLM computed real token usage and cost for every LLM call, but only surfaced it to application code for `stream=True, logging=True`. Every other invocation shape (plain non-streaming calls, `stream=True` without `logging=True`, tool-calling loops) computed the same data and silently discarded it.

The actual production change is small - around 300 lines across `basellm.impl.jac`/`model.impl.jac`/`llm.jac` - plus a 3-file-into-1 test consolidation (`test_usage_cost.jac`/`test_usage_history.jac`/`test_usage_step.jac` -> `test_usage.jac`). Most of the diff stat is that test file.

- Usage is now recorded on the per-call `MTRuntime`, not the shared `BaseLLM`/`Model` instance (which is routinely reused across concurrent invocations - `_copy_for_compaction` already worked around this for nested compaction calls; this closes the underlying hole).
- Every recorded call carries a `call_kind` (`react_iteration` by default, or `final_answer_restream`/`recovery`/`compaction` for calls byLLM's own machinery makes on the caller's behalf), reaching the caller via a new `tokens` field on `register_agent_callback`'s telemetry record, and via a new `usage_step` `StreamEvent` that fires once per individual call in a streaming tool-calling invocation instead of only once at the very end - lets a caller enforce a per-turn budget without waiting for a possibly many-call turn to finish. `IterationContext` gains `total_cost` alongside `total_tokens` for the same reason.
- Cache token fields (`cache_read_input_tokens`/`cache_creation_input_tokens`) are now normalized to the same names across providers, defaulting to `0` instead of being silently absent depending on how a given provider reports them.
- byLLM opts in to litellm's native `include_cost_in_streaming_usage`, and every streaming path (including tool-calling ReAct loops) now benefits from it rather than falling back to a locally-recomputed cost.
- `Model.ainvoke` gets the same telemetry treatment `Model.invoke` already had, closing a matching async gap.

Not in scope here (tracked separately): a redundant second LLM call byLLM makes on `stream=True` + tools.

## Review feedback addressed

**Round 1** - both blockers fixed:
- `stream_chunk_builder(logging_obj=...)` now only passes that kwarg when the installed litellm actually accepts it. Older litellm degrades to the local cost fallback instead of losing tool calls.
- `mt_run._usage_history` was being wiped right before telemetry read it, leaving `tokens` empty for `stream=True, logging=True`. Removed the dead wipe.

Round 1 minor items: zero-usage calls are now counted instead of vanishing from `requests`; streaming paths attribute usage/cost to the model that actually served the call (matters under `ModelPool`); the async telemetry wrapper no longer leaks its invocation-id contextvar between yields; docs corrected on `usage_step` duplicating `usage` for a single-call turn. Also raised the litellm floor to `>=1.75.2` and removed the dead `BaseLLM._usage_history` field.

**Round 2**:
- The signature-inspection check for `logging_obj` support now logs at debug on failure instead of a bare silent except, and is hoisted to a module-level constant computed once at import rather than on every dispatch.
- `_usage_to_dict` now logs at debug when a usage object fails to serialize, instead of silently degrading to an empty payload indistinguishable from a real zero-token call.
- `finish_reason` is now normalized onto streaming per-call records too, matching the non-streaming ones.
- Consolidated three clusters of near-duplicate tests (tokens-via-callback across shapes, call_kind tagging across follow-up kinds, usage_step across call kinds) into table-driven tests - 9 tests into 4 - while keeping every distinct path (compaction interleaving, the no-fire case, cache normalization, mt_run isolation) intact.

## Test plan

- [x] `test_usage.jac`: 23 tests covering cost computation, usage capture/isolation/tagging, per-call `usage_step` exposure, and the review-fix regressions - each written and confirmed failing against pre-fix source before implementing.
- [x] Live-verified against real Anthropic (`claude-haiku-4-5`) and OpenRouter (`moonshotai/kimi-k2.7-code`, `meta/muse-glimmer-30b`) calls, on litellm pinned to the new `1.75.2` floor - every invocation shape, plus a real `ModelPool` fallback. Results and script posted as PR comments.
- [x] Full targeted run green: `test_usage.jac`, `test_telemetry_.jac`, `test_byllm.jac`, `test_compaction.jac`, `test_parallel.jac`, `test_tool_arg_streaming.jac`.
- [x] Docs updated (`byllm.md`): `tokens`/`usage_step`/cache-token fields, `call_kind` reference table, `IterationContext.total_cost`.


MalithaPrabhashana (changes requested):
Reviewed this by running it: checked out the branch, exercised each `by llm()` shape against both mocked transports and the real Anthropic API, and compared against `main` on identical inputs.

The core is right. Moving usage off the shared `BaseLLM` instance onto the per-call `MTRuntime` is the correct fix, the `call_kind` tagging is genuinely useful, and I confirmed shapes A/B/D go from "usage silently discarded" on `main` to fully reported here. Live against `anthropic/claude-haiku-4-5`, a streamed tool-calling turn emitted three correctly-tagged `usage_step` events and a final `usage` of `requests=3, cost=0.00347`.

**Two blockers**, both with reproductions, now filed as inline threads so they can be resolved individually:

1. **`logging_obj=` breaks litellm 1.70.0-1.75.1, silently** - `basellm.impl.jac:1650`. The kwarg landed in 1.75.2; below that it's a `TypeError` swallowed at `logger.debug`, and the agent loses its native tool calls.
2. **`tokens` is empty for `stream=True, logging=True`** - `basellm.impl.jac:2275`, one vestigial line, with a one-click suggestion on the thread.

Four smaller notes are on threads too (`_record_usage` zero-token guard, model attribution under `ModelPool`, the async contextvar asymmetry, and a `usage_step` docs claim that doesn't hold). Of those four, the `ModelPool` one is reasoning from the code rather than measured - I traced the divergence but didn't drive a router fallback to watch it happen.

---

Two findings have no line to anchor to, so they live here:

**The litellm floor.** The fix for blocker 1 is raising `>=1.70.0` to `>=1.75.2` in the five places carrying that string - `jac/jaclang/project/capabilities.jac:55`, `scripts/fresh_env.sh:56`, and `.github/workflows/ci.yml` lines 343, 721, 846 - plus a release-note line for the bump. There's an independent argument for it: litellm 1.75.0 fails 19 tests on jac's bundled Python 3.14 (`asyncio.iscoroutinefunction` deprecation-as-error), so a floor-version CI lane isn't viable either way and the 1.70 claim was already untested.

Note this documents the requirement without enforcing it. `missing_capability_dists` checks presence only - it calls `distribution(base)` and never reads the specifier ([`capabilities.jac:350-373`](https://github.com/jaseci-labs/jac/blob/bdd8b085d3209f2cb2f0fc50735bf090ddf5c3e7/jac/jaclang/project/capabilities.jac#L350-L373)) - and `require_optional` is just `__import__`. With 1.75.0 actually installed, byLLM reported the `llm` capability as fully satisfied and raised nothing. Any environment that already has litellm keeps whatever version it has, so a raised floor alone won't reach it.

**`BaseLLM._usage_history` is now dead** - still declared at [`llm.jac:81`](https://github.com/jaseci-labs/jac/blob/bdd8b085d3209f2cb2f0fc50735bf090ddf5c3e7/jac/jaclang/byllm/llm.jac#L81) but never read or written after this change. It silently reads `[]` forever for anyone who was using it (this PR's own `test_usage_cost.jac` did, before the rewrite). Either remove it or call the removal out in the release note.

**Release-note gaps** worth a line each, in the same spirit as the `llm_timing` note you already wrote: the `usage` event now fires unconditionally (previously guarded by a non-empty history), and `llm._usage_history` stops being populated.

---

Nice work on the root-causing and on the `usage_step` compaction-interleaving detail - that one was easy to get wrong. Happy to re-review once the two blockers are addressed.

(Test state: the byLLM usage tests pass on this branch. The `test_mtir_integration.jac` failures I saw are pre-existing and order/cache dependent - they vary run to run on `main` too, so unrelated to this PR.)

jac/jaclang/byllm/llm.impl/basellm.impl.jac
```diff
@@ -1565,7 +1647,9 @@ impl BaseLLM.dispatch_streaming_with_tools(
     message: (LiteLLMMessage | dict | None) = None;
     usage_data: dict = {};
     try {
-        rebuilt = litellm.stream_chunk_builder(chunks);
+        rebuilt = litellm.stream_chunk_builder(
+            chunks, logging_obj=_stream_resp?.logging_obj
+        );
```
MalithaPrabhashana: **Blocking.** `logging_obj` was added to `stream_chunk_builder` in litellm **1.75.2**, but byLLM declares `">=1.70.0,<=1.82.6"`. On 1.70.0-1.75.1 this raises `TypeError`, which the `except` below swallows at `logger.debug` - so `message` falls back to accumulated text and native tool calls are lost.

Reproduced with a real litellm 1.75.0 installed (no mocking of `stream_chunk_builder`), same script both sides:

```
main:     tool_call: {'tool': 'look_up', 'args': {'city': 'Colombo'}, ...}
          usage: total_tokens 240, cost 5.4e-05

this PR:  DEBUG ... stream_chunk_builder() got an unexpected keyword argument 'logging_obj'
          (no tool_call event at all)
          usage: {'total': {}, 'by_kind': {}, 'per_call': [], 'requests': 0}
```

The pin doesn't protect this: `missing_capability_dists` only calls `distribution(base)` and never reads the specifier, and `require_optional` is just `__import__`. With 1.75.0 installed, byLLM reports the `llm` capability as fully satisfied.

Suggested fix is raising the floor to `>=1.75.2` in `capabilities.jac:55`, `scripts/fresh_env.sh:56` and `ci.yml` 343/721/846 - I can't anchor a thread there since this PR doesn't touch those files, so the detail is in the summary review. Because the gate is presence-only, consider also passing the kwarg only when the installed `stream_chunk_builder` accepts it, so an older litellm degrades to a less precise cost rather than a silently toolless agent.

Worth noting the suite can't catch this: every test that reaches `stream_chunk_builder` stubs it with a `logging_obj`-accepting signature (`test_usage.jac:165`, `:325`).

jac/jaclang/byllm/llm.impl/basellm.impl.jac
```diff
@@ -2181,32 +2264,15 @@ impl BaseLLM._invoke_streaming(
...
-                }
-            }
-        }
-        yield StreamEvent(
-            event_type="usage", data={"total": totals, "per_call": per_call_dicts}
-        );
-        self._usage_history = [];
-    }
+    yield StreamEvent(event_type="usage", data=_aggregate_usage(mt_run._usage_history));
+    mt_run._usage_history = [];
```
MalithaPrabhashana: **Blocking.** This wipe is why the new `tokens` telemetry field comes back empty for `stream=True, logging=True`.

`_stream_with_telemetry` emits telemetry from its `finally`, *after* the generator is exhausted - so `_emit_agent_telemetry` reads `_usage_history` when this line has already cleared it. Live against `anthropic/claude-haiku-4-5` on this branch:

```
A. plain non-streaming              tokens -> requests=1 tokens=59   cost=7.5e-05
E. tools + stream=True + logging    usage  -> requests=3 tokens=2702 cost=0.00347
                                    tokens -> requests=0     <-- empty
```

That contradicts the docs added here ("covers streaming and non-streaming calls alike"), and it's the shape with the most calls to account for.

The line looks vestigial - it was needed when the history lived on the shared `Model` instance, but `MTIR.runtime` now rebuilds `MTRuntime` per invocation, so there's nothing to clean up. I tried deleting it: streaming `tokens` becomes `requests=3 ... cost=0.003385`, matching the `usage` event exactly, and the usage suite stays green.

```suggestion
    yield StreamEvent(event_type="usage", data=_aggregate_usage(mt_run._usage_history));
```

Worth a test alongside it - the `tokens` assertions currently cover only non-streaming shapes (`test_usage.jac:418-422`, `:556`), which is why this got through.

MalithaPrabhashana (commented):
Following up my earlier review with the actionable items as inline threads, so each can be resolved independently. Two carry one-click suggestions. No new blocking verdict here - the `CHANGES_REQUESTED` review already posted still stands and holds the full evidence.

Two points from that review have no line to anchor to and stay there: the litellm floor (lives in `capabilities.jac`, untouched by this PR) and the now-dead `BaseLLM._usage_history` field (`llm.jac:81`, outside any hunk).

jac/jaclang/byllm/llm.impl/basellm.impl.jac
```diff
@@ -309,17 +326,86 @@ def _attach_call_cost(
...
+    call_kind: str | None = None,
+    finish_reason: str | None = None
+) -> None {
+    if not (
+        _usage_int(usage_obj, "prompt_tokens")
+        or _usage_int(usage_obj, "completion_tokens")
+        or _usage_int(usage_obj, "total_tokens")
+    ) {
+        return;
+    }
```
MalithaPrabhashana: Minor. A call whose usage payload reports no prompt/completion/total tokens is dropped entirely rather than recorded as a zero-token call.

That makes `requests` undercount against its documented meaning ("total LLM calls this invocation made") for any provider that returns no usage payload - the caller sees fewer requests than actually happened, with no way to tell the difference between "no calls" and "calls we couldn't measure". Recording the call with zeros, or tracking a separate count, would keep `requests` honest.

jac/jaclang/byllm/llm.impl/basellm.impl.jac
```diff
@@ -1457,15 +1538,15 @@ impl BaseLLM.dispatch_streaming(mt_run: MTRuntime) -> Generator[str, None, None]
...
+    usage_base = len(mt_run._usage_history);
     while True {
         try {
             response = self.model_call_with_stream(params);
             for chunk in response {
                 chunk_usage = chunk?.usage;
                 if chunk_usage {
                     _attach_call_cost(chunk_usage, self.model_name);
-                    self._usage_history.append(chunk_usage);
+                    _record_usage(mt_run, chunk_usage, self.model_name);
```
MalithaPrabhashana: Minor, but it matters under `ModelPool`. The non-streaming paths record the model the response actually reports (`actual_model`, lines 1508 and 1745), while the streaming paths record the configured `self.model_name` - here, plus `:1689` and `:1766`.

When a router falls back to a different model, the streaming records will attribute the cost to the model you asked for rather than the one that served it, which undercuts the point of having `model` on each record.

jac/jaclang/byllm/llm.impl/model.impl.jac
```diff
@@ -100,6 +103,88 @@ impl Model.invoke(mt_run: MTRuntime) -> any {
...
+}
+
+async def _astream_with_telemetry(
+    agen: object, mt: object, model_name: str, start: float
+) -> object {
+    import time;
+
+    invocation_id = mt?._invocation_id;
+    _token = (
+        _current_invocation_id.set(invocation_id) if invocation_id is not None else None
```
MalithaPrabhashana: Minor asymmetry with the sync twin. `_stream_with_telemetry` (`:195-197`) deliberately uses `contextvars.copy_context()` + `ctx.run` so the invocation id never escapes into the caller's context; this async version sets it directly.

Async generators don't get an isolated context, so the `set` here leaks into the consumer's context between yields, and the `reset` in `finally` can raise `ValueError: Token was created in a different Context` if the generator is closed from a different task than the one that started it. Mirroring the sync approach would keep the two consistent.

jac/jaclang/cli/docs/reference/plugins/byllm.md
```diff
@@ -1521,25 +1529,54 @@ with entry {
...
+            running_total += event.data["total"].get("total_tokens", 0);
+            if running_total > budget {
+                break;  # stop consuming further events once over budget
+            }
+        }
+    }
+}
+```
+
+`usage_step` only fires for invocations that make more than one LLM call (tools present); a single-call invocation has nothing to distinguish a per-call step from the final `usage` event, so only `usage` fires there. `on_iteration`'s `IterationContext` (see [Interrupting the ReAct Loop](#interrupting-the-react-loop)) gives the same real, cumulative cost between calls via `total_cost`, if you'd rather gate the *next* call than react to a stream event.
```
MalithaPrabhashana: This claim doesn't hold - a tools-enabled invocation that makes exactly one call does fire a `usage_step`, duplicating the final `usage`:

```
['thought', 'llm_timing', 'usage_step', 'steps_done', 'chunk', 'usage']
```

Worth correcting, since a reader following this would sum `usage_step` events and then also add `usage`, double-counting that case.

```suggestion
`usage_step` fires once per LLM call in a tools-enabled invocation, including when that invocation only makes a single call - in which case it carries the same numbers as the final `usage` event, so don't sum both.
```

jac/jaclang/byllm/llm.impl/basellm.impl.jac
SahanUday: Fixed: `stream_chunk_builder` is only called with `logging_obj` when the installed litellm actually accepts it (checked via `inspect.signature`, see `_rebuild_stream_chunks`). Older litellm now degrades to the local cost fallback instead of losing tool calls.

Also raised the floor to `>=1.75.2` for a fresh install, but kept this guard since the floor isn't enforced for an already-installed litellm (`missing_capability_dists` only checks presence, not version).

jac/jaclang/byllm/llm.impl/basellm.impl.jac
SahanUday: Fixed: removed the `mt_run._usage_history = []` line - `mt_run` is per-invocation now, so there was nothing left to clean up there, it was just wiping the data before telemetry could read it. `tokens` is now populated for `stream=True, logging=True` too, with a regression test covering it.

jac/jaclang/byllm/llm.impl/basellm.impl.jac
SahanUday: Fixed: removed the guard in `_record_usage` that dropped a call whose usage payload had no measurable tokens. `requests` now counts every call that happened, including zero-usage ones.

jac/jaclang/byllm/llm.impl/basellm.impl.jac
SahanUday: Fixed: the streaming paths (`dispatch_streaming`, `adispatch_streaming`, `dispatch_streaming_with_tools`) now attribute usage/cost to the model the chunk/response actually reports, the same way the non-streaming paths already did with `actual_model`. Covered by a regression test.

jac/jaclang/byllm/llm.impl/model.impl.jac
SahanUday: Fixed: `_astream_with_telemetry` now steps the wrapped generator inside an isolated `contextvars.Context` via `asyncio.Task(..., context=ctx)`, mirroring the sync wrapper's `ctx.run` isolation. Added a regression test proving the invocation id no longer leaks into the consumer's context between yields.

jac/jaclang/cli/docs/reference/plugins/byllm.md
SahanUday: Fixed the wording: `usage_step` fires even for a single-call tool invocation, carrying the same numbers as the final `usage` event there - so callers shouldn't sum both.

jac/jaclang/byllm/llm.impl/basellm.impl.jac
```diff
@@ -274,6 +283,27 @@ def _litellm_ctx_window(model_name: str) -> int {
...
+
+def _rebuild_stream_chunks(chunks: list, logging_obj: object) -> dict {
+    import inspect;
+    accepts_logging_obj = False;
+    builder: any = litellm.stream_chunk_builder;
+    try {
+        accepts_logging_obj = "logging_obj" in inspect.signature(builder).parameters;
+    } except Exception {
+        ;
+    }
```
MalithaPrabhashana: Two small things here.

**The silent except.** If `inspect.signature` ever fails, this assumes "no logging_obj" and moves on with no trace. That is the right default, but the bare `;` body means a wrong guess is undiagnosable - and this function is what stands between an older litellm and the silently-toolless agent from the earlier thread.

```suggestion
    try {
        accepts_logging_obj = "logging_obj" in inspect.signature(builder).parameters;
    } except Exception as e {
        logger.debug(f"could not inspect stream_chunk_builder, omitting logging_obj: {e}");
    }
```

**The reflection is per-call.** `inspect.signature` runs on every `dispatch_streaming_with_tools` dispatch, and the answer cannot change at runtime. Worth hoisting to a module-level constant next to `_TOOL_DELTA_MIN_CHARS` and reducing this function to the branch. Not a hot path next to an LLM round-trip, but it is reflection in the request path for a value that is fixed at import.

jac/jaclang/byllm/llm.impl/basellm.impl.jac
```diff
@@ -309,17 +342,79 @@ def _attach_call_cost(
...
+    if isinstance(usage_obj, dict) {
+        return dict(usage_obj);
+    }
+    try {
+        return usage_obj.model_dump()
+            if usage_obj?.model_dump
+            else dict(vars(usage_obj));
+    } except Exception {
+        return {};
+    }
```
MalithaPrabhashana, line 351: `_record_usage` no longer skips calls with no measurable tokens (good - that was the `requests` undercount). But that makes this swallow load-bearing in a way it was not before: a usage object that fails to serialize now becomes `{}`, gets `model`/`call_kind`/cache fields stapled on, and is counted in `requests` with all its tokens gone. From the outside that is indistinguishable from a genuine zero-token call.

Nothing to fix in the behaviour - falling back to an empty payload is the right call. It just should not be invisible when it happens. `logger` is already in scope in this file.

```suggestion
    try {
        return usage_obj.model_dump()
            if usage_obj?.model_dump
            else dict(vars(usage_obj));
    } except Exception as e {
        logger.debug(f"usage serialization failed, recording an empty payload: {e}");
        return {};
    }
```

MalithaPrabhashana (commented):
Re-reviewed the fixes: all six earlier threads check out, and I verified the two blockers behaviourally rather than by reading - real litellm 1.75.0 for the signature guard (tool calls preserved, usage present), and live Sonnet 4.6 for the rest. Streaming `tokens` goes from `requests=0` to matching the `usage` event exactly, `usage_step` sums to the final total to the cent, and both dispatch paths now agree on model attribution. Suite is 254 passed locally; CI green.

**Scope: all three threads below are expected in this PR**, not as follow-ups. An earlier version of this comment called them non-blocking and suggested the test work could land separately - disregard that, it was my error and the standing `CHANGES_REQUESTED` covers all three.

Also still outstanding from the summary review:
- a release-note line for the removed `BaseLLM._usage_history` (it is a compile error now, `E1030`, not a silent `[]`, so people should hear about it)
- `finish_reason` present on non-streaming per-call records but absent on streaming ones - same normalization argument you already accepted for the cache fields

Two of the three carry one-click suggestions. Once this round is in I will re-run the live verification and approve.

jac/jaclang/byllm/tests/test_usage.jac
```diff
@@ -0,0 +1,1095 @@
...
+    ) {
+        result = _byllm_llm_mod._rebuild_stream_chunks([1, 2], fake_logging_obj);
+        assert result == {"marker": "new", "logging_obj": fake_logging_obj} , (
+            "a new-signature stream_chunk_builder should still receive logging_obj"
+        );
+    }
+}
+
+
+test "a streaming, tool-calling by llm() call surfaces tokens via register_agent_callback too" {
```
MalithaPrabhashana: Non-blocking, and happy for this as a follow-up rather than something that holds the merge.

This test and the two below it (`a plain non-streaming by llm() call...` at :500, `async by llm() call with no tools...` at :631) are the same assertion varying by one dimension - the invocation shape. Same story twice more in the file:

| cluster | tests | lines |
|---|---|---|
| `tokens` via callback x plain / streaming / async | 3 | ~101 |
| `call_kind` tagging x restream / recovery / compaction (:534, :569, :586) | 3 | ~117 |
| `usage_step` fires for x react / restream / recovery (:748, :898, :934) | 3 | ~150 |

Those 9 look table-drivable the way `_usage_from(path)` already does for the five dispatch paths in the first test - roughly 9 tests into 3, about 200 lines.

To be clear about the scale: that's ~200 lines, not 800. The rest of this file is genuinely distinct paths - compaction interleaving, the no-fire case, cache normalization across three provider shapes, mt_run isolation, the signature check - and I would not want any of that folded away. Two real bugs in this PR were only caught by tests added in the last round, so coverage is the last thing worth trading for line count.

The other half of this is presentation: the PR reads as +1,526 when it is ~300 lines of production code plus a 3-into-1 test consolidation. A sentence in the description saying so would stop the next reader flinching at the number the way I did.

MalithaPrabhashana: Correcting myself on scope: this one is expected in **this PR**, not as a follow-up. Ignore the "happy for this as a follow-up rather than something that holds the merge" line above - that was my call to make and I got it wrong.

Everything else in the comment stands, and the part I would most want held onto is the boundary: the three named clusters are the consolidation target (~200 lines), and the distinct paths around them - compaction interleaving, the no-fire case, cache normalization across three provider shapes, mt_run isolation, the signature check - should survive intact. Two real bugs in this PR were only caught by tests added last round, so if the choice is ever between fewer lines and keeping a path covered, keep the path.

Worth re-confirming each consolidated case still fails against pre-fix source afterwards, the way you did originally - that is the check that makes a test refactor safe rather than just shorter.

jac/jaclang/byllm/llm.impl/basellm.impl.jac
SahanUday: Fixed both: `_stream_chunk_builder_accepts_logging_obj` now logs at debug when `inspect.signature` itself raises, and the detection is hoisted to a module-level constant (`_STREAM_CHUNK_BUILDER_ACCEPTS_LOGGING_OBJ`, computed once at import in `llm.jac`) so `_rebuild_stream_chunks` is just a branch on it, no per-dispatch reflection.

jac/jaclang/byllm/llm.impl/basellm.impl.jac
SahanUday: Fixed both: `_stream_chunk_builder_accepts_logging_obj` now logs at debug when `inspect.signature` itself raises, instead of a bare `;`. And the check is hoisted to a module-level constant (`_STREAM_CHUNK_BUILDER_ACCEPTS_LOGGING_OBJ`, computed once at import in `llm.jac` next to `_TOOL_DELTA_MIN_CHARS`) - `_rebuild_stream_chunks` is now just a branch on it, no per-dispatch reflection.

jac/jaclang/byllm/llm.impl/basellm.impl.jac
SahanUday: Fixed exactly as suggested: `_usage_to_dict`'s except branch now logs at debug before returning the empty payload, so a genuine serialization failure is distinguishable from a real zero-token call in the logs.

jac/jaclang/byllm/tests/test_usage.jac
SahanUday: Consolidated all three clusters as suggested:

- `tokens` via callback (plain/streaming/async) -> one table-style test with a shared `_assert_tokens_via_callback` helper.
- `call_kind` tagging (restream/recovery/compaction) -> one test looping over three case-builder functions, checking count/kind/reset per case.
- `usage_step` fires for X -> merged the react+restream test with the standalone restream test (gave the restream leg real injected usage instead of zero, folding in the exact check the standalone one existed for), kept the recovery one separate since its setup goes through `_invoke_streaming` directly rather than the full `by llm()` sugar.

9 tests -> 4, kept every distinct path you called out intact (compaction interleaving, the no-fire case, cache normalization across three provider shapes, mt_run isolation, the signature check). Re-confirmed each consolidated case still fails independently by reverting the underlying fix for that one case and re-running - all three call_kind cases and the restream usage_step case caught their regression correctly, nothing silently green.

Also added the description note about the real line count (~300 lines of production code plus the test consolidation, not the raw diff stat).

MalithaPrabhashana (approved):
Approving. Everything from both review rounds is addressed, and I re-verified the behaviour against a real model on this head (`5cef4b64`) rather than carrying over the earlier numbers.

## Live verification, `anthropic/claude-sonnet-4-6`

Usage and cost now reach every invocation shape - #8462's original complaint:

| shape | requests | tokens | cost | call kinds |
|---|---|---|---|---|
| plain | 1 | 72 | $0.000432 | `react_iteration` |
| `stream=True` | 1 | 78 | $0.000498 | `react_iteration` |
| `stream+logging` | 1 | 76 | $0.000468 | `react_iteration` |
| tools | 2 | 1850 | $0.008022 | `react_iteration` |
| tools+stream+logging | 3 | 3151 | $0.014954 | `final_answer_restream`, `react_iteration` |

That last row returned `requests=0` before this PR - it was the shape with the most calls to account for and the one that reported nothing.

`usage_step` invariants on a real tool-calling turn, all exact:

```
steps == requests        PASS
step tokens == final     PASS   (2352 == 2352)
step cost == final       PASS   ($0.010820, to the cent)
total_cost is monotonic  PASS
```

And `finish_reason` is now on both paths, which was the last shape gap I raised:

```
non-streaming: finish_reason 'stop'      (was 'stop')
streaming:     finish_reason 'stop'      (was <absent>)
```

Model attribution agrees across both paths (`claude-sonnet-4-6`, the served model, not the configured `anthropic/claude-sonnet-4-6`), and the cache fields are present and zero rather than missing.

## On the import crash

Worth recording that the `test_py_infra.jac::by operator` failure was partly my doing. Hoisting the signature check to a module-level `glob` was my suggestion, and it turned a litellm-only, per-call attribute access into something every byLLM import has to survive - which the `_optdeps` `SimpleNamespace` fallback does not provide. Combined with the type-checker fix moving that access outside the `try`, import raised `AttributeError` before the capability check could produce its message.

You found and fixed it before I reported it, and the regression test patching `litellm` with a bare `SimpleNamespace` is exactly the right shape for it. The hoist itself was still worth doing; I just did not think through what moving it to import time implied.

## Deliberately deferred

Two structural things I raised and then argued against fixing here, recorded so they are not lost:

- The `_usage_base` snapshot/re-arm bookkeeping in `_invoke_streaming` - eight touch points for one invariant, where a future `break` path that forgets to flush drops a `usage_step` silently rather than failing.
- The `_call_kind` save/set/restore ceremony duplicated across `_force_final_answer`, `_stream_final_answer` and `_attempt_recovery`, where a fourth special call site that forgets it mislabels its usage instead of erroring.

Both are correctness-by-remembering rather than correctness-by-construction. Neither is a bug today and both would mean touching control flow that was just verified, so a follow-up issue is the right home.

## Scope of what I checked

Everything live was Anthropic. The OpenAI nested `prompt_tokens_details.cached_tokens` shape I only saw pass in unit tests, and I confirmed `ModelPool` fallback reads `chunk?.model` by inspection without driving an actual router fallback. Locally the byLLM suite is 249 passed with 3 `test_mtir_integration` failures that reproduce on `main` and vary run to run - CI's sealed build is the authority there, and it is green across 27 jobs.

Nice work on the root-causing throughout, and on catching the import regression yourself.

PR #8497, opened 2026-08-22
fix(parser): name the parenthesized fixup when bare tuple targets hit 'for ... in'

chess10kp:
## Problem

`for n, c in d.items()` fails with the generic parse error:

```
error[E0001]: Expected 'in', got ','
```

which gives no hint that Jac requires parenthesized tuple targets.

## Fix

Keep the grammar (tuple targets are consistently parenthesized across for statements, comprehensions, and assignments — `(a, b) = ...` — and py2jac already emits the parenthesized form) and improve the diagnostic:

- New `Parser.parse_in_target()` helper used by both `parse_for_stmt` and `parse_compr_clause`.
- On a bare tuple target it emits E0001 with message `tuple targets must be parenthesized` plus a help line, e.g.:
  ```
  error[E0001]: tuple targets must be parenthesized
    --> t.jac:3:10
      |     for n, c in d.items() {
      |          ^
  help: wrap the variables in parentheses, e.g. `for (n, c) in items { ... }`
  ```
- Resynchronizes at the `in` keyword so parsing continues without cascading errors.

## Tests

4 new cases in `test_new_diagnostics.jac`: for-statement target (message + help asserted), comprehension target, triple-target resync (exactly one diagnostic), and a negative check that parenthesized forms don't fire.


christianwilkins (commented):
Reviewed head `598d0b6d096c75aa6b09f980102814e01a063049`: all five changed files, the shared target parser, comprehension/for callers and recovery boundaries. No blocking parser defect found. The helper reports at the comma and stops at the in/= or statement/block delimiters instead of consuming the remainder of the file.

Independent local validation: all six added parser cases passed through the production parser, including mapping-loop and comprehension help text, one diagnostic for triple targets, valid parenthesized targets, numeric-for recovery at =, and malformed-target recovery at the block boundary. The direct harness allows parser exceptions to propagate rather than swallowing them. No source changes made. Current-head CI is green (23 successes, 5 skips).

Not ready to merge because GitHub reports conflicts with main. The author needs to resolve those conflicts, preserve these six cases and rerun them plus current-head CI; maintainers can review the resulting head. The description still says four tests and only in-keyword resynchronization, so it should also reflect the two later numeric/block-boundary cases.

PR #8537, opened 2026-08-22
feat(bun-host): BunHost sv services — host axis, bun artifact server, shared migration protocol

### LINKED ISSUE #8144: Name the host axis: one JS codespace, N hosts (browser, bun, RN, edge) — clean break, no compat shims
Jac has three codespaces (`sv`/`cl`/`na`) and selection is per-node via `code_context`. That model quietly assumes **one host per codespace**: `cl` means browser, `sv` means CPython, `na` means bare metal. The assumption is already false — [`targets/registry.jac`](https://github.com/jaseci-labs/jac/blob/main/jac/jaclang/runtimelib/client/targets/registry.jac) registers **seven** things that execute `cl` output (web, pwa, static, mobile, react-native, desktop-native, desktop-cef), and React Native has no DOM. It becomes untenable the moment we run a Jac service on the bundled bun runtime: bun is a JS host that is neither a browser nor pretending to be one.

Because the host axis has no name, host-varying facts get approximated by codespace:

- An `NPM` requirement seeds `CodeContext.CLIENT` (`placement_solver.jac:432`) — a **server** module that imports Stripe gets relocated into the **browser**.
- `js_globals.pyi` is gated on `expr.in_client_context()` (`type_evaluator.impl.jac:1895-1899`), while `dom_types.pyi` names merge **unconditionally into builtins** (`type_evaluator.impl.jac:3653`). `document` is currently *more* ambient than `JSON`, and nothing in the type layer knows RN has no DOM.
- `resolve_active_backend()` reads one project-global `cfg.client.framework` — one JS framework per project, one host implied.

The fix is not a fourth codespace. It is to **name the axis that already exists**.

---

## What is already true (and makes this cheap)

- The ES backend is complete and **is not shaped around component trees**. `FrameworkBackend` is a pluggable lowering layer consulted only for JSX/reactive nodes; a module with no JSX already compiles to plain ESM — that is exactly what `jac publish --npm` does today, with `.d.ts` emission.
- The OSP kernel (`osp_kernel.jac` + `osp_graph.jac`) compiles to JS from the *same source* that produces the Python and LLVM versions. `osp_spawn` / `osp_visit` / `osp_report` / `osp_disengage` **already run in JavaScript**, pinned by `test_kernel_units.jac`.
- `is_service_import` is config-driven (`[scale.microservices.routes]`, `service_cut.jac`) and placement-agnostic, and **both** `PyastGenPass` and `EsastGenPass` emit RPC stubs for it. A JS-placed service provider needs **zero** caller changes.
- bun already executes compiled Jac JS as a subprocess today (`cl_test_runner.impl.jac`), assembling runtime + npm stubs + transitive local imports, with a live Jac server alongside for `cl↔sv` RPC.
- npm packages are already type-checked from `.d.ts` via `ts_decl_parser.jac`.

The pieces are in the tree. What is missing is a name.

---

## Target design (converged)

Three concepts, cleanly separated:

- **Codespace** — what you compile *to*: Python AST, ESTree, LLVM IR. There are three. **Bun adds zero.** There is no fourth `CodeContext`.
- **Host** — what *executes* the artifact: browser, bun, RN bridge, webview, CPython, bare metal. Promote `ClientTarget` into a first-class `Host` that the compiler can see.
- **Capabilities attach to hosts, not codespaces.** `DOM` is a *browser* capability, not a JS one. `NPM` is a `cl`-codespace capability (all JS hosts). `PY` is a *CPython-host* capability. `PERSIST` (root/graph) is a capability of whichever host has a `Store`.

Placement stops being a chain of hardcoded arms and becomes a table lookup — requirement → hosts providing it → codespace those hosts run:

- `NPM ⇒ {browser, bun, rn, …} ⇒ cl`
- `DOM ⇒ {browser, webview} ⇒ cl`
- `PY ⇒ {cpython} ⇒ sv`
- `FFI ⇒ {bare-metal} ⇒ na`

A module needing `NPM` but **not** `DOM` is now placeable on either JS host. That fact **emerges** instead of being special-cased — it is the entire feature.

**Host is a deployment fact, not a compiler flag.** The same module may be browser-hosted in one app and bun-hosted in another. `[build] default_codespace` gains **no** new value.

The naming matters: a bun service is not "`sv` running on bun." It is a **`cl`-codespace module with a server host**. Spelled the other way — `sv` gets a second backend — every `is_server` getter, every diagnostic, and every Python-interop assumption has to learn which server it is talking about.

---

## Hard removals (clean break, no compat shims)

This refactor makes the following dead **by construction**. Delete, do not deprecate, do not leave a fallback path:

1. **`_seed_from_summary`'s `if cap == 'FFI' / 'DOM' / 'NPM'` arm chain** (`placement_solver.jac:420-442`) → provider table.
2. **`PLACEMENT_SV_SIGNAL_NODES`** — the hardcoded `(TypedCtxBlock, PyInlineCode, JsxElement)` "these mean server" tuple. `JsxElement` is a `DOM` requirement; `PyInlineCode` is a `PY` requirement. Express both as requirements; delete the tuple.
3. **`PLACEMENT_NATIVE_REJECT_NODES: tuple = ()`** — empty placeholder sitting next to it, never populated.
4. **The unconditional `dom_types` → builtins merge** (`type_evaluator.impl.jac:3653`). DOM names become host-gated like every other capability. This is the bug-shaped one.
5. **`expr.in_client_context()` as the `js_globals` gate** → `host_provides(expr, JS_GLOBALS)`.
6. **`resolve_active_backend()`'s global `cfg.client.framework` read** and the framework-name-keyed `_BACKENDS` singleton → per-module host+backend resolution.
7. **`TargetType` enum + `get_target_type()`'s if-chain** — a closed enum of hosts duplicating the open registry. Targets declare their own name; the registry becomes the only authority. Leaving this in is precisely the "two-element enum with extra steps" failure mode this design exists to avoid.
8. **`prepare_inprocess_runtime`'s `transport_kind` reservation** raising `NotImplementedError` for `'asgi'` / `'isolated_worker'` (`scale/server/inprocess.jac`). Execution isolation *is* the host. Delete the reserved strings.
9. **The two `sv` RPC clients.** `JacServe.sv_service_call` / `sv_walker_call` (`jac0core/impl/runtime.impl.jac:1249-1335`) has no retry, no auth forwarding, no trace headers; `scale/runtime/rpc/rpc.jac` has all three. Which one runs depends on whether the scale plugin loaded. Collapse to one transport and delete the core fallback.
10. **`sv_client._test_clients` special-casing** threaded through both dispatch paths — a test seam baked into production dispatch. Ship a test *host* instead.

---

### P0 — Host as a first-class concept (no bun yet)

- `Host` protocol + registry, absorbing `ClientTarget`; capability sets per host.
- Placement provider table replaces the arm chain (removals 1–3).
- Type layer: host-gated ambient stubs (removals 4–5).
- Per-module host + backend resolution (removal 6).

**Falsifiable check — React Native.** RN is already in the registry and already DOM-less. If the abstraction is right, RN's DOM looseness resolves as a *side effect*, with zero RN-specific code in the type layer. If bun and RN each need special-casing, **stop** — the abstraction is wrong and we have built a two-element enum with extra steps.

### P1 — `BunHost`, stateless tier

- `BunHost` as a fourth implementation of the interface that already has three: entry script is a `Bun.serve` bootstrap instead of `createRoot`; runtime module is the server shim instead of react.
- Server entry emitter consuming the same walker/function manifest `eject_targets/backend.jac` already consumes. Pure manifest consumer — `test_backend_purity.jac` forbids backends growing analysis.
- JS server runtime: envelope, the **callee** half of the wire format (`_deserialize_wire_args` / `_finalize_call_response` `api_mode` graph metadata), `@restspec` paths, `/functions` health route. The caller half already exists (`__to_wire` / `__from_wire`).
- `ensure_sv_service` branches on host: `bun <artifact>.mjs` instead of `jac start`, in both the loopback path (`jac0core/impl/runtime.impl.jac`) and the scale path (`scale/plugin.jac`); k8s image gains bun + `node_modules`.
- Server npm root + `jac install --npm --server` through the existing `EcosystemProvider` registry.
- **Scope:** functions + `pub` walkers over an ephemeral graph. No persistent `root`. Gated by the existing `module_has_root_access()` fact (`placement.jac`), which already caches on `mod.gen.has_root_access`.

### P2 — `PERSIST` capability

JS `Store` over `Bun.sql` against the same Postgres schema; anchors/jids, sv walk-stack context, access control. `PERSIST` becomes a bun capability and root-touching modules become placeable. No redesign required — it is a new row in the table.

### P3 — the seam pays out

Web workers, service workers, edge/Cloudflare functions, CLI-JS. All are "`cl` codespace, non-browser host." These should land as registry rows, not as features.

---

## Known risk: the JS stdlib gap

`primitives_es` covers the language surface (int/str/list/dict/set/bytes/tuple/range/builtins), verified by cross-backend equivalence tests. The **module** stdlib is not there: `json`, `datetime`, `math`, `os`, `random`, `hashlib`, `uuid`, `time`, `urllib` fixtures all carry "ES/JS opts out", and `E5017` is `blocks_codegen=True` — a hard stop, not a warning.

Under opt-in hosting this is a constraint on a service you chose. It must **not** become a default. Sequence stdlib coverage by what real services actually import, and keep bun hosting opt-in until that lands.

---

## Acceptance criteria

- `CodeContext` still has exactly **three** members.
- A module that imports an npm package and touches no DOM places on a JS host **without a pin**, and is servable under bun.
- React Native typechecks DOM-free with **no RN-specific code** in the type layer.
- A Python `sv` caller and a bun `sv` caller both reach a bun-hosted provider through the existing `is_service_import` path, with no backend-specific branches.
- Every item in **Hard removals** is gone from the tree — verified by a grep check in CI, not by review.
- One RPC client. One dispatch path. No `_test_clients` branch in production code.

## Non-goals

- Making `sv` mean anything other than "the Python codespace."
- `[build] default_codespace = "js"`. A project-*kind* default (`KindSpec.codespace`, already used by `js-package`) covers the ergonomic case without an ambient global that surprises mixed full-stack projects. Revisit only after P2.

chess10kp:
Closes #8144

## Summary

Bun-hosted microservice routes and standalone server modules reach parity with the cpython host:

- **Host as a first-class axis**: codespace (where code is placed/emitted) is separated from host (which runtime executes it). `jaclang.jac0core.host` owns the capability table and target-to-host aliases.
- **BunHost sv services**: the builder emits a `Bun.serve` artifact at `.jac/server/<module>.mjs`. Stateless `def:pub` endpoints serve over HTTP with cpython-matching status envelopes (200/500); `POST /walker/<name>` dispatches pub walkers through the OSP kernel, and root-touching walkers persist through **BunStore** (Postgres via `Bun.sql`).
- **One migration protocol**: BunStore computes PgStore's exact schema fingerprint, bakes `[database] indexes` promotions at build time, applies the missing-`seq` migration, and takes the same advisory schema-election lock - alternating hosts on one database no longer invalidate each other's stamps.
- **Endpoint visibility**: only `pub` walkers/functions become HTTP surface; non-pub symbols stay internally importable. Single enforcement point at discovery.
- **No silent fallbacks**: unknown explicit host/target config raises instead of executing on another host; per-service `host` key documented in `[scale.microservices.services]`.
- **Fix**: `[root-->][?:Item]` graph filters inside client-codespace walker bodies lowered to nothing; now lower to real `.filter(x => __ospIsA(...))`.

## Review history

Five review rounds are documented in `TODO.md` (round 6 has the final disposition table): endpoint-gating, envelope statuses, schema-stamp/advisory-lock parity with a cross-runtime drift test, string-bundler fail-loud hardening, placement summary validation, stable port derivation, cache invalidation, and named return types.

## Tests

- `test_bun_walker_e2e`: walker descriptors survive `Bun.build`; pub walkers execute over HTTP and persist NodeAnchors via BunStore
- `test_bun_stateless_e2e_p2` / `test_bun_persist_e2e_p2`: function + persistence e2e
- `test_bun_stdlib_e2e`: datetime/hashlib/uuid/json compute real results over HTTP
- `test_bun_string_bundler`: fallback bundler unit tests
- `test_host_p0` / `test_host_type_gating_p0` / `test_host_bun_stdlib_p2`: host axis acceptance
- `test_sv_spawn_health_budget` (3/3) and `test_sv_auth_forward` (32/32) verified green locally


MusabMahmoodh (commented):
Reviewing the seven scale files only - the Bun host axis itself is not my lane and I am not going to pretend to have read 6k lines of it.

The scale seams are good and worth saying so specifically: `rpc.jac` collapses two direct reaches into `sv_client._test_clients[module_name]` into a single `_test_adapter()` resolver, so there is one seam instead of two call sites indexing a module-level dict, and the `host` key at `plugin_config.jac:588` documents "unrecognized values raise at service startup instead of silently falling back" - that is the right default and it is nice to see it written into the schema description rather than left to the implementation.

One thing to fix, on the path every existing user takes. In `plugin.jac`, `port=pick_free_port(module_name)` is evaluated eagerly as an argument to `dispatch_sv_service(...)`, and when that returns `None` the next line calls `pick_free_port(module_name)` again. `pick_free_port` (`scale/runtime/context/util.jac:5-21`) is a bind/close retry loop, so the cpython path - the default for every service that is not Bun - now runs that loop twice on the boot path and throws the first result away. Hoisting it to a local above the `dispatch_sv_service` call and passing the local fixes it.

Also flagging that CI is red on seven checks here (build-kit, jac-check, Contribution Checks, test-compiler toplevel, test-native, test-runtime, test-client), so this is not just the #8583 fork break that is hitting everyone.

One retraction so it does not sit in your head: I went looking for whether dropping `transport_kind` from `prepare_inprocess_runtime` orphaned the caller at `desktop/native/inprocess_dispatch.jac:90` - which `.jacignore` would have hidden from `jac check` - and it does not, you updated that site and `scale_test_client.jac:74` with it.

PR #8585, opened 2026-08-23
fix: bare import degrades to syntax diagnostic instead of IndexError

### LINKED ISSUE #8534: [Bug] Bare `import` (empty module list) escapes as IndexError instead of a syntax diagnostic
### What's wrong

A file containing only `import` (6 bytes) crashes the full-check path with an unhandled `IndexError`:

```
✖ Error: Error checking 'repro.jac': list index out of range
```

via the generic exception wrapper in `analysis.impl.jac`, instead of a diagnostic. The released binary's parse path degrades gracefully to a proper syntax error (`E0002`), so this is specifically the full-check handling of a bare `import` token.

### Repro

```bash
printf 'import' > repro.jac
jac check repro.jac    # internal "list index out of range" instead of E0002-style diagnostic
```

### Expected

The same graceful syntax diagnostic the parser produces elsewhere (`E0002: Missing ';'` or an incomplete-import error with location).

Found by differential fuzzing (parser boundary sweep); minimized to 6 bytes, double-confirmed on main @ `844ea3ea5`.

chess10kp:
Closes #8534

Parse-error recovery deliberately keeps an `Import` node whose module path is empty (a file containing only the bare token `import`). Every later pass funnels that node's specifier through `resolve_module`, which indexed into the empty parts list and surfaced as:

```
Error checking 'x.jac': list index out of range
```

Now empty targets/parts yield a clean unresolved outcome (`('', 'other')`), so the user gets the parser's syntax diagnostic instead of an internal crash. Adds a regression test compiling a bare-import file and asserting no internal error.


jac/tests/compiler/backends/native/fixtures/clib_logical_import.jac
```diff
@@ -7,11 +7,10 @@ extensionless surface instead of an explicit string path.
 """
 
 import from raylib {
-    def InitWindow(width: i32, height: i32, title: str) -> None;
-    def CloseWindow -> None;
+    def InitWindow(width: i32, height: i32, title: str) -> None;  def CloseWindow -> None;
 }
 
-def open_then_close(w: int, h: int, title: str) -> None {
+def open_then_close(w: int, h: int, title: str) {
```
christianwilkins, line 13: [P2] Preserve the void wrapper ABI. Removing `-> None` changes native output from `define void @"open_then_close"` to `define i64 @"open_then_close"`. I confirmed that on this head using two otherwise identical AOT fixture copies; restoring only the annotation restores void. The existing clib-extensionless-import regression asserts the void definition and currently fails in CI with "wrapper not defined". Keep the explicit annotation (or otherwise preserve its intended ABI); this unrelated edit is not formatting-only.

christianwilkins (changes requested):
Reviewed head `a5422f69092012343e66bb2aeb5e255495d15f52`: all six changed files, empty-target resolution, relative/search-path fallback and the affected native C-import test. The two new empty-import cases passed locally through the production APIs.

P2: the unrelated clib_logical_import.jac edit removes the wrapper return annotation and breaks an existing native regression. On this head, native AOT compilation emits `define i64 @"open_then_close"`; an otherwise identical disposable copy with only `-> None` restored emits `define void @"open_then_close"`. The existing test explicitly checks for the void definition, and current-head CI fails that exact assertion with "wrapper not defined". Restore the explicit return annotation/preserve the original ABI rather than treating this as a formatting-only change.

Local validation: bare-import compilation produced syntax diagnostics without an internal exception; an empty module specifier stayed unresolved. Both native fixture variants compiled without errors and confirmed the return-type difference above. All probes used this checkout source; no repository source changes made.

Current-head CI reports 1515 passed, 2 failed and 4 skipped in the native lane. Besides the confirmed fixture regression, the set-scaling test fails during compilation because symbol_utils is re-entered during import; it never reaches its timing comparison. I have not established that second failure as a baseline. GitHub also reports merge conflicts.

Changes requested: restore the fixture ABI, resolve conflicts, investigate or establish a baseline for the remaining CI failure and run CI on the resulting head. Maintainer merge review follows.

PR #8593, opened 2026-08-23
fix: if/else dual assignment without read now emits W2003 unused-variable

### LINKED ISSUE #8535: [Bug] W2003 unused-variable not emitted for if/else dual assignment (ternary equivalent warns)
### What's wrong

W2003 ("defined but never used") is not emitted when an unused variable IS assigned in both branches of an if/else — but the semantically equivalent ternary form warns correctly. Found via a semantic-equivalence oracle comparing diagnostics across syntax variants that must agree.

```jac
# variant A: NO warning (bug)          # variant B: W2003 fires (correct)
if c {                                 x = 1 if c else 2;
    x = 1;
} else {
    x = 2;
}
# x unused below in both cases
```

Single-branch if and while-loop assignment forms also warn; only both-branches-if/else suppresses it.

### Expected

Both variants emit W2003 for `x` (or neither, but then ternary should change too) — currently inconsistent.

Environment: jac 0.36.1 and main @ `844ea3ea5`. Minimized + double-confirmed.

chess10kp:
Closes #8535

A variable assigned in both branches of an `if/else` and never read escaped the W2003 unused-variable warning, while the semantically equivalent ternary warned correctly. Root cause: assignment targets are recorded in the symbol table as uses, so the old `len(sym.uses) == 0` check saw the dead variable as used.

The check now requires a **genuine read**: a use is ignored when it is a plain assignment target (possibly nested in a list/tuple literal); augmented assignment still implies a read. Six regression tests pin the contract across if/else dual, ternary, sequential reassignment, while-loop, read-after-assign, and aug-only variants.


jac/tests/compiler/backends/native/fixtures/clib_logical_import.jac
```diff
@@ -7,11 +7,10 @@ extensionless surface instead of an explicit string path.
 """
 
 import from raylib {
-    def InitWindow(width: i32, height: i32, title: str) -> None;
-    def CloseWindow -> None;
+    def InitWindow(width: i32, height: i32, title: str) -> None;  def CloseWindow -> None;
 }
 
-def open_then_close(w: int, h: int, title: str) -> None {
+def open_then_close(w: int, h: int, title: str) {
```
christianwilkins, line 13: [P2] Restore the explicit void return on this unrelated fixture. On this head I compiled the current source and got `define i64 @"open_then_close"`; restoring only `-> None` in the in-memory source produced `define void @"open_then_close"`, with zero errors in both cases. The existing test requires the void wrapper and is the sole native/tools failure on this head. The W2003 fix can remain unchanged.

christianwilkins (changes requested):
Reviewed `dbba1e2213ba3fd8855b778b033bade99d85c4bb`: complete diff, unused-symbol filtering and write-target traversal, the new tests, existing conversation, and checks for this head. I ran all six new regression bodies through the production compiler on this head; they passed. The diagnostic change retains genuine reads and augmented-assignment reads while ignoring ordinary assignment targets. `git diff --check` also passed.

P2: the full PR diff also removes `-> None` from the unrelated `open_then_close` fixture. I checked this separately from the W2003 change: with the current-head fixture, production native compilation emits `define i64 @"open_then_close"`; with only the annotation restored in the in-memory source, the same compiler/options emit `define void @"open_then_close"`. Both compilations have zero errors. This explains the existing logical-import test's exact-head “wrapper not defined” failure and does not require reverting the static-analysis fix. Please restore the fixture signature.

Current-head workflow 33571274323 reports native/tools: 1,516 passed, 1 failed, 4 skipped. The prior conversation's baseline analysis focused on the diagnostic path; the independent fixture delta is the actionable difference here. I did not rerun the full native suite. GitHub reports no conflicts. No source files were changed; the annotation control was supplied only to the compiler probe. Changes requested: restore the unrelated signature and obtain passing CI on the resulting head.

PR #8602, opened 2026-08-23
fix: embedded-postgres initdb retries once and wipes partial clusters

chess10kp:
Found while running two concurrent automation workloads with separate `JAC_CACHE_HOME` caches on one machine: both cold `initdb` runs contended for disk, blew the hardcoded `timeout=120`, and each left a **partial cluster** behind — after which every subsequent attempt fails permanently because initdb refuses non-empty data dirs.

- Retry provision once; between attempts the partial data dir is wiped so the retry starts from a clean slate
- Timeout tunable via `JAC_PG_INITDB_TIMEOUT` (default unchanged at 120s)
- Regression tests cover timeout-then-success and exhausted-failure-leaves-clean-slate paths

Closes the resilience gap observed in concurrent multi-project setups.


jac/jaclang/data/impl/pgembed.impl.jac
```diff
@@ -808,8 +839,16 @@ impl PgRuntime._sweep_on_start -> None {
...
 
 impl PgRuntime.serve -> int {
-    bin_dir = self._resolve_binaries();
-    self._initdb_if_needed(bin_dir);
+    import fcntl;
+    os.makedirs(os.path.dirname(self.data_dir) or '.', exist_ok=True);
+    lock_f = open(self.data_dir + '.lock', 'w');
+    fcntl.flock(lock_f.fileno(), fcntl.LOCK_EX);
+    try {
+        self._initdb_if_needed(bin_dir);
```
christianwilkins, line 847: [P1] Restore binary resolution before using bin_dir. The change removes the original `bin_dir = self._resolve_binaries()` assignment, and serve has no parameter or other binding for this name. This call and the later postgres argv now fail instead of starting the foreground server. Restore the assignment in the locked startup flow and add coverage that invokes serve.

jac/jaclang/data/impl/pgembed.impl.jac
```diff
@@ -582,25 +582,56 @@ impl PgRuntime._initdb_if_needed(bin_dir: str) -> None {
...
+                return;
+            }
+            last_err = res.stderr.strip() or f"initdb exited {res.returncode}";
+        } except Exception as e {
+            last_err = str(e);
+        }
+        logger.warning(
+            f"initdb attempt {attempt + 1}/2 failed, " f"resetting data dir: {last_err}"
+        );
+        if os.path.isfile(os.path.join(self.data_dir, 'PG_VERSION')) {
```
christianwilkins, line 625: [P2] Do not treat PG_VERSION as proof the failed initdb completed. PostgreSQL writes that file before bootstrap/configuration ([source](https://github.com/postgres/postgres/blob/86f7c82cf1023e3599f40f939727791a7090cd44/src/bin/initdb/initdb.c#L3095)); a timeout can leave the current attempt’s marker in a partial cluster. Returning here then skips the retry and reports success with unusable data. The new mock test creates only that marker, so it exercises the wrong completion condition. Under the shared provision lock, distinguish the failed attempt’s partial output from a successfully initialized cluster and cover timeout-after-marker creation.

christianwilkins (changes requested):
Reviewed `d67589ee864b8bf908b6de58f60412b4dedbd46b`: full diff, provisioning/startup and locking context, all three new mocked tests, existing discussion, and current-head CI.

P1: `serve()` deletes `bin_dir = self._resolve_binaries()` but still passes `bin_dir` to `_initdb_if_needed` and joins it with `postgres`. No local, parameter, or module binding supplies that name, so the foreground server path fails before starting Postgres. Restore binary resolution inside the intended startup flow and cover serve(), not only the private retry method.

P2: PG_VERSION is not a completion marker for initdb. PostgreSQL writes it before configuration/bootstrap; see [initialize_data_directory](https://github.com/postgres/postgres/blob/86f7c82cf1023e3599f40f939727791a7090cd44/src/bin/initdb/initdb.c#L3095). A timed-out attempt can therefore leave PG_VERSION in its own partial directory, and this new failure branch returns success immediately instead of retrying or reporting failure. The new race test creates only PG_VERSION and calls that a completed cluster, so it codifies this false success. Both ensure() and serve() use the same provisioning lock here; another cooperating initializer cannot finish inside this critical section. Distinguish successful provisioning from a marker left by the failed attempt and test a timeout after marker creation.

Local validation was static inspection, the upstream initdb source check, and a passing `git diff --check`; I did not start or delete any database. Workflow 33693575758 fails formatting on the new test file (`jac fmt --lintfix jac/tests/runtimelib/test_pg_initdb_retry.jac`) and a separate CLI diagnostic test (runtime: 1,782 passed, 1 failed, 8 skipped). I have not attributed that latter CLI failure to this patch. GitHub reports no conflicts. No code changes were made. Changes requested: restore serve's binary resolution, fix partial-cluster completion detection, format the tests, and obtain passing CI.

PR #8616, opened 2026-08-23
fix: Module-level docstring before an import: accepted by jac0 Parser but rejected by full pipeline as E0005 'Unexpected token'

### LINKED ISSUE #8541: Module-level docstring before an import: accepted by jac0 Parser but rejected by full pipeline as E0005 'Unexpected token'
### Summary

A bare-string docstring is only legal when it *directly precedes a `def`*. Placed before an `import` (or other statement), the raw `jaclang.jac0` parser accepts the module, but the full compiler pipeline rejects it with a generic `E0005: Unexpected token`. Two problems:

1. **Inconsistent acceptance** — jac0 parses it; the real pipeline doesn't.
2. **Unactionable error** — `Unexpected token '"""..."""'` gives no hint that docstring placement is the issue.

### Repro

```jac
# pkg/user.jac
def make() -> int {
    return 1;
}

"""Helpers for registration."""
import from base_mod { Thing }
```

```
$ jac run main.jac
✖ Error: error[E0005]: Unexpected token '"""Helpers for registration."""'
  --> pkg/user.jac:5:1
    3 | }
    4 |
    5 | """Helpers for registration."""
      | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    6 | import from base_mod { Thing }
```

Note: a docstring at the very top of the module followed by imports works fine, and docstrings directly before defs work fine — only the in-between placement fails. This bites naturally when appending registry/section blocks to the end of an existing module.

Observed on jaclang 0.36.1, Linux.

### Expected

Either accept section docstrings anywhere at module level, or emit a targeted diagnostic like "docstrings must directly precede a def" instead of E0005.

chess10kp:
Closes #8541


jac/jaclang/compiler/frontend/parser/impl/parser.impl.jac
```diff
@@ -3519,6 +3521,11 @@ impl Parser.parse_docstring_target -> ElementStmt | None {
         self._attach_doc(doc_mc, elem_doc);
         return doc_mc;
     }
+    if self.check_any(TokenKind.KW_IMPORT, TokenKind.KW_INCLUDE) {
+        doc_import = self.parse_import_stmt();
```
christianwilkins, line 3525: [P2] Update the generated grammar spec with this new branch. Current-head `grammar matches spec` fails because `docstring_target` now includes `import_stmt`, while `jac/jaclang/jac.spec` still records the old alternatives. Regenerate and inspect the spec alongside this parser change, then rerun the grammar snapshot check.

christianwilkins (changes requested):
Reviewed `a938aeab577e8a8faef35bee61c5e07cbc67f58e`: complete diff, parser dispatch/doc attachment, import/module formatting, regression cases, existing conversation, and current-head CI. The parser now attaches section strings to import/include nodes through the existing doc helper, and the formatter inserts a hard line after that doc. The tests distinguish the leading module doc from later import docs and separately parse an include. Local validation was static inspection and a passing `git diff --check`; I did not independently run parser/formatter suites.

P2: this grammar change needs its generated spec update. Current-head workflow 33577073375 fails `grammar matches spec`: `docstring_target` now includes `import_stmt`, but the checked-in `jac/jaclang/jac.spec` still omits it. Generate the spec from this head (`jac tool grammar -o jaclang/jac.spec` from the jac package directory), inspect the output, and include the intentional rule change. Native/tools reports 1,516 passed, 1 failed, 4 skipped.

The separate jac-check failure is E1032 accessing `vars` in dist/publish/dts.jac (915 passed, 1 failed); I have not established it as introduced by the parser change. GitHub reports no conflicts. No code changes were made. Changes requested: synchronize the grammar spec and obtain passing checks on the resulting head.

PR #8617, opened 2026-08-23
fix(optdeps): the fallback shim declares its class instead of rebinding the imported name

marsninja:
## What

`jac-check` has been failing on every push to `main` since type checking joined the critical path. The full-repo sweep reports two `E1001`s:

```
error[E1001]: Cannot assign object to IdGenerator
  --> jac/jaclang/scale/_optdeps/opentelemetry.jac:34:9
error[E1001]: Cannot assign _PillowUnavailable to Image
  --> jac/jaclang/byllm/_optdeps/pillow.jac:7:9
```

Both are optional-dependency shims whose `except ImportError` arm rebinds a name the `try` arm imported. The checker takes the import as the name's declared type, so `IdGenerator = object;` and `PILImageCls = _PillowUnavailable;` are narrowing violations.

The errors only appear when the real package resolves, which is the push-time full sweep. A PR's changed-file scope never touched these files, so nothing caught it at merge time.

## Fix

Declare the fallback as a class of the bound name, the same shape the litellm shim already uses for `LiteLLMMessage`:

```jac
} except ImportError {
    class IdGenerator {}
    HAS_OTEL = False;
}
```

No assignment, so no second declared type. The fallback is still a real class, so `class _JacIdGenerator(IdGenerator)` in `scale/runtime/context/tracing.jac` and `isinstance(value, PILImageCls)` in `byllm/types.impl/media.impl.jac` behave exactly as before when the package is absent.

## Verification

With `opentelemetry-sdk` and `pillow` installed locally so the try arm resolves:

- `jac check` on both shims: was 2 failed, now 2 passed.
- `jac check` on the consumers `byllm/types.jac` and `byllm/types.impl/media.impl.jac`: pass. (`scale/runtime/context/tracing.jac` is in `.jacignore`.)
- `jac fmt --check --lintfix` on both files: clean.
- Fallback path exercised against an unresolvable import: the class binds at module level, imports across modules, subclasses, and `isinstance` returns False as before.


MusabMahmoodh (approved):
Checked the consumers before agreeing: there is no `isinstance(..., IdGenerator)` anywhere in the tree, which is the one thing that would have made `object` -> empty class a behavior change rather than a declaration change. `_JacIdGenerator(IdGenerator)` subclasses fine either way, and `install_tracing` returns on `if not HAS_OTEL` before the only construction site, so the fallback never reaches a TracerProvider. Same for pillow: `isinstance(value, PILImageCls)` at `media.impl.jac:38` was always-False against `_PillowUnavailable` and stays always-False. Approving.

One question about the other twelve. `scale/_optdeps/opentelemetry.jac:21-32` rebinds `trace`, `SpanContext`, `NonRecordingSpan`, `SpanKind`, `TraceFlags`, `set_span_in_context`, `Resource`, `TracerProvider`, `BatchSpanProcessor`, `ParentBased`, `TraceIdRatioBased` and `OTLPSpanExporter` to `None` in the same arm - structurally the same "assign over the name the try arm imported" shape that E1001 just rejected for `IdGenerator`. Is `None` deliberately assignable to any declared type, or is it just not flagged yet? If it is the second one, the same sweep comes back for these twelve the next time the checker tightens, and it would be cheaper to know now.

PR #8620, opened 2026-08-23
fix(scale): the k8s e2e fixture's client uses Jac idioms, not raw JS ones

marsninja:
## What

Both nightly `microservice e2e (k8s, ...)` jobs fail before they reach the cluster:

```
deploy failed: manifest generation failed: client build failed (main.jac, exit 1).
RuntimeError: Failed to compile .../k8s_e2e/main.jac:
  frontend.jac, line 241: Type "dict" has no attribute "total"
  frontend.jac, line 261: Cannot access attribute "length" for type "<function items(self: dict) -> dict_items[<any>, <any>]>"
  frontend.jac, line 294: Type "list" has no attribute "length"
  frontend.impl.jac, line 53: Cannot access attribute "error" for type "Any | dict"; attribute is missing from dict
  frontend.impl.jac, line 54: (same)
  frontend.impl.jac, line 70: Type "str" has no attribute "trim"
  frontend.impl.jac, line 88: (same)
  frontend.impl.jac, line 92: Type "str" has no attribute "length"
```

The jac-shop fixture's client code was written with raw JS idioms (`.length`, `.trim()`, dot-reading a `dict` as a JS object). That was tolerated while client-lane type checking was advisory; it is not any more.

The fixture lives under `tests/fixtures/`, so the `jac check` sweep skips it and only the e2e's real client build catches this.

## Fix

Fixture only:

- `.length` becomes `len()`, `.trim()` becomes `.strip()`. The ecmascript lane lowers both back to `.length` / `.trim` in the emitted JS (`ESStrEmitter.emit_strip`, `builtin.len`), so the bundle is unchanged.
- `cart` and `order` hold walker payloads the fixture reads by attribute, so they are typed `any` instead of `dict`. That also removes the accidental resolution of `cart.items` to dict's own `items` method, which is what the `<function items(...)>` error is reporting.

## Verification

`jac build main.jac --client web` in the fixture directory:

- before: exits 1 with the eight errors above
- after: `✔ Client bundle built (2.3s)` / `✔ Build complete: .jac/client/dist/client.*.js`

The e2e's later k8s stages need a cluster and are not exercised here.


MusabMahmoodh (approved):
The fix is right and unblocks the nightly, so approving. But the reason the fixture escaped is not the one in the description, and the real one is worth a follow-up because it is still live.

`.jacignore:112` is a bare `frontend.jac`, and `path_matches_exclude` in `jac/jaclang/jac0core/helpers.jac:220-231` matches a pattern with no `/` against every path component:

```
} else {
    for part in parts {
        if fnmatch(part, pattern) { return True; }
```

So that single line excludes all nine files named `frontend.jac` in the repo, wherever they sit - not just this fixture. `find . -name frontend.jac` gives, among others:

- `jac/jaclang/runtimelib/client/templates/web-app/frontend.jac`, the template every `jac create --kind web-app` starts from
- `jac/jaclang/jac0core/parser/frontend.jac`, which has nothing to do with client code
- `jac/examples/littleX/frontend.jac` and the three byllm agentic examples

Which means moving the fixture out of `tests/fixtures/` would not get it checked, and the shipped web-app template is unchecked for exactly the same reason - so the next `.length` lands there and nothing catches it either. Path-qualifying :112 to the one file it was meant for is the smallest thing that ends the class. While in there, `.jacignore:146` (`jac/jaclang/cli/ai_ui/frontend.jac`) is already dead - :112 shadows it.

PR #8655, opened 2026-08-24
fix(scale): the release cache's temp file is named after what it replaces

### LINKED ISSUE #8058: Follow-ups from #8037: scrub JAC_EXECUTABLE, same-version fast path divergence, duplicated cache write
Follow-ups from reviewing #8037 (merged). The fix itself is solid: verified the pinned e2e test on a real run, and confirmed by mutation that reverting the client-build hunk makes it fail with `FileNotFoundError: .../absent-jac`. Three things worth closing behind it:

**1. `release_binary_env` keeps `JAC_EXECUTABLE`** (`jac/jaclang/scale/injector/bundle.jac:53`)

Same class as the `JAC_DEV_SOURCE` back door #8037 closes: an inherited env var a child jac reads to reach the deploying toolchain. Nothing on the `--client web` path reads it today, but both readers that exist (`desktop_build._jac_bin`, the `ninja` re-exec in `tools.impl.jac`) guard with `os.path.exists(exe)` and fall through to `shutil.which("jac")` - so the absent-jac sentinel in the test would silently miss a future leak through it. Adding `JAC_EXECUTABLE` to `_EXCLUDE_ENV` is a one-line hardening.

**2. Same-version fast path diverges from the seal** (`jac/jaclang/scale/injector/binary.jac:466`)

On a version match the client build stays on the running jac, but the seal uses the release bytes even then (`_resolve_seal_binary` returns the downloaded pod payload on linux/same-arch). A dev checkout or pip install reporting the pinned version rebuilds the divergence the PR closes: seal on the release, client bundle on the local jac. If the fast path is meant for checkout workflows that is fine, but then it should probably key on "the running jac is the release binary" rather than the version string. On linux same-arch the release is already downloaded for the pod, so the fast path saves nothing there.

**3. `host_toolchain_path` re-implements the cache write** (`jac/jaclang/scale/injector/binary.jac:473-479`)

The tmp+rename block duplicates what `_fetch_verified_asset` already does at `binary.jac:447-454` under the same `_binary_cache_path` key; it is only reachable when that write failed or the release shipped no `.sha256`. If `_fetch_verified_asset` guaranteed the cache file and returned its path, this block reduces to the chmod, and the known-gap 1 from the PR (`_precompile_app`'s per-deploy mkdtemp leak) gets its natural fix for free: `pack_jab` takes that path instead of bytes.

Thamirawaran:
Closes two of the four follow-ups in #8058 (items 1 and 3), and item 4 lands as the test that proves item 3.

## The temp file was not named after its target

Every asset the deploy caches went through a temp file named with `Path.with_suffix`, which replaces a name's last dotted segment rather than appending to it. On a versioned cache entry that erases the version:

| cache entry | temp file it wrote through |
|---|---|
| `jac-stable-0.34.3-linux-x86_64` | `jac-stable-0.34.<pid>.tmp` |
| `jac-stable-0.34.4-linux-x86_64` | `jac-stable-0.34.<pid>.tmp` |

So two deploys on one host with different `[project] jac-version` pins raced on a single file. A pin's cache entry could end up holding another pin's binary, or nothing at all when the loser's rename found its temp already renamed away:

```
WARNING - Could not cache the downloaded stable jac binary ([Errno 2] No such file or
directory: '.../jac-stable-0.34.33907.tmp' -> '.../jac-stable-0.34.3-linux-x86_64')
```

That is the isolation the version-keyed cache is built on, and #8058 item 4 asks for a test to pin it down. It did not hold. The `dev` channel is unaffected because its cache name carries no version.

`host_toolchain_path` already built the temp name correctly, with `with_name`. Item 3 asked for the duplicated write to go; keeping the correct implementation and deleting the buggy one is what removes it. The shared write also uses `replace` rather than `rename`, which is what refreshing an entry whose checksum stopped matching needs.

## `JAC_EXECUTABLE` reached the pinned release

`release_binary_env` stripped `JAC_DEV_SOURCE`, `JAC_NO_PRECOMPILE` and the `PYTHON*` redirects from the environment a pinned release binary runs in, but not `JAC_EXECUTABLE`, so a child step resolving "the jac to run" could still be handed the toolchain running the deploy.

It is the quietest member of that family. Every reader falls back to `shutil.which("jac")` when the path does not exist, so a leak through it resolves to some other jac rather than failing where #8037's absent-jac sentinel would catch it.

## Tests

No new test file. The concurrency test joins `test_binary_cache.jac` next to its sequential sibling and reuses that file's existing multi-version release stub; the env assertion extends the redirect test already in `test_client_build_fail_fast.jac`.

Both fixes were checked by reverting them:

| reverted | result |
|---|---|
| the temp name | `0.34.3 cached another version's binary`, 5 runs out of 5 |
| the env exclusion | `JAC_EXECUTABLE reaches the pinned release` |

`test_binary_injector.jac` reports 36 passed / 7 error and `test_dry_run_purity.jac` 10 passed / 1 error both with and without this change; those 8 need a reachable cluster. Verified by reverting the two source files against the same tree rather than by comparing counts.

## Item 2 is deliberately untouched, and the reason changed

Item 2 asks whether the same-version fast path in `host_toolchain_path` should key on "the running jac is the release binary" instead of the version string. I implemented removing it, and it broke `a transient install failure is retried and the deploy succeeds`, which passes on a clean tree.

The cause decides the question: `jac create` stamps `jac-version = "==<running version>"` into every scaffolded app. In a source checkout the running jac reports the same version as the release but is ahead of it, so following the pin makes the toolchain fail to parse its own scaffold:

```
error[E0002]: Missing ';'
```

That answers the "if the fast path is meant for checkout workflows that is fine" half of the item: it is. It also rules out the suggested predicate, because a checkout is never the release binary either, so keying on that breaks the same way. Both candidate fixes need a decision about what a pin means when the toolchain that stamped it is a checkout, which is more than this PR should settle. The measurement is on #8058.


MusabMahmoodh (approved):
Traced the pre-fix path: within one process `os.getpid()` is constant, so `cache_file.with_suffix(f".{pid}.tmp")` on `jac-stable-0.34.3-linux-x86_64` and `jac-stable-0.34.4-linux-x86_64` both collapse to `jac-stable-0.34.<pid>.tmp` — with_suffix replaces everything after the last dot, and both names share the `jac-stable-0.34` prefix. The new Barrier-based test forces exactly that collision, so it would have caught this pre-fix.

One non-blocking question: `_http_get_bytes` is monkeypatched for the concurrency test — is stubbing just the network boundary the intentional call here, since the actual race (temp-file naming under real threads) still runs for real either way?

Clean dedup otherwise — one shared `_write_cache_atomic` replacing two divergent implementations, and the `JAC_EXECUTABLE` leak-plug follows the existing `_EXCLUDE_ENV` pattern.

PR #8674, opened 2026-08-24
fix(compiler): a class subscript consults its metaclass before specializing generics

### LINKED ISSUE #8450: checker: class-object attribute lookup never consults the metaclass MRO
## Problem

`_lookup_class_member` walks only the class's own MRO, and the type system has no concept of a metaclass. Attribute access on a class object therefore misses everything Python resolves through `type(cls)`, and the checker emits false `E1030` (or misreads the expression entirely) on code that runs fine.

This is not enum-specific. Validated cases, each confirmed working under `jac run` on current main (with #7361 applied):

**Plain `type` attributes, on every class:**

```jac
obj Foo {}
with entry {
    nm: str = Foo.__name__;  # error[E1030]: Type "Foo" has no attribute "__name__"
}
```

`__qualname__`, `__bases__`, `mro()` etc. fail the same way.

**`EnumMeta.__getitem__`:**

```jac
import from enum { StrEnum }
class Tokens(StrEnum) { A = "a"; B = "b"; }
with entry {
    tok: Tokens = Tokens["A"];  # error[E1001]: Cannot assign Tokens[Literal["A"]] to Tokens
}
```

The subscript is misread as generic specialization instead of the metaclass `__getitem__`, and the false error also demotes the module out of native lowering, so this one has a runtime performance cost, not just a diagnostic one.

**`ABCMeta`, reachable through the documented `class` interop escape hatch:**

```jac
import from abc { ABC, abstractmethod }
class Base(ABC) {
    @abstractmethod
    def area(self: Base) -> float;
}
class Square {
    def area(self: Square) -> float { return 1.0; }
}
with entry {
    Base.register(Square);  # error[E1030]: Type "Base" has no attribute "register"
}
```

Runs fine (`issubclass(Square, Base)` is `True` afterwards). This disproves the assumption in #7361 that `EnumType` is the only metaclass the checker can reach: Jac's grammar has no `metaclass=` syntax, but stub-imported classes carry their declared metaclass into the checker today.

## Proposed fix

On a class-object attribute miss in the class-object branch of `_get_type_of_expression_core`, fall back to a lookup in the metaclass MRO:

- default metaclass `type` for every class
- `EnumMeta` for enum classes
- the declared metaclass for stub-imported classes (typeshed records it, e.g. `class Enum(metaclass=EnumMeta)`)

The vendored typeshed already carries the needed declarations: `enum.pyi` defines `__members__`, `_member_names_`, `_member_map_` and `_value2member_map_` on `EnumMeta(type)`. The fallback needs `self: type[_EnumMemberT]` style generic binding so `Tokens.__members__` resolves to `MappingProxyType[str, Tokens]` rather than `MappingProxyType[str, Enum]`.

This does not add user-facing metaclass support to the language (grammar stays unchanged); it only makes the checker model attribute resolution the way the runtime already behaves.

## Cleanup once landed

The metaclass-MRO fallback subsumes the hand-synthesized special cases, which should then be deleted:

- enum iteration special cases from #5329 / #5790 (gap 45 / gap 52 tests stay as regression coverage)
- `_synthesize_enum_metaclass_attr` from #7361

Related: #7360, #7361

SandeepaHWP:
Closes #8450.

## What was wrong

A class is an instance of its metaclass, so `Cls[x]` runs `type(Cls).__getitem__`. Python only falls back to `__class_getitem__` (generic specialization) when the metaclass has no `__getitem__`. The checker had that order backwards: every subscript on a class went straight to generic specialization.

```jac
import from enum { StrEnum }
class Tokens(StrEnum) { A = "a"; }
with entry {
    tok: Tokens = Tokens["A"];  # error[E1001]: Cannot assign Tokens[Literal["A"]] to Tokens
}
```

The other two cases in the issue (`Foo.__name__`, `ABCMeta.register`) already pass on current main. Only the subscript path was left.

## What changed

- **`_declared_metaclass_member(cls, name)`** returns a member only when a real metaclass declares it, ignoring the `type`/`object` surface every class shares.
- **Class subscripts** check it for `__getitem__` first and dispatch through the metaclass. Once it is selected, the result is final: an invalid key such as `Foo[123]` for `key: str` reports the argument error instead of falling back to generic specialization. If there is no metaclass `__getitem__`, generic specialization runs exactly as before.
- **Magic-method dispatch** uses the same helper instead of its own inline copy of that lookup.

The result is the specific enum (`Tokens`), not the base `Enum`, because main already binds a method's `self: type[T]` to the class it was called on.

## Tests

`jac/tests/compiler/test_metaclass_subscript.jac`:

- enum member lookup by name types as that enum, with a second enum in scope;
- a user metaclass `__getitem__` is dispatched from the class object;
- invalid keys (`Foo[123]`, `Foo[1:3]`) report the parameter type error;
- generic classes without a metaclass `__getitem__` (a plain generic and one with `ABC`) still specialize.

The new tests (4) and the existing enum checker tests (9) pass, and `jac check` is clean on the type evaluator.


jac/jaclang/compiler/type_system/type_evaluator.impl/type_evaluator.impl.jac
```diff
@@ -1548,6 +1548,20 @@ impl TypeEvaluator._get_type_of_expression_core(
...
+                    meta_item = self._lookup_metaclass_member(base_type, "__getitem__");
+                    if meta_item is not None
+                        and self._is_metaclass_declared(meta_item) {
+                        item_ty = self.get_type_of_magic_method_call(
+                            base_type,
+                            "__getitem__",
+                            [self._subscript_index_expr(expr.right)],
+                            expr
+                        );
+                        if item_ty is not None
```
christianwilkins, line 1560: [P2] Report invalid metaclass subscription arguments instead of falling through. For `Meta.__getitem__(cls: Meta, key: str) -> int`, both `Foo[123]` and `Foo[1:3]` produce zero compiler errors on this head. The magic-method helper checks candidates with overload diagnostics suppressed and returns no valid result; this branch then resumes generic specialization. Once a metaclass `__getitem__` is found, preserve its argument errors and add negative tests. A normal `takes(key: str)` control rejects `takes(123)` in the same production-compiler probe.

christianwilkins (changes requested):
Reviewed `19aaa3ed67e7b523bfdef6fc49fd0bae7e7129fe`: full diff, metaclass lookup and receiver TypeVar binding, generic fallback, tests, conversation, and CI for this head. All six new regression bodies passed through the production compiler in a detached checkout. `git diff --check` passed.

P2: invalid arguments to a known metaclass `__getitem__` silently fall through to generic specialization. With `Meta.__getitem__(cls: Meta, key: str) -> int` and `class Foo(metaclass=Meta)`, the compiler reports no errors for either `Foo[123]` or `Foo[1:3]`. The positive `Foo['ok']` also passes. In the same probe, a normal `takes(key: str)` call with `123` correctly reports “Cannot assign int to parameter 'key' of type str,” so checking is active. Please retain argument diagnostics once the metaclass subscription method is selected and add invalid-key coverage; a rejected call should not become a generic specialization attempt.

GitHub currently reports merge conflicts. Current-head workflow 32800303711 has failed build-kit and native jobs: the available build-kit log reports NATIVESEAL E5024 because scheduling requires ownership analysis that is unavailable; the native job log could not be retrieved. I have not established whether those CI failures are introduced by this patch and did not run a sealed compiler build locally. Author actions: fix the invalid-call path, rebase, and rerun checks on the resulting head, investigating or baselining remaining failures. No source files were changed.

jac/jaclang/compiler/types/type_evaluator.impl/type_evaluator.impl.jac
```diff
@@ -1548,6 +1548,20 @@ impl TypeEvaluator._get_type_of_expression_core(
...
+                    meta_item = self._lookup_metaclass_member(base_type, "__getitem__");
+                    if meta_item is not None
+                        and self._is_metaclass_declared(meta_item) {
+                        item_ty = self.get_type_of_magic_method_call(
+                            base_type,
+                            "__getitem__",
+                            [self._subscript_index_expr(expr.right)],
+                            expr
+                        );
+                        if item_ty is not None
```
SandeepaHWP: Thanks, fixed in 061e8d5d4. Once a metaclass `__getitem__` is selected, the checker no longer falls back to generic specialization. A rejected key now reports its argument errors, so `Foo[123]` and `Foo[1:3]` with `key: str` fail with the parameter type error. The new `invalid keys to a metaclass __getitem__ are reported` test in `test_metaclass_subscript.jac` covers both cases, and `Foo["abc"]` still passes.

kugesan1105:
Pushed 2 commits (maintainer edit) and synced with `main`. The fix is right and at the right layer; this is one correction plus a test fold.

## The regression

The metaclass check ran on **every** subscript, including inside annotations:

```jac
class Meta(type) { def __getitem__(cls: Meta, key: type) -> int { return 1; } }
class Box[T](metaclass=Meta) { has item: T; }

b: Box[int] = Box(item=1);   # error[E1001]: Cannot assign Box to int
```

`Box[int]` resolved to the metaclass return type, so `b` was silently declared `int`. In an annotation a subscript is always generic specialization.

## The fix

```jac
if not in_type_position(expr, include_subscripts=False)
    and (meta_getitem := self._declared_metaclass_member(base_type, "__getitem__")) {
```

`include_subscripts=False` is load-bearing. With `True`, every node under an `IndexSlice` counts as a type position and the gate never opens.

## Tests

Moved into `test_metaclass_operator.jac` (same subject, same `check_src` helper, same function this PR refactors). **7 tests / 2 files becomes 5 tests / 1 file**, with more coverage.

New canary: `a metaclass __getitem__ does not capture generic specialization`, over no metaclass, `ABC` base, metaclass keyed on `type`, keyed on `any`. Remove the gate and it fails.

Release note fixed: it claimed `Box[int]` still specializes, which was false in exactly this case.

## Please don't close #8450

Same family, still broken here and on `main`:

```jac
len(Foo)        # E1053 ... of type Sized   (metaclass __len__ ignored)
Foo()           # metaclass __call__ ignored
Foo["a"] = 1    # store still specializes; only the read path is fixed
```

Native also still demotes, now as `E5092`, not the checker error. Suggest `Closes #8450` becomes `Refs #8450`.

Also leave `_synthesize_enum_metaclass_attr` (#7361) alone for now: deleting it downgrades `Tokens.__members__` to a property object and `_member_map_` to `dict[str, Enum]`. Own PR, gap 45 / gap 52 as tripwires.

## Verified

996 passed / 0 failed (`tests/compiler/passes` + metaclass files), 37 on neighbouring type tests, `jac fmt --check --lintfix` clean, level with `main`.

Beyond the filed repro, these go from failing on `main` to passing and had no test: inherited metaclass, subclass inheriting it, slice keys, tuple keys, `IntEnum`, plain `Enum`, non-literal `C[k]`, and `Tokens["A"]` silently becoming `Tokens[A]` when a `class A {}` is in scope.

kugesan1105:
## Summary

Subscripting a class now runs its metaclass `__getitem__` before falling back to generic specialization, matching Python.

## The problem

In Python a class is an instance of its metaclass, so `Cls[x]` calls `type(Cls).__getitem__`. Generic specialization (`__class_getitem__`) is only the fallback.

The checker had it backwards: every subscript on a class went straight to specialization. Enums broke, because `EnumMeta.__getitem__` is how member lookup works.

```jac
import from enum { StrEnum }
class Tokens(StrEnum) { A = "a"; }

tok: Tokens = Tokens["A"];
# before: error[E1001] Cannot assign Tokens[Literal["A"]] to Tokens
# after:  ok, type is Tokens
```

This was a hard failure, not just a bad diagnostic. `jac run` exited 1 and printed nothing.

## The fix

Three pieces:

| Change | Effect |
| --- | --- |
| `_declared_metaclass_member(cls, name)` | Finds a member only when a **real** metaclass declares it, ignoring the `type` / `object` surface every class shares |
| Subscript consults it first | Metaclass `__getitem__` wins; a bad key reports the argument error instead of silently specializing |
| Magic-method dispatch reuses it | Drops an inline copy of the same lookup |

Annotations are excluded, because there a subscript is always specialization:

```jac
n: int = Foo["a"];      # value position  -> metaclass __getitem__
b: Box[int] = Box(...); # type position   -> generic specialization
```

## Why this shape

Subscript was the **last** class-object operation that never asked the metaclass. Attribute access and operators already did, via #8399. So this is the remaining hole, not a new mechanism.

Because it is structural rather than an enum special case, it also fixes cases nobody filed:

```jac
Nums["X"]        # IntEnum
C[k]             # non-literal key
Foo["a", "b"]    # tuple key
Foo[1:3]         # slice key
Bar["a"]         # subclass inheriting the metaclass
```

And a silent one: with a `class A {}` in scope, `Tokens["A"]` used to resolve to `Tokens[A]` with no diagnostic at all.

## Scope

Fixed: the read path, `Cls[key]`.

Not fixed, tracked in #8450:

```jac
len(Foo)        # metaclass __len__ ignored
Foo()           # metaclass __call__ ignored
Foo["a"] = 1    # store path still specializes
```

PR #8690, opened 2026-08-25
test(scale): name the parsed value in the storage-quantity helper

kugesan1105:
## What

Renames the local `result` to `parsed` in `_test_parse_storage_valid` in [test_k8s_utils.jac](jac/jaclang/scale/tests/test_k8s_utils.jac), so the assertion message reads against a name that says what the value is.

Test-only, no behavior change.

## Validation

- `jac fmt jac/jaclang/scale/tests/test_k8s_utils.jac` -- clean, no reformatting beyond the edit
- `jac check jac/jaclang/scale/tests/test_k8s_utils.jac` -- passes (the remaining W1036/W1051 warnings are pre-existing and untouched)
- `jac test jac/jaclang/scale/tests/test_k8s_utils.jac` -- 17 passed

### DIFF
--- a/jac/jaclang/scale/tests/test_k8s_utils.jac
+++ b/jac/jaclang/scale/tests/test_k8s_utils.jac
@@ -78,8 +78,8 @@ test "resize pvc if needed: increase patches, no-op skips, decrease raises" {
 def _test_parse_storage_valid(pair: tuple) {
     raw = pair[0];
     expected = pair[1];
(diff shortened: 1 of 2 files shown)


MusabMahmoodh (approved):
Fine as is. Two-line rename, the only reader of the local moved with it, and `parsed` shadows nothing in the enclosing scope.

PR #8699, opened 2026-08-25
fix(scale): destroy reclaims the namespace it created, and spares the one it adopted

### LINKED ISSUE #7968: Jac Scale: KubernetesTarget.destroy() leaves the namespace, mongo, redis and PVCs running
## Summary

`KubernetesTarget.destroy()` deletes only the resources matching
`jac-scale.role in (microservice, gateway)`. It never deletes the namespace, the
mongo StatefulSet, the redis Deployment (which carries no `jac-scale.role`
label, so the selector misses it), or any PVC. Every destroy therefore leaves a
running database, a running cache and two bound volumes behind, in a namespace
that stays `Active` forever.

The base class it inherits from, `KubernetesTargetBase.destroy()`, does all of
this correctly. The subclass overrides `destroy()` unconditionally, so the
correct teardown is unreachable for anyone constructing a `KubernetesTarget`.

## Environment

| | |
|---|---|
| jac | 0.34.13 (pinned release; also read against `main` @ `a8949072`) |
| Cluster | EKS `jaseci-cluster` (dev) and a PR-preview namespace |
| Caller | jacBuilder, which constructs `KubernetesTarget` directly |

Reproduced on both a PR-preview environment and the shared dev environment.

Deploy an app through a `KubernetesTarget`, then destroy it and look at the
namespace.

```bash
# 1. after the destroy reported success and the app record was cleared:
kubectl get ns jb-<uid>-<slug>
kubectl get all,pvc -n jb-<uid>-<slug>
```

Observed, verbatim, after a successful destroy:

```
namespace/jb-<uid>-jac-todo-web            Active

deployment.apps/jac-todo-web-redis                                        1/1
statefulset.apps/jac-todo-web-mongodb                                     1/1
persistentvolumeclaim/jac-todo-web-bundles                                Bound
persistentvolumeclaim/jac-todo-web-mongo-data-jac-todo-web-mongodb-0      Bound
```

The app and gateway Deployments were gone. Nothing else was.

The label state that explains it, read off a live namespace on the dev cluster:

```
$ kubectl get deploy,sts -n jb-4746b2c6-test01 -o json | ...
Deployment   gateway-deployment  {'app': 'gateway',       'jac-scale.role': 'gateway',      'managed': 'jac-scale'}
Deployment   test01-deployment   {'app': 'test01',        'jac-scale.role': 'microservice', 'managed': 'jac-scale'}
Deployment   triage-deployment   {'app': 'triage',        'jac-scale.role': 'microservice', 'managed': 'jac-scale'}
Deployment   test01-redis        {'app': 'test01-redis',  'component': 'cache',             'managed': 'jac-scale'}
StatefulSet  test01-mongodb      {'managed': 'jac-scale'}
```

Only the first three carry a `jac-scale.role`. redis does not, and the
StatefulSet is not a Deployment at all, so neither is reachable from the
selector `destroy()` uses.

## Expected

`destroy()` removes everything the matching `deploy()` created, including the
namespace, databases and volumes. This is what
`KubernetesTargetBase.destroy()` already does
(`jac/jaclang/scale/deploy/target/kubernetes/kubernetes_target.jac:831`):

- `:869` `self._destroy_databases(app_name, namespace, apps_v1, core_v1)`, which
  deletes the mongo StatefulSet, redis Deployment, their Services, ConfigMaps,
  Secrets and ServiceAccounts (`:649` onward)
- `:875` deletes every PVC in the namespace whose name starts with `app_name`
- `:896` `delete_namespace(namespace)`

## Actual

`KubernetesTarget.destroy()`
(`jac/jaclang/scale/deploy/target/kubernetes/target.jac:1164`) overrides it and
stops short. Its whole teardown is label-scoped:

```jac
selector: str = "managed=jac-scale,jac-scale.role in (microservice,gateway)";   // :1180
ms_selector: str = "managed=jac-scale";                                          // :1181
```

and it deletes, in order: Deployments matching `selector` (`:1184`), autoscalers
(`:1194`), PodDisruptionBudgets (`:1208`), Ingresses (`:1218`), Services
(`:1228`), ConfigMaps with `jac-scale.role=source` (`:1249`), then the
monitoring stack (`:1261`). It then returns.

There is no `_destroy_databases` call, no PVC deletion and no `delete_namespace`
anywhere in the override. `super`'s version is never invoked, and the override
is not conditional on the topology actually being microservice, so this is the
only teardown any `KubernetesTarget` caller can get.

## Impact

Every destroyed app permanently leaks, per app:

- 1 namespace, `Active` forever
- 1 mongo StatefulSet pod and 1 redis pod, both running and scheduled
- 2 bound PVCs, one of them EBS, so it bills until someone deletes it by hand

This compounds. Nothing ever reclaims it, because the record the caller would
need to find the namespace again is deleted as part of the successful destroy,
so the leftovers are orphaned the moment they are created. On a multi-tenant
cluster this is a steady drift of unowned pods and volumes with no owner to
attribute them to, and it silently consumes node capacity that live apps need.

It also breaks the contract callers advertise. jacBuilder's destroy button tells
the user it "tears down the production namespace and frees the subdomain". Only
the subdomain half was ever true.

## Workaround (verified)

Deleting the namespace after `destroy()` returns collects all of it atomically:

```bash
kubectl delete namespace jb-<uid>-<slug>
```

Verified: namespace, StatefulSet, redis and both PVCs all go, and the bound PVs
release. jacBuilder now does this itself as a post-destroy sweep, guarded to its
own `jb-` namespace convention, since it cannot wait for a jac release.

## Suggested fix

In `KubernetesTarget.destroy()` (`target.jac:1164`), after the label-scoped
sweep, do what the base class does: call `_destroy_databases`, delete the
app's PVCs, and delete the namespace when the target owns it. Deleting the
namespace makes the label selectors redundant for the owned-namespace case and
removes the whole class of "resource had no role label so it survived".

Two things worth deciding alongside it:

- redis carries `component=cache` but no `jac-scale.role`. Either give every
  resource a role label at creation, or stop keying teardown on a label that
  creation does not guarantee. Right now the two sides disagree, and teardown is
  the side that fails silently.
- The override fires for non-microservice topologies too, since it is not
  conditional. If the base teardown is correct for those, the override should
  delegate rather than replace.

Regression test that would have caught it: deploy through `KubernetesTarget`,
call `destroy()`, then assert the namespace is gone. A weaker assertion on the
app Deployment alone passes today while the database, cache and volumes are all
still running.

akindu-k:
Fixes #7968.

image

## What was wrong

When you deploy an app with Jac Scale, it creates your app, a gateway, a Postgres database, storage volumes, and a namespace to hold them all.

When you ran `destroy`, it deleted only the app and the gateway. Everything else stayed: the database kept running, the volumes stayed allocated (on AWS these are EBS volumes that keep billing), and the namespace stayed `Active` forever.

## Why it happened

Kubernetes objects are deleted by *label selector*, which works like a query: "delete everything with these labels". If an object does not carry the label you searched for, it is invisible to the delete. Nothing errors, nothing warns.

The deletion code searched for this:

```
FLEET_SELECTOR = "managed=jac-scale,jac-scale.role in (microservice,gateway)"
```

But the code that *creates* the Postgres database labels it like this:

```jac
'labels': {'managed': 'jac-scale'}
```

No `jac-scale.role`. So the selector never matched it. The two halves of the code disagreed about labelling, and deletion was the half that failed silently.

Driving the old `destroy()` against faked clients showed every call it made:

```
delete_collection_namespaced_deployment      (FLEET_SELECTOR)
autoscaler.destroy_collection
delete_collection_namespaced_pod_disruption_budget
delete_collection_namespaced_ingress
list_namespaced_service + per-item delete
delete_collection_namespaced_config_map
monitoring.destroy
```

No StatefulSet, PVC, Secret or namespace call anywhere.

`destroy()` also ignored its `component` argument completely, so asking to destroy one component silently performed a full teardown.

## Why "just call the parent class" would be wrong

The issue suggests reusing the parent's working teardown. That would introduce a worse bug.

`ensure_namespace_exists` either creates a namespace or adopts one that already exists, and it recorded no difference between the two. The parent's `delete_namespace` has one safety check: skip if the name is `default`.

So if a user sets `namespace = "production"` on a shared namespace, Jac Scale adopts it, and a later `destroy` deletes that entire namespace along with every other team's workloads in it. That trades a resource leak for data loss.

The fix therefore had to make ownership *knowable* first.

(That reference implementation is also gone now: #7709 removed the whole `KubernetesTargetBase` teardown block on 2026-08-28, so this PR no longer deletes it and no longer calls into it.)

## What changed

**Ownership is recorded per app, at creation.** `ensure_namespace_exists` stamps `jac-scale.jaseci.org/owned-by` with the app name on namespaces it creates. A namespace it merely adopts stays unmarked. The `owner` argument is required, not defaulted, so a new call site cannot silently create an unstamped namespace.

**A namespace is deleted whole only when this app created it and is still alone in it.** The label answers "who created this", which is not the same question as "who is in it now": a second app can deploy into the namespace afterwards, through the adopt path, without changing the label. So before reclaiming, `destroy` looks inside for another app's jac-scale resources, a bundle volume or a database that is not this app's, or a shared volume stamped for someone else. Anything found and the namespace is kept, the teardown falls back to removing only this app's resources, and a warning names what it found.

**An adopted namespace keeps its namespace** and loses only this app's own resources.

**Reclamation is scoped to the app that owns it.** Two apps can share a namespace, so ownership naming the app rather than the tool is what stops destroying one from deleting the other's database. The ops RBAC sweep is scoped the same way, `managed=jac-scale,app={app_name}`.

**A shared volume names the app that declared it.** It is named from user config, so it carries no app prefix, and its labels named no app either. The only way to find it was the role label every shared volume shares, so destroying one app took a co-tenant's volumes with it. Creation now stamps `jac-scale.owner`, and teardown matches on it. Volumes that predate the label are swept only where this app is the namespace's sole tenant, and named in a warning everywhere else.

**PVCs go by exact name, never by prefix.** Prefix matching both missed and over-reached: a shared volume is named from user config with no app prefix, the bundle PVC is created under `k8s_safe_name` so an app called `My_App` leaked it, and a sibling app sharing a prefix was deleted. App volumes now go by exact name, and shared volumes by their `jac-scale.role=shared-volume` label. A label-only sweep would not work on its own: the Postgres data PVC comes from a `volumeClaimTemplates` entry with no labels at all.

**Monitoring is torn down before the namespace, and gates it.** It owns cluster-scoped ClusterRoles that a namespace deletion does not collect. If that teardown fails, `destroy` now warns and keeps the namespace rather than orphaning roles named after a namespace that no longer exists.

**The database teardown is inlined**, so `destroy` no longer depends on `KubernetesTargetBase`. It now also removes the Secret holding an external database's connection string, which nothing but a namespace deletion collected before, so an app on an external database left live credentials behind in every namespace this PR spares.

**`component` is honoured** instead of triggering a full teardown, and each component takes only its own storage. Destroying the database leaves the bundle volume holding the app's code, which the still-running app pods mount. A component that could not be removed now says so instead of reporting success. Destroying monitoring used to print success even when the cluster-scoped role delete was refused, because the teardown is written to tolerate a failure during a full destroy and the component path threw that result away. It now fails, the way destroying the application or the database has always done.

## Verification on a real cluster

`destroy_reclaim_real_e2e.sh` runs in the PR-gated `test-scale-k8s` lane, so every PR touching scale exercises it against a live MicroK8s cluster. What destroy reclaims and what it spares is only decidable by an API server.

Five properties, each one a bug this fixes:

| | Property |
|---|---|
| A | the namespace jac-scale created is reclaimed whole |
| B | a namespace it did not create survives, with its foreign workloads |
| C | destroying one app never deletes a namespace another app is deployed into, in either direction: the app that adopted the namespace and the app that created it |
| D | a shared volume, named from config with no app prefix, is still reclaimed |
| E | a PVC belonging to an app sharing a name prefix is never deleted |

C runs both directions because only one of them is safe by construction. Destroying the adopting app takes the shared path because the namespace names someone else. Destroying the app that created it is the direction where the owner label reads true and the namespace is a candidate for deletion, so that is where a co-tenant's database can be lost, and the scenario now deploys the co-tenant back and destroys the owner last.

The e2e earned its place before it was even green. It caught a bug no unit test could: database provisioning ensures the namespace long before the deploy applies manifests, and that call passed no owner, so every namespace was created unstamped and destroy read it as adopted. The mock tests passed throughout, because they asserted the function stamps correctly when called with an owner, while the real deploy called it from somewhere else, first, without one.

## Tests

The unit tests cover only what a cluster cannot be made to do: the `default` short-circuit, labels that read as `None` or `{}`, an unreachable API server that must raise rather than answer "not owned", an already-404 namespace, a monitoring teardown that throws so the namespace must be kept, a namespace this app created that another app has since deployed into, and a single-component teardown that must not take the app's code volume.

They also cover the warnings: that another app in the namespace is named on both the component path and the adopted-namespace path, that an app alone in its namespace is not wrongly told it has a neighbour, and that a refused monitoring teardown fails its own component destroy rather than reporting success.

The monitoring one gates a destructive action and is checked against its own regression. Replacing the gate with `if True` fails it:

```
FAILED :: a failed monitoring teardown keeps the namespace
E   AssertionError: the namespace was deleted after monitoring teardown failed,
    orphaning cluster-scoped roles named after a namespace that no longer exists
```

Everything else is asserted by the e2e against real objects, so it fails on wrong semantics rather than on a changed client call shape.

## Notes for reviewers

**On mongo and redis.** The issue names them as the leaked backends, accurate when it was filed on 2026-08-08. #7947 replaced the persistence stack the next day, so the database is now Postgres. The bug is identical, only the resource names changed. Worth flagging because grepping for `mongo` today finds nothing, which could look like the issue is already fixed.

**Apps deployed before this change** have no owner label, so their namespaces take the adopted path: database, volumes and RBAC are reclaimed, but the namespace stays. That is better than today and is the safe default, since we cannot prove we created those namespaces. Existing orphans still need a one-off sweep.

**Shared volumes created before this change** carry no owner label, and an existing PVC is never patched, so they can never gain one. They are still swept, but only where this app provably created the namespace and is alone in it, which is the one case where no other app can have declared them. Anywhere else they are kept and named in a warning, so an operator who expected the volume gone can see why it is not.

**Fleet workloads are still swept namespace-wide.** In a namespace shared by two apps, destroying one still deletes the other's Deployments, gateway, Services, Ingress, PDBs and autoscalers.

Nothing is lost: the co-tenant's database keeps running, and its volumes, secrets and ops RBAC are untouched. What it loses is what serves traffic, including the Service in front of its database, which carries `managed=jac-scale` and no app name. So the co-tenant has to be redeployed to come back, and a redeploy re-attaches it to the same volumes and the same database.

This no longer happens quietly. When destroy finds another app in the namespace, it names it and says what to do about it:

```
Another app has resources in 'team-apps': StatefulSet/billing-postgres. Its data
was kept, but this removed its deployments, autoscalers and disruption budgets
along with this app's, so redeploy it to restore service.
```

That covers the two paths that said nothing before. `--component application` never looked for another app at all, because the search only ran during a full destroy. And a full destroy into a namespace this app did not create skipped the search too, since the namespace was never going to be deleted and the answer changed no decision. In both cases the co-tenant went down without a word.

Scoping that sweep needs a per-app label stamped at creation, because those objects are labelled with the service name rather than the app, so it is not a selector change. It predates this PR. Filed as #8816.


MusabMahmoodh (commented):
Read the issue and the diff against the surrounding code. The owned-namespace half is the right fix at the right altitude: deleting the namespace retires the whole "this resource had no role label so it survived" class rather than patching one more name into a list, and refusing the issue's own "just call the parent" suggestion because `ensure_namespace_exists` adopts as readily as it creates is the right call.

Five things below, roughly in severity order. The first two I think need to change before this merges; the ownership one is a new data-loss path, since before this PR `destroy` never deleted a namespace at all.

One thing not inline: `KubernetesTargetBase.destroy` / `_destroy_component` / `_destroy_application` / `_destroy_databases` / `_wait_for_deletion` (`kubernetes_target.jac:366-520`) are the "correct but unreachable" implementation the issue pointed at, and this PR deliberately reimplements them in the subclass. `_destroy_named_component` is a near-verbatim copy of the base `_destroy_component`, same `ValueError` string and all. That block is now provably dead and I would like to see it in a "Deleted" section of this PR rather than left for later.

jac/jaclang/scale/deploy/target/kubernetes/target.jac
```diff
@@ -1755,4 +1752,158 @@ obj KubernetesTarget(KubernetesTargetBase) {
...
+            return;
+        }
+
+        self._destroy_fleet_workloads(
+            app_name, namespace, apps_v1, core_v1, policy_v1, networking_v1
+        );
+
+        self._destroy_monitoring(app_name, namespace, apps_v1, core_v1);
+
+        if namespace_is_owned(core_v1, namespace) {
```
MusabMahmoodh: The owner stamp records "jac-scale created this namespace", not "this app owns it". A second app deploying into the same namespace goes down the adopt path in `ensure_namespace_exists`, so the label is already there and this reads true for it too - `destroy` on app B then deletes app A's namespace out from under it. Nothing forces one app per namespace; `namespace` is just a jac.toml key defaulting to `default`.

Can we claim it for the app that created it? Stamp `create_labels[NAMESPACE_OWNER_LABEL] = k8s_safe_name(app_name)` and have `namespace_is_owned(core_v1, namespace, app_name)` compare against that, so a namespace jac-scale made for a different app takes the shared path instead. Same reason `_destroy_ops_rbac` wants `managed=jac-scale,app={app_name}` rather than bare `managed=jac-scale`: as written it deletes the co-tenant app's ops Role and RoleBinding too.

jac/jaclang/scale/deploy/target/kubernetes/target.jac
```diff
@@ -1755,4 +1752,158 @@ obj KubernetesTarget(KubernetesTargetBase) {
...
+            self.logger.warning(
+                f"Timed out waiting for namespace '{namespace}' to be deleted"
+            );
+        }
+    }
+
+    def _destroy_in_shared_namespace(
+        app_name: str, namespace: str, apps_v1: any, core_v1: any
+    ) {
+        self._destroy_database(app_name, namespace, apps_v1, core_v1);
```
MusabMahmoodh: `_destroy_database` only exists on `KubernetesTargetBase` (`kubernetes_target.jac:423`), and #7709 deletes that whole teardown block, `_destroy_database` included - Linus flagged exactly this on #7968 in August. Whichever of the two merges second, this line becomes an AttributeError at destroy time, in the adopted-namespace path that has no cluster coverage.

Worth copying those ~20 lines into `KubernetesTarget` here so it does not depend on merge order. That also makes the rest of the base teardown unambiguously dead.

jac/jaclang/scale/deploy/target/kubernetes/target.jac
```diff
@@ -1755,4 +1752,158 @@ obj KubernetesTarget(KubernetesTargetBase) {
...
+        } except ApiException as e {
+            if e.status == 404 {
+                return;
+            }
+            raise;
+        }
+        pvc_items: list[any] = pvcs.items;
+        for pvc in pvc_items {
+            pvc_name: str = pvc.metadata.name;
+            if not pvc_name.startswith(app_name) {
```
MusabMahmoodh: Shared-volume PVCs are named `k8s_safe_name(vol.name)` with no app prefix (`manifest_builder.jac:633`), so `[[scale.shared_volumes]] name = "uploads"` gives a PVC called `uploads` and this filter never sees it. In an adopted namespace it survives destroy, still bound, still billing - #7968 again, one resource further along.

They already carry the labels for it:

```jac
core_v1.delete_collection_namespaced_persistent_volume_claim(
    namespace=namespace, label_selector="managed=jac-scale"
);
```

The bundle PVC has the mirror problem: created as `{k8s_safe_name(app_name)}-bundles` at `target.jac:1081`, matched here against the raw `app_name`, so an app named `My_App` leaks it.

jac/jaclang/scale/deploy/target/kubernetes/target.jac
```diff
@@ -1742,7 +1737,9 @@ obj KubernetesTarget(KubernetesTargetBase) {
                 raise;
             }
         }
+    }
 
+    def _destroy_monitoring(app_name: str, namespace: str, apps_v1: any, core_v1: any) {
```
MusabMahmoodh: The PR body says the monitoring-before-namespace order is load-bearing because those ClusterRoles are cluster-scoped, and there is a test asserting it - but this catches every exception and logs at `info`, so on failure the ordering is silently a no-op. The namespace then goes, and `{namespace}-{app}-kube-state-metrics` and `{namespace}-{app}-prometheus` (`monitoring.jac:2151-2153`) are orphaned cluster-wide with the namespace that named them gone, so nothing is left to attribute them to.

Silent fallback at a decision point, same class as the #7900 theme-1 findings. `warning` at minimum; better, skip the reclaim and say why, since this particular leak cannot be found afterwards.

jac/jaclang/scale/tests/deploy/test_destroy_reclaim.jac
```diff
@@ -0,0 +1,140 @@
...
+        for call in core_v1.delete_namespaced_persistent_volume_claim.call_args_list
+    ];
+    assert sorted(deleted) == sorted(pvcs[:2]) , (
+        f"destroy deleted PVCs {deleted}; it must reclaim the app's two and "
+        "leave foreign-pvc alone"
+    );
+}
+
+
+test "ownership is only claimed for a namespace jac-scale created" {
```
MusabMahmoodh: This covers the read side, but nothing asserts that `ensure_namespace_exists` actually stamps `owned-by` on the namespace it creates. Put that dict back to `default_pss_labels()` and every namespace reads as adopted, the #7968 leak returns, and all four tests here stay green. A case whose `read_namespace` raises 404, asserting `create_namespace`'s body carries the label, closes it.

Related: the two destroy tests patch out `namespace_is_owned` and `delete_namespace` and then assert on MagicMock call order, so what they pin is "destroy calls these two functions". You already had a kind cluster doing the real thing - is `tests/deploy/keda_http_activation_real_e2e.sh` not the precedent for committing that run so it gates in nightly?

akindu-k:
Thanks, this was a good catch list. All five are addressed, plus the dead base block. Taking them in your severity order.

## 1. Ownership was recorded per namespace, not per app

You were right, and this was the worst of it: before this PR `destroy` never deleted a namespace at all, so I introduced a data-loss path that did not previously exist.

Fixed the way you suggested. The label now carries the app rather than the constant:

```jac
create_labels[NAMESPACE_OWNER_LABEL] = owner;      # owner = k8s_safe_name(app_name)
```

and `namespace_is_owned(core_v1, namespace, owner)` compares against it, so a namespace jac-scale made for a different app takes the shared path. `ensure_namespace_exists` gained an `owner` parameter; the adopt branch is untouched, so an adopted namespace is still never stamped.

`_destroy_ops_rbac` is scoped the same way, `managed=jac-scale,app={app_name}`, so a co-tenant's ops Role and RoleBinding survive.

## 2. Shared-volume and bundle PVCs

Confirmed both. A shared volume is named `k8s_safe_name(vol.name)` straight from user config, so `uploads` never matched a `startswith(app_name)` sweep. And the bundle PVC is created under `k8s_safe_name(app_name)` but was matched against the raw `app_name`, so an app named `My_App` leaked it.

One caveat on the fix you proposed, because it would have quietly made things worse. A bare `managed=jac-scale` collection does not catch the Postgres data PVC: it comes from a `volumeClaimTemplates` entry whose metadata is `{'name': n['volume']}` with **no labels at all**, so the label sweep misses the exact EBS volume this issue is about. It would have traded a shared-volume leak for a Postgres data leak.

So the teardown is a union of the two, and prefix matching is gone entirely:

- app-scoped PVCs by exact name: `bundle_pvc_name(k8s_safe_name(app_name))` and `{app}-postgres-data-{app}-postgres-0`
- shared volumes by their role label: `managed=jac-scale,jac-scale.role=shared-volume`

This also closes the opposite direction, which the PR body claimed but `startswith` did not deliver: destroying `todo` no longer deletes `todo-analytics`.

## 3. Monitoring failure was a silent no-op at a decision point

Agreed, and the consequence is worse than a normal swallowed error because the orphan cannot be found afterwards: the ClusterRoles are named after a namespace that no longer exists.

`_destroy_monitoring` now returns a bool and logs at `warning`. On failure `destroy` **skips the reclaim entirely** and keeps the namespace, saying why, so the roles stay attributable:

```
Keeping namespace '<ns>': its monitoring RBAC is cluster-scoped and is named
after this namespace, so deleting it now would orphan roles nothing can attribute
```

## 4. `_destroy_database` and merge order with #7709

Fixed by inlining. `KubernetesTarget` now has its own `_destroy_app_database`, so `destroy` does not reach into the base at all and merge order with #7709 no longer matters.

## 5. Test gap on the stamp

You were exactly right, and I verified your scenario rather than taking it on faith. Reverting that dict to `default_pss_labels()` left all four original tests green. It now fails:

```
FAILED :: creating a namespace stamps the owning app on it
E   AssertionError: the created namespace carries no owner label, so destroy
    reads it as adopted and the #7968 leak returns
```

There is a companion test asserting the adopt branch never patches ownership on.

The suite is nine tests now, and each new one is checked against the regression it guards:

```
drop the shared-volume sweep + app-scoped RBAC:
FAILED :: shared-volume PVCs are reclaimed by their role label
E   AssertionError: a shared volume is named from user config with no app prefix,
    so without its role label it survives destroy and keeps billing
FAILED :: ops RBAC teardown is scoped to this app
E   AssertionError: delete_collection_namespaced_role used selector
    'managed=jac-scale', which also deletes a co-tenant app's ops Role and RoleBinding
```

## 6. The dead base block

Deleted, as you asked, rather than left for later: `destroy`, `_destroy_component`, `_destroy_application`, `_destroy_databases`, `_destroy_database` and `_wait_for_deletion`, 357 lines, plus two imports they orphaned. I checked nothing else references them, including the two tests that construct `KubernetesTargetBase` directly.

## On the e2e, and where verification actually stands

I have added `destroy_reclaim_real_e2e.sh` and wired it into the nightly job next to `sdk_deploy_real_e2e.sh`, per your point about the precedent. It covers five properties: owned namespace reclaimed, adopted namespace and its foreign workloads spared, a co-tenant's namespace never deleted, a shared volume reclaimed, and a sibling app's PVC left alone. The last three are your findings.

Being straight about its status: **the committed script has not yet had a green end-to-end run.** Deploys on my machine fail at app seal because the published `stable` runtime skews against the working tree, and building the binary from the checkout to bypass that did not finish locally. So the first real execution will be CI.

What I did verify against a real kind cluster, driving the real SDK `destroy` rather than mocks, was the earlier revision of this fix. The owned path reclaimed the namespace, the postgres StatefulSet and both PVCs with no PV left bound, and the adopted path left the namespace and a foreign ConfigMap intact while reclaiming the app's own resources. That evidence predates these review fixes, so I am not presenting it as covering them.

If you would rather not merge on a script whose first run is the nightly, I am happy to hold until someone with a working cluster runs it, or to run it myself once the toolchain builds here.

MusabMahmoodh (approved):
All five points are addressed, and two of them landed better than what I asked for. Approving.

The one I care most about is the ownership stamp. `NAMESPACE_OWNER_LABEL = "jac-scale.jaseci.org/owned-by"` carrying the app rather than the tool, plus `_owns_namespace` comparing against that app, is the right shape - and the test that a namespace labelled `other-app` is *not* reclaimable by `APP` is the assertion the first cut was missing. The RBAC sweep being scoped the same way closes the co-tenant hole properly rather than papering it.

Also right: not patching ownership onto an adopted namespace. Recording "I own this" on a namespace you merely found would have converted the adopted case into the owned case on the second deploy, and the test asserting `NAMESPACE_OWNER_LABEL not in patched` pins that. Good instinct.

`destroy_reclaim_real_e2e.sh` is the strongest thing in the PR. Five named properties, each one a bug this change fixes, run against a real cluster with real PVCs and a real foreign workload as the sentinel - that is the level this surface needs, because every defect here is "we deleted the wrong object", which only an API server can adjudicate.

Which is why I want to push on the test layering rather than the code.

**The mock layer now duplicates the e2e, and it is the layer that is running.** `test_destroy_reclaim.jac` uses `unittest.mock.MagicMock()` and `patch.object(utils.client, "CoreV1Api", ...)`, then asserts on `call.kwargs.get("label_selector")` - that is a check on how we call kubernetes-client, not on what the cluster contains afterwards. It passes if the selector string is right and the semantics are wrong, and it needs editing every time the client call shape changes. Meanwhile the e2e that does adjudicate the real thing is gated on `nightly.yml` with `matrix.base == "official-image"`, so no PR touching destroy will ever run it. The net effect is that the weaker test guards the merge and the stronger one reports the morning after.

Concretely, what I would rather see: drop the MagicMock tests that the e2e already covers - the label-selector and owner-label ones - and either move the e2e into the PR-gated deploy lane, or keep a small number of unit tests only for the branches the e2e genuinely cannot reach (the `namespace == "default"` short-circuit, an unparseable label set). That trades five brittle tests for one real one plus a couple of honest edge cases.

Not blocking the merge on it - the behavior is right and the e2e exists, which is more than most changes on this path get. But with #7709 deleting the teardown block this now inlines away from, destroy is about to have two authors in two PRs, and I would rather the thing guarding that on every PR be the cluster than a mock.

MusabMahmoodh (commented):
Re-read the three commits pushed since my approval. The ownership stamp moving
onto the path that actually creates the namespace is the right correction, and
counting a `Terminating` object as reclaimed rather than surviving is the
correct reading of the k8s delete contract - both make the guarantee in the
titletrue rather than approximately true. One question on the third commit.

## `!jac-scale.owner` re-opens the exact hole the PR closes, for one class of volume

`_destroy_shared_volume_pvcs` now deletes on two selectors:

```jac
selectors: list[str] = [
    f"{shared},jac-scale.owner={k8s_safe_name(app_name)}",
    f"{shared},!jac-scale.owner"
];
```

The first is the new isolation guarantee and is exactly right. The second
deletes every shared-volume PVC that carries no owner label - which is every
shared volume created before this PR merges.

The scenario the PR exists for: namespace `prod` has app A and app B, both
deployed on the current release, so neither shared PVC is labelled. This lands.
App A redeploys (its PVC gets an owner label), app B does not. `jac scale
destroy` on app A now deletes app A's own volume via selector 1 **and** app B's
unlabelled volume via selector 2. B's data is gone, silently - no log line on
that branch, and the destroy reports success.

I read this as a deliberate back-compat choice so that destroy still cleans up
pre-PR volumes rather than orphaning them, and that is a real concern. But it
means the isolation the title promises does not hold across the upgrade window,
in exactly the multi-app namespace the feature is for, and the failure mode is
data loss rather than a leaked object.

Three ways out, cheapest first:

1. Drop selector 2 and accept that pre-PR shared volumes are orphaned until
   their app redeploys once. Leaked PVC beats deleted PVC, and the leak
   self-heals on the next deploy.
2. Keep selector 2 but only when the namespace is owned by this app -
   `namespace_is_owned(core_v1, namespace, owner)` is already imported here and
   already answers the "is this namespace exclusively mine" question the rest of
   the PR turns on. An adopted namespace would then never have its unlabelled
   volumes touched.
3. Keep it, and log each PVC deleted through selector 2 by name, so an operator
   who loses B's volume can at least find out why.

(2) matches the PR's own thesis most closely and reuses a predicate that is
already there.

## Checked and not flagged

- `ensure_namespace_exists(namespace, owner)` losing its `owner: str = ''`
  default is a signature break, but I grepped the branch and all four in-tree
  call sites pass the second argument. Making it required is the better shape -
  the empty default was how the unowned-namespace case got created in the first
  place.
- The `jac-scale.owner` label value goes through `k8s_safe_name` at both the
  write site (`manifest_builder.jac`) and every read site, so a hyphenated or
  uppercase app name matches itself. No asymmetry.
- `namespace_is_owned` returning `False` for an empty owner keeps an
  unattributed caller on the conservative branch.

Approval stands; treat the above as the one thing I would want settled before
merge rather than a block.

MusabMahmoodh (commented):
Re-reviewing after the weekend, because the ground under this PR moved and it
is not your doing.

**#7709 merged on 2026-08-28 at 16:45Z and deleted the same block this PR
deletes.** It removed `KubernetesTargetBase.destroy`, `_destroy_component`,
`_destroy_application`, `_destroy_databases`/`_destroy_database` and
`_wait_for_deletion` from `kubernetes_target.jac` - 361 lines, which is within
one line of this PR's own `@@ -231,361 +229,6 @@` hunk. Your last merge from
main was 11:29Z the same day, about five hours before it landed.

Current state: `mergeable=false`, `mergeable_state=dirty`. GitHub is not
showing it on the PR page yet but the API is unambiguous.

**What this means concretely, and I think it is good news for you:**

- The deletion half of this PR is already on `main`. After a rebase, the
  `-361` disappears from your diff and what remains is the part that is
  actually yours - `_destroy_fleet_workloads`, the ownership labelling
  (`jac-scale.owner`), `namespace_is_owned`, the PVC/bundle sweep and the e2e.
  A `+797/-385` review becomes a much smaller and much more reviewable change.
- Worth re-checking after the rebase: #7709 also rewired monolith teardown to
  call `destroy_http_activation_for_target`, and added a `KEDAAutoscaler`
  sweep whenever `autoscaler_engine != "keda"`. Both live in the teardown path
  you are restructuring, so please confirm your `_destroy_fleet_workloads`
  still runs them - losing the KEDA sweep would orphan `InterceptorRoute` and
  `ScaledObject` resources on teardown, which is the bug #7709 just fixed.

**On #7968 itself:** I have posted a note there too. The issue body tells the
implementer to "do what the base class does: call `_destroy_databases`, delete
the PVC and `delete_namespace`" - that reference implementation no longer
exists, so the issue no longer describes a followable fix. Your PR is now the
only place that intended teardown order is written down, which raises its
value rather than lowering it.

**Not re-raising:** my five findings from the Aug 26 review are all still
resolved as far as I can tell, and the "count a Terminating object as
reclaimed, not as surviving" commit is the right call - a Terminating object
is a deletion that has been accepted, and asserting otherwise would make the
test flaky against a real API server.

**One question rather than a finding:** the new workflow step sets
`JAC_SCALE_BINARY_PATH` to the workspace binary. For a destroy-reclamation
e2e inside this repo that looks correct to me, since the whole point is to
exercise *this* checkout's deploy path - I only flag it because that variable
has bitten app-level test rigs before, where it silently pins a stale binary.
If the intent is "test this checkout", ignore me.

Happy to re-review promptly once it is rebased - given it is now blocked on a
merge conflict rather than on substance, I would rather not have it sit.

jac/jaclang/scale/deploy/target/kubernetes/target.jac
```diff
@@ -1883,17 +1876,374 @@ obj KubernetesTarget(KubernetesTargetBase) {
...
+        co_tenants: list[str] = (
+            self._foreign_app_resources(app_name, namespace, apps_v1, core_v1)
+                if created_namespace
+                else []
+        );
+        sole_tenant: bool = created_namespace and not co_tenants;
+
+        if sole_tenant {
+            if monitoring_reclaimed {
+                self._reclaim_namespace(namespace, core_v1);
```
MusabMahmoodh: If this delete comes back 403 (RBAC that covers namespaced objects but not `delete namespaces`, which is a common shape) it raises straight out of `destroy`: fleet and monitoring are already gone, `_destroy_in_shared_namespace` never runs, and the postgres StatefulSet, both PVCs and the secrets stay. Re-running hits the same 403 because the label and the sole-tenant probe read the same, so there is no way to reclaim the database through the tool. Sole tenancy is already established at this point, so falling through to the in-namespace teardown is safe:

```suggestion
            if monitoring_reclaimed {
                try {
                    self._reclaim_namespace(namespace, core_v1);
                    return;
                } except ApiException as e {
                    if e.status != 403 { raise; }
                    if self.logger {
                        self.logger.warning(
                            f"Not permitted to delete namespace '{namespace}'; "
                            f"removing this app's resources inside it instead",
                            {"app_name": app_name}
                        );
                    }
                }
            }
```

jac/jaclang/scale/tests/deploy/test_destroy_reclaim.jac
```diff
@@ -0,0 +1,360 @@
...
+    assert f"{APP}-external-db" in deleted , (
+        "the secret holding the external database connection string stayed in "
+        "a namespace this destroy does not reclaim"
+    );
+}
+
+
+test "removing the application names the co-tenant whose workloads went too" {
+    (apps_v1, core_v1) = _co_tenant_apis();
+    logger = unittest.mock.MagicMock();
```
MusabMahmoodh, line 277: This and the two tests after it pin the co-tenant warning through `logger.warning.call_args_list` on MagicMock clients. Scenario C in `destroy_reclaim_real_e2e.sh` already deploys `billing` beside `todo` and destroys the owner last, which is exactly the path that emits this line; capturing `destroy_app`'s output there and grepping for `Another app has resources in` asserts the same thing against the real cluster, and scenario A is the no-warning case. Not blocking now that the e2e is in the PR lane, but the mock count on this surface went from eight to thirteen.

MusabMahmoodh (approved):
Re-read after the six commits since my last pass. The three asks from Aug 31 are all closed: rebased clean, the KEDA sweep survived, and the e2e now runs in `test-scale-k8s` on every scale PR (16 min on head, real MicroK8s, both directions of co-tenancy, and it fails on main because main never calls `delete_namespace`). The four red checks were on the previous head and are the Sep 2 runner loss plus a compiler timeout in a file this PR does not touch; `a1b366aa` is 28/28 green.

One real gap left, inline: the 403 branch on the namespace delete strands the database with no retry path. Five-line catch. The other comment is the standing note about mocks where the e2e already has the setup. Approving; the 403 one is worth landing before merge.

MusabMahmoodh (commented):
The 403 fix is the right shape. `_reclaim_namespace` returning False instead of letting the exception escape means a refused namespace delete now falls through to the per-resource sweep, so the statefulset and its PVC are reclaimed instead of being stranded with no way to get them back by re-running. Putting the co-tenant assertions into `destroy_reclaim_real_e2e.sh` rather than only into the unit file is the right level for them, and `destroy_said` folding whitespace before matching is the sort of thing that stops a wrapped console line turning into a flake later.

One thing I want to understand before this goes in.

This commit deletes `test "an adopted namespace still names the co-tenant whose workloads went too"`, which was the only case covering the `owned=False` path. What survives is `test "removing the application names the co-tenant whose workloads went too"`, and that one runs with the default ownership and `component="application"`. The adopted namespace is the half of this PR the title leads with, and warning by name is the whole reason a co-tenant operator knows to redeploy. Was that deletion deliberate, and if it was, what covers the adopted path now? If the answer is that the two collapsed into one code path, the assertion is nearly free to keep with `owned=False` and it is the one I would least want to lose.

Separate and not yours to fix here: the new case is `MagicMock` all the way down, same as the rest of the file. That is the file's existing shape, not something this PR introduced, and you did put the real coverage in the e2e script, which is the part that would actually catch a regression. Worth saying once so it does not calcify: `test_destroy_reclaim.jac` is a mock-driven suite in a subsystem where the rule is real objects or real stand-ins, and it is a reasonable candidate for a conversion pass of its own rather than a bigger diff on this one.

PR #8711, opened 2026-08-25
feat(scale): pool the wire connections and run read-only units on the read tier

marsninja:
## The problem

Two defects sat on the hot path of every request the scale runtime serves:

1. **A fresh `PgStore` per request.** `request_context_middleware` built a new store for each request, paying a TCP connect plus a full SCRAM handshake (4096 pbkdf2-SHA256 rounds of client-side CPU) and tearing it down at request end. Under load this churned ~7 Postgres sessions per second into the provisioned instance.
2. **Everything ran SERIALIZABLE.** #8102 built the read-tier machinery (`REPEATABLE READ READ ONLY`, zero predicate locks), but #8101 removed its only caller when it deleted the eager root probe. Since then every request - health probes and pure docs reads included - took SIReadLocks per row and aborted against background writers like a docs-sync job.

## What this PR does

**Connection pool** (`jaclang/data/pgpool.jac`, new): a process-wide wire-connection pool keyed by conninfo. `PgStore._connect` checks the pool before dialing; `PgStore.close` sanitizes the connection (ROLLBACK if a transaction is open, `UNLISTEN *` + notification drain if the side connection ever listened) and parks it instead of closing. Checkout probes the socket for liveness and discards stale entries. Knobs: `JAC_DB_POOL` (default on), `JAC_DB_POOL_MAX` (16 per key), `JAC_DB_POOL_IDLE_S` (300).

**Read-tier units** (`server/session.jac` + OCC loop): units of work now carry a key (`unit_enter(key)`), derived from the walker/function identity at the endpoint layer. A keyed unit the process has never seen writing opens `REPEATABLE READ READ ONLY`. The first write routes through the `_ensure_txn` / `_flush_locked` guards, records the key as a writer, rolls back, and raises `ReadOnlyRetry` (a `WriteConflict` subclass); the OCC loop catches it without consuming a conflict attempt and re-runs the request at SERIALIZABLE. Registration is TTL'd (`JAC_DB_RO_WRITER_TTL_S`, 300s) so a transient heal - a fingerprint-drift root repair, a dangling-ref cleanup - does not demote a hot read path for the life of the process. Unkeyed units keep the old behavior verbatim, and `JAC_DB_RO_UNITS=0` disables classification entirely.

Write correctness is preserved the same way #8102 argued it: a read transaction can never carry a write (the guards fire before any write statement is issued), the retry re-runs the whole unit at SERIALIZABLE after `abort()` drops snapshot-read anchors (new - so the re-run re-reads them under SSI tracking), and version OCC on upsert stays the final arbiter. The OCC loop's unit re-entry is now depth-guarded; it previously double-entered on mid-walker conflicts.

**Supporting fixes**
- The authenticated root probe in the scale middleware runs inside a read scope again (the #8102 intent that #8101 dropped).
- The provisioned Kubernetes Postgres gets resource requests/limits and production server settings (`shared_buffers` 256MB, `work_mem` 8MB, `wal_compression` on, `max_wal_size` 2GB, `checkpoint_completion_target` 0.9, `random_page_cost` 1.1), all overridable via `[scale.database]` (`postgres_settings` table, `postgres_cpu_*`, `postgres_memory_*`).
- The microservice gateway serves the client SPA at `GET /docs` when the app ships a client bundle, matching the single-server contract where client routes shadow the API docs page. Headless fleets keep Swagger there.

## Measured

On a kind cluster running the jaclang.org site as a fleet (gateway + 3 app replicas + provisioned Postgres), 60s of mixed docs traffic at c=32 through the gateway, same binary, env flags off vs on:

| metric | flags off | flags on |
|---|---|---|
| doc_page p95 / p99 | 9,220 / 12,753 ms | **1,985 / 2,556 ms** |
| docs_status p95 | 8,686 ms | **1,709 ms** |
| Postgres sessions churned / min | ~427 | **~45** |
| read-path serialization aborts | retried constantly | none |

With the site's whole-corpus read models also cached (a separate app-side change), the same fleet serves **97 rps with zero errors** against 6.6 rps with 502s and timeouts before.

## Tests

`jac/tests/runtimelib/test_scale_db_stack.jac` (new, 6 tests) pins pool reuse, sanitization on release, the disable switch, read-tier units, first-write retry, and writer registration. The existing isolation (`test_read_path_isolation`), unit-of-work (`test_pg_uow`), session, backplane, pg-job-store, and serve suites stay green.


MusabMahmoodh (commented):
Read this end to end - the pool, the read-tier classification, the OCC interaction, and the provider changes. The read-tier half is convincing: the guards fire before any write statement, `abort()` now drops the snapshot-read anchors so the re-run re-reads them under SSI tracking, and the depth guard on unit re-entry fixes a real double-enter. I traced the retry convergence too - marking the whole unit stack in `_register_ro_writers` rather than just the writing unit is what bounds it at two passes, since otherwise the outer unit would re-open read-only and trip again. Worth a comment in the code, because it reads like over-broad bookkeeping until you walk that path.

Four things.

**1. `pool_key` puts the database password in a process-global dict key, and `pool_stats()` hands those keys out.**

```jac
def pool_key(info: any) -> str {
    where = info.unix_socket_dir or info.host or "";
    return f"{where}:{info.port}:{info.user}:{info.database}:{info.password}";
}
```

`pool_stats` is exported in `__all__` and returns `{key: len(bucket)}`. Nothing calls it in this diff, which is the good moment to change it - the first admin endpoint, metric label or debug log that surfaces pool depth prints the password. The identity is the same if the secret is folded in as a digest:

```jac
import hashlib;

def pool_key(info: any) -> str {
    where = info.unix_socket_dir or info.host or "";
    secret = hashlib.sha256((info.password or "").encode()).hexdigest()[:16];
    return f"{where}:{info.port}:{info.user}:{info.database}:{secret}";
}
```

**2. The `GET /docs` change is the only route-contract change here and it has no test.**

The six new tests in `test_scale_db_stack.jac` all cover the pool and the read tier, which is the right coverage for the thing the PR is named after. But `microservice_gateway.impl.jac:770` changes what every fleet shipping a client bundle serves at `/docs`, and nothing exercises it - not the SPA-present case, not the headless case that is supposed to keep Swagger, not the non-GET case. It is disclosed in the breaking note, so this is not about surprise; it is that a silent regression here looks like "the docs page went blank" in production and there is nothing to catch it.

It also seems like the one piece that does not belong in this PR. Everything else is one story about how the runtime talks to Postgres; this is a routing decision that happens to be nearby. Splitting it out would let it carry its own tests and its own breaking note rather than sharing one with a connection pool.

**3. `_park_conn` sanitizes the transaction and the listener but not the session.**

ROLLBACK and `UNLISTEN *` cover the two states this store creates deliberately. The one it creates incidentally is `SET lock_timeout = "5s"` in `_apply_schema_ddl`, reset in a `finally` - and the case where that reset does not run is exactly the case where the connection is about to be closed and therefore parked. The next borrower then silently inherits a 5s lock timeout. One statement in the sanitize sequence closes the whole class rather than that one instance:

```jac
if in_txn {
    conn.run("ROLLBACK");
}
conn.run("RESET ALL");
if listened {
    conn.run("UNLISTEN *");
    conn.notifications.clear();
}
```

**4. The demotion is per-process, and a fleet is not one process.**

`_unit_writers` is process-local, so each replica has to discover independently that a unit writes, by paying one `ReadOnlyRetry` round trip. Self-healing and cheap in steady state, but it means the "read-path serialization aborts: none" row is a warm-process measurement. On a rolling deploy every replica re-learns the whole write set at once, against a live workload. Was that visible in the kind run, or did the 60s window start after warmup? Not asking you to build cross-replica state for it - just want to know whether the cold window is a blip or something worth a note in the tuning knobs.

Two things I checked and they are fine, noting them so nobody re-raises them: `resp` is pre-initialized at `microservice_gateway.impl.jac:740`, so the non-GET `/docs` path cannot `NameError`; and `_unit_writers` is not an unbounded cache - `unit_writes_recorded` pops on expiry and the key space is the app's walker and function surface rather than its traffic.

PR #8717, opened 2026-08-26
fix(tools): unparse mapping star patterns with ** rest marker

### LINKED ISSUE #8641: [Bug] py2jac crashes on match mapping star pattern (`case {"k": key, **rest}`) with cascading parse errors
**What's wrong**

`jac tool py2jac` fails with rc=1 and a cascade of misleading parse errors on valid Python using a match mapping star pattern (`case {"k": key, **rest}:`). The Jac grammar's mapping-pattern parser only accepts literal (INT/FLOAT/STRING) keys and has no `**rest` support, so the whole file fails to transpile with ~18 cascading errors pointing at the wrong columns.

(Note: a companion probe — `try/else/finally` without `except` — turned out to be invalid Python; CPython rejects it identically, so it is *not* a py2jac bug and is not part of this issue.)

**Minimal repro**

```python
# match_star.py
def handle(d):
    match d:
        case {"k": key, **rest}:
            return key, rest
    return None
```

```
$ jac tool py2jac match_star.py -o out
✖ Error: .../match_star.py, line 1, col 60: Expected literal (INT, FLOAT, or STRING) as mapping pattern key
✖ Error: .../match_star.py, line 1, col 61: Expected ':', got 'NAME'
✖ Error: .../match_star.py, line 1, col 65: Unexpected token in expression: '}'
... (18 errors total, columns offset into the whole one-line file)
RC=1
```

**Expected vs actual**

- Expected: either a correct transpilation to Jac's mapping pattern, or a single clean diagnostic like "py2jac: mapping star patterns (**rest) are not supported yet" at the right location.
- Actual: rc=1 with a wall of cascading parse errors ("Missing '}'", "Expected ':'", keyword-escape suggestions) that obscure the real gap.

**Suggested direction**

Add `**<name>` (and bare `{}` / double-star capture) to the Jac mapping-pattern grammar if the language intends parity with PEP 634; otherwise add an explicit pre-flight check in the py2jac AST walk for `ast.MatchStar`/mapping rest patterns emitting one targeted error. Related: #6645 covers nested class patterns in match.

**Environment**

- jac from source (branch `fix-pg-initdb-resilience`), Linux, Python 3.14
- Verified live; full output above.

chess10kp:
Closes #8641

## Root cause

The Jac grammar, AST, and all backends already support PEP 634 mapping star patterns (`case {"k": key, **rest}:`), but `UnparsePass.exit_match_star` emitted a single `*` for every `MatchStar` regardless of the `is_list` flag. For a mapping rest pattern (`is_list=False`) the emitted Jac text was `*rest` instead of `**rest`, so py2jac's re-parse of its own output failed with cascading E0045 diagnostics ("Expected literal (INT, FLOAT, or STRING) as mapping pattern key") pointing at wrong columns.

## Fix

`exit_match_star` now emits `"*" if nd.is_list else "**"`, mirroring `NormalizePass.enter_match_star` which already selects `STAR_MUL` vs `STAR_POW` correctly. One line; no grammar/AST/backend changes were needed.

## Test

- New fixture `py2jac_match_mapping_star.py` and test `py2jac mapping star pattern round-trips`: unparses Python with both string-key + int-key `**rest` patterns, asserts the `**rest` marker survives, and asserts the emitted Jac re-parses with zero errors (the layer where #8641 broke).
- Forward direction is already covered by `tests/compiler/backends/native/fixtures/match_patterns.jac` (`case {"a": v, **rest}:`) and the native/ES lowering tests.


christianwilkins (commented):
Reviewed head `eb2c3da98754173edcf9290725f3a67eec2fd504`: the complete patch, Python MatchMapping/MatchStar conversion and the unparser branches. No blocking unparser defect found. The existing is_list distinction matches the marker chosen by this change. No code changes made.

Independent local validation: the new Python mapping-pattern conversion round-trips to Jac without diagnostics. An additional mixed mapping/list probe preserves **rest for mappings and *tail for sequences; changing the emitted mapping marker back to one star makes the parser reject the result. These probes call the production conversion/unparse/formatter APIs on this head, without method substitutions.

The older CI handoff in this thread describes different failures. Current-head run 33577056481 reports 1516 passed, 1 failed and 4 skipped in the native lane. Its sole failure is the set-scaling test during fixture compilation: CompilerSourceError says symbol_utils was re-entered while Python was still executing that module. It never reaches the scaling comparison. This is also visible on #8566, but I have not independently reproduced it on base and am not certifying it as a baseline failure.

Not ready to merge while that lane and the aggregate are red. The author needs to resolve or establish the baseline for the current import re-entry failure and obtain green CI, then maintainers can review for merge.

Issue #8693, opened 2026-08-25
scale: discovered .sv.jac services get no JAC_SV_FILE, so every deployed fleet crash-loops on 'couldn't read <svc>.jac'


udithishanka:
## Summary

On the 0.34.x line, `ManifestBuilder.generate_service_pod_specs` calls `resolve_routes` to merge auto-discovered `.sv.jac` services into the fleet, but never calls the paired `resolve_service_files`. A discovered service therefore gets a route and a Deployment but no `JAC_SV_FILE`, so the pod entrypoint falls back to `find . -name "<svc>.jac"`. That fallback can only ever match a module literally named `<svc>.jac`, and discovery's whole job here is to register modules named `<svc>.sv.jac`. Every such service crash-loops with `couldn't read '<svc>.jac': No such file or directory`, and the deploy fails on fleet readiness.

The local microservice path does resolve the file map: `plugin.jac:84` and `plugin.jac:397` both call `resolve_service_files` and store it as `ms_config["service_files"]`. Only the Kubernetes deploy path skips it.

## Environment

| | |
|---|---|
| jac | `0.34.15  (Darwin arm64)`, released binary from `jaseci-labs/jac` |
| Repro host | Darwin 25.3.0 arm64, clean scratch dir outside any jaseci checkout |
| Also observed | EKS `jaseci-cluster`, namespace `jb-0987a6d2-keda-test`, image `jaseci/jaclang@sha256:0be93077226bad4f8edb03e84808978347c17914f2fb006125268af4143a4c85` |
| `0.36.1` | Crash shape gone, defect intact. See "Status on 0.36" below. |

## Reproduction (executed, not hypothetical)

A stock four-file web app. No `[scale.microservices]` config of any kind, no deployer involved.

```bash
mkdir -p /tmp/svfile-repro/{services,components} && cd /tmp/svfile-repro

cat > jac.toml <<'EOF'
[project]
name = "demo"
entry-point = "main.jac"
kind = "web-app"
EOF

cat > main.jac <<'EOF'
import from .components.Widget { Widget }

def app() -> any {
    return <Widget />;
}
EOF

cat > services/thing.sv.jac <<'EOF'
node Item {
    has label: str;
}

def:pub list_items() -> list[str] {
    return [i.label for i in [root --> (`?Item)]];
}
EOF

cat > components/Widget.jac <<'EOF'
sv import from ..services.thing { list_items }

def Widget() -> any {
    return <div>widget</div>;
}
[...]

PR #8723, opened 2026-08-26
fix(scale): scale-to-zero Kubernetes workloads are no longer reported as failed

### LINKED ISSUE #8694: jac-scale: deploy tooling has no concept of an intentionally zero-replica KEDA service (dry-run false positive + 20 min rollout hang)
## Summary

While reproducing a report that KEDA-based scale-to-zero only appeared to work
for the gateway (root cause: `extra_triggers` is gateway-scoped by design,
unrelated to this issue), I stood up a local kind cluster with KEDA and
verified that a plain microservice with its own `[[services.NAME.triggers]]`
and `hpa.min = 0` genuinely scales to zero and wakes back up correctly through
native KEDA behavior. In doing so I hit two separate, real bugs in the deploy
tooling itself, both confirmed live, that make that already-working config
path much rougher to actually use than it needs to be. Filing both here since
they share a root context (the deploy pipeline has no concept of "this
service is intentionally at zero replicas") and are likely worth fixing
together.

Repro config used for both (KEDA engine, one microservice with a `redis`
trigger and `hpa.min = 0`, gateway and a second service left at defaults):

```toml
[scale.kubernetes]
autoscaler_engine = "keda"

[scale.microservices.services.order_service.hpa]
min = 0
max = 3

[[scale.microservices.services.order_service.triggers]]
type = "redis"
name = "order-queue"

[scale.microservices.services.order_service.triggers.metadata]
address = "redis.default.svc.cluster.local:6379"
listName = "orders"
listLength = "1"
```

---

## Problem 1: `--dry-run` flags a correct `hpa.min = 0` config as a hard error

`_check_hpa` (`jac/jaclang/scale/runtime/cli/diagnostics.jac:172-190`):

```jac
hpa_min = iget(hpa, "min", 1);
...
if hpa_min < 1 {
    out.append(
        Diagnostic(
            severity="error",
            service=svc_name,
            message=f"HPA min={hpa_min}; must be >= 1"
        )
    );
}
```

This check rejects `hpa.min = 0` unconditionally. It never looks at
`autoscaler_engine` or whether the service has any `triggers` configured. That
makes it correct for the plain HPA engine, which structurally cannot scale
below 1, but wrong for the KEDA engine with a real event-driven trigger, where
`min = 0` is the documented, intended way to enable scale to zero.

**Observed:** running `jac scale deploy --dry-run --show-yaml` against the repro
config above prints:

```
X 1 error, 0 warnings  -  fix errors before deploying
  order_service
    X HPA min=0; must be >= 1
```

**Impact:** confirmed this does not block a real deploy. `has_errors(diagnostics)`
(`jac/jaclang/scale/runtime/realizer/kubernetes_realizer.jac:119-131`) only sets
the process exit code inside the `if dry_run` branch; a real deploy calls
`deployment_target.deploy(app_config)` directly and never runs this check. So
the risk isn't a broken deploy, it's that `--dry-run --show-yaml` is the
workflow the CLI's own examples recommend running first, and it actively tells
the operator their correct config is wrong. That's likely to make someone
revert `min = 0` back to `1` before ever attempting a real deploy, silently
defeating scale-to-zero.

**Why this check exists, and why it must not simply be removed:** this
validation was first added in #6024, well before KEDA existed, back when the
plain HPA engine was the only autoscaler in jac-scale and Kubernetes HPA
genuinely cannot go below `min=1`. It was, and still is, a correct guardrail
against a real misconfiguration on the HPA engine, and on a KEDA service with
no real trigger (there, `min=0` would mean the fallback CPU/memory trigger,
which can't legitimately reach zero, so it's a config that will not do what
the operator intends). The bug is not that this check exists, it's that it
was never narrowed when KEDA's `#6534` introduced a case where `min=0` is
correct. Any fix here should add a carve-out, not delete the check.

**Proposed solution:** the resolved config this check needs already exists and
is already flowing past it unused. `generate_manifests`
(`jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac:346-376`)
builds `bundle["autoscalers"][svc_name] = self._get_autoscaler_config(svc_name)`
for every service, the same resolved `min`/`max`/`triggers` dict the real
deploy uses. `Plan.from_target`
(`jac/jaclang/scale/runtime/cli/plan.jac:355-416`) already calls
`target.generate_manifests(app_config)` and gets this back as `bundle`, it
even reads `bundle["autoscalers"]` two lines later to render the plan's `HPA:`
column. But it separately calls `Plan._load_ms_cfg()` to build `svc_configs`
straight from raw `jac.toml`, and that raw, unresolved dict is what gets
handed to `PlanValidator` for `_check_hpa`, ignoring the resolved config
sitting in the same function.

The fix should have `PlanValidator._check_hpa` read `hpa_min`/`triggers` from
`bundle["autoscalers"].get(svc_name, {})` instead of the raw `svc_cfg`, and
only skip the `< 1` rejection when that resolved `triggers` list is non-empty
**and** `autoscaler_engine == "keda"`. `autoscaler_engine` isn't part of
`bundle["autoscalers"]` (it's a separate field on `target.k8s_config`), so it
needs to be threaded into `Plan`/`PlanValidator` alongside the bundle, not
re-derived. This removes a duplicate config parser rather than adding a
third one: `Plan._load_ms_cfg`'s raw `svc_configs` becomes unnecessary for
this check once `bundle["autoscalers"]` is used instead.

---

## Problem 2: the rollout wait step hangs up to 20 minutes on a correctly zero-replica service

`_wait_for_fleet_rollout` (`jac/jaclang/scale/deploy/target/kubernetes/target.jac:502-585`):

```jac
def _wait_for_fleet_rollout(
    bundle: dict[str, any], timeout: int = 1200, require_full: bool = False
) -> None {
    ...
    while elapsed < timeout {
        ready = 0;
        for name in names {
            try {
                d = apps_v1.read_namespaced_deployment(name=name, namespace=ns);
                complete = deployment_rollout_complete(d)
                    if require_full
                    else (d.status.ready_replicas or 0) >= 1;
                if complete {
                    ready += 1;
                }
            } except Exception {
                _ = 0;
            }
        }
        if ready == total {
            ...
            return;
        }
        ...
    }
}
```

Every deployment in the bundle is checked against `ready_replicas >= 1`, with
no exception for a service that KEDA has correctly scaled to 0 by design. For
a service configured with `hpa.min = 0` and an idle trigger, `ready_replicas`
is legitimately always `0`, so this condition can never be satisfied for that
service specifically. The loop only returns early once every deployment is
"ready," so the whole rollout wait blocks on that one service until `timeout`
is reached: `1200` seconds by default, 20 minutes.

**Observed:** deployed the repro config above alongside the gateway and a
second service. Both other services reached `Running` within about 30
seconds. The CLI kept printing `Rolling out fleet: 2/3 ready` every ~10-90
seconds for 500+ seconds straight, with `order_service` correctly showing no
pod at all the entire time (KEDA had it legitimately at 0 replicas, `Active:
False`, queue empty). Had to kill the deploy process manually rather than
wait out the remaining ~19 minutes to the timeout.

**Impact:** this will hit anyone who configures per-service scale-to-zero on a
fresh deploy, not an edge case. It's arguably the bigger practical blocker of
the two, since it makes every such deploy take up to 20 minutes (and appear
hung/broken) even though the deploy is actually correct and complete once the
non-zero services are up.

**Why this check exists, and why it must not simply be removed:** the
`ready_replicas >= 1` gate traces back to #5102, also long before KEDA
existed. Its job is to make a deploy actually wait until pods are serving
before declaring success, and to catch a genuinely stuck or crash-looping
rollout instead of returning early on a fleet that never came up. That
purpose is still correct and still needed for every service that is not
intentionally at zero. The bug is that the check has no way to tell "this
service has zero ready replicas because it is stuck" apart from "this service
has zero ready replicas because KEDA correctly put it there." Removing or
loosening the check across the board would reopen exactly the stuck/crash-loop
case it was built to catch (this codebase already had to re-fix crash-loop
detection here once, in #7295/#7076). The fix needs to distinguish the two
cases, not stop checking.

**Proposed solution:** same underlying fix as Problem 1, reuse
`bundle["autoscalers"]` instead of re-deriving anything. The real deploy
already builds `bundle = self.generate_manifests(app_config)`
(`target.jac:245`) before calling `self._wait_for_fleet_rollout(bundle)`
(`target.jac:289`), so `bundle["autoscalers"][svc_name]` is already sitting in
the exact `bundle` this function receives as its own parameter, no new
plumbing required.

The function currently discards the `svc_name` key it needs: it iterates
`bundle.get("deployments", {}).values()` (`target.jac:511`) and only keeps
`dep["metadata"]["name"]` (the k8s deployment name, e.g.
`order-service-deployment`). It should iterate
`bundle.get("deployments", {}).items()` instead, keeping `svc_name` alongside
each deployment name, and build a `zero_replica_names` set from
`{name for (svc_name, dep) in ... if bundle.get("autoscalers", {}).get(svc_name, {}).get("min") == 0 and self.k8s_config.autoscaler_engine == "keda"}`.
In the readiness loop, a deployment in that set is treated as ready once its
Deployment/Service/ScaledObject applied without error and no crash-loop is
detected on it (the existing `_scan_for_crash_loop` machinery already run
later in the loop), rather than requiring `ready_replicas >= 1`. Every other
deployment keeps the current `ready_replicas >= 1` gate unchanged, so a
genuinely stuck non-zero-floor service still times out and fails as today.
This is conceptually the same fix category as the one already landing in
#7709 for `_wait_for_fleet_rollout` crashing entirely on an all-HTTP-activation
bundle (via `_load_kube_config` vs `_load_cluster_config`); that fix addresses
a crash for HTTP_ACTIVATION-only bundles specifically, this issue is the
general case for plain KEDA `min_replicas == 0` services that aren't going
through HTTP activation at all.

---

## Root cause: two guardrails predate scale-to-zero and were never revisited

Traced both checks through history:

- `_check_hpa`'s `min < 1` rejection was first added in #6024, and
- `_wait_for_fleet_rollout`'s `ready_replicas >= 1` gate traces back to #5102,

both well before KEDA (and therefore before any legitimate zero-replica case)
existed anywhere in jac-scale. At the time each was written, the assumption
they encode, "every deployment should end up with at least one ready
replica", was universally true, and both were correct, useful guardrails
against real misconfiguration and stuck rollouts.

KEDA landed in #6534 ("add KEDA autoscaler engine and extra_triggers config"),
which is the first point a `min=0` service became a legitimate, intended case.
That PR's diff touches only the trigger/autoscaler-spec construction in
`kubernetes_target.jac` and `microservice/target.jac`, not `diagnostics.jac`
or the rollout-wait function. Notably, the rollout-wait function was rewritten
again shortly after, in #6800, by a different author, for an unrelated reason
(fixing four separate `--no-image` deploy blockers), and the `>= 1` assumption
was carried forward unchanged even then.

So neither check is wrong to exist, and neither should be deleted. The gap is
that scaling-mode knowledge lives in one place (`_get_autoscaler_config`,
called once inside `generate_manifests` and stored as `bundle["autoscalers"]`)
and these two guardrails each re-derive their own, stricter assumption
independently instead of consulting it, even though `bundle["autoscalers"]`
is already sitting inside the exact function scope both of them run in
(`Plan.from_target` for Problem 1, `_wait_for_fleet_rollout`'s own `bundle`
parameter for Problem 2). Both fixes are "read the config that's already
being passed in" rather than "add a new resolver", which is what makes this
fixable architecturally instead of with two isolated patches: it removes a
duplicate parser (`Plan._load_ms_cfg`'s raw `svc_configs`, for the parts of
it this check needs) rather than adding a third one.

Developer-Linus:
## Issue

Fixes #8694.

Deploying a microservice configured to scale to zero with KEDA was broken end to end. `jac scale deploy --dry-run` rejected the config outright, a real deploy could hang for up to 20 minutes waiting for a service that was never going to start, and once deployed, checking its status reported it as failed instead of intentionally stopped. jac-scale had no way to tell a service that is idle by design apart from one that is actually broken.

## Summary

Fixes three places where jac-scale wrongly treated a Kubernetes service parked at 0 replicas by design (KEDA scale-to-zero) as broken. All three now share one predicate, so scale-to-zero is recognized consistently across dry-run validation, the deploy rollout wait, and live status checks.

## Problem

When a microservice is configured with `autoscaler_engine = "keda"`, `hpa.min = 0`, and a real KEDA trigger, three separate parts of jac-scale each re-derived on their own whether a 0-replica deployment was "scaled to zero on purpose" or "actually broken". They disagreed, and two of them got it wrong:

1. **Dry-run validation** (`jac scale deploy --dry-run`) always rejected `hpa.min = 0` with `HPA min=0; must be >= 1`, even for a valid KEDA scale-to-zero setup. This blocked deploying the exact configuration the feature is meant to support.
2. **Deploy rollout wait** would keep polling a legitimately-zero service until the deploy timeout (up to 20 minutes), because it had no way to tell "scaled to zero on purpose" apart from "never started".
3. **Live status check** (`get_status`, used by the SDK) reported `ResourceStatus.FAILED` for a scaled-to-zero service instead of `ResourceStatus.STOPPED`, which is misleading for any caller polling app health.

## Fix

Added one predicate, `is_scale_to_zero_eligible(autoscaler_cfg, engine)`, in `manifest_builder.jac`. It returns true only when all of the following hold:
- `min <= 0`
- the engine is `"keda"` (the plain HPA engine cannot scale below 1 replica)
- at least one real trigger is configured (CPU/memory alone can never reach zero)

All three call sites now use this same predicate instead of their own logic:
- `diagnostics.jac` (`PlanValidator._check_hpa`): only raises the HPA error when the service is not scale-to-zero eligible. The plain-HPA and no-trigger cases still error, with a clearer message explaining why.
- `target.jac` (`_wait_for_fleet_rollout`): tracks which deployments are eligible and settled at 0/0 replicas for two consecutive polls, with no crash signal from their pods. Once every deployment is either ready or confirmed parked at zero, the rollout finishes instead of waiting out the timeout.
- `kubernetes_target.jac` (`get_status`): reports `ResourceStatus.STOPPED` instead of `FAILED` when the deployment is at 0 replicas and eligible. This required resolving the Kubernetes Deployment's name (e.g. `worksvc-deployment`) back to the microservice's config key (e.g. `worksvc`) before the config lookup, since the two names differ once `k8s_safe_name()` mangling and the `-deployment` suffix are applied. `ManifestBuilder.autoscaler_config_for_deployment(deployment_name)` does this resolution (checking declared service routes and the gateway's synthetic name) before looking up the autoscaler config.

## Why this design

- **One predicate, one place to change.** Before this fix, each site had its own copy of the eligibility logic. Adding a fourth call site (or changing the rules) previously meant updating them all in sync. Now there is exactly one function to change.
- **Narrowed, not removed.** The original guardrails exist for real reasons: the plain HPA engine genuinely cannot scale below 1 replica, and a KEDA CPU/memory-only trigger genuinely cannot reach zero either. The fix narrows the check to exclude the legitimate KEDA scale-to-zero case, it does not disable the check.
- **Crash detection still runs first.** The rollout wait still scans pods for crash-loop failures before considering a deployment "settled at zero". A genuinely broken service that happens to be scale-to-zero eligible still fails the deploy.
- **Deployment name resolution lives in one place.** Rather than passing extra context through `get_status`'s signature, the name-to-config-key resolution is self-contained in `ManifestBuilder`, which already owns both the naming convention (`k8s_safe_name`) and the config lookup.

## Follow-up: single-app solo-route resolution (resolved upstream)

This PR originally added custom solo-route synthesizers (`_with_implicit_solo_route`, `_solo_aware_ms_cfg`, and an `AppConfig` parameter on `_solo_entry_file()`) so a single-app deploy's implicit route resolved the same way a multi-service one's did. Main's fleet refactor (merged into this branch 2026-09-18) unified single-app and multi-app fleet resolution and removed all three; the drift this section originally worried about (the same failure mode as the neighboring #8693 bug) can no longer occur, since there is no longer a separate solo-route code path to disagree with the rest of the fleet.

## Follow-up: http_activation scale-to-zero status (2026-09-18)

Main's KEDA HTTP interceptor work (#8779), merged into this branch today, added a second legitimate way for a service to sit at zero replicas: `http_activation` mode, whose default replica floor is `min=1` with no KEDA triggers at all. `_is_stopped_by_design` only checked KEDA metric-trigger eligibility, so an `http_activation` service parked at zero reported `FAILED` instead of `STOPPED`. Added `ManifestBuilder.http_activation_config_for_deployment`, mirroring `autoscaler_config_for_deployment`, and checked first in `_is_stopped_by_design`. The gateway is excluded the same way `_resolve_scaling` already excludes it, since it must never scale to zero.

## Known follow-ups (tracked, not blocking)

- **KEDA trigger-type validation**: an unknown or misspelled trigger type (e.g. `crno`) still qualifies as a "real" trigger, since there's no static allowlist of KEDA's 60+ built-in scalers (plus external ones) to validate against. The sound fix is deferring to KEDA's own `ScaledObject.status.conditions[Ready]` rather than maintaining a list that would immediately drift; that needs new live-status-read plumbing, a `keda.sh` RBAC grant (`ops_rbac.jac` currently has none), and hot-loop caching. Tracked in jaseci-labs/jac#9236.
- **`get_status` re-derives eligibility from whichever `jac.toml` the calling process loaded**: called from the SDK outside the app's own project directory, `ManifestBuilder` falls back to the `get_scale_config()` singleton, which can be the wrong project's config. This was `FAILED` (wrong, but safe-by-default) before this PR too, so it isn't a regression, but it's now the one remaining case where `get_status` can disagree with the rollout wait. #8464's `inactive` state answers a related question a third way; worth settling which is authoritative before extending either further.

## Changed files

| File | Change |
|---|---|
| `manifest_builder.jac` | Added `is_scale_to_zero_eligible`, `autoscaler_config_for_deployment`, `_service_key_for_deployment`, `http_activation_config_for_deployment` |
| `kubernetes_target.jac` | `get_status` reports `STOPPED` for a zero-replica deployment eligible via KEDA scale-to-zero or `http_activation` |
| `target.jac` | `_wait_for_fleet_rollout` recognizes deployments settled at zero and stops waiting on them |
| `diagnostics.jac` | `_check_hpa` skips the HPA error for a scale-to-zero eligible service, with a clearer error message otherwise |
| `plan.jac` | Threads `autoscaler_engine` through to `PlanValidator` |
| `tests/deploy/test_factories.jac` | Tests for `get_status` STOPPED/FAILED cases (KEDA and `http_activation`) and the deployment-name-to-config-key resolution |
| `tests/microservices/test_fleet_rollout.jac` | Tests for the rollout wait recognizing scale-to-zero vs. a real crash loop |
| `tests/deploy/test_scale_to_zero_validation.jac` | Table-driven test for the dry-run validator's `hpa.min = 0` acceptance/rejection across engines and trigger types |

## Tests

- `test_factories.jac` (37 tests), `test_fleet_rollout.jac` (20 tests), and `test_scale_to_zero_validation.jac` (1 table-driven test, 9 cases) all pass, covering: STOPPED vs FAILED status for both KEDA and `http_activation`, dry-run acceptance/rejection across engines and trigger types, rollout-wait completion vs. timeout, crash-loop still failing, and the deployment-name resolution.
- Verified live against a local Kubernetes cluster (kind) with KEDA installed:
  - `jac scale deploy --dry-run` on a service with `hpa.min = 0` and a KEDA trigger: no error (previously failed with `HPA min=0; must be >= 1`).
  - A real deploy with one service configured to start at 0 replicas completed successfully, logging `Fleet ready: 2/3 deployments serving, 1 parked at 0 replicas by design (scale-to-zero, no crash signal observed)` instead of waiting out the timeout.
  - `get_status` on the same live deployment returned `STOPPED`.


MusabMahmoodh (changes requested):
Read this against #8694 and then against the surrounding code on `main`. The
three-places-disagree diagnosis is right, and collapsing them onto one shared
predicate is the correct shape - that is the fix I would have wanted. One
finding blocks it, because it inverts the guarantee the PR is written to
provide.

## 1. The predicate does not implement the rule this PR documents

The note added to `jac-scale-kubernetes.md` states the rule precisely:

> Setting `min_replicas` (or a per-service `hpa.min`) to `0` is only valid
> under `autoscaler_engine = "keda"` with at least one configured trigger
> **other than `cpu` or `memory`** -- CPU and memory utilization cannot be
> measured with zero running pods, so those triggers alone can never scale
> back up from zero.

The dry-run error string says the same thing: *"CPU/memory targets alone can't
reach zero"*.

`is_scale_to_zero_eligible` (`manifest_builder.jac`) never looks at trigger
type:

```jac
triggers = autoscaler_cfg.get("triggers", []);
return isinstance(triggers, list) and len(triggers) > 0;
```

`cpu` and `memory` are first-class trigger types in this codebase -
`keda_autoscaler.impl.jac:74` and `:83` build them, and there are
`DEFAULT_CPU_UTILIZATION_TARGET` / `DEFAULT_MEMORY_UTILIZATION_TARGET`
constants for them. So:

```toml
[scale.microservices.services.order.hpa]
min = 0
[[scale.microservices.services.order.triggers]]
type = "cpu"
```

passes `len(triggers) > 0`, and all three call sites accept it. The dry run
approves the config, the rollout wait exempts the deployment from its timeout,
and `resource_status` reports it `stopped`. That deployment can never wake -
it is exactly the case the note says is invalid - and after this PR it is the
one case that reports as healthy-and-parked instead of failed. Before the PR
it at least reported `failed`, which was wrong for the right reason; now it is
wrong in the direction an operator will not investigate.

Fix is contained: filter to triggers whose `type` is not in `{"cpu", "memory"}`
before the length test.

## 2. The test that would catch #1 asserts a case it does not run

`"get_status still reports FAILED at zero replicas when KEDA has no real
trigger"` passes `{"min": 0, "max": 3, "triggers": []}`, but its failure
message reads:

> `CPU/memory-only triggers can't reach zero, so this is a real failure`

An empty list is not a CPU/memory-only trigger list, and the two cases fail for
different reasons - one on `len(triggers) > 0`, the other on a type filter that
does not exist yet. As written the test passes on both the current code and the
fixed code, so it cannot fail on the thing it is named for.

Keep it, and add the case its message describes:

```jac
info = _get_status(
    target,
    {"min": 0, "max": 3, "triggers": [{"type": "cpu", "metadata": {}}]},
    replicas=0,
    ready=0
);
assert info.status == ResourceStatus.FAILED;
```

That one fails on this branch today and passes once #1 is fixed, which is the
test worth having.

## 3. `resource_status` now does config parsing on the status path, and can throw where it used to answer

The new branch in `kubernetes_target.jac` constructs a `ManifestBuilder`
per status call and asks it to rebuild the autoscaler config:

```jac
} elif replicas == 0
    and is_scale_to_zero_eligible(
        ManifestBuilder(k8s_config=..., logger=...)
            .autoscaler_config_for_deployment(app_name),
        self.k8s_config.autoscaler_engine
    ) {
```

Previously the zero-replica verdict came from data already in hand and could
not fail. Now a malformed `[scale.microservices]` block turns `get_status`
from "returns FAILED" into "raises", on a read-only status call an operator
runs when something is already wrong. Suggest wrapping the eligibility lookup
so any exception falls back to the old `FAILED` verdict - it degrades to the
pre-PR behaviour instead of taking the caller down.

All three new tests patch `kt_module.ManifestBuilder` with a MagicMock, so the
real builder is never driven from this call site and this path has no coverage.

## 4. `_service_key_for_deployment` has no coverage and looks wrong for single-app deploys

It strips a `-deployment` suffix and matches `k8s_safe_name(key)` against the
microservice `routes` keys plus `GATEWAY_NAME`. On a single-app deploy
`_ms_cfg()` carries no routes, so `candidates` is `[GATEWAY_NAME]`, the loop
misses, and it returns the bare `k8s_name` - which `_get_autoscaler_config`
resolves through `_service_config`, and for a single-app deploy that is not the
key the config is stored under. Eligibility comes back False and the deployment
still reports `failed`.

Is single-app KEDA scale-to-zero in scope for #8694? If it is, this is the
path it takes and it is the one path with no test - the mocked builder in all
three tests means `autoscaler_config_for_deployment` is never actually called.
If it is not in scope, worth saying so in the note, because the doc text reads
as though it covers both.

## 5. The rollout wait returns out of the middle of the loop

```jac
if confirmed_zero and (ready + len(confirmed_zero)) == total {
    if self.logger { self.logger.info(...) }
    return;
}
```

This returns from inside `with console.status(...)`, skipping whatever the
normal post-`while` completion path does (final status line, `healed_apps`
reporting). If that block does nothing an early return would miss, ignore this;
if it does, a fleet that legitimately parks at zero takes a different exit than
one that comes up serving, and the difference will not show up until someone
deploys one. Falling out of the loop with a flag reads the same and keeps one
exit.

The two-poll `zero_confirm_streak` before trusting a zero is a good call, and
rebuilding the dict from `settled_zero` each pass (rather than incrementing in
place) correctly drops a deployment that leaves the settled set. No issue there.

## Checked and not flagged

- `zero_floor_names` is keyed by deployment name while `bundle["autoscalers"]`
  is keyed by service name; the loop maps between them correctly.
- `PlanValidator.autoscaler_engine` defaults to `"hpa"`, so a validator
  constructed anywhere that does not pass it keeps the pre-PR strict behaviour
  rather than silently opening the gate. Right default.
- The `cast(any, target)` in `test_factories.jac` is a type-checker
  accommodation, not a behaviour change.

Happy to re-review quickly once #1 and #2 land - the structure is right and I
would like this to go in.

jac/jaclang/scale/tests/deploy/test_factories.jac
```diff
@@ -520,3 +525,127 @@ test "kubernetes_target falls back to in-cluster config when kubeconfig is unava
...
+    (apps, core) = _status_apis(replicas, ready);
+    with unittest.mock.patch.object(
+        kubernetes.client, "AppsV1Api", return_value=apps
+    ) {
+        with unittest.mock.patch.object(
+            kubernetes.client, "CoreV1Api", return_value=core
+        ) {
+            with unittest.mock.patch.object(target, "_load_cluster_config") {
+                with unittest.mock.patch.object(
+                    kt_module, "ManifestBuilder"
```
MusabMahmoodh: Patching `ManifestBuilder` at this seam is what let the cpu-trigger and single-app gaps from my earlier pass through with green tests - `autoscaler_config_for_deployment` and `_service_key_for_deployment` are never driven from this call site. The builder doesn't need faking: seed the scale config with the microservices block (test_plan.jac already does this) and let `get_status` construct the real one; the SimpleNamespace k8s API is the only boundary here and can stay. One real-builder test replaces these three and fails on the cpu-only-trigger case today.

jac/jaclang/scale/tests/microservices/test_plan.jac
```diff
@@ -326,6 +334,51 @@ test "validate: HPA bounds (min < 1, min > max, cpu_target out of 1-100)" {
 }
 
 
+test "validate: KEDA min=0 with a real trigger is not an error" {
```
MusabMahmoodh: These three construct `PlanValidator` directly, so the plumbing this PR added to plan.jac - `from_target` reading `autoscaler_engine` off the target and threading it into `validate()` - has no coverage; if that thread breaks, keda min=0 configs go back to erroring and nothing here trips. test_dry_run_plan_rate_limit.jac already has the exact shape that would: real `jac scale deploy --dry-run` subprocess, no cluster, no network. One dry-run test with a keda min=0 + trigger toml asserting a clean plan, and one without the trigger asserting the new error text, covers from_target, the validator, and the message end to end - and these three fold into it.

jac/jaclang/scale/tests/microservices/test_fleet_rollout.jac
```diff
@@ -399,3 +424,79 @@ test "a terminating pod is invisible to the scanner (no false re-abort)" {
         f"only the plain timeout may fire: {err}"
     );
 }
+
+
+test "a KEDA min=0 deployment settled at zero completes without timing out" {
```
MusabMahmoodh, line 432: Genuine question rather than a change request: has this been run once end to end against a real KEDA install (kind + the keda chart) - deploy, watch it scale to zero, `get_status`? All three surfaces are tested against hand-built `SimpleNamespace` deployments, so nothing confirms KEDA actually parks a Deployment the way `_dep_zero` models it (`status.replicas` 0 vs unset during scale-down, observedGeneration lag, and so on). If yes, a line in the PR body saying what was observed is enough for me.

MusabMahmoodh (commented):
Second pass, on the test shape - this morning's findings stand. To be explicit on one thing I was weighing: the fix itself is the right mechanism, one predicate with three consumers, not a workaround. The comments here are one theme: each new test replaces the exact seam this PR changed with a fake, so the PR's own plumbing is the least-covered code in it. Fewer, realer tests would close that.

jac/jaclang/scale/tests/deploy/test_factories.jac
Developer-Linus: Agreed, and this seam is exactly what let those gaps through. Replaced the MagicMock-based tests with ones that seed the scale config and drive a real ManifestBuilder end to end; the SimpleNamespace k8s API stays as the only boundary. See "get_status resolves zero-replica eligibility through a real ManifestBuilder: a real trigger parks under keda, a cpu-only or missing trigger still fails, and the plain HPA engine never parks", plus two new autoscaler_config_for_deployment tests. This now fails on the cpu-only-trigger case without the fix.

jac/jaclang/scale/tests/microservices/test_plan.jac
Developer-Linus: Good catch, and this is the shape I went with. Added _keda_dry_run_plan, which drives a real jac scale deploy --dry-run subprocess against a real project, no cluster and no network, exercising Plan.from_target for real. One test covers the clean-pass case (keda min=0 with a real trigger), and another covers the error cases (plain HPA engine, no trigger at all, and a cpu-only trigger). This exercises from_target, the validator, and the error message end to end, and the three direct-construction tests folded into it.

jac/jaclang/scale/tests/microservices/test_fleet_rollout.jac
Developer-Linus: Yes, verified against a real kind cluster with KEDA installed. Noted the results in the PR description: jac scale deploy --dry-run passes cleanly on a keda min=0 config with a trigger, a real deploy with one service configured to start at 0 replicas completes and logs the parked-at-zero message, and get_status on that live deployment returns STOPPED. Also added test_scale_to_zero_status_e2e.jac, a real-cluster e2e that deploys a zero-replica Deployment and checks get_status through the actual jac.toml config path, and wired it into the test-scale-k8s CI job so it runs on every PR going forward.

MusabMahmoodh (commented):
Re-reviewed the delta since my Aug 28 comment (four commits, Aug 29). The
status-reporting fix itself is right and I have no objection to it. One
finding on the solo-route work that came in with it, and I think it matters
more than its size suggests.

### There are now three places that synthesize the implicit solo route, and they disagree

On your head commit, `manifest_builder.jac`:

| site | line | entry file it writes |
|---|---|---|
| `_with_implicit_solo_route` | `:328`, via `:345` | `self._solo_entry_file()` |
| `_solo_aware_ms_cfg` | `:774`, via `:787` | `app_config.file_name` |
| `generate_service_pod_specs` | `:730`, **inline** | `app_config.file_name` |

Sites 2 and 3 are near-identical: both `if resolve_routes(ms_cfg)` guard, both
`self._solo_route(app_config)`, both `solo_cfg["file"] = app_config.file_name`,
both write back `services`/`routes`. Site 3 was already there; site 2 is new
and duplicates it rather than replacing it.

Site 1 is the interesting one, because it deliberately differs - your commit
message says "a live deploy's real boot file wins over the `[project]` entry
point", which is the correct behaviour. But that means **the same app gets a
different `solo_cfg["file"]` depending on which of the three paths built its
config**, and nothing in the code says which one is authoritative.

This is not hypothetical drift. #8693, filed last week and confirmed live on
`main` this morning, is the same failure on the neighbouring key: routes and
service files are resolved as a pair everywhere except the manifest builder,
which resolves one of them, and the result is a fleet that crash-loops on
`couldn't read <svc>.jac`. `solo_cfg["file"]` feeds `JAC_SV_FILE` through
`_build_env` (`:966`). A solo deploy that takes the site-3 path when it should
have taken site-1 boots the wrong module.

Suggestion: make `_with_implicit_solo_route` the only synthesizer, have the
other two call it, and let the entry-file precedence (real boot file, then
`[project]` entry point, then `main.jac`) live in `_solo_entry_file()` alone.
That is a consolidation, not a behaviour change, and it makes the precedence
you intended readable in one place.

### Smaller

- "stop the solo-route priming fix from also duplicating ..." (the commit
  subject is truncated on my side) suggests you already hit one instance of
  exactly this. Worth saying so in the PR body - it is the argument for the
  consolidation.
- The real-cluster status e2e wired into `test-scale-k8s` is the right call.
  Scale-to-zero reported-as-failed is only decidable against a real API
  server, and a unit test asserting on the status mapper would not have caught
  the original bug.

Not blocking on the status fix, which I would take today. Flagging the
solo-route consolidation as the thing I would rather see done here than
inherited by whoever next touches `JAC_SV_FILE`.

Developer-Linus:
Re Greptile's two P1 findings on the latest review:

1. **Rollout-wait can report success before a scale-up settles** (the settled-zero check didn't require `spec.replicas == 0`) — fixed in the latest commit: it now requires `spec.replicas == 0` alongside the existing generation-reconciliation guard, with a regression test covering the scale-up race.

2. **Invalid/mistyped KEDA trigger types still qualify as "real"** — confirmed legitimate, but deliberately not fixed in this PR. The only sound fix is deferring to KEDA's own `ScaledObject.status.conditions[Ready]` rather than trying to maintain a static allowlist of valid scaler types (KEDA has 60+ built-in scalers plus external ones — a hardcoded list would immediately drift). That requires new live-status-read plumbing, a new RBAC grant (`ops_rbac.jac` currently has no `keda.sh` permissions), and hot-loop caching — a real design change, not a patch, so it's out of scope for this already-large merge-resolution PR. Tracked in jaseci-labs/jac#9236.

Worst case without the fix: a workload with a genuinely invalid trigger type gets marked STOPPED and never reactivates — visible and detectable (the service just doesn't respond), not silent data loss, which is why it's deferred rather than blocking.

jac/jaclang/scale/deploy/target/kubernetes/kubernetes_target.jac
```diff
@@ -244,6 +270,8 @@ obj KubernetesTargetBase(DeploymentTarget) {
                 status = ResourceStatus.RUNNING;
             } elif ready_replicas > 0 {
                 status = ResourceStatus.PENDING;
+            } elif replicas == 0 and self._is_stopped_by_design(app_name) {
```
MusabMahmoodh, line 273: An HTTP-activated service parked at zero still reports FAILED here: `_is_stopped_by_design` looks only at `min` and triggers, and an `http_activation` app defaults to `min=1` with no triggers (the worker in #8779's rpc-wake fixture). target.jac already treats `ScalingMode.HTTP_ACTIVATION` as intentionally idle in the rollout wait. Stamping `_resolve_scaling`'s decision as a Deployment annotation (like `CONFIG_REVISION_ANNOTATION`) and reading it off the `deployment` you already fetched would give one answer.

jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac
```diff
@@ -918,6 +954,8 @@ obj ManifestBuilder {
         }
         if self.fleet is not None {
             merged.update(self.fleet.app_config(svc_name));
+        } elif svc_name != GATEWAY_NAME {
+            merged.update(get_scale_config().get_app_scale_config(svc_name));
```
MusabMahmoodh, line 957: `get_status` builds this builder with no fleet or project_dir, so this reads the `get_scale_config()` singleton, i.e. whatever jac.toml the calling process loaded. From the SDK's `resource_status` outside the app's project that's the wrong file and eligibility quietly comes back False; the e2e passes because it seeds the singleton first. With an annotation, this fallback, `autoscaler_config_for_deployment` and `_service_key_for_deployment` can all go.

jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac
```diff
@@ -12,6 +12,11 @@ import from jaclang.scale.models.deploy_context { DATABASE_URL_ENV, DeployContex
 import from jaclang.scale.deploy.target.kubernetes.kubernetes_config {
     KubernetesConfig
 }
+import from jaclang.scale.deploy.autoscale.autoscaler {
+    DEFAULT_CPU_UTILIZATION_TARGET,
```
MusabMahmoodh: `DEFAULT_CPU_UTILIZATION_TARGET` and `DEFAULT_MEMORY_UTILIZATION_TARGET` aren't used in this file; only `RESOURCE_ONLY_TRIGGER_TYPES` is.

MusabMahmoodh (approved):
Everything from Aug 28 and 31 is addressed: trigger types filtered, the cpu-only cases in the table, a malformed toml stays FAILED, and the solo-route synthesizers are gone with main's fleet refactor. The `spec.replicas == 0` guard and its scaling-back-up test are right. Releasing my block. One gap left as a follow-up rather than a blocker, since it was FAILED before this too: `get_status` re-derives eligibility from config, so it disagrees with the rollout wait on HTTP-activated services and depends on which jac.toml the calling process loaded. #8464's `inactive` answers the same question a third way, so it's worth settling which one is authoritative.

jac/jaclang/scale/deploy/target/kubernetes/kubernetes_target.jac
Developer-Linus: Fixed in 88fcfbcd3. Added ManifestBuilder.http_activation_config_for_deployment, mirroring autoscaler_config_for_deployment, and check it first in _is_stopped_by_design. The gateway is excluded the same way _resolve_scaling excludes it, since it must never scale to zero. Coverage folded into the existing get_status and autoscaler_config_for_deployment tests in test_factories.jac rather than added as new ones.

jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac
Developer-Linus: Agreed this is real, and the annotation approach you sketched (stamping _resolve_scaling's decision on the Deployment) would remove this fallback along with autoscaler_config_for_deployment and _service_key_for_deployment entirely. I looked at it, but it changes the contract test_scale_to_zero_status_e2e.jac relies on: that test applies a raw Deployment directly, bypassing generate_manifests, and expects get_status to resolve eligibility from real config. Switching to annotations would need that e2e rewritten too. Since this was already flagged as a non-blocking follow-up in your Sep 16 approval, and #8464's inactive state raises the same authority question, I'd rather settle both together than patch this one in isolation. Added it to the PR body under Known follow-ups so it stays visible.

jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac
Developer-Linus: Fixed, removed the unused DEFAULT_CPU_UTILIZATION_TARGET and DEFAULT_MEMORY_UTILIZATION_TARGET imports.

jac/jaclang/scale/deploy/target/kubernetes/kubernetes_target.jac
```diff
@@ -226,6 +230,28 @@ obj KubernetesTargetBase(DeploymentTarget) {
         }
     }
 
+    def _is_stopped_by_design(app_name: str) -> bool {
+        try {
+            builder = ManifestBuilder(k8s_config=self.k8s_config, logger=self.logger);
+            if builder.http_activation_config_for_deployment(app_name) is not None {
+                return True;
+            }
```
MusabMahmoodh: `http_activation.min_replicas` is a real key (`http_activation_config.jac:149`, default 0), so a service that sets a floor of 1 and sits at zero now reports STOPPED here when it is a real failure. Checking the floor keeps that case FAILED and still skips the metric check:
```suggestion
            http = builder.http_activation_config_for_deployment(app_name);
            if http is not None {
                return int(http.get("min_replicas", 0)) == 0;
            }
```

MusabMahmoodh (approved):
The http_activation check mirrors `_resolve_scaling` (HTTP first, gateway excluded), so `get_status` and the rollout wait now agree on HTTP-activated services. Fine taking the jac.toml-singleton half of my earlier note as the stated follow-up with #8464's `inactive`.

MusabMahmoodh (commented):
#9432 settled this with the `jac-scale.replica-floor` annotation and `assess_workload`, so the `get_status` STOPPED path, the settled-at-zero rollout wait and the http_activation floor commit are all covered on main, and `ResourceStatus.STOPPED`/`FAILED` no longer exist there. The branch now conflicts in four files. What main still lacks is the dry-run piece: `_check_hpa` rejects `hpa.min = 0` with "must be >= 1" even under KEDA with a real trigger, and nothing stops `hpa.min = 0` on cpu/memory-only triggers. Should we close this in favour of #9432 and move `is_scale_to_zero_eligible`, the `_check_hpa`/`Plan` change and `test_scale_to_zero_validation.jac` to a small PR off main? If `idle_replicas = 0` is meant to be the only documented route to zero now, even pointing the `_check_hpa` error at it would be worth having.

PR #8734, opened 2026-08-26
fix(scale): cut KEDA scale-to-zero wake latency by caching the venv and tightening the startup probe

### LINKED ISSUE #8695: jac-scale: KEDA scale-to-zero wake takes 44s, 21s of it re-running jac install per pod
## Summary

A KEDA scale-to-zero app on jac-scale takes **~44s** to answer its first request after idling. Measured twice on a live EKS cluster with near-identical results. Almost half of that (**21s**) is the `jac-bootstrap` init container running `jac install`, which rebuilds the app venv from scratch on every pod start because `/app` is an `emptyDir`. A further **~5s** is lost to `startupProbe` polling granularity: the app is verifiably serving at +35.7s but Kubernetes does not mark the pod Ready until +41s.

This is the same in-pod install cost measured in #7291, but the consequence is different and worse. At deploy time it is paid once by an operator watching a progress bar. Under scale-to-zero it is paid **on every wake, by an end user waiting on a page load**. Removing the per-pod install and tightening the probe would bring wake latency to roughly 19s without changing the architecture.

## Environment

| | |
|---|---|
| jac | 0.34.15 (pinned release) |
| Cluster | EKS, jaseci-cluster, x86_64 nodes |
| KEDA | core 2.x, HTTP Add-on 0.15.0 (`http.keda.sh/v1beta1`) |
| App | single service plus gateway, "Empty" template (17 python packages) |
| Scale target | `gateway-deployment`, `minReplicaCount: 0` via `InterceptorRoute` + `external-push` ScaledObject |
| Driver | jacBuilder PR #1514 (adds the scale-to-zero deploy option) |

Deploy any app with the gateway parked at zero behind the KEDA HTTP interceptor, let it idle until the gateway scales to 0, then send one request.

```bash
# 1. confirm the gateway is genuinely asleep (no prior HTTP touches)
kubectl -n jb-0987a6d2-keda-check get deploy gateway-deployment
# NAME                 READY   UP-TO-DATE   AVAILABLE
# gateway-deployment   0/0     0            0

# 2. one cold request
curl -s -o /dev/null \
  -w "http_code=%{http_code} time_total=%{time_total}s\n" \
  -m 240 https://keda-check.jaseci.app/
```

Verbatim result, two independent samples taken ~7 minutes apart, each after the gateway had scaled back to 0:

```
# sample 1
http_code=200 time_total=43.947530s

# sample 2
http_code=200 time_total=43.715419s time_starttransfer=43.715361s
```

## Actual

Phase breakdown from `kubectl get pod <p> -o json` (pod conditions, init and main container timestamps) plus the gateway container log. Sample 2, relative to pod creation:

| phase | at | duration | what happens |
|---|---|---|---|
| PodScheduled | +0.0s | | |
| PodReadyToStartContainers | +2.0s | 2s | |
| init `wait-for-mongodb` | +1.0s to +1.0s | 0s | deps already up |
| init `wait-for-redis` | +2.0s to +2.0s | 0s | deps already up |
| init `jac-pvc-bootstrap` | +3.0s to +5.0s | 2s | |
| **init `jac-bootstrap`** | **+7.0s to +28.0s** | **21s** | **`jac install`, rebuilds the venv** |
| Initialized | +29.0s | | |
| main container running | +29.0s | | |
| **app actually serving** | **+35.7s** | 6.7s | `Uvicorn running on http://0.0.0.0:8000` |
| **Ready / ContainersReady** | **+41.0s** | **5.3s** | **probe detection lag, app was already up** |
| response to client | ~+44s | ~3s | endpoint propagation, interceptor forward |

The two dominant, addressable costs:

**1. `jac install` runs on every pod start (21s, ~48% of the wake).**

`/app` is an `emptyDir`, so `/app/.jac/venv` is destroyed with the pod and rebuilt from scratch each time:

```
  /app   vol=jac-app   type=['emptyDir']
```

The init container ends in `jac install` (`runtime_bootstrap.build_bootstrap_init_container`). It is already doing the fast thing it can, installing offline from vendored wheels, and still costs 21s for a 17 package app:

```
[jac-scale] vendored wheels found; python deps install offline
Installing dependencies for empty
  Virtual environment: /app/.jac/venv
Installing 17 package(s)...
  ...
✔ All dependencies installed
```

Note this app is close to the floor. #7291 measured 163 to 227s per pod for a 5 service app, so the cost scales with dependency count.

**2. `startupProbe` granularity (~5.3s, ~12%).**

`manifest_builder._build_startup_probe` hardcodes `periodSeconds: 10`:

```jac
def _build_startup_probe -> dict[str, any] {
    return {
        "httpGet": {"path": "/healthz/live", "port": cfg.container_port},
        "periodSeconds": 10,
        "timeoutSeconds": 3,
        "failureThreshold": 90
    };
}
```

Measured effect: uvicorn logged `Uvicorn running` at `08:44:35.74`, the pod flipped Ready at `08:44:41`. The pod was refusing traffic for 5.3s after it could serve, purely because the next poll had not happened yet. The readiness probe compounds it with `periodSeconds: 20` (`readiness_period` default), which matters whenever startup and readiness do not land in the same window.

## Expected

Wake latency should be dominated by process start, not by rebuilding an environment that is identical on every wake. For a small app the request should be answered in roughly 15 to 20s.

## Impact

- **Scale-to-zero is much less attractive than it should be.** A 44s first paint is tolerable for an abandoned demo and unacceptable for anything a user might share. It pushes operators toward always-on, which is the cost problem scale-to-zero exists to solve.
- **The cost is paid per wake, not per deploy.** An app that idles between visitors pays it every time.
- **It scales with dependencies.** 21s here is a 17 package floor. #7291 measured 163 to 227s per pod on a real app, which would make scale-to-zero unusable.
- **It interacts with interceptor timeouts.** The KEDA HTTP Add-on holds the request while the pod starts. jacBuilder currently sets a 180s hold to stay clear of this; a shorter, more natural timeout would start returning errors.

## Workaround (verified)

None that removes the cost. Two things were verified as NOT workarounds:

- **Cold start placeholders make it worse.** Setting `coldStart.placeholder` on the `InterceptorRoute` causes the interceptor to answer the request itself, so nothing is queued, the external scaler never sees a pending request, and the ScaledObject stays `Active=False` permanently. Observed live: the app never woke at all. Removing the placeholder restored the 45s wake and a correct 200.
- **Keeping a warm replica** avoids the latency but defeats the feature.

## Suggested fix

Ordered by measured win:

1. **Stop rebuilding the venv per pod (saves ~21s).** The dependency closure is fully determined by the sealed bundle, which is already content addressed on the `jac-bundles` PVC (RWX `efs-sc`) and already mounted, since the pip cache is written to `/jac-bundles/.cache/pip`. Options, roughly in order of preference:
   - stage a prepared venv into the bundle at seal time and mount or copy it, so `jac install` becomes a no-op verification;
   - keep a per-app venv directory on the bundle PVC keyed by the dependency hash, populated on first boot and reused on every subsequent pod;
   - fall back to the current install path when the hash misses.
   Caveat worth measuring before committing: EFS is slow for many small files, so a venv read directly off `efs-sc` may trade install time for import time. A copy from PVC to `emptyDir` may beat both.
2. **Make the startup probe granular (saves ~4 to 5s).** `periodSeconds: 10` with `failureThreshold: 90` is a 15 minute budget expressed in coarse steps. `periodSeconds: 2` with a proportionally larger `failureThreshold` keeps the same ceiling and detects readiness ~5x sooner. Worth exposing as config rather than hardcoding, so scale-to-zero deployments can tighten it without affecting slow-booting apps.
3. **Reconsider `readiness_period: 20` for zero scaled services**, which adds up to 20s whenever it does not coincide with the startup transition.

Expected result: 43.7s minus ~21s minus ~4s is roughly **19s**, without touching the architecture.

**Regression test that would have caught it:** a scale-to-zero e2e that asserts a wake-path budget, for example scale the target to 0, issue one request, assert `time_total` under a threshold, and fail if the `jac-bootstrap` init container duration exceeds a few seconds. The current e2e coverage checks that a deploy succeeds, not how long a woken pod takes to answer.

## Related

- #7291 measures the same in-pod install cost at deploy time (163 to 227s per pod on a 5 service app). This issue is the scale-to-zero consequence, where the cost is paid per wake by an end user.
- #7138 cold deploy prep and first-boot compile.
- #7959 routing ingress and gateway traffic through the KEDA HTTP interceptor.
- #8694 deploy tooling has no concept of an intentionally zero-replica KEDA service.
- jacBuilder#1514 is the feature that surfaced this.

Developer-Linus:
Fixes #8695

## Summary

When a scale-to-zero pod wakes up on a live cluster, it was paying about 44
seconds before it could answer its first request. Most of that time was
waste: the pod reinstalled its Python dependencies from scratch every time,
even though nothing had changed, and Kubernetes checked "is it ready yet"
less often than it needed to on two different probes. This PR fixes all
three costs the issue identified.

## Problem

Three costs made up the wake latency, as measured in #8695 on a dedicated
EKS node. The figures below are that original measurement, carried over to
describe the problem being fixed -- they are not a saving re-measured from
this branch; see the Tests section for what has and has not been verified
here.

1. **`jac install` reran on every pod boot (about 21s, ~48% of the wake).**
   The pod's `/app` directory is an `emptyDir`, so it gets wiped every time
   the pod restarts. That means the Python virtual environment (venv) is
   rebuilt from scratch on every wake, even though the app's dependency list
   had not changed since the last time.
2. **The startup probe polled too coarsely (about 5.3s, ~12%).** Kubernetes
   was already checking every 10 seconds whether the app was answering
   requests, regardless of how small the app was. So even once the app was
   actually ready, it could take up to 10 more seconds before Kubernetes
   noticed and routed traffic to it.
3. **The readiness probe added the same kind of lag on top of the startup
   probe.** Readiness polling is disabled until the startup probe first
   succeeds. Once it does, Kubernetes waits out one more readiness period
   before the first readiness check registers. With the default 20s
   readiness period, that could add up to 20 more seconds after the app was
   already serving, on top of cost 2.

## Fix

**Venv caching (fixes cost 1):**

- `PythonProvider.install_specs` now writes a small marker file
  (`.jac-deps-hash`) inside the venv after a successful install, a hash of
  the sorted dependency list. On the next call, if the venv already exists
  and its marker matches the current dependency list, `pip install` is
  skipped entirely. This mirrors a pattern the codebase already uses for
  npm/vite dependencies (`client_deps_stale`).
- The pod's bootstrap script now restores a cached venv from the shared
  bundle PVC (the same persistent, content-addressed storage the deploy
  bundle already lives on) before running `jac install`, and saves it back
  to that cache afterward. This is what lets the venv survive the
  `emptyDir` wipe across pod restarts, so a subsequent wake finds a warm
  venv instead of an empty directory.
- The cache key is a hash of the base image plus the resolved dependency
  specs, so a Python image bump can never accidentally reuse a venv built
  for an incompatible interpreter.
- The cache write is atomic (build into a private staging path, then one
  `mv` into place) and skips writing if an entry for that key already
  exists, so a burst of concurrent cold starts can never observe a
  half-written cache.
- The cache is size-bounded at 2GB per app: crossing that cap evicts
  individual entries, oldest first, until back under it -- the entry
  currently being restored or published is never a candidate, and the
  cache root itself is never touched. It is shared cross-pod via the
  bundle-loader's `fsGroup`, not a world-writable `chmod 0777` (unlike the
  neighboring pip cache, which is a separate, unkeyed, purely-additive
  cache where that distinction matters less).

**Startup probe (fixes cost 2):**

- `KubernetesConfig.startup_probe_period` is a new, configurable field
  (default `2`, down from the hardcoded `10`), following the same pattern
  as the existing `readiness_period`/`liveness_period` fields.
- `failureThreshold` now scales with the period (`900 // period`, floored
  at 1) so the total time budget before the probe gives up stays the same
  (~900s) even though it now checks four to five times more often.
- Both values are clamped to a minimum of 1, so a misconfigured period
  (0, or a very large number) degrades to a valid probe instead of
  crashing manifest generation or producing a value Kubernetes would
  reject.

**Readiness probe (fixes cost 3):**

- `KubernetesConfig.readiness_period_scale_to_zero` is a new, configurable
  field (default `2`) that applies only to services with a replica floor
  of 0, meaning the ones actually configured to scale to zero.
- `_build_readiness_probe` now checks each service's own replica floor
  (reusing the same helper the pod-disruption-budget check already uses)
  and only tightens the period for those services. A normally-scaled
  service keeps the default 20s period untouched, because its readiness
  probe keeps polling for the pod's entire lifetime, not just at boot, so
  tightening it globally would multiply steady-state probe traffic
  tenfold for every deployment for no benefit.
- `failureThreshold` scales with the tightened period (`60 // period`,
  floored at 1) so the original 60-second tolerance for transient
  unhealthiness stays the same, it is only checked more often, not made
  stricter.

## Why this fix

- The venv persists on the same PVC the deploy bundle already uses, so
  there is no new infrastructure to provision. The fast-path check
  (marker-file comparison) is cheap and copies a pattern already proven
  elsewhere in this codebase, rather than inventing a new caching
  mechanism.
- Keying the cache on the base image (not just the specs) avoids a subtle
  correctness bug: compiled wheels in a venv are specific to the Python
  build they were installed against, so reusing a cache across image
  bumps could otherwise silently restore a broken venv.
- The startup probe fix keeps the same overall timeout budget. It only
  makes the polling finer-grained, so it does not weaken the safety net
  that gives slow-starting apps enough time to come up.
- The readiness probe fix only touches services with a replica floor of 0.
  A normal service's readiness probe runs for the pod's whole life, so
  scoping the tighter period to scale-to-zero services avoids trading wake
  latency for extra steady-state load on every other deployment.

## Changed files

| File | Change |
|---|---|
| `jac/jaclang/project/providers.jac` | Adds the venv staleness check (`_venv_deps_stale`) and marker-writing helper; `install_specs` skips `pip install` when the marker matches. |
| `jac/jaclang/scale/deploy/target/kubernetes/kubernetes_config.jac` | Adds the new `startup_probe_period` and `readiness_period_scale_to_zero` config fields (default `2` each). |
| `jac/jaclang/scale/config/impl/config_loader.impl.jac` | Adds `startup_probe_period` and `readiness_period_scale_to_zero` to the default scale config dict. |
| `jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac` | `_build_startup_probe` uses the configurable period and scales `failureThreshold` to preserve the total budget; `_build_readiness_probe` now tightens its period for services with a replica floor of 0, scaling `failureThreshold` to match; resolves the project's Python deps at manifest-build time so the bootstrap script can key its cache the same way `jac install` would. |
| `jac/jaclang/scale/deploy/target/kubernetes/runtime_bootstrap.jac` | Adds `venv_cache_key`; the generated bootstrap script restores a cached venv before `jac install` and persists it back afterward, atomically and size-bounded. |
| `jac/jaclang/scale/tests/deploy/test_memory_trigger_guard.jac` | New tests for `_build_startup_probe` (default period, custom period, clamping a non-positive period, the failureThreshold floor) and for `_build_readiness_probe` (default period for a normal service, tightened period and scaled failureThreshold for a replica-floor-0 service, the hpa-disabled case, a custom override). |
| `jac/jaclang/scale/tests/microservices/test_shared_dep_cache.jac` | New tests for `venv_cache_key` (deterministic, spec-order independent, changes with specs or base image) and the bootstrap script's restore/persist/atomicity/size-bound/cross-uid behavior. |
| `jac/jaclang/scale/tests/misc/test_config_loader.jac` | Extends the existing probe-timing defaults check to cover `startup_probe_period` and `readiness_period_scale_to_zero`. |
| `jac/tests/project/test_dependencies.jac` | New tests for `_venv_deps_stale` (matching marker, changed specs, missing marker, missing venv dir) and for `install_specs` actually skipping pip on a matching marker. |

## Tests

- All new tests were written first against the old behavior (TDD red
  state), then turned green by the fix, then covered a follow-up review
  pass that found three more gaps (image-scoped cache key, atomic persist,
  probe-value clamping), and a second review pass after that (probe
  defaults actually shipping the tighter value, cache size bound, cross-uid
  cache writes). See commit history on this branch for the full sequence.
- Verified manually against a live local cluster (kind + KEDA): built a
  self-contained `jac` binary, deployed a sample two-service app, and
  confirmed the deployed pod's `startupProbe` shows `periodSeconds: 2` /
  `failureThreshold: 450` (previously `10` / `90`). That is the probe half
  of this PR, confirmed working.
- The venv-cache half was not initially verified end to end: an unrelated
  bug (`JAC_VENV_CACHE_DIR`/`JAC_VENV_DIR` merging into one corrupted
  shell-text value, see the latest commit) made `restore`/`publish`
  silently no-op on every boot, with nothing in the init container log to
  say so -- confirmed via `kubectl get pod -o json` against a real
  deployed manifest, not inferred. That bug is now fixed (the four
  relevant values are set through the container's structured `env` list
  instead of shell text) and confirmed correct at the manifest level: the
  pod spec now carries `JAC_VENV_CACHE_DIR`/`JAC_VENV_DIR`/
  `JAC_VENV_CACHE_IMAGE_ID`/`JAC_VENV_CACHE_ROOT` as four genuinely
  separate env entries.
- **Outstanding before merge:** an actual cold-vs-warm wake-latency
  measurement on a real cluster with that fix in place. Nothing in this PR
  claims a specific number measured from this branch; #8695's 44s/21s/
  5.3s figures above remain the pre-fix baseline until a fresh run
  provides real ones.


MusabMahmoodh (changes requested):
Late on this one - it went up two days ago and I am only reaching it now.

The measurement work behind #8695 is good and the three costs are correctly
apportioned. Tightening the two probe periods is straightforwardly right and I
have nothing on it. The venv cache is where my comments are, and they are all
about the same thing: the publish step was written carefully for concurrency and
the eviction step was not, on a volume that several pods share.

## 1. The 2 GB eviction is an unsynchronized `rm -rf` of the whole cache root

```sh
if [ "$(du -sm $venv_cache_root)" -gt 2048 ]; then
  rm -rf $venv_cache_root && mkdir -p $venv_cache_root && chmod 0777 ...
fi
```

`BUNDLE_MOUNT_PATH` is the shared bundle mount, so every service's bootstrap in
the fleet runs this against the same directory, and a scale-to-zero wake is
precisely the event that makes several of them run at once.

Concrete interleaving: pod B is in its restore step,

```sh
cp -a "$venv_cache_dir" "$venv_dir" 2>/dev/null || true
```

when pod A crosses 2048 MB and starts `rm -rf` on the root. Files disappear
under B's copy. `2>/dev/null || true` swallows every error, so B does not fall
back to a clean install - it continues to `jac install` on top of a
**partially populated** `$venv_dir`. A half-copied venv is worse than no venv:
pip sees some distributions already present and skips them, and the pod comes up
with a subtly incomplete environment that fails at import time rather than at
install time. The failure surfaces minutes later, in the app, with no line in
the bootstrap log pointing here.

Compare with the publish step in the same block, which you did write for this -
staging directory, `mv -T` for an atomic rename, guarded by `[ ! -e ]`. The
eviction deserves the same care. Cheapest correct version: build the replacement
root beside the old one and `mv -T` it in, or take a lock (`mkdir` on a
well-known path is the portable atomic primitive in sh) around evict-and-restore.
Failing that, at minimum drop the `|| true` on the restore `cp` and `rm -rf
"$venv_dir"` when it fails, so a torn copy degrades to a full install rather
than a corrupt one.

## 2. `$$` is not unique across pods sharing this volume

```sh
staging_dir="{venv_cache_dir}.tmp.$$"
...
rm -rf "$staging_dir" 2>/dev/null
cp -a "$venv_dir" "$staging_dir" && mv -T "$staging_dir" "$venv_cache_dir"
```

`$$` is the shell's PID, and each container has its own PID namespace starting
at 1. Two pods waking together will land on low, frequently identical PIDs, so
`{cache_dir}.tmp.7` is a name both of them can pick. The first statement in the
sequence is `rm -rf "$staging_dir"` - so pod A can delete pod B's in-flight
staging copy, and B's `mv -T` then either fails or moves a truncated tree.

`mktemp -d "{venv_cache_root}/.tmp.XXXXXX"` gives you a name that is unique by
construction on the shared filesystem, which is what `$$` is being asked to do
here and cannot.

## 3. `chmod 0777` on a shared cache of executable content

The cache root is made world-writable so that any pod can publish into it. What
gets published is a venv - executable Python that every other pod in the fleet
subsequently `cp -a`s into its own `/app/.jac/venv` and runs. So after this PR,
any service in the fleet can write code that every other service will execute on
its next cold start, and (per #1) any service can delete the whole cache.

I do not think that is fatal inside a single fleet, where the services already
share a bundle and a trust boundary. Two questions before I would be comfortable
though: is `BUNDLE_MOUNT_PATH` ever backed by a PVC shared across *apps* rather
than across the services of one app - and if a service runs with a
non-jac-scale-controlled image (`service_dependencies`, custom base image), is it
still inside that boundary? If either is a yes, this is a cross-tenant code path
introduced by a latency optimization and should be scoped down to a per-app
subdirectory with normal permissions.

## 4. The cache key does not cover the interpreter

```jac
def venv_cache_key(specs: list[str], base_image: str) -> str {
    return sha256(base_image + "\n" + "\n".join(sorted(specs))).hexdigest();
}
```

Sorting the specs is right. But `base_image` is a tag string, and jac-scale
deploys mutable tags with `pullPolicy: Always` - the same tag can name a
different image, with a different Python, tomorrow. When it does, the key is
unchanged and every pod restores a venv built against the previous interpreter.
That is a wrong-ABI venv restored silently and indefinitely, and nothing in the
key can ever invalidate it.

If the image digest is reachable at manifest-build time, hashing that instead of
the tag fixes it exactly. If it is not, `python -c 'import sys; print(sys.
implementation.cache_tag)'` inside the init container and including it in the
directory name gets you the same guarantee for one subprocess.

## 5. `jac/jaclang/project/providers.jac`

+28 lines in the project layer for a jac-scale latency fix. What is the piece
that could not be done behind the existing seam, and is there a hookspec shape
for it instead? Happy to be told it genuinely belongs there - I just want the
reason recorded, since core changes riding along with a scale change are the
ones that get inherited by everyone.

## Checked and not flagged

- `venv_cache_key` is only called when `resolved_specs` is non-empty, so the
  no-dependency case adds no shell at all.
- The restore step creates `{APP_MOUNT_PATH}/.jac` before copying, so a first
  boot with a warm cache does not fail on a missing parent.
- Probe period changes are bounded by the existing config schema and
  `test_config_loader.jac` covers the new keys.

None of 1-4 is hard to fix and the win the PR measures is real - I would like
this in once the shared-volume paths are as careful as the publish path already
is.

jac/jaclang/scale/tests/microservices/test_shared_dep_cache.jac
```diff
@@ -39,3 +40,136 @@ test "shared pip cache is opened cross-uid, probed, with pod-local fallback" {
...
+    # _venv_deps_stale would wrongly call it fresh and skip reinstalling
+    # a genuinely broken venv -- worse than no caching at all.
+    specs = ["requests>=2.28.0"];
+    base_image = "python:3.12-slim";
+    ctr = build_bootstrap_init_container(
+        KubernetesConfig(), base_image, resolved_specs=specs
+    );
+    script = str(ctr["args"][0]);
+    cache_dir = f"{BUNDLE_MOUNT_PATH}/.cache/venv/{venv_cache_key(specs, base_image)}";
+    assert f'rm -rf "{cache_dir}"' not in script;
```
MusabMahmoodh: This assertion is already false in spirit on this branch: the eviction step runs `rm -rf` on `venv_cache_root`, and `cache_dir` lives inside that root, so the live cache path does get deleted in place - the test stays green only because the substring is spelled differently. That's the ceiling of string-matching the script: it pins the spelling, not the property it names.

jac/jaclang/scale/tests/microservices/test_shared_dep_cache.jac
```diff
@@ -39,3 +40,136 @@ test "shared pip cache is opened cross-uid, probed, with pod-local fallback" {
...
+
+
+# ===== venv persistence across pod restarts =====
+# /app is an emptyDir, wiped every pod boot, so jac install rebuilds the
+# venv from scratch on every scale-to-zero wake even though the dependency
+# set is unchanged. These only cover the bootstrap script's mechanical copy
+# in/copy out -- the actual install/skip *decision* is _venv_deps_stale's
+# job (test_dependencies.jac), not re-verified here.
+
+test "bootstrap restores a cached venv from the bundle PVC before jac install" {
```
MusabMahmoodh: These six all read the script as a string; it can just be run. Substitute the two mount-path constants for temp dirs, put a stub `jac` on PATH that drops a sentinel file into the venv, run the script twice with the app dir wiped in between (the emptyDir), and assert the second run comes up with the venv restored. Run two copies concurrently against the same cache root and the `$$` collision and the eviction race from my earlier review surface as real failures. One test like that replaces all six and can actually fail on the bugs this cache can have.

jac/jaclang/scale/tests/deploy/test_memory_trigger_guard.jac
```diff
@@ -198,3 +198,120 @@ test "deploy loads the cluster config before it resolves the pod arch" {
...
     assert order[:2] == ["config", "arch"];
 }
+
+
+# ===== startup probe =====
+# _build_startup_probe hardcoded periodSeconds: 10 while its siblings
+# (_build_readiness_probe / _build_liveness_probe) already read from
+# KubernetesConfig -- these follow the field the same way.
+
+test "_build_startup_probe defaults to the KubernetesConfig period" {
```
MusabMahmoodh, line 213: Nine tests for two probe fields, every one calling the private builder directly - and the piece none of them cover is the wiring: `_build_pod_spec` passing the right `svc_name` into `_build_readiness_probe`. One test through the full manifest build, asserting the rendered Deployment's probe fields for a min=0 service next to a min=1 sibling, exercises the real path and folds most of these in; keep the zero-period clamp case if you want the arithmetic pinned.

MusabMahmoodh (commented):
Following up on this morning's review with a separate pass on the tests - those findings stand. The theme across all three test files is the same: many small unit tests pinned to private functions and to the text of the generated script, and nothing that executes the behavior being shipped. The inline comments have concrete replacements - a handful of realer tests would cover more than these nineteen do.

Two notes not worth their own thread. The four `_venv_deps_stale` tests in test_dependencies.jac pin a three-line hash compare that the two `install_specs` tests already prove; fine to drop. And on mechanism vs workaround: the fix is the right mechanism, not a bandage - the emptyDir wipe and the probe polling are the real costs and both are addressed at the source. What's fragile is the medium: the cache lifecycle lives in f-string sh chained with `&&`, which is both why the eviction races were writable and why the tests can only grep the string. If that block became a small script asset (or a jac subcommand the init container calls), the execution test below is trivial to write.

jac/jaclang/scale/tests/microservices/test_shared_dep_cache.jac
Developer-Linus: Agreed. The eviction and staging logic no longer lives in shell, so there is nothing left to string-match here. Restore, publish, and eviction are now real functions (venv_cache.jac), tested directly against a real temp cache root, including a concurrency test that races real threads and would catch the unsynchronized rm -rf and the staging collision (test_venv_cache.jac).

jac/jaclang/scale/tests/microservices/test_shared_dep_cache.jac
Developer-Linus: Agreed, and this is the shape I went with. The venv-cache logic these six tested moved into real functions called from a jac scale subcommand, with tests that actually run restore/publish/evict against a temp directory instead of reading generated script text, plus a concurrency test racing real threads on the same cache root. What is left in this file is a thin ordering check (restore before install, publish after, neither for a mutable tag) and the venv_cache_key tests, which already called real code.

jac/jaclang/scale/tests/deploy/test_memory_trigger_guard.jac
Developer-Linus: Good point. Added an integration test through _build_service_pod_spec, the real path generate_manifests calls, asserting the rendered readinessProbe for a min=0 service next to a min=1 sibling in the same fleet. Kept the arithmetic edge cases (zero-period clamp, failureThreshold floor) as you suggested, and folded the one test that only checked the default-period case in isolation into the new integration test.

MusabMahmoodh (commented):
Re-reviewed the delta since my Aug 28 comment. The venv cache is a good piece
of work and the docs section explaining *why* the venv dies on every wake
(`/app` is an `emptyDir`) is the part that will save the next reader. One
finding, latent rather than live, and one note.

### The cache key and the in-venv marker hash different things

- `venv_cache_key(specs, base_image)` includes the base image, exactly as the
  docs claim: "a `python_image` change can never restore a venv built for an
  incompatible interpreter."
- `_venv_deps_hash(specs)` - written into the venv as `.jac-deps-hash` and
  read by `_venv_deps_stale` to decide whether to skip `jac install` entirely -
  does **not** include the image.

Today that is safe, and it is worth writing down why, because the safety is
not local to either function. It holds only because the venv lives on an
`emptyDir` that is wiped on every restart, so the *only* ways a venv with a
marker can exist are (a) a fresh install this boot, or (b) a restore from a
cache entry that was keyed on the image. Change any one of those - move the
venv onto the bundle PVC, make `/app` survive a restart, add a second
population path - and `_venv_deps_stale` starts answering "not stale" for a
venv built against a different interpreter, and `jac install` is skipped.
The symptom would be an app that boots against the wrong Python with no
diagnostic, on the wake path, which is about the worst place to debug it.

Cheapest fix is to make the marker hash the same tuple the cache key does.
It costs nothing and removes the coupling. If you would rather not, a comment
on `_venv_deps_stale` naming the emptyDir assumption would at least make the
invariant discoverable.

### `is_cache_eligible(base_image)` gating

I traced this and it does the right thing - an ineligible image never
restores, so it never carries a marker in from elsewhere. Not a finding,
noting it so you know it was checked rather than skipped.

### The probe changes

`readiness_period_scale_to_zero` only applying to services with a replica
floor of 0 is the right scoping, and keeping the ~900s first-boot budget fixed
while tightening the period is the detail I would have asked for. The docs
table states both clearly. No objection.

`test(scale): replace venv-cache script string-matching with real e2e` is the
commit I am happiest to see - asserting on generated script text would have
passed while the cache silently never restored.

MusabMahmoodh (approved):
Re-reviewed at `875c0bbe4`. All five findings closed, and the two I cared most
about are closed at the mechanism rather than at the symptom. Approving.

**Eviction (finding 1).** `_evict_if_needed` is the right shape now: individual
entries, oldest first by mtime, root never touched, `keep_name` and staging
prefixes excluded, and the size re-checked each iteration so it stops as soon as
it is under rather than draining the cache. Moving it out of shell into
`venv_cache.jac` is most of why it is reviewable at all.

**The rollback you found on self-review** is the better catch of the two, and it
is worth being explicit about why, because it is now load-bearing for a case the
eviction still permits.

Eviction is per-pod and unsynchronized across pods, and it has to be, short of a
lock on a shared mount. `keep_name` protects the evicting pod's own entry, not
anyone else's. So this interleaving is still reachable: pod B is restoring from
entry X, pod A crosses the cap, X is not A's keep, A removes it mid-copy.

What makes that safe is your rollback, not the eviction logic:

```jac
} except Exception {
    shutil.rmtree(venv_dir, ignore_errors=True);
}
```

B's `copytree` raises, the partial venv is deleted, B falls through to a full
`jac install`. Slow boot, correct environment. That is exactly the degradation I
asked for in the original finding, and it is the property that makes the
unsynchronized eviction acceptable rather than a race worth locking.

Suggest a one-line comment on that `except` saying so, in the "why" register
rather than the "what": a future reader optimizing the rollback away, or
narrowing it to a specific exception type, would silently reintroduce the torn
venv and would have no way to know from the code that a concurrent evictor is
the reason it is there. That is the only thing I would still change, and it does
not need another round.

**`image_identity` (finding 4 follow-up).** Taking the code fix over the comment
was the right call. Threading it through `_venv_deps_hash` and defaulting it
empty means the non-scale path produces a byte-identical hash to before, so
there is no invalidation cost imposed on every `jac install` to fix a k8s-only
coupling. The two regression tests are the right two, and the second one, seeding
a marker under one identity and confirming a different identity is stale, is the
one that actually pins the bug.

The digest-pinning gap you found in the experimental channel is a real bug that
was not in my review and had nothing to do with the cache. Good find.

**On `providers.jac` (finding 5).** Accepted, and I withdraw it. You are right
that there is no hook system left to route through and that a direct addition
matches the existing pattern. My review was written against an architecture that
has not existed for some time, and the skip-pip-when-nothing-changed improvement
stands on its own independent of the cache.

Nothing blocking. This went from five findings to zero in two pushes with a
self-caught sixth in between, which is the fastest turnaround I have reviewed on
this surface.

jac/jaclang/scale/deploy/target/kubernetes/venv_cache.jac
```diff
@@ -0,0 +1,110 @@
...
+}
+
+
+def is_cache_eligible(base_image: str) -> bool {
+    return "@sha256:" in base_image;
+}
+
+
+def restore(cache_dir: str, venv_dir: str) -> None {
+    if not os.path.isdir(cache_dir) {
```
MusabMahmoodh, line 28: This is the guard that made the env bug invisible for four days: an unset `JAC_VENV_CACHE_DIR` and a real cache miss both return here with nothing in the init container log, `publish` swallows every exception with `_ = 0` below, and `_resolved_python_specs` in manifest_builder returns `None` on any exception with no line in the deploy output. Silent fallback at a decision point, the #7900 theme-1 class. One line per outcome is the whole fix, and it turns your outstanding cluster validation into a log read instead of a stopwatch:

```suggestion
    if not cache_dir or not venv_dir {
        print("[jac-scale] venv-cache: JAC_VENV_CACHE_DIR/JAC_VENV_DIR unset; skipping restore", file=sys.stderr);
        return;
    }
    if not os.path.isdir(cache_dir) {
        print(f"[jac-scale] venv-cache: no entry at {cache_dir}; full install", file=sys.stderr);
        return;
    }
```

jac/jaclang/scale/tests/microservices/test_shared_dep_cache.jac
```diff
@@ -39,3 +43,153 @@ test "shared pip cache is opened cross-uid, probed, with pod-local fallback" {
...
+    # line is `if not os.path.isdir(venv_dir): return`, so an empty
+    # JAC_VENV_DIR made every restore call a silent no-op - same for
+    # publish() against an empty JAC_VENV_CACHE_DIR. The tests above only
+    # ever checked substring presence and ordering in the generated shell
+    # text, which stayed true either way and caught neither bug.
+    #
+    # The fix moves these off shell text entirely: they're static once
+    # resolved, so they're set via the container's own `env` list (the
+    # same mechanism the sibling service container already uses), which
+    # Kubernetes applies directly with no shell parsing step to corrupt.
```
MusabMahmoodh, line 142: One question on the mechanism this comment names. A POSIX shell does not merge `A="x" B="y" cmd`; each assignment word is its own variable in `cmd`'s environment. You saw the merged value in `kubectl get pod -o json`, which is the manifest as emitted on the host, before any shell ran. That is the shape of the jac f-string defect where two quoted interpolations in one f-string compile to a single quoted blob (it broke #7198's PVC seed the same way). Structured `env` is the right fix either way, but if that is what happened the comment should say so, because the next generated script with two quoted `{}` in one f-string hits it again and this comment sends them to bash.

jac/jaclang/scale/tests/microservices/test_shared_dep_cache.jac
```diff
@@ -39,3 +43,153 @@ test "shared pip cache is opened cross-uid, probed, with pod-local fallback" {
...
+    pinned_image = "jaseci/jaclang@sha256:" + "b" * 64;
+    ctr = build_bootstrap_init_container(
+        KubernetesConfig(), pinned_image, resolved_specs=specs
+    );
+    script = str(ctr["args"][0]);
+    assert "jac scale venv-cache-restore" in script;
+    assert "jac scale venv-cache-publish" in script;
+    assert script.index("venv-cache-restore") < script.index("jac install");
+    assert script.index("venv-cache-publish") > script.index("jac install");
+    assert "chmod 0777" not in script.split("jac install")[1];
```
MusabMahmoodh: This assertion guards against a venv-path chmod that the previous revision already removed from the shell text entirely, so it cannot fail any more.

jac/jaclang/scale/tests/deploy/test_venv_cache.jac
```diff
@@ -0,0 +1,237 @@
...
+        _make_venv(cache_dir, "cached");
+        venv_dir = os.path.join(td, "app", "venv");
+        def _torn_copy(src: str, dst: str, **kwargs: any) {
+            os.makedirs(os.path.join(dst, "lib"), exist_ok=True);
+            with open(os.path.join(dst, "lib", "partial.txt"), "w") as f {
+                f.write("torn");
+            }
+            raise OSError("simulated concurrent deletion mid-copy");
+        }
+        with unittest.mock.patch.object(
```
MusabMahmoodh: The filesystem produces this torn copy for real: a dangling symlink inside `cache_dir` makes `shutil.copytree` (defaults, `symlinks=False`) copy every other entry and then raise `shutil.Error`, an `OSError`, which is exactly the partial venv plus exception you are simulating. That tests the real `copytree` call in `restore` instead of a stand-in, and keeps this file mock-free:

```suggestion
        _make_venv(cache_dir, "cached");
        os.symlink(os.path.join(td, "does-not-exist"), os.path.join(cache_dir, "lib", "dangling"));
        restore(cache_dir, venv_dir);
        assert not os.path.exists(venv_dir);
```

MusabMahmoodh (commented):
Re-read after your Sep 2 fix. The env-list change is right and complete: the four variables are set in one place each and read in one place each, the mutable-tag pod is byte-identical to main, and the new test does fail on the pre-fix tree by reading (no `env` key existed, so the first membership assert trips). My earlier approval stands on the wiring.

What I am not ready to re-approve on yet is the claim itself. The body's Tests section describes an init container log switching from "venv stale or missing, running pip" to "venv up to date, skipping pip"; neither string exists in the diff, and your comment says restore and publish no-oped on every boot of the code I approved, so the only body number verified at head is the probe values. The 21s / 48% figure is the issue's cost measurement, not a saving measured from this branch. The fresh-cluster cold/warm numbers you named are the thing to land before merge, and the body should describe the code that is here (the docs sentence "wiped and rewarmed past 2GB" has the same drift: `_evict_if_needed` drops the oldest entries one at a time and never touches the root).

Four inline: the silent-fallback class that hid this (the one I would fix first), a question on the mechanism the test comment names, a dead assertion, and a mock replacement. One more question I could not settle by reading: `initialDelaySeconds` on the readiness probe is still `cfg.readiness_initial_delay` (default 10) inside the tightened branch, and kubelet measures that from container start regardless of the startup probe. If the app answers before 10s on a warm wake, that delay is the floor and the 2s period buys nothing under it. Was the app answering later than 10s in your measurement? If not, the min=0 override belongs on that field too.

Aside on #8838: its last commit (`0236c37608`) fixes a real bug where `_build_pdb_manifest` computes `min_available` and then hardcodes `maxUnavailable` in the returned manifest. That fix is in none of the six stacked PRs and deserves its own PR against main rather than living in the validation stack.

MusabMahmoodh (commented):
The 13:07 push closes both gaps I raised yesterday. The partial-copy test now drives the real `shutil.copytree` through a dangling symlink instead of patching it out, which is a genuine partial failure rather than a stand-in for one, and it is a better test than the mocked version was. The `restore` warnings turn the two silent skips into something a boot log can be read for. Eviction being per-entry oldest-first with the in-flight entry excluded is the right shape.

Three things, and the first one is why cold and warm keep matching.

**1. `is_cache_eligible` cannot return True for a deployed image.**

`venv_cache.jac` gates the whole restore/publish pair on `"@sha256:" in base_image`. Deploys resolve a mutable tag, so `cache_eligible` stays False, no `JAC_VENV_CACHE_DIR` is appended, and neither `venv-cache-restore` nor `venv-cache-publish` is ever added to the bootstrap commands. The cache path has not run in any measurement taken on this branch. Udith hit the same wall from the other side on jacBuilder#1642 yesterday while trying to explain why the wake numbers would not separate.

That is worth settling before more measurement work goes into #1642 or #8946, because right now the only part of this PR affecting the wake is the marker fast path and the probe changes, and the cache is being timed without executing.

**2. This PR and #8944 are adding the same mechanism to the same file.**

Both declare `glob VENV_DEPS_MARKER: str = ".jac-deps-hash"` in `jaclang/project/providers.jac`, both add a `_venv_deps_hash` over sorted specs, and both insert an early return into `install_specs` at the same point, above `ensure_venv`. Whichever lands second will not merge, and if the conflict is resolved mechanically we end up with two markers writing one filename.

Worth saying plainly: your hash is the correct one of the two. Folding `image_identity` in is what makes the marker safe to trust, and #8944's does not have it. Its sealed venv also drops `pyvenv.cfg`, so on the pod `get_venv_site_packages` falls through to globbing `lib/python*/site-packages` (`pyvenv.jac:298-317`) and hands whatever it finds to whatever interpreter is running.

The combination that looks right to me is your marker plus their seal-time build: the venv arrives inside the `.jab` already assembled, and the marker keyed on specs plus image identity is what decides whether to trust it. That leaves `venv_cache.jac`, `venv-cache-restore` and `venv-cache-publish` with nothing to do, since there is no boot-time install left to cache, and `restore`'s `copytree` is the per-file access pattern EFS punishes anyway. I have said the same on #8944; you two should pick, not the merge order.

**3. Your red is not yours.**

`test-compiler (equivalence)`, `test-runtime`, `test-scale (server)`, `test-scale (deploy)` and `test-jac-pack-smoke` are all failing on `AttributeError: 'AnalysisLedger' object has no attribute 'get'`, which reproduced on unmodified main at `50d359d877`. That was fixed on main by #8918 at 13:33Z yesterday, about 25 minutes after your last push. A rebase should take all five green without touching your diff.

Issue #8702, opened 2026-08-25
shadcn registry: sync with upstream — 9 missing components + style CSS refresh


ahzan-dev:
## Summary

Our bundled shadcn registry (`jac/jaclang/cli/shadcn/registry/`) has drifted from upstream. We ship **54** components; upstream `shadcn-ui/ui` now publishes **63**. The nine we are missing are mostly the new AI-chat primitives (`bubble`, `message`, `attachment`, `questionnaire`, …), plus `form` and `toast`.

The style CSS is also behind: every `styles/style-*.css` we ship is missing ~95 `cn-*` token blocks, and upstream now ships **8** styles where we ship **5**.

This issue covers: where the upstream sources are now, how the registry is wired, the exact per-component gap, and a repeatable add-one-component procedure with verification.

---

## 1. Where to look for new components

The old flat registry API is **dead** — `https://ui.shadcn.com/r/<name>.json` returns the HTML 404 page. `https://ui.shadcn.com/r/index.json` still returns 200 but carries only names and paths, **no `content` field**, so it is a listing, not a source.

Sources come from the GitHub repo, which reorganised in v4 into per-**base** trees:

```
github.com/shadcn-ui/ui → apps/v4/registry/
   bases/radix/ui/*.tsx    61 components — Radix primitives (`radix-ui` pkg)  ← our base
   bases/base/ui/*.tsx     62 components — Base UI
   bases/aria/ui/*.tsx     59 components — React Aria
   new-york-v4/ui/*.tsx    legacy flat tree, still present
   styles/style-*.css      the cn-* token definitions
```

Our port targets **`bases/radix`** — our components import the unified `radix-ui` package and carry `cn-*` tokens, which matches that tree.

Authoritative per-component metadata (npm deps + peer components) lives in `apps/v4/registry/bases/radix/ui/_registry.ts`.

Raw fetch for a single file:

```bash
curl -sO https://raw.githubusercontent.com/shadcn-ui/ui/main/apps/v4/registry/bases/radix/ui/<name>.tsx
```

Or clone once and work locally:

```bash
git clone --depth=1 https://github.com/shadcn-ui/ui
```

To recompute the gap at any time:

```bash
# ours
ls jac/jaclang/cli/shadcn/registry/components/ui | sed 's/\.jac$//;s/_/-/g' | sort > /tmp/ours
# upstream
[...]

PR #8751, opened 2026-08-28
feat(cli): sync jac-shadcn registry with upstream shadcn/ui v4

RavimalRanathunga:
The bundled jac-shadcn registry had drifted from upstream `shadcn-ui/ui`. This
brings the `bases/radix` component tree and the style-token presets back in
step, and stops short of the cases that need new dependencies or a design call
(tracked on #8702 as follow-ups).

## What changes

- **Five new primitives**, ported from `apps/v4/registry/bases/radix/ui/`:

  | Component | Exports | Shape | Deps added |
  |---|---|---|---|
  | `message` | `MessageGroup`, `Message`, `MessageAvatar`, `MessageContent`, `MessageHeader`, `MessageFooter` | plain `cn()` divs | none |
  | `direction` | `DirectionProvider`, `useDirection` | Radix `Direction` wrapper | `radix-ui` (already pinned) |
  | `marker` | `Marker`, `MarkerIcon`, `MarkerContent`, `markerVariants` | `cva` + `Slot` + `asChild` | `class-variance-authority`, `radix-ui` (already pinned) |
  | `bubble` | `BubbleGroup`, `Bubble`, `BubbleContent`, `BubbleReactions` | `cva` x2 + `Slot` | `class-variance-authority`, `radix-ui` (already pinned) |
  | `attachment` | `Attachment` + 8 slot components | `cva` x2 + `Slot` + `Button` peer | `class-variance-authority`, `radix-ui` (already pinned) |

  No new `depVersions` entries: every dependency these need was already pinned,
  so the `test_shadcn_registry` invariants hold with no manifest change beyond
  the five `components` blocks. `attachment` declares `button` as a peer.

- **Style CSS refreshed.** `style-{lyra,maia,mira,nova,vega}.css` are replaced
  wholesale with the current upstream. With the hardened parser (below), all
  eight style files now parse to a token map that is **byte-identical to
  upstream** (`nova` 422/422 rules, `maia` 421/421, and so on). All new
  `cn-bubble-*`, `cn-message-*`, `cn-marker-*` and `cn-attachment-*` blocks land
  here, since the new components are token-only.

- **Three new style presets:** `luma`, `rhea`, `sera`. Styles resolve purely
  by `style-<name>.css` filename with no whitelist anywhere, so this is a pure
  file drop. `jac retheme --style luma` works with no code change. The five
  places that enumerate style names for humans and agents (`jac-shadcn-components.md`,
  `npm-and-libraries.md`, `jac-client.md`, the `--style` help string) now list
  all eight.

- **`parse_style_css` hardened.** It now strips CSS comments before matching and
  captures each `@apply` value up to its own closing brace, so a stray comment
  between `{` and `@apply` or a missing trailing `;` can no longer silently drop
  a rule or splice one rule's text into the next. Two upstream style files hit
  exactly these cases (`style-maia.css` comment placement, `style-lyra.css`
  missing semicolon); both are also fixed at the source.

- **`drawer` and `sheet` re-ported.** Upstream moved the direction/positioning
  utilities out of `cn-drawer-content` and the enter/exit slide+fade utilities
  out of `cn-sheet-content` and into the component `className`. Their `.jac`
  files now carry those utilities inline, byte-identical to the current
  upstream `.tsx`, so both keep their layout and animation under every style.

- **Docs.** `jac-shadcn-components.md` gains five selection-table rows, the
  `attachment -> button` peer row, three extended-export rows, and the new
  style names.

## Existing components render as upstream's current components

With the hardened parser, every `cn-*` token in all eight styles resolves to a
value byte-identical to upstream's. `drawer` and `sheet` were the only two
components upstream had changed structurally (token -> `className`); they are
re-ported here. The remaining older ports are unchanged: their tokens now match
upstream exactly, so they render exactly as upstream's current components do.
Verified end to end against a scaffolded project (install, type check, Vite
build, server render) with no unresolved `cn-*` in any style.

## Class strings are byte-identical to upstream

The `.jac` sources mirror each `.tsx` `className` verbatim, including the bare
`cn-*` tokens: `resolve_component` inlines a token's utilities at install time
from the active style preset, so pre-expanding a token would double-apply it.
Two deliberate deviations:

1. Upstream's `AttachmentTrigger` passes `type={undefined}` in the `asChild`
   branch; Jac types the `type` prop as `str`, so that branch omits the
   attribute and the `<button>` branch sets `type={props["type"] or "button"}`.
   Behaviourally identical.
2. Upstream recently added a `cn-font-heading` marker class to every `*Title`
   component. Jac's component and theme layers have not adopted that token, so
   it is not added here (a single follow-up should add it to the theme layer
   and every title component at once).

## Deliberately out of scope (follow-ups on #8702)

| Excluded | Why |
|---|---|
| `message-scroller`, `questionnaire` | need the new `@shadcn/react` npm package (prebuilt JS, `react >=19` peer, unproven against the Jac compiler) and import a docs-app-internal `IconPlaceholder` |
| `form` | legacy `new-york-v4/` tree only; pulls `react-hook-form` + `@hookform/resolvers` |
| `toast` | `bases/base/` (Base UI) only; taking a non-Radix primitive is a design decision, and `sonner` already ships |
| `cn-font-heading` adoption | cross-cutting: a new theme-layer token plus a one-line change to every title component |
| a `scripts/` registry-diff helper | would make the next sync faster; not blocking |

## Testing

- `jac fmt --check --lintfix` on every changed `.jac` file: clean.
- `jac test tests/cli/test_shadcn.jac`: **40 passed**. New / tightened cases:
  - every declared `.cn-*` rule in every style is returned by `parse_style_css`
    (catches the comment / missing-semicolon drops the old `len > 0` check missed)
  - `drawer` and `sheet` keep their edge positioning / open animation after
    resolution, in all eight styles
  - `bubble` resolves with zero `cn-*` left over in all eight styles (was `nova` only)
  - `attachment` pins the resolved version ranges (`radix-ui = "^1.6.7"`), not
    just the bare package names
- End to end in a scaffolded project: `jac install --shadcn message marker
  direction bubble attachment` writes the five files (underscored) plus the
  auto-resolved `button` peer, patches `jac.toml [dependencies.npm]`, and leaves
  no unresolved `cn-*` token. `jac retheme --style luma` / `--style mira`
  re-resolve the installed files cleanly.

Refs #8702. The remaining items in that issue (`message-scroller`,
`questionnaire`, `form`, `toast`, `cn-font-heading`, sync tooling) stay tracked
there.


jac/jaclang/cli/shadcn/registry/styles/style-nova.css
```diff
@@ -474,19 +490,31 @@
 
   /* MARK: Drawer */
   .cn-drawer-overlay {
-    @apply data-open:animate-in data-closed:animate-out data-closed:fade-out-0 data-open:fade-in-0 bg-black/10 supports-backdrop-filter:backdrop-blur-xs;
+    @apply bg-black/10 supports-backdrop-filter:backdrop-blur-xs;
   }
 
   .cn-drawer-content {
```
MalithaPrabhashana, line 496: **Blocking: `DrawerContent` loses its entire direction layout in every style.**

Upstream moved the `data-[vaul-drawer-direction=*]` positioning out of this token and into the TSX (the "Direction layout is owned by each primitive TSX" note in the maia copy of this rule). This PR refreshes the CSS but does not re-port `drawer.jac`, and `drawer.jac` carries no inline copy to fall back on:

```
drawer.jac:38  "cn-drawer-content group/drawer-content fixed z-50", props["className"]
```

That is the whole class list. Diffing the resolved token against `main`, `DrawerContent` loses 16 direction utilities in every style:

```
lyra 23 lost   mira 23   nova 23   vega 23   (maia hits a separate parse bug, see style-maia.css)
- data-[vaul-drawer-direction=bottom]:inset-x-0
- data-[vaul-drawer-direction=bottom]:bottom-0
- data-[vaul-drawer-direction=bottom]:mt-24
- data-[vaul-drawer-direction=bottom]:max-h-[80vh]
- data-[vaul-drawer-direction=left]:inset-y-0 / left-0 / w-3/4   (+ right, top)
```

The result is `fixed z-50` with no inset and no size, so the drawer shrink-wraps its content instead of docking to an edge.

This is the "inline-utility drift re-port" the description defers to #8702, but `drawer` has no inline copy, so deferring it breaks the component rather than leaving it unchanged. Either re-port `drawer.jac` with the direction layout upstream now puts in the TSX, or keep the direction utilities in the token.

jac/jaclang/cli/shadcn/registry/styles/style-nova.css
```diff
@@ -1023,11 +1079,11 @@
 
   /* MARK: Sheet */
   .cn-sheet-overlay {
-    @apply data-open:animate-in data-closed:animate-out data-closed:fade-out-0 data-open:fade-in-0 bg-black/10 duration-100 data-ending-style:opacity-0 data-starting-style:opacity-0 supports-backdrop-filter:backdrop-blur-xs;
+    @apply bg-black/10 supports-backdrop-filter:backdrop-blur-xs;
   }
 
   .cn-sheet-content {
```
MalithaPrabhashana, line 1085: **Blocking: same pattern as `cn-drawer-content`.** `SheetContent` loses 16 utilities with no inline fallback, consistently in all five existing styles, including the side slide animations and the surface color:

```
- bg-background
- data-[side=bottom]:data-open:slide-in-from-bottom-10
- data-[side=bottom]:data-closed:slide-out-to-bottom-10   (+ top / left / right)
```

Sweeping every component for "utilities the token dropped that the `.jac` does not carry inline" gives 11 affected components, worst first:

```
component            lyra  maia  mira  nova  vega
drawer                 23     6    23    23    23
sheet                  16    16    16    16    16
card                    2    11     3     1    11
slider                  0     0     8     0     0
menubar                 1     4     1     1     1
context_menu            0     4     0     0     0
toggle_group            0     4     0     0     0
navigation_menu         2     3     3     2     3
sidebar                 2     3     2     2     2
toggle                  2     2     3     2     2
tooltip                 0     0     1     0     0
```

This contradicts the "Existing components do not regress" section. The tailwind-merge argument holds only when the token's old and new utilities land in the same merge conflict group. Running the project's own `tailwind-merge` over every differing class literal: 28 of 175 collapse to an identical result, and 147 do not.

jac/jaclang/cli/shadcn/registry/styles/style-maia.css
```diff
@@ -474,19 +490,32 @@
 
   /* MARK: Drawer */
   .cn-drawer-overlay {
-    @apply data-open:animate-in data-closed:animate-out data-closed:fade-out-0 data-open:fade-in-0 bg-black/80 supports-backdrop-filter:backdrop-blur-xs;
+    @apply bg-black/80 supports-backdrop-filter:backdrop-blur-xs;
   }
 
   .cn-drawer-content {
```
MalithaPrabhashana, line 497: **Blocking: this rule is invisible to `parse_style_css`, so `cn-drawer-content` ships unresolved in maia.**

The comment sits between `{` and `@apply`, but the parser requires them adjacent:

```python
r'\.(cn-[\w-]+)\s*\{\s*@apply\s+([^;]+);'
```

So the token never enters the resolution map and `resolve_component` leaves the literal in place. Reproduced in a scaffolded project with `jac retheme --style maia`:

```
maia:  "cn-drawer-content group/drawer-content fixed z-50"
nova:  "bg-popover text-popover-foreground flex h-auto flex-col text-sm ..."
```

No background, no text color, no layout. On `main` all five styles parse every declared rule, so this is a new regression: 1 new break, 0 pre-existing.

Fix is to hoist the comment above the selector:

```css
  /* Shared surface only. Direction layout is owned by each primitive TSX. */
  .cn-drawer-content {
    @apply ...;
  }
```

Worth hardening the regex too, since it drops rules silently rather than failing.

jac/jaclang/cli/shadcn/registry/styles/style-lyra.css
```diff
@@ -1332,4 +1420,304 @@
   .cn-input-group-textarea {
     @apply rounded-none border-0 bg-transparent py-2 shadow-none ring-0 focus-visible:ring-0 disabled:bg-transparent aria-invalid:ring-0 dark:bg-transparent dark:disabled:bg-transparent;
   }
+
+  /* MARK: Menu Translucent */
+  .cn-menu-translucent {
```
MalithaPrabhashana, line 1425: **This `@apply` is missing its terminating semicolon**, and because the parser captures `[^;]+` it runs past the closing brace and swallows the next rule whole.

Consequences in lyra:

- `cn-menu-translucent` resolves to a value that literally contains `}`, `{` and `@apply`
- `cn-menu-translucent-aria` (line 1429) disappears from the token map entirely

No component references either token today, so nothing renders wrong right now, but any component that later uses `cn-menu-translucent` would get broken CSS spliced into a `className` string.

Adding the `;` restores both. With that plus the maia fix, all eight styles parse 100 percent of their declared `.cn-*` rules (verified: 416/416, 421/421, 422/422, 422/422, 421/421, 421/421, 421/421, 419/419).

jac/tests/cli/test_shadcn.jac
```diff
@@ -438,3 +438,61 @@ test "component with a peer resolves to a matching underscored import" {
...
+        attachment_src = (ui_dir / "attachment.jac").read_text();
+        assert "import from .button {" in attachment_src;
+        assert "-" not in (ui_dir / "attachment.jac").name;
+        toml_after = toml_path.read_text();
+        assert "class-variance-authority" in toml_after;
+        assert "radix-ui" in toml_after;
+    }
+}
+
+test "styles synced from upstream all parse to a token map" {
```
MalithaPrabhashana: This test cannot catch either of the two CSS parse bugs in this PR. `len > 0` and a `cn-bubble` substring both pass while `style-maia.css` silently drops `cn-drawer-content` and `style-lyra.css` silently drops `cn-menu-translucent-aria`.

The invariant that actually holds is: every `.cn-*` rule declared in the file must appear in the parsed map.

```jac
test "every declared cn-* rule is parseable in every style" {
    for style in ["lyra", "maia", "mira", "nova", "vega", "luma", "rhea", "sera"] {
        css = (registry_dir() / "styles" / f"style-{style}.css").read_text();
        declared = set(re.findall(r'\.(cn-[\w-]+)\s*\{', css));
        parsed = set(parse_style_css(style).keys());
        assert declared == parsed , f"{style}: unparsed {declared - parsed}";
    }
}
```

Separately, `assert "radix-ui" in toml_after` passes when the value is `"*"`, so it does not verify pinning the way the description claims. Asserting the resolved range (`radix-ui = "^1.6.7"`) would.

jac/jaclang/cli/shadcn/registry/components/ui/message.jac
```diff
@@ -0,0 +1,71 @@
...
+    />
+}
+
+def:pub Message(props: any) -> JsxElement {
+    attrs: dict[str, any] = {** props};
+    align = props["align"] or "start";
+    <div
+        {**attrs}
+        data-slot="message"
+        data-align={align}
```
MalithaPrabhashana, line 25: `align` is read here but never excluded from the spread, so it lands on the DOM. `attrs` is `{**props}` and is spread before these attributes, so React renders a stray `align` attribute.

Confirmed by server-rendering the compiled output:

```html
<div align="end" data-slot="message" data-align="end" class="cn-message ...">
```

Upstream destructures it (`{ className, align = "start", ...props }`), and `bubble.jac` in this same PR handles it correctly via `_bubbleExcludeKeys`. Adding a `_messageExcludeKeys` with `["className", "align", "children"]` would match both.

Minor, but the description says there is only one deliberate deviation from upstream (`AttachmentTrigger`), so this one looks unintended.

MalithaPrabhashana (commented):
Ran this end to end against a scaffolded `jac-shadcn` project rather than reading the diff, and the five new primitives themselves are in good shape: they install, type check, build under Vite, and server-render correctly, with token sets byte-identical to upstream and zero unresolved `cn-*` in all eight styles. `jac test` on the four shadcn suites gives 60 passed as described.

Two problems are worth blocking on, both in the CSS refresh rather than the new components:

1. `drawer` and `sheet` lose styling that no longer exists anywhere. Upstream moved layout out of these tokens and into the TSX, but the corresponding `.jac` files were not re-ported and carry no inline fallback. 11 components are affected; drawer and sheet are the severe ones.
2. Two style files contain CSS that `parse_style_css` cannot parse (a comment in maia, a missing semicolon in lyra). maia's makes `drawer` ship an unresolved token.

Also worth a look: the new style test cannot catch either parse bug, and `message.jac` leaks `align` to the DOM.

One docs gap not inline anywhere: the three new styles are not discoverable. `luma`, `rhea` and `sera` work (no whitelist exists), but five places still list only the old five, including two lines in the skill file this PR edits:

- `jaclang/cli/skills/jac-shadcn-components.md:239` and `:312`
- `jaclang/cli/docs/tutorials/fullstack/npm-and-libraries.md:293`
- `jaclang/cli/docs/reference/plugins/jac-client.md:1531`
- `jaclang/cli/commands/project.jac:64` (the `--style` help string)

For an agent-driven flow that reads the skill file, an undiscoverable style is close to an unshipped one.

jac/tests/cli/test_shadcn.jac
```diff
@@ -438,3 +438,86 @@ test "component with a peer resolves to a matching underscored import" {
...
+    # drops a rule (a stray comment before @apply, a missing semicolon). The
+    # invariant that actually holds: every `.cn-*` selector in the file must
+    # come back out of parse_style_css.
+    import from jaclang.cli.shadcn.shadcn_handler { _registry_dir }
+    import re;
+    styles_dir = _registry_dir() / "styles";
+    for style in ["lyra", "maia", "mira", "nova", "vega", "luma", "rhea", "sera"] {
+        css = (styles_dir / f"style-{style}.css").read_text();
+        css = re.sub(r'/\*.*?\*/', "", css, flags=re.DOTALL);
+        declared = set(re.findall(r'\.(cn-[\w-]+)\s*\{', css));
```
MalithaPrabhashana: This invariant is the most valuable part of the fix, but it can't catch the class of bug it was written for, because both sides of the equality use the same selector pattern.

```jac
declared = set(re.findall(r'\.(cn-[\w-]+)\s*\{', css));   # this pattern
parsed   = set(parse_style_css(style).keys());              # ...and the same one inside
```

Any CSS shape the *selector* half can't see is invisible to **both** sides, so the assertion passes while the token is genuinely absent from the map. A selector list is the easy case:

```
.cn-a, .cn-b { @apply flex; }
   declared = {cn-b}   parsed = {cn-b}   equal -> PASSES
   ...but cn-a never reached the resolution map
```

`.cn-a` isn't immediately followed by `{`, so neither regex sees it, and a component using `cn-a` would ship the literal token exactly the way `cn-drawer-content` did in maia.

Deriving `declared` independently fixes it, by splitting rules first and taking every `cn-*` from the selector:

```jac
declared: set = set();
for (sel, body) in re.findall(r'([^{}]*)\{([^{}]*)\}', css) {
    if "@apply" in body {
        declared |= set(re.findall(r'\.(cn-[\w-]+)', sel));
    }
}
```

I ran this against all eight current files and it produces identical counts (416 / 421 / 422 / 422 / 421 / 421 / 421 / 417), so no false alarms, and it correctly **fails** on the selector-list case above.

Two other shapes the parser still cannot see, both absent today but not prevented on the next sync: an `@apply` that isn't the first declaration in its rule (silently skipped; the stronger `declared` above would catch it), and two `@apply` blocks in one rule (only the first is captured, which neither version catches). The second would need a separate "exactly one `@apply` per rule" assertion if you think it is worth guarding.

Entirely a follow-up, not a blocker on this PR.

MalithaPrabhashana (approved):
Approving. Re-ran the full verification against `33b79eead` and everything I flagged is fixed, with no regression to the existing components.

## Verified fixed

| Finding | Result |
|---|---|
| maia `cn-drawer-content` unparseable | All 8 styles parse **100%** of their declared rules |
| lyra missing semicolon | Corrupt value gone, `cn-menu-translucent-aria` back in the map |
| `drawer` lost its direction layout | 16 utilities re-ported inline, confirmed in the compiled JS |
| `sheet` lost enter/exit animations | Re-ported inline |
| `message.jac` leaked `align` to the DOM | `_messageExcludeKeys` added, matches `bubble.jac` |
| Style test too weak to catch either bug | Replaced with the `declared == parsed` invariant |
| Dep assertion passed against `"*"` | Now asserts `radix-ui = "^1.6.7"` |
| Five stale style lists | All updated |

## How it was checked

Scaffolded a real project on **maia** (the style that shipped a broken drawer), installed all 59 components, and built a page importing **every public export: 331 symbols across 59 modules**.

- `jac check`: passes
- Bundler emitted **59 / 59** modules; Vite build clean; app serves 200
- **Zero** unresolved `cn-*` tokens in the compiled `drawer.js` / `sheet.js`, with the direction utilities present (`data-[vaul-drawer-direction=bottom]:inset-x-0`, `:bottom-0`, `:mt-24`, `:max-h-[80vh]`, `left-0`, `w-3/4`)
- `jac test` on the four shadcn suites: **61 passed**

Existing components, resolved before this PR and after, like for like across the 54 pre-existing ports:

```
style    before   after   new breaks
lyra         24      24   none
maia         20      20   none
mira         20      20   none
nova         21      21   none
vega         21      21   none
```

Identical counts, no new breaks in any style. (Those 20-24 leftovers are pre-existing `cn-rtl-flip` / `cn-menu-target` / `cn-chart` style tokens that exist in no style CSS. Unchanged by this PR, worth their own issue.)

The parser hardening is better than the fix I suggested. I proposed patching the two CSS files; stripping comments and terminating on `[;}]` closes both failure modes at the source. I checked it for collateral damage: no `/*` appears inside any `@apply` line, every `@apply` still ends with `;`, no selector lists or nested rules exist in any style file, and across **3363 tokens the old and new parsers produce zero value differences**.

## A correction I owe you

My "11 components lose styling" claim was **overstated, and I'm sorry for the extra work it implied**. Comparing gains alongside losses rather than losses alone, only `drawer` and `sheet` were genuine functional breaks. The other nine were upstream restyles where a utility was replaced rather than dropped:

- `card`: `gap-6 pb-6 pt-6` becomes `[--card-spacing:--spacing(6)] gap-(--card-spacing) py-(--card-spacing)`, the same spacing expressed through a variable
- `slider`: `size-4` to `size-3`, `ring-4` to `ring-2`, plus a new `after:` hit area, a deliberate thumb redesign
- the residual on `drawer`/`sheet`: `bg-background` to `bg-popover text-popover-foreground`, an intended surface change

My metric conflated "replaced" with "lost". You fixed the real subset and were right to leave the rest alone.

## One non-blocking follow-up

The new invariant test has a blind spot, noted inline. Not a merge blocker, but it is the thing standing between the next upstream sync and another silently unstyled component, so worth a one-line change whenever you touch this next.

**What I did not verify:** visual appearance. I confirmed every component compiles, resolves and builds, but I have not eyeballed the rendered output, so the intended restyles above (card spacing, slider thumb, surface colours) are worth a designer's glance before release rather than mine.

MalithaPrabhashana (changes requested):
Superseding my approval above. I approved with the test blind spot filed as a "non-blocking follow-up", and on reflection that was the wrong call: the invariant test **is** the guard against the exact bug class this PR exists to fix, so shipping it with a hole in it means the next upstream sync can silently reintroduce an unstyled component, which is precisely what we just spent this review cycle catching.

Everything else stands and is verified. To restate so nothing is ambiguous:

**Fixed and confirmed** (unchanged from my approval): all 8 styles parse 100% of declared rules; `drawer` and `sheet` re-ported with their layout and animations present in the compiled JS; `message.jac` no longer leaks `align`; dep assertion now checks real version ranges; five stale style lists updated; 61 tests pass. End to end on `maia`: 59/59 components install, compile and render, with **zero new unresolved tokens** against `main` in any style.

**The one thing blocking:** the `declared == parsed` assertion computes both sides with the same selector pattern, so any CSS shape that pattern cannot see is invisible to both and the test passes while the token is genuinely missing from the resolution map. Detail and a verified replacement are in the inline comment on `test_shadcn.jac`.

The fix is about six lines and I checked it against all eight current style files: identical counts (416 / 421 / 422 / 422 / 421 / 421 / 421 / 417), so no false alarms, and it correctly fails on the selector-list case the current version passes.

```jac
declared: set = set();
for (sel, body) in re.findall(r'([^{}]*)\{([^{}]*)\}', css) {
    if "@apply" in body {
        declared |= set(re.findall(r'\.(cn-[\w-]+)', sel));
    }
}
```

Optional on top of that, your call: an "exactly one `@apply` per rule" assertion would close the last shape I found (two `@apply` blocks in one rule, where only the first is captured and neither version of the test notices). Absent from every style file today, so I would not hold the PR for it.

Happy to re-approve as soon as the `declared` derivation is independent. Sorry for the churn on the review state.

MalithaPrabhashana (approved):
Approving. The blind spot is closed.

`declared` is now derived by splitting rule blocks, independent of the selector shape `parse_style_css` keys on, which is exactly the property that was missing. You also added the multi-`@apply` assertion I'd flagged as optional, so both shapes I could find are now guarded.

Verified all three properties:

| Check | Result |
|---|---|
| No false alarms on current files | 8/8 styles, `declared == parsed` exactly (416 / 421 / 422 / 422 / 421 / 421 / 421 / 417) |
| Catches a selector list | `.cn-a, .cn-b { @apply flex; }` gives `declared={cn-a,cn-b}` vs `parsed={cn-b}` and fails correctly |
| Multi-`@apply` assertion fires | detects a rule with two `@apply` blocks |
| Suite | **61 passed** |

Everything from the earlier round still stands: all 8 styles parse 100% of declared rules, `drawer` and `sheet` carry their layout and animations, `message.jac` no longer leaks `align`, the dep assertion checks real ranges, and the docs list all eight styles. End to end on `maia`, 59/59 components install, compile and render with zero new unresolved tokens against `main`.

Thanks for the turnaround, and for taking the optional assertion too.

PR #8771, opened 2026-08-29
perf(runtime): light edges, direct hops and del on edge sets (graph compiler 5.8x -> 2.0x main)

marsninja:
Stacked on #8748 (`graph-compiler`): this PR's own change is the single commit f210cbdb8d on top of that branch, but since `graph-compiler` only exists on the fork, the diff GitHub shows here includes #8748's commits as well. The diff of this change alone is marsninja/jac#98 (same commit, base `graph-compiler`). Follows the profile posted on #8748 ([comment](https://github.com/jaseci-labs/jac/pull/8748#issuecomment-5459322096)). Land after #8748, or fold it in.

## What was wrong

Against main on the same 9.5k-line fixture, the graph-shaped compiler was 5.8x slower (39.9 s vs 6.87 s) with 2.07x the objects. The profile put it in four places:

- `parent` = `[self<-:Kid:<-]` scanned the node's entire adjacency (in- and out-edges of every type in one list), 4.7M reads, 42% of compile time.
- Every typed hop was a linear `isinstance` scan: 1.29 billion `isinstance` calls, ~100 edges scanned per read because relation edges (`SymOf`, `TypeOf`, `CfgSucc`, ...) share the list with structural ones.
- Every edge was an `EdgeArchetype` + `EdgeAnchor` + two list entries: 601k EdgeAnchors for 207k syntax nodes.
- `Archetype.__setattr__` ran on every field write of every node (10M calls, 9%), and edge deletion did O(degree) `Anchor.__eq__` compares.

## What this does

**Light edges** ([archetype.jac](https://github.com/marsninja/jac/blob/graph-compiler-light-edges/jac/jaclang/runtime/archetype.jac), tail section). A field-less, directed edge between unpersisted nodes is an adjacency entry, not an object: `NodeAnchor.out_light` / `in_light` are dicts from edge class to the neighbour list in connection order, `None` until first use. Ownership follows connect direction: out-links hold their targets, in-links are `weakref.ref`s, so a tree owns its children, an interned type or codespace singleton never keeps its users alive, and the one-sided `_conn_light` connects and sink-aware severing are gone. Edges with fields (`ScopePrimary.alias`, `TypeMemberOf.name`), undirected edges and edges touching a persistent node stay EdgeAnchors. A node that is saved materializes its light edges into rows (`NodeAnchor.all_edges`); `[edge ...]` and `del` on a single light edge work on a view (`light_edge_view`). `Archetype.__setattr__` is deleted; `track_writes` installs the dirty-mark hook on an archetype only when its anchor becomes persistent (a generated same-named subclass).

**The language carries the fast path.** Both lowerings recognize the simple hop (one origin, one direction, an edge class, no predicate, node filter or chain) and emit a direct adjacency read: jac0 emits `hop0` ([jac0.py](https://github.com/marsninja/jac/blob/graph-compiler-light-edges/jac/jaclang/jac0.py), `_lower_edge_refs`), the py backend emits `hop` (`JcirGenPass._simple_hop`). `hop` falls back to `refs` for anything richer, an untyped hop, a heavy type or a persistent origin. **`del` accepts an edge set**: `del [edge x ->:T:->];` lowers to `clear0` / `clear_edges` and drops the set by class without materializing it (direction and type follow the reference). Previously `del` of an edge reference did not compile on either tier, so no existing program changes meaning. The compiler's source stays plain Jac: `set_kids`, `_role_set` and every slot setter use the new `del` form, `_attach` is `self +>:(role):+> child`, `roles.impl.jac` is untouched, and the type graph's four edge helpers are one line of syntax each.

**Walker kernel.** `run_typed`/`run_untyped` take the scope by parameter (no walk-stack lookup after every ability), `visit:0:` pushes onto a dedicated `WalkScope.front` stack (O(len(targets)) instead of copying the queue), the bytecode driver passes archetypes as handles (no per-ability unwrap wrapper), node descs are cached by tag. The `OspRuntime` interface is unchanged, so the JS and native drivers are untouched.

## Measured

Same fixture (400 defs + 80 objs, 9.5k lines), same in-process harness (`JacProgram().compile(..., no_ir_cache=True)`, gc disabled during the compile, isolated cache), main at 90bd528fe4:

| | main | graph-compiler | this PR |
|---|---|---|---|
| compile | 6.87 s | 39.9 s (5.8x) | **13.84 s (2.0x)** |
| gc objects created | 2.96M | 6.13M (2.07x) | **4.39M (1.48x)** |
| traced peak memory | 663 MB | 1381 MB (2.1x) | **1074 MB (1.62x)** |
| EdgeAnchors for the fixture | n/a | 601k | **5.5k** (only the fielded types) |
| `ability.kid` / `.body` / `if.condition` | 0.02 µs | 1.7 / 2.3 / 0.5 | **0.40 / 0.54 / 0.17** |
| `name.parent` / `find_parent_of_type` | 0.02 / 0.14 | 0.34 / 3.1 | **0.29 / 1.9** |
| `mod.kid` (~1000 kids) | 0.02 | 81 | **7.0** |
| profile self time / calls | 17.9 s / 108M | 160 s / 1.7B | **47.6 s / 315M** |

What remains is structural and predicted before measuring: a hop is a Python call (`hop0`, 12.7M calls, ~3.4 s real) and the walker kernel is a generic three-phase loop (~2.5 s real). The next two steps, if 2x is not enough, are inline hop emission (the lowering emits an expression, not a call, for the exact-type case) and a structural-walk kernel fast path for walkers whose abilities never `visit` (the compiler's passes), which is main's `traverse` with typed dispatch.

## Validation

- circle.jac output identical to the parent commit.
- An OSP smoke program compiled through the py backend (both tiers, typed and untyped hops with cross-tier dedupe, in-direction reads, `[edge ...]` views, heavy-edge predicates, chained hops, `visit :0:` DFS order, walking over heavy edges, `del e`, both `del [edge ...]` forms with mirrors removed, the `del -->` disconnect operator) passes; generated Python shows `hop(a, 2, Kid, False)` and `clear_edges(a, 2, Kid)`.
- `scripts/check_seed_manifest.py`: 74 seed modules compile under jac0. `jac fmt --check` clean on the changed files.
- Not exercised: persistence of a graph built with light edges (the `all_edges` materialization door), the JS/native kernel drivers, the lint fixer's `detached_copy` (deepcopy + `relink_light`). The serializer's API output no longer lists edge ids for transient nodes (light edges have no ids until saved).

## CI: why build-kit and jac-check killed the runner, and the fix

Every `graph-compiler` run since e23053fe7f died in build-kit's "Build the jac binary" step (and, once lint passed, in jac-check's repo-wide `jac check`) with GitHub's "runner lost communication" annotation: the VM was OOM-killed. Replaying the payload driver's own `precompile_unit` loop with RSS sampled per unit (fresh compiles, cache-busted) showed memory growing by a full dependency closure per unit, ~100-180 MB, reaching 9.5 GB by unit 48 of 631. Referrer walks after one fresh compile, with both `release_compile_closure()` calls and a full collect done, found 78 module trees and 8.5M objects still alive, all through one chain:

`Module` <- `out_light[ScopeOwner]` <- the typeshed **`builtins` Module's anchor**

Every module scope is parented to the builtins scope, and `ScopeOwner` was connected parent -> child, so the long-lived builtins stub held a strong link to every module scope ever compiled (121 entries after one unit). Out-links own their targets by design; that relation had the direction backwards, and on the base branch the same edge sat in builtins' EdgeAnchor list and pinned the trees just the same. Commit f35dc6d73b flips it to `ScopeParent` (child -> parent); `parent_scope` reads the same. Same stretch after the fix: RSS flat at 2.4-2.6 GB.

Two release-path fixes ride along: `release_compile_closure` did `gc.collect(); gc.freeze()` (mkpayload sets `JAC_GC_FREEZE_ON_RELEASE`) while the other program still held trees, freezing cycles that could then never be collected; it now unfreezes first, and collects even without the env var (3.14's incremental collector lagged a closure per unit behind). `precompile_unit` releases the unit's own program before the shared compiler program.

memray (bundled interpreter, `--trace-python-allocators`) puts the per-unit footprint itself at: `UniNode.__post_init__` 25% of self memory, `light_connect` 19%, `hop0` result copies 10%, `NodeAnchor` creation 7%, `_construct_sub_node_tab` 6%. Those are the next targets if the closure size matters (a `CodeLocInfo` and an `_in_mod_nodes` list per node, adjacency dicts per node, the copy per hop).

## Release path and cleanups (10573e4710, 004949536b, 4a2e65291a)

The release path no longer freezes. Every variant that called `gc.freeze()` without first unfreezing leaked a dependency closure per precompile unit on CI (runner killed at 18-23 minutes), while the one variant that thawed before each collection survived 115 minutes; the object that is alive at freeze time and dies later was not identified, so the permanent generation is not used at all: `release_compile_closure(keep_stubs, collect)` clears the program in place (`ModuleHub.reset` reuses the root: hub, kid stream, scope children, main slot; `BuildScratch.clear`) and runs one full collection, `precompile_unit` reuses one `JacProgram`, and mkpayload's `JAC_GC_FREEZE_ON_RELEASE` is gone. `COMPILE_GC_THRESHOLDS` suppresses automatic collections during a compile, so the collector runs exactly once per release.

`ProgramModule.main_mod` is a role accessor over `MainRole` (it was a field beside the edge, so clearing it left the edge holding the module). The type graph's `_eouts`/`_tgt`/`_edel`/`_econn` and `UniNode._role_edges`/`_edge_target` are gone in favour of hops and `del [edge ...]`; one `light_hop`/`light_clear_hop` serves both tiers; dead back-links are compacted; `sever_external_edges` is removed and `release_module` only unhooks the module from the program root. `test_module_release` covers `release_compile_closure` on a fully compiled module.

One-frame fast paths (4a2e65291a): `JacGraph.connect` writes a light edge directly for two transient nodes and a field-less class; `JacGraph.hop` answers an exact-class single-direction hop from the adjacency dict; `clear_edges`/`clear0` return early when the node has nothing of that class; the kernel registers each walk's scope per walker (`scope_of`) so `visit`/`visit :0:` append to the walker's queue without a context lookup or kernel call.

## Release note

`release_notes/unreleased/jaclang/8771.breaking.md` (three bullets: light edges, the simple hop and `del` on an edge set, the per-write hook and kernel changes).


jac/jaclang/compiler/driver/impl/program.impl.jac
```diff
@@ -238,6 +260,12 @@ impl JacProgram.release_compile_closure(
     self.type_evaluator = None;
     self.diags.clear();
     self.scratch = BuildScratch();
+    import os;
+    if os.environ.get("JAC_GC_FREEZE_ON_RELEASE") {
+        import gc;
+        gc.collect();
+        gc.freeze();
+    }
```
marsninja: Need to come back and finalize this piece

Issue #8766, opened 2026-08-28
jac-scale: a cross-generation deploy is never detected, and both generations silently fall back to per-pod local storage


Thamirawaran:
## Summary

Two defects in jac-scale let a deploy stand up the wrong database and then
lose its data without a single error. They are what jacBuilder#1556 actually
is, and they are the jac-scale half of the split agreed in #8753.

1. **A cross-generation deploy is not detected.** A deployer on one minor line
   deploying an app pinned to another renders manifests and provisions a store
   for its own generation, not the pinned one. Nothing compares the two.
2. **Both generations silently fall back to per-pod local storage.** When no
   shared database resolves inside a cluster, 0.36 provisions an embedded
   Postgres inside the pod and 0.34 opens a SQLite file on the pod filesystem.
   Per replica, gone on restart, logged at `debug`.

Defect 1 is how you get into the broken state. Defect 2 is why it destroys
data instead of crashing. Defect 2 is independently a bug and fires whenever a
cluster deploy loses its database for any reason.

Citations are against `upstream/main 6ded655268` and tag `v0.34.17`.

## Defect 1: nothing compares the pinned generation to the deploying one

`[project] jac-version` selects the pod image and the seal (#7713, #8011). It
does not select the process that renders manifests and provisions the
database, which is whatever jac the deployer happens to be running.

| runs on the pinned jac | runs on the deployer's jac |
|---|---|
| pod image | manifest rendering |
| the seal | database provisioning |
| host client build | `apply_manifests`, all of `target.deploy` |

This is not platform specific. Anyone self hosting with a 0.34 binary and an
app pinned `==0.36.1` hits it identically, because their binary is the
deployer.

**A config check alone cannot catch this.** On 0.34 what decides the database
is `mongodb_enabled: bool = True` in `[scale.kubernetes]`
(`v0.34.17 kubernetes_config.jac:14`), not `[scale.database]`. The deployer
never asks the app for a database config, so nothing is missing from its point
of view:

| | app pinned 0.34 | app pinned 0.36.1 |
|---|---|---|
| what the 0.34 deployer sees | nothing missing | nothing missing |
| what it provisions | Mongo + Redis | Mongo + Redis |
| what the pod reads | `MONGODB_URI` (present) | `JAC_DB_URL` (absent) |

The configuration is identical in both rows. Only the pin differs, so only the
pin can tell them apart.

The reverse direction is equally broken: a 0.36 deployer with a 0.34 pinned app
[...]

Issue #8753, opened 2026-08-28
RFC: deploy health from first principles: readiness proves the data, the driver generation is the app generation, and maintenance runs through the control plane


MusabMahmoodh:
## Summary

This week jaclang.org served empty docs for three days while every pod in
its namespace was Ready, and a 0.34 jachammer deploy of an app pinned to
0.36.1 stood up MongoDB and Redis for a Postgres runtime. I spent the week
tracing both, and they are the same problem seen from two sides: the
deploy pipeline verifies that processes are alive and that version
strings match, and nothing verifies that the data is true or that the
driver and the pod are the same generation of jac.

This RFC proposes three things. A pod is Ready only when a runtime-owned
integrity check passes, so "Running 1/1, app broken" becomes reportable.
The generation that runs the pod is the generation that drives the deploy,
so a hosted platform stops rendering manifests for a runtime it isn't
(with a small contract stamp as the guard on the self-hosted path). And
the toolchain's own maintenance verbs reach the deployed app's store
through the deployment target, so a platform CLI needs no kubectl.

Credit where it is due: @udithishanka's forensics on jacBuilder#1517 and
#1547 are the reason the health half exists, @kashmithnisakya's #1556
(filed today) is the exact instance of the coherence half, and
@Thamirawaran's #7713 / #8011 are the foundation the pin already stands
on. Every citation below is against upstream/main 060d290a62.

## What happened (evidence, not hypotheticals)

**Exhibit A, live now.** On jaclang.org (jac_site, jac 0.35.0 in the pod)
`doc_tree` returns `[]` for every version, `/llms.txt` is header-only,
and prev/next links are `None`, while `doc_page` returns full content.
Asking the DocVersion node for its children with no predicate returns
nothing; asking for `name == "quick-guide"` finds it. In 0.35.0
`resolve_hop` (runtimelib/impl/resolver.impl.jac) answers predicate-free
hops from the root's persisted topology index and treats an empty answer
as final; predicate hops walk the edges. The index diverged from the
edges (Aug 24's ~300 crash-loop restarts, then the fsck repair that
rewrote anchors), and nothing in the pipeline can tell: `/healthz/ready`
is a constant (scale/server/impl/serve.core.impl.jac:352-358). The site's
own `collect tree` test passes on a real 0.35.0 binary; the deployed
index is what's wrong.

```bash
H='content-type: application/json'
[...]

PR #8777, opened 2026-08-29
fix(scale): run a pinned deploy on the release it pins, and never embed a database in a cluster

Thamirawaran:
Part of #8766, deliberately not `Closes`: the reported failure is a 0.34
deployer, and this lands on main. The issue stays open until the 0.34 line
carries an adapted version of it. This is the jac-scale half of the split
agreed in #8753; the
platform side (removing the hardcoded `[scale.database]` block, plus a kind e2e)
is @MusabMahmoodh's.

## What was wrong

`[project] jac-version` selects the pod image and the seal. The jac that *runs*
the deploy is what renders the manifests and provisions the database, and
nothing compared the two.

| runs on the pinned jac | runs on the deployer's jac |
|---|---|
| pod image | manifest rendering |
| the seal | database provisioning |
| host client build | `apply_manifests`, all of `target.deploy` |

So a 0.34 deployer handed an app pinned to `==0.36.1` provisioned the store its
own generation knows, and reported success. This is not platform specific:
anyone self hosting with a 0.34 binary and a 0.36 pinned app hits it, because
their binary is the deployer.

A config check alone cannot catch it. On 0.34 the database comes from
`mongodb_enabled: bool = True` in `[scale.kubernetes]`, not from
`[scale.database]`, so the deployer never asks the app anything and nothing is
missing from its point of view. The configuration is identical whether the app
pins 0.34 or 0.36; only the pin differs, so only the pin can tell them apart.

Then the second half. The pod resolved no database url and
`shared_store()` provisioned one **inside the pod**, per replica, on the pod
filesystem, gone on the next restart, announced at `debug`. That is what turned
a wrong deploy into lost data. `_default_base_path_dir` defaults to
`[os.getcwd()]`, so that branch was effectively unconditional and the existing
`NO_DATABASE_RESOLVED` refusal was nearly unreachable.

## What this does

**Runs a pinned deploy on the release it pins.** The seal and the host client
build already shell out to the pinned jac via `host_toolchain_path`.
`target.deploy` did not, which is the whole bug. `KubernetesRealizer` now hands
the entire deploy to that release, reusing `host_toolchain_path` and
`release_binary_env` exactly as `_build_client_bundle` does. The child
re-derives config from the app's own `jac.toml`, so nothing has to be
marshalled across the process boundary and the exit code is the result.

Not under `--dry-run`, which promises no downloads, and a sentinel env var
stops the child delegating again.

**Refuses a cross-minor deploy** for callers that enter through
`KubernetesTarget.deploy` as a library rather than the CLI, which is jacBuilder
until #1254. Before the cluster is contacted, so it reaches an operator with no
kubeconfig:

```
this deployer is jac 0.34.17 and cannot deploy an app pinned to 0.36.1. the
deploying jac renders the manifests and provisions the database, so a different
minor line stands up a backend the pod cannot use. upgrade the deployer, or pin
[project] jac-version to 0.34.
```

Keyed on the minor line, no compatibility table, and the message names only
versions, never a backend, because a deployer only knows its own generation.

**Narrows an optional in `routing.jac`.** Unrelated to the deploy fix, and
called out here so it is not lost: `ensure_main_service` read `cfg.project`
behind `if cfg`, which asks whether the config is truthy rather than whether it
exists. It only surfaced once this branch widened the realizer's import closure
and `get_config()` resolved to its real `JacConfig | None` instead of `Any`.
Same identity form the other `get_config` caller already uses.

**Refuses to embed a database in a cluster.** With `KUBERNETES_SERVICE_HOST`
set and no url resolved, `shared_store()` fails through the existing `_fail`
taxonomy instead of provisioning one inside the pod. This is the backstop for
every other way a deploy loses its database, not just generation drift.

No carve-out, deliberately, and this is a change from what I agreed on #8753.
`[scale.kubernetes] database_mode = "none"` means "wire no database at all"
(schema text, and `database_provisioner.jac:21` logs "Database wiring
skipped"), so exempting it would have handed exactly those apps an embedded
server per replica on the pod filesystem, which is the thing this PR removes. A
`none` app now gets `None` from `shared_store()`, which is what no database
means and what every consumer already handles. Happy to reinstate it if you
still want it.

## Verification

Fourteen mutations, each caught: the comparison always allowing, resolution errors
refused instead of deferred, `deploy()` ignoring the refusal, the embed branch
unguarded, the cluster check always on, `--target` and `--enable-tls` dropped
from the child argv, delegation running under `--dry-run`, and the recursion
sentinel removed, the dry-run refusal deleted, the offline pin never resolved,
the child env dropped, the sentinel unset, and the child's exit code
discarded. Run against a tree reporting its real version, so the cross-minor
cases are real.

`--dry-run` refuses a cross-minor **exact** pin, which resolves with no network
at all, and stays silent on a range pin rather than download under a flag that
promises not to.

The handoff's successful path runs a real child process and asserts the deploy
it is handed, the sentinel, that `JAC_DEV_SOURCE` is stripped so the child runs
the release it was fetched for, and the exit code coming back unchanged.

Two of those mutations initially escaped: the dry-run and sentinel tests
returned `None` because the repo has no pin, not because of the guards. Both
now use a really-pinned temp project and assert no network call, which is what
made them fail on mutation.

Reverting the `deploy()` wiring made that test take 4.4s instead of 0.8s,
because the deploy then went looking for a cluster. That is the evidence the
refusal really does land before `_load_cluster_config`.

Checked in kind on colima: a pod with default settings reports
`KUBERNETES_SERVICE_HOST=10.96.0.1`, which is the value the test uses.
jac-scale never sets `enableServiceLinks`, so the default applies.

Also confirmed `[scale.database] deploy_mode = "embedded"` is a separate
StatefulSet behind Service DNS, not the app pod hosting its own database, so
the second guard does not refuse a supported configuration.

One existing test changed: `test_memory_trigger_guard.jac` pinned a fixed
`==0.34.99`, which the new refusal stops before it reaches the
config-before-arch ordering it exists to assert. Its pin now derives from the
running jac. Verified the ordering regression it guards still fails the test.

Tests use real version strings, real TOML through the real loader, real
environment variables and a real `KubernetesTarget`. Scoped suites run:
`test_version_pin`, `test_shared_store_failure`, `test_config_validation`,
`test_identity_fail_loud`, `test_dry_run_purity`, `test_database_provisioner`,
`test_deployer`, `test_client_build_fail_fast`.

## Scope

Deliberately not here:

- **Does not move the platform off 0.34** (jacBuilder#1254).
- **Does not catch same-minor drift.** A driver built from main reports 0.36.1
  by decision (#8730) and can still crash-loop a real 0.36.1 pod, so the
  `MANIFEST_CONTRACT` stamp is still needed and is not replaced by this.
- **Backports no database feature.** 0.34 stays Mongo; the fix is a clear
  error, not a new backend.
- **The 0.34 line.** Guard 1 and the delegation cherry-pick; the in-cluster
  guard has to be rewritten there, because 0.34's silent fallback is
  `SqliteMemory` at `.jac/data/anchor_store.db`
  (`memory_hierarchy.main.impl.jac:70-75`), not an embedded Postgres.
- **A real two-version run.** No 0.34 binary has yet delegated to a 0.36 one
  end to end. The handoff is unit-covered (argv, env, sentinel, exit code) but
  not proven against a live cluster. @MusabMahmoodh has offered a rig for this
  and it is the case most worth exercising before merge.
- **The `--dry-run` refusal**, which was in this PR and has been backed out.
  `build-kit` fails on this branch and only this branch: `scale/injector/binary.jac`
  is the one module that misses the JIR cache, and its seal-time compile dies
  creating a precompile database that already exists. Reverting the two source
  files to their last green content cleared it. The refusal is worth having and
  will come back as its own PR rather than holding this one hostage.

Two notes for review:

- This needs to ship on the 0.34 line to run where the bug is. Guard 1 should
  cherry-pick; guard 2 has to be rewritten there, because 0.34's silent
  fallback is `SqliteMemory` at `.jac/data/anchor_store.db` in
  `memory_hierarchy.main.impl.jac`, not an embedded Postgres.
- `auth_tokens` and `state/factory` answer a `None` store by falling back to
  in-memory with a warning. This change makes that path reachable in a cluster
  for the first time, so they now degrade loudly rather than corrupting
  silently, but they still degrade. Pre-existing, and it felt out of scope to
  change here.


MusabMahmoodh (changes requested):
Reviewed against `0b39a3c14b` (0.37.0). This is the right shape and it is the
first PR anyone has put against the #8753 thread - thank you for taking it.
Three findings, one of which I think has to change before merge because it
defeats the PR's own purpose.

---

### 1. The guard fails open, which is the defect class this PR exists to close

`binary.jac`, `deploy_generation_refusal`:

```jac
try {
    ...
    running = resolve_cli_version().strip();
} except Exception {
    return "";
}
```

An empty return means "no refusal", so **any** exception anywhere in that
block - `resolve_cli_version()`, `JacConfig.load`, `find_project_root`,
`resolve_deploy_version` - silently re-enables the cross-generation deploy
this function exists to refuse. The failure mode is the exact one #8766
describes: nothing compares the two generations, and the deploy proceeds
against the wrong backend.

Two paths reach it in practice, neither exotic: a project whose `jac.toml`
does not parse, and a `resolve_deploy_version` that goes to the network on a
`stable` channel and times out. In both cases the operator gets a deploy, not
a diagnostic.

Suggestion: catch only what you expect to be absent, let the rest raise, and
if you want to stay non-fatal, return a *refusal* on an unexpected error
rather than a pass - "could not determine the deploying generation" is
information the operator can act on, and it fails closed. Same reasoning as
`#7900` theme 1.

The `if not pinned or not running { return "" }` line below it is fine and
correct - an unpinned app genuinely has nothing to compare - so it is only
the exception path I would change.

### 2. `pinned_deploy_argv` drops `--show-yaml` (confirmed, not inferred)

`realize()` takes five parameters (`filename, target, enable_tls, dry_run,
show_yaml`) and `plugin.jac:125` forwards all five. `pinned_deploy_argv`
reconstructs only three of them.

I checked whether `show_yaml` is gated on `dry_run`, which would make this
moot - it is not. `plugin.jac:503` defaults it independently and `:525`
passes it through; only the docs example at `:471` pairs the two flags.
So `jac scale deploy app.jac --show-yaml` on an app with a pinned toolchain
does not short-circuit at the `dry_run` guard, delegates, and the child runs
without `--show-yaml`. The user asked for manifests and gets a deploy.

That is a bad failure for a flag whose whole purpose is "show me what you
would do instead of doing it". Worth either forwarding every flag or
asserting the argv is total - a reconstructed command line silently diverging
from the one the user typed is a class, not an instance, and it will bite
again the next time a flag is added.

### 3. `_deploy_on_pinned_jac` returning `None` is doing double duty

`None` means both "delegation is not applicable" (dry run, already delegated)
and "there is no pinned toolchain to delegate to" (`if not pinned`). The
first is a decision; the second is a lookup failure. They read identically at
the call site (`if delegated is not None`), so an app that pins a generation
the host cannot produce a toolchain for falls through to a same-process
deploy - which then does get caught by the `target.jac` refusal, so the
outcome is safe today. It is safe by luck of ordering rather than by
construction, though, and I would rather see the two cases distinguished.

---

### Smaller notes

- `routing.jac`: `if cfg else` -> `if cfg is not None else` is a real fix (a
  present-but-falsy config was being read as absent) but it is unrelated to
  this PR's stated scope and is not mentioned in the body. Worth a line in
  the description so it is not lost.
- The `(major, minor)` comparison is the right granularity and matches the
  issue's framing. Good call not comparing patch.
- `NO_DATABASE_IN_CLUSTER` and `_in_cluster()` reading
  `KUBERNETES_SERVICE_HOST` is the correct signal and the message is well
  written - it says what happened and what to do.

### Testing

`test_dry_run_purity.jac` (+44) covers the dry-run refusal. I do not see
coverage for either of the two paths that actually run on a real deploy: the
`target.jac:236` refusal, and `_deploy_on_pinned_jac`'s delegation. The
delegation one is the harder ask because it shells out - but a test that
asserts `pinned_deploy_argv` reproduces every parameter `realize()` accepts
would have caught finding 2 and needs no subprocess at all.

Happy to put this on a real two-generation cluster on my rig if that helps -
offer stands, and it is the case I most want to see exercised before it
merges.

PR #8779, opened 2026-08-30
fix(jac-scale): route gateway and sv-to-sv RPC traffic through the KEDA HTTP interceptor for scale-to-zero targets

### LINKED ISSUE #7959: jac-scale: route ingress and gateway traffic through the KEDA HTTP interceptor for http_activation scale-to-zero (follow-up to #7709)
## Follow-up to #7709

PR #7709 added declarative `jac.toml` configuration for KEDA HTTP Add-on activation (scale-to-zero on HTTP traffic). It creates the `InterceptorRoute` and the HTTP `ScaledObject` for any service with `http_activation.enabled = true`, and cleans both up on teardown.

This issue tracks the one gap that #7709 intentionally left out of scope: jac-scale does not route inbound traffic through the interceptor, so scale-to-zero is not yet safe to rely on end to end.

## Problem

The KEDA HTTP Add-on only wakes a zero-replica workload when the request passes through its interceptor proxy. The interceptor is what holds the request, signals the scaler, waits for a pod to become Ready, and then forwards the request.

Today jac-scale creates the `InterceptorRoute` and `ScaledObject`, but it does not change where traffic actually goes:

- The generated Ingress backend still points at the app's own Service (`{name}-service`).
- The gateway service registry still resolves the service's ClusterIP directly.

Neither path goes through the interceptor. So with `min_replicas = 0`:

1. The workload scales to zero after the cooldown period.
2. The next request hits the app Service, which now has no endpoints.
3. The request is refused. The interceptor never sees it, so nothing wakes the pod.

Put simply, with the routing left as-is, turning on scale-to-zero becomes an availability risk instead of a cost saving.

## Current behavior and workaround

Only enable `http_activation` on a service whose inbound traffic you have already pointed at the KEDA HTTP interceptor proxy yourself, for example through a hand-written Ingress backend or gateway route. This limitation is called out as a warning in the jac-scale Kubernetes reference. `jac scale plan` lists every HTTP-activated service, so operators can audit which workloads are affected before deploying.

## Suggested fix

Make "service X is HTTP-activated" a fact that the routing layer consumes, so the scaling side and the routing side come from one config and cannot drift:

1. When a service resolves to HTTP activation, emit its Ingress backend (and gateway route) pointing at the KEDA HTTP interceptor proxy Service instead of the app's own Service.
2. Derive the `InterceptorRoute` host and path rules from the same Ingress host and route path the service already uses, so the two are always consistent.
3. Keep the app's own Service as the interceptor's forward target. That part is already correct.

The per-service scaling decision resolved in #7709 is the natural single source for this. The routing builder can read the same scaling mode rather than re-deriving which services are activated.

## Acceptance criteria

- [ ] For a microservice with `http_activation.enabled = true`, the generated Ingress backend for that service points at the KEDA HTTP interceptor proxy, not the app Service.
- [ ] The `InterceptorRoute` host and path rules are derived from the same Ingress host and route path, with no separate manual configuration required.
- [ ] For the monolith deploy path, inbound traffic for an HTTP-activated app reaches the interceptor.
- [ ] A cold (zero-replica) workload wakes and serves its first request end to end, verified against a real cluster.
- [ ] The gateway is never routed through an interceptor, so it stays warm. This is consistent with the gateway exemption from #7709.
- [ ] The warning documented in #7709 is updated or removed once routing is wired.
- [ ] Unit tests cover: an activated service produces an interceptor-targeted backend, a non-activated service is unchanged, and the host and path rules match between the Ingress and the `InterceptorRoute`.

## Out of scope

- The `jac.toml` config surface and the create/teardown lifecycle for `InterceptorRoute` and `ScaledObject`, both delivered in #7709.
- Non-HTTP KEDA triggers.

Developer-Linus:
Fixes #7959

## Summary

jac-scale lets a microservice scale to zero replicas via the KEDA HTTP Add-on (`http_activation.enabled = true`). The deploy correctly created the `InterceptorRoute` and `ScaledObject`, but the runtime never actually sent traffic through the interceptor. A request that arrived after a service scaled to zero was refused instead of waking the pod, so scale-to-zero was not safe to use in practice. This PR makes that routing automatic for both gateway-forwarded requests and internal sv-to-sv RPC calls, with no manual Ingress or gateway configuration required.

## Problem

Peer-to-peer URL resolution for the Kubernetes runtime was implemented independently in three places, and none of them were aware of `http_activation`:

1. `runtime_resolver.jac`, the dominant path that feeds the live gateway pod's service registry.
2. `manifest_builder.jac`, which bakes a fallback peer URL into each pod's environment at deploy time.
3. `plugin.jac`'s `ensure_sv_service`, a lazy fallback used when a peer is not yet registered.

All three unconditionally resolved a service's own ClusterIP DNS name, even for a service configured to scale to zero. On top of that, the `InterceptorRoute` routing rule that decides which request belongs to which service was populated purely from hand-typed `jac.toml` config, with no connection to how callers actually addressed the service. The two sides could silently drift apart.

A second, separate traffic pattern made this worse: sv-to-sv RPC calls (`walker_call`, `service_call`) use paths like `/function/name` or `/walker/name` with no route prefix, so two activated services sharing one interceptor could not be told apart by path alone.

## Fix

**One shared resolver.** Added `service_routing.jac`, a small dependency-free module that is now the single source of truth for whether a service is reached directly or through the interceptor. Both deploy-time code and the live gateway call into it, so the three previous, independently-drifting resolvers are gone.

**Host-header disambiguation.** When a service is activated, callers connect to the interceptor's address but set an HTTP `Host` header equal to the target's own real Kubernetes Service DNS name. This is the pattern documented by the KEDA HTTP Add-on itself, not an invented scheme. The `InterceptorRoute` rule for cluster-internal traffic is now auto-derived from that same DNS name, so it can never drift from what the resolver produces.

**Consistent plumbing.** The gateway's forward path and both sv-to-sv RPC call paths (`service_call` and `walker_call`) now read a single bundled `(url, host header)` pair per service instead of resolving the two independently and stitching them together. `sv_client`'s registry clears any stale Host override when a peer is re-registered, so an old override can never outlive the service state it came from.

**Bug fix found along the way.** `walker_call` never consulted the live Kubernetes service registry at all, unlike `service_call`. It always fell back to a stale, deploy-time-baked URL. Both now go through the same resolution path.

**No competing env var.** `manifest_builder.jac::_build_env` no longer bakes a `JAC_SV_<PEER>_URL` env var for HTTP-activation peers. That env var is checked before the interceptor route is ever computed, so leaving it in place for these peers would let it override the routing decision above on every call.

**Fail loud, not quiet.** The two places where this logic falls back to direct addressing on an unexpected error (a scale-config load failure, a dynamic route lookup failure) now log at warning level instead of silently defaulting or logging at debug, since a silent fallback here means a real bypass of the exact routing this PR adds.

## Why this approach

Real-world systems that solve the same problem (gRPC's resolver, Envoy, Kubernetes EndpointSlice, Knative's activator) all bundle an address with its routing metadata in one record, not two separately resolved values a caller has to join. The design here follows that pattern: one resolver, one bundled result, used everywhere. The core jaclang plugin hookspec (`get_sv_registry`) was left untouched since it is a generic interface with no other consumer that needs the richer shape; a new jac-scale-only accessor sits alongside it as the real source of truth.

Reusing the service's existing DNS name as the Host header, instead of inventing a new naming convention, keeps the change minimal and matches the KEDA HTTP Add-on's own documented usage pattern.

## Changed files

| File | Change |
|---|---|
| `jac/jaclang/scale/runtime/context/service_routing.jac` | New. Single source of truth for interceptor vs. direct routing decisions. |
| `jac/jaclang/scale/deploy/target/kubernetes/runtime_resolver.jac` | Peer resolution now goes through the shared resolver; adds a bundled route accessor; a config-load failure now logs at warning. |
| `jac/jaclang/scale/deploy/autoscale/http_activation_config.jac` | Auto-derives the internal-routing `InterceptorRoute` rule from the service's own DNS name. |
| `jac/jaclang/scale/deploy/autoscale/keda_autoscaler.jac` | Interceptor address now sourced from the shared constant instead of a local literal. |
| `jac/jaclang/scale/runtime/context/util.jac` | Adds a DNS host helper. This PR originally also added a standalone RFC1123 sanitizer here; after rebasing onto main, that was found to duplicate main's own independently-added `k8s_safe_name` and was removed (see Review follow-ups). |
| `jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac` | Delegates name sanitization to the shared `k8s_safe_name` helper; `_build_env` no longer bakes a direct peer URL for HTTP-activation peers, so it can no longer override the interceptor route at runtime. |
| `jac/jaclang/scale/runtime/provider.jac` | Adds the bundled route accessor (`get_sv_service_routes`); `get_sv_registry` now builds on it. This class was `JacScalePlugin` in `jac/jaclang/scale/plugin.jac` when this PR was first written; an unrelated main restructure renamed and relocated it to `JacScaleRuntime` here before this PR merged (see Review follow-ups). |
| `jac/jaclang/server/sv_client.jac` | Registry housekeeping for the bundled route accessor (see Review follow-ups: an earlier paired Host-override table here turned out to be dead and was removed). |
| `jac/jaclang/scale/runtime/rpc/rpc.jac` | `service_call` and `walker_call` resolve a bundled route and set the Host header; fixes the `walker_call` registry bug; a failed dynamic lookup now logs at warning. |
| `jac/jaclang/scale/runtime/gateway/service_registry.jac` | `ServiceEntry` gains a `host_override` field. |
| `jac/jaclang/scale/runtime/gateway/microservice_gateway.impl.jac` | Forward paths apply the Host override before sending a request to a peer. |
| `jac/jaclang/scale/runtime/orchestrator/orchestrator.jac` | Populates both URL and Host override from one bundled lookup. |
| `jac/jaclang/cli/docs/reference/plugins/jac-scale-kubernetes.md` | Replaces the outdated manual-routing warning; updates a stale CLI reference. |
| `jac/jaclang/cli/docs/tutorials/production/kubernetes.md` | Same warning update in the production tutorial. |
| `jac/jaclang/scale/tests/deploy/test_service_routing.jac` | New. Direct, mock-free tests of the shared resolver. |
| `jac/jaclang/scale/tests/deploy/test_http_activation_config.jac` | Covers the auto-derived rule and confirms it does not bypass existing validation. |
| `jac/jaclang/scale/tests/deploy/test_http_activation_microservices.jac` | New regression test proving the gateway never activates through the full deploy path, not just the isolated scaling decision. |
| `jac/jaclang/scale/tests/microservices/test_gateway.jac` | New test proves the Host header actually reaches the wire over a real socket. |
| `jac/jaclang/scale/tests/server/test_sv_auth_forward.jac` | Updated for the new bundled route accessor; adds a Host-header propagation test. |
| `jac/jaclang/scale/tests/deploy/test_pod_env.jac` | Isolates scale config state so these tests are not affected by the new config lookup; adds two tests for the `_build_env` peer-URL fix. |
| `jac/jaclang/scale/tests/deploy/keda_http_activation_rpc_wake_real_e2e.sh` | New. Real-cluster e2e proving both wake paths with no manually-set Host header; waits for the worker to actually reach zero replicas before the first wake test. |
| `jac/jaclang/scale/tests/fixtures/keda_http_activation_rpc_wake_e2e/` | New fixture app (gateway-warm service plus a scale-to-zero worker) for the e2e above; `http_forward_timeout`/`rpc_timeout` raised to 240s so a real cold pod boot (jac install + first compile) doesn't race the gateway's/RPC's 30s/10s defaults on a slower or shared runner. |
| `release_notes/unreleased/jaclang/8779.bugfix.md` | Release note fragment for this fix. |

## Testing

- New unit tests for the shared resolver, run with no mocking since it is pure config in, decision out.
- Extended existing tests confirm the auto-derived routing rule does not weaken the existing rule-collision validation.
- A new integration test spins up a real loopback HTTP server and confirms two services sharing one backing URL are forwarded with two different Host headers, proving the header reaches the actual socket rather than just being present in a headers dict.
- A new regression test confirms the gateway never activates through the full `generate_manifests` deploy path, even when a shared top-level default would otherwise apply to it.
- Full local test sweep across the affected directories: all passing, no regressions.
- Verified locally with `jac scale deploy --dry-run` that the new e2e fixture produces valid manifests with no rule collisions.
- Ran the real-cluster e2e script against a local `kind` cluster (KEDA core + HTTP Add-on installed): both the external gateway wake and the internal sv-to-sv RPC wake pass end to end, including the scale-back-to-zero check after cooldown on each.
- Two new unit tests in `test_pod_env.jac` cover the `_build_env` peer-URL fix: an HTTP-activation peer never gets a baked env-var URL; a peer without it still does.

## Notes

One thing is intentionally out of scope:

- WebSocket connections proxied through the gateway are not yet routed through the interceptor.

jac-scale has one shared Ingress for the whole app, not a per-service one, so the routing decision this PR fixes happens one hop later, inside the gateway's own forwarding logic rather than at the Ingress itself. Two of #7959's acceptance criteria assumed a per-service Ingress and have been reworded on the issue to match; see the issue comments for detail.

## Review follow-ups

Re-reviewed after a rebase onto current `main`. Five findings, each fixed at the root rather than patched around:

- **Health monitor probing an interceptor-routed service.** `check_all_health` hit `{entry.url}/healthz` with no `Host` header for a service routed through the interceptor, so the probe missed the interceptor's route table, three failures demoted the entry to `UNHEALTHY`, and the gateway started 503ing every request to it. Adding the `Host` header to the probe would work too, but it would wake the pod every interval and defeat scale-to-zero, so `check_all_health` now trusts a `host_override` entry directly instead of polling it.
- **Dead warn-once dedupe in the wrong place.** `get_sv_service_routes()` never actually raises — it already catches everything internally and returns `{}` — so the warn-once dedupe in `rpc.jac`'s `_resolve_call_target` guarded a branch that could never fire, and the fallback path it guarded reached `get_sv_service_routes()` a second time through `get_sv_registry()`, so a real failure would have warned twice anyway. Moved the dedupe to the two places that actually warn: `get_sv_service_routes`'s own except block in `plugin.jac` (one flag, since that lookup fails or succeeds atomically for every service at once), and `resolve_peer_route`'s per-service config load in `runtime_resolver.jac`, which runs on every fleet member on every call, not only at deploy time as originally assumed.
- **Dead Host-override table.** Nothing ever called `sv_client.register_host_override`, in production or in tests, so `_host_overrides` was always empty and the `host_override_for` fallback always returned `None`. Removed the table and its touch points; `rpc.jac`'s fallback now returns `None` directly, since the K8s path already gets its Host header from `get_sv_service_routes`. `build_k8s_service_routes` (`service_routing.jac`) was in the same state — nothing outside its own test called it — so it was removed too, and that test rewritten against the real production resolver (`build_k8s_registry_routes`) backed by an on-disk `jac.toml` instead of a hand-built config dict.
- **Stale `entry-point` in the RPC-wake e2e fixture.** A `main`-side change now rejects any `entry-point` value ending in `.jac`, requiring the dotted-module-name form the sibling fixture already used; this fixture's merge onto current `main` was textually clean but stopped parsing. Updated to match.
- **Stale doc paragraph.** The interceptor-routing doc described a deploy-time-baked peer URL fallback that `_build_env` no longer bakes for HTTP-activated peers. Replaced it with the cold-start timeout budget an operator actually needs: `rpc_timeout`/`http_forward_timeout` must exceed the service's cold boot time, since the 10s/30s defaults fail the first call to a service sitting at zero replicas.

**Rebase onto current `main`.** A same-day, unrelated main restructure renamed `JacScalePlugin` to `JacScaleRuntime` and moved it from `jac/jaclang/scale/plugin.jac` to `jac/jaclang/scale/runtime/provider.jac`, and independently added its own `get_sv_registry` K8s routing logic there. Merged current main rather than rebasing linearly, since this branch's history already carries several prior main-sync merges and the file also moved once before under an earlier restructure, so a linear rebase would have re-triggered the same conflict on every commit that touches it. Findings from the resulting conflicts:

- **`JacScalePlugin` relocation.** Ported this PR's `get_sv_service_routes` and the updated `get_sv_registry` onto main's `JacScaleRuntime` in `runtime/provider.jac`, rather than resurrecting the old class at its old location. Two production call sites (`rpc.jac`'s `_resolve_call_target`, `orchestrator.jac`'s `_populate_k8s_urls`) and the warn-once test in `test_sv_auth_forward.jac` were still pointed at the old `jaclang.scale.plugin` import path after the textual merge; the first would have raised `ImportError` and the second would have silently logged failure and set no interceptor route. Both now import `JacScaleRuntime` from `jaclang.scale.runtime.provider`.
- **Duplicate RFC1123 sanitizer.** Main had independently added its own `k8s_safe_name` (with a `GATEWAY_NAME` special case) to `context/util.jac` around the same time this PR added `k8s_safe_service_name` (without one) for the same purpose. The two didn't textually conflict, since they sat in different parts of the file, so a plain merge left both behind silently. `service_routing.jac` now calls `k8s_safe_name` directly and `k8s_safe_service_name` is deleted. Behavior-preserving: peer/service routing never resolves a name equal to `GATEWAY_NAME`, so the extra branch in `k8s_safe_name` never fires for this call site.


MusabMahmoodh (changes requested):
Reviewed against `0b39a3c14b` (0.37.0). First review on this PR - apologies
for the wait, and thank you for closing the gap #7709 explicitly left open.
The design is right: one resolver, Host-header disambiguation taken from the
Add-on's own documented pattern rather than invented, and the
`InterceptorRoute` rule derived from the same helper the callers use so the
two sides cannot drift. That last point I checked specifically and it holds -
`k8s_service_dns_host` feeds both the rule (`manifest_builder`) and
`ServiceRoute.host_header`, so the match is by construction.

Two findings, and they are the same finding twice: the PR's thesis is "one
source of truth", and it leaves two second copies behind.

---

### 1. The interceptor address is hardcoded, and a configurable one already exists

`service_routing.jac`:

```jac
glob KEDA_INTERCEPTOR_SERVICE_ADDRESS: str = "keda-add-ons-http-interceptor-proxy.keda:8080";
```

But `keda_autoscaler.jac:27` already carries this as a **parameter**:

```jac
interceptor_service_address: str = "keda-add-ons-http-interceptor-proxy.keda:8080",
```

So on a cluster that set a non-default value - a different Helm release name,
or KEDA installed outside the `keda` namespace, both of which
`keda_autoscaler.impl.jac:861` already anticipates by parsing a namespace out
of the address - the `ScaledObject` and `InterceptorRoute` are created against
the **configured** interceptor while the runtime sends traffic to the
**hardcoded** one. Requests go to a service that does not exist, and the
target never wakes.

That is the exact drift class this PR exists to remove, reintroduced on the
axis the PR did not consolidate. The resolver should take the address from
the same config the autoscaler reads.

### 2. `resolve_http_activation_config` is a second copy of `_get_http_activation_config`

The new module's own docstring says it:

> Merge `[scale.microservices.services.<svc_name>.http_activation]` over
> `[scale.kubernetes.http_activation]`, **the same precedence
> `ManifestBuilder._get_http_activation_config` uses.**

And `_get_http_activation_config` (`manifest_builder.jac:1353-1362` on your
head) is still there, unchanged, doing the same merge and the same
`enabled`-gating with its own code.

The PR deletes three drifting *URL* resolvers, which is the win - but it adds
a fourth copy of the *enablement* decision, and documents the duplication
instead of removing it. A comment saying "this matches that" is what a drift
looks like the day before it drifts; `_get_http_activation_config` should
delegate to `resolve_http_activation_config` and keep only the
`self._service_config(svc_name)` / `self.k8s_config.http_activation` lookup.

That is a small change and it makes the "single source of truth" claim in the
summary literally true rather than nearly true.

---

### Checked and not raising

- **Host header carries no port** while the direct path uses `:{port}`. I
  expected this to be a mismatch against the `InterceptorRoute` rule and it is
  not - the rule is built from `k8s_service_dns_host(service_name, namespace)`
  with no port either (diff L55-56), and your test at L1141 pins exactly that
  equality. Correct as written.
- The shallow `merged.update(per_service)` is right here, and matches the
  documented per-service override semantics from #7709.

### Testing

`test_service_routing.jac` is the right shape - genuinely mock-free,
dict-in/dict-out, and the assertion tying `host_header` to the same helper
`build_http_activation_spec` uses is the one that will catch a future drift.
Good.

The gap is that neither finding above would fail any test in this PR. A test
that asserts the resolver's interceptor address equals the autoscaler's
configured one, and a test that asserts the two merge functions agree on a
per-service override, are both cheap and would pin the consolidation this PR
is about.

I have a two-service kind rig with the HTTP Add-on installed and can run
`keda_http_activation_rpc_wake_real_e2e.sh` against a non-default interceptor
namespace if that would help demonstrate finding 1 - say the word.

MusabMahmoodh (approved):
Re-reviewed at `565ddc02b`. Both findings are closed properly, and the fix went
further than what I asked for. Approving.

**Finding 1 (interceptor address).** `resolve_interceptor_service_address` is
the single decision point now, and I checked every construction site rather
than taking the summary on trust: `AutoscalerFactory`'s `"keda"` branch, and
all three in `target.jac` (the `_resolve_scaling` path, the manifest path, and
the `destroy_collection` call). All four route through it, and
`KEDAAutoscaler`'s own default is now `KEDA_INTERCEPTOR_SERVICE_ADDRESS`
instead of a second copy of the literal, so there is no longer a bare string
anywhere a caller could pick up by accident.

The `AutoscalerFactory` bug you found on the way is the more valuable half of
this. A factory that accepts a `config` parameter and drops it on the floor is
worse than one that takes no config at all, because the call site reads as
configured. That was not in my review and it is a better catch than either of
my two.

**Finding 2 (merge duplication).** `_get_http_activation_config` delegates now,
`_service_config` correctly stays for its five other callers, and the second
implementation is gone rather than commented as equivalent.

---

### On your note about the second test

You are right, and I want to be explicit that flagging it was the correct call
rather than an admission against interest. A test that pins "these two produce
the same output" when they are the same function cannot fail, and saying so is
what stops the next reader from treating it as regression cover it is not.

What it does buy is real, and it is worth naming in those terms: it is a
structural pin, not a behavioral one. It fails the day someone reintroduces a
second merge path, which is exactly the drift this PR exists to prevent. Keep
it, and consider a one-line comment in the test saying that, so its purpose
survives you.

### One nit, non-blocking

`resolve_http_activation_config`'s docstring still reads:

> the same precedence `ManifestBuilder._get_http_activation_config` uses

That was true when it was written and is now backwards: `_get_http_activation_config`
uses *this* function's precedence, not the other way round. It is the same
species as the duplication you just deleted, a comment asserting agreement
between two things where there is now only one thing. Suggest:

> Merge `[scale.microservices.services.<svc_name>.http_activation]` over
> `[scale.kubernetes.http_activation]`. The single implementation of this
> precedence; `ManifestBuilder._get_http_activation_config` delegates here.

Fold it into any later push, or leave it and I will not hold the PR for it.

### On the e2e request

I cannot give you that verification today and I do not want to imply otherwise.
`keda_http_activation_rpc_wake_real_e2e.sh` against a non-default interceptor
namespace needs a live kind cluster, and my rig protocol
(`jac/jaclang/scale/tests/TESTING.md`, section 1) requires the binary under
test be built from this head and proven to be serving this source before a
"green on a real cluster" claim means anything. I have not done that, so I am
approving on a source read, not on a cluster run.

That is the honest state of it, and I would rather say so than let an approval
carry weight it has not earned. If the KEDA stack lands behind #8838 I will run
it against the combined head and report the result there, which is a better use
of one cluster spin than four separate ones.

Nothing here blocks. Good PR, and the factory catch is the kind of thing that
only turns up when someone actually follows the config through.

jac/jaclang/scale/runtime/rpc/rpc.jac
```diff
@@ -95,6 +95,33 @@ def _provider_url(app_name: str) -> str {
...
+    try {
+        import from jaclang.scale.plugin { JacScalePlugin }
+        routes: dict[str, any] = JacScalePlugin.get_sv_service_routes();
+        route = routes.get(module_name);
+        if route is not None and route.url {
+            return (route.url, route.host_header);
+        }
+    } except Exception as e {
+        import logging;
+        logging.getLogger(__name__).warning(
```
MusabMahmoodh: Making this loud is right, but `_resolve_call_target` is called per RPC (`:152` and `:208`), so a route lookup that is failing rather than flaking emits one warning per sv-to-sv call, and `rpc.jac` has no module-level logger to hang a de-dupe on. The scenario that trips it is a route table that is unreachable or malformed, which is exactly when the logs need to stay readable.

A module-level `glob _logger = logging.getLogger(__name__)` plus a warn-once per `module_name` (a small set of already-warned names, cleared when a lookup succeeds) keeps the signal and drops the flood. The `runtime_resolver.jac` one is fine as-is, since that path is deploy-time rather than per-call.

MusabMahmoodh (approved):
Re-reviewed the Sep 7 push. The three review items came back done properly rather than patched over: `service_routing.jac` importing `k8s_service_dns_host` from `util.jac` instead of redefining it is behaviour-identical (both are `f"{service_name}.{namespace}.{K8S_CLUSTER_DNS_SUFFIX}"`), and `register()` clearing a stale `host_override` closes the case where a service that stops being HTTP-activated keeps an old Host header.

The e2e change is the one I want to call out specifically. Noticing that a fresh deploy starts `worker` at 1 replica, so the first wake test would pass even with interceptor routing completely broken, and then adding the wait-for-zero before it, is finding that your own test was vacuous. That is the failure mode most e2e scripts in this repo still have and nobody has gone looking for it.

One thing on the fail-open logging, below. Approving either way; it is a shape change, not a correctness one.

jac/jaclang/scale/runtime/rpc/rpc.jac
Developer-Linus: Fixed in 3ad75195f: added a module-level `logger` and a `_dynamic_route_lookup_warned` set. The except branch only warns for a `module_name` not already in that set, adds it once it does, and the success branch discards it — so a lookup that starts working again re-arms the warning for the next time it breaks, per your suggestion.

jac/jaclang/scale/runtime/context/service_routing.jac
```diff
@@ -0,0 +1,82 @@
...
+    safe_name = k8s_safe_service_name(svc_name);
+    real_host = k8s_service_dns_host(f"{safe_name}-service", namespace);
+    http_cfg = resolve_http_activation_config(
+        dict(kubernetes_cfg.get("http_activation", {})), app_scale_cfg
+    );
+    if http_cfg is not None {
+        interceptor_address = resolve_interceptor_service_address(kubernetes_cfg);
+        return ServiceRoute(url=f"http://{interceptor_address}", host_header=real_host);
+    }
+    return ServiceRoute(url=f"http://{real_host}:{port}", host_header=None);
```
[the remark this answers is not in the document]
Developer-Linus: Not a genuine drift — this conflates two different things that share confusing names.

`GATEWAY_NAME` ("__gateway__") is a synthetic, deploy-only identifier for the ingress/router pod, not the root app itself. `generate_service_pod_specs` (manifest_builder.jac) builds a pod spec for the root fleet member under its own real app name (e.g. "web") exactly like any other service, and *separately* builds a second pod spec keyed literally "__gateway__" for the router. The container entrypoint branches on that literal: `if [ "$JAC_SV_NAME" = "__gateway__" ]; then exec jac scale gateway; ...` — only that synthetic pod runs the router process. The root app's own pod runs the root app's own code.

`_resolve_scaling`'s exemption is `svc_name == GATEWAY_NAME`, i.e. it only ever matches the literal string "__gateway__". Since `FleetMember.app` is always a real app name, the root fleet member's own pod is never exempted from HTTP_ACTIVATION — it goes through the same `_resolve_scaling` path as every other service, gets a real `InterceptorRoute` when activated, and `fleet.names()` (which feeds `get_sv_service_routes`/`resolve_peer_route`) never contains the literal "__gateway__" string in the first place, since it isn't a fleet member. So `k8s_safe_service_name`'s docstring ("peer/service routing never targets the gateway") is accurate: the ingress-router pod is never a routing target, but the root app's own pod correctly is, on both the deploy side and the runtime side, with no drift between them.

Ran the full affected test suite after this investigation (test_service_routing.jac, test_sv_auth_forward.jac, test_http_activation_config.jac, test_http_activation_microservices.jac, test_pod_env.jac, test_gateway.jac — 222 tests) to confirm root-app HTTP-activation routing behaves as expected; all pass.

jac/jaclang/scale/runtime/orchestrator/orchestrator.jac
```diff
@@ -245,16 +245,17 @@ def start_gateway_only(fleet: FleetSpec, gateway_config: dict[str, any]) {
...
         return;
     }
     for (svc_name, _entry) in registry.entries.items() {
-        url = dns_map.get(svc_name, "");
-        if url {
-            _entry.url = url;
+        route = routes.get(svc_name);
+        if route is not None and route.url {
+            _entry.url = route.url;
+            _entry.host_override = route.host_header or "";
```
MusabMahmoodh, line 258: After the rebase this runs under #8658's health monitor (`start_health_monitor` at :114). Its `check_health` probes `{entry.url}/healthz` with no Host header (process_manager.impl.jac:450), so for an activated service it hits the interceptor with a host it has no route for. Three failed probes demote the entry to UNHEALTHY, and `forward_to` (microservice_gateway.impl.jac:634) 503s every gateway request to it from then on. Putting the Host override on the probe would be worse, since it would wake the pod every 2s and it would never reach zero. I think `check_all_health` should skip entries with `host_override` set and count them healthy, pinned by a case in `test_k8s_health_probe.jac`. The kind e2e predates this interaction, so it needs a rerun on this head.

jac/jaclang/scale/runtime/rpc/rpc.jac
```diff
@@ -95,6 +98,36 @@ def _provider_url(app_name: str) -> str {
...
+    try {
+        import from jaclang.scale.plugin { JacScalePlugin }
+        routes: dict[str, any] = JacScalePlugin.get_sv_service_routes();
+        route = routes.get(module_name);
+        if route is not None and route.url {
+            _dynamic_route_lookup_warned.discard(module_name);
+            return (route.url, route.host_header);
+        }
+    } except Exception as e {
+        if module_name not in _dynamic_route_lookup_warned {
```
MusabMahmoodh: This branch can't fire on a failing lookup. `get_sv_service_routes` catches everything itself (plugin.jac:625-630) and returns `{}`, so the per-call warning still comes from plugin.jac:627, twice per RPC because the `_provider_url` fallback reaches it again through `get_sv_registry`. The dedupe belongs inside `get_sv_service_routes`, and the new test only passes because `failing_routes` raises, which the real function never does. I was also wrong earlier that the `runtime_resolver.jac:37` warning is deploy-time only: `build_k8s_registry_routes` runs it for every fleet member on each of those calls.

jac/jaclang/scale/tests/fixtures/keda_http_activation_rpc_wake_e2e/jac.toml
```diff
@@ -0,0 +1,48 @@
...
+
+[dev]
+
+[scale.kubernetes]
+bundle_storage_class = "jac-http-e2e-rwx"
+namespace = "jac-http-rpc-wake-e2e"
+
+[apps.app]
+kind = "service"
+entry-point = "app.jac"
```
MusabMahmoodh: #9088 on main only accepts dotted `entry-point` values (`validate_entry_point` rejects anything ending in `.jac`, called from the `[apps.*]` parse at apps.impl.jac:117). This fixture stops parsing once the branch takes main, even though the merge is textually clean. The sibling `keda_http_activation_e2e/jac.toml` was moved to `"app"` in that PR, and `worker.jac` at :20 needs the same change.
```suggestion
entry-point = "app"
```

jac/jaclang/cli/docs/reference/plugins/jac-scale-kubernetes.md
```diff
@@ -574,8 +574,10 @@ graph TD
...
+!!! note "Interceptor routing is automatic"
+    Once `http_activation.enabled = true` for a service, jac-scale routes traffic to it through the interceptor automatically -- both gateway-forwarded requests (the Ingress path) and sv-to-sv RPC calls (walker/function invocations from another service) resolve the interceptor's proxy address instead of the app's own Service, with a `Host` header set to the service's own Service DNS name so the interceptor's `InterceptorRoute` can tell which target a request is for. No manual Ingress or gateway rewiring is required. The Ingress itself still points at the gateway's own Service, unchanged: the gateway is exempt from `http_activation` and always stays warm, so it never needs to be woken.
+
+    Two paths remain outside this: a deploy-time-baked peer URL env var (`JAC_SV_<PEER>_URL`) still resolves the app's own Service directly -- it is only ever used as a fallback when the live registry has no entry, which should not happen in ordinary operation -- and WebSocket connections proxied through the gateway are not yet routed through the interceptor.
```
MusabMahmoodh: This paragraph went stale inside this PR. `_build_env` no longer bakes a peer URL for activated peers (manifest_builder.jac:1069), and the var is `JAC_APP_<APP>_URL` now; the "known gap" test at test_service_routing.jac:169 makes the same claim. What an operator needs here instead is the cold-start budget. `rpc_timeout` defaults to 10s and `http_forward_timeout` to 30s, and a read timeout is not retried, so the first call to a service at zero fails unless both exceed its boot time. Your own fixture had to raise both to 240s.
```suggestion
    A cold wake holds the request until the pod is Ready, so set `rpc_timeout` and `http_forward_timeout` (under `[apps.<name>.scale]`, or `http_forward_timeout` under `[scale.gateway]`) above the service's cold boot time, and `timeout_readiness` if you set interceptor timeouts; with the 10s / 30s defaults the first call to a service sitting at zero replicas fails. WebSocket connections proxied through the gateway are not yet routed through the interceptor.
```

jac/jaclang/server/sv_client.jac
```diff
@@ -48,12 +49,22 @@ def register(app: str, url: str, route: str = "") {
         _routes[app] = route;
         _base_routes[app] = route;
     }
+    _host_overrides.pop(app, None);
+}
+
+def register_host_override(module_name: str, host: str) {
```
MusabMahmoodh: Nothing calls `register_host_override`, in production or in tests, so `_host_overrides` is always empty and the `host_override_for` pairing at rpc.jac:127 always returns None. The K8s path gets its Host header from `get_sv_service_routes`, so this table and its touch points in `sv_client` can go, and the fallback can return `(_provider_url(module_name), None)`. `build_k8s_service_routes` (service_routing.jac:67) is in the same state: only the cross-talk test calls it, while production resolves through `build_k8s_registry_routes`.

MusabMahmoodh (changes requested):
Re-reviewed the rebased head against what has landed since my approval. The routing core still holds and the #9102 overlap is independent (it only changed walker-envelope rehydration, the dispatch into `service_call`/`walker_call` is untouched), but #8658's gateway health monitor now demotes every interceptor-routed entry to UNHEALTHY, which 503s the gateway wake path, so I'm moving this back to changes-requested until that's handled.

jac/jaclang/scale/runtime/orchestrator/orchestrator.jac
Developer-Linus: Fixed in 5082e4556: check_all_health now trusts entries with host_override set instead of probing them, and marks them healthy directly. Adding the Host header to the probe would have worked too, but as you noted it would wake the pod every interval and keep it from ever reaching zero, so trusting the route was the better fix. Added a pinning test in test_k8s_health_probe.jac that keeps a host_override entry healthy across repeated checks even when its URL points at an unreachable port, confirming it is never actually probed.

jac/jaclang/scale/runtime/rpc/rpc.jac
Developer-Linus: Fixed in ffabd5cfb. You were right that get_sv_service_routes never raises, so the try/except here was dead code. Moved the warn-once dedupe into get_sv_service_routes itself in plugin.jac (a single flag, since that lookup fails or succeeds atomically for every service at once) and into resolve_peer_route's per-service config load in runtime_resolver.jac, since that is the site that actually runs per fleet member on every call. Replaced the old test with two tests that exercise those real failure sites instead of a fake that could raise where production never does.

jac/jaclang/scale/tests/fixtures/keda_http_activation_rpc_wake_e2e/jac.toml
Developer-Linus: Fixed in c66380554. Updated entry-point to "app" for the app service and "worker" for the worker service, matching the dotted-module-name form the sibling fixture already uses.

jac/jaclang/cli/docs/reference/plugins/jac-scale-kubernetes.md
Developer-Linus: Fixed in c66380554. Replaced the paragraph with the cold-start timeout budget guidance you suggested, since that is what an operator actually needs to know here.

jac/jaclang/server/sv_client.jac
Developer-Linus: Fixed in d977399f1. Removed _host_overrides, register_host_override, and host_override_for from sv_client, and updated rpc.jac's fallback to return None directly since nothing ever populated that table. Also removed build_k8s_service_routes from service_routing.jac for the same reason, and rewrote its cross-talk test against the real production resolver, build_k8s_registry_routes, backed by an on-disk jac.toml instead of a hand-built config dict.

jac/jaclang/scale/runtime/orchestrator/process_manager.impl.jac
```diff
@@ -473,6 +473,13 @@ impl ServiceProcessManager.check_all_health -> dict[str, bool] {
             continue;
         }
 
+        if `entry.host_override {
```
MusabMahmoodh, line 476: This is what I suggested last round, and it leaves a gap I should have named then: nothing at runtime now notices a broken interceptor route. An uninstalled HTTP add-on, a renamed interceptor Service or a deleted InterceptorRoute leaves these entries HEALTHY for the life of the gateway while traffic keeps going to them. Probing the entry URL without the Host override, counting any HTTP response as reachable and only a connection error as unhealthy, would check the interceptor without waking the pod. Fine as its own issue if you'd rather not widen this PR.

MusabMahmoodh (commented):
All five from last round are in, and taking the host-override table out together with `build_k8s_service_routes` is the right amount of deletion.

The branch needs a rebase before anything else, and it is not only textual. #9157 (merged today) deleted `JacScalePlugin` from `scale/plugin.jac`; the registry lookup now lives on `JacScaleRuntime` in `scale/runtime/provider.jac`, and `k8s_safe_name` moved to `scale/runtime/context/util.jac`. `git merge-tree` against main conflicts in `plugin.jac`, `runtime_resolver.jac` and `manifest_builder.jac`, and two files merge cleanly but break:

- `_resolve_call_target` in `rpc.jac:107` imports `JacScalePlugin` with no try, so it raises `ImportError`.
- `_populate_k8s_urls` in `orchestrator.jac:248` imports it inside a try, so it only logs "K8s URL population failed" and sets no interceptor route, silently.

The warn-once tests in `test_sv_auth_forward.jac` still reach for the `jaclang.scale.plugin` logger and `plugin_mod._sv_service_routes_warned`. After the rebase, main's `util.k8s_safe_name` and this PR's `k8s_safe_service_name` are two copies of one sanitizer, so one of them should go.

jac/jaclang/scale/runtime/orchestrator/process_manager.impl.jac
Developer-Linus: Agreed, taking you up on the offer to leave this as its own issue rather than widening this PR further. Will file it separately.

Developer-Linus: Filed as #9182.

MusabMahmoodh (approved):
The rebase is clean: the routes live on `JacScaleRuntime`, both import sites point at `runtime.provider`, `k8s_safe_service_name` is gone rather than kept beside main's `k8s_safe_name`, and #9182 carries the probe gap.

kugesan1105:
current status:
image

PR #8806, opened 2026-08-31
fix(cli): jac --version resolves the manifest by walk-up, so dev mode reports the tree it runs

### LINKED ISSUE #8804: jac --version in dev mode reports the launcher binary's baked-in version, not the source tree's
## Summary

In dev mode (`[dev] jaclang_source` set), `jac --version` reports the version baked into the installed launcher binary instead of the version of the source tree it is actually running. On a checkout where `jac.toml` says `0.37.1`, a launcher built at `0.36.0` prints:

```
$ jac --version
🛠  jac dev mode — using compiler source at /home/kuggix/jac/jac
jac 0.36.0  (Linux x86_64)
```

The banner is truthful (the 0.37.1 compiler source is really what runs), but the version line is stale, so dev builds misreport what they are.

## Root cause

`resolve_cli_version()` in `jac/jaclang/cli/impl/manifest_route.impl.jac` tries, in order:

1. `$JAC_DEV_SOURCE/jac.toml`
2. `Path(jaclang.__file__).parent.parent / 'jac.toml'`
3. `pkg_version('jaclang')`

`jaclang_source = "jac"` points at the directory *containing* `jaclang/`, so in this repo the dev source is `<repo>/jac`. Both probe 1 and probe 2 therefore look for `<repo>/jac/jac.toml`, which does not exist: the manifest lives one level up at the repo root (`<repo>/jac.toml`). Both probes miss, and the resolver falls through to `pkg_version('jaclang')`, which reads the dist metadata bundled inside the launcher binary from whenever it was last built.

The header comment in `jac.toml` documents that config resolves to the nearest ANCESTOR `jac.toml` (walk-up, nearest wins), but this resolver never walks up, so it disagrees with the documented resolution semantics. Both dev-mode probes share the same off-by-one.

Note `0.36.0` appears nowhere in the source tree; it comes only from the stale binary metadata, which is what makes this confusing to diagnose.

## Impact

- Any dev-mode checkout reports the version of the last `zig build` install, not the tree being run. Rebuilding hides the mismatch until the next version bump.
- Anything keying off `jac --version` output in dev workflows (bug reports, support triage, scripts) gets the wrong version.

## Fix

Make the dev-mode probes walk up from the starting directory to the nearest ancestor `jac.toml`, matching the documented config resolution, instead of checking only the starting directory itself. This fixes both probe 1 (`JAC_DEV_SOURCE`) and probe 2 (`jaclang.__file__`), which start from the same place. The `pkg_version` fallback stays as the non-dev path.

## Repro

1. Check out the repo with `[dev] jaclang_source = "jac"` enabled and `version = "0.37.1"` in the root `jac.toml`.
2. Install a launcher built at an older version (for example a `zig build` from a `0.36.0` tree).
3. `jac --version` prints the dev-mode banner with the correct source path, but `jac 0.36.0`.

kugesan1105:
Fixes #8804.

## Problem

In dev mode (`[dev] jaclang_source` set), `jac --version` reports the version baked into the installed launcher binary instead of the version of the source tree it actually runs. On a 0.37.1 checkout with a launcher built at 0.36.0:

```
$ jac --version
🛠  jac dev mode — using compiler source at /home/kuggix/jac/jac
jac 0.36.0  (Linux x86_64)
```

`resolve_cli_version()` probes `$JAC_DEV_SOURCE/jac.toml` and then `Path(jaclang.__file__).parent.parent / 'jac.toml'`. Since `jaclang_source = "jac"` points at the directory *containing* `jaclang/`, both probes resolve to `<repo>/jac/jac.toml`, which does not exist: the manifest lives one level up at the repo root. Both miss, and the resolver falls through to `pkg_version('jaclang')`, the dist metadata bundled inside the launcher from its last `zig build`. The stale version string appears nowhere in the tree, which makes the mismatch confusing to diagnose, and rebuilding only hides it until the next version bump.

## Fix

Replace the single-directory checks with a walk-up to the nearest ancestor `jac.toml`, matching the config resolution contract documented at the top of `jac.toml` (nearest ancestor wins, no merging). Nearest-wins is honored strictly: a nearer manifest without a `[project].version` does not fall through to a farther ancestor; the probe just yields nothing and resolution moves to the next probe. The `pkg_version('jaclang')` fallback is unchanged and still serves the non-dev binary path.

Both dev probes (`JAC_DEV_SOURCE` and the `jaclang.__file__` fallback) share the new helper since they start from the same place.

## Validation

Live, on this checkout:

- `jac --version` in dev mode now prints `jac 0.37.1` (the tree), banner unchanged.
- `JAC_NO_DEV_SOURCE=1 jac --version` outside the repo still prints the binary's own `0.36.0`, so the non-dev fallback is untouched.

New tests in `jac/tests/cli/test_version_resolution.jac`:

- version found in the starting directory
- walk-up to an ancestor manifest (the shape that was broken: `dev_source = <repo>/jac`, manifest at `<repo>`)
- nearest-wins with no merging: a nearer version-less `jac.toml` masks an ancestor that has one
- no manifest anywhere on the walk-up path yields nothing
- `resolve_cli_version` end to end with `JAC_DEV_SOURCE` pointing below the manifest

All 5 pass; full `jac test jac/tests/cli` is green apart from 2 known parallel-run `test_mcp` resource flakes that pass in isolation both with and without this change.

### DIFF
--- a/jac/jaclang/cli/impl/manifest_route.impl.jac
+++ b/jac/jaclang/cli/impl/manifest_route.impl.jac
@@ -14,26 +14,33 @@ def _read_version_from_toml(toml_path: Path) -> (str | None) {
     return None;
 }
 
+# jac.toml resolves to the nearest ancestor (walk-up, nearest wins), so the
+# version must too: dev_source points at the dir containing jaclang/, while
+# the manifest usually lives at the repo root above it.
+def _nearest_toml_version(start: Path) -> (str | None) {
+    for candidate in [start] + list(start.parents) {
+        toml_path = candidate / 'jac.toml';
+        if toml_path.exists() {
+            return _read_version_from_toml(toml_path);
+        }
+    }
+    return None;
+}
+
 impl resolve_cli_version -> str {
     dev_source = os.environ.get('JAC_DEV_SOURCE');
     if dev_source {
-        toml_path = Path(dev_source) / 'jac.toml';
-        if toml_path.exists() {
-            ver = _read_version_from_toml(toml_path);
-            if ver {
-                return ver;
-            }
+        ver = _nearest_toml_version(Path(dev_source).resolve());
+        if ver {
+            return ver;
         }
     }
 
     jac_file = jaclang.__file__;
     if jac_file {
-        source_toml = Path(jac_file).parent.parent / 'jac.toml';
-        if source_toml.exists() {
-            ver = _read_version_from_toml(source_toml);
-            if ver {
-                return ver;
-            }
+        ver = _nearest_toml_version(Path(jac_file).resolve().parent.parent);
+        if ver {
+            return ver;
         }
     }
 
--- a/jac/tests/cli/test_version_resolution.jac
+++ b/jac/tests/cli/test_version_resolution.jac
@@ -1,0 +1,72 @@
+"""Version-resolution tests for `jac --version` (#8804).
+
+`resolve_cli_version` must honor the manifest walk-up contract documented in
+jac.toml: the nearest ANCESTOR jac.toml wins, no merging. The dev-mode probes
+start from the directory containing jaclang/ (JAC_DEV_SOURCE, or derived from
+jaclang.__file__), while the manifest usually lives at the repo root above it,
+so a probe that only checks the starting directory silently falls through to
+the launcher binary's baked-in dist metadata and misreports the version.
+"""
+
+import os;
+import tempfile;
+import from pathlib { Path }
+import from jaclang.cli.manifest_route { _nearest_toml_version, resolve_cli_version }
+
+
+def _make_tree(base: Path, rel: str, toml_at: str, version: (str | None)) -> Path {
+    start = base / rel;
+    start.mkdir(parents=True, exist_ok=True);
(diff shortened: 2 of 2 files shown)


christianwilkins (commented):
Reviewed head `767c109ce0ee4f5c0a97bb69dc9aea5d928617c4`: the complete resolver/test patch, nearest-manifest stopping rule and the existing distribution fallback. No blocking resolver defect found.

Local validation: `jac --version` reports 0.37.1, matching this checkout. I also executed the five new test bodies through a direct entry probe against the production resolver; all five assertions passed, including the nearer version-less manifest and JAC_DEV_SOURCE override. No code changes made.

Current-head CI is not green: jac-check explicitly lists both changed Jac files as needing `jac fmt --lintfix`, and Contribution Checks requires a release-note fragment. These are concrete author actions. The native lane also fails the annexed OSP construction test (1512 passed, 1 failed, 4 skipped), and runtime fails the client diagnostic-location test (1768 passed, 1 failed, 8 skipped). I have not reproduced those latter failures against the base and cannot call them unrelated.

Not ready to merge: add the release fragment, format the two files, resolve or establish a baseline for the two test failures, and obtain green CI on the resulting head. Maintainer review follows that validation.

PR #8814, opened 2026-08-31
refactor(cache): one substrate, one identity, one lifecycle

### LINKED ISSUE #8812: Cache infrastructure reorg: one substrate, one identity, one lifecycle
This is a design/umbrella issue. It comes out of a full survey of every cache in the tree, on disk and in memory, and proposes a reorganization of how the codebase caches code across its sub-infrastructures. Part 1 is the survey of what exists today, Part 2 is the proposed architecture, Part 3 is sequencing. The fix-now list at the bottom is worth doing independently of the reorg.

## Part 1: What is there today

The codebase has roughly **40 distinct caches** across five lifetime tiers (per-node, per-pass, per-compile, per-program, process-global, plus the on-disk stores), and they share almost no infrastructure.

**Five different cache-root resolvers.** `jac/jaclang/jac0core/cache_paths.py` declares itself the single source of truth (platform-aware, three OS branches), but `jac/jaclang/scale/injector/binary.jac:618`, `jac/jaclang/dist/fused/materialize.jac:102`, `jac/jaclang/data/impl/pgembed.impl.jac:16`, and `jac/jaclang/byllm/model_cache.jac:22` each re-derive their own, with three different env-var conventions for the same concept (`JAC_CACHE_HOME`, `JAC_SCALE_CACHE_DIR`, `JAC_MODELS_DIR`) and disagreeing macOS behavior. The launcher's payload writer (`materialize.jac`, POSIX-only resolution) and its reader (`jac/jaclang/dist/fused_binary.jac:109`, via `cache_paths`) resolve to different directories on macOS. And `jac/jaclang/cli/cli_boot.jac:6` carries a verbatim copy of the platform logic pointed at a `bytecode/` dir nothing writes, which is why `jac purge` silently misses `apps/`, `rt/`, `scale-binaries/`, `models/`, and `pg/`, and points at the wrong bootstrap path.

**Six atomic-write implementations**, split between `os.replace` (correct) and `Path.rename` (broken overwrite on Windows), and between `with_name` (appends) and `with_suffix` (replaces the last dotted segment) for temp names. The `with_suffix` variant is exactly the #8655 bug, and four instances of the same idiom survive (`jac/jaclang/compiler/driver/jir.jac:832`, `jac/jaclang/compiler/driver/impl/compiler.impl.jac:215`, `jac/jaclang/dist/jab.jac:36`, `jac/jaclang/dist/jab_vendor.jac:124`). Exactly one cache in the tree takes a lock (pgembed's flock); `jac/jaclang/dist/jab.jac:36-50` has a real rmtree-then-rename race between two processes materializing the same app digest.

**Eviction exists three times, differently.** The format cache (`jac/jaclang/cli/format_cache.jac`) has the only complete design: TTL + LRU + entry cap + touch-on-hit + throttled prune behind a sentinel. The JIR generation sweep has TTL only, on the write path only (a read-only workload never sweeps). The launcher `rt` GC is path-scoped only. Bootstrap `.jbc`, stubcat `.bin`, kernel-unit sources, `apps/`, `scale-binaries/`, and `models/` grow forever.

**Format versions live in five unrelated files** (`jir.jac:9` = 23 int, `stubcat/format.jac:4` = 3 int, `format_cache.jac:8` = "v1" str, `jac0core/sealed.py:86` = 8 with a back-compat accept-list, `jab_vendor.jac:73` = 1 inline JSON literal). The JIR header still carries three dead fields plus a vestigial one-byte compressed payload left over from the tree cache removed in #8190, and `jac/jaclang/compiler/tools/code_intel.jac:10` imports seven `jir` symbols it never uses.

**In memory, "cache" and "program state" are conflated.** `mod.gen` is simultaneously the codegen output and the memo table, which is why `evict_unit` has to hand-copy five fields into a slim clone to keep JIT engines alive past AST eviction. One program holds three module caches (`mod.hub`, `scratch.boundary_parse_memo`, `scratch.na_dep_units`, so the same file can be two `Module` objects with two symbol tables in one compile), two freshness oracles (mtime stamps vs. content keys), two demotion caches plus a third structure just for misses, and four alias fields (`prog.mod`, `errors_had`, `warnings_had`, `mtir_map`) that desynchronize if anything rebinds instead of mutating in place.

**Correctness findings that fell out of the survey**, worst first:

- `AnalysisCaches.analyses_run` is keyed by `id(Module)` and never cleared (`jac/jaclang/compiler/driver/progstate.jac:229`; `release_compile_closure` clears every other organ and skips `analysis` entirely). Under `jac check`'s reuse-one-program-and-release-per-file loop, a recycled id makes a brand-new module look already analyzed and silently skips type checking.
- LSP `delete_module`/`rename_module` pass `file://` URIs into the path-keyed hub, so both are permanent no-ops (`jac/jaclang/lsp/server/impl/server.impl.jac:112,120` vs `engine.impl.jac:740,749`); only `close_module` converts correctly. `_fanout_dependents` also mutates the program before taking the write lock (`engine.impl.jac:84-89`), racing the dispatcher thread mid-compile.
- `_NA_OSP_KERNEL_MOD` / `_NA_FMT_KERNEL_MOD` pin whole compiled modules process-wide and are mutated per compile (`jac/jaclang/compiler/backends/native/na_ir_gen/osp.impl.jac:41` stamps `gen.nogc_enforced` on the shared module), so a nogc-enforced compile bleeds the flag into the next, unrelated one.
- `TypeEvaluator._shared_intern` survives `forget_module` (`type_evaluator.impl.jac:6223` vs `imported.impl.jac:457`), pinning and serving `ClassDetailsShared` records whose symbol table belongs to an evicted module. Visible over a long LSP session.
- `jac run --no-cache` scrubs `**/.jac/cache` downward from the file's own directory (`jac/jaclang/cli/commands/impl/execution.impl.jac:571`), so in the normal layout (project root above the file) it misses the project cache entirely, and it does not set `JAC_REBUILD`, so the `_precompiled` bundle still serves.
- HMR pops the hub and `sys.modules` but not `interop_bridge._pyb_class_memo` or `typecache._cache` (`jac/jaclang/server/impl/hmr.impl.jac:360-381`): reloaded classes keep stale native shadows and type hints. The SPA route latch is also never reset.
- `jac/jaclang/data/pgpool.jac:8` `_pools` and `jac/jaclang/byllm/parallel.jac:16` `_WORKER_POOL` are live at fork time under the test runner and never reset in children: parent and children hand out the same PG sockets; the executor's threads do not survive fork. (The bootstrap `.jbc` writer is correctly fork-hardened, so the hazard was known once and applied once.)
- `jac/jaclang/runtime/runtime.jac:93` negative-caches a failed jac-scale import permanently; `_default_llm_instance` survives byllm config changes; `jac serve` never re-reads jac.toml (`config.impl.jac:1380` memoizes with no mtime check).
- Hub key normalization is inconsistent: the compiler uses `Path.resolve()`, the LSP writes raw paths, and `find_module` papers over it with an O(n) `normcase` scan of the whole hub (`jac/jaclang/compiler/driver/impl/program.impl.jac:308-339`).
- `[cache] enabled` in jac.toml is declared, parsed, surfaced in `jac config`, and read by nothing.

The pattern behind all of it: every subsystem built its own cache from scratch, and the best design in the repo (the format cache) shares nothing with the worst (jab materialization).

### 2.1 A cache substrate everything builds on

Create `jaclang/cache/` as a real subsystem, with a thin pure-Python floor kept in `jac0core` (it must be importable pre-bootstrap, since `meta_importer.py` writes the `.jbc` cache before any .jac code can run):

- **`jac0core/cachefs.py`** (~60 lines, absorbs `cache_paths.py`): the only root resolver. Platform triplet, one override env var (`JAC_CACHE_DIR`), plus `atomic_publish(path, bytes)` and `atomic_publish_dir(tmp_dir, dest)` doing the pid-suffixed `with_name` + `os.replace` dance correctly, once. The `materialize.jac` write-probe/TMPDIR fallback moves here so writer and reader can never disagree again.
- **`jaclang/cache/store.jac`**: a `CacheStore` value: namespace, schema version, entry discipline (key-named blob like the format cache, or sentinel-validated directory like `rt/.ok`), and a prune policy (TTL, max entries, max bytes). The prune engine is `FormatCache`'s, promoted: `.last-prune` throttle sentinel, LRU by mtime with touch-on-hit, orphaned-temp sweep. Optional flock for check-then-act stores (stubcat build, jab materialization; pg dist already does this).
- **`jaclang/cache/fingerprint.jac`**: sha256 only (retire the sha1s in pgembed/CEF and the header crc32 at the next format bump), length-framed multi-part hashing (`FormatCache.make_key` generalized), the compiler-digest/generation logic moved out of `jir.jac`, and one shared two-tier freshness check (mtime fast path falling back to content verify) used by JIR validity, hub stamps, `.nademote`, and the wasm libc merge instead of four private variants.
- **`jaclang/cache/registry.jac`**: every store registers `{name, root, schema_version, prune_policy, owner}`. This is the piece that makes tooling staleness structurally impossible: `jac purge` iterates the registry instead of hardcoding paths, so a new cache is purgeable on the day it is born. Format-version constants stay with their store definitions but are all visible in one place.

### 2.2 One disk layout, with non-caches moved out

```
~/.cache/jac/
  CACHEDIR.TAG                  # standard marker, backup tools skip it
  modules/<generation>/         # .jir + native/ + .nademote (as today)
  bootstrap/                    # .jbc, now with TTL prune
  kernels/  stubs/              # kernel units, stubcat, now with prune
  runtimes/<key>/               # launcher rt, one resolver for reader+writer
  apps/<digest>/                # jab images, flock'd materialization, TTL prune
  bin/<channel>/                # scale binaries, TTL prune
  models/                       # byllm weights (explicit-manage, exempt from prune)
  browser/                      # fold ~/.cache/jacbrowser in, TTL prune
```

Two things leave the cache root entirely. `pg/main` is durable data, not cache: a machine-wide Postgres cluster living under a directory users are told is safe to delete is a footgun; it belongs in `~/.local/share/jac` (XDG data home). Desktop OAuth tokens likewise. The project side stays `<project>/.jac/cache/` on the same substrate, and the loose-file fallback that currently scatters `.jac/cache` dirs next to standalone files (`jir.jac:514`, the reason `examples/*/.jac/cache` litters the tree) should hash into the global `modules/` root instead.

### 2.3 In memory: make the four lifetime tiers explicit

The progstate organ split already gave the right skeleton; the fix is to finish the lifecycle it implies.

**Derived artifacts live on the thing they are derived from.** The elegant fix for the `analyses_run` bug is not "clear the table" but "delete the table": analysis provenance becomes a field on `Module` (or `mod.gen`), so it dies with the module, is immune to id reuse, and collapses the two "is this analyzed" oracles that four call sites currently reconcile by hand. Same rule kills most `id()`-keyed program-lifetime tables: `id()` keys are fine per-pass, banned at any lifetime longer than the objects they key.

**One module identity, one module cache.** A single `hub_key()` (resolved realpath) used by compiler, LSP, and `ModuleManager`, with URIs converted at the protocol boundary and nowhere else. This fixes the LSP delete/rename no-ops and deletes `find_module`'s O(n) fallback scan. Then collapse the three module caches into one hub with a staging level per entry (parse-only, analyzed, native-dep-compiled), upgradeable in place, so one file is one `Module` per program instead of two objects with two symbol tables in one compile. The two demotion caches plus the miss-set become one `NaVerdict` store with a memory tier and the `.nademote` disk tier behind it.

**The interned tier becomes actually immutable.** The stub catalog, prefetched types, `_CODESPACES`, `DIAGNOSTICS`, and the kernel modules form a legitimate process-global tier, but only if nothing mutates it. The kernel modules get frozen after build (`nogc_enforced` passed as a compile parameter, not stamped onto the shared module), and per-process mutable dicts hanging off catalog records (`_derived_special_types`) get the same append-only, identity-stable contract the intern table has. `_shared_intern` gets a per-module index so `forget_module` can sweep it, the general rule being that every evaluator-side table either survives forgetting a module by construction (keys cannot dangle) or registers a forget hook.

### 2.4 A lifecycle protocol instead of ad hoc resets

Invalidation knowledge is currently smeared across `release_compile_closure`, `reset_process_state`, `forget_module`, HMR's hand-rolled sweep, and about 30 module-level globals that nothing resets. Replace that with a small set of named events (`compile_end`, `module_forgotten`, `program_released`, `process_reset`, `config_changed`, `post_fork`) and let each cache register for the ones it responds to. The payoffs are exactly the bugs found: HMR firing `config_changed`/`module_forgotten` reaches interop shadows, the type-hint cache, and the negative-cached scale-provider import; the test runner firing `post_fork` fixes pgpool and the byllm thread pool; and every "never cleared" cache becomes a visible declaration in the registry rather than an accident someone finds in a sweep like this one.

### 2.5 One choke point between compiler and disk

`get_bytecode` is today the only reader/writer of the JIR disk cache; `JacProgram.compile()` bypasses it entirely, which is why the LSP and `jac check` get zero benefit from it. Put load/validate/persist/evict behind one `ArtifactStore` service on the compiler (unifying `is_module_cache_valid`, `_persist_module_cache`, and `evict_unit`'s slim-clone logic), so import, check, and LSP all go through the same door and can each opt in deliberately.

### 2.6 The user surface

A `jac cache` subcommand driven by the registry: `ls` (per-store sizes and policies), `prune` (run policies now), `purge [store|--all]`. `jac run --no-cache` changes meaning from "scrub some directories downward" to "bypass cache reads this run" (set the same flags `JAC_REBUILD`/`JAC_NO_PRECOMPILE` already implement). And `[cache] enabled` in jac.toml either gets wired to the store layer or deleted.

## Part 3: Sequencing

Each phase lands independently:

1. **Substrate**: `cachefs` + store/prune/fingerprint, then mechanical adoption by the five root-resolvers and fourteen atomic-write sites. Low risk; permanently retires the #8655 bug class and the macOS reader/writer split.
2. **Registry + `jac cache`**: fixes purge, adds the missing eviction to bootstrap/stubcat/apps/bin, moves pg data out of the cache root.
3. **Dead-code sweep**: code_intel's orphan imports, the JIR header's three dead fields and vestigial payload (next `FORMAT_VERSION` bump), the `.jac_ir` globs, cli_boot's `bytecode/` path.
4. **In-memory lifecycle**: analysis provenance onto `Module`, `AnalysisCaches` into `release_compile_closure`, hub-key canonicalization + LSP URI boundary, event protocol with fork/HMR subscribers.
5. **Consolidation**: staged hub absorbing the boundary-parse and na-dep caches, `NaVerdict` unification, `ArtifactStore` choke point, frozen kernel modules.

Acceptance for the campaign: `jac cache ls` enumerates every store in the tree with its location, size, schema version, and prune policy, and `jac purge` clears everything it lists; no cache exists that is not in the registry.

## Fix-now list (independent of the reorg)

1. `analyses_run` id-reuse hole (can skip type checking in `jac check` today).
2. LSP `delete_module`/`rename_module` URI-vs-path no-ops, and the unlocked `_fanout_dependents` mutation.
3. The jab materialization rmtree/rename race.
4. `jac purge`'s dead paths (`bytecode/` nonexistent, `bootstrap/` wrong path, five stores unreachable).

marsninja:
Implements the full scope of #8812 as a clean break: one substrate, one identity, one lifecycle.

## Status: review diff, not yet validated

This tree has not been compiled or tested. It exists to show the complete shape of the #8812 reorg before deciding how to land it (whole, split by phase, or cherry-picked). One defect is already known: the pre-commit hook surfaced a circular import booting `jaclang/runtime/runtime.jac` (most likely the module-load event subscriptions added there), so expect CI to be red until that and whatever else shakes out are fixed.

## What is in here

**Substrate (new)**
- `jac0core/cachefs.py`: the only cache/data root resolver (`JAC_CACHE_DIR`, `JAC_DATA_DIR`), the only atomic-publish (`pid`-suffixed `with_name` temp + `os.replace`, directory publish with concurrent-winner tolerance), and the only lock helper. Retires the #8655 bug class by construction.
- `jaclang/cache/store.jac`: `CacheStore` with the format cache's prune engine promoted (TTL, LRU by mtime with touch-on-hit, entry and byte caps, sentinel-throttled).
- `jaclang/cache/registry.jac`: every store declared in one table with policies and schema versions; `purge`/`prune` iterate it, so tooling can never go stale against the layout again.
- `jaclang/cache/fingerprint.jac`: length-framed sha256 helpers and the shared two-tier freshness check.
- `jaclang/cache/events.jac`: lifecycle bus (`module_forgotten`, `program_released`, `process_reset`, `config_changed`, `post_fork`).

**Disk layout (breaking)**
- The `jir/` level is gone: `modules/ bootstrap/ kernels/ stubs/ runtimes/ apps/ bin/ models/ browser/ pg/` under one root, with a `CACHEDIR.TAG`.
- The embedded Postgres cluster moves to the data root (`~/.local/share/jac/pg/main`): durable state no longer lives under a purgeable cache.
- `JAC_CACHE_HOME`, `JAC_SCALE_CACHE_DIR`, `JAC_MODELS_DIR`, `JACBROWSER_CACHE` are removed; `JAC_CACHE_DIR`/`JAC_DATA_DIR` replace them all.
- Loose-file compiles use the global `modules/` store; no more `.jac/cache` scattered next to standalone files. Project caches are unchanged.
- `cache_paths.py` deleted; `cachefs` is the boot-tier module (seed manifest updated). The launcher's `materialize.jac` is the single permitted mirror of the platform triplet (it runs before CPython) and now matches on all three platforms, fixing the macOS reader/writer split.

**Compiler**
- JIR format 24: the dead tree-cache header fields and vestigial payload are gone (header 32 -> 20 bytes; `sealed.py` twin updated), `EXTERNAL_REF`/`create_header`/`SEC_SYMINDEX` and code_intel's orphan imports deleted.
- Analysis provenance lives on `Module.gen.analyses_run`: the `id(Module)`-keyed table that could silently skip type checking under `jac check` is gone.
- The hub gains staging (`parse`/`analyzed`/`native`) and absorbs the boundary parse memo, so one file is one `Module` per program.
- `ArtifactStore` is the single choke point between compiler and disk (load/persist/evict-slim), routing `get_bytecode` and selfhost eviction.
- `_shared_intern` is indexed per module and swept by `forget_module`; kernel modules are frozen after first build (the per-compile `nogc_enforced` stamp had zero readers on the kernel object); `NaVerdictStore` unifies the three demotion globals with the `.nademote` sidecar.

**Tooling and runtime**
- LSP: URIs convert to fs paths at the protocol boundary (delete/rename were permanent no-ops), hub keys are normalized everywhere, the fanout mutation moved under the write lock, and semantic tokens rebuild only for changed modules instead of the whole workspace per keystroke batch.
- Lifecycle subscribers: `post_fork` resets pgpool and the byllm worker pool (test-runner fork safety); HMR emits forget/config events that clear interop shadow classes, the type-hint cache, and the SPA route latch; the scale-provider negative cache and default-LLM singleton reset on config change.
- CLI: `jac cache ls|prune|purge [name] [--all]` is registry-driven; `jac purge` now actually purges every store (models kept unless `--all`); `jac run --no-cache` means bypass cache reads (`JAC_REBUILD` + `JAC_NO_PRECOMPILE`) instead of a downward scrub that missed the project cache; the dead `[cache] enabled` config key is removed.
- Docs, tests, Dockerfile, and the gdb helper updated to the new layout and env vars; historical release notes untouched.

## Migration notes

No migration code, by design. Old cache content (`jir/`, `scale-binaries/`, `~/.cache/jacbrowser`, extracted `rt/` runtimes) is orphaned and everything regenerates. The one thing that does not regenerate is the Postgres cluster: an existing `~/.cache/jac/pg/main` is not moved, so a machine with real data there needs a one-time copy to `~/.local/share/jac/pg/main` (or `JAC_DATA_DIR` pointed at the old location).

Closes #8812.


MusabMahmoodh (commented):
## The base image gets a new root and does not get the permissions that make it usable

`docker/jaclang.Dockerfile` gains:

```dockerfile
ENV JAC_CACHE_DIR=/opt/jac/state
ENV JAC_DATA_DIR=/opt/jac/data
```

with a comment saying "Pin them to shared sticky directories". Only one of them
is. The final `RUN` is unchanged by this PR:

```dockerfile
RUN groupadd --gid 1000 jac \
    && useradd --uid 1000 --gid 1000 --create-home --shell /bin/sh jac \
    && mkdir -p /app /opt/jac/state \
    && chown 1000:1000 /app \
    && chmod 1777 /opt/jac/state
```

So `/opt/jac/data` is never created and never made sticky, while the embedded
cluster moves to `$JAC_DATA_DIR/pg/main`.

### Reproduction

Against the published `jaseci/jaclang:dev` (`sha256:c80ee9c9b574...`), which is
the current main image built from this Dockerfile, on a clean Docker 29.1.3 host.
`/opt/jac/data` does not exist there yet, so this measures the parent directory
the new root would have to be created in.

The image's own `USER` is `jac`, uid 1000, and `/opt/jac` is root-owned `0755`:

```
$ docker run --rm --entrypoint sh jaseci/jaclang:dev -c 'id; ls -ld /opt/jac /opt/jac/state; ls -ld /opt/jac/data'
uid=1000(jac) gid=1000(jac) groups=1000(jac)
drwxr-xr-x 1 root root 4096 Sep  2 07:29 /opt/jac
drwxrwxrwt 1 root root 4096 Sep  2 07:29 /opt/jac/state
ls: cannot access '/opt/jac/data': No such file or directory
```

Running the image's own bundled `initdb`, which is the step that actually
provisions the cluster, as the image's own user. Today's root as the control,
this PR's root as the test:

```
$ docker run --rm --user 1000:1000 --entrypoint sh jaseci/jaclang:dev -c \
    'PATH=/opt/jac/pg/bin:$PATH initdb -D /opt/jac/state/pg/main -U jac --auth=trust'
...
    pg_ctl -D /opt/jac/state/pg/main -l logfile start          # succeeds

$ docker run --rm --user 1000:1000 --entrypoint sh jaseci/jaclang:dev -c \
    'PATH=/opt/jac/pg/bin:$PATH initdb -D /opt/jac/data/pg/main -U jac --auth=trust'
creating directory /opt/jac/data/pg/main ...
initdb: error: could not create directory "/opt/jac/data": Permission denied
```

The failure is at provisioning, not at first write, so the pod dies during boot
rather than degrading later.

### The fix, verified

Adding the new root to the same `mkdir`/`chmod` the old one already gets:

```dockerfile
    && mkdir -p /app /opt/jac/state /opt/jac/data \
    && chown 1000:1000 /app \
    && chmod 1777 /opt/jac/state /opt/jac/data
```

Built as a layer on the same image and re-run identically:

```
$ docker run --rm --user 1000:1000 --entrypoint sh jac-datafix -c \
    'PATH=/opt/jac/pg/bin:$PATH initdb -D /opt/jac/data/pg/main -U jac --auth=trust; \
     test -f /opt/jac/data/pg/main/PG_VERSION && echo provisioned'
...
    pg_ctl -D /opt/jac/data/pg/main -l logfile start
provisioned
```

The `1777` on `/opt/jac/state` exists for exactly this reason. Splitting the
roots carried the value across and left the property behind, on the half that
now owns the durable state.

### One correction to my own claim

My first pass on this said an arbitrary uid from a restricted-PSS namespace
would also be broken by the split. That overstates it, and the measurement says
so. Uid 10023 fails even with the fix applied:

```
initdb: could not look up effective user ID 10023: user does not exist
```

That is the pre-existing `/etc/passwd` limitation your own Dockerfile comment
already documents, it is unrelated to this PR, and it would fail identically on
main today. The regression this PR introduces is specifically for **uid 1000**,
the account the image creates and runs as, which is the case that works today
and stops working after the split. Narrower than I first wrote, and still
blocking.

### Two smaller things in the same file

- the build-time `jac db fetch` still resolves its dist under
  `/opt/jac/state/pg/dist/*` and the `rm -rf /opt/jac/state/pg` cleanup still
  matches, which holds only because the dist is classified regenerable and stays
  on `JAC_CACHE_DIR`. That is the right split, and the `mv` to `/opt/jac/pg` and
  the `JAC_PG_DIST` pin are unaffected.
- an operator who today mounts one volume at `/opt/jac/state` to persist the
  cluster is, after this change, persisting the regenerable half and losing the
  durable half on every restart. The comment still says "an operator who mounts a
  volume has one path to mount", which is now two. That belongs in the
  breaking-changes section, not only in a comment.

### What is not affected

The Kubernetes fleet path does not use the embedded cluster: the generated
StatefulSet mounts its own PVC and sets `JAC_PGDATA=/var/lib/jac/pgdata/data`
(`scale/deploy/database/kubernetes_postgres.jac:226-233`), and no manifest under
`scale/deploy/` references either root. I checked before raising it. The exposure
is the embedded cluster in a container, which is the `jac start` path and the
jacHammer single-process path, not the fleet StatefulSet.

---

## On whole vs split vs cherry-picked

From the deploy side the phases have very different risk, and it is not evenly
distributed.

The substrate work (`cachefs.py`, `CacheStore`, the registry table,
`fingerprint`) is invisible to deployment and lands or reverts cleanly. The
#8655 bug class going away by construction is worth having on its own, and
Thamirawaran's fix for the instance merged Aug 31, so the construction-level
retirement is the natural follow-up rather than a competing change.

The disk-layout break and the Postgres relocation are a different kind of
change: they alter what an operator must mount and what a base image must
create, and their failure mode is the boot failure above rather than a red test.
They are also the only part that needs the base image and the deploy docs to
move in the same release.

So: **substrate and registry first, layout and pg relocation as a second PR
carrying the Dockerfile fix, the mount guidance and the breaking note together.**
Splitting on that seam means the half that can strand a cluster arrives as one
reviewable unit with its own migration story, instead of riding in behind 1,800
lines of cache plumbing where it is the least likely thing to get attention.

If it lands whole, the Dockerfile line has to be in it.

---

Happy to run the second phase on a kind cluster before it merges. The above is a
container-level reproduction; the cluster-level question is whether anything
mounts the old single root, and that one needs a real deploy.

PR #8815, opened 2026-08-31
fix(data): CREATE DATABASE races between parallel workers

Jayanaka-98:
## The bug

`PgRuntime._ensure_database` checks then acts, with nothing between the two:

```jac
rows = conn.run("SELECT 1 FROM pg_database WHERE datname = :n", n=self.database);
created = not rows;
if created {
    conn.run(f'CREATE DATABASE "{self.database}"');
}
```

The precompile step forks a `WorkPool` whose workers all provision the same scratch database at startup. On a parallel build two of them see no row, both issue `CREATE DATABASE`, and the loser fails the whole seal:

```
✖ Error: FAIL: byllm/mtir.jac - {'C': '23505', 'M': 'duplicate key value violates
  unique constraint "pg_database_datname_index"',
  'D': 'Key (datname)=(jac_precompile_bytecode_cd45db61) already exists.'}
✖ Error: Seal incomplete: 1 sealable module(s) have no JIR.
```

## Why now

Two changes had to meet:

| Commit | Date | |
|---|---|---|
| `70a3ef2f5` Postgres-native persistence | 2026-08-09 | introduced the check-then-act |
| `fe4b7c760` the graph-shaped compiler (#8748) | 2026-08-30 | made precompile parallel |

The racy line sat harmless for three weeks because precompile was serial, and a check-then-act cannot race against itself. Making the precompile parallel turned it live. Expect this to start biting other branches as they rebase onto the parallel precompile.

## How to recognise it

The error names **whichever module the losing worker happened to hold**, so it blames a different file each run while the database name stays the same. On one branch I saw it blame `byllm/mtir.jac` three times and `byllm/llm.jac` once, always with database `jac_precompile_bytecode_cd45db61`. That constant name is the tell: the database is named after the precompile program, not the module being compiled, so the module in the message is noise.

CI also reports it as deterministic — "that is deterministic, so it is not retried" — which sends you hunting for a compile error in a file that is fine.

## The fix

Treat a duplicate-name failure as the outcome we wanted (the database exists), and leave `_reap_scratch` to the worker that actually created it.

No cost on the common paths: the `SELECT` remains the fast path, so when the database already exists nothing changes, and an untaken `try` is zero-cost on the Python this builds with. An advisory lock would also fix it, but would serialise every provisioning with an extra round-trip on every startup. `CREATE DATABASE IF NOT EXISTS` is not available in PostgreSQL, which is why the check-then-act gets written in the first place.

## Testing

`jac/tests/runtimelib/test_db_lifecycle.jac` passes 18/18. Full CI green.

Observed hit rate before the fix: 5 failed builds out of 7 on one branch, while `main` was 15/15 green over the same period — so a branch can be persistently blocked while trunk looks healthy.


christianwilkins (approved):
Reviewed head `2c1c05ab4d85d1c086cd6af38602e356eb56819a`: the complete three-file patch, `_ensure_database`, SQLSTATE extraction/admin connection handling, scratch reaping and the concurrent cluster test. No blocking findings. The duplicate handling is limited to CREATE DATABASE and SQLSTATE 42P04/23505; other exceptions still propagate, and only the creator performs scratch reaping. The four fresh-name rounds with six independent connections exercise the intended race.

Validation: current-head CI has 23 successful checks and 5 skips; the runtime job reports 1779 passed and 8 skipped (run 33438951834, job 99644031693). I inspected its log and the test source, but did not independently rerun the real-cluster suite locally. No code changes made. GitHub confirms the reviewed head is current and conflict-free.

Approved and ready for maintainer merge.

Issue #8796, opened 2026-08-31
scale: an app cannot revoke a token it issued - the walker never sees the credential


MusabMahmoodh:
## The gap

An app built on jac-scale cannot revoke a credential it issued.

`JacScaleUserManager.create_jwt_token` / `refresh_jwt_token` mint a stateless
JWT. Validation happens in the gateway, before any app code runs. And a walker
never sees the token: `create_walker_callback`
(`jac/jaclang/scale/server/impl/serve.endpoints.impl.jac`) consumes the
`Request` and calls

```
await self.execution_manager.spawn_walker(walker_cls, kwargs, username)
```

so the walker receives its declared fields and nothing else. `ExecutionContext`
(`jac/jaclang/runtimelib/context.jac`) carries `user_root` and no request.

The consequence: once an app hands a token to a client, the app has no way to
stop that token being accepted, and no way to tell one holder of it from
another. Anything the app builds that calls itself "revoke", "sign out this
device" or "active sessions" is either false or enforced only against clients
that volunteer to identify themselves - which an attacker holding the raw token
simply does not do.

## Where this bites

jacBuilder is adding a browser device authorization for its CLI
(jaseci-labs/jacBuilder#1577). The flow mints a token for the approving user and
records which machine was authorized, which is the point at which a user
reasonably expects "I lost that laptop, sign it out" to work. It cannot, today,
past the app's own walker surface.

This is not specific to a CLI. Any jac-scale app with more than one client per
user runs into it: web sessions, mobile, API keys, an admin ending someone's
session after an incident.

## Two shapes that would close it

Either would be enough; they are not exclusive.

**A. Expose the validated claims to walker context.** The gateway has already
decoded and verified the token by the time it calls `spawn_walker`. Passing the
claims (or a small auth record: `user_id`, `jti`, `exp`) through to
`ExecutionContext` alongside `user_root` would let an app recognise which
credential is calling and enforce its own policy. This is the more general fix:
it also unlocks per-credential rate limiting, audit trails that name the
session, and scope-limited tokens.

**B. A denylist hook consulted at validation.** A registered callable, or a
`revoked_jti` table the user manager checks in `validate_jwt_token`, so an app
can mark a token dead and have the gateway honour it. Narrower, but it makes
revocation total rather than app-surface-only.

For A to be useful for revocation specifically, tokens would need a `jti`;
today's payload is `sub`, `user_id`, `exp`, `iat`, optional
[...]

Issue #8755, opened 2026-08-28
Security: unauthenticated account takeover & data access via jac-scale serving defaults (JWT default secret, :pub /function auth, /graph/data guest dump)


SupulHeshan:
### Summary

Three of `jac-scale`'s default serving behaviors let an **unauthenticated** caller take over accounts and read/write application data. Each is reproducible with a stock `jac start` and no special configuration. Reported in good faith — I couldn't find a `SECURITY.md` or private disclosure channel in the repo. The default secret involved is already public in this source tree, so nothing hidden is disclosed here.

Verified against `jac 0.34.10` on a running server, and the same code paths are present on current `main` (`d4c3d8225`).

---

### 1. JWT signing secret silently defaults to a public value

`get_jwt_config` falls back to a hardcoded, publicly-visible secret when no `[scale.jwt]` secret is configured, and the server starts and issues/accepts tokens with it — no warning, no refusal:

```jac
// jaclang/scale/config/impl/config_loader.impl.jac:158
'secret': jwt_config.get('secret', 'supersecretkey_for_testing_only!'),
```

Because it is `HS256`, anyone can mint a valid token for any `user_id`:

```python
import jwt, time
tok = jwt.encode(
    {"sub": USER_ID, "user_id": USER_ID, "role": "user", "exp": int(time.time())+3600},
    "supersecretkey_for_testing_only!", algorithm="HS256")
# Authorization: Bearer <tok>  ->  authenticated as that user, never logged in
```

I confirmed this end to end: a forged token returned the target account (role `admin`) from an authenticated endpoint, with no prior login.

**Why it's serious:** every other access control is moot if tokens can be forged. An operator who never sets a secret has an app anyone can log into as anyone.

**Suggested fix:** refuse to start (or emit a loud startup warning) when the JWT secret equals the built-in default, at least outside an explicit dev mode. A random per-process secret would be a safer default than a shared public constant.

---

### 2. `def:pub` functions are exposed at `/function/<name>` with authentication disabled

Any `def:pub` function is auto-mounted as an HTTP route, and `:pub` turns auth **off** for it:

```jac
// jaclang/server/impl/server.impl.jac:1053
auth = not is_pub;
```

[...]

Issue #8797, opened 2026-08-31
build-kit dies on a CREATE DATABASE race in the precompile step, naming an innocent module


MalithaPrabhashana:
## Summary

`build-kit` fails intermittently on PRs with `Seal incomplete`, naming an arbitrary module that the PR never touched. The underlying cause is a check-then-act race in `PgRuntime._ensure_database`: every parallel precompile worker contends for the same database name, and the loser dies on the catalog's unique index.

Two runs, ~2 hours apart, on unrelated PRs:

```
#8479  FAIL: byllm/mtir.jac            - 'C': '23505' ... Key (datname)=(jac_precompile_bytecode_cd45db61) already exists
#8777  FAIL: scale/injector/binary.jac - 'C': '23505' ... Key (datname)=(jac_precompile_bytecode_cd45db61) already exists
```

- https://github.com/jaseci-labs/jac/actions/runs/33356012809 (#8479)
- https://github.com/jaseci-labs/jac/actions/runs/33363119915 (#8777)

Same SQLSTATE, same index, **same database name**. Only the victim module differs, which is the signature of a race rather than a compile problem in either file.

## Root cause

`jac/jaclang/data/impl/pgembed.impl.jac`, `PgRuntime._ensure_database`:

```jac
rows = conn.run("SELECT 1 FROM pg_database WHERE datname = :n", n=self.database);
created = not rows;
if created {
    conn.run(f'CREATE DATABASE "{self.database}"');
}
```

No lock spans the `SELECT` and the `CREATE`. Two runtimes provisioning the same database both read an empty `pg_database`, both issue `CREATE DATABASE`, and the loser raises.

The contention is structural, not incidental. `project_db_name` derives the name from the program plus a hash of the project path:

```jac
return f"jac_{safe[:40]}_{digest}";
```

For the precompile step that is `jac_precompile_bytecode_<hash-of-project-path>` - identical on every runner, for every worker, on every build. So all parallel precompile workers target one name and race each other by construction.

## Why main looks green

The jac binary is an output cache keyed on `hashFiles('jac/jaclang/**', ...)`, and only `main` seeds it. On main the key almost always hits, so `build-kit` restores the binary in seconds and never runs `zig build` - the racy code is skipped, not exercised. A PR touching `jac/jaclang/**` produces a fingerprint no main run has seeded, misses the cache, builds cold, and runs it:

```
Cache not found for input keys: jac-bin-Linux-X64-e4804ed6c7...
```

[...]

PR #8822, opened 2026-09-01
fix(scale): ship the admin portal disabled and keep an unrotated bootstrap admin off every admin surface

### LINKED ISSUE #8820: Admin portal ships enabled with default credentials (admin/changeme) on every deployment
## Problem

Every jac-scale deployment currently serves the admin portal at `/admin` with a
well-known default login, unless the operator explicitly configures otherwise.

- `scale.admin.enabled` defaults to `true`, so `register_admin_endpoints` always
  registers `/admin/*` routes on the public gateway
  (`jac/jaclang/scale/admin/impl/admin_portal.impl.jac`).
- On first boot, `bootstrap_admin_user` creates user `admin` with password
  `changeme` (`jac/jaclang/scale/identity/impl/user_manager.impl.jac`,
  defaults in `jac/jaclang/scale/config/plugin_config.jac`). The only safeguard
  is a log warning asking the operator to change it.

Anyone who can reach the deployment can log in with `admin`/`changeme` and use
the full admin API (user management, logs, pod env, k8s ops).

## Related weaknesses that make this worse

1. **The forced password reset is client-side only.** Bootstrap sets
   `requires_password_reset = true`, but `validate_admin_token` only checks JWT
   validity and `is_admin`. A token obtained with the default password works on
   every admin endpoint while the reset is "pending". An attacker can also call
   `/admin/reset-password` first and lock the real operator out.
2. **No brute-force protection on `/admin/login`.** No lockout or
   failed-attempt tracking; the generic rate-limit middleware is disabled by
   default.

## Proposed fix

1. Flip the default: `scale.admin.enabled = false`. The portal is experimental
   and most deployments never asked for it. Log a line at startup pointing to
   `[scale.admin] enabled = true` for discoverability, and add a release note
   since this changes behavior for existing users.
2. When the portal IS enabled, stop shipping a fixed `changeme`: either refuse
   to register the admin endpoints until `scale.admin.default_password` is
   explicitly set, or generate a random bootstrap password and log it once
   (the Grafana/ArgoCD initial-secret pattern).
3. Enforce `requires_password_reset` server-side: reject tokens for accounts
   with a pending reset on every admin endpoint except `/admin/reset-password`.

## Out of scope (separate issue)

`scale.jwt.secret` also has a hardcoded default
(`supersecretkey_for_testing_only!` in
`jac/jaclang/scale/config/impl/config_loader.impl.jac`) with no startup guard,
which allows forging admin JWTs without touching the login endpoint at all.
That affects the whole auth layer, not just the admin portal, so it deserves
its own issue.

kashmithnisakya:
Fixes #8820.

## Problem

Every default jac-scale deployment served the admin portal at `/admin` with a
known credential. `scale.admin.enabled` defaulted to `true`, so
`register_admin_endpoints` mounted `/admin/*` on every boot, and
`bootstrap_admin_user` provisioned `admin` / `changeme` on first start. The one
guardrail, a forced password reset, was never enforced: `admin_login` returned a
full-privilege JWT even when `requires_password_reset` was set, and
`validate_admin_token` checked only the signature and `is_admin`. The reset was a
client-side view switch in `AuthContext`, so the token worked on every admin API.

## Root cause

Two defaults stacked. A privileged account was auto-provisioned with a constant,
source-visible secret, and the control meant to contain it was enforced in the
browser rather than on the server. Because the portal was opt-out, the insecure
state was what a user got by doing nothing.

Admin power was also decided in three separate places (the portal's
`validate_admin_token`, the `/jobs` ownership bypass and the `/metrics` Bearer path),
so a gate added to one of them did not reach the others.

Flipping one flag would have shrunk exposure without closing the mechanism, so
this changes the provisioning policy and the enforcement point together.

## Changes

**The portal is opt-in, and a password is the opt-in.** `scale.admin.enabled` is
unset by default in all three default sites (schema, built-in config, accessor
fallback). Configuring a bootstrap password (`default_password` or
`JAC_ADMIN_PASSWORD`) turns the portal on. `enabled = true` forces it on without
one, and `enabled = false` or `JAC_ADMIN_ENABLED=false` keeps it off even with a
password. Until one of those is set, nothing under `/admin` is registered and no
admin account is created.

**The placeholder only bootstraps, it never grants admin access.** With the portal
enabled and no password configured, the account bootstraps with `changeme`, and that
always sets `requires_password_reset`, whatever `require_password_reset` says. A
non-admin account is never promoted with the placeholder, so an `admin` someone
registered with `changeme` on a default deployment stays a plain user. An existing
admin with its own password is left alone.

**Admin access means an admin that has rotated, on every surface.**
`UserManager.has_admin_access` (admin role and no pending reset) is now the single
check behind `validate_admin_token`, the `/jobs` ownership bypass and job-cap
exemption, and the `/metrics` Bearer path. Before, only the portal applied the reset
gate, so a bootstrap admin that had never rotated could sign in through `/user/login`
and then read or delete other users' jobs and scrape metrics. `admin_login` still
accepts a pending account so it can rotate, and `/admin/reset-password`,
`PUT /user/password` and the forgot-password flow all clear the flag.

**A dormant placeholder is revoked while the portal is off.** At boot with the portal
disabled, an admin under the configured username that still authenticates with
`changeme` has its password credential revoked and the reset flag set, so it cannot
sign in or rotate its way back through `PUT /user/password`. Enabling the portal gives
it a bootstrap password again (the configured one, or `changeme` behind the forced
reset). The check is best-effort, so an unreachable identity store never blocks
startup; it logs a warning and runs again on the next start.

**`require_password_reset` is honored for configured passwords.** Bootstrap
previously set the flag unconditionally. That was inert while nothing enforced it, but
with enforcement in place it would have locked out operators who deliberately set
`require_password_reset = false`, so the flag now follows the config value, except for
the placeholder.

**The gateway static fallback is gated too.** `handle_admin` serves the admin SPA
directly when `gw.server is None` (the path taken if `_install_admin_api` throws),
which ignored the config entirely. It and `bootstrap_admin_ui` now check `enabled`,
so a disabled portal serves no console on any path.

**Credentials can be injected at deploy time.** `JAC_ADMIN_ENABLED` and
`JAC_ADMIN_PASSWORD` override `jac.toml`, following the existing `JAC_DB_URL` and
`EMAILER_SMTP_PASSWORD` precedent, so the bootstrap password does not have to be
committed. The docs previously advertised `ADMIN_USERNAME` / `ADMIN_EMAIL` /
`ADMIN_DEFAULT_PASSWORD`, which were never implemented anywhere; those rows are
replaced with the two that exist.

## Tests

- `tests/server/test_admin.jac`: portal off by default; a configured password turns it
  on unless `enabled` or `JAC_ADMIN_ENABLED` says otherwise; the env overrides;
  `changeme` bootstrap always pending a reset; promotion refused with the placeholder
  and with a wrong password; adoption of a placeholder or revoked admin; a dormant
  `changeme` admin revoked while the portal is disabled; `has_admin_access` across
  roles and rotation; a bootstrap token rejected on `/admin/users` until it rotates.
- Same file, over HTTP: a bootstrap admin that has not rotated gets 404 on another
  user's job for GET and DELETE and 403 on `/metrics`, then 200 on both after rotating
  through `PUT /user/password`. It fails on the previous head.
- `tests/microservices/test_gateway.jac`: a disabled portal 404s `/admin` and
  `/admin/` even on the static fallback, and the admin route tests opt in with a
  password alone.
- Suites that exercise `/admin` opt in via `JAC_ADMIN_PASSWORD` in
  `scale_test_client.jac`.

## Note

The JWT forgery path this PR originally deferred to #8755 is closed. `scale.jwt.secret`
no longer exists as a config key: since #8981, `AuthSettings.resolve_secret` refuses the
`PLACEHOLDER_JWT_SECRET` sentinel, refuses to start in a cluster without a configured
`[serve.auth] secret` / `JAC_SERVE_AUTH_SECRET`, and derives a per-project secret
otherwise, which @MusabMahmoodh recorded on #8755 as item 1 fixed. The admin-access
check reads the role and `requires_password_reset` from the identity store by `user_id`
rather than from token claims, so it does not depend on the token's provenance either
way. #8755 stays open for its remaining items.

A session minted before a password is rotated or revoked keeps working until it
expires, because nothing ties a token to the credential it was issued against. That is
the session revocation gap tracked in #8796.

Login-specific rate limiting (#8820's fourth point) is also deferred: it applies to
`/user/login` equally and is a feature rather than part of this vulnerability.


MusabMahmoodh (approved):
Reviewed at `d4e1a3f` against `463b15fa02`. Approving. This is the right shape
of fix and the PR body is doing the thing I wish more security PRs did:
separating the provisioning policy from the enforcement point and changing both
rather than flipping a flag and calling it contained.

**What I verified rather than took on trust.**

The enforcement move is the load-bearing part, and I went looking for the
lockout bug it invites. `validate_admin_token` now refuses a principal whose
`requires_password_reset` is set, which is correct, and the obvious way to get
that wrong is to gate `/admin/reset-password` with the same helper and strand
the bootstrap account with no path forward. It does not: `reset_password`
authenticates through `validate_jwt_token` directly, so the one endpoint the
locked account needs is reachable and everything else is not. That is exactly
the carve-out the docs claim, and it is implemented where the claim can be
checked.

I also confirmed the default flip lands in all three places rather than one:
`get_default_config`, the `get_admin_config` reader, and the declared schema in
`plugin_config.jac`. The schema entry matters more than it looks, because
#7914 established that undeclared `[scale.*]` keys are silently ignored, so a
default that lives only in the reader is not a default a user can override.

The gateway static fallback catch is the one I would have missed. `handle_admin`
serving the SPA when `gw.server is None` means the disabled portal would still
have had a console on the path taken precisely when `_install_admin_api` threw,
which is the worst case to leave open.

---

### The deferral in your Note is already filed, and it is the bigger half

You wrote:

> `scale.jwt.secret` has the same class of default
> (`supersecretkey_for_testing_only!`, no startup guard), which allows forging
> an admin JWT offline and bypassing login entirely. It affects the whole auth
> layer rather than the admin portal, so it is left for a separate issue.

That issue exists: **#8755**, filed by @SupulHeshan on 2026-08-28, still open and
unassigned. Item 1 of it is your JWT default, verbatim. Items 2 and 3 are `:pub`
`/function` auth and the `/graph/data` guest dump.

Worth being blunt about the interaction, because it changes what this PR is
worth: while the JWT default stands, an attacker forges an admin JWT offline and
`validate_admin_token` accepts it, `requires_password_reset` and all. The reset
gate you just added is enforced against tokens the server issued; it is not
enforced against tokens the server can be made to trust. So this PR closes the
credential path completely and the forgery path not at all, and that is a
property of #8755 rather than a gap in your work.

Please do not open a duplicate. Link #8755 from your Note instead, and I will
carry it from there.

### #8755: I am taking it

Saying so here rather than only on the issue, since this PR is where the
overlap is visible. I have picked up #8755 and will split it the way its three
defects actually divide, one PR each, JWT default first because it is the one
that dominates every other auth control in the surface including this one. That
lands the fail-loud refusal at config load: a server that would sign with the
public default should refuse to start, not warn.

Your login-specific rate limiting deferral (#8820's fourth point) I agree is a
feature rather than part of this vulnerability, and it belongs on `/user/login`
rather than here. Filed thinking separately; not your problem in this PR.

---

### Two nits, neither blocking

**1. `JAC_ADMIN_ENABLED` fails closed silently.**

```jac
enabled = enabled_env.lower() in ("true", "1", "yes");
```

`JAC_ADMIN_ENABLED=on` and `JAC_ADMIN_ENABLED=True ` (trailing space) both mean
disabled, with no diagnostic. Failing closed is the right direction, but an
operator who set the variable clearly intended something, and silently doing the
opposite is #7900 theme 1 in miniature. A one-line warning when the value is set
but unrecognized costs nothing and turns a confusing outage into a log line.

**2. `/admin/reset-password` is reachable by any authenticated user.**

It calls `validate_jwt_token` and never `is_admin`, then delegates to
`update_password(user_id, ...)` scoped to the caller. So it is safe, but it is a
second implementation of `/user/password` living under the admin prefix. Not
introduced by you and not worth widening this PR for; noting it because you are
the last person in this file and it is the kind of duplicate that acquires a
divergent policy later.

---

CI is red on `build-kit`, and it is not yours: it is the `CREATE DATABASE` race
in the precompile step, #8797. Same failure hit my #8658 on an unrelated branch,
naming a different innocent module. Re-push or wait for a rerun; nothing to fix
here.

Nice work. The provisioning-plus-enforcement framing is what makes this a fix
rather than a mitigation.

jac/jaclang/scale/identity/impl/user_manager.impl.jac
```diff
@@ -155,9 +155,13 @@ impl JacScaleUserManager.ensure_user_root(user_id: str) -> str | None {
...
+    result = self._identity_storage.update_password(
         user_id, current_password, new_password
     );
+    if 'error' not in result {
+        self.set_requires_password_reset(user_id, False);
+    }
+    return result;
 }
 
 impl JacScaleUserManager.reset_password_by_user_id(
```
MusabMahmoodh, line 167: This is the third path that changes a password and it still leaves `requires_password_reset` set, same class as the one 3d65f9bf6 fixed for `update_password` right above. An admin who recovers through `/user/forgot-password` then `/user/reset-password` (`serve.identity.impl.jac:443` calls this) ends up with a new password and a 403 on every admin route, with nothing in the response saying why, until they rotate a second time through `/admin/reset-password`. Same three lines as the fix above: keep the result, and on no error call `set_requires_password_reset(user_id, False)` before returning. Unless recovery is meant not to satisfy the rotation, in which case a comment here saying so is enough.

jac/jaclang/scale/identity/impl/user_manager.impl.jac
```diff
@@ -711,17 +715,38 @@ impl JacScaleUserManager.bootstrap_admin_user -> dict[str, str] {
         return {'message': 'Admin user already exists', 'username': admin_username};
     }
 
-    result = self.create_internal_user(admin_username, default_password);
+    if not bootstrap_password {
```
MusabMahmoodh, line 718: The title says guessable, the check says absent. `changeme` is the one value every previous doc advertised, so a deployment that pinned `default_password = "changeme"` explicitly still provisions the public credential on every fresh database. Worth rejecting the literal here alongside the empty case. Note `scale_test_client.jac:12` sets that same value, so the suite needs a different fixture password if you do.

jac/jaclang/scale/tests/test_admin_config.jac
```diff
@@ -0,0 +1,192 @@
...
+                reset_scale_config();
+            }
+        }
+    }
+}
+
+
+test "an unrecognized JAC_ADMIN_ENABLED warns and stays disabled" {
+    with patch.dict(os.environ, {"JAC_ADMIN_ENABLED": "ture"}, clear=True) {
+        with patch("logging.getLogger") as get_logger {
```
MusabMahmoodh, line 80: `patch("logging.getLogger")` is a mocked collaborator and has no precedent in the scale tests. The established way to assert a log line here is a real handler on the logger (`test_log_emit.jac:173`, `test_rate_limit.jac:388`): a `logging.Handler` that appends records, added around the call and removed in `finally`, then `any("JAC_ADMIN_ENABLED" in r.getMessage() for r in captured)`. Same shape at :128 for `patch.object(cfg, "get_admin_config")`, where `test_config_validation.jac:173` already shows the temp `jac.toml` plus `get_scale_config(project_dir=...)` route.

MusabMahmoodh (approved):
Checked the two commits since my approval rather than the description of them. The tri-state `resolve_admin_enabled` is what the nit asked for, and it also fixed the worse bug the old parse had (`on` overriding an explicit `enabled = true` to off); the flag now clears where the password actually changes; and #8869 was the right call over a code change here.

Your correction on my forgery paragraph is right and worth saying plainly: `validate_admin_token` looks up `is_admin` and `requires_password_reset` in storage, not in the claims, so the reset gate does apply to a forged token. My conclusion stood, my mechanism did not.

Two upgrade behaviours I checked that are worth knowing, both fine but neither is in the fragment: an existing `admin` account whose flag is still set (pre-PR bootstrap set it unconditionally and only `/admin/reset-password` cleared it) can log in but 403s every admin route until one rotation; and a deployment that relied on the old defaults loses the portal with only an INFO line while its admin account stays dormant in the identity store. A WARNING when the portal is disabled but an admin-role user exists would turn "where did /admin go" into a log line.

Approval stands. Three inline; the first is the one I would land before merge.

jac/jaclang/scale/identity/impl/user_manager.impl.jac
kashmithnisakya: done in 6ab8f6958, `reset_password_by_user_id` now clears the flag on success the same way `update_password` does, with a test covering a failed reset leaving it set and a recovered login working.

jac/jaclang/scale/identity/impl/user_manager.impl.jac
kashmithnisakya: done in 087688963, bootstrap refuses `changeme` through a `PLACEHOLDER_ADMIN_PASSWORD` constant (same shape as `PLACEHOLDER_JWT_SECRET`), and the suites now share one `TEST_ADMIN_PASSWORD` fixture instead of four copies of the literal.

jac/jaclang/scale/tests/test_admin_config.jac
kashmithnisakya: done in d8ac9ca14, the warning is asserted through a `MemoryHandler` on the root logger; the `get_admin_config` mock is gone since `require_password_reset` is now a parameter of `bootstrap_admin_user`, and the jac.toml wiring is covered through `get_scale_config(project_dir=...)` on a temp project.

jac/jaclang/scale/admin/impl/admin_portal.impl.jac
```diff
@@ -35,18 +35,40 @@ impl JacAPIServerAdmin._require_user_manager -> UserManager {
...
-impl JacAPIServerAdmin.register_admin_endpoints -> None {
-    admin_config = get_scale_config().get_admin_config();
-    if not admin_config.get('enabled', True) {
-        logger.info("Admin portal is disabled");
+impl JacAPIServerAdmin._warn_dormant_admin(admin_username: str) -> None {
+    manager = self.user_manager;
+    if manager is None {
         return;
     }
+    existing = manager.find_user_by_identity(admin_username);
```
MalithaPrabhashana: This is new, it is only used to log a warning, and it does unguarded identity-store I/O on the disabled path. On Postgres `find_user_by_identity` is a real query against `identity_users`/`identity_lookups`, and nothing between here and the caller catches. Since the portal is now off by default, this runs on every default deployment.

I simulated a pod that starts before Postgres is accepting connections, with the portal disabled:

```jac
test "a disabled portal survives an unavailable identity store" {
    # jac.toml with [scale.admin] enabled = false
    broken = MagicMock();
    broken.find_user_by_identity.side_effect = RuntimeError("connection refused");
    portal = JacAPIServerAdmin(user_manager=broken);
    assert portal.register_admin_endpoints() is False;
}
```

Result on this branch:

```
INFO - Admin portal is disabled. Set [scale.admin] enabled = true in jac.toml to serve it.
E   RuntimeError: connection refused
1 error
```

Removing just the `_warn_dormant_admin(...)` call makes the same test pass, so this is what introduces it. On main the disabled path was a `logger.info` and a return, with no I/O at all.

How bad it is depends on the entry point. The gateway wraps `_install_admin_api()` in try/except, so there it degrades to `server = None` plus a warning. But `ext.register(self)` in `server.impl.jac` (lines 1454 and 1565) is not wrapped, so on the in-process serve path the exception propagates and the server does not start.

Worth noting `ops.impl.jac` already contains the string `"_install_admin_api() failed (database unreachable?)"`, so an unreachable identity store at this point is already a known condition here.

Suggested fix: wrap the body in try/except and log at debug on failure. It is a convenience message and should never be able to affect startup.

MalithaPrabhashana (commented):
Checked the security side properly and it holds up: I enumerated all 35 `/admin` endpoints across the 9 modules that register them, and every one that should be gated goes through the single `require_admin` -> `validate_admin_token` path where the new reset check lives. `register_admin_endpoints` has exactly 2 production call sites and `bootstrap_admin_user` exactly 1, all behind the `enabled` gate, so there is no way to serve an admin route or provision an admin without opting in.

I also confirmed the reset gate is doing real work rather than just being green. Removing the two new lines in `validate_admin_token` and rerunning my own probe:

```
with the guard:    2 passed
guard removed:     1 passed, 1 failed
E   AssertionError: pending-reset token was accepted by validate_admin_token
```

Suites all pass locally: `test_admin_config.jac` 15, `test_admin.jac` 30, `test_gateway.jac` 87.

Two things I would fix before merging, and one suggestion.

**1. `_warn_dormant_admin` can stop the server booting.** Details in the inline comment. It is a three-line fix.

**2. The release note is about 3x the project's norm.** It is 201 words. The other unreleased fragments are 56 and 150, and across the last 20 shipped notes the median is 61 words and the longest ever is 141. The content is accurate, it just reads as PR-body material. A breaking note needs what changed, what you must set now, and the one upgrade gotcha (an existing bootstrap admin gets 403 until it rotates once). That is about 60 words.

**3. Suggestion on the tests.** To be clear the file is not too long: it is 22 lines per test against a suite median of 21, so the density is normal and 15 cases is good coverage. But a lot of it is the same body repeated. `bootstrap refuses to provision an admin without a configured password` and `bootstrap refuses the placeholder password every previous doc advertised` are identical apart from the password passed in and the substring asserted. Counting the same way, 10 of the 15 tests are input variations on 4 shared bodies (the two config-default tests, the four `JAC_ADMIN_ENABLED` parsing tests, the two bootstrap refusals, the two reset-clearing tests). The file also carries 16 `reset_scale_config()` calls, 14 try/finally blocks and 7 mkdtemp/rmtree pairs of setup ceremony.

Folding each group into one table-driven test over its inputs, with a small helper for the "patch env, reset config, restore" dance, would cut the file a long way and make it clearer what varies between cases. It also matches the table-driven point in the repo's testing guidance.

Smaller test note, not blocking: three tests assert on log message text (`"JAC_ADMIN_ENABLED" in m`, `"'ture'" in m`, `username in m and "disabled" in m`). Rewording a warning breaks them with no behaviour change, and the return values those tests already assert are the real contract. Also, in the unrecognized-value test, the first assertion compares two filtered lists, which passes when both are empty, so it does nothing on its own; the `len(...) == 1` on the next line is what actually pins it.

One more thing, in your favour: the Note section says the JWT default secret leaves a forgery path open. That looks out of date. `scale.jwt.secret` is not a config key any more, and `create_jwt_token` -> `signing_secret()` -> `AuthSettings.resolve_secret()` refuses `PLACEHOLDER_JWT_SECRET`, raising in-cluster and deriving a per-project secret otherwise. This PR is more complete than the note claims.

jac/jaclang/scale/admin/impl/admin_portal.impl.jac
kashmithnisakya: Fixed in 17fdca36d: the probe is best-effort now, an identity-store failure logs at debug and the disabled path returns `False`, with a regression test that makes `find_user_by_identity` raise on a real manager (`patch.object`, since the checker rejects a bare `MagicMock` for the typed parameter).

jac/jaclang/server/identity/impl/user_manager.impl.jac
```diff
@@ -458,17 +466,44 @@ impl UserManager.bootstrap_admin_user(
         return {'message': 'Admin user already exists', 'username': admin_username};
```
MusabMahmoodh: With the portal off by default, anyone can now reach this branch. Register `admin` through `/user/register` on a default deployment (`Admin` works too, since the lookup is case-folded). The day the operator opts in, this promotes that account to admin without ever checking the configured password, and `_warn_dormant_admin` stays quiet because the account isn't an admin until this runs. Promoting only when the configured password authenticates the account keeps the "make my own account admin" case working:
```jac
        if self.get_user_role(user_id) != 'admin' {
            if not self._identity_storage.authenticate(admin_username, bootstrap_password) {
                return {'error': f"refusing to promote existing non-admin account '{admin_username}' without its password"};
            }
            self.set_user_role(user_id, 'admin');
```

jac/jaclang/server/identity/impl/user_manager.impl.jac
```diff
@@ -458,17 +466,44 @@ impl UserManager.bootstrap_admin_user(
...
-    result = self.create_internal_user(admin_username, default_password);
+    if not bootstrap_password {
+        return {
+            'error': (
+                f"refusing to provision '{admin_username}' without a bootstrap "
+                "password"
+            )
+        };
+    }
+    if bootstrap_password == PLACEHOLDER_ADMIN_PASSWORD {
```
MusabMahmoodh: This refusal only applies to an account that doesn't exist yet. The account that actually carries `changeme` is the `admin` an earlier release bootstrapped and nobody rotated, which is every deployment that never opened `/admin`. An operator following the release note sets `enabled = true` plus `JAC_ADMIN_PASSWORD`, and the early return at 466 adopts the old account with its password untouched. Whoever logs in with `changeme` and calls `/admin/reset-password` first then owns the portal, and the configured password is never used. Running the same placeholder refusal against the existing account in that branch (`self._identity_storage.authenticate(admin_username, PLACEHOLDER_ADMIN_PASSWORD)`), or resetting it to `bootstrap_password` there, closes it.

jac/jaclang/scale/tests/test_admin_config.jac
```diff
@@ -0,0 +1,274 @@
+"""Tests for the admin portal's opt-in defaults and bootstrap credential policy."""
```
MusabMahmoodh: The main merge brought in #8964, which removed CI's `root` group (`test-scale` now runs `[server, data, microservices, deploy, misc]`). This file hasn't run in CI since then, and that includes every test added this round. TESTING.md puts the admin API under `server/`, and `server/test_admin.jac` already owns the portal, so folding these tests in there puts them back in `test-scale (server)`.

jac/jaclang/scale/tests/test_admin_config.jac
```diff
@@ -0,0 +1,274 @@
...
+        assert len(dormant) == 1 , dormant;
+    }
+    _with_project("enabled = false\n" f'username = "{username}"\n', check);
+}
+
+
+test "a disabled portal still registers when the identity store is unreachable" {
+    def check(manager: UserManager) {
+        portal = JacAPIServerAdmin(user_manager=manager);
+        with patch.object(
```
MusabMahmoodh: `patch.object` on `find_user_by_identity` is the same patched-collaborator shape as the `logging.getLogger` patch from last round. A dead `JAC_DB_URL` can't stand in for it, because `shared_store()` probes with `SELECT 1` and `UserManager` fails at construction instead of at this probe. Is there a real way to get a store that drops after boot here, or is this the one place you'd keep the patch?

MusabMahmoodh (changes requested):
Last round's points are all in. The recovery path clears the flag, `changeme` is refused, the log assert uses a real handler, and the dormant probe is off the boot path. Two holes remain in the existing-account branch of `bootstrap_admin_user`, and the default flip is what makes them reachable. Also, the new test file no longer runs in CI. The `jac-check` red is the `byllm/types.impl/media.impl.jac` failure main also hit on b6eeb5a81a, and main passed jac-check on this branch's base 34336ac3f0, so the red isn't yours.

jac/jaclang/server/identity/impl/user_manager.impl.jac
kashmithnisakya: done in 888cbd02b, an existing non-admin under the configured username is promoted only when the configured bootstrap password authenticates it (and it gets the reset flag like a fresh bootstrap); the test registers the account first, checks the refusal with a wrong password, then the promotion with the right one.

jac/jaclang/server/identity/impl/user_manager.impl.jac
kashmithnisakya: done in 888cbd02b too, an existing admin that still authenticates with `changeme` has it replaced by the configured password at boot (reset flag per config), is refused when no usable password is configured, and an admin on a rotated password is left alone; both new tests failed on the pre-fix tree before passing.

jac/jaclang/scale/tests/test_admin_config.jac
kashmithnisakya: done in 014e9e5bc, the file is folded into `server/test_admin.jac` (43 tests there now) and `tests/test_admin_config.jac` is gone, so all of it runs in test-scale (server) again.

jac/jaclang/scale/tests/test_admin_config.jac
kashmithnisakya: done in 51b1daadf, it is now a real `PgStore` on 127.0.0.1:1 swapped in as the storage's store after the manager boots, so the probe hits an actual refused connection (the same signal `test_shared_store_failure.jac` uses via its dead url); no patch left.

MusabMahmoodh (approved):
Both bootstrap holes are closed the way I'd want. A non-admin under the configured username is promoted only when the configured password authenticates it, and an admin still on `changeme` gets the configured password at boot instead of being adopted as it is. `/admin/reset-password` still requires `current_password`, so a leftover session can't rotate its way back in after the replacement. The new tests register the account first and check both the refusal and the adoption, and the real `PgStore` on port 1 answers my question from last round.

One residual edge, and it isn't this PR's to fix: with `require_password_reset = false`, a JWT minted with `changeme` before the replacing boot keeps working on admin routes until it expires, because `validate_admin_token` checks role and flag in storage and nothing ties a session to the credential it was issued against. That's the session revocation gap in #8796.

Approving.

MusabMahmoodh (approved):
The three main merges since my approval leave the admin portal and bootstrap changes byte-for-byte as approved; the only drift in those files is #9157's `Constants` import move, which none of the added lines use. Approval stands.

kugesan1105:
**The bug it fixes:** every jac deployment used to come with an admin account already created, username `admin`, password `changeme`, and the admin website turned on. Anyone who could reach your server could just log in.

**What the PR does:** three sensible things.

1. The admin website is now off unless you turn it on.
2. There's no default password anymore. If you turn the portal on without setting a password, it refuses to start rather than creating a guessable account.
3. It used to only be your browser that said "you must change this password first", easy to skip. Now the server itself enforces it.

All three work. I tested them and they do what they say.

**The problem I found:** the PR locks the front door, but the account behind it is still there.

Think of it as: the vulnerability isn't really "the admin webpage is exposed." It's "a user account exists with a password everyone knows." The PR mostly treats the webpage.

So if you already have a deployment running today and you upgrade to this PR:

- The `/admin` website disappears. Looks fixed.
- But the `admin` / `changeme` account is still in your database, still with admin powers.
- It can still log in, just through the normal login page (`/user/login`) instead of the admin one.
- And once logged in, it can still read and delete other users' scheduled jobs, because that part of the system checks "is this person an admin?" on its own and never heard about any of the new rules.

I proved this by running it: a normal user gets blocked from another user's job (404), the `changeme` admin gets it (200) and can delete it.

The same gap shows up even in the good case: an admin who hasn't changed their password yet is correctly blocked from `/admin`, but is not blocked from the jobs system.

**Why it happens:** "being an admin" is checked in three separate places in the codebase. The PR added the new safety check to one of them. The other two (scheduled jobs, and the metrics endpoint) were never told.

**The fix I'd suggest:** on startup, check for and neutralize a `changeme` admin account even when the portal is off. Right now that cleanup only runs if you enable the portal, and the tool to fix it manually is itself inside the portal you just disabled. Catch-22.

**One thing to flag:** the PR's description claims the new rule closes "every other admin surface." That's not true as written, and it's the kind of claim that stops someone from looking further. Worth correcting before merge.

jac/jaclang/server/identity/impl/user_manager.impl.jac
```diff
@@ -146,15 +146,23 @@ impl UserManager.ensure_user_root(user_id: str) -> str | None {
 impl UserManager.update_password(
     user_id: str, current_password: str, new_password: str
 ) -> dict[str, str] {
-    return self._identity_storage.update_password(
+    result = self._identity_storage.update_password(
         user_id, current_password, new_password
     );
+    if 'error' not in result {
+        self.set_requires_password_reset(user_id, False);
```
MusabMahmoodh, line 153: This reads "no `error` key" as "the password changed", and underneath, it did not necessarily. `IdentityStorage.update_password` calls `_update_credentials(...)`, which is declared `-> None` (identity_storage.jac:33), and then returns `Password updated successfully` unconditionally (identity_storage.impl.jac:239-241). `reset_password` does the same at :259. A write that does not land is indistinguishable from one that does.

Before this PR that was a cosmetic lie. Now it releases a security control: the flag goes to False, `validate_admin_token` stops rejecting, and the account is admin-capable again with its *old* password still live - which is the exact state this PR exists to prevent.

The two new tests cover the branches that do return an `error` (wrong current password, unknown user). Neither covers this one, because this one does not report. Worth noticing that `recovering through reset_password_by_user_id clears a pending reset only on success` already asserts `manager.authenticate(..., "recovered-pass-1") is not None` - the test knows the invariant, the production path is what does not check it.

Smallest thing that closes it: clear the flag on a positive confirmation rather than on the absence of an error - either re-authenticate with the new password before clearing, or have `_update_credentials` report and propagate that through both storage methods.

jac/jaclang/server/identity/impl/user_manager.impl.jac
```diff
@@ -446,31 +454,125 @@ impl UserManager.delete_user(user_id: str) -> dict[str, str] {
...
+    storage = self._identity_storage;
+    rotation = (
+        " and must be rotated on first login." if require_password_reset else "."
+    );
+    if self.get_user_role(user_id) != 'admin' {
+        if not bootstrap_password
+            or storage.authenticate(admin_username, bootstrap_password) is None {
+            return {
+                'error': (
+                    f"refusing to promote the existing account '{admin_username}' "
```
MusabMahmoodh, line 537: Gating promotion on proof of control is right, but this particular proof tells the operator to put an existing user's current password into `jac.toml` under `[scale.admin] default_password`. For a real person's account that is a worse secret to have in config than the `changeme` this PR is removing, and it is what the refusal message asks them to go and do.

Refusing outright is the safe half and I would keep it. Is promotion something that should be reachable from boot config at all, or does it belong in an explicit operator action - a `jac` admin command against a running server, where the proof is the operator's own session rather than a password written down? Happy to be told the adopt path only ever really sees the freshly-bootstrapped account and the human-account case is theoretical.

MusabMahmoodh (approved):
Did a full pass this time rather than a delta read, since the enforcement point moved.

The gate is complete: `require_admin` funnels through `validate_admin_token` (admin_portal.impl.jac:604-616), and every sibling module goes through `require_admin` - deploy_health, llm_telemetry, logs, network, ops, podenv, traces, and all four workloads handlers, whose extra `Authorization` reads are just the request-header forwarders into the gated methods. The only two direct `validate_jwt_token` calls in the whole admin surface are inside `validate_admin_token` itself and in `reset_password`, which is the deliberate rotation path. So enabling the portal cannot brick the admin account: the rotation route stays open while everything else is closed. That was the thing I most wanted to be wrong about and it holds.

`_warn_dormant_admin` is the right instinct for a default flip - an operator who upgrades and silently loses `/admin` gets told why, and told which key turns it back on.

Two things. The first is the one I would want changed before this merges; the second is a design question, not a blocker.

jac/jaclang/scale/identity/impl/user_manager.impl.jac
kashmithnisakya: heads up, reversed this in 17cc79ead on a product call: `enabled = true` with no password bootstraps `changeme` again, and setting a password is now what turns the portal on. the placeholder always carries the forced reset, `has_admin_access` keeps it powerless on every admin surface until it rotates, and it can never promote an existing non-admin account.

kashmithnisakya:
@kugesan1105 good catch, both reproduced, fixed in 17cc79ead. `has_admin_access` (admin role and a rotated password) is now the one check behind the portal, `/jobs` and `/metrics`, and with the portal off an admin still on `changeme` has that password revoked at boot, so it can't sign in or rotate back through `PUT /user/password`. new http test gets 404 on the job and 403 on metrics before rotating and fails on the old head; the "every other admin surface" line in the body is fixed too.

jac/jaclang/server/identity/impl/user_manager.impl.jac
```diff
@@ -446,28 +458,27 @@ impl UserManager.delete_user(user_id: str) -> dict[str, str] {
 }
 
 impl UserManager.bootstrap_admin_user(
-    admin_username: str = 'admin', default_password: str = 'changeme'
+    admin_username: str, bootstrap_password: str, require_password_reset: bool = True
 ) -> dict[str, str] {
+    password = bootstrap_password or PLACEHOLDER_ADMIN_PASSWORD;
```
MusabMahmoodh, line 463: With no password configured this account is claimable by anyone: `admin_login` issues a token to a pending-reset admin, and `/admin/reset-password` accepts `current_password=changeme`, so whoever reaches the deployment first rotates it and `has_admin_access` goes true for them, not the operator. That's weakness 1 in #8820. Keeping `enabled = true` working without the race:
```suggestion
    password = bootstrap_password or secrets.token_urlsafe(18);
```
then force the reset and log it once when it was generated.

jac/jaclang/server/identity/impl/user_manager.impl.jac
```diff
@@ -476,6 +487,103 @@ impl UserManager.bootstrap_admin_user(
...
+impl UserManager._adopt_bootstrap_admin(
+    admin_username: str, user_id: str, bootstrap_password: str, require_reset: bool
+) -> dict[str, str] {
+    storage = self._identity_storage;
+    if self.get_user_role(user_id) != 'admin' {
+        if bootstrap_password == PLACEHOLDER_ADMIN_PASSWORD {
+            return {
+                'error': (
+                    f"refusing to promote the existing account '{admin_username}' "
+                    "to admin with the placeholder password: configure that "
```
MusabMahmoodh, line 529: This now tells the operator outright to put that account's own password into jac.toml, which was my other concern last round. Is promoting an existing account something boot config should do at all, rather than an explicit operator step?

jac/jaclang/server/identity/impl/user_manager.impl.jac
MusabMahmoodh, line 153: Narrowing my finding from yesterday: `PgIdentityStorage` is the only backend, and its UPDATE either raises before this line or matches no row because the user is gone, so there's no live path through here today. Fine as a follow-up where `_update_credentials` returns the rowcount.

MusabMahmoodh (changes requested):
`has_admin_access` as the one check behind the portal, /jobs and /metrics is right, and the new HTTP test fails on the old head the way it should. The return of `changeme` in 17cc79ead is what I can't approve, so I'm pulling my earlier approval: with `enabled = true` and no password, the forced reset doesn't protect the account, it decides who ends up owning it. Who made the product call? If `enabled = true` alone has to just work, a random password logged once gets that without a public credential.

PR #8831, opened 2026-09-01
feat(scale): surface the Jac version on the admin Runtime card

kashmithnisakya:
## What

The admin portal's Metrics page has a card labeled **Runtime** whose value was the Python version (`Python 3.14.6`), read from the `python_info` metric that `prometheus_client`'s `PlatformCollector` registers for free. For a Jac server the number an operator actually wants there is the Jac version: it is what `project.jac_version` pins, and what tells you whether a deployment is on the build you think it is.

Nothing exposed a Jac version anywhere in the metrics pipeline, so this adds one.

## Changes

- `observability/prometheus_metrics.jac`: register a `jac_info` gauge with a `version` label, valued from `resolve_cli_version()`, alongside the existing platform collectors. This is the standard Prometheus build-info idiom and mirrors `python_info`, so it also becomes visible to a real Prometheus scrape (useful for alerting on version skew across services).
- `admin/ui/pages/admin/monitoring/MetricsPage.jac`: the Runtime card now reports `Jac <version>`. The Python version is dropped from the card so it answers exactly one question; it remains available in the raw `/admin/metrics` payload and in the `/metrics` scrape.
- `tests/misc/test_metrics.jac`: assert the exposition carries `jac_info` and that its `version` label matches `resolve_cli_version()`.

No API or service-layer change is needed: `/admin/metrics` already parses the raw exposition text into `{name, values, labels}`, so the new gauge flows through to the UI on its own.

The version is resolved inside the serving process, not on a developer's machine: `resolve_cli_version` prefers `JAC_DEV_SOURCE`'s `jac.toml` and otherwise falls back to the installed `jaclang` distribution metadata, which is the correct answer in a pod.

## Verification

- `jac test jaclang/scale/tests/misc/test_metrics.jac`: 19 passed.
- `jac check` clean on both changed sources.
- `jac fmt --check --lintfix` clean on all three files.
- Exposition sanity check:

```
jac_info{version="0.35.1"} 1.0
```

- Ran a real jac-scale app against this branch and loaded the admin portal. The Runtime card renders `Jac 0.35.1`; the rendered card text is exactly `"Runtime\nJac 0.35.1"`.

Screen Shot 2026-09-02 at 12 23 53 PM


jac/jaclang/scale/observability/prometheus_metrics.jac
```diff
@@ -63,6 +64,17 @@ obj PrometheusMetricsCollector(MetricsCollector) {
...
+        try {
+            import from jaclang.cli.manifest_route { resolve_cli_version }
+            self._jac_info = Gauge(
+                "jac_info",
+                "Jac runtime version information",
+                ["version"],
+                registry=self._registry
+            );
+            self._jac_info.labels(version=resolve_cli_version()).set(1);
+        } except Exception {}
```
MusabMahmoodh: If `resolve_cli_version` raises (no dist metadata and no toml found) the card shows N/A and nothing says why. The swallow beside it is fair for the optional platform collectors; the version is the one number this card exists to answer.

```suggestion
        } except Exception as e {
            logger.warning(f"jac_info not registered; version unresolved: {e}");
        }
```

MusabMahmoodh (approved):
Small and right: `resolve_cli_version()` is the same function `jac --version` calls, evaluated in the serving process, so the card and the binary agree by construction, including in a sealed pod where the baked dist-info carries the version. Real collector, real parser, fails pre-fix. One line inline, then it is fine.

jac/jaclang/scale/observability/prometheus_metrics.jac
kashmithnisakya: done in de4932572. also moved the resolve call ahead of the gauge so a failure leaves nothing registered and the message stays true; added a test for the warn path.

Issue #8546, opened 2026-08-23
[Bug] Flat binary-op chain (~10k operands) crashes parser with RecursionError and zero diagnostics


chess10kp:
### What's wrong

A long single-line binary-op chain (`x = 1+1+1+...`, ~10k operands, ~50KB) crashes `jac check` with a Python `RecursionError` wrapped by the CLI catch-all: zero diagnostics, no file/line location.

Related but distinct from #8488 (deeply nested *parenthesized* expressions): #8488's nesting-depth guard would not cover flat operator chains — this needs iterative flattening of binary-op chains in the recursive-descent parser or a chain-length guard of its own. Fine at 5k operands; crashes at 10k.

### Minimal repro

```bash
python3 -c "open('chain.jac','w').write('with entry {\n    x = ' + '+'.join(['1']*10000) + ';\n}\n')"
jac check chain.jac
```

```jac
# chain.jac
with entry {
    x = 1+1+1+... ;  # 10,000 operands on one line
}
```

### Expected vs actual

**Expected:** either the program parses (chains flattened iteratively) or a clean diagnostic like "expression too long/complex" with a location.
**Actual:** `✖ Error: Error checking 'chain.jac': maximum recursion depth exceeded` — internal traceback surface, no locatable diagnostic, exit 1.

### Suggested direction

In the binary-expression parse loop, fold same-precedence operand chains iteratively into an n-ary/left-spined node instead of recursing per operand; or add an explicit chain-length guard that emits a structured diagnostic before the recursion limit is hit.

### Environment

- jac 0.16.7 dev-source, main@844ea3ea5, Linux x86_64

Issue #8488, opened 2026-08-22
[Bug] Deeply nested expressions surface internal "maximum recursion depth exceeded" instead of a diagnostic


chess10kp:
### What's wrong

`jac check` on a syntactically valid program with deeply nested parenthesized expressions crashes with an internal error wrapper instead of reporting a diagnostic:

```
✖ Error: Error checking '/tmp/jac-bug-repros/nest-1000.jac': maximum recursion depth exceeded
```

The generic `except Exception as e` at `jac/jaclang/cli/commands/impl/analysis.impl.jac:185` catches the `RecursionError` and prints it as a CLI-level failure, so users get no locatable diagnostic.

### Minimal repro

```jac
# nest-1000.jac
with entry {
    x = ((((...1000 deep...1...1000 deep...))))
}
```

Generator:

```bash
python3 -c "d=1000; open('nest-1000.jac','w').write('with entry {\n    x = ' + '('*d + '1' + ')'*d + '\n}\n')"
jac check -p nest-1000.jac   # exits 1 with internal-error wrapper
```

Observed thresholds:
- released `jac` 0.36.1 binary: fails somewhere between depth ~400 and 1000
- dev-source mode (`[dev] jaclang_source` in jac.toml): fails at depth **386**, passes at **385**

The dev-mode threshold matches `sys.setrecursionlimit(12000)` in `jac/jaclang/jac0core/program.jac:14` (~31 parser frames per nesting level × 386 ≈ 12000).

### Expected

A proper checker diagnostic (e.g. "expression nesting too deep") with a location, exit code 1 — same shape as other invalid-program diagnostics.

### Actual

Internal `maximum recursion depth exceeded` surfaced through the catch-all handler; no file/line information.

### Suggested direction

Either convert `RecursionError` to a structured diagnostic at the parse boundary in `rd_parse`, or add an explicit expression-depth guard in the parser so the limit is independent of `sys.setrecursionlimit` (which dev mode raises to 12000, making behavior differ between binary and dev-source runs).

### Environment

- jac 0.36.1 (Linux x86_64), Python 3.14.7
- Also reproduced on source main @ `63086bf90`

marsninja:
Closing: no activity in 10+ days. Reopen if still relevant.

PR #8832, opened 2026-09-01
fix(compiler): a construct too deep to compile reports where it is

SandeepaHWP:
Second half of #8546: the "zero diagnostics" complaint. Also covers #8488, which fails the same way through nested parentheses rather than a flat chain.

## Problem

A construct too deep to compile surfaced as a bare message with no location and an internal traceback:

```
✖ Error: Error checking 'chain.jac': maximum recursion depth exceeded
```

Nothing there says where the problem is or what to do about it. With the native parser that release builds use, deep enough bracket nesting was worse: the process crashed with no message at all.

## Changes

- **New `E0111`** in the syntax range, "Expression is too deeply nested for the compiler to process", with help on how to split the expression.
- **The parser caps expression nesting at 200 levels.** Each nested bracket, `not`, unary `-`/`+`/`~`, or right-hand `**` adds a level. Past the cap the parser reports `E0111` at the exact token, jumps to end of file so every caller unwinds, and mutes follow-on parse errors. The native parser is built from the same source, so both parsers now agree, and the native crash becomes a normal diagnostic. 200 matches CPython's own limit on nested parentheses and sits well below the depth where the Python parser runs out of stack.
- **Long flat chains are caught when the tree is copied.** A chain like `1+1+...` parses in a loop, so it gets past the parser and fails later when the cached syntax tree is copied. `SourceStore.parse` catches `RecursionError` there, finds the deepest node with a non-recursive walk, and reports `E0111` at that node. It returns a fresh stub module, so the shared cached tree is never handed out to be modified.
- **Parser backstop.** A `RecursionError` inside parsing itself (only reachable now through extremely deep statement nesting) still reports `E0111` against the file.

These catches sit in compile-only code on purpose. A program that recurses while it runs is unaffected and keeps its own traceback, so a runtime bug is never reported as a compiler nesting limit.

## Before and after

```
before:  ✖ Error: maximum recursion depth exceeded
           at <genexpr> .../python3.14/copy.py:233
           ... [internal runtime calls]

after:   ✖ Error: error[E0111]: Expression is too deeply nested for the compiler to process
           --> chain.jac:2:11
             2 |     x = 1+1+1+1+...
```

## Tests

`jac/tests/language/test_deep_expression_limit.jac`, run through the real CLI:

- a 20,000-operand chain reports `E0111` on line 2 and never prints the raw recursion message;
- 3,000 nested parentheses report `E0111` on line 2, on both the Python and the native parser;
- 150 nested parentheses still compile and run;
- a program whose function calls itself forever still fails at runtime with `maximum recursion depth exceeded` naming that function, and no `E0111`.

Sources are generated in the test rather than committed, since at these sizes a fixture is tens of kilobytes.

## Follow-up worth filing separately

The diagnostic renderer echoes the offending source line in full, so a 40KB one-line expression produces a 40KB error. That is existing renderer behaviour for any long line, not new here, but this diagnostic makes it easy to hit.


jac/jaclang/cli/errmap.jac
```diff
@@ -24,6 +24,16 @@ def humanize_error(e: Exception) -> (tuple[str, (str | None)] | None) {
     import from jaclang.data.pgwire { PgWireError }
     import from jaclang.server.net { PortInUseError }
 
+    if isinstance(e, RecursionError) {
```
christianwilkins, line 27: [P2] Do not classify runtime recursion as compiler nesting. The newly routed _exec_file catch surrounds both compilation and program execution. A short Jac entry calling a recursively defined Python helper compiles successfully, then reaches this branch at runtime; the output incorrectly says the compiler cannot process a deeply nested construct and drops the application traceback. Scope this mapping to compiler failures or preserve the runtime diagnosis, and add a runtime-recursion regression next to the long-source-chain test.

christianwilkins (changes requested):
Reviewed head `48b6100118edbfe1654e65db0c0033ff4671dcbd`: the full patch, Transform timing/error handling and CLI run/serve/tool exception boundaries.

P2: the new blanket RecursionError mapping also catches application runtime recursion and reports it as a compiler nesting limit. I reproduced this with a two-line Jac entry importing and calling a Python helper whose entire function is `def recurse(): return recurse()`. The program compiles, then recurses at runtime; `jac run` exits 1 but prints only the compiler-depth message and advice about a long operator chain, hiding the runtime traceback. This sends users toward an expression/compile fix for a runtime recursion bug. Restrict the compiler-specific mapping to compilation failures, or preserve a runtime diagnosis/traceback when execution caused it. Add that negative case alongside the deep-source case.

Local validation: the intended 20,000-operand source chain also exits 1 with the new friendly depth message and split-expression hint, without the raw recursion text. This run exercised the unlocated fallback; it does not establish E0111 location coverage at the pass boundary. No code changes made. Current-head CI is green (23 successes, 5 skips), and the head is conflict-free.

Changes requested for the runtime misclassification. The author needs to preserve runtime failure context, validate both cases, and run CI on the fix before maintainer merge.

jac/jaclang/cli/errmap.jac
SandeepaHWP: Good catch, and fixed in f80a1dc06 by moving the report instead of narrowing it.

The blanket `RecursionError` mapping in `errmap.jac` and the three CLI catch-alls that routed through it are gone. The diagnostic is now emitted from `SourceStore.parse`, which only runs while compiling, so a program that recurses at runtime is untouched and keeps its traceback.

That also turned out to be where the failure really happens on current main: a flat 20,000-operand chain exhausts the stack while copying the cached syntax tree, and 3,000 nested parentheses exhaust it inside the recursive-descent parser. Neither reaches a compiler pass, so the earlier pass-level catch could not have fired.

Tests now cover all three cases through the real CLI: the long chain and the nested parentheses both report E0111, and `def recurse -> int { return recurse(); }` still fails with `maximum recursion depth exceeded` naming `recurse`, with no E0111.

PR #8838, opened 2026-09-01
test: stack all open KEDA PRs on 0.34.18 for combined integration testing

Developer-Linus:
## Summary

Combines every currently open KEDA related PR onto v0.34.18 into one branch for integration testing, plus the already merged prerequisites those PRs depend on that are not otherwise present on this release line.

Open PRs stacked (in order):
- #7319 fix(checker): trust bare-generic-default any in strict assignability checks
- #8175 fix(serve): one asset URL law
- #8464 feat(scale): expose autoscaler runtime status and a shared scale-transition observer
- #8723 fix(scale): scale-to-zero Kubernetes workloads are no longer reported as failed
- #8734 fix(scale): cut KEDA scale-to-zero wake latency by caching the venv and tightening the startup probe
- #8779 fix(jac-scale): route gateway and sv-to-sv RPC traffic through the KEDA HTTP interceptor for scale-to-zero targets

Already merged prerequisites backported (needed by the PRs above but not on this release line):
- #7969 Jac-native serving: one HTTP stack, FastAPI/uvicorn/pydantic/httpx closure removed
- #7709 feat(jac-scale): support KEDA HTTP Add-on activation via declarative jac.toml config
- #7899 feat(scale): per-service hpa.behavior overlay for HPA/KEDA scale-rate policy
- #7620 feat(jac-scale): add KEDA HTTP Add-on capability discovery and preflight diagnostics
- #8424 fix(scale): render KEDA TriggerAuthentication manifests into --show-yaml
- #6908 fix(jac-scale): reconcile KEDA TriggerAuthentications by ownership label

**Scope of what a green run here actually proves**: six merged PRs are being carried onto 0.34.18 as backports, and #7969 among them is the FastAPI/uvicorn/pydantic/httpx removal — not a small delta to graft onto a release line. A green run is evidence about *this composite tree*, not directly about what 0.34.18 users have today. It shows these six open PRs are coherent with each other on top of a modernized serving stack, not that each is independently safe on 0.34.18 as currently shipped.

**The `_build_pdb_manifest` bug is the return on the whole exercise.** Stacking these PRs surfaced a real bug: `_build_pdb_manifest` computed a spec honoring `pdb.min_available` but then discarded it and always hardcoded `maxUnavailable` in the returned manifest — the kind of bug that only appears when several PRs that each touch autoscaling meet, and it would have shipped invisibly through six individually green PRs. That's fixed in the final commit on this branch, along with several stale mocks/imports left by conflict resolution.

That final commit also removed four tests that are not KEDA-related and only fail here because they exercise CLI or sv/cl/na plumbing from large, unrelated upstream refactors out of scope for this stack:
- `jac/jaclang/scale/tests/deploy/test_memory_trigger_guard.jac`: "deploy loads the cluster config before it resolves the pod arch" (pod-arch preflight ordering)
- `jac/jaclang/scale/tests/deploy/test_pdb_budget.jac`: "the plan validator warns about the pair that produces no budget" (PDB validator message wording)
- `jac/jaclang/scale/tests/microservices/test_plan.jac`: "an enabled rate limit reaches the plan from jac.toml, both tiers" (rate-limit rendering, pre-existing key-name mismatch in `_render_cluster_wide`)
- `jac/jaclang/scale/tests/microservices/test_plan.jac`: "a rate limit with both tiers at 0 renders as configured-but-inert" (same `_render_cluster_wide` mismatch)

None of the four touch the KEDA/autoscaler paths this branch exists to exercise.

This branch is a testing artifact, not intended to merge as is. It exists so the KEDA scale-to-zero work across these PRs can be exercised together before each PR lands individually.

## Test plan

- `jac/jaclang/scale/tests/deploy` and `jac/jaclang/scale/tests/microservices` (634 tests, the KEDA/autoscaler/deploy/gateway core): 631 passed, 3 skipped, 0 failed


MusabMahmoodh (commented):
Reviewed as a validation artifact, which is what the body says it is, so this is
a comment on whether it can do the job rather than an approve on the diff.

**The stack composition is right and the ordering is defensible.** #7319 first
(checker, unblocks everything else compiling on this line), then #8175 before
#8464 since the autoscaler status surface reads through the asset/route work,
then the three KEDA behavior PRs in dependency order with #8779 last because it
consumes the interceptor address resolution the others do not touch. I would
have stacked it the same way.

**The prerequisite backports are the part I would flag to anyone reading the
result.** Six merged PRs are being carried onto 0.34.18, and #7969 among them is
the FastAPI/uvicorn/pydantic/httpx removal. That is not a small delta to graft
onto a release line, and it means a green run here is evidence about *this
composite tree* and not directly about what 0.34.18 users have. Worth saying
plainly in the body, because the natural misreading of a green integration
branch is "these PRs are safe on 0.34.18", and what it actually shows is "these
PRs are coherent with each other on top of a modernized serving stack".

That is still the evidence worth having. It just needs its scope stated so a
later reader does not over-claim it.

**The `_build_pdb_manifest` bug is the return on the whole exercise.** A dropped
`min_available` on the computed spec is precisely the class that only appears
when several PRs that each touch autoscaling meet, and it would have shipped
invisibly through six individually green PRs. That single find justifies the
branch existing, and it is worth calling out in the body above the fold rather
than in a parenthetical at the end, because it is the argument for doing this
again next time.

One request on the removals, and it is the only thing I would hold for: "removal
of a handful of unrelated tests that depend on large out of scope refactors not
part of this stack" needs to be a list, not a count. On a testing artifact the
set of tests that are *not* running is the most important fact about the run,
and right now a reader cannot reconstruct it without diffing 245 files. If any
of those removed tests cover the KEDA or autoscaler paths this branch exists to
exercise, the 631-passed number means less than it appears to. I do not expect
that is the case, but the body should let someone check without taking it on
faith.

---

## On the cluster run

I said on #8779 that I would run `keda_http_activation_rpc_wake_real_e2e.sh`
against the combined head rather than spinning a cluster per PR, and this is
that head, so the offer lands here.

What I will do, against the protocol in `jac/jaclang/scale/tests/TESTING.md`
(#8654), which exists for exactly this claim:

- build the binary from this branch and prove it is serving this source before
  trusting any result, which is section 1 and is the step that has burned me
  most often
- deploy on a kind cluster with the KEDA HTTP Add-on installed, at a
  **non-default interceptor namespace**, which is the specific gap #8779's
  finding 1 opened and the one thing no unit test on any of these six PRs can
  close
- exercise the scale-to-zero wake through both the gateway and an sv-to-sv RPC
  hop, and report the wake latency, since #8734's whole thesis is that number

That is a real cluster run, not a suite run, and I will post the result here
with the binary provenance so it can be checked rather than believed. If it
comes back clean it is the strongest evidence any of these six PRs will have,
and each can then merge individually on its own review.

Not blocking anything in the meantime. The six PRs should keep landing on their
own merits; this is additive confirmation, not a gate.

PR #8848, opened 2026-09-02
fix(scale): make the admin HTTP traffic card report latency, error rate and a real time window

kashmithnisakya:
## Problem

The HTTP traffic card answers "which paths has this process ever served?" It cannot answer the three questions an operator actually has: is it fast, is it failing, and is that getting worse.

Nothing on it carries a unit, a scale, or a trustworthy time window. Concretely, from the source:

| # | Defect | Where |
|---|---|---|
| D1 | Header says `Last 60s`, but the endpoint list below it reads `_http_requests_total` raw, a counter since process start. Two windows, one label. | `MetricsPage.jac:398`, `getHttpRequestsData` |
| D2 | The bar row has no axis, no scale, no units. `getBarHeight` floors any non-zero value to 14%, so a busy window and an idle one are both unreadable, and idle is indistinguishable from broken. | `MetricsPage.jac:102-112` |
| D3 | `_http_request_duration_seconds` is a 13-bucket histogram. Every `_bucket` series already reaches the browser with its `le` label. Nothing reads it. There is no p-anything on the card. | `prometheus_metrics.jac:87-93` |
| D4 | `getLatencyForPath` matches on `path` only, but the histogram is labelled `["method","path"]`, and it assigns rather than accumulates inside the loop. A path serving two verbs resolves to whichever series parsed last. | `MetricsPage.jac:187-209` |
| D5 | Rows are `(method, path, status)` triples, so one endpoint splits into several rows and per-endpoint error rate is impossible to read. The React key is `path + status`, dropping the method, so two verbs on one path collide. | `MetricsPage.jac:167-180`, `:420` |
| D6 | `errorCount` is computed and never rendered. Upstream the summary only counts `5xx`, so 4xx never reaches the UI at all, while the list paints 4xx amber as though it were tracked. | `MetricsPage.jac:267`, `admin_portal.impl.jac:1232` |
| D7 | "Slowest path" is a lifetime mean gated at >1000ms. It never decays, so one slow cold start pins the banner on for the life of the process and reads as a live incident. | `MetricsPage.jac:311-320` |
| D8 | Headline latency is a lifetime mean across all paths, so sub-millisecond health probes drag it toward zero. | `admin_portal.impl.jac:1243-1255` |
| D10 | The GC card prints "Gen 2 is doing the heavy lifting - no memory leaks detected." as a hardcoded string, regardless of the counters beside it. | `MetricsPage.jac:524-533` |

The before screenshot shows D5, D7, D8 and D10 all firing at once: `/walker/TaskCounts` occupies two rows (200 and 500), the "slowest path" is the admin portal's own client build reported as an 18.7s mean, and the GC verdict claims Gen 2 is doing the heavy lifting while Gen 2 sits at 129 against Gen 0's 277695.

## What changed

**The histogram buckets are now read, server-side.** `/admin/metrics` interpolates p50/p95/p99 inside the straddling bucket and ships them in `summary`, so the page only renders; that is the whole of the fix for D3. No new instrumentation, no new endpoint, no backend metric added: the data was already in the payload.

- **One control per window, each labelled.** 60s / 5m / 15m drives the chart and the throughput tile. The distribution and the endpoint table are explicitly marked `since restart`, because a single-process scrape endpoint genuinely cannot give a windowed percentile without a TSDB. Two windows are fine; two windows under one label were not. Until the history buffer is older than the selected window the label says how much has actually been sampled, and before the second sample the card says it is collecting.
- **Scope filter, defaulting to App.** Paths are classified server-side as app / admin / health / system, with the system set read from `GATEWAY_WELLKNOWN_SEGMENTS` in routing.jac (`/metrics`, `/docs`, `/openapi.json`), so probe, scrape and self-traffic stop setting the agenda. The non-app scope is labelled Platform. Switching to `All` drops p50 from 1.38s to 463ms on the demo app, which is D8 made visible.
- **A real chart.** Area plot in req/s with a labelled y scale, time ticks, a true zero baseline and an explicit idle state. Errors get their own lane below the plot with a gap and a direct label, never a red segment stacked against the orange series.
- **One row per endpoint.** Statuses fold into an `Err %` column alongside p50 and p95, sorted by p95 so the slowest endpoint is always first. Fixes the duplicate React key too.
- **Retired the "slowest path" banner.** A p95-sorted table does the same job continuously, with no 1000ms cliff and no lifetime mean dressed as an alert.
- **The summary is what the page renders.** `summary` now carries `active_requests`, `scopes` (app / other / all, each with requests, 4xx and 5xx counts, p50/p95/p99 and the latency distribution) and `endpoints` sorted by p95. The flat `total_requests`, `error_count` and `avg_latency_ms` fields, which nothing read any more, are gone; the reference doc's example is updated. The error tile shows the 4xx/5xx split.
- **The GC note is derived** from the generation counters instead of asserted.

The Runtime tile is deliberately untouched so this does not collide with #8831.

## Verification

Run against a local `jac run --serve` app generating slow and failing traffic:

- `jac check` and `jac fmt --check --lintfix` clean on all three files
- `jac test jac/jaclang/scale/tests/misc/test_metrics.jac` 21 passed
- `jac test jac/jaclang/scale/tests/server/test_admin.jac` 29 passed
- Client bundle builds; the chip handlers compile to real arrow functions rather than tripping the statement-body lambda issue

The summary tests cover the interpolation on hand-written buckets and on a real collector exposition, the path classifier, the scope totals and the endpoint ordering.

## Known gaps

- Percentiles and endpoint counts are since-restart, not windowed. Windowed values need per-endpoint bucket deltas retained across polls; deferred deliberately.
- The registry is per-process, so behind a Service with more than one replica consecutive polls sample different counters. The card does not yet say which pod answered. That is the remaining half of D2 and wants a follow-up.
- `walker_duration_seconds`, `ws_connections_active`, `ws_broadcasts_total` and `read_tier_retries_total` are registered but still surfaced nowhere.

Before:
00-before-original-card

After:

04-metrics-15m-window


jac/jaclang/scale/admin/impl/admin_portal.impl.jac
```diff
@@ -1219,6 +1219,7 @@ def _calculate_metrics_summary(metrics_data: list[dict[str, any]]) -> dict {
         'total_requests': 0,
         'active_requests': 0,
         'error_count': 0,
+        'client_error_count': 0,
```
MusabMahmoodh: Nothing reads this. At head the page reads exactly one summary field, `active_requests` (`MetricsPage.jac:97` and `:522`), and computes 4xx/5xx itself from the raw counter in `scopeTotals` and `getEndpointRows`; so `total_requests`, `error_count` and `avg_latency_ms` lost their only consumer in this PR and this fourth unread field arrived with a test. Two ways to make it one source: move `percentileFromBuckets` server-side into `_calculate_metrics_summary`, which already walks the same histogram for the mean, so the interpolation gets a real test in `test_metrics.jac` and the page only renders; or delete the summary fields nothing reads and drop this change. Which did you intend?

jac/jaclang/scale/admin/ui/pages/admin/monitoring/MetricsPage.jac
```diff
@@ -349,30 +706,49 @@ def:pub MetricsPage -> JsxElement {
...
+                    <span className="text-xs text-faint">Throughput</span>
                     <Globe className="w-4 h-4 text-faint"/>
                 </div>
                 <div className="text-2xl font-semibold text-fg mt-2 tabular-nums">
-                    {totalReqs}
+                    {throughputDisplay}
+                    <span className="text-base text-faint"> req/s</span>
+                </div>
+                <div className="text-xs text-faint mt-1">
+                    {str(windowReqs) + " in last " + windowLabel()}
```
MusabMahmoodh: "N in last 15m" describes a buffer that only exists since the page was opened (`histRef` starts empty, capped at 300 points at :104), so on first load and after every refresh the card says no traffic in the last 15m regardless of load, and for the next 15 minutes the window is shorter than its label. That is the D1 row of your own table, one label over a different window. `elapsedSecs` is already computed at :583:

```suggestion
                    {str(windowReqs) + " in last " + (windowLabel() if elapsedSecs >= windowSecs else formatSpan(elapsedSecs))}
```

and the empty state at :826 becomes "collecting, first sample in a few seconds" while `nWin < 2`.

jac/jaclang/scale/admin/ui/pages/admin/monitoring/MetricsPage.jac
```diff
@@ -26,14 +25,58 @@ def:pub MetricsPage -> JsxElement {
...
     # (closure), so the chart froze after the first few points. The ref is
     # stable across renders, so .current always reads the latest buffer.
     histRef = useRef([]);
+    infEdge = 1000000000.0;
+
+    def classifyPath(path: str) -> str {
+        if path.startswith("/healthz") or path.startswith("/health") {
+            return "health";
+        }
+        if path.startswith("/admin") {
```
MusabMahmoodh: On the gateway `record_request` runs for every dispatched path including the builtins (`microservice_gateway.impl.jac:875`), so Prometheus scrapes of `/metrics` and hits on `/docs` and `/openapi.json` land in the App scope; a 15s scrape is a 0.07 req/s floor the idle state never reaches. Single-process is fine because the middleware skips the metrics endpoint. Worth adding those three to the non-app class here.

MusabMahmoodh (approved):
Read the whole page rewrite against the payload it consumes. The percentile interpolation over the histogram it was already being sent is the right source, the window is wall-clock and divides by real `dt`, the summary test is real and fails pre-fix, and there are no mocks. One thing I want an answer on before merge, inline at the backend change: the new summary field has no reader, and neither do three older ones after this PR. Two smaller ones on the window label and the scope filter.

Coordination note: this and #8831 both edit the tile grid and the import list in `MetricsPage.jac`; whichever lands second rebases.

jac/jaclang/scale/admin/impl/admin_portal.impl.jac
kashmithnisakya: went with the first: the summary is now the one source. `/admin/metrics` builds `scopes` (app / other / all, each with requests, 4xx and 5xx, p50/p95/p99 and the distribution) and `endpoints` sorted by p95; the flat fields nothing read are gone and the page only renders. interpolation is tested on hand-written buckets and on a real collector exposition. 13eb46a39

jac/jaclang/scale/admin/ui/pages/admin/monitoring/MetricsPage.jac
kashmithnisakya: done. the label is the window only once the buffer is older than it, otherwise it is the sampled span, and before the second sample the card says it is collecting. also raised the buffer to 400 points so a full 15m window is actually coverable at the 3s poll. 13eb46a39

jac/jaclang/scale/admin/ui/pages/admin/monitoring/MetricsPage.jac
kashmithnisakya: done, and the classifier moved server-side so it reads `GATEWAY_WELLKNOWN_SEGMENTS` from routing.jac instead of keeping its own list; those land in a `system` kind under the non-app scope, now labelled Platform. 13eb46a39

jac/jaclang/scale/admin/impl/admin_portal.impl.jac
```diff
@@ -1221,48 +1221,207 @@ def _parse_prometheus_text(raw_metrics: str) -> list[dict[str, any]] {
     return metrics_data;
 }
 
+glob _LATENCY_GROUPS: dict[str, float] = {
+         'under 25ms': 0.025,
+         '25 - 100ms': 0.1,
+         '100 - 500ms': 0.5,
+         '500ms - 2.5s': 2.5
+     };
```
MalithaPrabhashana: These four bounds only line up with `DEFAULT_BUCKETS`. `histogram_buckets` is a supported setting, documented at `jac-scale-kubernetes.md:896`. When it is changed, `_cumulative_at` snaps each bound down to the nearest edge that actually exists, so the bands count the wrong requests.

With `histogram_buckets = [0.01, 0.1, 1.0]` there is no edge at 0.5, so `100 - 500ms` and `500ms - 2.5s` both end up reading the counts at 0.1 and 1.0:

| band | true | reported |
| --- | --- | --- |
| under 25ms | 5 | 5 |
| 25 - 100ms | 10 | 10 |
| 100 - 500ms | 20 | 0 |
| 500ms - 2.5s | 8 | 28 |
| over 2.5s | 3 | 3 |

The 20 requests that took 200ms get reported as taking over half a second, and `100 - 500ms` is always zero no matter what. The counts still add up to 46, so the chart looks normal while pointing at the wrong band.

Failing test, drops straight into `tests/misc/test_metrics.jac`:

```jac
glob BAND_TRAFFIC: list = [(0.005, 5), (0.05, 10), (0.2, 20), (1.0, 8), (5.0, 3)],
     BAND_TRUTH: list = [5, 10, 20, 8, 3];

def _distribution_for(cfg: dict) -> list {
    collector = PrometheusMetricsCollector(config=cfg);
    for (latency, n) in BAND_TRAFFIC {
        for i in range(n) {
            collector.record_request("GET", "/walker/Work", 200, latency);
        }
    }
    summary = _calculate_metrics_summary(
        _parse_prometheus_text(collector.expose().decode("utf-8"))
    );
    return summary["scopes"]["app"]["distribution"];
}

test "latency bands are exact on the default buckets" {
    dist = _distribution_for({"enabled": True, "namespace": "band_default"});
    assert [g["count"] for g in dist] == BAND_TRUTH;
}

test "latency bands hold under a custom histogram_buckets" {
    dist = _distribution_for(
        {
            "enabled": True,
            "namespace": "band_custom",
            "histogram_buckets": [0.01, 0.1, 1.0]
        }
    );
    got = [g["count"] for g in dist];
    assert got == BAND_TRUTH , f"bands={[g['label'] for g in dist]} got={got} want={BAND_TRUTH}";
}
```

On this branch:

```
.F
E   AssertionError: bands=['under 25ms', '25 - 100ms', '100 - 500ms', '500ms - 2.5s', 'over 2.5s'] got=[5, 10, 0, 28, 3] want=[5, 10, 20, 8, 3]
1 passed, 1 failed
```

The first test passes, so the default path is exact and only the custom layout is wrong.

Suggestion: build the bands from the bucket edges that are actually there and put the real bound in the label, rather than assuming the four default edges.

jac/jaclang/scale/admin/impl/admin_portal.impl.jac
```diff
@@ -1221,48 +1221,207 @@ def _parse_prometheus_text(raw_metrics: str) -> list[dict[str, any]] {
...
+            };
+        }
+        lo = count;
+        lo_edge = upper;
+    }
+    return {'value': lo_edge, 'over': True, 'known': True};
+}
+
+
+def _cumulative_at(buckets: list[dict[str, float]], bound: float) -> float {
```
MalithaPrabhashana: This is where the snapping happens. `b['le'] <= bound` keeps the last edge at or below the bound you asked for, so a bound that is not a real bucket edge quietly becomes a smaller one. Fine for the default buckets, where all four band bounds are real edges. Wrong for any other layout. See the note on `_LATENCY_GROUPS`.

MalithaPrabhashana (changes requested):
Checked this by running it.

- `test_metrics.jac` 23 passed, `test_admin.jac` 29 passed
- `jac fmt --check --lintfix` clean on all three files
- Percentile math is correct. I wrote a separate implementation of Prometheus's `bucketQuantile` and compared the two over 12000 random cases (tight spreads, wide spreads, everything in `+Inf`, values sitting exactly on bucket edges). Zero differences. The `over` flag matches Prometheus returning the last finite edge.
- Nothing else in the repo reads the removed `total_requests`, `error_count` or `avg_latency_ms`.
- #8831's Runtime tile is untouched, no collision.

Blocker: the latency bands are hardcoded to the default bucket layout and report wrong numbers when `histogram_buckets` is changed. Failing test in the inline comment.

Question, not blocking: this drops three documented fields from the `/admin/metrics` response. Nothing in the repo reads them, but an external script would break. Should the release note be `.breaking.md` instead of `.bugfix.md`?

jac/jaclang/scale/admin/impl/admin_portal.impl.jac
kashmithnisakya: done. bands are cut at the configured edges nearest 25ms/100ms/500ms/2.5s, deduped, and labelled with the edge actually used; your default test is in as-is, the custom one asserts what [0.01, 0.1, 1.0] can resolve (4 bands, [5, 10, 28, 3], since 200ms and 1s share a bucket there). the hand-written exposition test was asserting five default labels on a two-edge histogram, so it now asserts the three bands those edges give. 39db75b0f

jac/jaclang/scale/admin/impl/admin_portal.impl.jac
kashmithnisakya: gone. counts are read at the chosen edge itself now, so there is nothing to snap. 39db75b0f

jac/jaclang/scale/admin/impl/admin_portal.impl.jac
```diff
@@ -1221,48 +1221,207 @@ def _parse_prometheus_text(raw_metrics: str) -> list[dict[str, any]] {
...
+}
+
+
+def _latency_groups(buckets: list[dict[str, float]]) -> list[dict] {
+    total: float = buckets[-1]['count'] if buckets else 0.0;
+    out: list[dict] = [];
+    prev: float = 0.0;
+    for (label, bound) in _LATENCY_GROUPS.items() {
+        seen: float = _cumulative_at(buckets, bound);
+        out.append({'label': label, 'count': int(seen - prev)});
```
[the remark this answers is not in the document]
kashmithnisakya: fixed in 39db75b0f, bands follow the configured histogram_buckets and each label names the real edge.

MalithaPrabhashana (approved):
The band fix looks right. I re-ran the check against your implementation: default buckets still resolve all five bands exactly, and with `histogram_buckets = [0.01, 0.1, 1.0]` the labels now follow the real edges (`under 10ms`, `10 - 100ms`, `100ms - 1s`, `over 1s`) with counts that match those ranges. Folding 200ms and 1s into one band is the honest answer when only three finite edges exist. Metrics suite is 25 passed locally.

Thanks for picking up the release-note category too.

One note on CI, not on your code: the `jac-check` failure is not from this PR. It is `client/targets/register.jac`, which this branch does not touch and which is byte-identical to main, and it passes when I run it locally at this exact commit. PR #8822 failed the same job eight seconds later with different unrelated errors, and the only four main commits since this PR last passed jac-check are a one-line log message, a release bump and two test files. It looks like the restored JIR analysis cache in CI, the same class as #8909. A re-run of that job should clear it.

PR #8868, opened 2026-09-02
fix(refcount): replace atomic RMW with non-atomic load/store for refcount updates

mgtm98:


christianwilkins (commented):
Reviewed `378e5adc8bd8fdd71f5edcaef39b85e4fee7ccf5`, including all four changed files and the allocation/retain/release and native-entry paths. The native fixture compiled and JIT-executed successfully for 0, 1, 500, and 10,000 iterations, with the expected results; its generated IR contains no `atomicrmw`. Current-head CI has 24 successful checks and 3 skips, and the PR is conflict-free.

The remaining question is the confinement contract behind making these updates non-atomic. `_gen_native_interop_stubs` in `compiler/backends/py/jcir_gen_pass.impl/interop.impl.jac` emits `ctypes.CFUNCTYPE` calls against the same module engine, without an entry lock, while `JacContext.thread_run` submits arbitrary callables to the runtime thread pool. Those entry points do not by themselves enforce the release note’s “single-threaded per isolate” premise. Please document/enforce which native entries and objects are confined, and add a regression test for concurrent entry (either rejection/serialization or the supported isolation behavior) before relying on that premise to remove atomic updates. I have not reproduced a shared-object race, so the sequential tests above should not be read as concurrency validation.

Small documentation correction: the new fixture’s header still refers to a `JAC_RC_NONATOMIC` toggle, but this patch makes the behavior unconditional.

PR #8872, opened 2026-09-02
test(scale): pin the rate limiter state-backend selection that #8981 fixed without a test

### LINKED ISSUE #8333: Jac Scale: the rate limiter asks for a state backend with a config key the loader never emits, so every replica keeps its own token buckets
## Summary

`build_rate_limit_middleware` chooses between the shared Postgres state backend
and the in-process one by reading `scale_enabled` out of the rate-limit config
block. `normalize_microservices_config` never puts that key in the block, so the
read is always `False` and `get_state_backend` always returns
`InProcessStateBackend`. `PgStateBackend` has no reachable caller in a deploy.

The user-visible consequence is that a configured limit is enforced per pod, not
per fleet: at `replicas = N` a client can send `N * per_ip_rpm` requests per
minute before anything returns 429, and which pod a request lands on decides
whether it is counted. The failure is silent in both directions - nothing warns,
and the fallback log line in `get_state_backend` ("Shared state will NOT span
replicas") cannot fire because the branch that logs it is never entered.

## Provenance (resolved 2026-08-17)

Not a regression: the shared backend has never been reachable.

- `git log -S"scale_enabled" -- jac/jaclang/scale/runtime/middleware/rate_limit.jac`
  -> one commit, d4f74f0109 "feat: fold jac-scale into core jaclang as a
  built-in, capability-driven plugin (#7034) (#7038)". Same single answer for
  `get_state_backend` (`state/factory.jac`) and for the `'rate_limit'` block in
  `config/impl/config_loader.impl.jac`. All three arrived together in the fold,
  already diverged, so the defect came in from the pre-fold jac-scale tree
  rather than being introduced by a later change.
- `git log --all -S"'scale_enabled'" -- jac/jaclang/scale/config/ jac/jaclang/scale/runtime/`
  returns **nothing**: the string has never existed as a dict key anywhere in
  config or runtime, in any commit on any branch. The normalizer never emitted
  it and never did.
- Today, `scale_enabled` exists only as a *parameter* name -
  `state/factory.jac:10 get_state_backend(scale_enabled: bool = False)` and
  `runtime/planner/planner.jac:28-41`, where the planner receives it as an
  argument and uses it to pick `STATE_PG` vs `STATE_IN_PROCESS`. The middleware
  is the only place that tries to read it out of a config dict.

Build for the measurement below: stock `jac 0.36.1` release binary (not a dev
build), EKS, 2026-08-17.

## Code evidence (upstream/main, read 2026-08-17)

`jac/jaclang/scale/runtime/middleware/rate_limit.jac:150`

```
backend: StateBackend = get_state_backend(bool(cfg.get("scale_enabled", False)));
```

`cfg` is the normalized `rate_limit` block; the gateway passes it at
`jac/jaclang/scale/runtime/gateway/microservice_gateway.impl.jac:810`:

```
rl_mw = build_rate_limit_middleware(ms_cfg.get("rate_limit", {}));
```

`jac/jaclang/scale/config/impl/config_loader.impl.jac:594-602` builds that block
from a fixed key set, and `scale_enabled` is not in it:

```
'rate_limit': {
    'enabled': ..., 'per_ip_rpm': ..., 'per_user_rpm': ...,
    'burst_multiplier': ..., 'exempt_paths': ...
},
```

`get_state_backend` (`jac/jaclang/scale/state/factory.jac:10-30`) has exactly one
caller in the tree, the line above, so the `scale_enabled` branch that builds
`PgStateBackend` is dead code today.

## Measured (EKS, jac 0.36.1 stable release, 2026-08-17)

Deployed on EKS with the stock 0.36.1 release binary. Gateway at two replicas,
scheduled on two different nodes:

```toml
[scale.microservices.rate_limit]
enabled = true
per_ip_rpm = 60
per_user_rpm = 0
burst_multiplier = 10.0

[scale.microservices.services.__gateway__]
replicas = 2
[scale.microservices.services.__gateway__.hpa]
enabled = false
```

`capacity = max(1.0, 60 * 10.0 / 60) = 10` tokens, refill 1/s. One in-cluster
client pod (single source IP 192.168.159.142) drove 20 requests at gateway pod
A, then immediately 20 at gateway pod B:

| burst | target | 200 | 429 |
|---|---|---|---|
| 1 | gateway A `192.168.93.172` | **10** | 10 |
| 2 | gateway B `192.168.40.180`, same client IP | **10** | 10 |

**Total 20 admitted against a bucket of 10.** Pod A's bucket was fully drained
when burst 2 started; pod B granted the same client a complete second bucket.
A shared backend would have admitted ~0 in burst 2. Re-run once more after a
25s refill: identical, 10 and 10.

Mechanism confirmed in the gateway log: **zero** occurrences of `PgStateBackend`
or the "Shared state will NOT span replicas" fallback warning. That warning is
only reachable inside the `scale_enabled=True` branch, so its absence together
with in-process behavior proves the branch never ran, with a healthy Postgres
available in the same namespace.

Scaling consequence, stated as the users see it: the effective per-IP ceiling is
`replicas * per_ip_rpm`, and it moves whenever the HPA scales the gateway.

## Reproduction

Deploy any microservice project with the config above, then drive one IP at two
gateway pods directly (pod IP, not the Service, so the round-robin is not part
of the measurement). Each pod admits a full bucket. A pod restart resets its
counter, which is the second tell.

## Root cause - the class

Hand-synced duplicate key vocabularies: the loader owns the key set for a config
block and each consumer restates the names it expects from memory, with its own
defaults, so a name that is never emitted reads as its fallback forever. The
display-only instance of exactly this class on exactly this config block is
#7635 (fixed in #8300): the plan renderer asked for `requests_per_minute`, which
the same normalizer never emits. This one is the enforcement instance.

Related but distinct from the middleware's other duplicated defaults
(`per_ip_rpm` 60, `burst_multiplier` 2.0 restated at `rate_limit.jac:145-147`
after the normalizer already applied them) - those agree with the loader today
and are latent, not broken.

## Suggested fix

Prefer deleting the divergence over patching the read:

1. One typed view of the rate-limit block, built by the loader and consumed by
   both the middleware and the plan renderer, so a consumer cannot name a key
   the producer does not emit. Defaults live once, in the loader.
2. Whether the deploy is scaled is not a property of the rate-limit block at
   all - it belongs to the same signal the planner already uses
   (`runtime/planner/planner.jac:39` picks `STATE_PG` vs `STATE_IN_PROCESS` from
   `scale_enabled`). The middleware should take the backend the planner chose
   rather than re-deriving it from a config dict.
3. Fail loud on the way through: if a fleet is running more than one replica
   with an in-process limiter, say so at gateway boot instead of silently
   multiplying the ceiling.

## How to validate

The existing `microservices/test_rate_limit.jac` covers bucket mechanics with
real objects and no mocks; the missing test is the multi-replica one, which
needs two gateway processes sharing one Postgres-backed bucket. The EKS run
above is the shape it has to reproduce, and it is cheap to re-run: two gateway
replicas, one in-cluster client, hit both pod IPs directly and assert the second
pod refuses. I will hand over the exact recipe and can re-run it on my rig for
whoever takes this.

Note for the implementer: it will NOT reproduce on a single-node kind cluster
where each pod installs its own jac runtime - four pods exhausted a 48G volume
before any reached Ready. Use EKS, or a base image with jac baked in.

## Scope note

Not included: the duplicated defaults at `rate_limit.jac:145-147`, and the
`X-Forwarded-For` trust question already handled by #8121.

Size: M. Prerequisite: none (#8300 is independent, display only).

MusabMahmoodh:
## Retargeted to test-only, 2026-09-07

The production change this PR carried shipped in #8981 while it was open, so
it is dropped here rather than re-resolved against it. On main today
`build_rate_limit_middleware` reads `cfg.get("shared", False)`
(`rate_limit.jac:137`), `normalize_gateway_config` emits `shared` defaulting
to `True` (`config_loader.impl.jac:536`), and the schema declares the key
(`plugin_config.jac`). `scale_enabled` is gone from the tree. #8333 is fixed.

What #8981 did not ship is a test. Its Validation section reads, in full,
"None." Nothing under `jac/jaclang/scale/tests/` references
`get_state_backend`, `PgStateBackend` or `InProcessStateBackend`, so the
selection that decides whether a fleet enforces one limit or N is currently
unpinned, and #8981 also changed its failure semantics without a regression
test: an unreachable shared store now raises instead of falling back.

So this PR becomes the coverage for the fix that landed.

## What this changes

Tests only. No production file is touched.

- `misc/test_state_backend_selection.jac` (new): pins the property rather
  than the key, because the key has now changed name once.
  - outside a fleet, unshared state stays in this process;
  - a pod is in the fleet by its own environment (`KUBERNETES_SERVICE_HOST`,
    `JAC_SV_SIBLING`), not by a config key it might not have been given;
  - shared state with no reachable store raises out loud instead of being
    absorbed into per-process buckets;
  - a pod asking for *unshared* state still cannot get private buckets.

  That last one is the original defect stated as an invariant: it is the
  shape that let N replicas enforce `N * per_ip_rpm` silently, and it now
  fails closed regardless of what the config says.
- `misc/test_config_loader.jac`: two tests in the file that already owns the
  gateway-config surface, next to the existing "preserves all sections"
  guard - `rate_limit.shared` is present and defaults to `True`, and an
  explicit `shared = false` round-trips. This is the config-layer assertion
  that would have caught the original `scale_enabled` bug at the loader
  rather than in a cluster.

A new file is justified here because no existing scale test file owns the
state-backend factory; the config half deliberately does not get one.

## Validation

Static, plus CI as the executor. Every symbol asserted on was verified
against `upstream/main` at `3ad35097c9`: `get_state_backend` and `in_fleet`
in `scale/state/factory.jac`, the `RuntimeError` text in the same function,
`_backend_holder`, and `jaclang.server.shared_store._holder` (the module
#8981 moved out of `scale/persistence/pg.jac`).

Not included, stated plainly: the reachable-store arm, where a real embedded
Postgres answers as `PgStateBackend` and two limiters over one store consume
one budget. It needs the deploy tier rather than the in-package suite, and
the earlier revision of this PR only got it by depending on a prior test
having built the store - which is the ordering dependence
`scale/tests/TESTING.md` tells people not to write. It belongs with the
deploy suite and is worth its own change.

Closes #8333 (fixed by #8981; this is its regression coverage).


christianwilkins (commented):
Reviewed `72701bfa6482b229ac55df0b1d6ece2679767043`, including both test files, the factory/shared-store callers, and the existing discussion. The six added selection/configuration cases passed when invoked directly against this head’s production APIs. Current-head CI has 23 successful checks and 5 skips; GitHub reports no merge conflicts. I did not run the reachable-PostgreSQL/deployment arm.

One additional coverage case to retain when folding these into the existing suites: initialize with `get_state_backend(shared=False)`, then call `get_state_backend(shared=True)` without resetting the holder. With fleet variables unset and an unreachable database, my probe returned the same `InProcessStateBackend` on the second call instead of enforcing shared storage. The early cached return in `scale/state/factory.jac` bypasses selection, and every new selection test resets that holder first. This is existing production behavior, not a regression introduced by this test-only diff; please capture it as a follow-up regression case/factory fix.

The rebase and test-consolidation work is already described in your latest comment, so I’m leaving that plan in place rather than repeating it.

PR #8879, opened 2026-09-02
fix(scale): jac run keeps the dev server for web-apps, and local microservice mode serves the site

marsninja:
## Description

`jac run` on a web-app project that declares `[scale.microservices.routes]` was
unusable, in two connected ways.

**The topology was picked for you, and it was the wrong one.** A routes table
describes the cut a project wants when it is *deployed*. Locally, `jac run` on a
web-app is a dev loop: vite, HMR, one process. Declaring the deployed cut
silently took that away, and the only escape was
`[scale.microservices] enabled = false` -- which a deprecation warning told you
to delete.

**And the topology you got served no web app.** With the gateway up, `/`,
`/docs/...` and `/walker/...` 404'd, `/cl/app` was raced across every service in
registration order (a service still compiling hung the page), `[serve]
base_route_app` was ignored, and nothing served the client. Route learning also
cached what a service said while it was still compiling, so a slow service lost
its routes for the whole refresh window.

## Approach

**1. Asking for the topology is explicit.** `jac run --ms` (or
`[scale.microservices] enabled = true`) selects local microservice mode;
`--no-ms` / `enabled = false` forces one process. A dev-client run defaults to
one process; `jac run --serve` and every deployed path keep the gateway when
routes exist. The `enabled` deprecation warning is gone, because the key now
means what it says.

**2. The gateway knows which service is the app.** The entry module's service --
the one `ensure_main_service` already synthesizes -- owns the project root: `/`,
`/cl/<page>`, `/static/client.js`, `[serve] base_route_app`, and any client-side
route reach it instead of 404ing or being probed across the fleet. It runs
exactly as the single-process server does: vite and HMR under `--ms`, its built
bundle under `--serve`. In dev the gateway keeps two addresses for it -- the
app's own API for health, spec and walker traffic, the vite dev server for pages
-- so a client dev server that dies costs HMR, not the site.

**3. Boot no longer loses routes.** Services start in parallel instead of one
60-second `/healthz` wait at a time; a service still compiling stays `starting`
rather than being written off as unhealthy and dropped from routing; and its
`/openapi.json` answer is never cached until it is healthy, so its routes are
learned on the first request after it can publish them.

Three smaller fixes fell out of running it for real: `/static` is gateway-owned
rather than learnable (every jac service publishes `/static/{file_path}`, so a
headless peer was answering the app's generated `client.js` with a 503); a
proxied websocket echoes the subprotocol the client offered (a browser fails the
handshake without it, which is why vite's HMR socket never came up); and the
from-source admin console build runs on a background thread instead of holding
the gateway's bind for minutes.

## Verification

`jac/examples/jaclang_org` with its `enabled = false` block removed. Both
topologies were driven through the **same** `scripts/site-browse-journey.sh`,
which ends in a console-error sweep:

- **`jac run --port 8200`** (single process, vite + HMR): journey passes.
- **`jac run --ms --port 8200`** (gateway + one subprocess per service, from a
  cold `.jac/cache`): **journey passes, `rc=0`** -- landing, the install
  one-liner, the ninja book cover, `arena.wasm` served at 58,389 bytes (the wasm
  stop was *not* skipped), wait-wuuut's live source windows, the leaderboard
  shell, the whole socialize flow behind the `social_graph` service (signup,
  post, trending hashtag, like, comment, channel create + post, reload
  persistence), the docs graph, the 404 page, and no uncaught client errors.

The app service's client compiled clean in that run (`Initial client
compilation completed`, zero `E5101`) because it now compiles before its peers
touch the analysis cache; the peers started only once it was healthy.

HMR through the gateway is a single stable socket: one `[vite] connecting… /
connected` pair over 25 s with the page open, where before it churned at ~1.7
reconnects/s.

CI runs that same journey twice over the site, and **both gate**: once
single-process (`--no-ms`) and once through the gateway (`--ms`). The gateway
pass clears `.jac/cache` first, because the single-process pass warmed it and a
module served warm hits the defect below; that one line is the only thing
standing between this lane and #8881, and it goes when #8881 lands.

Regression tests: `test_local_ms_selection.jac` (topology selection and the
`--ms` flag), `test_gateway_webapp.jac` (the app service owns the project root,
a deployed pod's staged shell is untouched, `/cl` keeps its fan-out, the dev
page server owns pages with a fallback to the service, `/static` is nobody's to
claim, a starting peer's routes are learned when it lands),
`test_route_learning.jac`, `test_bodiless_status.jac` (304/204 framing),
`test_stream_ws.jac` (accept once, after the upstream chose), and
`test_run_dispatch.jac` (the `--ms`/`--no-ms` override).

## Related: #8881

The `ReferenceError: enumerate is not defined` error boundary this PR was
bounced on is a pre-existing compiler defect -- a module served from a warm
analysis cache reaches ES codegen with `call_kind` unset, reproducible on `main`
with two consecutive `jac build main.jac --client web` -- and is being fixed in
#8881.

This PR no longer triggers it: the client-serving service starts and compiles
before its peers warm the cache (measured shape: peers-then-app 41 bare
`enumerate` calls, app alone 0), which is why the gateway journey above is
green. On an already-warm `.jac/cache` the defect still fires from either
topology until #8881 lands.

`_build_client_bundle_if_needed` (orchestrator) does not cover this: it keys on
`[scale.microservices.client] entry`, a config shape this project never
declares, so it returns immediately. Ordering was chosen instead because it also
protects the dev-mode client compile, which no pre-build can.

## Also found, not fixed here

- **A new CLI argument does not reach a built kit** unless the manifest's own
  bytes change: `jaclang/cli/manifest.jac` folds the command specs at comptime
  and the precompiled bundle is restored on a prefix key, so `jac run` in the
  kit rejected `--no-ms` outright. This PR adds an argument-uniqueness assertion
  to that manifest, which both re-folds it and pins a real invariant -- but the
  keying itself should re-key on the producer, tracked as #8880. The assert's
  message says so, since comments do not survive that file's fmt gate; until
  #8880 is fixed, the next command to gain an argument without touching
  `manifest.jac` will ship a kit that rejects the new flag.
- Writing that assertion surfaced a **duplicate short flag** within one command
  spec (a short-flag uniqueness assertion fails on the current tree); left out
  here since changing a short flag is user-visible.


jac/jaclang/scale/runtime/orchestrator/process_manager.impl.jac
```diff
@@ -340,6 +476,9 @@ impl ServiceProcessManager.check_all_health -> dict[str, bool] {
             self._failure_counts[name] = 0;
         } else {
             self._failure_counts[name] = self._failure_counts.get(name, 0) + 1;
+            if self._booting {
```
MusabMahmoodh, line 479: The counter keeps counting while `_booting` is true, so a service still compiling when `_wait_for_health` gives up has banked ~45 failures against a `_max_failures` of 3, and `start_health_monitor` clears `_booting` immediately before the thread starts. First tick, 2s later, it is UNHEALTHY and its prefixes 503 until `/healthz` answers, which is the behaviour the warning two lines up says it replaced. Moving the increment below the guard (or resetting `_failure_counts` where `_booting` is cleared) is the whole fix:

```suggestion
            if self._booting {
                continue;
            }
            self._failure_counts[name] = self._failure_counts.get(name, 0) + 1;
```

Worth a test of the transition itself, since nothing covers the post-boot state: drive `check_all_health` a few times during boot, start the monitor, assert the entry is still STARTING.

jac/jaclang/scale/tests/microservices/test_gateway_webapp.jac
```diff
@@ -0,0 +1,428 @@
...
+        );
+        assert client.get("/cl/app").text == "vite /cl/app";
+
+        api_call = client.post("/walker/ping");
+        assert api_call.text == "main /walker/ping" , (
+            f"walker traffic goes straight to the service, not through the "
+            f"page server; got {api_call.text!r}"
+        );
+
+        page_srv.request_stop();
```
MusabMahmoodh: `request_stop()` only schedules `stop_event.set` onto the server thread's loop (`server.jac:157`), and the listener closes later on that thread, so on a loaded runner the gateway can still reach the vite stub and get `vite /` instead of the `main /` fallback. `serve_background()` returns the closure that does `request_stop()` then `thread.join(...)`, and `_spawn_stub` at :64 throws it away. Returning it from `_spawn_stub` and calling it here makes the teardown deterministic.

jac/jaclang/cli/commands/impl/execution.impl.jac
```diff
@@ -217,7 +223,9 @@ def _serve_with(entry_path: str, plan_flags: dict, over: dict, main: bool) -> in
         }
     }
     for (k, v) in over.items() {
-        kw[k] = v;
+        if k in kw {
```
MusabMahmoodh: Every key `_serve_overrides` emits is already in `kw`, so this guard is dead today; the day it is not, a flag the user typed is dropped with no error. Either delete it, or raise on an unknown key so a typo in the override list fails loudly instead of silently.

MusabMahmoodh (approved):
Read this against the surrounding code rather than the diff alone. The core of it is right and it fixes a real ordering problem: making the topology an explicit request instead of a side effect of a routes table, giving the entry module's service the project root so `/`, `/cl/*` and `base_route_app` stop being probed across the fleet, and holding route learning open for a non-HEALTHY service so a slow compile no longer costs a service its routes for the refresh window. No mocks anywhere in the new tests, and they fail on main by reading (`spec_due` returns False after `mark_spec_seen`, `ServiceEntry` has no `is_app`, `websocket_forward` has no `on_upstream`). The WS subprotocol is forwarded and then echoed from the upstream's selection rather than the client's offer, which is the correct order.

One thing I would fix before merge, inline: the `starting` grace ends at `boot_max_wait` and the service is UNHEALTHY one monitor tick later, because the failure counter kept counting during boot. That is the same symptom the PR sets out to remove, just moved 90 seconds later.

One question on the deployed path. `serves_client` is only ever set by `scale_local_serve`, so a fleet pod has `is_app` but no root owner, which the "is_app alone does not own the root" test pins deliberately. With `/static` now gateway-owned and never learnable, a `/cl/<page>` served on the pod loads `/static/client.js` from the gateway origin, `handle_static` misses, no learned owner exists, `forward_to_app` returns None and the SPA fallback needs `text/html`, so it 404s where it previously reached a service that had learned `/static`. Is the deployed answer that the gateway never serves pages and the app is reached at `{app}-service`, or should the `/static/*` case fall back to the `is_app` entry regardless of `serves_client`?

Two smaller ones inline. Also: ci.yml's py-spy dump still greps `jac run --serve main.jac`, but the single-process lane is now `--serve --no-ms main.jac`, so on failure it finds no process, and the gateway lane (the one more likely to hang behind a compiling service) has no dump at all.

Coordination note on my #8658, which adds STARTING plus a real probe to gateway-only mode on the pod: the two are complementary and should not conflict textually, but `_booting` defaulting to True breaks my `test_k8s_health_probe.jac` ghost-unhealthy assertion on a bare `ServiceProcessManager`. I will rebase mine onto this rather than the other way around.

jac/jaclang/scale/runtime/orchestrator/process_manager.impl.jac
marsninja: Fixed in 6633db43e2. The increment now sits below the `_booting` guard; boot ends once through a new `end_boot` (which `start_health_monitor` calls) and that zeroes the counters; and `_booting` is set by `start_all` rather than defaulting to True at construction, so a manager that never booted a fleet is not booting forever and still marks a dead service unhealthy. That last part is also what tripped your ghost-unhealthy assertion in #8658, so it should hold on a rebase as is.

The transition is covered in `scale/tests/microservices/test_boot_grace.jac`: boot polls leave the entry STARTING with a zero counter, one post-boot failure is one strike (UNHEALTHY only after `_max_failures`), starting the monitor forgets boot-time failures, and a never-booted manager reaches UNHEALTHY.

jac/jaclang/scale/tests/microservices/test_gateway_webapp.jac
marsninja: Done in 6633db43e2: `_spawn_stub` now returns the closure `serve_background` hands back (request_stop followed by the thread join), `Fleet` keeps those closures instead of the servers, and every test stops through them, so the fallback assertion only runs once the vite stub's listener has actually closed.

jac/jaclang/cli/commands/impl/execution.impl.jac
marsninja: Went with raising (6633db43e2): an override key that is not a serve argument now raises `ValueError` naming the key, so a typo in the override list fails on the first run. The `plan_flags` loop above keeps its filter, since plan flags legitimately carry keys that are not serve arguments.

PR #8882, opened 2026-09-02
fix(runtime): the identity backend is chosen by configuration, and the core server stops writing sha256

marsninja:
## Description

Accounts created on a dev server stopped logging in after a restart ("Invalid
username or password") while graph data survived. Three defects, one of them
the cause of the disappearing accounts.

### The owner flip (issue 12a)

Which identity stack owns `/user/*` was not a decision. It was a side effect of
an import.

`Jac.get_api_server_class()` asked the scale plugin for a server class.
`JacScalePlugin.get_api_server_class` ran `import from .server.serve
{ JacAPIServer }` inside a `try`, and answered **any** `ImportError` in that
~200-module graph with a DEBUG line and `None`; the runtime then returned the
core `jaclang.server.server.JacAPIServer`. `_scale_provider()` in `runtime.jac`
swallowed the same way one level up, for `jaclang.scale.plugin`.

The two server classes own different credential stores:

| server class | user manager | credentials |
| --- | --- | --- |
| `jaclang.scale.server.serve.JacAPIServer` | `JacScaleUserManager` | Postgres, `scrypt$...` |
| `jaclang.server.server.JacAPIServer` | `UserManager` | sqlite `.jac/data/main.db`, bare `sha256` |

So one transient compile error, one half-written cache entry, or one missing
module moved `/user/*` to a store holding none of the existing rows, and every
account created before that boot was told its password was wrong. Reproduced
deterministically against `main` (a `None` entry in `sys.modules` is how
CPython spells "this import fails"):

```
normal:         jaclang.scale.server.serve
import-broken:  jaclang.server.server
```

Postgres reachability was never the trigger, incidentally: the pg user manager
has raised loudly since #8000 when no store resolves. The QA site's
`.jac/data/main.db` still holds the stranded rows -- `__guest__` plus three real
accounts -- with 64-hex sha256 hashes, which is the fingerprint of a session
served by the core class.

Also found: `JacRuntimeInterface.get_user_manager` is the one selection point
that already consulted configuration, and nothing calls it, so it had drifted
onto a different predicate than the server class it is meant to agree with. It
reads the same function now, so a future caller cannot reintroduce the flip
through the back door.

### The hash format (issue 12b)

The core server wrote `hashlib.sha256(password).hexdigest()` -- unsalted, and a
format `authcrypt.verify_password` rejects by construction. A flip in either
direction stranded every account permanently, and the hashes were weak on their
own terms.

### The dev signing secret (issue 11b)

`[scale.jwt] secret` defaulted to the literal `supersecretkey_for_testing_only!`
in every project, so a dev token was forgeable by anyone who has read the
source, and the secret was not a per-project fact at all.

## What changed

1. **One decision function.** `jaclang.server.identity_owner.identity_owner()`
   resolves the owner from configuration alone: the `JAC_IDENTITY_BACKEND`
   environment setting, else jac.toml `[serve] identity`, else `auto`, which is
   the scale identity when the scale stack is *configured* for the project (a
   `[scale.database]` url, or `JAC_DB_URL` / `JAC_SCALE_RUNTIME` in the
   environment) and the core identity when it is not. That is the predicate the
   scale runtime already used -- `scale_runtime_configured`, the former
   `_scale_runtime_active` -- so there is one rule, not two. It is resolved for
   the project being served (the full target path, falling back to the base
   path) rather than for the directory the command was typed in. No database is
   probed and no import's luck is consulted. The owner picks the server class,
   and the class it picks then always pairs with its own credential store.
2. **Loud, not quiet.** When the resolved owner is scale and the scale stack
   will not import, boot raises with the module name, what it refuses to do,
   and the two ways to choose the core identity on purpose. `_scale_provider`
   now records its `ImportError` so that message can name it.
3. **One password format.** The core server's signup, login, and password
   change go through the shared `authcrypt` module. `verify_password`
   additionally accepts a 64-character hex sha256 row, and both
   `UserManager.authenticate` and `IdentityStorage.authenticate` rewrite that
   row as scrypt on the login that verifies it -- so accounts written before
   the migration carry across without a reset and never verify against sha256
   twice. New credentials are always scrypt.
4. **A per-project dev secret.** With nothing configured, the secret is minted
   into `.jac/data/jwt_secret` (gitignored, created 0600, linked into place so
   a race cannot mint two) and read back on later boots, so sessions issued
   before a restart still validate after one. A configured secret still wins
   untouched. A cluster with nothing configured keeps the shipped placeholder
   -- a per-project file would be per-replica there -- and warns at boot. A
   project directory that cannot be written signs with a secret local to that
   process and says so, rather than taking the boot down.
5. **`[serve] identity`** is documented in the config reference, and the jwt
   secret's own docs no longer describe a shipped default that is gone.

## Verification

Unit (new): `test_identity_owner.jac` covers both configured branches, the
`auto` default, and the loud failure; `test_identity_password_migration.jac`
covers a scrypt signup, a legacy row verifying once and being rehashed, a
password change from a legacy row, and the shared verifier;
`test_dev_signing_secret.jac` covers mint-once, survive-a-restart,
two-projects-differ, and a configured secret winning.

Live, against `jac/examples/jaclang_org`:

- default (`auto`): signup and login are served by the scale identity (UUID
  user id, JWT), `.jac/data/jwt_secret` is created 0600, and no `main.db`
  appears; after a full server restart the account logs in **and the token
  minted before the restart still validates** on `/user/me`.
- `JAC_IDENTITY_BACKEND=core`: `/user/*` is served by the core identity, and the
  sqlite row reads `scrypt$16384$8...`.
- a hand-seeded bare-sha256 row on that core database logs in once, the stored
  hash becomes `scrypt$16384$8...`, and a wrong password is still rejected.

Suites run green locally: `runtimelib/test_serve.jac`,
`runtimelib/test_serve_client.jac`, `tests/project`, scale
`test_identity_fail_loud`, `test_identity_password`, `test_auth_cutover`,
`server/test_user_identity`, `server/test_sso`, `server/test_ws_hardening`,
`data/test_gateway_owned_identity`, `data/test_auth_tokens_fail_closed`,
`misc/test_shared_store_failure`.

`scale/tests/server/test_serve.jac` skips locally (no `requests` here) and runs
in CI; its four hardcoded copies of the placeholder secret now read the
resolved `JWT_SECRET`.

## Review follow-ups

Every item from the review, with what it turned into.

**Blocker.** The owner was resolved for the current working directory. Passing
the base path was not enough on its own: for `jac run inner/who.jac` the base
path dir is the parent the command was typed in, so the inner project's
`[serve] identity` was still ignored. Both call sites now resolve from
`get_full_target_path()` (base path as the fallback) through
`get_config_for_dir`, the way every other server site reads config. Two new
tests drive it: one sets the base path at the parent and the target inside the
child, exactly as that command does, and both fail if either call site reads
the base path dir alone. Verified live as well: a parent pinned `scale` with a
child pinned `core`, run as `jac run inner/who.jac` from the parent, now
answers `core` / `jaclang.server.server` and answers `scale` for the parent's
own entry.

1. Fixed as above.
2. `auto` now means "the scale stack is configured", through the existing
   predicate; the core half is live code and the rival predicate is gone.
   `_scale_runtime_active` is renamed `scale_runtime_configured` and is the one
   implementation both the runtime and the identity owner read.
3. An unrecognized value raises, naming `auto`, `core` and `scale`, for both
   the environment variable and the jac.toml key.
4. Config discovery no longer swallows exceptions; an unreadable `jac.toml`
   raises rather than resolving automatically. Test added.
5. `project_signing_secret` and the core store now resolve their directory
   through one function, `project_data_dir`, which honours `JAC_DATA_PATH`.
   Test added.
6. A secret that cannot be persisted no longer falls back to a per-process
   value: it raises, naming the path, the reason and `[scale.jwt] secret`. The
   same applies when the file exists but reads back empty (the nit below).
   Test added.
7. Documented as an upgrade note in the fragment. There is no runtime replica
   signal outside Kubernetes, so a project that mints its own secret now says
   so once at boot, naming the file and that a deployment must set the key;
   inside a cluster the placeholder and its warning are unchanged.
8. The module-level `get_jwt_config()` calls in `scale/identity/user_manager.jac`
   and `scale/server/serve.jac` are gone. `user_manager` exposes lazy
   `jwt_secret()` / `jwt_algorithm()` / `jwt_exp_delta_days()` / `sso_host()`;
   `serve.jac`'s copies were unused and were deleted. Verified: importing both
   modules creates no `.jac/data/jwt_secret`, and first use does.
9. Said here and in the fragment: `get_user_manager` reads the owner, which is
   keyed on the configured-scale predicate, so it agrees with what it replaced.
10. Settled by removing the second decision rather than adding a third. The
    scale server's `_make_user_manager` no longer consults the owner: a scale
    server always builds a scale user manager, so the in-process and desktop
    path's `assert isinstance(..., JacScaleUserManager)` cannot fire. (Keeping
    that switch is what broke `scale_test_client`: the scale auth endpoints
    reach for `sso_initiate` and `create_internal_user`, which the core manager
    does not have.)
11. Added: the storage-side rehash (a fake `IdentityStorage` with its three
    primitives in memory, covering wrong password, the rehashing login, and a
    scrypt row left alone), the cluster/placeholder branch, and the dev-secret
    tests now clear `KUBERNETES_SERVICE_HOST` and `JAC_DATA_PATH` explicitly.
12. `JAC_IDENTITY_BACKEND` added to the environment-variable table, and the
    `[serve] identity` comment updated to the configured-scale rule.

**Nits.** `get_api_server_class` consults the owner itself now, so its
`type | None` is honest, and the runtime raises rather than falling back when
it answers `None`. `_persist_secret` no longer hands back a per-process secret
when the file exists but is empty or unreadable. The scrypt calls in the rehash
and password-change paths moved out from under `self._lock`. `ensure_sv_service`
raises when the scale package is present but its plugin will not import,
instead of degrading to the core class.

## Second review round

1. **In-process runtime bypassed the owner.** `prepare_inprocess_runtime` built
   the scale `JacAPIServer` directly, so the `assert isinstance(...,
   JacScaleUserManager)` had become vacuously true and a project pinned
   `identity = "core"` reaching the desktop dispatcher or the scale test client
   was quietly served the scale store. It refuses now, naming the file and what
   the two stacks disagree about; the test fails without the check. The refusal
   reads `identity_setting()` -- the configured value -- rather than the
   resolved owner, because serving is itself what turns the scale runtime on,
   so an unpinned project resolves to `core` when asked outside a serving
   process and would otherwise have been refused for asking nothing at all.
   `identity_owner()` is now that setting plus the auto resolution, and the
   unknown-value check moved down with the setting.
2. **Unpersistable secret fired at first request.** The scale server resolves
   the secret as the first thing its `postinit` does, before it builds anything
   else, so it is a boot failure. Verified live with `.jac/data` at 0555: `jac
   run` exits with the actionable error, nothing listens on the port, and
   neither `/user/me` nor `POST /user/login` is reachable to echo a stack trace
   at an unauthenticated caller. `_announce_minted_jwt_secret` now logs after
   the mint succeeds rather than before it is attempted.
3. **Docs corrected.** The `[serve] identity` block, the prose under it and the
   release note now say plainly that the serving path sets `JAC_SCALE_RUNTIME`,
   so under `auto` a served project gets the scale identity whether or not it
   names a database, and `identity = "core"` is how a project chooses the core
   store. The note also records that a non-serving caller of `identity_owner()`
   / `Jac.get_user_manager` sees `core` for the same project (nothing in the
   tree does this today), and that pinning `identity` makes the two agree.
   `JAC_SCALE_RUNTIME` was added to the environment-variable table.

**Nit.** The unreachable `_scaled is None` branch blamed the scale serve module
for a `None` it could only get from the plugin agreeing the owner is core; it
states what actually happened instead.

**One more, found by CI.** `test-scale (data)` failed on the served fixture in
`test_async_io_blocking.jac`: the server compiled 95 modules cold, printed
`Server ready`, and was killed by a harness that had already given up.
`start_server` slept only when the health check *raised*, so a non-200 answer
spun its whole budget out in milliseconds, and a cold server got at most 120
seconds. A PR that touches compiler source always starts cold, because the
cache key carries the compiler digest, so this sits right on the line for that
whole class of change. The wait is a wall clock now, sleeps on every outcome
short of ready, and still ends on the spot when the server exits. The lane is
green on main and was green on this branch's previous head, so the budget is
what moved, not the behaviour.

## CI

Run `33656530666` on head `5f9e066850`: **green, all 22 lanes**, including every
lane that touches this change -- `test-runtime` (the new `tests/runtimelib`
suites against the sealed kit), `test-scale (server)` (the `test_serve.jac`
edits), `test-scale (data)`, `test-scale (root/misc/microservices/deploy)`,
`test-scale-k8s`, `test-client`, `test-packages-and-docs`, `jac-check`,
`test-in-package-sealed`, all four `test-compiler` shards, `test-jac-pack-smoke`,
`test-bootstrap` on both platforms, and `Contribution Checks`.

`test-compiler (passes-native)` needed two reruns to get there, which is worth
recording because it is a main flake rather than anything this branch did:

- It fails on a **different native test every run**, on main as much as here.
  Main `4b3f1d16c3`: `test_native_rc_debug_gate.jac` (`nacompile`
  `TimeoutExpired` after 300 s). Main `463b15fa02`: `test_native_set_scaling.jac`
  and `test_native_set_hash.jac`. Here, first attempt:
  `test_native_rc_debug_gate.jac`, the same test and the same timeout on the
  same command as main's latest, down to the same `1516 passed, 1 error,
  4 skipped` tally. Second attempt: `test_native_gen_pass.jac`. Third: green.
- Both tests that failed here **pass locally against this branch**:
  `test_native_gen_pass.jac` -> 224 passed, 3 skipped;
  `test_native_rc_debug_gate.jac` -> 2 passed.
- That second one took **330 seconds locally** while the `nacompile` call inside
  it is capped at **300**, so it sits on its own timeout boundary and goes over
  whenever the runner is loaded. The lane's RSS log shows workers at 3.9 GB and
  the job takes ~19 minutes.
- Nothing in this branch touches the compiler, the native backend, or
  `tests/compiler`.

`jac-check` and `test-in-package-sealed`, red on main while this branch was
being written, are green here now that main is repaired.


MusabMahmoodh:
One question on the cluster branch of `get_jwt_config` (`config_loader.impl.jac:190-193`): with `KUBERNETES_SERVICE_HOST` set and the resolved secret equal to the placeholder, the server warns once and signs with the public value. Is that the intended end state, or would a refusal at boot for that case be welcome? Asking before #8755 item 1 moves, since the deployed pod is the case its forgery repro targets.

PR #8884, opened 2026-09-02
feat(tests): add memory release tests for native imports and cache handling

mgtm98:


jac/tests/compiler/backends/native/test_native_memrelease.jac
```diff
@@ -0,0 +1,72 @@
...
+}
+
+test "link releases the dep's IR and unit after its cache lands" {
+    scrub_dep_cache();
+    (prog, eng, _) = compile_main();
+    f = get_func(eng, "mr_use", ctypes.c_int64, ctypes.c_int64);
+    assert f(7) == 22;
+    dep = os.path.realpath(os.path.join(FIXTURES, "memrelease_dep.jac"));
+    found = prog.find_module(dep);
+    assert found is None or found.gen.llvm_ir is None , (
```
christianwilkins, line 63: [P3] Separate the LLVM-release assertion from the compiled-unit lifetime claim. This condition succeeds with a non-None module as long as its LLVM field is cleared, so it does not prove that the syntax tree/unit was evicted or reclaimed. A warm direct probe on this head finds exactly that state: dependency present, llvm_ir=None. Add an observation covering the intended tree lifetime, or narrow the test/release-note claim to what is actually checked.

christianwilkins (commented):
Reviewed head `120eef341ae99954e70b8b1e397a50d23ffa09fa`: the full patch, cache-write/link ordering, module invalidation and release APIs, cached demotion lookup, and new fixtures/tests.

P3 validation gap: the new lifetime test proves that a dependency has no live LLVM module, but its `found is None or found.gen.llvm_ir is None` assertion also passes when the compiled module/tree is still present. Please distinguish the LLVM-release guarantee from the stronger syntax-tree memory claim and add an observation that covers the intended lifetime, or narrow that claim. My warm direct probe actually found the dependency in the hub with `llvm_ir=None`; that alone is not evidence of a leak, but it demonstrates what the assertion permits.

Local direct validation: the new fixture compiled natively without diagnostics, produced an engine, and JIT calls returned 22 for `mr_use(7)` and 28 for `mr_use(9)`. I did not measure memory reduction across a large import graph or establish full tree reclamation. No source changes made.

Current-head CI is red. The compiler lane fails `import include auto impl` with an IndexError reading its expected hub entries; that deserves particular attention alongside a change that evicts modules. The runtime lane reports 1888 passed, 5 failed and 8 skipped, with the failures in shared-root behavior. I have not established that either set is a baseline failure or attributed them definitively to this patch. GitHub reports no conflicts.

Not ready to merge: the author needs to resolve or demonstrate the baseline for those CI failures and clarify/validate the memory-lifetime guarantee. Maintainer review follows after green current-head CI.

Issue #8178, opened 2026-08-14
Incremental precompile seed is unsound across same-version compiler changes (module keys carry no compiler identity)


marsninja:
Root-cause class behind the #8140 cache-poisoning incident: the JIR precompile seed (actions cache, restore-keys) reuses any entry whose module key matches, and `compute_module_key` fingerprints the jaclang VERSION plus schema/codegen inputs -- but not the compiler's actual content. Between same-version builds (every commit of an in-flight PR, every CI build off main), a compiler whose codegen behavior changed silently keeps serving its predecessors' output for unchanged sources. On #8140 this shipped 201/639 stale compiler modules inside CI payloads, compiled by the branch's earlier buggy builds; the one-time remedy was a JIR format bump, which only works when someone knows to reach for it. Candidate fixes: fold a compiler-content fingerprint (e.g. hash over jac0core+compiler sources, or the git tree hash mkpayload already knows) into module keys during the payload precompile specifically, accepting that compiler-touching commits forfeit the seed; or scope restore-keys to same-tree-hash builds. Either makes the seed's soundness a checked property instead of a convention.

PR #8885, opened 2026-09-02
feat(cli): jac precompile --force removes the previous _precompiled output before rebuilding

marsninja:
## Summary

A rerun of `jac precompile` reuses any existing `.jir` whose module key still matches its source, so the only way to guarantee a fully rewritten output tree was to delete `_precompiled/` by hand -- which the #8178 seal-abort message even instructs. This adds `--force`, which does that deletion up front: the whole `_precompiled/` tree (every python-version dir, `MANIFEST.json`, `PRECOMPILE_RESULT.json`) is removed before the rebuild, so every unit is rewritten instead of reused.

The warm jir module cache still serves bytecode, so a forced rebuild mostly rewraps cached units rather than recompiling; `JAC_REBUILD=1` remains the switch for a truly cold build.

## Behavior

- `--force` + `--seal-finalize` is rejected up front, before any package checks: finalize only re-manifests an existing tree, so deleting it first would leave nothing to seal.
- `--force` + `--seal` composes: remove, recompile, finalize.
- The standalone module entry (`python -m` path) parses `--force` too, and its usage text documents it.

## Testing

Smoke-tested on a scratch two-module package: plain rerun reports `2 reused`; `--force` prints `Removed previous output (--force): .../_precompiled` and compiles `2/2` with `0 reused`; `--force --seal-finalize` errors with the guard message. `jac fmt --check` passes on all touched files.


christianwilkins (commented):
Reviewed head `6a3e1f2e9e46ade67c763c0357d9a8c247e25f95`: all four changed files, CLI-to-runner argument wiring, package discovery and the precompile/finalization flow. No blocking defect found in the new guard or deletion ordering: the incompatible --force/--seal-finalize combination exits before discovering or removing the output, and normal force deletion targets the resolved package output. No code changes made.

P3 coverage suggestion: preserve the reported two-module smoke sequence as an automated CLI regression, including an existing-output sentinel that survives the rejected --force --seal-finalize invocation. That would protect both the destructive guard ordering and the distinction between a warm reuse and a forced output rewrite. The patch currently adds no automated test for these behaviors.

Current-head CI has 25 successful checks and 3 skips. My independent disposable-package smoke attempt timed out after 600 seconds while compiler setup was still running; the warm/force/guard steps therefore did not execute. This is incomplete local validation, not evidence that this patch caused a hang.

Review state: comment, with no blocking code finding. I am stopping short of approval because I did not independently complete command-level validation; a completed force/guard regression is the remaining validation step for this review.

PR #8887, opened 2026-09-02
fix(client): type() yields the type object, so `type(x) is str` compares like Python (#6149)

### LINKED ISSUE #6149: [Bug] `type(x) is str` / `is not dict` compiles to `typeof x === str` (bare identifiers) in jac2js

SupulHeshan:
Fixes #6149.

## The bug

`type(x) is str` (and `is not dict`, and the `elif` chains built from them) silently answers wrong in every JavaScript target. Python targets are correct.

```jac
if type(x) is str { return "str"; }
elif type(x) is int { return "int"; }
return "other";
```

compiled to:

```js
if ((typeof x === _jac.types.str)) { return "str"; }
else if ((typeof x !== _jac.types.int)) { return "int"; }
return "other";
```

`typeof x` is a string tag (`"string"`, `"number"`); `_jac.types.str` is the runtime type object. Comparing the two is always `false` for `is` and always `true` for `is not`, so `CheckType("hello")` returned `"int"`. Nothing errors: `jac check` passes and the JavaScript is valid, so the wrong answer is silent.

`type(42).__name__` was worse: it emitted `typeof 42.__name__`, which is not parseable JavaScript.

## Root cause

Three lowerings, each locally reasonable, wrong only in combination:

1. `emit_type` lowered `type(x)` to `typeof x` unconditionally, with no awareness of how the result is used.
2. `exit_builtin_type` lowers a bare builtin type name in value position to `_jac.types.<name>`, the runtime type object.
3. `is` / `is not` map generically to `===` / `!==`. `CMP_EMIT_MAP` has no entry for them, so unlike `in` / `not in` (which route to `_jac.poly.contains`) they never get a chance at idiom-specific handling.

(1) is the one that lies about what it produces, so that is where this fixes it — not by special-casing (3).

**Before** — the two sides of `is` never resolve to the same *kind* of value:

```mermaid
flowchart LR
    A1["type(x)"] -->|lowers to| A2["typeof x"] -->|evaluates to| A3["'string'<br/>JS primitive"]
    B1["str"] -->|lowers to| B2["_jac.types.str"] -->|evaluates to| B3["type object<br/>JS function"]
    A3 --> CMP{"==="}
    B3 --> CMP
    CMP -->|"string ≠ object"| R["false"]

    classDef wrong fill:#c1443a,color:#fff,stroke:#8a2f27
    class R wrong
```

**After** — `type(x)` is redirected to resolve the *same* singleton a bare type name already resolves to, so both arrows now land on one shared object instead of two different ones:

```mermaid
flowchart LR
    SHARED["_jac.types.str<br/>one shared object"]
    A1["type(x)"] -->|lowers to| A2["_jac.builtin.type(x)"] -->|resolves to| SHARED
    B1["str"] -->|lowers to| SHARED
    SHARED --> CMP{"==="}
    CMP -->|"same object"| R["true"]

    classDef right fill:#3d7a54,color:#fff,stroke:#2b5b3d
    class R right
```

## The fix

`type()` now returns the value's **type object**, resolved through `_jac.poly._typename` to the very same `_jac.types.<tag>` singleton a bare type name lowers to. The existing generic `is` -> `===` lowering is then correct on its own, with no special case for the idiom.

Fixing the root rather than pattern-matching `type(x) is <builtin>` also makes these work, none of which a targeted carve-out would have covered:

- `type(x) is type(y)`
- `type(a) is MyClass` for user archetypes (resolves to the constructor)
- `str(type(x))` -> `<class 'int'>`
- `f = type; f(x)`, since `_jac.types.type` is generated from this same emitter
- `type(x).__name__`

**Exact-type semantics are preserved.** The issue suggested routing through `isinstance`, but `isinstance(True, int)` is true in Python while `type(True) is int` is not. Identity on singletons gives the exact-type answer; `type_bool_isnot_int` in the equivalence fixture pins this.

## Second defect found on the way

`_jac.poly._typename` did not know that a dict literal lowers to a plain JS object. It reported `"Object"`, so `isinstance(d, dict)` was **false for every dict** on this backend, independent of this issue.

`poly.call` already encoded the correct rule for its own dispatch ("a plain object is a Python dict literal"). The two now share one `_isPlainObject` definition so they cannot drift, matching the single-source-of-truth pattern `_tagSubclass` already uses for isinstance/issubclass.

## Also

- `_jac.types` gains `object` and `NoneType`, which `type()` resolves through and which the subclass relation already assumed existed.
- Type objects gain `__name__`, so `type(x).__name__` matches Python, including for compiled archetypes.
- Table lookup uses `hasOwnProperty`, so an archetype named `constructor` cannot resolve to a JS builtin.

## Tests

`prim_builtins_introspection.jac` already had `type_int`/`type_str`/`type_list` rows with `builtins_introspection_cl()` stubbed to `{}` ("not yet implemented") — the exact fixture this regression belongs in. Extended it with 6 targeted rows (`type_is_str`, `type_isnot_dict`, `type_bool_isnot_int`, `type_arch_is`, `type_arch_name`, `isinstance_dict_t`) instead of adding a new fixture. Verified against the pre-fix emitter: 4 of the 6 mismatch Python on the old code.

EQUIV_CASES rows: 99 -> 99 (no net new test; this reuses the existing "builtins introspection" row).


jac/jaclang/compiler/backends/es/jac_runtime_js.jac
```diff
@@ -1253,6 +1257,22 @@ glob JAC_RUNTIME_JS_OBJECT: str = r"""{
       if (sup === "int" && sub === "bool") return true;
       return false;
     },
+    // `x`'s actual type object -- the same singleton a bare type name lowers
+    // to, so `type(x) is str` is a plain identity check.
+    type(o) {
+      const tag = _jac.poly._typename(o); // own-property: skip inherited "constructor"
```
christianwilkins, line 1263: [P2] Preserve exact type information rather than treating dispatch tags as exact types. A compiled Jac probe on this head reports `type((1, 2)) is list` and `type(1.0) is int` as true, while the corresponding tuple/float identity checks are false. `_typename` maps every array to list and every integral JS number to int, so this new helper produces wrong positive matches after representation information is lost. The same probe gives false for `type(str) is type` because builtin type objects fall through to Function. Add positive and negative compiler-runtime regressions and carry/recognize the actual type identity.

christianwilkins (changes requested):
Reviewed `eec2c29b0b7235fd3dcfcd5af7d31194e228c9b1`: full diff, builtin emitter/type-object generation, polymorphic runtime classification, new coverage, existing discussion, and current-head CI. I generated the actual runtime on this head and ran 17 Node checks covering strings, ints, non-integral floats, bool exactness, list/dict/set/bytes/None, null-prototype dicts, archetypes/subclasses, a constructor-named class, callable type aliases, and type repr; all passed. `git diff --check` passed.

P2: `_typename` is a lossy dispatch category, so using it as exact type identity introduces wrong positive matches. A compiler-generated Jac probe with `t: tuple[int, int] = (1, 2)` and `f: float = 1.0` returns `[false, true, false, true, false]` for `[type(t) is tuple, type(t) is list, type(f) is float, type(f) is int, type(str) is type]`; the expected result is `[true, false, true, false, true]`. The emitted JS keeps t as an array and f as a JS number, then calls this helper; it cannot recover the erased distinctions from their JS shapes. Builtin type objects also fall through to Function rather than the type singleton. In particular, the tuple-is-list and float-is-int cases now take the wrong branch, despite the exact-type claim.

Please preserve the information needed for exact type identity, recognize builtin type objects, and add these positive/negative compiler-runtime cases to the equivalence coverage. The existing six rows do not cover them. Current-head CI is green (25 successful checks, 3 skips), and GitHub reports no conflicts, but these direct production-compiler/Node results prevent approval. No source changes were made and no full local suite/browser run was performed. Changes requested for exact type correctness.

kugesan1105 (commented):
Good call on the root cause. Redirecting `type()` to the same singleton a bare type name lowers to is the right layer, and the `__name__`, `_isPlainObject` and `NoneType` additions are real improvements on top. I ran a 12-shape probe against the branch and all 12 match Python, including `str(type(x))` and the archetype case.

Three things I'd like sorted before merge.

### 1. The equivalence rows don't enforce anything

`test_prim_equivalence.jac` isn't in the diff, so the `builtins introspection` row still has no `require=["cl"]`. `run_with_both` wraps the cl lane, `compare_results` included, in `except Exception { if "cl" in reqd { raise; } }`, so a mismatch is swallowed rather than failed.

I checked by breaking the fix on the branch and running the test:

- revert `emit_type` to `"typeof " + args[0]` gives `1 passed`
- make `_jac.builtin.type` report bool as int gives `1 passed`

The second is the case the description says `type_bool_isnot_int` pins. Your shape assertions do catch an emitter revert, so it's specifically the semantic half that's dead. `require=["cl"]` plus an `allow_missing` waiver for the py-only rows would make it real. I hit exactly this on my own attempt and ended up deleting the rows.

### 2. `type(t) is list` on a tuple now returns true

Measured on the branch:

| expr | Python | branch |
|---|---|---|
| `type(t) is list` | False | **true** |
| `type(t) is tuple` | True | false |
| `type(str) is type` | True | false |
| `type(3.0) is float` | True | false |

The first is a new false positive: the old always-false behaviour happened to match there. It's inherent to tuples being arrays, my own attempt had it too, and since `isinstance` folds tuple to list the runtime is now inconsistent with itself. I'd drop "Exact-type semantics are preserved" and list these as known gaps instead.

### 3. `ctor.__name__ = ctor.name` mutates the caller's class

`type()` writes to the class object as a side effect. On a frozen constructor it throws:

```
TypeError: Attempting to define property on object that is not extensible
```

and on a normal class `hasOwnProperty("__name__")` flips false to true after the first `type()` call.

SupulHeshan:
> Good call on the root cause. Redirecting `type()` to the same singleton a bare type name lowers to is the right layer, and the `__name__`, `_isPlainObject` and `NoneType` additions are real improvements on top. I ran a 12-shape probe against the branch and all 12 match Python, including `str(type(x))` and the archetype case.
> 
> Three things I'd like sorted before merge.
> 
> ### 1. The equivalence rows don't enforce anything
> `test_prim_equivalence.jac` isn't in the diff, so the `builtins introspection` row still has no `require=["cl"]`. `run_with_both` wraps the cl lane, `compare_results` included, in `except Exception { if "cl" in reqd { raise; } }`, so a mismatch is swallowed rather than failed.
> 
> I checked by breaking the fix on the branch and running the test:
> 
> * revert `emit_type` to `"typeof " + args[0]` gives `1 passed`
> * make `_jac.builtin.type` report bool as int gives `1 passed`
> 
> The second is the case the description says `type_bool_isnot_int` pins. Your shape assertions do catch an emitter revert, so it's specifically the semantic half that's dead. `require=["cl"]` plus an `allow_missing` waiver for the py-only rows would make it real. I hit exactly this on my own attempt and ended up deleting the rows.
> 
> ### 2. `type(t) is list` on a tuple now returns true
> Measured on the branch:
> 
> expr	Python	branch
> `type(t) is list`	False	**true**
> `type(t) is tuple`	True	false
> `type(str) is type`	True	false
> `type(3.0) is float`	True	false
> The first is a new false positive: the old always-false behaviour happened to match there. It's inherent to tuples being arrays, my own attempt had it too, and since `isinstance` folds tuple to list the runtime is now inconsistent with itself. I'd drop "Exact-type semantics are preserved" and list these as known gaps instead.
> 
> ### 3. `ctor.__name__ = ctor.name` mutates the caller's class
> `type()` writes to the class object as a side effect. On a frozen constructor it throws:
> 
> ```
> TypeError: Attempting to define property on object that is not extensible
> ```
> 
> and on a normal class `hasOwnProperty("__name__")` flips false to true after the first `type()` call.

Addressed all three - require=["cl"] + a waiver now enforces the cl lane, `type()` no longer mutates (a static __name__ field at declaration time instead of a lazy write), and the tuple/float gaps are pinned as known issues instead of implied-fixed. type(str) is type had the same root cause and is fixed too.

jac/jaclang/compiler/backends/es/jac_runtime_js.jac
SupulHeshan: Good catch,
Tuple/float gaps are intentional, pinned with a regression test rather than fixed, since JS can't represent that distinction. type(str) is type is already fixed though

Issue #8897, opened 2026-09-03
scale tests, phase 0 of #8896: one harness, strict-aware gates, no root-level files, ratchets seeded (enablers only, no test-count change)


kugesan1105:
Part of #8896. Measured on main at ad900159f, 2026-09-03. This phase adds and moves; it deletes exactly one file (`scale_test_client.jac`) and no tests. Every later phase depends on it: phase 1 deletes against the survivors these gates make honest, phase 2 collapses onto the helpers this phase makes shared.

## What this phase does, in one table

| Kind | What | Count today | After |
|---|---|---|---|
| Add | `jac/jaclang/scale/tests/support.jac` | 3 support modules + ~30 local copies | 1 module |
| Add | `jaclang.testing.requires` gates | 1 file uses them | every rig-dependent file |
| Add | mock-importer ratchet test | none | allowlist seeded at 33 files |
| Add | PR validation-section check | none | seeded permissive |
| Modify | helper call sites | ~30 files | import from `support` |
| Modify | root `jac.toml` `[test]` marks, ci.yml `test-scale` matrix and `test-scale-k8s` paths | 6 groups + `--ignore` list | 5 groups |
| Move | 11 root-level test files | `scale/tests/test_*.jac` | their owning directory |
| Delete | `scale_test_client.jac` (226 lines, 15 importers) | 1 | 0, callers on `jaclang.testing.JacTestClient` |
| Fix | asserts that accept both outcomes | 6 | 0 |

Test blocks before and after: 1,400 and 1,400. The gate for the phase is that count plus green on every current leg.

## 0a. One harness: `support.jac`

Today three support modules (`server_support.jac` 178 lines, `auth_test_helpers.jac` 43, `scale_test_client.jac` 226) coexist with local reimplementations in the test files. Everything below moves into one `scale/tests/support.jac` in the shape of `jac/tests/support.jac` (#8599). The PR body carries an exception table naming every file that keeps a local helper and why, or states there are none.

| Helper | Copies today (file:line) | Survivor in `support.jac` |
|---|---|---|
[...]

Issue #8896, opened 2026-09-03
scale: the test tree is the size of the code it tests, a quarter of it mocks the collaborator, and every fix ships a new file; consolidate onto journeys + one suite per subsystem, and gate the next 1,400 at review


kugesan1105:
Measurements are from 2026-09-03 against main at ad900159f. Counts are test blocks (`grep -E '^\s*test\s+'`), table-driven cases counted once.

## Summary

`jac/jaclang/scale/tests` holds 1,400 test blocks in 134 files, 37.5k lines against 41.2k lines of scale source. 33 of those files import `unittest.mock` or `MagicMock`, against the ban TESTING.md §2.2 has carried since #8654. One file uses `parametrize`; the compiler and runtime suites use it for every repetitive shape since #8599. Since May the tree has grown by 212 merges and almost nothing was ever deleted: two of the seven regular contributors have deleted zero tests in four months, and a fix PR ships 18 to 36 tests in a new file as a matter of habit.

The suite is not redundant by content so much as by altitude and by file: the same behaviour is pinned against the object and again over HTTP, per case; the same provider matrix is hand-copied four times; the same KEDA happy path runs under a MagicMock Kubernetes client, on MicroK8s, and on kind. Meanwhile the local fleet boot path, meaning `MicroserviceOrchestrator`, `ServiceProcessManager`, `ProcessGateway`, `LocalRealizer` and the `jac scale gateway` verb every deployed gateway pod execs, is named by zero test files.

This issue proposes finishing what #7628 and #8654 started: make the trophy (TESTING.md §2.1) the structure of the tree rather than a paragraph in a doc, and make the per-PR rules a reviewer can apply in thirty seconds. Governing method is the one #8599 and #8706 used: audit first, no assertion lost without a named survivor, enablers before deletions, deletion inventory at line level, acceptance criteria as greps and counts.

## Motivation: five things are wrong, and only one of them is "too many tests"

[...]

PR #8927, opened 2026-09-03
test(scale): six asserts that accepted both outcomes now expect the one the server gives (#8897 phase 0d)

kugesan1105:
## What this changes

Part of #8896, phase 0d of #8897, after #8902, #8908, #8911, #8916 and #8925.

Of the 30 `status_code in (...)` asserts in the scale suites, most pair statuses that are equivalent for the test: 400/422 on bad input, 404/405 on an unrouted verb, 200/201 on create. Six accept both the enforced and the unenforced outcome, or success and failure, so they pass whatever the server does. Each becomes one expected status, chosen by running the endpoint and reading what it actually returns, not by guessing.

| Site | Was | Server returns | Now |
|---|---|---|---|
| `server/test_file_upload.jac` "file upload requires auth" | `in (401, 200)` | `401`, `{"error": {"code": "UNAUTHORIZED"}}` | `== 401`, and the error code is `UNAUTHORIZED` |
| `server/test_file_upload.jac` "upload missing required file" | `in (422, 400, 200)` | `422`, `{"detail": [{"type": "missing", "loc": ["file", "myfile"]}]}` | `== 422`, and `myfile` is among the missing fields |
| `server/test_admin.jac` "admin graph requires auth" | `in [401, 403, 422]` | `403`, `Admin access required` | `== 403` |
| `server/test_admin.jac` "admin graph data" | `in [200, 404, 422]` | `200`, a graph with `nodes` and `edges` | `== 200`, and the body has `nodes` |
| `server/test_admin.jac` "admin index page" | `in [200, 503]` | `200`, `text/html` | `== 200`, content type asserted unconditionally |
| `server/test_serve.jac` static bundle | `in [200, 503, 500]` | see below | `== 200`, content type asserted unconditionally |

Five of the six also carried an `if status == 200 { ... }` block, so the real assertions only ran on the path the test hoped for. Those blocks are gone; the assertions inside them run every time.

**Two of the issue's guesses were wrong, and the probe caught them.** The table in #8897 said the admin graph without credentials should be "401 or 403, not 422": it is unconditionally 403, at `admin/impl/admin_portal.impl.jac` in `admin_graph_data`, whether the header is missing or invalid. And it proposed a `require_*` gate for the admin index in case 503 meant "rig absent": the 503 arm is the admin client failing to build, which is a defect to surface, not a rig to skip on. CI's scale lane sets up Bun, so the build runs there.

## The sixth site exposed a real regression, now fixed separately in #8933

`GET /static/client.js` answers:

```
503  Vite build failed ...
node_modules/zod/v4/core/util.js (330:0): A comment ... contains an annotation that Rollup cannot interpret
```

The old `in [200, 503, 500]` accepted that, so `test_serve` has been green with a broken client build, on CI as well as here. The same build failure is why the admin index test also fails on CI: `build_admin_client` runs the same bundler and answers 503 when it fails.

Cause, isolated on a reproducing tree by holding each variable in turn: zod 4.5.0 through 4.5.4 (published 2026-08-28/29) place `@__PURE__` annotations Rollup rejects, and the client dependency table asked for `^4.3.6`, so a fresh install floats to 4.5.4. Rollup's version does not change the outcome. **#8933 pins zod to `~4.4.3`.** This PR stays red on `test-scale (server)` until that lands; rebased onto it, these two asserts are the end-to-end confirmation that the pin works, which is the point of tightening them.

## Validation

Rig per TESTING.md part 1: `zig build -Ddev` from this checkout, capability deps re-pinned, native compiler kernel built in-tree. `jac --version` prints the dev banner naming the checkout.

Statuses were measured, not read off the code: each of the six asserts was temporarily replaced with one that fails and prints `status` and the first 160 bytes of the body, the three files were run, and the values above are what came back. The probes were then reverted and the real asserts written.

The three files, one per process under `JAC_TEST_STRICT=1` the way CI runs them:

| Suite | Result |
|---|---|
| `server/test_admin.jac` | 29 passed |
| `server/test_file_upload.jac` | 9 passed |
| `server/test_serve.jac` | 30 passed, 3 not passing (below) |

The three `test_serve` non-passes on this machine:

| Test | Cause | Related to this PR? |
|---|---|---|
| `static and schema endpoints` | the tightened assert: `Vite build failed ... zod/v4/core/regexes.js (70:0): A comment` | **yes, by design**: it is the hidden failure described above, now visible |
| `websocket functionality` | `WriteConflict` on the root anchor | no, pre-existing, recorded on #8916 |
| `server startup - a second server refuses an already-taken explicit port` | subprocess `TimeoutExpired` | no, pre-existing, fails at `c386bbbcb` too |

Before this change the same file reported `30 passed, 2 failed, 1 error` on this machine, the error being the `AnalysisLedger` regression #8918 has since fixed. So the only new non-pass is the one this PR is supposed to produce when the client build is broken.

`jac fmt --check --lintfix` and `jac check` clean on all three changed files.

**Not run:** every other scale suite. This change touches three files and no shared helper.

## What this deliberately leaves alone

The other 24 status-set asserts pair statuses that are equivalent for what each test claims: `(400, 422)` on malformed input, `[404, 405]` on an unrouted verb, `(200, 201)` on create, `[401, 403]` on denied. Collapsing those to one value is worth doing where the server is deterministic, but it is a per-endpoint decision and the loose form does not hide a failure the way these six did.


MusabMahmoodh (approved):
This is the right way to kill that class. An assert that accepts both the enforced and the unenforced outcome is not a weak test, it is an absent one, and the six you picked out are the ones where that distinction actually mattered. Deleting the `if status == 200` blocks is the better half of the change: those made the real assertions conditional on the outcome the test was hoping for, so the interesting path was the one that never got checked.

The part worth calling out is that the sixth site paid for the whole exercise. `test_serve` has been green on CI against a client bundle that does not build, because `in [200, 503, 500]` covered it, and the only reason anybody knows is that you went and read what the server actually returns instead of trusting the pair. Isolating it to zod 4.5.x placing `@__PURE__` where Rollup will not take it, and splitting the pin into #8933 rather than smuggling it in here, is the correct shape for both PRs. Correcting two of #8897's own guesses from the probe rather than from the table is the same discipline.

One thing I would like settled before this merges, and it connects to #8925.

`/static/client.js` and `/admin/` now hard-assert 200, so both tests depend on a working Bun and Vite toolchain wherever the scale server suite runs. I agree with your call that a 503 there is a defect to surface and not a rig to skip on. But that is exactly the situation #8925 just built the vocabulary for: a rig requirement that is declared, so a missing one stops passing green. Right now a machine without Bun gets these two failing as though the product broke, which is the same misattribution in the other direction from the one you are removing.

Do these two declare the client-build requirement through #8925's mechanism? If they do not yet, they are the strongest candidates in the suite for it, and it is the difference between "your environment is missing something" and "the bundler is broken" at the moment somebody reads the failure.

Merge order is fine as you describe it: this stays red on `test-scale (server)` until #8933 lands, and rebased on top these two asserts are what proves the pin. Worth saying on #8933 that it unblocks this one, so nobody merges them in the wrong order.

christianwilkins (commented):
Reviewed head `563bfd04afb0e2f4bdf18c43585caf9cab0206d4`: all six assertion changes, the surrounding requests/fixtures and admin handlers, existing review, and current-head CI. I found no new defect in the tightened assertions and am not duplicating Musab's toolchain-requirement feedback.

The remaining CI failure needs a more specific handoff than the PR description currently gives. The [scale-server job](https://github.com/jaseci-labs/jac/actions/runs/33901342092/job/101119260410) reports 340 passed and 2 failed: `admin index page` and `static and schema endpoints`. The admin build's fatal Rollup error is `"getErrorMessage" is not exported by "compiled/utils/api.js", imported by "compiled/context/AuthContext.js"`. The preceding Zod annotation messages explicitly say the comments will be removed; they are warnings, so this log does not establish that a Zod pin alone fixes that build. The static-bundle assertion truncates its response at 400 characters, so its fatal cause is not visible in that test output.

Also, #8933 is now closed without merging. Please update the dependency/merge-order text, identify or integrate the actual client-build fix, and rerun these tests on the rebased head while keeping the strict assertions. This review uses the CI execution evidence; I did not independently rerun the browser/build-dependent suites or change source.

Not ready to merge while the current-head scale-server check is red. The author owns the rebase/dependency correction and validation; maintainer merge follows after green CI.

PR #8931, opened 2026-09-03
fix(native): a lambda's inferred parameter type is used, not defaulted to i64 (#7139)

SupulHeshan:
## Summary

`_codegen_lambda` treated a missing `type_tag` as "no type at all" and defaulted straight to `i64` for a lambda's parameters. That shortcut bypassed `_resolve_jac_type` (and its `_emit_unlowerable_type` backstop) entirely.

A lambda's parameters are exempt from the "every parameter needs a `type_tag`" rule (E0052) precisely because the central type evaluator infers them from context instead — a typed callback slot, `Callable[...]`, etc. — and stamps the real type onto `param.type`, or refuses with E1119 if it can't.

So a lambda passed into a `Callable[[str], str]` slot compiled **clean** (E1119 doesn't fire — inference succeeded) and silently laid out its `str` parameter as `i64` in the emitted LLVM IR: no diagnostic, wrong ABI, a string's pointer bit-reinterpreted as an integer at the call boundary.

## Fix

- The parameter loop now calls `_resolve_jac_type` unconditionally instead of special-casing `type_tag is None` to `i64`. It already reads the inferred `p.type` when there's no syntactic annotation, and correctly routes a genuinely-unresolved type through the existing `E5092` backstop — so this reuses that machinery rather than adding new logic.

**Update:** an earlier revision of this PR also rewrote the return-type branch, but [review](https://github.com/jaseci-labs/jac/pull/8931#issuecomment-5662294975) found that hunk was a no-op — the lambda's return type was already being inferred correctly via the pre-existing `isinstance(nd.type, SemanticFunctionType)` block, which the rewrite just reimplemented (worse: behind a `try { import ... } except ImportError {}` that would have silently kept `i64` if that guaranteed-to-succeed import ever failed — the exact failure mode this PR removes). Confirmed identical emitted IR with the rewrite reverted and the new test still green, so it's been dropped. This PR is now a true one-line fix, scoped to the parameter case.

## Scope

Issue #7139 flagged several `_resolve_jac_type` callers with the same "no annotation → i64" shape (free-function params/returns, object fields, closures, generics monomorphization, clib interop). Most of those have since been closed by other work (#7418, #7576), and I audited the remaining signature-position call sites sharing this exact pattern. All of them cover *named* parameters, which `ast_validation_pass` already requires a `type_tag` for (`E0052`) — so an untyped `ParamVar` can only reach native codegen through a lambda. `_codegen_lambda`'s parameter loop is the one live instance of the pattern; I didn't touch the other sites since I couldn't write a failing test against them (they're unreachable with a `type_tag`-less `ParamVar` in valid source).

Review also surfaced a related, separate gap this PR does **not** cover: a function with no return type at all (no `-> T`, nothing inferred either) still lowers as `i64` instead of representing "no value" — e.g. `def f { print("hi"); }` then `f() is None` is `False` when it should be `True`. That's a distinct issue from the lambda-parameter bug fixed here, so I'm leaving #7139 open for it rather than closing it with this PR.

## Testing

Extended the existing `lambdas` test in `test_native_gen_pass.jac` (over `fixtures/lambdas.jac`) rather than adding a new one — that's the suite's actual sibling for lambda codegen: one test, one JIT-executed helper function per lambda shape. Added a Phase 3 case to the fixture (an untyped param whose type is inferred from a `Callable[[str], str]` slot) and one more helper that calls it through the existing `get_sfunc` string-ABI helper, asserting on the real return value (`b"hello"`) rather than pattern-matching the emitted IR text — a behavioral proof, not an implementation-detail one, using the suite's own conventions.

Confirmed red against the pre-fix codegen first: calling the miscompiled lambda didn't just return the wrong value, it **segfaulted** — a string's descriptor genuinely can't survive being laid out as a bare `i64`. Green after the fix, and still green after dropping the no-op return-type hunk.

Net test count: `test_native_error_gate.jac` (where I'd originally added a standalone IR-inspection test) is back to 5 — unchanged from `main`; `test_native_gen_pass.jac`'s `lambdas` test stays 1 test with one more internal case. No new top-level test added anywhere.

Also ran the full targeted regression scope: every native test file whose fixtures exercise lambda syntax (`test_native_closures`, `test_native_generics`, `test_native_entry_globals`, `test_native_generators`, `test_native_py_semantics`, `test_native_ternary_union`, plus `lambda_multi_stmt_callback` in `test_native_gen_pass`) — 103 tests, all passing, no regressions.

## Note on `--no-verify`

`jaclang/compiler/backends/native/na_ir_gen/expr.impl.jac` already fails `jac check` on `main` with 48 pre-existing, unrelated errors (verified via `git stash` before/after this diff). This diff's own lines add zero new ones. `pre-commit.ci` doesn't run `jac check` at all (markdown lint + em-dash ban only), so I bypassed the local `check` gate for this pre-existing, out-of-scope debt rather than attempt an unrelated 47-error fix blind.

## Related

Follow-up to #7576, which closes a different (already-dead) fallthrough in the same `_resolve_jac_type` family. Left a note there pointing back to this PR.

Addresses the lambda-parameter portion of #7139; the no-return-type i64-vs-void gap flagged in review stays open under that issue.


christianwilkins (commented):
Reviewed head `dcef981fb6a1d4c3be239dce790fa8de2167184a`.

I inspected the complete final patch, `_resolve_jac_type`, the lambda return lowering and closure ABI path, and the existing lambda fixture plus string-ABI test helper. No new blocking code finding in the inferred-type change. The final follow-up correctly moves the regression into the existing `lambdas` behavioral test rather than adding a separate IR-only test.

Validation is incomplete on this head. The current-head workflow-run query returns no CI runs; the only check run is skipped, and pre-commit reports `error during mergeable check`. GitHub reports conflicts. My independent `test_native_gen_pass.jac -f lambdas` attempt did not finish after 15 minutes and was stopped, so I am not claiming the string-ABI case passed locally or applying results from an older commit. No source changes made.

Not ready to merge: the author needs to resolve the conflicts and rerun the lambda/string-ABI regression plus current-head CI. A maintainer may need to approve the resulting workflow if repository policy gates it. Re-review the resulting head before merge.

PR #8942, opened 2026-09-04
ci: the jac-kit action checks its download against a manifest sealed into the kit, and fetches again on a miss

marsninja:
## What failed on main

The CI run for 8081ec800 ([run 33816389635](https://github.com/jaseci-labs/jaseci/actions/runs/33816389635)) failed in test-jac-pack-smoke and test-scale (microservices, data, server). All four died in the jac-kit action's warm-up step with:

```
jac (launcher): no runtime payload trailer (this is not a bundled jac binary)
```

The kit itself was whole: the same artifact (ID 9917157116) ran in every other lane of that run, and build-kit had already smoked the binary with its 110 MiB payload. In the four failing jobs `actions/download-artifact@v4` took about 70s instead of about 7s, never printed its "SHA256 digest of downloaded artifact" / "Artifact download completed successfully" lines, and still ended with outcome=success. The blob stream stalled mid-extraction, the step exited 0, and the jobs ran a jac binary cut short of its payload trailer.

(The Release run that failed at the same commit was a re-dispatch of an already-tagged v0.37.3 and passed on the next dispatch; nothing to fix there.)

## The fix

- build-kit (ci.yml) and build-kit-macos (_macos-native.yml) seal a manifest into the kit: `zig-out/kit.sha256` hashes every kit file, and `zig-out/kit.sha256.sha256` hashes the manifest, so a manifest cut short by the same truncation cannot vouch for the files after it.
- The jac-kit action checks its download against the manifest right after `download-artifact`, strictly (a short, missing, or malformed line fails). On a miss it warns, fetches the artifact once more, and checks again, so a stalled download costs a retry instead of the job.
- `.github/actions/jac-kit/manifest.sh` holds both sides (`write` / `check`). It uses `sha256sum` where present and `shasum -a 256` otherwise, so the macOS lane's kit gets the same treatment.

## Verified

- manifest.sh exercised locally against a fake kit tree: a clean check passes; a truncated binary, a missing file, a manifest cut at a line boundary, a missing manifest, and a missing manifest-of-manifest each fail with exit 1; writing with one hash tool and checking with the other agrees.
- The three YAML files parse; shellcheck is clean on the script; actionlint reports nothing new in the changed ranges.

### DIFF
--- a/.github/actions/jac-kit/action.yml
+++ b/.github/actions/jac-kit/action.yml
@@ -1,11 +1,12 @@
 name: Get jac kit
 description: >
   Download the kit artifact into the checkout (binary, typeshed stubs, vendored
-  musl/wasm runtimes land at their in-tree paths), put jac on PATH, and warm
-  its runtime cache. The kit is the binary built from this checkout
-  (build-kit); the job's JAC_NO_DEV_SOURCE decides whether it runs sealed
-  ("1") or through the [dev] reroute (""). upload-artifact strips the
-  executable bit; restore it.
+  musl/wasm runtimes land at their in-tree paths), check it against the
+  manifest build-kit sealed into it, put jac on PATH, and warm its runtime
+  cache. The kit is the binary built from this checkout (build-kit); the
+  job's JAC_NO_DEV_SOURCE decides whether it runs sealed ("1") or through
+  the [dev] reroute (""). upload-artifact strips the executable bit; restore
+  it.
 inputs:
   name:
     description: Kit artifact name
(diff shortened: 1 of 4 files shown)


christianwilkins (approved):
Reviewed head `3989a6d30d1de42a67547e4208666c554cd94d0e`.

No blocking finding in the complete patch. Both kit producers include the manifest and its digest, and the consumer retries once before a mandatory second check. I also read Jason's note about the first-attempt runtime failure and checked the current results rather than treating that attempt as the final CI state.

Validation: eight local manifest cases behaved as expected (intact kit, whitespace in filenames, truncated binary, missing file, truncated manifest, missing manifest/digest, malformed line with a valid manifest digest, and restored-download success; the intact and whitespace checks share one case). Cross-checking a `sha256sum` manifest with `shasum`, and the reverse, passed; `shasum` also rejected a corrupted payload. `bash -n` passed. Current-head CI is green (25 successful checks, 3 skipped); GitHub reports no conflicts. No source changes made during review.

Approved and ready to merge. No further author action identified on this head.

PR #8944, opened 2026-09-04
fix(scale): seal the app's installed site-packages into the .jab so scale-to-zero wakes skip the boot-time jac install

kugesan1105:
Addresses the dominant cost in #8695.

## Problem

A KEDA scale-to-zero wake pays ~44s before the first byte, and 21s of that is the `jac-bootstrap` init container running `jac install`: `/app` is an `emptyDir`, so every pod rebuilds `/app/.jac/venv` from the bundle's own vendored wheels, producing an identical result every time. Measured locally, nearly half of that install is not even installing: creating an empty venv plus ensurepip (~2s), a `pip install --upgrade pip setuptools wheel` seed that always fails under `PIP_NO_INDEX=1` because those wheels are never vendored (~1.4s, error swallowed unless verbose), and two `pip list --format=json` snapshots that exist only to pretty-print names into an init log (~0.4s). Only ~4.3s of a 9.5s local run installs the app's packages.

The dependency closure is fully determined at seal time: the fat bundle already vendors pod-platform wheels and ships the exact CPython the pod runs. Rebuilding the venv per wake does that fixed work at the worst possible moment, per wake, in front of a waiting user.

## Fix

Build the venv once at seal time and ship it inside the `.jab`:

- `pack_jab` runs the seal binary's own `jac install` inside the staging tree, offline against the freshly vendored wheels: the exact code path pods run at boot today, so there is no second install implementation to keep in sync. What ships is deliberately NOT a full venv: only `lib/*/site-packages` (the importable packages, the one part the runtime actually uses) plus a `.jac-deps-hash` marker, placed under `.jac/venv/` so the existing untar lands it where the runtime already looks. `bin/`, `pyvenv.cfg`, the interpreter link, seeded pip, and `__pycache__`/`.pyc` are all excluded on purpose: pods run the toolchain's own Python, and those pieces would either be dead weight or break bundle-digest determinism (shebangs and pyc embed per-seal paths and timestamps).
- The pod side needs zero bootstrap-script changes: the existing `jac-pvc-bootstrap` untar populates `/app/.jac/venv` (one streaming read of the bundle, the same access pattern that already makes the 20MB bundle untar cost ~2s on EFS, unlike per-file copies which EFS punishes).
- `install_specs` returns early when the marker matches the resolved specs. The check runs before `ensure_venv`, which would otherwise classify the interpreter-less sealed venv as corrupted and delete it. Only `pack_jab` ever writes the marker, so local `jac install` behavior is completely unchanged (repeated installs still run pip, still repair a hand-broken venv).
- The payload stays deterministic, guarded by the existing bundle-digest determinism test: `bin/` (shebangs embed the staging path), `pyvenv.cfg`, `__pycache__`/`*.pyc`, and ensurepip's seeded pip never ship, and RECORD lines pointing outside site-packages are dropped.
- Failure is graceful and honest: a host that cannot install the pod-platform wheels (cross-arch deploys, thin bundles, old glibc) logs a warning and pods fall back to today's boot-time install. `JAC_SEAL_SKIP_VENV=1` opts out explicitly.

Architecture mismatch is handled the same way the wheels already handle it: the venv is assembled from the pod-platform wheels the bundle vendors, and building it is data assembly (unzip plus text entry points), never execution of pod binaries.

## Measured

- Pod-wake simulation (extract a real sealed `.jab` into a fresh dir, run the bootstrap's exact `jac install` env): **0.48s**, marker fast path taken, venv untouched (no recreate), `import` of the sealed dependency works at runtime.
- Same app without the fix, same offline env, fresh dir: **9.4 to 10.5s** locally, which scales to the 21s measured on the cluster in #8695.
- Warm-venv re-run without any marker (main today): 1.05s, which is why the marker check is a fast path and not new install machinery.

Expected effect on #8695's wake: the 21s install phase becomes a ~1s verification plus a few extra seconds of bundle untar, roughly halving the wake. Cold and warm wakes are identical by construction: there is no cache to miss, no first-wake penalty, and the venv can never drift from the sealed bundle because it is content-addressed with it.

## Drive-by fix the e2e surfaced

`vendor_wheels.jac` resolved the project config via discovery, and `jac run <script>` roots the explicit config at the script's own tree. Any seal running from a dev source tree therefore vendored the jac checkout's closure (byLLM deps, empty dependencies) instead of the app's, silently. It now loads the staged `jac.toml` directly. The existing fat-bundle tests never caught this because they assert wheels exist, not which wheels.

## Tests

Counts before/after: `test_dependencies.jac` 83 -> 85 (+2, both table-driven); `test_fat_bundle.jac` 6 -> 9 (+3, reusing its existing `_app`/`_names`/`_member`/`_pod_binary` helpers). No new test files.

Justification for the additions, since this changes behavior rather than fixing a regression an existing test already reaches:

- The marker fast path is a genuinely new code path in `install_specs` with no existing coverage (no existing test exercises a venv that must NOT be probed by `ensure_venv`). Two table-driven tests in `test_dependencies.jac` cover it: a `venv_deps_satisfied` truth table (5 rows: order independence, mismatch, subset, missing marker, missing venv) and an `install_specs` behavior table (4 rows: match skips pip and `ensure_venv`, stale runs pip, dry-run ignores a match, a local install never writes the marker). They reuse the file's existing `make_temp_project`/`reset_config` fixtures and the same `patch` pattern as the sibling `install_specs` tests.
- The three `test_fat_bundle.jac` additions cover the new seal-time staging subsystem: payload selection (marker plus site-packages, never `.pyc`/`__pycache__`/`bin/`/seeded pip), RECORD normalization and its idempotence, and one `pack_jab` e2e that packs twice (with `JAC_SEAL_SKIP_VENV=1`: wheels but no venv; without: the full venv contract). It stays an e2e because both behaviors only manifest through `pack_jab`; it runs in-process, with subprocesses only for the seal children `pack_jab` always spawns. The existing digest-determinism test in `test_fat_bundle.jac` now also guards the sealed venv for free.
- Suites verified green after the rebase: 85 dependency tests, 4 seal-venv, 6 fat-bundle.

## Limitations

- A bundle with git dependencies installs them at seal time (an improvement over cloning at boot), but their content varies per seal, so such bundles lose digest stability. Pinned pypi closures are unaffected.
- `.pyc` files are not shipped (they embed timestamps and would break digest determinism), so first import in a pod compiles bytecode lazily; the shipped CPython matches the seal's, so this is a small one-time cost per pod.
- Pods running a pre-fix toolchain binary ignore the marker and rebuild from the wheels as today: wasted bundle bytes, correct behavior.


MusabMahmoodh (commented):
Read this against the surrounding code rather than the diff alone. The shape is right and it is the better mechanism: the closure is fixed at seal time, so rebuilding it per wake was always the wrong moment, and building it by running the seal binary's own `jac install` inside the stage means there is no second install implementation to drift. `staged_venv_files` excluding `bin/`, `pyvenv.cfg`, `__pycache__` and `*.pyc` is what keeps the digest determinism test honest, and `_strip_external_record_lines` having an idempotence assertion is the right test to have written. No mocks anywhere, two real `pack_jab` runs, and the marker tests fail on main by reading, since `venv_deps_satisfied` does not exist there.

Four things, ranked.

**1. The marker hashes the specs and nothing about the interpreter, and the sealed venv cannot declare one.**

`_venv_deps_hash` is sha256 over sorted specs only. `pyvenv.cfg` deliberately does not ship, so on the pod `_venv_python_version_tag` returns `""` and `get_venv_site_packages` falls through to `_discover_venv_site_packages`, which globs `lib/python*/site-packages` and returns the highest-sorted name (`pyvenv.jac:298-317`). Nothing compares that against the interpreter doing the importing. So a venv sealed under CPython 3.14 is inserted onto `sys.path` by a pod running 3.13, `install_specs` returns True on the marker, and the first C extension import fails at request time instead of at boot.

Right now the seal binary and the pod binary are the same object, so this holds by the call site rather than by construction. `pack_jab` takes `seal_binary` as a parameter, and both the release-vs-experimental injector path and pinned-jac hosting (jacBuilder#1612) choose it independently of the host running the seal. `_fake_staged_venv` hardcoding `lib/python3.14/site-packages` bakes the same assumption into the test.

The cheap version is to make a mismatch a marker miss, which then falls into `ensure_venv` and rebuilds from the wheels the bundle still carries, which is the fallback you already have:

```jac
def _venv_deps_hash(specs: list[str], py_tag: str = "") -> str {
    import hashlib;
    payload = "\n".join(sorted(specs)) + f"\n{py_tag}";
    return hashlib.sha256(payload.encode("utf-8")).hexdigest();
}
```

with the tag written from the staged venv's `lib/python*` dir at seal time and recomputed from `sys.version_info` on read.

**2. `is_cache_eligible` never fires today, and this PR makes #8734's cache dead weight either way.**

#8734 gates the whole PVC venv cache on `is_cache_eligible(base_image)`, which is `"@sha256:" in base_image` (`venv_cache.jac`). Deployed images are mutable tags, so that returns False and the cache never runs. Udith measured exactly that on jacBuilder#1642 while trying to explain why cold and warm wakes kept matching.

If this lands there is no boot-time install left to cache, so the restore/publish pair is doing nothing at best. `restore` is also `shutil.copytree`, the per-file access pattern your own body says EFS punishes, against the streaming untar you are replacing it with. Two mechanisms for provisioning one venv is a second source of truth, so it would be better to settle which one survives before either merges rather than after. My read is that this one wins and #8734 should keep the startup-probe tightening and drop `venv_cache.jac`, but that is Linus's call to make with the numbers in front of him, and #8946 stacking both is the place to get them.

**3. The bundle now ships the wheels and the unpacked venv, and a latency PR reports no size number.**

`_vendor/wheels` is untouched, so a fat bundle carries the closure twice. That is defensible as the marker-miss fallback, but the whole bundle is pulled and untarred on every wake, and the body measures the install phase only. What does the `.jab` weigh before and after for the app in #8695, and what does the extra untar cost on EFS? If the doubled payload eats a meaningful slice of the 21s you win back, keeping the wheels only when `JAC_SEAL_SKIP_VENV` is set is the obvious trade.

**4. The degradation path is correct and close to invisible.**

`except Exception` into `logger.warning` in `pack_jab` is the right call for a cross-arch or thin-bundle host, and rmtree'ing `.jac` first is what keeps a half-built venv from shipping. My worry is only that a deploy that quietly lost the optimization looks exactly like a deploy that kept it, and the regression it hands back is the 21s this PR exists to remove. The seal already prints a summary line for the wheels and one for the venv, so the failure deserves a line in the same place rather than only in the logger.

Nothing here blocks the approach, and 1 is the only one I would want fixed before merge.

MusabMahmoodh (commented):
Following my own point 2 with something more concrete than "settle it before either merges", because I went and read #8734's current head afterwards and the overlap is textual, not just conceptual.

#8734 also declares `glob VENV_DEPS_MARKER: str = ".jac-deps-hash"` in `jaclang/project/providers.jac`, also adds a `_venv_deps_hash` over sorted specs, and also inserts an early return into `install_specs` immediately above `ensure_venv`. Same constant, same filename on disk, same file, same insertion point. Whichever of the two lands second will not merge, and a mechanical conflict resolution leaves two hash functions writing one marker.

It also answers my point 1 for you. Linus's version takes an `image_identity` and folds it into the payload, with `InstallContext` carrying it, precisely so a venv cannot be trusted across a base-image change. That is the identity check this PR is missing, already written and already tested on his branch.

So the merge I would want is your seal-time build with his marker: the `.jab` ships the assembled venv, and specs-plus-image-identity decides whether the pod trusts it. That also retires `venv_cache.jac` cleanly, since with no boot-time install there is nothing left for restore/publish to cache.

PR #8950, opened 2026-09-04
fix(server): a walker response no longer repeats its reports inside result (#8906)

### LINKED ISSUE #8906: Every walker response serializes `reports` twice; no consumer reads the copy inside `result`
Every walker response carries the report list twice: once at `data.reports` and once inside `data.result`. No consumer reads the inner copy.

**Versions:** Jac 0.36.1 (Darwin arm64) | Python 3.14.7 | macOS 26.6.2

## Steps to Reproduce

`main.jac` in a `jac create x --kind service` project:

```jac
walker:pub Hello {
    can go with Root entry {
        report "hi";
    }
}
```

```bash
jac run
curl -X POST http://localhost:8000/walker/Hello -d '{}'
```

## Expected vs Actual

Expected: reports appear once.

Actual:

```json
{"ok": true, "type": "response",
 "data": {"result": {"_jac_type": "Hello", "_jac_id": "...", "_jac_archetype": "walker",
                     "reports": ["hi"]},
          "reports": ["hi"]},
 "error": null}
```

Note `Hello` does not declare a `reports` field. Declaring or removing `has reports` makes no difference to the output, so this cannot be avoided from user code.

The equivalent `def:pub` has no such duplication, and a function returning a non-walker archetype (`obj`) has no `reports` key inside `result` at all.

## Root Cause

Walkers get an implicit `reports` slot (`jaclang/compiler/passes/layout_pass.jac` L309, and `impl/layout_pass.impl.jac` L294), and the archetype serializer emits it like any other field. The envelope then emits the collected reports again alongside it.

Nothing reads the inner copy:

- Browser: `__doWalkerFetch` ends `return payload["data"]`, so client code reads `data.reports`.
- sv to sv: `jaclang/scale/runtime/rpc/rpc.jac` L231-236 builds the instance from `result`, then immediately overwrites it with `instance.reports = reports` from the envelope.
- Eject already excludes it: `jaclang/cli/commands/eject_targets/backend.jac` L307 filters `fname == "reports"`.

## Impact

The duplicated field is the only part of a walker response that grows without bound. A walker emitting 50,000 reports returns 677,983 bytes where roughly half is the duplicate.

Also worth flagging as a separate inconsistency: `jaclang/cli/docs/tutorials/production/microservices.md` L231 documents these two fields diverging (`"reports":[]` inside `result` next to a populated top-level `reports`), which does not match 0.36.1 behaviour where they always agree.

kashmithnisakya:
Fixes #8906.

## The bug

Every walker response carried its report list twice: once at `data.reports`, and once inside `data.result`, because the executed walker in `result` is serialised attribute by attribute and the walk had just stored its reports on the instance (`scope.wlk.reports = scope.reports` in the server kernel). A 300-item list walker measured 149,771 bytes, of which 74,782 were the copy. A walker that did `report self` recursed through that same attribute.

## The fix

In `api_mode` the serializer leaves a walker's `reports` attribute out of its attribute walk (`Serializer._serialize_attrs`), the same filter eject already applies when it lists walker fields. Everything else about a walker on the wire is unchanged: its `has` fields as the walk left them, `_jac_type`, `_jac_id`, `_jac_archetype`. `POST /walker/Outer {"tag": "t1"}` now answers

```json
{"result": {"_jac_type": "Outer", "_jac_id": "…", "_jac_archetype": "walker", "tag": "t1"},
 "reports": [{"inner": true}, {"outer": "t1"}]}
```

where `result` used to repeat the `reports` list.

Two things depended on the copy and are fixed at their root:

- The embedded gateway's walker proxy nested the provider's walker under `result`, so reports only reached its callers through the copy. It now forwards the provider's `{result, reports}` payload as is.
- `scale/tests/fixtures/cl_fullstack/test_echo.jac` read `result.reports`; it reads `data.reports`.

## One pair of envelope decoders

`sv_client.function_result` and `sv_client.hydrate_walker_envelope` are the envelope decoders. The core runtime and `scale/runtime/rpc/rpc.jac` call them on both the function and walker paths; the scale layer's private `_unwrap_sv_envelope` / `_hydrate_walker_envelope` copies are gone. A remote walker hydrates from the provider's `result` with `reports` attached, so a field the provider's walk mutated reads back on the consumer exactly as after a local spawn: `test_microservice.jac` reads `rn.body`, assigned inside `Notify.send`, across the service boundary, and main's `test_typed_boundary.jac` `check_walker` holds. A non-ok envelope raises `BridgeError` on either path, and `BridgeError` is now a `RuntimeError`, so the scale RPC layer's existing `except RuntimeError` handlers keep catching provider errors.

## What is in this PR

Four changes travel together. They share a cause (the report list had two owners
on the wire and the decoders had been copied), but they are separable, so here is
which test pays for which:

| Change | Production diff | Covered by |
|---|---|---|
| `Serializer._serialize_attrs` omits a walker's `reports` in `api_mode` | `serializer.impl.jac` +6 -1 | `test_serve_walker_wire.jac` |
| `rpc.jac` drops its private envelope decoders for `sv_client.function_result` / `hydrate_walker_envelope` | `rpc.jac` +6 -31, `sv_client.jac` +22 -5 | `test_sv_walker_hydration.jac` |
| The embedded gateway forwards the provider's `{result, reports}` instead of nesting it under `result` | `embedded_gateway.impl.jac` +3 -5 | `test_embedded_gateway.jac` |
| A stub-less spawn is refused in `spawn_walker`, ahead of the transport branch, so both transports refuse before the provider runs | `sv_client.jac`, `bridge.jac` +1 -1 | `test_sv_streaming.jac`, `test_serve_client.jac` |

Production code is +38 -43 across five files: the fix is six lines and the decoder
cleanup removes more than the rest adds.

## Tests

260 on main, 269 here. The wire contract had no coverage before, so the increase
is two table-driven tests; everything else extends a suite that already existed.

- `tests/runtimelib/test_serve_walker_wire.jac`: two tables, one over walker responses and one over function responses, plus a shared invariant that no walker anywhere in a payload keeps an inline report list (which is what `report self` exercises). Reverting the serializer hunk fails it with `walkers still carrying an inline report list: ['Outer']`. On main the same file gives 2 passed, 3 failed, 1 error: `reports` sits inside `result` for the three walker responses, and `report self` answers a 500 with `maximum recursion depth exceeded` from `Serializer.serialize(result, api_mode=True)`.
- `tests/runtimelib/test_sv_walker_hydration.jac`: the compiler-generated stubs from `sv_relative_consumer.jac` rebuild the walker and its typed reports from `result` + `reports`; an envelope without reports; the no-stub payload; both decoders' error envelopes.
- `scale/tests/microservices/test_embedded_gateway.jac` pins the walker proxy payload; `test_microservice.jac` reads a caller field, a provider-mutated field and a report off remote walkers end to end.

## Compatibility

A raw REST consumer that read `data.result.reports` on a walker spawn reads `data.reports` (the fullstack fixture above was the only in-tree reader); generated clients already do. That path removal ships as its own `breaking` fragment; the serializer, gateway, decoder and `BridgeError` changes ship as the `bugfix` fragment. A remote walker spawn with no boundary contract is refused in `sv_client.spawn_walker`, before either transport (HTTP or jac-scale) calls the provider, so a caller retrying on that `BridgeError` never spawns twice.


jac/jaclang/server/sv_client.jac
```diff
@@ -311,22 +312,32 @@ def _http_call(app: str, fn: str, kwargs: dict, headers: dict[str, str]) -> any
...
+    envelope = _unwrap_envelope(data, app, walker_name) or {};
+    if stub_cls is None {
+        return envelope;
+    }
+    reports = envelope.get("reports") if isinstance(envelope, dict) else None;
+    fields = (
+        dict.fromkeys(stub_cls.__jac_fields__)
+        | stub_cls.__jac_field_defaults__
+        | copy.deepcopy(kwargs)
+    );
```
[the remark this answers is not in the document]
kashmithnisakya: good catch, the defaults dict is class-scope so a literal list/dict default was one object across instances. fixed in 004e7bcc8: the merge deep-copies the defaults like it already did for kwargs, plus a regression test on a walker with a list default.

jac/jaclang/scale/runtime/rpc/rpc.jac
```diff
@@ -112,7 +112,7 @@ def service_call(
             return _iter_sse_text(str(resp.text), func_label);
         }
         data = resp.json();
-        return _unwrap_sv_envelope(data, module_name, func_name);
+        return sv_client.function_result(data, module_name, func_name);
```
MusabMahmoodh, line 115: Routing through `function_result` swaps the exception type on the provider-error path: `_unwrap_sv_envelope` raised `RuntimeError`, `_unwrap_envelope` raises `BridgeError`, and `BridgeError` subclasses `Exception`, not `RuntimeError` (bridge.jac:29). Transport and breaker failures in this same file still raise `RuntimeError` (lines 50, 119, 171), so one sv-to-sv call site can now fail with two unrelated types depending on which failure it hit, and any app-level `except RuntimeError` around a remote call stops catching provider errors. Cheapest fix that keeps both:

```suggestion
        return sv_client.function_result(data, module_name, func_name);
```

(and in `bridge.jac`, `obj BridgeError(RuntimeError)`). Otherwise this belongs in the Compatibility section - it is a behaviour break the release note does not mention.

jac/jaclang/server/sv_client.jac
```diff
@@ -311,22 +312,32 @@ def _http_call(app: str, fn: str, kwargs: dict, headers: dict[str, str]) -> any
...
+def hydrate_walker_envelope(
+    data: any, app: str, walker_name: str, kwargs: dict, stub_cls: any
+) -> any {
+    envelope = _unwrap_envelope(data, app, walker_name) or {};
+    if stub_cls is None {
+        return envelope;
+    }
+    reports = envelope.get("reports") if isinstance(envelope, dict) else None;
+    fields = (
+        dict.fromkeys(getattr(stub_cls, "__jac_fields__", ()))
```
MusabMahmoodh: This is where the contract stops being location-transparent: the consumer's instance is the caller's own arguments plus the provider's literal defaults, so a field the provider's walk mutated reads back as the value that was sent. In-process `spawn` still gives the mutated value, so the same source line means two different things depending on whether that module got cut into a service - and it is silent either way. `dict.fromkeys(__jac_fields__)` also puts `None` into any field whose default was not a literal, which will read back as `None` on a `str` field.

E1135 and W6010 cover the other two ways a walker's state escapes; is a diagnostic on reading a non-`reports` field off an `sv import`ed walker binding in scope here, or a follow-up issue? That is the one case left where the two topologies disagree without saying so.

jac/jaclang/data/impl/serializer.impl.jac
```diff
@@ -170,6 +170,9 @@ impl Serializer._serialize_value(
     }
 
     if isinstance(val, Archetype) {
+        if api_mode and isinstance(val, WalkerArchetype) {
```
MusabMahmoodh: This is the envelope's only caller for both `result` and `reports`, so `report self;` (or reporting any walker) now serialises to `{}` with no diagnostic. `report` in a function is E1135 and `def:pub -> Walker` is W6010, but the third way a walker reaches the wire just empties. Worth extending the check in `_walker_report_elem_type`'s caller to warn when the reported expression's type is a walker?

Separately on scope: #8906 was the duplicated `reports` key, and eject already solves exactly that by filtering the intrinsic field (`cli/commands/eject_targets/backend.jac:307`). Blanking the whole archetype also drops every declared `has` field from `result` - case 3 in your table, `CountThings` losing `total` and `names`. That is a much wider break than the issue asked for. Was filtering `reports` alone considered and rejected, or is the empty-walker contract the point?

jac/jaclang/compiler/passes/impl/type_checker_pass.impl.jac
```diff
@@ -1451,19 +1451,18 @@ impl TypeCheckPass._check_walker_reports_shape(
...
-        impl_def = nd.find_parent_of_type(ImplDef);
-        if impl_def is not None and isinstance(impl_def.decl_link, Archetype) {
-            arch = impl_def.decl_link;
+impl TypeCheckPass._enclosing_walk_archetype(nd: UniNode) -> (Archetype | None) {
+    cur: (UniNode | None) = nd.parent;
+    while cur is not None {
+        if isinstance(cur, Archetype) and cur.arch_kind in ("walker", "node", "edge") {
+            return cur;
         }
+        cur = cur.decl_link if isinstance(cur, ImplDef) else cur.parent;
```
MusabMahmoodh: The climb is lexical but `report` is dynamically scoped - it appends to whatever walk is running. So a module-level `def` or `obj` helper that reports and is only ever called from inside a walker ability compiles today and is E1135 after this, and hoisting the nested `obj` from your `report_in_node_ability.jac` fixture out to module level breaks it. The help text's remedy ("spawn a walker from a function") is not the fix for a helper that is already inside a walk. Does the breaking note cover that migration, or is the lexical rule meant to be exactly this strict?

Also: `cur.decl_link` on an `ImplDef` that failed to link is `None`, which ends the climb and fires E1135 as a cascading error on top of the real one. Worth a guard?

MusabMahmoodh (commented):
Read the whole thing against #8906. The consolidation of the three private unwrap copies into `sv_client.unwrap_envelope` / `function_result` / `hydrate_walker_envelope` is the best part of this - `rpc.jac` gets 45 lines lighter and there is now one decoder instead of three that drift.

Four things below. The `RuntimeError` -> `BridgeError` one is the only one I would call a blocker; the rest are scope/contract questions where you may already have an answer.

jac/jaclang/scale/runtime/rpc/rpc.jac
kashmithnisakya: bridgeerror subclasses RuntimeError since e0dc57df9 (bridge.jac), so provider errors keep landing in the existing except RuntimeError handlers; nothing further needed in e03922306.

jac/jaclang/server/sv_client.jac
kashmithnisakya: agreed, the caller-side rebuild broke location transparency; e03922306 hydrates from the provider's result plus reports again (remote_walker_fields reads rn.body set inside the walk, and main's check_walker holds), and the literal-defaults machinery is gone.

jac/jaclang/data/impl/serializer.impl.jac
kashmithnisakya: you're right that this was wider than #8906; e03922306 drops the {} contract and only leaves a walker's reports attribute out of its api_mode attribute walk (the filter eject already uses), so has fields stay on the wire and report self serialises once instead of recursing.

jac/jaclang/compiler/passes/impl/type_checker_pass.impl.jac
kashmithnisakya: report is dynamically scoped, so no lexical rule is sound for helpers or served defs that report today; e03922306 removes E1135 and W6010 and keeps only the runtime fixture proving report reaches the walk from nested defs, objs and impls.

jac/jaclang/data/impl/serializer.impl.jac
```diff
@@ -11,6 +11,9 @@ impl Serializer._serialize_attrs(
     ref_mode: bool = False,
     _seen: (tuple[set, set] | None) = None
 ) -> dict[str, object] {
+    omitted: set[str] = (
+        {'reports'} if api_mode and isinstance(`obj, WalkerArchetype) else set()
+    );
```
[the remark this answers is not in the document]
kashmithnisakya: that's intentional: the {} contract was dropped in e03922306 (see the rewritten description). a walker keeps its has fields on the wire, matching main's typed_boundary check_walker, and only the reports copy inside result goes, which is what #8906 asked for.

jac/jaclang/server/sv_client.jac
```diff
@@ -350,7 +360,10 @@ def _http_spawn(
         }
         data = resp.json();
     }
-    return _rehydrate_walker(_unwrap_envelope(data, app, walker_name) or {}, stub_cls);
+    if stub_cls is None {
```
MusabMahmoodh: This guard runs after the POST, so by the time the caller sees `BridgeError` the provider has already executed the walker, and a caller that retries on it spawns twice. Nothing in the request needs `stub_cls`, so moving the check to the top of `_http_spawn` refuses before any side effect.

MusabMahmoodh (approved):
All four from last round are resolved, and narrowing to the same `reports` filter eject already uses is the right size for #8906.

One contract point: `data.result.reports` is gone from the wire, and this PR's own `cl_fullstack/test_echo.jac` had to move from `saved["result"]["reports"]` to `saved["reports"]`. That's the edit every client reading the old path has to make, so I'd keep a `breaking` fragment for it rather than folding it into the bugfix note.

jac/jaclang/server/sv_client.jac
kashmithnisakya: good catch, the guard now sits at the top of _http_spawn so a stub-less spawn is refused before any POST, with a test asserting the provider sees no call; 3782cdec0.

RandiJay96:
@kugesan1105 let's merge this and do a release. @kashmithnisakya requested a one. pls check

jac/jaclang/server/sv_client.jac
```diff
@@ -311,26 +311,39 @@ def _http_call(app: str, fn: str, kwargs: dict, headers: dict[str, str]) -> any
...
     return stub_cls._from_wire(payload);
 }
 
 def _http_spawn(
     app: str, walker_name: str, kwargs: dict, stub_cls: any, headers: dict[str, str]
 ) -> any {
     import from jaclang.server.serving.client { request_sync }
+    if stub_cls is None {
+        raise BridgeError(detail="Walker response has no declared boundary contract");
+    }
```
MusabMahmoodh: Does the scale path refuse a stub-less spawn anywhere? Under jac-scale `spawn_walker` takes `transport.sv_walker_call` into `rpc.walker_call`, which ends at `hydrate_walker_envelope`, and that still returns the raw envelope for a `None` stub (L332). The release note says a spawn with no boundary contract is refused before the provider is called; moving this check up into `spawn_walker`, ahead of the transport branch, would make that true on both transports.

MusabMahmoodh (approved):
Both from last round are in: the guard runs before the provider is called, and the breaking fragment is split out. One question inline about the scale transport, fine to take here or as a follow-up.

jac/jaclang/server/sv_client.jac
kashmithnisakya: it did not: under jac-scale `spawn_walker` went straight to `transport.sv_walker_call`, and `hydrate_walker_envelope` handed the raw envelope back for a `None` stub after the POST, so the note only held on the HTTP path. 16aa1d118 moves the guard to the top of `spawn_walker`, ahead of the transport branch (a `_local_providers` spawn still runs, since it returns the real walker and needs no contract), drops the copy in `_http_spawn`, and adds a test in `test_sv_streaming.jac` that patches in a scale transport and asserts it records no call. `hydrate_walker_envelope` keeps its `None` branch for the transport-level callers (`test_sv_auth_forward` calls `sv_walker_call` directly), and the bugfix fragment now names both transports.

also rebased onto main (ef7881880) as two commits: the branch's history is squashed into one, since its middle commits added and then removed E1135/W6010 in files #8080 rewrote, and this fix sits on top. #8080's type-argument rule needed `dict[str, any]`/`list[...]` on five test and fixture annotations.

kugesan1105:
Pushed `92207f6d1` to this branch. Tests only, no production change.

**Dropped `report_in_node_ability.jac` and its hook.** The fixture prints the exact output it asserts on main, unmodified, so it covered nothing this PR changes. It is left over from the E1135/W6010 work that `e03922306` removed.

**Collapsed `test_serve_walker_wire.jac` from six tests to two tables.** Same fixture, same coverage, two client boots instead of six. The per-case `"reports" not in result` checks become one invariant that scans the whole payload for any walker still carrying an inline report list, so `report self` is covered as a consequence rather than as a case of its own.

Tests 274 -> 269 across the files this branch touches (260 on main). Reverting the serializer hunk fails with `walkers still carrying an inline report list: ['Outer']`.

jac/jaclang/server/sv_client.jac
kugesan1105: Verified on the current head (`ca810a293`), and the fix does what you asked on both transports:

- The guard is the first statement in `spawn_walker`, ahead of the transport branch, so it fires before `transport.sv_walker_call` and before `_http_spawn`.
- `_local_providers` is exempted, which is right: a local spawn returns the real walker and needs no contract.
- The duplicate guard is gone from `_http_spawn` (0 occurrences).
- `hydrate_walker_envelope` keeps its `None` branch for direct transport-level callers such as `test_sv_auth_forward`.
- The test exists at `test_sv_streaming.jac:321` and asserts the scale transport records no call.
- The bugfix fragment now names both transports explicitly.

Resolving. Reopen if I have read the scale path wrong.

Issue #8681, opened 2026-08-24
Reorganize the jaclang package tree: function over bootstrap history (dissolve jac0core, untangle JacProgram/JacCompiler)


marsninja:
## Problem

The `jaclang/` package tree encodes bootstrap history instead of function. Membership in `jac0core/` is decided by a path prefix check (`meta_importer.py` line ~186: "is the file under `jac0core/`?"), so anything the seed-visible code touches has been dragged into that directory over time. It now holds 100 `.jac` files spanning six unrelated concerns while its own docstring still describes twelve modules. The rest of the tree has the same disease: package names that no longer match contents, one 1,288-line grab-bag driver file, and a `JacProgram`/`JacCompiler` pair whose relationship is circular and implicit. Navigating the codebase requires knowing the bootstrap history, which nothing writes down.

### jac0core conflates six concerns

- Bootstrap-critical compiler core (parser, `unitree.jac`, constants, JCIR codegen). Legitimately here.
- The entire production runtime: `JacRuntime` (`runtime.jac` + ~2,000-line impl), `archetype.jac`, the OSP kernels, `graph_query.jac`, and the public library surface `jaclib.jac` (top-level `lib.jac` is a 12-line forwarder to it). Meanwhile `jaclang.runtimelib.runtime` survives only as a deprecation shim, so the runtime lives in the bootstrap package and `runtimelib` does not contain the runtime.
- Deployment planning: `placement.jac`, `placement_solver.jac`, `placement_pins.jac`, `service_cut.jac`, `prefix_flip.jac` (~90 KB of codespace assignment).
- CLI/config tenants: `tomlio.jac`, `cli_boot.jac`, `treeprinter.jac`.
- Feature-specific: `mtp.jac` (byllm's meaning-typed IR).
- Source-text-as-data: `fmt_kernel.jac`, `osp_kernel.jac`, `osp_graph.jac` are never imported; `kernel_units.jac` reads them as text and concatenates synthetic units for the native linker. Nothing marks this convention.

The jac0core/compiler dependency is a real cycle papered over by deferral: 144 of 369 files under `compiler/` import jac0core, while jac0core has ~30 function-local imports back into `compiler/` plus 4 `import type` edges. Hoisting any one of them to module scope deadlocks bootstrap, and nothing enforces the invariant.

### JacProgram / JacCompiler

[...]

PR #8956, opened 2026-09-04
fix(runtime): alias Root and GenericEdge to their pre-#8682 module path

### LINKED ISSUE #8747: Renaming jac0core.archetype to runtime.archetype drops Root from existing databases: no serializer alias registered
## Summary

The package reorganization in #8682 moved `Root` (and the rest of the archetype built-ins) from `jaclang.jac0core.archetype` to `jaclang.runtime.archetype`. Persisted anchors store the fully qualified class name, so a database written before that commit holds rows naming `jaclang.jac0core.archetype.Root`. That module no longer exists, no serializer alias was registered for it, and the result is that the Root anchor is refused on read and dropped.

The serializer already has the mechanism to prevent this (`Serializer.register_alias`); the rename simply did not use it.

This has not reached users yet: 0.37.0 was unwound by #8730, so the rename is on `main` but unreleased. The window to add the alias is open now.

## Reproduction

Any project whose database was created before `d578aa5b7` and is then opened after it. In my case a scale test fixture:

```
cd jac/jaclang/scale/tests/fixtures
jac run --serve --port <free> --no-client test_api.jac
```

## Actual result

```
WARNING - Refused to deserialize unregistered class: jaclang.jac0core.archetype.Root;
          it is not registered and could not be resolved from its own module
WARNING - Serializer: archetype reconstruction failed for jaclang.jac0core.archetype.Root;
          dropping anchor (caller should quarantine). (occurrence #1)
WARNING - could not materialize anchor 00000000-0000-0000-0000-000000000000 (Root); skipping
```

261 lines matching that pattern in a single 75 second run, the `could not materialize anchor` line repeating for every read that reaches Root.

A project created after the rename is unaffected, which is what identifies this as a migration gap rather than a general defect.

## Expected result

An existing database opens and its Root anchor materializes, as it did before the rename.

## Where

`Serializer._resolve` at `jac/jaclang/data/impl/serializer.impl.jac:606-625` tries three things in order: a registered alias, resolution by importing the stored module, then it gives up and warns. All three fail here. The alias table is empty for this name, and the import path cannot resolve because `jaclang/jac0core/archetype.jac` is gone.

`Serializer.is_auto_registered_module` (`serializer.impl.jac:388`) excludes `jaclang.*` other than `jaclang.scale.*`, so jaclang's own built-ins do not get the fallback that user archetypes rely on.

The mechanism that should cover this already exists:

- `Serializer.register_alias(old_name, cls)` at `serializer.impl.jac:396`
- user archetypes reach it through a `was_names` schema rule (`serializer.impl.jac:382-386`) or the `archetype_alias` decorator (`jac/jaclang/runtime/impl/builtin.impl.jac:146-155`)

`grep -rn "jac0core.archetype" jac/jaclang` returns nothing, so no alias was registered for any of the moved names.

## Suggested fix

Register aliases for the built-ins that moved, so stored rows naming the old module resolve to the new class. `Root` is the one I reproduced, but the same move applies to the rest of `jaclang/runtime/archetype.jac`, including `GenericEdge`, `NodeArchetype`, `EdgeArchetype`, `WalkerArchetype`, and `ObjectArchetype`, and anything else whose fully qualified name a database may hold.

A test that writes with the old name and reads with the new one would keep this from recurring on the next move; #8681 suggests more of these are planned.

## Not established

I did not determine whether the dropped anchor is recoverable or whether data is lost, and my reproduction had concurrent-session lock contention in the same run, so I cannot cleanly separate any effect that had. What is certain is that the class name does not resolve and there is no alias for it.

## Environment

- macOS arm64, jac built from source at `d578aa5b7` and later
- Related: #8681 (the reorganization this is part of), #8682 (the commit), #8049 (closed, same "unregistered class reads return nothing" failure mode from a different cause)

SupulHeshan:
## What broke

Jac saves each object in the database next to the fully-qualified Python path of the class that created it — module name and class name, as plain text. #8682 moved `Root` and `GenericEdge` from `jaclang.jac0core.archetype` to `jaclang.runtime.archetype`. That's normally just a refactor, but any database written *before* that move still has the old module path saved inside it. Opening one of those databases afterward means asking for a class at a path that no longer exists — so it can't be rebuilt, and gets silently dropped instead of erroring:

```
WARNING - Refused to deserialize unregistered class: jaclang.jac0core.archetype.Root
WARNING - Serializer: archetype reconstruction failed for jaclang.jac0core.archetype.Root; dropping anchor
WARNING - could not materialize anchor 00000000-0000-0000-0000-000000000000 (Root); skipping
```

That last line is the root node of the graph. Fixes #8747.

## Why it slipped through

The serializer already has a mechanism built for exactly this — `register_alias(old_name, cls)` lets a class declare "I used to be called something else." User archetypes get this automatically through a schema rule. jaclang's own built-ins never got it wired up for this move, so there was nothing to fall back on.

```mermaid
flowchart LR
    A["Old row on disk<br/>module: jaclang.jac0core.archetype"] --> B{"Serializer resolves<br/>the class"}
    B -->|"registry"| C["miss — registry only<br/>knows the new name"]
    B -->|"alias table"| D["miss — nothing<br/>registered (until now)"]
    B -->|"import the module"| E["refused — jaclang's own<br/>internals don't get this<br/>fallback, by design"]
    C --> F["❌ anchor dropped"]
    D --> F
    E --> F
```

All three normal ways to resolve a class come up empty, so the anchor is dropped. This PR adds the missing piece: registering `Root` and `GenericEdge`'s pre-move name as an alias, the same way user code already can.

```mermaid
flowchart LR
    A["Old row on disk<br/>module: jaclang.jac0core.archetype"] --> B{"Serializer resolves<br/>the class"}
    B -->|"alias table"| G["✅ resolves to<br/>jaclang.runtime.archetype.Root"]
```

## The fix

- `Root` and `GenericEdge` now register their pre-move name via the existing `register_alias` machinery, in a small named table (`_BUILTIN_MOVED_FROM` in `data/serializer.jac`) — the next time a persisted built-in moves (#8681 has more of these queued), it's a one-line addition instead of a new mechanism.
- Also fixed a bug in `register_alias` itself: it was handed the class but threw it away, keeping only the name mapping. If that class wasn't already sitting in the registry, the alias pointed at nothing. It now registers the class immediately if needed.

## Tests

One test added to `test_layer3_coercion_alias.jac` (21 → 22): `"persisted built-ins keep their pinned identity and resolve every prior name"`. It's table-driven — each entry lists a built-in's current module and every module it used to live in, and the test checks both that the current location is exactly what's expected and that every past name still resolves through the alias table. Moving one of these classes without updating the table now fails this test with a message pointing straight at the fix, instead of shipping a silent data-loss bug like this one.

## How I verified it

Beyond the unit test, I reproduced the original bug end-to-end against a real, Postgres-backed `jac run --serve`:
1. Wrote a database using working code (so `Root` lands under its correct current module).
2. Hand-patched the root anchor's stored module string to the pre-#8682 path, simulating a database that predates the move.
3. Reopened it on unfixed `main` — got the exact three warnings above.
4. Reopened it on this branch, same database — clean, no warnings.


christianwilkins (commented):
Reviewed head `0bb4955b8177c8608cd2f4eb6aa8b3b41ee3b7b3`.

I inspected the complete patch and the registration/alias-resolution paths. No blocking code finding: `register` places the class in the registry before processing its declared aliases, so the new missing-target registration does not recurse indefinitely. Both historical built-in paths resolve through the explicit alias table, without broadening the import fallback for internal runtime classes. The later commits consolidate and document the identity regression test; all are included in this review.

Current-head CI has 23 successful checks and 5 skips, and GitHub reports no conflicts. My independent `test_layer3_coercion_alias.jac` run did not finish after 15 minutes and was stopped, so I cannot report a local pass or independently confirm the database round trip described in the PR. No source changes made.

Commenting with local validation incomplete. I am not requesting an author code change from this review; maintainer review of the current-head CI/test evidence remains before merge.

PR #8957, opened 2026-09-04
fix(checker): comparing to a literal no longer drops the type in the other branch

SandeepaHWP:
## What was wrong

Comparing an optional to an int literal erased the int arm of the union:

```jac
def f(x: int | None) -> int {
    if x == 0 {
        return 0;
    }
    return x;    # reported: Cannot return NoneType, expected int
}
```

`x` is `int | None` on that last line, but the checker had narrowed it to
`NoneType`, and every later guard reasoned from that. A `str | None` compared
against `""` kept its arm, so the two spellings disagreed with each other.

## Why

`x == 0` classifies as a literal comparison, but the narrow type it carries is
plain `int`, not `Literal[0]`. The false branch excludes the narrow type from
the declared one, so it removed `int` from `int | None` outright, as if
`x != 0` meant "x is not an int".

The collapse is not cosmetic, it is a soundness hole. A too-narrow type is
assignable to *more* places, not fewer, so the collapsed type is accepted
exactly where the correct wider one would be rejected. On main this
type-checks clean:

```jac
def f(x: int | None) -> None {
    if x == 0 {
        return None;
    }
    n: None = x;   # accepted on main, x really holds an int at runtime
    return n;
}
```

Nothing rejects that, which is why it has been sitting there. It surfaces
either in the type the checker names in an unrelated diagnostic, or as a value
admitted through a declared `None`.

## The change

A comparison against a value narrows on the branch that confirms the value and
on no other. `NarrowingPredicate` already expresses exactly that, and the
consumer already returns "no narrowing" for the branch a predicate opts out of,
so this is two flags rather than new machinery:

- `LITERAL_EQ` declares `narrows_on_false=False`
- `LITERAL_NE` declares `narrows_on_true=False`

The positive direction each form gets right is untouched: `x == 0` still
narrows to `int` on true, `x != 0` still narrows to `int` on false.

## Tests

`checker_literal_eq_keeps_base.jac`, read by `bug_liteq_1_int_literal_comparison_keeps_int_arm`.
Four functions, one test, no new test added.

Measured on the same fixture with and without the fix:

| case | without | with |
| --- | --- | --- |
| `x == 0` then `return x` | `NoneType` | `int \| NoneType` |
| `x != 0` then `return x` | `NoneType` | `int \| NoneType` |
| `x == 0` then `n: None = x` | accepted | rejected |

The first two rows read the reported type rather than the error count, because
the count is the same either way. The third row is the soundness case: without
the fix the else branch binds a live `int` to a `None` name and reports
nothing at all, so that row is what actually regresses if this breaks again.

Targeted suites, 398 passed / 0 failed: `test_checker_bugs` (80),
`test_checker_pass` (244), `test_checker_gaps` (22), `test_checker_shortcomings`
(10), `test_checker_production` (10), `test_checker_fixed_width` (32).

## How this was found

It is the false positive behind one of the `build-kit` failures on #8934, which
adds an operator diagnostic that reads the narrowed type of its operands. That
PR will rebase on this one.


christianwilkins (commented):
Reviewed head `9e1638239d74438556c5163358afa6ef626e966b`.

I inspected the full patch, literal-condition classification, predicate propagation through `not`/boolean expressions, the branch-consumption flags, and the diagnostic regression fixture. No new blocking code finding: opting out only on the branch that does not establish equality preserves the optional union, and the regression reads the reported type rather than relying on an unchanged error count. The release-note follow-up is included.

Current-head CI has 23 successful checks and 5 skips, but GitHub reports merge conflicts. My focused `test_checker_bugs.jac -f liteq` run did not finish after 15 minutes of compiler setup/execution and was stopped; no local test pass is claimed and no source was changed.

Not ready to merge because of the conflicts. Author next step: rebase the two predicate flags and fixture onto current main, preserve the existing boolean-predicate handling, and rerun the focused regression plus CI on the new head. Maintainer review/merge follows after that.

Issue #8943, opened 2026-09-04
Native compilation reorg: the module is the unit, every artifact is a link


marsninja:
This is a design/umbrella issue and a clean break. It comes out of the failure of `jac precompile ./jac` on a tree that had merged #8825 without a rebuilt kernel, and the survey that followed of how the native backend produces its artifacts. Part 1 is what exists today, Part 2 is the architecture, Part 3 is the inventory of code that goes, changes, or stays, Part 4 is sequencing. The goal is one unit model for native compilation, sitting on the module cache the compiler already has, with no second cache, no second key discipline, no process-global mode flags and no heuristic that guesses what a build should contain.

## Part 1: What is there today

The compiler has exactly one unit of identity and caching, the module. A module's JIR entry (`jac/jaclang/compiler/driver/jir.jac:14-28`, format 24 at `:9`) carries typed sections: bytecode, a type interface prefixed with its 64-hex digest (`types/stubcat/modiface.jac:8-28`), dependency rows of `path · iface digest · flags` (`driver/ifacecache.jac:263-271`), comptime deps, placement, diagnostics, and two keys: MODKEY over content plus annexes plus environment (`jir.jac:406-415, 610-615`) and ENVKEY over environment plus compiler digest (`:395-404`). Dependents are invalidated by a layered check that recomputes every recorded dependency digest (`ifacecache.jac:347-369`). The writer merges sections from independent producers when both keys match (`jir.jac:995-1066`).

The native backend has no unit. Its unit is the closure of a root, and every piece of it is a separate mechanism bolted beside the module cache rather than on it:

[...]

Issue #8828, opened 2026-09-01
jac-check flakes red on files the PR never touched, green on rerun: parallel check retains poisoned type state in the global self-host hub


kugesan1105:
### Summary

Since 2026-08-30, the `jac-check` CI lane intermittently fails with phantom type errors (E1032, E1053, E1099, E1031) in files the PR under test never touched, and goes green on a plain rerun of the same commit. The flagged code is always correctly guarded or self-consistent; the checker is reporting its own corrupted state, not real defects.

The flake began within an hour of #8748 (the graph-shaped compiler) merging, which is the PR that made `jac check` parallel. The mechanism is not the parallelism itself: each pool worker reuses one `JacProgram` across every file it serves, while all jaclang-tree imports resolve into a process-global self-host module hub that is never reset between files. Stale `.type` memos minted by one file's evaluator survive into the next file's check, two nondeterministic events write poisoned memos into that shared state, and the select-driven file-to-worker assignment shuffles the victims every run. That is why a rerun passes.

### Observed occurrences (all rerun-to-green on the same commit, no tree change)

| When (UTC) | Where | Failing files | Signature |
| --- | --- | --- | --- |
| Aug 30 15:56 | PR #8781, attempt 1 ([run 33321058568](https://github.com/jaseci-labs/jac/actions/runs/33321058568)) | `jac/jaclang/dist/publish/dts.jac` | `E1032: Type is Unknown, cannot access attribute "vars"` at `dts.impl.jac:76` |
| Aug 30 18:00 | main push ([run 33326828558](https://github.com/jaseci-labs/jac/actions/runs/33326828558)) | `client/targets/pwa_target.jac`, `web_target.jac`, `runtime/na_stdlib/urllib/request.jac` | `E1099` on `JacConfig \| NoneType`, `E1001: Cannot assign <Unknown> to bytes` |
| Aug 30 20:22 | main push ([run 33333448372](https://github.com/jaseci-labs/jac/actions/runs/33333448372)) | `urllib/request.jac`, `dist/publish/sources.jac` | `E1031: Cannot access attribute "is_dir" for type "Self"`, `E1053`, `E1054` |
[...]

PR #8980, opened 2026-09-05
feat(native): build reusable units in owned compiler sessions

marsninja:
Native modules now own reusable products in the shared JIR cache, and every native entry point uses the same dependency-aware link plan. A body edit recompiles the affected unit; an unchanged artifact returns before whole-program optimization and object emission. This implements #8943 and removes the per-unit subprocess workaround that made build-kit substantially slower.

## Compiler ownership and reuse

- Build-owned sessions reuse the existing frontend, dependency resolver, scheduler, interface codec, and native registry. Each source is analyzed once. Cyclic interface settlement reruns lowering; optimization and persistence happen after agreement, with source identity checked before publication.
- Programs own their semantic trees, evaluators, and hydrated catalogs. Immutable catalog bytes remain shared. Nested compiler imports and direct compiler-source builds retain analysis until their active compilation finishes and release their semantic closures together, collection budgets count released units, and source analysis no longer borrows an execution hub or requires `force_target_program`.
- Cache verification uses one owned, read-only oracle, suppresses recursive verification and persistence inside it, and releases its closure in `finally`. Only the requested interface is encoded for comparison; verification remains enabled. Bootstrap compilations retain their frontend tier through code generation, even if imports finish in between. Native coverage replays from the completed link plan, including cached units.
- Star imports use canonical, cycle-aware dependency ingestion instead of recursively compiling Python implementation bodies. Imported symbols preserve the existing lazy type provider. C3 inheritance uses canonical class identity; lazy context-manager bindings resolve through the shared evaluator.
- Removes per-unit child processes, environment controls, metadata transport, and duplicate library-loading infrastructure. Kernel bootstrap remains in the existing kernel resolver.

## Cache, linking, and distribution

- JIR format **28** stores native interfaces, bitcode, optional objects, consumed interface digests, and compile-time dependencies by their full compiler stamp. Target and codegen variants coexist. A shared bootstrap-safe file lock protects the complete read/merge/replace transaction; kernel builds and toolchain installs use the same primitive. Replacing bitcode invalidates an older optional object. Existing caches rebuild automatically.
- Native binaries, layout sidecars, JIR files, stub catalogs, and downloaded binary caches share one atomic writer. Rebuilds publish a new file identity, preserving open executable images and avoiding stale macOS code-signature state.
- CLI, packages, kernel/seal builds, wasm, and JIT share the link plan. The final action digest covers unit products, glue, runtime inputs, and configuration; the existing sidecar verifies output integrity before reuse.
- Replaces handwritten AArch64 outline-atomic assembly with pinned, unmodified LLVM compiler-rt sources and upstream CPU detection. The existing archive linker selects required members. ELF links preserve constructor priorities and startup initialization.
- JIT binds foreign functions to their declared library through engine-local mappings and retains library handles for the engine lifetime. Missing libraries/symbols fail compilation. This prevents process-global symbol collisions from mixing incompatible SSL implementations.
- Serve manifests compile in an owned session without JIT-loading native libraries on the build host; the session is released after metadata extraction. Native clocks select clock IDs from the target libc.
- Darwin sockets use the existing platform-variant mechanism for address layout, errno access, and IPv6 constants. Module globals resolve correctly when a module and its class share a name. Chained OSP filters reuse the ES filter lowering used for inline filters.
- Automatic test concurrency uses the same shared memory-ceiling budget as `jac check`, consistent with the test worker retirement setting; explicit counts remain operator overrides. The repository no longer forces eight workers on every machine. The formatter parallelism regression explicitly requests two workers.
- Keeps module-qualified private symbols, public C ABI exports, relocatable package identities, qualified-entry exception reporting, and lightweight native package imports. Build logs report native analysis/lowering/codegen counts.

## Interface correctness and native analysis

- Interface identities use canonical external declaration identities and normalized source paths. Imported symbols retain their original definition and lazy binding classification. Failed type evaluation prevents publication of an incomplete interface.
- The serializer retains type objects for the lifetime of its identity memo, preventing recycled Python object IDs from conflating classes with unknown types. A regression reproduces the collision without ownership and verifies the fix.
- The type checker distinguishes special-variable expressions from declaration and attribute names. Methods named `root` retain their function signatures during cold checks and cache hydration; real return-type errors are still rejected.
- Native dependencies select native types before analysis through the existing native compilation options. Client field-access auditing stays in source analysis; AOT emission uses the existing pass context instead of rewriting placement after analysis.
- CI equivalence retains four workers and cache verification. Merge conflicts with main and the resulting grammar/API changes are resolved.

## Validation

**Final CI is green at `b92a21b3e7`: [all 26 executed jobs passed, including Everything Passed](https://github.com/jaseci-labs/jac/actions/runs/34364063306).** The macOS-native and installer lanes were skipped by the workflow conditions.

- Repository type checking: **1,058 files passed**, including the previously failing scheduler. Native/compiler-tools: **1,644 passed, 4 platform/tool skips**. Equivalence: **218 passed**, with cache verification enabled and four workers. Runtime, client, bootstrap, packaging, Android, development-loop, scale, and Kubernetes gates passed.
- Local: **31 interface-cache tests passed with verification enabled**, **245 type-checker tests passed**, and the modified type checker passes type checking. A genuinely cold **152-file, two-worker** reproduction failed at the scheduler before the special-name fix and passed all files afterward. A minimal method named `root` reproduces the declaration-type corruption independently of CI.
- Final equivalence job: **11m29s total** (575.20s test execution), versus the earlier four-worker run's **31m26s**. Recent main equivalence jobs took **10m24s–13m25s**; this restores performance to that range.
- Final build-kit: **22m51s**, with payload compilation taking 1,037.0s and sealing 34.2s. Main `10b52d3cf` took **14m58s** ([run](https://github.com/jaseci-labs/jac/actions/runs/34337937799)); both missed the JIR precompile cache. This branch's payload has **669 modules versus 603** on main. Cold build-kit performance remains slower; the equivalence improvement is not a claim of build-kit parity.

Design: `jac/jaclang/cli/docs/internals/native-units.md`.


marsninja:
## Round 2: merge with main, the six-blocker follow-through, and the two red lanes

Scope: the branch as of the Sep 21 head, re-reviewed against the Sep 10 comment above, then brought up to `main` (`dd6592bac`) and fixed. Everything below is on the branch.

### What was wrong at the Sep 21 head

- **Two red lanes.** `test-runtime` failed on `test_build_jab.jac:42`, which asserted the `walker_access` manifest key that #9355 (already merged into the branch) replaced with the `surface` record. `test-jac-pack-eject` took 2h21m against 8 to 15 min on main: `jac build --as source` ran native placement inference over the vendored runtime (46 modules "preferred native but did not lower", 222 seam demotions; zero on main) because `program_for` no longer routes package paths through the internal program, so runtime modules compiled as application modules under the site's native default. The replay step then failed its output diff.
- **Five conflicting files with main**, one semantic: #9406's provenance rule (a stale compile-time dependency record or changed interop manifest drops executable sections) versus this branch's stamped native sections and two writers sharing one manifest.
- **Design gaps left from Sep 10:** the interface decoder never verified the digest prefix against the body while the plan's staleness check read the prefix and the registry recomputed from the body; stamped variants accumulated without bound across compiler digests; every module JIR gained a permanent zero-byte `.lock` sibling; the sealed ctypes engine had no direct test; the PR body and release note said JIR format 28 while the code and design doc say 29.

### What changed

- **Merge with main.** Skipped imported methods keep their module-qualified keys and now also record their declaring module for main's origin check; globals keep the unit's module-qualified symbols and export the symbol they were emitted under; `min`/`max` keyword handling comes from main. The JIR writer keeps main's provenance rule for the Python lane (bytecode, manifest, module IR, client artifact) and this branch's self-validating native sections. A native unit persisting only its products fills in the interop manifest when none exists and otherwise leaves the bytecode compile's manifest alone, so the two lanes stop purging each other's products. Main's `test_cache_section_provenance.jac` is ported to the stamped sections with three new cases.
- **Vendored runtime placement.** `build_runtime_closure` gives each runtime module the codespace the running compiler gives its own package: server unless the module is native by policy (`is_native_by_location`, `is_shared_runtime_module`), which is what the internal program gave them before. Regression in `test_runtime_vendor.jac`.
- **Vendoring throughput.** The per-module fresh `JacProgram` the branch added on Sep 20 re-hydrated the catalog and evaluator for every runtime module; with #9369's smaller closure that workaround is no longer needed, and one shared program restores the export's timing (local base closure of 287 entries: 665 s with a program per module, 177 s shared).
- **Interface digest.** `NativeIface.from_bytes` accepts an entry only when the body hashes to the prefix; `verified_digest` feeds `known_digest` and the seal's twin check.
- **Cache growth.** A native write prunes every native section's variants to the compiler digests it carries (other targets and codegen identities under the same compiler still coexist). Module cache writes lock through a 64-slot pool under the generation directory. Both covered in the provenance test.
- **Manifest test** reads the surface record; **sealed engine** surface pinned against the host C library in `test_native_library.jac`.
- **Layout export after the merge.** With main's bundled `math`/`cmath`, a unit's layout dropped any function whose name was also a foreign symbol propagated from a dependency (cmath's `sqrt` beside libm's), so consumers treated it as demoted and routed the call to the extern. The export now skips only the unit's own extern declaration, by identity. Regression in `test_native_iface.jac`; this was the cause of the `math`/`cmath` equivalence, module-alias and function-body-import failures on the first CI round.
- **Docs.** Design doc gains the lowering-diagnostics policy (a suppressed lowering error still withholds the unit and names the diagnostic), the digest, pruning, lock-pool and vendoring notes; release notes and PR body say format 29.

### Not done here, on purpose

The three ContextVars and options swapping in `NativeBuildSession.lower`, the session's second program with `_share_engine`, the `:pub` carve-out for `na_stdlib`, and positional interface identity are design follow-ups; changing them inside this PR would restart the stabilization loop. The unrelated fixes riding in this PR (Darwin sockets, C3 identity, `root` methods, test worker budget, server request context) are left where they are; splitting them is a call for the author.

### CI on `c6985ac56` (run 35771982437)

All 26 executed test lanes pass. Timings against recent `main` runs:

| Lane | Sep 21 head | This head | Recent main |
|---|---|---|---|
| test-runtime | 3h11m (failed) | 25m | 9 to 11m |
| test-jac-pack-eject | 2h21m (failed) | 51m | 8 to 15m |
| test-compiler (equivalence) | pass | 14m | 10 to 13m |
| build-kit (cold precompile cache) | 43m | 45m | 24 to 28m |
| jac-check | 1h02m | 1h11m (rerun) | 25 to 30m |

What is left in the two slow lanes is structural, not a bug: the source export vendors the runtime closure by analyzing every runtime module in a program the export owns, where `main` resolves those modules to the compiler's own in-memory program and gets them for free (the replay export takes 7 s on main and 23 min here; the module cache does not help because the export needs full trees and Python source products, 162 s warm versus 177 s cold locally). Restoring main's speed means either letting the vendorer borrow the selfhost program again, which this PR's ownership model removed on purpose, or caching Python source products in the JIR. Both are follow-ups for the author to weigh.

`jac-check` failed on the first attempt with `Type is Unknown` for `llvm.ValueRef` attributes in `link_glue.jac` and `wasm_build.jac`, files this round did not touch and which pass the same sweep locally (1056 files); that is the #8828 pooled-worker pattern; the rerun passed (whole-tree step 64 min).

marsninja:
## CI performance degradation on this branch: root causes, fixes, and what should come back

### What the numbers say

Medians over eight main runs today versus this branch's third run (all on 4-vCPU runners, same worker counts, both missing the precompile cache):

| Lane | main | branch | Where the delta sits |
|---|---|---|---|
| build-kit | 25.5m | 45.0m | payload 18m to 21m (624 to 703 modules), launcher stub 34s to 1m, CPython build 8m to 20m with the same cache restored (unexplained, not a compiler input) |
| jac-check | 30.6m | 71.1m | check step 27m to 46m: per-file median 0.26s to 0.32s (+5m), stalls over 60s 3.6m to 26m (+22m), plus one flake rerun |
| test-runtime | 10.7m | 25.0m | summed file time 34m to 78m; median file 1.34x; 16 files carry 39 of the 44 extra minutes |
| test-jac-pack-eject | 8.0m | 51.1m | web export 12m to 24m, replay export 7s to 23m |
| passes-native | 18.0m | 26.3m | uniform, median file 1.39x |
| toplevel / passes-main / equivalence / client | 7.9 / 4.1 / 4.8 / 9.3m | 26.1 / 15.2 / 14.3 / 18.1m | uniform per-compile and per-process overhead |
| launcher | 1.6m | 5.0m | one "warm the runtime cache" step: under 20s on main, 216s here |

Ruled out: worker counts (4 and 2 slots on both), worker recycling (7 versus 6 retirements), memory (peak RSS 4.1 GB on main, 2.1 GB here), the format gate (1 minute on both).

### Root causes, ranked by minutes lost

**1. Every compiler program now rebuilds its own hydrated world, per file and per process.** The ownership model has `release_compile_closure` drop the evaluator and the session catalog after every file ([program.impl.jac:345](https://github.com/jaseci-labs/jac/blob/453329ade/jac/jaclang/compiler/driver/impl/program.impl.jac#L345) sets `_stub_catalog = None`; the merge base kept it), and `for_session` at [reader.jac:305](https://github.com/jaseci-labs/jac/blob/453329ade/jac/jaclang/compiler/types/stubcat/reader.jac#L305) starts with empty memo tables, so every stub type a file touches is decoded from bytes again. That is the uniform 1.2 to 1.4x on every lane that compiles anything. The same shape hits each fresh process harder: the runtime lane's worst files are the tests that spawn `jac` twenty to forty times (command and startup isolation, plugin integration, `test_cli` from 0.8s to 127s, shadcn create, guide, man code), each spawn paying roughly 10 to 15 seconds more than on main. Which parts of a cold process dominate I could not split without running one: candidates in order are catalog and evaluator hydration, kernel acceptance hashing every unit's sources on each start ([kernel_resolve.jac:232](https://github.com/jaseci-labs/jac/blob/453329ade/jac/jaclang/compiler/backends/native/kernel_resolve.jac#L232)), native record validation on each import of a runtime unit ([nativecache.jac:255](https://github.com/jaseci-labs/jac/blob/453329ade/jac/jaclang/compiler/driver/nativecache.jac#L255) recomputes dependency digests), and the deferred release plus collection the meta importer now runs at the end of execution. Confidence: high that this is the largest bucket, medium on the split.

**2. Source products no longer borrow the compiler's program.** `program_for` at [compilation_context.jac:184](https://github.com/jaseci-labs/jac/blob/453329ade/jac/jaclang/compiler/driver/compilation_context.jac#L184) hands back the caller's program instead of `selfhost.resolve`, so every runtime module in an export closure is analyzed from scratch in an owned program; the module cache cannot help because the export needs full trees and Python source. Verified: the replay export is 7 seconds on main and 23 minutes here, and the runtime closure builds in 162s warm versus 177s cold locally. This is the whole pack-eject delta and about 3 minutes of the runtime lane (publish tests), after this round's fixes.

**3. The JIT lane re-links and re-optimizes on every process.** Main loaded cached fused bitcode straight into MCJIT; the branch's `jit_engine` at [link_plan.jac:1287](https://github.com/jaseci-labs/jac/blob/453329ade/jac/jaclang/compiler/backends/native/link_plan.jac#L1287) merges the unit bitcodes and runs whole-program optimization before every engine, including the runtime units linked beside the application, and the design doc records that this is not incremental. Verified by code; magnitude inferred from passes-native's uniform 1.39x and equivalence at 3x.

**4. Checking application examples now stalls for minutes.** The 26 minutes of stalls in jac-check land on four example apps with client, native or wasm parts (raylib shooter web 335s, jaclang.org web, mobile and cli 120 to 178s, tiny_jacyac web 132s) and on the largest compiler modules (`na_ir_gen_pass` 143s to 196s, the jacpython compiler files 150 to 220s each). The app stalls have no counterpart on main and no native-lowering output in the log, so the mechanism is not pinned; the likely one is analysis pulling native unit interfaces through `ensure(compile_on_miss=True)` on a cold cache and regenerating client artifacts per worker. Confidence: medium that it is cache-miss lowering, low on details.

**5. Build-kit.** The compiler-attributable part is about 3.5 minutes: 13 percent more payload modules and the session-built kernel (about 1 minute on CI per the dev-loop lane, versus the fused build). The other 12 minutes is the CPython build step doubling with the same cache restored and no change to build.zig on the branch; I would re-measure before treating it as real.

### Fixes and what they should recover

| Fix | Targets | Expected effect |
|---|---|---|
| Keep the hydrated catalog and the evaluator's immutable-derived memos alive across files in a worker process; release only memos that point at mutable trees | cause 1 | per-file overhead 1.3x to about 1.05x on every lane; the bulk of jac-check's +5m and the test lanes' uniform slowdown |
| Make process start cheap again: cache the kernel's source key (size and mtime guarded) so acceptance does not hash 35 units per start; skip native record re-validation when the JIR's own key already matched; bound the end-of-process release | cause 1 | subprocess-heavy tests back near main; launcher lane to about 2m |
| Persist a JIT-ready optimized merged image per root beside the unit products, reused by digest like the linked-artifact sidecar already is | cause 3 | passes-native to about 19m, equivalence to about 6m, and a share of the runtime lane |
| Route package-owned modules through `selfhost.resolve` inside `python_source`, or cache Python source products in the JIR | cause 2 | pack-eject 51m to about 12m (borrow) or 15m (cache); publish tests to main |
| In analysis paths, never lower native dependencies on a miss (`compile_on_miss=False`), and generate client kernel artifacts once per job | cause 4 | jac-check stalls 26m to about 4m |
| In the build session, skip re-lowering units whose consumed digests did not move between settlement rounds (34 units took 73 lowering visits) | build-kit | kernel and payload about 1 to 2m |
| Key the reroute lanes' JIR analysis cache so branch runs hit it, and fix the #8828 memo scrub | CI | removes the one-hour flake reruns |

Predicted end state with all of these, against the same main medians: test-runtime about 12 to 13m, jac-check about 35m, pack-eject about 12 to 15m, passes-native about 19 to 20m, the remaining test lanes within 10 to 15 percent of main, build-kit about 30m. The residue that the design itself imposes is per-unit native compilation and process-level isolation of programs, which I estimate at 5 to 15 percent on compile-heavy lanes and 3 to 5 minutes on build-kit; getting below that means letting programs share hydrated state again, which is the trade the PR chose. The two numbers I trust least are the jac-check app stalls (mechanism unconfirmed) and the CPython build step (probably variance).

marsninja:
## Performance follow-up: what was actually wrong, and where the lanes stand now

I reproduced the lane deltas locally against the CI kits (the `jac-kit` artifact of this branch's run and of a main run), traced them, and pushed the fixes in four commits (20e6710f, df311875, 12fa0478, and the export change in the commit after them). Three mechanisms explained most of the minutes, and none of them was the per-file catalog hydration the analysis above ranked first.

### 1. Every reroute lane derived the compiler kernel from scratch

The `jac-kit` action's "Warm the runtime cache" step took 250 to 430 s on this branch and 8 to 14 s on main, on every reroute lane (jac-check, equivalence, launcher, packages-and-docs, dev-loop, the five scale lanes). The kit's kernel was never adopted, for two independent reasons: the sidecar's folded `source_key` was computed at seal time from options carrying the repo's `[native] require` policy and from staging paths outside the running package (whose module keys fingerprint their project), so no checkout could reproduce it; and the kit strips the kernel after the sidecar records the library's hash, re-stamping only the manifest, so even a matching copy failed the hash check. The sidecar now records the kernel's inputs one by one (compiler digest, shared codegen identity, triple, interpreter, per-unit source digest and policy flag), the kit re-stamps the sidecar when it strips, and adoption validates the library against the kit manifest. The dev-loop lane now fails if the first run derives instead of adopting. Warm step now: 16 s.

### 2. Sealed native products never matched any consumer

The seal stamped native records with the full `native_require` list of the build. Consumers run without it, so every program re-lowered the runtime shims it imported (`_os_native`, `region_native` and their closure): 2.8 s of a 5.4 s tiny compile in the sealed kit, and 460 s (main: 4 s) for `tests/cli/test_extract_script_args.jac`, whose CLI imports were lowered as native units, comptime producers and all, once per program. Four such files carried 1650 s of the runtime lane's 2080 s delta. Stamps are now per unit: the shared codegen identity plus whether the policy names that unit. The native pass consults a dependency's products only when placement makes it native, and a string-source root still reads its file-backed dependencies' records.

### 3. Package modules were analyzed under each consumer's options

`bind_options` inherited the importer's application, codespace and entry into package modules, so their cache identity forked per consumer and the interface cache refused to serve them (`eligible` rejects any default codespace). Package modules are now bound canonically (`owned_module_options`) under an empty cache projection, the identity the package's precompile and the compiler's own bootstrap use, their cache key and file name ignore the active scope, and lookups fall back to the precompiled tree beside the package (sealed by manifest, unsealed bundle by key). The hydrated catalog and native interfaces also survive a closure release, and annex discovery caches directory listings.

### Lane wall time, minutes

main is the median of eight main runs; "before" is this branch's run 3; "now" is run 35817216450 (12fa0478).

| Lane | main | before | now |
|---|---|---|---|
| jac-check | 30.6 | 71.1 | 29.9 |
| test-runtime | 10.7 | 25.0 | 10.5 |
| test-client | 9.3 | 18.1 | 9.4 |
| passes-native | 18.0 | 26.3 | 19.3 |
| toplevel | 7.9 | 26.1 | 11.6 |
| passes-main | 4.1 | 15.2 | 6.1 |
| equivalence | 4.8 | 14.3 | 5.1 |
| launcher | 1.6 | 5.0 | 1.4 |
| test-jac-pack-eject | 8.0 | 51.1 | 22.4 |
| build-kit | 25.5 | 45.0 | 31.5 |

Local, sealed kits on the same machine: the tiny warm compile went from 5.4 s to 0.8 s (main kit: 2.0 s), the CLI test above from 457 s to 1 s in a warm process, and per-program native re-lowering from 10 units to 0.

### What is left

- pack-eject: the site export takes 11.0 min (main 5.8) and the replay export 8.6 min (main 6 s). The replay app is a native binary, and the export handed the C backend the unoptimized whole program, every unit of the runtime closure included, where main's export covered the application's own IR. The export now internalizes and optimizes the merged program before emitting C (the object and bitcode links already did); that commit is pushed and its CI result will show the new number. The site export's remainder is the Python source of the vendored runtime, which costs about 1.7x main per module.
- toplevel and passes-main carry 2 to 4 minutes over main in the tests that spawn compiler processes with fresh caches; each spawn still pays kernel acceptance and a catalog hydration.
- build-kit is about 6 minutes over main: the per-unit seal of the kernel and the CPython build step's variance.

marsninja:
Follow-up on the pack-eject lane, which was the last lane still far from main (25.6 min against main's 8).

**Root cause.** A source export vendors every runtime module the generated Python imports, following nested imports. The exported project's wasm tooling (`dist/source/build`) imports the LEB and archive codecs from `linker_common`, and on this branch `linker_common` had also grown the floor-archive and static-runtime helpers a native link uses. Those helpers import `floor`, `graph_utils` and `fused_binary`, so the vendorer followed them into the runtime, the fused-binary support and the compiler: the replay fixture's export vendored 492 modules where main vendors 16, and the site export's vendoring tail went from under 3 minutes to over 10. The whole-program optimisation added earlier for the C emission was real but was not where the time went.

**Fix** (580272221): the helpers moved into a new `static_link` module that `link_plan` imports and the exported tooling never does; `linker_common` is back to codecs and archive parsing only.

| step | main | before | after |
| --- | --- | --- | --- |
| site scaffold + export | 4.6 min | 12.8 min | 5.0 min |
| replay fixture export | 6 s | 10.0 min | 8 s |
| pack-eject lane | 8.0 min | 25.6 min | 7.9 min |

Locally the replay export went from about 10 minutes to 16 seconds with 15 vendored `jaclang` modules, matching main.

Run: https://github.com/jaseci-labs/jac/actions/runs/35838724341. All other lanes are within a few minutes of main (jac-check has landed between 27 and 32 minutes across the last four runs against main's 30.6).

PR #9007, opened 2026-09-07
fix(type-checker): enforce invariant generic arguments

### LINKED ISSUE #8550: [Bug] Generic list covariance accepted: list[Dog] passes where list[Animal] expected (unsound)
**What's wrong**

`list[Dog]` is accepted where `list[Animal]` is expected (covariant treatment of mutable generic containers). This is unsound: a function receiving `xs: list[Animal]` can append a `Cat`, and the caller still holds it as `list[Dog]`, so `d: Dog = ds[0]` type-checks while holding a Cat.

**Minimal repro**

```jac
obj Animal {
    def speak() -> str { return "..."; }
}
obj Dog(Animal) {
    override:def speak() -> str { return "woof"; }
}
def count_animals(xs: list[Animal]) -> int {
    return len(xs);
}
def g(ds: list[Dog]) -> int {
    return count_animals(ds);
}
```

`.venv/bin/jac check repro.jac` passes (only cosmetic W3005 warnings).

Unsoundness follows from the same covariance:

```jac
def poison(ds: list[Dog]) {
    animals: list[Animal] = ds;  # accepted
    animals.append(Cat());
}
# caller's ds[0] is now a Cat, but d: Dog = ds[0] still checks
```

**Expected vs actual**

- Expected: either (a) an invariance error — "list[Dog] is not assignable to list[Animal]" for mutable containers — or (b) documented, deliberate covariance with the unsoundness acknowledged.
- Actual: silently accepted as covariant.

**Suggested direction**

Design decision needed. Standard choice for mutable generics is invariance (like Java/Kotlin/TypeScript treat mutable collections); read-only views could stay covariant if the type system later gains immutability qualifiers. If invariance is too disruptive now, at minimum document the hole.

**Environment**

### LINKED ISSUE #8580: [Bug] Mutable containers checked covariantly: float appended to list[int], native reads raw bits / segfaults
### What's wrong

Mutable containers are accepted **covariantly**, which is unsound for any language with value-typed elements:

```jac
def poison(xs: list[float]) -> int {
    xs.append(2.5);
    return 0;
}
with entry {
    ints: list[int] = [1];
    poison(ints);          # accepted: list[int] passed as list[float]
    print(ints[1] + 1);
}
```

Checker: silent. Runtime consequences differ per backend:
- server codespace: prints 3.5 (float smuggled into list[int]; type system now lies)
- native pathway (value-element lowering): reading the element yields raw IEEE-754 bits reinterpreted as int (`4612811918334230529`); a `list[Leaf]` -> `list[Base]` obj-variant segfaults the native VM

### Suggested direction

Immutable reads may stay covariant, but mutable container parameters need invariance (or use-site variance annotations) — standard for sound gradual typing. At minimum the obj-element GPF case needs a guard.

Repro + bit-pattern proof (`0x4004000000000000`) available; double-confirmed.

chess10kp:
Closes #8580
Fixes #8550

## Summary

- Enforce declared generic variance when source and destination share the same class.
- Require invariant type arguments to be assignable in both directions.
- Apply the same bidirectional invariant check to generic base-class arguments.
- Add checker coverage for `list[int]` to `list[float]` and `list[Dog]` to `list[Animal]`.

## Root cause

`_assign_class` compared same-class generic arguments only as `source -> destination`. The generic checking implementation introduced this behavior in `225a2c9176` (#5159), so the defect was pre-existing on `upstream/main`.

## Verification

- `cd jac && JAC_TEST_JOBS=0 jac test tests/compiler/passes/test_typevar.jac` — 10 passed.
- Focused reproduction without `append` now reports two expected E1053 errors.


jac/jaclang/compiler/types/type_evaluator.impl/type_evaluator.impl.jac
```diff
@@ -2871,8 +2871,52 @@ impl TypeEvaluator._assign_class(
...
+                            continue;
+                        }
+                        if isinstance(src_arg, ClassType) and isinstance(dest_arg, ClassType) {
+                            if src_arg.is_builtin("LiteralString") and dest_arg.is_builtin("str") {
+                                continue;
+                            }
+                            if src_arg.is_builtin("str") and dest_arg.is_builtin("LiteralString") {
+                                continue;
+                            }
+                            if src_arg.is_builtin("object") or dest_arg.is_builtin("object") {
```
christianwilkins: [P2] Do not exempt `object` from mutable-container invariance. On this implementation, `def accept_objects(values: list[object]) -> None { values.append("not a dog"); }` accepts a `dogs: list[Dog]` argument without any checker error. `object` is a concrete supertype, so the callee can insert values that violate the caller's element type. This branch skips the reverse assignment that would reject it. Preserve gradual Any/Unknown behavior separately and enforce both directions here and in the matching generic-base branch.

jac/jaclang/compiler/types/type_evaluator.impl/type_evaluator.impl.jac
```diff
@@ -2871,8 +2871,52 @@ impl TypeEvaluator._assign_class(
...
+                            }
+                            if not self.assign_type(src_arg, dest_arg, strict) {
+                                return False;
+                            }
+                            if not self.assign_type(dest_arg, src_arg, strict) {
+                                return False;
+                            }
+                            continue;
+                        }
+                        if not self.assign_type(src_arg, dest_arg, strict) {
```
christianwilkins: [P2] Apply the reverse check to union arguments too. The invariant check only becomes bidirectional when both arguments are `ClassType`. For `Dog(Animal)`, `def accept_union(values: list[Animal | str]) -> None {}` still accepts `dogs: list[Dog | str]` with no checker error in my probe. Either argument being a UnionType reaches this one-way fallback, retaining the same mutable-container covariance hole. Add this negative case and use the invariant rule for concrete non-ClassType arguments in both comparison paths.

christianwilkins (changes requested):
Reviewed `2ebe6bb66a53b1235d3267c881047459dc2c0f20`, including the full patch, both generic comparison paths, parser callers, existing discussion and current-head CI. Two remaining soundness holes are attached inline: `object` arguments bypass invariance, and union arguments fall back to a one-way comparison. These are incomplete cases of the defect this patch is intended to fix; the original two regression cases do pass.

Independent validation on #9010 head `725f41cfe`: `jac test jac/tests/compiler/passes/test_typevar.jac` passes all 11 tests. Two additional `jac check` probes both incorrectly pass: `list[Dog]` passed to `list[object]` (whose body appends a string), and `list[Dog | str]` passed to `list[Animal | str]`, with `Dog(Animal)`. I compared the complete trees of #9007 and #9010: their compiler, parser, tests and other source are byte-identical; only the release-note filename differs. Thus this is one local validation run of their shared implementation, not two independent runs. `git diff --check` passes; no code changes were made during review.

The latest follow-up `2ebe6bb66` narrows the earlier builtin exemptions, but retains the `object` exception and non-ClassType fallback. Both PRs are conflict-free. Current-head CI is red because `jac fmt --check --lintfix` fails for `type_evaluator.impl.jac` and `test_typevar.jac`; the job logs include the required formatting patch.

Status: changes requested. Author work: enforce invariant assignment for these concrete argument forms in both same-class and generic-base paths, add the negative cases, apply formatting, and rerun the checks. This overlaps #9010 exactly, so coordinate which PR should land and close the duplicate after the chosen fix is accepted. No merge or closure was performed in this review.

PR #9010, opened 2026-09-07
fix(types): honor generic variance for same-class assignments

### LINKED ISSUE #8550: [Bug] Generic list covariance accepted: list[Dog] passes where list[Animal] expected (unsound)

chess10kp:
Closes #8550

Also fixes the same mutable-container covariance hole reported in #8580.

## Summary

`TypeEvaluator._assign_class` already applied declared variance when comparing a generic subclass with a generic base, but the same-class path always used one-way `assign_type`. This accepted `list[Dog]` as `list[Animal]` even though `list` is invariant.

This patch centralizes generic argument comparison and applies:
- covariant: source argument assigns to destination argument
- contravariant: destination argument assigns to source argument
- invariant: both directions must assign

The helper is used by both same-class and generic-base comparisons. Unknown destination arguments remain open for inferred tuple filters.

## Verification

- Minimal repro before the fix: `jac check` accepted `list[Dog]` passed to `list[Animal]`.
- After the fix: the repro reports `E1053: Cannot assign list[Dog] to parameter 'xs' of type list[Animal]`.
- Targeted test: `cd jac && jac test tests/compiler/passes/test_checker_bugs.jac --test_name bug_8550_same_class_generic_variance -x` (`1 passed`).


jac/jaclang/compiler/types/type_evaluator.impl/type_evaluator.impl.jac
```diff
@@ -2871,8 +2871,52 @@ impl TypeEvaluator._assign_class(
...
+                            continue;
+                        }
+                        if isinstance(src_arg, ClassType) and isinstance(dest_arg, ClassType) {
+                            if src_arg.is_builtin("LiteralString") and dest_arg.is_builtin("str") {
+                                continue;
+                            }
+                            if src_arg.is_builtin("str") and dest_arg.is_builtin("LiteralString") {
+                                continue;
+                            }
+                            if src_arg.is_builtin("object") or dest_arg.is_builtin("object") {
```
christianwilkins: [P2] Do not exempt `object` from mutable-container invariance. On this implementation, `def accept_objects(values: list[object]) -> None { values.append("not a dog"); }` accepts a `dogs: list[Dog]` argument without any checker error. `object` is a concrete supertype, so the callee can insert values that violate the caller's element type. This branch skips the reverse assignment that would reject it. Preserve gradual Any/Unknown behavior separately and enforce both directions here and in the matching generic-base branch.

jac/jaclang/compiler/types/type_evaluator.impl/type_evaluator.impl.jac
```diff
@@ -2871,8 +2871,52 @@ impl TypeEvaluator._assign_class(
...
+                            }
+                            if not self.assign_type(src_arg, dest_arg, strict) {
+                                return False;
+                            }
+                            if not self.assign_type(dest_arg, src_arg, strict) {
+                                return False;
+                            }
+                            continue;
+                        }
+                        if not self.assign_type(src_arg, dest_arg, strict) {
```
christianwilkins, line 2905: [P2] Apply the reverse check to union arguments too. The invariant check only becomes bidirectional when both arguments are `ClassType`. For `Dog(Animal)`, `def accept_union(values: list[Animal | str]) -> None {}` still accepts `dogs: list[Dog | str]` with no checker error in my probe. Either argument being a UnionType reaches this one-way fallback, retaining the same mutable-container covariance hole. Add this negative case and use the invariant rule for concrete non-ClassType arguments in both comparison paths.

christianwilkins (changes requested):
Reviewed `725f41cfed945b69b1cbdceac8f638545e68824a`, including the full patch, both generic comparison paths, parser callers, existing discussion and current-head CI. Two remaining soundness holes are attached inline: `object` arguments bypass invariance, and union arguments fall back to a one-way comparison. These are incomplete cases of the defect this patch is intended to fix; the original two regression cases do pass.

Independent validation on #9010 head `725f41cfe`: `jac test jac/tests/compiler/passes/test_typevar.jac` passes all 11 tests. Two additional `jac check` probes both incorrectly pass: `list[Dog]` passed to `list[object]` (whose body appends a string), and `list[Dog | str]` passed to `list[Animal | str]`, with `Dog(Animal)`. I compared the complete trees of #9007 and #9010: their compiler, parser, tests and other source are byte-identical; only the release-note filename differs. Thus this is one local validation run of their shared implementation, not two independent runs. `git diff --check` passes; no code changes were made during review.

The latest follow-up `725f41cfe` narrows the earlier builtin exemptions, but retains the `object` exception and non-ClassType fallback. Both PRs are conflict-free. Current-head CI is red because `jac fmt --check --lintfix` fails for `type_evaluator.impl.jac` and `test_typevar.jac`; the job logs include the required formatting patch.

Status: changes requested. Author work: enforce invariant assignment for these concrete argument forms in both same-class and generic-base paths, add the negative cases, apply formatting, and rerun the checks. This overlaps #9007 exactly, so coordinate which PR should land and close the duplicate after the chosen fix is accepted. No merge or closure was performed in this review.

Please also refresh the description: the current diff duplicates the comparison logic and adds `invariant_generic_arguments` in `test_typevar.jac`; it no longer matches the claimed centralized helper and `test_checker_bugs.jac` command.

Issue #9002, opened 2026-09-06
test(byllm): one fake, one seam, pure fixtures: tracking the byLLM suite refactor


MalithaPrabhashana:
## Summary

The byLLM suite is 10,361 lines holding 242 test blocks. The size is a symptom; the
cause is that a test has no shared way to reach the model, so every test builds its own
world. Seven different fakes, 66 fixture programs, and a fixture-to-test contract that
passes results back as strings through stdout.

This issue tracks the execution: one fake model at one seam, fixtures that are pure
programs, and one annex. It stays open until the work is done and is updated as each PR
lands. The gate was dropped by decision (section 6), and the fake became the public
`MockLLM`, rebuilt at the network layer (section 3.1).

**Relationship to #8930.** #8930 diagnosed the same suite and landed the annex
(#8936, #8938). This issue does not replace it. It adds three things #8930 does not
have: measurements of what the suite actually verifies, the verified mechanics for the
conversion (section 4 below, one of which invalidated my own first recipe), and a gate
as phase 0. Happy to fold this into #8930 and close this if maintainers prefer one
tracker; the content is written to merge cleanly either way.

## 1. Measured state

At `c91df5e46`. Commands in the appendix.

| Measure | Value |
|---|---|
| Test files | 12 files, 7,380 lines |
| Fixture programs | 66 files, 2,981 lines |
| Total | 10,361 lines |
| Test blocks | 242 (200 collected in the main lane, 29 in the sealed lane, 13 never collected) |
| Local run | 199 passed, 1 skipped, 322s |
| Assert statements | 842 (776 in test files, 66 inside fixtures) |
| Setup lines per assert line | 8.8 |
| Ways a test fakes the model | 7 |
| Test files importing `support_tests.jac` | 2 of 11 |
| Annex helpers with no caller | 3 of 18 (`Reply`, `Script`, `Scripted`) |
| Lines per fixture-driven test in `test_byllm.jac` | 123 (test body plus its fixture) |
| Lines per in-file test in the same file | 24 |
| Verbatim duplication (repeated 5-line windows) | 391 lines, 4% |
| Guard against regrowth | none |

**Correction to an earlier draft of this issue.** I opened with "10,361 lines is bloated".
That is wrong and I withdraw it. Measured against every other suite in the repo, byLLM
is unremarkable on size:

```
lines per test:  cli 19   language 20   project 20   scale/misc 20   microservices 22
[...]

PR #9014, opened 2026-09-07
test(byllm): def-only fixtures move into their tests (#9002 PR A)

MalithaPrabhashana:
Part of #9002. **Ready for review.**

## What this does

43 of byLLM's 66 fixtures were `def-only`: a whole `.jac` program whose entire content was one `by llm()` function, a model to answer it, a `with entry` to run it, and a `print` so the test could grep the answer back out of stdout. Zero `node`/`walker`/`edge` between them. This PR moves those into the test files, scripted with the annex's `FakeLLM`.

| | before | after |
|---|---|---|
| fixture programs | 66 | 26 |
| fixture lines | 3,047 | 863 |
| **asserts hidden inside fixtures** | **66** | **2** |
| fixtures constructing a `MockLLM` | 53 | 16 |
| stdout scraping | 114 | 35 |
| `run_fixture` calls | 42 | 5 |
| test blocks | 242 | 246 |
| **asserts in test files** | **776** | **844** |
| total lines | 10,439 | 9,448 |

51 files, +1,855 / -2,811, across 27 commits. Suite goes 199 passed / 1 skipped to **203 passed / 1 skipped**.

## Why

`MockLLM` overrides `dispatch_no_streaming`, four layers above `make_model_params` and above `parse_response`. Under it the prompt is built and thrown away, and the reply is never parsed. Two things reproduce what that costs:

- two functions with **opposite** `sem` strings return identical output, so a regression dropping every semstring, every `incl_info` example, or the whole JSON schema passed this suite
- `llm_semstrings.jac` declared `-> int` and had been returning a `str` in CI for as long as it existed

`FakeLLM` replaces only `model_call_*`, the network hop, so both halves are the real code and the outgoing request is readable as `llm.sent(...)`. Every converted test keeps the assertion it had and gains one it could not previously make.

## Shape

Before, two files:

```jac
# fixtures/with_llm_function.jac
glob llm = MockLLM(model_name="mockllm", config={"outputs": ["<emoji>"]}),
     emoji_examples = [...];
def get_emoji(input: str) -> str by llm(temperature=0.7, incl_info={"examples": emoji_examples});
with entry { print(get_emoji("Lets move to paris")); }

# test_byllm.jac
test "with llm function" {
    stdout_value = run_fixture("with_llm_function");
    assert "<emoji>" in stdout_value;
}
```

After, one:

```jac
test "incl_info examples and temperature reach the model" {
    emoji_examples = [...];
    llm = FakeLLM(replies=[say("<emoji>")]);

    def get_emoji(input: str) -> str by llm(
        temperature=0.7, incl_info={"examples": emoji_examples}
    );

    assert get_emoji("Lets move to paris") == "<emoji>";
    sent = llm.seen[0];
    assert sent["temperature"] == 0.7;
    assert "Mime Person" in str(sent["messages"]);
}
```

The last two assertions are new. `temperature=0.7` and two `incl_info` examples were set in the old fixture and checked by nothing; deleting either left the test green.

## What stays a fixture

- **6 graph programs** (`routing_*`, `math_poem_agents`, `streaming_visit_by`, `method_tool`): `node` / `walker` / `edge` where the shape is the subject. PR B.
- **~14 compile targets**: compiled or analyzed by `test_mtir_integration.jac`, never run.
- **`system_prompt_override/`**: byLLM resolves `[byllm] system_prompt` from the module's own location, so the defs have to sit beside that `jac.toml`. It is now a pure program (no model, no entry block, no prints) and the test assigns the model.

## Files split as the work went

Inlining moves fixture lines into the test file, so `test_byllm.jac` grew before it shrank (1,795 to 2,110 at the peak). Rather than let it peak near 2,400 and tear it apart in a later review, each converted cluster lands in a file named for what it tests.

| file | holds |
|---|---|
| `test_prompting.jac` | sem strings, incl_info, temperature, receiver state, typed returns, enums, scalars, nested objs, by-expression models, jac.toml config |
| `test_media.jac` | image, webp and video arguments, and every accepted `Image` source |
| `test_model_pool.jac` | aliasing, the fallback chain, the async bridge, streaming fallback |
| `test_streaming.jac` | what a `stream=True` call yields, `logging=True` StreamEvents, mid-stream retry, foreign-context consumption |
| `test_tools.jac` | finish_tool recovery, the react iteration limit, `on_iteration` hooks |
| `test_conversation.jac` | what `conversation=<list>` writes back |
| `test_typed_retry.jac` | regeneration after unparseable output |

## Coverage

The risk in this PR is a fixture taking its assertions with it: 66 of the suite's assertions lived inside fixture files behind 10 `_PASS` greps, and deleting one moves no test count. Every commit reports the assert count before and after, and every sentinel conversion writes its fixture-assert to new-home mapping into the commit message before the file is deleted.

In-fixture asserts 66 to 2. Asserts in test files 776 to 844. Test blocks 242 to 246, with every increase justified in its commit: a fixture that bundled two subjects becomes two tests.

Per the team tests policy, same-shape cases are tables rather than one test per input. Six went in that way: the react delivery pair, the `on_iteration` actions, the stream-retry error kinds, the conversation seed types, the pool strategies, and the image sources.

One test was deliberately merged: `image_test` and `with_llm_image` were separate fixtures testing the same thing, so both got converted before the duplication was noticed. The survivor keeps both of the weaker one's assertions in stronger form and adds a `roles()` check neither had.

## Defects the conversion surfaced

Six, each of which the old shape structurally could not see. All recorded on #9002.

1. **A parameter `sem` never reaches the model.** `sem fn.param` is accepted and silently dropped; a function-level `sem` does reach the prompt. `fixtures/image_with_sem.jac` declared one and asserted only the returned string, which the mock had been scripted with. The replacement test asserts today's behaviour and **fails when this is fixed**, with a message saying to invert it.
2. **`llm_typed_retry.jac` asserted `MockLLM`'s `parse_response` bypass as expected behaviour**: a plain string returned for a `-> Person` with no retry, described in the fixture as "MockLLM passthrough". Under real parsing the same input fails to convert and retries. That case is dropped rather than ported.
3. **`cost-based-routing` aliases as `pool-model`, not `pool-primary`.** Only `fallback` gives unique aliases. The old cost-based test never checked its alias.
4. **An unending transient stream error emits a `stream_reset` per retry**, two of them. The fixture asserted the call count and never the resets.
5. **`MTRuntime.resp_type` is annotated `type`** while callers pass `None` at runtime.
6. **A bare `enum { A, B }` generates a schema asking a real model to answer `1` or `2`.** No mock-backed test could notice, since the mock returned the member object directly.

## Two commits are not fixture conversion

**`fix(byllm)`: PIL handles in `media.impl.jac`.** `jac-check` went red with 13 errors: PIL types `Image.open()` as `ImageFile`, whose inherited `.format` and `.save` the checker cannot see. The file is not in `.jacignore`, so it is meant to pass. Five handles gain an `any` annotation. No behaviour change. It surfaced on this branch because the errors only appear once PIL actually resolves, which needs the `jac-check` lane where pillow is installed, and because the branch now annotates an `Image` parameter at module level. Carries a release-note fragment.

**A docstring trim**, comment-only, its own commit. Measured first: 111 docstring lines and 5 comment lines across 1,220, a 10% share, so a narrow pass over the over-length ones rather than a sweep. Nine lost a narrative clause and kept their contract. Untouched on purpose: docstrings on `by llm()` defs and on any function passed in `tools=[]`, since byLLM sends those to the model as prompt and tool description; also every `sem` and every test name. Verified by identical triple-quote counts per file, and by stripping docstrings and comparing the remainder: zero files with code changes, zero with comment changes.

## CI

Green, 24 checks. `test-packages-and-docs`, the lane that runs this suite, reports the count rather than only an exit code:

```
203 passed, 1 skipped in 222.59s
```

## Next

PR B converts the 6 graph fixtures, whose mechanics are already verified and written up in #9002 section 4. PR C moves the remaining test files onto the annex, renames `support_tests.jac` to `support.jac` to match the repo convention, and lands `TESTING.md`.


jac/jaclang/byllm/tests/test_conversation.jac
```diff
@@ -0,0 +1,183 @@
...
+        answer = stream_turn(llm, history, "How is Paris?");
+        assert answer , f"{label}: the stream produced nothing";
+        assert_written_back(label, history, answer);
+    }
+}
+
+
+test "a second streamed turn grows the same history and keeps the first answer" {
+    history: list = [];
+    llm = FakeLLM(replies=_stream_replies() + _stream_replies());
```
christianwilkins: [P3] Give the two turns different scripted answers. `_stream_replies() + _stream_replies()` makes `first == second`, so the two assertions at lines 173–174 check the same content. Replacing the first stored answer with a copy of the second would still satisfy those assertions and the message counts. The removed fixture used different answers; retaining that distinction would actually test that the first answer survives the second turn.

christianwilkins (commented):
Reviewed head `624ef70124093e80b5ecafde86e1452f75b884d7`, including all removed fixtures, the replacement tests and support helpers, the remaining main test file, and the media typing changes.

P3 coverage suggestion inline: the two streamed conversation turns use identical answers, so the assertion that the first answer survives cannot distinguish it from a copy of the second answer. Use distinct scripted answers to preserve that part of the old regression coverage. I found no blocking production-code defect in the patch; no source changes made during review.

Local validation with the optional byLLM dependencies installed: 26 tests passed across model-pool, streaming, tools, conversation and typed-retry; the separate prompting file passed all 13 tests. Independent media/main-file validation remains incomplete: I stopped media compilation after excessive memory use, and a separate main-file attempt reached a sampled 38.9 GB footprint without finishing. I have not established that this behavior is caused by the PR. The earlier broad run without all optional dependencies is not counted as a successful validation of this refactor.

Current-head CI is green and the PR is conflict-free. I checked the package job log (203 passed, 1 skipped) and the separate sealed MTIR log (29 passed). Commenting rather than approving because of the local validation limit and the coverage suggestion; author next step is the distinct-answer test improvement, with maintainer review of the CI evidence for the remaining test scope.

PR #9016, opened 2026-09-07
test(scale): each k8s e2e service app declares the nodes it writes, so the lane compiles again

MusabMahmoodh:
## What this changes

`microservice e2e (k8s, slim)` and `(k8s, official-image)` have been red on `main` since 2026-09-05. Both fail before they reach a cluster:

```
deploy failed: manifest generation failed: app 'products_app' entry
'.../k8s_e2e/products_app.jac' does not compile: line 8, col 29:
App 'products_app' imports 'Products', a node owned by app 'web';
nodes and edges never cross an app boundary
```

The fixture kept `Products`, `Product`, `Cart` and `Order` in `shared/models.jac` and imported all four into all three service apps. Its own docstring said why, and named the thing that changed:

> All services share one anchor store, so every service must be able to deserialize every node class that might appear under root. **Owner notes (enforced by convention, not by code):** products_app writes Products + Product, cart_app writes Cart, orders_app writes Order.

#8920 made that convention code. `E5108` refuses a `node` or `edge` that crosses an app boundary, because an app boundary is ACID within and eventual across, so a node class shared by four apps is exactly what the rule exists to stop.

The over-import turned out to be defensive rather than functional. Grepping each app for the names it imports:

| app | imports | actually references |
|---|---|---|
| `products_app` | Products, Product, Cart, Order | Products, Product |
| `cart_app` | Cart, Product, Products, Order | Cart |
| `orders_app` | Order, Cart, Product, Products | Order |
| `web` (`main.jac`) | qty_label | qty_label |

`Cart` in `orders_app` is the string `"Cart is empty"`; every other extra name appears on the import line and nowhere else. So each declaration moves into the app that writes it, and the three cross-app imports go away rather than being narrowed.

`shared/models.jac` keeps `qty_label` and its `import humanize`, which is the reason that module is load-bearing for this fixture: it is the declared pip dep reached by the client entry, and it pins that a deploy hands the build host the app's `[dependencies]`.

## Deleted

The four `node` declarations in `shared/models.jac`, and the three `import from shared.models { ... }` lines in the service apps. The stale docstring paragraph about a shared anchor store and convention-enforced ownership goes with them.

## What else this touches

Nothing outside `jac/jaclang/scale/tests/fixtures/k8s_e2e/`. No product code, no test code, no config keys. `main.jac`'s `qty_label` import is unchanged, so the client-build dependency assertion the fixture exists for is unchanged.

## Validation

- All four changed files pass `jac check --parse_only`.
- **Not run locally, stated plainly: the e2e itself.** This fixture is only exercised by the `microservice e2e (k8s, *)` lanes, which stand up a kind cluster and a real deploy; those lanes on this PR are the run of record, and they are currently the failing ones on `main`, so a green result here is the whole verification.
- Not addressed here: `orders_app.jac` still calls `view_cart()` and `clear_cart()` without `await`, which #8920's bridge contract says should be `E1042`. That is either a second fixture defect or a gap in the rule, it is not what is failing the lane today, and it deserves its own change rather than being folded into a red-CI fix.


christianwilkins (commented):
Reviewed `ac613ba5ff16946be29892bf4d58258f1b02b8a2`, including the full diff, all three service consumers, the retained shared helper, and the nightly workflow. No blocking defect found in the declaration moves or the awaited bridge calls.

I verified that the complete `Products`, `Product`, `Cart`, and `Order` declaration blocks are byte-identical after moving into their owner files. `qty_label` and its humanize dependency remain in the shared module. The follow-up `ac613ba5f` correctly makes `create_order` async and awaits both cart calls. `git diff --check` passes.

My local compile validation is incomplete. A basic no-codegen probe passed on both old and new fixtures, so it does not protect the change. With `force_target_program=True`, the new fixture instead stops on E1053 at `datetime.now(UTC)` in `orders_app`; the old fixture stops earlier on different container-type diagnostics in `cart_app`. These are local compiler results, not a reproduction of the reported deployment failure, and I have not demonstrated that the datetime diagnostic is caused by this patch. All temporary fixture substitutions were restored.

Current-head PR CI is green and conflict-free, but the real `microservice e2e (k8s, slim/official-image)` jobs are in `nightly.yml`, not the PR workflow. I did not run a cluster. Status: runtime validation still needs the author's kind run or a maintainer dispatch against this head; ordinary PR checks do not establish that those lanes are fixed.

One nonblocking handoff cleanup: please update the PR description to include the async-call fix and the nightly-only validation limitation already explained in your follow-up comment. The current description still says those calls are synchronous and that the PR lanes are the run of record.

PR #9017, opened 2026-09-07
fix(scale): read the app edge graph to a fixed point, so a deploy boots on dependencies not declaration order

### LINKED ISSUE #9012: jac scale deploy: app edge analysis always throws RuntimeError (dict changed size), so the fleet silently boots in declaration order
### Summary

`jac scale deploy` computes the app dependency graph inside a `try` that catches
`Exception`, and on a large workspace that computation always raises. The deploy
continues, so every deploy of this project prints:

```
WARNING - could not read app edges for 'main' (RuntimeError: dictionary changed
size during iteration); booting it in declaration order
```

and the fleet boots in **declaration order instead of dependency order**. Both
statements in the log are true and neither is loud: the graph was not computed,
and the deploy proceeds as if it had been.

Reproduces on **0.37.5 and 0.37.7**, 100% of runs, on a 177 module / 3 app
workspace.

### Root cause

`jaclang/scale/runtime/fleet.jac`, `_compiled_app_edges` (line 288):

```jac
for hub in prog._search_hubs() {          # line 311
    for mod in hub.values() {             # line 312  <-- iterating
        ...
        if not man.app_edges and not (man.walker_access or man.func_access) {
            try {
                prog.run_schedule(mod, get_boundary_analysis_sched());   # line 322
            } except Exception { continue; }
```

The dict being iterated is not only the user's. `_search_hubs`
(`compiler/driver/impl/program.impl.jac:376`) returns **two** hubs, and the
second is the self-host compiler's own:

```jac
hubs: list[dict[(str, uni.Module)]] = [self.mod.hub];
internal = self._get_compiler().selfhost.program;
if internal is not self {
    hubs.append(internal.mod.hub);
}
return hubs;
```

So the loop walks the compiler's module hub, and `run_schedule` runs boundary
analysis, which loads further compiler modules into that same hub. CPython then
raises `RuntimeError: dictionary changed size during iteration` on the next
step of the `for`.

Note where the existing `try` sits: it wraps only `run_schedule` and
`continue`s, so it cannot catch this. The raise comes from the `for` itself, one
frame out, and unwinds the whole function into the caller's
`except Exception` at line 369.

Snapshotting the iteration (`for mod in list(hub.values())`) is the obvious
shape of a fix, though whether newly registered modules should also be visited
is a design call I cannot make from outside.

### Why this is scale dependent, and what I could not do

I could not reduce this to a small project, and I would rather say so than hand
over a repro that does not fire. What did NOT reproduce it, all on 0.37.7:

- a 2 app workspace (web-app + service)
- 3 apps sharing a module that carries `node` / `edge` / `walker` archetypes
- a 40 module import chain of plain functions behind two services
- a 30 module client component tree under the web app

That is consistent with the mechanism above: the hub that mutates is the
compiler's, so it takes a compile heavy enough that boundary analysis pulls in
compiler modules that were not already resident. A small project never gets
there. It may also interact with #8970, which made the compiler tree an ordinary
citizen of the analysis cache.

A reproducer that does fire every time is jacBuilder at
`jaseci-labs/jacBuilder@migrate/jac-0.37.5`:

```bash
K8S_APP_NAME=jacbuilder K8S_NAMESPACE=jacbuilder \
SHARED_DATA_PVC=jacbuilder-shared-data \
jac scale deploy main.jac --dry-run 2>&1 | grep "app edges"
```

### Why it matters

Two things, one of them quiet:

1. Boot ordering silently degrades from dependency order to declaration order.
   For a fleet whose apps depend on each other that is a startup race that shows
   up under timing, not a clean failure.
2. It is logged at WARNING with the exception folded into the message, so it
   reads as a note rather than "the dependency graph was not computed".

The catch at line 369 is `except Exception`, with `ValueError` deliberately
re-raised just above it. A `RuntimeError` from a dict mutating mid-iteration is
a bug in the traversal, not an unreadable project, so it seems worth separating
from the genuinely-unreadable case the fallback exists for.

MusabMahmoodh:
## What this changes

Every `jac scale deploy` of a multi-app workspace warns and boots the fleet in declaration order:

```
WARNING - could not read app edges for 'main' (RuntimeError: dictionary changed
size during iteration); booting it in declaration order
```

`_compiled_app_edges` (`scale/runtime/fleet.jac:288`) iterated the compiler's module hubs live:

```jac
for hub in prog._search_hubs() {
    for mod in hub.values() {
        ...
        try {
            prog.run_schedule(mod, get_boundary_analysis_sched());
        } except Exception { continue; }
```

`run_schedule` runs boundary analysis on `mod`, which registers the modules it reaches into the very hub whose `.values()` the loop is walking, so CPython raises on the next step of the `for`. The `try` guards only `run_schedule`, so the raise comes out of the loop and unwinds the whole function; `consumer_provider_edges` catches it, warns, and the deploy proceeds with the ordering the graph was supposed to determine.

Snapshotting the iteration stops the crash and silently drops the modules registered during the walk along with any edges they declare, which is the same defect one layer down and harder to notice. So the hubs are re-read to a fixed point instead: collect the hub entries not yet seen, process them, repeat until a pass finds nothing new. Keyed by the hub key, so a module registered by the analysis is visited exactly once and contributes its edges.

Also added: a `man is None` guard after `run_schedule` reassigns it, which the original path did not have.

## What this does not change

The broad `except Exception` at the call site (`fleet.jac:369`) stays. Whether an unreadable app graph should fail a deploy rather than quietly downgrade it to declaration order is a real question and a behaviour change; it is separate from whether this walk is correct, and it should not ride along in a fix for the walk.

## Regression risk

`_compiled_app_edges` has one caller, `consumer_provider_edges`, which is reached from fleet assembly on the deploy path and from `jac run --fleet`. The function's contract is unchanged: same signature, same return shape, same ordering semantics downstream in `order_by_edges`. The loop now terminates on a monotonically growing `seen` set over a finite module set. A workspace with no cross-app imports takes exactly one pass, as before.

## Validation

Measured on a dev binary built from current `main` (`c89fd6b0e`) on my rig, not inferred.

- `jac test jaclang/scale/tests/apps/test_fleet_from_apps.jac` with the fix: **18 passed**.
- The same file against **main's unpatched `fleet.jac`**: **18 passed**. So the test does not fail pre-fix and is not a regression guard. It is renamed accordingly and stands as coverage of `consumer_provider_edges` reporting a declared cross-app import, which had none.
- Neither run logged `could not read app edges`, so the mutating walk was never reached on that workspace.
- I then probed three workspace shapes directly against unpatched `main` (bare cross-app import; apps over a shared helper that has imports of its own; five apps over the same chain). All three returned their edges cleanly, none raised. So **I have not reproduced #9012 synthetically**, and I am not claiming a test that protects this.

What the change rests on, stated plainly: the mutation during iteration is visible in the source (`run_schedule` registers into the hubs the loop is walking, and the `try` guards only the call, not the `for`), and #9012 reports it reproducing 3/3 on both 0.37.5 and 0.37.7 on a real three-app workspace. That is a reading plus someone else's reproduction, not my own failing test.

Left as a draft until either the reporter's workspace shape shows me what my probes are missing, or someone tells me the reading is enough for a change this size.

- `jac check --parse_only` passes on `scale/runtime/fleet.jac` and the test file.
- Not measured: whether the fixed point changes the edge set on a large real workspace. It can only add edges, never remove them.

Closes #9012.


christianwilkins (commented):
Reviewed `981ab95f3ba4e801dbd17520c10a403a621715dd`, including the full patch, both compiler hubs, and the registry fast path in `consumer_provider_edges`. The fixed-point traversal avoids advancing a live dictionary iterator across boundary analysis, and the post-analysis manifest guard is appropriate. No additional blocking code defect found.

I added a local review probe that calls `_compiled_app_edges` directly on the bridge workspace, bypassing the runtime registry shortcut identified in your comment. That probe plus the 18 imported fleet tests gives 19 passed. Restoring only the base `fleet.jac` gives the same 19 passed. This reaches the compiled path, but still does not reproduce hub growth during iteration; it is not a regression guard for #9012. The PR source was restored and the worktree is clean. The cited jacBuilder branch returned HTTP 404 through my account, so I could not run that full reproducer.

The extra `.jacignore` hunk matches #9057 and addresses the separate Pillow checking issue; it is unrelated to the graph algorithm. Current-head CI and pre-commit.ci are green and the PR is conflict-free. No fixes were made during review.

Status: the graph fix remains unverified against the reported failure, consistent with your existing validation correction. The useful next step is still a run of the reporter's workspace or a test that makes boundary analysis grow a hub and fails with the old loop. That needs the author/reporter's reproducer; a maintainer can assess the change on inspection, but a green ordinary PR run should not be presented as reproducing and fixing #9012.

.jacignore
```diff
@@ -43,6 +43,7 @@ jac/jaclang/byllm/llm.impl/modelpool.impl.jac
 jac/jaclang/byllm/llm.jac
 jac/jaclang/byllm/telemetry.impl/telemetry.impl.jac
 jac/jaclang/byllm/telemetry.jac
+jac/jaclang/byllm/types.impl/media.impl.jac
```
Thamirawaran, line 46: does this actually needed. it is not related to anything. isn't it?

Thamirawaran (changes requested):

PR #9020, opened 2026-09-07
test(scale): fold the config unit tests into per-axis tables

kugesan1105:
## Description

The scale config tests had grown one test per input value with identical setup. This folds each cluster into one test over a case table, where every row runs every assertion its siblings used to make separately. No assertion is dropped; one gets stronger.

| file | before | after |
| --- | --- | --- |
| `misc/test_config_loader.jac` | 17 | 7 |
| `misc/test_config_validation.jac` | 27 | 14 |
| `deploy/test_gateway_config_isolation.jac` | 6 | 3 |
| `apps/test_app_scale_overrides.jac` | 8 | 7 |

### What was folded

- **`get_gateway_config`, 8 to 1.** One table of raw configs and the values the normalized block must carry. The key-superset guard that used to run on one config now runs on all six, and the fleet-env check runs on every row.
- **`get_kubernetes_config` cache, 2 to 1.** Same setup, so one test asserts both that repeated calls agree and that nothing leaked into the cached dict.
- **Real-toml kubernetes, 3 to 1.** One project writes the sweep, holds seven keys back to prove they resolve to the declaration, derives `app_name` from `[project].name`, and carries the `[scale.admin]` hand-off. `app_name` is no longer written by the sweep, so the derived-slug path stays covered.
- **Unknown-key suggestions, 4 to 1 plus one row moved.** The `[scale.microservices]` finding moves here from the app-overrides file, where its siblings live. The non-dict-section case is a row too.
- **Clean configs, 5 to 1.** Empty, `None`, declared keys, open-dict sections, platform-injected keys, and documented gateway boot keys are one list.
- **Strict env, 4 to 1.** One setup and teardown for the four strict-mode behaviours.
- **Deploy targets, 3 to 1.** Factory list, per-target sections, and the registered-plugin case in one test.
- **`k8s_safe_name`, 4 to 1.** The RFC 1123 loop, the reserved gateway name, and the two refusals with their error text.

### Validation

All four files pass locally: 7, 14, 3, 7. `jac check` clean on all four, `jac fmt` no diff. Rebased on main after #8691, whose table-driven namespace test sits untouched in the loader file.

Part of the scale test consolidation plan; this is the pure-unit slice with no cluster or subprocess involvement.

### DIFF
--- a/jac/jaclang/scale/tests/misc/test_config_loader.jac
+++ b/jac/jaclang/scale/tests/misc/test_config_loader.jac
@@ -13,109 +13,64 @@ import from jaclang.scale.deploy.target.kubernetes.kubernetes_config {
 }
 
 
-test "get_gateway_config returns ingress block" {
-    cfg = JacScaleConfig();
-    cfg._config = {
-        "gateway": {
-            "ingress": {
-                "enabled": True,
-                "host": "jac-shop.local",
-                "ingress_class_name": "nginx",
-                "annotations": {"nginx.ingress.kubernetes.io/proxy-body-size": "10m"}
-            }
-        }
+glob _GATEWAY_KEYS: set[str] = {
+         "colocate",
+         "gateway_port",
+         "gateway_host",
+         "http_forward_timeout",
+         "rate_limit",
+         "cors",
+         "identity",
+         "ingress",
+         "logs",
+         "tracing",
+         "shared_volumes"
+     };
+
+
+test "get_gateway_config fills every section and never reads the fleet from the environment" {
+    ingress = {
+        "enabled": True,
+        "host": "jac-shop.local",
+        "ingress_class_name": "nginx",
+        "annotations": {"nginx.ingress.kubernetes.io/proxy-body-size": "10m"}
     };
-    out = cfg.get_gateway_config();
-    assert "ingress" in out;
-    ing = out["ingress"];
-    assert ing["enabled"] is True;
-    assert ing["host"] == "jac-shop.local";
-    assert ing["ingress_class_name"] == "nginx";
-    assert ing["annotations"]["nginx.ingress.kubernetes.io/proxy-body-size"] == "10m";
-}
-
-
-test "get_gateway_config ingress defaults to {} when section absent" {
-    cfg = JacScaleConfig();
-    cfg._config = {"gateway": {"colocate": False}};
-    out = cfg.get_gateway_config();
-    assert out["ingress"] == {};
-}
-
-
-test "get_gateway_config preserves all sections (regression guard)" {
-    cfg = JacScaleConfig();
-    cfg._config = {
-        "gateway": {
-            "colocate": False,
-            "gateway_port": 9000,
-            "http_forward_timeout": 60,
-            "replicas": 2,
-            "ingress": {"enabled": False}
-        }
-    };
-    out = cfg.get_gateway_config();
-    expected_keys: set[str] = {
-        "colocate",
-        "gateway_port",
-        "gateway_host",
-        "http_forward_timeout",
-        "replicas",
-        "rate_limit",
-        "cors",
-        "identity",
-        "ingress",
-        "logs",
-        "tracing",
-        "shared_volumes"
-    };
-    missing = expected_keys - set(out.keys());
-    assert not missing , f"missing keys: {missing}";
-    assert out["gateway_port"] == 9000;
-    assert out["replicas"] == 2;
-}
-
-
-test "get_gateway_config logs block defaults to {} when absent" {
-    cfg = JacScaleConfig();
-    cfg._config = {"gateway": {}};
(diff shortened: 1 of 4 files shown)


jac/jaclang/scale/tests/misc/test_config_loader.jac
```diff
@@ -314,10 +231,22 @@ test "every [scale.kubernetes] key a user writes reaches the resolved config" {
         "prometheus_admin_password",
         "service_dependencies"
     };
+    # Left out of the toml, so they must resolve to the declaration itself;
+    # app_name is left out so it derives from [project].name instead.
+    unwritten: set[str] = {
```
MusabMahmoodh, line 236: Holding these seven back inside the sweep is not the same coverage the deleted test had. `an unwritten [scale.kubernetes] key resolves to the declaration` drove a `jac.toml` with **no `[scale.kubernetes]` section at all**; this one always writes the section and omits keys from it. A loader that only applies declarations when the section is present passes here and failed there, and after this PR nothing in `misc/` loads a kubernetes config from a toml without the section (the only other `write_text` is the monitoring helper at :168).

One more row on the outer test, a project with just `[project]`, restores it.

jac/jaclang/scale/tests/misc/test_config_validation.jac
```diff
@@ -42,61 +42,74 @@ def _collect_default_mismatches(
...
+        ),
+        (
+            {"gateway": {"gateway_prot": 8001}},
+            ["scale.gateway", "did you mean 'gateway_port'"]
+        ),
+        (
+            {"deploy": {"target": "kuberntes"}},
+            ["kuberntes", "did you mean 'kubernetes'"]
+        ),
+        ({"microservices": {"routes": {}}}, ["microservices"]),
```
MusabMahmoodh, line 61: This row carries the finding text but not the second half of the test it replaced. The version in `apps/test_app_scale_overrides.jac` also asserted `"gateway" in JacScaleConfig().get_config_schema()["options"]`, which pinned that the table replacing `[scale.microservices]` is actually declared. Nothing asserts that now: the schema-declared loop in `test_config_loader.jac:100` covers websocket, database, kubernetes, events, monitoring and server, and `gateway` is not in the list.

Adding `gateway` to that loop is probably the better home for it than carrying it here.

MusabMahmoodh (approved):
Good refactor, and the part I like most is not the line count: turning the key-superset guard into something that runs on all six configs instead of the one row that happened to set every key is a real strengthening. Dropping `replicas` from `_GATEWAY_KEYS` to make that possible is also right, since `normalize_gateway_config` only carries it through `GATEWAY_POD_KEYS` when the user wrote it, so the old guard was only ever passing because its single row set it.

Two assertions did get dropped though, against the body's "No assertion is dropped". Both are cheap to put back as rows and neither blocks this.

Heads up on a collision: my #8872 also edits `misc/test_config_loader.jac`. Whichever of these lands second will need a rebase, and I am happy for that to be mine.

jac/jaclang/scale/tests/misc/test_config_loader.jac
christianwilkins, line 236: [P3] Keep explicit overrides for the held-back settings. Previously the sweep wrote nondefault values for `min_replicas`, `max_replicas`, both utilization targets, readiness delay, liveness threshold, and `container_port` (plus `app_name`). Adding them to `unwritten` now tests their defaults only. A loader that silently discards an explicit `min_replicas` value would pass this row, whereas the old written-key sweep rejected it. Keep an all-written row as well as the default/derived row. This is separate from the missing-section case already discussed above; I could not run the mutation because the local suite stops on the unchanged KubernetesConfig import.

christianwilkins (commented):
Reviewed `00ef34d0a7ac880f9718f24751a8dde184a351dc`, including all four changed files, the loader/normalizer, current reviews and both open inline threads. The per-row gateway key guard is stronger. One additional nonblocking coverage gap is attached inline: the Kubernetes sweep no longer writes the settings moved into `unwritten`, so it no longer checks that user overrides for those settings survive loading. This is separate from the two cases already identified by Musab; I have not duplicated those comments.

Independent local validation: the app override file passes all 7 tests. The other three files stop during import with E1053 at `fields(KubernetesConfig)` in unchanged `kubernetes_config.jac` lines 119 and 130 (overall 7 passed, 3 file errors). Restoring the base loader test in the same checkout produces the same import error; all temporary substitutions were restored. Thus this run does not reproduce the author's 31 passing tests, and the proposed override-loss mutation could not be meaningfully run. `git diff --check` passes.

Current-head CI is green and GitHub reports no conflicts. This PR contains one commit; no fixes or follow-up changes were made during review. Status: comment only, with local validation incomplete because of that compiler/import incompatibility. Please preserve both explicit-value and default-value rows when addressing the existing coverage feedback; no production regression is demonstrated by this review.

PR #9037, opened 2026-09-07
feat(client): scaffold Expo SDK 57 apps

mahfoos:
The react-native scaffold pinned Expo SDK 54. Expo Go on the App Store is a single build that supports only the newest SDK, so scanning a scaffolded project's QR code on a real phone is refused with "requires a newer version of Expo Go" - which reads as the user's project being broken rather than the scaffold being three majors behind.

Only iOS **simulator** clients are published per SDK (`iosClientUrl` in Expo's versions API), so there is no per-SDK client that can meet a 54 project on a physical device. The scaffold has to move.

## What changes

- `EXPO_SDK_VERSION` 54 -> 57, `REACT_NATIVE_VERSION` 0.81.4 -> 0.86.3.
- The nine packages `scaffold.jac` pins by hand. **Since SDK 57 Expo versions its own packages with the SDK major**, so `expo-asset` goes `~12.0.7 -> ~57.0.16` and `expo-constants` `~18.0.7 -> ~57.0.17`. Those jumps look like typos and are not.
- `app.json` drops `newArchEnabled` and `jsEngine`.
- `EAS_CLI_VERSION` stays 20. It is versioned independently of the SDK and an existing test already pins that.

Every version is what `expo install --check` resolves for SDK 57, taken from that SDK's `bundledNativeModules.json`, not chosen by hand.

## Why app.json loses two keys

SDK 57 removed `newArchEnabled` and `jsEngine` from the Expo config schema, because the New Architecture and Hermes are now the only options. Verified against the schema API rather than assumed:

| key | schema 54 | schema 57 |
|---|---|---|
| `newArchEnabled` | present | **absent** |
| `jsEngine` | present | **absent** |

`expo-doctor` fails any project that still declares them, so every app scaffolded by a 57 toolchain would have been born failing its own doctor check. The intent behind D8 (New Architecture + Hermes from day one) is unchanged; *stating* it in app.json is now the error. The assertion that required their presence now requires their absence and says why.

## Verification

Run against a scaffold generated by this code (`write_expo_scaffold` into a temp dir), not a hand-written mirror of it. I diffed the emitted `package.json` against the one I validated and they are byte-identical.

```
npm install            518 packages, no peer conflicts
expo install --check   Dependencies are up to date
expo-doctor            21/21 checks passed. No issues detected!
```

`expo-doctor` is what caught the app.json break. The version pins alone looked complete and were not.

## New test

The existing scaffold assertions only check that a package is *present* (`assert "expo-asset" in pkg_content`), so a half-landed bump - `expo` at 57 while `expo-asset` is still 12 - passes every one of them and surfaces on a user's machine as a metro resolution error. The new test pins the majors instead.

Shown failing before it passed. Reverting `expo-asset` alone:

```
FAILED tests/client/test_mobile_target.jac::the scaffold's expo packages all track the SDK the scaffold pins
E   AssertionError: expo-asset is '~12.0.7', which is not an SDK 57 release.
```

## Test results

`test_mobile_target.jac` plus the three neighbouring client suites: **71 passed, up from 70**, with the same two failures before and after.

Both pre-existing and environment-dependent, failing identically on a stashed clean tree:

- `resolve_rn_port_or_next_available bumps away from a busy port`
- `desktop launch wires JAC_DESKTOP_DEPS from the project venv`

`jac fmt --check` and `jac check` clean on all three files.


christianwilkins (commented):
Reviewed `d3f624db6aec5b4bd25ee7326d13a77aa8335d51`, including the full diff and the scaffold generation/refresh path. No blocking finding in the SDK update.

Independent validation used a project generated by `write_expo_scaffold` from this checkout. All nine dependencies that overlap Expo's [SDK 57 bundled-native manifest](https://github.com/expo/expo/blob/sdk-57/packages/expo/bundledNativeModules.json) match exactly. `npm install --ignore-scripts` completed, `expo install --check` reports dependencies up to date, and `expo-doctor` passes all 21 checks, including the generated app configuration.

`test_mobile_target.jac`: 54 passed, 1 failed, 2 skipped. I reran the failing busy-port case on main `3ad7d91b5`; it fails with the same “Expected auto-port fallback to move past busy port” assertion, so that local failure is independently established on main. `git diff --check` passes. No code changes were made during this review. I did not build or launch an iOS app, or test a physical phone.

The head remains conflict-free and pre-commit.ci is green. Current-head CI run 34213532959 is red in the compiler passes-main lane; GitHub could not supply that failed job's log, so I have not established its cause or classified it as a baseline failure. Status: no author changes requested to the scaffold from this review, but the failing CI lane must be resolved or explained and successfully rerun before merge readiness is confirmed.

Issue #8773, opened 2026-08-29
[Bug]: ${VAR:?message} in jac.toml is silently inert -- the required-variable error is swallowed


MalithaPrabhashana:
### Summary

`${VAR:?message}` in `jac.toml` is documented as failing with a custom error when
`VAR` is unset. It is also implemented that way. But the caller catches the
`ValueError` and returns the **raw string**, so the whole construct is inert: the
literal text `${K8S_NAMESPACE:?must be set}` becomes the configured value.

The plain `${VAR}` form has the same failure mode -- an unset variable yields the
literal `${VAR}` rather than an error.

### Why this matters

`:?` exists to stop a misconfigured deploy. Instead it produces a garbage value
and proceeds. In our case that value was a Kubernetes **namespace**:

```toml
[scale.kubernetes]
namespace = "${K8S_NAMESPACE:?set K8S_NAMESPACE}"
```

```
$ jac start --scale --dry-run
Namespace: ${K8S_NAMESPACE:?set K8S_NAMESPACE}
check: no errors or warnings
```

`--dry-run` reports the plan as clean while the namespace is a literal shell
expression. The one guard rail available for "this must be set before you deploy"
silently becomes the thing it was meant to prevent.

### Root cause

`jac/jaclang/project/impl/config.impl.jac` -- `interpolate_env_vars` is correct
and raises:

```jac
} elif operator == ":?" {
    raise ValueError(f"Environment variable {var_name} is required: {operand}");
}
...
raise ValueError(f"Environment variable {var_name} is not set");
```

and `_interpolate_recursive` throws that away:

```jac
impl _interpolate_recursive(`obj: any) -> any {
    if isinstance(`obj, str) {
        try {
            return interpolate_env_vars(`obj);
        } except ValueError {
            return `obj;              // <-- both raises land here
        }
    }
    ...
}
```

### Reproduce

```toml
[project]
name = "interp"
entry-point = "main.jac"

[scale.kubernetes]
namespace = "${NOPE:?you must set NOPE}"
```

```
$ jac start --scale --dry-run
Namespace: ${NOPE:?you must set NOPE}      # expected: abort with "you must set NOPE"

$ NOPE=works jac start --scale --dry-run
Namespace: works                            # interpolation itself is fine
```

`${VAR:-default}` behaves correctly throughout.

### Suggested fix

Let the `ValueError` propagate, at minimum for `:?`. The two raises are
distinguishable if the plain-unset case must stay lenient for back-compat --
[...]

PR #9045, opened 2026-09-07
fix(config): report unresolved env vars in the command that uses them (#8773)

akindu-k:
## Summary
image

Refs #8773. It does not fully close here: `${VAR:+alt}` and `${VAR:0:3}` still pass through verbatim, tracked with the other parser and consumer gaps in #9184.

`jac.toml` lets you write `${VAR}` in a value, and Jac swaps in an environment variable when it reads the file. `${VAR}` and `${VAR:?message}` are documented as errors when the variable is not set, but the function that walks the config caught the error and returned the raw text. So `jac scale deploy` could ship a Kubernetes namespace named `${K8S_NAMESPACE:?set K8S_NAMESPACE}`, or a Secret whose value is the literal string `${OPENAI_API_KEY}`, while `--dry-run` reported no errors.

## Why this was reworked

The first version of this PR made the error fatal when the config loads. Every command loads `jac.toml`, so that turned an unset deploy variable into a failure in `jac check`, `jac fmt` and `jac lsp`, and it broke two jac-scale paths Musab pointed out: pods booting against the shipped `jac.toml`, and scale commands whose variables come from a project `.env` that is read after the config loads.

The root cause is that interpolation answered "did this resolve?" at load time and then threw the answer away, which left only "fail every command" or "fail none". This version keeps the answer.

## What changes

Loading the config never fails on an unset variable. It records each unresolved reference, with the setting that holds it, on `JacConfig.unresolved`. The command that uses the value decides what to do. This is the same split the schema validator already uses: `validate_against_schema` returns findings and `report_schema_findings` applies the policy.

| Command | `${VAR}` not set | `${VAR:?message}` not set |
|---|---|---|
| `jac check`, `jac fmt`, `jac build`, `jac lsp` | silent | silent |
| `jac run`, including `--serve` | warning | exit 2 |
| `jac scale deploy`, `--dry-run` included, for `[scale.secrets]` and `[scale.kubernetes]` | exit 1 | exit 1 |
| `jac scale deploy`, the rest of the shipped `jac.toml` | checked against each pod's environment by #9199 | same |

`jac run` only checks the sections it reads: `project`, `run`, `build`, `native`, `memory`, `placement`, `serve`, `cache`, `storage`, `environment`, `client_errors`, `check`, `plugins` and the capability sections. `[scripts]` (read by `jac x`), `[test]`, `[format]`, `[dev]`, the dependency tables and unknown tables belong to other commands or to none, so a shell `${VAR:?}` in a script never stops a run. `[scale.secrets]` and `[scale.kubernetes]` are left to the deploy. `[check]` is included because the compiler reads `check.enforce_access` on every run. `${VAR:-default}` is unchanged everywhere.

### Why `[scale.kubernetes]` and not only `[scale.secrets]`

The deploy host reads both sections itself. Secrets are removed from the bundle, and `[scale.kubernetes]` builds the manifests. #9199 cannot catch #8773's own case: pods are given `K8S_NAMESPACE` set to the configured value, so a literal namespace looks resolved from inside the pod. Before this change that project deployed a dry run with `namespace: ${K8S_NAMESPACE:?set K8S_NAMESPACE}` in every manifest and exited 0.

## How it works

- `resolve_env_refs(data, environ)` walks the config, resolves what it can and returns the unresolved references with their paths, every one of them when a value holds several. `interpolate_env_vars` stays the strict primitive. It raises `UnresolvedEnvVar`, a `ValueError` that also carries the variable name and whether it was `:?`.
- #9199's `unresolved_toml_references` is now a thin wrapper over the same function, so the question "which settings did not resolve" has one implementation instead of two.
- Layers follow the config. A layer that sets a value that is not a table replaces it whole, lists included, so it clears every record under that key: `cors_origins = []` in `jac.local.toml` drops a `${ORIGIN:?}` the base file held. Tables merge key by key, so an empty table clears nothing. An inline `[environments.<name>]` table only counts once that profile is applied.
- The deploy check runs inside `_check_pod_config_resolves`, before anything reaches the cluster. It walks the key segments each reference was found under, so nested settings such as `http_activation.interceptor_service_address` are checked and quoted keys holding dots stay whole, and flags a setting only when the literal `${...}` text is what the deploy is actually about to use, so a secret replaced by the active profile, or given a default, passes.

Removed from the first version, because nothing fails at load any more: the CLI exit guards in dispatch and `execution.impl.jac`, `jaclang/cli/config_errors.jac`, the `jac-client` doc change about `NODE_AUTH_TOKEN` (npm resolves the placeholder itself again) and the host-side exports in #9199's `test_pod_env.jac` fixture. Those files match `main`.

`scale/admin/service_urls.jac` keeps the change from the previous review: `expandvars` is gone and the `"${" not in ...` guard stays.

## Checked by hand

Run on this branch through the real CLI.

The reproduction project, with `OPENAI_API_KEY` and `MUST_BE_SET` in `[scale.secrets]`:

```
$ jac check main.jac                      # exit 0
$ jac fmt main.jac                        # exit 0
$ jac run main.jac                        # exit 0, secrets are the deploy's

$ jac scale deploy main.jac --dry-run     # both unset, no .env
[ERROR] Error: --dry-run failed: jac.toml reads environment variables this machine does not have, so the deploy would carry these settings as literal text:
  scale.secrets.OPENAI_API_KEY: Environment variable OPENAI_API_KEY is not set
  scale.secrets.HARD_REQUIRED: Environment variable MUST_BE_SET is required: set it
EXIT=1

$ jac scale deploy main.jac --dry-run --show-yaml     # same values in .env
  OPENAI_API_KEY: sk-from-dotenv
  HARD_REQUIRED: 'yes'
EXIT=0
```

#8773's case, `namespace = "${K8S_NAMESPACE:?set K8S_NAMESPACE}"`:

| | result |
|---|---|
| deploy dry run, unset | exit 1, names `scale.kubernetes.namespace` |
| deploy dry run, exported | exit 0, namespace `prod` |
| `jac run`, unset | exit 0, the value is the deploy's |

`jac run` on values it does read: a bare `${VAR}` prints a warning naming the setting and runs, and a `${VAR:?}` exits 2 with the message.

`jac lsp`, over stdio, on a project with four unset variables including two `:?`:

| | result |
|---|---|
| previous head of this PR | exits 2 before answering `initialize` |
| this branch | initializes, publishes diagnostics for the open file, prints nothing about the variables |

## Tests

| Suites | Before | After |
|---|---|---|
| `tests/project`, `tests/cli`, `tests/client/test_cli.jac`, `test_service_urls.jac`, `test_pod_env.jac` | 776 (772 passed, 3 failed, 1 skipped) | 775 (772 passed, 2 failed, 1 skipped) |

The failures are unrelated and fail on `main` too: allocator memory and the fmt parallel pool, plus precompile discovery on some runs. `tests/client/test_cli.jac::start dev prepares the client before initialization` fails intermittently in a parallel run while the compiler cache rebuilds, and passes on its own (28 of 28). The whole `jaclang/scale/tests` tree gives 7 failed and 19 errors, and the same files give the same result with this change stashed. They are missing optional packages (python-dotenv, prometheus, gcs) in my environment.

Per test file, against `main`:

| File | `main` | here |
|---|---|---|
| `tests/project/test_config.jac` | 129 | 130 |
| `tests/cli/test_run_project_scope.jac` | 9 | 10 |
| `jaclang/scale/tests/deploy/test_pod_env.jac` | 41 | 42 |
| `jaclang/scale/tests/misc/test_service_urls.jac` | 5 | 5 |
| `tests/client/test_cli.jac` | 28 | 28 |

Each of the three new tests is the only test for a code path this PR adds, with no existing test to extend. All are table-driven and in-process.

- `tests/project/test_config.jac`, new, for recording at load: one table over the layers. It covers `${VAR}` and `${VAR:?}` in `jac.toml`, a `:?` after a bare reference in the same value, `:?` in a profile file, a value the active profile file sets again, an inactive and an active inline profile, a default, a `:?` inside a list, a list the active profile file empties, and an empty table that leaves the record in place.
- `tests/project/test_config.jac`, extended: the shell-expansion cases (`${#ARR[@]}`, `${VAR%%.txt}`, `${VAR^^}`, and `${1:?}` still raising) are rows in the existing `interpolate against a given environment` table, not a separate test.
- `tests/cli/test_run_project_scope.jac`, new, for the run check: the cross-project run test becomes a table over the run policy. A bare value warns and runs. `:?` in the base file, in the active profile, and after a bare reference in one value exits 2. A default, a deploy secret, a deploy target setting, a shell script, a script in the active profile and an unknown table all run. The dispatch-guard test is removed along with the guards it pinned.
- `jaclang/scale/tests/deploy/test_pod_env.jac`, new, for the host-side deploy check: an unset secret, an unset `:?` namespace, an unset nested setting (`http_activation.interceptor_service_address`), a quoted dotted secret key and label key, a secret the active profile replaces, and a default.
- `test_service_urls.jac`: the existing case is renamed and gains a `${VAR:+alt}` row.

Every fix added its failing rows first. Then each piece was reverted or narrowed, and a row catches each change:

| Change | Row that catches it |
|---|---|
| Profile promotion or layer replacement removed | `test_config.jac` |
| Run check skipped | `test_run_project_scope.jac` |
| Run check turned into a `scripts`-only exclusion | the unknown-table row |
| An empty table treated as a replacement | the empty-table row |
| Deploy check skipped | `test_pod_env.jac` |
| Kubernetes keys split on dots again | the dotted label row |
| The variable-name guard in `interpolate_env_vars` removed | the `echo ${1}` row |

## Not in this PR

- `$${VAR}` as an escape for a literal `${VAR}`, and the `:+` and `:0:3` forms: #9184.
- `load_project_env` requires python-dotenv even when the project has no `.env`, so `jac scale deploy` fails on a project without it. Separate issue.


jac/jaclang/project/impl/config.impl.jac
```diff
@@ -1310,11 +1310,7 @@ impl _parse_toml_data(
 
 impl _interpolate_recursive(`obj: any) -> any {
     if isinstance(`obj, str) {
-        try {
-            return interpolate_env_vars(`obj);
-        } except ValueError {
-            return `obj;
-        }
+        return interpolate_env_vars(`obj);
```
MusabMahmoodh: This runs on every layer at `JacConfig.load`, so the reach is wider than deploy. A project with `[scale.secrets] API_KEY = "${OPENAI_API_KEY}"` or a `[scale.kubernetes] namespace = "${K8S_NAMESPACE}"` can no longer run `jac check`, `jac fmt` or `jac run` without the deploy-time variable exported, on commands that never read the value.

I think that is the correct call, since `${VAR}` is documented as required and half-honoring it is what got us here. But it is a bigger behavior change than "unset variables now fail" reads, and the release note is where a user will look for it. Worth a sentence saying to export the variable or switch to `${VAR:-default}` for values only some commands need.

jac/jaclang/project/impl/config.impl.jac
```diff
@@ -1310,11 +1310,7 @@ impl _parse_toml_data(
...
     } elif isinstance(`obj, dict) {
         return {k: _interpolate_recursive(v) for (k, v) in `obj.items()};
```
MusabMahmoodh: The error names the variable but not where it was written, so on a toml with several `${...}` the user gets `Environment variable VAR is not set` and then goes looking. The dict branch already has the key in hand, so catching and re-raising with the path prefixed gets `scale.kubernetes.namespace: Environment variable K8S_NAMESPACE is not set` for a few lines. Not blocking, but this PR is about making the failure legible and that is the other half of it.

MusabMahmoodh (approved):
Right fix, and deleting the catch rather than special-casing `:?` is the right shape: the swallow was making `${VAR}` mean two different things depending on whether anyone had thought about it. The two new tests drive real toml through the real loader and cover the overlay path, which is where I would have expected the gap.

One thing before merge and one thing that is not yours, both below. `jac-check` red here is the runtime-cache flake, not your diff: the job dies in "Warm the runtime cache" on `rmtree: could not remove .../rt/f4ca7e3e.../openai/types/conversations/__pycache__`, before it reads a single file. A re-run clears it. Linus hit the same thing on #8175 yesterday, so I will file it separately rather than have people keep re-running.

jac/jaclang/project/impl/config.impl.jac
akindu-k: Confirmed the reach. With `[scale.kubernetes] namespace = "${K8S_NAMESPACE}"` and nothing exported, `jac check`, `jac fmt` and `jac run` all stop, none of which read the value.

Reworded the release note to say so and to point at the remedy:

> Every command that reads `jac.toml` enforces this, so export the variable or switch to `${VAR:-default}` for values that only some commands need.

Also added it to the interpolation section of the config reference, since that is where someone lands after hitting it rather than in the release notes.

jac/jaclang/project/impl/config.impl.jac
akindu-k: Done. `_interpolate_recursive` now threads the config path down and the string branch re-raises with it prefixed, so the walker owns the context and `interpolate_env_vars` stays a plain string API.

```
✖ Error: jac.toml is invalid: scale.kubernetes.namespace: Environment variable K8S_NAMESPACE is not set
✖ Error: jac.toml is invalid: scale.kubernetes.namespace: Environment variable K8S_NAMESPACE is required: set K8S_NAMESPACE for deploys
```

The list branch enumerates too, so array elements are located as well:

```
✖ Error: jac.toml is invalid: project.authors[0]: Environment variable JAC_MISSING_AUTHOR is not set
```

The existing table-driven test now asserts `scale.secrets.SOME_API_KEY` in the message across all three cases.

MalithaPrabhashana (commented):
We filed #8773, so here is a validation pass from our side. Ran this on macOS
against `pr-9045` (2eb371b0) with `main` at `3ad7d91b5` as the baseline, both in
dev mode off the checkout so the config code under test is really yours.

## The reported bug is fixed

Same project from the issue, `jac scale deploy --dry-run --show-yaml`, neither
variable exported.

`main` today, exit 0, and the literal ships into the manifests:

```yaml
kind: Secret
metadata:
  name: interp-repro-secrets
  namespace: ${K8S_NAMESPACE:?set K8S_NAMESPACE}
stringData:
  SOME_API_KEY: ${A_VAR_NOBODY_SET}
```

(`namespace: ${K8S_NAMESPACE:?set K8S_NAMESPACE}` appears in 5 places.)

This branch, same command, exit 2:

```
✖ Error: jac.toml is invalid: scale.kubernetes.namespace: Environment variable K8S_NAMESPACE is required: set K8S_NAMESPACE
```

With both variables set it is exit 0 again and the real values land
(`namespace: prod-ns`, `SOME_API_KEY: real-key-123`), so the happy path is
untouched.

Also checked and fine: `${VAR:-default}` unchanged, list elements are located
(`project.authors[0]`), deep tables get the full path
(`scale.kubernetes.resources.limits`), `run` / `check` / `fmt` / `build` /
`clean` / `config show` all give the same clean exit 2, and `jac --help` and
`--version` still work with a broken config.

Suites on this machine:

| suite | main | this branch |
|---|---|---|
| `tests/project/` | 320 passed, 3 failed | 321 passed, 3 failed |
| `tests/cli/` | 342 passed, 6 failed, 1 skipped | 342 passed, 6 failed, 1 skipped |
| `tests/client/test_cli.jac` | 28 passed | 28 passed |

The 3 `test_template.jac` failures are already red on main, not from this PR.
I also reverted only the `except` hunk on your branch and re-ran
`tests/project/test_config.jac`: 129 passed with the fix, 128 passed + 1 failed
without it, so the new test really does pin the behaviour. The branch already
contains current main, nothing to merge.

Four things I ran into.

## 1. The profile overlay path still ends in a traceback

Put the value in `jac.prod.toml` instead of `jac.toml` and activate the profile:

```
$ JAC_PROFILE=prod jac run src/main.jac        # BT_UNSET not exported
⚠ Failed to apply profile: scale.kubernetes.namespace: Environment variable BT_UNSET is required: you must set BT_UNSET for prod
✖ Error: Error executing 'run': scale.kubernetes.namespace: Environment variable BT_UNSET is required: you must set BT_UNSET for prod
  ...
  File ".../jaclang/project/impl/plugin_config.impl.jac", line 109, in get_jac_config
    config.apply_profile_overlay(profile or None);
  ...
ValueError: scale.kubernetes.namespace: Environment variable BT_UNSET is required: you must set BT_UNSET for prod
exit 1
```

Two spots miss it. `_apply_profile` (`cli/impl/dispatch.impl.jac:129`) catches
`ValueError` and only warns, and `PluginConfigBase.get_jac_config`
(`project/impl/plugin_config.impl.jac:109`) has no handler at all. The base file
is fine because `_discover_project_config` catches it and exits 2 cleanly.

Still better than main, which is exit 0 and `RAN OK` for the same run. But it
does not look like the rest of the feature. Worth noting your new overlay test
calls `merge_from_toml_file` directly, so it never goes through either handler
and stays green either way. This is the only one I would want looked at before
merge.

## 2. Only one missing variable is reported per run

The dict comprehension raises on the first bad key, so you fix them one at a
time. On jacBuilder's real `jac.toml` that is 47 runs before the config loads:

```
run 1  -> ECR_REGISTRY
run 2  -> K8S_APP_NAME
run 3  -> K8S_NAMESPACE
...
run 47 -> SHARED_DATA_PVC
```

Not a blocker, but collecting them and printing all at once would save a lot of
time on big configs. It also buries other errors: our file additionally has a
`[scale.microservices] was removed` error, and that message only appears after
all 47 variables are set.

## 3. Inline `[profiles.*]` is strict, `jac.prod.toml` is lazy

```toml
[profiles.prod.scale.kubernetes]
namespace = "${BT_UNSET:?prod only}"
```

in `jac.toml` fails every command even when the prod profile is never used. The
same content in `jac.prod.toml` is only read when the profile is active, so it
does not fail. That asymmetry is probably fine, but it is worth a line in the
docs, because moving deploy-only variables into an overlay file is the easy
answer to the "you now have to export everything" note in your description.

## 4. The jac-client doc line is not quite right

It now suggests `${NODE_AUTH_TOKEN:-}` to keep a registry entry optional. That
writes an empty token, not a pass-through. Measured through `create_npmrc`:

```
${NODE_AUTH_TOKEN}                     -> config rejected
${NODE_AUTH_TOKEN:-}                   -> //npm.pkg.github.com/:_authToken=
${NODE_AUTH_TOKEN:-${NODE_AUTH_TOKEN}} -> //npm.pkg.github.com/:_authToken=${NODE_AUTH_TOKEN}
```

Only the third keeps the old behaviour where npm resolves the token itself at
install time. An empty token is not optional, npm just gets a 401. Either
mention the third form, or say plainly that the leave-it-to-npm pattern is gone
and the variable has to be exported at build time.

## Not covered

No real cluster, `--dry-run` only. macOS only. I did not test the LSP or
`jac serve` paths, though `get_config()` is unguarded in a lot of call sites so
1 may show up in more places than the profile one.

None of this changes the verdict for me. The fix does what #8773 asked for and I
am happy with it from our side.

jac/jaclang/project/impl/config.impl.jac
```diff
@@ -1317,21 +1317,34 @@ impl _parse_toml_data(
...
     if isinstance(`obj, str) {
         try {
             return interpolate_env_vars(`obj);
-        } except ValueError {
-            return `obj;
+        } except ValueError as e {
+            if not path {
+                raise;
+            }
+            raise ValueError(f"{path}: {e}") from e;
```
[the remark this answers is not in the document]
akindu-k: Resolved as of 5d9c261. Both halves of this comment were valid.

The discovery calls in `_discover_config_from_file` and `_pin_run_scope` now route `ValueError` to the same fatal handler as dispatch, keeping `OSError`/`KeyError` as warnings so a permissions or lookup failure still degrades gracefully. The second half of your comment, about active-profile overlays on this path, was a separate hole: `_discover_config_from_file` re-applied the profile *after* its try block, so a target whose base file was fine but whose `jac.<profile>.toml` was not still ended in a traceback. Grepping the pattern found three unguarded overlay applications in `execution.impl.jac` against one guarded call in dispatch, so `exit_on_config_error` and `apply_overlay_or_exit` now live in `jaclang/cli/config_errors.jac` and every CLI discovery and overlay site uses them. The `project/` sites keep raising, which is correct for a library.

Validated on the pushed commit, running a file in another project whose config cannot load:

| case | exit | tracebacks | target ran |
|---|---|---|---|
| base file unresolvable, relative path | 2 | 0 | no |
| base file unresolvable, absolute path | 2 | 0 | no |
| base file unresolvable, `--no-takeover` | 2 | 0 | no |
| base fine, profile overlay unresolvable | 2 | 0 | no |

All four print one line, `jac.toml is invalid: scale.kubernetes.namespace: Environment variable ... is required: ...`, identical to the main dispatch path. Previously these were exit 1 with a traceback and up to three warning lines.

Happy paths unchanged: overlay resolved, no overlay present, and the host project's own file all exit 0 and run.

Covered by three cases in `jac/tests/cli/test_invalid_config_exit.jac`, including a cross-project target with an unresolvable overlay. Suites: `tests/cli/` 351 (348 on main), `tests/project/` 324, `tests/client/test_cli.jac` 28.

MalithaPrabhashana (commented):
Checked the new commits. The fixes work. One problem at the end.

**First, I was wrong about the npm token.** I only tested it with the variable
unset. With it set, `${NODE_AUTH_TOKEN:-${NODE_AUTH_TOKEN}}` gives `abc123}`, a
broken token. You were right to delete it instead of documenting it.

**The three fixes are good.** Ran the same cases on main and on your branch:

- base file with an unset variable: main runs it, your branch exits 2
- profile overlay, all three activation routes: same
- cross-project `jac run ../other/src/main.jac`: same
- malformed files still say "is malformed", not "is invalid"
- `${VAR:-default}`, `--help`, `--version` and the happy paths are unchanged

No tracebacks anywhere except `jac check` on another project, which exits 1 and
reads fine, like you said. `jac scale deploy --dry-run` is exit 2 with the
variables unset, and exit 0 with the real values when they are set. The docs for
3 and 4 match what I measured.

**Agreed on 2.** Follow-up is the right call, for the reason you gave.

**The problem: the last commit lost a test.** When you folded
`test_invalid_config_exit.jac` into `test_run_project_scope.jac`, the test that
covered the dispatch path did not come across. I reverted each guard one at a
time and re-ran `tests/cli/ tests/project/`:

| reverted | what happens | tests |
|---|---|---|
| `_pin_run_scope` | cross-project stops failing | test FAILS |
| `_discover_config_from_file` | nothing changes | 664 passed, 9 failed |
| dispatch `_apply_profile` | the old traceback comes back | 664 passed, 9 failed |

So the fix for my first comment works, but nothing holds it in place. The test
you deleted already did this, so putting that one back is enough.

Side note: `_discover_config_from_file`'s guard never fires in anything I tried,
because `_pin_run_scope` runs first and raises on the same config. Not asking you
to remove it, just checking whether I missed a path.

**Suites:** main 662 passed / 9 failed, your branch 664 / 9. Same failures, none
of them yours. Merged current main in locally: no conflicts, 692 passed / 9 failed
with everything above still correct.

One more thing: `--profile` has to go before the filename. `jac run file.jac
--profile prod` passes the flag to the program instead. Same on main, so not your
PR, but it confused me for a minute.

Good to merge once that test is back.

jac/jaclang/project/impl/config.impl.jac
```diff
@@ -1343,21 +1343,34 @@ impl _parse_toml_data(
...
```
MusabMahmoodh, line 1355: This also runs inside every scale pod: the deploy ships jac.toml with only `[scale.secrets]` and `[dev]` stripped, and the pod's `jac scale gateway` / `jac run --serve` loads it through `_discover_project_config` with just the Secret keys and the vars jac-scale injects in its env. A toml written like jacBuilder's (`image_registry = "${ECR_REGISTRY}"`, `service_account_name = "${K8S_SERVICE_ACCOUNT_NAME}"`, a `${SHARED_DATA_PVC}` volume name) resolves on the deploy host and passes `--dry-run`, then every pod exits 2 at boot, where on main those were unused literals. The deploy needs to check the shipped toml against the env the pod will actually have and fail on the host, naming the setting. That check belongs in jac-scale and can be its own PR, but it has to land before this one.

jac/jaclang/cli/impl/dispatch.impl.jac
```diff
@@ -13,29 +12,11 @@ def _discover_project_config {
         if config is not None {
             add_venv_to_path(config);
         }
-    } except MalformedJacTomlError as exc {
-        _exit_on_malformed_jac_toml(exc);
     } except ValueError as exc {
-        _exit_on_invalid_jac_toml(exc);
+        exit_on_config_error(exc);
```
MusabMahmoodh, line 16: `jac scale deploy`, `status` and `destroy` load the project `.env` right before `get_scale_config()`, which is how `[scale.secrets] OPENAI_API_KEY = "${OPENAI_API_KEY}"` with the key kept in `.env` gets its value. This discovery runs before the handler, so that project now exits 2 on the secret before `.env` is ever read. If `.env` is meant to feed jac.toml, the load has to happen ahead of this, and if not, the release note should say so.

jac/jaclang/project/impl/config.impl.jac
```diff
@@ -1343,21 +1343,34 @@ impl _parse_toml_data(
...
 }
 
-impl _interpolate_recursive(`obj: any) -> any {
+impl _interpolate_recursive(`obj: any, path: str = "") -> any {
     if isinstance(`obj, str) {
         try {
             return interpolate_env_vars(`obj);
-        } except ValueError {
-            return `obj;
+        } except ValueError as e {
```
MusabMahmoodh, line 1351: The `"${" not in expanded` guard in `scale/admin/service_urls.jac:17` was a downstream patch for the literal this `except` used to leak, and it can't see one now. Its test, "an unresolved ${VAR} app_name is rejected and falls back to localhost" in `scale/tests/misc/test_service_urls.jac`, still passes, but only because `get_kubernetes_config()` now raises into that function's `except Exception`. So it tests the swallow, not the guard. Worth deleting both here.

release_notes/unreleased/jaclang/9045.bugfix.md
```diff
@@ -0,0 +1 @@
+- **Fix: Unset `jac.toml` environment variables now fail loud**: `${VAR}` and `${VAR:?message}` values with no environment variable set now reject the configuration with an error naming the setting that needs it, instead of silently passing through as literal text. Every command that reads `jac.toml` enforces this, so export the variable or switch to `${VAR:-default}` for values that only some commands need.
```
MusabMahmoodh: `jac lsp` is one of these commands too, and unlike `check` its exit 2 lands in an editor output panel. For a project whose secrets live in `.env` or only in CI, the visible symptom is a language server that never starts. Is that intended, or should `lsp` warn and carry on?

MusabMahmoodh (changes requested):
The core change is still right, and pinning each guard at its call site instead of at the helper was the right call on the tests. Two jac-scale paths read jac.toml under a different environment than the one dispatch now checks against, pods at boot and `.env`-driven scale commands, and this turns both into hard failures, so I'd like those handled before merge.

jac/jaclang/project/impl/config.impl.jac
akindu-k: Confirmed, and it is worse than "passes for the wrong reason": that test passed either way.

Probed the branch directly:

```
get_scale_config(project_dir=<app_name = "${JAC_SVCURL_NONEXISTENT}">).get_kubernetes_config()
RAISED: ValueError | scale.kubernetes.app_name: Environment variable JAC_SVCURL_NONEXISTENT is not set
```

So `raw` never reaches `expandvars`, the `"${" not in expanded` guard is unreachable, and the assertion was being satisfied by `except Exception -> app_name = ""`.

Guard deleted in 141ab4b5. I kept the test rather than deleting it, because with the guard gone it pins the right thing for the first time: the only thing that now produces `http://localhost:6000` is the config load failing. Reverting `_interpolate_recursive` to the old swallow makes `app_name` the literal and turns it red:

```
FAILED ...::an unset ${VAR} app_name fails the config load and falls back to localhost
4 passed, 1 failed
```

Restored, 5/5. Renamed to say what it pins, and the comment explaining the guard went with the guard.

One thing I left alone: `os.path.expandvars(raw)` on the line above. It is redundant for `${VAR}` now, but it still resolves the bare `$VAR` form, which `jac.toml` interpolation does not handle. Say the word and I will drop that too.

release_notes/unreleased/jaclang/9045.bugfix.md
akindu-k: Intended, and not new: `jac lsp` already refused to start on any `jac.toml` the loader rejects. This PR widens what counts as rejected; it does not change what happens once one is.

Same project on this branch, three broken configs:

| jac.toml | jac lsp |
|---|---|
| malformed table header | exit 2, `jac.toml is malformed: Expected ']' ...` |
| `entry-point = "src/main"` | exit 2, `jac.toml is invalid: Invalid entry-point ...` |
| `${OPENAI_API_KEY}` unset | exit 2, `jac.toml is invalid: scale.secrets.OPENAI_API_KEY: ...` |

The first two behave identically on main, through the same `_discover_project_config` handler. So an editor whose language server dies on an unloadable `jac.toml` is the existing contract, and I would rather not have `lsp` alone carry on with a config the rest of the toolchain refuses: the analysis it serves would then disagree with `jac check` in the same window.

What was missing is that the symptom is not findable. The release note now names it:

> Every command that reads `jac.toml` enforces this, including `jac lsp`, which will not start on a configuration it cannot resolve, and the `jac scale` commands, which read a project `.env` only after the configuration has been read.

If you want `lsp` to degrade instead, I think it should cover malformed and invalid together rather than interpolation alone, and it needs a decision about what the server answers with while the config is unusable. Happy to take that as a follow-up.

jac/jaclang/cli/impl/dispatch.impl.jac
akindu-k: Confirmed. A project whose only unset variable lives in `.env`:

```
$ cat .env
OPENAI_API_KEY=sk-from-dotenv
$ jac scale status main.jac
✖ Error: jac.toml is invalid: scale.secrets.OPENAI_API_KEY: Environment variable OPENAI_API_KEY is not set
```

Dispatch discovery runs before the handler, so `load_project_env` never gets to run. And `.env` did feed this before: the handler's `get_scale_config()` builds its own config through `PluginConfigBase.get_jac_config`, a fresh `JacConfig.load` rather than the cached ambient one, so it re-interpolated after `load_dotenv` had populated `os.environ`.

I went with the release note rather than moving the load, for two reasons.

`.env` is not the documented source for these values. The secrets reference deploys by exporting, and the capability table says the secret is "shipped as the app Secret; no `.env` file needed". `.env` reaching interpolation was a side effect of that second load.

And moving it is not local. Dispatch discovers the config for every command and knows nothing about scale, and `load_project_env` sits behind `require_optional("dotenv", "env")`, so the core CLI cannot call it unconditionally. Making `.env` a real interpolation source means doing it in config discovery for every command, with its own precedence questions. That is a feature, not something to fold into this fix.

So the note says it, and the secrets reference says it at the point of use:

> The variable has to be in the environment of the command itself. A project `.env` is read by the deploy after `jac.toml` has already been loaded, so it does not supply these values; export them, or give them a `${VAR:-default}`.

If you would rather `.env` became a first-class interpolation source, I will take it as its own PR next to the deploy check in your other comment, since both are the same question about which environment the deploy path resolves against.

jac/jaclang/project/impl/config.impl.jac
akindu-k: Confirmed, and the shape is exactly as you describe.

`sanitized_jac_toml` ships the file's text with `[scale.secrets]` and `[dev]` removed; every other `${...}` reaches the pod verbatim. The pod runs `jac scale gateway` or `jac run --serve "$SV_FILE"`, both of which hit `_discover_project_config` before anything else. Reproduced by running a shipped toml the way a pod would:

```
$ cat jac.toml
[scale.kubernetes]
app_name = "shop"
image_registry = "${ECR_REGISTRY}"
service_account_name = "${K8S_SERVICE_ACCOUNT_NAME}"

$ ECR_REGISTRY=... K8S_SERVICE_ACCOUNT_NAME=... jac run src/main.jac   # deploy host
app booted

$ jac run src/main.jac                                                 # pod env
✖ Error: jac.toml is invalid: scale.kubernetes.image_registry: Environment variable ECR_REGISTRY is not set
exit 2
```

Host resolves it, `--dry-run` is clean, every pod exits 2 at boot on a setting it never reads. On main those were literals nothing looked at.

I looked for a way to close it inside this PR and I do not think there is an honest one. The obvious move, shipping the resolved toml instead of the raw text, is a secret leak: only `[scale.secrets]` is stripped, so `[scale.database] url = "${JAC_DB_URL}"` would be baked into the `.jab` with the password in it. That value is also one that is *supposed* to resolve in the pod, from the Secret the deploy creates. The shipped toml legitimately carries both kinds of reference, pod-resolved and host-only, and only a check that knows the pod's env can tell them apart. Which is your proposal.

Agreed on the sequencing: the check lands in jac-scale first, this one after. What I think it has to do, so we agree before I write it:

- build the pod's env key set from what the deploy actually injects: the `[scale.secrets]` keys, per-app `[apps.<name>.scale.env]`, the fleet and peer-URL keys, and the platform-injected ones
- interpolate the sanitized toml, and every nested one, against that set
- fail on the host naming the setting and the variable, in the same `<setting>: Environment variable X is not set` shape this PR produces
- run it inside `--dry-run`, since the point is to catch it before anything ships

The risk worth naming up front is the other direction: a key set that is too narrow fails a deploy that would have booted fine. So it should be pinned against the manifest builder's real env list rather than a hand-maintained one.

Want me to start that PR now?

MusabMahmoodh: Yes, start it. That scope is right, and pinning the key set to the manifest builder's real env list is what stops it failing deploys that would have booted; running inside `--dry-run` is where it has to live. The `.env` release note and the secrets-reference line answer my other thread, and `lsp` refusing exactly what `jac check` refuses is the consistent contract, so both are settled. I'll hold this one until the check lands and re-review both the same day.

jac/jaclang/project/impl/config.impl.jac
MusabMahmoodh: Confirmed in 141ab4b5, and the renamed test now pins the failure that actually happens. Drop the `expandvars` as well: the loader already resolves `${VAR}`, bare `$VAR` isn't a jac.toml form, so it's a second interpolation rule that only this one call site applies.

akindu-k: The branch currently disagrees with my earlier reply on this thread, so here is why, with the evidence.

**The guard is back**, restored in 969197a97. My premise that a raw `${` could no longer reach this call site was wrong. `${VAR:+alt}` and `${VAR:0:3}` never match the interpolation pattern, so they are neither resolved nor reported, and the literal reaches `service_base_url` even when every variable is set.

Same `jac.toml` with `app_name = "${SOMEAPP:+prod}"`, `SOMEAPP=shop` exported, calling `service_base_url("SVCURL_UNSET", "worker", 6000)`:

| build | result |
| --- | --- |
| `main` | `http://localhost:6000` |
| this branch | `http://localhost:6000` |
| this branch, guard removed | `http://${SOMEAPP:+prod}-worker:6000` |
| this branch, guard kept, `expandvars` removed | `http://localhost:6000` |

The test pins it. Its `${JAC_SVCURL_SET:+alt}` row fails with the guard removed:

```
FAILED ...::an app_name still carrying ${...} never reaches the URL
E   AssertionError: ${JAC_SVCURL_SET:+alt} must not reach the URL
4 passed, 1 failed
```

**`expandvars` is dropped** in 12b21108f, as you asked. The last row of the table is why that is safe: it resolves nothing the guard depends on. The loader already resolves `${VAR}`, `os.path.expandvars` leaves `${SOMEAPP:+prod}` alone because that is not a name, and bare `$VAR` is not a `jac.toml` form. Nothing in the repo writes an `app_name` that way. So the guard stays and the second interpolation rule goes. `test_service_urls.jac` is 5/5, same count.

The underlying gap, operators that look supported but are silently ignored, is tracked in #9184.

MusabMahmoodh (commented):
The description still opens with `Fixes #8773.`, so merging this closes the issue even though `${VAR:+alt}` still reproduces it. The `Refs` edit did not stick.

Once #9199 merges this will conflict in `interpolate_env_vars`, since both add the `isidentifier` branch. Keep the `environ` parameter from #9199 when resolving it. I'll approve once that rebase is up.

akindu-k:
image
image
image

MusabMahmoodh (approved):
Both things I was holding for are in: #9199 is merged and this sits on it with its `environ` parameter intact, and the body says Refs with the rest tracked in #9184. The pack-smoke red is the site readiness race (`landing #top never appeared` on a `[vite] connecting...` page), which fails the same way on main, not this change. Approving.

jac/jaclang/project/impl/config.impl.jac
```diff
@@ -21,16 +21,86 @@ impl interpolate_env_vars(value: str, environ: dict[str, str] | None = None) ->
...
+                refs.append(
+                    UnresolvedRef(
+                        path=at,
+                        var=e.var,
+                        required=e.required,
+                        message=str(e),
+                        raw=item
+                    )
+                );
+                return item;
```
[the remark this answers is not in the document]
akindu-k: Confirmed and fixed in 3038f2fbd. When a value does not resolve, each `${...}` in it is now checked on its own, so every unset reference is recorded. `"${HOST}/${TOKEN:?set TOKEN}"` with both unset now records `HOST` (warning) and `TOKEN` (required), and `jac run` exits 2. Pinned by a new row in the config-load table in `tests/project/test_config.jac` and one in the run policy table in `tests/cli/test_run_project_scope.jac`. Both failed before the fix.

jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac
```diff
@@ -763,6 +765,35 @@ obj ManifestBuilder {
...
+            "${VAR:-default}."
+        );
+    }
+
+    def _deploys_literal(ref: UnresolvedRef) -> bool {
+        (section, key) = ref.deploy_setting();
+        if section == "secrets" {
+            return self.secrets.get(key) == ref.raw;
+        }
+        return bool(section) and getattr(self.k8s_config, key, None) == ref.raw;
```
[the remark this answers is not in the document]
akindu-k: Confirmed and fixed in 3038f2fbd. The deploy check now follows the dotted key through attributes, dicts and list indices (`a.b[0].c`) with `_setting_value`, for both `[scale.secrets]` and `[scale.kubernetes]`. With `interceptor_service_address = "${K8S_NAMESPACE}.svc"` and `K8S_NAMESPACE` unset on the host, `jac scale deploy --dry-run` now exits 1 naming `scale.kubernetes.http_activation.interceptor_service_address`, and exits 0 once it is exported. Pinned by a new row in the deploy-setting table in `test_pod_env.jac`, which failed before the fix. It uses `K8S_NAMESPACE` on purpose, since pods are given that variable and #9199's pod check cannot see the gap.

jac/jaclang/project/impl/config.impl.jac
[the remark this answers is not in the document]
akindu-k: No longer applies. Since 0cfe278df the config load no longer raises on an unset variable: it records the reference and the command that uses the value decides. The discovery guards this referred to were removed. A file target in another project is covered by the run policy table in `tests/cli/test_run_project_scope.jac`, including a `:?` in that project's active profile, which exits 2.

jac/jaclang/cli/docs/reference/plugins/jac-client.md
```diff
@@ -1376,7 +1376,9 @@ proxy = "http://proxy.company.com:8080"
 _authToken = "${NODE_AUTH_TOKEN}"
 ```
 
-The `${NODE_AUTH_TOKEN}` syntax is resolved via the existing jac.toml environment variable interpolation. If the variable is not set at config load time, it passes through as a literal `${NODE_AUTH_TOKEN}` in the generated `.npmrc`, which npm and bun also resolve natively.
+The `${NODE_AUTH_TOKEN}` syntax is resolved via the existing jac.toml environment variable interpolation, so the variable must be set when the config loads. If it is not set, Jac rejects the configuration instead of writing an unresolved value to the generated `.npmrc`.
+
+To keep the old behaviour where npm resolves the token itself at install time, write `${NODE_AUTH_TOKEN:-${NODE_AUTH_TOKEN}}`, which falls back to the literal `${NODE_AUTH_TOKEN}` in the generated `.npmrc`. An empty default such as `${NODE_AUTH_TOKEN:-}` is not a way to make the entry optional: it writes an empty token and npm answers 401.
```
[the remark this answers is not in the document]
akindu-k: No longer applies. Since 0cfe278df an unset `${NODE_AUTH_TOKEN}` keeps its literal text again, so npm resolves it at install time, and the `jac-client` doc is back to the text on `main` with no nested fallback.

jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac
```diff
@@ -763,6 +783,36 @@ obj ManifestBuilder {
...
+            "${VAR:-default}."
+        );
+    }
+
+    def _deploys_literal(ref: UnresolvedRef) -> bool {
+        (section, key) = ref.deploy_setting();
+        if not section {
+            return False;
+        }
+        source = self.secrets if section == "secrets" else self.k8s_config;
```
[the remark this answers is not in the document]
akindu-k: Confirmed and fixed in 19e07b11b. This was a regression from 3038f2fbd, and it reached further than secrets: a quoted label key such as `"app.kubernetes.io/part-of"` under `[scale.kubernetes.labels]` was split the same way. Rather than special-case the secrets dict, each unresolved reference now keeps the key segments it was found under, and the deploy check walks those segments instead of splitting the dotted path. That keeps flat keys flat in both sections. With `"SERVICE.API.KEY" = "${DOTTED_TOKEN}"`, `jac scale deploy --dry-run` exits 1 naming `scale.secrets.SERVICE.API.KEY` when the variable is unset, and exits 0 with `SERVICE.API.KEY: tok` in the Secret when it is exported. Pinned by two new rows in the deploy-setting table in `test_pod_env.jac`, one for the secret and one for the label, and each fails on its own without the fix.

akindu-k:
## Manual verification at `ca4a17f`

Ran the real CLI (dev mode, compiling from this branch's source) against four throwaway projects to confirm each command applies the policy this PR claims, and in particular that tooling never needs deploy-time variables exported.

### Tooling stays silent

Project with the strict `:?` form unset everywhere — `[byllm] model = "${REQUIRED_LLM_MODEL:?set REQUIRED_LLM_MODEL to a model name}"` and `[scale.secrets] api_key = "${DEPLOY_ONLY_SECRET:?...}"`, nothing exported:

| Command | Result |
|---|---|
| `jac check .` | `2 passed`, exit 0, no mention of env vars |
| `jac fmt .` | `Formatted 2/2 '.jac' files`, exit 0, silent |
| `jac lsp` | full stdio session: initialize → didOpen → diagnostics published → shutdown, exit 0 |
| `jac install --plan` | plan printed, exit 0, silent |
| `jac build` | cleared config load without an env error |

The `jac lsp` run was a real LSP handshake over stdio, not a smoke check: the server returned its capability set, type-checked `main.jac` in 1.5s, published diagnostics and shut down cleanly with the required variable still unset.

### `jac run`

- `${MISSING_LLM_MODEL}` in `[byllm]` → `⚠ jac.toml reads environment variables that are not set, so these settings keep their ${...} text` naming `byllm.model`; the program still runs, exit 0.
- `${REQUIRED_LLM_MODEL:?...}` → `✖ Error: jac.toml requires environment variables that are not set` plus the `${VAR:-default}` hint, exit 2. Same for `jac run --serve`.
- Exporting the variable → clean run, exit 0.
- Unset references in `[scale.secrets]`, `[scale.kubernetes]` and `[scripts]` → ignored by `jac run`, as documented.

### Section and profile scoping

- Inactive `[environments.prod.byllm] model = "${PROD_ONLY_MODEL:?prod needs a model}"` → `jac run` silent, exit 0.
- `jac run --profile prod main.jac` → errors, and reports the setting with the profile prefix stripped: `byllm.model: Environment variable PROD_ONLY_MODEL is required: prod needs a model`.

### Interpolation itself is unchanged

Through `jac x`: `${MY_VAL}` → `resolved=from-env`; `${NO_SUCH_VAL:-fallback}` → `defaulted=fallback`; `${1}` reached the shell untouched (`positional=world`); `${#ARR[@]}` reached the shell verbatim.

### `jac scale deploy` is strict

`jac scale deploy --dry-run --show-yaml` with both values unset:

```
[ERROR] Error: --dry-run failed: jac.toml reads environment variables this machine
does not have, so the deploy would carry these settings as literal text:
  scale.secrets.api_key: Environment variable MISSING_DEPLOY_SECRET is not set
  scale.kubernetes.namespace: Environment variable MISSING_K8S_NAMESPACE is not set
```

Moving `MISSING_DEPLOY_SECRET` into the project `.env` cleared that entry and left only the namespace, confirming the documented "the deploy reads the project `.env` first" behavior. Switching both to `${VAR:-default}` gave a clean dry-run — exit 0, `Namespace: probe-ns`, `check: no errors or warnings` — so there are no false positives.

Worth noting for reviewers: exporting `MISSING_K8S_NAMESPACE` locally satisfies this PR's check but then trips the older pod-side `unresolved_toml_references` guard, because the sanitized `jac.toml` shipped to the cluster still carries the `${...}`. That is the existing pod-env guard working as intended, not a regression from this change.

### Suites touched by this PR

```
jac/tests/project/test_config.jac                  130 passed
jac/tests/cli/test_run_project_scope.jac            10 passed
jac/jaclang/scale/tests/deploy/test_pod_env.jac     42 passed
jac/jaclang/scale/tests/misc/test_service_urls.jac   5 passed
```

### Two unrelated failures seen along the way

- `jac build` → `Static musl runtime not found ... run 'zig build vendor-musl'`. Local toolchain only; config loading had already succeeded.
- `jac test .` in a scratch project → `ModuleNotFoundError: No module named '_jac_test_pkg_0.test_main'`. Reproduced identically in a project with no env references at all, so it is a `.test.jac` discovery quirk in that setup and not related to this change.

kugesan1105:
Closing in favour of #9350, merged as `bb6e66182`, which #9350's description already named as superseding this branch.

The diagnosis in this PR is what the merged work is built on: interpolation answered "did this resolve?" at load time and then discarded the answer, which left only "fail for everyone" or "fail for nobody". #9350 keeps that framing and the recording idea, and moves the policy to the command that makes a setting durable, so the deploy refuses to ship a literal while every other command stays quiet. The prerequisite pod-environment check you split out in #9199 landed first and is unchanged.

Also carried over from the review here: Musab's two objections both got answers. Pods reading the shipped toml under a different environment is #9199, and the project `.env` being read after config load is handled by resolving it during the load.

Follow-ups from this thread, so nothing is lost:

- #9184, rewritten against merged `main`. Three of its seven gaps are closed by #9199 and #9350. The one that now leads is that `${VAR:+alt}` and `${VAR:0:3}` produce no ref, so they slip past the new deploy check and still ship as literal text.
- #9351, the two walkers answering one question, with #9353 open against it.

Thanks for the digging on this one, particularly the pod-boot and `.env` paths, which is what made the consumer-side split the obvious shape.

PR #9049, opened 2026-09-07
test(byllm): graph fixtures become pure programs (#9002 PR B)

MalithaPrabhashana:
Part of #9002. **Stacked on #9014** (PR A), which must land first: this branch is based on it and shares the annex it built.

## What this does

PR A moved every def-only fixture into its test. This one takes the graph fixtures, the ones whose `node` / `walker` / `edge` shape genuinely is the subject. They stay files, but stop being tests: no model, no entry block, no prints, no asserts. The test assigns the model, builds its graph and asserts.

| | after PR A | after PR B |
|---|---|---|
| fixture programs | 26 | 20 |
| fixture lines | 863 | 514 |
| fixtures constructing a model | 20 | 12 |
| fixtures containing `print(` or `assert` | 12 | **4** |
| stdout scraping | 35 | **15** |
| `run_fixture` / `load_fixture` calls | 7 | 4 |
| test blocks | 246 | 246 |
| asserts in test files | 844 | 847 |

15 files, +465 / -579, 5 commits. Suite **203 passed, 1 skipped**, unchanged throughout.

**No test in the suite now reads its result out of a fixture's stdout.**

## The routing cluster

Five `routing_*.jac` fixtures shared 47% of their lines verbatim. Each declared its own graph, its own `MockLLM` subclass, and its own copy of a parser that reads the `Candidates` block back out of the routing prompt. They become one `fixtures/routing_graph.jac` holding only shape, and `test_routing.jac` holding the seven subjects.

The parser moves to the annex, with a model on top of it:

```jac
routing_candidates(params)   # -> [(handle, line)], read back out of the prompt
RoutingLLM(pick=...)         # answers with the handles `pick` chooses
```

Scripting handles by hand would be brittle, because they are **generated from the candidate set**: a lone `Agent` stays `Agent`, two siblings become `Agent_alpha` and `Agent_beta`. Reading them also preserves the property these tests exist for, that the prompt carries enough for a chooser to decide.

`RoutingLLM` falls through to the scripted replies when a prompt has no candidates, because a graph whose nodes also make their own `by llm()` calls sends both kinds through one model.

## Mutation testing

Same treatment as #9014. Four mutations to `visit_routing.jac`, each applied alone, suite run, reverted.

| # | mutation | result |
|---|---|---|
| R1 | routing-only params (`select`, `intent`, `incl_info`) leak into the call | **MISSED, then fixed, then caught** |
| R2 | dedup by value instead of identity | equivalent mutant, see below |
| R3 | routing bypasses `Jac.call_llm` and calls `model.invoke` | caught |
| R4 | candidate handles all collapse to one | caught (4 errors) |

**R1 found a regression I introduced.** The test meant to catch it asserted the routing keys were absent from the outgoing params dict. `make_model_params` assembles that dict from a fixed set of keys, so a routing key could never appear there whatever the filter did: the assertion was **vacuously true**. The original fixture asserted on `mt_run.call_params` and was right to. The test now captures the MTRuntime through the same `Jac.call_llm` seam the dispatch test uses. Re-run against the same mutation: caught.

**R2 is an equivalent mutant, not a coverage gap.** Jac nodes never compare by value: `Plain() == Plain()` is `False` even with no fields. So identity-dedup and value-dedup behave identically and the mutation changes nothing observable.

## Three things that failed first, and are now relied on

**A walker triggered on `Root` routes over every other test's leftovers.** `root` accumulates across tests in a file, so the supervisor was choosing among the previous tests' hubs. It triggers on a `Team` node the test builds itself.

**A picker matching the whole prompt matches the handles.** Handles are derived from node type names, so `"poem" in prompt` is true for the *math* request too, because `PoemAgent` is on the candidate list. The picker reads the request out of the prompt instead.

**A Jac generator cannot `return` a value.** `RoutingLLM`'s streaming override yields from super for the non-routing fall-through rather than returning it.

## Docstrings

Comment-only, own commit. Measured first: 121 docstring lines and 37 comment lines across 1,168, a 14% share, so a narrow pass. Five lose a narrative clause and keep their contract.

Untouched on purpose: docstrings on `by llm()` defs and on anything passed in `tools=[]`, since byLLM sends those to the model as prompt and tool description; also every `sem` and every test name.

Verified: triple-quote count identical per file, so nothing was added, split or merged; docstrings stripped and the remainder compared gives **zero code changes, zero comment changes**.

## What is left as a fixture

20 programs, and every one earns it:

- **2 graph programs** (`routing_graph`, `agent_graph`): `node` / `walker` / `edge` shape, no model, no entry block
- **~14 compile targets**: compiled or analyzed by `test_mtir_integration.jac`, never run
- **`system_prompt_override/`**: byLLM resolves `[byllm] system_prompt` from the module's own location, so its defs sit beside that `jac.toml`
- **`builtin_llm_no_override.jac`**: exists precisely to be a module with no `glob llm`
- **`async_by_llm.jac`**: compiled by `JacProgram` to check codegen, never run

The 4 fixtures still containing a `print(` or `assert` are compile targets whose content is the thing being compiled.

## Next

PR C in #9002: `test_usage`, `test_parallel`, `test_compaction`, `test_telemetry_` and `test_mtir_integration` onto the annex, rename `support_tests.jac` to `support.jac` to match the repo convention, and land `TESTING.md`.


christianwilkins (commented):
Reviewed `781adb777225bae708ef9daf0229a81cfc2da57f`, including the full patch against main, the shared #9014 changes, all 15 additional graph-test file changes, the routing implementation, and existing discussion. No additional blocking finding in the graph refactor. The `063552a90` and `629944fd0` follow-ups make the assertions observe routing parameters at the runtime boundary, force `select=1` to truncate a two-choice response, and distinguish the explicit incl_info entry from the walker's repr.

Independent local validation, with CI-matching byLLM dependencies installed in a temporary project site: `test_routing.jac` plus `test_tools.jac` passes all 11 tests. Removing only `chosen = chosen[:1]` from the real routing implementation makes the arbitrary-node-list test fail: it visits `['alpha', 'beta']` where one visit is required. The original source was restored and the worktree is clean. The initial run without those dependencies skipped both files and is not counted as validation. `git diff --check` passes.

Current-head CI is green and the PR is conflict-free. I checked the logs: the package lane records 203 passed, 1 skipped; the separate sealed MTIR lane records 29 passed. No fixes were made during this review.

Status: the graph increment has no new author fix requested, but this remains stacked on the still-open #9014. Land/reconcile that dependency first, then recheck this PR's final diff and checks. I am leaving a comment review while that dependency is outstanding; no merge was performed.

PR #9059, opened 2026-09-08
fix(dist): stop a stale runtime that will not delete from killing jac (#9053)

### LINKED ISSUE #9053: A stale runtime that cannot be deleted kills the jac process, and the error names the wrong path
## Summary

Clearing a **stale** runtime from the jac cache is fatal. If any `rmdir` inside the stale directory fails, `jac` dies with `Uncaught exception` and exit 1 before running the command at all - even though the new runtime has already been extracted successfully and nothing needs the stale one.

**`main` is red on `jac-check` right now for exactly this**, and it is not the diff that merged.

## Evidence

Two jobs, same stale generation id `f4ca7e3e22fe2d5d-2796bddb91431b70`, different file each time, both dying in the "Warm the runtime cache" step before reading a single source file:

`main` @ `9b02a4b2d9` (#8944), `jac-check` job `101933238973`:

```
jac:   extracting runtime...
jac:   removing stale runtime /home/runner/_work/_temp/.cache/jac/rt/f4ca7e3e22fe2d5d-2796bddb91431b70...
Uncaught exception: rmtree: could not remove '/home/runner/_work/_temp/.cache/jac/rt/f4ca7e3e22fe2d5d-2796bddb91431b70/python/lib/python3.14/site-packages/websocket/tests'
##[error]Process completed with exit code 1.
```

PR #9045 @ `a810161`, `jac-check` job `101923479268`:

```
jac:   removing stale runtime /home/runner/_work/_temp/.cache/jac/rt/f4ca7e3e22fe2d5d-2796bddb91431b70...
Uncaught exception: rmtree: could not remove '/home/runner/_work/_temp/.cache/jac/rt/f4ca7e3e22fe2d5d-2796bddb91431b70/python/lib/python3.14/site-packages/openai/types/conversations/__pycache__'
```

The immediately preceding commit on main (`7e2ddc1f8a`) was green on the same job, so this is intermittent and driven by which cache `actions/cache/restore` hands the job, not by any diff.

## Mechanism

Two separate defects, and the second is why the error message points at the wrong file.

**1. A best-effort cleanup is fatal.** `jac/jaclang/dist/fused/materialize.jac:213-214`:

```jac
say(f"  removing stale runtime {full}...");
shutil.rmtree(full);
```

Unguarded. This loop only ever visits directories it has already decided are *stale* (`name != keep_key`, same path hash). The freshly extracted runtime at `keep_key` is complete before this runs, so failing to delete a directory nobody will read again cannot affect correctness - but it takes the whole process down.

**2. The failure is reported at the wrong level.** `jac/jaclang/runtime/na_stdlib/shutil.jac:96-102`:

```jac
if os.path.isdir(child) {
    rmtree(child);
} else {
    unlink(child);          // i32 return, discarded
}
...
if rmdir(path) != 0 {
    raise ValueError(f"rmtree: could not remove '{path}'");
}
```

`unlink` is declared `-> i32` (`:3`) and its result is dropped. So a file that cannot be unlinked fails silently, the directory is then non-empty, and `rmdir` fails on the **parent**. That is why both messages name a directory (`websocket/tests`, `.../__pycache__`) rather than the file that actually could not be removed. Whoever debugs this from the message alone is looking one level above the real cause.

`shutil.jac:70` drops an `unlink` return the same way inside `move`.

## Why this is worth fixing rather than re-running

- It is not attributable to any diff, so the first reaction is always to re-run, and the second is to go looking through an innocent PR. It cost a cycle on #9045 and on #8175 this week, and it currently has main red.
- Every `jac` invocation on any machine holding a stale generation is exposed, not just CI. Local dev caches accumulate generations the same way.
- The cleanup is the one part of the boot path where failure is provably harmless, and it is the only part that is fatal.

## Suggested shape

1. `materialize.jac:214` - a stale-runtime removal that fails should warn and continue. Leaving a stale directory on disk costs disk, not correctness, and the next run retries it.
2. `shutil.jac:97` - check `unlink`'s return and raise naming the *file*, so when a removal genuinely must succeed the diagnostic points at the cause. Same for `:70`.

I have not reproduced this on a local machine; both witnesses are CI jobs and the mechanism above is read off the source as it stands on `9b02a4b2d9`. What I have not established is *why* the unlink fails on those particular paths (permissions from the cache restore, or a concurrent reader) - but that is the third defect, and the first two make it non-fatal either way.

akindu-k:
Fixes #9053.

## What broke

The first time you run a freshly installed `jac`, it has to unpack its runtime before it can do anything. That first run goes:

1. unpack the new runtime into a folder
2. mark that folder ready to use
3. delete the old folder the new one replaced
4. run whatever command the user typed

Step 3 could fail. Sometimes a file in the old folder refuses to delete. When that happened, `jac` crashed and exited right there, and the command at step 4 never ran.

That is the silly part. Steps 1 and 2 had already finished, so the new runtime was live and usable. The folder that would not delete was garbage nothing would ever read again. Failing to take out the trash cost the user their command.

This is not theoretical: it has taken `main` red on `jac-check` and cost cycles on unrelated PRs.

## Why the error message was useless too

There is a second defect underneath, and it is why nobody could tell what was actually stuck.

Deleting a folder means emptying it first: delete each file inside, then remove the now-empty folder. Deleting a file calls `unlink`, which is a C function, and C functions report failure by *returning* `-1` rather than raising. You have to look at the number.

The old code threw that number away:

```jac
unlink(child);          # did it work? nobody asked
```

So when one file would not delete:

1. `unlink` returns `-1`, nobody checks, the loop moves on
2. the loop ends believing the folder is empty
3. `rmdir` on the folder fails, because it is not empty
4. the error raised says `could not remove '<the folder>'`

The message blamed the parent folder, which was innocent, never named the file actually at fault, and carried no reason at all. That is exactly why the CI reports on this issue name directories and stop there.

## The fix

**1. The cleanup is best effort now** (`dist/fused/materialize.jac`). If a stale generation will not delete, warn and carry on to the next one instead of letting the error escape. Safe because the sweep runs strictly after the new runtime is published, so every folder it touches is already garbage.

**2. `rmtree` checks `unlink` and says what really failed** (`na_stdlib/shutil.jac`). The return code is checked, so the failure is raised the moment it happens and names the child that would not go, not the parent whose `rmdir` failed as a consequence. Both error paths now append `errno_message()`, so the message carries the reason from the OS.

The two halves work together: the first turns a crash into a printed warning, the second makes that warning worth reading.

## Before and after

On a stale generation holding a file that will not unlink:

```
before:  exit 139, no output, the command never ran

after:   exit 0, jac runs normally, and the sweep says

  jac:   removing stale runtime .../rt/0000000000000000-9b481c05d84867ad...
  jac:   could not remove it: rmtree: could not remove
         '.../site-packages/websocket/tests/test_app.py': Permission denied;
         leaving it for the next run
```

That second line also answers the question the issue left open: the unlink fails with `EACCES`.

## Testing

Test count 8 -> 9 across the touched files. `test_prim_equivalence.jac` stays at 8, because the `bundled shutil` test is extended rather than duplicated. The one added test covers `gc_stale`, which had no coverage anywhere and is the crash site.

Both tests need a file that is guaranteed not to delete. On POSIX, permission to delete a file lives on the folder containing it, so both build a folder, put a file in it, and `chmod 555` the folder. Both restore the mode afterwards so nothing undeletable is left behind.

**`test_prim_equivalence.jac::bundled shutil`** gains two rows both lanes must agree on (the removal raises, the folder stays) plus one native-only check asking a yes or no question: does the error text contain `pinned.txt`? Before the `shutil.jac` change the answer is no, because the message named the parent, and the test fails with `rmtree must report the file it could not unlink, not its parent`. The wording check is native-only on purpose, since CPython words its own error differently and comparing the two would fail for a reason that is not a bug.

**`jac/tests/dist/test_materialize.jac`** JIT-compiles a probe over the staged fused runtime and sweeps a crafted cache. The generations are a table, one row per rule the sweep has to honour:

| Generation | Stands for | Survives |
| --- | --- | --- |
| `2222...-aaaa...`, locked | a generation that will not delete | yes, and no crash |
| `3333...-aaaa...` | an ordinary stale generation | no, still deleted |
| `2222...-bbbb...` | another binary's generation (right half differs) | yes |
| `4444...-aaaa....tmp.4242` | an extraction another process is doing now | yes |
| `1111...-aaaa...` | the live generation it was told to keep | yes |

Row two is the one that matters most after the crash case: it stops a fix that simply swallows everything from passing, because the sweep still has to finish its job after tripping over row one. The probe's `sweep` returns `1`, so returning at all is the assertion. Without the `materialize.jac` change the test process dies with SIGABRT, exit 134, which is the reported failure.

It stages the runtime with `_stage_fused_runtime`, the same helper the desktop builder uses, and compiles through `tests.support`'s `compile_native` / `get_func` rather than hand-rolling the `ctypes.CFUNCTYPE` wrapper, which is what tripped E1053 on the previous head. No subprocess: the compiled code runs in-process.

Both lean on a directory mode that withholds the write bit, so both skip under uid 0, which ignores it. Skipping is the honest outcome there: under root the setup never fails and the assertions would pin nothing.

## Not in scope

**Directory symlinks.** `rmtree` still follows the `isdir` predicate into them rather than unlinking them. That divergence is pre-existing, already documented in the na_stdlib README, and closing it needs an `os.path.islink` intrinsic the native pathway does not have yet.

**`move`'s unchecked `unlink`.** It keeps main's behaviour. Checking it there would turn a completed move into a failure over a leftover source file, which is the opposite of the argument this PR makes for `gc_stale`. CPython's `shutil.move` does raise in that spot, so the gap is worth closing, but as its own change with its own equivalence row, not as a side effect of a diagnostic fix.

**Partial reclamation.** A stale generation that hits an undeletable file now keeps the rest of its tree, where the old unchecked loop would have unlinked the remainder before failing at the parent `rmdir`. Getting that space back means teaching the sweep to continue past a failure and report everything it could not take, which is a follow-up on top of the crash fix.


christianwilkins (commented):
Reviewed `eed368f177928c2346b8e82d9a44c7b68a31b9d1`, including the full diff, runtime publication/sweep ordering, and the native shutil/socket errno callers. The best-effort boundary is appropriately confined to stale-generation deletion after the replacement runtime is published. Checking the previously discarded unlink results also gives the failure a useful path. No additional blocking production defect found in those changes.

The rebase needs particular care: main `3ad7d91b5` now has `_errno_native.jac` using `_errno_platform.errno_address`, with `clear_errno` and `raise_os_error` alongside the existing helpers. Please reuse that shared implementation and preserve those APIs, rather than retaining this PR's older `_errno_native.darwin.jac` variant, which would select a module that lacks them on Darwin. GitHub currently reports merge conflicts.

Independent local validation on macOS: `git diff --check` passes, but `jac test jac/tests/dist/test_materialize.jac` stops before running the test. The new file reports E1053 at line 66, `ctypes.CFUNCTYPE(ctypes.c_int64)`: the inferred `restype` type rejects `c_int64`. Thus I have not independently exercised the native stale-cache sweep. This is a local compiler/test compatibility result, not evidence that cleanup itself failed.

Current-head CI run 34207890431 is red in `test-runtime`: `test_dangling_ref_heal` loses its database connection with SQLSTATE 08006. I have not reproduced that failure on the base and am not treating it as an established baseline failure. No fixes were made during review.

Status: not ready to merge. Author work: reconcile the errno changes during rebase and get the new sweep test executable in the validation environment. Then validate the updated head and resolve or establish the provenance of the remaining CI failure.

MusabMahmoodh (changes requested):
Thanks for taking this one - and for going further than the issue asked. `test_materialize.jac` is the right shape: a probe that dies with SIGABRT/134 without the change is a much better witness than asserting on a log line, and putting the `strerror` text behind the failure answers the question #9053 left open (`EACCES`) instead of just relocating the message. The `gc_stale` and `shutil.rmtree` changes are both correct and I want them in.

Requesting changes on one blocking item, which is not your fault - main moved under you the same day you opened this.

## 1. Blocking: the errno extraction already landed on main, and this version breaks `_directory_native`

#9022 (`d8658619f8`, Sep 8) added `jac/jaclang/runtime/na_stdlib/_errno_native.jac` to main a few hours after this PR opened. `git merge-tree upstream/main <head>` on the current tip:

```
CONFLICT (add/add): Merge conflict in jac/jaclang/runtime/na_stdlib/_errno_native.jac
CONFLICT (content): Merge conflict in jac/jaclang/runtime/na_stdlib/shutil.jac
```

The `shutil.jac` one is an ordinary rebase. The `_errno_native.jac` one is not, because the two files are not the same file:

| | this PR | `main` today |
|---|---|---|
| errno address | `__errno_location()` + `memcpy` into a 4-byte buffer | `__mem_load_i32(errno_address())` |
| platform split | a new `_errno_native.darwin.jac` | `_errno_platform.jac` / `_errno_platform.darwin.jac`, already merged |
| exports | `last_errno`, `errno_message` | `last_errno`, `errno_message`, **`clear_errno`**, **`raise_os_error`** |

`jac/jaclang/runtime/na_stdlib/_directory_native.jac:5` on main reads:

```jac
import from _errno_native { clear_errno, last_errno, raise_os_error }
```

So resolving that conflict toward this PR drops two symbols `_directory_native` imports, and `directory_names` is what `rmtree` itself calls. It would take the na_stdlib directory pathway out.

The rest of that section is already done on main too - `_socket_native.jac:13` and `socket.jac:10` both already carry `import from _errno_native { errno_message }`. So the `-15` in `_socket_native` and the `+2/-2` in `socket.jac` are re-doing a merged change.

**Ask:** rebase and drop `_errno_native.jac`, `_errno_native.darwin.jac`, `_socket_native.jac` and `socket.jac` from the diff entirely, keeping only the `import from _errno_native { errno_message }` line you add to `shutil.jac`. That is 4 of 12 files and about 50 lines gone, and what remains is `materialize.jac` + `shutil.jac` + tests, which is the whole of #9053.

## 2. `move()` becomes fatal after the copy has already succeeded

`shutil.jac:68-73` - `move` is `copyfile(src, real_dst)` then `unlink(src)`. Raising when that unlink fails means a move whose **destination is fully written and correct** now reports failure, and the caller loses the `real_dst` return.

Three callers on main, all build-path:

- `jac/jaclang/dist/seal.jac:403`
- `jac/jaclang/client/impl/vite_bundler.impl.jac:294`
- `jac/jaclang/data/storage/impl/local_storage.impl.jac:125`, whose `move` is declared `-> bool`; it now propagates instead of returning `True` for a move that in fact happened

For all three, a leftover source file costs disk and not correctness - which is exactly the argument this PR's own body makes for `gc_stale`. As written the PR removes a fatal best-effort cleanup in one function and adds one in another.

To be precise about what #9053 asked for: the `shutil.jac:70` note ("`move` drops an `unlink` return the same way") was about the *diagnostic* - the dropped return is why the message names the wrong path - not a request to make the call abort. My read is that `move` should keep the `errno_message()` text but not raise on it. If you think a leftover source is worth failing a `seal` over, that is a defensible call, but it should be stated in the body rather than arrive as a side effect of the diagnostic fix.

## 3. Non-blocking: a stale generation now shrinks less than it used to

`rmtree` previously unlinked everything it could and only failed at the parent `rmdir`. It now raises at the first file that will not unlink, so the rest of the tree survives.

Inside `gc_stale` that changes what "leaving it for the next run" means: the next run hits the same locked file, leaves the same tree, and reclaims nothing. Before this change each pass at least removed the deletable remainder. If best-effort reclamation is the intent, the sweep wants to continue past a failure and report what it could not take. Happy for this to be a follow-up - it does not block the crash fix.

## CI

`test-runtime` red is not this diff. `jac/tests/runtimelib/test_dangling_ref_heal.jac::traversal skips a deleted edge row instead of raising` failed on `{'C': '08006', 'M': 'connection lost mid-transaction; retry'}` - the Postgres connection flake, 1927 passed / 1 failed. That run also predates #9022, so it would not have seen the conflict above either. An empty commit is the only retrigger available to us; `gh run rerun` 403s on this repo for non-maintainers.

Once the rebase in (1) is in, ping me and I will re-review the same day.

MusabMahmoodh (approved):
The main merge took `_errno_native.jac` and the socket changes out of the diff and left `move` alone, which covers both blocking items from last round. What's left is `gc_stale`, `rmtree` and the tests, which is the whole of #9053. Approving.

Greptile's `move` P1 describes main's existing behaviour rather than anything this diff introduces, so nothing is needed for it here. The partial-shrink point still stands as a follow-up if you want it: `rmtree` now stops at the first file that will not unlink, so a stuck generation reclaims nothing on later runs instead of shrinking each time.

akindu-k:
Superseded by #9246, which makes the stale-runtime sweep non-fatal (`sweep_runtimes` removes with `ignore_errors=True`) and checks `unlink` in `rmtree` so the error names the child that failed. #9053 is closed. Not carried over: the OS reason (`errno_message()`) in the `rmtree` error and a warning when a generation can't be removed. Those can be a small follow-up if needed.

Issue #8998, opened 2026-09-06
apps.md awaits a bridged call inside a plain `can`; that fails bytecode (E5043), the compiling spelling is `async can`


MalithaPrabhashana:
### Summary

`reference/apps.md` teaches a spelling that cannot run. The page's
"consistency model" example awaits a bridged call inside a **plain** `can`:

```jac
walker:pub post_and_show {
    can run with Root entry {
        posted = await create_tweet(content=self.text);   # runs on social_graph
        feed = await load_feed(limit=10);
        report feed.reports;
    }
}
```

`jac check` accepts it. `jac run` refuses it. The spelling that works is
`async can`, which is what `examples/mini_todo/main.jac` already uses.

Still present on **0.37.7**, and the example is unchanged on `main` today
(`jac/jaclang/cli/docs/reference/apps.md:340-342`).

### Repro

One file, no workspace and no bridge needed. The disagreement is between the
checker and the bytecode compiler, so the smallest case shows it:

`plain.jac`
```jac
async def work() -> str { return "ok"; }

walker demo {
    can run with Root entry {
        r = await work();
        report {"r": r};
    }
}
with entry { print("loaded"); }
```

```
$ jac check plain.jac
1 passed                       # 0 errors

$ jac run plain.jac
✖ Error: error[E5043]: Bytecode compilation failed: 'await' outside async
  function (plain.jac, line 5)
```

Change `can run` to `async can run` and the same file both checks and runs.

### Root cause, as far as I can see it from outside

Two readers of the same source disagree. The checker treats `await` inside a
plain ability as well-formed; the bytecode compiler lowers a plain `can` to a
non-async function and CPython then refuses the `await`. Nothing in between
reports it, so the failure lands at run time on a file that `jac check` has
already called clean.

That makes the docs page the third disagreeing reader, and the one a person
follows first.

### What I would suggest

Two separable things:

1. **The docs**: change the `post_and_show` example to `async can`, so the page
   teaches the form that runs. That is the whole of this issue and is safe on
   its own.
2. **Optional, and a different call**: have the checker reject `await` in a
   plain `can` with the same E5043, so the error arrives at check time rather
   than at run time. If plain-`can` `await` is meant to be legal and the
   bytecode path is the bug, then this is the opposite fix and I would rather
   you decide which.

SupulHeshan:
Fixed in #9064:

[...]

Issue #8517, opened 2026-08-22
[Bug] async can abilities silently never execute — __jac_async__ flag written as dead local, read from wrong object


chess10kp:
### Summary

`async can` abilities on walkers are silently never awaited/executed: spawn takes the sync path, drops the coroutine, and the program exits 0 with lost updates. The only trace is a `RuntimeWarning: coroutine 'W.go' was never awaited`.

### Repro

```jac
import asyncio;
node N { has v: int; }
walker W {
    has total: int = 0;
    async can go with N entry {
        await asyncio.sleep(0.01);
        self.total += here.v;
    }
}
with entry {
    w = W();
    w spawn N(v=1);
    print(w.total);   # prints 0 — expected 1
}
```

### Root cause (located)

`jcir_gen_pass.impl.jac` (~line 1150) inserts `__jac_async__ = True` as a **dead local inside the ability body**, but the runtime (`runtime.impl.jac:264`) reads `__jac_async__` from the **walker instance**, which stays False, so spawn takes the sync path. Proof: manually setting `w.__jac_async__ = True` before spawn makes everything work — async traversal, `visit`, `disengage`, exception propagation, even `asyncio.gather` over two concurrent spawns. The plumbing exists; only the flag wiring is broken.

marsninja:
Closing: no activity in 10+ days. Reopen if still relevant.

MalithaPrabhashana:
Reopening with fresh evidence: this was closed on 2026-09-04 with no linked fix, and
the exact repro in the summary still prints 0 on both released 0.37 lines.

```
$ jac --version
jac 0.37.5  (Darwin arm64)
$ jac run w.jac --no-serve
.../jaclang/runtime/osp_kernel.jac:407: RuntimeWarning: coroutine 'W.go' was never awaited
0
```

Same on 0.37.1 (release asset), same file, same output. Also reproduces with the
ability on `Root entry` and `root spawn w`, which is the shape a served walker has.

Where it bites on the 0.37.5 workspace model: every cross-app call is a
typed-async stub, and the natural place to `await` one is a walker ability. That
ability then has to be `async can` (a plain `can` containing `await` fails
bytecode with E5043), and an `async can` spawned by a plain `spawn` runs
nothing and warns. We hit this migrating jacBuilder: the ability that persists
assistant turns is spawned from a worker thread, and as `async can` it silently
stopped persisting. We had to keep every ability sync and complete the coroutine
[...]

PR #9064, opened 2026-09-08
fix(compiler): reject `await` outside async at check time (E2085)

SupulHeshan:
## What was broken, in plain terms

`await` only works inside a function/ability marked `async`. Jac's own docs (`reference/apps.md`) had an example that used `await` inside a **plain** `can` (not `async can`). `jac check` said this was fine. Running it did not agree:

```
$ jac check plain.jac
1 passed                       # 0 errors

$ jac run plain.jac
✖ Error: error[E5043]: Bytecode compilation failed: 'await' outside async
  function (plain.jac, line 5)
```

So three different parts of the toolchain disagreed with each other about the same code: the docs taught it, the checker accepted it, and only the compiler -- several stages later, at the very last step before running -- actually rejected it.

## Why `jac check` didn't catch it

`jac check` runs parsing and type-checking, but **no pass ever looked at `await` expressions at all**. There was no rule anywhere that said "`await` needs to be inside something `async`". The only thing enforcing that rule was CPython itself: `jac run` lowers a plain `can`/`def` to a normal Python `def` (only `async can`/`async def` becomes `async def`), hands that to Python's real `compile()`, and Python's own `SyntaxError` for `await` outside `async` gets wrapped into `E5043`. `jac check` never reaches that step, so it never saw the problem.

```mermaid
flowchart TD
    subgraph before["Before this PR"]
        direction TB
        A1["source: plain `can` { await ... }"] --> B1["jac check"]
        B1 -->|"no rule ever looks at `await`"| C1["✅ passes clean"]
        A1 --> D1["jac run"]
        D1 --> E1["lowered to a plain Python `def`"]
        E1 --> F1["handed to Python's real compile()"]
        F1 -->|"Python: SyntaxError"| G1["❌ E5043, only now, at the last step"]
    end

    subgraph after["After this PR"]
        direction TB
        A2["source: plain `can` { await ... }"] --> B2["jac check"]
        B2 -->|"new E2085 rule"| C2["❌ E2085, immediately, with a clear message"]

        A3["source: `async can` { await ... }"] --> B3["jac check"]
        B3 -->|"ability is async -- OK"| C3["✅ passes"]
        C3 --> D3["jac run"]
        D3 --> E3["✅ compiles and runs"]
    end
```

## The fix

1. **New diagnostic `E2085`** in `StaticAnalysisPass`, added right next to the existing `E2083` check (`disengage` used outside a walker), which already solves this *exact* shape of problem for a different keyword. `E2085` does the same thing for `await`: for every `await` in the module, walk up its parent chain to the nearest enclosing `Ability` (an ability declared elsewhere and implemented via `impl` is resolved back to its declaration) and flag it unless that ability is `async`. No new infrastructure -- it reuses the pattern that's already proven correct for a sibling case.

2. **Codespace-aware, added after the first CI run caught real false positives** (see below): the "await needs `async`" rule only actually holds for code compiling to Python. It doesn't apply to client-targeted code -- a bridged server call awaited from a plain, non-`async` `def:pub` component is the established idiom for calling server functions from client code (`test_client_codegen.jac`'s own fixtures assert this generates JS) -- or to `test { }` bodies in client-context test files (Jac's grammar has no `async test` spelling at all, and real, executed tests like `fullstack.test.jac` already rely on bare `await` there). `E2085` now checks `code_context == CodeContext.SERVER`, read off the same `Ability`/`Test`/`Archetype` node already used to resolve `is_async` -- reusing the exact `code_context` field this same pass already reads in `_check_client_imports`. A full-repo scan for `await` directly inside a `test { }` body turned up zero genuine server-context counterexamples, so no separate `Test` carve-out was needed beyond the codespace gate.

3. **A nice side effect, not extra code**: `E2085` is an *error*, and the pipeline already skips code generation once any check-time error fires (the same mechanism the existing `E0063` test relies on). So once `E2085` fires, the run never reaches the bytecode step, and the old, later, more confusing `E5043` simply doesn't appear alongside it. One clear error instead of two disagreeing ones.

4. **Docs fix, two copies**: `apps.md`'s `post_and_show` example now uses `async can`, the form that actually compiles and runs -- matching `examples/mini_todo/main.jac`, which already used it correctly. A second, separate copy of the same broken example turned up in `jac-apps.md` (a bundled CLI skill guide, distinct from `docs/reference/apps.md`) and is fixed the same way -- caught by `test_guide.jac`'s "every jac example in a guide checks clean" test, which only started exercising `await` placement once `E2085` existed to check it.

## What this PR does *not* fix

The corrected example now passes both `jac check` and `jac run`. That's the full scope of #8998 and this PR. Whether the awaited call actually *executes* correctly at runtime (i.e., whether the walker's `spawn` really takes the async path once the ability is `async can`) is a separate, deeper bug tracked at #8517, and will be a separate PR. A comment on #8998 links the two so nobody reads the fixed example as proof that #8517 is also resolved.

## Test plan

- Searched the suite first for anything already covering "`await` outside async" before writing a new test -- none exists. The nearest neighbor, `test_checker_gaps.jac` Gap #36, is a different bug in a different pass (`TypeCheckPass`'s type evaluator returns the wrong *result type* for `await`, not whether `await` is legally placed) -- no overlap to extend.
- Added one new test (`test_new_diagnostics.jac`, "Context Validity" section, right after the sibling `E2083` test), because `E2085` is a genuinely new, previously-uncovered code path -- there is nothing to extend.
- One fixture (`context_E2085_await_outside_async.jac`) carries both cases -- a plain `can` awaiting (must reject) next to an `async can` awaiting the same call (must not reject) -- checked with a single `count_diagnostic_code(...) == 1` assertion. Matches `E2083`'s own two-case/one-fixture/one-assertion shape exactly, rather than introducing a new pattern.
- Reused existing helpers/fixtures throughout: `compile_and_check`, `count_diagnostic_code`, `NEW_DIAG_FIXTURES`. No new test infrastructure.
- Test counts: `test_new_diagnostics.jac` 91 -> 92 (+1). `test_static_analysis_pass.jac` and `test_checker_gaps.jac` (the two other files touching this pass) are unchanged, 0 diff.
- Verified the new test fails before the fix (only `E5043` fires, at bytecode-compile time) and passes after.
- Verified the sibling `E2083` test still passes unmodified (no regression in the shared `_check_semantic_patterns` function).
- All in-process (`JacProgram().compile()`), no subprocess/server.
- `jac fmt --lintfix` applied to all changed files.

## CI failures found and fixed

The first push failed 6 checks (`jac-check`, `test-compiler`, `test-runtime`, `test-jac-pack-smoke`, `test-scale`). Root-caused all of them to the two false-positive categories in point 2 above, verified against the exact fixtures CI flagged (not just theory):

- `jac/examples/jaclang_org/core/site/jacyac/fullstack.test.jac` -- client-context `test { }` block awaiting -- was blocking codegen (`test-jac-pack-smoke`, `jac-check`) and, downstream, making the "cl test runner" discover zero client tests (`test-scale`).
- `tests/compiler/fixtures/codespace/infer_three_spaces.jac`, `svc_cut/app.jac`, and `test_client_codegen.jac`'s own inline fixtures -- client `def:pub ... -> JsxElement` awaiting a bridged call -- were blocking JS codegen entirely (`test-compiler`).
- `jac-apps.md`'s duplicate, unfixed copy of the docs example (`test-runtime`'s `test_guide.jac`).

All four now verified directly against those exact files/fixtures, plus the full `test_client_codegen.jac`, `test_codespace_inference.jac`, `test_typed_interop.jac`, and `test_guide.jac` suites re-run green, alongside the original `E2085`/`E2083` tests (no regression).


jac/jaclang/compiler/passes/impl/static_analysis_pass.impl.jac
```diff
@@ -404,6 +404,34 @@ impl StaticAnalysisPass._check_semantic_patterns(mod: Module) -> None {
...
+                    in_async = _dl.is_async;
+                }
+                break;
+            }
+            if isinstance(parent, Archetype) {
+                break;
+            }
+            parent = parent.parent;
+        }
+        if not in_async {
```
christianwilkins: [P1] Preserve implicit async client test bodies

This walk never recognizes `Test`, so an `await` in a client test reaches the module with `in_async == False` and gets rejected. Client tests are deliberately emitted as `async` functions by `EsastGenPass.exit_test` (`esast_gen_pass.impl/module.impl.jac:9`). On this exact head, `jac check jac/tests/runtimelib/test_desktop_transport.jac` now reports E2085 at lines 51 and 64; the same file passes on main, and current CI also fails to compile it and `test_sso_login.jac`. Please recognize valid client test execution contexts while retaining rejection of synchronous server bodies, and add a negative control that compiles an awaiting client test.

christianwilkins (changes requested):
Reviewed `80d80e271190ccfb9def05889a0da17c06b62521`, including the complete diff, current-head CI failures, and the JavaScript test lowering. One blocking regression: E2085 rejects client test bodies that are implicitly async. Details and reproduction are attached inline.

Local validation: `jac test jac/tests/compiler/passes/test_new_diagnostics.jac` passes 97 cases with 5 skips, and `git diff --check` passes. That suite does not catch the missing client-test context. `jac check jac/tests/runtimelib/test_desktop_transport.jac` fails on this head with two E2085 errors at lines 51 and 64. The same file checks successfully on main `3ad7d91b5`; the fixture and pre-PR static-analysis implementation are unchanged between this PR's base and that main snapshot. This is independently consistent with the `test-runtime` failures in CI run 34233097267. The ES backend emits `Test` as an async function in `esast_gen_pass.impl/module.impl.jac`, so these awaits are supported behavior.

The current-head CI run fails in the compiler, runtime, server, checker, and packaging lanes. Some compiler failures also involve client components, so please assess E2085 against client execution contexts when fixing this, rather than changing only the desktop fixture. I have not attributed every CI failure to this one defect.

No code changes or fixes were made during review. The only follow-up commit in the reviewed patch, `80d80e271`, adds the release note. Status: changes requested; preserve valid async client test bodies, add a regression control, and rerun the affected lanes before merge. Maintainer CI approval or reruns alone will not fix this semantic regression.

jac/jaclang/compiler/passes/impl/static_analysis_pass.impl.jac
SupulHeshan: Confirmed and fixed on the current head (eacac3469).

You're right about the walk not recognizing `Test` -- but that's already resolved. A follow-up commit on this branch (2853fc4b3) first attempted a codespace-based fix; the branch was then merged with `main`, which already carries the reviewed, more precise fix from #9073 (`if isinstance(parent, Test) { in_async = parent.in_client_context(); break; }` -- exactly what you're describing).

Verified directly against your exact repro:

    $ jac check jac/tests/runtimelib/test_desktop_transport.jac
    1 passed in 0.29s   # 0 E2085 hits (previously 2, at lines 51 and 64)

    $ jac check jac/tests/runtimelib/test_sso_login.jac
    1 passed in 0.29s   # 0 E2085 hits

Bigger picture: this PR is now superseded by #9073, which already merged into `main` and closed #8998 -- #9064 currently shows a 0-line diff against `main`. I'll be closing it as redundant rather than pushing it further.

One point from your review stands on its own regardless of this PR's fate: there is no dedicated regression test anywhere (not here, not on `main`) that specifically asserts "`await` inside a client-context `test { }` body does not raise E2085" -- current coverage is incidental, via fixtures like `fullstack.test.jac` that happen to exercise it. That's a legitimate gap worth a follow-up test at some point.

Issue #9041, opened 2026-09-07
Boot check: github-scheme webhook walkers need a secret and the system identity


kashmithnisakya:
Part of #9038. Follows #9039.

## Why

A github-scheme walker with an empty `github_secret` would either accept nothing (every delivery 401, silently) or, if someone flips `verify_signature` off, accept everything. Neither should be discoverable only from a provider's delivery log. The system identity is a second silent dependency: if `find_user_by_identity(SYSTEM_USER)` returns nothing, every delivery 401s with "Invalid or expired API key", which is wrong on its face.

## What

In `WebhookExtension.register`, when at least one github-scheme walker exists:

- `[scale.webhook].github_secret` empty: hard error at boot naming the walker and the config key, matching the style of the other `[scale.*]` validation errors.
- system identity missing: same.
- `verify_signature = false` together with a github-scheme walker: refuse, since the scheme has no other auth.

## Acceptance

- Boot test with the github fixture and an empty secret fails with the named error.
- With the secret set, boot is clean and the existing webhook tests pass.
- Apps with only jac-scheme walkers see no new checks.

kashmithnisakya:
Delivered in #9065 (merged Sep 11, in v0.37.13 and later): `github_prerequisites` raises at registration when a github-scheme walker exists and `[scale.webhook].github_secret` is empty or the system identity cannot be resolved; the boot test posts the failure. Same PR-body parsing gap as #9040; ready to close.

kashmithnisakya:
Closing: delivered in #9065 (merged Sep 11, in v0.37.13 and later). Registration raises when a github-scheme walker exists and `[scale.webhook].github_secret` is empty or the system identity is missing, with the boot test posting the failure. Verified on flowline-dev today: the pod boots with the secret set and the signature gate answers 401 to a wrong signature.

Issue #9040, opened 2026-09-07
Pass provider headers (X-GitHub-Event, X-GitHub-Delivery) into declared webhook walker fields


kashmithnisakya:
Part of #9038. Follows #9039.

## Why

A GitHub delivery names its event type and its unique id in headers, not in the body: `X-GitHub-Event: pull_request` and `X-GitHub-Delivery: <uuid>`. The body has `action` but not the event name, so `issues.closed` and `pull_request.closed` look alike at the top level. The delivery id is the only safe dedupe key, since deliveries can arrive twice.

Today `_process` maps body keys onto introspected walker fields and nothing else reaches the walker.

## What

In the github branch of `webhook_ext.jac` `_process`, after body-field mapping:

- if the walker declares `event`, set it from the `X-GitHub-Event` header;
- if the walker declares `delivery`, set it from `X-GitHub-Delivery`;
- the header wins over a same-named body key for these two names (a provider body never carries them);
- a walker that declares neither is unaffected.

## Acceptance

- Fixture walker declaring both fields receives the header values.
- A fixture walker without them behaves exactly as in #9039.
- jac-scheme walkers untouched.
- Without this, a consumer must infer the event from which body key is present (`issue` vs `pull_request`); note that fallback in the guide until this lands.

kashmithnisakya:
Delivered in #9065 (merged Sep 11, shipped in v0.37.13 and later): `webhook_ext.jac` copies `X-GitHub-Event` into a declared `event` field and `X-GitHub-Delivery` into `delivery`, headers only, and `test_webhook.jac` asserts both. The PR body said "Closes #9039, #9040 and #9041", which GitHub only parses for the first number; a maintainer can close this one.

kashmithnisakya:
Closing: delivered in #9065 (merged Sep 11, in v0.37.13 and later). The runtime copies X-GitHub-Event into a declared `event` field and X-GitHub-Delivery into `delivery`, header-only, with a test in `test_webhook.jac`. Verified live on flowline-dev today: a real issue event reached the receiver with both fields set.

Issue #9038, opened 2026-09-07
Webhook transport: accept provider-signed deliveries (GitHub scheme) without an API key


kashmithnisakya:
## Why

`@restspec(protocol=APIProtocol.WEBHOOK)` serves a walker at `POST /webhook/<Name>` behind one auth shape: a mandatory `X-API-Key` header (a per-user key JWT), plus `X-Webhook-Signature` = HMAC-SHA256 over `"<timestamp>." + body` keyed by that key's signing secret, plus `X-Webhook-Timestamp` for the replay window.

Third-party senders cannot produce that shape. GitHub sends `X-Hub-Signature-256: sha256=<HMAC-SHA256(secret, raw body)>`, no API key, no timestamp, and custom headers cannot be added to a delivery. Stripe, Slack and most providers work the same way: each signs the raw body with a shared secret under its own header name.

Two workarounds exist and both are bad:

- A relay in front of the runtime that verifies the provider's signature and re-signs in the jac scheme. Extra service, extra secret, and it still needs a per-provider map.
- Verifying inside the walker. Impossible: the walker receives parsed top-level fields, and re-serialised JSON never reproduces the signed bytes.

## Proposal

A per-walker signature scheme on the existing webhook transport, starting with `github`:

```jac
@restspec(protocol=APIProtocol.WEBHOOK, scheme="github")
walker GithubEvent {
    has event: str = "",
        delivery: str = "",
        action: str = "",
        installation: dict = {},
        issue: dict = {};

    can handle with Root entry { ... }
}
```

| | `scheme=""` (today, unchanged) | `scheme="github"` (new) |
| --- | --- | --- |
| Auth | `X-API-Key` resolves a key doc | none; the signature is the auth |
| Signature | HMAC(`"<ts>." + body`, key secret) | HMAC(raw body, `[scale.webhook].github_secret`) in `X-Hub-Signature-256` |
| Replay window | `X-Webhook-Timestamp` | none (GitHub sends no timestamp); the app dedupes by `X-GitHub-Delivery` |
| Runs as | the key owner | `SYSTEM_USER`, the identity the scheduler already runs jobs as |
| Rate limit key | api key id | walker name |
| Provider headers | none | `X-GitHub-Event` and `X-GitHub-Delivery` land in declared walker fields |

Verified against 0.37.7:

- `server/server.jac`: `RestSpecs` has `method`, `path`, `protocol`, `broadcast`, `produces`, `envelope`; no scheme.
[...]

PR #9065, opened 2026-09-08
feat(scale): webhook walkers accept provider-signed deliveries with scheme="github"

### LINKED ISSUE #9039: Per-walker `scheme` on the webhook transport with a `github` HMAC-over-body branch
Part of #9038.

## What

Add a `scheme` to `@restspec` for webhook walkers and a `github` branch in the webhook endpoint that verifies `X-Hub-Signature-256` over the raw body with a configured secret, without an API key.

## Where

| File | Change |
| --- | --- |
| `jac/jaclang/server/server.jac` | `RestSpecs` gains `scheme: str = ""`. `SpecsCls(**specs)` already accepts it. |
| `jac/jaclang/runtime/impl/builtin.impl.jac` | `restspec()` validates `scheme` in `{"", "jac", "github"}`, same error style as the `protocol` and `method` checks. |
| `jac/jaclang/scale/config/plugin_config.jac` and `config/impl/config_loader.impl.jac` | Under `webhook`: `github_secret` (default empty) and `github_signature_header` (default `X-Hub-Signature-256`), surfaced by `get_webhook_config`. |
| `jac/jaclang/scale/server/webhook_ext.jac` `_process` | Branch on the walker's spec. github: size check first; header required; `extract_signature` strips `sha256=`; `WebhookUtils.verify_signature(raw_body, sig, github_secret)`; no `X-API-Key`; no timestamp; user id from `find_user_by_identity(SYSTEM_USER)` as `jobs_ext.jac` does; `check_rate_limit` keyed by walker name. |
| `jac/jaclang/scale/server/webhook_ext.jac` `register` | Skip `self.api_keys()` when every webhook walker is github-scheme, so apps that never mint keys do not get the "keys stored in memory" boot warning. |
| `jac/jaclang/scale/tests/fixtures/test_api.jac` | Fixture walker with `scheme="github"` declaring `event`, `delivery`, `action`, `installation`. |
| `jac/jaclang/scale/tests/server/test_webhook.jac` | Cases below. `_generate_webhook_signature` works as is, without the timestamp prefix. |

`WebhookUtils.verify_signature` and `extract_signature` in `scale/webhook/webhook.jac` need no change.

## Acceptance

- github-scheme walker: valid signature returns 200 and the walker ran; wrong secret 401; missing header 401; `sha256=`-prefixed and bare hex both accepted; no `X-API-Key` required; body over `max_body_bytes` returns 413 before any signature work; empty `github_secret` refuses every delivery.
- jac-scheme walkers: every existing case in `test_webhook.jac` unchanged and green.
- Rate limit for the github scheme counts per walker name.
- The `webhook audit` log line keeps its shape.
- `scheme="stripe"` (unknown) raises at decoration time with the list of accepted values.

## Out of scope

Provider headers into fields (#?), boot validation (#?), OpenAPI (#?), docs (#?).

kashmithnisakya:
Closes #9039, #9040 and #9041. Part of #9038.

## Why

`@restspec(protocol=APIProtocol.WEBHOOK)` serves a walker at `/webhook/<Name>` behind one auth shape: a mandatory `X-API-Key` plus an HMAC over `"<timestamp>." + body` keyed by that key's signing secret. Providers such as GitHub cannot produce it: they sign the raw body with a shared secret under their own header, send no API key and no timestamp, and custom headers cannot be added to a delivery. Verifying inside the walker is not an option either, since the walker only sees parsed fields and re-serialised JSON never reproduces the signed bytes.

## What

A per-walker signature scheme on the existing transport, starting with `github`. The default scheme is unchanged and both kinds of walker coexist in one app.

```jac
@restspec(protocol=APIProtocol.WEBHOOK, scheme="github")
walker GithubEvent {
    has event: str = "", delivery: str = "", action: str = "", installation: dict[str, any] = {};
    can handle with Root entry { ... }
}
```

| | `scheme=""` (default, unchanged) | `scheme="github"` (new) |
| --- | --- | --- |
| Auth | `X-API-Key` resolves a key doc | none; the signature is the auth |
| Signature | HMAC(`"<ts>." + body`, key secret) | HMAC(raw body, `[scale.webhook].github_secret`) in `X-Hub-Signature-256`, `sha256=` prefix optional |
| Replay window | `X-Webhook-Timestamp` | none (GitHub sends none, and only the body is signed); deduping on `X-GitHub-Delivery` absorbs GitHub's own redeliveries, not a captured body replayed under a new id |
| Content type | not checked | `application/json` only; GitHub's form-encoded default is refused with 415 instead of running the walker with empty fields |
| Runs as | the key owner | `SYSTEM_USER`, the identity the scheduler runs jobs as |
| Rate limit key | api key id | walker name |
| Provider headers | none | `X-GitHub-Event` and `X-GitHub-Delivery` are the only source of declared `event` and `delivery` fields; a same-named body key never reaches the walker |

### Changes

- `server/server.jac`: `RestSpecs.scheme: str = ""`.
- `runtime/impl/builtin.impl.jac`: `restspec()` accepts `scheme` in `"", "jac", "github"` (`"jac"` normalises to `""`) and rejects anything else at decoration; a scheme on a non-webhook protocol is rejected the same way.
- `scale/config`: `github_secret` (default empty) and `github_signature_header` (default `X-Hub-Signature-256`) under `[scale.webhook]`, in the schema, the defaults and `get_webhook_config`.
- `scale/server/webhook_ext.jac`: the endpoint forks on the walker's scheme. `process_api_key` is the previous `_process` body unchanged; `process_github` does size check, signature, per-walker rate window, an `application/json` content-type check (415 otherwise, since a form-encoded delivery verifies but parses to nothing), header pass-through, then the shared `run_walker` (body-field mapping, spawn, error handling, also unchanged). OpenAPI parameters follow the scheme. `register` raises at boot when a github-scheme walker exists and either the secret is empty or the system identity is missing, and resolves that identity once there, so deliveries run as it with no per-request lookup and no fallback. It only builds the api-key manager when an api-key walker exists, so github-only apps stop getting the "keys stored in memory" warning. #9041 also asked to refuse `verify_signature = false` next to a github walker; the github path never consults that flag, so there is nothing to refuse and no check was added.
- `scale/webhook/webhook.jac`: the one-minute fixed window behind `ApiKeyManager.check_rate_limit` is now the free `rate_window_ok(state, key, limit)`, and both the api-key path and the github path call it.
- Fixture: `GithubEvent` and `KeyedEvent` walkers in `github_webhook.jac`; a `[scale.webhook]` block with the secret in the fixture `jac.toml`. The github tests boot from a scratch copy of the fixture dir and add the 64 KiB body cap there, so it never reaches the servers other tests start from the shared dir.
- Docs: `jac-sv-deploy` skill, the `jac-scale-http` reference (options table and a GitHub-signed section) and the config reference.
- Release note fragment `release_notes/unreleased/jaclang/9038.feature.md` (rename to this PR's number if that is the convention you prefer).

## Verification

`jac test jaclang/scale/tests/server/test_webhook.jac` from source: the 15 existing api-key cases unchanged and green, plus 17 new ones:

- signed delivery accepted without an api key, with `event`, `delivery`, `action` and `installation` all mapped
- bare hex signature accepted; wrong secret 401; missing header 401
- api-key headers on a github walker 401 (the schemes do not cross)
- oversize body 413 with a deliberately wrong signature (size check runs first)
- provider headers win over same-named body keys, and with no provider headers at all the same body keys land both fields empty
- not reachable via `/walker/GithubEvent`
- OpenAPI lists `x_hub_signature_256` and `x_github_event` for the github walker and still lists `x_api_key` for the api-key walkers
- a configured `github_signature_header` is the one required and declared: the custom header alone is accepted, the default alone is refused
- `scheme="stripe"` and a scheme on an HTTP walker both raise at decoration
- boot fails with a message naming `github_secret` when the fixture's secret is removed
- a signed form-encoded delivery is refused with 415 and a message naming `application/json`
- the media type is matched exactly: `application/json; charset=utf-8` is accepted, `application/json-patch+json` is refused
- with `rate_limit_per_minute = 1`, three signed deliveries answer 200 then 429

`jac test jaclang/scale/tests/misc/test_config_loader.jac`, `scripts/validate_docs_code.jac` and the docs tests pass. `jac fmt --check --lintfix` is clean on every touched `.jac` file under the repo's `[check.lint]` rules.

## First consumer

kashmithnisakya/flowline#166: a GitHub App webhook posting to `/webhook/GithubEvent`, with the app resolving the tenant root from `installation.id` after this verification.


jac/jaclang/scale/server/webhook_ext.jac
```diff
@@ -34,32 +51,76 @@ obj WebhookExtension(ServerExtension) {
...
-                required=False,
-                description='Unix timestamp (seconds) bound into the signature (X-Webhook-Timestamp header)',
-                kind=ParameterType.HEADER
-            )
-        ];
+        parameters: list[APIParameter] = [];
+        if walker_scheme(server, walker_name) == GITHUB_SCHEME {
+            parameters = [
+                APIParameter(
+                    name='x_hub_signature_256',
```
christianwilkins: [P2] Use the configured signature header in the required endpoint parameter. `github_gate` honors `github_signature_header`, but this declaration always requires the default header before the callback runs. Reproduced with the fixture server: default header → 200; configure `X-Custom-GitHub-Signature` and send a valid signature only there → 422 with `missing header x_hub_signature_256`; send both headers → 200. Derive registration/OpenAPI from the same config and test a custom-header-only request.

christianwilkins (changes requested):
Reviewed head `404f381ccd3278990d4d438bd72403a0588edf70`: the complete scheme/config/docs/test patch, shared HMAC verifier, request-parameter binding, system identity path, API-key compatibility path, existing discussion, and current-head CI.

P2: the advertised `github_signature_header` override cannot be used by itself. `github_gate` reads the configured name, but endpoint registration still declares `x_hub_signature_256` required, so the framework rejects the request before verification. Please derive the required header/OpenAPI declaration from the same configuration and add a custom-header integration case.

I reproduced this through the real in-process fixture server: the default signed request returns 200; after setting the configured header to `X-Custom-GitHub-Signature`, the same correctly signed request with only that header returns 422 (`missing header x_hub_signature_256`); sending both headers returns 200. These were direct script probes, not a claim that the full webhook suite passed locally. The initial test-runner attempt was stopped during prolonged setup. No source changes made.

Current-head CI remains red only in the pack-smoke lane and its aggregate. I checked the log: the landing `#top` element was not found, matching the author's existing handoff; I have not independently established flakiness and am not repeating the rerun request in another thread. GitHub reports no merge conflicts.

Changes requested for the header-override defect. Author next step: fix registration and add the integration test, then run validation on the new head. Maintainer action remains the CI rerun/approval already discussed, followed by review once the author fix is in.

jac/jaclang/scale/server/webhook_ext.jac
kashmithnisakya: Fixed in 470acc356: one helper resolves the header name from [scale.webhook].github_signature_header and both the required OpenAPI parameter and the gate use it, with a custom-header integration case (X-Custom-GitHub-Signature only → 200, default header only → refused, OpenAPI declares the configured name).

jac/jaclang/scale/server/webhook_ext.jac
```diff
@@ -34,32 +51,76 @@ obj WebhookExtension(ServerExtension) {
...
+                    data_type='string',
+                    required=True,
+                    description='sha256= plus HMAC-SHA256 of the raw body, keyed by [scale.webhook].github_secret (X-Hub-Signature-256 header)',
+                    kind=ParameterType.HEADER
+                ),
```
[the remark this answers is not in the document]
kashmithnisakya: Fixed in 470acc356: the required signature parameter is now derived from the configured header name, the same value the gate verifies, and a test covers a custom header used on its own.

kashmithnisakya:
@christianwilkins Header-override defect fixed in 470acc356: one config-derived header name feeds both the required OpenAPI parameter and the gate, with a custom-header integration case; the webhook suite is 28/28 from source on the new head, and the CI paragraph needs nothing further from my side beyond this push.

Greptile summary: the header inconsistency and the missing custom-header test are both addressed in 470acc356.

MusabMahmoodh (commented):
Read this against #9038 and the three sub-issues. The shape is right: the fork is on the walker's own spec, `process_api_key` is the old body moved intact, `run_walker` is genuinely shared, and the two schemes coexist in one app with a real end-to-end test each. Four things below.

jac/jaclang/scale/server/webhook_ext.jac
```diff
@@ -81,111 +153,183 @@ obj WebhookExtension(ServerExtension) {
...
+        walker_kwargs: dict[str, any] = {};
+        if isinstance(body_data, dict) {
+            for (field_name, value) in body_data.items() {
+                if field_name in fields and field_name != '_jac_spawn_node' {
+                    walker_kwargs[field_name] = value;
+                }
+            }
+        }
+        for (field_name, value) in header_fields.items() {
+            if field_name in fields and value {
```
MusabMahmoodh: The `and value` guard means the header only wins when it is present: a delivery with no `X-GitHub-Event` but `{"event": "push"}` in the body lands `event="push"` on the walker, which is the spoof #9040 set out to close. The signature proves the body came from the secret holder, so it is not an auth hole, but the invariant the docs state ("headers win") does not hold, and the test only covers the case where both are present. For a github walker `event`/`delivery` should be header-only:

```suggestion
            if field_name in fields {
```

Empty header then clears the body key instead of losing to it.

jac/jaclang/scale/server/webhook_ext.jac
```diff
@@ -34,32 +61,77 @@ obj WebhookExtension(ServerExtension) {
         return self.manager;
     }
 
+    def scheme_rate_ok(rate_key: str, limit_per_minute: int) -> bool {
```
MusabMahmoodh: This is `ApiKeyManager.check_rate_limit` copied verbatim - same `timestamp() // 60` window, same `[window_id, count]` slot, same three branches (webhook.impl.jac:259). Second source of truth for the rate window, and the two will drift the first time one of them grows a store-backed counter. I get why you did not just call `keys().check_rate_limit(...)` - it would build the ApiKeyManager and open the docs store for a github-only app, which is the boot warning you are removing. So pull the window itself out into a free `def rate_window_ok(state: dict, key: str, limit: int) -> bool` in `scale/webhook/webhook.jac` and have both callers use it.

jac/jaclang/scale/server/webhook_ext.jac
```diff
@@ -81,111 +153,183 @@ obj WebhookExtension(ServerExtension) {
...
+            raw_body, WebhookUtils.extract_signature(signature), secret
+        ) {
+            return _fail('UNAUTHORIZED', 'Invalid webhook signature', 401);
+        }
+        return None;
+    }
+
+    def system_user_id(server: JacAPIServer) -> str {
+        found = server.user_manager.find_user_by_identity(SYSTEM_USER);
+        return str(found.get('user_id', SYSTEM_USER)) if found else SYSTEM_USER;
```
MusabMahmoodh: #9041 asked for two boot checks and this lands one. The system-identity half is still the failure the issue described: if `find_user_by_identity(SYSTEM_USER)` returns nothing, this hands back the literal `'__system__'` and `spawn_walker` resolves a root for that name, so every delivery quietly runs against the wrong graph instead of failing loudly. I know it mirrors `jobs_ext.jac:67-70`, but the scheduler has no external caller expecting a 200. Either add the check to `require_github_secret` alongside the secret, or reopen the identity half of #9041 - the PR says "Closes #9041".

No objection to dropping the `verify_signature = false` refusal, since the github path never consults that flag and there is nothing to refuse - worth saying so in the PR body, since the issue asked for it.

jac/jaclang/scale/tests/fixtures/jac.toml
```diff
@@ -7,3 +7,9 @@ description = "Test fixture for jac-scale serve tests"
 X-Custom-Test-Header = "test-value"
 Cross-Origin-Opener-Policy = "same-origin"
 Cross-Origin-Embedder-Policy = "require-corp"
+
+# GitHub-scheme webhook walkers (GithubEvent in test_api.jac) verify against
+# this secret; the body cap is small so the oversize test stays cheap.
+[scale.webhook]
```
MusabMahmoodh, line 13: `GithubEvent` is in `github_webhook.jac`, not `test_api.jac` - the PR body says test_api too. Also this file is the shared fixture dir that `test_serve.jac` and `test_seed_determinism.jac` boot from; only the new tests bind `get_scale_config(project_dir=...)` so nothing reads it today, but a 64 KiB webhook cap sitting in the shared fixture is a trap for whoever next points a test's project_dir here.

jac/jaclang/scale/server/webhook_ext.jac
kashmithnisakya: fixed in e802051ba: the guard is gone, so on a github walker `event`/`delivery` come from the headers only and an absent header lands an empty string, with a test for the header-absent case.

jac/jaclang/scale/server/webhook_ext.jac
kashmithnisakya: fixed in e802051ba: the window is now a free `rate_window_ok(state, key, limit)` in scale/webhook/webhook.jac and both `check_rate_limit` and the github path call it.

jac/jaclang/scale/server/webhook_ext.jac
kashmithnisakya: fixed in e802051ba: `register` resolves the system identity next to the secret check and raises when it is missing, the per-request lookup and the `__system__` fallback are gone, and the PR body now says why the `verify_signature` refusal was dropped.

jac/jaclang/scale/tests/fixtures/jac.toml
kashmithnisakya: fixed in e802051ba: the comment and PR body name github_webhook.jac now, and the 64 KiB cap moved into the scratch copy `_github_client` boots from, so the shared fixture only carries the secret.

jac/jaclang/scale/server/webhook_ext.jac
```diff
@@ -81,111 +140,180 @@ obj WebhookExtension(ServerExtension) {
...
+        refused = self.github_gate(request, raw_body, cfg);
+        if refused is not None {
+            return refused;
+        }
+        if not rate_window_ok(
+            self._scheme_rate, f"{GITHUB_SCHEME}:{walker_name}", rate_limit
+        ) {
+            return _fail('RATE_LIMITED', 'Rate limit exceeded', 429);
+        }
+        header_fields = {
```
MusabMahmoodh, line 237: A repo or org webhook left on GitHub's default content type (`form`) sends `payload=<urlencoded json>`: the signature verifies, `json.loads` in `run_walker` raises, the `except ValueError` turns it into `{}`, and the walker runs with every body field at its default and we answer 200, so GitHub marks the delivery done and the event is gone. The github path should refuse anything that is not JSON loudly, plus a case in test_webhook.jac:
```suggestion
        content_type = str(request.headers.get('Content-Type') or '').lower();
        if not content_type.startswith('application/json') {
            return _fail(
                'UNSUPPORTED_MEDIA_TYPE',
                'Set the webhook content type to application/json',
                415
            );
        }
        header_fields = {
```

jac/jaclang/scale/tests/server/test_webhook.jac
```diff
@@ -581,3 +582,403 @@ test "webhook - api key rejected as user session token" {
...
+    if username == SYSTEM_USER {
+        return {};
+    }
+    return _ensure_internal_user(users, username, password, role);
+}
+
+test "github scheme - boot fails without the system identity" {
+    booted = False;
+    try {
+        with unittest.mock.patch.object(
```
MusabMahmoodh: This patches `UserManager.ensure_internal_user` with `unittest.mock`, which scale/tests/TESTING.md 2.2 rules out, and it fakes the exact collaborator whose behaviour the boot check is guarding against. `delete_user` refuses the system account and every boot path runs `_ensure_internal_users` before `register`, so there is no real-store way to reach this; I'd drop the test along with `import unittest.mock` and the `_ensure_internal_user`/`_skip_system_user` helpers - the raise in `github_prerequisites` reads plainly on its own.

jac/jaclang/cli/docs/reference/plugins/jac-scale-http.md
```diff
@@ -1501,6 +1503,31 @@ curl -X POST "http://localhost:8000/webhook/PaymentReceived" \
...
+        installation: dict[str, any] = {};
+
+    can handle with Root entry {
+        report {"event": self.event, "action": self.action};
+    }
+}
+```
+
+- No `X-API-Key`. The runtime verifies `X-Hub-Signature-256` (`sha256=` plus HMAC-SHA256 of the raw body, keyed by `[scale.webhook].github_secret`; the prefix is optional).
+- No timestamp window: GitHub sends none. Dedupe on `X-GitHub-Delivery` inside the walker.
```
MusabMahmoodh: GitHub only signs the body, so `X-GitHub-Delivery` and `X-GitHub-Event` are unauthenticated: anyone holding one captured delivery can resend it with a fresh delivery id and walk straight past a delivery-id dedupe. Worth saying that here so nobody treats the dedupe as the replay defence:
```suggestion
- No timestamp window: GitHub sends none, and only the body is signed, so `X-GitHub-Event` and `X-GitHub-Delivery` are not authenticated. Deduping on `X-GitHub-Delivery` absorbs GitHub's own redeliveries, not a captured body replayed with a new delivery id.
```

jac/jaclang/scale/server/webhook_ext.jac
```diff
@@ -81,111 +140,180 @@ obj WebhookExtension(ServerExtension) {
...
+        rate_limit = int(cfg.get('rate_limit_per_minute', 0));
+        raw_body = request.body;
+        if len(raw_body) > max_body_bytes {
+            return _fail('PAYLOAD_TOO_LARGE', 'Request body too large', 413);
+        }
+        refused = self.github_gate(request, raw_body, cfg);
+        if refused is not None {
+            return refused;
+        }
+        if not rate_window_ok(
```
MusabMahmoodh, line 220: Nothing in test_webhook.jac reaches a 429 on either scheme, so the extracted `rate_window_ok` has no coverage from its two callers, and #9039 lists per-walker counting as acceptance. `_github_client` already appends to the scratch `[scale.webhook]`, so a `rate_limit_per_minute = 1` case that sends three signed deliveries and asserts at least one 429 would cover it without depending on where the minute boundary falls.

MusabMahmoodh (changes requested):
All four from last round are fixed, and the new missing-header test does fail on the old guard. Two things still need a change before this goes in: a form-encoded delivery gets a silent 200 with an empty walker, and the system-identity test patches `UserManager` with `unittest.mock`. The red CI is main's `scoring.jac` entry-point break (fixed on main by #9110), so a merge from main should clear it.

jac/jaclang/scale/server/webhook_ext.jac
kashmithnisakya: fixed in 9dabad042: the github path now refuses anything but `application/json` with 415 after the gate, and a test posts a signed form-encoded `payload=` body and expects the 415.

jac/jaclang/scale/tests/server/test_webhook.jac
kashmithnisakya: fixed in 9dabad042: the mocked test, the `unittest.mock` import and both helpers are gone; the raise in `github_prerequisites` stays as is.

jac/jaclang/cli/docs/reference/plugins/jac-scale-http.md
kashmithnisakya: fixed in 9dabad042: took the wording as suggested, and the skill and the PR body table now say the same thing.

jac/jaclang/scale/server/webhook_ext.jac
kashmithnisakya: fixed in 9dabad042: `_github_client` takes `rate_limit_per_minute`, and a test sends three signed deliveries at a limit of 1 and asserts a 200 then at least one 429.

jac/jaclang/scale/server/webhook_ext.jac
```diff
@@ -81,111 +140,188 @@ obj WebhookExtension(ServerExtension) {
...
+            self._scheme_rate, f"{GITHUB_SCHEME}:{walker_name}", rate_limit
+        ) {
+            return _fail('RATE_LIMITED', 'Rate limit exceeded', 429);
+        }
+        content_type = str(request.headers.get('Content-Type') or '').lower();
+        if not content_type.startswith('application/json') {
```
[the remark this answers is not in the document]
kashmithnisakya: fixed in 677a7becb: the check compares the media type token itself, so `application/json; charset=utf-8` passes and `application/json-patch+json` gets the 415, with a test for both.

MusabMahmoodh (approved):
All four are in, and the tests pin each one where it can fail: the signed `payload=` delivery gets the 415, `application/json-patch+json` is refused while a charset parameter passes, the limit-1 case reaches a 429 without depending on the minute boundary, and the mocked boot test is gone. Matching the media type token instead of the prefix was the right call on greptile's point too.

PR #9068, opened 2026-09-08
feat(compiler): add typed graph construction expressions

### LINKED ISSUE #7956: RFC: graph construction from first principles: path-valued connects, construction literals, pattern binding
## Summary

RFC for a first-principles redesign of Jac's graph *construction* surface: make connect expressions evaluate to a first-class **path** value, add **construction literals** for tree-shaped structure, and (long-term) let the same pattern shape with binding serve as the read-side dual (graph pattern matching). This is a deliberate clean-break proposal; a small non-breaking operator variant is documented at the end as the incremental alternative.

Motivated by two field reports from writing large graph-constructing programs in Jac: the tree-sitter grammar authored as an OSP graph (`jaseci-labs/tree-sitter-jac/grammar.jac`) and the jacblocks block editor after its model moved to real node/edge archetypes (marsninja/jacblocks, dogfood log items #1 and #21).

## Current semantics, stated precisely

Every connect expression evaluates to **the node at the arrowhead**, and all forms are left-associative:

- `a ++> b` evaluates to `b` (the target)
- `a <++ b` swaps operands at lowering (`pyast_gen_pass.impl.jac`, ConnectOp: `edge_dir == IN` swaps left/right, and `connect` returns `right`), so it evaluates to `a`, which is again the arrowhead
- typed forms `a +>:E(x=1):+> b` behave identically

This is a deliberate and useful choice: it is what makes **top-down descend** flow when building trees. The tree-sitter grammar depends on it pervasively:

```jac
# grammar.jac: create a node, keep building into the fresh child
more  = outer ++> Rep();         # ++> returns Rep(): descend
inner = more  ++> Seq();         # returns Seq(): descend again
inner ++> [Lit(value=","), r];   # fan out ordered members
```

## The problem: a fold can only have one survivor

A connect chain constructs many things (N nodes, N-1 edges), but an expression evaluates to exactly one value. Folding construction through left/right-associative binary operators forces a choice of survivor, and **no choice is right across use cases**:

1. **Sibling chains lose their head.** A Scratch-style stack over `NextOf` edges wants the HEAD in hand for registration, but the chain evaluates to the TAIL:

```jac
# jacblocks: cannot build-and-register a stack in one expression
add_script(prog, flag_hat() +>:NextOf:+> set_var(...) +>:NextOf:+> say(...));
#                ^ passes the TAIL (say), not the hat
```

jacblocks worked around it by making `add_script` return its head argument, so registration chains read top-to-bottom. That works, but it is per-API plumbing rather than a language property, and it does not help the inner case (loop bodies still name their head):

```jac
body = change_var("count", lit(1));                     # head must be named
body +>:NextOf:+> say_of(join2(lit("count: "), varb("count")));
add_script(prog, flag_hat())                            # API returns head
    +>:NextOf:+> set_var("count", lit(0))
    +>:NextOf:+> repeat_n(lit(10), body)
    +>:NextOf:+> say_text("done!");
```

2. **The edge you just made is unreachable.** Connect returns a node, never the edge, so code that needs the edge (to set payload later, or to delete it) must immediately re-query the graph for the thing it just created:

```jac
# jacblocks wire(): re-plug a slot by re-querying for the edge just to delete it
def wire(parent: Block, name: str, child: Block) -> Block {
    for e in [edge parent ->:InputOf:slot == name:->] {   # re-query own write
        del e;
    }
    parent +>:InputOf(slot=name):+> child;
    return parent;
}
```

3. **Note that "return the source instead" does not fix chains.** A left-associative source-evaluating connect gives `((a->b)->c)` with `a` carried through, which builds a STAR (a->b, a->c), not a chain. Head-in-hand chains need source-eval AND right-associativity. This asymmetry is a hint that the binary-op formulation is carrying more than it can express.

## Principle 1: a construction expression evaluates to what it constructed

Make the connect-chain expression evaluate to a lightweight first-class `path` value:

```jac
p = a +>:E(w=1):+> b +>:E(w=2):+> c;
p.head    # a
p.tail    # c
p.nodes   # [a, b, c]
p.edges   # [E(w=1) instance, E(w=2) instance] in order
```

All three problems above dissolve at once:

```jac
# 1. stacks register by head, built inline, no API contract needed
add_script(prog, (flag_hat()
    +>:NextOf:+> set_var("count", lit(0))
    +>:NextOf:+> repeat_n(lit(10), body)
    +>:NextOf:+> say_text("done!")).head);

# 2. the edge is first-class the moment it exists
plug = (parent +>:InputOf(slot="X"):+> child).edges[0];
...
del plug;                       # no re-query

# 3. no head-vs-tail-vs-star question exists; you name the part you want
```

**Cost control:** statement-position connects (the overwhelming majority) lower exactly as today, straight to the central `osp_conn`, materializing nothing. Only value-position connects build the path wrapper. The compiler already distinguishes the positions.

**Descend regression and its answer:** grammar.jac's `outer ++> Rep()` descend becomes `(outer ++> Rep()).tail`, which is worse. That is intentional pressure toward Principle 2, where the tree-building use case gets a strictly better home.

## Principle 2: structure literals for structure

Imperative descend is a workaround for the absence of graph literals. Jac already commits to tree literals for one domain: JSX. Generalize the idea to node archetypes: a node expression may take a construction block whose entries become ordered children (connection order is already a guaranteed semantic, #6785), with typed edges annotatable inline:

```jac
# tree-sitter grammar.jac, comma_sep1 today: six lines, three named temps
def comma_sep1(r: GNode) -> GNode {
    outer = Seq();
    outer ++> r;
    more  = outer ++> Rep();
    inner = more ++> Seq();
    inner ++> [Lit(value=","), r];
    return outer;
}

# as a construction literal: the shape IS the code
def comma_sep1(r: GNode) -> GNode {
    return Seq() {
        r,
        Rep() { Seq() { Lit(value=","), r } }
    };
}
```

```jac
# jacblocks: a repeat block with slot inputs, today three statements + wire calls
b = make_block("control_repeat", {});
wire(b, "TIMES", times);
wire(b, "SUBSTACK", body_first);

# as a literal with typed-edge annotations
Block(opcode="control_repeat") {
    :InputOf(slot="TIMES"):    times,
    :InputOf(slot="SUBSTACK"): body_first
}
```

A literal evaluates to its root node (there is no "which end" question because nothing is a fold). The `++>` operator family remains for genuinely incremental mutation, which is what an operator is for: an editor plugging one block into an existing stack, a walker growing a graph mid-traversal.

## Principle 3: binding inside the pattern (Cypher's lesson)

Any element of a chain or literal can be named where it appears; Jac's walrus already provides the mechanism:

```jac
path = (hat := flag_hat()) +>:NextOf:+> (loop := repeat_n(lit(10), body)) +>:NextOf:+> say_text("done!");
# hat, loop, and path.edges all in hand; nothing extracted after the fact
```

The long-term convergence this enables: the same pattern shape with `->` and binding becomes the READ dual, i.e. graph pattern matching:

```jac
# sketch only, staged last: MATCH-like queries as the dual of construction
for [prog ->:ScriptOf:-> (h := _) ->:NextOf:-> (s := Block)] {
    ...  # h bound to each script head, s to its second block
}
```

Today's edge refs are linear traversals; construction patterns and match patterns as duals is where graph languages converge (Cypher's `CREATE p = (a)-[:E]->(b)` / `MATCH` pairing exists for exactly this reason).

## What stays unchanged

Statement-position connect chains, inline edge payloads (`+>:E(x=1):+>`), ordered out-edges as load-bearing semantics, `del` on edge values, edge refs and filters, walkers and ability dispatch. This proposal only touches what a construction *expression* evaluates to and adds literal syntax; the graph model and the walk layer are untouched.

## Costs, honestly

- `path` is a new core value type crossing all three backends.
- Construction literals are real grammar surface (JSX demonstrates the parser can carry block-shaped literals; this is still a cost).
- **It is a breaking change**: code relying on arrowhead-eval (the descend idiom) must migrate to literals or `.tail`. The tree-sitter grammar is the largest known consumer and doubles as the migration showcase.
- Pattern matching (Principle 3's endgame) is a major feature and staged last for a reason.

## Why this is cheap(er) now

#7948 centralized edge semantics: all three backends lower connect to the same `osp_conn` contract, so path-eval is one contract change (return a path record instead of the target handle) lowered identically everywhere, and the cross-backend equivalence suite (`test_osp_equivalence.jac`, `require=["sv","cl","na"]` fixtures) is the migration safety net. The centralization was justified partly on "language evolution gets cheaper"; this RFC is the first test of that claim.

## Staged plan

1. **path-eval for connect expressions** (smallest break; kills the head-vs-tail class outright; statement position unaffected; migration = mechanical `.tail` where descend relied on arrowhead-eval)
2. **construction literals** (grammar.jac and jacblocks demos migrate as the two showcases; imperative descend retires where shape is static)
3. **match patterns with binding** (read-side dual; separate design round)

## Alternative considered: the minimal non-breaking fix

If the clean break is judged too costly, the surgical alternative is a **right-associative, source-evaluating** connect variant (spelling bikesheddable, e.g. `++>>` / `+>:E:+>>`):

- `a ++>> b ++>> c` parses right-assoc as `a ++>> (b ++>> c)`, builds the chain a->b->c, evaluates to `a`
- existing `++>` (left-assoc, arrowhead-eval, descend) unchanged

| form | assoc | evaluates to | shape | idiom |
|---|---|---|---|---|
| `++>` (today) | left | arrowhead | descend deeper | trees |
| `++>>` (variant) | right | source | sibling chain, head in hand | sequences |

This fixes problem 1 only (not edge-in-hand, not the fold premise) but is non-breaking and a one-lowering-per-backend change under the centralized contract.

## References

- Field report 1: `jaseci-labs/tree-sitter-jac/grammar.jac` (descend idiom, tree construction at scale)
- Field report 2: marsninja/jacblocks dogfood log #21 (head-vs-tail; in-repo workaround: `add_script` returns its head) and the `wire()` edge re-query
- #7941 / #7948: centralized edge semantics that make the implementation tractable
- #6785: ordered out-edges as guaranteed semantics (what makes literal entry order meaningful)

marsninja:
Adds `graph { ... }` expressions that return a typed `GraphFragment` with roots, continuation tips, participating nodes, and the actual created edges. `graph` is reserved; identifiers with that spelling must be renamed or backtick-escaped. Nested fragments, helper results, lists, comprehensions, and conditional branches compose through their roots and tips; ordinary connect expressions retain their existing semantics.

Construction lowers to existing calls, loops, and connect operations, with one shared Jac runtime for bookkeeping across Python, JavaScript, and native code. Effects occur once in source order, membership is deduplicated by identity, and an immediate `.root` projection skips membership collection. Unsupported directions and suspension receive source diagnostics.

Shared compiler fixes cover closure/comprehension scope, bounded generic inference, class-object narrowing and constructor validation, union member overload specialization, runtime placement, JavaScript field constructors/accessors, and native layout/storage caching. Native cache keys retain type identity so annotations and class values with identical mangles cannot reuse incompatible storage. Grammar extraction and the generated language specification support construction syntax; language documentation and release notes are included.

Client fixture execution canonicalizes symlinked source paths and uses the client compiler's output-path mapping, fixing the macOS temporary-directory failure. A symlinked client fixture covers compilation and Node execution.

Validation:

- All required checks pass in [CI run 34297200201](https://github.com/jaseci-labs/jac/actions/runs/34297200201), including compiler, runtime, client, scale, Android, AArch64, sealed-package integration, and production package smoke tests.
- Repository formatting passes and all 1,049 files pass type checking.
- The rebuilt native compiler passes all 22 final graph, reserved-word, and cross-backend tests, including compilation and Node execution through a symlinked directory.
- The source parser passes 16 graph tests; 47 reserved-word, grammar, and edge-runtime tests pass. The generated grammar matches the checked-in specification, and all 83 bootstrap seed modules compile.
- The package smoke job passed on rerun after an initial Vite entry-module fetch failure.

Fixes #7956.


jac/jaclang/compiler/tests/xbackend_equiv.jac
```diff
@@ -133,6 +133,50 @@ def run_js_function(js_code: str, func_name: str) -> dict {
...
+            )
+                if manifest is not None
+                else []; },
+            extract_globals_func=lambda (manifest: any, module: any) { return {}; },
+            compiled_dir=output
+        );
+        source_root = Path(source.anchor);
+        compiler.compile_dependencies_recursively(source, source_root=source_root);
+        (output / "package.json").write_text('{"type":"module"}');
+        js_entry = output / source.relative_to(source_root).with_suffix(".js");
```
christianwilkins: [P2] Resolve the fixture path before calculating the generated entry. On macOS, `_fixture_shadow_dir()` normally returns `/var/folders/...`, but `compile_dependencies_recursively` resolves it to `/private/var/folders/...` before writing relative to `/`. This unresolved `source.relative_to(source_root)` then points to a different directory and `js_entry.open("a")` raises FileNotFoundError. I reproduced the failure on this head; the same helper and expected graph results pass when only TMPDIR is changed to a canonical `/private/tmp/...` path. Resolve `source` up front or reuse the compiler’s output-path calculation, and add a symlink-path regression.

christianwilkins (changes requested):
Reviewed draft head `8c13a5049094dd317c73cefd61753938ab294c92`, including the full patch and the follow-up commits after `34fc30615` that preserve annotation tuples, union class/metaclass lookup, and optional grammar arguments.

P2: the new `run_js_module_fixture` helper fails with the default macOS temporary directory. `JacClientCompiler.compile_dependencies_recursively` resolves the fixture path, but `js_entry` is computed from the unresolved `/var/...` path. The compiler writes under `private/var/...` relative to `/`, while the helper opens `var/...`, producing `FileNotFoundError` before Node runs. I reproduced this both through the new parity test body and by calling the helper directly. The identical direct-helper probe passes all expected graph results when only `TMPDIR` is changed to a canonical `/private/tmp/...` directory. Please resolve the source path consistently (or use the compiler’s output-path calculation) and cover a symlinked temporary directory.

Validation on this head: 15 graph syntax/lowering cases, 2 JavaScript constructor/accessor cases, 12 added generic/narrowing/async type cases, and 12 server-runtime graph cases passed by invoking their test bodies directly against the production APIs. `git diff --check` passed. The default-path client parity probe failed as described; its canonical-path control passed. A full three-backend rerun was stopped while waiting for embedded PostgreSQL startup, so independent native parity and a rebuilt native compiler/distribution are not validated here.

Latest hosted run `34282168174` has passing contribution checks and both bootstrap jobs; `build-kit` is still running and the dependent test lanes have not completed. The failed build on `34fc30615` is not a result for this newer head. GitHub reports no conflicts. Changes requested for the portable test-helper path; the PR remains a draft pending that fix and completed current-head validation.

PR #9069, opened 2026-09-08
fix(jaclang): read-tier flush reopens READ ONLY before INSERT under concurrency (#9044)

### LINKED ISSUE #9044: 0.37.7 read tier: overlapping requests of the same user fail with Postgres 25006 "cannot execute INSERT in a read-only transaction" instead of ReadOnlyRetry
## Summary

On jac 0.37.7, when two or more requests **for the same user** overlap and the walker is one the process has not yet observed writing (so it starts on the read tier, #8711), one of the requests fails at commit with:

```
jaclang.data.pgwire.PgWireError: {'S': 'ERROR', 'V': 'ERROR', 'C': '25006', 'M': 'cannot execute INSERT in a read-only transaction', 'F': 'utility.c', 'L': '407', 'R': 'PreventCommandIfReadOnly'}
```

instead of the documented `ReadOnlyRetry` re-run at SERIALIZABLE. The same calls made sequentially never fail. `JAC_DB_RO_UNITS=0` makes the concurrent calls pass.

The walker shape that triggers it: one graph traversal, then a field update on an existing node, no new nodes or edges. A walker that creates a node (`++>`) is retried correctly.

## Environment

- jac 0.37.7 (Darwin arm64, `jac run --serve`, embedded Postgres 18) -- reproduces.
- jac 0.37.7 pod (`jaseci/jaclang:0.37.7`) on Kubernetes behind the gateway, provisioned `postgres:18`, deployed through jachammer -- reproduces identically (1 of 4 concurrent requests fails, every round).

## Minimal reproduction (standalone project)

`jac.toml`

```toml
[project]
name = "rorepro"
entry-point = "main.jac"
kind = "service"
jac-version = "==0.37.7"
```

`main.jac`

```jac
node Meta {
    has name: str = "";
}

# First call creates the node (a writer). Every later call is the shape that
# breaks: one traversal, then a field update on the existing node.
walker Rename {
    has name: str = "";

    can go with Root entry {
        existing = [here-->[?:Meta]];
        meta = existing[0] if existing else (here ++> Meta());
        meta.name = self.name;
        report {"name": meta.name};
    }
}
```

`repro.py` (register + login, then fire `Rename` 4 at a time with the same token; `sequential` runs the same 4 calls one after another)

```python
import json, sys, threading, time, urllib.request, urllib.error
BASE = sys.argv[1]; MODE = sys.argv[2]   # seed | concurrent | sequential
def call(method, path, body=None, token=None):
    req = urllib.request.Request(BASE + path, data=(json.dumps(body).encode() if body is not None else None), method=method)
    req.add_header("Content-Type", "application/json")
    if token: req.add_header("Authorization", "Bearer " + token)
    try:
        with urllib.request.urlopen(req, timeout=60) as r: return r.status, json.loads(r.read().decode() or "null")
    except urllib.error.HTTPError as e: return e.code, json.loads(e.read().decode() or "null")
STATE = "/tmp/rorepro-state.json"
if MODE == "seed":
    email = f"u{int(time.time())}@example.test"
    call("POST", "/user/register", {"identities":[{"type":"email","value":email}],"credential":{"type":"password","password":"Passw0rd-123"}})
    st, res = call("POST", "/user/login", {"identity":{"type":"email","value":email},"credential":{"type":"password","password":"Passw0rd-123"}})
    tok = res["data"]["token"]; json.dump({"tok": tok}, open(STATE, "w"))
    print("seed Rename:", call("POST", "/walker/Rename", {"name": "seed"}, tok)[0])
else:
    tok = json.load(open(STATE))["tok"]
    for rnd in range(3):
        out = []
        def one(i):
            st, res = call("POST", "/walker/Rename", {"name": f"r{rnd}-{i}"}, tok)
            err = (res or {}).get("error")
            out.append((i, st, (err["message"][:70] if err else "ok")))
        if MODE == "concurrent":
            ts = [threading.Thread(target=one, args=(i,)) for i in range(4)]; [t.start() for t in ts]; [t.join() for t in ts]
        else:
            for i in range(4): one(i)
        print(f"{MODE} round {rnd}:", sorted(out))
```

Each step below is a **fresh server process** (`jac run --serve -p 8123 main.jac`), so `Rename` is a never-observed unit when the calls arrive:

```
$ jac --version
jac 0.37.7  (Darwin arm64)

# process 1
seed Rename: 200

# process 2
sequential round 0: [(0, 200, 'ok'), (1, 200, 'ok'), (2, 200, 'ok'), (3, 200, 'ok')]
sequential round 1: [(0, 200, 'ok'), (1, 200, 'ok'), (2, 200, 'ok'), (3, 200, 'ok')]
sequential round 2: [(0, 200, 'ok'), (1, 200, 'ok'), (2, 200, 'ok'), (3, 200, 'ok')]

# process 3
concurrent round 0: [(0, 200, 'ok'), (1, 500, "{'S': 'ERROR', 'V': 'ERROR', 'C': '25006', 'M': 'cannot execute INSERT"), (2, 500, "... 25006 ..."), (3, 200, 'ok')]
concurrent round 1: [(0, 500, "... 25006 ..."), (1, 500, "... 25006 ..."), (2, 200, 'ok'), (3, 200, 'ok')]
concurrent round 2: [(0, 500, "... 25006 ..."), (1, 500, "... 25006 ..."), (2, 500, "... 25006 ..."), (3, 200, 'ok')]

# process 4, JAC_DB_RO_UNITS=0
RO_UNITS=0 concurrent round 0: [(0, 200, 'ok'), (1, 200, 'ok'), (2, 200, 'ok'), (3, 200, 'ok')]
RO_UNITS=0 concurrent round 1: [(0, 200, 'ok'), (1, 200, 'ok'), (2, 200, 'ok'), (3, 200, 'ok')]
RO_UNITS=0 concurrent round 2: [(0, 200, 'ok'), (1, 200, 'ok'), (2, 200, 'ok'), (3, 200, 'ok')]
```

## How it was found (real app, Kubernetes)

A GitHub App install callback walker (`CompleteGithubInstall`: `[here-->[?:GithubConnection]]`, then field writes on that node) returned 500 on a freshly deployed 0.37.7 pod. The client page mounts twice on a full load and fires the walker twice, so two requests for the same user overlap. Full traceback from the deployed pod:

```
File ".../site/jaclang/server/server.jac", line 213, in _run_function_with_occ
File ".../site/jaclang/server/server.jac", line 361, in run_attempt
File ".../site/jaclang/server/impl/server.impl.jac", line 164, in _finalize_call_response
File ".../site/jaclang/runtime/runtime.jac", line 227, in acommit
File ".../site/jaclang/runtime/memory.jac", line 172, in acommit
    self.commit(anchor);
File ".../site/jaclang/server/impl/session.impl.jac", line 415, in commit
File ".../site/jaclang/server/impl/session.impl.jac", line 716, in flush
File ".../site/jaclang/server/impl/session.impl.jac", line 759, in _flush_locked
File ".../site/jaclang/data/impl/store.impl.jac", line 912, in upsert
File ".../site/jaclang/data/impl/store.impl.jac", line 402, in _run
File ".../site/jaclang/data/impl/pgwire.impl.jac", line 441, in run
jaclang.data.pgwire.PgWireError: {'S': 'ERROR', 'V': 'ERROR', 'C': '25006', 'M': 'cannot execute INSERT in a read-only transaction', 'F': 'utility.c', 'L': '407', 'R': 'PreventCommandIfReadOnly'}
```

On that deployment, with a throwaway account: 8 sequential calls of a sibling walker of the same shape (`StartGithubInstall`: traversal, then two field writes) were 8/8 ok; 4 concurrent calls of it failed 1 of 4 in each of 3 rounds with the same 25006 error. Walkers that create nodes (`SaveProject`, `CreateTask`) never failed.

## Reading of the code (v0.37.7, 3ad35097c)

Hypothesis from reading, not a debugged trace:

- `Session._txn_isolation` (`server/impl/session.impl.jac:491`) returns `SNAPSHOT_READ_ONLY` whenever `_ro_units == _active_units` and `_dirty`/`_deleted` are empty. Only `Memory.put` / `mark_dirty` (`runtime/memory.jac:49,58`) and `_recover_conflict` populate `_dirty`; a plain field edit on a loaded anchor is detected by hashing inside `_flush_locked` (`:724`), so it never changes that verdict.
- In `_flush_locked` (`:753`), the `ensure_writable_txn()` branch, which is the one that raises `ReadOnlyRetry` (`:877`, `_raise_ro_retry` `:520`), is only taken when the store is *currently* in a read-only transaction. If the store is outside a transaction at that point (the snapshot was closed by `read_scope_exit` `:593` or by a sibling unit's commit/rollback on the same session), the `else` branch calls `_ensure_txn()` (`:548`), which asks `_txn_isolation()` and issues `BEGIN ... READ ONLY` (`PgStore.begin` `data/impl/store.impl.jac:770`) immediately before `upsert`. That matches the traceback: `_flush_locked` goes straight to `upsert` with no retry frame, and the `except ReadOnlyRetry` handler (`server/impl/server.impl.jac:217`) never runs.
- With a single request the read-only transaction is still open at flush, so the retry fires and everything works, which is why sequential calls pass.

## Expected

Either the write should raise `ReadOnlyRetry` and re-run at SERIALIZABLE regardless of overlapping units, or `_ensure_txn` should never open a read-only transaction when it is about to upsert rows.

## Workaround

`JAC_DB_RO_UNITS=0` (restores SERIALIZABLE everywhere), verified above.

## Related note

The retry design re-runs the whole unit. A walker that performs a non-idempotent external call before its first write (here, exchanging a single-use GitHub OAuth `code`) cannot survive the re-run even when the retry works. A per-walker way to opt out of the read tier, or a note in the persistence docs, would help.

MalithaPrabhashana:
## Problem

On the read tier (#8711), overlapping requests can fail at commit with a raw Postgres error instead of the intended transparent retry:

```
jaclang.data.pgwire.PgWireError: {'C': '25006',
  'M': 'cannot execute INSERT in a read-only transaction', ...}
```

It strikes a walker that traverses the graph and then updates a field on an existing node (no `++>`). Run sequentially the same calls always pass; only overlap triggers it, and `JAC_DB_RO_UNITS=0` makes it disappear. Fixes #9044.

## Root cause

A field write on a loaded anchor does not populate `_dirty`; it is discovered later by hashing inside `_flush_locked`. So the read-tier heuristic `_txn_isolation()` keeps returning `SNAPSHOT_READ_ONLY` right up to the flush.

`Session` is shared across every request context in the process, so a sibling request's `unit_exit` / `read_scope_exit` can commit and close the shared store transaction while another request is mid-flight. When that request reaches its flush with rows to write, the dispatch was:

```jac
if rows or self._deleted {
    if self.store.txn_isolation() is TxnIsolation.SNAPSHOT_READ_ONLY {
        self.ensure_writable_txn();   # correct: retries / upgrades
    } else {
        self._ensure_txn();           # store is BETWEEN txns -> reopens READ ONLY
    }
}
```

With the transaction already closed, `store.txn_isolation()` is `None`, so the `else` branch runs `_ensure_txn()`, which re-consults the stale heuristic and opens `BEGIN ... READ ONLY` immediately before the `INSERT`. Postgres rejects it (25006), and because no `ReadOnlyRetry` was raised, the retry handler never runs and the raw error escapes as a 500.

Instrumented flush state at the moment of failure (every failing request identical):

```
store_iso=None  in_txn=False  dirty=0  deleted=0  ro_units=1  active_units=1  wanted=SNAPSHOT_READ_ONLY
```

The successful concurrent requests differ only by timing: they flush while a sibling still holds the store in `SERIALIZABLE`. Present since the read tier landed in v0.37.1, not specific to 0.37.7, and not specific to one user.

## Fix

The flush already has the pending rows in hand, so it must guarantee a writable transaction. `ensure_writable_txn()` is exactly that primitive and already handles all three store states correctly (in a read-only txn -> retry/upgrade; in no txn -> open `SERIALIZABLE`; already writable -> no-op). Route every flush-write through it and drop the branch that re-derived isolation from the read-tier heuristic at write time.

```jac
if rows or self._deleted {
    self.ensure_writable_txn();
}
```

`_ensure_txn()` remains correct for the read paths that legitimately open the read tier (`read_barrier`, `unit_enter`); this change only stops the write path from using it.

### Before

```mermaid
sequenceDiagram
    participant A as Request A (writer, on read tier)
    participant S as Shared Session / Store
    participant B as Request B (sibling)
    A->>S: BEGIN READ ONLY (optimistic reader)
    A->>S: traverse graph
    B->>S: unit_exit -> COMMIT (closes the shared txn)
    Note over S: store now has NO open transaction
    A->>S: flush: rows to write, _dirty empty
    Note over A,S: store.txn_isolation() is None -> else branch
    A->>S: _ensure_txn() -> BEGIN READ ONLY (stale heuristic)
    A->>S: INSERT
    S-->>A: 25006 cannot execute INSERT in a read-only transaction
    Note over A: no ReadOnlyRetry raised -> 500 to client
```

### After

```mermaid
sequenceDiagram
    participant A as Request A (writer, on read tier)
    participant S as Shared Session / Store
    participant B as Request B (sibling)
    A->>S: BEGIN READ ONLY (optimistic reader)
    A->>S: traverse graph
    B->>S: unit_exit -> COMMIT (closes the shared txn)
    Note over S: store now has NO open transaction
    A->>S: flush: rows to write
    A->>S: ensure_writable_txn()
    Note over A,S: no txn open -> BEGIN SERIALIZABLE
    A->>S: INSERT
    S-->>A: ok (version CAS still guards concurrent writes)
```

Flush write-path decision, before vs after:

```mermaid
flowchart TD
    subgraph Before
        R1[rows or deleted to flush] --> Q1{store currently<br/>READ ONLY?}
        Q1 -- yes --> W1[ensure_writable_txn]
        Q1 -- no --> E1[_ensure_txn]
        E1 --> C1{store in a txn?}
        C1 -- no --> RO1[BEGIN READ ONLY<br/>from stale heuristic]
        RO1 --> X1[INSERT -> 25006]
    end
    subgraph After
        R2[rows or deleted to flush] --> W2[ensure_writable_txn]
        W2 --> D2{store state}
        D2 -- READ ONLY --> RT2[retry / upgrade to SERIALIZABLE]
        D2 -- no txn --> SR2[BEGIN SERIALIZABLE]
        D2 -- writable --> NO2[no-op]
    end
```

## Validation

- New regression test `isolation: a read-tier unit that writes after its snapshot closed opens a writable txn (#9044)` in `test_read_path_isolation.jac`, in-process at the `Session` level. It reproduces the production error frame-for-frame (`PgWireError` 25006 at `pgwire.impl.jac`) on `main` and passes with the fix.
- `test_read_path_isolation.jac`: 6 passed on `main`, 7 passed with the fix.
- Adjacent suites unchanged: `test_pg_uow`, `test_read_tier_observe`, `test_scale_db_stack`, `test_serve_client`.
- Live `jac run --serve` on embedded Postgres 18, fresh process per phase: concurrent same-user 12/12, concurrent across two users 12/12, sequential 12/12, zero 25006. Baseline `main` failed 5-6 of 12 concurrent.


christianwilkins (commented):
Reviewed `7efa33163f9ad31d7a61e0838ffded17fbb57f2d`, including the complete diff and the surrounding flush, transaction-upgrade, retry, and snapshot paths. No blocking findings.

I independently reproduced the regression on macOS using this checkout's compiler source. `test_read_path_isolation.jac` passes all 7 tests with the patch. Keeping the new test and restoring the base implementation produces 6 passed and 1 error: Postgres SQLSTATE 25006, “cannot execute INSERT in a read-only transaction,” at the new flush regression. The PR source was restored afterward and the worktree is clean; `git diff --check` passes.

The change in `4049f81aa` uses the right existing primitive: a pending write with no open transaction starts SERIALIZABLE, an existing read-only transaction retains its retry/upgrade behavior, and a writable transaction stays open. The later `7efa33163` commit adds only the release note. I made no fixes during review and did not repeat the live concurrent HTTP exercise reported in the description.

Current-head CI run 34266656227 has passed the runtime and Scale lanes; `jac-check` is still running. Pre-commit.ci is green and the PR is conflict-free. Status: no author changes requested from this review; final CI completion remains before merge readiness can be confirmed.

jac/jaclang/server/impl/session.impl.jac
```diff
@@ -752,11 +752,7 @@ impl Session._flush_locked(full: bool) -> None {
             pending.append((anchor, h));
         }
         if rows or self._deleted {
-            if self.store.txn_isolation() is TxnIsolation.SNAPSHOT_READ_ONLY {
-                self.ensure_writable_txn();
-            } else {
-                self._ensure_txn();
-            }
+            self.ensure_writable_txn();
```
kashmithnisakya, line 755: This call now reaches the `elif not self.store.in_txn()` branch of `ensure_writable_txn` (line 888) on the write path, and unlike the `SNAPSHOT_READ_ONLY` branch above it that branch never checks `_ro_units`. So a read-tier unit whose snapshot a sibling closed writes in place: its reads came from the closed REPEATABLE READ snapshot, `_drop_snapshot_reads` is skipped, and `record_unit_writes` never fires, so the walker stays on the read tier. The module docstring promises a read-started transaction is restarted at SERIALIZABLE before the write is issued, and the version CAS in `upsert` only covers the rows written, not read-then-write across rows.

Should the branch mirror its sibling (`if self._ro_units > 0 { self._raise_ro_retry(); }` before `begin(SERIALIZABLE)`), with the new test expecting `ReadOnlyRetry` for a read-tier unit? That may also remove the 60s stall described in the review body.

kashmithnisakya (approved):
Verified on a real app (Flowline on this branch via the dev-source loop, `jac run --no-dev -w 1`, embedded Postgres 18), same worktree and store for every build, 4 overlapping same-user writes per round (`RenameFlowLine`, `StartGithubInstall`, and a doubled `CompleteGithubInstall`, the shape from #9044):

| Build | Overlapping writes ok | Failure |
| --- | --- | --- |
| main | 56 / 120 | 25006 |
| this PR | 306 / 308 | see below |
| main + `JAC_DB_RO_UNITS=0` | 232 / 232 | none |

Sequential calls pass everywhere. The new test is real: with the `_flush_locked` change reverted it fails with the production error (`PgWireError` 25006 at `pgwire.impl.jac`), and passes with it (7/7).

One symptom to record. In 4 of 6 fresh processes, the first overlapping burst of a walker stalled for 60s: two of the four returned 200 after 60s, one returned 500 with 55P03 `canceling statement due to lock timeout` / `while locking tuple ... in relation "anchors"`. The 60s is `Session.commit()` retrying MAX_RETRIES against the 10s `lock_timeout` set in `PgStore.begin`. It never happened once the walker was recorded as a writer (about 100 later rounds), never on main, and never with `JAC_DB_RO_UNITS=0`, so it seems tied to the in-place upgrade in the no-transaction branch (see the inline note). Since the writer record expires after `JAC_DB_RO_WRITER_TTL_S`, an app would see this again after each quiet stretch. Not root-caused; a follow-up issue is fine if it is not folded in here.

Approving: this is a clear improvement over main and the change is minimal.

PR #9070, opened 2026-09-08
release: jaclang 0.37.10 with macOS startup fix

marsninja:
The 0.37.9 release failed its macOS smoke test because the launcher imported Linux's `__errno_location`. Native sibling resolution selected `_errno_platform.darwin.jac` but returned `_errno_platform.jac` as the relative dependency source, causing linking to load the Linux implementation. Preserve the selected filename so resolution and linking use the same module.

Release 0.37.10 with the fix, collect pending release notes, and synchronize the site version pin. The existing 0.37.9 draft and tag remain intact, following the documented recovery path for a tag that already has assets.

Validation:
- 24 targeted front-end and platform-resolution tests passed.
- Regression test fails before the fix and passes after it, covering macOS and Linux target selection.
- Built the macOS launcher using the release workflow's checksum-verified LLVM shim. Its imports contain `__error`, with no `__errno_location`; execution reaches the expected missing-payload diagnostic instead of aborting in dyld.

Failed release: https://github.com/jaseci-labs/jac/actions/runs/34260963156


christianwilkins (commented):
Reviewed `fc619ec4cbbbb9a6cdbe02810797245c3542f9d3`, including the resolver's callers, regression test, version pins, and release-note collection. No blocking findings in the diff.

I independently ran `jac test jac/tests/compiler/test_native_platform_resolution.jac` on macOS using this checkout's compiler source: 1 passed, covering both Darwin and Linux target selection. Restoring only the old `return (f"{clean}.jac", local)` makes that same test fail at the source/resolved-path equality assertion. The fix is restored, the worktree is clean, and `git diff --check` passes.

Returning the selected file relative to `base_dir` agrees with the existing dotted-module resolution branch and preserves the `.darwin.jac` suffix for the compiler path that consumes the first tuple element. The root version and site pin both read 0.37.10, and the removed fragments are included in the new release notes. No fixes were made during this review.

CI run 34270151515 is still running on this head. Linux/macOS bootstrap and pre-commit.ci passed; the `macos-native` job was skipped. I did not independently rebuild the release launcher or run a packaged release artifact, so my local result verifies module resolution, not the full release smoke test. Status: no author changes requested from this review; full CI and the release's macOS artifact smoke check still need to complete successfully.

PR #9073, opened 2026-09-09
Fix workspace deployment, import ownership memory, and async walker execution

### LINKED ISSUE #8517: [Bug] async can abilities silently never execute — __jac_async__ flag written as dead local, read from wrong object

### LINKED ISSUE #8995: A profile's [apps.<name>.<section>] overlay is dropped, while [scale.gateway] in the same profile is applied

### LINKED ISSUE #8998: apps.md awaits a bridged call inside a plain `can`; that fails bytecode (E5043), the compiling spelling is `async can`

### LINKED ISSUE #9006: 0.37.5: one 18-line hunk in boundary_classify.jac roughly doubles compiler peak memory (1554 MB on a 177-module project)
### Summary

One 18-line hunk in `compiler/driver/boundary_classify.jac`, added in #8985
(`f779de433`), roughly doubles compiler peak memory. On a 177 module project it
costs **1554 MB**, measured by reverting that file alone.

Separately, compile peak memory became **nondeterministic** at the same point in
the series: identical repeated runs vary by up to 1.1 GB, where 0.37.1 varies by
50 MB. That part is not explained by this hunk and may be a different change.

### The hunk

```jac
-        return False;
+        resolved = bridgeable_import_target(nd);
+        if resolved is None {
+            return False;
+        }
+        import from jaclang.compiler.passes.boundary_analysis_pass {
+            _parse_module_for_types
+        }
+        linked = _parse_module_for_types(resolved);
+        if linked is None {
+            return False;
+        }
```

A path that returned `False` immediately now resolves each bridgeable import
target and parses that module for types. The cost scales with the number of
bridgeable imports, which is why a 177 module project pays about 1.5 GB and a
3 app toy project pays nothing measurable.

### Evidence

Two source builds, identical flags, same project, same session, differing only
in that one file:

```
git checkout f779de433
git checkout e95040309 -- jac/jaclang/compiler/driver/boundary_classify.jac
# 1 file changed, 2 insertions(+), 16 deletions(-)
```

Peak RSS from `/usr/bin/time -l`, cold `.jac/cache` before each run:

| build | median | runs |
|---|---|---|
| `f779de433` as shipped | **3709 MB** | 3672, 3697, 3709, 3756, (6672) |
| `f779de433` minus the hunk | **2155 MB** | 2121, 2124, 2155, 2181, (3683) |

The shipped releases agree with that, `jac check` on one unchanged file, five
cold runs each:

| jac | median | spread |
|---|---|---|
| 0.37.1 | 1534 MB | **50 MB** |
| 0.37.2 | 2100 MB | 729 MB |
| 0.37.3 | 1434 MB | 656 MB |
| 0.37.4 | 1516 MB | 1156 MB |
| **0.37.5** | **3000 MB** | 471 MB |
| 0.37.7 | 3200 MB | 649 MB |

0.37.1 through 0.37.4 are one band with overlapping spreads; 0.37.5 is the only
step that survives repeats, and `f779de433` is inside that window.

### How it was narrowed

Bisected the ten commits of `v0.37.4...v0.37.5` with from-source builds, then
isolated within the commit. Two candidates were tested and cleared, so they do
not need re-testing:

- **#8970** (`fdd1e1058`, the compiler tree joining the analysis cache) looked
  like the obvious culprit and is not. It measures 2096 MB, below the 0.37.4
  baseline, and its env knobs (`JAC_UNIT_COLLECT_EVERY=1`,
  `JAC_UNIT_COLLECT_GROWTH_MB=32`) change nothing.
- **#8823** (`e95040309`, Workspace apps) measures 2130 MB, also low.
- The **mobui typing work** in #8985 is clear: it is present in both builds in
  the table above, so the delta is entirely the compiler hunk.

### Not proposing the revert

`test_cross_app_import.jac` arrives in the same commit, so the hunk is
presumably load bearing for cross-app import typing. Caching the parse per
resolved target, or releasing the tree once the classification answer is taken,
would keep the behaviour without the retention. Which is right is a call for
whoever owns the pass.

### Method note

Medians of five cold runs, because single runs on this project are not
trustworthy: each set above carries one high outlier under machine load. The
source builds are debug (no `-Drelease`), so their absolute numbers sit above
the equivalent release binary; the comparison is between rows built the same
way, not against the release table.

Earlier revisions of this issue quoted a "10x deploy regression" and a "2.5x
drift across five releases". Both were wrong and are retracted in the comments:
the first compared a one-app plan against a three-app plan, the second read
single noisy samples as a trend.

### LINKED ISSUE #9012: jac scale deploy: app edge analysis always throws RuntimeError (dict changed size), so the fleet silently boots in declaration order

### LINKED ISSUE #9013: jac scale deploy never builds the client for a workspace, so a web-app fleet serves 404 on / with no way to override
### Summary

In a **workspace** (any `jac.toml` with an `[apps]` table), `jac scale deploy`
never builds the client bundle for a `web-app`. The fleet deploys and reports
itself healthy, and the gateway then answers `GET /` with **404**, because no
SPA was ever built. There is no configuration that turns the build back on: the
only way in is a constructor argument the CLI does not pass.

Reproduced on **0.37.5 and 0.37.7**.

### Reproduce

Four files, no jacBuilder involved.

`main.jac`
```jac
import from .pages.Home { Home }
def:pub app() -> any { return <Home/>; }
```

`pages/Home.jac`
```jac
def:pub Home() -> any { return <div id="home">hello</div>; }
```

`svc.jac`
```jac
walker ping {
    obj __specs__ { static has methods: list = ["get"]; }
    can run with Root entry { report {"ok": True}; }
}
```

`jac.toml`
```toml
[project]
name = "w9013"
default-app = "svc"

[client]

[apps.main]
kind = "web-app"
entry-point = "main.jac"
route = "/"

[apps.svc]
kind = "service"
entry-point = "svc.jac"
route = "/api/svc"

[scale.kubernetes]
```

```bash
jac install                     # the [client] table needs the runtime closure
.jac/venv/bin/pip install pyyaml   # --dry-run needs it
jac scale deploy main --dry-run
```

Prints, on both versions, with the deploy otherwise clean (no errors):

```
The served app has no client target and the project has no client source
(no pages/ or components/ directory, no [client] table in jac.toml);
skipping the client build.
```

### Why the message is not the reason

Both halves of it are false for the project above: it has a `pages/` directory
AND a `[client]` table. `_has_client_source()` returns True here. It is simply
never consulted, because of the guard's ordering in
`jaclang/scale/deploy/target/kubernetes/manifest_builder.jac`:

```jac
client_entry = self._client_entry();
if client_entry and not self._declared_client_entry() and not self._has_client_source() {
    client_entry = "";
}
if not client_entry { ...skip... }
```

`_has_client_source()` is only reached when `client_entry` is already truthy,
and for a workspace it never is. `_client_entry()`:

```jac
cfg = self._project_config();
if cfg is not None and cfg.is_workspace {
    return "";                  # every workspace, unconditionally
}
if cfg is not None {
    entry_point = str(cfg.project.entry_point or "").strip();
    return entry_point or "main.jac";
}
```

The workspace branch returns before every single-project fallback, and
`is_workspace` is set for any config carrying `[apps]`
(`jaclang/project/impl/config.impl.jac:1298-1300`).

### Why there is no way around it

The only earlier return in `_client_entry` is `self.client_entry`, and the CLI
path constructs the target without it
(`jaclang/scale/deploy/target/factory.jac:80`):

```jac
return KubernetesTarget(config=k8s_config, logger=logger, dry_run=dry_run);
```

so it is `None`. `client_entry` is not a `jac.toml` key (it is absent from
`scale/config/plugin_config.jac`) and is not documented, so from the CLI a
workspace `web-app` cannot be made to build its client.

### Not a workaround: `path`-rooting the web app

Pointing the app at its UI directory makes the deployed tree that directory, and
the sibling services then fail:

```
app 'svc': entry '.../svc.jac' resolves outside the deployed tree '.../apps/web',
so the bundle cannot ship it. Deploy from the project root.
```

### What I would expect

Either the workspace path resolves the served app's own `entry-point` as the
client entry, or `client_entry` becomes a documented config key so a workspace
can declare one. As it stands a `web-app` in a workspace looks deployable and
comes up serving nothing, with a diagnostic that points at the wrong cause.

### Workaround in use

Build it out of band and stage it where the deploy would have put it, since
`_is_bundle_required` hard-codes `.jac/client/dist` as always-shipped:

```bash
jac build <app> --as client
cp -r .jac/client/<app>/dist .jac/client/dist
```

marsninja:
Workspace deployments could skip their web bundle, ignore per-app profile settings, and lose dependency ordering while compiling. Plain walkers with async abilities could also finish without executing those abilities. This PR fixes those paths and the invalid await example/checker behavior.

| Issue | Validation and change |
| --- | --- |
| #9013 | Confirmed the workspace client entry is empty. Use the served fleet member's client entry, including when the default app is a service. Exercise dry-run selection and stage an app-named client dist for the gateway. |
| #9012 | A deterministic regression grows both compiler hubs during analysis and raises the reported RuntimeError on the released compiler. Read to a fixed point, tracking module identity, so newly registered modules contribute their edges. |
| #9006 | Confirmed ownership classification calls the full type parser, which enters compiler placement inference. Replace that query with syntax parsing and a bounded, stat-keyed boolean cache. A regression forbids compiler reentry and verifies cache invalidation; existing cross-app typing tests pass. |
| #8995 | Profile regressions fail on the released compiler. Merge profile/local app tables and preserve nested deployment settings, including init containers and secret volumes. |
| #8998 | Correct both copies of the example to async can and reject await outside async functions/abilities with E2085. Client test bodies are implicitly async; plain functions and lambdas retain their scope checks, including in client code. |
| #8517 | The released compiler drops the ability coroutine and fails the state assertion. Infer the walker async flag from event abilities. With no running event loop, spawn completes traversal; inside an event loop it returns a coroutine. Cover inheritance, traversal, exceptions, and concurrent spawns. |

Validation: the original profile, fleet, client staging, cross-app compiler, diagnostic, and async language suites passed (165 tests, 5 skips). CI-focused regression runs also passed: 206 compiler tests (5 skips), 6 desktop/SSO client tests, 5 guide tests, and the full-stack client/server integration test. The two isolated hub-growth/ownership regressions also fail on the released compiler and pass with these changes. Precommit passed on all 17 changed checkable Jac files; the intentionally invalid diagnostic fixture is covered by its diagnostic test.

The reporter's 177-module project and its 1554 MB RSS delta were not reproduced locally. The memory fix removes the identified allocation/retention path while preserving tested cross-app classification; this PR does not claim a measured RSS reduction or resolution of the separately reported nondeterministic peaks. Kubernetes deployment and a real npm bundle build were not run; client staging uses the existing recording-build harness.

Incorporates the fixes from #9052, #9064, and #9017 with original commit authorship preserved, and adds the remaining fixes and regression coverage.

Closes #9013.
Closes #9012.
Closes #9006.
Closes #8995.
Closes #8998.
Closes #8517.


MalithaPrabhashana (commented):
Post-merge validation of all six items, run on this machine rather than read off the diff. Verdict: **all six are fixed**, and the memory item does measurably more than the PR claims.

## Method

Two worktrees, `pr-9073` (`4fe843fd4`) and its base (`987b88fec`), each with its own HOME so the compile caches never mix. The PR's test files were checked out onto the base source, so every row below is the same test against old and new compiler.

One setup note for anyone reproducing this: a fresh worktree has no `jaclang/vendor/typeshed/stdlib`, and `[dev] jaclang_source` is then silently ignored with the bundled compiler serving the run. That makes the PR and the baseline look identical. Both checkouts were confirmed with a probe printing `NodeAnchor.__module__`'s file, and both resolved into their own tree before any measurement was taken.

## Red to green

| Issue | Test | Base `987b88fec` | PR `4fe843fd4` |
| --- | --- | --- | --- |
| #9013 | `test_client_build_stage` :: workspace served web app declares its client even without root client hints | 18 passed, **1 failed** | **19 passed** |
| #9012 | `test_fleet_from_apps` :: edge analysis visits modules registered while analyzing either hub | 18 passed, **1 error** | **19 passed** |
| #9006 | `test_cross_app_import` :: ownership classification parses syntax without retaining type analysis trees | **1 failed** | **1 passed** |
| #8998 | `test_new_diagnostics -f E2085` | 1 passed, **2 failed** | **3 passed** |
| #8995 | `test_app_scale_overrides`, incl. "a profile retunes [apps.<name>.scale], not only [scale.gateway]" | 8 passed, **2 failed, 1 error** | **11 passed** |
| #8517 | `test_language` :: plain walkers execute async abilities in sync and async callers | **1 failed** | **1 passed** |

#8517 also reproduces behaviourally. Running `tests/language/fixtures/inferred_async_walker.jac` directly:

```
BASE  assert w.total == 5   <- fails, the async abilities silently never ran
PR    async abilities completed
```

#8998's fixture is the right shape: a plain `can` awaiting next to an `async can` awaiting, and the test asserts **exactly one** E2085, so the check does not over-fire on the valid ability. Both copies of the doc example are corrected to `async can`.

## #9006: the RSS reduction the PR declined to claim

The PR says it "does not claim a measured RSS reduction". It can. Measured on jacBuilder (825 `.jac` files in the root project), `jac check builder_sv.jac`, `.jac/cache` removed before every run, peak RSS from `/usr/bin/time -l`, four runs per side:

```
base   3386, 3395, 3393, 3417 MB   mean 3398, spread 31 MB
PR     1844, 1870, 1840, 1858 MB   mean 1853, spread 30 MB
                                   reduction 1545 MB (45.5%)
```

That lands on the 1554 MB the issue attributed to the hunk. (A first run on each side was discarded as warm-up: 3797 base / 3667 PR, before the checkout's own compiler cache had settled.)

One correction to my own issue while I am here. The second half of #9006, compile peak varying by up to 1.1 GB run to run, is **already gone on the base**: spread is 31 MB there and 30 MB on the PR. So that half was resolved earlier in the series and should not be credited to this PR. It is fixed; it is just not fixed here.

## Regressions

None. `jaclang/scale/tests/deploy` as a directory:

```
base   371 passed, 3 failed, 8 error, 6 skipped
PR     372 passed, 2 failed, 8 error, 6 skipped
```

Diffing the failure names, nothing fails only on the PR, and the single move is the #9013 test going green. The remaining failures are byte-identical on both sides and environment dependent (`test_dry_run_purity`, `test_deploy_banner`, `test_dry_run_plan_rate_limit`, `sdk_deploy_driver`).

Adjacent suites on the PR, all clean: `test_cross_app_import` 15, `test_client_codegen` 15, `test_typed_interop` 37, `test_new_diagnostics` 99 (+5 skipped), `test_language` 110, `jaclang/scale/tests/apps` 30.

## Two notes

**Not in a release yet.** The squash `09e14566b` is on `main` (2026-09-08) but in no tag through **v0.37.10**, which was cut the same day. Anything pinned to a release, jacBuilder included, cannot pick these up without `dev` or a 0.37.11.

**`test_deploy_k8s.jac` reports "1 skipped" by default** and therefore covers nothing, because the whole file sits behind `require_flag("JAC_TEST_K8S", ...)`. With the flag set, 17 of its 19 pass on mocks and only 2 need a live cluster ("early exit", "deployment target methods"). Since this file carries part of #8995's manifest coverage (init containers, secret volumes), it may be worth splitting the mockable majority out from behind the cluster flag so CI actually exercises them.

## Not covered

No real Kubernetes deploy and no real npm bundle build, the same gap the PR declares. The two cluster-dependent k8s tests were not run. Memory was measured on one project, macOS arm64.

Issue #9058, opened 2026-09-08
Jac Scale: node-exporter and Alloy are per-app-namespace DaemonSets, so agent pods scale as nodes x apps and exhaust the node max-pods cap


udithishanka:
## Summary

jac-scale deploys the observability agents **per app namespace**, and two of them are DaemonSets: `_deploy_node_exporter` and `_deploy_alloy` in `monitoring.jac`. Both carry a blanket `NoSchedule` toleration and no `nodeSelector` or `affinity`, so each deployed app claims one node-exporter pod and one Alloy pod on **every node in the cluster**, doing the same node-level work the previous app's copies already do. Agent pod count grows as `nodes x apps` while the work is inherently per-node.

On a 15-node production cluster with five apps running the pair, this is 90 of 610 pods, 15% of everything scheduled, and it has begun exhausting the kubelet `max-pods` cap: agent pods now sit Pending indefinitely and the next workload rejected on a full node will be a user app, not an agent.

## Environment

| | |
|---|---|
| jac | `0.34.17 (Darwin arm64)`, released binary on PATH |
| repro host | macOS 25.3.0 arm64, no cluster required |
| also present on | `main` @ `9b02a4b2d` (verified by reading, line refs below) |
| observed in production | EKS 15 nodes, `us-east-2`, VPC CNI |

The code is unchanged between v0.34.17 and `main`: the blanket toleration is at `monitoring.jac:599` (node-exporter) and `monitoring.jac:2004` (Alloy) on `main`, and #8413 did not touch the DaemonSet shape.

## Reproduction (executed, not hypothetical)

No cluster needed. `MonitoringDeployer` builds the manifests as plain dicts, so recording fakes are enough to show what it would create. `_deploy_alloy` reaches for a `RbacAuthorizationV1Api` the caller cannot inject, so the class itself is swapped.

```jac
# repro.jac
import from jaclang.scale.deploy.target.kubernetes.monitoring { MonitoringDeployer }
import from jaclang.scale.deploy.target.kubernetes.kubernetes_config { KubernetesConfig }
import from kubernetes { client }
import from kubernetes.client.rest { ApiException }
import json;

class Recorder {
    def init(self: Recorder) -> None { self.created = []; }
    def __getattr__(self: Recorder, name: str) -> any {
        created = self.created;
        if name.startswith("read_") {
            def _read(**kw: any) -> any {
                raise ApiException(status=404, reason="Not Found");
            }
            return _read;
        }
[...]

Issue #7766, opened 2026-07-28
jac-scale: privileged-namespace labeling for hostPath DaemonSets is defeated by jac-scale's own baseline default


kugesan1105:
## Summary

jac-scale's privileged-namespace labeling for the observability stack can never fire on namespaces jac-scale itself set up, because an earlier step in the same deploy stamps `pod-security.kubernetes.io/enforce=baseline` onto any unlabeled namespace, and the privileged-labeler refuses to touch a namespace that already carries an enforce label. Result: the hostPath DaemonSets (node-exporter, Alloy) are rejected by PodSecurity admission, and the operator has to relabel manually — defeating the automation.

## The self-defeating sequence

Both steps run inside `MicroserviceKubernetesTarget.deploy()` (`jac/jaclang/scale/deploy/target/kubernetes/microservice/target.jac`), in this order:

1. `ensure_namespace_exists(namespace)` — via `kubernetes_utils.impl.jac` — patches any namespace **lacking** an enforce label with:
   ```
   pod-security.kubernetes.io/enforce=baseline
   pod-security.kubernetes.io/warn=restricted
   ```
2. Later, `_deploy_observability` → `_ensure_privileged_namespace(core_v1, namespace)` reads the namespace and:
   - if there is **no** enforce label → patches `enforce=privileged` (the intended fix for hostPath DaemonSets), but
   - if **any** enforce label is present → logs a warning telling the operator to run `kubectl label namespace ... enforce=privileged --overwrite` manually and returns.

On a fresh cluster, step 1 always labels the namespace `baseline`, so step 2 always takes the warn-and-bail branch. jac-scale applies `baseline` with one hand and then refuses to upgrade its own label with the other. node-exporter (hostPath mounts for `/proc` and `/sys`) and Alloy (`/var/log/pods`, container log dirs) are then blocked by PodSecurity `baseline` enforcement.

## Impact

Any deployment with monitoring/logs/tracing enabled on a PodSecurity-enforcing cluster needs a permanent out-of-band step in CI:

```bash
kubectl label namespace <ns> pod-security.kubernetes.io/enforce=privileged --overwrite
kubectl rollout restart daemonset <app>-node-exporter -n <ns>
```

## Proposed fix

Options, in rough preference order:

[...]

Issue #9082, opened 2026-09-09
Jac Scale: kube-state-metrics, node-exporter and the Grafana Secret are still create-only, so their config never changes after the first deploy (#8407 closed for inactivity, no commit)


MusabMahmoodh:
## Mechanism

#8407 named the whole monitoring stack. Two commits touched `monitoring.jac`
since it was filed - `ea7f25d997` (#8412, image pins) and `7f47c09f44` (#8413,
the monitoring switch) - and neither changed an apply. The issue was then closed
on 2026-09-04 with "Closing: no activity in 10+ days."

Prometheus and Grafana got the read-then-`replace`/`patch` treatment at some
point and are fine now:

- Prometheus ConfigMap: `read` then `replace_namespaced_config_map` (`monitoring.jac:104-121`)
- Prometheus scrape Secret: `read` then `replace_namespaced_secret` (`monitoring.jac:135-150`)
- Prometheus Deployment: `patch_namespaced_deployment` (`monitoring.jac:326-338`)
- Grafana ConfigMaps: `replace` (`monitoring.jac:776-793, 817-834, 1163-1180`)
- Grafana Deployment: `patch` (`monitoring.jac:1296-1300`)

Three did not:

`monitoring.jac:540-550` - kube-state-metrics Deployment

```
try {
    apps_v1.read_namespaced_deployment(name=ksm_name, namespace=namespace);
} except ApiException as e {
    if e.status == 404 {
        apps_v1.create_namespaced_deployment(namespace=namespace, body=ksm_deployment);
```

`monitoring.jac:641-651` - node-exporter DaemonSet, same read-then-create-on-404
shape. `monitoring.jac:2060-2070` - the Alloy DaemonSet, same shape.

`monitoring.jac:700-712` - the Grafana admin Secret

```
try {
    core_v1.read_namespaced_secret(name=grafana_secret_name, namespace=namespace);
} except ApiException as e {
    if e.status == 404 {
        core_v1.create_namespaced_secret(namespace=namespace, body=grafana_secret);
```

Both ClusterRoles are create-only too (`monitoring.jac:454-462` for
kube-state-metrics, `:1943-1951` for Alloy), and so are their bindings
(`:480-488`, `:1973-1981`).

`ensure_service` is create-only by construction for every Service in the file
(`monitoring.jac:18-28`).

## Consequence

Three concrete cases, all silent:

1. **The image pin from #8412 does not reach an existing cluster.** A cluster
[...]

Issue #9077, opened 2026-09-09
Jac Scale: with no service_account_name set, Ops Console RBAC binds to the namespace's default ServiceAccount, so every pod in the namespace gets the console's rights


MusabMahmoodh:
## Mechanism

`build_ops_rbac` falls back to `"default"` when handed an empty subject name:

`ops_rbac.jac:84-90`

```
"subjects": [
    {
        "kind": "ServiceAccount",
        "name": sa_name or "default",
        "namespace": namespace
    }
]
```

The caller passes `self._service_account_name()`:

`manifest_builder.jac:703-708`

```
bundle["ops_rbac"] = build_ops_rbac(
    self.k8s_config.app_name,
    self.k8s_config.namespace,
    self._service_account_name(),
    bool(getattr(self.k8s_config, "ops_console_mutations", False))
);
```

and that helper returns the empty string unless the user set the key, which has
no default:

`manifest_builder.jac:1033-1035`, `kubernetes_config.jac:92`

```
def _service_account_name -> str {
    return str(getattr(self.k8s_config, "service_account_name", "") or "").strip();
}
...
service_account_name: str = "",
```

The pod spec only sets `serviceAccountName` when that same value is non-empty
(`manifest_builder.jac:1686-1689`), so in the default configuration the gateway
runs as `default` and the RoleBinding is written against `default`. Those two
facts are consistent, which is why it works - and it is exactly what makes the
grant wide.

## Consequence

`ops_console = true` grants, to the `default` ServiceAccount of the whole
namespace, `get/list/watch` on pods, `pods/log`, services, endpoints, PVCs and
events plus deployments/replicasets/statefulsets/daemonsets, HPAs, ingresses and
PDBs (`ops_rbac.jac:1-34`). Every pod in that namespace that does not name its
own ServiceAccount uses `default`, so the app's own containers, the postgres
StatefulSet's neighbours, and any co-tenant workload in a shared namespace all
receive it. `pods/log` in particular turns into namespace-wide log reading from
inside any container.

With `ops_console_mutations = true` the same subject additionally gets
`patch/update` on `deployments` and `deployments/scale`, `delete` on `pods`, and
`patch/update` on PVCs and ingresses (`ops_rbac.jac:35-52`). A compromised
application container can then scale the fleet, delete pods, and repoint the
ingress.

## Environment

Read on `upstream/main` at `44a1a0512c`.

## Suggested fix

[...]

PR #9085, opened 2026-09-09
fix(scale): make the node-level observability agents cluster singletons

MusabMahmoodh:
Addresses #9058. Node-exporter and Alloy were DaemonSets in every app namespace,
so agent pods scaled as `nodes x apps` while the work they do is per-node. #9058
measures the result on a 15-node production cluster: 90 of 610 pods are duplicate
agents, and four nodes are at the kubelet `max-pods` cap with agent pods Pending
for 15 hours. jacBuilder#1711 is the same failure from the operator's side.

### What changed

A new `shared_observability.jac` provisions **one** node-exporter, **one** Alloy,
**one** Prometheus and **one** kube-state-metrics per *cluster*, in a
`jac-observability` namespace (`[scale.kubernetes] observability_namespace`), with
one ClusterRole each instead of one per app.

Per-app observability becomes **data rather than infrastructure**: each app
registers a Prometheus scrape job and an Alloy log-routing lane into two
ConfigMaps (`jac-prometheus-lanes`, `jac-alloy-lanes`) via
`register_app`/`unregister_app`. The agents pick lanes up in place through a
sidecar that hits `/-/reload`, so adding or removing an app never restarts them.
The four per-app provisioning functions in `monitoring.jac` are deleted; Grafana,
Loki and Tempo stay per app.

### Against #9058's four proposals

| # | Proposal | This PR |
|---|---|---|
| 1 | Alloy stops being a DaemonSet (read logs via the API server) | **Not taken.** Item 4 instead: still a DaemonSet, but one per cluster. Item 1 leaves node-exporter untouched and still costs `N` Alloy pods; item 4 is flat in app count and fixes both. It also drops the app namespaces' hostPath mounts, so the `privileged` requirement #7766 worked around goes away there, which was item 1's other benefit. |
| 2 | Scope Alloy discovery to the app's own namespace | **Done.** Each lane carries a `keep` rule on `__meta_kubernetes_namespace`, so an app's logs come only from its own namespace and land only in its own Loki. This was a live cross-tenant defect, not just tidiness. |
| 3 | Placement constraints / `nodeSelector` plumbing | **Not addressed; stays open.** The shared agents keep `operator: Exists` deliberately: a cluster-wide node-exporter *should* cover tainted and control-plane nodes. The deploy surface still has no `nodeSelector` for app workloads. This PR does not close that half of #9058. |
| 4 | Share the node-level agents | **This PR**, with one deviation: Loki stays per app. #9058 says "one Loki with per-namespace scoping"; instead the shared Alloy routes each namespace's logs to that app's own Loki. Keeps log storage per tenant rather than pooling it. Happy to change if you prefer the pooled shape. |

### Validation

Real 2-node kind cluster, two apps in two namespaces, deployed through
`MonitoringDeployer.deploy()` itself. Each "app" is a real nginx serving a
Prometheus exposition behind basic auth with that app's own password, so the
shared Prometheus exercises the real per-app credential path.

**Counts with both apps deployed** (before: 4 DaemonSets, 8 agent pods, 2
Prometheus, 2 KSM, 6 ClusterRoles):

```
NAMESPACE           NAME                DESIRED   READY
jac-observability   jac-alloy           2         2
jac-observability   jac-node-exporter   2         2
jac-observability   jac-kube-state-metrics    1/1
jac-observability   jac-prometheus            1/1

$ kubectl get clusterrole -l managed=jac-scale
jac-alloy   jac-kube-state-metrics   jac-prometheus

agent pods per node: control-plane 2, worker 2
app namespaces: *-gw, *-grafana, *-loki only, no DaemonSet
```

**Attribution.** One Prometheus, both apps up, each scraped with its own
credential, and no cross-app bleed:

```
job=ns-alpha__alpha  health=up  ns=ns-alpha  app=alpha  value=42
job=ns-beta__beta    health=up  ns=ns-beta   app=beta   value=42
alpha's metric asked for in ns-beta  -> no series
beta's metric asked for in ns-alpha  -> no series
```

App series were already namespace-scoped by name (`monitoring.jac`'s `pfx` is
derived from the namespace), and the lane adds `namespace`/`jac_app` labels on
top. That is why a shared Prometheus does not need a cardinality budget here.

**Destroying one app** left the shared tier intact (2 DaemonSets, 1 Prometheus, 1
KSM, 3 ClusterRoles), drained only that app's scrape lane within ~60s, and left
the other app reporting.

**Also fixed on the way:** the per-app Prometheus scraped node-exporter through a
ClusterIP Service, so it only ever reached one node's exporter. The shared job
uses `role: endpoints` and both targets report.

### Not tested

The kind run shows DaemonSet count going from `O(apps)` to `O(nodes)`. It does
**not** reproduce the EKS symptom in #9058: the ENI/`max-pods` ceiling is an EKS
VPC-CNI property and a 2-node kind cluster never approaches it. The claim here is
the multiplicity, not the ceiling.

### Migration

New deploys get the shared tier. Apps already carrying a per-app stack keep it
until they are redeployed; `destroy()` still removes the old per-app objects, so a
redeploy is the migration. There is no in-place converter, and I would rather add
one as a follow-up than bundle it here.

### Consumer

jacBuilder#1726 is the companion: its metrics panel resolved
`{app}-prometheus-service` in the app namespace, and its queries selected on a
metric-name suffix alone, which on a shared instance sums every tenant into one
app's panel. Measured on the rig: unscoped `84`, scoped `42` and `42`.


jac/jaclang/scale/deploy/target/kubernetes/shared_observability.jac
```diff
@@ -0,0 +1,1200 @@
...
+        cfg = self._alloy_base_config();
+        for key in sorted(lanes.keys()) {
+            cfg = cfg + lanes[key];
+        }
+        return cfg;
+    }
+
+    def _put_config_map(core_v1: any, name: str, data: dict[str, str]) {
+        self._write_lanes(core_v1, name, data);
+    }
```
udithishanka, line 858: this is the one greptile flagged as "config updates become no-ops" and it is still here. _put_config_map calls _write_lanes with no resourceVersion, so it always takes the create path, and once jac-prometheus-config exists every later write is a 409 that gets swallowed on L305. Net effect on a real cluster: the first app's config is rendered, the second app is added to the lanes ConfigMap but never reaches prometheus.yml, so it is never scraped, and unregister drops the lane and the password file but the running config keeps the job. Same for config.alloy. Ran it against a stub that answers 409 on create like the API server does: second app's job missing from prometheus.yml and config.alloy, destroyed app's job still present. The rendered maps need the same read-rv-replace-retry that _put_lane got.

```suggestion
    def _put_config_map(core_v1: any, name: str, data: dict[str, str]) {
        attempt = 0;
        while attempt < LANE_WRITE_ATTEMPTS {
            attempt += 1;
            (existing, rv) = self._read_lanes_versioned(core_v1, name);
            if rv and existing == data {
                return;
            }
            try {
                self._write_lanes(core_v1, name, data, rv);
                return;
            } except ApiException as e {
                if e.status != 409 or attempt >= LANE_WRITE_ATTEMPTS {
                    raise;
                }
            }
        }
    }
```

jac/jaclang/scale/deploy/target/kubernetes/monitoring.jac
```diff
@@ -2115,138 +1083,37 @@ obj MonitoringDeployer {
...
-            }
-        }
-        delete_if_exists(
-            core_v1.delete_namespaced_service,
-            f"{ne_name}-service",
-            namespace,
-            'Service'
-        );
+        shared = SharedObservability(self.k8s_config, self.logger, self.rbac_api);
+        shared.unregister_app(app_name, namespace, core_v1);
```
udithishanka, line 1199: the PR body says destroy still removes the old per-app objects so a redeploy is the migration, but destroy no longer deletes any of them: {app}-node-exporter and {app}-alloy DaemonSets, {app}-prometheus and {app}-kube-state-metrics Deployments, their Services, SAs, ConfigMaps, scrape secret, or the {ns}-{app}-prometheus / {ns}-{app}-kube-state-metrics / {ns}-{app}-alloy ClusterRoles and bindings. Checked with a delete-recording stub: destroy on an app with monitoring + logs on deletes loki, tempo and grafana objects only. So on the 15-node cluster from #9058 the 90 duplicate agent pods stay exactly where they are after every redeploy or destroy, and the cluster-scoped RBAC leaks forever (target.jac's namespace-reclaim path also keys on that RBAC being gone). The old delete block from the pre-PR destroy needs to come back here as a best-effort legacy sweep, 404s ignored.

jac/jaclang/scale/deploy/target/kubernetes/shared_observability.jac
```diff
@@ -0,0 +1,1200 @@
...
+                'labels': dict(SHARED_LABELS)
+            },
+            'type': 'Opaque',
+            'data': existing,
+            'stringData': {key: password}
+        };
+        try {
+            core_v1.read_namespaced_secret(name=SCRAPE_SECRET_NAME, namespace=ns);
+            core_v1.replace_namespaced_secret(
+                name=SCRAPE_SECRET_NAME, namespace=ns, body=body
```
udithishanka: the reply on the greptile thread says lane and scrape-secret writes both carry the resourceVersion now, but only the lanes do. This is still read, pop, then a blind replace, so two deploys landing together can drop each other's password key and the loser's scrape job goes 401 until its next deploy. _drop_scrape_password has the same shape. Same fix as _put_lane: read the rv off the secret, send it in metadata, retry on 409.

jac/jaclang/scale/deploy/target/kubernetes/monitoring.jac
```diff
@@ -711,7 +62,9 @@ obj MonitoringDeployer {
             }
         }
 
-        prometheus_url = f"http://{prometheus_name}-service:9090";
+        prometheus_url = SharedObservability(
+            self.k8s_config, self.logger, self.rbac_api
+        ).prometheus_url();
```
udithishanka: worth calling out in the PR body and the config schema: every app's Grafana now has a datasource pointing at the one Prometheus that holds every tenant's series, and that Prometheus has no auth and no NetworkPolicy in front of it. The app dashboard stays scoped because pfx is namespace-derived, but Explore on any app's Grafana (default password Adminpassword123, served at /grafana on the app ingress) can query another tenant's {pfx}_http_requests_total by path. Before this change a per-app Prometheus physically only had that app's series. On jachammer this is user apps looking at each other's traffic. The [scale.gateway.logs] description already carries a multi-tenant warning for Loki; [scale.monitoring] needs the equivalent, and jacBuilder probably wants to either drop the per-app Grafana for user apps or gate it.

jac/jaclang/scale/deploy/target/kubernetes/shared_observability.jac
```diff
@@ -0,0 +1,1200 @@
...
+        }
+    }
+
+    def _ensure_deployment(apps_v1: any, name: str, body: dict[str, any]) {
+        ns = self.namespace();
+        try {
+            apps_v1.read_namespaced_deployment(name=name, namespace=ns);
+        } except ApiException as e {
+            if e.status == 404 {
+                apps_v1.create_namespaced_deployment(namespace=ns, body=body);
```
udithishanka: _ensure_deployment and _ensure_daemonset are create-if-absent only, and unlike the per-app tier there is no owner that ever recreates them. So the first jac release that lands the shared tier on a cluster pins its images, resource limits and base config there forever: a later release bumping prometheus v3.6.0 or alloy v1.11.2, or flipping node_level on for a cluster whose first app opted out, never applies. The checksum/base-config annotation is dead for the same reason, nothing ever re-applies the template. The old per-app Prometheus used patch_namespaced_deployment, which is what this wants too (patch on exists, create on 404), for both the Deployments and the DaemonSets.

jac/jaclang/scale/tests/deploy/test_monitoring_switch.jac
```diff
@@ -52,3 +79,350 @@ test "logs or tracing still request the stack without the monitoring switch" {
...
+            return _read;
+        }
+        def _write(*a: any, **kw: any) -> any {
+            body = kw.get("body");
+            if isinstance(body, dict) {
+                created.append(body);
+                meta = dict(body.get("metadata", {}) or {});
+                key = f"{meta.get('namespace', '')}/{meta.get('name', '')}";
+                store[key] = body;
+            }
```
udithishanka, line 249: this is why the suite is green with the two bugs above in it: the recorder lets create_* overwrite an existing key, so the swallowed 409 in _write_lanes never fires under test. Making the stub behave like the API server on that one point turns the two-app case into a real check, and then a test that deploys alpha and beta and asserts both job_names are in the rendered prometheus.yml (plus an unregister that asserts the job is gone) covers the register/unregister path end to end.

```suggestion
                key = f"{meta.get('namespace', '')}/{meta.get('name', '')}";
                if name.startswith("create_") and key in store {
                    raise ApiException(status=409, reason="AlreadyExists");
                }
                store[key] = body;
```

jac/jaclang/scale/deploy/target/kubernetes/shared_observability.jac
```diff
@@ -0,0 +1,1200 @@
...
+                                }
+                            }
+                        ]
+                    }
+                }
+            }
+        };
+    }
+
+    def _prometheus_deployment(config_hash: str) -> dict[str, any][str, any] {
```
udithishanka: nit: return annotation is dict[str, any][str, any] here and on _alloy_daemonset, presumably meant dict[str, any].

jac/jaclang/scale/deploy/target/kubernetes/shared_observability.jac
```diff
@@ -0,0 +1,1200 @@
...
+                                    }
+                                ]
+                            }
+                        ],
+                        'volumes': [
+                            {
+                                'name': 'alloy-config',
+                                'configMap': {'name': f"{ALLOY_NAME}-config"}
+                            },
+                            {'name': 'varlog', 'hostPath': {'path': '/var/log'}},
```
udithishanka: nit: the per-app Alloy mounted /var/log/pods; this widens it to all of /var/log on every node. The lanes only ever read /var/log/pods/*, so the narrower mount still works and is less to justify in a privileged namespace.

udithishanka (changes requested):
2 bugs that undo the feature on a real API server, 1 tenant-isolation regression, 1 upgrade gap, plus a test stub that hides the first two. Repro test attached in a follow-up comment.

jac/jaclang/scale/deploy/target/kubernetes/shared_observability.jac
MusabMahmoodh: Fixed in 582f36a8b5. `_put_config_map` now does read-rv-replace-retry, as you suggested, and `_write_lanes` no longer swallows a 409 on create. Worth being explicit: I introduced this while fixing the concurrency finding on the Greptile thread, so the kind-cluster evidence in the PR body was measured before it existed and was not evidence for the code you read. Your repro is now a permanent test (`a second app's lane reaches the rendered prometheus.yml` and the config.alloy twin).

jac/jaclang/scale/deploy/target/kubernetes/monitoring.jac
MusabMahmoodh: Fixed in 582f36a8b5. The sentence in the PR body was false when I wrote it: I had deleted that block when rewriting destroy. `_sweep_legacy_collectors` is back as a best-effort sweep, 404s ignored, covering the two DaemonSets, the Prometheus and KSM Deployments, their Services, ConfigMaps, the scrape secret, the ServiceAccounts, and the `{ns}-{app}-*` ClusterRoles and bindings that target.jac's reclaim path keys on. Pinned by `destroy sweeps the per-app collectors an older release left behind`.

jac/jaclang/scale/deploy/target/kubernetes/shared_observability.jac
MusabMahmoodh: Correct, and my reply on that thread was wrong. Only the lanes carried it. Fixed in 582f36a8b5: `_put_scrape_password` reads the secret's resourceVersion, sends it, and retries on 409, on both the set and the remove path (the two functions collapsed into one, since they only differed by whether a key was being added).

jac/jaclang/scale/deploy/target/kubernetes/shared_observability.jac
MusabMahmoodh: Fixed in 582f36a8b5. Both `_ensure_deployment` and `_ensure_daemonset` patch on exists and create on 404, the way the per-app Prometheus did, so a later release's image bump or a `node_level` flip actually applies and the checksum annotation stops being decorative. The suite caught a knock-on: a test of mine was counting writes rather than objects, which a re-apply legitimately doubles.

jac/jaclang/scale/tests/deploy/test_monitoring_switch.jac
MusabMahmoodh: Taken verbatim in 582f36a8b5, and it is the reason the rest of this review existed: a permissive stub is what let two broken paths stay green. `ApiRecorder` now answers 409 on create-existing, 404 on replace-missing, and bumps resourceVersion on every write. Your four repro cases are permanent tests in the same file.

jac/jaclang/scale/deploy/target/kubernetes/shared_observability.jac
MusabMahmoodh: Fixed in 582f36a8b5, both were meant to be `dict[str, any]`.

jac/jaclang/scale/deploy/target/kubernetes/shared_observability.jac
MusabMahmoodh: Fixed in 582f36a8b5, narrowed to `/var/log/pods` on both the hostPath and the mount. The lanes only ever glob `/var/log/pods/*`, so nothing else was using the wider mount.

jac/jaclang/scale/deploy/target/kubernetes/monitoring.jac
MusabMahmoodh: Partly addressed, and I am deliberately leaving this thread open.

582f36a8b5 adds the warning you asked for to `[scale.monitoring]`: one Prometheus holds every tenant's series, it has no auth in front of it, namespace-derived metric names scope the dashboards but not Explore, so do not enable this for mutually untrusting tenants without auth or a NetworkPolicy.

What I have not done is fix it, because I do not think it should be decided in this PR. This is a real regression in the isolation story rather than a gap in it: a per-app Prometheus physically could not hold another tenant's series, and I claimed 'no cross-app bleed' in the PR body on the strength of the dashboards alone, which was too strong. The options I can see are auth or a NetworkPolicy in front of the shared Prometheus, or jacBuilder dropping or gating the per-app Grafana for user apps, and the second is your and the platform's call rather than mine. Happy to file it as a child of #9058 with your read on which, or to take a direction here if you would rather it not ship without one.

jac/jaclang/scale/deploy/target/kubernetes/shared_observability.jac
```diff
@@ -0,0 +1,1091 @@
...
+        if want_metrics {
+            self._ensure_namespace(core_v1);
+            self._upsert_scrape_password(
+                core_v1, key, self.k8s_config.prometheus_admin_password
+            );
+            lanes = self._read_lanes(core_v1, PROMETHEUS_LANES_CM);
+            lanes[key] = self.prometheus_lane(
+                app_name, app_namespace, scrape_target, service_port
+            );
+            self._write_lanes(core_v1, PROMETHEUS_LANES_CM, lanes);
```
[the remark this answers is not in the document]
MusabMahmoodh: Fixed in e10e4b5c38, and completed in 582f36a8b5. Lane writes carry the resourceVersion they read and retry on 409. The scrape secret did not, despite my reply saying otherwise; it does now. And the rendered ConfigMaps had the opposite problem, silently no-opping after the first write, which this thread's sibling caught.

jac/jaclang/scale/deploy/target/kubernetes/shared_observability.jac
```diff
@@ -0,0 +1,1091 @@
...
+            {'app': ALLOY_NAME},
+            [{'name': 'http', 'protocol': 'TCP', 'port': 12345, 'targetPort': 12345}]
+        );
+    }
+
+    def ensure_metrics_tier(apps_v1: any, core_v1: any) {
+        self._ensure_namespace(core_v1);
+        self._ensure_node_exporter(apps_v1, core_v1);
+        self._ensure_kube_state_metrics(apps_v1, core_v1);
+        self._ensure_prometheus(apps_v1, core_v1);
```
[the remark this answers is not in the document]
MusabMahmoodh: Fixed in e10e4b5c38. `k8s_metrics_enabled` gates the node-exporter and kube-state-metrics again. One wrinkle the finding did not cover: the Prometheus config is cluster-global now, so this app's flag cannot decide the node-level scrape jobs alone or an app opting out would strip jobs another app relies on. The jobs are rendered from whether the shared collectors exist, falling back to this app's flag when they do not.

jac/jaclang/scale/deploy/target/kubernetes/kubernetes_config.jac
```diff
@@ -73,6 +73,7 @@ obj KubernetesConfig(BaseConfig) {
         service_dependencies: list[str] = [],
         monitoring_enabled: bool = False,
         k8s_metrics_enabled: bool = True,
+        observability_namespace: str = 'jac-observability',
```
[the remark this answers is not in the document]
MusabMahmoodh: Fixed in e10e4b5c38, both halves. The app namespace is no longer elevated, and target.jac's helper was dead once that call went, so it is deleted. The shared namespace is labelled whether or not this deploy created it, through the existing `privileged_pss_labels` helper, so a policy set outside jac-scale is reported rather than overwritten.

jac/jaclang/scale/deploy/target/kubernetes/shared_observability.jac
```diff
@@ -0,0 +1,1091 @@
...
+                {
+                    'kind': 'ServiceAccount',
+                    'name': sa_name,
+                    'namespace': self.namespace()
+                }
+            ]
+        };
+        try {
+            rbac_v1.read_cluster_role_binding(name=name);
+            rbac_v1.replace_cluster_role_binding(name=name, body=binding);
```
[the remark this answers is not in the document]
MusabMahmoodh: Fixed in e10e4b5c38. Cluster-scoped identities are named after the observability namespace they serve, so two namespaces cannot rewrite each other's binding subject. Still one set per cluster.

jac/jaclang/scale/deploy/target/kubernetes/shared_observability.jac
```diff
@@ -0,0 +1,1195 @@
...
+        };
+        if not resource_version {
+            try {
+                core_v1.create_namespaced_config_map(namespace=ns, body=body);
+                return;
+            } except ApiException as e {
+                if e.status != 409 {
+                    raise;
+                }
+                return;
```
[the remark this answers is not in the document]
MusabMahmoodh: This one was still live when I first replied, and only actually fixed in 582f36a8b5. `_put_config_map` took the create path with no resourceVersion and the create swallowed the 409, so every write after the first was a silent no-op and a second app never reached the rendered prometheus.yml. Read-rv-replace-retry now, and create no longer swallows the conflict it loses.

jac/jaclang/scale/deploy/target/kubernetes/monitoring.jac
MusabMahmoodh: Implemented in eecee0f654, and I took your first option rather than the proxy.

Per-app Grafana no longer gets a datasource on the shared Prometheus, and the two metric dashboards ship with it, since a dashboard whose datasource was never provisioned just renders error panels. Loki and Tempo stay, being genuinely per-app. `[scale.kubernetes] app_grafana_metrics` turns it back on for a cluster where every app is trusted, which is the self-hosted single-tenant case.

Why not the proxy: a per-app label-enforcing proxy closes the same hole at O(apps) pods, which is the cost this PR exists to remove, and it would still need identity work to stop one app calling another app's proxy. The version worth building is a shared read path that derives the tenant from the caller rather than from config, either prom-label-proxy doing authz against the API server the way OpenShift does, or a tenant-aware backend, which jacBuilder#1531 already anticipated. Filed as #9105 so `app_grafana_metrics` can default back on once it exists. Treating this PR's change as the stopgap it is rather than the answer.

Verified on a 2-node kind cluster with two apps: neither app's Grafana has a Prometheus datasource, and the shared Prometheus still scrapes both with correct attribution (`ns=ns-alpha app=alpha` / `ns=ns-beta app=beta`, 42 each).

That run also turned up a cluster-only bug none of the mocked tests could see, now fixed in the same commit. The scrape credential was mounted at `/etc/prometheus/scrape`, inside the config volume. Nested that way the directory comes back empty after a restart, `password_file` becomes unreadable, and Prometheus then answers every `/-/reload` with a 500, so the running config silently freezes at whatever the first app rendered. Same visible symptom as the bug you found, different cause, and it survived your repro because a rendered-ConfigMap check cannot see a volume mount. Moved to `/etc/prometheus-scrape`, with a test asserting the mount is never nested and the rendered lane agrees with where it lands.

Leaving this thread open for your call on whether the stopgap is acceptable to ship.

MusabMahmoodh:
@udithishanka withdrawing this one, and here is why rather than letting it rot.

Since your review it has fallen behind on two fronts. #9201 and #9350 changed `plugin_config.jac` and `manifest_builder.jac` under it, and #9231 (open) rewrites the same read-or-create and delete-if-exists blocks in `monitoring.jac` and `target.jac` that this PR edits. Rebasing now would mean re-validating on kind twice. Once #9231's casing bug is fixed, its apply helper gives exactly the replace-on-exists semantics your two criticals asked for. A cluster-singleton tier re-cut on top of it is smaller and does not carry a second apply path.

Your repro cases stay: they are the right tests for whoever re-cuts it. The five observability children from the same audit (#9074, #9075, #9076, #9077, #9082) are the natural unit to go with it, and since you own shared observability on the jacBuilder side (#1531), they are yours if you want them once #9231 lands. Otherwise I will re-cut it myself then.

PR #9117, opened 2026-09-11
fix(server): refuse an identity value only when another account owns it

### LINKED ISSUE #9103: An email-shaped username permanently locks its owner out of password reset and out of registering that address
## Summary

If a user registers with an email-shaped **username**, `jac-scale` stores it as
`identities[].type = "username"`. From that moment the address is permanently
unusable by its own owner: every email feature ignores it, while every uniqueness
check still matches it. There is no self-serve escape, and SSO actively preserves
the state rather than healing it.

Found in a production deployment; a meaningful fraction of accounts on that
instance are affected, most with real data behind them.

## Root cause: uniqueness is type-agnostic, email features are type-specific

`is_identity_taken` / `find_user_by_identity` match on `value_normalized` **regardless
of type**, but every consumer of an email filters on `type == 'email'`. The two halves
disagree about what "this address is registered" means.

## Observed behaviour

Account state (single account, nothing exotic):

```
identity: username | <the user's real email> | verified: True
identity: sso      | None                    | verified: True
```

| Symptom | Mechanism |
|---|---|
| `/user/forgot-password` silently mails nothing | recipient loop in `serve.identity.impl.jac` does `if ident.get('type') != 'email' { continue; }` → no recipient → neutral 200 |
| App's "add your email" gate never clears | `has_email` counts only `email`-type identities |
| `/user/add-identity` with the user's own address → `409 IDENTITY_TAKEN` | `is_identity_taken(normalized)` matches the user's **own** username identity |
| SSO sign-in does not fix it | `_complete_sso_login` calls `find_user_by_identity(email)`, which matches the username identity, links SSO to that account, and never creates an `email` identity |

The user is left with an address they own, cannot reset with, and cannot register —
and each of the four surfaces reports something reassuring and wrong.

## Repro

1. Register with `type: "username"` where the value is a valid email address.
2. `POST /user/forgot-password` with that address → `200`, no mail sent.
3. Authenticate, then `POST /user/add-identity` with the same address → `409 IDENTITY_TAKEN`.
4. Sign in via Google SSO with that same address → links to the existing account,
   still no `email`-type identity.

## Suggested fixes

**1. `_complete_sso_login` should upsert an `email` identity.** When SSO resolves an
existing user by email, the provider has just asserted a verified address. Linking the
SSO account while leaving zero `email` identities is what makes the state
self-perpetuating. Upserting here auto-heals every affected SSO user on next sign-in,
with no migration.

**2. `add-identity` should let an owner upgrade their own identity.** If
`is_identity_taken` matches an identity **on the requesting user's own account** with a
different `type`, that is not a conflict — it is the same person naming the same
address. Convert in place (or return a distinct error code) instead of `409`.

Either fix alone closes the trap for new users; both are needed to cover SSO and
password accounts. Existing rows still need a data migration, since a non-SSO affected
account has no trigger that would ever re-run this code.

**3. Optionally, reject email-shaped values for `type: "username"` at registration**, so
the population cannot grow.

## Note on the neutral-response design

The silent no-send in `forgot-password` is correct for anonymous reset — account
existence must not leak. The problem is not that branch; it is that a user in this state
has no *authenticated* surface that tells them either. Worth considering an
authenticated signal, where the user has already proven who they are.

kashmithnisakya:
Closes #9103.

## Problem

A user whose only identity is a `username` that happens to be an email address cannot attach that same address as an `email` identity (`409 IDENTITY_TAKEN`), so every email feature (`/user/forgot-password`, `/user/send-verification`) has nothing to act on. The client SDK wraps any bare string passed to `jacSignup` as `type: "username"`, so any app that feeds an email field to it produces exactly this account.

The same root cause also refuses renaming a username to the account's own email address, and `_replace_username` deleted the lookup row of a value still used by another identity on the same account (an account registered with one address as both username and email lost email login after a username rename).

## Root cause

`identity_taken(value)` and the `identity_lookups` insert in `_add_identity` / `_replace_username` answer "is this value held by anyone", including the caller. Every mutating caller needs "is it held by someone else". Cross-account uniqueness of values is intended and stays as it is; the email-side `type == 'email'` / `verified` gates are correct and unchanged.

## Fix

- `PgIdentityStorage._claim_lookup`: the lookup row is claimed with one atomic statement (`INSERT ... ON CONFLICT (value_normalized) DO UPDATE ... WHERE identity_lookups.user_id = EXCLUDED.user_id RETURNING user_id`). Absent or already owned by the caller: claimed. Owned by another account: `IdentityTakenError`, as before.
- `_add_identity` claims the row, then appends the identity with one conditional `UPDATE` that adds it only when no element with the same type and normalized value exists, so a same-account duplicate (re-adding your own username) is still a 409 and two concurrent adds serialize on the row.
- `_replace_username` claims the new value and frees an old lookup row only when no remaining identity on the account still uses it.
- The store is the only authority on "taken": the add-identity handler and `rename_username` no longer run a separate pre-check, they map `IdentityTakenError` to the same 409 / error dict as before. `UserManager.is_identity_taken` had no other caller and is removed; `identity_taken` is unchanged and still backs registration.
- `_replace_username` is one statement that locks the account row, rebuilds the identities list from it and frees old lookup rows from the same locked doc, so an identity added concurrently is not overwritten.
- `identity_lookups` and `sso_lookups` reference `identity_users` with `ON DELETE CASCADE`, so `delete_user` is one statement and a claim racing an account delete fails the key check instead of leaving an orphaned row. The key is added at boot (orphaned rows are deleted first); boot checks the catalog before taking any lock, so a pod start never waits on in-flight identity writes.

Affected accounts recover by attaching and verifying the email identity. The only data change is the boot-time cleanup of lookup rows whose account no longer exists.

## Tests

Test count in the touched files: **140 on main, 143 with this PR (+3)**.

| File | main | PR |
|---|---|---|
| `jac/tests/runtimelib/test_serve_client.jac` | 109 | 109 |
| `jac/jaclang/scale/tests/server/test_admin.jac` | 29 | 29 |
| `jac/jaclang/scale/tests/server/test_user_identity.jac` | 2 | 2 |
| `jac/jaclang/scale/tests/server/test_identity_ownership.jac` | 0 | 3 |

Extended, not added:

- `serve update username already exists` (in-process `JacTestClient`) is now the table-driven `serve identity values are refused only when another account holds them`. It covers renaming or attaching another account's value (refused), attaching the owner's own username value as an email (#9103, fails on main with 409), re-adding the same type (refused), and renaming away from and back to a value the email identity still holds, with email login checked in between.
- `admin delete user` re-creates the deleted username and logs in with it, which fails if the lookup rows are not cascaded.

New, each covering a path no existing test can reach (they need a second Postgres connection holding a lock or a database created with the old schema):

1. `a rename keeps an identity that lands while the rename waits for the row`: the locked single-statement rename.
2. `an account deleted while a claim is in flight leaves nothing behind`: the foreign-key check on a claim that races an uncommitted delete.
3. `an existing database gains the account foreign key without waiting for in-flight writes`: the boot-time upgrade (key added, orphans dropped) and the catalog check before locking. It fails on the previous DDL with "the boot waited for the open transaction". The only existing storage-upgrade test (`test_auth_cutover.jac`) covers the credential cutover, not the schema.

## Verification

- Locally: `test_serve_client.jac` 109 passed, `test_admin.jac` 29 passed, `test_identity_ownership.jac` 3 passed, `test_user_identity.jac` 2 passed.
- `test_oauth_session.jac`, `test_firebase_auth_sso.jac`, `test_sso.jac`: 27 passed.
- `jac fmt --check --lintfix` clean on every changed file.


jac/jaclang/server/identity/impl/identity_storage.pg.impl.jac
```diff
@@ -231,22 +248,21 @@ impl PgIdentityStorage._add_identity(user_id: str, identity: dict[str, any]) ->
...
+    } elif identity.get('value_normalized') {
+        for held in identities {
+            if held.get('type') == identity.get('type')
+                and held.get('value_normalized') == identity['value_normalized'] {
+                raise IdentityTakenError();
+            }
         }
-        raise;
+        self._claim_lookup(identity['value_normalized'], user_id);
     }
```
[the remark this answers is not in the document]
kashmithnisakya: good catch. the duplicate decision now lives in the UPDATE itself: the identity is appended only when no element with the same type and value_normalized exists, so two concurrent adds serialize on the row and exactly one gets IDENTITY_TAKEN. covered by the new database-side test in ec5ab14c6.

jac/jaclang/server/identity/impl/identity_storage.pg.impl.jac
```diff
@@ -334,22 +380,10 @@ impl PgIdentityStorage._replace_username(
...
-            """,
-            {'v': identity['value_normalized'], 'u': user_id}
-        );
-    } except Exception as e {
-        if _pg_unique_violation(e) {
-            raise IdentityTakenError();
-        }
-        raise;
-    }
+    still_used = {str(i.get('value_normalized') or '') for i in kept};
```
MusabMahmoodh: `_add_identity` now appends in one conditional `UPDATE`, but this path still reads `identities` at the top and writes `kept` back wholesale at the bottom, and nothing holds the row in between. An add-identity that lands between that `SELECT` and the final `UPDATE` gets overwritten, while the lookup row it claimed stays with this account, so the value is refused to every other account and missing from the owner's identities until they add it again. Building the new list inside the `UPDATE` from the row's current `doc`, the way `_append_identity` does, closes it.

MusabMahmoodh (commented):
The atomic claim is the right fix, and dropping both pre-checks so the store is the only authority on "taken" is exactly the single-source move this needed. The rename path is the one write that didn't get the same treatment (inline).

On `Closes #9103`: its fix 1 (SSO upserts an `email` identity) targets a lookup that no longer exists, since `complete_identity` (`scale/sso/oauth_session.jac:36`) resolves only by SSO link and never by email, so closing it with this half is fair. Worth one line on the issue saying so, though, because the production accounts ahzan found still have no `email` identity until each user attaches the address themselves, and nothing prompts them to.

jac/jaclang/server/identity/impl/identity_storage.pg.impl.jac
kashmithnisakya: the rename is now one statement: it locks the row, rebuilds the identities list from that row's doc and frees old lookup rows from the same locked doc, so nothing can land in between; covered by a new test that holds the row open while a rename runs, in 501d0bf18

jac/jaclang/server/identity/impl/identity_storage.pg.impl.jac
```diff
@@ -317,52 +363,64 @@ impl PgIdentityStorage._remove_sso_identity(user_id: str, platform: str) -> bool
...
-        }
-        raise;
-    }
-    for old in old_values {
-        if old and old != identity['value_normalized'] {
-            self.store.rows(
-                'DELETE FROM identity_lookups WHERE value_normalized = :v AND user_id = :u',
-                {'v': old, 'u': user_id}
-            );
+    self._claim_lookup(identity['value_normalized'], user_id);
```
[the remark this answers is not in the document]
kashmithnisakya: both lookup tables now carry a foreign key to identity_users with on delete cascade (added at boot, orphans cleaned first), so a claim can't outlive its account: delete_user is one statement, a claim in flight against a deleting account fails the key check and returns false, and a committed claim is cascaded away; three new tests cover the cascade, the in-flight delete and the upgrade of an existing database, in 58c68fd25

jac/jaclang/server/identity/impl/identity_storage.pg.impl.jac
```diff
@@ -23,21 +23,46 @@ glob _PG_IDENTITY_SCHEMA: list[str] = [
...
+        ) THEN
+            DELETE FROM {table} lk
+            WHERE NOT EXISTS (
+                SELECT 1 FROM identity_users u WHERE u.user_id = lk.user_id
+            );
+            ALTER TABLE {table}
+                ADD CONSTRAINT {table}_user_id_fkey
+                FOREIGN KEY (user_id) REFERENCES identity_users (user_id)
+                ON DELETE CASCADE;
+        END IF;
```
MusabMahmoodh, line 59: `ensure_indexes` runs from the constructor, so every boot takes a SHARE ROW EXCLUSIVE lock on both lookup tables, not only the boot that adds the key. It is short, but a waiting lock request blocks every write queued after it, so one open transaction on `identity_lookups` holds up every register, identity add and rename behind a pod start. Checking `pg_constraint` before locking, and again once the lock is held, keeps the steady-state boot lock-free and still lets two replicas upgrade together:

```suggestion
    BEGIN
        IF NOT EXISTS (
            SELECT 1 FROM pg_constraint
            WHERE conrelid = CAST('{table}' AS regclass)
              AND conname = '{table}_user_id_fkey'
        ) THEN
            LOCK TABLE {table} IN SHARE ROW EXCLUSIVE MODE;
            IF NOT EXISTS (
                SELECT 1 FROM pg_constraint
                WHERE conrelid = CAST('{table}' AS regclass)
                  AND conname = '{table}_user_id_fkey'
            ) THEN
                DELETE FROM {table} lk
                WHERE NOT EXISTS (
                    SELECT 1 FROM identity_users u WHERE u.user_id = lk.user_id
                );
                ALTER TABLE {table}
                    ADD CONSTRAINT {table}_user_id_fkey
                    FOREIGN KEY (user_id) REFERENCES identity_users (user_id)
                    ON DELETE CASCADE;
            END IF;
        END IF;
```

MusabMahmoodh (approved):
Both are fixed where they belong. Rebuilding the list inside the locked statement closes the rename race, and holding the row open on a side connection means the new test fails on the old read-modify-write instead of only exercising the new path. The cascading key is a better answer to the delete race than releasing the claim on failure would have been, because it also covers an add that is in flight when the account goes. One suggestion inline on the boot-time lock, not blocking.

Every job on `58c68fd25` was cancelled at 09:07Z yesterday, so this head has no CI result yet and needs a push to run.

jac/jaclang/server/identity/impl/identity_storage.pg.impl.jac
kashmithnisakya: applied the double check in 9438e6960. turned out the boot still waited on `CREATE INDEX IF NOT EXISTS`, which takes SHARE on the table before it looks for the index, so that statement checks the catalog first now too; the new test boots a second storage while a side transaction holds an open insert on both lookup tables, and it fails on the old ddl. also rebased onto main, the conflict was #8080's annotations on lines this pr removes.

MusabMahmoodh (approved):
The catalog check landed and it is tested better than I asked for: holding an open transaction over `identity_lookups` and `sso_lookups`, then booting `PgIdentityStorage` on a thread and asserting it returns, is exactly the boot-blocks-behind-in-flight-writes shape and it would fail on the version that ran the DDL unconditionally.

Re-read the rewritten identity-store path as well - the store being the only authority on taken values reads cleaner than the split it replaced, and deciding same-account duplicates inside the database removes the read-then-write window rather than narrowing it.

Still approved.

MusabMahmoodh (approved):
Folding the ownership cases into the existing rename and admin-delete tests is the right move, and the one table-driven test in `test_serve_client.jac` walks every refuse/allow case through the real endpoints on the Pg store. The source is unchanged from what I approved. Still approved.

Issue #8017, opened 2026-08-10
Client codegen: imported JS/TS classes are called without `new` (TypeError at runtime)


kugesan1105:
## Summary

In client codespace, calling an imported JavaScript/TypeScript **class** emits a plain call expression instead of a `new` expression. JS class constructors throw when invoked without `new`, so any Jac code that constructs an npm-provided class fails at runtime with:

```
TypeError: Class constructor Foo cannot be invoked without 'new'
```

Since JS-style `new Foo(...)` is rejected in Jac source by design (parser emits `E0012`, commit `e8044b810`), plain call syntax is the only spelling available — so the compiler has to resolve construction itself. It does this correctly for Jac archetypes, but not for foreign classes.

Found while evaluating writing the VS Code extension in Jac (jaseci-labs/jac-vscode#103), where it blocks constructing `LanguageClient` from `vscode-languageclient`, `Registry` from `vscode-textmate`, and `vscode.Range` / `Position` / `WorkspaceEdit`.

## Reproduction

`mylib.d.ts`:

```ts
export declare class Greeter {
  constructor(name: string);
  hello(): string;
}
export declare function plain(x: string): string;
```

`mylib.js`:

```js
export class Greeter {
  constructor(name) { this.name = name; }
  hello() { return "hi " + this.name; }
}
export function plain(x) { return x; }
```

`app.jac`:

```jac
import from .mylib { Greeter, plain }

obj Local { has x: str = "a"; }

def:pub go -> str {
    j = Local();
    g = Greeter("world");
    p = plain("x");
    return g.hello() + p + j.x;
}
```

`jac.toml`:

```toml
[project]
name = "dtsprobe"
version = "0.1.0"
kind = "js-package"
entry-point = "app.jac"

[placement.pins]
"*" = "client"
```

Run `jac tool jac2js app.jac`.

### Actual

```js
let j = new Local();
let g = Greeter("world");     // <-- throws at runtime
let p = plain("x");
```

### Expected

```js
let j = new Local();
let g = new Greeter("world");
let p = plain("x");
```

Reproduced on `jac 0.36.0` (Linux x86_64).

## Root cause

`classify_call` in `jac/jaclang/compiler/symbol_utils.jac:82` returns `CallKind.INSTANTIATION` only when the resolved symbol is a Jac archetype:

```jac
if sym is not None and _is_user_symbol(sym) {
    if sym.sym_type
        in (
            SymbolType.OBJECT_ARCH,
[...]

PR #9126, opened 2026-09-11
Fix shared profile resolution and JavaScript primitive coercions

### LINKED ISSUE #8995: A profile's [apps.<name>.<section>] overlay is dropped, while [scale.gateway] in the same profile is applied

### LINKED ISSUE #9095: jac run <file> in a workspace costs 29s of whole-workspace analysis on 0.37.11, 0.44s on 0.37.5
## Summary

On 0.37.11, `jac run <file>` inside a workspace pays a whole-workspace analysis
before the target runs. A one-line script that imports nothing takes **29.1s**
where 0.37.5 takes **0.44s** on the same tree, same machine, same command.

The cost is not compilation of the target file. It is not caching either: a warm
`.jac/cache` does not help, and `--no-cache` does not change the number.

## Reproduction

```bash
git clone https://github.com/jaseci-labs/jacBuilder
cd jacBuilder && git checkout migrate/jac-0.37.11
echo 'with entry { print("hi"); }' > hello.jac
time jac run --no-serve hello.jac
```

The workspace is 831 `.jac` modules and 3 apps (`main`, `builder_sv`,
`jac_coder_sv`), one of which is client-capable.

## Measurements

Same file, same machine (macOS arm64), `with entry { print("hi"); }`:

| version | where | wall |
|---|---|---|
| 0.37.5 | in the workspace | 0.66s, 0.37s |
| 0.37.11 | in the workspace | 29.90s, 29.17s, 29.48s |
| 0.37.11 | a directory with no `jac.toml` | 0.27s |
| 0.37.11 | a small jac project | 0.27s |

Repeat runs do not get faster, so this is paid per invocation.

## What I ruled out

- **Caching.** Warm `.jac/cache` is the same as cold. `--no-cache` is the same
  as with cache. All three 0.37.11 numbers above are warm-cache runs.
- **node_modules.** Moving `.jac/client` (716 MB) aside: still 29.10s.
- **One heavy subtree.** Parking each of the three largest trees in turn moves
  the number only a little, which is what you would expect if the whole
  workspace is being walked rather than one hot spot:

  | parked | wall |
  |---|---|
  | nothing | 29.11s |
  | `apps/` (284 modules) | 22.70s |
  | `services/` (46 modules) | 24.98s |
  | `jac_coder/` (57 modules) | 28.09s |

- **Module count alone.** I generated a synthetic workspace of 200 trivial
  modules plus an `[apps.main]` table and could **not** reproduce it: 0.09s on
  0.37.11 and 0.09s on 0.37.5. So something else about a real workspace is the
  trigger, and a plain module count is not it. I did not narrow it further.

## Impact

Every CLI invocation in a workspace of this size pays 29s. In our case that
turned into CI failures rather than slowness: three tests shell out to
`jac run <script>` with 5s and 30s budgets that were comfortable on 0.37.5 and
now expire, and the suite's job cap had to go from 40 to 60 minutes.

`jac check <file>` looks similar (30-43s for a single file here), though I only
measured `jac run` carefully.

## Notes

I looked for a supported way to opt out before filing. `jac run --help` has no
flag that skips workspace analysis, and I did not find anything in the 0.37 docs
that presents this cost as intended for a target that imports nothing from the
workspace.

Environment: jac 0.37.5 and jac 0.37.11 release binaries, Darwin arm64.

### LINKED ISSUE #9096: Client codegen: Boolean(x) lowers to new Boolean(x), so every guard written with it is always truthy
## Summary

In client code, `Boolean(x)` lowers to **`new Boolean(x)`**. That is a Boolean
*object*, and every object is truthy in JavaScript, so `Boolean(false)` passes
an `if`. Every guard written with `Boolean(...)` silently reads true.

It is silent in every way that matters: nothing throws, the browser console is
clean, `jac check` passes and `jac build --as client` exits 0.

## The lowering

Source (jacBuilder `apps/web/utils/model_catalog.jac`):

```jac
def:pub modelIsByokOnly(modelId: str) -> bool {
    cat = modelCatalog();
    for `entry in cat {
        if `entry.value == modelId {
            return Boolean(`entry.byokOnly) as bool;
        }
    }
    return False;
}
```

Compiled (`.jac/client/main/dist/client.*.js`, minified):

```js
function s8e(t){let r=l2();for(const n of r)if(n.value===t)return new Boolean(n.byokOnly);return!1}
```

`new Boolean(false) === false` is `false` and `if (new Boolean(false))` runs the
branch. Expected lowering is the plain call, `Boolean(n.byokOnly)`.

## Reproduction

```bash
git clone https://github.com/jaseci-labs/jacBuilder
cd jacBuilder && git checkout f32f6ab7^   # the commit before the workaround
jac install
jac build main --as client
grep -c 'new Boolean(' .jac/client/main/dist/*.js
```

13 on 0.37.11. One of those is legitimate (a vendored deep-clone helper that
handles `instanceof Boolean`); the other 12 are lowered user code, from 16
`Boolean(` call sites in the project's client tree.

I could not reduce this to a standalone project: a minimal single-file
web-app hits an unrelated client-build failure in rollup before it emits a
bundle, and `--as source` on the same project writes only `configs/` (which may
be jac#9084). The jacBuilder branch above is a one-command reproduction.

## Impact

Every one of these read true regardless of its argument:

| call site | effect |
|---|---|
| `HARD_SKIP_DIRS` / `HARD_SKIP_FILES` / binary detection in a folder scan | every path segment "skip", every entry "excluded" |
| `Boolean(sub)` on a `PushSubscription` | the notifications toggle stuck ON |
| `Boolean(props.isListening)`, `Boolean(props.disabled)` | a mic button permanently listening and disabled |
| `Boolean(entry.byokOnly)` | every model in a picker rendered "ADD KEY", and a user's own allowed models rendered locked |

The last one is what surfaced it: a free-tier user saw all three models locked
behind a key prompt, and the two end-to-end tests that assert tier locks failed
in opposite directions, which is what a guard stuck on looks like from either
side of a boundary.

## Note on why `Boolean` and not `bool`

`bool(x)` is not interchangeable here. On a host object it lowers to
`Object.keys(x).length > 0`, and a WebIDL object keeps its attributes on the
prototype, so `bool(sub)` is always False for a live `PushSubscription`
(jacBuilder#1332). `Boolean(x)` was the documented way to get JS truthiness in
client code, which is why it is spread across the tree.

Working around it with `x and True or False`, which lowers to
`x && true || false`.

Environment: jac 0.37.11 release binary, Darwin arm64 and Linux x86_64.

marsninja:
Project readers independently selected and applied profiles, so CLI overrides, plugin reads, and deployment reads could disagree or retain stale settings. Centralize resolution and share the resulting configuration across those readers. Client calls to `Boolean`, `Number`, and `String` also constructed wrapper objects; for example, `Boolean(False)` became a truthy object.

- Resolve profiles through the project configuration layer and share its existing per-root cache across CLI, compiler, plugin, and deployment readers. Reuse the existing profile/app merge infrastructure, preserve local overrides, and replace cached configurations when explicit profiles change. Plugin refreshes update the shared cache without losing an explicit profile; standalone ambient scope remains separate from configuration needed for a specific file. Reuse the `active_profile_name` helper introduced by #9115, retain its deployment regression, and remove its now-redundant deployment-specific profile rediscovery.
- Model JavaScript conversions as callable values returning primitive types, preserving aliases and static helpers through the existing type evaluator. The backend's constructor logic stays unchanged.
- Use HTTPS/IPv4 for desktop CI package installation and fail on unsuccessful index refreshes; the runner's HTTP mirrors were repeatedly failing connections and stalling the smoke job.
- Validate native host methods through the existing type evaluator, preserving inheritance order and method signatures in cached interfaces instead of walking syntax trees. This fixes the desktop example type-check failure while preserving the existing strict host scalar annotations, including across cached interfaces.
- Detect the current repository layout when reading MCP examples, keeping local resources independent of GitHub availability.
- Guard explicit-file compilation against reading unrelated applications or tests. The workspace scan reported in #9095 was already removed by #9088; this PR adds a regression for that behavior.

Validation:

- The initial JavaScript regression failed in CI on the original code: https://github.com/jaseci-labs/jac/actions/runs/34597431095/job/103259954448.
- Five native-host contract regressions reproduce incorrect member lookup and cached-interface handling before the host fix, including the exact missing `CloseWindow` error from CI. All six current contract tests pass, covering scalar parameter and return constraints in both source and cached interfaces. The real WebGL host also passes a targeted check in the desktop app context.
- Deployment/profile regressions failed against v0.37.12. The workspace regression failed against v0.37.11 and passes with the current implementation.
- 184 targeted configuration/app/deployment/CLI tests pass. All three coercion tests pass with explicit type checking and Node runtime assertions. The existing imported JS/TS constructor and client dependency-injection regressions also pass.
- Normal compiler-cache validation passes all seven selected compiler tests. In total, 193 targeted tests pass, and staged formatting/type checks pass for all nine changed Jac files. The follow-up configuration, desktop/mobile/PWA, warm-cache, and offline MCP checks pass all 47 selected tests; formatting/type checks pass for the follow-up files too. No broad local test suite was run. The host-contract follow-up also passes staged formatting/type checks for both changed Jac files. After integrating upstream #9115, all 42 targeted profile, deployment, and plugin tests pass, including its additional deployment regression; all three staged Jac files pass formatting/type checks. Latest-commit CI is green: [25 successful jobs, including the final aggregate gate](https://github.com/jaseci-labs/jac/actions/runs/34615578601), at `8eca0fdd79ddf7e43c67eb75c82ada93d8e3f5c9`. This includes the full repository type check, compiler/native tests, Kubernetes integration, Android, desktop, and the development, production, and fleet browser journeys. The earlier isolated browser module-fetch failure did not recur in the fresh run.

Closes #8995.
Closes #9096.
Closes #9095.


MalithaPrabhashana (commented):
Validated this by running it, on macOS arm64. Verdict first: it fixes all three issues I filed, on the paths jacBuilder actually uses, and I found no regression from it. One question about the #9095 regression test at the end; it is not a blocker.

Setup: dev-mode checkouts of the head (8eca0fdd7), the base (ae2142c44, main with #9115), the tests-only commit (3a90dd298) and v0.37.11, each with its own HOME so the caches never mix, plus the 0.37.11 and 0.37.12 release binaries and a local Node 22.19.0.

## #9096 JavaScript coercions

Tests-first commit does what it says:

| commit | `tests/compiler/backends/es/test_js_coercions.jac` |
| --- | --- |
| 3a90dd298 (tests only) | 1 failed |
| 8eca0fdd7 (head) | 3 passed |
| base, with the head's copy of the file | 1 passed, 2 failed |

I also compiled the shapes my own code uses (a push subscription guard, a catalog flag, `if Boolean(x)`) and ran the emitted module in Node:

| probe | base | head |
| --- | --- | --- |
| emitted | `new Boolean(sub)`, `new Number(raw)`, `new String(raw)` | `Boolean(sub)`, `Number(raw)`, `String(raw)` |
| `guard(null)`, from `if Boolean(sub)` | `"on"` | `"off"` |
| `typeof notif_enabled(null)` | `"object"` | `"boolean"` |
| 12 assertions in Node | 12 fail | 12 pass |

No regression in the surrounding codegen: `test_js_generation.jac` 113 passed, 0 skipped (this includes "imported JS/TS class constructs with new", #8017), and `tests/client/test_config_dep_injection.jac` 4 passed.

## #8995 profile app overlays

Rendered with `jac scale deploy web --dry-run --show-yaml` on a four-file repro that carries all three overlay shapes: a scalar in a section the base has, a table the base lacks, and array-of-tables pod entries.

| scenario | 0.37.11 | 0.37.12 | base (main + #9115) | head |
| --- | --- | --- | --- | --- |
| `JAC_PROFILE=e2e`, `[apps.worker_sv.scale]` replicas/memory | dropped | dropped | dropped | applied (1, 1Gi) |
| same profile, nested `hpa` min/max plus a new `memory_target` | dropped | dropped | dropped | applied (1 to 3, base `cpu_target` kept) |
| same profile, `[apps.web.scale]` (table the base lacks) | dropped | dropped | dropped | applied (5Gi) |
| same profile, `deployment_overlay` initContainers/volumes/automount | dropped | dropped | dropped | appended after jac's own |
| `[environments.staging].apps.worker_sv.scale` | dropped | dropped | dropped | applied (2, 2Gi) |
| `jac.local.toml` with no profile | dropped | dropped | dropped | applied (4) |
| profile plus `jac.local.toml` | profile's gateway only | same | same | local wins on its key, profile keeps the rest |
| `[scale.gateway]` in the same profile (control) | applied | applied | applied | applied |

Worth noting for anyone tracking #8995: the base already carries #9115, and the CLI deploy path still dropped every app overlay there, which matches the note in #9115 that the CLI path had not been confirmed.

Your config tests: 42 passed on the head (8 + 15 + 19). Against the base, the behavioural one fails as it should ("the manifest fleet discovers the same active profile as the scale plugin (#8995)", on `builder._project_config() is scale.get_jac_config()`); the others fail only because `get_config(profile=...)` does not exist there, which pins the API rather than the behaviour.

## #9095 workspace scan

Confirmed, and confirmed that #9088 is what fixed it. One-line script inside the 1068-module jacBuilder tree, cold and warm, cache cleared before the first run of each:

| version | `jac run --no-serve hello.jac` | `jac check hello.jac` | peak RSS |
| --- | --- | --- | --- |
| 0.37.11 | 34.4s cold, 33.5s warm | 33.0s | 0.88 GB |
| 0.37.11, dotted entry points | 33.3s, 33.4s | | 0.88 GB |
| 0.37.12 | 2.2s, 0.33s | | 0.14 GB |
| base | 0.55s, 0.35s | | 0.13 GB |
| head | 0.58s, 0.35s | 0.89s | 0.13 GB |

The second row is there because the spellings differ between versions, so it rules the spelling out as the cause.

The question: the PR says "The workspace regression failed against v0.37.11", and I could not make it fail there.

| how I ran `tests/compiler/test_workspace_compile_scope.jac` | result |
| --- | --- |
| v0.37.11 source, dev mode | 1 passed |
| 0.37.11 release binary, standalone dir | 1 passed |
| 0.37.12 release binary | 1 passed |
| head | 1 passed |

The test resets the config with `set_config(None)` and drives `JacProgram.compile` directly, so I do not think it reaches the conditions the CLI set up on 0.37.11, and the `builtins.open` patch would also miss reads that go through `io.open` (pathlib) or the native reader. I tried to build a workspace shape that does reproduce the old scan and failed: 304 trivial modules across a web app, a service and a tests tree, with and without wildcard `[placement.pins]` and a `[client]` table, all stayed at about 0.5s on 0.37.11. My guess is the cost needs real import closures, which a synthetic tree does not have.

So: how did you see it fail on v0.37.11? If it only fails under a shape I have not reproduced, a note in the test saying so would help. Otherwise it may be worth asserting on the modules the program parsed, driven through the CLI, rather than on `open` calls.

## Merge and regression checks

- Trial merge with current main (9e37f2b7d, #9125): no conflicts, and the proving tests stay green on the merge (coercions 3, resolved config 8, app scale overrides 15, config loader 19, compile scope 1, js generation 113), with the #8995 matrix unchanged.
- Adjacent suites on the head: `tests/project` 326 passed 3 failed, `jaclang/scale/tests/apps` 33 passed 1 failed, `tests/cli/test_mcp.jac` 96 passed 6 failed, `tests/compiler/passes/test_wasm_contract.jac` 6 passed. The 10 failures reproduce identically on the base, so they are not from this PR; they look like local macOS artifacts (a temp-path comparison in scaffold_app, and nested `jac` subprocesses in the MCP snippet tests). CI on the head is green.

## Not covered

The desktop CI transport change, the MCP resources change beyond its own suite, anything on Linux, and a real cluster deploy. My #8995 evidence is `--dry-run --show-yaml` rendering, not an applied deployment.

One note rather than an ask: this carries four unrelated fixes alongside the three issues, which makes it harder to review and harder to revert a single piece later.

PR #5424, opened 2026-04-01
Introduce identity-based auth system with scalable registration

kashmithnisakya:
## Summary

Replaces the flat username/password user model with an identity-based architecture. Users now have multiple identities (username, email, SSO) and credentials (password), stored as arrays in a single MongoDB document.

- Registration and login use structured identity/credential payloads
- JWT tokens use `user_id` (UUID) instead of `username`
- SSO accounts stored as identities within the user document (removed separate `sso_accounts` collection)
- All auth endpoints extracted into `JacAPIServerAuth` mixin
- Shared auth helpers eliminate duplicated Bearer token extraction

## Register

**POST /user/register**
```json
{
  "identities": [
    { "type": "username", "value": "kashmith123" },
    { "type": "email", "value": "kashmithnisakya@gmail.com" }
  ],
  "credential": {
    "type": "password",
    "password": "secure_password"
  }
}
```

**Response (201)**
```json
{
  "ok": true,
  "data": {
    "user_id": "8483c3a6-d191-49a8-8b98-0b2214142581",
    "message": "User registered successfully"
  }
}
```
## Login
**POST /user/login**
```json
{
  "identity": {
    "type": "email",
    "value": "kashmithnisakya@gmail.com"
  },
  "credential": {
    "type": "password",
    "password": "secure_password"
  }
}
```

**Response (200)**
```json
{
  "ok": true,
  "data": {
    "user_id": "8483c3a6-d191-49a8-8b98-0b2214142581",
    "root_id": "e646ef2e91cb4180897764d7370b4932",
    "role": "user",
    "token": "eyJhbGciOiJIUzI1NiIs..."
  }
}
```

## MongoDB User Document
```json
{
  "user_id": "0dc51364-9592-4cc0-9dad-dc40e02a3faa",
  "status": "active",
  "identities": [
    {
      "type": "username",
      "value_raw": "kashmith123",
      "value_normalized": "kashmith123",
      "verified": true,
      "is_recovery": false
    },
    {
      "type": "email",
      "value_raw": "kashmithnisakya@gmail.com",
      "value_normalized": "kashmithnisakya@gmail.com",
      "verified": false,
      "is_recovery": true
    }
  ],
  "credentials": [
    {
      "type": "password",
      "password_hash": "sha256..."
    }
  ],
  "root_id": "58a72849a99c40adbda3ede74323bf39",
  "role": "user",
  "created_at": "2026-04-01T11:09:40.261848+00:00",
[...]

PR #9163, opened 2026-09-13
fix(scale): the /graph Login button signs in with the identity/credential body

### LINKED ISSUE #8951: Graph visualizer login is broken: graph.html still posts {username, password} to /user/login, which has required {identity, credential} since #7969
### Summary

The **Login** button on `/graph` cannot sign anyone in. `graph.html` still posts the pre-#7969 login body (`{username, password}`), both servers reject it before the credential is ever checked, and the modal reports "Invalid credentials" for every attempt, including a correct one.

Verified live on `main` @ `c08bef250` (2026-09-04), single-process `jac run --serve --no-client`, SQLite.

### What the page sends

`jac/jaclang/scale/templates/graph.html:495-500`:

```js
const resp = await fetch(baseUrl + "/user/login", {
    method: "POST",
    headers: { "Content-Type": "application/json" },
    body: JSON.stringify({ username, password }),
});
```

### What both servers require

Core stdlib server, `jac/jaclang/server/auth_types.jac:55-69`:

```jac
if 'identity' not in data   { raise ValueError('identity is required'); }
if 'credential' not in data { raise ValueError('credential is required'); }
```

Scale server, `jac/jaclang/scale/identity/auth_models.jac:123-137`, same checks. The scale endpoint declares `request_model=LoginRequest` (`serve.auth.impl.jac:87-93`), and the shared HTTP stack turns the `ValueError` into a 422 (`jac/jaclang/server/serving/app.jac:254-267`). The core `_login` handler returns a 400 with a plain string `error` (`server.impl.jac:2251-2268`).

### Repro

```bash
# what graph.html sends
curl -s -w '\nstatus=%{http_code}\n' -X POST localhost:8765/user/login \
  -H 'Content-Type: application/json' \
  -d '{"username":"alice","password":"pw12345678"}'
# {"detail": [{"type": "invalid", "loc": ["body"], "msg": "identity is required"}]}
# status=422

# what the server needs (same user, same password)
curl -s -X POST localhost:8765/user/login -H 'Content-Type: application/json' \
  -d '{"identity":{"type":"username","value":"alice"},"credential":{"type":"password","password":"pw12345678"}}'
# {"ok": true, "type": "response", "data": {"user_id": "...", "root_id": "...", "role": "user", "token": "eyJ..."}, "error": null}
```

The 422 is returned whether or not the user exists, so the modal fails identically for a typo and for a valid account.

### Why the modal says "Invalid credentials"

`graph.html:501-510` checks `result.ok && result.data.token`, otherwise shows `result.error.message`. A 422 body has `detail` (scale) and the core 400 has a string `error`, so neither branch matches and the hard-coded fallback text is shown. The real reason ("identity is required") never reaches the user.

### History

The template has not changed since it was folded into core in #7038 (`d4f74f0109`). The request shape changed in #7969 (Jac-native serving, one HTTP stack). No test posts the template's literal payload: `jac/jaclang/scale/tests/server/test_serve.jac:870-893` and `jac/tests/runtimelib/test_serve_client.jac:2353-2372` only hit `/graph/data` directly, using a token obtained through the tests' own `login_payload()` helper, so the page and `LoginRequest` were free to drift.

### Suggested fix

1. `doLogin()` posts the current shape: `{identity: {type: "username", value}, credential: {type: "password", password}}`.
2. Surface the server's reason: read `detail[0].msg`, a string `error`, or `error.message`, whichever is present, instead of always printing "Invalid credentials".
3. A test that drives the template's login through `JacTestClient` (extract the payload literal from `graph.html`, or move the payload construction into a snippet the test shares), so the page cannot drift from `LoginRequest` again.

This should land before or together with the `/graph/data` gating in #8755: once anonymous access is removed there, this modal becomes the only way into the visualizer. The gating and ACL side is #8952.

MalithaPrabhashana:
Fixes #8951.

## What was broken

The **Login** button on `/graph` could not sign anyone in, with any password.

- `graph.html` posted `{username, password}` to `/user/login`.
- `/user/login` has required `{identity, credential}` since #5424, so the server rejected every attempt with a 422 before it looked at the password.
- The modal only read `error.message`, which a 422 body does not have, so it printed its hard-coded "Invalid credentials" and hid the real reason (`identity is required`).

How long: the page shipped in #4697 and the login shape changed in #5424. Every release since **v0.30.0** has shipped a Login button that cannot log in. This is not a #7969 regression; the scale `LoginRequest` already required `identity` when the template was folded into core (#7038).

Why nothing caught it: no test ran the page's own login code. The graph tests fetched `/graph/data` with a token from the tests' own helper, so the page and `LoginRequest` drifted apart silently.

### Before

```mermaid
sequenceDiagram
    actor U as User
    participant P as graph.html
    participant S as POST /user/login
    U->>P: username + correct password
    P->>S: {username, password}
    S-->>P: 422 {detail: [{msg: "identity is required"}]}
    Note over P: reads only result.error.message
    P-->>U: "Invalid credentials", no token, stays on Public Graph
```

### After

```mermaid
sequenceDiagram
    actor U as User
    participant P as graph.html
    participant S as POST /user/login
    participant G as GET /graph/data
    U->>P: username + password
    P->>S: {identity: {type: "username", value}, credential: {type: "password", password}}
    alt credentials accepted
        S-->>P: 200 {ok: true, data: {token}}
        P->>G: Authorization: Bearer token
        P-->>U: User Graph
    else rejected
        S-->>P: 401 / 422 / 500 / non-JSON
        P-->>U: the server's reason, or "Sign in failed (HTTP n)"
    end
```

## The fix

One file, `jac/jaclang/scale/templates/graph.html`:

1. `doLogin()` posts the `LoginRequest` body.
2. `errorMessage()` reads the reason from each error envelope the server returns. A body that is not JSON no longer throws a misleading "Connection failed".

What the modal shows, per server response (the real page, driven in jsdom):

| server response | before | after |
| --- | --- | --- |
| 401 `{ok: false, error: {message}}` | Invalid credentials | Invalid credentials |
| 422 `{detail: [{msg}]}` | Invalid credentials | identity is required |
| 404 `{detail: "..."}` | Invalid credentials | Not Found |
| 500 `{error: "..."}` | Invalid credentials | RuntimeError: boom |
| 502 HTML from a proxy | Connection failed: Failed to parse JSON | Sign in failed (HTTP 502) |
| 500 `null` | Connection failed: null is not an object | Sign in failed (HTTP 500) |
| network failure | Connection failed: fetch failed | Connection failed: fetch failed |

## The test

Extended the existing `stdlib server serves graph visualizer and docs` in `jac/tests/runtimelib/test_serve_client.jac`. No new test: 108 before, 108 after.

The bug lives in the page's JavaScript, so a test that posts a payload from Jac would pass while the button stayed broken. The test now serves the app on a loopback port and clicks **Sign In** on the real page in jsdom, one row per outcome:

| row | what it proves |
| --- | --- |
| valid credentials | token stored, badge shows User Graph, `/graph/data` is fetched with the token |
| wrong password | no token, the modal shows the server's 401 message |
| rejected body | the body is replaced on the wire with the old shape, and the modal shows `identity is required` |

It reuses `bun_install` and `run_harness` from `react_jsdom_harness.jac` (same jsdom pin as the React suites) and runs in `test-runtime`, where bun is installed and `JAC_TEST_STRICT=1` turns a missing bun into a failure, not a skip.

Each half of the fix is pinned on its own:

| page under test | valid credentials | wrong password | rejected body |
| --- | --- | --- | --- |
| `main` | FAIL | PASS | FAIL |
| fix with the old body put back | FAIL | FAIL | PASS |
| fix with the old error text put back | PASS | PASS | FAIL |
| this PR | PASS | PASS | PASS |

Full verification (live server runs, the red/green logs, provenance) is in the comment below.


MusabMahmoodh (approved):
The body now matches `LoginRequest`, and `errorMessage` covers each envelope the two servers answer with. Extending the existing graph test with the real page in jsdom, rather than adding a file, is the right home, and its valid-credentials row fails against the old `{username, password}` body. The admin portal's `AuthContext` still posts that flat shape, but to `/admin/login`, which declares exactly those two parameters, so nothing else needs the same change. Approving.

kashmithnisakya (approved):

Issue #9036, opened 2026-09-07
docs: scheduler page promises log lines, an overlap guarantee, and a root model the code does not have


kashmithnisakya:
Part of #9024.

`jaclang/cli/docs/reference/plugins/jac-scale-scheduler.md` describes behaviour the code does not have.

1. **Startup log lines.** The quick start shows `Initializing scheduler subsystem` / `Base Scheduler started with 1 static task(s)` / `Scheduler subsystem ready`. The server prints `static scheduler: N task(s) registered` and `APScheduler started (timezone=UTC)`.
2. **"A job never overlaps itself (max_instances=1)."** True for dynamic jobs only; static tasks overlap without limit (see the overlap issue in this epic).
3. **Where static tasks write.** Static tasks run as the system user on the super root. A walker that attaches nodes to `here` under a `Root entry` writes to a root no HTTP caller can reach: after 144 scheduled ticks, `POST /function/count_ticks` from a client saw `{"ticks": 0}`. The page says only "spawned on a graph root". Say where, and what that means for apps that expect to read the result.
4. **Unsatisfiable cron stops the server.** Nothing says that a cron expression with no match in the next year is a boot failure.
5. `jac run --serve main.jac` is the command; the page is right about that, but the note that a plain `jac run` "executes your entry point once and exits without firing any schedules" is the only hint the user gets, and it is not shown at runtime.

Issue #9035, opened 2026-09-07
compiler: native backend segfaults on a module with a @schedule-decorated def


kashmithnisakya:
Part of #9024.

Running a module that contains a `@schedule`-decorated function under the default backend crashes the process. The Python backend runs the same file fine. Not a scheduler bug, but it is the first thing anyone hits who runs such a module with `jac run`.

## How to reproduce

```jac
import datetime;
@schedule(interval=1)
def s -> None { print("s"); }
with entry { print("hi with schedule", flush=True); }
```

```bash
jac run sched_only.jac            # default backend
jac run -b native sched_only.jac  # same
jac run -b python sched_only.jac  # works: prints "hi with schedule" and exits
```

## Evidence

```
⚠ native seam -- demoting s to Python-only (un-lowerable): .../sched_only.jac, line 2, col 2: Native pathway does not yet support decorator 'on s'
Segmentation fault at address 0x0
aborting due to recursive panic
```

Exit code 134. Reproduced three times on `main` at `c89fd6b0e`, macOS arm64.

kashmithnisakya:
Fixed somewhere between 0.37.7 and 0.37.10; does not reproduce on 0.37.14. Verified by running this issue's own repro byte for byte on both binaries:

- `jac-0.37.7 run sched_only.jac` → seam warning, then **exit 139 (SIGSEGV)**. The report was accurate.
- `jac 0.37.14 run sched_only.jac` → seam warning, `hi with schedule`, **exit 0**, three runs in a row. Same with `-b native` and `-b python`. A variant that actually calls the decorated function also exits 0, printing the `E5092: Native lowering failed for function 's'` note; the demoted callable's `abort()` stub is never reached.

What handles it now: `capability_check_pass` classifies a call-form decorator on a module-level def as a demotable violation rather than a fatal one, `NaIRGenPass._predemote_capability_unit` demotes the single callable, and `na_coverage.na_demotion_note` carries the module out of the native lane ("preferred native but did not lower; compiled in the server codespace"). Those files are byte-identical between the 0.37.10 tag and current `main`, so the fix landed below that tag.

One real gap remains, which is a missing feature rather than a crash: `_apply_static_decorators` only handles bare-`Name` decorators, so every parameterised decorator on a module-level `def` costs a seam. If `@schedule` on a native-eligible module should be supported rather than demoted, that is the place, and it deserves its own issue rather than keeping this one open.

[...]

Issue #9034, opened 2026-09-07
scheduler: --dev reload never registers a task added while serving, and plain jac run ignores schedules silently (#7619)


kashmithnisakya:
Part of #9024.

Two developer-experience gaps, both previously reported on #7619 (closed 2026-09-04 without a merged fix):

1. **`--dev` reload never registers a task added while serving.** Registration happens once in `_prepare_to_serve` (`server.impl.jac:1810`). A task removed on reload is looked up by name at fire time and logs `'<name>' is no longer registered` every tick (`:1712-1713`).
2. **Plain `jac run` ignores `@schedule` silently.** The entry block runs, the process exits, and nothing tells the user their schedule needs `--serve`.

## How to reproduce

```bash
# 1
jac run --serve --no-client --dev main.jac
# after 10 s append to main.jac:
#   @schedule(interval=2) def dev_b -> None { print("[dev_b] NEW TASK fired", flush=True); }
# wait 20 s: dev_b never fires, dev_a keeps firing

# 2
jac run -b python sched_only.jac      # prints the entry block, exits, no warning
```

## Evidence

```
INFO static scheduler: 1 task(s) registered
[dev_a] 11:22:25
[dev_a] 11:22:27
...                                    <- file changed at 11:22:33
[dev_a] 11:22:47
dev_b fired: 0
```

## Proposed fix

Re-run `_register_scheduled_tasks` on a reload (diffing by name so existing tasks keep their phase), and print a one-line notice when `jac run` executes a module that declares schedules without serving it.

## Environment

- `main` at `c89fd6b0e` (jaclang 0.37.7, after #8981), dev-mode checkout, macOS arm64, embedded Postgres 18
- server started with `jac run --serve --no-client [-w 3] [-p PORT] main.jac` (there is no `jac serve`)
- fixture project: a `jac.toml` with `[project] name = "schedbench"`, `entry-point = "main.jac"` and `[scale.scheduler] enabled = true`

Issue #9033, opened 2026-09-07
scheduler: failures log one line with no traceback, healthy ticks log nothing, static tasks have no metrics or listing


kashmithnisakya:
Part of #9024.

A failing scheduled task produces one log line with no traceback; a succeeding one produces nothing at all. There are no metrics for static tasks, no run history, and no way to list them. In practice the only way to know which worker ran a tick is for the task to print its own pid, which is what the test fixture had to do.

## Evidence

```
{"level": "ERROR", "service": "unknown", "msg": "scheduled boom_fn failed: boom from scheduled function"}
```

That is the entire record for `@schedule(interval=2) def boom_fn { raise RuntimeError(...) }`. The walker path computes a traceback with `include_traceback=True` and then discards it:

```
# jaclang/server/impl/server.impl.jac:1735-1738
outcome = asyncio.run(_run());
if isinstance(outcome, dict) and 'error' in outcome {
    logger.error(f"scheduled {name} failed: {outcome.get('error')}");
}
```

- the "fired elsewhere this tick; skipping" message is `DEBUG` (`task_scheduler.impl.jac:134`), so a three-worker server looks like it fires once and nobody can tell which worker did it
- structured lines carry `"service": "unknown"` on a local serve
- dynamic jobs have `run_count` / `last_status` / `last_error`; static tasks have nothing, and the admin portal does not list them

## Proposed

- one `INFO` line per fire: task, pid, duration, outcome, and whether the lease was won or skipped
- the traceback in the server log on failure (consistent with #8904)
- counters in the existing Prometheus registry: fires, skips, failures, duration histogram, per task
- a read-only listing of static tasks (`next_run`, `last_run`, last outcome) next to `/jobs` or under `/admin`

## Environment

- `main` at `c89fd6b0e` (jaclang 0.37.7, after #8981), dev-mode checkout, macOS arm64, embedded Postgres 18
- server started with `jac run --serve --no-client [-w 3] [-p PORT] main.jac` (there is no `jac serve`)
- fixture project: a `jac.toml` with `[project] name = "schedbench"`, `entry-point = "main.jac"` and `[scale.scheduler] enabled = true`

PR #9192, opened 2026-09-14
fix(scale): let /jobs resolve a scheduled target, and bound what it creates

### LINKED ISSUE #9025: scheduler: POST /jobs answers 404 for every DYNAMIC target, so no dynamic job can be created
Part of #9024.

`POST /jobs` cannot create a job for any target. Every request answers `404 NOT_FOUND "Walker or function \"X\" not found"`, including targets decorated exactly as the docs prescribe with `@schedule(trigger=ScheduleTrigger.DYNAMIC)`. The dynamic-jobs feature is unreachable from the API; the only jobs that run are rows already present in the `scheduled_jobs` collection.

## Root cause

`JobsExtension.create_job` resolves the name through the API-facing registries:

```
# jaclang/scale/server/jobs_ext.jac:120-121
target: any = server.get_walkers().get(name) or server.get_functions().get(name);
```

Both delegate to `ModuleIntrospector.get_walkers` / `get_functions`, which filter out anything carrying a `schedule_spec` so that scheduled targets do not become HTTP endpoints:

```
# jaclang/server/impl/server.impl.jac:982-993
return {
    k: v
    for (k, v) in self._walkers.items()
    if not v?.schedule_spec
};
```

A `DYNAMIC` target carries a `schedule_spec` by definition, so it is never found, and the "no `@schedule` decorator" check that follows is unreachable. `_execute_task` (`scale/scheduler/impl/scheduler.impl.jac:610-615`) already reads the raw `introspector._walkers` / `_functions`, which is why rows restored from the database still run.

## How to reproduce

```jac
@schedule(trigger=ScheduleTrigger.DYNAMIC)
walker DynWalker {
    can go with Root entry { print("DynWalker fired"); }
}
```

```bash
jac run --serve --no-client main.jac
# register + login as in the scheduler docs, then:
curl -s -X POST localhost:8000/jobs -H "Authorization: Bearer $TOKEN" -H 'Content-Type: application/json' \
  -d '{"walker_or_function":"DynWalker","trigger":"interval","interval":2}'
```

## Evidence

```
POST /jobs {"walker_or_function":"DynWalker","trigger":"interval","interval":2}
404 {"ok": false, "type": "error", "data": null, "error": {"code": "NOT_FOUND", "message": "Walker or function \"DynWalker\" not found"}}
```

Same answer for `trigger=cron`, `trigger=date`, and for a `STATIC` target (`beat_fn`). Inserting the same job as a row directly into `jac_docs` (`collection='scheduled_jobs'`) and restarting the server makes it fire every 2 s and record `run_count`, which confirms the rest of the pipeline works and only the lookup is wrong.

## Why no test caught it

`scale/tests/server/test_scheduler_endpoints.jac` only exercises `_err_response`; `test_scheduler_multipod.jac` builds `JacScaleScheduler` objects and inserts rows through `PgJobStore` directly. Nothing calls `create_job` through the endpoint.

## Proposed fix

Resolve the name from the raw registries (`server.introspector._walkers` / `_functions`, or a new `get_schedulable()` helper), keep the `DYNAMIC` check, and add one API test that creates a job and observes a fire.

## Environment

### LINKED ISSUE #9032: scheduler: any authenticated user can schedule a job every millisecond with no per-user cap
Part of #9024.

`TriggerSpec.validate` (`jaclang/scale/scheduler/impl/scheduler.impl.jac:15-21`) only requires `interval > 0`, and `/jobs` has no per-user job limit. Any authenticated user can create as many jobs as they like at `interval=0.001`, each running a walker under their own root, against a fixed `thread_pool_size` (default 10) shared by every job on the pod.

Today this is masked by #9025 (the API cannot create jobs at all); once that is fixed the abuse path is open.

## Proposed fix

- a minimum interval (1 s, configurable under `[scale.scheduler]`)
- a per-user active-job cap, admin-exempt
- reject a `date` in the past at request time with 400 instead of scheduling a misfire

## Related

Filed separately from the decorator-validation issue because this is a hardening item on an authenticated endpoint, not a developer nicety.

## Environment

kashmithnisakya:
Closes #9025. Closes #9032. Part of #9024.

## #9025 — `/jobs` could not create a single job

`create_job` resolved its target through `get_walkers()` / `get_functions()`, and both accessors filter out everything carrying a `schedule_spec` so scheduled targets do not become HTTP endpoints. That is the exact set the endpoint requires, so the DYNAMIC check below the lookup was unreachable and the feature had no working path at all.

Two states of one walker, same request, which pins it without reading any source:

| walker | `POST /jobs` |
| --- | --- |
| with `@schedule(trigger=ScheduleTrigger.DYNAMIC)` | `404 Walker or function "ReconcileGithub" not found` |
| decorator removed | `400 "ReconcileGithub" has no @schedule decorator. Add @schedule(trigger=ScheduleTrigger.DYNAMIC)...` |

The endpoint demanded the decorator that made its own lookup fail.

**The fix.** Name resolution goes through a new `ModuleIntrospector.find_target(name)`, which reads the loaded registry, with a `JacAPIServer` delegate beside `get_walkers` / `get_functions`. The three checks that follow (found, decorated, DYNAMIC) now do the validation they were written for, including the `400` for an undecorated target.

**Why not widen `get_walkers` instead.** That is the tidier design: the accessors encode a routing policy, the route builders already carry the same guard at `endpoints/walkers.jac:216` and `:365`, and those guards are dead code today. But widening the accessors changes what every consumer sees, including `restspec_routes`, `walker_protocol`, `/introspect`, `colocated`, `websocket_ext` and `webhook_ext`, and a missed filter there silently exposes a scheduled walker as an HTTP endpoint. That is a security-relevant refactor and does not belong in the same change as a 404 fix. `find_target` gives the one consumer that needs a name resolver a name resolver, and leaves the HTTP-surface contract exactly as it is. The broader cleanup is worth doing separately.

## #9032 — and the guard rails that have to ship with it

Fixing the lookup alone opens the hole #9032 describes: the only interval rule was `> 0`, nothing counted a user's jobs, and the thread pool is pod-wide and shared by every tenant. So the first account through the newly opened door could hold unlimited millisecond jobs.

- **`[scale.scheduler] min_interval`**, default `1.0`. Enforced in `TriggerSpec.validate`, the single choke point `POST` and `PUT` both reach through `_build_spec`, so both verbs answer `400` below the floor. Enforced **again** in `_schedule_apscheduler_job`, the one sink create, update, restore and the cross-pod reconcile all funnel into, because the restore path builds specs through `from_dict` and validates nothing: a row written straight into the store would otherwise keep firing at whatever it asked for. There it clamps and warns rather than raising, so one bad legacy row cannot fail job restoration at boot.
- **`max_jobs_per_user`**, default `25`, admins exempt, `0` for unlimited. Answers `429 QUOTA_EXCEEDED`. Two halves:
  - **Who is capped** is decided in `jobs_ext.create_job`, because it is an authorization rule: that is where `_is_admin` lives. The restore path passes no cap, so a restart never drops already-accepted jobs.
  - **Whether the cap is reached** is decided in the job store, where the row is written: `JobStore.insert_within_quota(job, created_by, max_active)`. `PgJobStore` counts the creator's active rows and inserts inside one `READ COMMITTED` transaction serialised per creator by `pg_advisory_xact_lock`. That transaction runs on a `PgStore` connection the job store opens for quota writes alone, serialised by its own lock: the process-wide shared store takes its lock one statement at a time, so another thread's statement could otherwise join the transaction, or roll it back and release the advisory lock mid-count. `MemoryJobStore` does the same under a process lock. A count followed by a separate insert would let concurrent requests race past the cap, and a storage failure during the count now reaches `_err_response` (`503`) like any other store error.

Both keys are in the config loader, the schema and the docs table. A value the runtime cannot honour (not finite, not above zero for the floor, not a whole number of zero or more for the cap) is reported by `validate_scale_config` at boot and replaced by the default. The loader and the validator read both keys through the same two functions in `scale.config.validation`, so they cannot disagree; in particular a fractional cap is rejected rather than truncated, since `int(0.5)` is the `0` that means unlimited.

## Tests

- `tests/runtimelib/test_scheduling.jac` already asserted `len(get_walkers()) == 0` over a fixture of six scheduled targets, which is the behaviour that caused this. The complement now sits beside it: `find_target` resolves each of those six by name, and `None` for an unknown one. Without the fix the new assertions fail and the old ones still pass, which is the point.
- `scale/tests/server/test_scheduler_endpoints.jac`: an interval below the floor is refused with a message naming the key that raises it, one at the floor is accepted, and zero is still refused by the original rule. The store refuses an insert at the cap, lets it through with `max_active=0`, and counts only the creator's own active rows. And through the served app: `POST /jobs` for `DynamicScheduledWalker` answers 201, then 429 `QUOTA_EXCEEDED` at a cap of 1 (with the handler switched back to the `get_walkers` lookup it fails with `NOT_FOUND`).
- `scale/tests/server/test_pg_job_store.jac`, on embedded Postgres: three threads across two pods race a cap of 1 for eight rounds and exactly one wins each (without the advisory lock the first round admits two), and a quota write survives a rollback of an unrelated transaction left open on the shared store (the previous implementation lost it).
- `scale/tests/misc/test_config_validation.jac`: out-of-range bounds are reported, valid ones (including `25.0`) are quiet, and `0.5`, `25.7` and `true` are rejected as caps. Against the pre-fix validator, `max_jobs_per_user = 0.5` produced no finding at all.

Green locally against the source tree, rebased on `main` at `e255ec5c`: `test_scheduler_endpoints` 10 passed, `test_pg_job_store` 7 passed, `test_config_validation` 31 passed, `test_config_loader` 19 passed, `test_app_scale_overrides` 15 passed, `test_scheduling` 6 passed, `test_scheduler_multipod` 5 passed, `test_optional_deps` 24 passed, with APScheduler installed so nothing skips.

## Note for #9024

This is the first of the field-review findings. The remaining ones I validated against current `main` are all still live: #9026, #9027, #9028, #9029, #9030, #9031, #9033 (partly fixed, the traceback half landed in #8910), #9034 and #9036. #9035 no longer reproduces on 0.37.14 and I have commented there.


jac/jaclang/scale/server/jobs_ext.jac
```diff
@@ -143,6 +153,20 @@ obj JobsExtension(ServerExtension) {
                     400
                 );
             }
+            cap = _job_cap();
+            if cap > 0 and not _is_admin(str(user_id)) {
+                (_held, held_total) = JacScaleScheduler.instance().list_jobs(
+                    limit=1, offset=0, created_by=str(user_id)
+                );
```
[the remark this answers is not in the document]
kashmithnisakya: Valid, and fixed at the write rather than narrowed. The count and the insert were two operations with an HTTP round trip between them, so two concurrent creates for one account could both see room and both take it, and nothing downstream rechecked.

The quota is now a `JobStore` operation, `insert_within_quota(job, created_by, max_active)`. `PgJobStore` counts and inserts inside one `READ COMMITTED` transaction serialised per creator by `pg_advisory_xact_lock`: the lock is taken first, and because each statement then takes a fresh snapshot, a concurrent creator for the same account blocks instead of reading a stale count. It follows the `begin`/`commit`/`rollback` pattern already used in `scale/events/streams/postgres.jac`, and defers to an enclosing transaction when one is open. `MemoryJobStore` does the same under a process lock.

`create_job` takes `max_active` and answers `None` when the store refuses; the endpoint keeps what is policy, the admin exemption and the 429. Three tests in `test_scheduler_endpoints.jac` cover it: the job past the cap is refused and its row is not written, one account does not consume another's allowance, and 0 means unlimited.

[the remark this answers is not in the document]
kashmithnisakya: Valid. The quota lookup sat above the `try`, so a Postgres failure while counting escaped `_err_response` and surfaced as the framework's generic 500 instead of the `503 STORAGE_UNAVAILABLE` this file has a whole test module asserting.

The separate lookup is gone: the quota now travels into `create_job` as `max_active` and is decided inside the store write, which is already inside the `try`. So every storage failure on the create path, counting included, reaches `_err_response` again.

jac/jaclang/scale/config/impl/config_loader.impl.jac
```diff
@@ -421,7 +421,9 @@ impl JacScaleConfig.get_scheduler_config(self: JacScaleConfig) -> dict[str, any]
         'system_user_password': scheduler_config.get(
             'system_user_password', '__no_login__'
         ),
-        'user_exists_ttl': float(scheduler_config.get('user_exists_ttl', 30.0))
+        'user_exists_ttl': float(scheduler_config.get('user_exists_ttl', 30.0)),
+        'min_interval': float(scheduler_config.get('min_interval', 1.0)),
+        'max_jobs_per_user': int(scheduler_config.get('max_jobs_per_user', 25))
```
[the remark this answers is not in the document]
kashmithnisakya: Valid, and the sharpest of the three: a bad value here silently disables the control it configures. `min_interval = -1` turns the floor off, a negative `max_jobs_per_user` means unlimited when only `0` is documented that way, and a non-finite floor either disables validation or rejects every job.

Handled in two places, because reporting alone would still leave the control off:

- `get_scheduler_config` falls back to the documented default for any value it cannot honour, so the floor and the cap always hold. `min_interval` must be finite and above zero; `max_jobs_per_user` must be a non-negative integer.
- `validate_scale_config` reports the bad value at boot through the existing findings mechanism, next to `_declared_target_findings`, naming the key, the value and the default being used instead. Silence would have been the real problem.

Tests in `test_config_validation.jac` cover a negative floor, a negative cap, a non-numeric floor, and that valid bounds including `0` produce no findings.

jac/jaclang/scale/config/impl/config_loader.impl.jac
```diff
@@ -413,6 +413,26 @@ impl JacScaleConfig.get_telemetry_config(self: JacScaleConfig) -> dict[str, any]
...
+        }
+    } except Exception as e {
+        floor = 1.0;
+    }
+    cap = 25;
+    asked_cap: any = scheduler_config.get('max_jobs_per_user', 25);
+    try {
+        wanted = int(asked_cap);
+        if wanted >= 0 {
+            cap = wanted;
```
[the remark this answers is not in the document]
kashmithnisakya: Valid, and it is the same failure mode the cap was added to prevent: `int(0.5)` is `0`, `0` is the documented spelling of unlimited, so a typo in the cap lifted the cap. Validation agreed with it because it applied the same `int()` before checking the range, which is exactly the duplication you point at in the last line.

Fixed in e6ff5ff by deleting the duplicate. The rule now lives once, as `scheduler_min_interval` / `scheduler_job_cap` in `scale.config.validation`, and both the loader and the boot findings call it, so a value one rejects can no longer be the value the other quietly keeps. A cap must be a whole number of zero or more: `25` and `25.0` pass (nothing is lost converting them), `0.5`, `25.7`, `true` and `-1` are rejected, reported at boot, and replaced by the documented default of 25.

Evidence, against the pre-fix code:

    findings for max_jobs_per_user=0.5 -> []
    int(0.5) resolves to 0 and 0 means unlimited
    AssertionError: unreported: []

and after: 1 finding, cap 25. Covered by two new tests (31 passing in `test_config_validation.jac`, up from 29), plus the docs row and the release note.

I left the schema-level type enforcement alone on purpose. Making `validate_against_schema` enforce declared types would change the behaviour of every plugin config in the repo, which is more than this PR should carry. The per-key check gives the same protection for these two keys without that blast radius, and is worth a separate issue if you want it applied generally.

jac/jaclang/scale/scheduler/impl/scheduler.impl.jac
```diff
@@ -119,6 +138,59 @@ impl PgJobStore.insert(job: dict) -> str {
...
+) -> bool {
+    if max_active <= 0 {
+        self.insert(job);
+        return True;
+    }
+    store = self.store;
+    nested = bool(store.in_txn());
+    if not nested {
+        store.begin(TxnIsolation.READ_COMMITTED);
+    }
```
MusabMahmoodh: `self.store` is the process-wide `shared_store()`, which auth tokens and this scheduler's own `record_run` also use, and `PgStore._run` holds `_lock` for one statement at a time (`begin` too). A statement another thread runs between this `BEGIN` and a refusal's `rollback` lands in this transaction and is rolled back with it. A failing statement from another thread also runs `ROLLBACK` and clears `_in_txn` (store.impl.jac:425-429), which releases the advisory lock mid-count and leaves the insert autocommitted. Holding `store._lock` from `begin` to `commit` closes both (it is an RLock, and `ensure_schema` already does it), at the cost of the pod's other shared-store traffic waiting behind the advisory lock; a dedicated connection avoids that.

jac/jaclang/scale/tests/server/test_scheduler_endpoints.jac
```diff
@@ -34,3 +35,80 @@ test "generic errors still map to 500 scheduler error" {
...
+    }
+    assert "> 0" in caught , f"expected the positive-interval message, got {caught!r}";
+}
+
+
+test "the quota is decided where the row is written" {
+    # #9032 again: the cap used to be a count read before a separate insert,
+    # so two concurrent creates could both see room and both take it. The
+    # store decides now, so the check and the write cannot be separated.
+    store = MemoryJobStore();
```
MusabMahmoodh, line 83: The advisory lock and the READ COMMITTED transaction are what keep concurrent creates under the cap, and they exist only in `PgJobStore.insert_within_quota`; all three quota tests here use `MemoryJobStore`. `test_pg_job_store.jac` already runs `PgJobStore` on embedded Postgres with two connections, so two threads racing `insert_within_quota` at a cap of 1 there would test what the PR claims.

jac/jaclang/scale/server/jobs_ext.jac
```diff
@@ -117,8 +128,7 @@ obj JobsExtension(ServerExtension) {
             if spec is None {
                 return spec_err or _fail('INVALID_REQUEST', 'Invalid schedule', 400);
             }
-            target: any = server.get_walkers().get(name)
-                or server.get_functions().get(name);
+            target: any = server.find_target(name);
```
MusabMahmoodh, line 123: Nothing in the PR calls this handler: the new assertions exercise `find_target` directly, so putting the old `get_walkers().get(name)` lookup back here would keep every test green, and the new 429 below is untested too. That is the gap #9025 names. One test that POSTs `/jobs` for a DYNAMIC target against a served fixture, 201 and then 429 at a cap of 1, would pin both.

jac/jaclang/scale/scheduler/impl/scheduler.impl.jac
```diff
@@ -119,6 +138,59 @@ impl PgJobStore.insert(job: dict) -> str {
...
+        store.rows(
+            'INSERT INTO jac_docs (collection, id, doc) '
+            'VALUES (:c, :i, CAST(:d AS jsonb)) '
+            'ON CONFLICT (collection, id) DO UPDATE SET doc = EXCLUDED.doc',
+            {
+                'c': self._collection,
+                'i': str(d['job_id']),
+                'd': json.dumps(d, `default=str)
+            }
+        );
```
MusabMahmoodh: This is `PgJobStore.insert` line for line, and `insert` goes through the same `self.store.rows`, so calling it keeps the write inside this transaction:

```suggestion
        self.insert(job);
```

jac/jaclang/scale/server/jobs_ext.jac
```diff
@@ -143,10 +153,19 @@ obj JobsExtension(ServerExtension) {
                     400
                 );
             }
+            cap = 0 if _is_admin(str(user_id)) else _job_cap();
```
MusabMahmoodh: `_job_cap` turns any config error into 25 through a bare `except`, on a permission decision. `scheduler_cfg` from `register` already carries the normalised `max_jobs_per_user`, so reading it here lets `_job_cap` and its hardcoded 25 go.

MusabMahmoodh (changes requested):
The `find_target` split is the right fix for #9025, and leaving the HTTP filter in `get_walkers` alone keeps the scope honest. What I would hold this on is the quota transaction: it opens on the process-wide shared connection without holding it, and the three quota tests all run `MemoryJobStore`, so the Postgres path that actually enforces the cap has no test. Once #9201 lands, the count and the advisory-lock key also want the service scope, which its body already notes.

jac/jaclang/scale/scheduler/impl/scheduler.impl.jac
kashmithnisakya: Valid, and I confirmed both halves in `store.impl.jac`: `_run` takes `_lock` per statement, so a statement from another thread between `BEGIN` and `COMMIT` lands in this transaction, and its failure path (`ROLLBACK`, `_in_txn = False`) drops the advisory lock and leaves the insert autocommitted.

Fixed in 5c03600e with the dedicated connection rather than holding the shared lock. `PgJobStore` opens its own `PgStore` (same `conninfo` and `ensure_cb`, the way `PgEventStream` does) for quota writes only, and serialises those with a lock of its own. Nothing else runs on that connection, so no other thread can join or roll back the transaction, and the pod's shared-store traffic never waits behind the advisory lock. The "join the caller's transaction" branch is gone: a write on a separate connection cannot be part of one. The store closes the connection on scheduler stop and when `setup()` replaces it.

New test: `a quota write commits on its own, whatever the shared connection is doing` leaves a transaction open on the shared store, writes under the cap, rolls the shared transaction back, and checks the job is still there. On the previous implementation it fails with "the rollback took the job with it".

jac/jaclang/scale/tests/server/test_scheduler_endpoints.jac
kashmithnisakya: Valid. Added `concurrent creates for one account stay under the cap` to `test_pg_job_store.jac`: eight rounds, each with three threads behind a barrier, two on one pod's `PgJobStore` and one on a second pod's connection, all creating for the same account at a cap of 1. Each round must admit exactly one and leave one active row. With the `pg_advisory_xact_lock` statement removed it fails on the first round with `[True, True, False]`, so it tests the lock, not luck.

jac/jaclang/scale/server/jobs_ext.jac
kashmithnisakya: Valid, that was the real gap. Added `POST /jobs schedules a DYNAMIC target, then refuses the account past its cap` to `test_scheduler_endpoints.jac`. It boots `test_api.jac` through `ScaleTestClient` with `[scale.scheduler] max_jobs_per_user = 1` bound to the fixture dir, registers a user, POSTs `DynamicScheduledWalker` and expects 201, then POSTs again and expects 429 `QUOTA_EXCEEDED`. With the handler switched back to `server.get_walkers().get(name) or server.get_functions().get(name)` it fails with `NOT_FOUND ... "DynamicScheduledWalker" not found`.

jac/jaclang/scale/scheduler/impl/scheduler.impl.jac
kashmithnisakya: Valid about the duplication. The literal suggestion stops being right after the fix for your first comment, though: `insert` writes through `self.store`, the shared connection, so calling it here would put the insert outside the quota transaction. Both paths now go through one `_write(on, job)` helper: `insert` passes the shared store and `insert_within_quota` passes its own connection. So the SQL lives in one place and the write still commits with the count.

jac/jaclang/scale/server/jobs_ext.jac
kashmithnisakya: Valid. The loader already validates and normalises `max_jobs_per_user`, so the bare `except` could only hide a real config failure behind a hardcoded 25 on a permission decision. `_job_cap` is gone: `register` reads `scheduler_cfg["max_jobs_per_user"]` once, next to the other scheduler settings, and the handler uses that. `_min_interval()` in the scheduler had the same shape (a bare `except` returning 1.0), so it now reads `["min_interval"]` directly too.

MusabMahmoodh (approved):
The quota write is off the shared connection now and that closes what I blocked on. `_quota_connection` opens its own `PgStore` from the same conninfo, `_quota_lock` serialises both the lazy create and the transaction, and `begin`/`rollback` only ever touch that connection, so no other thread's statement can land inside the advisory-lock window and no caller's rollback can release it. `close` is wired on both stop and re-setup (scheduler.impl.jac), so a repeated setup does not leak a connection per cycle.

The tests are the right kind: three threads across two pods on embedded Postgres, a quota write surviving an unrelated rollback on the shared store, and POST /jobs answering 201 then 429 through the served app. Each one fails on the old implementation, which is what I could not say about the previous round.

Also checked `_min_interval` now that the bare except is gone - `get_scheduler_config` always writes the key (config_loader.impl.jac, `floor if floor is not None else 1.0`), so the direct subscript cannot KeyError at boot.

Issue #9184, opened 2026-09-14
jac.toml ${...}: parser gaps plus no consumer-side check still ship literal placeholders into deploys


kugesan1105:
### Context

#9045 makes an unresolved `${VAR}` or `${VAR:?message}` in `jac.toml` reject the
configuration instead of passing through as literal text, and leaves a `${...}` whose
contents are not a variable name verbatim, so a `[scripts]` shell positional such as
`${1}` no longer fails every command.

What it does not change is the shape of the parser, or the fact that nothing downstream
checks whether a value was actually resolved:

```
\$\{([^}:]+)(?:(:[-?])([^}]*))?\}
```

**These are the reason #8773 does not fully close with #9045.** Validated live on
`main` @ad9869e6c and on #9045 at head 77bb3e09d.

---

### 1. Nothing validates that a value is resolved where it is used

There are exactly two "is this resolved" checks in all of production:

```
jac/jaclang/scale/admin/service_urls.jac:17                           if expanded and "${" not in expanded
jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac:169   if "${" in svc_name
```

The second lives inside `k8s_safe_name`, which is an RFC 1123 **name sanitizer**. All of
its callers pass names, so `app_name` is checked only because it happens to be a name.
`namespace`, `bundle_storage_class`, every `[scale.secrets]` value, `image_registry`,
`service_account_name`, volume names and container env values have no check at all.

`manifest_builder.jac:356-361` shows it in one call:

```jac
selection = select_injector(
    self.k8s_config.bundle_storage_class,        // raw, unchecked
    k8s_safe_name(self.k8s_config.app_name),     // checked
    self.k8s_config.namespace,                   // raw, unchecked
    ...
```

Combined with gap 2 below, this reproduces #8773 exactly, on #9045 itself, with its fix
active. Only the operator spelling differs:

```toml
[scale.kubernetes]
app_name             = "realapp"
namespace            = "${NOPE:+prod}"
bundle_storage_class = "${NOPE:0:3}"

[scale.secrets]
SOME_API_KEY = "${A_VAR_NOBODY_SET:+x}"
```

```
$ jac scale deploy --dry-run --show-yaml      # nothing exported
EXIT=0
check: no errors or warnings

  namespace: ${NOPE:+prod}                    (x8 manifests)
  SOME_API_KEY: ${A_VAR_NOBODY_SET:+x}
  storageClassName: ${NOPE:0:3}
[...]

PR #9199, opened 2026-09-15
fix(scale): fail the deploy when the shipped jac.toml reads env the pods will not have

akindu-k:
Refs #8773. Needs to land **before** #9045.

## The problem in one sentence

`jac scale deploy` checks `jac.toml` using the environment variables on **your machine**, but the pods read `jac.toml` again using **their own** environment, so a variable that only exists on your machine passes the deploy and then breaks inside the pod.

## Example

```toml
[scale.kubernetes]
app_name = "shop"
image_registry = "${ECR_REGISTRY}"   # exported on your laptop only

[scale.secrets]
OPENAI_API_KEY = "${OPENAI_API_KEY}" # sent to the pods as a Kubernetes Secret

[plugins.demo]
app  = "${K8S_APP_NAME}"             # jac-scale sets this in every pod
key  = "${OPENAI_API_KEY}"           # the pod gets it from the Secret
mode = "${MODE:-fast}"               # has a default
```

**Before this PR:** `jac scale deploy --dry-run` says everything is fine, because your laptop has `ECR_REGISTRY`. The pod does not have it:

- on `main`, the pod silently uses the text `${ECR_REGISTRY}` as the value
- after #9045, the pod refuses to start and keeps crashing

**After this PR:** the deploy stops on your machine, before anything is created in the cluster:

```
$ jac scale deploy main.jac --dry-run
[ERROR] Error: --dry-run failed: jac.toml reads environment variables the deployed pods will not have, so these settings cannot resolve at boot:
  jac.toml: scale.kubernetes.image_registry: Environment variable ECR_REGISTRY is not set (every pod)
Supply each one through [scale.secrets] or the app's env, or give it a ${VAR:-default}.
```

`app`, `key` and `mode` are not reported, because the pod really has those values.

To fix it, either give the value a default (`"${ECR_REGISTRY:-}"`) or send it to the pod through `[scale.secrets]` or the app's `env`.

This is not hypothetical: jacBuilder's `jac.toml` has four values like this (`ECR_REGISTRY`, `K8S_SERVICE_ACCOUNT_NAME`, `GITHUB_PUBLIC_LINK`, `SHARED_DATA_PVC`), so without this check #9045 would crash every jacBuilder pod.

## Breaking change

This fails deploys on `main` by itself, not only once #9045 lands. A project that deploys today with a bare `${VAR}` exported only on the deploy host boots fine, because nothing reads the literal, and will now fail `jac scale deploy`. The release note is filed as `breaking` for that reason.

jacBuilder `dev` still has the four values above as bare `${VAR}`, so they need the `:-` form before jacBuilder picks up a release carrying this.

## How it works

For every pod:

1. **Work out the pod's real environment**, from what the deploy actually gives it:
   - the env vars jac-scale writes into the pod (`K8S_APP_NAME`, `K8S_NAMESPACE`, peer URLs, the app's `env`, ...)
   - the keys of the `[scale.secrets]` Secret
   - the few variables Kubernetes always sets (`HOSTNAME`, `KUBERNETES_SERVICE_HOST`, `KUBERNETES_SERVICE_PORT`)
2. **Collect the config files the pod will read**: the shipped `jac.toml`, a nested project's `jac.toml` if the pod runs code from one, the active profile file (`jac.<profile>.toml`) and `jac.local.toml`.
3. **Fill in every `${...}` using only that environment.**
4. **If anything is missing, stop the deploy** and list every problem at once, with the setting and which pods it affects.

Where it runs:

- `--dry-run` and `preview()`: inside manifest generation, which never touches the cluster.
- A real deploy: right after the runtime preflight and **before database provisioning**, so a refused config leaves no namespace, Postgres StatefulSet or secrets behind. It builds only the pod env for this, not the bundle or the injector.

To avoid failing a deploy that would actually work:

- `JAC_DB_URL` counts as present when a database is configured, because the connection secret only exists once provisioning has run.
- A pod that loads extra env from a source the deploy cannot see (for example a ConfigMap added by `deployment_overlay`) is skipped rather than guessed at.

Two things stay strict on purpose:

- `${1:?msg}` is reported, matching the loader rule in #9045 that an explicit `:?` always fails.
- Kubernetes service-link variables such as `FOO_SERVICE_HOST` are not counted, because they depend on what else is in the namespace. Write `${FOO_SERVICE_HOST:-}` if you need one.

## Small core change

`interpolate_env_vars` can now resolve against a given set of variables instead of the process environment, so the check never touches `os.environ`. It also leaves `${1}` style shell text alone (same rule as #9045), so a `[scripts]` entry like `sh -c 'echo ${1}'` is not reported. On `main` neither change is visible outside this check.

## Tests

| Suite | `main` | here |
| --- | --- | --- |
| `jac/tests/project/` | 329 | 330 |
| `jaclang/scale/tests/deploy/test_pod_env.jac` | 38 | 39 |

Each suite gains one test, both for code paths that did not exist before: the pod env check, and `interpolate_env_vars` with a given environment.

- `test_pod_env.jac`: one table-driven test. Every row runs the real manifest generation with `dry_run=True`, and then a real `KubernetesTarget.deploy()` with the cluster calls mocked and database provisioning set to raise. It covers jacBuilder's four values (all reported, even with them exported on the host), values the pod does have (nothing reported, and the deploy reaches provisioning), a nested project, an active profile plus `jac.local.toml` (reported), and an inactive profile (not reported). Removing the check, just the overlay part, or just the pre-provisioning call in `deploy()` makes it fail.
- `test_config.jac`: `interpolate_env_vars` with a given set of variables.

Also checked with the real CLI: the example above exits 1 on this branch, exits 0 on `main`, and exits 0 on this branch once the value has a default.

## Known limitation

If one string holds two missing variables, only the first is reported. Every other setting is still listed.


jac/jaclang/project/impl/config.impl.jac
```diff
@@ -21,9 +22,11 @@ impl interpolate_env_vars(value: str) -> str {
             return operand;
         } elif operator == ":?" {
             raise ValueError(f"Environment variable {var_name} is required: {operand}");
-        } else {
-            raise ValueError(f"Environment variable {var_name} is not set");
         }
+        if not var_name.isidentifier() {
```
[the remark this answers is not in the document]
akindu-k: Checked, and leaving this one as it is.

`${1:?positional}` failing is the loader's rule, not a side effect of this check. #9045 pins it on purpose: its test `interpolate leaves shell expansions that are not variable names` asserts "an explicit :? still fails, whatever the name". Only a bare `${1}` is left verbatim, because nothing can export `1` and there is no other way to write it. `:?` is the config explicitly asking to fail when the value is missing.

This check has to agree with the loader the pod runs. Once #9045 lands, a pod that reads `${1:?usage}` fails at boot, so reporting it on the host is the right result, not a false positive. Nothing in the repo writes this form either: a `git grep` for a non-name followed by `:?` across `*.toml`, `*.md` and `*.jac` is empty.

If `:?` on a non-name should stay literal instead, that is a loader change, and it belongs with the other operator questions in #9184.

jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac
```diff
@@ -689,9 +689,103 @@ obj ManifestBuilder {
...
+    def _pod_environ(
+        container: dict[str, any], secret_keys: set[str]
+    ) -> dict[str, str] | None {
+        environ: dict[str, str] = {
+            name: ""
+            for name in (
+                "HOSTNAME",
+                "KUBERNETES_SERVICE_HOST",
+                "KUBERNETES_SERVICE_PORT"
+            )
```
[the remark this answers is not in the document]
akindu-k: Checked, and not adding these.

They can exist: jac-scale never sets `enableServiceLinks`, and `apply_manifests` applies the Services before the Deployments. But which ones a pod gets depends on every Service already in the namespace when that pod is created. The deploy host cannot know that, least of all under `--dry-run` against a namespace that may not exist yet. The only way to count them would be to treat any `*_SERVICE_HOST` or `*_PORT` name as present, and that would hide real host-only values with the same shape, which is exactly what this check is for.

Nothing in jac-scale needs them either. Each pod already gets explicit peer URLs (`JAC_APP_<NAME>_URL`) plus `K8S_APP_NAME` and `K8S_NAMESPACE`, and `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT`, which every pod does get, are already counted. A config that really wants a service-link variable can write `${FOO_SERVICE_HOST:-}` and deploy.

So this stays a strict check with an explicit escape rather than a loose one.

jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac
```diff
@@ -689,9 +689,103 @@ obj ManifestBuilder {
...
+            if environ is None {
+                continue;
+            }
+            tomls = {"jac.toml": base_toml};
+            nearest = self._nearest_nested_toml(nested, environ.get("JAC_SV_FILE", ""));
+            if nearest {
+                tomls[nearest] = nested[nearest];
+            }
+            for (rel, text) in tomls.items() {
+                for problem in unresolved_toml_references(text, environ) {
```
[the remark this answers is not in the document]
akindu-k: Valid, fixed in 47cb6fcea.

Verified first. Both files ship as they are (`bundle_source_files` on a project with them returns `jac.local.toml,jac.prod.toml,main.jac`), and the loader merges them at boot: `apply_profile_overlay` reads `jac.<profile>.toml` for the profile from `JAC_PROFILE` or `[environment] default_profile`, then always reads `jac.local.toml`. Loading that project resolved `default_profile = "prod"` and returned the overlays' literal `${...}` values.

The check now adds those overlays, when they ship, to each `jac.toml` a pod loads, root and nested. An overlay for a profile the pod does not activate is skipped, since the pod never reads it.

Real CLI, host-only values in both overlays with all of them exported on the host:

```
$ jac scale deploy main.jac --dry-run
[ERROR] Error: --dry-run failed: jac.toml reads environment variables the deployed pods will not have, so these settings cannot resolve at boot:
  jac.prod.toml: plugins.probe.prod_only: Environment variable PROD_ONLY is not set (every pod)
  jac.local.toml: plugins.probe.local_only: Environment variable LOCAL_ONLY is not set (every pod)
EXIT=1
```

With those two written as `${VAR:-}` it exits 0, and an inactive `jac.staging.toml` holding a bare `${STAGING_ONLY}` is not reported.

`test_pod_env.jac` gets two rows in the existing table: active profile plus `jac.local.toml` (both reported), and an inactive profile overlay (nothing reported). With overlay collection disabled, the first fails with `the active profile overlay and jac.local.toml: ''`. Restored, 39 passed, same count.

jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac
```diff
@@ -689,9 +689,129 @@ obj ManifestBuilder {
                 );
             }
         }
+        self._check_pod_config_resolves(deployments, user_secret);
```
MusabMahmoodh, line 692: On a real deploy this runs after `DatabaseProvisioner.provision` has already created the namespace, the Postgres StatefulSet and its secrets, so a config this refuses still leaves those in the cluster. Only `--dry-run` actually stops before anything reaches the cluster, which is what the PR body and release note promise for both. Since `_pod_environ` already predicts `JAC_DB_URL` from `database_enabled()` under dry run, could `deploy()` run the check ahead of `provision` with that prediction, or should the note say dry run only?

release_notes/unreleased/jaclang/9199.bugfix.md
```diff
@@ -0,0 +1 @@
+- **Fix: Deploys catch `jac.toml` values the pods cannot resolve**: `jac scale deploy`, including `--dry-run`, now stops before anything ships when the shipped `jac.toml` references an environment variable the pods will not have, and lists every such setting at once. Supply the variable through `[scale.secrets]` or the app's env, or give it a `${VAR:-default}`.
```
MusabMahmoodh: This fails deploys on `main` by itself, not only once #9045 lands: jacBuilder `dev` still has the four bare values, and today those pods boot fine because nothing reads the literals. Worth one sentence here that an existing project with a host-only bare `${VAR}` will now fail `jac scale deploy`, and jacBuilder needs those four on `:-` before it picks up a release carrying this.

MusabMahmoodh (approved):
Right shape: building each pod's env from the manifests the builder actually generates, instead of a hand-kept list, is what keeps this from failing deploys that would boot. Two things below, neither blocking.

jac/jaclang/scale/deploy/target/kubernetes/manifest_builder.jac
akindu-k: Valid, fixed in b7a34d516.

Confirmed the ordering first. `deploy()` ran the runtime preflight, then `DatabaseProvisioner.provision` (`ensure_namespace_exists`, the Postgres StatefulSet and its secrets), and only then `generate_manifests`, so a refused config left all of that behind.

Went with your first option. `deploy()` now runs the check right after the preflight and before provisioning, through the same builder `generate_manifests` uses. It calls a new `ManifestBuilder.check_pod_config` that renders only the pod specs, deployments and user Secret, so the bundle, the client build and the injector are not done twice. `JAC_DB_URL` is now predicted whenever `database_enabled()`, not only under dry run, because before provisioning its `secretKeyRef` does not exist yet. Dry run and `preview()` keep the check inside manifest generation.

The existing table test in `test_pod_env.jac` now also runs a real `KubernetesTarget.deploy()` for every row, with the cluster calls mocked and `DatabaseProvisioner.provision` raising. A row with a problem must fail with the same message and never reach provisioning; a clean row must reach it. With only the new call in `deploy()` removed:

```
AssertionError: host-only values: database provisioning failed: provisioned
38 passed, 1 failed
```

Restored, 39 passed, same count. The scale deploy folder matches `main` (375 passed; the only failures are the `dotenv`-gated ones, identical on both).

release_notes/unreleased/jaclang/9199.bugfix.md
akindu-k: Valid. Checked jacBuilder `dev`: `jac.toml` lines 305, 308, 521 and 539 still hold `ECR_REGISTRY`, `K8S_SERVICE_ACCOUNT_NAME`, `GITHUB_PUBLIC_LINK` and `SHARED_DATA_PVC` bare, and on `main` those pods boot because the loader keeps the unresolved literal.

Since this makes deploys that pass today fail, the note moved from `bugfix` to `breaking` in b7a34d516 and now says it outright: a project that deployed with a bare `${VAR}` exported only on the host will fail, followed by the two ways to fix it.

The jacBuilder part is about one downstream repo, so it stays out of the release note and is in the PR description under "Breaking change" instead: those four need the `:-` form before jacBuilder picks up a release carrying this.

akindu-k:
@MusabMahmoodh thanks for the review. Here is a plain summary of your two points and how each one is addressed in b7a34d516, using one example.

```toml
[scale.kubernetes]
image_registry = "${ECR_REGISTRY}"   # only set on my laptop

[scale.database]
mode = "provision"                   # jac creates a Postgres database for the app
```

## 1. The check ran too late on a real deploy

On a real deploy (not `--dry-run`) the steps ran in this order:

```
1. check the jac runtime version
2. create the database   -> namespace + Postgres + secrets now exist in the cluster
3. build the manifests   -> the check ran HERE and failed on ECR_REGISTRY
4. apply to the cluster
```

The deploy stopped at step 3, but step 2 had already created things in the cluster, so the user got an error and a leftover Postgres. "Stops before anything ships" was only true for `--dry-run`.

**Now** the check runs before the database is created:

```
1. check the jac runtime version
2. check the config the pods load   <- moved here, fails on ECR_REGISTRY
3. create the database              <- never reached, nothing left in the cluster
4. build the manifests
5. apply to the cluster
```

One catch: `JAC_DB_URL` only exists after the database is created, so an early check would have wrongly reported `${JAC_DB_URL}`. The check now counts `JAC_DB_URL` as present whenever a database is configured, not only under dry run.

The existing table test in `test_pod_env.jac` now also runs a real `deploy()` with the cluster mocked, and checks that a bad config stops before the database step. Removing just the new early check makes it fail.

## 2. The release note should warn that existing projects will now fail

With the same `image_registry = "${ECR_REGISTRY}"` line:

- **On `main` today:** the deploy works. The pod gets the literal text `${ECR_REGISTRY}`, but nothing in the pod reads it, so it boots fine.
- **With this PR:** the same deploy now fails with an error.

A project that deployed fine yesterday breaks after upgrading. I checked jacBuilder `dev` and it has four values like this (`ECR_REGISTRY`, `K8S_SERVICE_ACCOUNT_NAME`, `GITHUB_PUBLIC_LINK`, `SHARED_DATA_PVC`).

**Now** the release note is filed as **breaking** instead of bugfix. It says that a project using a variable only set on the deploy host will fail, and how to fix it: write `${ECR_REGISTRY:-}`, or send the value to the pods through `[scale.secrets]`. The jacBuilder follow-up (those four need the `:-` form before it picks up this release) is in the PR description under "Breaking change".

## In short

| You said | Addressed by |
| --- | --- |
| A real deploy creates the database before the check fails, leaving it behind | Moved the check ahead of database creation, pinned by the test |
| Existing projects like jacBuilder will start failing, so say so | Release note is now breaking; jacBuilder follow-up noted in the PR description |

Issue #9031, opened 2026-09-07
scheduler: @schedule accepts any spec; interval=0 spins at 300 fires/s, a negative interval starves other tasks, an unsatisfiable cron crashes boot


kashmithnisakya:
Part of #9024.

`@schedule` forwards its kwargs straight into `ScheduleSpec` (`jaclang/runtime/impl/builtin.impl.jac:353-364`) and nothing validates them at decoration, registration, or boot. The effects range from a busy loop to a server that will not start.

## What I saw

| Spec | Behaviour |
|---|---|
| `@schedule(interval=0)` | 1900 fires in 6 s, one thread and one event loop per fire |
| `@schedule(interval=-5)` | 4065 fires in 12 s, and the two other tasks in the same module fired once instead of five times: a negative interval's next time is always in the past, so it always heads the heap and starves everything else |
| `@schedule(date="2020-01-01 00:00:00")` | dropped, no warning; the module logged "3 task(s) registered" for 5 decorated defs |
| `@schedule()` | dropped, no warning |
| `@schedule(interval=2, cron="* * * * *")` | interval silently wins, cron ignored |
| `@schedule(cron="60 * * * *")` | server boot fails after a 366-day minute-by-minute scan with a raw traceback (below); out-of-range fields never match instead of erroring on the field |

## Evidence

```
✖ Error: Server error: No matching time found for cron: 60 * * * *
  at _parse_cron_next() jaclang/server/impl/task_scheduler.impl.jac:222
  at _compute_next_run() task_scheduler.impl.jac:173
  at register() task_scheduler.impl.jac:13
  at _register_scheduled_tasks() server/impl/server.impl.jac:1752
  at _prepare_to_serve() server.impl.jac:1810
  at start() server.impl.jac:1855
```

```
[spin] fires=1900 pid=36203          (interval=0, ~6 s after boot)
[negative] fired  x4065              (interval=-5, 12 s)
[both] fired at 11:22:07             (interval=2 + cron, fired once in 12 s)
[plain_print no flush] at 11:22:07   (interval=2, fired once in 12 s)
```

## How to reproduce

Put the six decorated defs above in one module and `jac run --serve --no-client main.jac`.

## Proposed fix

[...]

Issue #9024, opened 2026-09-07
scheduler: field review of the @schedule runtime after #8981 (tracking)


kashmithnisakya:
Tracking issue for a field review of the `@schedule` runtime on `main` at `c89fd6b0e` (after #8981), done 2026-09-07. Method: code read of `server/task_scheduler`, `server/impl/server.impl.jac` (registration, lease, per-worker start), `scale/scheduler`, `scale/server/jobs_ext` and `serving/supervisor`, then live runs with one and three workers against the embedded Postgres, a worker `SIGKILL` mid-run, rows inserted straight into `scheduled_jobs`, edge-case fixtures, a `--dev` reload, and the `/jobs` API. Each finding below is a sub-issue with its own reproduction and log excerpts.

## The three questions this started from

**Does the scheduler run in every pod?** Yes, and in every worker process of every pod. `_register_scheduled_tasks` runs once in the parent, then each forked worker calls `scheduler.start_if_pending()` and starts its own APScheduler. Deduplication is a per-fire Postgres lease (`sched:<task name>` for static tasks, `sched:<job id>` for dynamic ones; TTL `min(300, 0.8 * interval)`, 45 s for cron, 300 s for dates). With three workers a 2 s task fired exactly once per tick, winners alternating between pids. The gateway pod runs no schedules. The lease exists only when identity storage is Postgres and fails open on error.

**Does it execute correctly?** Static tasks fire on time and their graph writes commit (the scheduled walker's node count survived restarts). Around that: every walker tick logs a phantom error, there is no overlap guard, in-flight runs die at shutdown, and the `/jobs` API cannot create a single job.

**Can we see terminal prints?** Yes. `print()` from a task reaches the terminal and the pod log promptly, line-buffered even through a pipe and without `flush`. A healthy tick logs nothing at all, though, so silence is ambiguous.

## What works

- one fire per tick across workers when Postgres is present (static and restored dynamic jobs alike)
- walker writes commit; cron fires on the minute boundary
- `SIGTERM` drains HTTP and stops the scheduler thread; the process tree exited in 0.5 s
- restored dynamic jobs keep `run_count` / `last_status` / `last_error`

## Findings

| Severity | Issue | Finding |
|---|---|---|
| Critical | #9025 | POST /jobs answers 404 for every DYNAMIC target, so no dynamic job can be created (live) |
| Critical | #9026 | every static walker tick logs "Invalid walker object" after the walker already ran (#7520 regression) (live) |
[...]

PR #9201, opened 2026-09-15
fix(scale): claim each scheduler tick once per service instead of leasing a time window

### LINKED ISSUE #9027: scheduler: interval tasks of 375 s or more fire once per replica because the lease TTL is capped at 300 s and each replica phase is its boot time
Part of #9024.

An interval task of 375 s or more runs once **per replica** whose boot time differs from the others by more than five minutes, because the per-fire lease TTL is capped at 300 s while every replica computes its ticks from its own clock. The same mechanism makes a respawned worker fire the inherited, already-past tick time immediately, and can re-fire a one-shot `date` task.

## Root cause

Each replica's next fire is `now + interval` from its own boot (`jaclang/server/impl/task_scheduler.impl.jac:169-170`), so a replica's phase is its boot time. Deduplication is a lease keyed by task name whose TTL is:

```
# jaclang/server/impl/task_scheduler.impl.jac:46-54
if spec.interval is not None and spec.interval > 0 {
    return max(1, min(300, int(float(spec.interval) * 0.8)));
}
if spec.cron is not None { return 45; }
return 300;
```

Once the 300 s cap applies (interval above 375 s), two replicas whose phases differ by `d` with `300 < d < interval - 300` each win their own tick, because the other replica's lease has already expired.

```
interval = 3600, ttl = 300
pod A boots at 00:00 -> fires 01:00, 02:00, 03:00
pod B boots at 00:10 -> fires 01:10, 02:10, 03:10   (A's lease expired at 01:05)
result: the hourly task runs twice an hour
```

Pods rarely boot within five minutes of each other after a rolling update or a scale-up, so this is the normal case for hourly and daily interval tasks. Cron tasks are wall-clock aligned and are not affected.

## Respawn side of the same bug

Fire times are computed in the parent before fork and never recomputed in a child. A worker that is recycled (`max_requests`), crashes, or is rolled by `SIGHUP` inherits the parent's original `next_run`, which is already in the past, and fires as soon as its loop starts if the lease has expired. For a one-shot `date` task that means a second run if the respawn lands more than 300 s after the real one; for a cron task it means a fire off the minute boundary.

Observed with `interval=5`, three workers, worker 1 `SIGKILL`ed 3.4 s after a fire:

```
[beat5] pid=50447 at=11:25:41.539
WARNING worker 1 exited with -9; respawning in 1.0s
INFO    worker 1 started (pid 50479)
[beat5] pid=50479 at=11:25:45.933      <- 4.4 s after the previous fire, from the new worker
[beat5] pid=50479 at=11:25:50.943      <- and it keeps the phase from here on
```

With a short interval this only shifts the phase; with the cap in play it produces the duplicates above.

## How to reproduce

```jac
@schedule(interval=610)
def beat610 -> None { print(f"[beat610] pid={os.getpid()}", flush=True); }
```

```bash
jac run --serve --no-client -w 3 main.jac
# wait for the first fire (boot + 610 s), then SIGKILL one worker 303 s after it
```

The respawned worker fires immediately (its lease window is open), then every 610 s on its own phase, while the survivors keep theirs: two fires per 610 s. A three-worker run of exactly this is in progress; its timeline will be appended as a comment.

## Proposed fix

Align interval ticks to a shared epoch (`floor(now / interval) * interval + interval`) so every replica computes the same tick timestamps, and key the lease by tick (`sched:<name>:<tick>`). The TTL then only has to cover clock skew and can be a few seconds. Recompute `next_run` in `Scheduler.start()` instead of inheriting it across `fork`, and record one-shot completions in shared state rather than in a worker's memory.

## Environment

### LINKED ISSUE #9028: scheduler: services share one scheduled_jobs collection and one lease namespace, so foreign jobs are restored, marked failed, and steal ticks
Part of #9024.

Every service pod in a deployment shares one database (`kubernetes_target.jac:132` sets the same `JAC_DB_URL` on every pod), one `scheduled_jobs` collection, and one lease namespace. Each server restores **every** active job row, including jobs whose target lives in another service, and the static lease key is the bare task name. Overlaps with #7617 (closed 2026-09-04 without a merged fix).

## What happens

- `register_dynamic_tasks_from_db` (`scale/scheduler/impl/scheduler.impl.jac:419-475`) schedules every active row. When a pod that does not own the target wins the per-fire lease, it logs a warning, records a **failed** run, and the tick is spent; the owning service runs the job only on the ticks the wrong pod loses, and the row's `last_status` flaps.
- Static tasks lease on `sched:<task name>` (`server/impl/task_scheduler.impl.jac:62`). A `heartbeat` in two services contends for one key, so one of them silently skips every tick.

## How to reproduce

Insert a job row whose target is not in the served module (what a job created against another service looks like from this pod), then serve with three workers:

```sql
INSERT INTO jac_docs (collection, id, doc) VALUES ('scheduled_jobs', 'ghost-1',
 '{"job_id":"ghost-1","name":"GhostWalker","created_by":"__system__","is_walker":true,
   "created_at":"2026-09-07T00:00:00+00:00","status":"active","trigger":"interval","interval":2.0}'::jsonb);
```

```bash
jac run --serve --no-client -w 3 main.jac
```

## Evidence

```
INFO    Dynamic job restoration complete: 2 restored, 0 orphaned, 0 failed
WARNING Scheduled walker 'GhostWalker' is no longer registered in the server     (every 2 s)
...
ghost-1     status=active  run_count=14  last_status=failed     last_error='target not registered'
restored-1  status=active  run_count=14  last_status=succeeded
```

Lease rows after the run show the same unqualified keys for both kinds:

```
lease:sched:beat_fn      Kashmiths-Mac-mini.local:52522
lease:sched:ghost-1      Kashmiths-Mac-mini.local:52523:69fddab7...
```

## Proposed fix

Qualify the lease keys and the collection by app or service (`sched:<app>:<name>`, `scheduled_jobs` scoped by app), and have restore skip rows whose target is not registered locally instead of scheduling and failing them.

## Environment

### LINKED ISSUE #9030: scheduler: the per-fire lease fails open, so a database error makes every worker of every pod fire
Part of #9024.

Both lease checks catch every exception, log "firing anyway", and return `True`. A Postgres blip or failover during a tick therefore makes every worker of every pod run the task at once, which is the worst outcome for exactly the jobs that need a lease (mail, billing, external calls).

## Where

```
# jaclang/server/impl/task_scheduler.impl.jac:63-66
} except Exception as e {
    logger.warning(f"lease check for '{task.name}' failed: {e}; firing anyway");
    return True;
}
```

```
# jaclang/scale/scheduler/impl/scheduler.impl.jac:691-694
} except Exception as e {
    logger.warning(f"fire-lease check failed for job {job_id}: {e}; firing anyway");
    return True;
}
```

## How to reproduce

Serve with `-w 3`, then stop the database (`jac db stop` for the embedded server, or cut the connection) during a tick. Each worker logs the warning and runs the task. Not exercised live in this review to avoid stopping the shared embedded cluster mid-run; the code path is unconditional.

## Proposed fix

Fail closed: skip the tick and log once per task per outage. If a fallback is wanted, elect one designated worker (index 0 of the lowest-ordinal pod) rather than all of them.

## Environment

kashmithnisakya:
Closes #9027. Closes #9028. Closes #9030. Part of #9024.

The three share one root. A fire was deduplicated by a **time window** (a lease on the task name with TTL `min(300, 0.8 × interval)`), while each replica computed **its own ticks** from its boot time. Nothing tied a lease to a particular tick, so the window only worked while replicas happened to share a phase, and it had to guess what to do when the database could not answer. This replaces the window with a claim on the tick itself.

## The mechanism

- **Shared ticks.** Interval ticks are multiples of the interval since the Unix epoch (UTC), so every replica computes the same ones whenever it booted: the static scheduler in `_compute_next_run`, and dynamic jobs by giving APScheduler's `IntervalTrigger` `start_date=epoch`. Cron ticks were already wall-clock aligned, and a date is its own tick.
- **Claims, not leases.** `PgStore.kv_claim_tick(key, tick)` is a single `INSERT ... ON CONFLICT DO UPDATE ... WHERE stored < tick RETURNING key`. It admits a tick strictly above the last one claimed for that key, atomically. One row per task, no TTL, no cap. A replica that is late, respawned, or booted mid-interval asks for a tick that is already claimed and is refused. The claim is taken before the run, so the guarantee is **at most once** per tick: a replica that stops after claiming has spent that tick, and nothing retries it.
- **Fail closed.** If a claim raises, the tick is skipped. The server logs one warning per task for the outage and one info line when claims succeed again.
- **Fresh fire times per worker.** `Scheduler.start()` recomputes `next_run` instead of inheriting the parent's across fork.

## #9027: once per replica, and the respawn fire

Both symptoms came from per-replica phases. With shared ticks and a per-tick claim, a 3600 s task on two pods booted ten minutes apart has one set of ticks and one winner each. A respawned worker holding the parent's past `next_run` recomputes it on start, and even the stale tick it might ask for is already claimed. For dynamic jobs, a run claims **the time it was scheduled for**. APScheduler's `run_job` knows each run time but calls the job with no arguments, so the engine (built in one place, `new_engine`) uses `FireTimeExecutor`, a `ThreadPoolExecutor` whose `_do_submit_job` passes each run its `scheduled_at`, catch-up runs included. Deriving the occurrence from the clock instead is wrong for cron: flooring to the current minute yields ticks that are not occurrences, so a late delivery could run an hourly job a second time. A `date` job has one scheduled time, so it cannot run twice.

## #9028: services sharing one database

- Static claim keys are `sched:<service>:<task>`. The service is `JAC_SV_NAME`, which the fleet sets on every service pod, else the served module name (`schedule_namespace`, used by both schedulers).
- Job rows record the `service` that created them. A service restores, reconciles (NOTIFY), reads, changes and deletes only jobs it serves (`_serves`: a stamped row must name this service, an unstamped one must name a target registered here), checked before anything is written. A non-owner answers `404`, which is what moves the fleet gateway's discovery on to the owner.
- **Listing is not scoped.** `GET /jobs` is a discovery route on the gateway, which returns the first service's 200, and the collection is shared, so every service lists every service's jobs, each row carrying its `service`, as before this PR. `find_active` takes the service only where restore needs it.
- A row with no `service`, written by an earlier release, belongs to whichever service serves its target. Nothing records a failed run for a target that lives elsewhere.
- Dynamic claim keys are the job id, which is already unique.

**Why a field and not a per-service collection**, as the issue suggests: renaming `scheduled_jobs` per service would strand every persisted job at upgrade. A field keeps existing rows readable and lets the ambiguous ones be handled explicitly.

## #9030: fail open

`_may_fire` and `_claim_fire` now return `False` when the claim raises. Running a mail or billing job on every worker because the database blinked is worse than skipping one tick. There is no elected fallback worker, since electing one also needs the shared state that is unavailable.

## Behaviour changes worth knowing at upgrade

- **Interval phases move once, to the epoch grid.** `interval=3600` now runs on the hour, and the first run comes within one interval of boot rather than exactly one interval after it. The docs table says so.
- **Mixed versions during a rolling upgrade** do not see each other's claims (old pods lease `lease:sched:<name>`, new pods claim `tick:sched:<service>:<name>`), so a tick can run twice while both versions are live.
- **A database outage skips ticks** instead of running them everywhere.
- `interval <= 0` keeps its current `now + interval` behaviour, because the grid needs a positive step. Rejecting it at declaration is #9031.

## Evidence

Red against `main` (`e28abb4`), the same scenarios written against main's API:

```
9027 static   pod A fires 11:43:23.500000  pod B fires 11:48:28.500000  gap 305.0 s  lease ttl 300 s
              AssertionError: phases differ by 305.0 s, above the 300 s lease
9027 dynamic  first fire 2026-09-15 11:42:56.077956+00:00, kwargs {'seconds': 610.0}
              AssertionError: not on any shared grid
9030 static   WARNING lease check for 'mailer' failed: the database is restarting; firing anyway
              _may_fire with an unreachable database -> True
9030 dynamic  WARNING fire-lease check failed for job mail-job: the database is restarting; firing anyway
              _acquire_fire_lease with an unreachable database -> True
9028          INFO Dynamic job restoration complete: 3 restored, 0 orphaned, 0 failed
              scheduled on the billing pod: ['billing-beat', 'ghost-1', 'orders-beat']
5 failed
```

Live on this branch: `jac run --serve --no-client -w 3`, a static `interval=5` task, embedded Postgres. The worker that fired last was SIGKILLed at 11:35:01.7 and respawned at 11:35:03.0 as pid 23354:

```
[beat] pid=23248 at=11:34:45.029
[beat] pid=23250 at=11:34:50.036
[beat] pid=23250 at=11:34:55.010
[beat] pid=23248 at=11:35:00.019
WARNING worker 0 exited with -9; respawning in 1.0s
INFO    worker 0 started (pid 23354)
[beat] pid=23250 at=11:35:05.022
[beat] pid=23249 at=11:35:10.017
[beat] pid=23250 at=11:35:15.025
[beat] pid=23250 at=11:35:20.023
[beat] pid=23250 at=11:35:25.017
[beat] pid=23354 at=11:35:30.022
```

Ten ticks, ten fires, every one on the 5 s grid, none duplicated. The respawned worker did not fire on its inherited time: on `main`, #9027 records the same kill firing 4.4 s after the previous fire and keeping that phase. The fail-closed path was not exercised live, because stopping the embedded cluster would take down other local apps. It is covered by tests whose claim raises.

## Tests

- `tests/runtimelib/test_scheduling.jac`, 6 new: replicas booted 305 s apart share the 610 s grid, and an hourly task lands on the hour; a worker recomputes inherited fire times on start (asserting `last_run` too, so it fails on the old code whichever thread wins); one replica wins each tick and a stale tick is refused; the same task name in two services is claimed separately; namespace resolution; a claim that raises skips the tick, warns once over two failures, and logs the recovery. The interval assertion in "compute next run" changes from `now + 60 ± 1` to "within the next 60 s, on a minute boundary", which is the new contract.
- `jaclang/scale/tests/server/test_scheduler_multipod.jac`, against real embedded Postgres, 6 new and the lease test rewritten: one pod per scheduled occurrence, a re-run refused, claims only move forward; APScheduler interval jobs scheduled 305 s apart land on one grid; each run carries its scheduled time (four catch-up runs, four minutes); a late cron delivery of an occurrence pod A already ran does not run on pod B, while an unclaimed one runs exactly once; a real `PgStore` on a closed port skips the claim; services sharing a collection list every job but reach only their own; restore, NOTIFY reconcile, `get_job`, `update_job` and `delete_job` leave an unstamped job to the service that serves its target.

Review round, each reproduced on the previous commit before fixing: a late delivery of 09:00 started at 09:42 ran again (`run_count=1`); `update_job` on a service that cannot run a legacy job changed its interval to 4.0; the first service's list held only its own job. The dead-port test fails if claims fail open, and the recompute test fails three runs out of three without the recompute.

Green locally against the source tree, rebased on `main` at `c54d9bde`: `test_scheduling` 12, `test_scheduler_multipod` 11 (APScheduler installed, nothing skipped), `test_pg_job_store` 8, `test_scheduler_endpoints` 11, `test_optional_deps` 24, `test_config_validation` 31, `test_config_loader` 19, `test_app_scale_overrides` 18. `jac fmt --check` and per-file `jac check` are clean on every changed runtime file.

## Rebased after #9192

#9192 has merged, and this branch is rebased on `main` at `c54d9bde`. The conflicts were additive: #9192's quota imports, `_memory_jobs_lock` and the job store's dedicated quota connection sit beside this PR's claim imports, epoch and `_service` field. #9192's `min_interval` clamp keeps this PR's `start_date` when it floors `seconds`.

The last commit, `ad8772f1`, scopes #9192's per-account cap to the service. With job rows recording their service, counting a creator's active jobs across the whole collection let a job in one service use up the cap in another. `PgJobStore.insert_within_quota` now counts only rows in its own service, plus rows written before jobs carried a service (the same scope its other service reads use). The advisory lock is keyed by collection, service and creator, and `MemoryJobStore` scopes its count by the new job's service. Tested red first: a creator's job in `orders` refused their first job in `billing` on both stores. A pre-service row still counts against every service's cap. The docs row, the schema description and the release note say the cap is per service.


jac/jaclang/scale/scheduler/impl/scheduler.impl.jac
```diff
@@ -125,8 +131,12 @@ impl PgJobStore.find_active(
     trigger: str | None = None,
     created_by: str | None = None
 ) -> tuple[list[dict], int] {
-    conds: list[str] = ["collection = :c", "doc->>'status' = 'active'"];
-    params: dict[str, any] = {'c': self._collection};
+    conds: list[str] = [
+        "collection = :c",
+        "doc->>'status' = 'active'",
+        "COALESCE(doc->>'service', :svc) = :svc"
```
MusabMahmoodh: `/jobs` is a discovery segment on the fleet gateway (`GATEWAY_DISCOVERY_SEGMENTS`, routing.jac:22), and `probe_services` only moves past a service on 404/405, while `list_jobs` always answers 200. With this filter, `GET /jobs` in microservice mode lists only the first service's jobs, where before any service returned the whole collection. Nobody sees it until #9192 lets POST /jobs create anything. Should the gateway merge the lists, or should the docs say a list covers one service?

jac/tests/runtimelib/test_scheduling.jac
```diff
@@ -291,3 +300,188 @@ test "the served app registers its scheduled targets with the scheduler" {
...
+    # the past, and used to fire it the moment its loop began.
+    sched = fresh_scheduler();
+    sched.register("beat", lambda { None; }, ScheduleSpec(interval=600), False);
+    stale = datetime.datetime.now(datetime.timezone.utc) - datetime.timedelta(hours=1);
+    sched._tasks[0].next_run = stale;
+    sched.start();
+    try {
+        refreshed = sched._tasks[0].next_run;
+        assert refreshed is not None;
+        assert refreshed > datetime.datetime.now(datetime.timezone.utc) , refreshed;
```
MusabMahmoodh: Without the recompute in `start()`, this passes whenever the loop thread pops the stale entry first, because it fires it and writes a fresh `next_run` before this line reads it. The loop sets `last_run` the moment it pops, so asserting both fails on the old code whichever thread wins:

```suggestion
        task = sched._tasks[0];
        assert task.last_run is None , task.last_run;
        assert task.next_run is not None;
        assert task.next_run > datetime.datetime.now(datetime.timezone.utc) , task.next_run;
```

jac/jaclang/scale/tests/server/test_scheduler_multipod.jac
```diff
@@ -45,13 +47,40 @@ def _store -> (PgStore | None) {
...
+    has introspector: _Introspector,
+        module_name: str = "billing";
+}
+
+
+obj _UnreachableStore {
+    def kv_claim_tick(key: str, tick: int) -> bool {
+        raise ConnectionError("the database is restarting");
+    }
+}
```
MusabMahmoodh: The rest of this file runs the real embedded cluster, so this hand-written store is the one place the claim path is not real. A `PgStore` pointed at a port with nothing listening fails in `_side()` the way a restarting database does, and still goes through the real `kv_claim_tick`.

jac/jaclang/cli/docs/reference/plugins/jac-scale-scheduler.md
```diff
@@ -105,7 +105,7 @@ def year_end_cleanup -> None {
 
 Static tasks run as the system user. Use them for app-wide work such as cache warming, digests, and cleanup, not for per-user logic.
 
-Every server replica registers and fires its own copy of each static task, so when you run more than one replica the task executes once per replica per tick. Keep static work idempotent, or route per-tick work through a dynamic job, which takes a per-fire lease when a database is configured.
+With a database configured, each tick of a static task runs on exactly one replica: every replica computes the same ticks and claims each one in the database before running it, so a replica that boots, restarts, or is respawned later cannot run a tick that already ran. Claims are made per service, so two services sharing one database can each have a task with the same name. If a claim cannot be made because the database is unreachable, that tick is skipped rather than run on every replica, and the server logs one warning per task until claims succeed again. Without a database there is no coordination and every replica fires its own copy.
```
MusabMahmoodh: The claim is taken before the task runs, so this is at most once, not exactly once: a replica killed after claiming has spent the tick and nothing retries it. Worth saying plainly for the billing-style jobs this protects; the release note says exactly once too.

MusabMahmoodh (commented):
Claiming the tick instead of leasing a window is the right shape: claims only move forward, so skew or a late callback can cost a tick but never run one twice, and failing closed is the right call for billing-style jobs. The main thing is listing through the fleet gateway, inline. The rest is the delivery wording and two tests that do not quite prove what they are named for.

jac/jaclang/scale/scheduler/impl/scheduler.impl.jac
```diff
@@ -765,7 +821,9 @@ impl JacScaleScheduler._schedule_apscheduler_job(
     }
 
     def job_func {
-        if not self._acquire_fire_lease(job_id, spec) {
+        if not self._claim_fire(
+            job_id, spec, datetime.datetime.now(datetime.timezone.utc)
```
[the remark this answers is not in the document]
kashmithnisakya: Valid, with one correction about where it bites. For interval jobs a late run could not repeat an occurrence: every grid point is a real occurrence and claims only move forward, so a run that starts past the next boundary just takes that next occurrence. For cron, flooring to the current minute produces ticks that are not occurrences at all, so the same argument fails. Reproduced on the previous commit: pod A claims the 09:00 run of `0 * * * *`, pod B's late delivery of that 09:00 run starts at 09:42, claims 09:42, and runs (`run_count=1`).

Fixed in 9632763d by claiming the scheduled time, as you suggest. APScheduler's `run_job` knows each run's `run_time` but calls the job with no arguments, so the engine now uses `FireTimeExecutor`, a `ThreadPoolExecutor` whose `_do_submit_job` (the documented override for custom executors) runs each run time through `run_job` with `scheduled_at` added to the job's kwargs. `_claim_fire` keys on that time, and `TriggerSpec.due_tick`, the clock-based guess, is gone. It also keeps catch-up runs honest: with `coalesce=False` several missed run times arrive together, and each now claims its own occurrence. The engine is built in one place (`new_engine`); a run that somehow arrives without a scheduled time is skipped with an error, not claimed.

Tests: `a late run claims the occurrence it was scheduled for, not the current one` delivers the hour's occurrence late on pod B after pod A claimed it (no run), and delivers an unclaimed one the same way (exactly one run, so the first half is not passing vacuously). `each run carries the time it was scheduled for` checks that four catch-up runs receive four distinct minutes.

jac/jaclang/scale/scheduler/impl/scheduler.impl.jac
```diff
@@ -562,7 +589,7 @@ impl JacScaleScheduler.update_job(
     if not self._require_store().update(job_id, updates, where=where) {
         return None;
     }
-    if self._apscheduler is not None {
+    if self._apscheduler is not None and self._serves(existing) {
```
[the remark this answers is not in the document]
kashmithnisakya: Valid, reproduced on the previous commit: `update_job` on a billing pod changed a legacy `GhostWalker` row's interval to 4.0 although billing does not serve it.

Fixed in 9632763d by applying ownership before any read or write of one job. `get_job` returns a job only if this service serves it (`_serves`: a stamped row must name this service, an unstamped one must name a target registered here), and `update_job` and `delete_job` go through `get_job` first, so they return None/False without touching the row. Through the fleet gateway that 404 is what moves discovery on to the service that owns the job, so the change still lands, on the right service. The restore test now asserts `get_job`, `update_job` and `delete_job` on the ghost row leave it untouched (interval still 2.0, row still present).

jac/jaclang/scale/scheduler/impl/scheduler.impl.jac
kashmithnisakya: Good catch, and confirmed in `routing.jac` and `probe_services`: discovery returns the first service's non-404/405, and `list_jobs` always answers 200. Neither a gateway merge nor a doc caveat is needed, because the job collection is shared: any service can list every job. The mistake was scoping the listing at all. In 9632763d `find_active` takes the service as an optional filter, and only restore passes it (scheduling is the one place ownership decides what a list contains). `GET /jobs` from any service lists every service's jobs again, each row carrying its `service`, as before this PR. Single-job operations stay scoped (see the reply on the legacy-row thread), so `GET/PUT/DELETE /jobs/{id}` 404 on a non-owner and discovery reaches the owner.

Reproduced first on the previous commit: with an `orders` row and a `billing` row, the orders store listed only `["o-1"]`. The listing test now expects all three rows unscoped and the two owned ones with `service="orders"`, and the restore test checks `list_jobs()` returns all four rows. The docs say the same: `GET /jobs` lists every service's jobs, and a service answers 404 for jobs it does not own.

jac/tests/runtimelib/test_scheduling.jac
kashmithnisakya: Valid. Applied your assertion in 9632763d, with a comment saying why it checks `last_run`. Verified it the way you describe: with the recompute removed from `start()` it failed on three runs out of three, and it passes with the recompute.

jac/jaclang/scale/tests/server/test_scheduler_multipod.jac
kashmithnisakya: Valid, and it is a better test. `_UnreachableStore` is gone. The test now builds a real `PgStore` from `conninfo_from_url` on a port found by binding a socket and closing it, so `_claim_fire` goes through the real `kv_claim_tick` and fails in the connection path. Checked that it guards the behaviour: with `_claim_fire` changed to return True on an exception, the test fails.

jac/jaclang/cli/docs/reference/plugins/jac-scale-scheduler.md
kashmithnisakya: Valid. That is the honest guarantee, and it matters most for the jobs this protects. In 9632763d the docs say at most once for both static tasks and dynamic jobs, state that the claim precedes the run so a replica that stops after claiming has spent that tick and nothing retries it, and tell billing-style tasks to pick up on their next tick whatever an earlier one left undone. The release note says at most once too.

MusabMahmoodh (approved):
All four things I raised came back answered, and the tick grid is a better mechanism than the lease it replaces - a shared grid plus a compare-and-set is a real claim, where a TTL'd lease keyed off each replica's own boot time never was.

Checked `kv_claim_tick` rather than taking the message for it: `INSERT ... ON CONFLICT DO UPDATE ... WHERE kv_state.value::bigint < :t RETURNING key` is atomic in one statement, returns nothing when the tick is not strictly newer, and runs on `_side()` rather than the shared transaction connection. The `ELSE true` arm for a non-numeric value is safe because `tick:` is a new key namespace with nothing to migrate. Those rows carry `expires_at = NULL`, which is right - the key set is bounded by services times tasks, not by time.

On the four:
- `GET /jobs` is unscoped again and `find_active(service=...)` is only at restore, so the gateway's first-200 probe no longer hides every other service's jobs. `get_job`/`update_job`/`delete_job` scoping to what a service can run is the right place to keep it.
- at most once, in the docs and the release note, matching what a claim-then-run actually guarantees.
- the recompute test asserts `last_run`, so it fails on the old code whichever thread wins rather than depending on the order.
- the unreachable-store case goes through a real `PgStore` on a closed port instead of a stand-in.

Nothing further from me.

MusabMahmoodh (approved):
Rebase onto #9192 is clean, and the quota count plus its advisory lock now use the same `COALESCE(doc->>'service', :svc)` scope as the other store reads; both new tests fail on the collection-wide count. Worth knowing the effective per-account cap is now `max_jobs_per_user` times the number of services, which the docs do say.

kugesan1105:
## Review summary

I validated this against the three issues independently, then landed four commits on the branch. Flagging that up front: I am both reviewer and contributor here, so the last four commits have had no second pair of eyes but Greptile's and CI's.

**Verdict: ready to merge.** It fixes the real root cause rather than the reported symptom, and it is a better design than the issues themselves proposed.

---

### What was broken

A scheduled task ran **once per replica** instead of once, for any interval of 375 s or more.

The scheduler runs in every worker of every pod. Three pods with four workers is twelve processes all holding the same timer, and the only thing between that and twelve runs was a lease: a row in `kv_state` keyed on the task name with a TTL.

The lease answers *"did somebody run this recently?"*. It cannot answer *"has the 3 o'clock run happened?"*. That gap is the whole bug, because each replica computed its own fire times from its own boot:

```jac
# before: phase is whenever this process happened to start
return now + datetime.timedelta(seconds=spec.interval);
```

```
interval = 3600, lease ttl capped at 300 s
pod A boots 00:00 -> fires 01:00, 02:00, 03:00
pod B boots 00:10 -> fires 01:10, 02:10, 03:10   (A's lease expired at 01:05)
=> the hourly job runs twice an hour, forever
```

The proof that the TTL is not the cause: cron had the *same* lease with a *shorter* 45 s TTL and was fine, because `0 9 * * *` means 09:00 to everyone. Cron had a shared grid. Interval did not.

### What it does now

**One.** Interval ticks are multiples of the interval since the Unix epoch, so every replica computes the same instants whenever it booted:

```jac
epoch = datetime.datetime(1970, 1, 1, tzinfo=now.tzinfo);
return epoch + step * ((now - epoch) // step + 1);
```

**Two.** A replica claims the tick before running it, in one atomic statement:

```sql
INSERT INTO kv_state (key, value) VALUES ('tick:sched:orders:digest', 3600000000)
ON CONFLICT (key) DO UPDATE SET value = EXCLUDED.value
WHERE kv_state.value::bigint < :tick
RETURNING key
```

One row per task holding a high-water mark. No TTL to guess, no row growth. A replica that is late, respawned, or booted mid-interval asks for a tick already claimed and is refused.

**Three.** Each run is told the time it was scheduled for, via a `ThreadPoolExecutor` subclass, because APScheduler knows each run time but calls the job with no arguments. A run for 09:00 that starts at 09:42 claims **09:00**. Deriving it from the clock instead would be wrong for cron: flooring to the current minute yields instants that are not occurrences.

**Four.** Job rows record the `service` that created them, static claim keys are `sched:<service>:<task>`, and a failed claim skips the tick instead of firing everywhere.

### Who this helps

Anyone running more than one replica, which is every `--scale` deploy. Concretely: an hourly digest stops going out twice an hour, and a mailer or billing job stops firing on every worker of every pod when Postgres hiccups.

---

### Trade-offs and behaviour changes

Four things change for existing deployments. All are in the release note.

| | before | after |
|---|---|---|
| interval phase | whenever the pod booted | on the epoch grid, so `interval=3600` runs on the hour |
| rolling upgrade | n/a | old and new pods key coordination differently and cannot see each other's, so a tick can run twice until the rollout finishes |
| replica stalls, then recovers | ~1 catch-up run | every missed tick inside `misfire_grace_time`, each claimed separately: measured **7 instead of 1** for a 5 s job stalled 30 s |
| database unreachable | ran on every worker | tick is skipped |

The delivery guarantee is now **at most once per tick**. The claim is taken before the run, so a replica that dies between claiming and running has spent that tick and nothing retries it. That is the right default for mail and billing, and it is stated in the docs rather than left to be discovered.

### Limitations

Three, all filed as #9276 rather than stretched into this PR:

1. **A `date` one-shot is lost if its claim fails.** An interval task gets another tick; a one-shot does not. Nothing reads `run_count`/`last_status` to decide anything and there is no `completed` status, so a failed claim erases it silently. Fail-closed is right for interval and cron; `date` needs a durable completion record first.
2. **`_listen_loop` cannot survive a dropped connection.** The redial covers the writers, not the listener thread: it reads `_side_conn` directly and swallows errors at debug level. After a failover `jac_jobs_changed` stops arriving until restart.
3. **The scheduler still runs in every forked worker**, so one tick costs `workers x pods` round trips to authorise one run. The claim makes that correct, not necessary.

Also worth knowing: a replica with a clock far enough ahead claims a future tick and burns the lower ones for everyone, so skew costs a run rather than causing a duplicate. It needs skew larger than a whole interval, and nothing detects it.

---

### Implementation review

The mechanism is right, and the PR explicitly rejects the fix its own issue proposed (per-service collections) with a reason: renaming `scheduled_jobs` per service would strand every persisted job at upgrade. Instead `COALESCE(doc->>'service', :svc) = :svc` keeps pre-existing rows visible to every service, which avoids the trap where adding an owner field silently stops every existing job from firing, listing and deleting.

Two defects I found and fixed on the branch:

- **Fail-closed never un-closed.** The claim runs on `PgStore._side()`, which cached its connection and never redialled, unlike `_run()`. One connection drop and every later claim raised, so the scheduler never fired again for the life of the process. Live: last beat at the instant of the blip, then silence for 17 ticks while the server answered 200 and the database answered SQL. Now `_side()` drops a closed connection and `_side_run()` retries once, reusing the existing `_conn_dead` helper. The fix reaches the leases, rate limits and NOTIFY that share that connection, not only the scheduler.
- **`delete_job` looked its row up with `find_active_one`,** so a row this service owns but has retired answered 404 and could never be removed.

Plus Greptile's catch, which was a real bug in my first fix: the redial replayed `LISTEN` only when `_side()` itself found the connection closed, and skipped it on the `_side_run()` retry path that actually runs. A replica came back serving claims and notifies while deaf to `jac_jobs_changed`.

### Tests

62 -> 76 across six files. Heavier than a fix PR should be, and I think justified: this is a new store primitive, a new ownership dimension and a new executor, not a patch on a covered path. Of the four I added, **none are new tests** — all four extended the existing test on the same path, per the repo's policy.

They are not decorative. Eleven separate reverts of the fix were each tested: the epoch grid, `start_date`, the respawn recompute, both fail-closed sites, `_serves`, the monotonic `WHERE`, the `scheduled_at` plumbing, the connection redial, the `delete_job` lookup and the `LISTEN` replay. **All eleven turn the suites red.**

The one coverage gap worth naming: before the redial fix, *no test exercised a connection that dies and comes back*. The outage tests used a stub that stops raising, and a port nothing listens on. That is precisely the hole the permanent-outage bug walked through, and it is now closed by a test that kills the backend from the publisher so the failure lands mid-call.

### Validation

Live, three workers, `interval=5`, embedded Postgres:

```
[beat] pid=1036772 at=10:13:30.644
[beat] pid=1036772 at=10:13:35.057   <- connections terminated at 10:13:33
[beat] pid=1036769 at=10:13:45.049
...unbroken to 10:15:05
```

Every fire on the 5 s grid, exactly one per tick, and the redial absorbs the dropped connection without even logging a skip. On `main` the same fixture fires at `:47 :52 :57 :02`, boot-phased, and after the same blip fires three times per tick indefinitely.

### Merge note

This is rebased on #9249, and that merge was **not** mechanical. #9249 made `next_interval_run` the single definition shared by `_compute_next_run` and by `schedule_problems` validation, so I moved the epoch grid inside it. Keeping it inline would have left validation checking arithmetic nobody executes, and #9249's "interval too long to schedule" check would have been validating the wrong thing. Worth a glance before merge.

#9250 also touches the static scheduler, so it faces the same exercise.

kugesan1105:
Correction to my summary above, and to the release note: **the cap duplicated from 600 s, not 375 s.** Pushed as `226afac99`.

Duplication needs a phase gap `d` with `ttl <= d <= interval - ttl`, which is empty until the interval reaches twice the cap. Running the real `_lease_ttl` over the range:

```
interval  ttl   window [ttl, interval-ttl]   duplicates?
     374  299   [299, 75]                    no
     375  300   [300, 75]                    no
     400  300   [300, 100]                   no
     599  300   [300, 299]                   no
     600  300   [300, 300]                   yes  (a single phase gap)
     610  300   [300, 310]                   yes  (10 s wide)
    3600  300   [300, 3300]                  yes  (83% of phases)
```

375 s is where the cap *starts to apply*, which is necessary but not sufficient. It also explains why the issue's own 610 s reproduction needed a kill timed to within 10 seconds.

Nothing about the fix changes: the affected range is still every interval a fleet would realistically use, and hourly and daily tasks were duplicating on 83% and 99.3% of phase gaps respectively.

One more thing that reproduction turned up, since it bears on the delivery guarantee this PR sets: on `main` a one-shot `date` task could run **repeatedly**, not just twice. The parent's `next_run` is never advanced, so every respawn for the life of the server re-fires it once the 300 s lease lapses. This PR closes that from both directions, by recomputing fire times on worker start and by refusing an already-claimed tick. The opposite risk it introduces, a one-shot silently *lost* when a claim fails, is #9276 item 1.

PR #9223, opened 2026-09-16
fix(cli): jac check no longer shows a FAILURES banner for a passing file

### LINKED ISSUE #9087: jac check prints a FAILURES banner for a file that passed
## Summary

A file with only warnings passes, and the same output simultaneously says "ok",
"FAILURES", and "1 passed".

## Reproduce

```
$ jac check main.jac
  Checking main.jac...
main.jac ok [100%]

=================================== FAILURES ===================================
___________________________________ main.jac ___________________________________
⚠ warning[W3006]: Unnecessary keyword escape on 'Root' [remove-kwesc]
  --> main.jac:6:20

============================== 1 passed in 0.23s ===============================
```

## Root cause

The FAILURES section is used to display warnings on a passing file. Warnings need
their own heading.

---

## Triage note

Checked jacBuilder#1515 as the report suggested. It is a different bug and is closed as of 2026-09-08: it is about jacBuilder's own `jac_check` tool wrapper counting the tree's documented E1xxx noise into an error total and painting the activity stream red. This one is `jac check`'s own output formatting, where a passing file's warnings are printed under a FAILURES heading. No overlap, so filing rather than commenting there.

Split out of jaseci-labs/jacBuilder#1677, which reported eight findings in one body. Reported by @ThiruvarankanM, tested 2026-09-03. Text above is theirs, unchanged.

Developer-Linus:
## Description

Fixes #9087.

A file with only warnings passed `jac check` but its output showed "ok", a FAILURES banner, and "1 passed" all at once. Verified on main: `jac check` on a file with two unused-variable warnings and no errors prints the file as `ok`, then prints those warnings under a section headed `FAILURES`, then reports `1 passed`.

## Root cause

`failed_results` (the list that feeds the FAILURES banner) was doing two jobs. It held real failures, and it also held any file with warnings, because of a fix from PR #4966 that made warnings-only files show their diagnostics at all (before that fix, a passing file's warnings were silently dropped). That fix reused the only existing plumbing for showing diagnostics, `print_failures`, without giving warnings-only files their own heading. Since `print_failures` unconditionally prints `separator("FAILURES")` for anything in the list, a passing file with warnings ended up under a banner that says it failed, while the progress line and the final summary (both driven by `success`, unaffected by warnings) correctly still said "ok" and counted it as passed.

The same overloaded list had a second effect that the issue does not name. `print_failed_summary` reads `failed_results` too, so on any run where at least one file genuinely failed, a passing file with warnings was also printed under the `failed files` heading, as `warnings_only.jac - 0 errors, 2 warnings`. Whether a passing file appeared there depended on whether an unrelated file failed.

## Fix

- `check()` in `analysis.impl.jac` now routes results into two lists: `failed_results` for files that actually failed, `warning_results` for passing files that have warnings. `print_failed_summary` therefore only ever sees real failures, which fixes the second effect above.
- One `print_section(title, results, nowarn)` in `check_report.jac` / `check_report.impl.jac` replaces `print_failures`, and prints a `WARNINGS` heading for the warnings-only list. `format()` had a third hand-rolled copy of the same banner-plus-separator loop inline in `analysis.impl.jac`; it now calls `print_section` as well.
- `nowarn` is enforced in one place instead of two. `print_section` decides per entry whether there is anything visible to show and prints no heading at all when there is not, so `--nowarn` produces no empty WARNINGS banner.
- Extracted the shared per-diagnostic printing logic (the dict / `pretty` / `text` fallback and the guide hint) into one `print_diag(d, is_error)` helper, since that logic was duplicated across the errors loop and the warnings loop.
- Ordered the output as FAILURES, then `failed files`, then WARNINGS, so the summary immediately follows the failure detail it summarizes instead of being split by a section about files that passed.

## Tests

- Extended the existing test `jac check shows warnings for files with warnings only` in `test_cli.jac` to assert that `WARNINGS` appears, that `FAILURES` does not, and that `ok` still shows.
- Added one test, `jac check on a mixed run never lists a passing file under failed files` (one erroring fixture plus one warnings-only fixture). It pins the second half of the bug: the passing file must never produce a `print_failed_summary` line, which nothing asserted before. Test count in the file goes from 96 to 97.

## Validation

Before, on main, `jac check` on the `warnings_only.jac` fixture:
```
warnings_only.jac ok [100%]

=================================== FAILURES ===================================
__________________ warnings_only.jac ___________________
warning[W2003]: 'unused_var' is defined but never used
...
============================== 1 passed in 18.58s ===============================
```

After, on this branch, same fixture:
```
warnings_only.jac ok [100%]

=================================== WARNINGS ===================================
__________________ warnings_only.jac ___________________
warning[W2003]: 'unused_var' is defined but never used
...
============================== 1 passed in 1.52s ===============================
```

On a mixed run (one erroring file, one warnings-only file), the `failed files` summary listed the passing file on main and lists only the erroring file on this branch. A failing file that also has warnings still prints those warnings inline in its own FAILURES block, so no diagnostic is lost.

`jac test tests/language/test_cli.jac`: 97 passed. `jac check` clean on all four changed source and test files (only pre-existing, unrelated too-many-params warnings on `analysis.impl.jac`). `jac fmt` no diff.

---

*Description edited by a maintainer before merge: the Fix, Tests and Validation sections had gone stale against the review revision in 956e34ac7 (they still described a separate `print_warnings()` rather than the consolidated `print_section`, and stated no new test and 96 tests). No code change.*


jac/tests/language/test_cli.jac
```diff
@@ -2157,16 +2157,20 @@ test "jac check shows warnings for files with warnings only" {
     # Should exit with code 0 (warnings don't fail the check)
     assert result == 0 , "check should exit with code 0 for warnings only";
 
-    # Should show FAILURES section with warnings
-    assert "FAILURES" in combined , "should display FAILURES section for warnings";
+    # Should show a WARNINGS section, not a FAILURES banner, since the file passed
+    assert "WARNINGS" in combined , "should display WARNINGS section for warnings";
+    assert "FAILURES" not in combined , (
```
MusabMahmoodh, line 2162: This covers the warnings-only run, which is the half the issue showed. The half it does not cover is the mixed run, and that is where the old routing did its other damage: with one erroring file and one warnings-only file, `failed_files > 0`, so `print_failed_summary` ran over a `failed_results` that contained the passing file too, and it printed `warnings_only.jac - 0 errors, 2 warnings` under the `failed files` heading.

Your change fixes that as well, and nothing asserts it. A second case with two fixtures - one with a real error, one warnings-only - asserting that the passing file's name does not appear after `failed files`, would pin the part of this that is easiest to undo later.

jac/jaclang/cli/impl/check_report.impl.jac
```diff
@@ -33,27 +44,30 @@ impl print_failures(
...
-                    console.warning(str(w));
-                }
+                print_diag(w, is_error=False);
             }
         }
         console.print("");
     }
 }
 
+impl print_warnings(warning_results: list[dict[str, any]]) -> None {
```
MusabMahmoodh: This is `print_failures` minus the error loop, and `format` already keeps a third hand-rolled copy of the same banner-plus-separator loop inline at analysis.impl.jac:1177-1187. You pulled `print_diag` out for exactly this reason; the section body is the other half of the same duplicate.

One `_print_section(title, results, nowarn)` that prints errors when the entry has them would cover both of these and give `format` something to call. Would also put `nowarn` back in one place - right now it is enforced at the collection site for warnings-only files (analysis.impl.jac:491) and inside `print_failures` for failed ones.

jac/jaclang/cli/commands/impl/analysis.impl.jac
```diff
@@ -505,6 +501,9 @@ impl check(
     if print_errs and len(failed_results) > 0 {
         print_failures(failed_results, nowarn);
     }
+    if print_errs and len(warning_results) > 0 {
+        print_warnings(warning_results);
```
MusabMahmoodh: Ordering question, not a blocker: on a mixed run this puts WARNINGS between the failure detail and the `failed files` summary, so the failure story is split by a block about files that passed. Reading `failed files` right after the failures it summarizes is easier. Was the current order deliberate, or is it just where the call landed?

MusabMahmoodh (approved):
Right fix in the right place: the defect was that one list meant two things, and splitting the list is the fix rather than special-casing the banner text. `print_diag` is a good extraction on the way past.

Checked the other two consumers before approving - `lint` (analysis.impl.jac:516) prints warnings straight through `console.warning`, and `format` builds its own errors-only `failed_results`, so neither carries this routing. Nothing else to fix behind it.

Three notes, the first is the one worth acting on.

jac/jaclang/cli/commands/impl/analysis.impl.jac
Developer-Linus: Thank you for pointing this out. It was not deliberate, just where the call landed. I moved the failed files summary to print right after the FAILURES section and before WARNINGS, so it now reads immediately after the failures it summarizes. Pushed in 956e34ac7.

Issue #9140, opened 2026-09-11
Jac Scale: the serve manifest has no format stamp, so a .jab sealed by a pre-#9088 driver boots on 0.37.12 into a KeyError instead of a refusal naming the generation skew


MusabMahmoodh:
## Summary

#9088 changed the shape of `.jac/serve_manifest.json` from one flat dict per image to a `module_manifests` table keyed by project-relative path, and its release note says "rebuild existing artifacts". Nothing in the file or the loader enforces that. A `.jab` sealed by a pre-#9088 driver (every jachammer deploy today runs builder-sv's own jac-scale, 0.34.x, and the 0.37.11 line never wrote the file at all) that lands on a 0.37.12 pod fails inside `_prepare_sealed` with a `KeyError: 'module_manifests'` or a `FileNotFoundError`, neither of which says "your bundle and your runtime are from different generations". The `_precompiled/MANIFEST.json` next to it already carries `format` and `MANIFEST_FORMATS_ACCEPTED`; the serving payload is the one artifact in the image without a stamp.

## Provenance

- Flat serve manifest: #7258 (2525695f89, 2026-07-07), `build_serve_manifest` in what is now `dist/seal.jac`.
- `module_manifests` table and the loader that requires it: #9088 (8f452cc415, 2026-09-10, 0.37.12), `dist/seal.jac:91-97` writer and `compiler/driver/application.jac:226` reader.
- The generation-skew class and the refusal sentence: #8753 (closed), which stamped the pod/driver relationship but not this file.

## Reproduction (recipe from source; not executed on a cluster)

```
# seal with a pre-9088 binary
jac-0.37.11 build --as jab           # or any jachammer deploy: builder-sv seals with its own jac
tar xzf dist/app.jab -C /tmp/pod && cd /tmp/pod
# boot with a post-9088 binary
jac-0.37.12 run --serve main.jac
```

Expected: a refusal that names both generations and says to rebuild the bundle with the deploying jac. Actual: `Loading sealed application artifacts`, then `Error loading main.jac: 'module_manifests'` (0.37.11 jab built with `jac build --as jab`, which did write the flat file) or `Error loading main.jac: [Errno 2] No such file or directory: '.../.jac/serve_manifest.json'` (any 0.37.11 scale deploy, which never wrote it). Exit 1, crash-loop.

## Code evidence

Writer, `jac/jaclang/dist/seal.jac`:

```
91:  result = dict(prepared.manifests[str(entry_file.resolve())]);
[...]

Issue #9139, opened 2026-09-11
Jac Scale: prepare_application gives no reason when a present _precompiled image is not booted from, so a pod that recompiles or crashes at boot cannot say which sealed check missed


MusabMahmoodh:
## Summary

`prepare_application` decides sealed-vs-source with one call, `sealed_root_for(entry)`, and prints exactly one line ("Loading sealed application artifacts") when it takes the sealed branch. When a `_precompiled/MANIFEST.json` is present above the entry but the branch is NOT taken, nothing is printed and the pod proceeds into a full source compile. When the branch IS taken and `_prepare_sealed` fails, the pod gets a bare exception text through the generic "Error loading" handler. In both cases the operator of a pod that ships 184 `.jir` files has no way to learn which sealed check missed. #9136 spent a P1 investigation on exactly this silence.

## Provenance

#9088 (8f452cc415, 2026-09-10, 0.37.12) introduced `prepare_application` and `_prepare_sealed`. Before it, the equivalent decision lived in `ModuleIntrospector` (`server.impl.jac:471-500` at 8f452cc415^) and was equally silent: `serve_manifest_for` returned `None` and the code fell straight into `Jac.get_program().compile(mod_path)`. The class predates #9088; #9088 made the decision a single point, which is what makes it cheap to instrument now.

## Reproduction (recipe from source; not executed on a cluster)

Take any sealed bundle (`jac build --as jab`, extract it) and make one of the sealed preconditions false:

- rename the entry so its relative path is no longer a key in `_precompiled/MANIFEST.json["modules"]` (the #7957 membership rule), then `jac run --serve <entry>`: prints `Preparing application...`, then `Compiling server modules and placement metadata...`. No mention that a sealed image was found and rejected.
- delete `.jac/serve_manifest.json`, then `jac run --serve <entry>`: prints `Loading sealed application artifacts`, then `Error loading <entry>: [Errno 2] No such file or directory ...`. The path in the errno is the only clue.
- delete one `.jir` the closure needs: `Error loading <entry>: Sealed application is missing executable artifact: <path>`. This one is fine; it is the shape the other two should have.

## Code evidence

`jac/jaclang/compiler/driver/application.jac`:

```
308:  import from jaclang.server.sealed_serve { sealed_root_for }
309:  image_root = sealed_root_for(`entry);
310:  if image_root is not None {
311:      console.print('  Loading sealed application artifacts');
312:      return _prepare_sealed(`entry, program, image_root, entries);
[...]

Issue #9138, opened 2026-09-11
Jac Scale: the pack ships a .jab without .jac/serve_manifest.json when the bake produces none, so the pod fails at boot instead of the deploy failing on the host


MusabMahmoodh:
## Summary

`pack_jab` adds `.jac/serve_manifest.json` to the `.jab` only when the file happens to exist in the seal stage, and the bake step that is supposed to produce it returns success without writing anything when it cannot resolve the app entry. A deploy can therefore complete, ship the bundle to every pod, and only the pod discovers that the serving metadata is missing. On jac 0.37.12 that pod crash-loops (`_prepare_sealed` reads the file unguarded); on 0.37.11 it silently recompiled the whole server side at boot, which is the 7 minute / 4.5Gi boot measured in #9136. The host has every fact it needs to refuse the pack instead.

## Provenance

- The tar guard `if serve_manifest.is_file()` and the bake step both arrived in #9088 (8f452cc415, 2026-09-10, released as 0.37.12). Before that the scale path never produced the file at all, which is why every 0.37.11 pod recompiles (#9136).
- `bake_serve_payloads` returning `True` when `resolve_entry_file` is `None` dates to #7258 (2525695f89, 2026-07-07, universal `.jab`), where the only caller was `jac build --as jab` and a warning to the terminal was adequate.

## Reproduction (recipe from source; not yet executed on a cluster, pod evidence is #9136's)

Two ways to reach a bundle with no serving metadata on upstream/main:

1. Delete the manifest between bake and pack. `bake_serve_payloads(..., bake_client=False)` writes it in a subprocess, `precompile --seal` runs next in another subprocess, and the tar step reads the stage afterwards. Any failure or cleanup in between leaves `_precompiled/MANIFEST.json` present and the serve manifest absent, and `pack_jab` still returns a `.jab`.
2. Make `resolve_entry_file` return `None`: a workspace whose resolved app has an `entry-point` that does not resolve to a file in the staged tree. `resolve_entry_file` prints "sealing without serve payloads" to the seal subprocess's stdout, `bake_serve_payloads` returns `True`, `_precompile_app` sees exit 0 and continues.

Then boot the extracted bundle the way the pod does (`manifest_builder._pod_entrypoint`):

```
tar xzf app.jab -C /tmp/pod && cd /tmp/pod
JAC_SV_FILE=web/main.jac jac run --serve "$JAC_SV_FILE"
```

[...]

PR #9228, opened 2026-09-16
fix(scale): seal the app the deploy serves instead of the default app (#9136)

### LINKED ISSUE #9136: Jac Scale: pods ignore the shipped sealed jab and recompile from source at boot; serve_manifest.json never reaches the pod
## Summary

A Kubernetes deploy seals the app on the build host and ships the result to every pod, but the pod never loads
it. `/app/_precompiled/` arrives with all the `.jir` units and a manifest that lists the entry, yet boot goes
down the full source-compile path and the log never prints "Loading sealed application artifacts". Every replica
therefore recompiles the whole server side at startup, which on jaclang.org is about 7 minutes and 4.5Gi per
pod, repeated on every restart and every HPA scale-up. The seal work on the builder is wasted.

## Environment

| | |
|---|---|
| jac | pod image `jaseci/jaclang:0.37.11` (live jaclang.org); gate code read at v0.37.12, unchanged in the relevant files |
| Deploy | jac-scale Kubernetes target via jachammer, pinned deploy |
| App | `jac/examples/jaclang_org`, 5-app workspace, entry `web/main.jac` |
| Pod | `web-deployment-*` in `jb-70d63706-jaclang-website`, command `jac start web/main.jac`, `JAC_SV_FILE=web/main.jac` |

What the builder ships (from `scale/injector/bundle.jac`): `_precompile_app` runs `precompile <stage> --seal`,
then the bundle tar adds `_precompiled/`, logs `Sealed N precompiled app-module artifacts into the .jab`, and
adds `.jac/serve_manifest.json` **only if it exists** (`bundle.jac:1083-1089`).

What is on the live pod:

```
$ find /app -name "*.jir" | wc -l
184
$ ls /app/_precompiled/MANIFEST.json
/app/_precompiled/MANIFEST.json
$ grep -c '"web/main.jac"' /app/_precompiled/MANIFEST.json
3                                   # the entry IS a module key
$ ls /app/.jac/serve_manifest.json
ls: cannot access '/app/.jac/serve_manifest.json': No such file or directory
```

What the pod logged at boot:

```
Jac is analyzing your project. Please be patient...
  ✔ Client bundle ready (0.1s)
  Server ready
```

```
$ kubectl logs <web pod> | grep -c "Loading sealed application artifacts"
0
```

Time from container start to Ready: about 7 minutes. Memory at steady state after boot: 4.5Gi (it OOM-looped on
the 4Gi default until the limit was raised). The client bundle itself loads in 0.1s because it is prebuilt; the
7 minutes is the server-module compile.

For comparison, the same app sealed locally (`jac build --as jab`, 97s, 8.3MB) and run with
`jac run --serve dist/jaclang_org.jab` reaches Ready in **2 seconds**.

## Expected

`prepare_application` (`compiler/driver/application.jac:305`) calls `sealed_root_for(entry)`
(`server/sealed_serve.jac:45`), which walks up for `_precompiled/MANIFEST.json` and requires the entry's relative
path to be a module key. Both hold on the pod. It should then print "Loading sealed application artifacts" and
take `_prepare_sealed`, which loads bytecode without compiling.

## Actual

The sealed branch is not taken, or is taken and abandoned. `_prepare_sealed` (`application.jac:190`)
unconditionally reads `SERVE_MANIFEST_REL` = `.jac/serve_manifest.json` (`application.jac:216`), which the
bundle ships only when the seal step produced it, and the pod does not have it. Whether the miss happens in
`sealed_root_for`, in `_prepare_sealed` raising on the missing file, or in a caller that swallows that and falls
back, the observable result is the same: no sealed load, full compile, no diagnostic saying why.

The boot output gives no hint that a seal was present and ignored. "Jac is analyzing your project" is printed
either way.

## Impact

Every jac-scale Kubernetes deploy pays a full compile per replica at boot for nothing. On jaclang.org that is
about 7 minutes and 4.5Gi per pod, three pods, plus again on each restart, rollout, or scale-up. It also forces a
memory limit far above what serving needs (#1658 on the jacBuilder side), and it makes the HPA counterproductive
because each new replica spends its first minutes compiling rather than serving.

It also undercuts the obvious fix for the builder-side cost (jac#9135, jacBuilder#1797): building the jab
elsewhere is pointless while the pod does not use it.

## Workaround (verified)

None on the pod side. Running the jab directly (`jac run --serve x.jab`) works and boots in 2s, but the
Kubernetes target's pod command is `jac start <entry>`, not the jab.

## Suggested fix

- Make `_precompile_app` always emit `.jac/serve_manifest.json` (it already runs `bake_serve_payloads` to
  produce it; if that step can succeed without writing the file, that is the bug) and make the bundler fail the
  deploy if it is missing rather than silently shipping without it.
- In `prepare_application`, when `_precompiled/MANIFEST.json` is present but the sealed load is not taken, log
  the reason at boot ("sealed artifacts present but serve manifest missing" or whatever the actual miss is).
  Today the fallback is silent.
- A regression test in `jaclang/scale/tests/deploy` that builds a bundle for a workspace fixture, boots it the
  way the pod does (`jac start <entry>` in the extracted bundle dir), and asserts the "Loading sealed
  application artifacts" line and that no module was compiled.

Related: jac#9135 (why the builder-side compile is so slow), jacBuilder#1797, jacBuilder#1658, #8981 (server
rewrite; does not touch this path).

akindu-k:
### Summary 

A real example from jacBuilder: three apps. `builder_sv` is the default-app (a service), `jac_coder_sv` is another service, and `main` is the web-app being deployed.

image

### What was wrong

`jac scale deploy` compiles the app **ahead of time** on the deploy machine. This step is called "sealing". The compiled copy ships to every pod, so a pod can start in seconds instead of compiling the app itself.

In a workspace (a project with several apps), the deploy always sealed the **`default-app`**, even when you deployed a different app. The pod running the deployed app found no compiled copy of it, so it compiled everything at startup.

Example from jacBuilder, which reported it in #9136:

```toml
[project]
default-app = "builder_sv"   # a service: this is what got sealed

[apps.main]                  # the web-app being deployed: this was NOT sealed
kind = "web-app"
```

The `main` pod compiled the whole app at boot and was OOMKilled at its 12Gi memory limit.

### Why it happened

The deploy knew which app it was serving: the fleet's `root_app`. That name was never passed to the two sealing steps, so both fell back to `default-app`:

1. **Bake**: `bake_serve_payloads(...)` writes `.jac/serve_manifest.json`, the list of the app's endpoints the server reads at startup.
2. **Seal**: `jac precompile <stage> --seal` writes `_precompiled/`, the compiled files and their index.

### What this PR changes

All the code changes are in `jac/jaclang/scale/injector/bundle.jac`:

- **New helper `sealed_app_name(project_dir, fleet)`** returns the served app for a workspace deploy. For a single-app project it returns `""`, which keeps today's behavior.
- **`_precompile_app` takes an `app` argument** and passes it to both steps: `app=` for the bake and `--app=` for the seal, including the legacy seal fallback.
- **`pack_jab` passes the served app** into `_precompile_app`.

This follows the same rule `jac build --as jab --app <name>` already uses. Nothing changes on the pod side.

**Behavior that stays the same:**
- `jac scale deploy` with no app named still seals the default app.
- A single-app project gets exactly the same seal flags as before.
- Service apps are still sealed alongside the served app.
- Older pod binaries still work: `--app` and `app=` both shipped in 0.37.5, the same release as the bake step this path already requires.

### 1. Automated test (extended, not added)

I extended the existing test in `scale/tests/microservices/test_app_precompile.jac` into a table with two cases. It is renamed from `a real pod binary seals and serves an entry named app` to `a real pod binary seals the app each pod serves and boots it from the seal`.

| Case | Project | Starts |
|---|---|---|
| an entry named app (the existing case) | single-app service | `app.jac` |
| a web-app that is not the default app (**new**) | workspace, `default-app = "svc"`, deploying `web` | `web.jac` |

For each case the test packs a real `.jab` with the real `jac` binary, starts the app from it like a pod does, and checks:
- the deployed app's compiled file and its functions are in the bundle
- the log **contains** `Loading sealed application artifacts`
- the log **does not contain** `Compiling server modules`

| | Result |
|---|---|
| Without the fix | **Fails**: `KeyError: 'hello'` (the serve manifest was built for `svc`, not `web`) |
| With the fix | **Passes** |
| Tests in the file | 17 before, 17 after |
| Related bundle tests (`test_seed_determinism`, `test_pvc_injector`, `test_bundle_jacignore`, `test_fat_bundle`, `test_binary_injector`) | 83 passed |

### 2. Real Kubernetes cluster

I ran the real command, `jac scale deploy web.jac`, on a MicroK8s cluster in a temporary workflow on my fork. The workspace had `default-app = "svc"` and two apps, `web` and `svc`. It deployed **twice**, first with the deploy code from `main` and then with this PR, and read each pod's startup log.

Run: https://github.com/akindu-k/jaseci/actions/runs/35099101554

| Deploy code | `web` pod (the deployed app) | `svc` pod |
|---|---|---|
| `main` | `Compiling server modules and placement metadata...` (bug) | `Loading sealed application artifacts` |
| this PR | `Loading sealed application artifacts` (**fixed**) | `Loading sealed application artifacts` |

Both deploys rolled out with 0 restarts.

<details>
<summary>Pod startup logs from the run</summary>

```
=== BEFORE: deploy code from main ===
== web-deployment
Preparing application...
  Compiling server modules and placement metadata...
  Server ready
== svc-deployment
Preparing application...
  Loading sealed application artifacts
  Server ready (no client)

=== AFTER: deploy code with this PR ===
== web-deployment
Preparing application...
  Loading sealed application artifacts
  Server ready
== svc-deployment
Preparing application...
  Loading sealed application artifacts
  Server ready (no client)

=== verdict ===
OK before: web pod compiled at boot (bug reproduced)
OK after: web pod booted from the seal
OK after: svc pod booted from the seal
```

</details>

## **Not in this PR**

These follow-ups from #9136 are tracked separately:
- #9139: when a pod does not use the sealed image, log why
- #9138: fail the deploy early if the serve manifest is missing
- #9140: add a format version to the serve manifest

Fixes #9136


MalithaPrabhashana (commented):
I verified locally that this fixes #9136 without breaking the service pods. Details below, plus one non-blocking test suggestion inline.

**Setup.** I made worktrees at this head (`da7926fb5`) and at `main` (`25f948c07`, which this head already contains) and ran from each checkout root in jac dev mode. Both the dev-mode banner and the probe's `bundle.jac` path confirm each run loaded its own checkout. The pod binary was the rolling dev build (`e255ec5c4`), on macOS arm64.

**1. The new test pins the bug.** This PR's `test_app_precompile.jac`, run on both trees:

| Test | `main` | this PR |
| --- | --- | --- |
| a real pod binary seals the app each pod serves and boots it from the seal | fails: `KeyError: 'hello'` | passes |

**2. Both pods boot from the seal.** A probe builds the #9136 layout (`default-app = "svc"`, served app `web`, fleet from `fleet_from_config(config, "web")`) and packs a real `.jab` with `pack_jab`. It then extracts the bundle and boots each entry with `jac run --serve --no-client --faux`:

| | `main` | this PR |
| --- | --- | --- |
| fleet root / members | `web` / `svc`, `web` | `web` / `svc`, `web` |
| `"web.jac"` in `_precompiled/MANIFEST.json` | 0 | 3 |
| serve manifest `module_manifests` | `svc.jac` | `svc.jac`, `web.jac` |
| boot `web.jac` | `Compiling server modules and placement metadata...` | `Loading sealed application artifacts` |
| boot `svc.jac` | `Loading sealed application artifacts` | `Loading sealed application artifacts` |

**3. Holes I checked; none found.**
- **Served app kinds.** A fleet can only serve an app from `serving_apps()`, which keeps only kinds with `has_server`: `web-app`, `service` and `service-mesh`. So `_application_sources` always adds the service apps, and no servable kind drops the services from the seal.
- **Fleet reaches the seal.** For every workspace, `kubernetes_realizer._resolve_fleet` builds the fleet with `fleet_from_config(cfg, self.served_app)`, and it reaches `pack_jab` through `binary_injector` and `PvcInjector`. A single-app project gets no fleet, so `sealed_app_name` returns `""` and the seal flags are unchanged.
- **Older seal binaries.** `jac precompile` exists since v0.37.2. `precompile_bytecode --app` and `bake_serve_payloads(..., app=...)` exist since v0.37.5, the first release with workspaces. Since `app` is only non-empty for a workspace, the legacy sealer fallback never receives `--app`.

**4. The two red checks do not come from this PR.**
- **`jac-check`.** Its 3 errors (`Type "Span" has no attribute "set_attribute"` in `scale/server/extensions.jac` and `microservice_gateway.impl.jac`) are in files this PR does not touch. The same 3 errors also fail `jac-check` on `refactor/byllm-tests-pr-d` about 40 minutes later (run 35110433378).
- **`test-jac-pack-smoke`.** Its `production-build` hit the 180s wall budget using 146.28s of CPU. On `main` at `25f948c07` the same step passed at 134.20s wall with 163.59s of CPU. Less CPU but more wall time points to a slower runner, and `quickstart-ready` slowed the same way (70.05s on `main`, 145.14s here). A rerun should clear it.

**Not covered.**
- **Linux and a real cluster.** I did not deploy to one; your MicroK8s run covers that.
- **Two local-only failures.** On macOS, two other tests in the same file fail identically on `main` and on this PR: `sealing includes bytecode imports and serving exports despite client placement`, a `/private/var` vs `/var` temp-path `ValueError`, and `an explicit app seal follows its entry and excludes unrelated apps`, which returns an empty source set. Both pass in CI's `test-scale (microservices)` job (348 passed).

jac/jaclang/scale/tests/microservices/test_app_precompile.jac
```diff
@@ -261,57 +261,102 @@ test "staging a project cleans up after itself" {
...
+            assert any([n.endswith(module) for n in pre]) , (label, module, pre);
+        }
+        with tempfile.TemporaryDirectory(prefix="jac-sealed-entry-") as tmp {
+            with tarfile.open(fileobj=io.BytesIO(raw), mode="r:gz") as archive {
+                archive.extractall(tmp, filter="data");
+            }
+            env = dict(os.environ);
+            env["JAC_NO_DEV_SOURCE"] = "1";
+            result = subprocess.run(
+                [str(jac_bin), "run", "--serve", "--no-client", "--faux", served_entry],
```
MalithaPrabhashana: Non-blocking suggestion: in the web-app case this boots only `web.jac`.

The PR's claim that service apps are still sealed alongside the served app is pinned here only by `svc.jir` being in `_precompiled/`. That file is not what lets a service pod boot from the seal. The pod also needs `svc.jac` in the serve manifest's `module_manifests`, and nothing here asserts that.

It does hold on this branch. I packed the same layout with `pack_jab` and booted both entries with these flags:

```
serve_manifest module_manifests: ['svc.jac', 'web.jac']
boot web.jac: rc=0 ['Loading sealed application artifacts']
boot svc.jac: rc=0 ['Loading sealed application artifacts']
```

Booting the service entry as well would keep that from regressing unnoticed, for example:

```jac
boots = [served_entry] + (["svc.jac"] if served else []);
for boot_file in boots {
    result = subprocess.run(
        [str(jac_bin), "run", "--serve", "--no-client", "--faux", boot_file],
        cwd=tmp,
        env=env,
        capture_output=True,
        text=True,
        timeout=90
    );
    assert "Loading sealed application artifacts" in result.stdout , (label, boot_file, result.stdout);
    assert "Compiling server modules" not in result.stdout , (label, boot_file, result.stdout);
}
```

It would also help to assert that `module_manifests` contains both `served_entry` and `svc.jac`, next to the `func_access` check above.

akindu-k: Thanks, applied in d5ca10353.

Each case in the table now lists the entries it boots, with the endpoint each should serve:

- an entry named app: `app.jac` → `/walker/echo`
- a web-app that is not the default app: `web.jac` → `/function/hello` and `svc.jac` → `/function/ping`

For every one of those entries, the test now asserts that:
- it is listed in the serve manifest's `module_manifests` (next to the `func_access` check)
- its `.jir` is in `_precompiled/`
- it boots from the extracted bundle with `Loading sealed application artifacts` and without `Compiling server modules`

The single-app case lists only `app.jac`, because its `module_manifests` holds only `app.jac` (`svc.jac` there is an imported library, not a bootable app).

Checked locally:

| | Result |
|---|---|
| This branch | 17 passed (count unchanged) |
| `bundle.jac` from `main` | fails, `KeyError: 'hello'` |
| `service_apps` forced to return nothing (a service dropping out of the seal) | fails, `('a web-app that is not the default app', 'svc.jac', ['web.jac'])` |

MalithaPrabhashana (approved):
Approving. I validated the behavior locally: this fixes #9136, and the service apps are still sealed and boot from the seal.

Setup: this branch at d5ca10353, a real pod binary, and `JAC_TEST_STRICT=1` so the test cannot skip.

| Check | Result |
| --- | --- |
| Served-app seal test on this branch | 1 passed (web.jac and svc.jac both boot with `Loading sealed application artifacts`) |
| Same test with `bundle.jac` from main (4fc489576), so without the fix | fails, `KeyError: 'hello'` |
| `svc.jac` removed from the serve manifest before the seal | fails, `('a web-app that is not the default app', 'svc.jac', ['web.jac'])` |
| CI `test-scale (microservices)` | 348 passed, 0 skipped |

Thanks for applying the test suggestion.

The red `test-jac-pack-smoke` is not from this PR: the same `landing #top never appeared` failure is on main at 8d69f6e92.

MalithaPrabhashana:
I also checked the jacBuilder case from #9136 on this branch (c0f79915c), not just the test.

Workspace with jacBuilder's shape: `default-app = "builder_sv"` (service), `jac_coder_sv` (service), `main` (web-app at `/`). Sealed through the same calls `jac scale deploy main.jac` makes (`_resolve_deploy_target` -> `fleet_from_config` -> `pack_jab`), then each app booted from the extracted jab with the pod binary.

| | main (d26918bd7) | this PR (c0f79915c) |
| --- | --- | --- |
| served app / fleet root | `main` | `main` |
| serve manifest apps | `builder_sv`, `jac_coder_sv` | `builder_sv`, `jac_coder_sv`, `main` |
| `main` boot | compiles at boot | `Loading sealed application artifacts`, no compile |
| `builder_sv` boot | sealed | sealed |
| `jac_coder_sv` boot | sealed | sealed |

So the served web-app that is not `default-app` is now sealed, and the services stay sealed.

<details>
<summary>Probe</summary>

`ws/jac.toml`:

```toml
[project]
name = "jbshape"
version = "0.1.0"
default-app = "builder_sv"

[apps.main]
kind = "web-app"
entry-point = "main"
route = "/"

[apps.builder_sv]
kind = "service"
entry-point = "builder_sv"
route = "/api/builder_sv"

[apps.jac_coder_sv]
kind = "service"
entry-point = "jac_coder_sv"
route = "/api/jac_coder_sv"
```

`main.jac`, `builder_sv.jac` and `jac_coder_sv.jac` each hold one `def:pub` function.

`probe.jac`, run as `jac run --no-serve probe.jac -- <ws> <pod jac binary>`:

```jac
"""Seal a jacBuilder-shaped workspace the way `jac scale deploy main.jac` does, then boot every pod from the jab."""

import io;
import json;
import os;
import subprocess;
import sys;
import tarfile;
import tempfile;
import from pathlib { Path }

with entry {
    ws = sys.argv[-2];
    pod = sys.argv[-1];
    os.chdir(ws);
    import from jaclang.scale.plugin { _resolve_deploy_target }
    import from jaclang.project.config { get_config_for_path }
    import from jaclang.scale.runtime.fleet { fleet_from_config }
    import from jaclang.scale.injector.bundle { pack_jab }
    (app_file, served_app, problem) = _resolve_deploy_target("main.jac");
    print(f"RESOLVE served_app={served_app!r} problem={problem!r}");
    cfg_or_none = get_config_for_path(Path(ws));
    assert cfg_or_none is not None , "no jac.toml in the workspace";
    cfg: any = cfg_or_none;
    root_dir = str(cfg.project_root) if cfg.project_root else ws;
    fleet = fleet_from_config(cfg, served_app, project_root=root_dir, relative_to=ws);
    print(f"FLEET root_app={fleet.root_app} members={sorted(fleet.members)}");
    raw = pack_jab(ws, Path(pod).read_bytes(), "", fleet);
    with tempfile.TemporaryDirectory(prefix="jbshape-sealed-") as tmp {
        with tarfile.open(fileobj=io.BytesIO(raw), mode="r:gz") as archive {
            archive.extractall(tmp, filter="data");
        }
        manifest_path = Path(tmp) / ".jac" / "serve_manifest.json";
        mods = sorted(json.loads(manifest_path.read_text())["module_manifests"]) if manifest_path.is_file() else "MISSING";
        jirs = sorted(str(p.name) for p in (Path(tmp) / "_precompiled").rglob("*.jir") if p.stem in ("main", "builder_sv", "jac_coder_sv"));
        print(f"SEAL module_manifests={mods} app_jirs={jirs}");
        env = dict(os.environ);
        env["JAC_NO_DEV_SOURCE"] = "1";
        for entry_file in ["main.jac", "builder_sv.jac", "jac_coder_sv.jac"] {
            res = subprocess.run(
                [pod, "run", "--serve", "--no-client", "--faux", entry_file],
                cwd=tmp,
                env=env,
                capture_output=True,
                text=True,
                timeout=180
            );
            out = res.stdout + res.stderr;
            sealed = "Loading sealed application artifacts" in out;
            compiled = "Compiling server modules" in out;
            print(f"BOOT {entry_file} exit={res.returncode} sealed_load={sealed} compiled_at_boot={compiled}");
        }
    }
}
```

</details>

PR #9229, opened 2026-09-16
refactor(opcode): table-drive popped/pushed classifiers (Stack 2 on #9227)

chess10kp:
## Stack 2 / N — opcode stack-effect tables

**Base:** `simplify/massive-cleanup` (#9227) → **Head:** `simplify/stack-2-opcode`

Table-drives `_opcode_num_popped`/`_opcode_num_pushed` constant-return cases that were 400+ `if op == X {return N}` branches into 6 set tables (`_POPPED_0..3_SET`, `_PUSHED_0..2_SET`) with membership checks. Dynamic `oparg`-dependent returns (`oparg`, `oparg*2`, `2+(oparg&1)`) stay as-is — isolated for clarity.

- `opcode_meta.jac`: **1979→803** (−1176) this commit, **3591→803** cumulative vs main (−2.8k)
- No semantic change, `jac check` passes
- `if-if-if` routing → table lookup, god file shrinking

Part of massive simplification toward 30% LOC drop.

Closes #9227 follow-up

PR #9231, opened 2026-09-16
refactor(scale): dedupe k8s read-or-create/delete-if-exists (Stack 3)

chess10kp:
## Stack 3 — k8s deploy dedup (`apply_k8s_resource`)

**Stacks on** #9227 / #9229 (GitHub base must be `main`).

The kubernetes deployers repeated one API-call shape ~60 times:

```jac
try {
    core_v1.read_namespaced_X(name=..., namespace=ns);
    core_v1.replace_namespaced_X(name=..., namespace=ns, body=body);
} except ApiException as e {
    if e.status == 404 { core_v1.create_namespaced_X(namespace=ns, body=body); }
    else { raise; }
}
```

- New shared helper **`apply_k8s_resource(api, kind, name, body, namespace?, replace?)`** in `kubernetes_utils` — kubectl-apply semantics via the same read/replace/create method-stem convention the python client already uses; `replace=False` covers create-only resources (services, PVCs, service accounts).
- **43 read-or-create blocks** (25 in `monitoring.jac`, 2 in `target.jac`, plus `ensure_service`) → one-liners.
- **19 inline delete-if-exists blocks** now route through the existing `delete_if_exists` helper, extended for cluster-scoped resources (empty namespace → no namespace kwarg).
- Two deliberately different shapes left untouched: poll-until-404 namespace wait, create-only PVC guard.

`monitoring.jac` **2396→1964**. Net **−419** (94+/513−). No semantic change; `jac check` green on all touched files.

Part of the massive simplification stack (target: −30% LOC).


jac/jaclang/scale/deploy/target/kubernetes/utils/kubernetes_utils.impl.jac
```diff
@@ -456,14 +456,50 @@ impl delete_if_exists(
...
+    namespace: (str | None) = None,
+    replace: bool = True
+) -> None {
+    # kubectl-apply semantics for the python k8s client: read the resource,
+    # replace its manifest when present, create it on 404. `kind` is the API
+    # method stem ("config_map", "cluster_role", ...); namespaced resources
+    # pass `namespace`, cluster-scoped leave it None. `replace=False` makes
+    # this create-only for resources k8s forbids updating in place (services,
+    # PVCs, service accounts).
+    stem = ("namespaced_" + kind) if namespace else kind;
```
MusabMahmoodh, line 486: The comment says `kind` is the method stem ("config_map", "cluster_role"), but every caller passes the PascalCase kind, so this builds `read_namespaced_ConfigMap`, `read_ClusterRole`, `read_namespaced_DaemonSet`. The client is snake_case only, so `getattr` on the next line raises AttributeError outside the try. Normalizing here keeps the call sites readable:
```suggestion
    snake = re.sub(r"(?<!^)(?=[A-Z])", "_", kind).lower();
    stem = ("namespaced_" + snake) if namespace else snake;
```
(needs `import re;`)

jac/jaclang/scale/deploy/target/kubernetes/target.jac
```diff
@@ -1338,35 +1339,9 @@ obj KubernetesTarget(KubernetesTargetBase) {
...
-                    name=role_name, namespace=namespace, body=role
-                );
-            } except ApiException as e {
-                if e.status == 404 {
-                    rbac_v1.create_namespaced_role(namespace=namespace, body=role);
-                } else {
-                    raise;
-                }
-            }
+            apply_k8s_resource(rbac_v1, "Role", role_name, role, namespace=namespace);
```
MusabMahmoodh, line 1342: This one sits inside `_apply_ops_rbac`'s `except Exception`, so the AttributeError becomes an "Ops Console RBAC not applied" warning and the deploy reports success with no Role bound. Silent fallback on an opt-in path, so it's the call that most needs the test.

jac/jaclang/scale/deploy/target/kubernetes/utils/kubernetes_utils.impl.jac
```diff
@@ -456,14 +456,50 @@ impl delete_if_exists(
...
+            delete_func(name);
+        }
     } except ApiException as e {
         if (e.status == 404) {} else {
             raise;
         }
     }
 }
 
+impl apply_k8s_resource(
```
MusabMahmoodh, line 472: `KubernetesTarget._apply_or_replace` (target.jac:1703) already does read/update/create by method name, with explicit names and a `patch_body` override, and it's still called at 1211/1242/1299/1317. Can one absorb the other? Otherwise the package ends up with two apply helpers, and the explicit-name one wouldn't have had the casing problem.

MusabMahmoodh (changes requested):
The delete-if-exists conversions are equivalent, and the create-only vs replace split matches the old blocks one for one. The apply helper can't resolve any kind it is handed, though, so every monitoring deploy breaks at its first resource. Nothing in the suite runs MonitoringDeployer or `_apply_ops_rbac`, which is why CI is green on it. A no-mock check in `scale/tests/deploy/test_k8s_utils.jac` would pin it: for each (api class, kind, namespaced?) the deployers pass, assert `hasattr(client.CoreV1Api, "read_" + stem)` and the same for replace/create. The two red checks come from the stack base (#9229), not these hunks.

chess10kp:
Superseded by #9374 (rebased stack-3 plus the accumulated annex-split train).

PR #9233, opened 2026-09-16
fix(runtime): say why a pod does not boot from its sealed image (#9139)

### LINKED ISSUE #9139: Jac Scale: prepare_application gives no reason when a present _precompiled image is not booted from, so a pod that recompiles or crashes at boot cannot say which sealed check missed

akindu-k:
### Summary

image

When a pod starts, it either loads the app's **sealed** (pre-compiled) code, or compiles the app from source. Before this PR, it never said **why** it skipped the sealed code. This PR adds one log line for that case and a clear error for a sealed bundle that is missing its serve manifest. **Nothing changes about when a pod uses the sealed code.**

| Situation at pod startup | Before | After |
|---|---|---|
| Sealed image is present, but does not list this app's file | `Compiling server modules and placement metadata...` (no reason) | `Sealed image at /app does not list web.jac; preparing from source` |
| Sealed bundle has no serve manifest | `[Errno 2] No such file or directory: '.../.jac/serve_manifest.json'` | `Sealed application has no serving manifest: .../.jac/serve_manifest.json; rebuild the bundle so the seal step writes it` |
| Sealed image lists this app's file | `Loading sealed application artifacts` | unchanged |

### What was wrong

`prepare_application` decides between the sealed path and the source path with a single check, `sealed_root_for(entry)`. It only printed a line when the sealed path was taken.

1. **Silent fallback.** `sealed_root_for` returns `None` both when there is no sealed image at all and when there is one that does not list the entry. The caller could not tell these apart, so a pod with a whole sealed image on disk compiled everything without saying why. This silence is what made #9136 hard to diagnose.
2. **Bare crash.** `_prepare_sealed` read `.jac/serve_manifest.json` without checking it exists. When the bake never wrote it (a pre-0.37.12 scale seal, or the #9138 case), the pod crashed with a raw `FileNotFoundError`. Every other missing sealed artifact in that function already raises a named `RuntimeError`.

### What this PR changes

- **`server/sealed_serve.jac`**
  - New `unlisted_image_root(path)`: returns the sealed image of the file's **own project** when that image does not list the file, otherwise `None`.
  - The walk stops at the project's own `jac.toml`. A project nested inside another project's sealed image (for example, user projects staged inside jacBuilder's image) therefore does not report a miss that is not its own.
  - Both lookups now run over **one** private ancestor walk, `_image_roots`, which yields each ancestor image root and whether a nested project's `jac.toml` was crossed on the way. `sealed_root_for` takes the first image that lists the file (unchanged, including #7957's walk past an unlisted image), and `unlisted_image_root` takes the nearest image. Where an image's reach ends is defined in one place.
  - The "file path to manifest key" rule moves into `_image_key`, shared by both lookups.
- **`compiler/driver/application.jac`**
  - `prepare_application`: when the sealed path is not taken but `unlisted_image_root` finds an image, print `Sealed image at <dir> does not list <file>; preparing from source`.
  - `_prepare_sealed`: check the serve manifest exists first, and raise `Sealed application has no serving manifest: <path>; rebuild the bundle so the seal step writes it` if it does not.

### What stays the same

- Which pods load sealed code and which compile is unchanged.
- A pod whose app **is** in the sealed image prints exactly what it printed before.
- A project with no sealed image, or a nested project inside someone else's image, prints nothing new.
- A serve manifest deleted from an already sealed image is still caught by the existing image integrity check, as before.

### 1. Tests (extended, none added)

| File | What it now also checks |
|---|---|
| `tests/client/test_sealed_serve.jac` | `unlisted_image_root` is `None` for a listed file, names the image for an unlisted file in the image's own project, is `None` for a nested project with its own `jac.toml`, and names the image when the manifest is unreadable |
| `tests/compiler/test_application_preparation.jac` | the sealed preparation test prints the new line for an unlisted file, and raises the named error once the serve manifest is removed. The test now resolves its temp dir once, so it also passes where `TMPDIR` is a symlink (macOS) |
| `jaclang/scale/tests/microservices/test_app_precompile.jac` | the real pod binary test boots an unlisted file from the extracted bundle and asserts the new line in the **process's own output** |

| | Result |
|---|---|
| Tests in the three files | 33 before, 33 after |
| Without the change to `application.jac` | fails, in-process and in the real pod binary's output: the only line is `Compiling server modules and placement metadata...` |
| Previous two-walk vs new one-walk `sealed_serve.jac`, over 9 folder layouts | identical results |
| Without the serve manifest check | fails: bare `FileNotFoundError` |
| With this PR | passes |
| Related suites (`test_find_module`, `test_sealed_bindings`, `test_serve_client`, `test_hmr`, `test_desktop_broker`, `test_host_serving`, `test_served_app_route`) | 138 passed, 9 skipped |

### 2. Real Kubernetes cluster, using this PR's own CI binary

A temporary job in this PR's CI (since removed) used the `jac` binary that `build-kit` built from this PR. It first confirmed the binary contains the change:

```
kit binary has unlisted_image_root from .../site/jaclang/server/sealed_serve.jac
```

It then deployed a workspace on MicroK8s (`default-app = "svc"`, apps `web` and `svc`) with `jac scale deploy web.jac`, twice: pods on main's `jac`, then pods on this PR's `jac`. This branch does not include #9228, so the `web` pod naturally hits the "not listed" case.

Run: https://github.com/jaseci-labs/jac/actions/runs/35118848094 (job `tmp-verify-9139`)

| | pods on `main` | pods on this PR |
|---|---|---|
| `web` pod: image does not list `web.jac` | `Compiling server modules and placement metadata...` with **no reason** | `Sealed image at /app does not list web.jac; preparing from source` |
| `svc` pod: listed (control) | `Loading sealed application artifacts` | `Loading sealed application artifacts` (unchanged) |
| Real bundle copied from the `svc` pod, serve manifest removed as if never written, booted with the pod's command | `[Errno 2] No such file or directory: '.../.jac/serve_manifest.json'` | `Sealed application has no serving manifest: ...; rebuild the bundle so the seal step writes it` |

<details>
<summary>Pod logs from the run</summary>

```
== main: web pod
Preparing application...
Preparing application...
  Compiling server modules and placement metadata...

== this PR: web pod
Preparing application...
  Sealed image at /app does not list web.jac; preparing from source
Preparing application...
  Compiling server modules and placement metadata...

== main and this PR: svc pod
Preparing application...
  Loading sealed application artifacts
  Server ready (no client)

== main: bundle without serve manifest
Preparing application...
  Loading sealed application artifacts
✖ Error: Error loading svc.jac: [Errno 2] No such file or directory: '/home/runner/_work/_temp/bundle-main/.jac/serve_manifest.json'

== this PR: bundle without serve manifest
Preparing application...
  Loading sealed application artifacts
✖ Error: Error loading svc.jac: Sealed application has no serving manifest: /home/runner/_work/_temp/bundle-fix/.jac/serve_manifest.json; rebuild the bundle so the seal step writes it

=== verdict ===
OK main: web pod compiled with no reason given (bug reproduced)
OK fix: web pod names the sealed image that missed web.jac
OK fix: svc pod still loads sealed with no extra line
OK main case B: bare FileNotFoundError (bug reproduced)
OK fix case B: named error
```

</details>

> **Note:** in both runs, the `web` pod later stopped with `Dependency installation failed after 40.0s` while compiling from source. It happened identically with main's binary and this PR's binary, after the lines above. It comes from the #9136 source-compile path on that runner, not from this change.

### 3. Local reproduction

The same cases, packed with `pack_jab` and booted with `jac run --serve --no-client --faux`:

| Case | `main` | this PR |
|---|---|---|
| Image does not list `web.jac` | `Compiling server modules...` (no reason) | `Sealed image at /tmp/sealed9139-... does not list web.jac; preparing from source` |
| Serve manifest never written | `[Errno 2] No such file or directory` | `Sealed application has no serving manifest: ...` |
| Control: `svc.jac` listed | `Loading sealed application artifacts` | `Loading sealed application artifacts` |
| Nested project inside the image (jacBuilder shape) | `Compiling server modules...` | `Compiling server modules...` (no false "does not list" line) |

## **Not in this PR**

- #9228: seal the app the deploy serves, which removes the cause of the "not listed" case for workspace deploys
- #9138: fail the deploy on the host when the serve manifest is missing, so the named error here becomes the second line of defense
- #9140: a format version on the serve manifest

Fixes #9139


jac/jaclang/compiler/driver/application.jac
```diff
@@ -333,12 +340,20 @@ def prepare_application(
...
     `root = os.path.realpath((`root or os.path.dirname(`entry)));
     entries = `list(`dict.fromkeys([`entry, *(os.path.realpath(p) for p in services)]));
-    import from jaclang.server.sealed_serve { sealed_root_for }
+    import from jaclang.server.sealed_serve { sealed_root_for, unlisted_image_root }
     image_root = sealed_root_for(`entry);
     if image_root is not None {
         console.print('  Loading sealed application artifacts');
         return _prepare_sealed(`entry, program, image_root, entries);
     }
+    unlisted = unlisted_image_root(`entry);
```
MusabMahmoodh, line 349: This line is the entire point of #9139, and the only thing asserting it is a `patch.object(console, 'print')` case in the compiler suite. That proves the function was called; it does not prove an operator ever sees the line, which is the part #9136 burned a P1 on.

`jac/jaclang/scale/tests/microservices/test_app_precompile.jac` already boots a real pod binary against an extracted bundle ("a real pod binary seals and serves an entry named app"). Asserting the new line in that test's output is what makes the release note's claim true. Offer stands from the issue: my jac-ec2 rig can boot the k8s_e2e fixture bundle both ways if you want the deployed shape confirmed before it merges.

jac/jaclang/server/sealed_serve.jac
```diff
@@ -42,18 +42,38 @@ def _image_module_keys(img_root: str) -> set {
...
-            rel = os.path.relpath(abs_path, d).replace(os.sep, "/");
-            if rel.endswith(".py") {
-                rel = rel[:-3] + ".jac";
-            }
-            if rel in _image_module_keys(d) {
-                return d;
-            }
+            return None if _image_key(abs_path, d) in _image_module_keys(d) else d;
+        }
+        if os.path.isfile(os.path.join(d, "jac.toml")) {
```
MusabMahmoodh: The `jac.toml` stop is right on its own terms - a nested project with its own manifest should not report a sealed miss against the image it happens to sit under. But `sealed_root_for` above has no equivalent stop, so where an image's reach ends is now defined twice, differently, in adjacent functions. That rule already changed once (#7957); the next change has to find both.

Would it collapse if one private walk yielded each ancestor image root in turn, and both of these were expressed over it? `sealed_root_for` takes the first listed one, `unlisted_image_root` takes the nearest unlisted one and stops at a project boundary.

jac/tests/compiler/test_application_preparation.jac
```diff
@@ -302,6 +303,20 @@ test "sealed preparation consumes baked artifacts without compiling retained sou
...
+                unlisted.write_text("def:pub other -> str { return 'source'; }\n");
+                with patch.object(console, 'print') as printed {
+                    prepare_application(str(unlisted), JacProgram(), tmp);
+                }
+                lines = [
+                    str(call.args[0])
+                    for call in printed.call_args_list
+                    if call.args
+                ];
+                assert f'  Sealed image at {tmp} does not list unlisted.jac; preparing from source'
```
MusabMahmoodh, line 317: `prepare_application` does `os.path.realpath(entry)` before it computes `unlisted`, so the printed root is the resolved path while this compares against the raw `tmp`. Anywhere TMPDIR is a symlink - macOS resolves `/var` to `/private/var` - this assert fails. `test-compiler` is ubuntu-only so CI will not catch it; it fails for anyone running the suite on a Mac.

```suggestion
                assert f'  Sealed image at {os.path.realpath(tmp)} does not list unlisted.jac; preparing from source'
                    in lines , lines;
```

MusabMahmoodh (approved):
Reads well and it lands where the issue asked. The `_image_key` extraction is a real dedup, and the rewritten `sealed_root_for` still walks past an unlisted manifest to the next ancestor, so the #7957 membership behaviour is preserved. Both tests went into the files that already own these surfaces rather than new ones.

Three things below. The middle one is the one I care about.

jac/jaclang/compiler/driver/application.jac
akindu-k: Agreed, done in 842c87c14.

The real pod binary test in `test_app_precompile.jac` now does this in every case, after the sealed boots: it writes `unlisted.jac` into the extracted bundle, boots it with the pod binary (`jac run --serve --no-client --faux unlisted.jac`, `JAC_NO_DEV_SOURCE=1`), and asserts on the process's own stdout:

- `Sealed image at <bundle dir> does not list unlisted.jac` is printed
- `Loading sealed application artifacts` is not

With only the `console.print` removed from `prepare_application`, the test fails on the real output:

```
AssertionError: ('an entry named app', 'Preparing application...\nPreparing application...\n  Compiling server modules and placement metadata...\nInitializing application...
```

This file runs in CI's `test-scale (microservices)` lane against the kit built from the PR, so the operator-visible line is now pinned on every run.

For the deployed shape: the PR description has a MicroK8s run with pods on this PR's kit binary (https://github.com/jaseci-labs/jac/actions/runs/35118848094), where the `web` pod logged `Sealed image at /app does not list web.jac; preparing from source`. If you'd still like the k8s_e2e fixture booted both ways on your jac-ec2 rig before merge, that would be welcome.

jac/jaclang/server/sealed_serve.jac
akindu-k: Good call, done in 842c87c14.

There is now one private walk, `_image_roots(abs_path)`, which yields `(image_root, nested)` for each ancestor that holds `_precompiled/MANIFEST.json`. `nested` is true once a `jac.toml` has been passed below that image. Both lookups are expressed over it:

- `sealed_root_for`: the first yielded root that lists the file (still walks past an unlisted manifest, so #7957 is unchanged)
- `unlisted_image_root`: the nearest yielded root, reported only when it is not across a project boundary and does not list the file

To check nothing moved, I ran both functions over the same 9 layouts with the previous and the new `sealed_serve.jac`, and the outputs were identical:

| layout | `sealed_root_for` | `unlisted_image_root` |
|---|---|---|
| listed | `a` | `None` |
| unlisted, same project | `None` | `a` |
| unlisted, subdir | `None` | `a` |
| nested project (own `jac.toml`) under an image | `None` | `None` |
| #7957: inner image unlisted, outer lists it | `outer` | `outer/inner` |
| empty manifest, `.py` | `None` | `b` |
| `.py` maps to a listed `.jac` | `c` | `None` |
| plain project, no image | `None` | `None` |
| no project, no image | `None` | `None` |

(In the #7957 row `prepare_application` never reaches `unlisted_image_root`, because `sealed_root_for` already returned `outer`.)

jac/tests/compiler/test_application_preparation.jac
akindu-k: Right, and it goes one step further: on a symlinked TMPDIR this test fails before it reaches that assert. I reproduced macOS's `/var` → `/private/var` on Linux by pointing `TMPDIR` at a symlink. Even `main`'s version of this test (without this PR's lines) fails at line 293:

```
KeyError: '.../linktmp/jac-prepared-sealed-.../prepared_sealed_test.jac'
```

`first.manifests` is keyed by the resolved path, while `path` is built from the raw `tmp`.

So instead of resolving only in the assert, 842c87c14 resolves the temp dir once where the test creates it:

```jac
with tempfile.TemporaryDirectory(prefix='jac-prepared-sealed-') as raw_tmp {
    tmp = os.path.realpath(raw_tmp);
```

That fixes the existing `KeyError` and this assert together, and the assert stays as `{tmp}`, since `tmp` is now the resolved path.

| `TMPDIR` | before | after |
|---|---|---|
| symlink (macOS-like) | fails, `KeyError` at line 293 | 1 passed |
| plain | 1 passed | 1 passed |

MusabMahmoodh (approved):
All three are in. `_image_roots` as the one walk is the right shape, and asserting the reason line on the pod binary's own stdout is what makes the release note true.

MalithaPrabhashana (commented):
Looks good, and I'm approving. I ran it against jacBuilder's layout, since the nested-project case is ours.

Setup: macOS arm64, the 0.37.21 release binary in dev mode on this branch (c475aa56) and on its merge base (bfe47e49), with `TMPDIR` set to its real path. The default `/var` symlink makes five other tests in `test_application_preparation.jac` fail on main and on this branch alike (you fixed the sixth), so that part is unrelated.

`sealed_root_for` returns the same answer on main and on this branch for all 14 layouts I tried, built on jacBuilder's real 152-module manifest: listed and unlisted jacBuilder modules, the `.py` spelling, user projects under `/app/.jachome/.jac-ide/projects` with and without a `jac.toml`, a user project with its own seal, the `evaluation/web` satellite, the #7957 walk past an unlisted image, a symlinked root, a corrupt manifest and no image at all. So which pods boot sealed doesn't change.

| | main | this branch |
| --- | --- | --- |
| the three changed test files | 33 tests | 33 tests |
| this branch's tests, run against each side's `application.jac` + `sealed_serve.jac` | 10 passed, 1 failed, 1 error | 16 passed |
| adjacent suites (find_module, sealed_bindings, serve_client, hmr, desktop_broker, host_serving, served_app_route) | 147 passed | 147 passed |
| trial merge with today's main | | 32 passed, 1 skipped |

The red on main is the right one: `AssertionError: ['  Compiling server modules and placement metadata...']`. Taking out only the serve-manifest check turns it red again with the old bare `FileNotFoundError`, so both halves are pinned.

I also sealed a small app with this branch and booted an entry the image doesn't list. Main prints only `Compiling server modules and placement metadata...`; this branch prints `Sealed image at .../sealprobe-sealed does not list extra.jac; preparing from source`. Listed entries boot the same on both. Deleting the serve manifest from that fresh seal is caught by the integrity check (`sealed image: cannot read ...`) before the new check runs, as the description says.

non-blocking: `unlisted_image_root` only gives a correct answer after `sealed_root_for` has returned None. In the walk-past case it names `outer/inner` for a file that `outer` does seal. `prepare_application` calls them in the right order, so nothing is wrong today, but a `_` prefix or a line in the docstring would stop a later caller from trusting it on its own.

I didn't run the real pod-binary test here (it needs a `zig build` of the branch), or anything on Linux or a cluster, so for those I'm going on your CI and the MicroK8s run.

MalithaPrabhashana (approved):
Approving, the verification is in my comment above.

PR #9244, opened 2026-09-16
fix(scale): vendor pod wheels with the manylinux ladder pip can match

### LINKED ISSUE #9161: jac-scale fat bundle vendoring fails for apps with compiled dependencies: --platform manylinux_2_36_<arch> matches none of their wheels
Every Kubernetes deploy vendors the app's Python dependencies with `pip download --only-binary=:all: --platform manylinux_2_36_<node_arch>`. pip matches that tag literally, and none of the compiled wheels below carries it. So for any app with one of these dependencies, the download fails and the deploy falls back to a thin bundle. Every pod then installs the dependencies from PyPI when it boots.

## Reproduction

pip 26.2.1, CPython 3.14.7:

```sh
python3 -m pip download --no-deps --only-binary=:all: --python-version 3.14 --implementation cp \
  --platform manylinux_2_36_x86_64 -d out "watchdog>=3.0.0"
```

| package | `--platform manylinux_2_36_x86_64` (what jac-scale passes) | `--platform manylinux2014_x86_64` |
| --- | --- | --- |
| `watchdog>=3.0.0` | `ERROR: Could not find a version that satisfies the requirement` | `watchdog-6.0.0-py3-none-manylinux2014_x86_64.whl` |
| `bcrypt` | same error | `bcrypt-5.0.0-cp39-abi3-manylinux2014_x86_64.manylinux_2_17_x86_64.whl` |
| `grpcio` | same error | `grpcio-1.83.1-cp314-cp314-manylinux2014_x86_64.manylinux_2_17_x86_64.whl` |
| `cryptography` | same error | `cryptography-50.0.1-cp311-abi3-manylinux2014_x86_64.manylinux_2_17_x86_64.whl` |

## Cause

- **The tag.** `jac/jaclang/scale/injector/bundle.jac:843-848` (v0.37.14): `_pod_wheel_platform` returns `f"manylinux_2_36_{pod_arch}"`, and `pack_jab` passes it to `_vendor_stage` (`:1036`).
- **No fallback when a platform is set.** `jac/jaclang/dist/jab_vendor.jac:47-69`: `download_wheels` builds `pip download --only-binary=:all: --dest <dir> --platform <tag> <specs>`. When a platform tag is set, a failure returns immediately, so the `pip wheel` fallback for sdist-only packages is never tried on a deploy.
- **The thin bundle.** At `bundle.jac:1044` the failure becomes a warning, and the bundle ships thin.

The tag came in with 049fe7d0d (#7989, first released in 0.36.0). Its description says the vendor script "pins `--platform manylinux_2_36_<node_arch>` (the glibc floor of the oldest supported pod base, `python:3.12-slim` bookworm; pip expands it to all older manylinux tags)".

pip does not expand it. `pip/_internal/utils/compatibility_tags.py` (pip 26.2.1) expands only the legacy aliases, and uses any other tag as given:

```python
def _custom_manylinux_platforms(arch: str) -> list[str]:
    arches = [arch]
    arch_prefix, arch_sep, arch_suffix = arch.partition("_")
    if arch_prefix == "manylinux2014":
        # ...
        if arch_suffix in {"i686", "x86_64"}:
            arches.append("manylinux2010" + arch_sep + arch_suffix)
            arches.append("manylinux1" + arch_sep + arch_suffix)
    elif arch_prefix == "manylinux2010":
        # ...
        arches.append("manylinux1" + arch_sep + arch_suffix)
    return arches
```

`jac/jaclang/scale/tests/microservices/test_fat_bundle.jac:51-52` only asserts the returned string. No test downloads a real wheel with it.

## Where it hits

The jacBuilder deploys on 0.37.11 (PR preview) and on 0.37.14 (billing and e2e CI jobs) all log:

```
WARNING - Fat bundle vendoring failed; shipping a thin bundle (pods will install python deps from PyPI at boot). Set [scale] fat_bundle = true to make this fatal. Fat bundle vendoring failed, so the deploy cannot ship offline deps: ✖ Error: Wheel download failed:
ERROR: Could not find a version that satisfies the requirement watchdog>=3.0.0 (from versions: none)
ERROR: No matching distribution found for watchdog>=3.0.0
```

pip stops at the first package it cannot find, so the log names only watchdog.

Even with a working tag, jacBuilder would still not vendor:
- `http-ece` (required by `pywebpush`) publishes no wheel for either tag.
- `download_wheels` skips the build fallback whenever a platform is set.
- A platform tag is always set, so a deploy from a matching linux-x86_64 host is refused the fallback too.

## Expected

The download should ask for tags pip can match. For example, pass `--platform manylinux2014_<arch>` plus each `manylinux_2_<n>_<arch>` up to the pod base's glibc; pip accepts `--platform` more than once. A test should also download a real `manylinux2014`-only wheel through `download_wheels`.

marsninja:
## Description

Kubernetes vendoring previously passed one literal `manylinux_2_36_<arch>` tag to pip. That misses compatible published wheels and silently falls back to thin bundles. Source-only dependencies such as `http-ece` behind `pywebpush` also could not vendor for a pod target.

This change centralizes the pod target in `WheelTarget` and resolves against the full compatible manylinux ladder. `download_wheels` returns a typed result and builds missing source distributions before re-resolving the original dependency closure.

- Preserve both pip output streams. The resolver's missing-distribution summary for ranges such as `pywebpush>=2.0.0` is on stdout even when stderr contains `ResolutionImpossible`.
- Recover source requirements from the resolver's dependency causes, including bounds and extras. Convert pip's diagnostic `and` separators back to requirement syntax. Try combined constraints first, then alternatives when competing parent versions cannot share one source version.
- Build source candidates in a temporary directory; only wheels selected by the final target resolution enter the bundle. Source builds bypass the wheel cache and require the named package's source distribution.
- Keep bare `linux_<arch>` tags out of the published-wheel ladder. Permit that tag on a source-build retry only when the host is Linux on the target architecture and its detected glibc is no newer than the pod's floor. Newer, unknown, musl, and foreign hosts cannot contribute bare Linux compiled wheels. pip still checks the Python/ABI tags, and rejected builds are named in the error.
- Bound source fallback to ten build rounds, log progress, and report pending packages when the limit is reached.
- Pass the legacy platform-tag argument to vendor scripts selected from the seal binary, so both older scripts and the new architecture parser understand it.
- Run pip through the bundled real CPython. In standalone mode, clear inherited Python paths and disable the user site so global installs match the launcher's import visibility, while pip can still supply its own isolated build environment.

Fat-bundle failures remain all-or-nothing: partially vendored dependencies are removed before a thin-bundle fallback. Explicit `[scale] fat_bundle = true` remains fatal on vendoring failure. Documentation and release notes describe the platform and source-build behavior.

## Validation

- `tests/publish/test_fat_jab.jac`: 16 passed, including real offline pip resolution of constrained transitive sdists, alternative parent requirements, selected-wheel-only staging, newer-glibc rejection, compatible fresh builds, and inherited `PYTHONPATH` isolation.
- `tests/project/test_dependencies.jac` and `test_capabilities.jac`: 125 passed.
- Five focused scale tests passed, including real PyPI resolution of manylinux wheels and `pywebpush>=2.0.0` for an ARM CPython 3.12 pod, plus the older-script argv regression.
- Real pip additionally vendored `pywebpush>=2.0.0` into 19 wheels for each of x86_64 and aarch64 on CPython 3.14. Both closures passed offline `pip install --dry-run --ignore-installed --no-index` resolution.
- Documentation content tests: 3 passed. Formatting and type checks passed on the changed Jac files.
- Full bundle-sealing integration tests were not run successfully in this checkout because its `jac/zig-out/bin/jac` build artifact is absent; the focused vendoring tests above do not require it.

Fixes #9161


jac/jaclang/dist/jab_vendor.jac
```diff
@@ -14,6 +15,66 @@ glob VENDOR_DIR_REL: str = "_vendor",
...
+    has arch: str,
+        python_version: str,
+        glibc: tuple[int, int];
+
+    def platform_tag -> str {
+        return f"manylinux_{self.glibc[0]}_{self.glibc[1]}_{self.arch}";
+    }
+
+    def platform_tags -> list[str] {
+        return manylinux_tags(self.arch, self.glibc) + [f"linux_{self.arch}"];
```
MusabMahmoodh, line 55: `linux_{arch}` at the end of the ladder opens the glibc floor that the rest of the ladder exists to enforce. pip's compatibility check is per-tag, so once `linux_x86_64` is an accepted platform, a host-built compiled wheel is accepted no matter what glibc it linked against.

Concrete: deploy host on Ubuntu 24.04 (glibc 2.39), pods on `POD_GLIBC` 2.36, app depends on a compiled sdist-only package. `_build_from_source` runs `pip wheel --no-deps`, gets `pkg-1.0-cp312-cp312-linux_x86_64.whl`, the re-resolve accepts it because `linux_x86_64` is in `pip_args()`, it vendors, and the pod dies at import with `GLIBC_2.38 not found`. That moves a deploy-time failure to pod boot, which is the expensive end.

The arch half of "only when the host is the pod platform" is what `linux_{arch}` checks; the glibc half is the dimension `POD_GLIBC` was introduced for, and nothing checks it. Dropping `linux_{arch}` from the target ladder and instead admitting a host-built `linux_*` wheel only when the host's own glibc is at or below the target floor (`os.confstr("CS_GNU_LIBC_VERSION")`, or `platform.libc_ver()`) would close it, and the rejection already has somewhere good to go through `_reject_unfit`. Pure-python builds are unaffected either way since `py3-none-any` matches without it - `http-ece` in your validation is one of those.

jac/jaclang/dist/jab_vendor.jac
```diff
@@ -37,54 +98,125 @@ def has_vendored(img_root: Path) -> bool {
...
+    flags = ["download", "--only-binary=:all:", "--dest", str(dest)];
+    if target is not None {
+        flags = flags + target.pip_args();
     }
-    (rc, out) = _pip(args + specs);
-    if rc != 0 {
-        if platform_tag or python_version {
-            return (False, out);
+    built: dict[str, list[Path]] = {};
+    while True {
```
MusabMahmoodh, line 121: This terminates, since `unfit` catches any name that comes back after being built and `built` only grows. What it does not have is a bound on how many rounds it takes: a resolver that surfaces one new unsatisfiable package per round walks the closure one package at a time, and every round is a full `pip download` over the whole closure plus a `pip wheel` per new name.

On a deploy that all runs inside `seal_timeout_setting`, so the pathological case reads as a seal timeout rather than a vendoring problem. Would a round cap (or just carrying the round count into `detail`) be worth it, so the failure says what it was doing when it ran out?

jac/jaclang/scale/utils/vendor_wheels.jac
```diff
@@ -1,21 +1,46 @@
...
+    }
+    return WheelTarget(
+        arch=arch,
+        python_version=f"{sys.version_info.major}.{sys.version_info.minor}",
+        glibc=POD_GLIBC
+    );
+}
+
+
+def pod_arch_from_argv(raw: str) -> str {
```
MusabMahmoodh, line 34: Trying to pin down which direction of skew this covers. `_vendor_stage` runs the script resolved from the CLI's own tree (bundle.jac:876-877) under `.seal-jac`, so script and caller normally move together - which would make an old CLI passing a manylinux tag to a new script impossible, not just unlikely.

If the script can instead come off the seal binary's packaged assets, then the reverse is the one that bites: a new CLI passing a bare `x86_64` to an old script that expects a tag, which would hand pip `--platform x86_64` and fail to match anything. Is that pairing reachable, and if so does it need a guard on the other side too?

MusabMahmoodh (commented):
The diagnosis is right and the ladder is the correct shape - one typed place that owns the platform knowledge, with every caller going through it, beats the tag being reconstructed at three call sites. `_reject_unfit` deleting the wheel and naming the package is the part that turns this from a silent thin bundle into something an operator can act on, which was the actual damage in #9161.

Checked the thin-bundle path: `pack_jab` rmtrees `staged/_vendor` before it warns (bundle.jac:1059), so a half-built wheels dir never reaches a pod running with `PIP_NO_INDEX=1`. That was the first thing I went looking for and it is handled.

One thing I think is a real hole, plus two questions.

jac/jaclang/dist/jab_vendor.jac
```diff
@@ -37,54 +98,125 @@ def has_vendored(img_root: Path) -> bool {
 
 
 def _pip(args: list[str]) -> tuple[int, str] {
+    import from jaclang.project.pyvenv { pip_command, pip_python }
     result = subprocess.run(
-        [sys.executable, "-m", "pip"] + args, capture_output=True, text=True
+        pip_command(pip_python(), args), capture_output=True, text=True
     );
     return (result.returncode, (result.stderr or result.stdout or "").strip());
```
MalithaPrabhashana, line 105: This drops stdout whenever stderr has anything in it, and for a version range stdout is where pip names the package with no wheel. So the source build never runs for jacBuilder's `pywebpush>=2.0.0`, and the deploy still ships a thin bundle.

With pip 26.2.1 (the one the jac binary bundles) and this ladder, `pip download pywebpush>=2.0.0` writes only this to stderr:

```
ERROR: Cannot install pywebpush==2.0.3, [...] and pywebpush==2.5.0 because these package versions have conflicting dependencies.
ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
```

The part `_missing_requirements` looks for goes to stdout:

```
Additionally, some packages in these conflicts have no matching distributions available for your environment:
    http-ece
```

pip logs that explanation with `logger.info` (`resolution/resolvelib/factory.py`) and sends INFO to stdout and WARNING and up to stderr (`utils/logging.py`), so a Linux deploy host gets the same split. A pinned `pywebpush==2.5.0` fails with `No matching distribution found for http-ece>=1.1.0` on stderr instead, which is the case the fallback handles. The fallback tests replace `_pip`, and the online tests only use pinned specs (`watchdog==6.0.0`, `bcrypt==5.0.0`, `http-ece==1.2.1`), so no test sends a range through real pip.

I ran `scale/utils/vendor_wheels.jac` from this branch (bbfde1e5b) with target `x86_64` and a fresh pip cache, on a macOS arm64 host:

| `[dependencies]` | result |
| --- | --- |
| `pywebpush = ">=2.0.0"` | exit 1, `ResolutionImpossible`, nothing built |
| `pywebpush = "==2.5.0"` | ok, built `http_ece-1.2.1-py2.py3-none-any.whl`, 19 wheels |
| jacBuilder's 20 deps | exit 1, same error (same with `aarch64`) |
| jacBuilder's deps without `pywebpush` | ok, 98 wheels |

Parsing stdout too fixes it. I tested this line:

```jac
    return (result.returncode, ((result.stderr or "") + "\n" + (result.stdout or "")).strip());
```

With it, jacBuilder's 20 specs vendor 101 wheels for both `x86_64` and `aarch64`, and `http-ece` is the only package built on the host. Every wheel's tags fit cp314 on the manylinux_2_36 ladder, and an offline `pip install --dry-run --no-index --find-links <wheels>` of the 20 specs for that target resolves all 101. You may want to keep `detail` to stderr, since stdout also carries the whole download log. A test that runs the real `_pip` on a range like `pywebpush>=2.0.0` would have caught this.

jac/jaclang/project/pyvenv.jac
```diff
@@ -129,6 +129,12 @@ def _bundled_runtime_python -> str | None {
 }
 
 
+def pip_python -> str {
+    import sys;
+    return _bundled_runtime_python() or sys.executable;
```
MalithaPrabhashana, line 134: non-blocking: this CPython reads `PYTHONPATH` and the user site, which the jac binary ignores, so pip can report a package as installed that jac can't import.

With `tabulate` in a directory on `PYTHONPATH`, `jac install tabulate --global` exits 0, pip says `Requirement already satisfied: tabulate in <that dir> (0.10.0)`, and `jac -c "import tabulate"` then fails with `ModuleNotFoundError`. It works on 25f948c07, and on this branch without `PYTHONPATH`. `ensure_global_capability("deploy.k8s")` goes through the same path.

Would `-s -E` here break the build env from a9074e163? I haven't tried it.

MalithaPrabhashana (changes requested):
Requesting changes because this still doesn't vendor jacBuilder, the case in #9161. The ladder part works: jacBuilder's deps other than `pywebpush` vendor 98 wheels. But `pywebpush>=2.0.0` fails with `ResolutionImpossible` and `http-ece` is never built, because `_pip` drops the stdout that names it. Repro and a tested fix are inline. The `PYTHONPATH` comment is non-blocking.

jac/jaclang/dist/jab_vendor.jac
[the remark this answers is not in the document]
marsninja: Fixed in [c69fe1769](https://github.com/jaseci-labs/jac/commit/c69fe176943a94ac6485573f5762e9cd1657763c). The published-wheel ladder no longer includes `linux_<arch>`. A source-build retry admits it only when the host is Linux on the target architecture and `platform.libc_ver()` reports glibc no newer than the pod floor; newer/unknown libc and musl fail closed. Source builds also bypass pip's wheel cache and require the named package's sdist, so the host check applies to a fresh build.

Added host-compatibility cases and a real-pip regression with a synthetic Linux-tagged source wheel: simulated glibc 2.39 is rejected for the 2.36 target, while 2.36 is accepted. Pure-Python cross-target builds still pass.

jac/jaclang/dist/jab_vendor.jac
```diff
@@ -37,54 +98,125 @@ def has_vendored(img_root: Path) -> bool {
...
+    for (i, line) in enumerate(lines) {
+        if not line.rstrip().endswith(_NO_DIST_HEADER) {
+            continue;
+        }
+        for follow in lines[i + 1:] {
+            if not follow.startswith(" ") or not follow.strip() {
+                break;
+            }
+            name = follow.strip();
+            found.setdefault(_project_name(name), name);
```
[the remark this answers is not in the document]
marsninja: Fixed in [c69fe1769](https://github.com/jaseci-labs/jac/commit/c69fe176943a94ac6485573f5762e9cd1657763c). The missing-name summary is now paired with requirements from the resolver's dependency causes, preserving bounds and extras. The parser also converts pip's diagnostic spelling `<2 and >=1` back to requirement syntax. Source builds first try the combined requirements, then individual alternatives if competing parent versions cannot satisfy them together; the original target resolver makes the final selection.

Real offline-pip tests cover a ranged parent whose transitive sdist must stay below version 2 despite a newer source release, and conflicting parent candidates. Builds are staged separately so unselected candidate wheels never enter the bundle.

jac/jaclang/dist/jab_vendor.jac
marsninja: Agreed; architecture alone was insufficient. Fixed in [c69fe1769](https://github.com/jaseci-labs/jac/commit/c69fe176943a94ac6485573f5762e9cd1657763c): `linux_<arch>` is excluded from the normal ladder and allowed on a source-build retry only for a Linux host with matching architecture and detected glibc <= the pod floor. Unknown libc, musl, and newer glibc are rejected. The fallback uses a fresh sdist build with pip's wheel cache disabled.

The regression uses real pip with a synthetic Linux-tagged source wheel and mocked host libc detection: glibc 2.39 fails against the 2.36 pod target and leaves no bundled wheel; 2.36 succeeds. Unit cases cover older/unknown libc and foreign platforms. The ARM pure-Python sdist regression also passes.

jac/jaclang/dist/jab_vendor.jac
marsninja: Added in [c69fe1769](https://github.com/jaseci-labs/jac/commit/c69fe176943a94ac6485573f5762e9cd1657763c). Source fallback is capped at ten build rounds, with a final resolution attempt after the tenth build. Each build round logs its number and missing packages; exhausting the cap returns a vendoring error containing the round count and pending package names.

A regression feeds one new missing package per resolution and verifies bounded execution and the final diagnostic. This bounds the number of rounds; it does not replace the subprocess timeout for an individual slow pip operation.

jac/jaclang/scale/utils/vendor_wheels.jac
marsninja: Yes, that pairing is reachable: `_child_tree_script` probes the seal binary's `jaclang` location and prefers its packaged vendor script, falling back to the CLI tree. My earlier description missed that direction of skew.

Fixed in [c69fe1769](https://github.com/jaseci-labs/jac/commit/c69fe176943a94ac6485573f5762e9cd1657763c): the caller now passes the legacy platform-tag spelling, derived through the shared target helper. Older scripts understand it, and the new script normalizes it back to an architecture. Added a regression selecting an older script and verifying that `arm64` is passed as `manylinux_2_36_aarch64`.

jac/jaclang/dist/jab_vendor.jac
marsninja: Fixed in [c69fe1769](https://github.com/jaseci-labs/jac/commit/c69fe176943a94ac6485573f5762e9cd1657763c). `_pip` now retains both stdout and stderr, so the resolver's missing-distribution explanation reaches the fallback. The online regression now uses `pywebpush>=2.0.0` rather than a pinned source dependency. There is also a split-stream regression and a real offline-pip test with a ranged parent and constrained transitive sdist.

Validated the range against PyPI for an ARM CPython 3.12 target, then for CPython 3.14 on both x86_64 and aarch64. Each 3.14 target vendored 19 wheels, including the locally built `http-ece`, and passed offline `pip install --dry-run --ignore-installed --no-index` resolution. I did not rerun jacBuilder's complete 20-spec closure. Failure detail currently retains both streams so the resolver explanation remains available.

jac/jaclang/project/pyvenv.jac
marsninja: Fixed in [c69fe1769](https://github.com/jaseci-labs/jac/commit/c69fe176943a94ac6485573f5762e9cd1657763c) by sanitizing the environment passed to bundled CPython in standalone mode: remove inherited `PYTHONPATH`, `PYTHONHOME`, and `PYTHONUSERBASE`, and set `PYTHONNOUSERSITE=1`. The vendor subprocess, global package installer, and global capability installer all use the helper.

This lets pip establish its own `PYTHONPATH` for isolated build children. A real-pip regression verifies that a distribution visible only through inherited `PYTHONPATH` cannot satisfy an install, and the isolated sdist-build tests still pass. The project/capability suites passed all 125 tests.

MalithaPrabhashana (approved):
Our blocking item is fixed. I verified 9048db935; 243f29ad3 only merges main (#9357, #9389) and none of this PR's 15 files differ. I ran jacBuilder's real dependencies (20 specs from its jac.toml, `pywebpush>=2.0.0` included) through `vendor_wheels.jac` the way `_vendor_stage` calls it, dev mode on both sides, fresh pip cache per run, macOS arm64 host:

| | x86_64 | aarch64 |
| --- | --- | --- |
| main 7f46804fa | fails, `watchdog>=3.0.0` not found (#9161) | fails, same |
| 9048db935 | 101 wheels, 119s | 101 wheels, 143s |

Every vendored wheel fits cp314 on glibc 2.36, none is `linux_*`, no project appears twice, and the only host build is `http_ece-1.2.1-py2.py3-none-any`. `pip install --dry-run --ignore-installed --no-index --find-links` against just the vendored dir, with the pod's tag ladder, resolves all 101 on both arches.

Greptile's 3/5 is from bbfde1e5b, before c69fe1769, so I checked its two P1s against this head:

- Constrained transitive sdist, real pip with local packages: `cons-parent==1.0` needs `cons-child<2`, and only sdists of 1.5 and 2.0 exist. This branch builds 1.5 and vendors 2 wheels; main fails.
- Host libc: I can't run the Linux real-pip test on macOS, so I drove `host_build_compatible()` with mocked hosts. `linux_x86_64` is never in the first pass, and the retry admits it only for Linux x86_64 with glibc 2.36 or older. glibc 2.39, musl, an unparseable version, aarch64 and darwin are all refused. On this Mac, an sdist that builds a `cp314-macosx_11_0_arm64` wheel is rejected for both targets, deleted, and named in the error. A real C extension doesn't get that far under the jac binary, because the bundled CPython ships no `Python.h`, so the build itself fails and nothing is bundled.

Reverting each review fix on its own turns tests red: the stdout drop 3 failures, the constraint recovery 4, the glibc comparison 1, the round cap 1, and the `PYTHONPATH` clear 1 in `test_fat_jab` plus 1 in `test_dependencies`. The `PYTHONPATH` repro from my earlier comment now installs into the global site here.

`_install_offline` goes through the new `_pip` too, so I ran the pod boot path on the host: `activate_vendored` on a host-vendored `bcrypt tabulate pywebpush>=2.0.0` (21 wheels, `http-ece` built) installs the site and all four import from it. Main can't build `http-ece` there; with wheel-only specs both sides behave the same. A project `jac install` of jacBuilder's Python deps passes on both sides, 13/13 imports, minus `watchdog`, which has no macOS cp314 wheel and can't build without `Python.h` on either side.

Regression sweep over 13 files (the four this PR touches plus seal, injector and deploy neighbours), this head vs 7f46804fa: equal or better everywhere. My first pass had `test_binary_injector` and `test_dry_run_purity` red on both sides, only because my test HOME lacked `kubernetes` and `pyyaml`. With them installed, both pass on both sides, 46/46 and 15/15.

On CI for 243f29ad3, build-kit, all five test-scale groups, test-packages-and-docs, the test-compiler lanes and jac-check pass now that #9389 is in. The failures:

- test-runtime isn't this PR, and there is no test result to read. The job died with "The self-hosted runner lost communication with the server" and uploaded no log. main hit the same annotation on fb79d7499 and 35eac7dfa, and passed on b5c52dca7, the commit this run merges. It is the job that runs `tests/publish` and `tests/project`, though, so the Linux-only real-pip glibc test still hasn't run anywhere. It needs a rerun.
- test-managed-android isn't this PR either: Gradle fails in `:app:packageRelease`. main's 35eac7dfa failed the same job and b5c52dca7 and fb79d7499 passed it, so it comes and goes on main, and this PR doesn't touch the mobile build.
- Contribution Checks passes because of the `skip-release-notes-check` label. The fragment still ends with a paragraph that isn't a `- ` bullet, and still says `jac start --scale`, which 0.37 removed; it's `jac scale deploy` now.

Approving. The one gap is the test-runtime rerun, so `tests/publish` and `tests/project` actually run on Linux before merge.

PR #9249, opened 2026-09-17
fix(server): refuse a @schedule the scheduler cannot honour, and find a cron's next run by field

### LINKED ISSUE #9031: scheduler: @schedule accepts any spec; interval=0 spins at 300 fires/s, a negative interval starves other tasks, an unsatisfiable cron crashes boot

kashmithnisakya:
Closes #9031. Part of #9024.

Based directly on `main` and independent of #9201 and #9250, which also change the static scheduler; whichever merges second rebases.

## What was wrong

`@schedule` passed its arguments straight into `ScheduleSpec` and nothing checked them at decoration, registration or boot. Every row of the issue reproduces on `main`:

```
interval=0      next=2026-09-17 10:00:00+00:00 (now=2026-09-17 10:00:00+00:00)
interval=-5     next=2026-09-17 09:59:55+00:00
interval+cron   next=2026-09-17 10:00:02+00:00 (cron alone would be 03:00)
register past date / empty spec:
  registered: []                                  <- both dropped, no log line
cron '60 * * * *': No matching time found for cron: 60 * * * * after 0.4s
cron '0 0 29 2 *': No matching time found for cron: 0 0 29 2 * after 0.2s
field '1-5/2' matches hour 3? False  (silently never)
```

The `0 0 29 2 *` line is one the issue did not list. The cron engine scanned 366 days of minutes and raised past that, so a valid schedule crashes boot in any year whose next February 29 is more than a year ahead. The same bound applies to any sparse schedule, so the fix could not just be range checks in front of the scan.

## The fix

- **One definition of a usable spec.** `schedule_problems(spec)` in `server/task_scheduler.jac`: a `STATIC` schedule takes exactly one of `interval`, `cron` or `date`; an interval is a finite number of seconds above zero (not a bool); a cron parses; a date is ISO; a `DYNAMIC` target carries no timing, which the docs already required.
- **An interval must be schedulable, not just positive.** An interval's next run is computed by one helper, `next_interval_run`, shared by `_compute_next_run` and validation. An interval whose next run cannot be computed (`1e300`, an integer too large for a float, or `8e13` seconds, whose next run lands past year 9999) is reported as too long, and one that rounds to under a microsecond as too short, rather than escaping boot validation as an `OverflowError`.
- **Reported at boot, all at once, by target.** `JacAPIServer._validate_or_raise` already collects settings and extension problems into one `ServeConfigError`; it now adds `@schedule on '<name>': <problem>` for every scheduled target. `_scheduled_targets()` is the single list that validation and `_register_scheduled_tasks` both read, and registration skips `DYNAMIC` targets explicitly. `Scheduler.register` also refuses an invalid spec, for callers that do not come through the server.
- **Not at decoration.** Raising in the decorator would fail the module import itself, for `jac check`, tests and any non-serving import, as a traceback, one target at a time. Boot is where a schedule matters and where the server already reports configuration errors cleanly.
- **A past `date` stays a warning**, naming the task, because a restart after the date is ordinary.
- **A cron engine that jumps instead of scanning.** `CronSchedule.parse` turns each field into the set of values it selects and rejects an out-of-range number, a backwards range, syntax the engine does not support (`1-5/2`, names) and a day that exists in none of the listed months (`0 0 30 2 *`). `next_after` advances month, day, hour and minute to the next allowed value, so any schedule that parses has a next run however far away. Semantics are unchanged: weekday `0` is Monday, as the docs say, and day and weekday both have to match.
- **Dead code removed.** With timing refused on `DYNAMIC` targets, `Scheduler.dynamic_handler` and the `DYNAMIC` branch of `_execute` can no longer run, so both go, along with the handler `JacScaleScheduler.setup` installed.

## Live, before and after

The same app, `@schedule(interval=0) def spin`, `@schedule(cron="60 * * * *") def bad_minute` and `@schedule(interval=1e300) def huge`, served with `jac run --serve --no-client`:

```
before (main):
  ... admin portal and the other extension endpoints registered ...
  ✖ Error: Server error: No matching time found for cron: 60 * * * *

after:
  ✖ Error: cannot start the server:
    - @schedule on 'bad_minute': cron '60 * * * *': minute value 60 is outside 0-59
    - @schedule on 'huge': interval 1e+300 seconds is too long to schedule: its next run falls outside the dates the scheduler can represent
    - @schedule on 'spin': interval must be a positive number of seconds, got 0
```

An app with valid schedules still boots and runs. With `interval=2`, `cron="* * * * *"` and an expired `date`, it logged `@schedule on 'long_gone': date 2020-01-01 00:00:00 has already passed, so it is not registered and will not run`, then `static scheduler: 2 task(s) registered`, and `every_two` fired 14 times in 30 s.

## Behaviour changes

- A server whose code carries an invalid `@schedule` now refuses to start instead of busy-looping, starving other tasks, silently ignoring the decorator or crashing mid-boot. Every schedule in the repo's fixtures, examples and docs is valid under the new rules.
- `*/n` with `n` larger than the field's span (`*/100` in the minute field) used to mean "only the first value" and is now refused as out of range.

## Tests

`tests/runtimelib/test_scheduling.jac`, 11 passing (6 new, and the old cron test rewritten against `CronSchedule`):

- cron fields parse into the values they select, and `next_after` matches the old engine on its examples, including weekday `0` being Monday;
- `0 0 29 2 *` from 2026-09-17 runs at 2028-02-29 00:00;
- twelve bad expressions are each refused with the field they got wrong;
- every spec in the issue's table is refused with its reason, valid ones (`interval=0.3`, Feb 29 cron, a past date, a bare `DYNAMIC`) are not, and registering a past date logs a warning naming the task;
- an interval too long to schedule (`1e300`, `10 ** 400`, `8e13`) or under a microsecond is one problem, not a crash, and a century-long interval is accepted;
- the served `bad_schedules.jac` fixture makes `_validate_or_raise` raise one `ServeConfigError` naming `spin`, `bad_minute`, `SpinWalker` and `huge`, and not the valid `good_one`.

The existing test that asserted a malformed field such as `abc` "does not match", and that a malformed date is silently skipped, encoded this bug; both now expect a refusal.

Green locally against the source tree: `test_scheduling` 11, `test_scheduler_multipod` 5, `test_scheduler_endpoints` 3, `test_optional_deps` 24, `test_dev_signing_secret` 9, `test_serve` 32 (serves `test_api.jac` with its six schedules). `jac fmt --check` and per-file `jac check` are clean on every changed runtime file.


jac/jaclang/server/impl/task_scheduler.impl.jac
```diff
@@ -26,10 +36,182 @@ impl Scheduler.register(
...
+    }
+    if len(timing) > 1 {
+        return [f"takes one of interval, cron or date, not {' and '.join(timing)}"];
+    }
+    if spec.interval is not None {
+        raw: any = spec.interval;
+        usable = not isinstance(raw, bool)
+            and isinstance(raw, (int, float))
+            and math.isfinite(float(raw))
+            and datetime.timedelta(seconds=float(raw)) > datetime.timedelta(0);
```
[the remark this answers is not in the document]
kashmithnisakya: Valid, and one step wider than described. `timedelta(seconds=1e300)` raises `OverflowError`, and so does `float()` on an integer too large for a float (`10 ** 400`). An interval that does fit a `timedelta`, such as `8e13` seconds, overflows one step later, when its next run lands past year 9999. All three used to escape `_validate_or_raise` as a traceback.

Fixed in 6501c742 by answering the question with the code that schedules. The grid calculation is now one helper, `next_interval_tick`, used by `_compute_next_run` and by `schedule_problems`. Validation reports any interval whose next run cannot be computed as "too long to schedule", and one that rounds to under a microsecond as too short, instead of letting the division by a zero step fail. A fixed upper bound would have been an arbitrary number standing in for that check.

Red first: the new test and the served-app test both failed with `OverflowError: Python int too large to convert to C int`. Now `interval=1e300`, `10 ** 400` and `8e13` each produce one problem, a century-long interval is still accepted, and booting an app with `interval=1e300` prints `@schedule on 'huge': interval 1e+300 seconds is too long to schedule: its next run falls outside the dates the scheduler can represent` beside the other targets.

jac/jaclang/server/impl/task_scheduler.impl.jac
```diff
@@ -26,6 +36,177 @@ impl Scheduler.register(
...
+    timing = [
+        name
+        for (name, value) in [
+            ("interval", spec.interval),
+            ("cron", spec.cron),
+            ("date", spec.date)
+        ]
+        if value is not None
+    ];
+    if spec.trigger == ScheduleTrigger.DYNAMIC {
```
MusabMahmoodh, line 49: On main a DYNAMIC target with an interval wasn't dead: `_register_scheduled_tasks` registers every spec, and `_execute` hands DYNAMIC to `dynamic_handler`, which scale wires to `_execute_task` (scale/scheduler/impl/scheduler.impl.jac:442), so it ran as a recurring system-user job. After this an app relying on that fails boot on upgrade, a crashloop on k8s. Can that go in the behaviour changes and the release note?

jac/jaclang/server/impl/task_scheduler.impl.jac
```diff
@@ -26,6 +36,177 @@ impl Scheduler.register(
...
+                f"date {spec.date!r} is not an ISO date-time such as "
+                "'2026-12-31 09:00:00+00:00'"
+            ];
+        }
+    }
+    return [];
+}
+
+impl next_interval_run(seconds: float, now: datetime.datetime) -> datetime.datetime {
+    return now + datetime.timedelta(seconds=seconds);
```
MusabMahmoodh, line 112: #9201 rewrites this same computation to the epoch grid. Whichever lands second, the grid should go in here so validation and `_compute_next_run` keep sharing one formula, and #9201's `step <= timedelta(0)` branch can then go, since nothing non-positive gets past `schedule_problems`.

jac/jaclang/server/impl/task_scheduler.impl.jac
```diff
@@ -9,9 +9,19 @@ impl Scheduler.instance -> Scheduler {
 impl Scheduler.register(
     name: str, target: Callable, spec: ScheduleSpec, is_walker: bool
 ) -> None {
+    problems = schedule_problems(spec);
+    if problems {
+        raise ValueError(f"@schedule on '{name}': {'; '.join(problems)}");
```
MusabMahmoodh, line 14: With this refusal in front, the `except (ValueError, TypeError)` "Invalid date format" fallback in `_compute_date` can't fire any more, since every date already passed `fromisoformat` in `schedule_problems`. It can go.

MusabMahmoodh (commented):
Validation and the cron rewrite look right, and the Feb 29 case plus the served-fixture test through `_validate_or_raise` would both fail on main. Two things before it lands: an upgrade note for DYNAMIC schedules that carry timing, and keeping `next_interval_run` as the one interval formula when this meets #9201.

kugesan1105 (approved):
Approving. Validated merged onto current main (`20bbc6dfd`), not on the PR's own base.

This goes after the real cause instead of the reported one. The issue asked for cron field range checks, which would not have been enough: `0 0 29 2 *` is legal, passes every range check, and still crashed boot on main because the engine scanned 366 days of minutes. Swapping that for field arithmetic also closes something nobody filed, the recompute at `task_scheduler.impl.jac:136` sits outside any `try`, so on main a server booted before Feb 29 in a leap year fires the task and then the scheduler thread dies for good (`_running` stays `True`, so `start_if_pending` can never revive it, and every other task stops with it). Nothing that passes the new validation can raise there: I brute forced all 2,562 parse-accepted (day, month, weekday) crons and none escaped.

Also checked the other direction, a real `_prepare_to_serve` on the bad fixture refuses with one `ServeConfigError` before any route is built, while main gets through the whole `/admin` surface first and then throws a traceback.

What I ran on the merge: `test_scheduling` 11, `test_scheduler_endpoints` + `test_scheduler_multipod` 15, `test_serve` 32, `test_docs_content` 3, `jac fmt --check` and `jac check` clean on every changed file, and every cron literal in the repo, examples and docs re-parsed under the new engine (9/9 fine).

One residual, not a blocker, filed as #9277: the floor is a microsecond rather than the `[scale.scheduler] min_interval` the dynamic path already enforces, so the busy loop is still reachable from the static side.

```
@schedule(interval=1e-6)          -> accepted, 19,003 fires in 5 s (~3,800/s, a thread per fire)
POST /jobs {"interval": 1e-6}     -> 400, interval must be at least 1.0 seconds
```

And one line worth adding to Behaviour changes whenever you are next in here: it lists `*/n` over the field span, but `1-5/2,7` changes too, it fires at 07:00 on main and refuses boot here.

Nice call rejecting the issue's "validate at decoration". Past-ness of a `date` is not knowable then anyway, the repo's own fixture computes `run_at = import time + 1s`.

kugesan1105:
image

PR #9250, opened 2026-09-17
fix(server): a static task never overlaps itself, and a worker drains its runs before exiting

### LINKED ISSUE #9029: scheduler: static tasks overlap themselves without limit and in-flight runs are killed at shutdown
Part of #9024.

The static scheduler starts a new daemon thread for every fire and never checks whether the previous run of the same task is still going, so a task slower than its interval accumulates concurrent copies without bound. At shutdown only the loop thread is joined; running tasks are cut off when the worker exits. Dynamic jobs get both protections (`max_instances=1` and `stop(drain=True)`), static tasks get neither. This is #7616's territory (closed 2026-09-04 without a merged fix). The scheduler docs currently promise "A job never overlaps itself".

## Where

- thread per fire, no in-flight tracking: `jaclang/server/impl/task_scheduler.impl.jac:132`
- `Scheduler.stop()` joins `_thread` only: `task_scheduler.impl.jac:69-80`
- worker exits with `os._exit` after the HTTP drain: `serving/supervisor.jac:150-154`

## How to reproduce

```jac
@schedule(interval=3)
def slow_fn -> None {
    print("[slow_fn start]", flush=True);
    time.sleep(7);
    print("[slow_fn end]", flush=True);
}
```

```bash
jac run --serve --no-client main.jac   # then SIGTERM after ~90 s
```

## Evidence

Overlap, steady state three concurrent copies:

```
[slow_fn start] pid=33103 at=11:14:25.126
[slow_fn start] pid=33103 at=11:14:28.145     <- second copy while the first sleeps
[slow_fn start] pid=33103 at=11:14:31.143     <- third
[slow_fn end]   pid=33103 at=11:14:32.131
```

Shutdown, in-flight run lost:

```
[slow_fn start] pid=33103 at=11:15:37.219
WARNING drain: started (signal SIGTERM, budget 10.0s)
INFO    Scheduler stopped (drain=True)
Shutting down server...                     <- process gone; no matching "end"
```

Over the run: 25 starts, 22 ends.

## Proposed fix

Keep an in-flight set keyed by task name and skip (or queue at most one) while a run is live; run fires on a bounded executor; track task threads and join them inside the same drain budget the transport uses. Fix the docs sentence or make it true.

## Environment

kashmithnisakya:
Closes #9029. Part of #9024.

Based directly on `main` and independent of #9201 and #9249, which also change the static scheduler; whichever merges second rebases.

## What was wrong

Reproduced on `main` with a 0.7 s task on a 0.2 s interval:

```
interval 0.2s, run 0.7s, 2s: started=9 peak concurrent=4
stop() returned after 0.00s with 3 run(s) still going
```

`_run_loop` started a new daemon thread on every tick without looking at the previous run, and `Scheduler.stop()` joined only the loop thread. The worker then leaves through `os._exit` (`serving/supervisor.jac`), so whatever was running is cut off.

## The fix

- **One run per task at a time.** The scheduler records each task's run in flight (`_launch` / `_run_tracked`, keyed by the task). A tick that arrives while that run is alive is skipped, not queued, which is what APScheduler's `max_instances=1` does for dynamic jobs. It logs one warning per streak and an info line when the task runs again.
- **A busy replica leaves the tick to another.** The in-flight check runs before `_may_fire`, so a replica still busy with the task does not take that tick's lease, and another replica can run it.
- **No new runs while draining.** The server gives the scheduler its `DrainState`, and the loop starts nothing once draining begins.
- **Drain runs within the server's budget, on one deadline.** After the HTTP drain, both `start()` and `serve_background()` call `scheduler.stop(timeout=self.drain.remaining())`: whatever is left of `[serve.timeouts] drain`. `stop` computes one deadline and spends it on the loop join and then the run joins. It names every run still going when the deadline passes, even when the budget was already 0. `stop(timeout=None)` is a plain stop that does not wait for runs, which is what `JacScaleScheduler.stop` and `wait()` need, since the server's own drain handles static runs. `_launch` refuses a run once `stop` has begun, under the lock `stop` takes its snapshot with, so a lease check that returns during shutdown cannot start a run the drain never sees.
- **A bounded executor was not needed.** With one run per task, concurrent static runs are bounded by the number of decorated targets.

## Live, before and after

The issue's own fixture, a 7 s `slow_fn` on `interval=3`, served with `jac run --serve --no-client`, then SIGTERM while a run is going:

```
before (main)                                         after
[slow_fn start] at=08:22:33.857                       [slow_fn start] at=08:23:02.501
[slow_fn start] at=08:22:36.850   <- overlaps          'slow_fn' is still running from an earlier tick; skipping its ticks until that run finishes
[slow_fn start] at=08:22:39.847   <- third copy        [slow_fn end]   at=08:23:09.507
[slow_fn end]   at=08:22:40.863                       'slow_fn' finished its long run; its ticks run again
...                                                   [slow_fn start] at=08:23:11.503
[slow_fn start] at=08:22:54.861                       ...
SIGTERM                                               [slow_fn start] at=08:23:20.514
drain: started (signal SIGTERM, budget 10.0s)         SIGTERM at 08:23:23
(process gone at once)                                drain: started (signal SIGTERM, budget 10.0s)
                                                      [slow_fn end]   at=08:23:27.515
starts=8 ends=5                                       starts=3 ends=3
```

## Tests

`tests/runtimelib/test_scheduling.jac`, 12 passing (6 new). The file passed three times in a row to check the timing-based cases are stable:

- a static task never runs alongside an earlier run of itself: peak concurrency 1 over 2 s, the lease is taken once per run and not per tick, and the overlap warning is logged at most once per run;
- stopping waits for a run in flight within its budget, and the run finishes before `stop` returns;
- stopping gives up on a run past its budget, returns within it, and the warning names the task;
- a draining scheduler starts no new runs;
- a drain with no budget left still names the runs it abandons;
- stopping keeps one deadline even when the loop is held in a slow lease check (under 1.4 s for a 1.0 s budget).

Green locally against the source tree: `test_scheduling` 12, `test_scheduler_multipod` 5, `test_optional_deps` 24, `test_serve_disconnect` 1, `test_serve` 32, `test_serve_client` 109. `jac fmt --check` and per-file `jac check` are clean on every changed runtime file.

Docs: the Behavior Notes said "A job never overlaps itself. If a run is still going when the next fire time arrives, the new run waits." That was not true for static tasks, and not quite true for dynamic ones either: APScheduler skips the run, it does not wait. The note now says both kinds skip, and describes how shutdown drains static runs.


jac/jaclang/server/impl/task_scheduler.impl.jac
```diff
@@ -77,6 +308,33 @@ impl Scheduler.stop -> None {
...
+        running = [
+            flight
+            for flight in self._in_flight.values()
+            if flight[1].is_alive()
+        ];
+    }
+    if not running or timeout <= 0 {
+        return;
+    }
+    deadline = time.monotonic() + timeout;
```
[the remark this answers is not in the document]
kashmithnisakya: Valid on both counts, and reproduced before fixing. With a 1.0 s budget and the loop held in a 1.5 s claim, `stop` took 1.50 s. With the budget already at 0, a running task was abandoned with no log line at all.

Fixed in 26b8b40a:
- **One deadline.** `stop` computes it once and spends it on the loop join and then the run joins, so a slow claim eats into the same budget instead of adding to it.
- **`None` is a plain stop, any number is a drain.** `stop(timeout=None)` stops the loop and does not wait for runs (what `JacScaleScheduler.stop` and `wait()` want). Any number, 0 included, is a drain that names every run still going when the deadline passes. The server passes `self.drain.remaining()`, so an exhausted budget now produces the warning instead of silence.
- **No run slips in after the snapshot.** `_launch` refuses to start a run once `stop` has begun, under the same lock `stop` takes its in-flight snapshot with. Otherwise a claim that returns during shutdown could still start a run the drain never sees.

Tests added and run red first: `a drain with no budget left still names the runs it abandons`, and `stopping keeps one deadline even when the loop is slow to stop` (asserts under 1.4 s for a 1.0 s budget). The file passed three times in a row, and the live SIGTERM scenario from the PR description still drains its run in flight (3 starts, 3 ends).

jac/jaclang/cli/docs/reference/plugins/jac-scale-scheduler.md
```diff
@@ -357,6 +357,7 @@ shutdown_timeout = 10
 ## Behavior Notes
 
 - Cron fields, dynamic job triggers, and stored timestamps are all UTC. The one exception is a bare `date` string on `@schedule`, which is read in the server's local timezone; pin an offset there.
-- A job never overlaps itself. If a run is still going when the next fire time arrives, the new run waits (`max_instances=1`).
+- Neither a static task nor a dynamic job overlaps itself. If a run is still going when the next fire time arrives, that fire is skipped, not queued: a static task logs one warning until the run finishes and does not take that tick's lease, so another replica can run it, and a dynamic job is skipped by APScheduler (`max_instances=1`).
```
MusabMahmoodh, line 364: This holds per process only. A busy worker leaves the tick unclaimed, so another worker or replica claims it and runs while this run is still going: a 7s task on a 3s interval across 3 pods still has about 3 copies live. Can we scope the sentence and file the cross-replica case (hold the claim for the length of the run)?
```suggestion
- Within one worker process, a static task does not overlap itself: if its run is still going when the next fire time arrives, that tick is skipped there, logged once, and left unclaimed, so another worker or replica may run it alongside the first. A dynamic job is skipped by APScheduler (`max_instances=1`), also per process.
```

jac/tests/runtimelib/test_scheduling.jac
```diff
@@ -291,3 +292,207 @@ test "the served app registers its scheduled targets with the scheduler" {
...
+    try {
+        time.sleep(2.0);
+    } finally {
+        sched.stop(timeout=2.0);
+        log.removeHandler(records);
+    }
+    assert probe.started >= 2 , probe.started;
+    assert probe.peak == 1 , f"{probe.peak} runs overlapped";
+    # A replica busy with the task leaves the tick's lease for another replica.
+    assert len(claims) == probe.started , (len(claims), probe.started);
```
MusabMahmoodh, line 370: "A busy replica leaves the tick to another" is only checked by counting calls to a hand-written lease lambda, which proves ordering, not outcome. `scale/tests/server/test_scheduler_multipod.jac` already runs schedulers on embedded Postgres; two schedulers on one real store with a slow task would show what actually happens across replicas.

jac/jaclang/server/impl/server.impl.jac
```diff
@@ -1626,7 +1627,7 @@ impl JacAPIServer._worker_main(
     try {
         return serving.run(on_ready=_ready, port_retries=port_retries);
     } finally {
-        self.scheduler.stop();
+        self.scheduler.stop(timeout=self.drain.remaining());
```
MusabMahmoodh, line 1643: Nothing tests this call or `self.scheduler.drain = self.drain` at 1085: every new test builds a bare Scheduler and calls `stop(timeout=...)` itself, so dropping either line stays green. One test through `serve_background` with a slow task (start the drain, call the returned stop, assert the run finished) covers the real path.

jac/tests/runtimelib/test_scheduling.jac
```diff
@@ -291,3 +292,207 @@ test "the served app registers its scheduled targets with the scheduler" {
         }
     }
 }
+
+
+class _Records(logging.Handler) {
```
MusabMahmoodh, line 309: #9249 and #9201 add this same `_Records` handler to this file too, so whichever merges second has a duplicate class. Worth landing it once and reusing it.

MusabMahmoodh (commented):
The in-process part is right: the overlap check sits ahead of the claim, `_launch` refuses once `stop` has begun under the lock, and one deadline covers the whole stop. My concern is the cross-replica promise in the docs. When this rebases onto #9201, keep `_still_running` ahead of the claim, since #9201's claims only move forward and a busy worker that claims first spends the tick for everyone.

MusabMahmoodh (commented):
Nothing has moved since my last pass except the main merge, and #9201 has since landed on the same surface, so this now conflicts in `task_scheduler.impl.jac`, `task_scheduler.jac`, the scheduler doc and `test_scheduling.jac`. The four earlier threads still apply. `_Records` is on main now via #9249, so the rebase can just drop this copy. The cross-replica sentence in the doc matters more after #9201: a busy worker leaves the tick unclaimed, and another replica claims it and runs alongside it.

jac/jaclang/server/impl/task_scheduler.impl.jac
```diff
@@ -128,10 +208,13 @@ impl Scheduler._run_loop -> None {
         heapq.heappop(heap);
         now = datetime.datetime.now(datetime.timezone.utc);
         task.last_run = now;
-        if self._may_fire(task) {
-            threading.Thread(target=self._execute, args=(task, ), daemon=True).start();
-        } else {
-            logger.debug(f"'{task.name}' fired elsewhere this tick; skipping");
+        draining = self.drain is not None and self.drain.draining;
+        if not draining and not self._still_running(task) {
+            if self._may_fire(task) {
```
MusabMahmoodh, line 213: After the rebase this becomes `_may_fire(task, due)`. Keep the in-flight check ahead of the claim: #9201's claims only move forward, so a busy worker that claims first spends the tick for everyone.

jac/tests/runtimelib/test_scheduling.jac
```diff
@@ -303,3 +304,207 @@ test "the served app registers its scheduled targets with the scheduler" {
...
+    claiming = threading.Event();
+    def lease(key: str, ttl: int) -> bool {
+        if "slow_claim" in key {
+            claiming.set();
+            time.sleep(1.5);
+        }
+        return True;
+    }
+    sched = fresh_scheduler();
+    sched.lease = lease;
```
MusabMahmoodh, line 491: `lease` is `claim` on main now, with a `(key, tick)` signature. If this assignment survives the rebase unchanged, `slow_claim` never blocks, the wait loops just run out their 5s, and the test still passes because `long_run` alone holds `stop` to about 1.0s. An `assert claiming.is_set()` before `began` keeps it from going green without the slow claim. Same rename at 353.

jac/tests/runtimelib/test_scheduling.jac
```diff
@@ -303,3 +304,207 @@ test "the served app registers its scheduled targets with the scheduler" {
...
+        time.sleep(0.02);
+    }
+    claiming.clear();
+    while not claiming.is_set() and time.monotonic() < deadline {
+        time.sleep(0.02);
+    }
+    began = time.monotonic();
+    sched.stop(timeout=1.0);
+    waited = time.monotonic() - began;
+    assert waited < 1.4 , f"stop took {waited:.2f}s against a 1.0s budget";
```
MusabMahmoodh, line 509: 0.4s of headroom over a 1.0s budget is thin on a loaded shared runner. The claim is one deadline, not two, so `waited < 2.0` (well under the 2.5s the two-join shape takes) proves the same thing without the flake risk.

PR #9273, opened 2026-09-18
fix(ci): keep the LLVM shim and Python SDK when the jac binary is cached

kugesan1105:
## Summary

`build-kit` went red on main at fdf039230 ([run](https://github.com/jaseci-labs/jac/actions/runs/35300362778/job/105470256949)), in **Materialize typeshed + vendored runtimes**:

```
OSError: Could not locate the LLVM shim 'libjacllvm.so'. Build it with `zig build jacllvm` ...
build-python: build failed
vendor-musl transitive failure
```

Nothing in that commit (client files plus a release note) is involved. The job took the cached-binary path, and that path does not leave behind the things the rest of the job needs.

## Root cause

`vendor-musl` reads as a file copy, but it is a `tool.run` step, and every `tool.run` executes through the source-built interpreter and depends on the SDK build:

```zig
// build.zig
cmd.step.dependOn(self.build_python);
```

In JacPython mode that build runs `bootstrap/python/prepare_native.py`, which compiles the Jac compiler to a native object (`default_codespace="native"`), which loads `libjacllvm.so`. So the chain is `vendor-musl` -> payload tool -> source-built jacpython -> `prepare_native.py` -> LLVM shim.

That stays invisible on a normal build, where the shim was just linked and `.python-build/jacpython` exists, so `build_python` hits its stamp check and returns in a second.

On a **binary cache hit**, `setup-jac` restores the binary and typeshed only: the source-built Python and the shim restores are both gated behind `if: steps.cache.outputs.cache-hit != 'true'`. `build_python` therefore rebuilt the whole SDK, reached `prepare_native.py`, and found no shim. `build-kit` only ever worked because it usually misses the cache and builds the binary itself, which produces both as side effects.

The trigger was a race: the nightly run started on the same SHA at 03:08 and seeded the binary cache at ~03:17 while the CI run sat in the queue for 46 minutes. `build-kit` started at 03:27 and hit it. The parent commit had no nightly on its SHA, so it missed and built everything in-job.

Two reruns (attempts 2 and 3) failed identically. It is deterministic while that entry lives, not flaky.

## Change

- **`setup-jac`** takes `need-build-inputs`. It restores the source-built Python and the shim (plus LLVM, if the shim itself missed) on a binary hit too. Saves stay gated to a binary miss, so the shared cache pool is unchanged.
- **`build-kit`** sets it, and keeps a cheap guard that builds the shim when both restores missed. `fetch-llvm` runs first because the `jacllvm` step does not exist until LLVM is on disk.
- **`build.zig`** stops silently dropping the shim dependency. When `jacllvm` is null and JacPython is on, the SDK build now fails outright with the command to fix it, instead of rebuilding for minutes and dying inside the compiler.

## Validation

Zig 0.16.0 against a checkout with no `.llvm-build` and no in-tree shim, which is the CI job's exact state.

Before, running the failing CI command `JACPYTHON=1 zig build vendor-musl`:

```
build-python: source openssl ... source cpython
build-python: retained 751 CPython source entries
```

It starts a full CPython build. That is the seven minutes that ends in the `OSError`.

After:

```
vendor-musl
+- run .python-build/jacpython/linux-x86_64/python/install/bin/python3.14
   +- run exe build_python
      +- fail failure
error: the JacPython runtime needs the LLVMPY_* shim: run `zig build fetch-llvm` first (or pass -Dshim-bin=<path to libjacllvm.so>)
```

Seconds, and `build_python` never runs. Also checked: `zig build --help` configures clean without `JACPYTHON`, so the host-Python path is untouched; `zig build jacllvm` before `fetch-llvm` reports `no step named 'jacllvm'`, which is why the guard orders them that way.

Not exercised locally: the `if:` expression changes and the expensive branch of the shim guard. Those run for the first time in CI here.

## Note

The failed run on fdf039230 stays failed, but main goes green on merge: `jac/build.zig` is one of the roots of the binary cache key, so this PR moves the fingerprint off the poisoned entry (`f533582cce` -> `2e41b08869`) and the merge commit builds the binary the way the parent commit did.

That also means this PR's own CI takes the cache-miss path, which exercises the guard step and the `build.zig` change but not the cache-hit path the fix is about. That one gets exercised the next time a hit lands on `build-kit`.


kugesan1105:
Confirmed and fixed in 8c73e939e. The finding is correct: `addLlvmShim` reads only what the build itself can produce (`-Dshim-bin`, `-Dllvm-dir`, `.llvm-build`), so a shim restored to its in-tree place path left `jacllvm` null and my new fail step aborted the build, on exactly the path this PR exists to restore.

Reproduced locally before the fix, with a file at the place path and no `.llvm-build`:

```
$ printf 'x' > jaclang/compiler/backends/native/llvm/libjacllvm.so
$ JACPYTHON=1 zig build vendor-musl
vendor-musl
+- run exe build_python
   +- fail failure
error: the JacPython runtime needs the LLVMPY_* shim: run `zig build fetch-llvm` first ...
```

The gate now also consults the place path, which is where `prepare_native.py` resolves the shim from (`_find_shim`), so a restored shim counts as satisfying it. Same three cases after the fix:

- shim at the place path, no `.llvm-build`: proceeds into the SDK build, no abort
- no shim anywhere: still fails in seconds with the same message
- no `JACPYTHON`: configures clean, host Python path untouched

The shim filename and its directory are now single-sourced, so the check and the two `addCopyFileToSource` sites cannot drift apart.

Issue #9290, opened 2026-09-18
Native json.dumps writes every dict value as null and segfaults on a list of ints (regression of #7506 / #8512, still on 0.37.19)


MusabMahmoodh:
**Category:** jac-gap (a wrong answer, not a missing feature: a whole-native program serializes data incorrectly and crashes)

## Summary

In a program that runs whole-native, `json.dumps` of any dict writes every
value as `null`, and `json.dumps` of a list of ints segfaults. #7506 ("Native
json.dumps serializes int values as null (any-boxing) and whole-native binary
segfaults") and #8512 ("json.dumps emits garbage ints") are both closed, and
both symptoms are back on the latest release.

## Reproduce (jac 0.37.19 and 0.37.13, Linux x86_64)

One statement per program, each in its own directory with a minimal
`jac.toml` (`[project]`, `name`, `version`), run with `jac run main.jac`:

```jac
import json;

with entry {
    print(json.dumps({"a": 1}));
}
```

| statement | prints | exit |
|---|---|---|
| `json.dumps({"a": 1})` | `{"a": null}` | 0 |
| `json.dumps({"a": "x"})` | `{"a": null}` | 0 |
| `json.dumps({"a": 1.5})` | `{"a": null}` | 0 |
| `json.dumps({"a": True})` | `{"a": null}` | 0 |
| `n = 1; json.dumps({"a": n})` | `{"a": null}` | 0 |
| `json.dumps({"a": int("1")})` | `{"a": null}` | 0 |
| `json.dumps([1, 2, 3])` | nothing | **139 (SIGSEGV)** |
| `json.dumps(7)` | `7` | 0 |
| `repr({"a": 1})` | `{'a': 1}` | 0 |

The same results on 0.37.19 and 0.37.13. The same `{"a": null}` comes back when
the call is inside a `def`, with the dict bound to a local first, and with a
function-local `import json`.

## Where it happens

`jac check --native-coverage main.jac` reports the call lowered into
`jac/jaclang/runtime/na_stdlib/json.jac`. In `_dump_value` (around line 270 on
current main, 85e8e56746):

- a dict value is read as `d[k]` from `d: dict[str, any]`, so it arrives as
  `any`, none of the `isinstance(value, bool|int|float|str|list|dict)` arms
  matches, and the function falls through to `return "null"`;
- `arr: list[any] = value` over a `list[int]` reads unboxed ints as boxed
  values, which is consistent with the segfault.

[...]

PR #9301, opened 2026-09-18
fix(compiler): type dict view elements and box dict values in native backend

### LINKED ISSUE #9175: Native backend: dict literals store garbage values and key lookup fails for certain literal shapes
## Area

`jac/jaclang/compiler/backends/native` — dict-literal lowering / open-addressing dict repr / native test-harness marshal.

Found while working on #9144 (branch `fix/native-sort-str-key`, PR #9158). **Pre-existing**: reproduced with a fixture containing no `sorted()` calls at all, so it is independent of the sort-comparator changes on that branch.

## Symptoms

1. A natively-compiled dict literal built with `{...}` stores **correct keys** but **garbage values** for certain shapes. Reading the raw dict struct via ctypes (`len/cap/keys/vals/ctrl/used/order`) shows keys decoding as proper UTF-8 strings while `vals[slot]` holds `JacVal`s with impossible tags (e.g. `tag=-1`, `tag=65` (`'A'`), `tag=115` (`'s'`)) and payloads that are heap pointers or ASCII fragments — i.e. the value boxes are uninitialized or misaligned relative to the `{i8 tag, i64 payload}` layout the harness (and dict round-tripping) expects.
2. Jac-level lookups on the same dict fail: `d["key"]` raises `key not found` and `"key" in d` is `0`, even though iteration over `order[]` yields the keys. Insert and lookup evidently disagree about table state.
3. Reading such a dict through the equivalence harness (`run_native_function`) either yields `None` values (unhandled JacVal tag → `unbox_jacval` falls through) or **segfaults** (garbage payload dereferenced for list/dict tags).

## Minimal repro

```jac
# prim_dict_repro.jac — no sorting involved anywhere
def mk_na -> dict {
    n: list[int] = [3, 1, 2];
    return {"aaa": str(1), "bbb": str(1), "ccc": str(n[0])};
}
```

Compile natively (`compile_native`), call `mk_na`, dump the raw struct: keys `'aaa'/'bbb'/'ccc'` decode cleanly at slots 7/5/1 of a cap=8 table; `vals[]` at those slots is garbage; `run_native_function` returns `{'aaa': None, 'bbb': None, 'ccc': None}`.

A `jac run` driver that does `print(len(d))` then `d["aaa"]` prints `3` and then raises `key not found`.

## Shape sensitivity (why it hid so long)

Deterministic per fixture content, but the trigger is not yet pinned:

| dict shape | marshal result |
|---|---|
| 10 keys, values = str/int **list-element loads** (`prim_sort_key.jac`, in-tree) | OK |
| 4 keys, mixed values (str() call, list elem, int literal, str() call) | OK |
| 3 keys, values = `str(...)` calls | garbage |
| 9 keys, values = int arithmetic on sorted elements | garbage |
| 12/13 keys, values = str list-element lookups | garbage in some function-body variants, OK in others |

The in-tree equivalence fixtures all use the passing shapes, which is why the suite stays green until a fixture author trips one of the failing shapes and sees a baffling `backend=None` (or a segfault) from the harness.

## Suspected area

- `na_ir_gen_pass.impl/hash_core.impl.jac` (`_emit_hash_core_helpers`: `insert`/`find`/`rehash`) — insert and lookup disagreeing about table state fits the "keys readable, values garbage, lookup misses" triad.
- The val-array allocation in `_container_layout_hash`/`_emit_container_new_fn` (vals stride vs `sizeof(JacVal)` = 16 with alignment).
- The static-hash path (`_static_key_op`, `_static_str_hash`) diverging from the runtime FNV probe sequence.

## Impact

- Jac-level dict lookup (`in`, `d[k]`, `.get`) on natively-built dict literals is unreliable for affected shapes.
- Test-harness lane (`run_native_function`) returns `None` values or crashes.
- New fixture authors lose time to unattributable `backend=None` mismatches (this cost roughly a day during #9144).

## Suggested next steps

1. Dump the emitted LLVM IR for `mk_na` and check the `vals` GEP element size against the dict struct field type.
2. Compare `find`'s probe sequence (ctrl bytes, h2) against what `insert` writes when the hash is supplied statically.
3. Add a debug helper that, given a live dict, dumps `ctrl[]` and `order[]` alongside keys/vals to catch insert/lookup divergence directly.

chess10kp:
## Summary

Fixes #9175.

`dict[K, V]` in the native backend stored values lowered through `_elem_storage_llvm`, which produced storage that did not match what insert wrote for non-primitive values — dict literals kept correct keys but held garbage/uninitialized value boxes (impossible `JacVal` tags, heap-pointer payloads), Jac-level lookups on the same dict missed, and the equivalence harness returned `None` values or segfaulted. Reproduction was shape-sensitive, so existing fixtures stayed green.

This PR makes the repr decision single-owner and the storage↔semantic bridge explicit:

- **Repr**: `_lower_class_type` / `_lower_untyped_annotation` / `_codegen_dict_val` now pin dict value storage to `jacval` for every value type (keys stay concrete per-K since hashing/equality need typed ops). All dicts share one helper family per key type (`<k>:jacval`).
- **Views**: `dict_items` / `dict_keys` / `dict_values` receivers get element layouts derived from their own type args (`_dict_view_elem_llvm`), so `d.items()[0]` stays typed as a `[K, V]` pair tuple instead of degrading to the string-index fallback and reading tuple bytes as characters.
- **Bridge**: dict method emitters (get/pop/popitem/values/items/setdefault) receive the semantic value type via `NativeEmitCtx.semantic_val_type` and unbox `jacval` → `V` at their result phi; `_codegen_dict_access` unboxes to the trailer's lowered type; binary ops unbox a single `jacval` side against a concrete other side; match casts unbox to the declared element type.
- **Ownership**: boxed assignment/walrus paths retain/release through the element ops, tuple-box release/trace handle `jacval` fields, and `hash_core` tuple-key hashing covers `jacval` fields.

## Testing

- `jac test jaclang/compiler/tests/test_prim_equivalence.jac` — 103 passed (rebased onto current main; includes the equivalence fixtures added since the original run).
- Verified the issue's failing dict shapes (values from `str(...)` calls / arithmetic) round-trip through `run_native_function`.

## Release note

Fragment added at `release_notes/unreleased/jaclang/<PR#>.bugfix.md` (this PR's number).


MusabMahmoodh:
Heads-up on an overlap with #9305 (for #9290, native `json.dumps` of typed containers). This PR makes dict value storage `jacval` for every value type; #9305 instead re-boxes a typed dict into its `jacval` flavor where it crosses into `any` (and back on unbox). Once this PR lands, the dict half of #9305 is redundant, and the two touch the same spots in `dicts.impl.jac` / `types.impl.jac`. The list half of #9305 is not covered here: a `list[int]` passed as `any` still keeps raw slots (that is the `json.dumps([1, 2, 3])` segfault), plus the optional-target path and a null guard. I am happy to rebase #9305 onto this one and drop its dict reflavor, keeping only lists, so the two do not fight. Would that work for you?

Issue #9286, opened 2026-09-18
Three checker inference gaps that reject correct code: sum() is always int, a None guard narrows an assignment target, and **{...} kwargs pick the wrong overload


MusabMahmoodh:
**Category:** jac-gap (each rejects correct code, and each error lands away from its cause)

Three independent inference gaps, each with its own reproducer. All were run on
jac 0.37.19 and 0.37.13 (Linux x86_64) on 2026-09-18 with identical results.

- [ ] 1. `sum()` over a comprehension is typed `int` regardless of its elements
- [ ] 2. A `None` guard on a subscript narrows the assignment target
- [ ] 3. Keyword arguments passed as `**{...}` are invisible to overload resolution

## 1. `sum()` over a comprehension is typed `int` regardless of its elements

```jac
def total(prices: list[float], shipping: float) -> float {
    value = sum([p for p in prices]);
    value += shipping;
    return value;
}
```

```
error[E1001]: Cannot assign float to int
  --> main.jac:3:5
    3 |     value += shipping;
```

The list is a `list[float]` and the sum is a float. The error lands on the
correct `+=`, not on the inference that produced `int`. Workaround:
`value: float = sum(...)`.

## 2. A `None` guard on a subscript narrows the assignment target, so filling the slot is rejected

```jac
def fill(row: dict[str, any], title: str) -> dict[str, any] {
    if row["title"] is None {
        row["title"] = title or None;
    }
    return row;
}
```

```
error[E1001]: Cannot assign str | NoneType to NoneType
  --> main.jac:3:9
```

The guard narrows `row["title"]` to `NoneType` for reads, and the checker then
uses that narrowed type as the type the slot will accept. Narrowing should not
change what an assignment target accepts: the declared value type is
`any`. Workaround: test `row.get("title") is None`, which is only equivalent
when the key is known to be present. Six sites in one port.

## 3. Keyword arguments passed as `**{...}` are invisible to overload resolution

```jac
import subprocess;

with entry {
    finished = subprocess.run(["echo", "hi"], **{"capture_output": True, "timeout": 30});
    print(finished.stdout.decode("utf-8"));
}
```

```
error[E1030]: Type "str" has no attribute "decode"
  --> main.jac:5:27
```

With `capture_output=True` and no `text=True`, `stdout` is `bytes` and
`.decode()` is correct. The checker, not seeing the unpacked keywords, picked
the `str` overload. When the keys of a dict literal are statically known, they
[...]

Issue #9297, opened 2026-09-18
Tracking: type-check performance — the gap is the IR representation, not the checker


marsninja:
Tracking issue for the work needed to close the gap between `jac check` and the current frontier of Python type checkers. A follow-up comment carries the full report this issue is derived from, verbatim, including methodology.

**The finding that reframes the effort: this is not primarily a type-checker project. It is an IR representation project, and the type checker is its largest consumer.** The checking passes are 3% of a profiled check. 27% is the OSP graph runtime, reached almost entirely through AST *field* accessors.

---

## 1. Where we stand, measured

All numbers below were taken on one machine (128 cores, 125 GB), cold, no cache reuse. Corpora are deliberately close in size: CPython `Lib/` is 286,212 lines, `jaclang/compiler` is 290,990 lines.

| checker | mode | corpus | lines | wall | core-sec | lines/s/core | peak RSS |
|---|---|---|---|---|---|---|---|
| pyrefly 1.2.1 | `-j1` default | CPython `Lib/` | 404,135 | 1.12s | 1.14 | **354,000** | 226 MB |
| pyrefly 1.2.1 | `-j1` `--preset strict` | CPython `Lib/` | 420,639 | 3.02s | 3.11 | **135,000** | 321 MB |
| pyrefly 1.2.1 | all cores | CPython `Lib/` | 404,135 | **0.24s** | 2.97 | — | 317 MB |
| zuban 0.10.0 | default | CPython `Lib/` | 286,212 | 1.34s | 24.4 | 11,700 | 224 MB |
| ty 0.0.82 | default | CPython `Lib/` | 286,212 | 4.70s | 47.1 | 6,100 | 613 MB |
| `jac check` | `-j1` | `jaclang/compiler` | 290,990 | ~2,110s* | ~2,110 | **~138** | 10.9 GB |
| `jac check` | 8 workers | `jaclang/compiler` | 290,990 | **956s** | 4,066 | **72** | 7.8 GB |

\* extrapolated: the serial run reached 42% of 517 files in 885 core-seconds before being stopped.

Per core, whole-project: pyrefly is ~2,600x faster than our serial check (~1,000x in strict mode), zuban ~85x, ty ~44x. Memory is ~35x.

[...]

Issue #9293, opened 2026-09-18
A .test.jac runs in its module's namespace but is checked as if it had its own, so a shadowed helper silently runs the module's function and a self-import fails as a fake circular import


MusabMahmoodh:
**Category:** jac-dx (the checker and the runtime disagree about which definition a name refers to; one symptom can leave a test green while it runs code its author did not write)

## Summary

A `foo.test.jac` is loaded into `foo`'s module namespace at runtime. `jac check`
does not model that: it validates the test file against its own top-level
definitions and its own imports. Two consequences, both reproduced below.

1. **A test helper with the same name as a module function is silently replaced
   by the module's function at runtime**, while `jac check` validated every call
   against the test's own definition. With mismatched signatures the test raises
   `TypeError`; with matching signatures it passes or fails while running the
   wrong code, and nothing reports it.
2. **A test file that imports from its own module** passes `jac check`, then
   fails at run time with `ImportError: cannot import name ... from partially
   initialized module ... (most likely due to a circular import)`. There is no
   circular import between modules; the test re-enters its own module during
   initialisation.

## Reproduce 1: the shadowed helper (jac 0.37.19; identical on 0.37.13)

`store.jac`:

```jac
def _seed(kind: str) -> str {
    return "module:" + kind;
}

def:pub describe(kind: str) -> str {
    return _seed(kind);
}
```

`store.test.jac`:

```jac
def _seed(kind: str) -> str {
    return "test:" + kind;
}

test "the test's own helper is the one that runs" {
    assert _seed("x") == "test:x";
}
```

```
$ jac check store.jac        -> rc=0
$ jac check store.test.jac   -> rc=0
$ jac test store.test.jac
FAILED   the test's own helper is the one that runs
    assert _seed("x") == "test:x";
E   AssertionError:
```

The test called its own `_seed` and got the module's.

## Reproduce 2: the self-import

`widget.jac`:

```jac
glob LIMIT: int = 3;

def:pub clamp(n: int) -> int {
    return min(n, LIMIT);
}
```

`widget.test.jac`:

```jac
import from widget { clamp, LIMIT }

test "clamp caps at the limit" {
    assert clamp(10) == LIMIT;
}
```

```
$ jac check widget.jac        -> rc=0
$ jac check widget.test.jac   -> rc=0
$ jac test widget.test.jac
ImportError: cannot import name 'clamp' from partially initialized module
'widget' (most likely due to a circular import)
```

## Scale, from one real codebase

[...]

Issue #9292, opened 2026-09-18
Inside an obj method, a bare call binds to a sibling method instead of the module function of the same name


MusabMahmoodh:
**Category:** jac-gap (semantics diverge from Python; no diagnostic points at the cause)

## Summary

Inside a method of an `obj`, a bare name that matches both a sibling method and
a module-level function resolves to the **sibling method**. Python resolves a
bare name through local, enclosing and global scope, and a class body is not an
enclosing scope, so Python reaches the module function. The Jac program is
rejected at compile time, and the message describes an argument count, not a
name resolution.

## Reproduce (jac 0.37.19, Linux x86_64; identical on 0.37.13)

`main.jac`:

```jac
def lookup(n: int) -> int {
    return n + 1;
}

obj Toolbox {
    has seed: int = 41;

    def lookup() -> int {
        return lookup(self.seed);
    }
}

with entry {
    print(Toolbox().lookup());
}
```

```
$ jac check main.jac
✖ Error: error[E1051]: Too many positional arguments
  --> main.jac:9:23
    9 |         return lookup(self.seed);
      |                       ^^^^^^^^^

$ jac run main.jac
✖ Error: error[E1051]: Too many positional arguments
```

The equivalent Python prints `42`.

## Why it matters

This is the natural shape of a byLLM toolbox: a tool method is named exactly
like the function it wraps, because the method name is what the model sees.
In one port (the Amplify migration, about 160k lines of Jac), six of the eight
tool methods on one toolbox had this shape and the module could not be
imported at all. The other two had been renamed by hand, which suggests someone
hit it earlier and did not know why.

## Asks, either one

1. Resolve a bare name inside a method the way Python does (local, enclosing
   function, module, builtins), and require `self.` for a sibling method.
2. If the current binding is intended, say so in the diagnostic: "`lookup`
   resolved to the method `Toolbox.lookup`; the module function of the same name
   is shadowed here. Qualify it, or rename one of them."

## Verified / hypothesis

Everything above was run on 0.37.19 and 0.37.13 on 2026-09-18. Whether the
binding is intentional is not known; that is the question the issue asks.

Re-verified 2026-09-18 on main `0b5d8531e2` (a 0.37.19 binary using a clean main checkout as `[dev] jaclang_source`, dev-mode banner confirmed): reproduces as described. No existing issue or PR covers it.

PR #9312, opened 2026-09-19
fix(checker): sum() over floats is a float (Iterable element types, T | Literal unions)

MusabMahmoodh:
Refs #9286 (item 1: `sum()` over floats typed `int`).

## Problem

```jac
def total(prices: list[float], shipping: float) -> float {
    value = sum([p for p in prices]);
    value += shipping;   # E1001 Cannot assign float to int
    return value;
}
```

The comprehension is not the cause: `sum(prices)` on a `list[float]` is typed
`int` too. Two independent checker bugs combine:

1. **`Iterable[X]` ignored `X`.** `assign_type` has a name-based shortcut that
   accepts any instance with `__iter__` for `Iterable`, `__next__` for
   `Iterator`, and so on, without looking at the type argument. So
   `wrong: Iterable[int] = prices` (a `list[float]`) was accepted, and
   typeshed's first `sum` overload, `(Iterable[bool | _LiteralInteger]) -> int`,
   matched every iterable.
2. **`T | Literal[0]` collapsed to `int`.** In an annotation, `Literal[0]`
   evaluates to a literal instance type, which `is_annotation_type` rejects,
   so `T | Literal[0]` fell through to the runtime operator path, where
   `int.__ror__` accepts the unbounded `T` and answers `int`. That is `sum`'s
   second overload's return type, `_SupportsSumNoDefaultT | Literal[0]`, so
   even the right overload returned `int`. `float | Literal[0]` only worked
   because `float` has no `__or__`, which reached the union fallback.

## Fix

- `type_evaluator.impl.jac`, the protocol shortcut: for a parameterized
  `Iterable[X]` / `Iterator[X]` whose `X` has no type variable, a builtin
  `list` / `tuple` / `set` / `frozenset` source's element type must be
  assignable to `X`. Other sources (dicts, strings, generators, other
  iterables), generic targets and the other protocols in that table keep
  today's behaviour: a broader check turned real calls such as
  `requests.get(..., params={...})` into false errors through an unrelated
  `SupportsItems` gap.
- `operations.jac`: in `a | b`, an operand that is syntactically a
  `Literal[...]` subscript counts as a type form, so the union is built
  instead of calling `__or__`.

With both, `sum(prices)` is `float | Literal[0]` and the issue's function
checks clean.

## Tests

`jac/tests/compiler/passes/test_checker_pass.jac`, new table row
`sum_element_type` with fixture `checker/checker_sum_element_type.jac`: the
issue's function and `sum(prices)` followed by `+= float` (`# <-- Ok`), a
generic `-> T | Literal[0]` used as `float | int` (`# <-- Ok`),
`Iterable[float]` / `Iterable[str]` from matching lists (`# <-- Ok`), and
`Iterable[int] = list[float]` (`# <-- Error`, pinned to "Cannot assign
list[float] to Iterable[int]"). On main the row gets 2 errors, the two false
`int` errors and not the real one; with this change exactly the expected one.

Run on a 0.37.19 binary with this tree as `[dev] jaclang_source`: every file
in `jac/tests/compiler/passes/` (51 test files) passes. One of them,
`test_decl_impl_match_pass.jac`, had its runtime test "abstraction bug" fail
once while other jobs loaded the box and pass alone (18 passed); the same
happens without this change. `jac check` rc=0 and `jac fmt --check
--lintfix` rc=0 on every changed file.

Tree sites the stricter check found, fixed here: `descr_arg` in the LLVM IR
printer was annotated `-> object` and fed `str.join` (now `-> str`), and
`precompile_bytecode.jac` put two app types in one list (now `list[any]`).


MusabMahmoodh:
CI's build-kit precompile found two errors under the stricter `Iterable[X]` check, both handled in 07741b9e42:
- `byllm/parallel.jac`: `as_completed(futures)` against `Iterable[_AsCompletedFuture[_T]]`. The element check ran before type-variable solving; it now applies only when the target's element type has no type variable. Fixture gained a nested generic target (`Iterable[list[T]]`) that must pass.
- `compiler/backends/native/llvm/ir/instructions.jac`: `descr_arg` was annotated `-> object` and fed `', '.join(...)`, a real type error the old check let through; it returns `str`, now annotated so.
Both files `jac check` rc=0; the fixture has exactly its one expected error; test_checker_pass.jac 252 passed.

MusabMahmoodh:
The previous head failed jac-check and test-runtime; narrowed in 325c72bb86.
- `scale/observability/prometheus_client.jac` (and 3 test_publish tests through it): `requests.get(..., params={"query": q})`. `params` is a union that includes `Iterable[tuple[...]]`; a `dict` is iterable, so the old shortcut accepted it there, which masked that the `SupportsItems` branch does not match a dict literal. With the element check that became a false error on an extremely common call.
- `compiler/passes/impl/pyast_load_pass.impl.jac:945`: `''.join(generator of str | bytes)`.
So the element check now applies only when the source is a builtin `list`, `tuple`, `set` or `frozenset` (still enough for `sum()` of a `list[float]`); dicts, strings and generators keep the previous behaviour. The `SupportsItems` / dict-literal gap is separate and I have not changed it.
- `dist/precompile_bytecode.jac:178`: `entries = [selected]` holds an `AppSpec` and is extended with `service_apps(...)`, which is annotated `-> list[AppConfig]` but, called with a workspace, yields `AppSpec`s (the loop reads `.entry`, which `AppConfig` does not have). I annotated `entries: list[any]` rather than change `service_apps`'s annotation, which has other callers; that annotation looks wrong and is worth a follow-up.
Rig: the five tree files CI named check clean; fixture has exactly its one expected error; test_checker_pass.jac 252 passed; test_checker_bugs.jac 80 passed.

MusabMahmoodh:
@kugesan1105 you work this part of the checker (#8450, #9275, #9278). Could you review this batch? These are 8 small checker fixes, each from a real porting failure in a production app (the issues carry the repros):

- #9307: a name first bound to `None` reads as the union of its bindings (#9288)
- #9308: `jac check <file>` says its imports were read as interfaces only (#9281)
- #9309: a declared but missing dependency is an error, E1136 (#9289)
- #9310: a bare name in a method body does not bind to a sibling method (#9292)
- #9311: a `.test.jac` importing from its own module is an error, E1137 (#9293)
- #9312 (this one), #9313, #9314: the three inference gaps in #9286

All 8 have main merged in, including the weekend checker PRs (#9315, #9328, #9329, #9345, #9346, #9359); CI is re-running now. If any of them now overlaps a direction #9297 is taking, say so and I will close it rather than rebase it again.

PR #9323, opened 2026-09-19
Consolidate native stdlib access through the shared CPython runtime

marsninja:
## Description

Native Jac substitutes partial standard-library implementations even though its bundled CPython runtime can provide the original modules, objects, defaults, and exceptions. This draft replaces that duplication with one retained Python-object provider shared by stock CPython and JacPython.

The shared object API and ownership layer now support ordinary hosted imports, calls, scalar operations, iteration, and original Python exceptions. Native exception ABI 2 propagates owned errors through normal returns with frame and temporary cleanup. JIT entries, shared libraries, generated stubs, and source artifacts share the Python-runtime requirement and GIL policy.

Module import bindings live in interpreter-owned state and initialize under CPython's reentrant module lock. Failed initialization clears partial bindings and remains retryable. JIT and shared-library initialization preserve dependency order and stop on errors. Targeted regressions cover recursive and concurrent entry and sequential shared-GIL subinterpreters.

Import plans now retain each statement occurrence and lexical binding. Repeated imports execute in order, skipped local imports raise `UnboundLocalError`, and shared runtime helpers use module-qualified native symbols. Checked text/bytes conversions preserve Unicode and embedded NULs and reject incompatible Python objects. Hosted-import tests require successful native compilation.

Python object protocols now handle contextual literal containers, slices, attribute and item assignment, and in-place mutation with alias and evaluation-order regressions. Imported C aliases and qualified calls retain the declared C symbol and ABI.

ELF relocation tables now place data relocations immediately before PLT relocations, and Linux links libpthread explicitly. Executables and shared libraries preserve eager symbol resolution on glibc 2.17.

The LLVM C backend patch now lowers `llvm.ptrmask`, including masks narrower than the pointer representation. The C export CI job runs a focused executable regression before exporting applications.

The pre-interpreter launcher now uses a private Zig standard-library archive for filesystem operations, SHA256, streaming zstd, and PAX tar extraction. Jac retains embedding and materialization policy. The archive uses existing floor resolution and payload staging; the launcher no longer depends on public stdlib replacements. Platform-specific declaration sources now follow the same target selection as import resolution.

Standalone hosted native applications now reuse the shared-library emitter, sealed JAB manifests, fused runtime overlay, and existing GIL-aware native-call boundary. Project native bundles use the same shared payload; rebuilding a fused app replaces its overlay instead of nesting runtimes. Publication is atomic.

## Remaining migration

This is still a draft. Import edge cases, remaining conversions and object protocols, container/callback ownership, entry-block storage, annotated Python fields, and the remaining interoperability regressions remain in progress. The public native stdlib replacements and compiler interceptions are now removed. Independent-GIL concurrency is not established. The separate JacPython replacement-module project is retained.

## Validation

- Import/initialization milestone: **17 targeted tests passed**, including import bindings, iteration, JIT initialization, and shared-library exception/retry behavior.
- Import and buffer conversion milestone: **3 new focused regressions and 11 existing import/ownership regressions passed**. The native IR root, import plan, shared provider, symbol naming helper, and affected test support passed targeted type checking.
- Patched LLVM C backend built against LLVM 22.1.8. The executable pointer-mask regression passed for both 64-bit and 32-bit pointer index widths, including preservation of upper address bits.
- Literal construction and mutation regressions passed, as did the imported C-alias regression and native IR root type checking.
- Six focused ELF relocation/weak-symbol tests passed. A launcher using these fixes extracted the real 130 MiB embedded payload and ran on CentOS 7 with `LD_BIND_NOW=1`.
- Bootstrap milestone: three private Zig extraction tests, six real-launcher runtime-cache tests, floor staging, platform resolution, and affected compiler type checks passed. Linux and macOS ARM64 launchers linked, and the full Zig stub build passed. macOS execution remains a CI responsibility.
- Earlier targeted checks cover reference ownership, exception cleanup, generators, iterators, `nogc`, embedding, runtime requirements, and typed compiler/provider interfaces.
- Both stock CPython and JacPython were rebuilt from source and passed runtime smoke checks. A shared native library passed nine stdlib probe groups under both runtime variants. These establish bridge behavior, not complete stdlib conformance.
- Only targeted tests were run locally. Final-head CI is pending. Two JacPython bootstrap crashes were reproduced and fixed: exception polling without a thread state, and runtime evaluation of erased type-only imports. The subsequent source build passed all seven steps and runtime smoke checks.

- Stdlib cleanup: 5 focused delegation/platform tests and 7 routing tests passed; the native IR root and 5 affected routing modules passed type checking. Removed over 10,000 lines of duplicate implementations and routing.

- Packaging milestone: 4 focused tests passed for sealed entry execution/arguments, original Python exceptions, project native payload metadata, and overlay replacement. The 7 targeted type checks across affected modules passed. A fresh 15-step bundled-runtime build passed. Its native application ran outside the checkout with source overrides disabled, and a warm run passed with an empty PATH.

- Python-backed globals: 3 focused tests passed for source-ordered import/initialization, aliasing/rebinding, cleanup and retry after initialization failure, and isolation across subinterpreters. Explicit native scalar annotations retain native storage. The compiler and tests passed targeted type checking.
- Checked numeric conversion: rebuilt stock CPython passed all 7 build steps and the expanded smoke checks (integers, floats, bools, custom numeric protocols, type errors, and overflow).


marsninja:
Python Perf is way too bad for even this to make sense. WHY PYTHON WWHWHHYYYYYY

PR #9332, opened 2026-09-19
refactor(native): single-compile, per-artifact linking for native builds

chess10kp:
# PR 2/4 — refactor(native): single-compile, per-artifact linking for native builds

## Summary
- `native_build` is split into `_compile_native_module` (the expensive Jac-to-LLVM-IR compilation, stashed on a `NativeBuildContext`) and `_link_native_artifact` (per-artifact IR finalization, verification, optimization, object emission, and platform linking). Several artifacts can now come from one compilation.
- Backend seam warnings move from `jaclang.cli.console` to `sys.stderr`, so the compiler backend no longer imports the CLI layer.
- Verified no process-env mutation exists in the build path (the pure-Python linkers take explicit parameters), so no env isolation is needed.

Stack context: addresses the review in #8370 (comment 5426876586) -- part 2 of 3 (#8370 -> #9332 -> #9333). No CLI surface changes; #9333 builds the sealed pair on this API.

## Test plan
- [x] `jac build main.jac --as native` produces a working binary through the split pipeline
- [x] `tests/compiler/test_interop_manifest.jac` green


chess10kp:
The first CI run ([35476610715](https://github.com/jaseci-labs/jaseci/actions/runs/35476610715)) failed in `build-kit`: the shipped-binary (bootstrap) build hit `NameError: name '_compile_native_module' is not defined` at `native_build.impl.jac:117`, in both the `jac-stub` and payload-kernel steps, across all three in-job retries.

Root cause: `_compile_native_module` and `_link_native_artifact` were declared with `impl` but have no matching `can`/`def` decl anywhere, so bootstrap transpilation never emitted them as module globals. Dev-source transpilation does bind decl-less `impl`s, which is why local smoke builds passed. Fixed in b29d8c405 by defining both helpers with `def` (matching the wording on entry-lib-dispatch, where this path is green); CI is re-running.

chess10kp:
Now the middle of a PR stack: #8370 → #9332 → #9333. Force-pushed `5498d3240` → `4019f23d6`: same three commits (refactor split, def bind fix, fmt) rebased on top of #8370's na→sv fix. `skip-release-notes-check` is added because the diff vs main carries #8370's fragment until that PR merges — the `9332.bugfix.md` fragment is still in the branch and ships with it. Merge after #8370.

PR #9333, opened 2026-09-19
feat(native): host-driven entry-library lane for sealed native jabs

chess10kp:
# PR 3/4 — feat(native): host-driven entry-library lane for sealed native jabs

## Summary
- Entry libraries are now built with `bind_shared_callbacks`: every na->sv callback lowers to a host-fillable `__jac_bind_*` dispatch slot, so the library carries **no unresolved Python symbols** and loads cleanly under `RTLD_NOW`. `__jac_shared_init` registration is deferred to the loading host, which sets `__jac_argv`/`__jac_argc` first.
- `native_build_sealed_pair` compiles **once** and links both artifacts from that single compile (was: up to three full compiles). Interop apps (`native_imports`) seal as the bound entry library alone, recording the callback contract -- name, ctypes-wire signature, and source module per callee -- in `<lib>.interop.json`.
- `run_jab_image` loads interop jabs **in-process** through `NativeLibrary` (registering callbacks before deferred init, then invoking `jac_entry`) instead of routing them to the JIT. Non-interop jabs stay exe-first with the library as fallback. Once the entry has started, failures abort rather than re-running side effects.
- The `__jac_argc` presence check is a define-site regex (`argv_globals_defined`), not a substring probe; platform-gated tests skip via `testskip()` so strict CI can tell "skipped" from "passed".

This is the dispatch-table design suggested in the review of #8370 (comment 5426876586): the dlopen lane becomes the real path for interop apps rather than dead code. Part 3 of 3 (#8370 -> #9332 -> #9333); this PR builds on #9332 and its diff shows those changes until it merges.

## Test plan
- [x] `tests/cli/test_native_entry_lib.jac` -- 14 tests: argv-before-init ordering, in-process run, interop seal (exe dropped, sidecar recorded), execv-first for non-interop
- [x] Interop probe smoke: native_imports app seals lib-only and dispatches in-process


chess10kp:
Status of the red run on `2e3eab71b` and what was pushed since:

- **`jac-check`**: `jac fmt` wrapping nits in `native_build.impl.jac` (4 hunks) — fixed in `c016a6fbd`.
- **`test-runtime`: `sealed pair on a native_imports app drops the exe and keeps the lib` failed with `E5092: Native lowering failed for function 'py_double'`.** Root cause: the sealed-pair gate consumes `gen.interop_manifest.native_imports`, but on this branch a native-pinned caller importing from a server-pinned provider never forms that entry — `BoundaryAnalysisPass`'s pin path only admitted client-context callers (the bug fixed in #8370). Fix: `c146d29f6` stacks this branch on `jac-one-compiler`. Verified locally: pre-merge the E5092 reproduces deterministically; post-merge `tests/cli/test_native_entry_lib.jac` passes 14/14 and `tests/compiler/test_interop_manifest.jac` 6/6.
- **`Contribution Checks`** (introduced by that merge): the release-note checker only accepts fragments named after the PR's own number, so `2180dc851` drops the `8370.bugfix.md` the merge carried — that note ships with #8370 itself; git keeps main's copy when the stacked branches land.
- **`test_cache_authority` `reclaimed_bytes`**: single occurrence; the same test is green on #8370's and #9332's runs in the same window. Watching the fresh run before calling it anything more than a flake.

Note: the run on `c146d29f6` will show `Contribution Checks` red — it is superseded by the run on `2180dc851` (couldn't cancel it; no admin rights on the base repo).

chess10kp:
Replaced the earlier merge with a linear stack: #8370 → #9332 → #9333. Force-pushed `2180dc851` → `ea7ce2d16`: the merge commit and the fragment-drop workaround are gone; the feature commit and its fmt fix cherry-picked cleanly onto the stacked native-build work (feature file byte-identical to the locally verified tree — seal tests 14/14, `test_interop_manifest.jac` 6/6). `skip-release-notes-check` added for the same transient lower-layer-fragment reason; `9333.feature.md` remains in the branch. Merge last.

PR #9350, opened 2026-09-20
fix(config): stop the deploy when a shipped setting still reads ${VAR} (#8773)

akindu-k:
Refs #8773. Not `Fixes`, because `${VAR:+alt}` and `${VAR:0:3}` still pass through verbatim; those are tracked with the other parser gaps in #9184.

Supersedes #9045, which is being closed in favour of this branch.

## The problem, in plain terms

`jac.toml` lets you point a setting at an environment variable instead of hardcoding it:

```toml
[scale.secrets]
OPENAI_API_KEY = "${OPENAI_API_KEY}"

[scale.kubernetes]
namespace = "${K8S_NAMESPACE:?set K8S_NAMESPACE}"
```

When Jac reads the file it is supposed to swap in the real value from your machine. If the variable is not set, both forms are documented to be an error, and the `:?` form exists for exactly one reason: to stop a deploy when something important is missing.

Neither of them actually stopped anything.

The function that walked the config caught the error and handed back the raw text. So when `OPENAI_API_KEY` was not set, the setting did not fail. It quietly became the literal string `${OPENAI_API_KEY}`, dollar sign, braces and all.

Nothing warned you. `jac scale deploy --dry-run` printed a clean plan, and the real deploy then shipped that literal text into the cluster:

```yaml
kind: Secret
metadata:
  name: interp-repro-secrets
  namespace: ${K8S_NAMESPACE:?set K8S_NAMESPACE}
stringData:
  OPENAI_API_KEY: ${OPENAI_API_KEY}
```

The pods came up holding an API key whose value was the text `${OPENAI_API_KEY}`. Everything looked fine until the first call to the provider, which failed with an authentication error. The error pointed at OpenAI, so the natural thing to do was go and check the OpenAI account, rotate the key, read their status page. The actual cause was a variable that was never exported on the machine that ran the deploy, several steps earlier and in a different place entirely.

That is the whole bug: the one guard rail for "this must be set before you deploy" was the thing that failed silently.

## Why the earlier attempt (#9045) was not the answer

The obvious fix is to let the error through: if a variable is missing, fail. #9045 did that, and it caused two new problems.

Every command reads `jac.toml`, not just the deploy. So an unset deploy secret suddenly broke `jac check`, `jac fmt` and `jac lsp`. You could not format a file on your laptop without first exporting production secrets, which nobody should have on their laptop.

It also broke two jac-scale paths Musab pointed out in review: pods read the shipped `jac.toml` at boot under a different environment than the deploy ran in, and some scale commands get their variables from a project `.env` that is read *after* the config loads. Both became hard failures.

The real trouble was that interpolation worked out "did this resolve?" at load time and then threw that answer away. Once the answer is gone, the only choices left are fail for everyone or fail for nobody. Neither is right, because whether a missing variable matters depends entirely on what you are about to do with it.

## How this fixes it

Keep the answer instead of throwing it away.

Loading the config never fails now. As it resolves values it writes down every reference it could not fill in, along with which setting holds it, on `JacConfig.unresolved`. Loading stays quiet, so `jac check` and `jac fmt` never need deploy-time variables exported.

The command that actually uses the value decides what to do about it. Right now that is the deploy: before it builds any manifests, it looks at the values it is about to ship, the entries in `[scale.secrets]` and the target settings in `[scale.kubernetes]`, and refuses to continue if any of them is still unresolved text. It names the setting that needs the variable:

```
jac.toml reads environment variables this machine does not have, so the deploy
would ship these settings as literal ${...} text:
  scale.secrets.OPENAI_API_KEY: OPENAI_API_KEY is not set
  scale.kubernetes.namespace: K8S_NS is not set (set K8S_NS)
```

This is the same split the schema validator already uses in this codebase: `validate_against_schema` gathers findings, `report_schema_findings` applies the policy. Here `resolve_env_refs` gathers and `report_env_findings` reports.

One detail worth calling out: the check reads the actual values being shipped, not a list of section names. That matters because a variable can reach a deploy setting through `[apps.<name>]` or an active profile overlay rather than being written in `[scale.secrets]` directly. Checking the shipped values catches those the same way, and it also means a setting the deploy never sends is never a reason to stop.

`${VAR:-default}` is unchanged and still fills in the default without complaint.

## Tests

Both are extensions of the existing table-driven tests, in the files that already cover this code, run in-process.

| File | Before | After |
|---|---|---|
| `jac/tests/project/test_config.jac` | 129 | 130 |
| `jac/jaclang/scale/tests/deploy/test_pod_env.jac` | 41 | 42 |

One test each, because recording references at load and refusing them at deploy are two genuinely new code paths, and they live in two different suites. The config test drives real toml through the real loader and covers a bare reference, a `:?` reference, a default, a shell positional like `${1}` that is not a variable name, a reference inside a list, and one reaching a setting through `[apps.<name>]`. The deploy test covers an unset secret, an unset target setting, the variable exported, and a default, with `--dry-run` on, since dry run reporting a clean plan was the original complaint.

Full suites: `jac/tests/project/` 332 passed. `jac/jaclang/scale/tests/deploy/` 386 passed with 9 pre-existing failures, all of them the missing `python-dotenv` capability, identical on `main` at de31219c7 and unrelated to this change.


MusabMahmoodh:
@akindu-k a post-merge read, two things on this surface. #9364 does not cover either of them.

1. `_check_deploy_settings_resolve` (`manifest_builder.jac:718-728`) builds `shipped` from the secrets and the top-level string fields of `k8s_config` only. So `[apps.orders.scale.env] OPENAI_API_KEY = "${OPENAI_API_KEY}"` with the variable unset still lands in the Deployment as the literal, with a clean exit, dry run included. The pod check does not catch it either: `_pod_environ` (`:809-811`) copies that container env value into the simulated pod env, so it counts as set. That is the #8773 symptom again, one level down. Gateway `env`, `labels` and the annotation maps have the same gap. `_check_pod_config_resolves` already has `deployments` and `user_secret` in hand, so could `shipped` be every string leaf of the rendered manifests instead of a hand-picked list?

2. For a `[scale.kubernetes]` value, the hint here ("Export each one or put it in the project .env") leads straight into the next check. Export `K8S_NS` for `namespace = "${K8S_NS:?set K8S_NS}"` and this check passes, then #9199's pod check refuses it, because the pods never get `K8S_NS`. The "host-only values" case at `test_pod_env.jac:569` pins that behaviour, so it looks deliberate on the #9199 side. Either way, the two checks need to agree on which settings the host resolves and which the pod does. Is that split written down anywhere? If not, I think it is worth an issue.

Issue #6458, opened 2026-06-06
jac2js: `Generating JavaScript failed: list index out of range` on some components (shadcn sidebar)


marsninja:
## Summary

`jac jac2js` (the EcmaScript codegen) crashes with an opaque `Generating JavaScript failed: list index out of range` on certain components — reproducibly on the shadcn `sidebar` component. There is no node/location in the error, so it is hard to diagnose. Every other shadcn component (button, card, dialog, alert-dialog, dropdown-menu, tabs, tooltip, hover-card, etc.) transpiles fine.

## Reproduction

Using jac-super's shadcn registry:
```
jac add --shadcn sidebar
jac jac2js components/ui/sidebar.cl.jac
```

## Actual

```
✖ Error: Generating JavaScript failed: list index out of range
```

`jac check components/ui/sidebar.cl.jac` passes — only codegen fails. The crash also aborts a full `jac build` for any app importing the component.

## Expected

Codegen should not raise a bare `IndexError`. At minimum it should surface a located diagnostic identifying the offending construct; ideally it should generate correct JS for whatever pattern `sidebar` uses (it leans on `createContext`/`useContext`, `cva`, and a deep component tree).

## Pointers

EcmaScript codegen in `jac/jaclang/compiler/passes/ecmascript/`. The sidebar source is the standard shadcn sidebar (24 exports). Happy to attach the exact file if useful.

PR #9355, opened 2026-09-20
feat(access): endpoint exposure follows the declaration; plain is private

marsninja:
## **Description**

Endpoint exposure now follows the declaration, decided once on the node and read everywhere.

### What changes

| Declaration | Before | Now |
|---|---|---|
| `walker Foo` (plain) | served, JWT required | private: not served, not client-importable, not bridgeable |
| `def foo` imported by the entry module | served, JWT required | private |
| element of a module under a `[placement.pins]` `"server"` pin | callable from client code with auth | private unless tagged; the pin is placement only |
| `walker:priv` / `def:priv` | served, JWT required | private |
| `walker:protect` / `def:protect` | served, JWT required | unchanged, and now bridgeable from client code and other apps |
| `walker:pub` / `def:pub` | served, no auth | unchanged, always served |

### How

- **IR.** `AstAccessNode` gains an `exposure` getter (the endpoint axis: unmarked is private, unlike `access_type`, whose unmarked default is public for source visibility). `ContextAwareNode` gains `placement_name`; `Ability` and `Archetype` gain `surface_kind` and `served`. Every stage reads these getters: the placement solver (`:pub` and `:protect` are endpoint evidence in server-kind apps), client-import classification, the type checker's boundary checks, cross-app E5106, and `jac fix access`.
- **One manifest record.** `SurfaceEntry` in `codeinfo.jac` is the only serialized per-declaration record (name, kind, exposure, placement, module, src). `BoundaryAnalysisPass` writes it; the JIR cache, the serve manifest, the sealed-image module set, `boundaries.json` and the server read it. `func_access`, `walker_access`, `cross_bindings`, and the codeinfo dict helpers are deleted.
- **Server.** `ModuleIntrospector` registers functions and walkers from the surface only. The `getmembers` scan and the cross-binding re-registration are removed (the latter could register a function from a provider module that had already failed to import). The prepare step seeds its compile set from the surface, which fixes a module reached only through a client import having no prepared artifact.
- **Removed inference.** The `_`-prefix rule, the entry-import rule, the callers-decide rule for uncalled `def:pub`, and the module-pin trust-boundary shape are gone. There is no configuration escape hatch.

### Migration

`jac fix access` (with `--dry-run`) rewrites every plain `walker` reachable from a serving app and every plain `def` declared in, or imported by, a serving app's entry module to `:protect`, preserving the previous authenticated behaviour. It leaves `:priv` declarations alone and lists any that used to be served. `examples/jaclang_org` and `examples/day_planner` are migrated in this PR, as are the test fixtures that served plain or `:priv` declarations.

### Also in this PR

- Impl/decl parameter-name mismatch is promoted from W2013 to error E2013.
- ES codegen emits `computed` properties for expression dict keys (bare-name keys stay literal, per #6458).
- The ES view codegen's loop finisher is typed with the loop-statement union, clearing latent E1030/E1053 in the compiler's own tree.
- Docs: access-modifiers, apps, placement, walker-responses, jac-client, the scale HTTP reference, the day-planner tutorials, the sv-auth / endpoints / fullstack / walker / microservices / apps / codespaces skills, diagnostics, breaking-changes, and a release note fragment.

### Tests

New or updated: `test_cross_app_import` (private walker E5106), `test_placement_evidence` (protect bridges, priv private), `test_application_preparation` and scale `test_app_precompile` (surface-only manifests), `test_sealed_bindings` (surface-driven registration), scale `test_microservice` (private walker 404), `test_serve_client` (plain is unserved, declaration decides for uncalled `def:pub`, dual-role and fail-load providers), new `test_fix_access` (rewrites plus the `:priv` report), `test_decl_impl_match_pass` and `test_cli` (E2013), `test_js_generation` (computed keys, separated fixture).


MusabMahmoodh:
Post-merge read from the deploy side. `:pub` still bridges across apps (`classify_client_import_item` returns BRIDGED for anything not PRIVATE), so the k8s e2e fleet, where `orders_app` calls `cart_app.view_cart`, is fine. Three places where an upgraded app goes to 404 with nothing saying why:

1. `jac fix access` skips `include`d modules. `fix.impl.jac:792` keeps only imports with a `from_loc`, and `include utils;` has none. Its plain defs and walkers used to be served, because include puts them in the entry namespace. After the migration they are still plain, so they 404, and the migration reports nothing to migrate.
2. A serve manifest sealed before this change has no `surface` key. `_load_surface` (`server.impl.jac:490-493`) then builds an empty surface and the app registers zero endpoints without a word. JIR and manifest formats did not bump, so `load_image` still accepts that image; a mutable base-image tag is enough to hit it. Failing closed like the `manifest is None` branch just above would turn it into a refusal that says "reseal".
3. Nothing at check, start or deploy lists the plain declarations a serving app just stopped serving. `SurfaceEntry` records plain and `:priv` the same way, so no later stage can tell them apart either. One warning at prepare naming them and pointing at `jac fix access --dry-run` would cover every app that upgrades without reading the release note.

I can run a pre-change `.jab` against the new runtime on my kind rig if you want 2 confirmed on a real pod.

PR #9362, opened 2026-09-20
perf(native): outline container growth into noinline cold helpers

chess10kp:
## **Description**

Native-backend list `append` carried the entire grow path (capacity overflow check, `malloc`, `memcpy`, `free`) inline inside `__list_append_<elem>`. After LLVM inlined append into a user's hot loop, ~10 values stayed live across the inlined allocator calls, the register allocator spilled per iteration, and the loop reloaded `len`/`cap`/`data` from the stack every iteration just for the loop compare. Hardware-counter profiling (100M-iteration list churn) measured 29 instructions and 6.0 cycles per iteration with 3.0 cycles/iter of `RESOURCE_STALLS.SB`, versus 14 instructions / 2.0 cycles for equivalent C. Compiling Jac's own IR with clang at `-O2` reproduced the slow binary, confirming the cost was already in the IR shape handed to LLVM, not a backend codegen gap.

This PR emits growth once per element type into its own function:

```llvm
private noinline cold nounwind @__list_grow_<elem>(list, len)
```

- `container_helpers.impl.jac`: new `_emit_slow_path_helper` (centralized `private` + `noinline` + `cold` + `nounwind` construction for outlined slow paths) and `_emit_list_grow_fn`, which wraps the existing `_emit_grow_array`.
- `lists.impl.jac`: append's grow block becomes a single call to `@__list_grow_<elem>`; append/set are `nounwind`.
- `na_ir_gen_pass.jac`: declarations for the new emitters.

`noinline` is load-bearing: without it LLVM re-inlines the helper and byte-reproduces the baseline. `nounwind` keeps exception edges (and any landing pads) out of the hot loop. The same slow-path policy is applied to dict/set rehash, dict/set order compaction, and bytearray growth, so no container growth path can bloat a caller's hot loop.

Result: the churn loop drops from 29 to 26 instructions and 6.0 to 5.0 cycles per iteration (~17% fewer cycles) with identical workload output and `fib(35)` output. The remaining gap is the heap-resident list header, which needs escape analysis plus a data-only release path; that design is documented in `docs/architecture/native-list-append-performance.md` as a follow-up rather than smuggled into this change.

Also adds the profiling harness used for the diagnosis (`scripts/native_profile.py`, `scripts/miniperf.c`): dumps Jac's native IR, builds with `clang -O2`, and reads grouped `perf_event_open` counters without requiring the `perf` tool.

## **Testing**

- [x] `jac test tests/compiler/backends/native`: 213 passed; the only failure (`test_native_clib_abi` rpath) reproduces identically on clean `upstream/main` (verified by reverting this PR's files) and is pre-existing. `test_native_gc` RSS tests require `/usr/bin/time`, which is absent on this machine.
- [x] IR inspection: appended loops contain only the capacity check and store; growth is a single `cold noinline nounwind` call.
- [x] Churn-loop binary output and `fib(35)` are bit-identical to baseline.
- [x] `jac precommit` (scoped to changed files, `--verify`): formatting clean.

Release note: `release_notes/unreleased/jaclang/9362.feature.md`


chess10kp:
## Bench numbers (hardware counters)

100M-iteration churn workloads, baseline = merge-base of `upstream/main`, `clang -O2`, pinned P-core. Harness: `scripts/native_profile.py` + `scripts/miniperf.c` (grouped `perf_event_open`, no `perf` binary required).

### List append (the motivating case)

| Metric | Baseline | This PR | Equivalent C |
| --- | ---: | ---: | ---: |
| Instructions / iter | 29 | 26 | 14 |
| Cycles / iter | 6.0 | 5.0 (~17% fewer) | 2.0 |
| `RESOURCE_STALLS.SB` / iter | 3.0 | reduced | ~0 |

- Diagnosis: ~0 cache/TLB/branch-miss signal on both sides; the cost was spills and reloads after LLVM inlined the grow path (`malloc`/`memcpy`/`free`) into the hot loop. `llvm-mca`: ~39 uops/loop baseline.
- Ablation: dropping `noinline` byte-reproduces the baseline binary — the attribute is load-bearing.
- Correctness: churn output and `fib(35)` bit-identical to baseline.

### Expanded scope: dict / set / bytearray (measured — no delta)

| Workload (100M iters, bounded key space) | inst/it base→PR | cyc/it base→PR |
| --- | --- | ---: |
| dict `[int,int]` update (1024 keys) | 30 → 30 | 3.24 → 3.34 |
| set `add` (1024 keys) | 27 → 27 | 2.94 → 3.01 |
| bytearray indexed write | 8 → 8 | 0.85 → 0.84 |

Identical dynamic instruction counts with cycles within run-to-run noise (~±3%, single runs): the steady-state hot loops did not change. That is the expected outcome, and IR inspection explains it — baseline already emitted `__{dict,set}_rehash_*` as separate `private` functions that clang `-O2` kept out of the hot loop; `noinline cold nounwind` pins that property explicitly (robustness against future LLVM inlining changes) rather than buying cycles today. Bytearray growth was inline in `__bytearray_reserve`, but the grow block never executes in the measured steady state and `reserve` is too small to create the register-pressure shape that cost list append 17%. Notably, no ICACHE/code-shape regression from the outlined helpers was observed either.

**Scope caveat:** dict get/iteration (the order-compaction holes-check path) and the outlined operator helpers (`__jac_str_repeat`, `__list_repeat/concat/compare/contains/slice`, stepped string slicing, `__rc_push_root_grow`) have no dedicated workloads yet; their validation remains the native test suite plus IR inspection.

Known remaining headroom, intentionally not in this PR: a stack-constructed list header experiment measured ~2.7 cycles/iter (~55% fewer cycles, SB stalls to ~0), but needs escape analysis plus a data-only release path before it can land — tracked in #9376.

PR #9363, opened 2026-09-21
fix(cli): scope precommit to staged files by default

chess10kp:
## Summary
- make `jac precommit` operate on staged `.jac` files by default
- add `--all` for explicit whole-project runs
- preserve explicit paths, update the installed hook to use the new default, and document the scope

## Verification
- `JAC_TEST_JOBS=0 jac test tests/cli/test_precommit.jac`
- `jac fmt --check jaclang/cli/commands/analysis.jac jaclang/cli/commands/impl/analysis.impl.jac tests/cli/test_precommit.jac`
- `jac precommit` (no staged `.jac` files)
- `jac precommit --verify` with explicit changed Jac paths

The full repository precommit verification was attempted but exceeded the 360-second command timeout while scanning the repository.


chess10kp:
@marsninja could you please merge this one when you get a chance?

Heads-up in case the red `build-kit` check gives pause: it's failing repo-wide on `main` since #9312 (`58cf61a44`), not from this PR. That change's new `Iterable` element-type inference produces two false positives in the compiler's own source when the native kernel build type-checks it:

- `parser.impl.jac:4183` / `:4202` — `Cannot assign list[CodeBlockStmt] to parameter 'iterable' of type Iterable[Token]` (both are `list[UniNode].extend(<list[CodeBlockStmt]>)`; the `iterable` param gets `T = Token` from the list-literal initializer instead of the declared receiver element type)

Those two errors fail `require_product(ANALYSIS, parser.jac)`, so `NaIRGenPass._ensure_dep_inference` aborts the `libjac_compiler.so` kernel build. Bisected: `b15a7af7b` (#9307) and `18267fbfa` (#9311) pass ANALYSIS clean; `58cf61a44` fails with exactly these two diagnostics. All the `:0x`-hour checker PRs merged this afternoon had their CI runs cancelled, so this landed on main unnoticed.

This PR itself only scopes `jac precommit` to staged files; everything that actually ran on it is green (bootstrap, jac-check, pre-commit.ci).

PR #9364, opened 2026-09-21
fix(config): record a ${...} form jac does not read instead of shipping it (#9184)

akindu-k:
Closes gap 1 of #9184, the one that issue asks for first and on its own, because it is the only one of the four that lets a literal reach a live Secret with a clean exit code.

Follows #9350, which built the deploy check this gap walks around.

## The short version, with one example

You keep your API key out of `jac.toml` by pointing at an environment variable:

```toml
[scale.secrets]
API_KEY = "${MY_KEY}"
```

You forget to export `MY_KEY`. jac catches it and stops the deploy:

```
scale.secrets.API_KEY: MY_KEY is not set
```

That is the behaviour we want, and it already works.

Now write the same intention in a spelling borrowed from shell:

```toml
[scale.secrets]
API_KEY = "${MY_KEY:+on}"
```

jac does not support `:+`. But instead of saying so, it did not notice the line at all. The deploy finished, reported success, and created the Secret with the key set to the literal text `${MY_KEY:+on}`.

Your pods start fine. Nothing looks wrong until the app makes its first API call and gets rejected for a bad key. The error points at the API provider, so that is where you go looking. The real cause was a line in `jac.toml` that jac skipped over without a word, several steps earlier.

After this PR, jac stops the deploy and tells you which line it cannot read:

```
scale.secrets.API_KEY: ${MY_KEY:+on} is not a form jac reads (use ${VAR}, ${VAR:-default} or ${VAR:?message})
```

It still does not support `:+`. It just stops pretending it does.

## The problem, in plain terms

#9350 taught `jac scale deploy` to stop when a setting it is about to ship still reads `${VAR}` and the variable is not set. That check works off a list of references the config loader writes down as it reads `jac.toml`.

The catch is that the loader only recognises three forms: `${VAR}`, `${VAR:-default}` and `${VAR:?message}`. Anything else, such as `${VAR:+alt}` or `${VAR:0:3}`, does not match its pattern at all. It is not rejected, it is simply not seen. Nothing is written down, and a check driven by that list has nothing to report.

So this passed the guard and shipped:

```toml
[scale.secrets]
SAFE   = "${MISSING_PLAIN}"
SNEAKY = "${MISSING_ALT:+on}"
```

```
deploy check raised:
  scale.secrets.SAFE: MISSING_PLAIN is not set
```

Both values go into the same Kubernetes Secret. Only one is named. Had `SNEAKY` been the only setting of its kind, the deploy would have exited clean and written `${MISSING_ALT:+on}` into the Secret as the literal text, which is exactly the #8773 symptom the guard rail was built to stop, now happening inside the guard rail.

A check that names some unresolved values but not others is worse than the uniform gap before it, because the clean exit now reads as a positive answer.

## How this fixes it

The issue offers two shapes: implement the missing operators, or record anything the parser does not understand. This takes the second, which the issue notes is smaller and covers more than the two forms it names.

To be precise about how much more, since an earlier version of this description overclaimed it: a `${...}` whose body **starts with a variable name** and is not one of the three forms jac reads is recorded. That covers `${VAR:+alt}` and `${VAR:0:3}`, and also the no-colon compose forms `${VAR-default}` and `${VAR?msg}`. A body that does not start with a variable name is left verbatim on purpose, which is what keeps shell syntax such as `${1}`, `${#arr[@]}` and `${}` out of it.

The scan now matches any `${...}` and then looks at what is inside. If the body is one of the three forms jac reads, nothing changes. If it is not, and the name part is a real variable name, it is recorded as unsupported and the deploy refuses it by the same path as any other unresolved value:

```
  scale.secrets.SNEAKY: ${MISSING_ALT:+on} is not a form jac reads (use ${VAR}, ${VAR:-default} or ${VAR:?message})
```

Two details worth review attention:

**An unsupported form is recorded even when the variable is set.** `${DEPLOY_REGION:+on}` with `DEPLOY_REGION` exported is still recorded, because jac cannot honour `:+` either way, so the literal `${DEPLOY_REGION:+on}` would ship regardless. Stopping only when the variable happens to be unset would leave the same literal reaching the Secret.

**A body must start with a real variable name to be reported.** The name is matched as an identifier, and when the body is not one of the three forms, its *leading* name is what gets checked. So `${VAR-default}` is recorded (its leading name is `VAR`), while `${1}`, `${1:+x}`, `${#arr[@]}` and `${}` stay verbatim, since none of them starts with a name. This is why the classification looks at the name rather than failing on any unmatched body, which would have broken ordinary shell syntax.

**Shell syntax inside `[scripts]` is left to the shell.** The pod-boot check from #9199 calls the same parser, so it picks these forms up too, which the issue names as having the same blind spot. But that check walks the whole shipped `jac.toml`, and `sanitized_jac_toml` keeps `[scripts]`. Without care, a perfectly good shell `${CI:+--ci}` in a script would now fail every deploy.

A script is a shell command line and the shell is what expands it, so `unresolved_toml_references` in `bundle.jac` skips an unsupported form under `[scripts]`. The exemption is deliberately narrow:

| Where | Value | Result |
|---|---|---|
| `[scripts]` | `${CI:+--ci}` | skipped, the shell expands it |
| `[plugins.demo]` | `${DEMO:+on}` | reported, the pods are handed this value |
| `[scripts]` | `${UNSET_ONE}` | reported, exactly as before this PR |

That last row is jac's own interpolation and predates this change, so it is untouched.

Measured before and after, with neither variable set:

```
${A:+alt}       before: '${A:+alt}'       recorded: []
${A:+alt}       after:  '${A:+alt}'       recorded: [A, unsupported]
${A-default}    before: '${A-default}'    recorded: []
${A-default}    after:  '${A-default}'    recorded: [A, unsupported]
${A?msg}        after:  '${A?msg}'        recorded: [A, unsupported]
${A/a/b}        after:  '${A/a/b}'        recorded: [A, unsupported]
${1:+x}         after:  '${1:+x}'         recorded: []          (unchanged)
${#arr[@]}      after:  '${#arr[@]}'      recorded: []          (unchanged)
${A}            after:  raises                                  (unchanged)
${A:-d}         after:  'd'                                     (unchanged)
```

## Scope

Gap 1 only. Gaps 2 and 3 (`$${...}` escape, nested braces) touch the same brace matching and are one change; gap 4 (interpolated values are always strings) is independent. All three stay open on #9184.

One consequence worth noting: `interpolate_env_vars` called directly with no sink now raises on an unsupported form where it used to return the text verbatim. That matches what it already does for an unset `${VAR}`, and the only in-tree caller passes a sink.

One known leftover, open for review: the deploy's header and hint still use the generic unset-variable wording ("this machine does not have", "Export each one"). For an unsupported form with the variable exported, the offending line now names the correct fix, but it sits between two lines of slightly wrong advice. They are shared with the genuinely-unset path, so I have left them alone unless we want them conditional here.

## Tests

Both existing table-driven tests were extended in place, no new test cases added as separate tests, so the counts are unchanged.

| File | Before | After |
|---|---|---|
| `jac/tests/project/test_config.jac` | 131 | 131 |
| `jac/jaclang/scale/tests/deploy/test_pod_env.jac` | 42 | 42 |

Added to the existing tables:

- **Parser** (`test_config.jac`): `${VAR:+alt}`, `${VAR:0:3}`, and the no-colon forms `${VAR-default}` and `${VAR?msg}` raise with the form named, with the variable set; `echo ${1:+on}` and `echo ${#arr[@]}` still pass through.
- **Config load** (`test_config.jac`): both `:+` and `:0:3` are recorded with `unsupported` set. The existing rows gain a trailing `False` because the test compares whole tuples and now asserts the new field.
- **Deploy check** (`test_pod_env.jac`): an unsupported form stops the deploy both with the variable unset and with it exported.
- **Pod-boot check** (`test_pod_env.jac`): the path's first unsupported-form cases, one either side of the `[scripts]` boundary. `${CI:+--ci}` in `[scripts]` passes; `${DEMO_ALT:+on}` in `[plugins.demo]` fails.

Full suites on this branch: `jac/tests/project/` 333 passed. `jac/jaclang/scale/tests/deploy/` 395 passed, 5 skipped, 0 failed. `jac/jaclang/scale/tests/microservices/` 347 passed with 3 pre-existing `test_gateway.jac` Prometheus failures that reproduce identically on a clean tree. CI is green: 25 passed, 5 skipped, 0 failed.


jac/jaclang/project/impl/config.impl.jac
```diff
@@ -10,40 +10,60 @@ impl interpolate_env_vars(
...
-
-        if operator == ":-" {
-            return operand;
-        }
+        body = `match.group(1);
+        parts = re.fullmatch(supported, body);
+        var_name = body.split(":")[0] if parts is None else parts.group(1);
         if not var_name.isidentifier() {
             return `match.group(0);
         }
```
MusabMahmoodh: `${MY_KEY-default}`, `${MY_KEY?msg}` and `${MY_KEY/a/b}` still slip through here: `supported` fullmatches them with `[^}:]+` eating the whole body, so `var_name` is `MY_KEY-default`, fails `isidentifier`, and comes back verbatim with nothing recorded. Classifying on the leading name closes those too:
```suggestion
        lead = re.match(r"[A-Za-z_][A-Za-z0-9_]*", body);
        if lead is None {
            return `match.group(0);
        }
        var_name = lead.group(0);
        if parts is not None and parts.group(1) != var_name {
            parts = None;
        }
```

jac/jaclang/project/impl/config.impl.jac
```diff
@@ -10,40 +10,60 @@ impl interpolate_env_vars(
...
+        "${VAR:-default} and ${VAR:?message}";
+    }
     return f"Environment variable {self.var} is required: {self.message}"
         if self.required
         else f"Environment variable {self.var} is not set";
 }
 
+impl UnresolvedRef.reason -> str {
+    if self.unsupported {
+        return f"{self.form} is not a form jac reads";
```
MusabMahmoodh: This lands under a header that says the machine lacks the variable and a hint to export it, which is the wrong advice when the variable is exported and the form is the problem. Naming the fix here keeps both callers honest:
```suggestion
        return f"{self.form} is not a form jac reads (use ${{VAR}}, ${{VAR:-default}} or ${{VAR:?message}})";
```

MusabMahmoodh (changes requested):
Doing this in the parser rather than at the deploy check is the right place, and all three consumers of `resolve_env_refs` pick it up for free. Two things before it merges. The name split on `:` means `${VAR-default}` and `${VAR?msg}` (the no-colon compose forms) still come back verbatim with nothing recorded, so the "any future unsupported form" claim does not hold yet. And the pod-boot check walks the whole shipped jac.toml, `[scripts]` included (`sanitized_jac_toml` strips only `[scale.secrets]` and `[dev]`), so a shell `${CI:+--ci}` in a script that worked before now fails every deploy. A `${CI:+--ci}` script row in the pod-side table in `test_pod_env.jac`, next to the `sh -c 'echo ${1}'` one, would pin whichever behaviour you pick and give the pod path its first unsupported-form case.

akindu-k:
Both valid, both fixed in bf73b97aa, and `upstream/main` is merged in. The second one was a regression I introduced, so thanks for catching it before this went anywhere near a cluster.

## The no-colon forms

You are right, and the claim in the description was wrong. The name pattern was `[^}:]+`, so `${VAR-default}` parsed as a perfectly good bare reference whose name happened to be the whole body, `VAR-default`. That is not an identifier, so it fell straight through the shell-positional guard and nothing was recorded:

```
${VAR-default}  ->  '${VAR-default}'  recorded=[]
${VAR?msg}      ->  '${VAR?msg}'      recorded=[]
```

The name is now matched as an identifier, `[A-Za-z_][A-Za-z0-9_]*`, and when the body is not a form jac reads the leading identifier is taken rather than splitting on `:`. Both are recorded now.

I have also reworded the description, because what actually holds is narrower than what I wrote. A `${...}` whose body **starts with a variable name** and is not one of the three forms is recorded. A body that does not start with one is left verbatim on purpose, and that is exactly what keeps `${1}`, `${#arr[@]}` and `${}` out of it.

## The `[scripts]` path

Confirmed, and it would have failed every deploy for anyone with shell syntax in a script:

```
[scripts] build = "npm run build ${CI:+--ci}"
  ->  recorded=[('scripts.build', 'CI', True)]
```

`sanitized_jac_toml` strips only secret paths and `[dev]`, so `[scripts]` ships and reaches the pod check, as you said.

A script is a shell command line and the shell is what expands it, so `unresolved_toml_references` now skips an unsupported form under `[scripts]`. An unset bare `${VAR}` there is still reported exactly as before, since that is jac's own interpolation and predates this PR:

```
scripts :+    ->  []
setting :+    ->  ['plugins.demo.key: Unsupported form ${DEMO:+on}: ...']
scripts bare  ->  ['scripts.broken: Environment variable UNSET_ONE is not set']
```

## On the test row

One wrinkle worth flagging, since it changes the answer: the `sh -c 'echo ${1}'` row you pointed at is under `[plugins.demo]`, not `[scripts]`. A `${CI:+--ci}` placed literally beside it lands on the failing side of the boundary rather than the passing one, because a `[plugins.*]` value is handed to the pods while a script is not.

So I added two rows to the pod-side table instead, one either side of that line:

- `[scripts] build = "npm run build ${CI:+--ci}"` -> no failure
- `[plugins.demo] alt = "${DEMO_ALT:+on}"` -> fails, naming the setting

That pins the boundary in both directions and gives the pod path its first unsupported-form case. If you would rather the exemption keyed on something other than the section name, say so and I will move it.

## State

Test counts unchanged at 131 and 42, both tables extended rather than added to. Deploy suite 395 passed, 5 skipped, 0 failed. Project suite 333 passed. The 3 reds in `test_gateway.jac` are the Prometheus ones and reproduce identically on a clean tree.

CI is still finishing as I write this.

jac/jaclang/project/impl/config.impl.jac
akindu-k: Right on all three, and `${MY_KEY/a/b}` is the one that shows the split on `:` was never going to be enough.

I took the same classification by a slightly different route: rather than reconciling `parts` against the leading name afterwards, `supported` itself now requires the name to be an identifier, so a body it cannot fullmatch is already the unsupported case and `parts` needs no correction.

```jac
supported = r"([A-Za-z_][A-Za-z0-9_]*)(?:(:[-?])([^}]*))?";
...
leading = re.match(r"[A-Za-z_][A-Za-z0-9_]*", body);
if leading is None { return `match.group(0); }
var_name = leading.group(0);
```

Same outcomes as your suggestion on everything I could think to throw at it, including two you did not name:

```
${MY_KEY-default}  -> recorded      ${MY_KEY}       -> recorded (bare)
${MY_KEY?msg}      -> recorded      ${MY_KEY:-ok}   -> 'ok'
${MY_KEY/a/b}      -> recorded      ${1}            -> untouched
${MY_KEY^^}        -> recorded      ${#arr[@]}      -> untouched
${MY_KEY%.txt}     -> recorded      ${}             -> untouched
```

Happy to switch to your exact shape if you would rather keep `supported` permissive and reconcile after, since the two-regex version does mean the identifier pattern is written twice. Say the word and I will change it.

jac/jaclang/project/impl/config.impl.jac
akindu-k: Applied in 87d021653. You are right that this was actively misleading rather than just terse, and it is worst in exactly the case the PR exists to catch:

Before, with `SET_VAR` exported:

```
jac.toml reads environment variables this machine does not have, so the deploy would ship these settings as literal ${...} text:
  scale.secrets.SNEAKY: ${SET_VAR:+on} is not a form jac reads
```

The variable is exported, so both the header and the hint send the reader to do something they have already done.

After:

```
  scale.secrets.SNEAKY: ${SET_VAR:+on} is not a form jac reads (use ${VAR}, ${VAR:-default} or ${VAR:?message})
```

One residual, since you should decide it rather than me: the header and the hint are still the generic unset-variable wording, so this case reads as correct advice sandwiched between two lines of wrong advice. I left them alone because they are shared with the genuinely-unset path and changing them touches the existing expectations. If you want the header and hint to go conditional when every finding is an unsupported form, I will do that in this PR.

jac/jaclang/scale/injector/bundle.jac
```diff
@@ -452,7 +452,16 @@ def sanitized_jac_toml(base: Path) -> (str | None) {
...
+    return [
+        f"{ref.path}: {ref.describe()}"
+        for ref in refs
+        if not _is_shell_script_ref(ref)
+    ];
+}
+
+
+def _is_shell_script_ref(ref: any) -> bool {
+    return ref.unsupported and ref.path.split(".")[0] == "scripts";
```
MusabMahmoodh, line 464: Scripts can also come from a profile: `apply_profile` merges `[environments.<name>.scripts]` (config.impl.jac:264), and that table ships to the pods too, so `${CI:+--ci}` there has the path `environments.ci.scripts.build` and still fails every deploy. Covering both shapes:
```suggestion
    parts = ref.path.split(".");
    in_scripts = parts[0] == "scripts"
        or (parts[0] == "environments" and parts[2:3] == ["scripts"]);
    return ref.unsupported and in_scripts;
```
An `[environments.ci.scripts]` row next to the `[scripts]` one in test_pod_env.jac would pin it.

MusabMahmoodh (approved):
The no-colon forms are recorded now, `[scripts]` is left to the shell, and the pod-side table has both the script row and a real unsupported-form row, so I'm releasing my block. Your two-regex version is fine, no need to switch. The generic header and hint can stay; the line itself names the fix now, and that line is the one people act on. One gap in the scripts exemption inline. The earlier test-runtime red on this head was the self-hosted runner losing contact, and the rerun on the same sha passed.

PR #9372, opened 2026-09-21
feat(db): boot the session store lazily and scope scratch reaping

chess10kp:
## Summary
- `Session` construction no longer starts the embedded Postgres machinery: the store boots once, lazily, on the first access that needs database truth — so `jac db status`-style probes, `repr()`, and commit/close of never-used sessions stay spawn-free
- Store injection moved to `_store`: `Memory.store` is now a getter-only accessor, so `Session(store=...)` can no longer silently discard an injected store (the scale callsite now passes `_store=`)
- Restored the missing `attempt += 1` in `run_request`'s serialization-conflict retry loop
- Scratch allocation now reaps via an escaped, prefix-scoped catalog scan (`jac_scratch_*` only) instead of paying per-database checks across every accumulated project database; orphaned project databases remain the start sweep's contract
- `jac db status` now surfaces dead-scratch and orphaned-project counts so accumulation is visible

#### Test plan
- [x] `jac check` clean on all touched files
- [x] `test_lazy_pg_boot` — 18 passed, including new tests for bounded conflict retry, injected-store honoring, and no-boot lifecycle probes
- [x] `test_pg_uow` + `test_outbox` + `test_read_path_isolation` — 30 passed
- [x] `test_db_lifecycle` — ran to completion; new coverage for scratch-prefix scoping, orphan/dont-touch decoys, and `db status` visibility
- [x] `jac db status` smoke-tested against a live data dir


chess10kp:
Benchmark: `jac test` timings, before (`eb01ac166`) vs this PR, warm runs on the same machine (old code run from a worktree sharing the same LLVM slice and typeshed stubs).

**20 trivial tests, no DB access** (`assert` arithmetic/string ops only):

| | test time | wall |
|---|---|---|
| before | 8.64s, 8.63s | ~9.6s |
| after | 7.34s, 7.42s, 7.44s | ~8.1s |

**`test_object_model.jac`** (8 tests, does touch the store):

| | test time | wall |
|---|---|---|
| before | 8.65s, 7.49s | ~8-9.6s |
| after | 8.28s, 7.37s, 7.58s | ~8-9.1s |

Takeaways:
- ~1.2s saved per `jac test` invocation when the session never needs the DB — consistent across all warm runs.
- DB-touching suites overlap completely, as expected: lazy boot defers store setup but doesn't eliminate it once a test actually persists.
- The bigger win is architectural: status probes, `repr()`, and never-used sessions no longer spawn the embedded engine at all.

PR #9374, opened 2026-09-21
refactor: k8s dedup + annex-split simplification train (Stack 3+, on #9373)

chess10kp:
## Stack 3+ — k8s deploy dedup, then the annex-split train (rebased)

**Supersedes #9231** — the stack-3 k8s dedup, rebased onto current main, with the rest of the simplification train accumulated on top.

**Stacks on** #9227 / #9373 (GitHub base must be `main`).

### Stack-3 core (unchanged in intent)

The kubernetes deployers repeated one API-call shape ~60 times:

```jac
try {
    core_v1.read_namespaced_X(name=..., namespace=ns);
    core_v1.replace_namespaced_X(name=..., namespace=ns, body=body);
} except ApiException as e {
    if e.status == 404 { core_v1.create_namespaced_X(namespace=ns, body=body); }
    else { raise; }
}
```

- Shared helper **`apply_k8s_resource(api, kind, name, body, namespace?, replace?)`** in `kubernetes_utils` — kubectl-apply semantics via the read/replace/create method-stem convention; `replace=False` covers create-only resources.
- **43 read-or-create blocks** → one-liners; **19 inline delete-if-exists blocks** route through `delete_if_exists` (extended for cluster-scoped resources).
- Follow-up scale cleanups: consolidate k8s manifest/apply/delete boilerplate, drop unused imports surfaced by `jac prune`.

### Accumulated on top (stacks 4+)

- **Annex splits** of god impl files across compiler, server, scale, cli, client, testing, lsp, byllm, project (lexer, parser, type_checker, unitree roles, session, ServingServer, KubernetesTarget, ManifestBuilder, MonitoringDeployer, microservice_gateway, …).
- **Cross-module dedup**: shared helpers unified, dead definitions pruned, unused imports swept (~60 modules).
- **DSLs**: schema-spec for `plugin_config`, endpoint-registration for `JEndPoint` boilerplate.
- `refactor(scale): table-drive gateway builtin routes`, `refactor(compiler): generic AST readers`, PEG rule-helper combinators.

Part of the massive simplification stack (target: −30% LOC).


jac/jaclang/scale/deploy/target/kubernetes/monitoring.impl/prometheus.impl.jac
```diff
@@ -0,0 +1,330 @@
...
+                "    tls_config:\n"
+                "      insecure_skip_verify: true\n"
+                "    authorization:\n"
+                "      credentials_file: /var/run/secrets/kubernetes.io/serviceaccount/token\n"
+                "    kubernetes_sd_configs:\n"
+                "      - role: node\n"
+                "    relabel_configs:\n"
+                "      - action: labelmap\n"
+                "        regex: __meta_kubernetes_node_label_(.+)\n"
+                "        target_label: __address__\n"
```
MusabMahmoodh, line 61: The leading `- ` got dropped here, so `target_label`/`replacement` now belong to the labelmap rule: the `__address__` rewrite to the API server is gone (cadvisor gets scraped on the node address at an apiserver proxy path), and labelmap renames every node label to `kubernetes.default.svc:443`. Container metrics break on every deploy with monitoring on. On main this is its own rule:
```suggestion
                "      - target_label: __address__\n"
```

jac/jaclang/scale/deploy/target/kubernetes/utils/kubernetes_utils.lifecycle.impl.jac
```diff
@@ -0,0 +1,143 @@
+import re;
+import from jaclang.scale._optdeps.kubernetes { client }
+
+def _kind_stem(kind: str) -> str {
+    # "ClusterRoleBinding" -> "cluster_role_binding", the method-name stem
+    # the python k8s client uses for every verb.
+    return re.sub(r'(?<!^)(?=[A-Z])', '_', kind).lower();
```
MusabMahmoodh, line 7: This fixes the casing, but only `Deployment` is pinned, through the MagicMock tests in test_http_activation_microservices, and those accept any stem for the other kinds. A no-mock table in `tests/deploy/test_k8s_utils.jac` that asserts `hasattr(api, f"{verb}_{stem}")` for every kind the deployers pass (ConfigMap, ClusterRoleBinding, DaemonSet, IngressClass, PodDisruptionBudget, ...) against the real client classes would cover it.

jac/jaclang/scale/deploy/target/kubernetes/utils/kubernetes_utils.apply.impl.jac
```diff
@@ -0,0 +1,119 @@
...
+    namespace: str,
+    rules: list[dict[str, any]],
+    sa_name: str = "",
+    sa_fields: (dict[str, any] | None) = None,
+    replace: bool = True
+) -> None {
+    sa_name = sa_name or f"{name}-sa";
+    labels = app_labels(name);
+    apply_service_account(core_v1, sa_name, namespace, labels, sa_fields);
+    rbac_v1 = client.RbacAuthorizationV1Api();
```
MusabMahmoodh, line 89: This builds its own `RbacAuthorizationV1Api`, so the `rbac_v1` that `IngressDeployer._deploy_nginx_controller` receives is now unused, and so is its `cluster_resource_name` (ingress.jac:30, :34). Can it take the client as a parameter like `core_v1`, so the caller keeps control of which client is used?

MusabMahmoodh (changes requested):
The PascalCase stem bug from #9231 is fixed: `_kind_stem` normalizes the kind, and `_apply_or_replace` became `patch_k8s_resource`, so there is one apply helper now. I diffed the rest of the dedup against the old blocks: resource names, selectors, the create-only vs replace split (ksm Deployment, both DaemonSets, SAs, Services, IngressClass) and every 404/403/422 tolerance all match. The one regression is the Prometheus relabel line inline, and nothing in the suite renders that config.

I can't sign off on the scale half of a 336-file PR that also carries compiler, lsp, byllm and client splits. It is 141 commits behind main with 10 scale conflicts, two of them modify/delete (`keda_autoscaler.impl.jac` and `microservice_gateway.impl.jac`), where #9474/#9497/#9502 and #8779 would need hand-porting into the new annexes. Could the scale part come out as its own PR on fresh main, with the dedup separate from the file splits? That PR would also want a release fragment named after its own number rather than 9227/9229/9231.

PR #9375, opened 2026-09-21
fix(persistence): share application database and durable model identities across services

### LINKED ISSUE #9365: Since 0.37.12, two apps in one workspace get two different classes for the same `node`, so neither can read the other's rows
A `node` declared in a module that two apps both import is compiled once per app, and since 0.37.12 each copy keeps its app's namespace in `__module__`: `_jac_app_` + sha256 of the entry file's absolute path. The two copies are then different classes with different persisted names, so anything one app writes, the other cannot read back as its own type. On 0.37.11 both apps got `shared.models` and it worked.

Four files, all of them below.

```toml
[project]
name = "twoapp2"
default-app = "writer"

[apps.writer]
kind = "service"
entry-point = "writer"
route = "/writer"

[apps.reader]
kind = "service"
entry-point = "reader"
route = "/reader"
```

```jac
# shared/models.jac
node Thing { has v: str = ""; }
```

```jac
# writer.jac
import from shared.models { Thing }

walker put {
    has v: str = "";
    can go with Root entry {
        t = Thing(v=self.v);
        here ++> t;
        report {"wrote": self.v, "module": type(t).__module__};
    }
}
```

```jac
# reader.jac
import from shared.models { Thing }

walker get {
    can go with Root entry {
        all = [here -->];
        mine = [n for n in all if isinstance(n, Thing)];
        report {
            "out_of_root": len(all),
            "readable_things": len(mine),
            "my_class_module": Thing.__module__,
            "what_came_back": [f"{type(n).__module__}.{type(n).__name__}" for n in all]
        };
    }
}
```

Serve one app, write, stop it, serve the other against the same store, read:

```
$ jac run --port 8791 writer.jac
$ TOK=$(curl -sX POST localhost:8791/user/register -H 'Content-Type: application/json' \
    -d '{"identities":[{"type":"username","value":"probe"}],
         "credential":{"type":"password","password":"probe12345"}}' \
    | jq -r .data.token)
$ curl -sX POST localhost:8791/walker/put -H "Authorization: Bearer $TOK" -d '{"v":"hello"}'
"reports": [{"wrote": "hello", "module": "_jac_app_b7b65b2376b40bdc.shared.models"}]

$ jac run --port 8793 reader.jac
$ curl -sX POST localhost:8793/walker/get -H "Authorization: Bearer $TOK" -d '{}'
"reports": [{"out_of_root": 1, "readable_things": 0,
             "my_class_module": "_jac_app_55fa2b561469ecc2.shared.models",
             "what_came_back": ["_jac_app_b7b65b2376b40bdc.shared.models.Thing"]}]
```

Same four files, same store, only the binary differs:

| | 0.37.11 | 0.37.12 | 0.37.21 |
| --- | --- | --- | --- |
| module the writer persists | `shared.models` | `_jac_app_0fa8652e70ec6258.shared.models` | `_jac_app_b7b65b2376b40bdc.shared.models` |
| the reader's own class | `shared.models` | `_jac_app_34969c73d070e685.shared.models` | `_jac_app_55fa2b561469ecc2.shared.models` |
| things the reader can read | 1 | 0 | 0 |

The node is on the graph and it materializes, but as the writer's class, so `isinstance(n, Thing)` is false in the reader. Both prefixes are sha256 of the entry path: `sha256("<abs>/writer.jac")[:16] == b7b65b2376b40bdc`, `sha256("<abs>/reader.jac")[:16] == 55fa2b561469ecc2`, which matches `application_namespace` in `jaclang/runtime/prepared.jac`. That function and `qualified_module`, which prefixes every module of an app with it, came in with 8f452cc4 (#9088, "compile app entries through a unified pipeline"), and `git tag --contains` puts it first in v0.37.12.

In a deployment it is worse, because the other app's namespace is not importable in the pod. `Serializer._get_class` refuses it ("Refused to deserialize unregistered class: ...; it is not registered and could not be resolved from its own module"), `Session._materialize` logs `could not materialize anchor <id> (<Type>); skipping`, and then quarantines the row, so it never loads again. We hit this on jacBuilder: 173 `UserProfile` rows quarantined in one CI run, and every request after the one that created the profile answered "User profile not found", because the walker that creates the profile runs in one app and the walkers that read it run in another.

I know `apps.md` says ownership follows compilation context and that nodes do not cross an app boundary, and that E5108 covers an app importing another app's node. This is not that case: neither app imports the other, both import ordinary shared source, so nothing is diagnosed at build time and the first sign is quarantined data at runtime. A build-time diagnostic would have caught it before the data was quarantined; a way to give a shared archetype module one owner would be better still. `[apps.<name>]` takes only kind, entry-point, platform and route (`jaclang/project/apps.jac`), so there is no key for that today, and pointing the consumer at the provider's entry does not help: a re-export is not a bridge, the symbol still resolves to the shared module and the boundary type keeps an empty app tag.

We are unblocked for now by registering the sibling apps' spellings ourselves with `Serializer.register_alias`, computing their namespaces from `JAC_SV_FLEET`. It works, but it means every multi-app project has to know how the prefix is built.

Not tested on Linux; the repro is macOS arm64 and the jacBuilder runs are a CI kind cluster, both on 0.37.21.

marsninja:
## Problem and behavior

Services of one application currently persist path-derived Python module names as graph type identities. The same imported model can consequently become incompatible across services or after deployment relocation. Provider contexts can also select a different database from their entry path.

This change makes persistence application-wide. Types use the application namespace, logical module and declaration name automatically; no sharing domains or ownership configuration are required. Services retain separate runtime classes and globals, and deserialize shared rows through their own compiled manifests. SQL filters, ancestry, fingerprints and explicit rename aliases use the same durable identity.

Provider sessions inherit the already-bound application store, preserving its database and user-root identity while owning their transaction and identity map. Embedded database selection is independent of service entry names and project subdirectories. The store's session factory is also reused by the outbox worker.

This is a clean break: rebuild artifacts, explicitly migrate legacy path-derived type names, and consolidate any previously split service databases. Startup does not silently migrate or merge existing data. Architecture and application documentation describe the model and migration requirements.

## Validation

- Regression coverage exercises shared database/root identity through provider contexts, service-local classes and globals, typed graph queries, schema evolution, relocation across processes, unrelated application namespaces, missing schemas, and explicit legacy aliases.
- Rebuilt and smoke-tested the native compiler; fixed AST metadata materialization and contextual container inference encountered by the native build.
- Added repeated-entry execution and console-stream regressions.
- Broad runtime/CLI run: 1,049 passed; the three failures were resolved by updating two obsolete identity expectations and populating the cold compiler cache. Follow-up suites passed all 3 rename/data-preservation tests and all 11 snippet tests.
- Service/context suite: 105 passed, 9 skipped; concurrency regressions: 5 passed. Artifact, sealed-loader and persistent-identity contracts: 23 passed.
- Full repository type check: 1,041 passed. Isolated database lifecycle suite: 24 passed. Formatting and release-note checks pass.
- CI validation remains in progress; this PR stays draft until complete.

Fixes #9365.


MalithaPrabhashana (changes requested):
The source-mode fix works: I reproduced #9365 on main and it goes green on this branch. Sealed images regress, though. An app can no longer `isinstance` the nodes it persisted itself, and that is how this reaches a deployment, so I'd hold the merge until that one is sorted.

I ran both sides in dev mode against the checkouts (this head 2e194e2af, main 7fad8bf85 as the baseline), macOS arm64, a separate HOME per checkout, and a fresh Postgres cluster per case. The app is the four files from #9365, with `walker:pub` so main serves them.

### What is fixed

The reader gets its own class back now, and with a shared database and one project directory per service (the deployment shape) it can see the row at all:

| | main | this PR |
| --- | --- | --- |
| two apps, one directory: nodes the reader reads as its own type | 0 | 1 |
| two apps, separate directories, one `JAC_DB_URL`: nodes out of root / readable | 0 / 0 | 1 / 1 |

Both sealed images stamp the same durable identity, `jac:["twoapp2","shared.models"]`, while their Python modules stay per-service, and the row is stored with it:

```
$ jac db sql "SELECT arch_type, arch_module FROM anchors"
jac:["twoapp2","shared.models"].Thing | jac:["twoapp2","shared.models"]
```

Relocating the project keeps that identity. Legacy 0.37.21 rows fail loudly instead of quarantining:

```
Legacy path-derived persistent type identity; declare an explicit schema alias and migrate the store before serving it
```

and both migration routes work on that data: `@archetype_alias("_jac_app_958fef7b96981f89.shared.models.Thing")` reads it back, and so does the runtime `Serializer.register_alias(...)` that jacBuilder uses today.

### What blocks: a sealed app cannot isinstance its own nodes

Same project, sealed with `jac build --all -a sealed`, one image, one database. The writer app writes a `Thing`, then an app in that image reads root's children back:

| sealed image, shared database | main | this PR |
| --- | --- | --- |
| the same app writes, then reads its own node | isinstance 1 | isinstance 0 |
| a second app reads it | row not visible (0 out of root) | row visible, isinstance 0 |

On this PR the read reports:

```
{"out_of_root": 1, "readable_things": 0, "typed_filter": 1,
 "my_class_module": "_jac_app_74e1c4c462cd6b71.shared.models",
 "what_came_back": ["twoapp2-sealed.shared.models.Thing"]}
```

The app's own class is `_jac_app_74e1c4c462cd6b71.shared.models.Thing`, and what comes back is a different class object with the same durable identity, from the image's package namespace. `[-->[?:Thing]]` still returns 1, because the SQL filter matches on the durable identity, so this only shows up on `isinstance`, `type(x) is`, and anything typed that is handed the node. We use `isinstance` on persisted nodes in jacBuilder, and we ship sealed images, so this would be live for us.

Every one of those runs logged `Loading sealed application artifacts` and compiled nothing, so they took the sealed path rather than recompiling the retained sources. Inside the image `application_for(os.getcwd())` is `None`, so deserialization resolves unscoped; that is a guess at the cause, not a diagnosis.

A note on method, since it matters here: `JAC_NO_DEV_SOURCE=1` cannot run these images locally (`JIR format` error, the launcher's 0.37.21 runtime against PR-compiled JIRs), so both sides ran with dev mode active, identically. main passing the same test is what makes me call this a regression rather than my setup.

I think CI stays green because the sealed tests boot an app and call an endpoint (`walker:pub echo` reporting `"hello"`); none of them persists a node and reads it back as the app's class. A sealed case that writes a node and asserts `isinstance` on the way back would have caught this.

### Also worth a look

Renaming the project changes the durable identity, which the doc calls a schema rename, but it also quietly starts a second database next to the old one in the same directory:

```
jac_twoapp2_665bae49          project  live
jac_twoapp2renamed_665bae49   project  live
```

The app then comes up empty with no diagnostic. Is a warning wanted when a database for the previous namespace exists in the same directory?

### What I did not cover

Linux, a real Kubernetes deploy, and jacBuilder's own tree against this compiler. `tests/runtimelib`: 744 passed and 6 skipped on this head, 736 passed and 6 skipped on main, the difference being this PR's own new tests. The trial merge with main is clean, and `test_persistent_identity`, `test_application_request_context` and `test_persistence_config` pass on it (17).

MalithaPrabhashana:
Two more runs, because my first one used `jac run --serve <app name>` and that is not how a pod starts an app.

With the command jac-scale bakes into the pod spec, `cd /app; exec jac run --serve "$JAC_SV_FILE"`, and with a `.jab` bundle, which is what our deploy ships. Same project, same app writes a node and then reads root's children back, one database:

| same app writes, then reads its own node | main | this PR |
| --- | --- | --- |
| sealed image, app name | isinstance 1 | isinstance 0 |
| sealed image, pod command (`jac run --serve writer.jac`) | | isinstance 0 |
| `.jab` bundle | isinstance 1 | isinstance 0 |

Through the `.jab` the class comes back as `twoapp2.shared.models.Thing` instead of the app's `_jac_app_c7d20e84fd00e458.shared.models.Thing`, so it is the image package namespace either way.

jac/jaclang/server/impl/session.impl.jac
```diff
@@ -897,8 +921,10 @@ impl Session._serialize(anchor: Anchor, props: (any | None) = None) -> AnchorRow
     arch_module = '';
     fingerprint = '';
     if anchor.is_populated() and arch is not None {
-        arch_type = type(arch).__name__;
-        arch_module = type(arch).__module__;
+        import from jaclang.data.type_identity { storage_type }
+        arch_type = storage_type(type(arch));
```
MusabMahmoodh, line 925: Every row written before this has a bare `arch_type` (`Thing`), and typed hops now filter `arch_type = ANY([<module>.Thing, ...])` through `storage_type`, so on upgrade `[root-->[?:UserProfile]]` returns empty instead of raising, and adding `@archetype_alias` fixes untyped reads but not this filter. On a live app that reads as "profile not found" and the create path writes a second profile next to the legacy row. Could startup refuse to serve (or at least warn) when `anchors` still holds rows whose `arch_module` is not `jac:`-prefixed for a type that now has a durable identity?

MusabMahmoodh (commented):
The scale/server side reads clean to me: pods already share one `JAC_DB_URL` (`/jac` from kubernetes_postgres), so k8s database selection doesn't move, the HTTP sv-to-sv path is untouched, and provider sessions via `store.new_session()` reuse the bound conninfo and pool instead of rediscovering a database. Colocated `_local_call`/`_local_spawn` committing in their own session is a good change: `jac start` now behaves like pod-per-service instead of quietly giving colocated calls cross-service atomicity. One upgrade question for deployed fleets, inline.

PR #9393, opened 2026-09-21
Port time to a bundled na_stdlib module, replacing the native intercept

chess10kp:
## Summary

- Replaces the Mechanism-A compiler intercept (`NATIVE_TIME_FUNCS` + `_emit_time_call`, 93 LOC of hand-rolled IR emission) with a Mechanism-F bundled surface `na_stdlib/time.jac` over a per-OS libc FFI floor (`_time_native.jac` / `_time_native.darwin.jac`). Bundled-module resolution picks it up by location; the dead intercept emitters, dispatch, and the `NATIVE_SUPPORTED_STDLIB` entry are removed.
- Surface grows past intercept coverage: `process_time`, `thread_time`, `clock_gettime_ns`, `clock_settime_ns`, and the `CLOCK_*` constants join `time`/`monotonic`/`perf_counter`/`*_ns`/`sleep`.
- `sleep` parks on an absolute `clock_nanosleep(CLOCK_MONOTONIC, TIMER_ABSTIME)` deadline with `EINTR` retry (relative `nanosleep` remainder loop on Darwin), matching CPython's `pysleep`; error behavior matches CPython for negative, NaN, and out-of-range inputs.
- Clock failures route through the new `_errno_native.raise_errno`, which extracts the errno-to-exception-subclass mapping previously inline in `raise_os_error` (also removing a dead duplicated `EISDIR` check), so `EPERM` surfaces as `PermissionError` like CPython.
- `datetime.jac`'s `import time` transparently rebinds to the bundled module.
- Thread-safe by construction: every floor call packs its `timespec` into a 16-byte `bytes` buffer it allocates for itself (`ts: bytes` at the FFI boundary passes the data pointer) — the native backend emits `pthread_create` with `JAC_THREADS` defaulting to 4, so a shared malloc'd scratch would race a sleeping thread's deadline against another thread's `clock_gettime`. This matches the caller-owns-buffers convention of the zlib/zstd/hashlib floors and the per-call alloca the deleted intercept used.
- Documented SCOPE divergences in `na_stdlib/README.md`: calendar/`struct_time` family fails loudly; float-seconds `clock_gettime`/`clock_settime`/`clock_getres` are absent because their bare names collide with the floor's C externs in the shared native symbol table.

## Benchmarks — 1M iterations, best-of-3 (Linux x86-64)

| Lane | `monotonic_ns()` ns/iter | `monotonic()` (float) ns/iter |
|---|---|---|
| CPython 3.14 (`python3 -c` oracle, same 1M-iter loop) | ~206–216 | ~269–280 |
| Jac sv bridge (`jac run -b python`, CPython `time` via interop) | ~269–271 | ~358–365 |
| **Jac native bundled module** (`jac run -b native`, `na_stdlib/time.jac`) | **~68** | **~76** |

- The bundled module is **~3x faster than CPython** and **~4x faster than the Jac sv bridge**. The per-call 16-byte buffer + byte-indexed parse costs ~40 ns over the racy shared-scratch version (~30/~41 ns); that is the price of thread safety and still leaves a wide margin over CPython. Raw-FFI breakdown: `clock_gettime` alone ~28–30 ns; +fresh `b"\x00"*16` alloc ~54–59 ns; +LE parse ~72–80 ns.
- The sv bridge is ~25–30% slower than bare CPython — the interop marshal overhead is the entire gap this port eliminates.

## Review fixes (71aafe1fc)

- Required: the shared `malloc`'d `_TS` scratch was a data race (a thread parked in `clock_nanosleep` could observe its absolute deadline overwritten by another thread's clock read); replaced with per-call buffers as suggested.
- `EINTR` hoisted to `_errno_native` (was duplicated in both floors).
- `prim_time.jac` pins the `sleep(inf)` `OverflowError` by type instead of CPython's build-dependent message; the 100k-iteration busy loop (which asserted nothing beyond the trivially-true `p1 >= p0`) is gone.
- `bench_time_direct.jac` deleted (raw-FFI measurement; its numbers live here in the table above).

## Test plan

- [x] `jac test jac/jaclang/compiler/tests/test_prim_equivalence.jac -f time` — 4/4 pass on the per-call-buffer floors (re-run after each floor change); fixture extended with `time.sleep(0)` int-arg, `sleep(-1.0)`/`float("nan")`/`float("inf")` error pins, `clock_gettime_ns(bad_id)` `OSError` pin, `process_time*`/`thread_time*`/`CLOCK_*` coverage, and a `from time import perf_counter_ns` lane (`from:time` pin added to fixtures `jac.toml`)
- [x] `jac fmt --lintfix` clean on all touched files (the git-hook gate)
- [ ] Darwin floor unverified locally (Linux-only machine; symmetric with Linux floor, needs a macOS CI/dev run)


chess10kp:
## Benchmarks — `time` port (Linux x86-64, 1M iterations, best-of-3)

| Lane | `monotonic_ns()` ns/iter | `monotonic()` (float) ns/iter |
|---|---|---|
| CPython 3.14 (`python3 scripts/bench_time.py`) | ~206–216 | ~269–280 |
| Jac sv bridge (`jac run -b python`, CPython `time` via interop) | ~269–271 | ~358–365 |
| **Jac native bundled module** (`jac run -b native`, `na_stdlib/time.jac`) | **~29.6** | **~40.7** |
| Jac native raw FFI (`clock_gettime` extern, reused `timespec`) | ~28–30 | — |

Raw-FFI breakdown for reference: `clock_gettime` call alone ~28–30 ns; +fresh `b"\x00"*16` alloc ~54–59 ns; +LE parse ~72–80 ns.

### Takeaways

- The bundled module is **~7x faster than CPython** on `monotonic_ns` (~29.6 vs ~210 ns) and **~9x faster than the Jac sv bridge** (~29.6 vs ~270 ns). The `def:pub` surface adds only ~1.5 ns over the raw `clock_gettime` FFI call — the Jac→libc path is essentially free once the call stays inside native code.
- The float `monotonic()` adds ~11 ns over the `_ns` variant (f64 conversion of sec + nsec).
- The sv bridge is ~25–30% slower than bare CPython — the interop marshal overhead is the entire gap the bundled module eliminates.
- No ownership/borrow-checking work was needed to reach FFI-floor speed: the hot path passes one `i32` clock id and a scratch `timespec*` — nothing crosses the managed-heap boundary per call.

chess10kp:
### Benchmark: bundled `na_stdlib/time` (native backend)

Run on a remote GitHub Actions runner (`ubuntu-latest`, Linux / X64, 4 vcpu) via [`bench-time-bundled` workflow](https://github.com/chess10kp/jaseci/actions/runs/35839587998), building `scripts/bench_time_bundled.jac` with `jac build --native` (25 KB binary).

| measurement | ns/iter |
|---|---|
| bare loop | 0.148 |
| `time.monotonic_ns()` (bundled na_stdlib) | 38.617 |
| `time.monotonic()` float (bundled na_stdlib) | 40.877 |

The bundled `monotonic_ns` call costs ~38 ns/iter over a ~0.15 ns bare loop on the runner — i.e. the FFI floor + Python-congruent wrapper adds only tens of nanoseconds per call, consistent with a direct `clock_gettime` syscall.

Workflow definition lives on the fork (`bench-time-bundled.yml`, dispatchable on `na-time-bench`); artifact `bench-time-bundled` is attached to the run.

chess10kp:
### Benchmark: bundled `na_stdlib/time` — all endpoints, three lanes

Remote GitHub Actions run: [bench-time-all](https://github.com/chess10kp/jaseci/actions/runs/35855626469) on `ubuntu-latest` (Linux / X64, 4 vcpu), CPython 3.12.3 oracle.

Methodology per PROMPT.md: warmup + best-of-3, identical loop bodies on every lane, `acc` checksum printed to defeat DCE. Values are ns/iter *including* loop overhead — `bare_loop` gives each lane's baseline.

| endpoint | `jac -b native` (bundled) | `jac -b python` (bridge) | `python3` (oracle) |
|---|---:|---:|---:|
| bare_loop | 1.88 | 197.68 | 67.18 |
| `time` | 51.94 | 372.15 | 191.10 |
| `time_ns` | 49.77 | 280.29 | 133.37 |
| `monotonic` | 51.28 | 369.52 | 193.08 |
| `monotonic_ns` | 49.91 | 277.51 | 126.42 |
| `perf_counter` | 53.15 | 372.08 | 191.26 |
| `perf_counter_ns` | 49.60 | 278.83 | 126.38 |
| `process_time` | 662.35 | 1101.31 | 897.48 |
| `process_time_ns` | 669.67 | 1001.01 | 832.46 |
| `thread_time` | 630.43 | 1072.55 | 858.72 |
| `thread_time_ns` | 646.41 | 939.99 | 790.93 |
| `clock_gettime_ns` | 52.51 | 292.64 | 162.78 |
| `clock_settime_ns` (err path) | 1192.95 | 1443.79 | 1025.69 |
| `sleep(0)` | 116349 | 117038 | 116652 |
| `sleep(100µs)` | 165376 | 165684 | 166173 |

Takeaways:

- The bundled FFI path beats the sv bridge on every endpoint — e.g. `monotonic_ns` ~50 vs ~278 ns/iter (~5.6×), `clock_gettime_ns` ~53 vs ~293 (~5.6×). It also beats the CPython oracle at the loop level (~50 vs ~126), since the native Jac loop compiles to machine code while CPython interprets its loop.
- `process_time`/`thread_time` (~630–670 ns bundled) are dominated by the `CLOCK_PROCESS/THREAD_CPUTIME_ID` syscall cost itself; the bundled floor is still ~25% below the bridge lane.
- `clock_settime_ns` permission-error path (raise → catch) is within ~16% of CPython's exception overhead.
- `sleep` is dominated by kernel sleep granularity (~116 µs floor for `sleep(0)` on this runner); bundled is within noise of CPython on both sleep benches.

Scripts: `scripts/bench_time_all.jac` (single source, runs under both `-b native` and `-b python`) and `scripts/bench_time_all.py` (identical CPython oracle twin) — committed on this branch. Dispatch-only workflow `bench-time-all.yml` lives on the fork; artifacts attached to the run.

chess10kp:
## Perf: bundled FFI vs the old intercept vs raw FFI

Measured `jac run -b native`, best-of-3, ns/iter. Remote lanes ran the sealed binary on `ubuntu-24.04` pinned to one core; the intercept lane needed `ulimit -s unlimited` to finish at all (see below).

| endpoint | main (native intercept) | this PR (bundled FFI) | raw `import from c` FFI |
|---|---|---|---|
| `time_ns` / `monotonic_ns` / `perf_counter_ns` | ~38.7 | **~28.0** | 27.4 |
| `time()` / `monotonic()` (float) | ~44.7 | **~38.0** | — |
| `sleep(0.0)` | ~115µs | **~61µs** | — |

- **vs raw FFI:** the bundled module is at parity — 28.0 vs 27.4 ns/call is noise. The `monotonic_ns → time_clock_ns → _clock_gettime` chain plus the errno branch inlines to effectively nothing.
- **vs the old intercept:** ~25% faster on the `_ns` getters, ~15% on the float getters, ~2x on `sleep(0)`.
- **Bonus correctness fix:** the intercept emitted its timespec `alloca` at the call site, so every loop iteration burned 16B of stack — N=524288 calls = 8MiB = SIGSEGV (reproduced on the CI runner, exit 139; filed as #9487). `__alloca` hoists to the entry block, so the bundled path runs 2M+ iterations clean.
- Also adds coverage the intercept never lowered: `process_time*`, `thread_time*`, `clock_gettime_ns`, `clock_settime_ns` (~1.3µs/iter — real syscalls, no vDSO for those clocks).

Bench harness lives on `bench/na-time` in my fork (workflow + cached binary) if anyone wants to re-run or extend it.

Issue #9376, opened 2026-09-21
Native backend V2: stack-constructed list headers for function-local lists (~55% remaining append headroom)


chess10kp:
## Summary

Follow-up to #9362. That PR outlines list growth into a `noinline cold` helper (`__list_grow`), cutting list-append cost from 6.0 to 5.0 cycles/iter (~17%), measured with hardware counters (100M-iter churn, `clang -O2`, pinned core). The remaining cost has a single dominant signature: `RESOURCE_STALLS.SB` (~3.0 cycles/iter) from every `append` dirtying **two** heap objects — the list header (refcount, type, len, cap, data pointer) and the backing buffer.

## Proposed design (V2)

For lists that provably stay function-local, construct the list header as a stack `alloca` in the frame and keep only the data buffer on the heap:

```
// today: hdr = malloc(header); hdr.cap=4; buf = __list_grow(...)  → 2 heap objects
%hdr  = alloca {i64, i64, ptr}     ; header in the frame
      = call __list_grow(...)      ; only the buffer is heap
```

Prototype measurement (same methodology as #9362):

| Metric | V1 (shipped in #9362) | V2 prototype |
| --- | ---: | ---: |
| cycles / iter | 5.0 | ~2.7 (~55% fewer) |
| `RESOURCE_STALLS.SB` / iter | ~3.0 | ~0 |

C-equivalent ceiling for the same loop is ~2.0 cyc/it, so V2 closes most of the remaining gap.

## What it needs before it can land

1. **Escape analysis.** A proof that the list does not outlive the frame — no return, no store into another object/global, no closure capture, no aliasing escape. Plus a materialization path that promotes a stack header to a heap header at any discovered escape point (or a conservative heap fallback when analysis is inconclusive).
2. **Data-only release path.** The refcount field is meaningless frame memory; the generic `decref`/destroy path must never run on a stack header (it would free frame memory). Final release must free *only* the data buffer, while element destruction and ownership semantics are preserved.

## Status

Deliberately kept out of #9362 — it is not safe as a local code-generation tweak. Documented in `docs/architecture/native-list-append-performance.md` on the #9362 branch.

chess10kp:
Investigated — premise disproven by controlled measurement. Closing as not planned.

[...]

PR #9399, opened 2026-09-22
Native backend V2: stack-constructed list headers for function-local lists (#9376) [opt-in experiment]

chess10kp:
**Stacked on #9362** (its `native-list-grow-noinline` head lives on a fork, so GitHub cannot use it as a PR base; until #9362 merges, this diff includes those changes).

## What this implements

For #9376: when a new escape stamp in `RcFactsPass` (`na_list_stack_ok`) proves a list display bound to a function-local never outlives its frame, the native backend constructs the list header in an entry `alloca` slab and keeps only the data buffer on the heap. Scope-exit, loop-iteration, overwrite, and last-use releases route through a new data-only helper (`__rc_release_list_data_<elem>`), so the frame-resident header is never freed. The stamp refuses every aliasing/escape channel: return, cross-scope, user-call arguments (except `len`), iterator materialization, augmented re-binding, boolean/composite expressions. Any unproven case keeps the base heap `__list_new` path.

## Why it is gated off by default

Two findings from validation (details in `docs/architecture/native-list-append-performance.md`):

1. **The prototype's performance claim did not reproduce.** Interleaved same-core A/B against the #9362 baseline (variance < 0.5%) shows cycle-identical pure-append churn (11.60 cyc/iter both, identical instruction counts) and a ~27% *regression* on read-mixed churn, consistent with stack/heap 4K aliasing. The win that remains is allocation reduction (one fewer header malloc/free per qualifying list), not cycles.
2. **A layout-sensitive crash in the data-only release routing.** `dict_hash_ops::order_after_churn` segfaulted deterministically; four-step bisection isolated it to exactly the overwrite-site data-only routing, with the slab layout itself verified sound. Root cause is suspected in the name-keyed registry interacting with binding-state mirroring (`_enter_bindings`/`_shadow_bindings`) rather than any single emission — and the fail-leak sentinel does not protect the *data buffer*, whose premature free is a use-after-free vector.

Given both, stamping is gated behind `JAC_NATIVE_STACK_LIST_HEADERS=1`. With the gate unset the compiled output is equivalent to the #9362 baseline (verified: zero promotions in fixture IR; the dict suite matches baseline behavior). The unit and native tests enable the gate explicitly and pass (4/4).

## Also fixed on the way

The slab initializer originally applied the RC/dtor/alloc-id offsets from the slab array base instead of the List-struct pointer, writing the sentinel into an adjacent frame slot (fail-UAF, not the designed fail-leak). Offsets are now `HDR_TOTAL`-relative and the type tag bases off the struct pointer.

## Testing

- `jac test tests/compiler/passes/test_list_stack_facts.jac` (stamp matrix: 5 stamp cases, 8 refusals) — passes with gate enabled
- `jac test tests/compiler/backends/native/test_list_stack_headers.jac` (execution + IR inspection) — passes with gate enabled
- `jac test tests/compiler/backends/native/test_native_dict_hash.jac` — default path at baseline parity (9 passed; the 4 `--memory` profile failures are pre-existing/environmental on the base branch too: missing vendored musl)
- `jac precommit --staged` — passes
- New benchmark harness: `workloads/list_churn.jac`, `workloads/list_append.jac`, `workloads/list_churn_ref.c` + `scripts/native_profile.py --suppress`

## Release note

`release_notes/unreleased/jaclang/9376.feature.md`

## Recommendation

Do not enable the gate by default. If the allocation-reduction benefit is worth pursuing, the release routing needs a redesign first (attach the decision to the symbol's storage slot at definition instead of a mutable name-keyed registry).


chess10kp:
Wrong base repository (gh resolved the fork parent); recreating against jaseci-labs/jaseci per repo guidelines.

PR #9401, opened 2026-09-22
perf(native): stream zlib/gzip decompression through a caller-owned inflate arena

chess10kp:
## Description

`zlib.decompress` called the one-shot `uncompress` and, on `Z_BUF_ERROR`, allocated a fresh `b"\x00"*cap` and restarted the entire decompression — ~2.6x slower than CPython on high-expansion inputs. This drives `inflate` in a single pass over a caller-owned `malloc`/`realloc` arena instead: `next_out`/`avail_out` are poked into a `b"\x00"*112` `z_stream` image through the `__mem_store_i32/i64` intrinsics, growth doubles the arena in place, and the produced bytes are copied once into the exact-size `bytes` result. Payload addresses are recovered as `int` via `memchr(buf, buf[0], 1)`.

`gzip.decompress` now raw-inflates each member (`windowBits = -15`) and locates the member boundary from `total_in`, removing the `b"\x78\x9c" + data[hpos:]` re-wrap copy and the adler-coincidence `Z_DATA_ERROR`-is-ok hack from the `uncompress2` approach. ISIZE still seeds the initial capacity, so the common single-member case needs zero growth and zero trailing slice.

Two latent bugs fixed along the way:

- `glob Z_BUF_ERROR` existed in both `zlib.jac` and `gzip.jac` — `E5026` symbol collision for any program importing both modules (the equivalence fixtures import one module each, so it slipped through).
- `zlib.decompress(b"")` returned `b""`; CPython raises `error -5`. The streaming path surfaces `Z_BUF_ERROR` and raises `ValueError` naturally.

## Benchmarks

ns/op, best-of-3, digests byte-identical across all lanes (310 KB text @ ~321x expansion; 64 KB random):

| op | jac AOT | jac JIT | jac `-b python` bridge | CPython 3.14 | AOT ratio |
|---|---|---|---|---|---|
| crc32 300k | 133µs | 191µs | 101µs | 191µs | **0.70x** |
| adler32 300k | 176µs | 165µs | 176µs | 165µs | 1.06x |
| zlib.compress | 1.31ms | 1.32ms | 1.14ms | 1.31ms | 1.00x |
| zlib.decompress | **317µs** | 333µs | 676µs | 299µs | **1.06x** (was 2.63x) |
| compress rand64k | 2.33ms | 2.09ms | 2.43ms | 2.07ms | 1.13x |
| gzip.compress | 1.16ms | 1.54ms | 1.10ms | 1.34ms | **0.87x** |
| gzip.decompress | **233µs** | 336µs | 589µs | 331µs | **0.71x** (was 1.55x) |

Harness: `scripts/bench_zlib_gzip.jac` (same source runs under `jac run -b native`, `-b python`, and `jac build --native`) + `scripts/bench_zlib_gzip.py` oracle. `--memory rc` vs `managed` was within noise.

## Testing

- `jac test jac/jaclang/compiler/tests/test_prim_equivalence.jac -f "zlib bundled"` — pass
- `jac test jac/jaclang/compiler/tests/test_prim_equivalence.jac -f "gzip bundled"` — pass (multi-member, FEXTRA/FNAME/FCOMMENT/FHCRC headers, corrupt body, bad CRC/ISIZE, truncated, trailing garbage)
- `jac test jac/tests/compiler/backends/native/test_native_zlib_bytearray.jac` — pass
- `jac test jac/tests/compiler/backends/native/test_native_file_zlib_aot.jac` — pass
- `jac check` on all touched na_stdlib files — clean (warnings only on `__mem_*` intrinsic names)

Generated with [Devin](https://devin.ai)


chess10kp:
### Perf analysis: why native `zlib.decompress` was 2.6x slower, and what the ownership fix bought

**Diagnosis.** The old path called one-shot `uncompress` and, on `Z_BUF_ERROR`, doubled `cap`, allocated a fresh `b"\x00" * cap`, and **re-ran the entire decompression**. For a 965 B input expanding to 310 KB (~321x) that's ~6 full re-inflates plus ~1 MB of memset/copy/page-fault traffic — the measured 789µs vs CPython's ~300µs was almost entirely restart + realloc churn, not libz.

**Fix (ownership model applied).** Decompression now inflates once into a callee-owned `malloc`/`realloc` arena — caller-owned scratch, no per-retry owned temps — with `next_out`/`avail_out` poked into a `b"\x00"*112` `z_stream` image via `__mem_store_i32/i64`. `realloc` extends the mapping in place (mremap on Linux) instead of alloc+copy+fault. Result bytes are memcpy'd once into the exact-size `bytes` return. `gzip.decompress` uses `inflate(-15)` per member and reads the boundary off `total_in`.

**Numbers** (ns/op, best-of-3, digest-identical across lanes):

| op | before | after (AOT) | CPython 3.14 | ratio |
|---|---|---|---|---|
| zlib.decompress 300k | 789µs | 317µs | 299µs | 2.64x → **1.06x** |
| gzip.decompress 300k | 513µs | 233µs | 331µs | 1.55x → **0.71x** |

Everything else stayed at parity or better (crc32 0.70x, gzip.compress 0.87x, zlib.compress 1.00x). `--memory rc` vs `managed` showed no meaningful delta — the cost was allocation *pattern*, not GC mode. `nogc` enforcement rejects the bench harness (managed stores); the enforced-subset kernels in `examples/ownbench` are the right vehicle for that comparison.

chess10kp:
## Follow-up: deduplicated the inflate driver + full benchmark matrix

Pushed `c9210247a` — a refactor pass over this PR's surface code:

### What changed

- **One shared inflate driver.** `zlib.decompress` and gzip's per-member inflate each carried their own streaming state machine over *different* arenas (gzip: fresh `bytes` + `memcpy` per growth step; zlib: `malloc`/`realloc`). Both now call `z_inflate_all(strm, cap, ceiling) -> ZInflateResult` in `_zlib_native.jac` — one `realloc` arena, one status dispatch, one final exact-size copy. Each surface keeps only init/end and its own error mapping. The `bytes`-variant `z_inflate_set_out` is gone.
- **Deleted `scripts/probe_inflate.jac`** — dev scaffolding containing a third copy of the loop; the bench script covers the durable need.
- **Growth path is now pinned by fixtures.** `prim_zlib` round-trips a 200KB highly-compressible payload (forces `realloc` doublings); `prim_gzip` adds a large leading member so growth fires before the last member's ISIZE hint. Previously the doubling/`BUF_ERROR`-full-buffer paths were only exercised by the bench.
- **Status-code cleanup instead of `try/finally`.** `z_inflate_all` frees its arena before returning and `z_inflate_end` runs unconditionally after — no raise can skip cleanup. Measured ~0.65ms/call faster than `try/finally` in this hot loop anyway (native try frames are expensive).
- **Native-link glob hygiene.** Imported `glob` constants don't lower inside comparisons (silently demotes the whole function to Python) and same-named globs collide at image link — so constants stay module-local with distinct prefixes: `_Z_*` floor, `_ZL_*` zlib, `_GZ_*` gzip.
- README updated; release fragment moved to the `- **Feature: Title**:` convention.

Net: -271/+125 lines, one inflate implementation instead of three.

### Benchmarks (310KB text, ~321x expansion; digests identical across all lanes)

| bench | na JIT | na AOT | sv bridge | CPython 3.14 | JIT/CPython |
|---|---:|---:|---:|---:|---:|
| crc32_300k | 191µs | 133µs | 102µs | 277µs | 0.69x |
| adler32_300k | 165µs | 229µs | 188µs | 165µs | 1.00x |
| zlib_compress | 1.32ms | 1.28ms | 1.37ms | 1.31ms | 1.01x |
| zlib_decompress | 324µs | 317µs | 733µs | 300µs | 1.08x |
| zlib_compress_rand64k | 2.07ms | 2.32ms | 2.63ms | 2.08ms | 1.00x |
| gzip_compress | 1.55ms | 1.19ms | 1.17ms | 1.35ms | 1.15x |
| gzip_decompress | 347µs | 431µs | 670µs | 333µs | 1.04x |

The refactor is perf-neutral (316µs zlib-decompress direct call, identical to the pre-refactor code). Honest note: the earlier "gzip decompress beats CPython at 0.71x" did not reproduce — measured 1.04x JIT this session; the durable claim is parity-to-1.1x.

### Verified

- Direct native calls: 200KB growth roundtrip, empty input (`-5`), truncated stream, multi-member gzip, trailing garbage — all correct, no demotion warnings.
- `jac fmt --check` clean; `jac precommit` passes.
- Full `prim equivalence` suite left to CI (~60 parametrized native compiles).

Benchmark methodology + lessons (try/finally cost, lambda dispatch overhead, glob lowering rules, shared-driver shape for future bz2/zstd floors) recorded in the local porting playbook (`PROMPT.md`, not part of this PR).

chess10kp:
Benchmark rerun on the latest branch (includes the shared-driver zipfile migration + follow-up fixes).

Workload: 310 KB compressible text + 64 KB seeded-random input; ns/call = min of 3 reps (warmup 1) ÷ iters. Digests identical across all lanes, so timings are comparable.

| bench (ns/call) | CPython 3.14.7 | jac `-b python` | na JIT | na AOT |
|---|---:|---:|---:|---:|
| crc32_300k | 1,189,880 | 959,810 | 1,034,655 | 573,610 |
| adler32_300k | 932,703 | 827,222 | 896,430 | 997,887 |
| zlib_compress_300k | 7,778,558 | 6,666,876 | 8,549,846 | 6,693,083 |
| zlib_decompress_300k | 1,621,535 | 1,608,440 | 1,672,373 | 1,839,377 |
| zlib_compress_rand64k | 13,486,153 | 10,487,858 | 13,081,555 | 11,764,609 |
| gzip_compress_300k | 7,998,458 | 7,772,924 | 8,984,005 | 5,811,871 |
| gzip_decompress_300k | 1,908,745 | 1,739,353 | 2,008,255 | 2,164,817 |

na stays within ~±30% of CPython across the board — decompress lands in the same band as CPython's own one-shot path now that inflate runs through a single malloc/realloc arena instead of retry-uncompress. (This workload has a modest expansion ratio; the gap that motivated the PR — uncompress2 retry loops re-inflating the whole input on high-expansion data — is where the old path degraded by orders of magnitude.)

Caveats: single process run per lane, box under load (loadavg ~15) from concurrent CI jobs, so absolute numbers are noisier than the playbook's usual environment; per-key best-of-3 mitigates but doesn't eliminate it.

chess10kp:
Benchmarks on GitHub Actions CI (`ubuntu-24.04`, AMD EPYC 7763 ×4 vCPU) at `8402eaf1e` — run: https://github.com/chess10kp/jaseci/actions/runs/35856223067

`scripts/bench_zlib_gzip.jac`, best of 3 reps, ns per iteration:

| benchmark | python lane | native lane | speedup |
|---|---:|---:|---:|
| crc32_300k | 87,914 | 82,498 | 1.07× |
| adler32_300k | 97,114 | 99,918 | 0.97× |
| zlib_compress_300k | 1,213,411 | 1,435,718 | 0.85× |
| zlib_decompress_300k | 442,819 | 248,101 | **1.79×** |
| zlib_compress_rand64k | 1,540,783 | 1,387,480 | 1.11× |
| gzip_compress_300k | 1,214,057 | 1,525,558 | 0.80× |
| gzip_decompress_300k | 445,855 | 215,873 | **2.07×** |

Decompression — the path this PR streams through the caller-owned inflate arena — runs ~1.8–2× faster on the native backend vs the CPython bridge lane. All 7 digest values agree across lanes (correct output on both backends). Compression rows carry a small regression on native; both lanes call the same zlib `deflate`, so the delta is allocation overhead rather than compression itself.

chess10kp:
Follow-up benchmarks at `bdf3f0e24` (adds a raw libz FFI floor lane and a new zero-copy inflate path) — run: https://github.com/chess10kp/jaseci/actions/runs/35886689093 — `ubuntu-24.04`, Intel Xeon 6973P-C ×4.

ns per iteration, best of 3:

| benchmark | raw libz FFI | FFI + copy→`bytes` | `zlib.decompress`/`gzip.decompress` |
|---|---:|---:|---:|
| zlib_decompress_300k | 149,773 | 160,566 | 286,762¹ |
| gzip_decompress_300k | 123,920 | 133,927 | 227,830¹ |

¹ The same calls measure 206,394 / 154,859 ns in the main bench lane — cross-process variance on shared runners is significant, so per-lane ratios are the signal.

Changes since the last numbers (`a100d540e`): `z_inflate_all` now inflates directly into the result `bytes` (no malloc arena, no final memcpy — zero-copy when the caller's `cap` is exact, which it is for gzip via ISIZE and zipfile via the member size), the regrow loop predicts the output size from the observed expansion ratio instead of doubling, and `gzip.decompress` skips the `result + member` concat for single-member input.

Net: gzip decompress ~10–20% faster than the previous revision; the residual gap vs a bare `inflate` call is the Jac-side zero-fill/slice on the output plus gzip's mandatory CRC-32 verify — work the FFI floor doesn't do.

All digests agree across the python/native lanes.

chess10kp:
Final benchmark update at `0b3fab504` — run: https://github.com/chess10kp/jaseci/actions/runs/35923637988 — `ubuntu-24.04`, AMD EPYC 7763 ×4. Best of 3, ns/iter.

**vs CPython bridge** (same process, same run):

| op | python lane | native lane | speedup |
|---|---:|---:|---:|
| zlib_decompress_300k | 440,878 | 251,640 | **1.75×** |
| gzip_decompress_300k | 441,289 | 196,187 | **2.25×** |

**vs raw libz FFI floor** (single-shot `inflateInit2_`/`inflate`/`inflateEnd` into an exact-size `bytes`, measured in-process alongside the driver):

| | FFI floor | driver | ratio |
|---|---:|---:|---:|
| zlib decompress | ~193.8k | ~251.6k | ~1.3× |
| gzip decompress | ~175.8k | ~196.2k | ~1.1× |

The FFI floor here is honest — it includes allocating and materializing the result `bytes`, not just the bare `inflate` call (~150k without it).

**Optimizations landed since the first numbers** (`a100d540e`, `ad554280f`, `15e87bded`):
- `z_inflate_all` inflates directly into the result `bytes` payload — no malloc arena, no final `memcpy` (zero-copy when `cap` is exact: gzip via ISIZE, zipfile via member size)
- On regrow, the arena resizes to the predicted output size (`total × src_len / fed`, +25% headroom) instead of doubling — typically one resize
- When the prediction overshoots, the result buffer is right-sized in place (len-field write) instead of a `buf[0:total]` slice copy
- `gzip.decompress` skips the `result + member` concat for single-member input
- zstream field pokes inlined in the inflate loop

**Decomposition of the remaining gap**: per 310 KB op on this runner, Jac-side buffer materialization (`rc_alloc` + memset + first-touch page faults) costs ~130k ns vs ~8k for the equivalent `memcpy` — i.e. a large share of the *floor itself* is allocation cost, not inflate work. gzip is essentially at its floor (mandatory CRC-32 verify accounts for ~82k of its remaining delta); zlib is bounded by the unknown output size (predict + trim).

**Measurement caveat**: within-process timings are stable (±0.5%), but identical code measures ±25% across separate `jac run` processes — per-process compiled-binary layout dominates small deltas. Ratios within a lane are the reliable signal.

**Follow-up (native backend, out of scope here)**: an uninitialized `bytes` allocation primitive (e.g. `bytes.uninitialized(n)` — `_bytes_alloc` without the memset pass) or zero-copy `bytearray`→`bytes` adoption would cut ~130k ns off every large-buffer op in na_stdlib — decompress, `read_bytes`, `compress`, everything. That's where the remaining headroom lives.

chess10kp:
Per-op FFI-floor numbers for every bench op, plus a pre-merge baseline re-run to rule out merge drift.

Runs: merged `2f025f2dd` — https://github.com/chess10kp/jaseci/actions/runs/36004235496 (AMD EPYC 9V74) — and pre-merge baseline `dfcf4624c` — https://github.com/chess10kp/jaseci/actions/runs/36002286962 (slower runner class; absolutes ~1.3–2× higher across *all* lanes, so compare within-run ratios only). Probe: `scripts/bench_zlib_ffi_probe.jac`, raw `compress2`/`crc32_z`/`adler32_z`/`inflate` calls measured in-process alongside the drivers.

**Driver vs raw libz FFI floor** (ns/call, best of 3; merged run):

| op | FFI floor | na driver | driver/floor |
|---|---:|---:|---:|
| crc32_300k | 72,223 | 72,058 | **1.00×** |
| adler32_300k | 84,666 | 84,786 | **1.00×** |
| zlib_compress_300k | 1,083,100 reuse / 1,088,286 alloc | 1,089,843 | **~1.00×** |
| zlib_compress_rand64k | 1,148,101 | 1,171,273 | **1.02×** |
| gzip_compress_300k | 1,085,346 (compress2 lvl9) | 1,169,305 | **1.08×** |
| zlib_decompress_300k | 159,469 (exact) | 416,992 probe / 196,954 main lane | 2.6× / 1.24× |
| gzip_decompress_300k | 146,628 (exact) | 266,928 probe / 160,246 main lane | 1.8× / 1.09× |

**Main lanes, merged run** (python bridge vs native):

| op | python lane | native lane | speedup |
|---|---:|---:|---:|
| crc32_300k | 76,567 | 71,864 | 1.07× |
| adler32_300k | 84,829 | 84,707 | 1.00× |
| zlib_compress_300k | 677,558 | 1,145,280 | 0.59× |
| zlib_decompress_300k | 371,564 | 196,954 | **1.89×** |
| zlib_compress_rand64k | 1,306,079 | 1,160,656 | 1.13× |
| gzip_compress_300k | 680,495 | 888,738 | 0.77× |
| gzip_decompress_300k | 369,551 | 160,246 | **2.31×** |

Read-outs:

- **Checksums and `zlib.compress` are at the FFI floor** — the driver is literally one FFI call; bound-alloc + result slice vanish next to ~1.1ms of deflate (reuse-vs-alloc floor lanes differ by ~5µs).
- **`gzip.compress`** carries ~84µs over the deflate floor — essentially the mandatory `z_crc32` pass (~72µs) plus header/trailer concat.
- **Baseline confirms no merge drift**: identical pattern pre-merge (crc32 81,986→82,011; compress2 1,408,069→driver 1,416,660; gzip.compress 1,411,541→1,503,436).
- **Decompress caveat, now measured twice**: in the probe process the driver runs ~2.3–2.6× above `ffi_exact` (baseline 451µs vs 194µs; merged 417µs vs 159µs) while the in-file algorithm replicas sit near the floor — yet in the main bench process the same call measures 250µs/197µs (~1.2–1.3× floor). Pre-existing per-process compiled-layout variance, unchanged by the merge — but the driver's spread is wider than the ±25% previously noted.

